PR #1887, opened 2018-06-16
luci-mod-admin-full: fix connection lookup_queue

Ansuel:
I notice that in the connection tab the lookup array keep checking all the entities everytime and adds to the dns_cache list only the resolved ones. This looks wrong as it's very unluckily that an addr thats not resolved in this instance gets resolved in the next one. This change how the dns_cache list works. We put in it every tried addr even if not resolved. Also this fix some error if for some reason the page doesn't receive the right data.

### DIFF
--- a/modules/luci-base/luasrc/sys.lua
+++ b/modules/luci-base/luasrc/sys.lua
@@ -340,6 +340,32 @@ function net.conntrack(callback)
 					if entry[key] == nil then
 						entry[key] = val
 					end
+					if entry["ptype"] == nil then
+						val = tonumber(val)
+						if key == "sport" then
+							if (val == 443) then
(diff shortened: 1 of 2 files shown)


jow-:
The resolver on the backend has a hard timeout limit which means that entries later in the queue are less likely to get resolved on the first attempt, hence the infinite retries.

We likely should fix that by trying each item three times or so

Ansuel:
Why not reduce the number of entities to check then ? (too much request?)

Ansuel:
@jow- i changed this to check 3 times the same entities then removes it...

Ansuel:
@jow- currently the connection table is broken... i also fixed this when you have time take a look at this too...

PR #1889, opened 2018-06-17
luci-nginx: add nginx support file

Ansuel:
This adds nginx-mod-luci as a dependency of this package.
@hnyman 


hnyman:
I will merge this, but i feel that more optimal solution would be to drop the "+nginx" dependency here, and have that "+nginx" in nginx-mod-luci-ssl (instead of "nginx").

PR #1893, opened 2018-06-18
luci-app-firewall: update Ukrainian translation

YuriPet:
Updated with the latest synchronization of the translation.


hnyman:
I merged the previous PR and now there is a conflict. Please rebase.

YuriPet:
Now everything is right?

hnyman:
You should still rebase it so that there is no merge commit, just the one topic commit.

PR #1895, opened 2018-06-19
Responsive OpenWrt theme

jow-:
This pull requests introduces, among some large markup cleanups, responsive design for the old OpenWrt.org theme.

The goal of this pull request is to modernize the OpenWrt theme to make it a first class citizen again, potentially allowing it to become the default choice in the future, after several years of using the Bootstrap theme.

### DIFF
--- a/applications/luci-app-shadowsocks-libev/luasrc/view/shadowsocks-libev/add_instance.htm
+++ b/applications/luci-app-shadowsocks-libev/luasrc/view/shadowsocks-libev/add_instance.htm
@@ -1,24 +1,17 @@
 <div class="cbi-section-create cbi-tblsection-create">
-	<br />
-	<div class="table cbi-section-table">
-		<div class="tr cbi-section-table-row">
-			<div class="td cbi-section-table-cell" style="width:140px">
-				<select class="cbi-input-select" id="_newinst.type" name="_newinst.type">
-					<option value="_dummy">-- instance type --</option>
-					<option value="ss_local">ss-local</option>
-					<option value="ss_tunnel">ss-tunnel</option>
-					<option value="ss_redir">ss-redir</option>
-					<option value="ss_server">ss-server</option>
-				</select>
-			</div>
-			<div class="td cbi-section-table-cell" style="width:110px">
-				<input type="text" class="cbi-input-text" id="_newinst.name" name="_newinst.name" placeholder="<%:Name%>"/>
-			</div>
-			<div class="td cbi-section-table-cell left">
-				<input type="submit" class="cbi-button cbi-button-add" name="cbi.cts.<%=self.config%>" value="<%:Add%>" />
-			</div>
-		</div>
+	<div>
+		<select class="cbi-input-select" id="_newinst.type" name="_newinst.type">
+			<option value="_dummy">-- instance type --</option>
+			<option value="ss_local">ss-local</option>
+			<option value="ss_tunnel">ss-tunnel</option>
+			<option value="ss_redir">ss-redir</option>
+			<option value="ss_server">ss-server</option>
+		</select>
 	</div>
+	<div>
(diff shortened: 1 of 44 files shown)


dibdot:
Thanks for the re-design! Just a short feedback after first quick look:

Due to removing the up/down button images, travelmate doesn't show Up/Down navigation any longer. Could you please add the values to stations.htm, too?

travelmate

There is a line break missing after section titles, e.g.:
firewall

The "Add" button is not well aligned, e.g.:
routes

jow-:
Which browser was that?

dibdot:
Firefox 60.0.2 (64-bit) / Linux
Chrome Version 67.0.3396.79 (Developer Build) built on Debian buster/sid, running on Debian buster/sid (64-bit)

OpenWrt SNAPSHOT r7093+185-4fdc6ca31b with the changes in this PR.

jow-:
Thanks! I pushed updates fixing the mentioned issues.

dibdot:
Many thanks, most of the issues are fixed. Still, the "Add" button alignment is not correct, e.g. bootstrap vs. openwrt theme:


bootstrap

openwrt

jow-:
The stretched add button is/was actually intentional to make it more obvious that it relates to adding another section item. After your remark I changed the CSS to make it inline-flex and left aligned. Maybe you need to clear the cache to force relaoding the CSS?

dibdot:
Ups, you are correct. After refreshing the browser cache, all looks good now. Sorry for the noise.

Ansuel:
@jow- if you want i will try to update the material theme with all this new things...  (a list of the broken thing would be good)

(and I think that you need to update also the luci-app-ddns with the row.push function instead of the append child)

jow-:
@Ansuel - I don't think there'y anything specifically broken in material, it just needs tuning here and there. My main point about it is the fact that it requires jquery.js - that should be replaced by vanilla JS and some CSS anmiations.

applications/luci-app-shadowsocks-libev/luasrc/view/shadowsocks-libev/add_instance.htm
```diff
@@ -1,24 +1,17 @@
...
+			<option value="ss_local">ss-local</option>
+			<option value="ss_tunnel">ss-tunnel</option>
+			<option value="ss_redir">ss-redir</option>
+			<option value="ss_server">ss-server</option>
+		</select>
 	</div>
+	<div>
+		<input type="text" class="cbi-input-text" id="_newinst.name" name="_newinst.name" placeholder="<%:Name%>"/>
+	</div>
+	<input type="submit" class="cbi-button cbi-button-add" name="cbi.cts.<%=self.config%>" value="<%:Add%>" />
```
yousong: Hi, the three widgets were expected to be placed on the same row, but now they were stacked closed together on the desktop chrome browser.

Another thing is that the `td` that contains the text "This section contains no values yet" seems to lack a `colspan=N` attribute causing the border-top not spanning the whole row.

jow-: @yousong - thanks fixed

yousong: @jow- yes, the layout restored but other worms sneak out ;)

The first being the unneeded border-top for the Status line of interface.



The second being that now the first `td` of an empty now takes up the whole viewpoint that the page is unclickable.

jow-: Note that the parent tr has "position relative" so the td should be confied to it. Might be some chrome/webkit specific issue though, will check.

jow-: @yousong - indeed Webkit specific as `display: table-row` Items have undefined behaviour in this context according to spec. Solved it differently now.

PR #1896, opened 2018-06-20
Add sort by signal feature to the wifi scanlist function

tt33415366:
It's common that people always want to join a better wireless network, so this sort by signal feature might be reasonable.


jow-:
Closing this as it does not apply anymore.

PR #1902, opened 2018-06-21
Update 40_luci-transmission

panda-mute:
Fix typo


panda-mute:
Other branch should have this issue too.

PR #1905, opened 2018-06-23
luci-app=bcp38: improve interface selection dropdown

ldir-EDB0:
Certain types of interface were excluded from bcp38's concept of
upstream wan interface e.g. pppoa-wan

Shameless copy the interface list generation from luci-app-sqm which
solved exactly the same problem and gives additional clues by displaying
the firewall zone (typically wan/lan etc)

Run-tested-on: DGN3500

Signed-off-by: Kevin Darbyshire-Bryant <ldir@darbyshire-bryant.me.uk>


jow-:
Merged, thanks!

PR #1913, opened 2018-06-26
luci-mod-admin-full: ralink support wpa/wpa2 encryption

Vonger:
current luci code is not well compatible with ralink wifi driver, so device based on ralink chip such like vocore2 will not able to setting up its wireless encryption mode to wpa/wpa2, this patch is used to fix such problem.

IMHO, WPA/WPA2 should not special for "broadcom", but it should for the all drivers without mac80211 support, currently almost every wireless router already supported wpa and wpa2, but wep is already abandoned.


tt33415366:
You must comprised the OpenWRT source with Ralink Specific Drivers to build your project, that's not the case where OpenWRT should officially fit.

jow-:
Does not apply to OpenWrt.

Issue #1914, opened 2018-06-27
luci-app-ddns: does not correctly determine that bind-host is installed et al


pleasantone:
https://github.com/openwrt/luci/blob/660e4f5504b91a7b5c51a6abb80d52c65b8528d2/applications/luci-app-ddns/luasrc/tools/ddns.lua#L116

Line 116 has a typo -- "type" not "tyep"

@Ansuel

Ansuel:
will fix this and another problem when anothe pr gets merged

pleasantone:
Pull request #1928

yangfl:
fixed in #2029

feckert:
Should be fixed

PR #1865, opened 2018-06-08
luci-app-ddns: various fix

Ansuel:
Rework overview page, sync with new style and simplify the code by removing unnecessary check

Signed-off-by: Ansuel Smith <ansuelsmth@gmail.com>

PR #1928, opened 2018-07-02
luci-app-ddns: fix recognizing nslookup tools

pleasantone:
Fix typo in luci-app-ddns that causes recognition of alternative nslookup utilities to fail.

Closes issue #1914 @Ansuel


Ansuel:
actually... there is already a pull #1865

pleasantone:
removed

PR #1931, opened 2018-07-03
base.po: Update Chinese translation

angusding:
Fix some Chinese translations.
Signed-off-by: Angus Ding angus.ding@gmail.com


yousong:
Hi @angusding , in case it may help, you can just force push to your branch to update it so that github will update existing pull requests here.  This workflow may save quite a few clicking and typing.

PR #1933, opened 2018-07-03
luci-app-upnp: add hostname info

Ansuel:
This adds the hostname info in the index page of luci.

### DIFF
--- a/applications/luci-app-upnp/luasrc/view/upnp_status.htm
+++ b/applications/luci-app-upnp/luasrc/view/upnp_status.htm
@@ -23,6 +23,7 @@
 						st[i].proto,
 						st[i].extport,
 						st[i].intaddr,
+						st[i].host_hint || "<%:Unknown%>",
 						st[i].intport,
 						st[i].descr,
 						E('<input class="cbi-button cbi-button-remove" type="button" value="<%:Delete%>" onclick="upnp_delete_fwd(%d)" />'.format(st[i].num))
@@ -41,6 +42,7 @@
 			<div class="th"><%:Protocol%></div>
 			<div class="th"><%:External Port%></div>
 			<div class="th"><%:Client Address%></div>
+			<div class="th"><%:Host%></div>
 			<div class="th"><%:Client Port%></div>
 			<div class="th"><%:Description%></div>
 			<div class="th cbi-section-actions">&#160;</div>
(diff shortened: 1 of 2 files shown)


jow-:
NACK. Please use the host_hints procedures from luci.sys.net to get mapping info. The ad-hoc lease parsing code here is not robust enough, it will also fail if another lease file path has been configured in /etc/config/dhcp.

jow-:
Use `luci.sys.net.ipv4_hints()` - this returns a table of table, with each table containing an IPv4 address as first item and a hostname (or MAC as fallback) as second item.

Ansuel:
didn't need the example (thx anyway) , you were super clear 

Anyway thx for the function, tested and works ;)

Think i should also include this in the upnp info page

jow-:
Looks good so far but needs some more improvements, please...

  - move the `luci.sys.net.ipv4_hints()` call out of the `while true` loop (its expensive!) and reuse the `ipv4_hints` variable in each iteration
  - declare both `_` and `tbl` as `local` (and maybe consider renaming `tbl` to `e` which I commonly use as iter variable abbreviation for `entry`
  - rename `intaddr_resolved` to `host_hint` as this same field name is already used for DHCP lease status and assoc list info

applications/luci-app-upnp/luasrc/view/upnp_status.htm
```diff
@@ -23,6 +23,7 @@
 						st[i].proto,
 						st[i].extport,
 						st[i].intaddr,
+						st[i].intaddr_resolved,
```
jow-: Maybe consider adding an ` || '-'` here since there might not always be a value available

applications/luci-app-upnp/luasrc/view/upnp_status.htm
```diff
@@ -41,6 +42,7 @@ <h3><%:Active UPnP Redirects%></h3>
 			<div class="th"><%:Protocol%></div>
 			<div class="th"><%:External Port%></div>
 			<div class="th"><%:Client Address%></div>
+			<div class="th"><%:HostName%></div>
```
jow-: Maybe just call this one `Host` as there can be either a hostname, or a MAC or nothing... depending on available info.

Ansuel:
Can you check this @jow- ?

jow-:
Merged, thanks!

PR #1936, opened 2018-07-04

YuriPet:
Updated with the latest synchronization of the translation, corrections and additions translation.


modules/luci-base/po/uk/base.po
```diff
@@ -2056,6 +2058,8 @@ msgid ""
 "Maximum length of the name is 15 characters including the automatic protocol/"
 "bridge prefix (br-, 6in4-, pppoe- etc.)"
 msgstr ""
+"Максимальна довжина імені становить 15 символів, включаючи префікс
```
musashino205: The double quotation at the end of the line is missing.

PR #1937, opened 2018-07-05
luci-proto-ppp: expose pppoe host_uniq tag setting

LuKePicci:
The host_uniq tag setting is supported by uci since CC15. It is actually the only setting that is not available within LuCi to support full internet service setup for some ISP to be completed by webui only.

The PPPoE RFC provide no upper bound to tag length, so its field datatype is unlimited length hex-encoded bytes string.

Signed-off-by: Luca Piccirillo <luca.piccirillo@gmail.com>

### DIFF
--- a/protocols/luci-proto-ppp/luasrc/model/cbi/admin_network/proto_pppoe.lua
+++ b/protocols/luci-proto-ppp/luasrc/model/cbi/admin_network/proto_pppoe.lua
@@ -114,6 +114,14 @@ keepalive_interval.placeholder = "5"
 keepalive_interval.datatype    = "min(1)"
 
 
+host_uniq = section:taboption("advanced", Value, "host_uniq",
+	translate("Host-Uniq tag content"),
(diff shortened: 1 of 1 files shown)


dibdot:
Thanks! Please use the translate() function for the placeholder string, like already used in the changed file.

LuKePicci:
There is "auto" also for AC and service name so I opted to be consistent. If omitted, the host_uniq tag will be set, not by auto-detection but by pppoe protocol design, to that one proposed by the pppoe server. 
If you would like, I can also change all three placeholders with something that make sense according to the standard definition https://tools.ietf.org/html/rfc2516
My placeholders proposal: 
ac -> "No AC-Name" 
service -> "Any service"
host_uniq -> "Proposed by server"

dibdot:
I think it's quite OK now. Before apply please squash both commits to one - thanks for your contribution.

dibdot:
Thanks!

PR #1938, opened 2018-07-06
luci-app-firewall: Ease creation of multiple day traffic rules

kristrev:
Creating a traffic rule that spans multiple days currently requires
creating three rules, unless the rule only covers complete days. You
need to create one rule for the first day, one rule for the complete
days, and then one rule for the final day. In addition to making
creating multi day-rules a bit of as hassle, performance suffers when
many multi day-rules are present on a system.

The --datestart/--datestop arguments to iptables, which is what
Start/Stop Date maps to, supports specifying a time of day. This PR
extends Start/Stop Date to also accept time. The new format of the
fields are yyyy-mm-ddThh:mm:ss. As before, Start/Stop Time will override
the time specified in Start/Stop Date.

I did not find a nice way to make the time-part of Start/Stop Data
optional. I don't believe this matters to much. If a user edits an
existing rule, he or she will get a validation error when the rule is
saved. However, all that has to be done to make the date(s) pass
validation (and keep the old behavior) is to add "T00:00:00" to the date
field(s).

The translations are also updated, with the exception of Russian and
Ukrainian. My knowledge of the cyrillic alphabet is a bit limited, to say
the least.


jow-:
Hmm ,a couple of thoughts...

  - I do not really like having any more markup in the translation strings, is the forced `<br>` really nescessary?
 - Instead of adding another datatime data type I'd rather extend the existing one with an "with time" flag or similar, see how e.g. the strict flag is handled for hostnames
 - I'd really prefer to have the rules accept both date only and datetime strings, iptables does it too, right?

kristrev:
> I do not really like having any more markup in the translation strings, is the forced <br> really nescessary?

No, not really. However, without it, the Start/Stop Date strings are broken in the middle of formatting.

> Instead of adding another datatime data type I'd rather extend the existing one with an "with time" flag or similar, see how e.g. the strict flag is handled for hostnames

Good idea, I was not aware of the strict flag and will take a look.

> I'd really prefer to have the rules accept both date only and datetime strings, iptables does it too, right?

Yes, that is how it works today. iptables will just insert 00:00:00 if not time is provided. I guess adding a flag will take care of this. Without flag, only date is accepted, and with flag, datetime is accepted.

kristrev:
I managed to royally mess up my branch, so I decided it was easiest to create a new PR and close thise one. I have incorporated the changes you asked for and created PR #2139.

PR #1940, opened 2018-07-06
luci-mod-admin-full: Respect default wifi section names

kristrev:
The default wifi-iface sections (the ones created by mac80211.sh) are
named "default_radioX". Respect this naming scheme when either replacing
the current wifi configuration, or when adding a network to a device
without any networks.

### DIFF
--- a/modules/luci-base/luasrc/model/network.lua
+++ b/modules/luci-base/luasrc/model/network.lua
@@ -1476,11 +1476,24 @@ function wifidev.get_wifinets(self)
 	return nets
 end
 
-function wifidev.add_wifinet(self, options)
+function wifidev.get_num_wifinets(self)
+	local num_wifinets = 0
+
+	_uci:foreach("wireless", "wifi-iface",
+		function(s)
+			if s.device == self.sid then
+				num_wifinets = num_wifinets + 1
+			end
+		end)
+
+	return num_wifinets
+end
+
+function wifidev.add_wifinet(self, name, options)
 	options = options or { }
 	options.device = self.sid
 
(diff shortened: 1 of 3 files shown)


jow-:
Hmm, on a first sight I am not extremely happy with it, can you elaborate some more on why this is desired? Is this related to the wifi join section mismatch issue?

kristrev:
Sorry for my late reply, I have been on holiday and not had time to sit down and write a proper reply.

Yes, this fix is an attempt to resolve the wifi replace config issue. On routers upgraded from earlier versions of OpenWRT and where anonymous sections are used for my 2.4GHz/5Ghz radios, I see incorrect behavior in Luci when doing the following:

* Connect 2.4 Ghz to a wifi network.
* Select replace config.
* The anonymous sections get out of sync.

What happens is that the names of the wifi-iface sections for the 2.4GHz/5Ghz radios are swapped. This swap is not seen by Luci, so the "Wireless Network"-page shows information about the 5Ghz wifi-device (but the correct wifi-iface). Also, on the "Wireless Network"-page it looks like a second network has been added to the 5GHz radio (there is a radio1-1 tab). 

The error is not that critical, as everything is correct when the configuration is applied. If I repeat the steps above again, then I also don't see the out of sync issue (I have not tried to recreate by messing around and creating anonymous sections manually). However, it is quite confusing when the issue appears.

I see that I missed one detail when preparing this code. On Broadcom-based devices, the device section is named `default_wlX`. So my code is anyway not correct and will have to be fixed either way.

kristrev:
This PR was based on an incorrect assumption and would anyway not fix the issue properly. I will close it.

PR #1944, opened 2018-07-08
luci-proto-wireguard: description field

rwalli:
For wireguard it would be handy to have a description field, especially when you have many peers.

Signed-off-by: Robert Walli <rwalli@gmx.net>


kofec:
@rwalli 
Maybe you can also add this field to "Status -> Wireguard Status" ?
Thanks

rwalli:
@kofec 
You can add this too and will be probably faster than me because i'm not familiar with the luci system.

PR #1946, opened 2018-07-11
luci-app-transmission: Allow user to define custom web interface

korzhyk:
There are exists several custom web ui for transmission. Such as:
https://github.com/ronggang/transmission-web-control
https://github.com/Secretmapper/combustion
https://github.com/fcsonline/react-transmission
But for now there are no way to use custom ui without editing `init.d` script.


korzhyk:
Also implemented in `transmission` package https://github.com/openwrt/packages/pull/6459

PR #1951, opened 2018-07-15
luci-base: network.lua: bridge information request functions fixes

namedun:
Fixes these bugs:

1. Functions interface.bridge_id() and interface.bridge_stp() uses self.br field
   that is not exists at all
2. Function interface.is_bridgeport() return true only for bridge interface, not for
   bridge port interfaces.
3. Function interface.ports() is not return anything

### DIFF
--- a/modules/luci-compat/luasrc/model/network.lua
+++ b/modules/luci-compat/luasrc/model/network.lua
@@ -1447,6 +1447,7 @@ function interface.ports(self)
 		for _, iface in ipairs(members) do
 			ifaces[#ifaces+1] = interface(iface)
 		end
+		return ifaces
 	end
 end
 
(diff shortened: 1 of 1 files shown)


namedun:
@hnyman @jow- 
That about this PR? Why not merge?

feckert:
@namedun please rebase your change if you still think this is needed.
I'm afraid I can't say anything about that because I have no idea why it's useful. Maybe you can say something about it

namedun:
Rebased.
 
Functions which are fixed by this PR are not used anywhere in luci-base and standard LuCI applications. However, these functions are used in our mstpd support LuCI application (https://github.com/tano-systems/luci-app-tn-mstpd).

jow-:
Merged it, sorry for the long wait time.

PR #1958, opened 2018-07-17
luci-base: update and improve Russian translation

namedun:
Add missing translations and update existing not quite correct translations.
Replaced hyphens on em dashes where it is required by the Russian rules.


jow-:
Merged - thank you!

PR #1964, opened 2018-07-18
luci-proto-ncm: remove mode 'auto' as default

hanipouspilot:
The default seting should be none.
Most modems have persistent mode setting that is stored in internal flash.
If option mode is unset the default modem setting is used.
If we choose some mode, then on every connection we write to modem flash.
In the current state we can't disable that from LuCI.

### DIFF
--- a/protocols/luci-proto-ncm/luasrc/model/cbi/admin_network/proto_ncm.lua
+++ b/protocols/luci-proto-ncm/luasrc/model/cbi/admin_network/proto_ncm.lua
@@ -32,7 +32,7 @@ end
 
 
 mode = section:taboption("general", Value, "mode", translate("Service Type"))
-mode.default = "auto"
+mode:value("", translate("Modem default"))
 mode:value("preferlte", translate("Prefer LTE"))
 mode:value("preferumts", translate("Prefer UMTS"))
 mode:value("lte", "LTE")


hanipouspilot:
At least I can select 'custom' and add an empty string.
I thought about adding exactly that, but I wasn't sure that gcom will work correctly when there is
option mode ''
in /etc/config/network

I 'll test it tomorrow. Probably 3g needs that too. I'll take care of that.

hanipouspilot:
Now the problem is that if the interface is set without using LuCI, and then go into GUI it is impossible to save it without adding some mode.

hanipouspilot:
Another way, probably a better one in terms of usability, is to leave Luci as is, but make a change in Openwrt code.
It is possible to check the current modem setting and send a command if the value is different.
But it is more complex. I can easily do it for Huawei, but other vendors need testing.

hanipouspilot:
I added a commot with "Modem default" as a default mode. It looks an easiest option.

jow-:
@hanipouspilot - Looks good, can you try if it also works properly if you remove the `mode.default = ""` line entirely? There should be no need to set it o the empty string.

hanipouspilot:
@jow- Yes, it does.
I rebased it to one commit.

PR #1972, opened 2018-07-20
luci-base, luci-app-upnp, luci-app-wol: update Ukrainian translation

YuriPet:
Updated with the latest resync translation, corrections and additions.


jow-:
Merged, thanks! 🙂

PR #1979, opened 2018-07-21
support host

jjm2473:
some shadowsocks server providers provide a host as server address.


yousong:
Hi, @jjm2473 ,

I would suggest just switching to using current rewritten version of luci-app-shadowsocks-libev.  The 17.01 branch is soon to be EOL.

However, if you are still interested in seeing this merged, please consider rewording the commit subject to something like `shadowsocks-libev: validate server address as "host" type` and sign off the commit.  Thank you.

PR #1980, opened 2018-07-21
luci-proto-ncm: add pdptype option

hanipouspilot:


jow-:
merged - thanks

hanipouspilot:
@jow Do we really need that "LCP echo", etc stuff in this proto? It doesn't work either way.

jow-:
Not at all, please feel free to.

PR #1981, opened 2018-07-21
luci-proto-ncm: remove unrelated options

hanipouspilot:
This removes all options 100% unrelated to NCM protocol.
Some options like 'dns' are not currently used by connection scripts,
but may be used in the future.


hanipouspilot:
I also changed 'maxwait' to 'delay' option that is relevant in this proto.

jow-:
I think `dns` is handled in a generic manner by netifd. Change looks good, thank you!

hanipouspilot:
@jow-  I doubt that neifd see dns settings. A virtual interface like wan_4, or wan_6 is created and parameters are passed by ncm.sh. Metric is passed, but I don't see anything else. I didn't look deep enough though.

PR #1982, opened 2018-07-21
luci-base: update Polish translation

Rixerx:
Updated Polish translations.

@jow- @musashino205 
I have a question to you. In commit https://github.com/openwrt/luci/commit/86660f92d12b963a05118a8f9471513a18773be4 and https://github.com/openwrt/luci/commit/524ce90c4e22dcb64ac8e6c2e012eb54cd5f098e, you have added "Alias interface" and "Alias Interface". Should there be almost identical expressions here?

Signed-off-by: Krystian Kozak <krystian.kozak20@gmail.com>


jow-:
@Rixerx - good point, will look into merging those two.

PR #1983, opened 2018-07-21
luci-app-adblock: sync with adblock 3.5.4

dibdot:
backend:
* add low priority mode (nice level 10), disabled by default
* enhance 'Force DNS' to redirect ports 53, 853 and 5353

frontend:
* switch to dynamic XHR polling for runtime information and logfile
viewing
* add new 'Refresh' button to reload blocklists
* various cleanups & small fixes

### DIFF
--- a/applications/luci-app-adblock/luasrc/view/adblock/runtime.htm
+++ b/applications/luci-app-adblock/luasrc/view/adblock/runtime.htm
@@ -1,10 +1,192 @@
 <%#
 Copyright 2017-2018 Dirk Brenken (dev@brenken.org)
 This is free software, licensed under the Apache License, Version 2.0
+local sys     = require("luci.sys")
+
 -%>
+<style type="text/css">
+.runtime
+{
+	color:#0069d6;
+	font-weight:bold;
+	display:inline-block;
+	width:100%;
+	padding-top: 0.5rem;
+}
+</style>
 
-<%+cbi/valueheader%>
+<script type="text/javascript">
+//<![CDATA[
+	function status_update(json)
+	{
+			var view  = document.getElementById("value_1");
+			var btn1  = document.getElementById("btn1");
+			var btn2  = document.getElementById("btn2");
+			var input = json.data.adblock_status;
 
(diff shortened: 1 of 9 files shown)


dibdot:
the frontend part has been tested with all three themes, the overview page / runtime information now looks like that:

adblock

runtime information & buttons are now updated via XHR.

jow-:
Merged - thanks!

applications/luci-app-adblock/luasrc/view/adblock/runtime.htm
```diff
@@ -1,10 +1,192 @@
...
 
-<%+cbi/valueheader%>
+	XHR.get('<%=luci.dispatcher.build_url("admin", "services", "adblock", "status")%>', null,
+	function(x, json_info)
+	{
+		if (!x || !json_info)
+		{
+			var btn1 = document.getElementById("btn1");
+			var btn2 = document.getElementById("btn2");
+			btn1.value = "Suspend";
```
jow-: Note that you'll lose translations here. Should it be `btn1.value = "<%:Suspend%>";` maybe? I noticed that you use the label strings on the server side to decide which action was invoked, which would fail if the buttons get translated. But you could fix that by using a hidden field or by maybe changing the `name` attribute instead so that you check for `if http.formvalue("do_suspend") then` instead of `if http.formvalue("fixed_name") == "Suspend" then`

applications/luci-app-adblock/luasrc/view/adblock/runtime.htm
```diff
@@ -1,10 +1,192 @@
...
+		}
+		status_update(json_info)
+	});
+//]]>
+</script>
 
-<%+cbi/valuefooter%>
+<h3>Runtime Information</h3>
+<div class="cbi-value" id="status_1">
+	<label class="cbi-value-title" for="status_1">Adblock Status</label>
```
jow-: Same here, this seems to have been translated before, I just noticed that strings like `Adblock Status` got dropped with the translation refresh.

dibdot:
@jow- translations - a valid point. Thanks, I will rework this part.

dibdot:
@jow- I don't think so, in 18.06 package branch is currently v3.5.2 and the LuCI part is completely untested with 18.06, too. I plan to backport & test all changes of 3.5.3 & 3.5.4 to 18.06 later on (after my vacation ;-)

PR #1989, opened 2018-07-23
luci-proto-ncm: suggest both tty and wdm devices

hanipouspilot:
Some modems are controlled only through a tty.
Autodetection of control device is broken in ncm.sh


hanipouspilot:
I am working on fixing 'wwan' and 'ncm" protocols.
Currently the first one doesn't really work with NCM for several reasons.
The pure NCM requires $ctl_device that is not there when called separately.

jow-:
Is the change really correct that you want *either* `ttyUSB*` or `cdc-wdm` ? Or did you mean to somehow concat both lists? Because your pull request title does not agree with the code change.

jow-:
Okay, I just wanted to be sure that you're aware that with your code change, LuCI will only ever suggest either `ttyUSB*` or `cdc-wdm*`. If any `/dev/ttyUSBx` exists on the system, it will use the result of `cdc-wdm*`.

This code:
```
local device_suggestions = nixio.fs.glob("/dev/ttyUSB*")
+	and nixio.fs.glob("/dev/cdc-wdm*")
```
is identical to
```
local device_suggestions
if nixio.fs.glob("/dev/ttyUSB*") then
    device_suggestions = nixio.fs.glob("/dev/cdc-wdm*")
else
    device_suggestions = nil
end
```

jow-:
Did you maybe mean something like that?

```
local dev
for dev in nixio.fs.glob("/dev/ttyUSB*") do
    device:value(dev)
end
for dev in nixio.fs.glob("/dev/cdc-wdm*") do
    device:value(dev)
end
```

hanipouspilot:
Hold on, I made this change and saw both. I can't test it now, because I don't have a modem here.
I want that it suggests tty if it exists (better all ttys) and cdc-wdm too.

hanipouspilot:
That one look correct. I am not good in this type of coding. I change it and it looked to work correctly.
But then I noticed that only one tty was suggested.

Your variant seem to be perfect.

hanipouspilot:
If you look into /lib/netifd/proto/wwan.sh and ncm.sh, you'll see that they are trying to find the correct tty. But they don't succeed in may cases, sent commands to a wrong revice, etc.

Will you do this change, or do I need to rebase?

hanipouspilot:
Jow,

What do you think about that autodetection? Is it worth the effort. I wrote it and it worked well until I changed the modem persistent setting to have both RAS and NCM ports. Then two ttys appeared in random order ttyUSB0 and ttyUSB1 and I had to probe which is the correct one. But it is possible to open a few more ports, etc. The script would be too heavy and still not universal, because other nodels need testing. so what do you think?

jow-:
I can't really comment on that as I do not know the details of NCM enough.

hanipouspilot:
@jow- You are correct. This shows all needed devs.

PR #1990, opened 2018-07-23
luci-app-nlbwmon: Add proper file names for json and csv exports.

ameer1234567890:
Currently, the json and csv exports are just named `data` which is not intuitive. This adds proper file name extensions using Content-disposition HTTP header.


jow-:
Merged with b3642f476 - thanks!

PR #2012, opened 2018-07-30
fix compilation failure when luasrcdiet is being copied to non existing directory

jempatel:
fix compilation failure when luasrcdiet is being copied to non existing directory


jow-:
Good catch! Can you amend your commit with a Signed-off-by? Use `git commit --signoff --amend; git push origin master --force` to update this PR inplace.

jempatel:
Done

jow-:
merged, thank you

PR #2013, opened 2018-07-30
luci-app-adblock: backport 3.5.4 to 18.06 branch

dibdot:
Tested with latest 18.06 branch (OpenWrt 18.06-SNAPSHOT, r7175+5-ca0c649a38)

### DIFF
--- a/applications/luci-app-adblock/luasrc/controller/adblock.lua
+++ b/applications/luci-app-adblock/luasrc/controller/adblock.lua
@@ -3,9 +3,12 @@
 
 module("luci.controller.adblock", package.seeall)
 
+local sys   = require("luci.sys")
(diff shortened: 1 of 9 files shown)


dibdot:
@jow- @hnyman I'm not sure if it's too late for 18.06 - the backports for adblock and travelmate are now tested & ready here and in package repo.

hnyman:
If you have tested them and they are ok, let's just merge them.

normal user-space packages (including LuCI modules) are constantly built for 18.06 branch by the 18.06 buildbot, so it is quite possible to fix/enhance things also after the release, and users will find the new versions with opkg. (The release is tightly fixed only for the firmware images, some core packages and kernel-dependent packages. That is simplification, but paints the picture.)

PR #2021, opened 2018-07-31
luci-app-travelmate: "final" fixes

dibdot:
during intense testing with different browsers (Chrome/Firefox/partly IE
in a VM) I found & fixed some more minor things:

* re-add accidently removed "apply_on_parse" attribute in overview cbi
* fixed a corner case where the "Restart" button not works correctly
* Removed leftovers from last commit


jow-:
@dibdot - feel free to merge but please use `git cherry-pick -x <commithash>` to pull it into 18.06.

Issue #2003, opened 2018-07-28
DSL Overview


Rixerx:
In Firefox and IE i got doubled table line (LuCI Master (git-18.208.43643-e5a1ac0)).

![2](https://user-images.githubusercontent.com/25949215/43354163-46931138-9247-11e8-92e0-cf7d3a551943.png)

Is it possible to rebuild DSL status, for example like wireless box?

PR #2026, opened 2018-07-31
luci-mod-admin-full: Fix DSL Stats display of dB statistics

alanswanson:
Line attenuation, signal attenuation, noise margin and aggregate transmit power really need to show
decimal digits. Fixes commit 88713f6 from issue #2003 due to excess changes from %s to %d.

Needs applied to both master and 18.06 branches.


jow-:
Please do not use `%s` for numeric data. Use something like `%0.2f` and replace `2` with the number of significant digits you want to show.

jow-:
Merged, thank you!

PR #1819, opened 2018-05-24
luci-app-ddns: improve performance

Ansuel:
Every request directed to the ddns app call ddns tools module.
Ddns tools module have lots of global variable that call slow os.execute function. This adds 10 second to every ddns request even if the function that is requested doesn't need that global variable. This commit introduce env_info function that execute os.execute command by executing what is actually requested and not process all the variables. Also remove 2 unecessary module that are not used. More researh find that major slowdown was caused by the calling of ddns script for the version check. Now we check if opkg is present and use it to check ddns-scripts version.

Signed-off-by: Ansuel Smith <ansuelsmth@gmail.com>

PR #2029, opened 2018-08-02
luci-app-ddns: fix huge typo in luasrc/tools/ddns.lua

yangfl:

### DIFF
--- a/applications/luci-app-ddns/luasrc/tools/ddns.lua
+++ b/applications/luci-app-ddns/luasrc/tools/ddns.lua
@@ -10,121 +10,120 @@ local SYS  = require "luci.sys"
 
 function env_info(type)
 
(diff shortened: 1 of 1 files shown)


hnyman:
Thanks for fixing the typo in tyep-->type originating from #1819 

This might be needed to be backported to 18.06, too.

@Ansuel , you authored the original changes. Somehow the app has worked with that typo, so I wonder if there have been some later workarounds around the effects of the typo, that could now be removed?
Or has it just been partially broken?

yangfl:
@hnyman It was partially broken, that is, no matter what packages you have installed, ddns app always complained about "Using specific DNS Server not supported" (in luasrc/model/cbi/ddns/hints.lua#140)

jow-:
The change looks legit to me, will merge it.

yangfl:
@jow- you might also want to merge #1865

PR #2030, opened 2018-08-02
luci-base: update Chinese translation

yangfl:

### DIFF
--- a/modules/luci-base/po/zh-cn/base.po
+++ b/modules/luci-base/po/zh-cn/base.po
@@ -1,7 +1,15 @@
+#
+# Yangfl <mmyangfl@gmail.com>, 2018.
+#
 msgid ""
 msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
-"Last-Translator: Hsing-Wang Liao <kuoruan@gmail.com>\n"
+"Last-Translator: Yangfl <mmyangfl@gmail.com>\n"
+"Language-Team:  <debian-l10n-chinese@lists.debian.org>\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=1; plural=0;\n"
+"PO-Revision-Date: 2018-08-07 14:59+0800\n"
+"X-Generator: Gtranslator 2.91.7\n"
 
 msgid "%.1f dB"
 msgstr "%.1f dB"
@@ -129,7 +137,7 @@ msgid "<abbr title=\"Media Access Control\">MAC</abbr>-Address"
 msgstr "<abbr title=\"Media Access Control\">MAC</abbr> 地址"
 
 msgid "<abbr title=\"The DHCP Unique Identifier\">DUID</abbr>"
-msgstr ""
+msgstr "<abbr title=\"The DHCP Unique Identifier\">DUID</abbr>"
 
 msgid ""
 "<abbr title=\"maximal\">Max.</abbr> <abbr title=\"Dynamic Host Configuration "
@@ -315,7 +323,7 @@ msgid ""
 "Always use 40MHz channels even if the secondary channel overlaps. Using this "
 "option does not comply with IEEE 802.11n-2009!"
 msgstr ""
-"即使辅助信道重叠，也始终使用 40MHz 信道。使用此选项不符合IEEE 802.11n-2009！"
+"即使辅助信道重叠，也始终使用 40MHz 信道。使用此选项不符合 IEEE 802.11n-2009！"
 
 msgid "Annex"
 msgstr "Annex"
@@ -647,10 +655,10 @@ msgid ""
 msgstr "在给定时间（秒）后关闭非活动链接，0 为保持连接"
 
 msgid "Close list..."
-msgstr "关闭列表..."
+msgstr "关闭列表…"
 
 msgid "Collecting data..."
-msgstr "正在收集数据..."
+msgstr "正在收集数据…"
 
 msgid "Command"
 msgstr "命令"
@@ -872,13 +880,13 @@ msgid "Device Configuration"
 msgstr "设备配置"
 
 msgid "Device is rebooting..."
-msgstr "设备正在重启..."
+msgstr "设备正在重启…"
 
 msgid "Device unreachable!"
 msgstr "无法连接到设备"
 
 msgid "Device unreachable! Still waiting for device..."
-msgstr "无法连接到设备！仍旧等待设备..."
+msgstr "无法连接到设备！仍旧等待设备…"
 
 msgid "Diagnostics"
 msgstr "网络诊断"
@@ -915,7 +923,7 @@ msgid "Disabled (default)"
 msgstr "已禁用（默认）"
 
 msgid "Disassociate On Low Acknowledgement"
-msgstr "在 Low Acknowledgement 时断开连接"
+msgstr "在低 Ack 应答时断开连接"
 
 msgid "Discard upstream RFC1918 responses"
 msgstr "丢弃 RFC1918 上行响应数据"
@@ -1042,7 +1050,7 @@ msgid ""
 "Enable <abbr title=\"Internet Group Management Protocol\">IGMP</abbr> "
 "snooping"
 msgstr ""
-"启用 <abbr title=\"Internet Group Management Protocol\">IGMP</abbr> 窥探"
+"启用 <abbr title=\"Internet Group Management Protocol\">IGMP</abbr> 嗅探"
 
 msgid "Enable <abbr title=\"Spanning Tree Protocol\">STP</abbr>"
 msgstr "开启 <abbr title=\"Spanning Tree Protocol\">STP</abbr>"
@@ -1134,7 +1142,7 @@ msgid "Enter custom values"
 msgstr "输入自定义值"
 
 msgid "Erasing..."
-msgstr "擦除中..."
+msgstr "擦除中…"
 
 msgid "Error"
 msgstr "错误"
@@ -1192,7 +1200,7 @@ msgid "FT protocol"
 msgstr "FT 协议"
 
 msgid "Failed to confirm apply within %ds, waiting for rollback…"
-msgstr "在 %d 秒内确认应用失败，等待回滚..."
+msgstr "在 %d 秒内确认应用失败，等待回滚…"
 
 msgid "File"
 msgstr "文件"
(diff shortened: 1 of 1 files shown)


yousong:
Please drop the translation of IPv4, DNS, SSH, etc.

yangfl:
@yousong That would be very unhelpful such that using English to explain English, given the situation that end users can't read English at all.

yousong:
It's not English literature essays we are talking about, it's technical glossaries we expect target users to know beforehand.  If they have no idea what IPv4, SSH, DNS or DHCP is, then more likely so that they will have no idea what "互联网协议第4版", "安全壳层", "域名系统“, “动态主机配置协议” refer to, these names hardly convey additional information yet it clutters the UI and causes pain for others who will have a hard time with the question of what "安全壳层" refers to.

yangfl:
@yousong Then I'd like to remove those full English names as well. While Chinese users might be aware of those glossaries like IPv4, SSH, DNS or DHCP, I heavily doubt they know anything about "Internet Protocol Version 4", "Secure Shell", etc. Often annoying is that you put your mouse on "SSH" and it gives you a tip of "Secure Shell", and try to work out what "Secure Shell" means before you realize it's of no use to you. These names are only useful for English readers, not Chinese.

yangfl:
@yousong ok fixed

modules/luci-base/po/zh-cn/base.po
```diff
@@ -2134,7 +2142,7 @@ msgid "Network without interfaces."
 msgstr "无接口的网络。"
 
 msgid "Next »"
-msgstr "下一步 »"
+msgstr "前进 »"
```
yousong: The original translation should be more conventional

modules/luci-base/po/zh-cn/base.po
```diff
@@ -2606,7 +2614,8 @@ msgstr "质量"
 msgid ""
 "Query all available upstream <abbr title=\"Domain Name System\">DNS</abbr> "
 "servers"
-msgstr "查询所有可用的上游 <abbr title=\"Domain Name System\">DNS</abbr> 服务器"
+msgstr ""
+"查询所有可用的上游 <abbr title=\"Domain Name System\">DNS</abbr> 服务器"
```
yousong: A typo sneaked in?

modules/luci-base/po/zh-cn/base.po
```diff
@@ -3278,10 +3287,10 @@ msgstr "本机的硬件不支持多 SSID，如果继续，现有配置将被替
 msgid ""
 "The length of the IPv4 prefix in bits, the remainder is used in the IPv6 "
 "addresses."
-msgstr "IPv4 前缀长度（bit），其余的用在 IPv6 地址。"
+msgstr "IPv4 前缀长度（位），其余的用在 IPv6 地址。"
```
yousong: Please keep the original unit notation.

modules/luci-base/po/zh-cn/base.po
```diff
@@ -3278,10 +3287,10 @@ msgstr "本机的硬件不支持多 SSID，如果继续，现有配置将被替
...
 
 msgid "The length of the IPv6 prefix in bits"
-msgstr "IPv6 前缀长度（bit）"
+msgstr "IPv6 前缀长度（位）"
```
yousong: ditto

modules/luci-base/po/zh-cn/base.po
```diff
@@ -3797,7 +3806,7 @@ msgid ""
 "or Safari."
 msgstr ""
 "您的 IE 浏览器太老了，无法正常显示这个页面！请更新到 IE7 及以上或使用其他浏览"
-"器，例如：Chrome、Firefox、Opera、Safari。"
+"器，如Firefox、Opera、Safari。"
```
yousong: Why drop out Chrome?

The final punctuation should be a period in this case because it ends with alphabetical words.  I remembered someone told me this but I am not professional in this regard.

modules/luci-base/po/zh-cn/base.po
```diff
@@ -3945,10 +3954,10 @@ msgid "unspecified"
 msgstr "未指定"
 
 msgid "unspecified -or- create:"
-msgstr "不指定或新建："
+msgstr "未指定 或 新建："
```
yousong: I guess the original translation suits more the context.  The whitespace around "或" is superfluous and should be removed.

modules/luci-base/po/zh-tw/base.po
```diff
@@ -1,18 +1,21 @@
...
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=1; plural=0;\n"
-"X-Generator: Pootle 2.0.6\n"
+"X-Generator: Gtranslator 2.91.7\n"
 
 msgid "%.1f dB"
-msgstr ""
+msgstr "%.1f dB"
```
yousong: This change can be safely left out.  Original text will be retained if translation does not exist.  Very likely we need to modify source code to unmark this as translation-needed

modules/luci-base/po/zh-cn/base.po
```diff
@@ -3921,16 +3930,16 @@ msgid "server mode"
...
-msgstr "无状态的"
+msgstr "无状态"
 
 msgid "stateless + stateful"
-msgstr "无状态的 + 有状态的"
+msgstr "无状态 + 有状态"
 
 msgid "tagged"
-msgstr "已标记"
+msgstr "有 VLAN 标记"
```
yousong: The added "VLAN" in the translation is not needed.  It's clear from the context it's about VLAN tag.

modules/luci-base/po/zh-cn/base.po
```diff
@@ -3945,10 +3954,10 @@ msgid "unspecified"
 msgstr "未指定"
 
 msgid "unspecified -or- create:"
-msgstr "不指定或新建："
+msgstr "未指定 或 新建："
 
 msgid "untagged"
-msgstr "未标记"
+msgstr "无 VLAN 标记"
```
yousong: ditto

modules/luci-base/po/zh-cn/base.po
yangfl: clearly "位" is much more widely used than "bit"; you say 32位, 64位, etc.

modules/luci-base/po/zh-cn/base.po
yangfl: cause you didn't mention Chrome in the original text.

I think it should be full-width period because you treat Firefox, Opera, Safari as proper nouns.

modules/luci-base/po/zh-cn/base.po
yangfl: it should be matched with "« Back" "« 后退"

modules/luci-base/po/zh-cn/base.po
yangfl: The translation of "unspecified" should match the same word above. And the whitespace is used to highlight it's two actions, just like the original text "-or-", which puts two dashes around "or".

modules/luci-base/po/zh-cn/base.po
yangfl: maybe not. It's just normalized po.

modules/luci-base/po/zh-cn/base.po
yangfl: maybe, but I also observer that many people are confused by this, since they are not VLAN professors.

modules/luci-base/po/zh-tw/base.po
yangfl: just filling untranslated strings, but nah.

modules/luci-base/po/zh-cn/base.po
yousong: Got it.  Nice catch

modules/luci-base/po/zh-cn/base.po
yousong: ok.  No strong preference.

modules/luci-base/po/zh-cn/base.po
yousong: In the context this text appears in, "unspecified" is actually an indication of choice of *action*, instead of *state*.  That's the subtle difference between "不指定" and "未指定".

It's conventional we write the sentence without whitespace between them, it will not cause any noticeable trouble at all.  It's overkill to manually add spaces here.

modules/luci-base/po/zh-cn/base.po
yousong: Still they will not in better situation with the added VLAN text.  Please leave the original translation there, it describes the intention just fine.

modules/luci-base/po/zh-cn/base.po
yousong: Okay, just find an example in section 5.1.2.1 at [1] to suggest using the full-width width circle period, though I still thought it's better to use dot period in this case, just like we use period when the sentence ends with numericals

 [1] http://www.moe.gov.cn/ewebeditor/uploadfile/2015/01/13/20150113092346124.pdf

modules/luci-base/po/zh-cn/base.po
yangfl: So "unspecified" is actually an indication of choice of _state_, otherwise it will be "unspecify" (without past tense) here. "create", however, is an action (without past tense). It actually is this, leave the interface _unspecified_ to any firewall zone, or _create_ a new firewall zone. One is state, and another is action.

For the whitespace, I'm ok with or without it.

yousong: "choice of state", "leave the interface unspecified" are still description about *action*, not state.

modules/luci-base/po/zh-cn/base.po
yangfl: This usage is removed from the latest GB/T 15834-2011.

[1] https://web.archive.org/web/20121020111229/http://www.sasac.gov.cn/n1180/n1271/n6716920/n6716950/n6723785/6723901.html , Section 4.1.1.
[2] http://www.moe.gov.cn/ewebeditor/uploadfile/2015/01/13/20150113091548267.pdf , Section 4.1.2.

modules/luci-base/po/zh-cn/base.po
yangfl: "unspecif**ied**" _itself_ is a _state_, not "unspecif**y**"

yangfl:
@yousong OK, I will not push that change this time, but I still hold my opinion on that. Other changes seem fine to both of us.

yousong:
Thank you.

Issue #869, opened 2016-12-09
Split modules into applications (ease transition)


cshoredaniel:
Hi all,

I'm feeling better and am thinking that I'd like to be useful by starting to take steps to make a smooth transition from current LuCI (Lua+HTML+a little javascript) to LuCI2 (ubus rpc + HTML5 + lots of javascript).

I think the first step is to decompose some of the major components like mod-admin-full into applications so that individual pieces can be replaced more easily, instead of having a situation where the all of mod-admin-full has to be completely rebuilt from the ground up for the replacement to be done.

So my plan is to identify common components and moving them into 'library' packages, and to move admin-full pages into individual applications under the applications directory.

As I move components out of mod-admin-full, I intend to make mod-admin-full depend on the new applications, so that from the user perspective nothing changes for now.

Any thoughts?

jow-:
@cshoredaniel - sounds good to me! I've noticed that there's a bit of confusion around "luci2" so I'd like to take the opportunity to clarify what "luci2" is going to be.

There's two repositories:
1) the "luci2/ui.git" repository (https://git.lede-project.org/?p=project/luci2/ui.git) which was a first prototype to build client side luci using a custom framework
2) the "luci-ng" repository on Github (https://github.com/jow-/luci-ng/)

To not fall into the custom code/framework trap again like we did with LuCI, I'd like to consolidate future work in jow-/luci-ng based on Angular.js to base our work on a well known, maintained foundation.

I will add a deprecation notice to the other repository to avoid future misunderstandings.

cshoredaniel:
@jow- Ok, thank you for the clarification; I didn't know about that.  Angular.js sounds like a good plan to me.  I've taken a quick pick at your luci-ng repo and it looks like like a plan I have ought to be doable (which is to have a luci-ng compatibility shim that allows using luci-ng alongside current luci with the same basic look, although the feel will be probably be different because of more responsive ui from luci-ng); basically my goal, long term, is to make it possible to shift individual apps (including decomposed mod-admin-full) without a lot of fuss, until enough has been shifted to deprecate luci and put it in maintenance mode and stop accepting new apps for the Lua stuff.

cshoredaniel:
Note, that I won't push the work in this regard until after the next release, unless that is being pushed significantly out due to possible merger.

cshoredaniel:
For personal reason no longer participating in OpenWrt/LEDE.

cshoredaniel:
Reopening since it's back on my radar.

[...]

PR #2032, opened 2018-08-02
modules: Make luci-base sufficient to use luci apps

cshoredaniel:
Per the discussion in https://github.com/openwrt/luci/issues/869, make
luci-base sufficient to login, logout, and review and apply or revert
uci changes.  This allows most luci-app-xxx to work without having
luci-mod-admin-full installed.

It has been tested with some apps and not luci-mod-admin-full, as well
as with luci-mod-admin-full (to make sure the usual case doesn't break).

Instead of creating a new module namespace (e.g. 'Base') we reduce the
opportunities for breakage by having luci-base take over the 'shell' of
the 'Administration' (admin/....) namespace.

Since admin is assumed by all current building LuCI components (including
Freifunk), this doesn't introduce the 'Administration' tab into any
situation where it would not already be present (but includes it where it
was before).


cshoredaniel:
Discovered some issues, will reopen after more extensive testing and fixing.

themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm
```diff
@@ -196,7 +196,9 @@
 				<div class="alert-message warning">
 					<h4><%:No password set!%></h4>
 					<p><%:There is no password set on this router. Please configure a root password to protect the web interface and enable SSH.%></p>
-					<div class="right"><a class="btn" href="<%=url("admin/system/admin")%>"><%:Go to password configuration...%></a></div>
+					<% if disp.lookup("amdin/system/admin") then %>
```
kyle30312: typo: `amdin`

PR #2035, opened 2018-08-03
luci-app-vnstat: Add fields for new database options

cshoredaniel:
The vnstat package supports some new options for restoring (on start)
databases from other locations, and for accessing the database through
symlinks at an alternate filesystem location.  This exposes those options
in the configuration tab.

### DIFF
--- a/applications/luci-app-vnstat/luasrc/model/cbi/vnstat.lua
+++ b/applications/luci-app-vnstat/luasrc/model/cbi/vnstat.lua
@@ -46,6 +46,12 @@ local s = m:section(TypedSection, "vnstat")
 s.anonymous = true
 s.addremove = false
 
+backup_dir = s:option(Value, "backup_dir", translate("Backup Dir"), translate("Directory from which to retrieve data backups"))
+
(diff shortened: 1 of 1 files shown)


dibdot:
I don't know this app, maybe it makes sense to add some sort of validation to these fields!?

cshoredaniel:
I'll have to check if there is validation of paths...

jow-:
There is a server side file / directory validator, but that requires the path to be present already. Not sure if that applies here.

Downside is that there's no client side realtime verification as this would requite checking the server side filesystem with each keystroke.

dibdot:
understood - sorry for the noise.

cshoredaniel:
@dibdot It's not noise!  I have to think about this since I'm not sure the existing code creates the directories, which means for a fully LuCI experience that make sure the paths exist and having the ability create them if they don't is an issue (otherwise to create the dirs you need to SSH or RTTY or such anyway).

cshoredaniel:
Closing until I rework to deal with non-existing directories and/or ensure reasonable directory chosen.  (Or should for this kind of thing should we insist on /etc/uci-upload with a manual (cli) symlink if user needs elsewhere)?

PR #2036, opened 2018-08-04
Make luci-base standalone and separate admin-full

cshoredaniel:
This PR contains two commits which are based on the discussion in https://github.com/openwrt/luci/issues/869.

luci-base is now sufficient to login and out and to review and apply or revert uci changes (made within LuCI).  It also has gained some additional code that are common among network and status.

The second comment splits luci-mod-admin-full into luci-mod-status and luci-mod-system and luci-mod-network.  For now luci-mod-admin-full is kept as a meta-package for upgrades, but I propose that the luci collections be modified to depend on the new modules and drop mod-admin-full entirely.

One thing that is in mod-admin-full that probably should be split into a separate app because
a) there is no good place for it, and
b) it's not necessarily good to have on by default (exposes every file to the web UI).
is the filebrowser thingy.


jow-:
Looking into this now....

cshoredaniel:
@jow- Thanks!  I was actually figuring it be at least a few more days because of 18.06.1

cshoredaniel:
Please ignore the next push until I've tested it; I'm just verifying something with GitHub.

cshoredaniel:
Still doing more complete testing though...I'm not sure I completely trust it yet, so consider it an RFC for now....

jow-:
I ave been flashing a build with this and so far it looks quite well.

cshoredaniel:
Great!  Thanks for looking at this.

cshoredaniel:
Rebased, now to retest...

jow-:
One little detail is that I'd like to redirect the default view to the first existing page and only then fallback to the placeholder page view. Not sure if I'll get around to it before holidays.

jow-:
Yep, correct. Was thinking about extending `dispatcher.lua` to simply remember the first node it registers, then add a new `firstnode()` target and simply use that for `/admin`

cshoredaniel:
Let me know if you don't get to it before your holidays and if I get time I'll try to whip something up (got a few other things earlier on my list right now).

jow-:
@cshoredaniel - it took me a while but I finally merged your changes along with my firstnode() logic on top to autoselect the first suitable page.

hnyman:
Looks nice at the first glance.

I fixed the fallout to i18n scripts with 90e749c6
(scan for base.pot needs also the three new modules, as luci-mod-admin-full is now empty)

jow-:
@hnyman - thanks a lot for fixing the fallout I left :)

cshoredaniel:
@jow @hnyman  Thanks for all your help...be back more regularly in a few weeks, all going well.

PR #2039, opened 2018-08-07
luci-base: add freq to channel calculation to tools

PolynomialDivision:
Translate iw function ieee80211_frequency_to_channel to lua
and add it to the tools section.

Signed-off-by: Nick Hainke <vincent@systemli.org>

### DIFF
--- a/modules/luci-base/luasrc/tools/status.lua
+++ b/modules/luci-base/luasrc/tools/status.lua
@@ -303,3 +303,19 @@ function switch_status(devs)
 	end
 	return switches
 end
+
+function ieee80211_frequency_to_channel(freq)
+	local channel = 0
+	if freq == 2484 then
+		channel = 14
+	elseif freq < 2484 then
+		channel = (freq - 2407) / 5
+	elseif freq >= 4910 and freq <= 4980 then
+		channel = (freq - 4000) / 5
+	elseif freq <= 45000 then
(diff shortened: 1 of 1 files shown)


dibdot:
@PolynomialDivision just a short formal feedback:

- Semi-colons in Lua are generally only required when writing multiple statements on a line
- I would suggest to define a local variable "channel" with default 0, change that variable in the function and finally return "channel" once at the end

PolynomialDivision:
@dibdot Thanks for your review! :)

If I add this function directly in my .../tools/status.lua script, it doesn't work.
I tested as local function in one of my lua scripts.
I think the dispatcher has to recognize this function somehow?

dibdot:
Just a rough guess: did you reset your luci caches (rm -rf /tmp/luci*)?

PolynomialDivision:
rm -rf /tmp/luci-*

is the solution. :)

Now you only have to accept my pull request. ^^

dibdot:
@PolynomialDivision regarding merging: I'm still a LuCI rookie and I'm not sure if we need such function at all. There is already a generic "getiwinfo" function available, therefore it might be better to re-use this one in your app to get the channel information!?

PolynomialDivision:
Ok.
I added this to my own app as tools. ;)
So I will close. :)

dibdot:
Fair enough - thanks for the info.

PR #2040, opened 2018-08-07
luci-app: update Chinese translation

yangfl:


xinfli:
I think "接口" is a better translation than "介面" for "interface".

yangfl:
@xinfli "介面" is used in Taiwan. You can confirm it in https://zh.wikipedia.org/wiki/%E7%B6%B2%E8%B7%AF%E4%BB%8B%E9%9D%A2

xinfli:
@yangfl Thanks, then I think it's better to use "接口" in zh-cn and use "介面" in zh-tw :-)

PR #2041, opened 2018-08-07
Correct grammar in apply_widget.htm

farmergreg:
This patch corrects "to get" to "to be" in apply_widget.htm
This shell command was used to find and make the change in
all impacted files:

find . -type f -exec sed -i 's/Waiting for configuration to get applied/Waiting for configuration to be applied/g' {} +

Signed-off-by: Gregory L. Dietsche <gregory.dietsche@cuw.edu>


jow-:
Merged - thank you!

PR #2042, opened 2018-08-07
luci-mod-admin-full: improve applyreboot page

Ansuel:
This improve applyreboot page and fix problem with luci-nginx that doesn't refresh the page when the router reboot.

### DIFF
--- a/modules/luci-mod-admin-full/luasrc/view/admin_system/applyreboot.htm
+++ b/modules/luci-mod-admin-full/luasrc/view/admin_system/applyreboot.htm
@@ -6,35 +6,46 @@
 
 <html>
 	<head>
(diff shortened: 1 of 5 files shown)


jow-:
Do we really need those inline styles? You could move them to the material theme CSS and I'll take care of adding appropriate styles to the other themes.

Ansuel:
2 class with that styles should be good right? 
applyreboot-container
applyreboot-section

Ansuel:
@jow- inline style removed... tested with all the other themes

jow-:
Almost fine :) Can you please replace `fieldset` with `div` ?

Ansuel:
done @jow-

jow-:
merged, thanks!

Issue #2048, opened 2018-08-08
Incorrect warning on adding SSH key


bvandergiessen:
LuCI openwrt-18.06 branch (git-18.210.69179-6df9a57).

On the page _System... Administration_, after you add an SSH key in the SSH-Key text box and click "Save and apply", the warning _there are no changes to apply_ appears. Nevertheless, the new key is correctly added to /etc/dropbear/authorized_keys.

bvandergiessen:
Fixed by #2049

PR #2049, opened 2018-08-08
luci-base: apply fix for admin.lua

dibdot:
* fix for #2048


dibdot:
@jow OK to merge?

jow-:
Hm, no not sure. I think it will cause SSH server changes to bypass the rollback mechanism. Will close this for now and come up with an alternative solution later.

jow-:
Seems it does the job indeed - will merge it.

jow-:
Merged with 32ee1873a - thanks!

dibdot:
Already forget about this one - thanks for merging!

PR #2061, opened 2018-08-13
luci-mod-admin-full: improve connection page lookup loop

Ansuel:
Currently the lookup loop is very inefficient. This improve the lookup loop by moving the entities to a separate array and adds them only one time. Every entities is checked 3 times, after that it will be moved to dns cache as a raw addres (not resolved). This also introduce the Type column to identify common packet using the source port.

### DIFF
--- a/modules/luci-mod-status/luasrc/view/admin_status/connections.htm
+++ b/modules/luci-mod-status/luasrc/view/admin_status/connections.htm
@@ -131,15 +131,21 @@
 				}
 
 				label_scale.innerHTML = String.format('<%:(%d minute window, %d second interval)%>', data_wanted / 60, 3);
+				
+				var recheck_lookup_queue = {};
 
 				/* render datasets, start update interval */
 				XHR.poll(3, '<%=build_url("admin/status/realtime/connections_status")%>', null,
 					function(x, json)
 					{
-						var rows = [];
+
+						if (!json.connections)
+							return;
+
 						var conn = json.connections;
 
 						var lookup_queue = [ ];
+						var rows = [];
 
 						conn.sort(function(a, b) {
 							return b.bytes - a.bytes;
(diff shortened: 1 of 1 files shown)


Ansuel:
@jow- any chance for this?

Ansuel:
@jow- can you check this?

Ansuel:
@jow- news?

feckert:
Please rebase against the master branch

Ansuel:
@feckert done but i think  this won't be part of default code as it's too shitty ahahhaha

feckert:
please fix also the intention and new line at end of file

Ansuel:
@feckert intention? if you mean indentation where is broken?

feckert:
Sorry for the confusion I meant of course indentation See for example the following lines.
https://github.com/openwrt/luci/blob/1db213e5d7ce89de319f020a6bc5509da42f9157/modules/luci-mod-status/luasrc/view/admin_status/connections.htm#L183-L185

https://github.com/openwrt/luci/blob/1db213e5d7ce89de319f020a6bc5509da42f9157/modules/luci-mod-status/luasrc/view/admin_status/connections.htm#L191-L193

https://github.com/openwrt/luci/blob/1db213e5d7ce89de319f020a6bc5509da42f9157/modules/luci-mod-status/luasrc/view/admin_status/connections.htm#L200-L205

Ansuel:
@feckert should be good now

modules/luci-base/luasrc/sys.lua
```diff
@@ -358,6 +358,32 @@ function net.conntrack(callback)
 					if entry[key] == nil then
 						entry[key] = val
 					end
+					if entry["ptype"] == nil then
```
jow-: This list is arbitrary and incomplete. I think it will cause confusion in the long run... ("Why is protocol X not show?")

Ansuel: @jow- so drop the entire port type part right?

jow-: Yes. If we want to introduce something like this it should be some kind of global configuration file similar to /etc/protocols

Ansuel: yes the first idea was to use the /etc/protocols file but also that file is incomplete... will take a look about that.

Ansuel:
news on this? @feckert

modules/luci-mod-status/luasrc/view/admin_status/connections.htm
```diff
@@ -131,15 +131,21 @@
 				}
 
 				label_scale.innerHTML = String.format('<%:(%d minute window, %d second interval)%>', data_wanted / 60, 3);
+
```
feckert: Please fix also this indentation

feckert:
Please fix the last indentation.
And it works as expected so it is LGTM
@jow- any hint from your side if no then i will merge this

Ansuel:
@feckert good now?

feckert:
Thank you for your contribution and for the long waiting time.

PR #2064, opened 2018-08-14
luci-base: avoid input field creation in value.htm

### LINKED ISSUE #2063: Edit field for WPA key.
Hi I dont know how this looks on the screen, but it is showing up to my screen reader that there is 2 edit fields to type in the WPA Key when setting up Wireless Security. 
I set Encryption to WPA2-PSK press tab, set Cipher
Force CCMP (AES) than tab again to the key field and typed in my key. 
After hitting Save & Apply  my wi-fi did not come up. When looking in the log I got:
Tue Aug 14 06:41:28 2018 daemon.notice netifd: radio0 (6816): command failed: Not supported (-95)
Tue Aug 14 06:41:28 2018 daemon.notice netifd: radio0 (6816): Interface 0 setup failed: INVALID_WPA_PSK
Tue Aug 14 06:41:28 2018 daemon.notice netifd: radio0 (6816): WARNING: Variable 'interfaces' does not exist or is not an array/object

Then when I looked at my key the edit field seemed to be empty.
If you tab again a seckend edit field shows up and then you can put in your WPA key and it will worke rite.

dibdot:
- avoid creation of needless password input field, should fix #2063

### DIFF
--- a/modules/luci-base/luasrc/view/cbi/value.htm
+++ b/modules/luci-base/luasrc/view/cbi/value.htm
@@ -1,9 +1,4 @@
 <%+cbi/valueheader%>
-	<%- if self.password then -%>
-		<input type="password" style="position:absolute; left:-1000px"<%=
-			attr("name", "password." .. cbid)
-		%> />
-	<%- end -%>
 	<input data-update="change"<%=
 		attr("id", cbid) ..
 		attr("name", cbid) ..


dibdot:
@jow not sure if this is a legit fix.

jow-:
This is needed to prevent browsers from filling wrong autocomplete credentials.

dibdot:
@jow- thanks for explanation. I thought that the already used autocomplete "new-password" is enough ... but maybe that doesn't work for all browsers. Just one note: the added aria tag didn't change the slightly wrong tab order for ordinary users. Not really important, but maybe a "display:none" is a better approach here.

jow-:
The "display:none" does not work anymore nowadays, hence the offscreen rendering.

dibdot:
It was only meant to fix the tab order for ordinary users, too - e.g.:

```
<%+cbi/valueheader%>
        <%- if self.password then -%>
                <input type="password" style="display:none" aria-hidden="true"<%=
                        attr("name", "password." .. cbid)
                %> />
        <%- end -%>
[...]
```
This works quite fine, quickly tested with firefox & chrome ...

jow-:
The point is that it renders the workaround useless as certain browser autocomplete versions are not fooled by "display:none" elements

jow-:
A simpler fix is to add `tabindex="-1"` to the dummy password field.

dibdot:
thanks for the coaching - still learning ... ;-)

PR #2067, opened 2018-08-15
luci-mod-admin-full: allow forced upgrade

psyborg55:
Enable checkbox to force flashing even if image header check fails. Unset by default.

Signed-off-by: Tomislav Požega <pozega.tomislav@gmail.com>

### DIFF
--- a/modules/luci-mod-admin-full/luasrc/controller/admin/system.lua
+++ b/modules/luci-mod-admin-full/luasrc/controller/admin/system.lua
@@ -271,7 +271,7 @@ function action_sysupgrade()
 	--
 	local step = tonumber(http.formvalue("step") or 1)
 	if step == 1 then
-		if image_supported(image_tmp) then
+		if image_supported(image_tmp) or http.formvalue("force") then
 			luci.template.render("admin_system/upgrade", {
 				checksum = image_checksum(image_tmp),
 				sha256ch = image_sha256_checksum(image_tmp),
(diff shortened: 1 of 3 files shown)


hnyman:
Thanks. I have sometimes wondered if that option should be supported in LuCI

I think that there should be an explanation text next to the option.  Something in style of 
"Force flashing even if the image format check fails. Use only if you are sure that the firmware is correct and meant for your device."

Even better would be to expose the option only after the image check first fails.

dibdot:
Just curious, what's the use case to expose/promote such possible dangerous feature in LuCI? 
If we really need this, I second the idea to show this option only if the first flash fails.

hnyman:
There are some devices, where flashing back to the OEM version requires this. 
E.g. the popular mvebu wrt1900acx series.  Similarly, switching betwen ar71xx and ath79 builds requires this.

Ps. It is not about "first flash fails", but that the image correctness check fails.

dibdot:
@hnyman thanks for the clarification

psyborg55:
I think the feature is self-explanatory and someone that is upgrading his/her device should know what it does. I've added this because I wanted to bypass header magic check on my device when switching between different images, and opened here because I thought there may be others interested in it too.

If somebody else can come up with additional checks so that it's visible after first flash fail this PR allows edit from maintainers so you can always push another commit.

dibdot:
superseded by #2075

PR #2071, opened 2018-08-16
luci-app-samba4: add macOS and Apple Timemachine support

Andy2244:
* add macOS compatibility option
* adds Apple Timemachine support
* dynamically disable invalid options

depends on openwrt/packages#6772


Andy2244:
@dibdot This is the associated PR for openwrt/packages#6772 and is needed for the new macOS options.

PR #2074, opened 2018-08-17
luci-app: update Chinese translation

yangfl:


jow-:
Merged with 3ed5dc0f6 - thanks!

PR #2075, opened 2018-08-17
luci-mod-admin-full: allow forced upgrade (revised)

dibdot:
* allow 'forced' firmware upgrades, even if the image format check fails. Useful where flashing back to the OEM versions or switching between ar71xx and ath79 builds
* option is only visible after first/failed image check
* added warning info (see screenshots below)

### DIFF
--- a/modules/luci-mod-admin-full/luasrc/controller/admin/system.lua
+++ b/modules/luci-mod-admin-full/luasrc/controller/admin/system.lua
@@ -269,15 +269,17 @@ function action_sysupgrade()
 	--
 	-- Initiate firmware flash
 	--
-	local step = tonumber(http.formvalue("step") or 1)
+	local step = tonumber(http.formvalue("step")) or 1
 	if step == 1 then
-		if image_supported(image_tmp) then
+		local force = http.formvalue("force")
+		if image_supported(image_tmp) or force then
 			luci.template.render("admin_system/upgrade", {
 				checksum = image_checksum(image_tmp),
 				sha256ch = image_sha256_checksum(image_tmp),
 				storage  = storage_size(),
 				size     = (fs.stat(image_tmp, "size") or 0),
-				keep     = (not not http.formvalue("keep"))
+				keep     = (not not http.formvalue("keep")),
+				force    = (not not http.formvalue("force"))
 			})
 		else
 			fs.unlink(image_tmp)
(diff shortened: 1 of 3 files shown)


dibdot:
@psyborg55 @hnyman @jow- 

Here is a slightly modified approach:
1. Start flashing a "wrong" image:
flash_01

2. New "force" option will be visible:
flash_02

3. Upgrade page with last warning:
flash_03

jow-:
LGTM, feel free to merge.

hnyman:
Interesting, the force from LuCI fails for me when I try to force-sysupgarde from ath79 build master-r7878-d1ea8ac3b4-20180819 to basic ar71xx build with WNDR3800. 

The dialog gets shown ok, and after force checkbox is used, the sysupgrade progresses to the hash verification screen. After pushing the button there, the flashing wait routine starts, but after a few seconds LuCI returns to the main overview screen and no flash or reboot happens.

System log shows just this:
```
Mon Aug 20 20:47:48 2018 authpriv.info dropbear[1408]: Early exit: Terminated by signal
Mon Aug 20 20:47:53 2018 authpriv.warn dropbear[10910]: Failed loading /etc/dropbear/dropbear_ecdsa_host_key
Mon Aug 20 20:47:53 2018 authpriv.info dropbear[10910]: Not backgrounding
```

Flashing the same image from console works ok:
```
root@OpenWrt:~# sysupgrade -F  /tmp/WNDR3800-master-r7890-40eb9bda44-20180820-2042-sqfs-sysupgrade.bin
Image metadata not found
Image check 'fwtool_check_image' failed but --force given - will update anyway!
Saving config files...
Commencing upgrade. Closing all shell sessions.
```

dibdot:
@hnyman I've never tested the flash process itself - I'll take a look and come back to you.

modules/luci-mod-admin-full/luasrc/controller/admin/system.lua
```diff
@@ -269,15 +269,17 @@ function action_sysupgrade()
...
+		local force = http.formvalue("force")
+		if image_supported(image_tmp) or force then
 			luci.template.render("admin_system/upgrade", {
 				checksum = image_checksum(image_tmp),
 				sha256ch = image_sha256_checksum(image_tmp),
 				storage  = storage_size(),
 				size     = (fs.stat(image_tmp, "size") or 0),
-				keep     = (not not http.formvalue("keep"))
+				keep     = (not not http.formvalue("keep")),
+				force    = (not not http.formvalue("force"))
```
psyborg55: @hnyman try remove force from here and flash again

psyborg55:
that way youre re-inverting already inverted force value. force value is already passed and is not needed in template.render which also keep the code smaller. and i think it can be removed from this line too: 			addr  = (#keep > 0) and (#force > 0) and "192.168.1.1" or nil

modules/luci-mod-admin-full/luasrc/controller/admin/system.lua
```diff
@@ -287,17 +289,19 @@ function action_sysupgrade()
...
 		local keep = (http.formvalue("keep") == "1") and "" or "-n"
+		local force = (http.formvalue("force") == "1") and "" or "-F"
 		luci.template.render("admin_system/applyreboot", {
 			title = luci.i18n.translate("Flashing..."),
 			msg   = luci.i18n.translate("The system is flashing now.<br /> DO NOT POWER OFF THE DEVICE!<br /> Wait a few minutes before you try to reconnect. It might be necessary to renew the address of your computer to reach the device again, depending on your settings."),
-			addr  = (#keep > 0) and "192.168.1.1" or nil
+			addr  = (#keep > 0) and (#force > 0) and "192.168.1.1" or nil
 		})
-		fork_exec("sleep 1; killall dropbear uhttpd; sleep 1; /sbin/sysupgrade %s %q" %{ keep, image_tmp })
+		fork_exec("sleep 1; killall dropbear uhttpd; sleep 1; /sbin/sysupgrade %s %s %q" %{ keep, force, image_tmp })
```
psyborg55: why msleep doesn't work here?

PR #2076, opened 2018-08-17
luci-mod-admin: make wifi encryption translatable

yangfl:


psyborg55:
sadly, your commit info is not translatable

PR #2079, opened 2018-08-18
luci-base: simplify luci base link

Ansuel:
@jow- 

### DIFF
--- a/modules/luci-base/root/www/index.html
+++ b/modules/luci-base/root/www/index.html
@@ -3,9 +3,9 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
 <meta http-equiv="Cache-Control" content="no-cache" />
-<meta http-equiv="refresh" content="0; URL=/cgi-bin/luci" />
+<meta http-equiv="refresh" content="0; URL=/luci" />
 </head>
 <body style="background-color: white">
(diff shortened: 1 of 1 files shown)


cshoredaniel:
@Ansuel Have you verified this works on a 'normal' OpenWrt installation?  This looks to me like it could break if you aren't using the uhttpd lua plugin...

Ansuel:
Lua plugin is installed and enabled by default in uhttpd configuration

cshoredaniel:
Oh! I didn't realize that had changed (I've been away a *while*).  In any event have you verified on 'normal' OpenWrt?  (IIRC you're doing the nginx thing).

Ansuel:
Actually on the wiki there is a guide to enable embedded lua... 

One part of it got merged (the uhttpd config) but the index changes never got merged... In the past I always used Luci with this configuration

cshoredaniel:
@Ansuel I knew about the guide, I didn't know it was half-meged though.  Seems a good plan to fully merge.  Knowing you've actually used it this way (on stock OpenWrt except these changes) is helpful.

In that case LGTM.

jow-:
NACK. Embedded Lua is *not* installed by default and your change would completely break normal LuCI for the majority of users.

Ansuel:
@jow- can i provide a specific index file for just luci-nginx?

Issue #1566, opened 2018-01-19
luci-app-minidlna asked to remove the colon


user7887:
Most OpenWrt applications do not have a colon in the input string header value.
I've created a template for translation into Russian of applications, many applications use the same words and phrases. The colon interfere with proper operation of the program transfer.
![bas](https://user-images.githubusercontent.com/28679841/35171346-962f98dc-fd74-11e7-91a5-593318bccd8c.png)
Ask you to remove if possible, thanks!

feckert:
@user7887 i have seen no maintainer for this app in the makefile and git history.
If you want to delete the colon then open a PR with the changes.
Changes:
update source and update translation msgid`s

PR #2080, opened 2018-08-18
luci-app-firewall: Colons removed from input headers

ameer1234567890:
Most OpenWrt applications do not have a colon in input headers.
This has been explained in #1566 as well.
This commit removes the said colons.


ameer1234567890:
#2082 supersedes this PR

PR #2097, opened 2018-08-19
luci-app-privoxy: Colons removed from input headers

ameer1234567890:
Most OpenWrt applications do not have a colon in input headers.
This has been explained in #1566 as well.
This commit removes the said colons.


dibdot:
@hnyman from my understanding these ".po-only" changes are pretty useless, cause they will be updated during the next sync translations round - correct?

jow-:
Correct, the change needs to be happen synchronously in both the affected template / Lua files, the pot and po files.

PR #2102, opened 2018-08-20
luci-mod-admin-full: wifi.lua: auto ft_psk_generate_local when using PSK

BigNerd95:
Easier config of IEEE802.11r Fast Transition when using PSK.

The local generation of PMK is now auto enabled by OpenWRT when a
PSK is used, this patch adapt the GUI to the new script.

Signed-off-by: Lorenzo Santina <lorenzo.santina@edu.unito.it>

### DIFF
--- a/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
+++ b/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
@@ -10,7 +10,7 @@ local fs = require "nixio.fs"
 local acct_port, acct_secret, acct_server, anonymous_identity, ant1, ant2,
 	auth, auth_port, auth_secret, auth_server, bssid, cacert, cacert2,
 	cc, ch, cipher, clientcert, clientcert2, ea, eaptype, en, encr,
-	ft_protocol, ft_psk_generate_local, hidden, htmode, identity,
+	ft_protocol, hidden, htmode, identity,
 	ieee80211r, ieee80211w, ifname, isolate, key_retries,
 	legacyrates, max_timeout, meshfwd, meshid, ml, mobility_domain, mode,
 	mp, nasid, network, password, pmk_r1_push, privkey, privkey2, privkeypwd,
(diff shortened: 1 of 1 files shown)


modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
```diff
@@ -818,21 +829,12 @@ if hwtype == "mac80211" or hwtype == "prism2" then
...
-
 	mobility_domain = s:taboption("encryption", Value, "mobility_domain",
-			translate("Mobility Domain"),
-			translate("4-character hexadecimal ID"))
+		translate("Mobility Domain"),
+		translate("4-character hexadecimal ID" ..
+			"<br />(Default generated from SSID)"))
 	mobility_domain:depends({ieee80211r="1"})
-	mobility_domain.placeholder = "4f57"
+	mobility_domain.placeholder = ut.exec("echo " .. ut.shellquote(wnet:get("ssid")) .. " | md5sum | head -c 4")
```
jow-: Note that this is only a placeholder; is it the same algorithm used by OpenWrt?

BigNerd95: Yes
https://github.com/openwrt/openwrt/blob/master/package/network/services/hostapd/files/hostapd.sh#L427

hnyman:
Why do you propose to remove the whole option ft_psk_generate_local  ?

The option is still quite valid, although the default behaviour in hostapd was changed (rather crudely by https://github.com/openwrt/openwrt/commit/3cc56a5534b8b49a7e9ba57edf9878ec32bdd27a )  to ignore all existing detailed configs by default. Old default was 0, so most 802.11r configs did not have that value set at all. Right now the new default is 1, so a missing value causes the old existing detail config (r0kh, r1kh etc.) to be ignored...

If ft_psk_generate_local is set to 0 on the config file, then the detailed settings are still available and should be configurable also in LuCI.

It is also quite possible that somebody wants to specifically configure the detail options even in PSK based configs, so trying to determine that solely by wpa/wpa2 is wrong.

I know that the current default value in LuCI is wrong (0), but deleting the whole option does not seem right.

BigNerd95:
> Why do you propose to remove the whole option ft_psk_generate_local ?

I removed that option because I think it can be confusing for users when they use a PSK.

> If ft_psk_generate_local is set to 0 on the config file, then the detailed settings are still available and should be configurable also in LuCI.

If a user set ft_psk_generate_local to 0 in the config file when using a PSK, then he should complete the configuration editing the config file.

> I know that the current default value in LuCI is wrong (0), but deleting the whole ft_psk_generate_local option does not seem right.

Any advice?

Thanks for your fast review :-)

hnyman:
> > Why do you propose to remove the whole option ft_psk_generate_local ?
>
> I removed that option because I think it can be confusing for users when they use a PSK.

Let's not remove already existing advanced options just because they may confuse some users. 

I feel that 
* the "local generation" option should be left there, but its default value should be corrected to match the changed default on Openwrt side.
* the 802.11r options could be arranged to logical order, having the local generation first, explanation then, and then the advanced 802.11r options.
* the "local generation" explanation text should be added, but the correct text (not the proposed one). Something like "with WPA(2)-PSK modes, the PMK is generated automatically by default. If PMK is locally generated, the 802.11r detailed options (keyholders etc.) are ignored. If PMK local generation is disabled, the detailed 802.11r options are needed". That is maybe not the optimal text, yet, but gives the core...
* the 802.11r advanced options should not depend on WPA/WPA2, as those options can be used also with PSK modes.
* the 802.11r advanced options should also not depend on "not local generation", as then the default local generation option would cause LuCI to delete already set values from user's config file (as LuCI would interpret the missing value (old default was 0) as the new default 1 and would think that the detail options are not needed).

PR #2104, opened 2018-08-21
luci-app-nut: Add support for NUT configuration

cshoredaniel:
Adds app to configure NUT (Network UPS Tools), including
nut-server, nut-monitor, and nut-cgi.


cshoredaniel:
Found a couple of things to fix....

cshoredaniel:
I've got this fixed, but RL is getting in the way for a week or two so will reopen when I can respond if there are problem when I post the final edition.

PR #2119, opened 2018-08-30
luci-mod-admin-full: add hint to overview page

feckert:
This PR will fix an issue with dynamic interfaces
See commits


jow-:
What does it fix exactly?

feckert:
If you have a dynamic interface.
For example a wwan interface which set up a wwan_4 interface, then you can not stop this because the button to deconfigure the connection is disabled.

jow-:
This was actually intentional as some more complex protocols will end up in an undefined state if you tear down virtual sub protocols

feckert:
I mean the base interface could also not get deconfigured because the button is disabled.
wwan is the base interface
wwan_4 is the dynamic interface

feckert:
Could we at least merge the following commits?
https://github.com/openwrt/luci/pull/2119/commits/c018b563c35d83b83181034ab40c128b0bf35b85
https://github.com/openwrt/luci/pull/2119/commits/95ff016cab3126e64d301fd1e4dc5029e7f9ea8f
https://github.com/openwrt/luci/pull/2119/commits/c54fbc49c74449dc337a43f549afc7260c76eb05
https://github.com/openwrt/luci/pull/2119/commits/4d1bfe12e13e7f0de1de8f25eb11041c46861767

Because we change the option **auto** implicit on pressing the button connect/disconnect i think we should inform the user that this interface is not auto booted on start. So I think this is also a good commit to get merged

https://github.com/openwrt/luci/pull/2119/commits/2b63078594f75daad4159bee47ba13ec55881c9e

If this has the change to get merged I will rebase my changes.

feckert:
@jow- I have rebased my changes and update my PR.
So this PR will only add a hint to the overview page because the auto option is implicit changed.
I think we should inform the user that we have changed the behaviour

feckert:
Rebase the changes against the current master.
@yousong @hnyman any hints from your side?
I would like to emphasize that this is not a fix but an important hint for the user.
Until now it was only implicitly clear when the interface is down that this is not started on booting by the netifd. Sine openwrt-18.06 the uci configuration **auto** is toggled every time on pressing connect and disconnect, in the background.

hnyman:
Haven't tested it, but the basic idea off adding the info seems ok to me.

feckert:
@hnyman thanks for feedback.
I will merge this now.

PR #2121, opened 2018-08-31
luci-mod-admin-full: performance fix

jokdarius:
*	uci:get replaced with direct access of value (functions has 3x better performance in that way)

Signed-off-by: Darius Joksas <jok.darius@gmail.com>


jow-:
Nice catch, thanks!

PR #2122, opened 2018-08-31
Backup: add firmware backup for mtdblock devices

rosysong:
This commit add the feature that allow user to backup the image of the flash. 

backup

### DIFF
--- a/modules/luci-base/luasrc/sys.lua
+++ b/modules/luci-base/luasrc/sys.lua
@@ -70,6 +70,24 @@ function mounts()
 	return data
 end
 
+function mtds()
+	local data = {}
+
+	if fs.access("/proc/mtd") then
+		for l in io.lines("/proc/mtd") do
+			local d, s, e, n = l:match('^([^%s]+)%s+([^%s]+)%s+([^%s]+)%s+"([^%s]+)"')
+			if s and n then
+				local d = {}
+				d.size = tonumber(s, 16)
+				d.name = n
+				table.insert(data, d)
+			end
+		end
+	end
+
+	return data
+end
+
 -- containing the whole environment is returned otherwise this function returns
 -- the corresponding string value for the given name or nil if no such variable
 -- exists.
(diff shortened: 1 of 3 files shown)


rosysong:
@hnyman @jow-

hnyman:
How is that ipv6 related change related to the backup functionality that is the topic of this PR?

rosysong:
@hnyman  sorry, will drop it.

rwalli:
Hello!

First I’d like to say that this PR is really essential in luci:
It allows user to make a real backup of the system and to restore the whole device with all packages an config's.
A perfect long outstanding feature!

For me it works perfect, but maybe for new users the interface may be confusing.
It’s not clear how to restore the backups and how to deal with mtd-partitions.
If you want to restore u-boot or art partition it is very likely that you already have serial adapter and soldering iron lying around, which is not the everyday case.

I would suggest to remove other partitions than “firmware” or make it visible via “Expert Options”  -checkbox in the Configuration-section (if you lockup the bootloader  it can’t be restored within luci).

Further more it could be handy if the downloaded images have “Model” and “Hostname” in their filenames. 

Maybe it could look like this:

(“Keep settings” is a little bit confusing when restoring a backup-firmware but did no harm in my tests.)

screenshot 2018-09-02 12 16 49

rosysong:
@rwalli  Having “Model” and “Hostname” in target file is a good suggestions. 

In addition, I think we really need the backup options for **u-boot** and **art** partitions, particularly for the developers who work on them. Though these features is not everyday case but it is useful and convenient once we need. Moreover, this can meet the different needs of people who use OpenWrt and luci. So I would prefer to **make it visible via “Expert Options” -checkbox** in the Configuration-section.

Thank you for your suggestions, @rwalli ! :-)

hnyman:
The help texts should indicate more clearly that this is low-level technical functionality accessing flash partition directly, and is not meant for everyday config backup/restore. I am pretty sure that if the current version is merged, there will be people making config backups this way and then wondering how to restore them.

Should it be in a separate section??? (instead of backup)
Maybe after backup and restore. 

Terminology may be a bit confusing, as both this new text and the firmware flash text talk about "image". So, people may try using this mtdblock image for firmware flash...

rosysong:
@hnyman @rwalli  Hi,  how about this ?

backupv3

rosysong:
@hnyman This is the latest modification. I added a caution sentence saying that "this feature is for professionals!" at the description. Feel free to correct me if you think it is not properly. Thanks.

backupv4

rosysong:
@hnyman @rwalli  I did not upload the code of the last screenshot, because I was not sure it seems OK.

Now I update my commit, and also fixed the issue (@hnyman) you mentioned. It won't miss the last 1024 bytes any longer. :-)

hnyman:
Thanks. Looks good enough for me.
Seemed to work ok with my wrt3200acm, so let's merge it.

PR #2126, opened 2018-09-01
luci-proto-ipv6: fix dependency for IPv6

rosysong:


rosysong:
@hnyman  Moved previous commit at here now.

jow-:
Merged, thanks!

Issue #2132, opened 2018-09-05
Dhcp static leases - IPv4 Address field too small


mbo2o:
While editing/browsing static leases on the Network→DHCP and DNS page all the IPv4 address are obscured by the pull-down menu button. While editing one of the address entries the column becomes wider and all entries are readable. The lease Time Field is wider than required. 
![screenshot_2018-08-03_17-22-53](https://user-images.githubusercontent.com/42272096/45066516-1b3f8a00-b102-11e8-96da-e189f20f3ffb.png)

A possible work around would be too increase the size of the table header by adding extra text.
replace

/usr/lib/lua/luci/model/cbi/admin_network/dhcp.lua:308
`ip = s:option(Value, "ip", translate("<abbr title=\"Internet Protocol Version 4\">IPv4</abbr>-Address"))`
with
`ip = s:option(Value, "ip", translate("<abbr title=\"Internet Protocol Version 4\">IPv4</abbr>-Address xxx.xxx.xxx.xxx"))`

dibdot:
I don't see any problem here. We're using dynamic, mostly responsive templates in LuCI - without any hardcoded table structure/measurements. During editing you'll get/see always the full content (same applies to other fields e.g. like MAC address), after that you'll see the full content or a part of it - depends on (mobile) device, screen resolution, browser window/tab size etc. ... in my opinion it works as intended. 

Your proposed "fix" doesn't change that at all.

mbo2o:
It doesn't work as intended, the last octet is always hidden by the button of the combo-box.
Inspecting the element in the browser shows the width is set to 100%
```
.table.cbi-section-table textarea, .table.cbi-section-table select {
    width: 100%;
}
```
Setting it to auto will resolve the issue however I am not sure where to make this change.
```
.table.cbi-section-table textarea, .table.cbi-section-table select {
    width: auto;
}
```

PR #2133, opened 2018-09-05
Increase IPv4 Addresses column width for dhcp static leases

mbo2o:
Refer to 
https://github.com/openwrt/luci/issues/2132


dibdot:
NAK from my side, comments see #2132

PR #2139, opened 2018-09-09
luci-app-firewall v2: Ease creation of multiple day traffic rules

kristrev:
Creating a traffic rule that spans multiple days currently requires
creating three rules, unless the rule only covers complete days. You
need to create one rule for the first day, one rule for the complete
days, and then one rule for the final day. In addition to making
creating multi day-rules a bit of as hassle, performance suffers when
many multi day-rules are present on a system.

The --datestart/--datestop arguments to iptables, which is what
Start/Stop Date maps to, supports specifying a time of day. This PR
extends Start/Stop Date to also accept an optional time. The format of
the fields are now either yyyy-mm-dd or yyyy-mm-ddThh:mm:ss. As before,
Start/Stop Time will override the time specified in Start/Stop Date.

The translations are also updated, with the exception of Russian and
Ukrainian. My knowledge of the cyrillic alphabet is a bit limit, to say
the least.

v1->v2:
* Update commit message
* Remove markup from translation strings.
* Extend existing datatype instead of adding a new type. I have added a
flag "allow_time" to "dateyyyymmdd" that, when set, enables users to
pass a time in addition to date.
* The time-part of "dateyyyymmdd" is optional. I decided to check if
time is passed or not by checking the length of val. By checking length,
we save a potentially redundant call to match.

### DIFF
--- a/applications/luci-app-firewall/po/ru/firewall.po
+++ b/applications/luci-app-firewall/po/ru/firewall.po
@@ -457,14 +457,14 @@ msgstr "Порт источника"
 msgid "Source zone"
 msgstr "Зона источника"
 
-msgid "Start Date (yyyy-mm-dd)"
-msgstr "Дата начала (год-мес-день)"
+msgid "Start Date (yyyy-mm-ddThh:mm:ss)"
+msgstr "Дата начала (гггг-мм-ддTчч:мм:сс)"
 
 msgid "Start Time (hh:mm:ss)"
 msgstr "Время начала (чч:мм:сс)"
 
-msgid "Stop Date (yyyy-mm-dd)"
-msgstr "Дата окончания (год-мес-день)"
+msgid "Stop Date (yyyy-mm-ddThh:mm:ss)"
+msgstr "Дата окончания (гггг-мм-ддTчч:мм:сс)"
 
 msgid "Stop Time (hh:mm:ss)"
 msgstr "Время окончания (чч:мм:сс)"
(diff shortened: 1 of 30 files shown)


applications/luci-app-firewall/po/ru/firewall.po
```diff
@@ -457,14 +457,14 @@ msgstr "Порт источника"
...
+msgid "Start Date (yyyy-mm-ddThh:mm:ss)"
 msgstr "Дата начала (год-мес-день)"
 
 msgid "Start Time (hh:mm:ss)"
 msgstr "Время начала (чч:мм:сс)"
 
-msgid "Stop Date (yyyy-mm-dd)"
-msgstr "Дата окончания (год-мес-день)"
+msgid "Stop Date (yyyy-mm-ddThh:mm:ss)"
+msgstr "Дата остановки (год-мес-день)"
```
namedun: msgstr "Дата окончания (гггг-мм-ддTчч:мм:сс)"

applications/luci-app-firewall/po/ru/firewall.po
```diff
@@ -457,14 +457,14 @@ msgstr "Порт источника"
 msgid "Source zone"
 msgstr "Зона источника"
 
-msgid "Start Date (yyyy-mm-dd)"
+msgid "Start Date (yyyy-mm-ddThh:mm:ss)"
 msgstr "Дата начала (год-мес-день)"
```
namedun: `msgstr "Дата начала (гггг-мм-ддTчч:мм:сс)"`

kristrev:
Hi,

I just noticed that this PR is still open, I should pay more attention :) Before I proceed with resolving the merge, is the datetype-extension an acceptable solution @jow- ?

kristrev:
I will close this one and create a v3, fixing up the conflicts.

PR #2140, opened 2018-09-09
luci-base: Show multiple upstream interface

kristrev:
Several devices have multiple upstream interfaces, for example a fixed
and a mobile broadband connection. Currently, only one upstream
interface is shown per address family in Luci. So in my example, one of
the interfaces would not appear on the Status-page.

This PR introduces support for showing multiple upstream interfaces on
the Status-page. The code is not very complicated. get_status_by_route()
has been extended to return a list of all routes, and
get_wannet()/get_wan6net() now returns all upstream interfaces.

I could not find any other (active) users of these three functions than
calls triggered from the Status-page, so changing the default behavior
should be fine. get_wandev()/get_wan6dev() called get_status_by_route(),
but I could not find any place where those functions were called. I
removed the dev-functions instead of keeping the old
get_status_by_route().

On the status page, the wan/wan6-variables have been replaced with
arrays. When populating the html, we now iterate through these arrays
and create one element for each interface.

I have tested the code with different interface types, v4, v6, as well as
disconnecting and connecting interfaces. The status is updated and the
correct interfaces (or sometimes none at all) are shown.

### DIFF
--- a/modules/luci-base/luasrc/model/network.lua
+++ b/modules/luci-base/luasrc/model/network.lua
@@ -813,6 +813,7 @@ function del_wifinet(self, net)
 end
 
 function get_status_by_route(self, addr, mask)
+	local route_statuses = { }
 	local _, object
 	for _, object in ipairs(utl.ubus()) do
 		local net = object:match("^network%.interface%.(.+)")
@@ -822,12 +823,14 @@ function get_status_by_route(self, addr, mask)
 				local rt
 				for _, rt in ipairs(s.route) do
 					if not rt.table and rt.target == addr and rt.mask == mask then
-						return net, s
+						route_statuses[net] = s
 					end
 				end
 			end
 		end
 	end
+
+	return route_statuses
 end
 
 function get_status_by_address(self, addr)
@@ -856,24 +859,28 @@ function get_status_by_address(self, addr)
 	end
 end
 
-function get_wannet(self)
-	local net, stat = self:get_status_by_route("0.0.0.0", 0)
-	return net and network(net, stat.proto)
+function get_wan_networks(self)
+	local k, v
+	local wan_nets = { }
+	local route_statuses = self:get_status_by_route("0.0.0.0", 0)
+
+	for k, v in pairs(route_statuses) do
+		wan_nets[#wan_nets+1] = network(k, v.proto)
+	end
+
+	return wan_nets
 end
 
-function get_wandev(self)
-	local _, stat = self:get_status_by_route("0.0.0.0", 0)
-	return stat and interface(stat.l3_device or stat.device)
-end
+function get_wan6_networks(self)
+	local k, v
+	local wan6_nets = { }
+	local route_statuses = self:get_status_by_route("::", 0)
 
-function get_wan6net(self)
-	local net, stat = self:get_status_by_route("::", 0)
-	return net and network(net, stat.proto)
-end
+	for k, v in pairs(route_statuses) do
+		wan6_nets[#wan6_nets+1] = network(k, v.proto)
+	end
 
-function get_wan6dev(self)
-	local _, stat = self:get_status_by_route("::", 0)
-	return stat and interface(stat.l3_device or stat.device)
+	return wan6_nets
 end
 
 function get_switch_topologies(self)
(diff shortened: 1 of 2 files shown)


jow-:
Can you please provide a screenshot displaying the changed behaviour?

modules/luci-base/luasrc/model/network.lua
```diff
@@ -822,12 +823,14 @@ function get_status_by_route(self, addr, mask)
 				local rt
 				for _, rt in ipairs(s.route) do
 					if not rt.table and rt.target == addr and rt.mask == mask then
-						return net, s
+						route_statuses[net] = s;
```
jow-: No semicolon please

modules/luci-base/luasrc/model/network.lua
```diff
@@ -857,23 +860,25 @@ function get_status_by_address(self, addr)
 end
 
 function get_wannet(self)
-	local net, stat = self:get_status_by_route("0.0.0.0", 0)
-	return net and network(net, stat.proto)
-end
+	local wan_nets = { }
+	local route_statuses = self:get_status_by_route("0.0.0.0", 0)
+
+	for k, v in pairs(route_statuses) do
```
jow-: Please add a `local k, v`

modules/luci-base/luasrc/model/network.lua
```diff
@@ -857,23 +860,25 @@ function get_status_by_address(self, addr)
...
 end
 
 function get_wan6net(self)
-	local net, stat = self:get_status_by_route("::", 0)
-	return net and network(net, stat.proto)
-end
+	local wan6_nets = { }
+	local route_statuses = self:get_status_by_route("::", 0)
+
+	for k, v in pairs(route_statuses) do
```
jow-: Please add a `local k, v`

modules/luci-mod-admin-full/luasrc/view/admin_status/index.htm
```diff
@@ -56,45 +56,58 @@
...
-				link    = wan:adminlink(),
-				mac     = dev and dev:mac(),
-				type    = dev and dev:type(),
-				name    = dev and dev:get_i18n(),
-				ether   = link and link.type == 1
-			}
+		if #wan_nets > 0 then
+			rv.wan = { }
+
+			for k, v in pairs(wan_nets) do
```
jow-: Please add a `local k, v`

modules/luci-mod-admin-full/luasrc/view/admin_status/index.htm
```diff
@@ -56,45 +56,58 @@
...
-				link      = wan6:adminlink(),
-				mac       = dev and dev:mac(),
-				type      = dev and dev:type(),
-				name      = dev and dev:get_i18n(),
-				ether     = link and link.type == 1
-			}
+		if #wan6_nets > 0 then
+			rv.wan6 = { }
+
+			for k, v in pairs(wan6_nets) do
```
jow-: Please add a `local k, v`

modules/luci-base/luasrc/model/network.lua
```diff
@@ -857,23 +860,25 @@ function get_status_by_address(self, addr)
 end
 
 function get_wannet(self)
```
jow-: Please rename to `get_wan_networks`

modules/luci-base/luasrc/model/network.lua
```diff
@@ -857,23 +860,25 @@ function get_status_by_address(self, addr)
...
+		wan_nets[#wan_nets+1] = network(k, v.proto)
+	end
 
-function get_wandev(self)
-	local _, stat = self:get_status_by_route("0.0.0.0", 0)
-	return stat and interface(stat.l3_device or stat.device)
+	return wan_nets
 end
 
 function get_wan6net(self)
```
jow-: Please rename to `get_wan6_networks`

lucize:
for me, the current is like this

image

jow-:
@lucize - thanks :) Will merge after the code issues are addressed

kristrev:
@jow- Thanks for the comments. I (should have) addressed all of them now. I am not sure if you would like me to rebase + force push, or just push. I chose the latter and then I can merge the commits if you want that instead.

jow-:
@kristrev - thanks for integrating the requested changes. Yes please squash the two commits and force-push.

kristrev:
Fixed, commits are squashed + pushed.

jow-:
Perfect - thanks!

Issue #2146, opened 2018-09-13
luci-app-openvpn unable to create new instances


davidc502:
luci-app-openvpn unable to create new instances.
LuCI Master (git-18.250.48934-18b4483) / Lede SNAPSHOT r8018-42f158314e

Users are reporting they are unable to save new Openvpn instances on build r8018  I've have also been able to duplicate and confirm this behavior.

dibdot:
The LuCI app itself was not changed, probably related to this core change where default config options have been removed: 
(https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=89b8ba96b404d2e412d5573e25348287255671dd )

dibdot:
@davidc502 My first assumption was wrong, please test proposed minimal patch - thanks.

davidc502:
Testing complete -- Added to parenthesis to line 62, and was able to create a new instance. 

Thank you

dibdot:
Great, thanks for re-testing!

PR #2149, opened 2018-09-14
luci-app-openvpn: fix template based config creation

dibdot:
* minimal fix to bring back openvpn config creation based on
openvpn_recipes template, fix for #2146


mwoolweaver:
how would one test this branch locally?

hnyman:
> how would one test this locally?

You can 
* edit the file in a live router. It is a text file, so editing is easy
* download the PR commit as patch and apply it to live router.
* download the PR commit as patch and apply it to sources and compile the update package and test it.

(as the patch is about adding { and } to one line, editing the file is likely easiest testing method. The file should be /usr/lib/lua/luci/model/cbi/openvpn.lua  )

PR #2158, opened 2018-09-18
luci-mod-admin-full: add dependency for channel auto

HenryGiraldo:
Do not show 'auto' channel option if hostapd is not compiled with CONFIG_ACS

Signed-off-by: Enrique Giraldo <enrique.giraldo@galgus.net>

### DIFF
--- a/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
+++ b/modules/luci-mod-admin-full/luasrc/model/cbi/admin_network/wifi.lua
@@ -174,6 +174,7 @@ if found_sta then
 else
 	ch = s:taboption("general", Value, "_mode_freq", '<br />'..translate("Operating frequency"))
 	ch.iwinfo = iw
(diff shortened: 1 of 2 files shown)


HenryGiraldo:
It is also necessary to apply this patch to the hostapd.

[0001-hostapd-add-acs-feature-indication.txt](https://github.com/openwrt/luci/files/2392534/0001-hostapd-add-acs-feature-indication.txt)

This new feature is very interesting because if you set the channel to auto and the hostapd hasn't been compiled with ACS, the radios don't work.

feckert:
Please remove the last commit
https://github.com/openwrt/luci/pull/2158/commits/adbb096e857896b6af547f4fd413d82f2331c10b
This doesn't look okay so we could not merge this.

HenryGiraldo:
I just did, but there seems to be some conflict left. I can't see them, I pulled the upstream/master and it doesn't give me any problem cherrypicking the commit.

feckert:
Maybe you could close this pullrequest and open a new one.

PR #2159, opened 2018-09-18
luci-theme-material: make colors configurable

feckert:
The var() css function can be used to insert the value of a custom property.
A custom.css file is added to change the colors. This file will get included
in the cascade.css file.

The following colors could be costomized in the file costom.css.

--header-bg
--header-color
--menu-bg-color
--menu-color
--submenu-bg-hover
--submenu-bg-hover-active


jow-:
Merged, thank you!

PR #2160, opened 2018-09-18
luci-app-mosquitto: import new application

karlp:
Originally developed and maintained at https://github.com/remakeelectric/owrt_pub_feeds/tree/master/luci-app-mosquitto

### DIFF
--- a/applications/luci-app-mosquitto/Makefile
+++ b/applications/luci-app-mosquitto/Makefile
@@ -1,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Mosquitto LuCI interface
+LUCI_DEPENDS:=+luci-mod-admin-full mosquitto
+LUCI_PKGARCH:=all
+LUCI_DESCRIPTION:=Provides a webadmin for most basic mosquitto parameters.
+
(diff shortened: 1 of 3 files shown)


karlp:
I've imported this ~as is from our original feed, but it's missing lots of translation.  I chose to leave that off for a second followup so I can get that in pieces.

jow-:
Looks okay on a first glance - the Makefile should be converted to use the shared `luci.mk` which would make it much shorter. Is the ucitrack entry needed yet?

karlp:
Honestly, I've not followed when luci_reload /ucitrack can go away.  This package worked without having to worry what branch it was on.  On current master, are you saying that it _shouldn't_ be required anymore, as long as the init script is setup properly?

jow-:
Correct, if Mosquitto's init script is procd enabled, then you do not need the ucitrack business anymore.

jow-:
It applies to both master and openwrt-18.06 but not to lede-17.01

karlp:
wow, this ```# call BuildPackage - OpenWrt buildroot signature``` magic string in a comment was.... not very visible to work out why my package wasn't showing up :)

karlp:
Ok, dropped ucitrack and confirmed no longer required, and moved to the luci.mk style.

jow-:
merged, thanks!

PR #2167, opened 2018-09-20

YuriPet:
luci-base: corrected and update Ukrainian translation with the latest resync.


jow-:
Merged, thanks!

PR #2168, opened 2018-09-20
luci-app-snmpd: Basic net-snmpd config options

karlp:
Imported from https://github.com/remakeelectric/owrt_pub_feeds/tree/master/luci-app-snmpd

Only provides configuration of basic authentication and system level
options, not even close to the entire slew of options that net-snmp
supports, but the basics are still helpful, and a base for future work.


jow-:
Merged, thanks!

PR #2173, opened 2018-09-23
luci-base: fix table align for mount-points

rosysong:
This commit fix the table align issue of mount-points module.

**Before**
fix-table-align-for-mount-points-before

**After**
fix-table-align-for-mount-points


rosysong:
@hnyman

PR #2174, opened 2018-09-23
luci-base: drop ipv6 lease status when IPV6 is not support

rosysong:
If the system does not have ipv6 support, just drop the relevant module, so that the page seems more concise. :-)


feckert:
LGTM

rosysong:
Thanks!

feckert:
Thanks

PR #2175, opened 2018-09-23
luci-mod-system: add feature that allow users to eject scsi devices safely

rosysong:
Hi all, 
    As we all known, there is a **"Safely Eject"** feature in desktop OS, but why not implement it in OpenWrt ? :-), This commit can help us to safely Eject the SCSI devices(USB, TF Card, SATA HD, ....) and also prevent from unplug them incorrectly, otherwise you will always see the below warning in logread:

......
**kern.warn kernel: [   11.900086] FAT-fs (sda1): Volume was not properly unmounted. Some data may be corrupt. Please run fsck.**

......

eject-on-desktop

**Here is the screenshot:** 

eject

**NOTE:**
This commit depends on the lsblk and eject utilities in util-linux (https://github.com/openwrt/openwrt/pull/1403). I also tested 
the simplified version provided by busybox, but they always delay or failed. 
Thanks @diizzyy for reminding me.

### DIFF
--- a/modules/luci-base/luasrc/sys.lua
+++ b/modules/luci-base/luasrc/sys.lua
@@ -70,6 +70,32 @@ function mounts()
 	return data
 end
 
+function scsis()
+	local data = {}
+	local ps = luci.util.execi("lsblk -d -n -P -S -o NAME,TYPE,VENDOR,MODEL,REV,TRAN,SIZE")
+
+	if not ps then
+		return
+	end
+
+	for line in ps do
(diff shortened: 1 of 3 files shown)


rosysong:
This feature depends on https://github.com/openwrt/packages/pull/7088 and https://github.com/openwrt/packages/pull/7087

rosysong:
@hnyman

feckert:
Please rebase your changes to the new javascript code

rosysong:
@feckert  OK, will update later. Thanks!

PR #1682, opened 2018-03-15
luci-proto-3g: fix error on proto selection

feckert:
If you select the proto 3g you will see the following error

> Missing protocol extension for proto "3g"
cannot open /usr/lib/lua/luci/model/cbi/admin_network/proto_3g.lua: No such file or directory

To fix this move the 3g proto from `luci-proto-ppp` to `luci-proto-3g` and add the missing file `proto_3g.lua` under `/model/network/`.

PR #2176, opened 2018-09-24
protocols: move hnet/pppossh to own packages

feckert:
This is a follow PR from the discussion with @yousong in https://github.com/openwrt/luci/pull/1682#issuecomment-387644397

A lot of protocols are mixed together which from my point of view should go into separate packages.
If the user has a special use cases then the missing protocols could be installed with opkg.


protocols/luci-proto-pppossh/Makefile
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (C) 2018 Florian Eckert <fe@dev.tdt.de>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Support for hnet
+LUCI_DEPENDS:=+pppossh
```
yousong: wrong title

protocols/luci-proto-pppossh/luasrc/model/network/proto_pppossh.lua
```diff
@@ -0,0 +1,48 @@
...
+	return true
+end
+
+function proto.get_interfaces(self)
+	if self:is_floating() then
+		return nil
+	else
+		return netmod.protocol.get_interfaces(self)
+	end
+end
```
yousong: Can this function simply return nil?  Or is that the default if we leave it out?

protocols/luci-proto-pppossh/luasrc/model/network/proto_pppossh.lua
```diff
@@ -0,0 +1,48 @@
...
+
+function proto.contains_interface(self, ifc)
+	if self:is_floating() then
+		return (netmod:ifnameof(ifc) == self:ifname())
+	else
+		return netmod.protocol.contains_interface(self, ifc)
+	end
+end
```
yousong: ditto

protocols/luci-proto-hnet/Makefile
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (C) 2018 Florian Eckert <fe@dev.tdt.de>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Support for hnet
+LUCI_DEPENDS:=
```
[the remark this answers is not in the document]
feckert: There are 3 packages selectable hnet-full, hnet-full-secure or hnet-full-l2tp I am not sure how to depend on a package with variants.
I think we need a extra package for the hnet netifd proto?
If so I will slice the netifd proto out of the variants https://github.com/openwrt-routing/packages/blob/master/hnetd/Makefile
and add a new package hnet which has only the proto of the netifd.

protocols/luci-proto-pppossh/Makefile
feckert: Thanks for the hint

protocols/luci-proto-hnet/Makefile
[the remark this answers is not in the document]
yousong: I see, seems the same question was also asked there in the comment ;)  I may resort to `PROVIDES` and try `depends` instead of `select`.  I think it's okay we leave it be and let LuCI test the existence of hnet.sh

protocols/luci-proto-pppossh/luasrc/model/network/proto_pppossh.lua
feckert: The luci-proto-pppossh is floating so we only have to change the functions to

> return (netmod:ifnameof(ifc) == self:ifname())

protocols/luci-proto-pppossh/luasrc/model/network/proto_pppossh.lua
feckert: Yes we have to return nil.
The luci-proto-qmi is doing the same so i will only return nil.
Thanks

protocols/luci-proto-hnet/Makefile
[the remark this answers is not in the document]
feckert: I have staged a PR on my github acount to move the hnet proto for netifd to an own package.
So we could depend on hnet?
https://github.com/TDT-AG/packages-1/commit/6100b458e9e615fe39d96b7f4cbdbe9bb8fb73d6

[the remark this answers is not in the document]
yousong: It's a bit overkill for the hnet package, plus I am afraid having only hnet.sh is not sufficient for the protocol to work in the first place.

yousong:
I plan to give the series a run and do the merge in the following days unless there are further comments on this change ;)  Thank you

yousong:
Thank you.

PR #2177, opened 2018-09-24
luci-mod-network: disable wifi config on new wifi-iface add

feckert:
If we add a new wifi-iface to the config then the iface will start at once.
But normaly we would configure the wireless security in the next step.
So do not start wifi iface on add.


jow-:
Merged, thanks!

PR #2178, opened 2018-09-24
community-profiles: change subnet of mesh_network option for Berlin

booo:
The mesh_network option is used to check the user input. The wizard
checks if the input ip address is part of the mesh_network. We use
multiple /16 networks. There is no support for multiple ranges so we the
10.0.0.0/8.

For reference the ip list:

https://wiki.freifunk.net/IP-Netze

Signed-off-by: Philipp Borgers <borgers@mi.fu-berlin.de>


jow-:
Merged, thanks!

PR #2179, opened 2018-09-24
luci-app-mosquitto: validate paths as path, not file

karlp:
luci has separate validation for directories and files, make sure to
use the right one.


jow-:
Merged, thanks!

PR #2181, opened 2018-09-25
luci-theme-material: fixes and add openwrt logo brand

feckert:
* Fix whitspaces and change file mode to 644 for footer.htm
* Update css and add openwrt brand.png
* Fix partly hidden logout navbar item


jow-:
Merged, thanks!

PR #2183, opened 2018-09-26
Parse method fixed in cbi.lua

jokdarius:
*	Method "parse" fixed. It's not necessary to execute all code if section already exists.

Signed-off-by: Darius Joksas <jok.darius@gmail.com>

### DIFF
--- a/modules/luci-base/luasrc/cbi.lua
+++ b/modules/luci-base/luasrc/cbi.lua
@@ -1200,7 +1200,7 @@ function TypedSection.parse(self, novld)
 				-- Ignore if it already exists
 				if self:cfgvalue(name) then
 					name = nil;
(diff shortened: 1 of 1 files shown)


modules/luci-base/luasrc/cbi.lua
```diff
@@ -1200,18 +1200,19 @@ function TypedSection.parse(self, novld)
 				-- Ignore if it already exists
 				if self:cfgvalue(name) then
 					name = nil;
-				end
-
-				name = self:checkscope(name)
+					self.err_invalid = tru
```
jow-: This looks like a typo.

jow-:
Merged with e5071c88f - thanks!

PR #2184, opened 2018-09-26
luci-base: fix sysauth cookie not removed on logout

kyle30312:
I noticed the `sysauth` cookie remains after logging out from luci, because the cookie path doesn't match when trying to remove it.  This change corrects the path so the cookie is removed.

In addition, because the cookie is being removed, the cookie value is moot, so this change uses `''` for the cookie value instead of the now-destroyed session ID.

# More details

## HTTP headers before the change:
1. at login, response to `POST https://192.168.8.1/cgi-bin/luci` :
`Set-Cookie: sysauth=45226885ec8484b575e4cd3e2c55ab22; path=/cgi-bin/luci/; HttpOnly; secure`
2. throughout the session, `Cookie` request header and value present (as expected) in requests to `https://192.168.8.1/cgi-bin/luci/...`
3. at logout, response to `GET https://192.168.8.1/cgi-bin/luci/admin/logout` :
`Set-Cookie: sysauth=45226885ec8484b575e4cd3e2c55ab22; expires=Thu, 01 Jan 1970 01:00:00 GMT; path=/cgi-bin/luci//` (wrong path)
4. Cookie header and value still present in subsequent request to `GET https://192.168.8.1/cgi-bin/luci/`

## HTTP headers after the change:
1. same as above
2. same as above
3. at logout, response to `GET https://192.168.8.1/cgi-bin/luci/admin/logout` :
`Set-Cookie: sysauth=; expires=Thu, 01 Jan 1970 01:00:00 GMT; path=/cgi-bin/luci/` (correct path)
4. Cookie header now absent (as expected) from subsequent request to `GET https://192.168.8.1/cgi-bin/luci/`

## How the bug was introduced
- In ec1a86977b1dc5cfc1c24ab1d54205531404087b (Feb 9, 2015), a slash was appended where needed for both the login and logout cases.
- In 79383f5a74bcff528a5270a930c36b58e310010c (Oct 21, 2015), the additional slash was no longer needed, and it was removed from the login case but not the logout case.


jow-:
Thank you for this patch! Can you append your Signed-off-by so that I can merge the PR? (`git commit -s --amend && git push origin logout-cookie-fix --force`)

kyle30312:
@jow- Sign-off done - does that look right?

PR #2185, opened 2018-09-27
luci-app-mwan3: start adding german translation a new max_ttl option

feckert:
- Start of adding german translation
- Add new max_ttl option see https://github.com/openwrt/packages/commit/c9d8fceb63a425f19fc12bee7ee4191fad514b6e#diff-87268a842bab5effbe181f9f18b2d348


jow-:
Merged, thanks!

PR #2192, opened 2018-09-29
luci-app-network: add rekey option

Ansuel:
This option is usefull to solve some problems with ath10k-ct and random client disconnect


jow-:
Merged - thanks!

Issue #2180, opened 2018-09-24
wireguard UI should provide QR code for setup


nerdoc:
The setup of an wireguard interface is a bit uneasy, as copying a public key from the cellphone is not that easy for most people, or manually typing it is error-prone. I hate doing either of it. There would be an easy way to add an Android setup, if LuCi would provide the user with a QR code that the Android App could easily scan.

TrashCanMan:
Or at the very least a .conf file generator of sort so we can download a conf file from the router and use it on the android client without the need to manually type the keys wich is prone to errors with 0 and O (zero and cap o) and l and I (lower L, cap i) etc...the file format can be seen here https://git.zx2c4.com/WireGuard/about/src/tools/man/wg.8
Also the command to generate the code can be seen here:
https://wiki.debian.org/Wireguard#A3._Import_by_reading_a_QR_code_.28most_secure_method.29
but im pretty sure that is going to require an extra package, on the other side the conf file AFAIK dont need anything extra...

yousong:
Ping luci-app-wireguard's current maintainer @danrl  ;)

If you just open LuCI in cellphone browser, the copy should be easy in both direction.  I can also think of many ways to transfer such info from desktop browser to cellphone, online qrcode generator, copy and send to IM.  There are also existing luci apps such as travelmate, fwknopd using qrencode to generate and embed svg qrcode.

But I cannot think of a better way to transfer public key from cellphone to desktop browser ;)  Usually I just send these literal strings, qrcode, encoded urls to IM then receive and use them in desktop system ;)

danrl:
Sounds like a nice idea. I am open to pull requests implementing the functionality. It's open source, get involved :)

TrashCanMan:
Im smart and good at googling but i know literally zero about coding, i posted the links to help the best i can with my """skills""" :D

nerdoc:
I think it's not about only the key. It's also the IP address. I always have troubles finding the right settings, only after trying a bunch of netmasks.
\<dream>I would be so a killer feature to just "hold" the phone camera onto the screen and *have a working wireguard connection*.\</dream>

dibdot:
@danrl Should the QR code added to the wg status page?

danrl:
[...]

PR #2195, opened 2018-10-01
luci-app-wireguard: add QR Code support plus fixes

dibdot:
* add optional QR Code support to status page (per wg Interface),
  submitted information via QR code are:
  - [Interface] (list of) Address / PrivateKey
  - [Peer] Endpoint / PublicKey / (list of) AllowedIPs
  - support multiple interfaces and peers
* fix XHTML rendering errors with OpenWrt standard theme
* fix logical glitch with empty descriptions, remove needless variables
* left align the output for better viewing esp. with material theme
* freshen up design a little bit
* switch space indentation to tabs

### DIFF
--- a/applications/luci-app-wireguard/luasrc/controller/wireguard.lua
+++ b/applications/luci-app-wireguard/luasrc/controller/wireguard.lua
@@ -4,5 +4,5 @@
 module("luci.controller.wireguard", package.seeall)
 
 function index()
-  entry({"admin", "status", "wireguard"}, template("wireguard"), _("WireGuard Status"), 92)
+	entry({"admin", "status", "wireguard"}, template("wireguard"), _("WireGuard Status"), 92)
 end
--- a/applications/luci-app-wireguard/luasrc/view/wireguard.htm
+++ b/applications/luci-app-wireguard/luasrc/view/wireguard.htm
@@ -1,224 +1,264 @@
 <%#
- Copyright 2016-2017 Dan Luedtke <mail@danrl.com>
- Licensed to the public under the Apache License 2.0.
+	Copyright 2016-2017 Dan Luedtke <mail@danrl.com>
+	Licensed to the public under the Apache License 2.0.
 -%>
 
 <%
-  local data = { }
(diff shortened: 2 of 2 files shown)


dibdot:
A few screenshots from the "new" status page wir QR Codes:

**Default view (with collapsed QR Code):**
wg_4

**QR Code view (of course usually without U2 tag ...;-)):**
wg2

**Hint for users without qrencode library:**
wg3

danrl:
I looked briefly over the proposed changes and they look good. I do not have access to a system where I could test this with. That means, the value of my review is limited.

Two minor questions:
* Would we need to escape the string that we send to `qrencode` somehow or can we trust it to not contain characters that would end the quotation of the string etc.?
* Do you mind saying a word about why you replaced `fieldset` by `div`. I think I can learn something here. Thank you!

danrl:
CC @zx2c4 
Look at this, the interface is getting cooler every day!

dibdot:
@danrl thanks for your feedback. Meanwhile I've added the interface address(es) to the provided QR Codes well - as requested by @nerdoc in #2180.
Regarding your questions:
- the qrcode input characters are coming from wg dump, therefore I would say they are safe. During my tests with the wireguard android app I didn't see any failures caused by unquoted chars or something like that.
- in the past "fieldset" was used for grouping & the "legend" tag - nowadays page structuring and grouping will be done by divs (we've removed the old table structure) and replaced legend with (a more screenreader friendly) "h3" tag

zx2c4:
This is super cool. Thanks for pointing it out to be @danrl. It's a bit hard to review, though, because it appears that every line has changed, due to formatting fixes?

dibdot:
@danrl ok to merge?

danrl:
@dibdot It is fine for me, but as I said, I currently lack the setup and hardware to properly test it.

dibdot:
OK, than let's merge it to get more feedback! ;-)

shelterx:
Is it possible to add more data to the qrcode, at least the IP range and perhaps set the router ip as dns?

PR #2197, opened 2018-10-01
Fix typo

yangfl:


hnyman:
Rather large amount of unrelated changes as one single commit.

At the first glance, some of the corrections look legit, but some may break functionality (e.g. by changing the uci config option name without doing the same change in the actual package using uci. Actually, the change in watchcat looks like a fix, while the change in dump1090 actually changes an option that has already been removed from the app itself) or break translations (as similar change is not done for the translated language files)

Some fix-like changes (like the correction of mwan3 function get_gateway argument name from "inteface") make me wonder if the app has worked ok this far by accident or has the app been failing all the time.  @feckert 

And some changes are rather irrelevant, e.g. grammatically misspelled word (childs/children) in a function internal argument name that is only used internally in that function, so the change does not actually do anything.

Similarly, there are spelling fixes for header files taken from upstream, and I am not sure if changing those is necessary/wise.

Not quite sure what would be the best approach. We had 1-2 years ago one large spelling fix like this, and it broke some things...

I think that this should maybe be split into several different commits & PRs, as fixing misspelling in documentation is quite different than changing a shell script or source code modifying the actual functionality. Some of the changes may also require attention from @jow- or the respective app maintainers (if any).

feckert:
@hnyman @yangfl thanks for the hint the code did work because the variable **interface** is a global definied. But indeed this should be fixed its not nice.

yangfl:
@hnyman It should be fine to merge it now since it only contains fixes for comments.

modules/luci-base/luasrc/view/cbi/tabmenu.htm
```diff
@@ -1,7 +1,7 @@
 <%- if self.tabs then %>
 	<ul class="cbi-tabmenu">
 	<%- self.selected_tab = luci.http.formvalue("tab." .. self.config .. "." .. section) %>
-	<%- for _, tab in ipairs(self.tab_names) do if #self.tabs[tab].childs > 0 then %>
+	<%- for _, tab in ipairs(self.tab_names) do if #self.tabs[tab].children > 0 then %>
```
jow-: I think this will break

PR #2198, opened 2018-10-02
luci-app-mwan3: fix i18n traceback

feckert:
The i18n translation did not work in the controller as expected, due missing i18n lua requirements.
This PR will fix this issue.


feckert:
This should fix https://github.com/openwrt/packages/issues/6982#issuecomment-421709798

PR #2201, opened 2018-10-05
luci-app-tinyproxy: upstream proxy URL:port

safari2:
Please add possibility to insert upstream proxy URL:port instead of ip_address:port 
It works if I add URL:port in config file manually. 

Example for /etc/init.d/tinyproxy:
adding echo 'upstream proxy.antizapret.prostovpn.org:3128 "rutracker.org"'  >> $CFGFILE

results in /var/etc/tinyproxy.conf:
upstream proxy.antizapret.prostovpn.org:3128 "rutracker.org"


dibdot:
Bogus PR, please try again.

dibdot:
Just note, your PR should apply probably to master, not 18.06 branch.

PR #2207, opened 2018-10-07
luci-proto-wireguard: permit storing private & preshared keys in files

LukaszDargiewicz:
luci corresponding changes to https://github.com/openwrt/openwrt/pull/1391
Signed-off-by: Lukasz Dargiewicz <lukasz.dargiewicz@gmail.com>

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -1862,13 +1862,18 @@ var CBIFileUpload = CBIValue.extend({
 	},
 
 	renderWidget: function(section_id, option_index, cfgvalue) {
+		var loadValueFn = this.loadValue;
 		var browserEl = new ui.FileUpload((cfgvalue != null) ? cfgvalue : this.default, {
 			id: this.cbid(section_id),
 			name: this.cbid(section_id),
 			show_hidden: this.show_hidden,
 			enable_upload: this.enable_upload,
 			enable_remove: this.enable_remove,
-			root_directory: this.root_directory
+			root_directory: this.root_directory,
+	                loadValue: function(value) {
+		                return (loadValueFn) ? loadValueFn(section_id, value) : value;
+			},
+			writeValue: this.writeValue
 		});
 
 		return browserEl.render();
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -2,6 +2,8 @@
 'require form';
 'require network';
 
+var root_directory = '/etc/wireguard';
+
 function validateBase64(section_id, value) {
 	if (value.length == 0)
 		return true;
@@ -10,6 +12,14 @@ function validateBase64(section_id, value) {
 		return _('Invalid Base64 key string');
 
 	return true;
+}
+
+function prependRootDirectory(value) {
+	return (value) ? root_directory + '/' + value : value;
+}
+
+function stripRootDirectory(value) {
+	return value.split(root_directory + '/')[1];
 }
 
 return network.registerProtocol('wireguard', {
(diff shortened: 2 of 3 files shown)


feckert:
I think we should iterate over **/etc/wireguard** and list all files in a dropdown for selection.
So we select the file we want to use.
If no file is in this directory we could also maybe not show this field.

Additional i think we should remove the placeholder comment.
https://github.com/openwrt/luci/pull/2207/files#diff-daba76e1569c21be9540dd6307bd578dR24
I think this is confusing and is not meant to display information there but rather default settings

feckert:
@LukaszDargiewicz Please rebase your changes so we can go on with this pullrequest?

jow-:
This needs to be ported to the JavaScript based ui here: https://github.com/openwrt/luci/blob/master/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js#L46

For the file selection, a new option of type `FileUpload` can be used, example here:
https://github.com/openwrt/luci/blob/master/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L1380

LukaszDargiewicz:
> @LukaszDargiewicz Please rebase your changes so we can go on with this pullrequest?

I will.
I have updated openwrt/openwrt pull request https://github.com/openwrt/openwrt/pull/2416 so far

LukaszDargiewicz:
> This needs to be ported to the JavaScript based ui here: https://github.com/openwrt/luci/blob/master/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js#L46
done
> 
> For the file selection, a new option of type `FileUpload` can be used, example here:
> https://github.com/openwrt/luci/blob/master/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L1380
files can now be selected

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -1862,13 +1862,18 @@ var CBIFileUpload = CBIValue.extend({
...
 		var browserEl = new ui.FileUpload((cfgvalue != null) ? cfgvalue : this.default, {
 			id: this.cbid(section_id),
 			name: this.cbid(section_id),
 			show_hidden: this.show_hidden,
 			enable_upload: this.enable_upload,
 			enable_remove: this.enable_remove,
-			root_directory: this.root_directory
+			root_directory: this.root_directory,
+	                loadValue: function(value) {
+		                return (loadValueFn) ? loadValueFn(section_id, value) : value;
```
feckert: Please fix indentation  to tab

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -114,9 +139,24 @@ return network.registerProtocol('wireguard', {
 		o.validate = validateBase64;
 		o.rmempty = false;
 
-		o = ss.option(form.Value, 'preshared_key', _('Preshared Key'), _('Optional. Base64-encoded preshared key. Adds in an additional layer of symmetric-key cryptography for post-quantum resistance.'));
+		var preshared_key_files = {},
+		    comment = ' Adds in an additional layer of symmetric-key cryptography for post-quantum resistance.';
```
feckert: Maybe we could wrap this in _('.....') and add this to lines with `, _('....') + comment )`?
I have not verified if this works but if we would have only one additional translation and not extend two with this additional text.

jow-:
Why is the `loadValue` and `writeValue` commit needed?

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -114,9 +139,24 @@ return network.registerProtocol('wireguard', {
...
+		o = ss.option(form.FileUpload, 'preshared_key_file', _('Preshared Key file'), _('Optional either this or below. Preshared key file.' + comment));
+		o.root_directory = root_directory;
+		o.enable_upload = false;
+		o.enable_remove = false;
+		o.writeValue = stripRootDirectory;
```
jow-: Replace `o.writeValue` with:

```
o.write = function(section_id, value) {
    form.Value.prototype.write.apply(this, [ section_id, stripRootDirectory(value) ]);
};
```

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -114,9 +139,24 @@ return network.registerProtocol('wireguard', {
 		o.validate = validateBase64;
 		o.rmempty = false;
 
-		o = ss.option(form.Value, 'preshared_key', _('Preshared Key'), _('Optional. Base64-encoded preshared key. Adds in an additional layer of symmetric-key cryptography for post-quantum resistance.'));
+		var preshared_key_files = {},
+		    comment = ' Adds in an additional layer of symmetric-key cryptography for post-quantum resistance.';
+		o = ss.option(form.FileUpload, 'preshared_key_file', _('Preshared Key file'), _('Optional either this or below. Preshared key file.' + comment));
```
jow-: This will not work, the argument to `_(...)` *must* be a literal string - so use `comment = _('Adds an additional ...')` above and `_('Optional either this ...') + ' ' + comment` here.

feckert:
Related pullrequest was closed https://github.com/openwrt/openwrt/pull/1391

Issue #2138, opened 2018-09-09
Would luci-app-statistics support collectd-mod-curl?


jinchizhong:
Hi everyone,

I'm interested in run collectd-mod-curl on OpenWRT.
I manually configured collectd-mod-curl, it works, and .rrd file appear on where I want.
But result doesn't appear on luci/Statistics/Graph.

After some research, I find it out, luci-app-statistics doesn't support collectd-mod-curl by now.
Is there any plan to support it?

If no, is it easy to implement? Can anybody guide me how to implement such a feature?

Thanks to all

hnyman:
> luci-app-statistics doesn't support collectd-mod-curl by now.
> Is there any plan to support it?

No.
You are the first one that has asked for it.

> If no, is it easy to implement? Can anybody guide me how to implement such a feature?

Rather easy.
You just do similar edits as for e.g. cpufreq and apcups module:
https://github.com/openwrt/luci/commit/df3c8728b995026a6df13efaf9a782b486de9ec5#diff-64c154d89ea097b12be97c2bc71b09da
https://github.com/openwrt/luci/commit/808f2617931dcd6f84637e1bff3a737b9368917c#diff-64c154d89ea097b12be97c2bc71b09da

You can do the editing in a live router, and once you get things working, you can then transfer your modifications to the git repo and submit a pull request here, to get your code merged into the official repo.

jinchizhong:
@hnyman Thank you,
I'll have a try next weekend.

jinchizhong:
Hi @hnyman 

I met a small problem, can you have a look?
![image](https://user-images.githubusercontent.com/1650368/46408482-e0526580-c744-11e8-8da2-701f97d87841.png)

In the first red box, I hope there should be '40 ms'. （the unit should be 'ms' not 'm')
And in the second red box, I hope there should be '30 ms'. (the value should x1000)

I can't find where to change them.

Below is my code by now.
```
-- statistics/rrdtool/definitions/curl.lua
module("luci.statistics.rrdtool.definitions.curl", package.seeall)

function rrdargs( graph, plugin, plugin_instance, dtype )
        return {
                title = "%H: cUrl Response Time for #%pi",
                y_min = "0",
                alt_autoscale_max = true,
                vlabel = "Response Time",
                number_format = "%5.1lfms",
                data = {
                        types = { "response_time" },
                        options = {
                                response_time = {
                                        title = ""
                                }
                        }
                }
        }
end
```

Thanks a lot

[...]

PR #2209, opened 2018-10-08
luci-app-statistics: add support for cUrl

jinchizhong:
Add collectd-mod-curl plugin support. Which can do some complex track,
such as grab stock, but by now, only response time are supported.

#2138

### DIFF
--- a/applications/luci-app-statistics/root/etc/config/luci_statistics
+++ b/applications/luci-app-statistics/root/etc/config/luci_statistics
@@ -66,6 +66,9 @@ config statistics 'collectd_cpu'
 config statistics 'collectd_cpufreq'
 	option enable '0'
 
+config statistics 'collectd_curl'
+	option enable '0'
+
 config statistics 'collectd_df'
 	option enable '0'
 	option Devices '/dev/mtdblock/4'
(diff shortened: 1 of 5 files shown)


applications/luci-app-statistics/root/etc/config/luci_statistics
```diff
@@ -66,6 +66,9 @@ config statistics 'collectd_cpu'
 config statistics 'collectd_cpufreq'
 	option enable '0'
 
+config statistics 'collectd_curl'
+    option enable '0'
```
hnyman: minor nitpick: tab instead of spaces

hnyman: I think I will fix it myself before committing this.

hnyman:
Apart from that one tab change, the code compiles and could be merged.
But I just noticed that you have not "signed-off-by" the commit. That is required.

Please amend your commit in the PR:
That one tab change, plus add the signed-off-by line into the commit message itself.

jinchizhong:
Hi @hnyman 

I have rewrite my commit to add signed-off-by.
And fixed the space/tab problem.

PR #2220, opened 2018-10-10
luci-mod-system: block install and search if no package lists

yangfl:


jow-:
Please use `disabled="disabled"` to not upset the XHTML OpenWrt theme :)

jow-:
Merged, thanks!

PR #2221, opened 2018-10-11
 treewide: Fix typos in UI strings

yangfl:


hnyman:
Please check that you do not break existing translations (and maybe use sed to edit them at the same time).

I fixed the fallout from your previous batch with https://github.com/openwrt/luci/commit/febda2e4a758d6d2869995c0c6df4b047e6def6a as there was one UI string included that actually was translated. The existing translations would have unnecessarily broken in the next translation sync.

You can see the effect of your changes on translation string with ./build/i18n-sync.pl

Ps.
I used sed in app's po directory with:
` sed -i -e "s/unneccesary downloads/unnecessary downloads/" */*.po*`

yangfl:
@hnyman I thought we should let it auto synced and mark old translations as fuzzy, but if you prefer directly fixing the original strings, I'll take care of them later.

hnyman:
> I thought we should let it auto synced and mark old translations as fuzzy

It makes no sense to cause the proper foreign language to get hidden just because the English original string changes due to a English typo fix. Better to fix the English key in each language, so that the translation stays working.

hnyman:
Thanks.

PR #2226, opened 2018-10-12
luci-app-dump1090: Fix typo of a dumb option

yangfl:
Option `aggressive` is supported by dump1090 binary, but oddly missing in
dump1090.init. Nevertheless, neither of them support the wrong
`agressive`.


hnyman:
`aggressive` was removed from the init script in 2016, I think.

https://github.com/openwrt/packages/commit/a708fa2f547815acda1fccc809cec1a1761835b5#diff-77942f98b42f0fb3d457266a32715350

When it is missing from the init script, the LuCI setting has no effect.

We can merge this, but I think that the change has no actual effect unless the init script is fixed, too.
Not sure if @Noltari removed that option intentionally.

PR #2238, opened 2018-10-21
luci-app-olsr-services: Don't use the nameservice with a specific ver…

mmouselli:
…sion.

The "Services" menu option of olsr on the top of the web interface
is not there.

A specific version of the App name was in the controller service
Removing the limitation will fix the problem

https://github.com/freifunk-berlin/firmware/issues/594


jow-:
Merged, thanks!

PR #2243, opened 2018-10-23
luci-app-openvpn: file name fix

dibdot:
* fix shellquote oversight of the last commit


dibdot:
@jow- sorry, I've tested this only with an old, cached controller before.

PR #2245, opened 2018-10-24
luci-mod-system: add support for additional filesystem

Ansuel:
The filesystem option doesn't include additional filesystem included with /etc/filesystems, this is usefull if someone have ntfs-3g installed and wants to force use the external utility to mount partition instead of the default present in the kernel (if it's supported)
@jow- 

### DIFF
--- a/modules/luci-mod-system/luasrc/model/cbi/admin_system/fstab/mount.lua
+++ b/modules/luci-mod-system/luasrc/model/cbi/admin_system/fstab/mount.lua
@@ -136,6 +136,13 @@ for fs in io.lines("/proc/filesystems") do
 	end
 end
 
+local ok, lines = pcall(io.lines, "/etc/filesystem")
+if ok then
+	local fs
+	for fs in lines do
+		o:value(fs)
+	end
+end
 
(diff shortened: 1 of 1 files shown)


modules/luci-mod-system/luasrc/model/cbi/admin_system/fstab/mount.lua
```diff
@@ -136,6 +137,11 @@ for fs in io.lines("/proc/filesystems") do
 	end
 end
 
+if has_add_filesystem then
```
jow-: Better use something like:

```
local ok, lines = pcall(io.lines, "/etc/filesystem")
if ok then
  local fs
  for fs in lines do
    o:value(fs)
  end
end
```

Ansuel:
@jow- done

Ansuel:
@jow- can this be merged?

jow-:
Please indent using tabs, then it should be fine

Ansuel:
@jow- sorry for the mess now should be good

jow-:
Yep, thanks :)

PR #2246, opened 2018-10-25
theme: add initial commit for rosy theme

rosysong:
Hi all, this is a theme new design for LuCI and OpenWrt. Below is the screenshots, hope you will like it.

deepin-screen-recorder_google-chrome_20181025112639

deepin-screen-recorder_select area_20181025112325

deepinscreenshot_20181025112418

deepinscreenshot_20181025112519


rosysong:
@jow-  @hnyman

jow-:
I don't have the capacity to maintain yet another theme so if this gets merged you have to take care of tracking master compatibility yourself.

rosysong:
@jow-  OK, got it, no problem! I will maintain this theme.

musashino205:
@rosysong Is it possible to drop ```Microsoft Yahei, WenQuanYi Micro Hei``` from the font specified by css?

Even in Japan, the same characters as Chinese are used, but in Chinese and Japanese there are characters with different glyphs.
example: https://en.wikipedia.org/wiki/Han_unification#Examples_of_language-dependent_glyphs

Since it is difficult to solve this problem by specifying the font, I propose to use the browser's default font.

### Comparison in luci-theme-rosy

default (```Microsoft Yahei, WenQuanYi Micro Hei```):
openwrt-luci_theme-rosy_font_cn

edited (drop *hei):
openwrt-luci_theme-rosy_font_jp

Glyphs after editing is appropriate for Japanese.

rosysong:
@musashino205  OK! will drop it later.

jow-:
So is this ready for merge now?

rosysong:
@jow- Yes, thanks :-)

rosysong:
And please don't forget to check this, https://github.com/openwrt/luci/pull/2193. Only add a id attribution for the div tag.

jow-:
Merged, thanks

PR #2249, opened 2018-10-25
luci-mod-system: add "KiB" to translatable strings

namedun:
Also changed original "kiB" to "KiB" in accordance with IEEE 1541-2002.


aparcar:
Do you have an example of an translation of "KiB" for any other language? Just curious

namedun:
Yes, for Russian "KiB" translates as "КиБ" in accordance with local standards.

aparcar:
thanks

PR #2250, opened 2018-10-25
get rid of library version numbers in luci olsrd code

booo:
Signed-off-by: Philipp Borgers <borgers@mi.fu-berlin.de>


jow-:
I'm a bit out of the loop with olsrd, is the init script taking care of the version suffixes now? If so the PR is fine for me to merge.

pmelange:
Here is the upstream change https://github.com/openwrt-routing/packages/commit/f4bc87dc65729ef663f6dc9b7e157a909c5eb92b

jow-:
Ah, perfect. Thanks!

PR #2257, opened 2018-11-07
luci-base: limit new pkg version on release branches

### LINKED ISSUE #2156: Luci should not generate new package releases with no change
Hello,

Currently, when any commit is made to a stable branch, all luci packages gets a new release, even those without changes. For those that install all "opkg list-upgradable" packages, there are some more precious kbytes consumed in /overlay. Since 18.06.1, there is already a new luci release. All luci-* packages in standard rom image have a an update but not real file change.

Would it be possible to have package versions be more conservative?

Could it use a fixed git-commit (a tag or a Makefile var) as a base commit when there is a openwrt release and use PKG_RELEASE as the "1+(number of commits since the base commit)". Each module/app/theme/(...) has its own directory and it would be simple to count the number of commits since the "fixed git-commit" inside each directories.

If there is a change that must recompile all packages, simply update the base git-commit. It will also automatically reset all "releases" back to 1.

luizluca:
This commits prevents new package version when no commit touches package
files. It is limited to branches and does not change master behavior.

Adds GIT_COMMIT to be used as reference for package version. For master
branch, GIT_COMMIT will be HEAD (as before). For branches, GIT_COMMIT
will be the most recent common ancestor between the branch and master.

Adds GIT_COMMIT_COUNT as the number of commits inside each package
directory since GIT_COMMIT. For master, it should be always 1.

Adds LUCI_RELEASE as the global LuCI release control. It should be
bumped to generate a new release for all packages. For master, it should
be always 1.

PKG_RELEASE will be the sum of LUCI_RELEASE and GIT_COMMIT_COUNT.

svn logic was removed and git command simplified.

closes #2156 

Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>

### DIFF
--- a/luci.mk
+++ b/luci.mk
@@ -47,41 +47,37 @@ LUCI_MENU.theme=4. Themes
 LUCI_MENU.proto=5. Protocols
 LUCI_MENU.lib=6. Libraries
 
+LUCI_RELEASE=1
 
 PKG_NAME?=$(LUCI_NAME)
 
-PKG_VERSION?=$(if $(DUMP),x,$(strip $(shell \
-	if svn info >/dev/null 2>/dev/null; then \
-		revision="svn-r$$(LC_ALL=C svn info | sed -ne 's/^Revision: //p')"; \
-	elif git log -1 >/dev/null 2>/dev/null; then \
-		revision="svn-r$$(LC_ALL=C git log -1 | sed -ne 's/.*git-svn-id: .*@\([0-9]\+\) .*/\1/p')"; \
-		if [ "$$revision" = "svn-r" ]; then \
-			set -- $$(git log -1 --format="%ct %h" --abbrev=7); \
-			secs="$$(($$1 % 86400))"; \
-			yday="$$(date --utc --date="@$$1" "+%y.%j")"; \
-			revision="$$(printf 'git-%s.%05d-%s' "$$yday" "$$secs" "$$2")"; \
-		fi; \
+GIT_COMMIT=$(shell \
+	branch="$$(git rev-parse --abbrev-ref HEAD)"; \
+	if [ "$$branch" != "master" ]; then \
+		git merge-base $$branch master; \
 	else \
-		revision="unknown"; \
+		echo HEAD; \
 	fi; \
-	echo "$$revision" \
-)))
+)
+GIT_COMMIT_COUNT=$(shell git rev-list --count $(GIT_COMMIT)..HEAD -- $(CURDIR))
 
+PKG_VERSION=$(if $(DUMP),x,$(strip $(shell \
+        set -- $$(git log -1 --format="%ct %h" --abbrev=7 $(GIT_COMMIT)); \
+        secs="$$(($$1 % 86400))"; \
+        yday="$$(date --utc --date="@$$1" "+%y.%j")"; \
+        revision="$$(printf 'git-%s.%05d-%s' "$$yday" "$$secs" "$$2")"; \
+        echo "$$revision" \
+)))
 PKG_GITBRANCH?=$(if $(DUMP),x,$(strip $(shell \
-	variant="LuCI"; \
-	if git log -1 >/dev/null 2>/dev/null; then \
-		branch="$$(git branch --remote --verbose --no-abbrev --contains 2>/dev/null | \
-			sed -rne 's|^[^/]+/([^ ]+) [a-f0-9]{40} .+$$|\1|p' | head -n1)"; \
-		if [ "$$branch" != "master" ]; then \
-			variant="LuCI $$branch branch"; \
-		else \
-			variant="LuCI Master"; \
-		fi; \
+	branch="$$(git rev-parse --abbrev-ref HEAD)"; \
+	if [ "$$branch" != "master" ]; then \
+		echo "LuCI $$branch branch"; \
+	else \
+		echo "LuCI Master"; \
 	fi; \
-	echo "$$variant" \
 )))
 
-PKG_RELEASE?=1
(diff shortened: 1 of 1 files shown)


feckert:
@hnyman any comment on this feature request? I have no idea if this is necessary.

hnyman:
I feel that it is mostly usable by persons using "opkg upgrade" blindly, who unnecessarily get new versions for all LuCI apps due to the automatic versioning that causes opkg to think that there is something to be updated.  Hopefully that is still a marginal user group.  (In general, it makes no sense to update e.g. luci-base after each small string change. )

Most users just install the firmware and use it, or build firmware with current LuCI by themselves and thus get the updated LuCI anyway. 


So, I am not that sure that we need the change and have not yet tested it myself.

luizluca:
> Most users just install the firmware and use it, or build firmware with current LuCI by themselves and thus get the updated LuCI anyway.

@hnyman, if this is true, why bother with package repositories, opkg. This change targets those users that manage their router as any other linux box, installing extra packages and keeping them updated. I just don't know if it is a marginal or major part of OpenWrt users. It is, however, how a non-developer enterprise user might use it.

OpenWrt does not provide an easy "package changelog" to let users know what a package update changes. Those who cares needs to search inside git history (and it is not for normal users). This PR only tries to reduce the number of (useless) updates an admin will take a look.

Yes, it will help those who upgrade "blindly" (because it saves flash space) but mostly those who check what each update will affect their routers (because it saves time).

jow-:
I am very sorry for the long time it took to look into this. I opted for a simpler, more robust solution now which derives the package version from the last commit done to the package sources.

luizluca:
Thanks @jow-

hnyman:
Nice one, especially that git log magic with `:(exclude)po`

However, I looked at the buildbot downloadable packages and noticed something strange:
https://downloads.openwrt.org/snapshots/packages/arm_arm1176jzf-s_vfp/luci/

Most of the packages are dated `20.069.54281-3fb8e3d_all.ipk`, while some are newer.  There are no older version strings, even for apps that have not been updated for ages. That commit 3fb8e3d is a merge of weblate translations, so it should actually be excluded from package versioning.

I think that the reason is a "shallow" git clone, where the history is not being pulled in, but just the current repo status. Thus git log shows the original clone commit of the repo as the oldest change. I haven't checked (and actually can't figure out where to check that), but I guess that the buildbot LuCI source repo has been re-initialised at that date (early March).

That measn that a casual user will get different versions in his own local repo, as the git clone in scripts/feeds is shallow by default (`--depth 1`). The user will get the newest LuCI commit at the clone moment as the base, and all package versions will be that commit or later.

Example: I have my build system in Ubuntu 19.10 initialised in late 2019, so I get older versions for the same apps. From my build:
```
luci-lib-ip_git-20.039.36236-c2eeb09_arm_cortex-a15_neon-vfpv4.ipk
luci-lib-ipkg_git-19.290.42471-b3965b9_all.ipk
luci-lib-jsonc_git-19.317.29469-8da8f38_arm_cortex-a15_neon-vfpv4.ipk
luci-lib-nixio_git-19.354.79968-39196b0_arm_cortex-a15_neon-vfpv4.ipk
luci-mod-admin-full_git-19.290.42471-b3965b9_all.ipk
luci-mod-network_git-20.100.77013-c099344_all.ipk
luci-mod-status_git-20.100.77013-c099344_arm_cortex-a15_neon-vfpv4.ipk
luci-mod-system_git-20.100.77013-c099344_all.ipk
luci-proto-ipv6_git-19.307.61018-284918b_all.ipk
```

From buildbot downloads:
```
luci-lib-ip_git-20.069.54281-3fb8e3d_arm_cortex-a8_neon.ipk
luci-lib-ipkg_git-20.069.54281-3fb8e3d_all.ipk
luci-lib-jsonc_git-20.069.54281-3fb8e3d_arm_cortex-a8_neon.ipk
luci-lib-nixio_git-20.069.54281-3fb8e3d_arm_cortex-a8_neon.ipk
luci-mod-admin-full_git-20.069.54281-3fb8e3d_all.ipk
luci-mod-network_git-20.100.77013-c099344_all.ipk
luci-mod-status_git-20.100.77013-c099344_arm_cortex-a8_neon.ipk
luci-mod-system_git-20.100.77013-c099344_all.ipk
luci-proto-ipv6_git-20.069.54281-3fb8e3d_all.ipk
```

That means that anybody using "uci list-upgradable" command will get a confusing list of upgradable packages eventhough there are no actual upgrades (as the build was from the HEAD). Well, the same has been with the old versioning scheme, so no negative change there, but the solution is not perfect, either.

Nothing to be alarmed of, and nothing that can be easily fixed without changing the feeds clone command's default, but I just mention this observation in case there will be later questions & issues about the differing LuCI versions.

hnyman:
@jow- 

One more quirk from the new LuCI package versioning scheme: 

Buildbot seems to have re-cloned all feed repos this week, maybe on April 21st? 
As OpenWrt default feed update action is a shallow clone of the repo, now all LuCI .ipks in the buildbot snapshot download repo show version as 20.112... or 20.111... and nothing is older (on this architectures that have been built after that).

I think that there is no good solution for this, as I can only think that the buildbot's clone action might need to be made a bit less shallow. Maybe with depth of 500 commits or so, in order to preserve at least some versioning. Could you manually deepen the luci feed repo on buildbot? Or is that futile, if it gets re-cloned every now and then.

Private builds on the other hand show much older versions.
I fear that there will be confusion later, as people will start opkg upgrading stuff unnecessarily and they will wonder why their brand new build shows older version than the download repo.
(I noticed on my new Ubuntu 20.04 buildhost that the luci-ssl collection version seems now to be from year 2017, as I have fetched the longer history to the repo...)

PR #2260, opened 2018-11-08
Update proto_vpnc.lua

joaofvieira:
Sometimes  people only wants VPN connection to be a route to a specific network, not the default gateway for all traffic.
I've tested this on my router and works fine!


jow-:
Merged with slight changes in f4212285d - thanks!

PR #2266, opened 2018-11-10
luci-app-wireguard: German translation

Matthias84:
by Matthias Meißer


hnyman:
Please amend the commit:

* The commit title should be:
   ` luci-app-wireguard: German translation`
   (something like that, so that the app context is clear)

* You need to have a proper Signed-off-by line with your real name in the commit message.

jow-:
Merged with 096caef87 - thanks!

PR #2269, opened 2018-11-11
luci-app-vpnbypass - German translation

Matthias84:
by Matthias Meißer


jow-:
Merged in 30e9f671a - thanks!

PR #2275, opened 2018-11-12
luci-app-travelmate: update Japanese translation

musashino205:
Updated Japanese translations.

Signed-off-by: INAGAKI Hiroshi <musashino.open@gmail.com>


hnyman:
thanks.

PR #2279, opened 2018-11-14
luci-theme-rosy: Remove redundant statements and modify the characters

yglb:
luasrc: Remove redundant statements
luasrc: modify the characters in the "set Password" button

Signed-off-by: Yanlan Shen <yanlan.shen.@rosinson.com>


yglb:
@jow-

PR #2281, opened 2018-11-14
luci-app-mwan3: add/remove options

feckert:
- Declare module variables local
- Remove deprecated option local_source
- Add option rtmon
- Add option rtlookup


feckert:
Explanation:
Mwan3 scans the main routing table for connected networks. All routes are added to the **mwan3_connected** IPset and will not processed by mwan3 and so will not get mangled.

With the new list option **rt_table_lookup** you could instruct mwan3 to scan also this additional routing tables and treat them as connected networks an so do not mangĺe them. We've considered that to use this with IPsec. I thought that would also be useful for other things maybe.

PR #2285, opened 2018-11-14
luci-ssl-nginx: fix dependencies

dengqf6:
Remove libustream-openssl because Nginx doesn't depend on it.


hnyman:
But other https related packages may depend on that, so adding that small wrapper can still be useful, I think. Otherwise some other packages may try to pull in libustream-mbedtls (which is the default)

cc @Ansuel

Ansuel:
@hnyman @dengqf6 it looks like libustream is needed only by uhttpd so think we can drop that dependency

hnyman:
Looking at the source code, also (luci-app-)ddns and (luci-app-)dnscrypt may need at least some libustream version, so it is not quite that clearcut. But as nginx is non-standard, I guess that the users using that will then survive by installing the needed libustream versions manually.

However, as openssl-utils are still installed for key generation, then the large openssl library gets installed in any case and the few additional kilobytes from libustream-openssl wrapper package have no significance.

So, I am still a bit hesitant although you are right, as nginx does not directly depend on libustream-openssl.

PR #2287, opened 2018-11-15
https_dns_proxy.lua:Fix the datatype of "subnet address".

alphakk:
The datatype ip4prefix only accept numbers from 0 to 32,not subnet/mask(i.e. 192.168.1.0/24),this affects option "-e".
https://github.com/openwrt/luci/wiki/Datatypes#ip4prefix the discription is wrong,according to src
http://luci.subsignal.org/trac/browser/luci/branches/luci-0.10/libs/web/luasrc/cbi/datatypes.lua#L76


jow-:
Merged with slight changes in 8aeb78d1c - thanks!

PR #2289, opened 2018-11-16
luci-app-setup-wizard: add new application

rosysong:
This application is design for the users who are new to OpenWrt or the people want to use OpenWrt but do not have professional background.

Setting parameters of OpenWrt devices has never been so easy as it is now, this app can guide users to initialize their device within 2 steps, including Internet Access, WiFi. That's most family users need (gain pppoe account from isp and use it connect to internet).

## Notes
notes

## Internet Access
internet

## WiFi
wifi

## Done
done

Enjoy it!!!

### DIFF
--- a/applications/luci-app-setup-wizard/Makefile
+++ b/applications/luci-app-setup-wizard/Makefile
@@ -1,0 +1,15 @@
+#
+# Copyright (C) 2008-2014 The LuCI Team <luci@lists.subsignal.org>
+# Copyright (C) 2018 Rosy Song <rosysong@rosinson.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Setup Wizard
+LUCI_DEPENDS:=+luci-base
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-setup-wizard/luasrc/controller/setup-wizard/setup-wizard.lua
+++ b/applications/luci-app-setup-wizard/luasrc/controller/setup-wizard/setup-wizard.lua
@@ -1,0 +1,15 @@
+-- Copyright 2018 Rosy Song <rosysong@rosinson.com>
+-- Licensed to the public under the Apache License 2.0.
+
+module("luci.controller.setup-wizard.setup-wizard", package.seeall)
+
+function index()
+	entry({"admin", "system", "setup-wizard"}, form("setup-wizard/setup-wizard"),
+		_("Setup Wizard"), 3)
+	entry({"admin", "system", "setup-wizard", "internet"}, form("setup-wizard/internet"),
+		_("Internet Access"), 10).leaf = true
+	entry({"admin", "system", "setup-wizard", "wireless"}, form("setup-wizard/wireless"),
+		_("WiFi Configuration"), 20).leaf = true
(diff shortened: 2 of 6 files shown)


rosysong:
@hnyman @jow-

hnyman:
The app seems to take for granted that there is pppoe and a single radio.

pppoe is not used that often, so the first step will be confusing for many users. 
Similarly, there is no info about which radio is being setup. It varies by the router, if 2.4GHZ or 5GHz is radio0

I feel that this app tries to oversimplify things.

aparcar:
I think this is generally a great idea!

Please add a dialog for default password.

Regarding 2.4 and 5GHz, I'd suggest to offer a single ESSID/Password form for both with an additional checkbox a la "Enable 5GHz" - if actually supported by hardware. Different names for 2.4/5GHz make no real sense for non-techies, huh?

Also it would be interesting to have this pop up instead of the default Luci Login page, giving whoever the possibility to hand a freshly flashed router to non-technical-users.

rosysong:
@aparcar  Thanks, and I agree with your advice that replacing the LuCI login page when the firmaware is freshly flashed or return to firstboot status. This really can issue users, especially non-technical-users, to setup their unconfigured router.

feckert:
I think this is a very good proposal.
We can't cover all the possibilities.
I think we should start with the standard scenario and if something is missing we could add the missing config options.

aparcar:
I'd be very much in favor to have a feature like that! However the current approach relies on Lua whereas ubus call would be a cleaner and more generic solution. @rosysong could you refactor your code to use ubus?

rosysong:
It's been a long time since I push this PR, well, I will refactor the code.

And, would you please comment in the code view or tell me where should I replace the approach with ubus call?

aparcar:
Dear @rosysong , sorry for the long delay! Some time ago [I posted in the forum](https://forum.openwrt.org/t/common-use-case-for-simplified-web-interface/38866) for ideas of a simplified web interface. The idea is basically a LCi web interface doing multiple steps at once, instead of 1 by 1 configuring of options. Therefore, pretty much what your wizard does.

Regarding refactoring, I'd be happy to see some rpcd shell scripts (or later even C) modifying UCI values to the desired state. The web interface could then call those scripts with parameters (like WiFi name or password).

I'd be in favor to avoid Lua scripts at this state, if that would work for you. If not I'm sure those backend scripts are easily portable to other languages in the later run.

rosysong:
@aparcar  Glad to see that you noticed this.
1. Both of them can works fine whether use rpcd Shell scripts or Lua scripts. For the As the first version, making it work is the first priority.

2. I agree with you opinion that to refactor this PR, and to have this pop up instead of the default Luci Login page. I'll update my commit once I done.

aparcar:
> Both of them can works fine whether use rpcd Shell scripts or Lua scripts. For the first version, making it work is the first priority. On the other hand, as @feckert said, we can't cover all the possibilities, just for the most common scenario.

Other router vendors already created (from my point of view) [very good](https://docs.gl-inet.com/en/3/setup/mini_router/first-time_setup/) interfaces with a focus on use cases rather than fine granular configuration. Maybe this could be used as an inspiration. 

> I agree with you opinion that to refactor this PR. The "simple" interface you described looks better than the wizard I pushed. Do you mean to have this pop up instead of the default Luci Login page ? I'll update my commit once I done.

Again, my personal view, having two tabs like "simple" and "advanced" allowing to switch between the two modes. Therefore yes, after a login it could pop up per default.

PR #2297, opened 2018-11-16
luci-mod-system: fstab: fix table cell rendering without unmount button

namedun:
Before:  
image

After:  
image

### DIFF
--- a/modules/luci-mod-system/luasrc/model/cbi/admin_system/fstab.lua
+++ b/modules/luci-mod-system/luasrc/model/cbi/admin_system/fstab.lua
@@ -117,14 +117,14 @@ function used.cfgvalue(self, section)
 end
 
 unmount = v:option(Button, "unmount", translate("Unmount"))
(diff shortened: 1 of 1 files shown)


modules/luci-mod-system/luasrc/model/cbi/admin_system/fstab.lua
```diff
@@ -117,16 +117,20 @@ function used.cfgvalue(self, section)
...
+function unmount.cfgvalue(self, section)
 	if non_system_mounts[section].umount then
-		self.title = translate("Unmount")
-		self.inputstyle = "remove"
-        	Button.render(self, section, scope)
+		return true
 	else
-		luci.http.write("&#160;")
+		return false
 	end
```
jow-: You can condense that into just `return non_system_mounts[section].umount`

namedun:
Updated

jow-:
Merged, thanks!

PR #2306, opened 2018-11-19
luci-theme-rosy: Repair form bug

yglb:
htdocs: Modifying the form does not fully display the bug of the dropdown box
htdocs: Match css to implement related styles

Signed-off-by: Yanlan Shen yanlan.shen.@rosinson.com


yglb:
@jow-

PR #2307, opened 2018-11-19
luci-app-openvpn: more changes & fixes

dibdot:
* fix possible exception in template based ovpn creation
* remove needless shellquote function in controller,
  the filename will be checked on client side with JS
* enhance FileUpload behaviour in basic/advanced mode:
  - change "auth_user_pass" to FileUpload
  - cfg entries (even with default values) will be shown
  - existing entries are now removable (incl. file unlink),
    simply clear the appropriate textbox
* change "key_direction" option to boolean ListValue
* add "config" option to basic/advanced edit, to make it possible to change the upload path in LuCI

### DIFF
--- a/applications/luci-app-openvpn/luasrc/controller/openvpn.lua
+++ b/applications/luci-app-openvpn/luasrc/controller/openvpn.lua
@@ -18,7 +18,7 @@ function ovpn_upload()
 	local util   = require("luci.util")
 	local uci    = require("luci.model.uci").cursor()
 	local upload = http.formvalue("ovpn_file")
-	local name   = string.gsub(util.shellquote(http.formvalue("instance_name2")), "'", "")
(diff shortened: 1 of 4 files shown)


mwarning:
The default for key_direction is "1". Shouldn't this be an integer?

dibdot:
@mwarning you mean boolean!? Maybe a Flag or ListValue would fit better but I don't care ... most of the users will use the direct ovpn upload anyway and not configure zillions of openvpn options by hand.

mwarning:
@dibdot boolean would be better. Since this commit is a bit of a collection. I thought to mention this small "error".

dibdot:
OK, I'll change that later today ...

dibdot:
done, it's now a boolean ListValue.

PR #2310, opened 2018-11-20
luci-app-openvpn: sort basic and advanced options

mwarning:
Options sorted by name will make it easier to navigate the option lists
and selection boxes.

Signed-off-by: Moritz Warning <moritzwarning@web.de>


mwarning:
I cannot verify the syntax right now, since luci-openvpn seems to be broken in nightly (`/usr/lib/lua/luci/model/cbi/openvpn.lua:76: attempt to index field 'extedit' (a nil value)`).

mwarning:
A different approach would be to sort at runtime:
```
-- sort params by name
for i, val in pairs(knownParams) do
  table.sort(val[2], function(a, b) return a[2] < b[2] end)
end
```

mwarning:
Anyway, the current order looks random to me. Finding a specific option is a bit of a hassle.

dibdot:
@mwarning to be honest, I don't see any benefit in an alphabetical order - I would close this PR.

mwarning:
Right now I find myself to have to parse the dropdown list twice, once quickly and once slowly, because I had overlooked the setting I was looking for. From my own experience it makes it easier.

Feel free to close the PR if you feel unconvinced.

PR #2314, opened 2018-11-21
luci-app-opkg: fix and update Japanese translation

musashino205:
Fixed and updated Japanese translations.

Signed-off-by: INAGAKI Hiroshi <musashino.open@gmail.com>


musashino205:
I mistook a branch to use for PR...

PR #2316, opened 2018-11-21
luci-app-aria2: add button for to AriaNg

sotux:
As AriaNg is another web UI for aria2, I add a button to open AriaNg if installed package ariang or ariang-nginx


jow-:
Merged, thanks!

PR #2317, opened 2018-11-21
luci-app-openvpn: redirect to overview page on save/apply, hide boolean by default and cleanup code

feckert:
On save and apply redirect to the overview page. This is the default behavior on LuCI.
Also make some code cleanup.


dibdot:
@feckert I don't think that it is desirable to always jump back to the overview page while configuring zillions of openvpn options. There is already the breadcrumb trail to going back to overview or switching between different edit views really fast.

Anyway, your PR needs to be rebased to the latest changes.

feckert:
@dibdot I will rebase my changes :-)
From my point of view i think if you press **save and apply** then we should go back to the overview page.
If youe press only **save** then you will stay on the page.
Other modules work the same way.

dibdot:
Maybe just a personal "problem" - I always use "Save & Apply" ... anyway, thanks for the additional rebase efforts!

feckert:
@dibdot i have added another commit which will hide all boolean option per default.
Till now all boolean option are shown. This is confusing.

dibdot:
Thanks! Probably we'll never get this really bullt proof - let's see what other users say ... ;-)

feckert:
> Thanks!  Probably we'll never get this really bullet proof - let's see what other users say ... ;-)
@dibdot  You're probably right.

Thanks for merging

PR #2330, opened 2018-11-24
luci-app-shadowsocksr: Add luci-app-shadowsocksr to Luci

OneHappyForever:
Add luci-app-shadowsockr to Luci repo so users can install it directly from the Luci panel.

This app has significant advantages over the shadowsocks-libev version, including obfuscation, the ability to add multiple servers at once (instead of manually), and preventing DNS leaks and poisoning in China.

Signed-off-by: Josue Masson <Josh.tesol@gmail.com>


hnyman:
The Makefile looks strange. Overly complicated. Almost like it would be targeting life outside the LuCI repo.

The package also includes in addition to the LuCI GUI part the underlying package itself ??? and other stuff like the main config file "/etc/config/shadowsocksr etc" that should be in the main packages repo along the core package itself.

There is also .gitignore file, lmo2po compiler sources etc.

Here in LuCI repo should be only the GUI components, and the actual shadowsocksr package should be in the packages repo.

OneHappyForever:
Thanks for the feedback!

I'm kind of new at this so I still have to get the hang of it. Let me go through your suggestions and edit the pull request.

Didn't realize that the packages and the GUI were supposed to be in different repos.

hnyman:
> Didn't realize that the packages and the GUI were supposed to be in different repos.

Yes.
packages and their config files  are in https://github.com/openwrt/packages
and only the possible GUI is here in the LuCI repo.
So, all config files related to the actual package should also be in the other repo, so that the app in the pckage repo is complete. This LuCI repo should only contain the GUI realted stuff for manipulating the already existing config of the actual package.

yousong:
afaik, dev work of ssr has stopped for white some time and even the original repo long gone.

PR #2333, opened 2018-11-26
luci-theme-bootstrap: fix tabs rendering

namedun:
If we have a lot of tabs that do not fit in width, we must wrap
tabs to next line.

Before:  
image

After:  
image


jow-:
I implemented a slightly different variant (to allow the line to wrap as well) but thank you for the report!

PR #2335, opened 2018-11-27
luci-theme-material: add new favicon

shunjou:
Notably improves visibility on dark colored tabs.
  
Link to source files for reference: https://drive.google.com/file/d/1vSneA5LAMec6veAQH5rQ-cI_19XhDmz9


jow-:
Merged, thanks!

PR #2336, opened 2018-11-27
luci-app-openvpn: Add missing config options

feckert:
- Add missing ncp_ciphers option
- Add missing verify_client_cert option
- Add missing route_gateway option
- Add missing compress option
- Predefining the drodown for the cipher option
- Add missing ncp_disable option
- Add missing allow_recursive_routing option
- Add missing resolv_retry option


dibdot:
@feckert is it intended that "cipher" is a normal Value and not a ListValue?

jow-:
@dibdot - hm, I assumed this was done to still allow entering custom cipher values (in case the list is incomplete).

feckert:
> @dibdot - hm, I assumed this was done to still allow entering custom cipher values (in case the list is incomplete).

exactly

@dibdot @jow- 
In the future i want to get the values from the openvpn itself with the following commands
**openvpn --show-ciphers** for **ciphers** option
**openvpn --show-tls** for the **tls-ciphers** option

What are your opinions?

dibdot:
@feckert @jow- Thanks, sounds reasonable!

Regarding your question: I think the list filling via program is OK in general (no auto-refresh), but please keep in mind that the current luci frontend package has no openvpn dependency specified.

feckert:
@dibdot thanks for the hint i will add openvpn do the dependency as well.
I think next week a nother PR with the changes is on the way

PR #2337, opened 2018-11-27
luci-lib-httpclient: Comply with RFC 2616 standard for Host header

hwinwuzhere:
Maintainer: @jow- 
Compile tested: ar71xx, GL-iNet AR750, OpenWrt master
Run tested: ar71xx, GL-iNet AR750, OpenWrt master

Description: According to the [RFC standard](https://tools.ietf.org/html/rfc2616#section-14.23) a Host header should always contain the host and the port. Also an IPv6 address could specify an interface (like %eth0). This commit adds the missing port to the Host header and ensures that IPv6 addresses are stripped of their interface before turned into a Host header.

Signed-off-by: hwinwuzhere <hwinwuzhere@hotmail.com>

### DIFF
--- a/libs/luci-lib-httpclient/luasrc/httpclient.lua
+++ b/libs/luci-lib-httpclient/luasrc/httpclient.lua
@@ -102,7 +102,9 @@ function request_raw(uri, options)
 		uri = uri .. '?' .. http.urlencode_params(options.params)
 	end
 
+	local ipv6_enabled = false
 	if uri:find("%[") then
+		ipv6_enabled = true
 		if uri:find("@") then
 			pr, auth, host, port, path = uri:match("(%w+)://(.+)@(%b[]):?([0-9]*)(.*)")
 			host = host:sub(2,-2)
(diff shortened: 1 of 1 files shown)


libs/luci-lib-httpclient/luasrc/httpclient.lua
```diff
@@ -162,7 +164,12 @@ function request_raw(uri, options)
 
 	-- Pre assemble fixes	
 	if protocol == "HTTP/1.1" then
-		headers.Host = headers.Host or host
+		if ipv6_enabled then
+			local ipv6_host = host:gmatch("[%w\\:]+")()
```
jow-: Why is the gmatch needed? Can't we simply use `host` directly?

hwinwuzhere: Good question!
The host string might contain an interface, e.g. `fe80::1%eth0` however this is not accepted as a Host header by HTTP. The gmatch removes the interface in case this interface was added.

jow-: Thanks, I only now noticed that you already explained it in the intro text.

jow-:
@hwinwuzhere - thanks again for your contribution. After thinking about it for a while I decided to fix the issue in a more complete manner in commit https://github.com/openwrt/luci/commit/a54baf7be0d293b50c6b2754c11d61e6e2ad82af

PR #2344, opened 2018-11-29
luci-theme-material: minor changes

shunjou:
9603990: This was missing from my previous commit that added a new favicon. Also removed other lines equivalent to useless fluff.

ba6f7b6: Same image, but sourced from the svg so the downscaled result is cleaner. Also uses more efficient png compression.


jow-:
Merged, thanks!

PR #2345, opened 2018-12-01
luci-app-iot: IoT helper API for OpenWRT/Luci

tjkolev:
Maintainer: TJ Kolev
Developed and tested: OpenWRT VM VirtualBox; OpenWrt 18.06.1 / LuCI openwrt-18.06
Deployed and tested: TP-Link TL-WR1043N/ND v1; LEDE Reboot 17.01.4 / LuCI lede-17.01
Description: A simple application intended to support home IoT devices by providing
an API to retrieve configuration parametrs or send email notifications. A UI page
can be used for set up. See the readme file for more details.

Signed-off-by: TJ Kolev <tjkolev@gmail.com>


jow-:
`luci-mod-iot-helper` sounds fine to me. Can you move/rename it accordingly?

tjkolev:
All right then. Hopefully I'll find the time to do it within the next couple of weeks.
Thank you.
tjk :)

stokito:
Why not to use a regular ubus over http and rpcd scripts? This api has nothing related to luci itself except of config screen.
Rpcd already have a file api so your devices can fetch any config.
The sending of email can be implemented with bash script in lambda/serverless style.
Also email is in fact too over complicated for such things. You can just send a post request to Telegram bot or even make a browser notification push.
Ubus over http has some basic authorization with tokens which can be easily improved to be OAuth compatible.
But it uses JSON-RPC which may be even to complicate for IoT devices.
Are you working in the IoT area?
Because I don’t know what is needed for the typical devices. But I know that they using mostly MQTT.

What this reminds for me is Parse/Firebase Backend As Service (BaaS) and in fact this is something that can be widely used even for creating a homepage on the router.

tjkolev:
@stokito, hi!

I am not working in the IoT area, not even close. This is my first attempt working with these microcontrollers (NodeMCU ESP8266 in my case). What I did works for me, so I thought I'd share. I didn't do much research, and picked what I thought would be the easiest for me to implement. Yes, MQTT is popular. But I figured that requires more infrastructure, and perhaps reaching out of my network, which I didn't want to do. Wasn't aware of the ubus until now.

Happy New Year!
tjk :)

Issue #2258, opened 2018-11-07
Display problems (Material theme, IE)


olimpiumarius:
Hello,
I want to show you possible display problems that persists on the Internet Explorer version of windows 10.
The problem persists in several submenus design Material, only on the Internet Explorer browser!
![openwrt3](https://user-images.githubusercontent.com/34411358/48148982-b4864a80-e2c3-11e8-9266-959ccda22a5f.PNG)

![openwrt](https://user-images.githubusercontent.com/34411358/48148601-d9c68900-e2c2-11e8-8f3d-61922b1ddeaf.PNG)
![openwrt1](https://user-images.githubusercontent.com/34411358/48148602-d9c68900-e2c2-11e8-85bb-d421fc98be0e.PNG)

jow-:
ping @LuttyYang

olimpiumarius:
done!

jow-:
I guess this is fixed now.

Issue #1795, opened 2018-05-15
Mount Points / Mounted file systems: Styling issue


he-ma:
There seems to be a little style problem which takes effect on certain themes. The attached picture should show what I mean. If no button is rendered (for system mount points) the table is truncated at the last field. Would be more nicer if all table rows have the same length.

![ipsec_status_virtual_access_bildschirmfoto3](https://user-images.githubusercontent.com/26056858/40045830-a68f983c-582b-11e8-8f4d-69749a02d361.jpg)

Code position where the button is created:
https://github.com/openwrt/luci/blob/f3b30512361584c61c9daa51650ad833072aa919/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/fstab.lua#L119-L125

The view which is used for the button:
https://github.com/openwrt/luci/blob/master/modules/luci-base/luasrc/view/cbi/button.htm

jow-:
This is a bug with the material theme I think. Iirc it has some global CSS rule to hide all empty fields, which would apply to empty `td` as well.

he-ma:
Hello @jow-
I did a research again and found out that the real problem is that there is no \<td\>\</td\> at the end of the line nor is a "hidden" button created if no button rendering is done (in button.htm).

So I modified button.htm to the following:

```
<%+cbi/valueheader%>
	<% if self:cfgvalue(section) ~= false then %>
		<% if self.inputstyle ~= "hidden" then %>
			<input class="cbi-button cbi-input-<%=self.inputstyle or "button" %>" type="submit"<%= attr("name", cbid) .. attr("id", cbid) .. attr("value", self.inputtitle or self.title)%> />
		<% else %>
			<input class="cbi-button cbi-input-button" type="hidden" />
		<% end %>
	<% else %>
		-
	<% end %>
<%+cbi/valuefooter%>

```
And modifying fstab.lua (here starting at line 119) to the following:

```
unmount = v:option(Button, "unmount", translate("Unmount"))
unmount.render = function(self, section, scope)
	if non_system_mounts[section].umount then
		self.title = translate("Unmount")
		self.inputstyle = "remove"
        	Button.render(self, section, scope)
	else
		self.inputstyle = "hidden"
		Button.render(self, section, scope)
	end
end

```
This works.

Maybe it's possible to extend the button class to use a type argument which defaults to submit, but can be modified to hidden? Alternatively could you kindly merge these modifications into the source?

feckert:
[...]

Issue #1231, opened 2017-07-10
[Material theme] Login form labels too narrow for Russian locale


K900:
Right now, the form looks like this when used with the Russian locale:

![](https://i.imgur.com/QQ2qkpz.png)

Making the labels wider produces a much nicer result:

```css
.node-main-login > .main .cbi-value-title {
    width: 10rem;
}
```

![](https://i.imgur.com/7UjclEs.png)

feckert:
should be fixed

PR #2347, opened 2018-12-01
luci-theme-material: partial style overhaul

shunjou:
The material theme is a mess in its current state. This attempts to solve a lot of the problems. A total rewrite would be the proper solution, but I think this is a good compromise.
  
It also reverts the removal of color fill to the buttons since the theme originally seemed to rely on color vibrancy to emphasize button actions. It already barely follows "material design" spec as it is and I don't think heavy use of empty buttons is a good idea without fancy animation that requires heavier js (dynamic ripple) and other significant changes to the button scheme.

Issues #1231, #1795, and #2258 can be closed.

### DIFF
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -428,13 +428,16 @@ header > .fill > .container {
 	user-select: none;
 }
 
-header > .fill > .container > img {
-	width: calc(0% + 10rem);
-	margin: -.1rem 3.5rem 0 1.5rem;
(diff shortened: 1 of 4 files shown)


shunjou:
I agree with you entirely, but then it approaches the point where I'd sooner opt for a total rewrite, which I'd rather not attempt at this time. I already held back from several other changes that would break its foundation, such as d07b3bc which would require significant rescaling of nearly ever text element. The fact that the original author decided to base the style on a non-english font is telling enough of its fundamental mess.

This is just intended as a "style overhaul". Someone else is more than welcome to submit similar changes for other sections of the theme in additional PRs.

PR #2350, opened 2018-12-04
luci-mod-system: add translation for led trigger names

feckert:
Not every user knows which trigger does what. To make this clear, the
triggers are now translated with a short help text.


LipkeGu:
lgtm

PR #2351, opened 2018-12-04
modules: use configurable poll interval

feckert:
If an interval configuration option is set under luci.main.interval,
then use this one.
If this is not specified then use the default specified in the source.

### DIFF
--- a/applications/luci-app-adblock/luasrc/view/adblock/logread.htm
+++ b/applications/luci-app-adblock/luasrc/view/adblock/logread.htm
@@ -10,7 +10,7 @@ This is free software, licensed under the Apache License, Version 2.0
 //<![CDATA[
 	function log_update()
 	{
-		XHR.poll(5, '<%=luci.dispatcher.build_url("admin", "services", "adblock", "logread")%>', null,
+		XHR.poll(-1, '<%=luci.dispatcher.build_url("admin", "services", "adblock", "logread")%>', null,
 		function(x)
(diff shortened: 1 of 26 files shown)


jow-:
Hmm, instead of replicating this everywhere I feel that it should got into `XHR.poll()` directly. Means when a value like `-1` or `0` is passed, we use the default value.

We could use the newly introduced `luci.js` initialization (https://github.com/openwrt/luci/blob/master/modules/luci-base/luasrc/view/header.htm#L16) to store the default value in the client side JavaScript context, then retrieve it from there (using `L.env.pollinterval`)

feckert:
@jow- I have update my PR with your suggestions.
The new poll interval will only be used if the first argument value of XHR.poll is lower or equal zero.
And if no poll interval option is set in uci the default value is set to 5 seconds which is the value most function uses.

Shouldn't I change the XHR.poll functions which use also 5 seconds?
They could be set to **-1** and then will also use the default value from L.env.pollinterval?

jow-:
Yep, I suggest to also change all `XHR.poll(5, ...)` to `XHR.poll(-1, ...)` calls.

feckert:
@jow- I have update the PR so that **all** XHR.poll functions who use the default value of 5 seconds using now the poll interval value from the config and if this is not set/found a default value of 5 seconds.

jow-:
Merged, thanks @feckert

PR #2353, opened 2018-12-06
luci-theme-material: refactor the luci-theme-material without jQuery.

brvphoenix:
Changelog:
1. Refactored the js codes all by DOM instaed of jQuery. 
2. All the features of material theme were reserved except the fadeOut and fadeIn effect when loading.
3. Some coeds removed which might be useless in the master and 18.0.1 branch.

Signed-off-by: Van Waholtz <vanwaholtz.gmail.com>


shunjou:
More than just the fade effects on loading were not preserved. The menu animations and dark mask that the side bar relies on to dim the main content on lower res were dropped. Although faster with no animation and lighter without the external library, this is incomplete and the effect is jarring. Ideally, the animations should be toned down but not entirely removed.

themes/luci-theme-material/luasrc/view/themes/material/header.htm
```diff
@@ -200,8 +210,79 @@
...
+
+			if (!document.querySelector('body').classList.contains('logged-in')) {
+				document.querySelector("body").classList.add("node-main-login");
+			};
+
+			var showSide = false;
+			document.querySelector(".showSide").addEventListener('click', function () {
+				if (showSide){
+					document.querySelector(".main-left").setAttribute("style", "width: 0");
+					document.querySelector(".main-right").setAttribute("style", "overflow-y: auto");
```
shunjou: `overflow-y` should be `visible` here, for mobile browsers to properly handle the scrollbar

themes/luci-theme-material/luasrc/view/themes/material/header.htm
```diff
@@ -200,8 +210,79 @@
...
+					showSide = false;
+				}else{
+					document.querySelector(".main-left").setAttribute("style", "width: 15rem");
+					document.querySelector(".main-right").setAttribute("style", "overflow-y: scroll");
```
shunjou: `overflow-y` should be `hidden` here, which makes sense when the dark mask is applied

shunjou:
I don't think you understand the way that the page is supposed to adapt as the resolution changes. Simply resizing the browser window to lower widths will demonstrate it.

1. The darkmask layer's purpose was to dim main-right's content when the main-left side bar is expanded. This only happens on lower width resolutions, such as on mobile devices, where main-left gets hidden and the showside menu button is visible.

2. Using `visible` instead of `auto` hides the scrollbar on mobile browsers, where page width is a premium. The `scroll` value is more than just whether the page can be scrolled or not; it makes the scrollbar visible at all times. Also, you break half the point of having a collapsible side bar menu if both it and the obscured page can both be scrolled at the same time while it is expanded.

3. What this change currently results in contradicts what the theme originally intended.

brvphoenix:
1. I think you should reread the script.js file for this theme and the jQuery.
2. The two parts can't be able to scroll at the same time and the html focus has only one. Besides, removing the daskmark, enabling the the scroll and etc is my former intension which I have said. If you think it is necessary, I will revert it back.
3. Expect of removing the animation effect including fadeIn, fadeOut, slideDown, slideUp and etc is temporal, I have already said I need more suggestions to decide whether revert it back considering the too much codes added. I don't think any other changes currently results in contradicts what the theme originally intended. You should realize that most the features are controlled by the css file which I have made no change.

I think, what we talk make no sense, if you involve others which have nothing to do with whether the animation effect and daskmark layer should be removed.

shunjou:
1. You're removing the overlay to dim the content that should be immutable. That is wrong. The script may have allowed for clicking it as another way to collapse the main-left side bar other than clicking the showside menu button, but that isn't its primary role.
2. I wrote "at the same time" but I didn't mean simultaneously with the same action. I'll edit my comment above to remove that. My point is that on low res when the side bar is expanded, both main-left and main-right are completely visible and scrollable. This is worsened by the fact that, without darkmask, they blend seamlessly since their backgrounds are both #fff.
3. I'll give you a relevant example so you can understand my point. Android is made by Google, who also made material design which this theme is supposedly based on. Within Android's developer options menu, there exist three settings to configure the animation scale. While reducing the scale may appear to boost performance, completely turning it off breaks the fluency of transitions. Animations are intended to serve a purpose in design; they are not merely cosmetic. As I said in my first comment, I'd be all in favor of reducing the animation scale to something like .5x but not removing it entirely.

PR #2359, opened 2018-12-09
luci-mod-status: add tagged includes to "Overview" page

namedun:
This allows to determine specific place on the "Overview" page for
the include. Including place is specified by tag in the filename of the
include in format:
```
[<tagname>]-custom-file-name.htm
```

Predefined known tags are:
  * before@system - place before "System" section
  * after@system - place after "System" section
  * before@memory - place before "Memory" section
  * after@memory - placte after "Memory" section
  * before@swap - place before "Swap" section
  * after@swap - place after "Swap" section
  * before@network - place before "Network" section
  * after@network - place after "Network" section
  * before@dhcp - place before "DHCP leases" section
  * after@dhcp - place after "DHCP leases" section
  * before@dsl - place before "DSL" section
  * after@dsl - place after "DSL" section
  * before@wireless - place before "Wireless" section
  * after@wireless - place after "Wireless" section
  * before@wireless-assoclist - place before "Associated Stations" section
  * after@wireless-assoclist - place after "Associated Stations" section
  * atend - place at end of the page (old behavior)

Files without tags will be placed at end of the page. Thus, the existing
functionality does not violated by this commit.

### DIFF
--- a/modules/luci-mod-status/luasrc/view/admin_status/index.htm
+++ b/modules/luci-mod-status/luasrc/view/admin_status/index.htm
@@ -127,11 +127,75 @@
 
 		return
 	end
+
+	local idx_includes = {}
+	local idx_knowntags = {
+		"before@system",
+		"after@system",
+		"before@memory",
+		"after@memory",
+		"before@swap",
+		"after@swap",
+		"before@network",
+		"after@network",
+		"before@dhcp",
+		"after@dhcp",
+		"before@dsl",
+		"after@dsl",
+		"before@wireless",
+		"after@wireless",
+		"before@wireless-assoclist",
+		"after@wireless-assoclist",
+		"atend"
+	}
+
+	local incdir = util.libpath() .. "/view/admin_status/index/"
+
+	if fs.access(incdir) then
+		local inc
+		for inc in fs.dir(incdir) do
+			if inc:match("%.htm$") then
+				local tag = inc:match('^%[(.*)%]')
+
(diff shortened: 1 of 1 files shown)


jow-:
This looks overly complex. Whats the use case for this?

namedun:
@jow- The simplest use case of this is that any custom LuCI application can choose a place on the page where to display information. For example, we may display some important messages from custom application at the top of the page, rather than at the bottom, where they may not be seen.

Another example is a custom LuCI applications for displaying extended information about switch ports. Information from this application would be logical to display about the "Network" section, and not at the very bottom of the page.

jow-:
I wonder if it would be simpler to split the existing sections up into partial includes and simply numerically prefix them. That would also be more in line with the way OpenWrt commonly handles partial configs.

namedun:
In fact, this is exactly what I did. Just the parts are indicated by a tag in the name of the included file.
A numerical prefix can be added immediately after tag.

Perhaps it would be better to use subfolders to indicate the parts? But in this case, applications using this feature will not work correctly with the older versions of LuCI. In the current implementation, applications using this feature will work with the old version of the LuCI, where they will simply be included at the bottom, as before.

jow-:
I'd rather like to split the existing status page into a series of partials. I do not like the rather convoluted tag system.

namedun:
Ok, I will think how it can be better implemented.

jow-:
I can also look into it tonight or tomorrow.

namedun:
@jow- I tried to make a different implementation of this feature in PR #2364. Please look at that.

PR #2361, opened 2018-12-09
CONTRIBUTING.md

stokito:
Little bit updated the CONTRIBUTING.md file

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,24 +11,24 @@ Regardless of whether you send a patch or open a pull request, please try to fol
     (E.g.: "luci-mod-admin-full: fix wifi channel selection on multiple STA networks")
 * Shortly explain the changes made and - if applicable - the reasoning behind them
 * Commit message of each commit should include a Signed-off-by line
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -53,7 +53,7 @@ Pull requests are the easiest way to contribute changes to git repos at Github.
 
 ## Release Branches:
 
-* Branches named "for-XX.YY" or "luci-X.Y" (e.g. "for-15.05") are release branches.
+* Branches named `for-XX.YY` or `luci-X.Y` (e.g. `for-15.05`) are release branches.
```
thess: Going forward, I believe branch names will be in the form: `openwrt-xx.yy` (Ex: `openwrt-18.06` is current release branch.)

thess:
Thanks for the update - see comments

CONTRIBUTING.md
stokito: so, should I update it?

thess: That was the idea.

PR #2362, opened 2018-12-10
luci-app-opkg: update Simplified Chinese translation

sotux:
Signed-off-by: Zheng Qian <sotux82@gmail.com>

### DIFF
--- a/applications/luci-app-opkg/po/zh-cn/opkg.po
+++ b/applications/luci-app-opkg/po/zh-cn/opkg.po
@@ -1,9 +1,10 @@
 #
 # Yangfl <mmyangfl@gmail.com>, 2018.
+# Zheng Qian <sotux82@gmail.com>, 2018.
 #
 msgid ""
 msgstr ""
-"PO-Revision-Date: 2018-08-18 12:39+0800\n"
+"PO-Revision-Date: 2018-12-11 09:07+0800\n"
 "Last-Translator: Yangfl <mmyangfl@gmail.com>\n"
 "Language-Team:  <debian-l10n-chinese@lists.debian.org>\n"
 "Language: \n"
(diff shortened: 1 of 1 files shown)


jow-:
@sotux - thank you! Some of the translated strings contain two consecutive `…` (e.g. `升级……`). Could you fix this up? Otherwise this is good to merge.

applications/luci-app-opkg/po/zh-cn/opkg.po
```diff
@@ -31,6 +32,9 @@ msgid ""
 "custom repository entries. The configuration in the other files may be "
 "changed but is usually not preserved by <em>sysupgrade</em>."
 msgstr ""
+"下面列出了 <em>opkg</em> 使用的各种配置文件。使用 <em>opkg.conf</em> 进行全局"
+"设置，使用 <em>customfeeds.conf</em> 进行自定义仓库。其他文件中的配置可能会更"
+"改，但通常不会由 <em>sysupgrade</em> 保留。"
```
yousong: 后半段需要修饰一下。比如

```
以下列出了<em>opkg</em>所使用的各个配置文件。<em>opkg.conf</em>用于全局配置，<em>customfeeds.conf</em>用于描述自定义仓库。其它配置文件的变更通常默认不会被<em>sysupgrade</em>保留。
```

yousong:
@jow- I may chime in here.  The six dots with two "…" is actually valid use of ellipsis of Chinese writing [1], though it's not uncommon these days people just use three dots.  The key is consistency of translation across different files.

 [1] https://en.wikipedia.org/wiki/Ellipsis#In_Chinese

jow-:
I didn't know, fine for me then.

sotux:
Hi @jow- , Hi @yousong , I've updated my commit.
Adjust some translations and replace "……" with "…"，because I reviewed the luci-base translation  and keep consistency of it.
Thank you for your review.

yousong:
Thanks!

PR #2363, opened 2018-12-10
luci-app-aria2: fix http/https protocol when opening aria2 webUI

sotux:
In the openWebUI function, the url is assumed as http protocol,
this will cause problem when we login luci-admin with https.

The variable localhostPath has already contained http/https
protocol, so we can use it directly and add the path of aria2
webUI.

Signed-off-by: Zheng Qian <sotux82@gmail.com>


jow-:
Merged, thanks!

PR #2364, opened 2018-12-10
luci-mod-status: split status page into a series of partials

namedun:
Split status page into a series of partials as discussed in PR #2359.


jow-:
That’s exactly what I had in mind, thanks! If we need more fine grained placing we can also split the headlines into separate partials.

PR #2367, opened 2018-12-11
luci-theme-rosy: Fixed insufficient container container width

yglb:
luci-theme-rosy: Fixed insufficient container container width

Signed-off-by: Yanlan Shen <yanlan.shen.@rosinson.com>


yglb:
@jow-

PR #2365, opened 2018-12-10
luci-base: fix typo in template

K900:
Makes tabs actually render correctly

PR #2368, opened 2018-12-11
Fix a bug when loading luci translation

heunghingwan:
Fix errors caused by an extra brackets
/usr/lib/lua/luci/template.lua:74: Failed to load template 'cbi/map'.


musashino205:
same fix as #2365

heunghingwan:
Sorry, didn't check before opening pull request

PR #2371, opened 2018-12-12
luci-base: clean up loading spinner icon

shunjou:
Result is much cleaner on dark backgrounds. Optimized with gifsicle.


jow-:
Hmm, I compared the old one with the proposed gif and imho the end result looks worse even on bright backgrounds while being a fair bit larger:

compare

Given that most direct uses of `loading.gif` through `` tags have been eliminated and that it is mostly used as background by CSS rules now, would it be a viable option to ship an alternative dark-optimized gif along with the themes that need it?

LipkeGu:
Dont forget the SoB line! (https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md)

PR #2383, opened 2018-12-20
luci-theme-material: make top level menu fontcolor configurable

feckert:
If you change the color of the submenues then the colors of the toplevel
should also be changed into the same color.


shunjou:
The main menu title is intended to be darker than the sub menus. Look at the side bar menu [here](https://material.io/design/introduction/#principles) for example.

If you want to make a variable for the menu title color, it could be something like `--main-menu-color`, but otherwise #202124 was correct.

shunjou:
@feckert @jow- ACK?

feckert:
sorry I overlooked it.

shunjou:
@feckert Sure, but is it agreed that this new variable should be #202124?

PR #2384, opened 2018-12-21
luci-app-ddns: check existence of nslookup util rather than making DNS request

davidjb:
This saves a DNS request whenever the nslookup check is performed.  For systems with nameservers configured in `/etc/resolv.conf`, nslookup will send a request to those servers (eg bypassing /etc/hosts), thus taking a lot more time than necessary.  For nameservers that are remote, this will take a *lot* longer than just checking if the nslookup util exists.

This change follows in the same vein as all the other checks that currently exist in the file which just use `which` without actually invoking each util's functionality.

Signed-off-by: David Beitey <david@davidjb.com>

### DIFF
--- a/applications/luci-app-ddns/luasrc/tools/ddns.lua
+++ b/applications/luci-app-ddns/luasrc/tools/ddns.lua
@@ -103,7 +103,7 @@ function env_info(type)
 		end
 
 		local function has_nslookup()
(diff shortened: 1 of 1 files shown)


jow-:
I think the intention of the original code was to look for a non-bugged version of `nslookup`, hence the check for the `(null)` output. But since any recent OpenWrt or LEDE ships with an improved nslookup in busybox, I guess we can merge it.

Will wait a bit to see if someone else chimes in.

dibdot:
I think it's OK to merge - thanks!

PR #2391, opened 2018-12-23
luci-app-uhttpd German translation

Matthias84:
German translation by Matthias Meißer

### DIFF
--- a/applications/luci-app-uhttpd/po/de/uhttpd.po
+++ b/applications/luci-app-uhttpd/po/de/uhttpd.po
@@ -1,0 +1,253 @@
+msgid ""
+msgstr "Content-Type: text/plain; charset=UTF-8"
+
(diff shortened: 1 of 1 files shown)


hnyman:
It should not be a .pot (template) file, but a normal .po

aparcar:
@hnyman so we only need to move the file from `.pot` to `.po`?

hnyman:
Basically so, and of course the commit should also be signed off.

feckert:
Please merge the two commits. So we have only the first one
Smal howto:
`git checkout <your pr branch>`
`git rebase -i HEAD~2`
in the vi replace the last commit `pick` with `fixup` and save the change
`git push --force`

Also add add a  `Signed-off-by: xxx xxx <xxx@xxx>` to the first commit

After this change we are ready to merge.

Matthias84:
Thanks for pointing me to the instructions! Usually I just use github to do some translations, but I wasn't clever enough to do it via frontend.

Issue #846, opened 2016-11-12
download files from luci ( use LuCI as file share )


mirjafarabedi:
I've uploaded some files in my OpenWrt root , /etc , /tmp directories.
I want user to be able to download uploaded files from from the above directories .
How can i **give a download URL to user** , to download that files ?

hnyman:
You should not give them access to /etc and root directories, as then they can also see your settings and other private stuff.

By default LuCI allows read access to stuff under /www directory
You should place the public stuff there.

Note that you can utilise symlinks, as explained here:
https://forum.openwrt.org/viewtopic.php?pid=298615#p298615

> Just create a symlink to the directory you want to share under /www like this:
> 
> `# ln -s /path/to/shared/folder /www/sharename`
> 
> And you can access it like this: http://192.168.1.1/sharename

mirjafarabedi:
I know this security risk (download from /etc and ...) . But in my app i monitor some sensor information and save them in /tmp folder.(in memory). and then i want access user download this exported info after authenticating with his username and password.Is this not possible ? (Thanks in advance for your response)

mirjafarabedi:
Unfortunately your answer doesn't work for me

hnyman:
Works for me, but naturally only from LAN. If you want it to work from wan, you need to adjust firewall etc.

Below is example of outputting tail of kernel log to file /www/test/test.txt and then how it is visible in the browser:

```
root@lede:~# cd /www
root@lede:/www# mkdir test
root@lede:/www# cd test/
root@lede:/www# dmesg | tail > test.txt
root@lede:/www/test# cat /www/test/test.txt
[67087.447386] [<c021f9bc>] (handle_IPI) from [<c0209474>] (gic_handle_irq+0x90/0xbc)
[67087.454677] [<c0209474>] (gic_handle_irq) from [<c020a594>] (__irq_svc+0x54/0x70)
[67087.462130] Exception stack(0xdd471f80 to 0xdd471fc8)
[67087.469712] 1f80: 00000001 00000000 00000000 c020b460 dd470000 c07c0480 c06d0818 00000000
[67087.474754] 1fa0: c07ba224 dd471fd8 00000000 c07bc100 1d25e000 dd471fd0 c021a0f0 c021a0f4
[67087.482885] 1fc0: 60000013 ffffffff
[67087.491043] [<c020a594>] (__irq_svc) from [<c021a0f4>] (arch_cpu_idle+0x38/0x5c)
[...]

PR #2404, opened 2018-12-28
luci-app-rosy-file-server: add new application

rosysong:
This application allows users to deploy a download server over http on their openwrt machine without installing external packages like samba, ftp, tftp. And this application takes up almost no memory unless users download files.

I wanted to add a http file server application so that we can deploy a download server on an openwrt machine and people can get the files without login to my machine. Unfortunately, the download action failed with below issue (can not access without authentication):

can-not-access

can-not-access-without-login
Is there any solution for this ?? thanks @jow- @hnyman 

So it turned to be a internal application. like this.

screen-file-server

### DIFF
--- a/applications/luci-app-rosy-file-server/Makefile
+++ b/applications/luci-app-rosy-file-server/Makefile
@@ -1,0 +1,15 @@
+#
+# Copyright (C) 2008-2014 The LuCI Team <luci@lists.subsignal.org>
+# Copyright (C) 2018 Rosy Song <rosysong@rosinson.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI Support for Rosy File Server
+LUCI_DEPENDS:=+luci-base +rosy-file-server
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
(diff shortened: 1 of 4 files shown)


hnyman:
Have you noticed discussion in #846 ?

LuCI allows download from /www

rosysong:
@hnyman  Thanks, it's done and seems works fine now. 

share-document-root

rosysong:
Here is the newer commit,
ChangeLog:
1. allow users to set the configuration of rosy file server.

jow-:
This application allows unauthenticated root level access to arbitrary files on the system, this looks like a huge security issue to me.

Please fix the `action_download()` procedure to actually enforce the `option target` of `/etc/config/rosyfs`.

rosysong:
@jow-  the ```action_download()``` is useless lines, sorry !
Please don't hesitate to tell me when you find any issues. Thanks!
Here is the newer commit, https://github.com/openwrt/luci/pull/2664

moocan:
Hello,

I'm trying Rosy File Server.
- Is it possible to keep luci administration portal as default web page and not access directly the shared files list ?
- is it possible access the shared file list after authentication ?

Kind Regards

firewire10000:
I've just installed luci-app-rosy-file-server - git-20.142.51595-0c6fd62-1 onto a Linksys EA6350v3 running v19.07.3 and it seems a little broken. I don't have any button in the bottom right corner to switch to the LuCI administration so when accessing my router it is going straight to the Rosy File Server

arky:
> I've just installed luci-app-rosy-file-server - git-20.142.51595-0c6fd62-1 onto a Linksys EA6350v3 running v19.07.3 and it seems a little broken. I don't have any button in the bottom right corner to switch to the LuCI administration so when accessing my router it is going straight to the Rosy File Server.
> 
> Any ideas?

@firewire10000  I have opened a new bug for this issue here: https://github.com/openwrt/luci/issues/5288

PR #2409, opened 2018-12-30
treewide: fix redirecting to System->Software

yousong:
By replacing url "admin/system/packages" with "admin/system/opkg"

While at it, also remove query arguments as they have been invalid for quite a while.

Signed-off-by: Yousong Zhou <yszhou4tech@gmail.com>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/network.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/network.js
@@ -113,7 +113,7 @@ L.poll(5, L.url('admin/network/iface_status', networks.join(',')), null,
 					var e = document.getElementById(ifc.id + '-ifc-edit');
 					if (e) e.disabled = true;
 
-					var link = L.url('admin/system/packages') + '?query=luci-proto&display=available';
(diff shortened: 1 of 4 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/network.js
```diff
@@ -113,7 +113,7 @@ L.poll(5, L.url('admin/network/iface_status', networks.join(',')), null,
 					var e = document.getElementById(ifc.id + '-ifc-edit');
 					if (e) e.disabled = true;
 
-					var link = L.url('admin/system/packages') + '?query=luci-proto&display=available';
+					var link = L.url('admin/system/opkg');
```
jow-: The query parameter still works and should be kept imho.

yousong:
v2 <- v1

 - Split it into 2 commits, with the second converting query arguments to using "query=xx" only.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/network.js
yousong: @jow- , Thank you, I just checked the code and it's indeed the case ;)

PR #2421, opened 2019-01-03
SSID escaping [RFC]

SibrenVasse:
I have a wireless network with an SSID containing an apostrophe. After escaping the SSID, "&#39;" is displayed on multiple locations in the interface. When removing the '%h'.format() calls, the apostrophe is displayed correctly.
Is this the correct fix?

There are a few more usages of '%h'.format() not changed by this commit.

Signed-off-by: Sibren Vasse <github@sibrenvasse.nl>


SibrenVasse:
@DHowett Yeah, but I was unsure where to look for escaping code.
When removing this piece of code and putting in an HTML-element (I tried <b>) it is correctly escaped.

Do you have any pointers where I can investigate further?

DHowett:
The more I look at this, the more I'm convinced it is the correct solution. `E` ends up piping plain text through `document.createTextNode` -- and `createTextNode` _does not_ require pre-escaped input.

Sorry for raising a red flag! :+1:

DHowett:
It's concerning, however. It only uses `createTextNode` if you use `E(node, ..., [ "whatever" ])`. Using `E(node, ..., "whatever")` _will_ end up putting raw text directly into the DOM... Perhaps that's a different bug that's worth fixing?

SibrenVasse:
@DHowett Yeah I see now. In the append function [luci.js:398](https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/luci.js#L398) the behavior is different if _children_ is an array. I think the following patch would fix this?

```
diff --git a/modules/luci-base/htdocs/luci-static/resources/luci.js b/modules/luci-base/htdocs/luci-static/resources/luci.js
index 4cb8bf4e5..3b3966989 100644
--- a/modules/luci-base/htdocs/luci-static/resources/luci.js
+++ b/modules/luci-base/htdocs/luci-static/resources/luci.js
@@ -415,7 +415,7 @@
 				return node.appendChild(children);
 			}
 			else if (children !== null && children !== undefined) {
-				node.innerHTML = '' + children;
+				node.appendChild(document.createTextNode('' + children));
 				return node.lastChild;
 			}
```

Before:
image

After:
image

jow-:
It actually is intentional that passing an array to `E()` will disable the HTML interpolation, the reasoning behind this is to allow for  `E('whatever', {}, [potentiallyUntrustedInput])`.

The proper fix for this issue would be removing the explicit `%h.format()` escaping in places where the value us part of an `E(..., [ ... ])` expression. I think the extraneous escaping is a remainder of previously refactored code.

SibrenVasse:
@jow- Thanks for explaining!
After a quick look at the entire tree, I only found 3 usages of %h.format() in a E(, []) expression. I've updated my branch. Could you take a look?

PR #2423, opened 2019-01-04
i18n: sync translations, add location annotatations

RoEng:
Signed-off-by: Roland Engert <Roland.Engert@web.de>


hnyman:
This only adds annotations, so not merging.
And currently already conflicts.

PR #2425, opened 2019-01-05
luci-base: Fix addr:port validate always fails

cshoredaniel:
In cbi.js there is an error which causes ipaddrport validation to always fail.
JS match() return the entire match as ret[0] and individual matches (for multiple ()) as the
subsequent list members.  So we fix it by just fixing the index in the calls that want the
individual parts.


cshoredaniel:
@jow- @hnyman  ?

cshoredaniel:
ping?

jow-:
Good catch, thanks!

PR #2427, opened 2019-01-05
luci-app-ddns: fix typo for has_curlssl()

leonghui:
Fixes the incorrect message "HTTPS not supported please disable !" when curl with https support is present.
grep returns exit code 0 when a match is found.

```
root@OpenWrt:~# ($(which curl) -V 2>&1 | grep "Protocols:" | grep -qF "https")
root@OpenWrt:~# echo $?
0
```

Signed-off-by: Leong Hui Wong <wong.leonghui@gmail.com>

### DIFF
--- a/applications/luci-app-ddns/luasrc/tools/ddns.lua
+++ b/applications/luci-app-ddns/luasrc/tools/ddns.lua
@@ -21,7 +21,7 @@ function env_info(type)
 		end
 
 		local function has_curlssl()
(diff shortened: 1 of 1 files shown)


dibdot:
Good catch. Please also remove the first grep in the pipe, it's needless, cause you are searching case sensitive for the literal 'https'. Thanks!

leonghui:
Done, thanks!

PR #2428, opened 2019-01-05
Update docs and packages wiki links

leonghui:
I found 3 instances of "wiki.openwrt.org" in this repository so I updated them to point to their new wiki pages.

Signed-off-by: Leong Hui Wong <wong.leonghui@gmail.com>


dibdot:
Thanks!

PR #2443, opened 2019-01-09
luci-mod-system: add password strength info

feckert:
Add a password strength hint.

### DIFF
--- a/modules/luci-mod-system/luasrc/view/admin_system/password.htm
+++ b/modules/luci-mod-system/luasrc/view/admin_system/password.htm
@@ -2,6 +2,27 @@
 
 <input type="password" aria-hidden="true" style="position:absolute; left:-10000px" />
 
+<script type="text/javascript">
+function checkPassword() {
+	var pw1 = document.body.querySelector('[name="pw1"]');
+	var view  = document.getElementById("passstrength");
+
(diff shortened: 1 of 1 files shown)


cshoredaniel:
I like the password strength check idea - not sure how other projects implement, but I'm of two minds about whether I think this regex based approach helps or is too simplistic to provide genuinely useful feedback.

cshoredaniel:
Also could this go in luci.util or such?  It'd be good to share this with luci-app-radicale2 for example (in progress) which has a password entry thingy.

feckert:
A lot of them use external javascript libraries. I think this is not applicable for LuCI. Keep it simple.


> Also could this go in luci.util or such? It'd be good to share this with luci-app-radicale2 for example (in progress) which has a password entry thingy.

Yes I could to this.

cshoredaniel:
I wonder if it could be called 'basic strength estimate' or some such to indicate that it's a fairly unsophisticated measure of strength.  Otherwise I like it.

yifu1128:
你们来自加拿大吗？    我来自福建省

feckert:
@yifu1128 ??
@jow- your assessment

cshoredaniel:
@feckert You might be interested in looking at https://github.com/openwrt/luci/pull/2463 where I (ab)use the datatype validation system to do password + confirm logic.  Depending on what @jow- thinks of that approach it might be a way to do this as well.

feckert:
@jow any comments from your side about this change?

jow-:
NO, feel free to merge.

PR #2444, opened 2019-01-09
Add PO for en 'translation' for apps I've created in the past

cshoredaniel:
This adds the 'en' 'translation' for various luci-apps I've created in the past.


jow-:
Merged, thanks!

PR #2462, opened 2019-01-17
luci-app-radicale2: A couple of fixes for radicale2 UI

cshoredaniel:
* In some case the password hash for existing users could be lost when updating other users
* The link the Radicale 2.x web UI sometimes was written has https:// when SSL not enabled.


cshoredaniel:
ping? @hnyman @jow-?

PR #2463, opened 2019-01-17
luci-base: Add generic password confirmation logic

cshoredaniel:
Entering passwords and their confirmation is a common task, so create
a generic mechanism for doing password + confirm dialogues.

To use this:

o = section:option(Value, "some_password", translate("Value Name"))
p = section:option(Value, "some_password_confirm", translate("Confirm Value"))

o.password = true
o.datatype="passconf"
o.confirmoption=p
p.password = true
p.datatype="passconf"
p.confirmoption=o

This will result in javascript flagging the password + confirmation as errors
until they match.  A non-JS (Lua) fallback is also included.

@feckert This might interest you in relation to #2443.

@jow- @hnyman Comments?

### DIFF
--- a/modules/luci-base/luasrc/cbi.lua
+++ b/modules/luci-base/luasrc/cbi.lua
@@ -1547,6 +1547,13 @@ end
 
 function Value.parse(self, section, novld)
 	if self.readonly then return end
+	if self.datatype == "passconf" and self.confirmoption then
+		if self:formvalue(section) ~=
+			self.confirmoption:formvalue(section)
+			then
+				return nil, _("Password confirmation doesn't match password!")
+		end
+	end
 	AbstractValue.parse(self, section, novld)
 end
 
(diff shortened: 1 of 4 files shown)


cshoredaniel:
oops forgot to update translation template...will do that shortly.

cshoredaniel:
ping @hnyman @jow- thoughts on this?

modules/luci-base/luasrc/cbi.lua
```diff
@@ -1547,6 +1547,13 @@ end
 
 function Value.parse(self, section, novld)
 	if self.readonly then return end
+	if self.datatype == "passconf" and self.confirmoption then
+		if self:formvalue(section) ~=
+			self.confirmoption:formvalue(section)
+			then
```
feckert: Wrong indentation?

modules/luci-base/luasrc/cbi.lua
```diff
@@ -1547,6 +1547,13 @@ end
...
+				return nil, _("Password confirmation doesn't match password!")
```
feckert: same her

cshoredaniel:
A combination of factors has interfered with working on much of anything non-local for 2-3 months now and in any event I am considering dropping OpenWrt from my roster of 'Things To Do' even when the difficulties are resolved, as I need to consider where too allocate my limited time.

PR #2465, opened 2019-01-17
luci-mod-network: Add mesh_rssi_threshold and encryption

HenryGiraldo:
* make mesh_rssi_threshold configurable
* add SAE encryption support

Signed-off-by: Enrique Giraldo <enrique.giraldo@galgus.net>


dangowrt:
authsea is deprecated in favour of wpa_supplicant's SAE implementation. Please make use of that instead.

PR #2469, opened 2019-01-20
luci-app-shadowsocks-libev: Fix HTML tag which will cause render error

moesoha:


yousong:
Hi @moesoha , thanks for spotting the issue.  Commits to OpenWrt repos require Signed-off-by line from the author and the commit subject line is better prefixed with `luci-app-shadowsocks-libev: xxx`.

moesoha:
I'm sorry for that. I've edit my commit now.

yousong:
Thank you.

PR #2486, opened 2019-01-29
luci-app-lxc: add aarch64 to target map

BKPepe:
Turris MOX uses aarch64 architecture and according to [stackoverflow.com](https://stackoverflow.com/questions/31851611/differences-between-arm64-and-aarch64), it is also called arm64.

With this commit, It was able to show me templates from images.linuxcontainers.org, create them, start them and so on. 
image

I think it would be great to ask review from @dibdot as he seems to be most active about luci-app-lxc even it's one row, however, I don't know anything about Lua so he may have a better solution.


dibdot:
LGTM, thanks!

PR #2488, opened 2019-01-30
luci-mod-status: Move Active Connections number to above group

peterwillcn:
Move Active Connections number to above group.

icon

Signed-off-by: xiaobo <peterwillcn@gmail.com>


jow-:
IMHO it makes no sense to group the active connections with memory stats just because it happens to be a progress bar as well.

PR #2490, opened 2019-01-30
[openwrt 18.06] luci-app-lxc: add aarch64 to target map

BKPepe:
Today this was merged to master and would be great to backport it as well to an upcoming version of OpenWrt 18.06, which should be released soon according to the mailing list.

I was asking about it on IRC and was told by @jow-  that it shouldn't be an issue.


dibdot:
@BKPepe I've cherry-picked both target map amendments (i686 & aarch64) to 18.06.

BKPepe:
ok, I wasn't sure about i686. Thank you!

PR #2494, opened 2019-02-01
18.06: luci-app-attendedsysupgrade: new api, unify status

aparcar:
to distinguish between installed packages and requested packages, the
API call is now "installed" instead of "packages".

Also unify "info_box" and "error_box" to "status_box"


hnyman:
If you do a backport, you might also pick up the two other fix commits for the app:
https://github.com/openwrt/luci/commits/master/applications/luci-app-attendedsysupgrade

At the first glance it looks like you have encorporated the typo fixes, right? but how about the variable declaration fixes from the other commit?

aparcar:
@hnyman thanks for the review, both commits added

PR #2496, opened 2019-02-01
luci-app-nlbwmon: Fixed incorrect period parameter/display

Aktanusa:
This is a fix to the issue in [jow-/nlbwmon#25](https://github.com/jow-/nlbwmon/issues/25) where parsing a date string with the constructor of the Date object would assume UTC time which would cause the wrong string to be passed when retrieving data and the wrong string to be displayed on the page.


Aktanusa:
I found bugs with the original fix which would show the wrong display using UTC timezone, so I've decided the simpler solution is to use all local timezones.

jow-:
Squashed/merged via bd56c1aac650c9d60d9f66bb0c923b0346f3fa41 - thank you!

Aktanusa:
Welcome!

PR #2497, opened 2019-02-01
luci-app-advanced-reboot - Add support for Linksys EA6350v3

imi2003:
With some experimenting I've tested this and works as intended, the only thing missing is partition labels they do show as unknown 
Tested with latest snapshot build that adds support for EA6350v3

not too sure what the offset variable does so left it as 32
First time doing any programming


hnyman:
cc @stangri

hnyman:
@imi2003 
You should sign-off your commit in the commit message.
see advice in 
https://github.com/openwrt/luci/blob/master/CONTRIBUTING.md
https://openwrt.org/docs/guide-developer/submittingpatches-tomerge#sign_your_work

(you can amend your commit and force push the PR branch to update the PR)

PR #2507, opened 2019-02-03
Rm luci-lib-luaneightbl. all functions ar also in luci-lib-ip

stargieg:
Signed-off-by: Patrick Grimm <patrick@lunatiki.de>


feckert:
No one is luci-lib-luaneightbl apart luci-app-olsr in the luci feed.
But in the source this is commented out.
I think we can merge, as I don't see any reference to luaneightbl in our code.

jow-:
Committed with slight changes.

PR #2514, opened 2019-02-05
luci-app-statistics: adding ability for luci_stats config to have multiple UPS's

sandinak:
collectd can support multiple UPSs .. 

 - altered the config builder to take a list
 - altered the template to use list

### DIFF
--- a/applications/luci-app-statistics/root/etc/config/luci_statistics
+++ b/applications/luci-app-statistics/root/etc/config/luci_statistics
@@ -137,7 +137,7 @@ config statistics 'collectd_netlink'
 
 config statistics 'collectd_nut'
 	option enable '0'
-	option UPS 'myupsname'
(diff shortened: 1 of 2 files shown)


feckert:
Please change the Topic of Pullrequest/Commit and add a prefix.
For example
**luci-app-statistics**: <description>
So we know the part of the source this commit changes.

feckert:
LGTM
I have checked that this is a list element.
In the collectd.conf I have not found any hint, but on puppet web side this config param is there an array [ ... ] this is also true for the ping config. So I think this is OK.
NUT:
https://forge.puppet.com/puppet/collectd/readme#class-collectdpluginnut
PING:
https://forge.puppet.com/puppet/collectd/readme#class-collectdpluginping

@jow to we need an upgrade script for this change?

PR #2515, opened 2019-02-07
luci-app-statistics: minor fixes

feckert:
- Fix whitspace issues
- Fix thermal definitions type has changed from temperatur to gauge

### DIFF
--- a/applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
+++ b/applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
@@ -26,34 +26,34 @@ function index()
 		apcups		= _("APC UPS"),
 		conntrack	= _("Conntrack"),
 		contextswitch	= _("Context Switches"),
-		cpu			= _("Processor"),
+		cpu		= _("Processor"),
 		cpufreq		= _("CPU Frequency"),
-		csv			= _("CSV Output"),
+		csv		= _("CSV Output"),
 		curl		= _("cUrl"),
-		df			= _("Disk Space Usage"),
+		df		= _("Disk Space Usage"),
 		disk		= _("Disk Usage"),
-		dns			= _("DNS"),
+		dns		= _("DNS"),
 		email		= _("Email"),
 		entropy		= _("Entropy"),
 		exec		= _("Exec"),
 		interface	= _("Interfaces"),
 		iptables	= _("Firewall"),
-		irq			= _("Interrupts"),
+		irq		= _("Interrupts"),
 		iwinfo		= _("Wireless"),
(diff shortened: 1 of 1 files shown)


hnyman:
> Fix thermal definitions type has changed from temperatur to gauge

What is the problem with thermal picture generation?

I think that the value name (collectd "type") is still "temperature".
https://github.com/collectd/collectd/blob/master/src/types.db#L244

The source code also references to "gauge" as alternative on some situations, 
https://github.com/collectd/collectd/blob/master/src/thermal.c

I see the thermal graphs quite normally.
image

feckert:
From the system i get the following error message.

> uhttpd[7505]: ERROR: opening '/tmp/rrd/G3000-LW/thermal-cooling_device0/temperature.rrd': No such file or directory

I got the following picture only if i change the type from temperature to gauge it is working again.

screenshot_2019-02-08 g3000-lw - thermal - luci

hnyman:
Well, R7800 has both 
* "cooling_deviceX" type of sensors (which do not seem to produce data), but graphs are visible with "gauge" type, as there is only that type rrd database. 
* "thermal_zoneX" types that produce measurements and only work with "temperature" type.


```
root@router1:~# ls /tmp/rrd/router1/
conntrack                memory                   thermal-thermal_zone2
cpu-0                    ping                     thermal-thermal_zone3
cpu-1                    thermal-cooling_device0  thermal-thermal_zone4
cpufreq                  thermal-cooling_device1  thermal-thermal_zone5
entropy                  thermal-cooling_device2  thermal-thermal_zone6
interface-br-lan         thermal-cooling_device3  thermal-thermal_zone7
iwinfo-wlan0             thermal-thermal_zone0    thermal-thermal_zone8
iwinfo-wlan1             thermal-thermal_zone1    thermal-thermal_zone9
load                     thermal-thermal_zone10

root@router1:~# ls /tmp/rrd/router1/thermal-thermal_zone0/
temperature.rrd
root@router1:~# ls /tmp/rrd/router1/thermal-cooling_device0/
gauge.rrd
```

So, it seems that the plugin supports several types of sensors and rrd data.

hnyman:
Those two types are visible in the source code:
```
enum dev_type { TEMP = 0, COOLING_DEV };
...
sstrncpy(vl.type, (dt == TEMP) ? "temperature" : "gauge", sizeof(vl.type));
```

The routers that I use (R7800 and WRT3200ACM) have been ok with "temperature", although there are some artefacts from some sensors unless you config those away. But "gauge" types are not is use in those routers.

feckert:
@hnyman thanks to clarify this. I will remove the commit **luci-app-statistics: fix thermal picture generation** So this will only PR to cleanup the code

PR #2519, opened 2019-02-08
luci-mod-status: Convert bytes into KB / MB / GB on progress bar

camarel:
Currently on the system status page the the memory usage is shown in kB only. If you have a system with 1 GB of Ram its hard to read the large numbers. This automatically formats the byte numbers more human readable into KB / MB / GB.

Signed-off-by: Claudio Marelli <camarelli@gmx.net>

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js
@@ -3,8 +3,8 @@ function progressbar(query, value, max, byte)
 	var pg = document.querySelector(query),
 	    vn = parseInt(value) || 0,
 	    mn = parseInt(max) || 100,
-	    fv = byte ? String.format('%.2mB ', value) : value,
-	    fm = byte ? String.format('%.2mB ', max) : max,
+	    fv = byte ? String.format('%1024.2mB', value) : value,
+	    fm = byte ? String.format('%1024.2mB', max) : max,
 	    pc = Math.floor((100 / mn) * vn);
 
 	if (pg) {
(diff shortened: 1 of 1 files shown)


camarel:
Current situation:

image

modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js
```diff
@@ -1,13 +1,15 @@
...
+function progressbar(query, value, max, byte)
 {
-	var pg = document.querySelector(q),
-	    vn = parseInt(v) || 0,
-	    mn = parseInt(m) || 100,
+	var pg = document.querySelector(query),
+	    vn = parseInt(value) || 0,
+	    mn = parseInt(max) || 100,
+	    fv = byte ? String.format('%.2mB ', value) : value;
+	    fm = byte ? String.format('%.2mB ', max) : max;
```
yousong: It should be comma as the line end, not semicolon.

camarel: oops, sorry for that one. Fixed it!

camarel:
So I cross checked again. In the JSON request I got the number 4124856320. Currently the values shown are in kB and are devided once by 1024:

- 4028180 kB (json value divided by 1024)
- 4.12 GB (String.format using strictly 1000)


But on the other side the String.format method (defined in cbi.js) calculates it differently. It just makes divisions by 1000.

When I check with htop, it shows 3.84G (which is always a division by 1024).

Isn't the calculation used in String.format wrong? Shouldn't it always use division by 1024 instead of 1000?

jow-:
Use %1024m to get a 1024 instead of 1000 divider

camarel:
> Use %1024m to get a 1024 instead of 1000 divider

Thank you, now I have the same values then in htop

feckert:
Please stash your last two commits into the first one this are only changes to your first commit.
Execute `git rebase -i HEAD~3`on this branche and then in the editor merge the last two with `fixup` into the first. After that execute a `git push -f` so that the branch is update with your changes.
Also add a Signed-off-by line to your commit.
If this is done I think we could merge this change

jow-:
Squashed, rewrapped and committed with 97da884240ce1a75c5a6e82dafba185ba5bb600d

PR #2520, opened 2019-02-08
luci-mod-status: Remove port suffix where not applicable

0x0aa:
In the Realtime Connections table, a :port suffix is added to the end of each address. Only TCP/UDP connections have these properties "sport" and "dport". As a result, addresses for non-TCP/UDP connections are displayed like this:

row1

This simple commit removes the :port suffix from connections that shouldn't have them, and it looks like this:

row2

The :port suffix is still shown for TCP/UDP connections.

Signed-off-by: Olli Asikainen [olli.asikainen@gmail.com](mailto:olli.asikainen@gmail.com)


jow-:
Committed with 8d839ddfb8f169c01e516e1a5b0c2e8de6948183 - thanks!

PR #2537, opened 2019-02-17
luci-mod-network: Add WPA3 SAE support

hauke:
This checks if the installed hostapd supports SAE and allows to select
WPA3 Personal SAE in that case.

Signed-off-by: Hauke Mehrtens <hauke@hauke-m.de>

### DIFF
--- a/modules/luci-mod-network/luasrc/model/cbi/admin_network/wifi.lua
+++ b/modules/luci-mod-network/luasrc/model/cbi/admin_network/wifi.lua
@@ -678,6 +678,10 @@ if hwtype == "mac80211" or hwtype == "prism2" then
 	local has_ap_sae  = (os.execute("hostapd -vsae >/dev/null 2>/dev/null") == 0)
 	local has_sta_sae = (os.execute("wpa_supplicant -vsae >/dev/null 2>/dev/null") == 0)
 
+	-- Probe OWE support
+	local has_ap_owe  = (os.execute("hostapd -vowe >/dev/null 2>/dev/null") == 0)
+	local has_sta_owe = (os.execute("wpa_supplicant -vowe >/dev/null 2>/dev/null") == 0)
+
 	if hostapd and supplicant then
 		encr:value("psk", "WPA-PSK", {mode="ap"}, {mode="sta"}, {mode="ap-wds"}, {mode="sta-wds"}, {mode="adhoc"})
 		encr:value("psk2", "WPA2-PSK", {mode="ap"}, {mode="sta"}, {mode="ap-wds"}, {mode="sta-wds"}, {mode="adhoc"})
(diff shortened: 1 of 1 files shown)


macromorgan:
Forgive me, but is it necessary to also check for 80211w support in hostapd?  I just tried this on my A5v11 (obviously running custom build) with the WolfSSL version of hostapd and while it does have SAE it doesn't have 802.11w as a limitation of the driver.  When I attempt to use it on this device I get a "Line 40: unknown configuration item 'ieee80211w'" in the log file and the interface isn't created.

I'll try testing this on a more conventional device later that can actually build without getting exotic.

hauke:
In the OpenWrt binary builds 802.11w is probably always activated because at least one driver supports it. It probably still makes sense to check for 802.11w in the lua script before, you can do SAE also without 802.11w, but in WPA3 802.11w is mandatory for WPA3 connections, for WPA2 clients it can be optional.

macromorgan:
I've confirmed that for the sunxi branch on an Orange Pi Zero I tested on WPA3 appears to work, at least on Luci.  When I set WPA3 I can see the network but my iPhone can't connect to it (expected).  When I set WPA2/WPA3 Mixed mode I can connect just fine with my iPhone and Windows 10 laptop with an Intel 7260 wireless card (expected), however I noticed that about every 5 minutes my laptop disconnects (unexpected).  When I switch to WPA2 this doesn't happen, and it's probably a bug with my laptop or hostapd though, and not your Luci code.

Doesn't 802.11w require hardware support?  For targets like the RAMIPS RT5350 would this be enabled?  How would one check for it?

Thank you.

hauke:
802.11w needs either hardware support or no hardware interference, but it looke like most drivers support it.

macromorgan:
Okay, so would the correct way to go be to add the line
`local has_80211w = (os.execute("hostapd -v11w 2>/dev/null || hostapd -veap 2>/dev/null") == 0)`
up with the area where we probe sae and owe support, then add another condition everywhere we check for both has_ap_sae or has_sta_sae to see if we also have has_80211w and only if so do we enable WPA3?

hauke:
DRIVER_11W_SUPPORT is now set by most drivers, I think it is not needed to check for this any more.

hnyman:
@hauke 
Does iwinfo already support providing WPA3 status as the AP type?

I configured WPA2/WPA3-mixed and LuCI showed the encryption status as WPA2.

I tried to look for root causes and looks like LuCI rather simply evaluates the iwinfo's output, and based on iwinfo source, I am not sure if that really supports WPA3 detection, yet.

hauke:
No, iwinfo currently does not handle WPA3 correctly. This needs some changes to support WPA3 and I am too lazy, would be nice if someone else could have a look at this problem.

PR #2558, opened 2019-02-23
luci-app-attendedupgrades: Fix errors when no upgrades or only package upgrades are available

rkunze:
Currently, there are errors in luci-app-attendedupgrades in any of these cases:
* Installed system is up to date
  * A misleading message is displayed to the user ("Package upgrades available" with an empty packages list)
  * "Edit packages" and "Request upgrade" buttons are displayed, but do not work
* No full upgrade available, but some packages can be upgraded
  * "Edit packages" and "Request upgrade" buttons are displayed, but do not work

See also aparcar/attendedsysupgrade-server/issues/152

The changes in this pull request fix three separate bugs that together are the cause for these problems.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/root/www/luci-static/resources/attendedsysupgrade.js
+++ b/applications/luci-app-attendedsysupgrade/root/www/luci-static/resources/attendedsysupgrade.js
@@ -157,19 +157,22 @@ function upgrade_check_callback(request_text) {
 
 	// create simple output to tell user what's going to be upgrade (release/packages)
 	var info_output = ""
+	var have_upgrades = false;
 	if(request_json.version != undefined) {
 		info_output += "<h3>New firmware release available</h3>"
 		info_output += data.release.version + " to " + request_json.version
 		data.latest_version = request_json.version;
+		have_upgrades = true;
 	}
-	if(request_json.upgrades != undefined) {
+	if(request_json.upgrades != undefined && Object.keys(request_json.upgrades).length > 0) {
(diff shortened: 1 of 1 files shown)


hnyman:
cc @aparcar

aparcar:
Hi @rkunze thank you very much for your debugging + code! I fixed the missing packages via https://github.com/aparcar/attendedsysupgrade-server/commit/20022e999fd6a76df96e932d85303f02aafe6280 and simplified your "empty upgrades" via https://github.com/openwrt/luci/pull/2562.

I'm somewhat unsure what is the better design, let the clients be *dump* and except the server to deliver all required params (version, packages, etc) or add more client logic to save some bandwidth. What would you suggest?

aparcar:
@dangowrt do you have an opinion on the question above?

rkunze:
> I'm somewhat unsure what is the better design, let the clients be _dump_ and except the server to deliver all required params (version, packages, etc) or add more client logic to save some bandwidth. What would you suggest?
Have the server deliver everything, IMO. Saving a handful of bytes on each request isn't worth the trouble. I only went for a client side solution because it was simple and I could test it without setting up my own server.

But make the server more tolerant against broken clients as well. Returning a 4xx for a malformed request is fine, but bombing out with HTTP 500 just because some properties are missing is a server side bug IMO.

aparcar:
Should both be fixed via https://github.com/openwrt/luci/commit/c318a10c14eba2eab616ab875a23722ef7655185

PR #2559, opened 2019-02-24
luci-base: update Chinese translate

dragon100a:
for msgid "Obtain IPv6-Address" at lede-17


hnyman:
You have not signed off your commit. I will still merge it, as it is just one-line translation.
But please remember to sign off your commit next time.

dragon100a:
> You have not signed off your commit. I will still merge it, as it is just one-line translation.
> But please remember to sign off your commit next time.

Sorry, I am not a developer. And I don't know how to use github correctly. I just edit the original page in Chrome. Never thought that it may cause some trouble.

PR #2583, opened 2019-02-27
luci-theme-material: fix loading view on href

feckert:
After opening an external hyperlink in a new browser tab, LuCI hangs in the
load screen. This commit will fix this issue.

### DIFF
--- a/themes/luci-theme-material/htdocs/luci-static/material/js/script.js
+++ b/themes/luci-theme-material/htdocs/luci-static/material/js/script.js
@@ -131,24 +131,6 @@
     $(".cbi-button-up").val("");
     $(".cbi-button-down").val("");
 
-
-    /**
-     * hook other "A Label" and add hash to it.
-     */
-    $("#maincontent > .container").find("a").each(function () {
-        var that = $(this);
-        var onclick = that.attr("onclick");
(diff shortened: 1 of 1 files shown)


shunjou:
Out of curiosity, what page with an external link does this happen to you on and with what browser? I can't reproduce it with the external link in the description on the Firewall > BCP38 page. Otherwise, I think the change is good since it doesn't seem to have any adverse effect.

feckert:
@shunjou sorry I have to mention a example. If you have luci-app-ddns https://gitlab.dev.tdt.de/dev/sys6/upstream/luci-trunk/tree/master/applications/luci-app-ddns  installed on your system and pressing on the external link on top of the page, then you will see this behavior.

feckert:
@hnyman Could you pleas have a look on this? From my point of view its where to merge this. I have also add an example https://github.com/openwrt/luci/pull/2583#issuecomment-468174180 to verify my changes.

hnyman:
Sorry, I am no real user interface javascript code specialist.
Better that @jow- looks at it.

jow-:
I have no idea about that theme, didn't understood what was broken in the first place and how or why this change fixes it so.. *shrug* :)

feckert:
Thanks for merging

PR #2370, opened 2018-12-12
theme: add initial commit for purple theme

yglb:
Signed-off-by: Yanlan Shen <yanlan.shen.@rosinson.com>

PR #2590, opened 2019-02-28
luci-theme-purple: add new theme

yglb:
Sorry, because my immaturity led to errors in code submission, which caused the previous submission to be interrupted. 

Below PR is only the bugs fix for Rosy theme, not for purple.
https://github.com/openwrt/luci/pull/2370

After a long time of deliberation, we have completely separated ourselves from the aura of material. The whole purple theme has been studied by the team for a long time. We consulted different frameworks and finally decided to write a set of CSS and JS by ourselves.

As it is completely rewritten, there may be many shortcomings. Later, we will slowly repair all kinds of shortcomings. please continue your favors toward the guidance in the future. Thank you.

@jow- @hnyman @rosysong @shunjou


themes/luci-theme-purple/Makefile
```diff
@@ -0,0 +1,14 @@
+#
+# Copyright (C) 2008-2014 The LuCI Team <luci@lists.subsignal.org>
```
jow-: Please assign your own copyright.

jow-:
Is there no way to implement this theme without a jQuery dependency? The few fade in / fade out effects can be realized with CSS and the DOM manipulation should be doable with `document.querySelector()`...

Also what is the relation of this theme to the rosy one? Does it supersede it? If so do we need to keep both?

rosysong:
> Is there no way to implement this theme without a jQuery dependency? The few fade in / fade out effects can be realized with CSS and the DOM manipulation should be doable with `document.querySelector()`...
> 

OK, thanks for your advises, will update it later.

> Also what is the relation of this theme to the rosy one? Does it supersede it ? If so do we need to keep both?

Purple theme is another solution that different from Rosy. Not only on color schemes but also on its CSS framework and user experience. Such as the Login UI (more concise), navigation menu (more amazing), So please keep them both. :-) 

Improvements remains to be done on purple theme, hope this will bring a different experience from Rosy theme. :-) Enjoy it !!!

jow-:
Closing due to timeout.

PR #2608, opened 2019-03-02
luci-app-openvpn: Update Spanish translation

castillofrancodamian:
Update Spanish translation


hnyman:
Does not look quite normal that it wants to rewrite everything.
What is the reason?
( wrong line-ends ? or something similar? )

castillofrancodamian:
> Does not look quite normal that it wants to rewrite everything.
> What is the reason?
> ( wrong line-ends ? or something similar? )

Error de escritura 😃

castillofrancodamian:
> Does not look quite normal that it wants to rewrite everything.
> What is the reason?
> ( wrong line-ends ? or something similar? )

Writing error 😃

PR #2611, opened 2019-03-02
luci-app-openvpn: Update Spanish translation

castillofrancodamian:
Update Spanish translation


hnyman:
The same thing again. all 844 lines deleted and re-created.
You have edited the same file last time two weeks ago, and then there was no problem. What has happened in the meanwhile?

castillofrancodamian:
> The same thing again. all 844 lines deleted and re-created.
> You have edited the same file last time two weeks ago, and then there was no problem. What has happened in the meanwhile?

I do not know what it could be. There must be a problem with the file. I will try another way.

hnyman:
Typical reason might be DOS-Unix line-ednings difference (CRLF vs. LF)
But hard to figure out why that would manifest now.

hnyman:
Please verify the file difference after committing to your own fork,  before publishing the actual PR.

PR #2618, opened 2019-03-08
luci-app-statistics: enhancements

feckert:
- Add rrdtool canvas height option
- Diskfree graph show also percent view
- Load graph menu entry name from the definitions
- Add new item callback for menu entry (needed for menu entry)
- Update i18n

### DIFF
--- a/applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
+++ b/applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
@@ -115,11 +115,15 @@ function index()
 		-- get plugin instances
 		local instances = tree:plugin_instances( plugin )
 
-		-- plugin menu entry
-		entry(
-			{ "admin", "statistics", "graph", plugin },
-			call("statistics_render"), labels[plugin], idx
-		).query = { timespan = span , host = host }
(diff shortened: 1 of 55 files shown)


hnyman:
Hmm.

The plugin titles defined in the old "labels" definitions are still used for the collectd setup menu, right?
(e.g .line 93)
https://github.com/openwrt/luci/blob/9c3132b97e0303df793a17786b48137061915454/applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua#L93

So, the new item thing in this PR creates a duplicate name to be used only in the graph display menu (line 121). So the plugin display name needs to be defined in two places for the two use cases in the menus?

feckert:
@hnyman i want to split the configuration labels from the graph labels. So if you have exe/python/lua/perl plugins for data collection you have no label and so it will not shown in the graph view (if you installed/written a definitions file). If we split up plugin configuration (collectd configuration) and graph view definition (rrdtool configuration) we have everything for the graph in one place and it so **generic**. We do not have to add a label to the source anymore

feckert:
@hnyman @jow ping

jow-:
Merged, thanks!

feckert:
@jow thanks

PR #2620, opened 2019-03-11
luci-mod-system: add button to sync with ntp server

feckert:
Especially for systems without RTC this change has two advantages

* manual time sync with time server during configuration
* test possibility of the time sync over the configured servers


feckert:
@hnyman ping
Good?
Bad?

hnyman:
Good idea, but no comment on the code :-(  
(as I have not tinkered with the XHR code etc.)

feckert:
I'll try. If no one likes this, then I can take the merge back again.
@hnyman thanks for your feeback

PR #2642, opened 2019-03-29
luci-base: add cbi tsection error msg option

feckert:
In some situation it is usefull to inform the use that this section
could not get delete, because this config is referenced or is in use.

Example pseudo code:

```
function s.remove(self, sid)
  <check if secation could get deleted>
  if output == nil then
    self.error_msg = translatef("Unable to remove this section")
    return nil
  end
end
```


jow-:
Merged, thanks!

PR #2656, opened 2019-04-03
[WIP] luci-app-attendedsysupgrade: use fetch

aparcar:
instead of using XMLHttp requests, now uses fetch.

`auto_search` is work in progress.


jow-:
Note that LuCI still supports browsers without `fetch()` support. Nothing that I consider blocking this PR but maybe you should somehow handle it, even if it is just a naive `alert()`

Issue #985, opened 2017-02-01
luci-app-statistics sensors - missing graphs


kofec:
@jow- 
You did a great job adding sensors for luci https://github.com/openwrt/luci/commit/cff2b99b4f9af0957ff6de4fe1ce48d30337b41b but one small thing is missing. You scan for all sensors and gather all data. But there is no definition for graphs - only for temperature. Maybe you can definitions for Volatge and fan speed graphs.
applications/luci-app-statistics/luasrc/statistics/rrdtool/definitions/sensors.lua
Thanks in advance

jow-:
@kofec - I know, in the commit message I wrote:

> Please send the output of "rrdtool info /tmp/rrd/*/sensors-*/*.rrd"
> if your system happens to support voltage, power or fanspeed sensors.

Do you have a system that supports voltage and/or fan speed? If so can you provide the output?

kofec:
Yes, i wilk do this today

02.02.2017 12:43 AM "Jo-Philipp Wich" <notifications@github.com> napisał(a):

kofec:
@jow- 
I attach file - https://github.com/kofec/openwrt-n2n/blob/master/files/sensors-lm85-i2c-0-2e.zip
simple output:
```
root@LEDEnsa310:~# sensors
lm85-i2c-0-2e
Adapter: mv64xxx_i2c adapter
in0:          +1.80 V  (min =  +0.00 V, max =  +3.32 V)
in1:          +1.08 V  (min =  +0.00 V, max =  +2.99 V)
in2:          +3.21 V  (min =  +0.00 V, max =  +4.38 V)
in3:          +5.05 V  (min =  +0.00 V, max =  +6.64 V)
in4:         +12.19 V  (min =  +0.00 V, max = +15.94 V)
fan1:        1009 RPM  (min =    0 RPM)
fan2:           0 RPM  (min =    0 RPM)
fan3:           0 RPM  (min =    0 RPM)
fan4:           0 RPM  (min =    0 RPM)
temp1:        +27.0°C  (low  = -127.0°C, high = +127.0°C)
temp2:        +30.0°C  (low  = -127.0°C, high = +127.0°C)
temp3:        +34.0°C  (low  = -127.0°C, high = +127.0°C)
cpu0_vid:    +0.000 V
```
At the beginning I monitor all sensors but after saome time I change it and monitor only in4, fan1 and all temp.
output of your command:

```
root@LEDEnsa310:~# /usr/sbin/sensors -u -A
[...]

PR #2657, opened 2019-04-03
luci-app-statistics - allow rrd files to contain :

msurovcak:
Fixes situations where RRD file name contains ":" (eg. _ping/ipv6_) in `rrdtool` it's unescaped - thus not able to render image. Adding simple escaping of `:` to `\\:` fixes the situation.

Might be a solution for https://github.com/openwrt/luci/issues/985

### DIFF
--- a/applications/luci-app-statistics/luasrc/statistics/rrdtool.lua
+++ b/applications/luci-app-statistics/luasrc/statistics/rrdtool.lua
@@ -62,7 +62,7 @@ function Graph._mkpath( self, plugin, plugin_instance, dtype, dtype_instance )
 end
 
 function Graph.mkrrdpath( self, ... )
-	return string.format( "%s/%s.rrd", self.opts.rrdpath, self:_mkpath( ... ) )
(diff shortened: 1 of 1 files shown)


feckert:
Could you give me an example page in the LuCI where I can reproduce this issue?

jow-:
I think we should start escaping backslashes as well then (`...:gsub("\\", "\\\\"):gsub(":", "\\")`)

msurovcak:
@feckert I have this on _luci-app-statistics/graph_ witht _ping_ to any _ipv6_ address, the rrd file is eg.: `/tmp/rrd/turris/ping/ping-2a02:598:2::1053.rrd`

@jow- I am not sure here, is there any case rrd file with `\` will be produced?

feckert:
I will merge this change and if we have a use case we could extent this to escaping backslashes as well.

jow-:
This is a matter of correctness. Whenever you introduce escape sequences in code, you need to ensure that the escape sequences itself are escaped as well. Regardless of whether such names can happen or not.

feckert:
@jow- sorry for this I will add this changes asap

jow-:
Np, I pushed a chage already.

PR #2667, opened 2019-04-08
[WIP] luci-mod-system: add application led trigger possibility

feckert:
Currently we only support native kernel LED triggers.
In some situations it is useful to have LED triggers for system events.
Let's call it application triggers.

An example would be the status of all connections on the mwan3.

There are the following possible states:
* If all mwan3 connections are connected, then the alarm LED is off
  (kernel trigger -> none).
* If all mwan3 connections are not connected, then the alarm LED is on
  (kernel trigger -> default-on).
* If one of the mwan3 connections is not connected (backup), the alarm LED
  should flash (kernel trigger -> timer).

The evaluation is done by a script in the directory "/etc/hotplug.d/iface"

Currently it is not possible to select this type of trigger. This change
makes this possible like with the Proto-Handlers in LuCI.

If this would be a nice feature I will add the trigger to mwan3 and open a pullrequest.

### DIFF
--- a/modules/luci-mod-system/luasrc/model/cbi/admin_system/leds.lua
+++ b/modules/luci-mod-system/luasrc/model/cbi/admin_system/leds.lua
@@ -185,4 +185,22 @@ end
 port_mask = s:option(Value, "port_mask", translate ("Switch Port Mask"))
 port_mask:depends("trigger", "switch0")
 
+-- Load application triggers
+local ledtriggers = string.format("%s/%s", util.libpath(), "model/cbi/admin_system/ledtriggers")
+if nixio.fs.access(ledtriggers) then
+	local net = require "luci.model.network".init()
+	local interface = s:option(ListValue, "ifc", translate("Network interface"))
+	for _, net in ipairs(net:get_networks()) do
+		if net:name() ~= "loopback" then
+			interface:value(net:name())
+		end
+	end
+	for file in fs.dir(ledtriggers) do
+		if file:match("led_") then
+			local form, ferr = loadfile(string.format("%s/%s", ledtriggers, file))
+			setfenv(form, getfenv(1))(m, s, trigger, interface)
+		end
+	end
+end
+
 return m


feckert:
@jow any hint from your side about this PR

feckert:
This is an example for the application triggers https://github.com/openwrt/packages/pull/9340

feckert:
Does not work anymore. So I will close this and rework/rebase my feature.

feckert:
@jow- I am working again on this feature, to add the application LEDs trigger feature. The reason why i want to add this, is that I want to signal over an LED if mwan3 is in connected, disconnected or in backup state.
Also a nice feature would be to show VPN (openvpn/ipsec) connection or SSH connection.
Could you give me an advice how to make this with the new javascript implementation in LuCI.

jow-:
I do not understand how these triggers are supposed to work. Can you explain how they look in uci, what is triggering them etc.?

feckert:
@jow- For now we have only kernel triggers support from **/sys/class/leds/<name>/trigger** in LuCI. But in the system there could be other sources who want to set the generic led triggers [timer] [none] [default-on]. In this pullrequest from https://github.com/openwrt/packages/pull/9340 I set the following kernel triggers on interface events if the mwan3 state changes
* on backup state the led kernel trigger [timer] is set with on/off time 500ms
* on offline state the led kernel trigger [default-on] is set
* on online state the led kernel trigger [none] is set

To make this possible I have to set the led trigger to **mwan3_status**  so that on an interface event this led could be changed by the script.
Led handling is not be done by the kernel it is done on an interface event.
But this is not possible because I can only select led kernel triggers.

jow-:
Is this something official in OpenWrt? To me it looks as if https://github.com/openwrt/openwrt/blob/master/package/base-files/files/etc/init.d/led#L67 will emit an error to the syslog whenever you try to use "custom" trigger values.

feckert:
@jow- I know this line. If we add application triggers we have to change this in the future so we could use other triggers. In the following line we already use an so called application trigger https://github.com/openwrt/openwrt/blob/8e197c7af2ef7eccfd1dd9396ab6a003992b183a/package/base-files/files/etc/init.d/led#L34

We could prefix this triggers with **app_\<name\>** so we could skip the led handling by the intit script because an user space process (for example hotplug scripts) will configure the led.

jow-:
Hmm ok. I see two possibilies here: the simple one is turning the trigger option from a `ListValue` into a `Value` widget, then users can at least enter custom values for now.

If you want to keep the list behavior, you need to define a proper way to "register" application triggers first, which seems to be an OpenWrt base system topic to me.

One could for example introduce an `/usr/libexec/led-trigger/` where each executable within this directory represents a custom application trigger. The `/etc/init.d/led` script will then invoke the executable to configure the led, e.g. 

```
config led
    option sysfs 'foobar:green:status'
    option trigger 'app_mwan3'
    ...
```

would cause /etc/init.d/led to execute/source `/usr/libexec/led-trigger/mwan3` which then programs the led (e.g. brightness, initial value) according to the remaining uci options.

Once such a framework is in place, you can extend LuCI to read the list of triggers by listing the contents of `/usr/libexec/led-trigger/` using `fs.list()` for example.

feckert:
> Hmm ok. I see two possibilies here: the simple one is turning the trigger option from a `ListValue` into a `Value` widget, then users can at least enter custom values for now.

However, this would mean that the following message https://github.com/openwrt/openwrt/blob/8e197c7af2ef7eccfd1dd9396ab6a003992b183a/package/base-files/files/etc/init.d/led#L67 would appear in the log at startup. Also I cannot set any options if a trigger needs another configuration parameter. I think this is not the way I want to go.


> One could for example introduce an `/usr/libexec/led-trigger/` where each executable within this directory represents a custom application trigger. The `/etc/init.d/led` script will then invoke the executable to configure the led, e.g.

Why is it necessary for /etc/init.d/led to call and execute this? Setting the kernel led trigger is supposed to make the application or hotplug event.
I agree with you that application triggers have to register under /lib/libexec/led-trigger/<name> with putting a file to this location. So the LuCi could read this file to registrate an application trigger and for /etc/init.d/led to drop the configuration of the led during start and restart because this is done by an application or hotplug event script which changes the leds by setting simple kernel led triggers [none] [time] [default-on].

> would cause /etc/init.d/led to execute/source `/usr/libexec/led-trigger/mwan3` which then programs the led (e.g. brightness, initial value) according to the remaining uci options.

If we change the configuration of an kernel configured led, for example setting a led to use the kernel netdev triggerthen the application leds would get changed too the default state again! This is not what I want because the led is controlled bei an application or hotplug event script.

> Once such a framework is in place, you can extend LuCI to read the list of triggers by listing the contents of `/usr/libexec/led-trigger/` using `fs.list()` for example.

Yes this could then be done.

PR #2670, opened 2019-04-10
luci-mod-network: move auto option to general tab

feckert:
Since openwrt-18.06 the auto option is set/unset if we press
connect/disconnect on the interface page. So I think we should move this
to the general tab so that we see at once if this is started on boot or
not.
This revers also to #2119


jow-:
I agree

feckert:
@jow- thanks for feedback on this topic

PR #2673, opened 2019-04-12
luci-app-mjpg-streamer: add link option 

jose1711:
Signed-off-by: Jose Riha <jose1711@gmail.com>


feckert:
Please update your commit message in the commit to

luci-app-mjpg-streamer: add link option

but I do net see the link option in the init script
https://github.com/openwrt/packages/blob/master/multimedia/mjpg-streamer/files/mjpg-streamer.init

jose1711:
> but I do net see the link option in the init script
> https://github.com/openwrt/packages/blob/master/multimedia/mjpg-streamer/files/mjpg-streamer.init

it is a chicken-vs-egg problem. the `link` option will only appear in `mjpg-streamer.init` if https://github.com/openwrt/packages/pull/8658 gets merged

feckert:
Thanks for the hint. And the other options?

jose1711:
> Thanks for the hint. And the other options?

All the other options were already in sync with the init file

feckert:
Related pullrequest https://github.com/openwrt/packages/pull/8658 was merged.

Issue #2671, opened 2019-04-10
luci-app-shadowsocks-libev cannot save from "Redir Rules"


jackcctse:
Have been getting this error for quiet some snapshot builds. 

1. Reproducing the error: just click save on "Redir Rules" screen, with or without editing makes no difference.
2. The error: failed to save anything, below error message appears.
`Failed to execute cbi dispatcher target for entry '/admin/services/shadowsocks-libev/rules'.
The called action terminated with an exception:
/usr/lib/lua/luci/cbi.lua:165: Datatype error, bad token "cidr"
stack traceback:
	[C]: in function 'error'
	/usr/lib/lua/luci/cbi.lua:165: in function 'compile_datatype'
	/usr/lib/lua/luci/cbi.lua:177: in function 'compile_datatype'
	/usr/lib/lua/luci/cbi.lua:189: in function 'verify_datatype'
	/usr/lib/lua/luci/cbi.lua:1493: in function 'validate'
	/usr/lib/lua/luci/cbi.lua:1414: in function 'parse'
	/usr/lib/lua/luci/cbi.lua:248: in function 'parse'
	/usr/lib/lua/luci/cbi.lua:1111: in function 'parse'
	/usr/lib/lua/luci/cbi.lua:248: in function 'parse'
	/usr/lib/lua/luci/cbi.lua:387: in function 'parse'
	/usr/lib/lua/luci/dispatcher.lua:881: in function </usr/lib/lua/luci/dispatcher.lua:860>`

Workaround: edit /etc/config/shadowsocks-libev with text editor

Lastest version tested

> OpenWrt SNAPSHOT r9782-09fe18f042 / LuCI Master (git-19.093.68468-d679b93) 
> luci-app-shadowsocks-libev git-19.093.68468-d679b93-1

dobetter:
The same here!

PR #2681, opened 2019-04-23
luci-app-shadowsocks-libev: fix cidr datatype

dibdot:
* fix unknown cidr datatype (#2671)

### DIFF
--- a/modules/luci-base/luasrc/cbi/datatypes.lua
+++ b/modules/luci-base/luasrc/cbi/datatypes.lua
@@ -132,6 +132,10 @@ function ip6prefix(val)
 	return ( val and val >= 0 and val <= 128 )
 end
 
+function cidr(val)
+	return cidr4(val) or cidr6(val)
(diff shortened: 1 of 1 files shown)


dibdot:
@yousong @jow-

yousong:
Hi @dibdot , thanks for taking care of this.  The issue was caused by commit a86c068870dea59844a282ca43885ef4fcbfa07f .  "cidr" as a validator is part of ubox validate command.  Maybe we should add 'cidr' as a alias for cidr4 and cidr6 into LuCI

jow-:
@dibdot - can you extend the client side JS validator in cbi.js as well?

dibdot:
@jow- There's already a generic 'cidr' function in cbi.js or did you mean something different!?

jow-:
@dibdot - ah nevermind then, sorry. I didn't bother to actually look due to spotty 3g connection.

PR #2687, opened 2019-05-03
luci-mod-admin-full: add dependency for channel auto

HenryGiraldo:
Do not show 'auto' channel option if hostapd is not compiled with CONFIG_ACS

Signed-off-by: Enrique Giraldo <hgiraldos@gmail.com>


robimarko:
This looks like really good idea, as auto has been default for years but it actually did not work.

jow-:
Actually `auto` was never meant to imply `automatic channel selection` but `use implementation defined default channel`. On OpenWrt it will typically cause the setup scripts to use the first available channel, depending on whether the radio is `a` or `b` band capable.

But I have no strong preference so I'll merge it.

PR #2730, opened 2019-05-27
[WIP] luci-base: Hack to bypass node users verification

wesraph:
Since openwrt-18.06, users different than root should be hardcoded in the code to be able to log in Luci. Because of this it is not possible to dynamically register/unregister users.

The check on "allowed_users" is done before checking RPCd. Why not just check RPCd and keep the rights inside ?


jow-:
I tend to agree that we should get rid of the hardcoded login user list but will need a more thorough review of the authentication code to avoid opening any security issues.

jow-:
This has been superseded with 184ea6230076bf4dd3abf6669bde17f87a051146 - the `root` username is not enforced anymore.

PR #2732, opened 2019-05-28
luci-app-hd-idle: Update Spanish translation

castillofrancodamian:
Update Spanish translation


feckert:
thanks

PR #2742, opened 2019-06-02
luci-base: handle dhcp lease from odhcpd

Ansuel:
Currently only the dnsmasq lease file is read for hostname hint. If someone use only odhcpd as dhcp daemon the hostname resolution on wifi_assoc list is broken.

Also i notice that very similar code is used in 2 different source.

dhcp_leases_common(family) in luci tools status.lua is very similar to what _nethints does... 
Some cleanup should be done as this error was already fixed in status.lua but was still present in sys.lua.

Wifi_assoc list use sys.lua 
lease_status use status.lua and include the odchpd lease file.

@jow- 

### DIFF
--- a/modules/luci-base/luasrc/sys.lua
+++ b/modules/luci-base/luasrc/sys.lua
@@ -137,7 +137,7 @@ end
 net = {}
 
 local function _nethints(what, callback)
-	local _, k, e, mac, ip, name
+	local _, k, e, mac, ip, name, duid, iaid
 	local cur = uci.cursor()
 	local ifn = { }
 	local hosts = { }
@@ -386,6 +386,26 @@ function net.devices()
 	return devs
 end
 
+function net.duid_to_mac(duid)
+	local b1, b2, b3, b4, b5, b6
+
+	if type(duid) == "string" then
+		-- DUID-LLT / Ethernet
+		if #duid == 28 then
+			b1, b2, b3, b4, b5, b6 = duid:match("^00010001(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)%x%x%x%x%x%x%x%x$")
+
+		-- DUID-LL / Ethernet
+		elseif #duid == 20 then
+			b1, b2, b3, b4, b5, b6 = duid:match("^00030001(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)$")
+
+		-- DUID-LL / Ethernet (Without Header)
+		elseif #duid == 12 then
+			b1, b2, b3, b4, b5, b6 = duid:match("^(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)$")
+		end
+	end
+
+	return b1 and luci.ip.checkmac(table.concat({ b1, b2, b3, b4, b5, b6 }, ":"))
+end
 
 process = {}
 
--- a/modules/luci-base/luasrc/sys.lua
+++ b/modules/luci-base/luasrc/sys.lua
@@ -189,6 +189,24 @@ local function _nethints(what, callback)
 			end
 		end
 	)
+	
+	cur:foreach("dhcp", "odhcpd",
+		function(s)
+			if type(s.leasefile) == "string" and fs.access(s.leasefile) then
+				for e in io.lines(s.leasefile) do
+					duid, iaid, name, _, ip = e:match("^# %S+ (%S+) (%S+) (%S+) (-?%d+) %S+ %S+ ([0-9a-f:.]+)/[0-9]+")
+					mac = net.duid_to_mac(duid)
+					if mac then
+						if ip and iaid == "ipv4" then
+							_add(what, mac, ip, nil, name ~= "*" and name)
+						elseif ip then
+							_add(what, mac, nil, ip, name ~= "*" and name)
+						end
+					end
+				end
+			end
+		end
+	)
 
 	cur:foreach("dhcp", "host",
 		function(s)
--- a/modules/luci-base/luasrc/tools/status.lua
+++ b/modules/luci-base/luasrc/tools/status.lua
@@ -6,21 +6,6 @@ module("luci.tools.status", package.seeall)
 local uci = require "luci.model.uci".cursor()
 local ipc = require "luci.ip"
 
-local function duid_to_mac(duid)
-	local b1, b2, b3, b4, b5, b6
-
-	-- DUID-LLT / Ethernet
-	if type(duid) == "string" and #duid == 28 then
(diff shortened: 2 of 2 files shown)


modules/luci-base/luasrc/sys.lua
```diff
@@ -189,6 +189,23 @@ local function _nethints(what, callback)
...
 	)
+	
+	cur:foreach("dhcp", "odhcpd",
+		function(s)
+			if s.leasefile and fs.access(s.leasefile) then
+				for e in io.lines(s.leasefile) do
+					local iface, duid, iaid, name, ts, id, length, ip = e:match("^# (%S+) (%S+) (%S+) (%S+) (-?%d+) (%S+) (%S+) (.*)")
+					if ip and iaid == "ipv4" then
+						ip = ip:gsub("/[0-9]+","")
+						mac = luci.ip.checkmac(duid:gsub("^(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)$", "%1:%2:%3:%4:%5:%6")) or "00:00:00:00:00:00"
```
jow-: If the mac cannot be parsed, skip the entry.
Also: `local mac`

modules/luci-base/luasrc/sys.lua
```diff
@@ -189,6 +189,23 @@ local function _nethints(what, callback)
 			end
 		end
 	)
+	
+	cur:foreach("dhcp", "odhcpd",
+		function(s)
+			if s.leasefile and fs.access(s.leasefile) then
+				for e in io.lines(s.leasefile) do
```
jow-: `local e`

modules/luci-base/luasrc/sys.lua
```diff
@@ -189,6 +189,23 @@ local function _nethints(what, callback)
 			end
 		end
 	)
+	
+	cur:foreach("dhcp", "odhcpd",
+		function(s)
+			if s.leasefile and fs.access(s.leasefile) then
```
jow-: Check for `type(s.leasefile) == "string"` otherwise, `fs.access()` will assert for `list leasefile foo`.

modules/luci-base/luasrc/sys.lua
```diff
@@ -189,6 +189,23 @@ local function _nethints(what, callback)
...
+				for e in io.lines(s.leasefile) do
+					local iface, duid, iaid, name, ts, id, length, ip = e:match("^# (%S+) (%S+) (%S+) (%S+) (-?%d+) (%S+) (%S+) (.*)")
+					if ip and iaid == "ipv4" then
+						ip = ip:gsub("/[0-9]+","")
```
jow-: I suggest to tweak the `e:match()` pattern above to not include `/[0-9]+` in the first place, e.g.
`e:match("^# (%S+) (%S+) (%S+) (%S+) (-?%d+) (%S+) (%S+) ([0-9a-f:.]+)/[0-9]+")`

modules/luci-base/luasrc/sys.lua
Ansuel: mac is already local

modules/luci-base/luasrc/sys.lua
Ansuel: e is already local

modules/luci-base/luasrc/sys.lua
Ansuel: this should also be done to dnmasq function then

jow-:
Another thing - when we parse this file anyway, shouldn't we use the IPv6 lease information as well?

E.g.:

```
if ip and iaid == "ipv4" then
    ...
elseif ip then
   ...
   _add(what, mac, nil, ip, name ~= "*" and name)
end
```

modules/luci-base/luasrc/sys.lua
jow-: Ack

modules/luci-base/luasrc/sys.lua
Ansuel: i removed the extra variables and moved duid iaid where the others locals are declared

Ansuel:
> Another thing - when we parse this file anyway, shouldn't we use the IPv6 lease information as well?
> 
> E.g.:
> 
> ```
> if ip and iaid == "ipv4" then
>     ...
> elseif ip then
>    ...
>    _add(what, mac, nil, ip, name ~= "*" and name)
> end
> ```

this way we should have hostname als for ipv6 right?

jow-:
Could you either add IPv6 handling or change the commit subject to `handle DHCPv4 lease from odhcpd` ? The it is ready to merge :)

jow-:
> this way we should have hostname als for ipv6 right?

Yeah

Ansuel:
@jow- find that too... and i was thinking to move that function to the sys.net (as it will be used by sys.lua and status.lua) and adds a way to translate duid from plain ipv4 (duid lenght = "12")

Ansuel:
I'm a little dubious if I should make another commit.

modules/luci-base/luasrc/sys.lua
```diff
@@ -386,6 +404,26 @@ function net.devices()
...
+		if #duid == 28 then
+			b1, b2, b3, b4, b5, b6 = duid:match("^00010001(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)%x%x%x%x%x%x%x%x$")
+	
+		-- DUID-LL / Ethernet
+		elseif #duid == 20 then
+			b1, b2, b3, b4, b5, b6 = duid:match("^00030001(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)$")
+			
+		-- DUID-LL / Ethernet (Without Header)
+		elseif #duid == 12 then
+			b1, b2, b3, b4, b5, b6 = duid:match("(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)(%x%x)$")
```
jow-: Please add a `^` to the match

modules/luci-base/luasrc/sys.lua
```diff
@@ -189,6 +189,24 @@ local function _nethints(what, callback)
...
+		function(s)
+			if type(s.leasefile) == "string" and fs.access(s.leasefile) then
+				for e in io.lines(s.leasefile) do
+					duid, iaid, name, _, ip = e:match("^# %S+ (%S+) (%S+) (%S+) (-?%d+) %S+ %S+ ([0-9a-f:.]+)/[0-9]+")
+					mac = net.duid_to_mac(duid)
+					if mac then
+						if ip and iaid == "ipv4" then
+							_add(what, mac, ip, nil, name ~= "*" and name)
+						elseif ip then
+							_add(what, mac, ip, nil, name ~= "*" and name)
```
jow-: You need to swap `nil` and `ip` in the IPv6 case

Ansuel:
@jow- should be good now

PR #2746, opened 2019-06-03
luci-theme-material: fix checkbox css StaticList view

feckert:
Without the change:

LED-Konfiguration-without

With the change:

LED-Konfiguration-with

### DIFF
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -2089,6 +2089,10 @@ label > input[type="radio"] {
 	vertical-align: bottom;
 }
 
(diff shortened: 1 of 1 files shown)


themes/luci-theme-material/htdocs/luci-static/material/cascade.css
```diff
@@ -2089,6 +2089,10 @@ label > input[type="radio"] {
 	vertical-align: bottom;
 }
 
+label[data-index][data-depends] {
+	padding-right:2em;
```
jow-: Please add a space before `2em`, otherwise LGTM :)

feckert: Thanks for feedback -> Merged

PR #2749, opened 2019-06-04
luci-app-wireguard: QR-Code enhancements

dibdot:
* explain the QR-Code 'use case' and the transfered information
* limit the QR-Code information to bare minimum (per interface):
  * [Interface] A random, on the fly generated 'PrivateKey',
                the key will not be saved on the router
  * [Peer] The 'PublicKey' of that wg interface and the
           'AllowedIPs' with the default of '0.0.0.0/0, ::/0'
           to allow sending traffic to any IPv4 and IPv6 address
* the QR-Code logic will be processed only once per page load,
  and not every 5 seconds with the regular XHR refresh

See corresponding forum thread:
https://forum.openwrt.org/t/luci-app-wireguard-qr-code-shows-private-key/38133

### DIFF
--- a/applications/luci-app-wireguard/luasrc/view/wireguard.htm
+++ b/applications/luci-app-wireguard/luasrc/view/wireguard.htm
@@ -4,19 +4,22 @@
 -%>
 
 <%
-	local uci = uci.cursor()
 	local data = { }
 	local last_device = ""
-	local enc = { }
+	local qr_pubkey = { }
 
-	local function wg_clean(value)
-		if value and value == "(none)" then
-			value = ""
+	local function qr_clean(qr_type, value)
+		if not value or value == "" or value == "(none)" then
+			return ""
 		end
-		return value
+		if qr_type == "privkey" then
+			return "PrivateKey = " ..value
+		elseif qr_type == "pubkey" then
+			return "PublicKey = " ..value
+		end
 	end
 
-	local wg_dump = io.popen("wg show all dump")
+	local wg_dump = io.popen("wg show all dump 2>/dev/null")
 	if wg_dump then
 		local line
 		for line in wg_dump:lines() do
@@ -30,17 +33,7 @@
 					fwmark		= line[5],
 					peers		= { }
 				}
-				local s = uci:get_list("network", line[1], "addresses")
-				local address = ""
-				local key, value
-				for key, value in pairs(s) do
-					if address ~= "" then
-						address = address.. ", " ..value
-					else
-						address = value
-					end
-				end
-				enc[line[1]] = "[Interface]\nPrivateKey = " ..wg_clean(line[2]).. "\nAddress = " ..address
+				qr_pubkey[line[1]] = qr_clean("pubkey", line[3])
 			else
 				local peer = {
 					public_key		= line[2],
@@ -60,7 +53,6 @@
 					end
 				end
 				table.insert(data[line[1]].peers, peer)
-				enc[line[1]] = enc[line[1]].. "\n\n[Peer]\nEndpoint = " ..wg_clean(line[4]).. "\nPublicKey = " ..wg_clean(line[2]).. "\nAllowedIPs = " ..wg_clean(line[5])
 			end
 		end
 	end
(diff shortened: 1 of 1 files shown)


dibdot:
New information text:
wg_info1

New error check example:
wg_info2

feckert:
Could we change the output so I can load this with my android wireguard app? In the prevision version it is not working. I have to say that I did a quick test so it could be I did some thing wrong.

dibdot:
I've tested all changes with this app ... and it works OK (at least for me)

feckert:
Thanks for feedback I will evaluate it again.

BlueMax:
I think the implementation is not correct. 
Instead of passing the servers private/public key pair it should:
- create new keypair
- create new peer (on server) and fill-in the public key (generated above).
- send keypair to client with public key of server (3 keys)
- forget/delete the clients private key (on server)

dibdot:
> I think the implementation is not correct.
> Instead of passing the servers private/public key pair it should:

First of all: please do not equate the terms "router" and "server". 
Wireguard is implemented per interface and there are a couple of possible setups (per interface!).

The current QR-Code implementation is only useful for road warrior/client setups, it's not intended for server setups e.g. to provide peer settings to a mobile for a router interface which is acting as a wg "server". Said that, you can create & mix multiple "server" and "client" wg interfaces with different key pairs on the same router etc.

Your approach won't work, cause you missed an important step: Every client setup needs a (different!) public key verification step with the vpn provider (Mullvad, Azire etc.)! It's no problem to generate a wg key on the fly, but if you try to automate different verification processes per provider, you need much much more error prone logic and additional vpn provider credentials on your router ... this is out of scope (at least for me) and in terms of security worse.

In general you have to trust your router environment, if it doesn't, don't store any keys or other credentials on your device, e.g.  for wireguard, openvpn etc.

BlueMax:
We are probably talking about different things here indeed. Many ppl use Wireguard as a drop-in replacement for OpenVPN to get into their local subnets for Samba, Nextcloud, Nginx or whatever _directly_. Not with any 3rd party VPN service. The QR Code is completely unsuitable for that. The Wireguard implementation itself has nothing to do with it. It works properly if set up manually indeed with asymmetrical keys and manual exchange. But not with the QR Code generator. Actually its not generating anything. It just passes the data along. And that's a dangerous and misleading thing. If used from scratch for the use case described above it results in symmetrical keys that is not intended nor recommended at all. It might raise security flaws or side-effect no one knows of. It has never been intended to be used that way and it should be avoided at all cost. If its intended to only pass 3rd party keys it should be stated appropriately. 
I trapped into that pit and many other did as well like [here](https://forum.openwrt.org/t/luci-app-wireguard-qr-code-shows-private-key/38133/1) and [here](https://forum.openwrt.org/t/wireguard-setup-help/37844).
I don't say its wrong how it is done atm but it should be stated clearly for what use case (3rd party key passing?) its intended and if its not possible to avoid symmetrical (same) keypairs it should be noted with bold letters as well that this is against official implementation intentions and might lead to significant degradation in security and encryption integrity.

feckert:
Regardless of your views. I have a simple question. Should the app work with the android app.

Nevertheless the private keys should be generated on the device and so should not leave them. So for the android app only the public information should be seen in the QR code (network config and router public key). After importing this information in the android app we could generate a device key. The private key stays on the device (phone) and the public key of them must be saved into the router. After that we are done and start a secured connection.

If I generate a file for the android app with the public information of the router and do a qr generation and scan them with the phone it works. I only have to publish the public key from the phone to the router.

BlueMax:
> If I generate a file for the android app with the public information of the router and do a qr generation and scan them with the phone it works. I only have to publish the public key from the phone to the router.
That's how it's intended to do be done.
The current OpenWrt QR Code generator though sends the routers private and public key. Both sides have the same keys then. Actually, it sends everything that has been set up (Router/Peer entries). The Android App cannot handle peer entries though and errors out (syntax error).

dibdot:
> So for the android app only the public information should be seen in the QR code (network config and router public key).

@feckert 
Please provide a working client.conf example that could be transfered via QR code to the mobile.

Thanks!

dibdot:
@feckert as discussed already offsite: I've rebased this PR and changed the th QR-Code logic as follows:

* limit the QR-Code information to bare minimum (per interface):
  * [Interface] A random, on the fly generated 'PrivateKey',
                the key will not be saved on the router
  * [Peer] The 'PublicKey' of that wg interface and the
           'AllowedIPs' with the default of '0.0.0.0/0, ::/0'
           to allow sending traffic to any IPv4 and IPv6 address
* the QR-Code logic will be processed only once per page load,
  and not every 5 seconds with the regular XHR refresh

Feel free to apply this PR if it works for you - thanks!

feckert:
@BlueMax any security hints from your side? If not I will give them a try

Borromini:
Seems to work fine here.

BlueMax:
Not sure if it might raise difficulties if the public key is missing on the far side (as just the private key is passed). Usually both sides have their own full keypair locally available. Technically not necessary but it might have potential to confuse apps (due to missing data). Other than that, looks good to me. Well done!

feckert:
Thanks all for your work and testing.
I will give them now a try, and wait for a feedback from the users/community :-)

Strykar:
Will adding a DNS option for the client to use be considered?

BlueMax:
Its better but not yet complete. 
The page reads:

> • [Interface] A random, on the fly generated 'PrivateKey', the key will not be saved on the router
• [Peer] The 'PublicKey' of that wg interface and the 'AllowedIPs' with the default of '0.0.0.0/0, ::/0' to allow sending traffic to any IPv4 and IPv6 address
Public Key: <openwrt_public_key>
Listen Port: 54321

The line:
> Public Key: <openwrt_public_key>

should be replaced by
> Peer Public Key: <peer_public_key_from_onthefly_privatekey>
Peer Private Key: <peer_onthefly_privatekey>
OpenWrt Public Key: <openwrt_public_key>

Currently, it only shows the <openwrt_public_key>. But we need the <peer_public_key> to be able to enter it into our peer configuration on OpenWrt side. We have to get it from Android side atm.
The <peer_public_key> can be generated with `wg pubkey < peer_onthefly_privatekey > openwrt_public_key`.

OrbitingOcelot:
The QR code as it is currently implemented is useless for the peer setup (for example on an android device).

The use case would be: Generate Peer with pairing info and QR code, open Android wireguard, scan the QR code, setup is done.

In fact this is how it works with another wireguard frontend implementation that I am using. When you add a peer, a QR code can be generated containing the following, setup using this is a breeze and takes a few seconds:

[Interface]
#ifname
PrivateKey=<generated>
Address=10.x.x.x/32

[Peer]
#name
PresharedKey=<generated>
PublicKey=<generated>
Endpoint=ip:port
AllowedIPs=.....

PR #2756, opened 2019-06-06
luci-mod-network: permit WPS push-button on WPA3

Ansuel:
Currently WPS push-button is dropped when SAE or SAE-Mixed is selected. WPS is still supported in WPA3.

@jow- 

### DIFF
--- a/modules/luci-base/po/ca/base.po
+++ b/modules/luci-base/po/ca/base.po
@@ -1762,7 +1762,7 @@ msgid "Enable VLAN functionality"
 msgstr "Habilita la funcionalitat VLAN"
 
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:1595
-msgid "Enable WPS pushbutton, requires WPA(2)-PSK"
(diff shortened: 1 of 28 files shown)


jow-:
Please resync the translations when you change labels. At least update the english key strings so that other translations do not get lost.

Ansuel:
@jow- should i manually update the already translated strings? (think it's bad that the translation got dropped)

hnyman:
> should i manually update the already translated strings? (think it's bad that the translation got dropped)

Yes. I think that for many of the few translated languages it would be relatively easy to add WPA3 to the translated string.  

Dropping existing translations is always bad. Preventing that usually requires some manual work (or sed + manual work).

feckert:
Please rebase your changes

Ansuel:
@feckert can you give me some advice on how to easily change po files? for this?

Ansuel:
@jow- i also updated this

jow-:
Strictly speaking there's no `WPA3-PSK` - a better wording would be `WPA(2)-PSK/WPA3-SAE`

Ansuel:
@jow- fixed

PR #2761, opened 2019-06-10
luci-mod-status: make Overview page configurable

Ansuel:
This is stil WIP. I'm pushing this to have some hint and a quick revision. 
I moved the LuCI setting to a separate lua file as the System page is already very heavy with settings. 
I still don't know if this is the right way of doing this. 
I think a good thing would be to have a shared function to get the list of the included files (as the code to get them is the same) 
Also to me having a cursor in the index page just to check that option looks very unefficent.  
By default every page is active. 
This creates a new section in luci config file called index_display. 
I'm also dubious about the "translation table" for the include file.

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/status/include/70_ddns.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/status/include/70_ddns.js
@@ -5,6 +5,8 @@
 return L.Class.extend({
 	title: _('Dynamic DNS'),
 
+	index_id:' ddns',
+
 	callDDnsGetServicesStatus: rpc.declare({
 		object: 'luci.ddns',
 		method: 'get_services_status',
--- a/applications/luci-app-minidlna/htdocs/luci-static/resources/view/status/include/80_minidlna.js
+++ b/applications/luci-app-minidlna/htdocs/luci-static/resources/view/status/include/80_minidlna.js
@@ -5,6 +5,8 @@
 return L.Class.extend({
 	title: _('miniDLNA Status'),
 
+	index_id: 'minidlna',
+
 	load: function() {
 		return uci.load('minidlna').then(function() {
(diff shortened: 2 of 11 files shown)


Ansuel:
@jow- sorry for the quote any hint about this?

Ansuel:
@jow- sorry for quote can you check this?

jow-:
Whats the use case for this? To me it looks like the common case (display all info) becomes way slower due to all the new filesystem stat syscalls involved for a very narrow, specific use.

What is the situation you want to address where there is a status page include present which should not be shown? If a user really does not want it, there's always the option to simply delete the file.

Ansuel:
Delete the file Is not so user friendly. This can be usefull if someone have lots of info in the overview page and needs to display only he wants.
For example someone wants to display only connected device as other tabs like ddns or upnp or WiFi stats

Ansuel:
Delete the file Is not so user friendly. This can be usefull if someone have lots of info in the overview page and needs to display only he wants.
For example someone wants to display only connected device as he is not interested in other tabs like ddns or upnp or WiFi stats. This would remove the load of unnecessary data. 

I understand the problem with the extra syscall. A solution would be to append _hide to the name of the index file. This should remove the extra syscall but it will be reset on every upgrade. But the main idea was to make the page full customizable to also select the order of the info page.

Have the possibility to hide overview section can also solve the problems of the page Bering overloaded of xhr request and have more Inddx elements for apps like AdBlock or IpBan

Ansuel:
@jow- now that we have client side implementation and luci actually works with promise... Now this should be possible and shouldn't cause any performance loss since the client will actually decide what to load right? 

(Also another suggestion... Would be good to make the index section mobile like we do for the table row)

Ansuel:
@jow- I rebased this. What do you think of this version ?

Ansuel:
approved with conflicts?

jow-:
Superseded by #5224

PR #2767, opened 2019-06-11
luci-mod-status: expand storage index page with mount points

Ansuel:
Expand storage index page with mount points. For custom mounts point we
use the device name and we reference the mount point between ().

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/25_storage.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/25_storage.js
@@ -7,6 +7,20 @@ var callSystemInfo = rpc.declare({
 	method: 'info'
 });
 
+var callMountPoints = rpc.declare({
+	object: 'luci',
+	method: 'getMountPoints',
+	expect: { result: [] }
+});
+
+var MountSkipList = [
+	"/rom",
+	"/tmp",
+	"/dev",
+	"/overlay",
+	"/",
+]
+
 function progressbar(value, max, byte) {
 	var vn = parseInt(value) || 0,
 	    mn = parseInt(max) || 100,
@@ -24,16 +38,33 @@ return baseclass.extend({
 	title: _('Storage'),
 
 	load: function() {
-		return L.resolveDefault(callSystemInfo(), {});
+		return Promise.all([
+			L.resolveDefault(callSystemInfo(), {}),
+			L.resolveDefault(callMountPoints(), {}),
+		]);
 	},
 
-	render: function(systeminfo) {
-		var root = L.isObject(systeminfo.root) ? systeminfo.root : {},
+	render: function(data) {
+		var systeminfo = data[0],
+		    mounts = data[1],
+		    root = L.isObject(systeminfo.root) ? systeminfo.root : {},
 		    tmp = L.isObject(systeminfo.tmp) ? systeminfo.tmp : {};
 
 		var fields = [];
 		fields.push(_('Disk space'), root.used*1024, root.total*1024);
 		fields.push(_('Temp space'), tmp.used*1024, tmp.total*1024);
+
+		for (var i = 0; i < mounts.length; i++) {
+			var entry = mounts[i];
+
+			if (MountSkipList.includes(entry.mount))
+				continue;
+
+			var name = entry.device + ' (' + entry.mount +')',
+			    used = entry.size - entry.free;
+
+			fields.push(name, used, entry.size)
+		}
 
 		var table = E('table', { 'class': 'table' });
 
--- a/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status-index.json
+++ b/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status-index.json
@@ -19,9 +19,10 @@
 	},
 
 	"luci-mod-status-index-storage": {
-		"description": "Grant access to Storage status display",
+		"description": "Grant access to Storage and Mount status display",
 		"read": {
 			"ubus": {
+				"luci": [ "getMountPoints" ],
 				"system": [ "info" ]
 			}
 		}
--- a/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status-index.json
+++ b/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status-index.json
@@ -18,6 +18,15 @@
 		}
 	},
 
+	"luci-mod-status-index-storage": {
+		"description": "Grant access to Storage status display",
+		"read": {
+			"ubus": {
+				"system": [ "info" ]
+			}
+		}
+	},
+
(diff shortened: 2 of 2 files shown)


Ansuel:
@jow- sorry for quote can you check this?

modules/luci-mod-status/luasrc/controller/admin/status.lua
```diff
@@ -5,6 +5,9 @@
 module("luci.controller.admin.status", package.seeall)
 
 function index()
+	
+	local hasAccess = nixio.fs.access
```
feckert: Is this necessary?

Ansuel:
Resolved by mistake... Anyway... Write that as the same function is repeated 3 times. Isn't better that way? I really don't know in performance perspective if a local variable is better than call 3 times the same function directly.

modules/luci-mod-status/luasrc/controller/admin/status.lua
feckert: I would like prefere useing the following checks:

local has_wireless = fs.access("/usr/bin/wireless")
local has_fstab = fs.access("/etc/config/fstab")
local has_block = fs.access("/sbin/block")

This will be used in LuCI

Ansuel:
@jow- @feckert I rebased this and made some changes... Can you review this? 

Can I use => or should I use the normal syntax? 
(As we use promise... i think => syntax is full supported right?)

jow-:
Please use the old syntax for now.

Ansuel:
Done.

ysc3839:
> As we use promise...

No, because there's polyfill for promise.

Ansuel:
@feckert can you check this?

feckert:
The mounts are shown for my system.
But with luci-theme-openwrt-2020, the whole thing doesn't look right to me?

**Screenshot for luci-theme-openwrt-2020:**
Screenshot_2021-08-05 VR2-106149 - Übersicht - LuCI

**Screenshot for luci-theme-bootstrap:**
Screenshot_2021-08-05 VR2-106149 - Übersicht - LuCI(1)

**Screenshot for luci-theme-material:**
Screenshot_2021-08-05 VR2-106149 - Übersicht - LuCI(2)

Ansuel:
Thx for testing. will check and fix the problem with the new theme (probably using old table definition?)
Any idea how to improve this? Or it's good like that and the info provided are sufficent?

Ansuel:
@feckert fixed display bug with 2020 theme

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/20_mount.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+'require fs';
+'require rpc';
+
+var callMountPoints;
```
feckert, line 5: You could delete this line and add var to this line.
https://github.com/openwrt/luci/pull/2767/files#diff-389d15d60f4efa0b48e4381ec566f58603780d102182d1c9071293a9be6c24bcR7

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/20_mount.js
```diff
@@ -0,0 +1,50 @@
...
+			E('tr', { 'class': 'tr table-titles' }, [
+				E('th', { 'class': 'th' }, _('Device')),
+				E('th', { 'class': 'th' }, _('Mount Point')),
+				E('th', { 'class': 'th' }, _('Available') + ' / ' + _('Size'))
+			])
+		]);
+
+		cbi_update_table(table, mounts.filter( function(mount) {
+				return mount.device.match('/dev/sd*') || 
+					mount.mount.match('/overlay')
```
feckert, line 39: On my system I have the following output for `ubus call luci getMountPoints`
```
{
        "result": [
                {
                        "avail": 0,
                        "device": "/dev/root",
                        "free": 0,
                        "mount": "/rom",
                        "size": 21495808
                },
                {
                        "avail": 61038592,
                        "device": "tmpfs",
                        "free": 61038592,
                        "mount": "/tmp",
                        "size": 62152704
                },
                {
                        "avail": 75231232,
                        "device": "/dev/ubi0_3",
                        "free": 79314944,
                        "mount": "/overlay",
                        "size": 80576512
                },
                {
                        "avail": 75231232,
                        "device": "overlayfs:/overlay",
                        "free": 79314944,
                        "mount": "/",
                        "size": 80576512
                },
                {
                        "avail": 13180928,
                        "device": "/dev/ubi0_0",
                        "free": 13918208,
                        "mount": "/permdata",
                        "size": 13955072
                },
                {
                        "avail": 524288,
                        "device": "tmpfs",
                        "free": 524288,
                        "mount": "/dev",
                        "size": 524288
                }
        ]
}
```
I would only filter for `mount.device` and not also for `mount.mount` if `mount.device` does not match. With your filter my device `/dev/ubi0_0` does not match and so will not get shown.
Interesting matches from my point of view would be:
`/dev/ubi*`
`overlayfs:/overlay`
`/dev/sd*`

Ansuel: I didn't include overlayfs:/overlay as it was redundant to have 2 row with the same values. Am I wrong?

feckert: You're right about that. I'm not sure which is better either.
I would just show everything and not add any special treatment.
At some poin in the feature something will change again and then you have to look at why it is not being displayed now. The user has to know what he is seeing and how to interpret the data.

I would report everything. If we adopt this, then my mount points will not be displayed.

Ansuel: Was thinkg of displaying only ubi and sd* mount. What do you think? Or display a copy of the table in the mount page?

I mean after all we receive all the mount so I think you are right and we should display all of them unconditionally.

feckert: Let's start like this. Can you please add the acl? And change the var (see above)?

Ansuel: Sorry you mean display all or display sda and ubi ?

feckert:
This is how it looks with me.
From my point of view this fits.

Screenshot_2021-08-11 st-dev-07 - Übersicht - LuCI

jow-:
Could you try to display the space usage per mount as progress bar and only display the actual numbers as tooltip?

Ansuel:
@jow- you mean like this?
image

or the value hidden and displayed only on the tooltip ?

Ansuel:
@feckert considering we have a variant of this implemented... should i close this or the implementation we have still lacks some feature that are here?

Ansuel:
@feckert @jow- I found time to refactor this. I decided to expand storage instead of adding a new index page. I would ask to also test this and tell me if the concept is ok?

PR #261, opened 2014-12-01
Add luci support package for sstp-client

harveyhu2012:
Add luci support package for sstp-client, make it easy to use.

<https://www.dropbox.com/s/n1wsem74pjcjhyu/luci-proto-sstp.png?dl=0>

PR #2768, opened 2019-06-11
Add SSTP Protocol Support for LUCI

rkkoszewski:
This PR is based on an older PR that can be seen here https://github.com/openwrt/luci/pull/261 (Commit: https://github.com/openwrt/luci/commit/3854b5ac2647a90839f075b8c8af4a4ea33bdf89)

PLEASE NOTE: In order for this to work the sstp-client needs to have the netifd sstp.sh file added. In other words, this PR depends on https://github.com/openwrt/packages/pull/9221 PR to be completed and won't work without it.


feckert:
Please stash  / fixup the commits into the first one so we have only on commit. 
And please add a new line on the button on some files. They are marked in github

rkkoszewski:
Hi @feckert, I rebased my commits into a single one, but I cannot do the same for the Merge commit as it seems that some commits have some untracked files in them (And appear as Signed off by me after the rebase) and other merge conflicts that surface during rebasing. 

Can you @feckert or anybody else suggest how to properly perform the rebase?

feckert:
Sorry I would merge this changes but the whole pullrequest is broken.
And it is not easy to solve this.

The new luci-proto handler looks good to me. :+1: 

Please open a **new** pullrequest with your changes.
And reference this one for the new fixed follow up pullrequest.
Thanks.

PR #2769, opened 2019-06-11
luci-base: update Traditional Chinese translation

rypervenche:
Update "gateway" translation using standardized word used by both Google and Microsoft.

### DIFF
--- a/modules/luci-base/po/zh-tw/base.po
+++ b/modules/luci-base/po/zh-tw/base.po
@@ -1314,7 +1314,7 @@ msgstr ""
 #: protocols/luci-proto-ipv6/luasrc/model/cbi/admin_network/proto_dslite.lua:32
 #: protocols/luci-proto-ipv6/luasrc/model/cbi/admin_network/proto_map.lua:67
 msgid "Default gateway"
-msgstr "預設匝道器"
+msgstr "預設閘道"
 
 #: modules/luci-mod-network/luasrc/model/cbi/admin_network/ifaces.lua:541
 msgid "Default is stateless + stateful"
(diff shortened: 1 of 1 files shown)


yousong:
@a962702 please help review, thanks

@rypervenche the commit needs to be signed off with your real name.  Check existing commits in this repo to find out what's it like.

yousong:
For ref, wikipedia says it should be ["閘道器"](https://zh.wikipedia.org/zh-tw/%E7%BD%91%E5%85%B3)

a962702:
1. In wikipedia, it should be [閘道器](https://zh.wikipedia.org/zh-tw/%E7%BD%91%E5%85%B3), and the [wiktionary](https://zh.wiktionary.org/wiki/gateway) say it can be 閘道 or 閘道器.
2. According to the dictionary by National Academy for Educational Research in Taiwan, it can be [閘道](http://dict.revised.moe.edu.tw/cgi-bin/cbdic/gsweb.cgi?o=dcbdic&searchid=Z00000114700) or [閘道器](http://dict.revised.moe.edu.tw/cgi-bin/cbdic/gsweb.cgi?o=dcbdic&searchid=Z00000114701)

So use 閘道 or 閘道器 are the same, but the original word 匝道器 is not correct.

rypervenche:
@yousong Sorry about that. I have signed off the commit. I chose the first version without 器 as both [Microsoft](https://www.microsoft.com/en-us/language/Search?searchTerm=gateway&langID=129&Source=true&productid=undefined) and [Google](https://support.google.com/a/answer/60730?hl=zh-Hant) seem to prefer the two-character version in most cases (see the two links).

yousong:
Thanks!

PR #2770, opened 2019-06-14
luci-proto-ncm: add dialnum option

geekinaboxx:
This option is useful for setting the dial number used for data service connection

Signed-off-by: geekinaboxx <aleem@tectonic.it>

### DIFF
--- a/protocols/luci-proto-ncm/luasrc/model/cbi/admin_network/proto_ncm.lua
+++ b/protocols/luci-proto-ncm/luasrc/model/cbi/admin_network/proto_ncm.lua
@@ -12,7 +12,7 @@ You may obtain a copy of the License at
 
 local map, section, net = ...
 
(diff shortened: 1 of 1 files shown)


feckert:
Please stash the two commits to one and also add a `Signed-off-by:` to your commit

geekinaboxx:
Thanks. Done.

feckert:
Sorry I have missed that we should also change the translation from `translate("Modem dial string")` to `translate("Dial number")` so we could reuse the translation from https://github.com/openwrt/luci/blob/master/protocols/luci-proto-3g/luasrc/model/cbi/admin_network/proto_3g.lua#L45 and do not have to introduce a new one to the po files.

geekinaboxx:
no problem; nice catch. Done.

feckert:
Thanks merged!

PR #2772, opened 2019-06-16
luci-app-aria2: fix: datatype

jjm2473:
just a spelling mistake.


hnyman:
Thanks.
Your "datetype" finding made me to do a github search for it, and there seems to be the same error in some other apps, too. :-(   I will fix them.
Thanks for pointing it out.

jjm2473:
Yes, is a common mistake.

PR #2775, opened 2019-06-17
luci-app-opkg: made 'force-overwrite' user-driven/optional plus bugfix

dibdot:
* made 'force-overwrite' flag optional, the user could set this
  like 'autoremove' today (see screenshots)
* fix erroneously as 'installed' reported packages (happens whenever
  a package installation runs into an error)
* add three dots to the first level 'Remove' button, cause it
  leads to a modal dialogue

### DIFF
--- a/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
+++ b/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
@@ -135,8 +135,9 @@ function display(pattern)
 
 		if (currentDisplayMode === 'updates') {
 			var avail = packages.available.pkgs[name];
-			if (!avail || compareVersion(avail.version, pkg.version) <= 0)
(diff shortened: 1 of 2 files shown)


dibdot:
New workflow/result without force:
opkg1
opkg2

New workflow with force:
opkg3
opkg4

Maybe that's useful ... at least the 'installed' bug should be fixed.

dibdot:
@jow- @BKPepe

jow-:
I think if a package is not installed it should not be part of the installed dictionary. The tests look like they paper over an underlying issue.

BKPepe:
I like this! :) Thank you.

jow-:
Merged with some changes. Thanks @dibdot

BKPepe:
Would it be possible to include it also in OpenWrt 19.07?

PR #2779, opened 2019-06-21
luci-app-statistics: add address family configuration for ping

hwti:
It allows to force ipv4 or ipv6 when the DNS returns both addresses, but
only one works (for example if there is no ipv6 connectivity).

Signed-off-by: Loïc Yhuel <loic.yhuel@gmail.com>


feckert:
I did not find any hint in the documentation about this option. But I have seen that this option is supported in the source https://github.com/collectd/collectd/blob/collectd-5.9/src/ping.c#L91 for version 5.9.
I will merge this.
Thanks.

PR #2791, opened 2019-06-26
luci-proto-3g: rename maxwait to delay option

feckert:
The option maxwait is not support by the 3g netifd proto.
To standardize that rename the maxwait option to delay as in the other
proto handlers luci-proto-qmi and luci-proto-ncm.


feckert:
This dependenc on https://github.com/openwrt/openwrt/pull/2166

feckert:
Dependency hast ween add to openwrt master. So I will merge this.

PR #2793, opened 2019-06-26
luci-theme-material: make 'Loading...' text translatable

sch-m:

### DIFF
--- a/themes/luci-theme-material/luasrc/view/themes/material/header.htm
+++ b/themes/luci-theme-material/luasrc/view/themes/material/header.htm
@@ -219,7 +219,7 @@
 	</div>
 </header>
 <div class="main">
-	<div style="" class="loading"><span><div class="loading-img"></div>Loading...</span></div>
+	<div style="" class="loading"><span><div class="loading-img"></div><%:Loading...%></span></div>
 	<div class="main-left">
 		<% render_topmenu() %>
 	</div>


jow-:
Merged, thanks!

sch-m:
@jow- Thanks for merging this.
While I wanted to integrate the msgid "Loading..." into the base.po I noticed that a msgid "Loading" already exists there. I don't want to add more variants of the same text.

What do you think about changing the 
`<%:Loading...%>` into `<%:Loading%>...`
This allows us to reuse the msgid "Loading" and the "..." are still there (untranslated).

jow-:
In principle okay, but I am not sure if the ellipsis is applicable like this to all languages (think Hebrew (rtl), Chinese etc...)

@yousong ?

sch-m:
> In principle okay, but I am not sure if the ellipsis is applicable like this to all languages (think Hebrew (rtl), Chinese etc...)

Oh, good point. Alternatively, we could just eliminate the ellipsis.

jow-:
Alternatively you maybe could use the `Loading data...` string.

sch-m:
Sorry, can't find any `Loading data...`string. Where is it? :-)

jow-:
Nvm, I was misremembering "Collecting data..."

sch-m:
Ah, nice. Ok, so will change the text in luci-theme-material to `<%:Collecting data...%>`

sch-m:
and send a new pr, ok?

jow-:
yes, please

PR #2803, opened 2019-06-28
luci-base: fix display of password input

### LINKED ISSUE #2624: Password input rendered wrong when embedded in table cell
The issue was originally reported as a followup in issue https://github.com/openwrt/luci/issues/2398#issuecomment-450495221.  but it got ignored there after the original issue got resolved and closed.

The issue is still present in current luci master.

Below is a copy of the original report

> I understand.  Previously browser autocomplete can indeed cause trouble.  But it seems when embedded in a table, the added input element will be visible, despite the css code `left: -1000px`
> 
> screen shot 2018-12-29 at 9 56 39 pm
> 
> Another minor undesirable change is that previously the `Reveal/hide password` button stick to the input element on the same row, not wrapping to the next line.
> 
> screen shot 2018-12-29 at 10 00 55 pm

yousong:
Should fix #2624

### DIFF
--- a/modules/luci-base/luasrc/view/cbi/cell_valueheader.htm
+++ b/modules/luci-base/luasrc/view/cbi/cell_valueheader.htm
@@ -3,7 +3,7 @@
 	local descr = luci.util.trim(striptags(self.description))
 	local ftype = self.typename or (self.template and self.template:gsub("^.+/", ""))
 -%>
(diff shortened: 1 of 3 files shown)


yousong:
Merged to fix the issue at hand.  The chance of causing unexpected effect should be low.

The change will also need to be backported to branch 19.07

yousong:
Backported to both 19.07 and 18.06

PR #2758, opened 2019-06-08
Fix a style issue on the dashboard page

HouCoder:
I just found a style issue on the dashboard pages, it happens in multiple themes, here is the PR to fix it.

image
image

PR #2804, opened 2019-06-28
Fix a overview page style issue for various themes

namedun:
This PR should fix a style issue with interface boxes, which is described in PR #2758.

Tested with browsers:
- Google Chrome 75.0.3770.100
- Mozilla Firefox 67.0.4
- Opera 60.0.3255.170
- Internet Explorer 11.0.9600.18817

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -1985,6 +1985,7 @@ table table td,
 
 .network-status-table .ifacebox-body > span {
 	flex: 10 10 auto;
+	height: 100%;
 }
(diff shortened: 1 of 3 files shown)


jow-:
LGTM, merged. Thanks!

hotpxl:
Thanks! This fixed it for me. While I'm really just learning this, could you please explain why adding `height: 100%` to the `span` under `.ifacebox-body` fixed the issue? What confuses me most is that the parent div has a top-down flex, and what would setting height to 100% on a child mean?

namedun:
I'm sorry, but I'm not a CSS expert and I can't explain why it works. I made this change intuitively and tests showed that it worked. I can only assume that the Сhrome has a bug, as for all other browsers this change does not affect anything and is unnecessary.

Issue #2800, opened 2019-06-27
 (E011) invalid escape sequence '\*'


rubo77:
https://github.com/openwrt/luci/blob/ae1652fc116f542dcefba275d1e7d2f9c6564e12/modules/luci-base/luasrc/cbi/datatypes.lua#L421

Shouldn't this be escaped with `%`?

     "^[0-9%*#!%.]+$"

I get some errors with `luacheck`:
 ```

 openwrt/feeds/luci/modules/luci-base/luasrc/cbi/datatypes.lua:417:26: (E011) invalid escape sequence '\*'
 openwrt/feeds/luci/modules/luci-base/luasrc/util.lua:210:34: (E011) invalid escape sequence '\-'
 openwrt/package/feeds/luci/luci-app-ocserv/luasrc/model/cbi/ocserv/users.lua:29:27: (E011) invalid escape sequence '\$'
 openwrt/package/feeds/luci/luci-app-travelmate/luasrc/model/cbi/travelmate/wifi_edit.lua:32:33: (E011) invalid escape sequence '\+'
```

jow-:
Fixed by #2806

PR #2806, opened 2019-06-29
system-wide: fix escaping of magic chars

dibdot:
* fix escaping of magic lua chars (#2800)
* fix redundant second gsub line in shellstartsqescape function
* fix return value of shellstartsqescape function


dibdot:
@jow- I'm quite unsure if the changes in util.lua are correct!?

jow-:
Looked correct - merged, thanks!

PR #2811, opened 2019-07-01
luci-proto-qmi: add missing mtu option

feckert:
Add missing mtu option.
This dependence on https://github.com/openwrt/openwrt/pull/2176


feckert:
Dependency hast ween add to openwrt master. So I will merge this.

PR #2817, opened 2019-07-03
luci-base: move some generic classes into a separate luci-base-libs package

SvenRoederer:
The new package luci-base-libs provides the modules that not strictly relate to the web-interface of luci. By separating these libs they can be used by other packages without having to install the web-components. 
This change was inspired by providing a shell-only interface for 4MB-flash devices, by keeping as much code common with a full install. Dropping the luci-web stuff reduces the package by 100kb.


SvenRoederer:
a later version of this code should not have a separate Makefile, the luci-base packge should be modified to generate both packages.

SvenRoederer:
@feckert  just seen that you marked it with WIP-flag and not gave it a NACK yet. So I'll continue to work on this.

SvenRoederer:
I pushed the missing changes to finish the integration of the rearranged libraries. I ran a test based on "luci-mod-admin-full".

@feckert can you review?

SvenRoederer:
Just rebased to recent master. Any other issues preventing from merging?

SvenRoederer:
@jow- @feckert any opinions on this change?
We are using it for some months in our freifunk firmware and were able to improve the support for 4MB devices.

feckert:
@SvenRoederer from my side ok, but that's up to @jow- if he wants to merge this.

A little question then. If you don't need the LuCI you have still include the whole LuCI feed repo during compilation and build. Doesn't it make then not more sense to move it into the package feed?

SvenRoederer:
our package ```luci-app-owm-cmd``` (https://github.com/freifunk-berlin/firmware-packages/blob/master/utils/luci-app-owm) is using the ```luci-lib-httpclient``` package which  pulls in luci-base. The 4MB devices we use w/o web-interface and don't need uhttpd and the web-framework. As ```luci-app-owm-cmd``` is just a cmdline-app to report some data, we like to use it on these devices, which will pull also the full web-framework as of the mentioned dependency.
Solving this ```luci-app-owm-cmd```-dependency by moving some packages to OpenWrt-packages would recursively require a bunch of packages to be moved. As our images for other routers will include the full LuCI, we need the luci-feed anyways.

SvenRoederer:
@jow- What do you think on this?

SvenRoederer:
@jow- just seen that your commit cf1219bd0790506279b14007595e8389e5eb99ff caused conflict of my PR.
What do you think of the intention of this PR?

SvenRoederer:
@jow- ping ... I would be happy to get your feedback on this idea.

SvenRoederer:
@jow- @NeoRaider @hnyman as @feckert mentioned above (https://github.com/openwrt/luci/pull/2817#issuecomment-558991698) this PR looks fine. 
What do you think on merging?

jow-:
Did you actually runtime test all LuCI views with this change? I do see that dispatcher.lua still references `util.striptags` and `util.pcdata` in the `tpl.context.viewns` meta table while these functions were moved to a new XML class. That will likely break rendering a bunch of templates.

SvenRoederer:
This was introduced when rebasing  from 83408a1 to ce8fb70.  Pushing an updated version which also updates the docs.

SvenRoederer:
Just pushed a rebased version.

SvenRoederer:
@jow- @NeoRaider @hnyman I fixed the missing lines mentioned in https://github.com/openwrt/luci/pull/2817#issuecomment-604872131 and tested the code on a recent master (551d839f90853b58f81d124c77442009d4df0) successfully.

Would be great if you accept this for master.

SvenRoederer:
@jow- @NeoRaider @hnyman @feckert ping

NeoRaider:
@SvenRoederer I'm really not the right person to ping about LuCI packaging - I personally don't use LuCI, and Gluon doesn't even include the feed anymore.

SvenRoederer:
@jow- @hnyman @feckert Any feedback on this would be great.

feckert:
@SvenRoederer Sorry I can't help you there either!
This was created by @jow-.
He has the last word on that.

jow-:
I don't think we should move `pcdata()` to a new `luci.xml` class. This has the potential to break a lot 3rd party out of tree code.

SvenRoederer:
As mentioned this change was driven by keeping some commandline tools working on 4MB devices which don't provide a web-ui (no uhhtp, no luci web-framework). 
I don't remember the exact details anymore, but it turned out that one of the libs in the new package depends on the striptags() and or the pcdata() function of luci.utils. But luci.utils depends on the luci.dispatcher and its dependencies. This way half of the stuff, required for the webframework ended up in the new "small" package. As this mixup of having web-framework related thing in the stripped down "non-web" package, which did not made sense at all. 
So the most clean solution I found, was to refactor the functions into this new module, to break the dependency chain.

>  This has the potential to break a lot 3rd party out of tree code.

@jow- Every change has this potential, but I think the change of having luci authenticate trough the rpcd had much more consequences. Here the worst case is adding a new "require" line and replacing "util" by "xml". 
And we can gain 100kb of flash.

jow-:
Please add wrappers to luci.util restoring this functionality. You can lazy-load luci.xml if needed.

SvenRoederer:
@jow- added a wrapper to luci.util which calls the luci.xml functions. Also added a message that the code should be updated to use luci.xml directly.

SvenRoederer:
@jow- can you check again for a merge? Added the compatibility-wrapper
https://github.com/openwrt/luci/pull/2817/files#diff-22b6303e1f39e747a12833e3814328d4R163; https://github.com/openwrt/luci/pull/2817/files#diff-22b6303e1f39e747a12833e3814328d4R190

SvenRoederer:
@jow- Did you had a chance to review this PR again after adding the compatibility-wrapper as requested?

jow-:
Merged via 8b8d83e42dd3d10d82e29a8614a7b3e3e94b16c6, 68521fca04e9b2ae801cc2daa4f920a2c43bc03f, 354e4cb4a70d576a4f02e024f3c8fb1e72f09350, 7edd635026594da577b73f059c0d8b95d653f8fc

SvenRoederer:
Thanks

PR #2818, opened 2019-07-03
luci-app-statistics: more extensibility

mberndt123:
Hi all, hi @jow-,

I recently wrote my own collectd plugin and I wanted to be able to display the resulting data in OpenWrt using RRDtool. Of course luci-app-statistics doesn't support my collectd plugin, and there isn't currently a way to non-intrusively add support for new collectd plugins to luci-app-statistics. Therefore I decided to create this PR to make it possible to extend luci-app-statistics without modifying the source code.

Today there's a .lua file for every supported collectd plugin in `/usr/lib/lua/luci/model/cbi/luci_statistics`. To add support for a new plugin one needs to add a new model there, but one also needs to add a bit of extra information in some hard-coded tables in the luci-app-statistics source code. The idea is to load this information from files in `/usr/lib/lua/luci/statistics/plugins` instead. A third-party developer can then simply dump some `.lua` files in those directories and luci-app-statistics will pick them up; zero source code modifications are necessary. It should also work nicely with package managers: you can package your collectd plugin with these two lua files and things will just work.

What do you think about the idea? I haven't done a lot of Lua before, so I'd be grateful for any comments you might have!

### DIFF
--- a/applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
+++ b/applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
@@ -22,52 +22,24 @@ function index()
 		s_output	= _("Output plugins"),
 		s_general	= _("General plugins"),
 		s_network	= _("Network plugins"),
-
-		apcups		= _("APC UPS"),
-		conntrack	= _("Conntrack"),
-		contextswitch	= _("Context Switches"),
-		cpu		= _("Processor"),
-		cpufreq		= _("CPU Frequency"),
-		csv		= _("CSV Output"),
-		curl		= _("cUrl"),
-		df		= _("Disk Space Usage"),
-		disk		= _("Disk Usage"),
-		dns		= _("DNS"),
-		email		= _("Email"),
-		entropy		= _("Entropy"),
-		exec		= _("Exec"),
-		interface	= _("Interfaces"),
-		iptables	= _("Firewall"),
-		irq		= _("Interrupts"),
-		iwinfo		= _("Wireless"),
-		load		= _("System Load"),
-		memory		= _("Memory"),
-		netlink		= _("Netlink"),
-		network		= _("Network"),
-		nut		= _("UPS"),
-		olsrd		= _("OLSRd"),
-		openvpn		= _("OpenVPN"),
-		ping		= _("Ping"),
-		processes	= _("Processes"),
-		rrdtool		= _("RRDTool"),
-		sensors		= _("Sensors"),
-		splash_leases	= _("Splash Leases"),
-		tcpconns	= _("TCP Connections"),
-		thermal		= _("Thermal"),
-		unixsock	= _("UnixSock"),
-		uptime		= _("Uptime")
 	}
 
 	-- our collectd menu
 	local collectd_menu = {
-		output  = { "csv", "network", "rrdtool", "unixsock" },
-		general = { "apcups", "contextswitch", "cpu", "cpufreq", "df",
-			"disk", "email", "entropy", "exec", "irq", "load", "memory",
-			"nut", "processes", "sensors", "thermal", "uptime" },
-		network = { "conntrack", "curl", "dns", "interface", "iptables",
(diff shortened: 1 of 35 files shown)


feckert:
Looks pretty good in principle.
But the translations for the labels are missing.

@jow Also I would rather write the configuration directly in future from the collectd because luci-app-statistics did hijack the config generation. I have already prepared add a uci binding to collectd. See https://github.com/openwrt/packages.

jow-:
@feckert - feel free to implement these changes, I don't have enough time to take care of this at the moment.

applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
```diff
@@ -36,7 +36,7 @@ function index()
 		local plugin = dofile(plugin_dir .. filename)
 		local name = filename:gsub("%.lua", "")
 		table.insert(collectd_menu[plugin.category], name)
-		labels[name] = plugin.label
+		labels[name] = _(plugin.label)
```
jow-: That will not work. The `_()` function must be invoked with a literal string since it is parsed by a static source code scanner to extract translatable phrases.

mberndt123: I see… So how should this be done? Should it be invoked in each individual .lua file in the plugins directory?

jow-: Right, in each individual file, you'll need something like  `label = translate("Foo")`, then it should work. But I didn't look in detail

mberndt123: Oh, why translate("Foo") and not _("Foo") like in the original file? Is the translate function globally visible?

mberndt123: OK, simply changing "Foo" to translate("Foo") doesn't work:
```
/usr/lib/lua/luci/statistics/plugins/apcups.lua:7: attempt to call global 'translate' (a nil value)
```
Any hints for how to make this work?

feckert: I have done the following to add the translation. See https://github.com/openwrt/luci/blob/master/applications/luci-app-statistics/luasrc/statistics/rrdtool/definitions/apcups.lua#L7
please try if this helps.

feckert:
Also please rebase your change with `git rebase -i HEAD~5` and then add a fixup to the last 4 commits in the vi. After that make a `git push --force`. We do not need the extra commits.

feckert:
> @feckert - feel free to implement these changes, I don't have enough time to take care of this at the moment.

Ok i could/will do that. The configuration is on some corners of collected difference to the luci-app-statistics configuration. Is this a Problem? Do we need a upgrade script for this cases?

applications/luci-app-statistics/luasrc/controller/luci_statistics/luci_statistics.lua
mberndt123: Hey @feckert, I think I got this working correctly. I'm now loading the file with loadfile rather than dofile, and I use `setfenv(f, { _ = luci.i18n.translate })` to give the loaded code access to the i18n stuff. The source code scanner that @jow- mentioned should probably work like before because the code looks the same `_("Bla")`.

mberndt123:
> Also please rebase your change with `git rebase -i HEAD~5` and then add a fixup to the last 4 commits in the vi. After that make a `git push --force`. We do not need the extra commits.

It should be OK now.

jow-:
I like this change, LGTM.

feckert:
I will do some test and if this is OK i will merge this in the next hours

feckert:
I have made some test and it works as expected.

Thanks or your work

feckert:
@mberndt123 I'm afraid I missed it!
Next time, please add a signed-off-by: to your work:

@jow- Sorry I was going too fast :-(
I think I need to do a checklist of what I need to consider.

PR #2822, opened 2019-07-05
luci-app-travelmate: update Japanese translation

musashino205:
- sync translations
- update Japanese translations


dibdot:
Thanks!

PR #2835, opened 2019-07-07
build: add circleci

aparcar:

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,0 +1,42 @@
+FROM debian:9
+
+
+# Configuration version history
+# v1.0   - Initial version by Etienne Champetier
+# v1.0.1 - Run as non-root, add unzip, xz-utils
+# v1.0.2 - Add bzr
+
+RUN apt update && apt install -y \
+build-essential \
+curl \
+jq \
+gawk \
+gettext \
+git \
+libncurses5-dev \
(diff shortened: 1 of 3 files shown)


BKPepe:
Is this used somehow in the LuCI repository? I am not able to see any check nor that it does compile anything. Anyway, it uses an old version of the docker image. The current is 1.0.5, but in LuCI, there is 1.0.2.

BKPepe:
Ping @aparcar!

Let's notify also @jow- and @hnyman as well.

hnyman:
I am not sure if @jow- or @aparcar ever actually activated the integration to circleci.

And in any case, if it is activated, the circleci recipe should be updated to the current one to handle 19.07 etc.

aparcar:
Ugh this is badly out of date. Should we update or remove it? Compiling LuCI packages is somewhat unlikely to fail (usually no real compiling) and checking a SoB line can be done by a simple GitHub action, so no need of some complex external scripting.

hnyman:
we could remove it.

Like you say, there is not much actual compiling involved in the LuCI repo itself.
I also feel that making circleci to compile the related dependency packages from the packages feed for e.g. a small change in statistics interface, is unnecessary.

aparcar:
@hnyman please merge this https://github.com/openwrt/luci/pull/3776

PR #2842, opened 2019-07-08
[18.06]build: add circleci

aparcar:

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,0 +1,42 @@
+FROM debian:9
+
+
+# Configuration version history
+# v1.0   - Initial version by Etienne Champetier
+# v1.0.1 - Run as non-root, add unzip, xz-utils
+# v1.0.2 - Add bzr
+
+RUN apt update && apt install -y \
+build-essential \
+curl \
+jq \
+gawk \
+gettext \
+git \
+libncurses5-dev \
+libssl-dev \
+python \
+subversion \
+bzr \
+time \
+wget \
+zlib1g-dev \
+unzip \
+xz-utils \
+&& rm -rf /var/lib/apt/lists/*
+
+RUN useradd -c "OpenWrt Builder" -m -d /home/build -s /bin/bash build
+USER build
+ENV HOME /home/build
+
+# LEDE Build System (LEDE GnuPG key for unattended build jobs)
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,144 @@
+version: 2.0
+jobs:
+  build:
+    docker:
+      - image: docker.io/openwrtorg/packages-cci:v1.0.2
+    environment:
+      - SDK_HOST: "downloads.openwrt.org"
+      - SDK_PATH: "snapshots/targets/arx71xx/generic"
+      - SDK_FILE: "openwrt-sdk-18.06.1-ar71xx-generic_*.Linux-x86_64.tar.xz"
```
hnyman: Is there a reason to specify the ancient 18.06.1 as the SDK version? Why not the recent 18.06.4 or the snapshot-based ongoing SDK from buildbot: 
https://downloads.openwrt.org/releases/18.06-SNAPSHOT/targets/ar71xx/generic/openwrt-sdk-18.06-SNAPSHOT-ar71xx-generic_gcc-7.3.0_musl.Linux-x86_64.tar.xz
as generically
openwrt-sdk-18.06*-ar71xx-generic_*.Linux-x86_64.tar.xz

I noticed that the 8 month old 18.06 packages repo circleci definition seems to contain 18.06.1, but that is likely just an oversight, or how? @thess  @champtar

.circleci/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# LEDE Build System (LEDE GnuPG key for unattended build jobs)
+RUN curl 'https://git.openwrt.org/?p=keyring.git;a=blob_plain;f=gpg/626471F1.asc' | gpg --import \
+ && echo '54CC74307A2C6DC9CE618269CD84BCED626471F1:6:' | gpg --import-ownertrust
+
+# LEDE Release Builder (17.01 "Reboot" Signing Key)
+RUN curl 'https://git.openwrt.org/?p=keyring.git;a=blob_plain;f=gpg/D52BBB6B.asc' | gpg --import \
+ && echo 'B09BE781AE8A0CD4702FDCD3833C6010D52BBB6B:6:' | gpg --import-ownertrust
+
+# OpenWrt Release Builder (18.06 Signing Key)
```
hnyman: The signing keys here do not match the new ones, recently updated by @jow- 
https://git.openwrt.org/?p=keyring.git;a=shortlog

I am not sure about their need here in circleci, but likely the new/current keys should be imported in addition to / instead of these old deprecated(?) keys.

PR #2843, opened 2019-07-08
[19.07]build: add circleci

aparcar:

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,0 +1,42 @@
+FROM debian:9
+
+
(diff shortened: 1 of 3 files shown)


hnyman:
This will not work any more, as for some reason @ynezz changed ath79 to be "source-only" in 19.07 in commit https://github.com/openwrt/openwrt/commit/2b074654b0f259518aa56e0975ca8e26c0c12bc9. 

So, apparently there will be no ath79 binaries from buildbot in 19.07, and thus ath79 should not be used as the SDK target.

PR #2844, opened 2019-07-09
luci-app-statistics: fix collectd config generation

feckert:
Fix config generation for the following sections:

* curl
* exec
* network
* iptables

This was added with the following commit.
https://github.com/openwrt/luci/commit/c1380ab36797dc4a9cba5a6aa738019452178d79

### DIFF
--- a/applications/luci-app-statistics/root/usr/bin/stat-genconfig
+++ b/applications/luci-app-statistics/root/usr/bin/stat-genconfig
@@ -276,15 +276,11 @@ plugins = {
 		{ },
 		{ }
 	},
-	curl = config_curl,
-	exec	= config_exec,
-	iptables = config_iptables,
 	logfile	= {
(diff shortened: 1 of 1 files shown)


jow-:
Please add `Fixes: c1380ab ("make luci-app-statistics more extensible")` before the signed-off-by. Otherwise LGTM

mberndt123:
@feckert , sorry for breaking this, and thanks for the fix.

However wouldn't it be simpler to just change the code from `plugins[name] = plugin.legend` to
```
if plugin.legend then
  plugins[name] = plugin.legend
end
```
Then you don't need a long cascade of if statements…

feckert:
@mberndt123 on problem I quickly found a solution to solve this issue.

> However wouldn't it be simpler to just change the code from plugins[name] = plugin.legend to

feel free to open another PR

PR #2845, opened 2019-07-09
luci-base: fix id for UICheckbox and UITextfield

namedun:
Clicking on labels of the checkbox or value edit does not work.
This is because the id is assigned to the parent div, not the input element.

I tried to fix it in this pull request, but I'm not sure that I did it right.


jow-:
Unfortunately this isn't the right fix, the ID has to stay at the toplevel element since it is used to find the entire widget markup.

PR #2857, opened 2019-07-13
luci-base: update Japanese translation

musashino205:
- fix translatable text
- sync translations
- update Japanese translation


yousong:
Thank you!

Issue #2878, opened 2019-07-16
LuCI showing all traffic rules as IPv4, including rules that are IPv6


knacky:
From a fresh clone of OpenWRT trunk, Traffic Rules in LuCI are all showing IPv4, even the rules that are in place for IPv6.

Steps to reproduce:
get a fresh clone of OpenWRT trunk
update and install packages feed
make clean (set target system ath79 and target profile wndr3800)
make defconfig
make menuconfig (add LuCI > Collections > luci)
make
flash factory image via tftp recovery for fresh install
get on luci web interface and go to Network > Firewall > Traffic Rules

all rules display as IPv4, even ones that are IPv6.

/etc/config/firewall still has "option family ipv6" for the v6 rules, so this appears to me to be a luci display issue and not a rule permissions being changed issue.

![Untitled](https://user-images.githubusercontent.com/8460814/61335043-09d30e00-a7fa-11e9-9b64-08de21458150.png)

lleachii:
I verify this also occurs on LuCI Master (git-19.194.40998-7b1aabc) / OpenWrt SNAPSHOT r10506-cbae306815.

PR #2880, opened 2019-07-17
luci-app-firewall: format protocol family for forwards and rules

namedun:
This should fix the issue #2878.

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js
@@ -211,6 +211,15 @@ return L.Class.extend({
 		return rv;
 	},
 
+	fmt_family: function(family) {
+		if (family == 'ipv4')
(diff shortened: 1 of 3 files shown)


applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js
```diff
@@ -211,6 +211,15 @@ return L.Class.extend({
 		return rv;
 	},
 
+	fmt_family: function(family) {
+		if (family == 'ipv4')
+			return _('IPv4');
+		else if (family == 'ipv6')
+			return _('IPv6');
+		else
+			return _('IPv4/6');
```
jow-: Can you change this to `IPv4 and IPv6` as this was the old translation label in the Lua page.

namedun: Changed

jow-:
Merged, thank you!

PR #2882, opened 2019-07-18
luci-app-openvpn: add new tls_ciphersuites option

sch-m:
This is used to configure ciphers for TLS 1.3 or newer.

Related to openwrt/openwrt#2185


jow-:
Merged, thanks!

PR #2883, opened 2019-07-18
address: Replace ualarm with setitimer

neheb:
ualarm has been deprecated and may not be available in the libc.

### DIFF
--- a/libs/luci-lib-nixio/src/address.c
+++ b/libs/luci-lib-nixio/src/address.c
@@ -24,6 +24,7 @@
 
 #ifdef __linux__
 
+#include <sys/time.h>
 #include <signal.h>
 #include <setjmp.h>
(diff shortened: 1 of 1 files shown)


libs/luci-lib-nixio/src/address.c
```diff
@@ -308,7 +309,8 @@ static int nixio_getnameinfo(lua_State *L) {
 			return 3;
 		}
 
-		ualarm(timeout * 1000, 0);
+		const struct itimerval t = { {timeout * 1000 * 1000, 0} , {0, 0} };
```
jow-: Please move the variable declaration to the top of the function.

neheb:
Done

jow-:
Merged, thanks.

PR #2884, opened 2019-07-19
luci-app-firewall: bring back 'MSS clamping' column/option

dibdot:
* 'MSS clamping' is missing in current Zones setup


jow-:
May I ask why? Is it really an urgently needed feature that needs to be edited by the majority of users? I am fine with the PR but I wonder if the MSS setting is important enough to be shown in the overview table (instead of being hidden in the modal).

dibdot:
sorry then, during testing I compared 19.07 vs. master and thought it was a typo. I'll close this one ...

jow-:
No need to be sorry, I honestly wonder if there is a use case for it.

PR #2892, opened 2019-07-20
[19.07] luci-app-https_dns_proxy: better handling of ports, update dhcp config

stangri:

### DIFF
--- a/applications/luci-app-https_dns_proxy/Makefile
+++ b/applications/luci-app-https_dns_proxy/Makefile
@@ -10,7 +10,7 @@ LUCI_TITLE:=HTTPS DNS Proxy Web UI
 LUCI_DESCRIPTION:=Provides Web UI for HTTPS DNS Proxy
 LUCI_DEPENDS:=+luci-mod-admin-full +https_dns_proxy
 LUCI_PKGARCH:=all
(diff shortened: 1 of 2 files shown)


feckert:
I think we should add https://github.com/openwrt/luci/pull/2890 into master first and then cherry-pick this with the option -x for reverence into to this stable branch -> openwrt-19.07

stangri:
@feckert I was previously advised to submit 3 different PRs (master, 18.06, 19.07) for the bugfix for a different package. I would appreciate some guidance in this regard for the future PRs.

feckert:
I think there is no guidance for this. There are advantage and disadvantage.

hnyman:
I feel that having a separate PR for each branch from the package maintainer makes it easier for commit access holders.

The PR might contain a simple cherry-pick commit, but in some cases there may be branch-specific modifications, so that a straight cherry-pick does not work.

feckert:
> I feel that having a separate PR for each branch from the package maintainer makes it easier for commit access holders.

Agree

> The PR might contain a simple cherry-pick commit,

Exactly so we could track the change if this is possible master -> stable

jow-:
Cherry-picked from 85989079f instead. Thanks!

PR #2895, opened 2019-07-20
luci-base: Update Ukrainian translation

YuriPet:
Update Ukrainian translation according the latest translation resync.

Signed-off-by: Yurii <yuripet@gmail.com>


jow-:
Thank you!

PR #2909, opened 2019-07-23
luci-app-lxc: Add Spanish translation

castillofrancodamian:
Add Spanish translation.


feckert:
Thanks

PR #2911, opened 2019-07-23
luci-app-ddns: Add English translation file

castillofrancodamian:
Add English translation file.


feckert:
I do not think we should translate english it is default and if no translation is found then the msgid is used. So I think from my side.
NOK

castillofrancodamian:
And why are there packages with "translation" into English?

PR #2934, opened 2019-07-29
luci-mod-network: 802.11w config in station mode

antoinedeschenes:
@hauke LIttle bugfix as I've been using two routers in WPA3-SAE mode for a few months.

802.11w has to be enabled in station mode to be effective. If the AP is using WPA3-SAE or requires 11w, station will not connect unless 11w is enabled via this option.


hauke:
This looks good to me.
WPA3 mandates ieee80211w when SAE is used, which is the case when you really use WPA3

feckert:
Could you please fix the whitespaces?
LuCI does have tabs.

antoinedeschenes:
@feckert Of course, fixed.

PR #2969, opened 2019-08-06
luci-base: add new VPN toplevel menu entry

feckert:
There is always more and more vpn services. To make the LuCI menu look cleaner, a new top level menu "VPN" will be added with this pullrequest. All luci-app-* that have something to do with VPN should move to this new menu entry.

* Add new vpn menu section
* Move luci-app-openvpn to the new vpn section
* Move luci-app-ocserv to the new vpn section

The luci-app-vpnbypass should also go to this section.

### DIFF
--- a/applications/luci-app-ocserv/luasrc/controller/ocserv.lua
+++ b/applications/luci-app-ocserv/luasrc/controller/ocserv.lua
@@ -10,24 +10,24 @@ function index()
 
 	local page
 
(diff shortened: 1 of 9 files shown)


feckert:
@jow Any objections from your part?
The background is I want to clean up the LuCI a little bit. I have had a look at opnsense and the WebUI looks tidier and straight forward. That is only a first step on cleanup.

jow-:
Fine with me.

PR #2975, opened 2019-08-11
luci-theme-bootstrap: Corrected overflowing inner div.

kugland:


jow-:
On which browser / theme / version combination did this occur? Iirc we fixed this very bug already

feckert:
Please update your commit message.
luci-theme-bootstrap: <message>

PR #2976, opened 2019-08-11
luci-app-firewall: rules: allow ICMPv6 ND types and sort alphabetically

aaronmdjones:
The "Match ICMP Type" dropdown had entries for router solicitation & router advertisements, but not the more generic neighbour solicitation & neighbour advertisements.

A LAN cannot function without Neighbour Discovery; this means that setting a LAN interface default input policy to REJECT breaks IPv6 WAN access for all hosts on that LAN; as they can no longer discover their gateway's MAC address.

This can be fixed with appropriate rules allowing ND input, which this patch allows one to do in LuCI.


jow-:
Merged, thanks.

Issue #2728, opened 2019-05-26
luci-mod-admin-full: Led not configurable through luci after upgrading to 18


stephengeorgewest:
Related to #971 
I just  updated to OpenWrt 18.06.2 r7676-cddd7b4c77 / LuCI openwrt-18.06 branch (git-19.020.41695-6f6641d) and now I only get one port as an option.
![image](https://user-images.githubusercontent.com/1442484/58383475-8f71d480-7f94-11e9-8583-49a5554a7267.png)

https://github.com/openwrt/luci/blob/master/modules/luci-mod-system/luasrc/model/cbi/admin_system/leds.lua#L151 says to use "/sys/bus/usb/devices/usb*" if present, but that is the usb port of the built in hub.
![image](https://user-images.githubusercontent.com/1442484/58383534-538b3f00-7f95-11e9-84d9-7593d3451eb4.png)

In my case I think it should be listing 1-1.1 as port 1 and 1-1.2 as port 2. I'm not sure how to fix it without breaking everyone else.

I guess not many other systems have a useless hub permanently attached to the root hub. Maybe I didn't have this extra hub before (I don't remember doing a lsusb), and openwrt/openwrt issue, or kernal updates...
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 001 Device 002: ID 05e3:0608 Genesys Logic, Inc. Hub
Bus 001 Device 003: ID 051d:0002 American Power Conversion Uninterruptible Power
Bus 001 Device 004: ID 051d:0002 American Power Conversion Uninterruptible Power

stephengeorgewest:
I read that bit wrong. That issue was just to read the values from the configuration. I think the issue is in one or both of the "for p in nixio.fs.glob". still trying stuff.

stephengeorgewest:
Model	TP-Link TL-WDR3600 v1
Architecture	Atheros AR9344 rev 2
Firmware Version	OpenWrt 18.06.2 r7676-cddd7b4c77 / LuCI openwrt-18.06 branch (git-19.020.41695-6f6641d)
Kernel Version	4.9.152

Probably should have put that in at the top.


If i put <code>usbport:value("usb1-port2", "Hub 1, Port 2*")<code> after the <code>for p in nixio.fs.glob("/sys/bus/usb/devices/*/usb[0-9]*-port[0-9]*") do  </code> .. <code>end</code> block I get:
[...]

PR #2979, opened 2019-08-12
luci-mod-admin-full: Update leds.lua to show all the possible ports

stephengeorgewest:
This will show all the values that can be applied as it matches /sys/class/leds/*/ports/
Signed-off-by: Stephen <stephen.george.west@gmail.com>

Possible fix for #2728

### DIFF
--- a/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
+++ b/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
@@ -139,7 +139,7 @@ for p in nixio.fs.glob("/sys/bus/usb/devices/[0-9]*/manufacturer") do
 end
 
 for p in nixio.fs.glob("/sys/bus/usb/devices/*/*-port[0-9]*") do
-	local port = p:match("/sys/bus/usb/devices/.*/(.*-port%d+)$")
+	local port = p:match("([^/]+-port%d+)$")
 	if port then
 		local attached = nixio.fs.stat("%s/device" % p) and " (%s)" % translate("device attached") or ""
 		usbport:value(port, port .. attached)
--- a/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
+++ b/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
@@ -121,7 +121,7 @@ usbport.size = 1
 function usbport.valuelist(self, section)
 	local port, ports = nil, {}
 	for port in util.imatch(m.uci:get("system", section, "port")) do
-			ports[#ports+1] = port
+		ports[#ports+1] = port
 	end
 	return ports
 end
--- a/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
+++ b/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
@@ -121,13 +121,7 @@ usbport.size = 1
 function usbport.valuelist(self, section)
 	local port, ports = nil, {}
 	for port in util.imatch(m.uci:get("system", section, "port")) do
-		local b, n = port:match("^usb(%d+)-port(%d+)$")
-		if not (b and n) then
-			b, n = port:match("^(%d+)-(%d+)$")
-		end
-		if b and n then
-			ports[#ports+1] = "usb%u-port%u" %{ tonumber(b), tonumber(n) }
-		end
+			ports[#ports+1] = port
 	end
 	return ports
 end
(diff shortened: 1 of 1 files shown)


feckert:
DCO was not successful.
Please add a new line in your commit before sign off.
And then make a push force.
I think this will pass the check.

feckert:
@jow I want to give a hint that on openwrt 18.06 as I know have kernel version 4.14 and 4.9
So If we add this commit to 18.06 I hope this will not break the other targets who are on other kernel versions and maybe have other led triggers.

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
```diff
@@ -121,13 +121,7 @@ usbport.size = 1
...
 	local port, ports = nil, {}
 	for port in util.imatch(m.uci:get("system", section, "port")) do
-		local b, n = port:match("^usb(%d+)-port(%d+)$")
-		if not (b and n) then
-			b, n = port:match("^(%d+)-(%d+)$")
-		end
-		if b and n then
-			ports[#ports+1] = "usb%u-port%u" %{ tonumber(b), tonumber(n) }
-		end
+			ports[#ports+1] = port
```
jow-: Please fix the indentation level here (remove one tab)

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
```diff
@@ -144,11 +138,11 @@ for p in nixio.fs.glob("/sys/bus/usb/devices/[0-9]*/manufacturer") do
 	usbdev:value(id, "%s (%s - %s)" %{ id, mf, pr })
 end
 
-for p in nixio.fs.glob("/sys/bus/usb/devices/*/usb[0-9]*-port[0-9]*") do
-	local bus, port = p:match("usb(%d+)-port(%d+)")
-	if bus and port then
-		usbport:value("usb%u-port%u" %{ tonumber(bus), tonumber(port) },
-		              "Hub %u, Port %u" %{ tonumber(bus), tonumber(port) })
+for p in nixio.fs.glob("/sys/bus/usb/devices/*/*-port[0-9]*") do
+	local port = p:match("/sys/bus/usb/devices/.*/(.*-port%d+)$")
```
jow-: Please use `p:match("([^/]+-port%d+)$")` as pattern.

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
```diff
@@ -144,11 +138,11 @@ for p in nixio.fs.glob("/sys/bus/usb/devices/[0-9]*/manufacturer") do
...
+	if port then
+		usbport:value("%s" %{ port },
```
jow-: This interpolation is redundant, simply pass `port` as-is (`usbport:value(port, ...)`)

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
```diff
@@ -144,11 +138,11 @@ for p in nixio.fs.glob("/sys/bus/usb/devices/[0-9]*/manufacturer") do
...
+		              "%s %s" %{ port, nixio.fs.stat("%s/device" %{p}) ~= nil and "(device attached)" or ""})
```
jow-: Please add helper variable a few lines above assigning the device presence state and make the string translatable:

```
local attached = nixio.fs.stat("%s/device" % p) and ' (%s)' % translate('device attached') or ''
usbport:value(port, port .. attached)
```

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
stephengeorgewest: fixed

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
stephengeorgewest: done

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
stephengeorgewest: done

modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/leds.lua
stephengeorgewest: done

Neustradamus:
@feckert: It was integrated in master?

stephengeorgewest:
I don't think it was, but it is probably a moot point since openwrt 18 is end of life and the logic has all moved to JS: applications\luci-app-ledtrig-usbport\htdocs\luci-static\resources\view\system\led-trigger\usbport.js

I upgraded the device to 21 a while ago, and checking now, I see that it lists 2 usb leds.
image
but there are no usb triggers (where before it listed the root usb port that was always connected).
image

So, yeah, this is probably a useless PR now.

PR #2981, opened 2019-08-14
[RFC] luci-nginx and luci-ssl-nginx: collect files from nginx and uwsgi-cgi packages

peter-stadler:
I am new to contributing and I am not sure if this change is a good idea. I tried to collect the files that are needed to run luci on nginx from the uwsgi-cgi and the nginx package and would move it here. See my other draft PRs in packages, too:
<strike>https://github.com/openwrt/packages/pull/9736 ([RFC] package etesync and its dependencies and move luci on nginx)</strike>
[uwsgi: add packages with modules](https://github.com/openwrt/packages/pull/9855) and [nginx: enable /etc/nginx/conf.d directory](https://github.com/openwrt/packages/pull/9859).
I would do this only if Ansuel approves it since it is his project.

See also the topic at:
https://forum.openwrt.org/t/package-etesync-and-its-dependencies/42269
I hope it is understandable what I try to do.

### DIFF
--- a/collections/luci-nginx/root/etc/uci-defaults/60-luci
+++ b/collections/luci-nginx/root/etc/uci-defaults/60-luci
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+if [ -n "$(pgrep uhttpd)" ]; then
+	/etc/init.d/uhttpd stop
+	/etc/init.d/uhttpd disable
+fi
+if [ -n "$(pgrep nginx)" ]; then
+	/etc/init.d/nginx restart
+else
+	/etc/init.d/nginx start
+fi
+if [ -n "$(pgrep luci)" ]; then
+	/etc/init.d/luci restart
+else
+	/etc/init.d/luci start
+fi
--- a/collections/luci-ssl-nginx/root/etc/uci-defaults/70-luci
+++ b/collections/luci-ssl-nginx/root/etc/uci-defaults/70-luci
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+if [ -n "$(pgrep uhttpd)" ]; then
+	/etc/init.d/uhttpd stop
+	/etc/init.d/uhttpd disable
+fi
+if [ -n "$(pgrep nginx)" ]; then
+	/etc/init.d/nginx restart
+else
+	/etc/init.d/nginx start
+fi
+if [ -n "$(pgrep luci)" ]; then
+	/etc/init.d/luci restart
+else
+	/etc/init.d/luci start
+fi
+if [ ! -f "/etc/nginx/nginx.key" ]; then
+	
+	NGINX_KEY=/etc/nginx/nginx.key
+	NGINX_CER=/etc/nginx/nginx.cer
+	OPENSSL_BIN=/usr/bin/openssl
+	PX5G_BIN=/usr/sbin/px5g
+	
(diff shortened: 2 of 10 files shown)


peter-stadler:
Sorry for the confusion by opening and closing this PR repeatedly. I wanted to change it to a draft PR.

collections/luci-nginx/root/etc/uci-defaults/60-luci
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+chmod +x /etc/init.d/luci
```
feckert: Why do we need this? We could install this file with the right mods during installation in the Makefile

collections/luci-ssl-nginx/root/etc/uci-defaults/70-luci
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+chmod +x /etc/init.d/luci
```
feckert: Same here

feckert:
Please add a prefix in front of your commits.  Have a look at the commit history
luci-<package>: <what have you done>

collections/luci-ssl-nginx/root/etc/uci-defaults/70-luci
peter-stadler: I am not sure, how to do it. We need a $(INSTALL_BIN) … somewhere, right? But where, if we want to use the Package/$(PKG_NAME)/install from luci.mk?

peter-stadler:
Thank you :-) I will change the prefixes. Right now I am changing the PR for the packages repository …

collections/luci-ssl-nginx/root/etc/uci-defaults/70-luci
feckert: Thats right. In LuCI you have to put the files into **root** with the right subdirectories and with the right mod flags and it will gets installed. See for example https://github.com/openwrt/luci/tree/master/applications/luci-app-adblock/root/etc/uci-defaults

peter-stadler: Oh, now I have a bunch of commits with the wrong username/email. Will continue tomorrow and change the mods of the files, too.

peter-stadler:
The uwsgi-cgi-plugin would be in the uwsgi Makefile of my other commits in the packages repository. Right now I am redoing them with better commit messages, it will be in https://github.com/peter-stadler/packages/tree/uwsgi – right now I want to change username/email before the PR …

peter-stadler:
Sorry for the late answer (it took me some time to fix the CircleCI build for the [etesync-server](https://github.com/openwrt/packages/pull/9865) work). 

I made the init files executable, changed the commit messages and included luci-app-opkg as dependency (it is in the standard luci collection, too). Thank you @feckert.

Both commits depend on the uwsgi-cgi-plugin of the [modular uwsgi](https://github.com/openwrt/packages/pull/9855) and that the [/etc/nginx/conf.d directory is enabled](https://github.com/openwrt/packages/pull/9859).

peter-stadler:
@jow- do you think this is a good idea (regarding also the openwrt guidelines)?

jow-:
Closing this as its outside of the scope of LuCI. The necessary work appears to have happened in the package repo now.

PR #2994, opened 2019-08-19
luci-base: add new Firewall toplevel menu entry

feckert:
To make the LuCI menu look cleaner, a new top level menu "Firewall" is added with this pullrequest. In a subsequent pullrequest we can also move the following application to the new firewall menu. These applications have everything to do with firewall from my point of view.

* luci-app-adblock
* luci-app-banip
* luci-app-clamav
* luci-app-nft-qos
* luci-app-privoxy
* luci-app-simple-adblock
* luci-app-tinyproxy

With this pullrequest I want to show the possibility that OpenWrt can also be used as a firewall solution like **opensense**. I realize that packages are still missing, but this is the first step

### DIFF
--- a/applications/luci-app-firewall/root/usr/share/luci/menu.d/luci-app-firewall.json
+++ b/applications/luci-app-firewall/root/usr/share/luci/menu.d/luci-app-firewall.json
@@ -1,5 +1,5 @@
 {
-	"admin/network/firewall": {
+	"admin/firewall": {
 		"title": "Firewall",
 		"order": 60,
 		"action": {
@@ -12,7 +12,7 @@
 		}
 	},
 
-	"admin/network/firewall/zones": {
+	"admin/firewall/zones": {
 		"title": "General Settings",
 		"order": 10,
 		"action": {
@@ -21,7 +21,7 @@
 		}
 	},
 
-	"admin/network/firewall/forwards": {
+	"admin/firewall/forwards": {
 		"title": "Port Forwards",
 		"order": 20,
 		"action": {
@@ -30,7 +30,7 @@
 		}
 	},
 
-	"admin/network/firewall/rules": {
+	"admin/firewall/rules": {
 		"title": "Traffic Rules",
 		"order": 30,
 		"action": {
@@ -39,7 +39,7 @@
 		}
 	},
 
-	"admin/network/firewall/custom": {
+	"admin/firewall/custom": {
 		"title": "Custom Rules",
 		"order": 40,
 		"action": {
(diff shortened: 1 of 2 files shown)


feckert:
@jow Any objections from your part?
Same reason as mention in https://github.com/openwrt/luci/pull/2969

hnyman:
I am hesitant to say either ACK or NAK.
Not all of the listed app are "firewall" related apps, although some are. (and you forgot BCP38 from the list).

nft-qos, firewall related? 

I do not like the idea of cluttering the top-level menu with new items. Already the VPN is questionable, as you could argue that similarly as VPR services got their own menu, also QoS tools could get, and privary related etc... 

The old Services menu is maybe a bit vague, but many of the apps would belong there.

feckert:
@hnyman thanks for your comment.

I have played around with opnsense some time ago, and I think there menu structure is very clear. it looks like it's made of one piece. They have the additional menu entries **VPN** and **Firewall** with the related applications for this cases.

Up to now the **Firewall** reladed apps and **VPN** related apps are scattered in LuCI. I think it would be more clear to add this to the top-level menu, so we could group the related applications together.

It would also make it more clear that OpenWrt could also acct as a VPN-Concentrator or Firewall-Solution. I think openWrt has everything on board to serve these scenarios as well. But that's not at the first look. But LuCI is the frontend where you have first contact with OpnWrt.

feckert:
@hnyman @jow- are there any additional comments/hints from your side if not I will remove the RFC and would change to request a merge.

aparcar:
@feckert I think this is no longer valid as everything moved on the JavaScript.

feckert:
@aparcar thans for the hint. I will rebase my changes

jow-:
I don't like that, especially since most services you propose to move there aren't actually firewall related. A more appropriate term would be "Network filtering" which then clashes with "Network" and will lead to even more confusion. Besides that I think privoxy, tinyproxy, clamav are actually services, not firewalls.

I do agree that we might need to reorganize the menu eventually but that Firewall menu proposed here seems arbitrary and - without an overhaul of the entire menu structure - will lead to more, not less clutter.

feckert:
@jow @hnyman I would like to take the ticket again, and point out that other systems also outsource the firewall in the top level directory. See pictures in Ticket https://github.com/openwrt/luci/issues/4747
Because the firewall menu is not so hidden, and we would show it off more that openwrt has also a good firewall config.
The other applications can remain in the service. Maybe the menu will evolve to include other firewall tasks in the future, that are now placed in the service menu entry

PR #3040, opened 2019-09-05
luci-proto-modemmanager: add luci-proto-modemmanager

nickberry17:
Signed-off-by: Nicholas Smith <nicholas.smith@telcoantennas.com.au>

### DIFF
--- a/protocols/luci-proto-modemmanager/Makefile
+++ b/protocols/luci-proto-modemmanager/Makefile
@@ -1,0 +1,46 @@
+#
+# Copyright 2019 Telco Antennas Pty Ltd <nicholas.smith@telcoantennas.com.au>
+# SPDX-License-Identifier: Apache-2.0
+#
+
+include $(TOPDIR)/rules.mk
+
+# LUCI_TITLE:=Support for ModemManager
+# LUCI_DEPENDS:=+modemmanager
+#
+# include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
+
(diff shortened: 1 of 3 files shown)


nickberry17:
For use when https://github.com/openwrt/packages/pull/9137 is merged

nickberry17:
I think this should actually go into the 18 branch, not master

feckert:
> I think this should actually go into the 18 branch, not master

I don´t think so 18.06 is feature complete.
We could add this to master and then backport this to openwrt-19.07 branch.

bobafetthotmail:
the PR this depends on was merged https://github.com/openwrt/packages/pull/9137
so I guess this can go in too now.

protocols/luci-proto-modemmanager/Makefile
```diff
@@ -0,0 +1,46 @@
...
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
+
+$(eval $(call BuildPackage,luci-proto-modemmanager))
```
feckert: Please update this Makefile have a look at https://github.com/openwrt/luci/blob/master/protocols/luci-proto-qmi/Makefile as pattern

feckert:
Sorry but we need to port this to javascript to get merged lua is not working anymore for proto!
We use now in master client side scripting.
All protocols in https://github.com/openwrt/luci/tree/master/protocols have been ported please use this as a pattern

PR #3061, opened 2019-09-13
[19.07] luci-base: update Spanish translation

castillofrancodamian:


feckert:
Could we fixup the commits in one?
And Please add a new line after the header in the future.

PR #3063, opened 2019-09-13
luci-app-statistics: Add new plugin options

### LINKED ISSUE #3013: luci-app-statistics: more selectable options for collectd like CPU and RAM in percent
As of release 5.5, collectd offers some new options like CPU usage in % instead of jiffies or RAM usage in % instead of bytes. I suggest to add these options for collectd configuration in LuCI. This would greatly help in preparing statistics with an external server like InfluxDB with Grafana as no longer extra math would be required to get meaningful graphs.
The relevant parts of `collectd.conf` look like this (already tested on my own):

```
LoadPlugin memory
<Plugin memory>
	ValuesPercentage true
</Plugin>

LoadPlugin cpu
<Plugin cpu>
	ReportByCpu true
	ReportByState true
	ValuesPercentage true
</Plugin>
```

### LINKED ISSUE #3053: luci-app-statistics: Ping plugin does not display stddev data
19.07

I have noticed that there are **ping_stddev-*.rrd** files generated by the **ping** plugin, but the data is not displayed on the chart like the other two datasets. Can **stddev** be added?

feckert:
- Add new cpu plugin options
- Add new memory plugin option
- Add ping stddev definition
- Minify memory license header

closes #3013 
closes #3053

### DIFF
--- a/applications/luci-app-statistics/luasrc/model/cbi/luci_statistics/cpu.lua
+++ b/applications/luci-app-statistics/luasrc/model/cbi/luci_statistics/cpu.lua
@@ -12,4 +12,25 @@ s = m:section( NamedSection, "collectd_cpu", "luci_statistics" )
 enable = s:option( Flag, "enable", translate("Enable this plugin") )
 enable.default = 0
 
+-- collectd_cpu.reportbycpu (ReportByCpu)
+reportbycpu = s:option( Flag, "ReportByCpu",
+	translate("Report by CPU"),
+	translate("By setting this, CPU is not aggregate of all processors on the system"))
+reportbycpu.default = 1
+reportbycpu:depends( "enable", 1 )
+
+-- collectd_cpu.reportbystate (ReportByState)
+reportbystate = s:option( Flag, "ReportByState",
+	translate("Report by state"),
+	translate("When set to true, reports per-state metric (system, user, idle)"))
+reportbystate.default = 1
+reportbystate:depends( "enable", 1 )
+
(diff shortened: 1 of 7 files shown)


feckert:
@hnyman any hints from your side?
I would go on and remove on the next pr the state-genconfig and the other related stuff, so we do not hijack the collectd anymore.
The result would be that the application luci-app-statistics will write a collectd config and then restart them.

hnyman:
Sorry for inactivity. I have been traveling, so no chance to test your proposals.

feckert:
@hnyman no problem
I'm not always how long I should wait. Is there policy in openwrt?

hnyman:
Nothing very clear.
A few days should be usually enough.

hnyman:
@feckert 
This can apparently lead into a config where neither option is enabled, although LuCI shows that the default "Value Absolute" is enabled:
image

As LuCI removes default values (as they are thought to be unnecessary as real defaults in the underlying app), editing configurstion in LuCI may lead into a config file, where neither option is enabled.

This probably needs an "property.optional = false" to disable the removal of the option by LuCI.
 
Additionally, there seems to be typo in the variable names, as "valuespercentage" is used for both ValuesPercentage and ValuesAbsolute.
https://github.com/openwrt/luci/blob/master/applications/luci-app-statistics/luasrc/model/cbi/luci_statistics/memory.lua#L14

hnyman:
I added 95ea1e67fe as a fix attempt.

feckert:
@hnyman thanks for your fix :-)

PR #3064, opened 2019-09-13
luci-app-commands: fix escaping of additional shell chars

### LINKED ISSUE #1622: luci-app-commands: does not support pipes
The pipe symbol `|` is somehow turned into `"|"`, so the piped commands fail.
exmample:
instead of `opkg list-upgradable | cut -f 1 -d " "`
gets executed and fails `opkg list-upgradable "|" cut -f 1 -d " "`

feckert:
Do not escape pipe on cmd line

closes #1622

### DIFF
--- a/applications/luci-app-commands/luasrc/controller/commands.lua
+++ b/applications/luci-app-commands/luasrc/controller/commands.lua
@@ -144,7 +144,7 @@ local function parse_cmdline(cmdid, args)
 		end
 
 		for i, v in ipairs(argv) do
-			if v:match("[^%w%.%-i/]") then
+			if v:match("[^%w%.%-i/|]") then
 				argv[i] = '"%s"' % v:gsub('"', '\\"')
 			end
 		end


jow-:
NACK, lets fix the underlying issue instead.

feckert:
@jow regex on parse_cmdline function not escaping the pipe.

jow-:
Are you sure the proposed patch is correct? Shouldn't `|` be put into the `[...]` ?

feckert:
I will try and disassemble the regex.
But with this test command **ls -la /dev | grep tty**  it works for me.

feckert:
@jow- As expected, you're right. My carelessness
I have update the pull-request

Before:

Screenshot-2019-9-18 G3000-LWV - Dashboard - LuCI

After:

Screenshot-2019-9-18 G3000-LWV - Dashboard - LuCI

jow-:
Should we add `<` and `>` as well?

feckert:
I am not sure if this make sense. Because the output of a command is already redirected. See source. I think the pipe is the missing part to concatenate the commands. But that`s only my point of view. 

https://github.com/openwrt/luci/blob/3c40a7450a9b7d340cdc9e233818c63a8092fec2/applications/luci-app-commands/luasrc/controller/commands.lua#L155-L170

feckert:
@jow- ping any objections from your side? If not I will merge this for now if we need also `<` and `>` I will add this later.

jow-:
No, please go ahead

Issue #2065, opened 2018-08-14
luci-base: OUI lookup in the DHCP leases table?


rlipscombe:
I've got a lot of devices that appear as '?' in the DHCP leases table. It'd be really useful if LuCI could look up the OUI and tell me the manufacturer. Ideally this'd be integrated in the table directly as an extra column, but I could live with an extra click.

jow-:
OUI prefix data is *huge* so the lookup would require online access. Neither is ideal.

49616e:
At the time of writing this comment, the OUI table available at [here](http://standards-oui.ieee.org/oui/oui.txt) about 4072KB in size. We can trim it down to less than half the size since we don't really need those addresses anyway. There are plenty of devices right now that comes with 32MB or bigger flash, I think <2MB is not that big of a deal if made available as an optional package. Please reconsider adding support for this as it'll be very nice to have in certain situations.

CTrabant:
As of this comment, the combined OUI lookup table from wireshark:
https://code.wireshark.org/review/gitweb?p=wireshark.git;a=blob_plain;f=manuf

is *1,632 KB*, and is an assemblage of the following sources:
```
# The IEEE public OUI listings available from:
# <http://standards-oui.ieee.org/oui/oui.csv>
# <http://standards-oui.ieee.org/cid/cid.csv>
# <http://standards-oui.ieee.org/iab/iab.csv>
# <http://standards-oui.ieee.org/oui28/mam.csv>
# <http://standards-oui.ieee.org/oui36/oui36.csv>
```

That size shouldn't be much of a problem, especially if a lookup were an opt-in feature to turn on.

ervee:
If I would download the OUI file on my AP or Router (both are pretty modern and have hundreds of MB storage available), how would I make use of this? I would love to see the MAC lookup used in the "Wireless Associated Stations" and "Active DHCP Leases"  view?

joshenders:
> That size shouldn't be much of a problem, especially if a lookup were an opt-in feature to turn on.

I think we could probably do even better with a binary format, as the OUI bytes could be stored as binary data and the strings compressed. Opt-in is icing on the cake.

n-thumann:
In case this get tackled, I'd appreciate it getting added to the list of associated stations (connected wireless devices) too :v:

systemcrash:
[...]

PR #3081, opened 2019-09-18
luci-base: show manufacturer in leases overview page

feckert:
This pull-request will add the feature, which was requested in the issue #2065.

**This works for now only for ipv4.**

I have added a new package for luci (ouidb) to lookup the company name with the oui part of the mac. I have also added the possibily, that the ouidb is not required for small devices. So if ouidb file is not installed then it will not shown in then status page.

To minify the oui database, I have added a python convert script to remove the unneeded information.
Only the Company name and the oui are written to the ouidb.

Size difference

- With all information 6.3MB
- With oui and manufacture information  0.9MB

### DIFF
--- a/contrib/package/ouidb/Makefile
+++ b/contrib/package/ouidb/Makefile
@@ -1,0 +1,51 @@
+#
+# Copyright (C) 2019 TDT AG <development@tdt.de> 
+# This is free software, licensed under the GNU General Public License v2.
+# See https://www.gnu.org/licenses/gpl-2.0.txt for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=ouidb
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL=https://github.com/jfisbein/ouidb-json.git
+PKG_SOURCE_DATE:=2019-09-18
+PKG_SOURCE_VERSION:=c45622bbcc29644a03202ce6ac0a618e8af417c3
+PKG_SOURCE_SUBDIR:=$(PKG_NAME)-$(PKG_SOURCE_VERSION)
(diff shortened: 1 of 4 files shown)


jow-:
Please hold this request for now, the corresponding code will be soon converted to JavaScript.

49616e:
> Please hold this request for now, the corresponding code will be soon converted to JavaScript.

I'm not sure if this is 100% on-topic but OUI info would be a good thing to have on the the wireless page as well.

lucize:
or in Routes section ?

kayrus:
+1 for this feature. @feckert could you pleas rebase?

tapper82:
+1 for this feature.

49616e:
> Please hold this request for now, the corresponding code will be soon converted to JavaScript

Since `modules/luci-base/luasrc/view/lease_status.htm` and `modules/luci-base/luasrc/tools/status.lua` are both gone, I assume the relevant lua code has already been converted to js now? If that's the case, please consider reworking on this PR now, thanks. @feckert

feckert:
@jow- I just wanted to update the pullrequest.
Not sure where to put this right now.
I could add the need information to luci-rpc in this function.
https://github.com/openwrt/luci/blob/0412e85a57c2f271c24fe837e3a6adcf60bc63f1/libs/rpcd-mod-luci/src/luci.c#L1739
We could check if the oui db file is present and if so, add the needed information to the ubus object.
This could then be used in the status pagehttps://github.com/openwrt/luci/blob/master/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js

jow-:
That would be one approach, yes. Another one would be loading the OUI database on the client side and use it there.

odkrys:
+1 It's closed, but I hope it gets added as a package.

PR #3117, opened 2019-09-25

YuriPet:
Updating Ukrainian translation according the latest sync translation. Сorrections and additions.

Signed-off-by: Yurii <yuripet@gmail.com>

### DIFF
--- a/modules/luci-base/po/uk/base.po
+++ b/modules/luci-base/po/uk/base.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: \n"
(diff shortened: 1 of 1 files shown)


feckert:
@YuriPet Could you please add in the future your real full name in den Signed-off-by line?

YuriPet:
I would do it with pleasure, but I can not find a way to do it.

jow-:
Do you use Git from the command line? If yes, try this within your clone of the LuCI repository:
`git config user.name "Your Real Name"`

YuriPet:
No, unfortunately, I work with github.com only through the browser.

PR #3125, opened 2019-09-28
[RFC] luci-app-cifsd: add new package

ysc3839:
As openwrt/openwrt#2368 said, I'm going to add luci-app-cifsd.
But I noticed that luci is moving to client-side view implemention. Should I implement it in client-side?

### DIFF
--- a/applications/luci-app-cifsd/htdocs/luci-static/resources/view/cifsd.js
+++ b/applications/luci-app-cifsd/htdocs/luci-static/resources/view/cifsd.js
@@ -1,0 +1,96 @@
+'use strict';
+'require fs';
+'require form';
+'require tools.widgets as widgets';
+
+return L.view.extend({
+	load: function() {
+		return Promise.all([
+			L.resolveDefault(fs.stat('/sbin/block'), null),
+			L.resolveDefault(fs.stat('/etc/config/fstab'), null),
+		]);
+	},
+	render: function(stats) {
+		var m, s, o;
+
+		m = new form.Map('cifsd', _('Network Shares'));
+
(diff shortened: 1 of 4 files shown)


ysc3839:
Ping @jow-

ysc3839:
Any updates?

jow-:
@ysc3839 - sorry, this slipped by me. The submission is fine to me. The only change request I have is converting the `callFile*()` procedures to `fs.js`.

applications/luci-app-cifsd/htdocs/luci-static/resources/view/cifsd.js
```diff
@@ -0,0 +1,113 @@
...
+
+
+		s = m.section(form.TableSection, 'share', _('Shared Directories'),
+			_('Please add directories to share. Each directory refers to a folder on a mounted device.'));
+		s.anonymous = true;
+		s.addremove = true;
+
+		s.option(form.Value, 'name', _('Name'));
+		o = s.option(form.Value, 'path', _('Path'));
+		if (callFileStat('/sbin/block') && callFileStat('/etc/config/fstab')) {
```
jow-: Note that these functions return promises, so you cannot use them synchronously. Instead, declare a `load:` function, and have it `return Promise.all([L.resolveDefault(fs.stat('/sbin/block')), L.resolveDefault(fs.stat('/etc/config/fstab'))])`.

Then change the render function to `render: function(stat_results) { ... }` and use `if (stat_results[0] && stat_results[1])` here.

ysc3839:
Because `!!{}` is not false, I changed it to `null`.

PR #3141, opened 2019-10-02
luci-app-statistics: switch to collectd uci handling

feckert:
* add lua script handling
* move to collectd uci backend

MISSING:
Upgrade script for old luci_statistics -> luci_collectd and collectd uci config.


applications/luci-app-statistics/luasrc/statistics/datatree.lua
```diff
@@ -17,13 +16,9 @@ function Instance.__init__( self, host )
...
-	if sections and sections.collectd then
-		self._host    = host or sections.collectd.Hostname or sys.hostname()
-		self._libdir  = sections.collectd.PluginDir        or "/usr/lib/collectd"
-	end
-	if sections and sections.collectd_rrdtool then
-		self._rrddir  = sections.collectd_rrdtool.DataDir  or "/tmp/rrd"
-	end
+	self._host    = host or uci:get("collectd", "globals", "hostname") or sys.hostname()
+	self._libdir  = uci:get("collectd", "globals", "PluginDir") or "/usr/lib/collectd"
+	self._rrddir  = uci:get("collectd", "rrdtool", "DataDir") or "/tmp/rrd"
```
hnyman: Please elaborate why do you see the safeguards added by d04f667383 can be now removed.

feckert: Thanks for your feedback. I think this is not neccessary because if the section is not present it will return **nil**. But this is only assumption I will have to check this. Thanks for the hint.

feckert: So I have checked it.

1. If collectd config file is not present then the default value after **or** is used because the uci:get returned nil
2. If the rrdtool config section is not presented then the default value after **or** is used because the uci:get returned nil too
3. If the rrdtool section does not have a the option then the default value after **or** is used as well because uci:get returned nil.

So I think this is not necessary anymore because we use a function call from **uci:get** and do not use a variable which could be nil.

feckert:
@hnyman @jow- any additional hints from your side?

hnyman:
Sorry,but  I have not had time to test your code.  
So this is just based on reading the code and trying to understand things.

I have not yet figured out where the possible new "LuCI defaults" that differ from "upstream collectd defaults" or from "OpenWrt collectd package defaults", are actually set for the collectd configuration.  

If I install the new luci-app-statistics package to a new system without any trace of collectd, how do the LuCI defaults get set sensibly. For example, memory, how does ValuesAbsolute=yes get set, as it is not the collectd default.  Or the ping target ?  Or the CPU temp sensors to be used, etc.?

(E.g. with community builds or private builds, it makes sense to tailor some device-specific settings at the build phase.)

Could be that I am just not getting/noticing something, but I can't figure it out. The old config file is deleted, but how is the corresponding thing rebuilt?

feckert:
> Sorry,but I have not had time to test your code.

I know that there can't be enough time. 

> I have not yet figured out where the possible new "LuCI defaults" that differ from "upstream collectd defaults" or from "OpenWrt collectd package defaults", are actually set for the collectd configuration.

Sorry I do not understand this?


> If I install the new luci-app-statistics package to a new system without any trace of collectd, how do the LuCI defaults get set sensibly. For example, memory, how does ValuesAbsolute=yes get set, as it is not the collectd default. Or the ping target ? Or the CPU temp sensors to be used, etc.?

Good point, I haven't thought about that yet.
We could change the default config in collectd https://github.com/openwrt/packages/blob/master/utils/collectd/files/collectd.uci so that we have the same configuration as in the old luci_statistics config file.

This needs to be configured like any other service.

hnyman:
> Sorry I do not understand this?

It is explained below more clearly: 

> >  If I install the new luci-app-statistics package to a new system without any trace of collectd, how do the LuCI defaults get set sensibly. For example, memory, how does ValuesAbsolute=yes get set, as it is not the collectd default. Or the ping target ? Or the CPU temp sensors to be used, etc.?

> Good point, I haven't thought about that yet.
> We could change the default config in collectd https://github.com/openwrt/packages/blob/master/utils/collectd/files/collectd.uci so that we have the same configuration as in the old luci_statistics config file.
> 
> This needs to be configured like any other service.

Sounds like you are now proposing converting collectd to something mimicing luci_statistics ...

So far luci_statistics has overriden the collectd default config if it has been installed. Plain collectd has been available by itself withotu any knowledge about LuCI stats needs if somebody has wanted just collectd.

Sounds strange that we would now convert collectd to be LuCI stats by default (if you just install collectd).

I think that we still need to ship the luci_statistics config file with specific defaults.

feckert:
@hnyman How to handle this? Should we add a preinstall script do move the installed config of collectd and install the config from luci-app-statisitcs?

hnyman:
What is wrong with the current approach? 
I guess that >98% of collectd users in OpenWrt context use it unknowingly for LuCI statistics, but have no actual interest for its other possible uses (and principally do not even know that collectd is the stats  backend).

The current approach that there is a separate luci_statistics config file and a script (stat-genconfig) that generates the collectd.conf for statistics usage just works. We need to be able to set some options tailored for luci stats that differ from the defaults in the separately installed independent collectd package, and similarly we should be able to ship some plugins included but disabled.  

Trying to serve the rare case where collectd is used for other purposes alongside the LuCI stats is a corner case from my perspective, so I wonder if this whole effort is needed. 

Trying to match all possible use case like "existing collectd config when installing luci stats", "luci stats and later some specific collectd config for other purposes" can get difficult rather soon. So trying to create a comprehensive post-install script could be difficult.

feckert:
> What is wrong with the current approach?

Nothing is wrong, but I have thought to myself, every app in the LuCI uses uci binding from the service it uses. Only luci-app-statisitics is dancing out of the series.

> I guess that >98% of collectd users in OpenWrt context use it unknowingly for LuCI statistics, but have no actual interest for its other possible uses (and principally do not even know that collectd is the stats backend).

That's what I think too. But nevertheless I think that the uci binding is better kept in service by the collectd than in the luci-app-statisitcs. I also think that the maintenance of the uci binding should be centralized and not in the collectd and in the luci-app-statisticcs (stat-genconfig). That brings more disadvantages than advantages since I have to edit a change of the uci binding twice.

> The current approach that there is a separate luci_statistics config file and a script (stat-genconfig) that generates the collectd.conf for statistics usage just works. We need to be able to set some options tailored for luci stats that differ from the defaults in the separately installed independent collectd package, and similarly we should be able to ship some plugins included but disabled.

We can update the default configuration of the collected package so that it meets the basic configuration of luci-app-statistics. Also the dependency of the collectd package could be updated so that the default plugins get installed.

> Trying to serve the rare case where collectd is used for other purposes alongside the LuCI stats is a corner case from my perspective, so I wonder if this whole effort is needed.

You may be right. But from my point of view the configuration creation simply belongs to the collectd package. I see it as preparation for other things I have thought about what we could do with collectd additional.

> Trying to match all possible use case like "existing collectd config when installing luci stats", "luci stats and later some specific collectd config for other purposes" can get difficult rather soon. So trying to create a comprehensive post-install script could be difficult.

I do not think so. If we update the collectd package with a default configuration that is the same as now the configuration from luci-app-statistics and update the dependency, then we met the requirements and we have the same tasks as before.
Only the configuration parameters in luci-app-statistics for the rrd view are then configured in luci-app-statistics. The configuration of the firewall configuration is a special case that must belong to luci-app-statistics because the generation of the configuration is more user friendly due the fact that it uses lua to get the exact rule number for the iptables collectd configuration.

Finally, I just want to emphasize that I think it makes sense to change luci-app-statistics to use the uci binding from the collectd. That it is advantageous for several reasons as mentioned above.

feckert:
@jow- @hnyman ping How do we proceed now with this PR?

hnyman:
> How do we proceed now with this PR?

Drop it?
I still do not see major benefits from it. Sorry.

I still see value in leaving the normal collectd clean for those who want to use it separately, and then having a separate dedicated luci_statistics uci config file for that stats part. Forcing luci_statistics defaults to be the collectd defaults sounds strage to me.

hnyman:
I will close this. The approach did not get traction, and wince the update from jow today, there are naturally massive conflicts.

feckert:
I find it a pity that this is not pursued further. It would have had certain advantages. First of all the double code maintenance would have been dropped. But that's the way opensource is, maybe I'll try it again another time when the time is right.
Thanks anyway

PR #3145, opened 2019-10-05
luci-mod-system: refine flash reset

ptpt52:
We determine flash reset capability by checking overlayfs in /proc/mounts

Traversing the ubi partitions is a little bad, it could not cover all the case
for some case if `rootfs_data` locates at `ubi_vol > 4`

and Traversing is not fast.

### DIFF
--- a/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json
+++ b/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json
@@ -34,7 +34,7 @@
 				"/proc/mtd": [ "read" ],
 				"/proc/partitions": [ "read" ],
 				"/proc/sys/kernel/hostname": [ "read" ],
-				"/sys/devices/virtual/ubi/*/name": [ "read" ]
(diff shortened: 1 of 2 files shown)


ptpt52:
patch rebased and up to date

jow-:
Hmm, the `jffs2reset` program which is eventually called explicitly looks for a `rootfs_data` MTD or UBI partition. The mount check you propose would match USB based overlay mounts too which we do not want.

So you say that simply increasing the amount of UBI volumes to scan is too slow on your device?

ptpt52:
> The mount check you propose would match USB based overlay mounts too which we do not want.

but it is acceptable if it is `USB based overlay`
that means it is safe to do reset even if we have `USB based overlay`

ptpt52:
> So you say that simply increasing the amount of UBI volumes to scan is too slow on your device?

first yes,
and second
increasing the amount of UBI volumes to scan, but what amount is Reasonable and complete?

jow-:
Merged, thanks!

PR #3146, opened 2019-10-06
luci-app-https_dns_proxy: bugfix: fix dnsmasq and subnet_addr configuration

antoinedeschenes:
- Port number requires a hash sign with dnsmasq
- Subnet address requires CIDR notation

Signed-off-by: Antoine Deschênes <antoine@antoinedeschenes.com>


feckert:
This is a duplicate https://github.com/openwrt/luci/pull/3134 @stangri

PR #3148, opened 2019-10-07
luci-app-https_dns_proxy: bugfix: subnet_addr option should be cidr

antoinedeschenes:
Subnet address requires CIDR notation

Signed-off-by: Antoine Deschênes <antoine@antoinedeschenes.com>

PR #3151, opened 2019-10-08
luci-app-https_dns_proxy: bugfix: subnet address datatype, add DNS.SB provider

stangri:
This addresses the issue mentioned in #3148.

### DIFF
--- a/applications/luci-app-https_dns_proxy/Makefile
+++ b/applications/luci-app-https_dns_proxy/Makefile
@@ -10,7 +10,7 @@ LUCI_TITLE:=HTTPS DNS Proxy Web UI
 LUCI_DESCRIPTION:=Provides Web UI for HTTPS DNS Proxy
 LUCI_DEPENDS:=+luci-mod-admin-full +https_dns_proxy
 LUCI_PKGARCH:=all
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 include ../../luci.mk
 
--- a/applications/luci-app-https_dns_proxy/luasrc/model/cbi/https_dns_proxy.lua
+++ b/applications/luci-app-https_dns_proxy/luasrc/model/cbi/https_dns_proxy.lua
@@ -1,4 +1,6 @@
 local uci = require("luci.model.uci").cursor()
+local dispatcher = require("luci.dispatcher")
+
 function uci_del_list(conf, sect, opt, value)
   local lval = uci:get(conf, sect, opt)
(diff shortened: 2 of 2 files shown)


feckert:
Could we split the commit into smaller peaces?
Also is it necessary that we check in code comments?

feckert:
> You want me to separate commits for bugfix and adding new DoH providers?

See it as a suggestion and improvement from my side. It is easier for the reviewer to have small commits. Your change makes a fix and a improvement in one commit. So if we move this in the future to a stable branch it es easier to have small commits because we only want to cherrry-pick the bugfix and not a new feature.

> It would make it easier for those willing to experiment with other DoH servers to do so. I'm waiting for feedback from @aarond10 on support on the main package.

A normal user without coding experience will not edit this file. So maybe we should read the values from a json file. So we do not have to edit the source file in the future. See https://github.com/openwrt/luci/tree/master/applications/luci-app-statistics/luasrc/statistics/plugins

```
return {
	value = "https://dns.adguard.com/dns-query?",
	i18n = _(AdGuard (Standard))
}
```
But this is only a suggestion from my side. How I would make this.

hnyman:
> Your change makes a fix and a improvement in one commit. So if we move this in the future to a stable branch it es easier to have small commits because we only want to cherrry-pick the bugfix and not a new feature.

I agree with the advice about separate commits. But I feel that on non-core apps it makes sense to pick up also the (small) improvements to stable branches. We do not need to be quite that strict there.  (Especially 19.07 has not yet seen any rc builds, so there is no locked-down feature set or anything like that)

For package maintainers it makes sense to keep branches in sync if possible, of course without breaking things in release branches.

feckert:
> > Your change makes a fix and a improvement in one commit. So if we move this in the future to a stable branch it es easier to have small commits because we only want to cherrry-pick the bugfix and not a new feature.
> 
> I agree with the advice about separate commits. But I feel that on non-core apps it makes sense to pick up also the (small) improvements to stable branches. We do not need to be quite that strict there. (Especially 19.07 has not yet seen any rc builds, so there is no locked-down feature set or anything like that)
> 

This was only an advice from my side.
So please go ahead

@hnyman @stangri if this is ok for you I will merge this and close https://github.com/openwrt/luci/pull/3148 because this fixes the same issue.

stangri:
> But this is only a suggestion from my side. How I would make this.

Thanks, I'll look into this for the future updates.

PR #3172, opened 2019-10-10
luci-base: widgets.js: add user and group select

ysc3839:
There's few applications provide user selection.
https://github.com/openwrt/luci/blob/07dbee37f858b93c10fe5114fbe55e36ddb0d654/applications/luci-app-aria2/luasrc/model/cbi/aria2/config.lua#L65 https://github.com/openwrt/luci/blob/9a489da90f1edeb01e88959c3092602bccf50a79/applications/luci-app-transmission/luasrc/model/cbi/transmission.lua#L28
However, after moving to client-siew view, we can't execute command directly from client-side. So I think we need corresponding widgets.
<del>I also removed the `-- Deprecated` comment of execl in util.lua, because rpcd/luci uses this function.</del>

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/tools/widgets.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/widgets.js
@@ -3,6 +3,19 @@
 'require form';
 'require network';
 'require firewall';
+'require fs';
+
+function getUsers() {
+    return fs.lines('/etc/passwd').then(function(lines) {
+        return lines.map(function(line) { return line.split(/:/)[0] });
+    });
+}
+
+function getGroups() {
+    return fs.lines('/etc/group').then(function(lines) {
+        return lines.map(function(line) { return line.split(/:/)[0] });
+    });
+}
 
 var CBIZoneSelect = form.ListValue.extend({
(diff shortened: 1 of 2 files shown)


jow-:
I'd prefer to whitelist read access to `/etc/passwd` and `/etc/group`, then use `fs.js` to read these files, e.g.:

```
'require fs';

...

function getUsers() {
    return fs.lines('/etc/passwd').then(function(lines) {
        return lines.map(function(line) { return line.split(/:/)[0] });
    });
}

function getGroups() {
    return fs.lines('/etc/group').then(function(lines) {
        return lines.map(function(line) { return line.split(/:/)[0] });
    });
}
```

ysc3839:
@jow- Updated.

jow-:
Merged, thanks!

brainstorm:
Still doesn't seem to allow selecting a group? I'd like to have `nobody:nogroup` by default but files appear as `-rw-r--r--    1 nobody   root` instead in the target filesystem?: 

Screenshot 2025-04-13 at 8 53 16 pm

Screenshot 2025-04-13 at 8 53 56 pm

PR #3178, opened 2019-10-13
luci-proto-qmi: add PDP type option

nickberry17:
Signed-off-by: Nicholas Smith <nicholas.smith@telcoantennas.com.au>

### DIFF
--- a/protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
+++ b/protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
@@ -68,7 +68,7 @@ return network.registerProtocol('qmi', {
 		s.taboption('general', form.Value, 'pincode', _('PIN'));
 
 		o = s.taboption('general', form.ListValue, 'auth', _('Authentication Type'));
-		o.value('both', 'PAP/CHAP (both)');
+		o.value('both', 'PAP/CHAP');
 		o.value('pap', 'PAP');
 		o.value('chap', 'CHAP');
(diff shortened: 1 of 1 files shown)


protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
```diff
@@ -97,5 +97,11 @@ return network.registerProtocol('qmi', {
 		o = s.taboption('advanced', form.Value, 'mtu', _('Override MTU'));
 		o.placeholder = dev ? (dev.getMTU() || '1500') : '1500';
 		o.datatype    = 'max(9200)';
+		
+		o = s.taboption('general', form.ListValue, 'pdptype', _('PDP Type'));
+		o.value('ipv4v6', 'IPv4/IPv6 (both)');
```
feckert: Could you please translate this? The value **(both)** is not common for all language.

nickberry17: Good point, I guess I'll get line 71 as well?  I based it on that format.  Will commit tomorrow

feckert: I would say omit "both" because I think it is self-explanatory

nickberry17: Done, but only for the IPv4/IPv6 value

feckert:
Thanks merged

PR #3188, opened 2019-10-15
luci-app-advanced-reboot: update Simplified Chinese translation

sotux:
update Simplified Chinese translation for luci-app-advance-reboot


applications/luci-app-advanced-reboot/po/zh-cn/advanced-reboot.po
```diff
@@ -1,5 +1,5 @@
 #
-# Yangfl <mmyangfl@gmail.com>, 2018.
+# Yangfl <mmyangfl@gmail.com>, 2018, 2019.
```
yousong: as noted before in #3184

sotux:
updated.

PR #3189, opened 2019-10-15
luci-app-opkg: update Simplified Chinese translation

sotux:
update Simplified Chinese translation for luci-app-opkg

### DIFF
--- a/applications/luci-app-opkg/po/zh-cn/opkg.po
+++ b/applications/luci-app-opkg/po/zh-cn/opkg.po
@@ -1,17 +1,19 @@
 #
 # Yangfl <mmyangfl@gmail.com>, 2018.
-# Zheng Qian <sotux82@gmail.com>, 2018.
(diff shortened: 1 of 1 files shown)


sotux:
Hi @yousong,
I noticed that  the translation of luci can merged by the hosted weblate.
It is very convenient for everyone to translate the project. But there is no review in the process, therefore the quality of the translation can't be guaranteed.
Is there a way to add a review step on the weblate?
Thanks.

yousong:
> Hi @yousong,
> I noticed that the translation of luci can merged by the hosted weblate.
> It is very convenient for everyone to translate the project. But there is no review in the process, therefore the quality of the translation can't be guaranteed.
> Is there a way to add a review step on the weblate?
> Thanks.

@aparcar may know more about this.

aparcar:
I think only project members can directly merge, everyone else can do suggestions

PR #3199, opened 2019-10-17
luci-app-cifsd: fix typo and add Simplified Chinese translation

ysc3839:
Ping @castillofrancodamian


castillofrancodamian:
Hi, what should I do with this?

ysc3839:
@castillofrancodamian Update Spanish translation.

ysc3839:
Something I forgot to say: I think the "Hide dot files" is not easy to explain in Chinese. (It means "hide files starting with a dot", which is too long.) So I leave it blink.

PR #3204, opened 2019-10-17
luci-base: Show a hint when WiFi key is from MAC

mgiganto:
Display a hint alerting about WiFi SSIDs configured using
the default key generated from label-mac or eth0-mac.

When all SSIDs are configured using a different key, the default_key
value is removed from system config and the alert gets disabled
completely for better performance.

This PR is related with https://github.com/openwrt/openwrt/pull/2408

Signed-off-by: Manuel Giganto <mgigantoregistros@gmail.com>

### DIFF
--- a/modules/luci-base/luasrc/sys.lua
+++ b/modules/luci-base/luasrc/sys.lua
@@ -553,6 +553,60 @@ function wifi.getiwinfo(ifname)
 	return { ifname = ifname }
 end
 
+function wifi.checkkey()
+	local default_key = uci:get("system", "@system[0]", "default_key")
+	if not default_key then
+		return
+	end
+
+	ntm.init()
+	local rt = ""
+	local cachecfg = {}
+	local handle = io.popen("date -r /etc/config/wireless +%s 2>/dev/null")
+	local configtime = "0"..handle:read("*a"):sub(0,-2)
+	handle:close()
+	handle = io.open("/tmp/luci-wireless-cachecfg", "r")
+	if handle then
+		cachecfg = luci.util.restore_data(handle:read("*all"))
+		handle:close()
+		if cachecfg.configtime >= configtime then
+			return cachecfg.ssids
+		end
+	end
+
+	for _, dev in ipairs(ntm:get_wifidevs()) do
+		for _, net in ipairs(dev:get_wifinets()) do
+			local wiface = {
+				ssid	= net:active_ssid(),
+				device	= dev:name(),
+				key	= net:get("key")
+			}
+
+			if wiface.key == default_key then
+				rt = rt .. ("%s (%s), "):format(wiface.ssid, wiface.device)
+			end
+		end
+	end
+
+	if rt ~= "" then
+		cachecfg.ssids 		= rt:sub(0,-3)
+		cachecfg.checktime 	= os.time()
+		cachecfg.configtime 	= configtime
+		handle = io.open("/tmp/luci-wireless-cachecfg", "w")
+		if handle then
+			handle:write(luci.util.serialize_data(cachecfg))
+			handle:close()
+		end
+		return rt:sub(0,-3)
+	else
+		local c = uci:changes("system");
+		uci:set("system","@system[0]","default_key","")
+		if c and not next(c) then
+			uci:save("system")
+			uci:commit("system")
+		end
+	end
+end
 
 init = {}
 init.dir = "/etc/init.d/"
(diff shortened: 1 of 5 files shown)


feckert:
Please fix your indentation first in the lua callback **wifi.checkkey**.
Does https://github.com/openwrt/openwrt/pull/2408 needs to be merged first for this feature?

mgiganto:
> Does [openwrt/openwrt#2408](https://github.com/openwrt/openwrt/pull/2408) needs to be merged first for this feature?
That PR sets the alert so the hint will be triggered. Both PRs can merge anytime in any order, they collaborate, but without that PR, this change won't do anything...

feckert:
Okay, got it.
Thanks.
But we should wait until the pullrequest is merged.
Because if not then we have a change in the LuCI that never comes to wear.

modules/luci-base/luasrc/sys.lua
```diff
@@ -8,6 +8,7 @@ local nixio  = require "nixio"
 local fs     = require "nixio.fs"
 local uci    = require "luci.model.uci"
 local ntm    = require "luci.model.network"
+local string = require "string"
```
CodeFetch: Importing "string" is not necessary.

modules/luci-base/luasrc/sys.lua
```diff
@@ -552,6 +553,39 @@ function wifi.getiwinfo(ifname)
...
 end
 
+function wifi.checkkey()
+	ntm.init()
+	local dev
+	local rt = ""
+	local default_key = uci:get("system", "@system[0]", "default_key")
+
+	for _, dev in ipairs(ntm:get_wifidevs()) do
+		local net
```
CodeFetch: Setting this variable local is not needed. Actually the variable in the for loop is a completely different one even if they both are called "net". So if you'd set "local net = '123'" a print(net) after the for-loop would still output "123". Do you understand? loop-variables are always local-scope in Lua.

modules/luci-base/luasrc/sys.lua
```diff
@@ -552,6 +553,39 @@ function wifi.getiwinfo(ifname)
...
+		for _, net in ipairs(dev:get_wifinets()) do
+			local wiface = {
+				ssid	= net:active_ssid(),
+				device	= dev:name(),
+				key	= net:get("key")
+			}
+
+			if wiface.key == default_key then
+				rt = rt .. string.format("%s (%s), ", wiface.ssid, wiface.device)
```
CodeFetch: "%s (%s), ":format(wiface.ssid, wiface.device)

modules/luci-base/luasrc/sys.lua
```diff
@@ -552,6 +553,39 @@ function wifi.getiwinfo(ifname)
...
+			end
+		end
+	end
+
+	if rt ~= "" then
+		return string.sub(rt,0,-3)
```
CodeFetch: rt:sub(0,-3)

modules/luci-base/luasrc/sys.lua
```diff
@@ -552,6 +553,39 @@ function wifi.getiwinfo(ifname)
...
+	else
+		local c = uci:changes("system");
+		uci:set("system","@system[0]","default_key","")
+		if c and not next(c) then
+			uci:save("system")
+			uci:commit("system")
+		end
+		return nil
```
CodeFetch: "return nil" is superfluous.

modules/luci-base/luasrc/sys.lua
mgiganto: Glass clear! 
I never used lua before and I got the wrong concept while looking at 
/usr/lib/lua/luci/tools/status.lua, so we may want to clear that code :)

jow-:
I don't like this code. You'll instantiate the network model class on every single view render. This adds a massive amount of overhead to every single LuCI operation.

We need to find a more lightweight way to check the default key condition.

mgiganto:
> I don't like this code. You'll instantiate the network model class on every single view render. This adds a massive amount of overhead to every single LuCI operation.

I agree with you @jow-, but when the user changes the key, it gets completely disabled, it will only check if system.@system[0].default_key and because it's null it ends. Isn't this enough?

mgiganto:
@jow- , although it was implemented as I mentioned before, it was removed from my last commit when I was testing it... 
Anyway it is back and this version also includes a method to cache the config and check if it is changed since it lasts check, avoiding to go through network model class... 
I thought about doing wireless md5sum but most probably it would be heavier than this implemented method based on checking the file modification time...

Please let me know your excellent thoughts :)

jow-:
Closing this as the related PR in OpenWrt core has been rejected.

PR #3206, opened 2019-10-17
Update from Weblate

weblate:


aparcar:
I'll now stop manual triggering of these PR, this was juts a final test if the SoB works correctly. From now on every 24 hours a new PR should appear (or the latest updated if unmerged) with all latest changes.

PR #3209, opened 2019-10-18
luci-app-banip: rename path for Simplified/Tranditional Chinese trans…

sotux:
As zh_Hans/zh_Hant is not the correct path for OpenWrt i18n path.


jow-:
Thats due to the new weblate system, it dictates the paths we have to use, I think.

@aparcar - any ideas how to solve it?

aparcar:
I think the problem only occurs on "newly" generated PO files as weblate uses zh_Hans instead of zh-cn. However the renamed files fitting OpenWrts i18n should still be detected as there is some support for [aliases](https://github.com/WeblateOrg/language-data/blob/master/aliases.csv#L63).

Therefore, I'll move all zh_Han{s,t} files and whenever there is a new component this step has to be performed again.

We could host weblate on our own and change the language definitions, however I wouldn't want to setup such a system.

aparcar:
Superseded by df6abf5

aparcar:
@jow- the solution is a bit unclear, the current bug boils down to [this](https://github.com/WeblateOrg/weblate/issues/3126) issue however other projects [move on](https://github.com/django/django/commit/e5e044da87800feb6ef63fef1765d8c05022d926) to "new language codes" a while ago. Firefox seems to discussing this [just recently](https://github.com/mozilla/addons-server/issues/5829).

Maybe the best would be to move on to zh_Han{s,t} and add a few lines of backward compatibility to the [dispatcher.lua](https://github.com/openwrt/luci/blob/master/modules/luci-base/luasrc/dispatcher.lua#L245). Sadly I'm not used to Lua so can't come up with a clean implementation for that...

PR #3215, opened 2019-10-18
luci-app-https_dns_proxy: better status/control/dhcp integration

stangri:
Improvements include:
 - Better integration with DHCP config: create the backup of the originally used servers and restore it on uninstall/service stop.
 - Better service control management in luci app, allowing user to start, reload, stop, enable and disable the service from WebUI.
 - Better service status display in WebUI.

PS. There's quite a bit of duplicate lua code between controller/cbi/view files. If anyone has a suggestion how can I keep it in just one of those files, I'd be much obliged.

### DIFF
--- a/applications/luci-app-https_dns_proxy/Makefile
+++ b/applications/luci-app-https_dns_proxy/Makefile
@@ -10,7 +10,45 @@ LUCI_TITLE:=HTTPS DNS Proxy Web UI
 LUCI_DESCRIPTION:=Provides Web UI for HTTPS DNS Proxy
 LUCI_DEPENDS:=+luci-mod-admin-full +https_dns_proxy
 LUCI_PKGARCH:=all
-PKG_RELEASE:=5
+PKG_RELEASE:=7
+
+define Package/luci-app-https_dns_proxy/postinst
+#!/bin/sh
+# check if we are on real system
+if [ -z "$${IPKG_INSTROOT}" ]; then
+	uci -q batch <<-EOF >/dev/null
+	delete ucitrack.@https_dns_proxy[-1]
+	add ucitrack https_dns_proxy
+	set ucitrack.@https_dns_proxy[-1].init=https_dns_proxy
+	commit ucitrack
+EOF
+	if ! uci -q get dhcp.@dnsmasq[0].doh_backup_server >/dev/null; then
+		for i in $$(uci -q get dhcp.@dnsmasq[0].server); do
+			uci -q add_list dhcp.@dnsmasq[0].doh_backup_server="$${i}"
+		done
+		uci -q commit dhcp
+	fi
+	rm -f /tmp/luci-indexcache
+fi
+exit 0
+endef
+
+define Package/luci-app-https_dns_proxy/prerm
+#!/bin/sh
+# check if we are on real system
+if [ -z "$${IPKG_INSTROOT}" ]; then
+	if uci -q get dhcp.@dnsmasq[0].doh_backup_server >/dev/null; then
+		uci -q del dhcp.@dnsmasq[0].server || true
+		for i in $$(uci -q get dhcp.@dnsmasq[0].doh_backup_server); do
+			uci -q add_list dhcp.@dnsmasq[0].server="$${i}" || true
+		done
+		uci -q delete dhcp.@dnsmasq[0].doh_backup_server
+		uci -q commit dhcp
+		/etc/init.d/dnsmasq restart >/dev/null 2>&1 || true
+	fi
+fi
+exit 0
+endef
 
 include ../../luci.mk
 
--- a/applications/luci-app-https_dns_proxy/luasrc/controller/https_dns_proxy.lua
+++ b/applications/luci-app-https_dns_proxy/luasrc/controller/https_dns_proxy.lua
@@ -1,7 +1,103 @@
 module("luci.controller.https_dns_proxy", package.seeall)
 function index()
-	if not nixio.fs.access("/etc/config/https_dns_proxy") then
-		return
+	if nixio.fs.access("/etc/config/https_dns_proxy") then
+		entry({"admin", "services", "https_dns_proxy"}, cbi("https_dns_proxy"), _("HTTPS DNS Proxy"))
+		entry({"admin", "services", "https_dns_proxy", "action"}, call("https_dns_proxy_action"), nil).leaf = true
 	end
-	entry({"admin", "services", "https_dns_proxy"}, cbi("https_dns_proxy"), _("HTTPS DNS Proxy"))
+end
+
+function https_dns_proxy_action(name)
+	local packageName = "https_dns_proxy"
+	if name == "start" then
+		servers_backup()
+		luci.sys.init.start(packageName)
+		servers_apply_doh()
+		luci.util.exec("/etc/init.d/dnsmasq restart >/dev/null 2>&1")
+	elseif name == "action" then
+		luci.util.exec("/etc/init.d/" .. packageName .. " reload >/dev/null 2>&1")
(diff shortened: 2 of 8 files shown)


applications/luci-app-https_dns_proxy/Makefile
```diff
@@ -10,7 +10,45 @@ LUCI_TITLE:=HTTPS DNS Proxy Web UI
...
+#!/bin/sh
+# check if we are on real system
+if [ -z "$${IPKG_INSTROOT}" ]; then
+	uci -q batch <<-EOF >/dev/null
+	delete ucitrack.@https_dns_proxy[-1]
+	add ucitrack https_dns_proxy
+	set ucitrack.@https_dns_proxy[-1].init=https_dns_proxy
+	commit ucitrack
+EOF
+	if ! uci -q get dhcp.@dnsmasq[0].doh_backup_server >/dev/null; then
```
feckert: What das doh_backup_server mean?

applications/luci-app-https_dns_proxy/luasrc/controller/https_dns_proxy.lua
```diff
@@ -1,7 +1,103 @@
...
+		luci.sys.init.disable(packageName)
+	end
+	luci.http.prepare_content("text/plain")
+	luci.http.write("0")
+end
+
+local uci = require("luci.model.uci").cursor()
+
+function uci_del_list(conf, sect, opt, value)
+  local lval = uci:get(conf, sect, opt)
```
feckert: Please fix the indentation

applications/luci-app-https_dns_proxy/luasrc/view/https_dns_proxy/buttons.htm
```diff
@@ -0,0 +1,56 @@
+<%#
+  Copyright 2019 Stan Grishin <stangri@melmac.net>
+-%>
+
+<%-
+	local packageName = "https_dns_proxy"
```
feckert: Do we really need this packageName checks?
Could we not add **https_dns_proxy** as a dependency to the Makefile?

stangri: I use similar code between different packages, it's easier for me to set the variable for the specific package when replace all entries in the code.

applications/luci-app-https_dns_proxy/luasrc/controller/https_dns_proxy.lua
stangri: I'm not sure I understand. The code of the function is indented, should it not be? Are you referring to something else?

applications/luci-app-https_dns_proxy/Makefile
stangri: It's the dnsmasq server setting before the package install. To be used on service stop and package uninstall. DoH is a short-hand for DNS-over-HTTPS.

applications/luci-app-https_dns_proxy/luasrc/controller/https_dns_proxy.lua
feckert: The indentation there is <space> but we prefer <tabs>

stangri: Ah, I see. I'll play with the tab setting of my editor in the future, I still think this should be merged as is for now.

feckert: I would like to do, but if we fix the indentation afterwards, then we lose the history if we want to find out with git blame when and what has been changed. This makes it more difficult to find errors when something was added and why. I'd rather see it if it is entered correctly.

applications/luci-app-https_dns_proxy/Makefile
feckert: I don't know if that's okay if we add new uci option?

applications/luci-app-https_dns_proxy/luasrc/view/https_dns_proxy/buttons.htm
feckert: I do not understand that!

stangri: There's no dependency outside of the Makefile.

applications/luci-app-https_dns_proxy/luasrc/controller/https_dns_proxy.lua
stangri: I believe I've been using the same editor for a while, so I believe double-spaces are in all of my packages, both luci and regular. What would be the best way to convert it to tabs now? Should I just run sed -i 's/  /	/g' on all files?

feckert: I am using the following *.vimrc*. This helps me to see the tab and the whitespace in my code and also the boundary for 80 chars.
```
" Show whitespace char
set listchars=tab:\|.,trail:.,extends:>,precedes:<
set list

" Set char highlight columne
set colorcolumn=80
```
If you like then you should convert it to tabs but this is not necessary because it was already accepted.

applications/luci-app-https_dns_proxy/luasrc/model/cbi/https_dns_proxy.lua
```diff
@@ -39,8 +48,129 @@ function uci_add_list(conf, sect, opt, value)
...
+end
+
+local tmpfsStatus, tmpfsStatusCode
+local ubusStatus = util.ubus("service", "list", { name = packageName })
+local tmpfsVersion = tostring(util.trim(sys.exec("opkg list-installed " .. packageName .. " | awk '{print $3}'")))
+
+if not tmpfsVersion or tmpfsVersion == "" then
+  tmpfsStatusCode = -1
+  tmpfsVersion = ""
+  tmpfsStatus = packageName .. " " .. translate("is not installed or not found")
```
feckert: You have already add a dependency to *httpy_dns_proxy* to your Makefile.
https://github.com/openwrt/luci/blob/7707b679e20df9806fbcbc58d12116f9ee9a3850/applications/luci-app-https_dns_proxy/Makefile#L11
So if we install *luci-app-https_dnx_proxy* then the dependency get installed by opkg wo why do we need to check if *https_dns_proxy* is installed?

applications/luci-app-https_dns_proxy/luasrc/view/https_dns_proxy/buttons.htm
feckert: That is the right source code place for this discussion https://github.com/openwrt/luci/pull/3215/files#r339785930

applications/luci-app-https_dns_proxy/luasrc/model/cbi/https_dns_proxy.lua
stangri: Ah, I see what you mean. Well, the luci app can still be force-installed without the principal package, that's why there's a code in place to handle missing version information from the principal package.

feckert: From my point of view if I am not an expert of LuCI. Does this really make sense to force an installation of an package if the package system wants to prevent me to install this package.

If so we have a problem in **all** other package that have an dependency to an package. So we have to check this in the code as well. Why query extra the dependency when the package system opkg already offers this feature?

stangri: > Why query extra the dependency when the package system opkg already offers this feature?

Because I want to display the version of the installed principal service. If opkg returns nothing when I query for the principal service version, what do you suggest luci app do?

PR #3236, opened 2019-10-23
[RFC] luci-app-opkg: add ipk upload feature

ysc3839:
Sometime people just want to upload and install a package using luci, however currently it's impossible without other tools (scp, WinSCP).
This pull request add this feature.

There are some minor issues I request for comments:
1. The confirm dialog reuses `handleManualInstall`, and shows `/tmp/upload.ipk` as filename. Should I make it shows it's real name?
image
2. Should confirm dialog shows checksums and size info?

@jow-

### DIFF
--- a/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
+++ b/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
@@ -781,6 +781,38 @@ function handleOpkg(ev)
 	});
 }
 
+function handleUpload(ev)
+{
+	var path = '/tmp/upload.ipk';
+	return L.ui.showModalUpload(path)
+	.then(L.bind(function(btn, res) {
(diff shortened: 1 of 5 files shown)


jow-:
Showing the real (before upload) name would be better UX I think. I see you reused the upload code from flash.js - I think instead of copying it, we should move it to ui.js and use it from there.

ysc3839:
Now it looks like this.
image

ysc3839:
There's another issue, the uploaded file is not deleted after install. `handleOpkg` doesn't provide a "after install" callback.

jow-:
The not deleted temp file is acceptable for now, but can you factor out the upload code into `ui.js` ? The rest looks fine to me

ysc3839:
Ping @jow-

jow-:
Merged with slight changes in c360cdd1f0b03ec208eaf1f8f85893aec50c7555

Issue #3243, opened 2019-10-28
luci-app-vpnbypass: action call violates dispatcher


ananas77777:
Mon Oct 28 19:01:21 2019 daemon.err uhttpd[909]: /usr/lib/lua/luci/dispatcher.lua:397: Access Violation
Mon Oct 28 19:01:21 2019 daemon.err uhttpd[909]: The page at 'admin/services/vpnbypass/action/enable/' has no parent node so the access to this location has been denied.
Mon Oct 28 19:01:21 2019 daemon.err uhttpd[909]: This is a software bug, please report this message at https://github.com/openwrt/luci/issues
Mon Oct 28 19:01:21 2019 daemon.err uhttpd[909]: stack traceback:
Mon Oct 28 19:01:21 2019 daemon.err uhttpd[909]: 	[C]: in function 'assert'
Mon Oct 28 19:01:21 2019 daemon.err uhttpd[909]: 	/usr/lib/lua/luci/dispatcher.lua:397: in function 'dispatch'
Mon Oct 28 19:01:21 2019 daemon.err uhttpd[909]: 	/usr/lib/lua/luci/dispatcher.lua:144: in function </usr/lib/lua/luci/dispatcher.lua:143>

feckert:
What version?

ananas77777:
That's latest from souce 1.3.1-4

Only sorted when I changed the location to force on admin/services tab not admin/vpn

feckert:
Did you reboot your system so that the ccache of LuCI is new?

ananas77777:
Yeah rebooted and the issue persists up until I changed the entry to be admin/services, This problem is that you call the admin/services entry even if admin/vpn is in place

feckert:
Could this be the root cause of your problem?
https://github.com/openwrt/luci/blob/7707b679e20df9806fbcbc58d12116f9ee9a3850/applications/luci-app-vpnbypass/luasrc/controller/vpnbypass.lua#L7-L10

If this is the problem we have to fix also the line https://github.com/openwrt/luci/blob/7707b679e20df9806fbcbc58d12116f9ee9a3850/applications/luci-app-vpnbypass/luasrc/view/vpnbypass/js.htm#L35 @stangri

ananas77777:
[...]

PR #3244, opened 2019-10-28
[19.07] luci-app-vpnbypass: bugfix: call to action has no parent node

stangri:


feckert:
Thanks for the fast fix :+1:

PR #3245, opened 2019-10-28
luci-app-vpnbypass: bugfix: call to action has no parent node

stangri:


feckert:
Thanks for the fast fix :+1: 
But there's one little thing I have to say. With this commit also the license description (no change) was updated. This has nothing to do with the fix. This should be an extra commit in the future.
Just a small comment

PR #3246, opened 2019-10-28
[18.06] luci-app-vpnbypass: bugfix: call to action has no parent node

stangri:

Only merge if the luci-app-openvpn which creates its entries under the VPN (and not Services) node will ever be merged to 18.06.


feckert:
Thanks for the fast fix :+1:

PR #3249, opened 2019-10-30
luci-app-adblock: Spelling: UDP, TCP, e-mail, DNS, cache

comradekingu:


feckert:
Please add add a Signed-off-by with your real name and mail.
Have a look at the other commits.
Please also update add your commit header.

**luci-app-adblock: \<your message\>**

hnyman:
please also note that changing the strings this way without also changing the corresponding translations (*.po files), will invalidate those existing translations.  You should maybe use sed and edit also the corresponding strings in the .po files.

cc @dibdot

jow-:
Implemented manually with 7b4efa9ba

PR #3251, opened 2019-10-31
luci-app-https_dns_proxy: remove dnsmasq integration, better service status/control

stangri:

As dnsmasq integration is moved to the principal service in https://github.com/openwrt/packages/pull/10413, luci app no longer needs that functionality.

### DIFF
--- a/applications/luci-app-https_dns_proxy/luasrc/model/cbi/https_dns_proxy.lua
+++ b/applications/luci-app-https_dns_proxy/luasrc/model/cbi/https_dns_proxy.lua
@@ -1,5 +1,14 @@
+local sys = require "luci.sys"
+local util = require "luci.util"
+local ip = require "luci.ip"
+local fs = require "nixio.fs"
+local jsonc = require "luci.jsonc"
+local http = require "luci.http"
+local nutil = require "nixio.util"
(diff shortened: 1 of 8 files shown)


applications/luci-app-https_dns_proxy/luasrc/model/cbi/https_dns_proxy.lua
```diff
@@ -39,8 +48,97 @@ function uci_add_list(conf, sect, opt, value)
   uci:set(conf, sect, opt, lval)
 end
 
-m = Map("https_dns_proxy", translate("HTTPS DNS Proxy Settings"))
-m.template="cbi/map"
+function get_provider_name(value)
+  if value:match("dns\.adguard") then
```
jow-: In Lua regular expressions, the escape character is `%` so a dot must be escaped as `%.`

stangri: I have another update ready where the providers are moved to the plugin-style directory (as @feckert suggested) and this code is not present. However I was previously told to split the PRs into smaller chunks/features, so I haven't pushed it yet.
What should I do? Easiest option for me would be to close this PR and create the new one with the new code. Let me know.

PR #3256, opened 2019-11-03
luci-base: form.js: fix GridSection not using user-input name when add

ysc3839:
@jow- 
https://github.com/openwrt/luci/blob/6c7b1e156d8a3342864a5e5e03a8331f6819109f/modules/luci-base/htdocs/luci-static/resources/form.js#L1393-L1395
https://github.com/openwrt/luci/blob/6c7b1e156d8a3342864a5e5e03a8331f6819109f/modules/luci-base/htdocs/luci-static/resources/form.js#L874-L877


jow-:
Merged, thank you!

PR #3258, opened 2019-11-04
luci-app-shadowsocks-libev: add server links import feature

ysc3839:
@yousong 
Related docs:
https://github.com/Jigsaw-Code/outline-shadowsocksconfig
https://shadowsocks.org/en/config/quick-guide.html
https://shadowsocks.org/en/spec/SIP002-URI-Scheme.html

### DIFF
--- a/applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/shadowsocks-libev.js
+++ b/applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/shadowsocks-libev.js
@@ -238,5 +238,71 @@ return L.Class.extend({
 			window.open(L.url('admin/system/opkg') +
 				'?query=' + opkg_package, '_blank', 'noopener');
 		};
+	},
+	parse_uri: function(uri) {
+		var scheme = 'ss://';
+		if (uri && uri.indexOf(scheme) === 0) {
+			var atPos = uri.indexOf('@'), hashPos = uri.lastIndexOf('#'), tag;
+			if (hashPos === -1) {
+				hashPos = undefined;
+			} else {
+				tag = uri.slice(hashPos + 1);
+			}
+
+			if (atPos !== -1) { // SIP002 format https://shadowsocks.org/en/spec/SIP002-URI-Scheme.html
+				var colonPos = uri.indexOf(':', atPos + 1), slashPos = uri.indexOf('/', colonPos + 1);
+				if (colonPos === -1) return null;
+				if (slashPos === -1) slashPos = undefined;
+
+				var userinfo = atob(uri.slice(scheme.length, atPos)
+					.replace(/-/g, '+').replace(/_/g, '/')),
+					i = userinfo.indexOf(':');
+				if (i === -1) return null;
+
+				var config = {
+					server: uri.slice(atPos + 1, colonPos),
+					server_port: uri.slice(colonPos + 1, slashPos ? slashPos : hashPos),
+					password: userinfo.slice(i + 1),
+					method: userinfo.slice(0, i)
+				};
+
+				if (slashPos) {
+					var search = uri.slice(slashPos + 1, hashPos);
+					if (search[0] === '?') search = search.slice(1);
+					search.split('&').forEach(function(s) {
+						var j = s.indexOf('=');
+						if (j !== -1) {
+							var k = s.slice(0, j), v = s.slice(j + 1);
+							if (k === 'plugin') {
+								v = decodeURIComponent(v);
+								var k = v.indexOf(';');
+								if (k !== -1) {
+									config['plugin'] = v.slice(0, k);
+									config['plugin_opts'] = v.slice(k + 1);
+								}
+							}
+						}
+					});
+				}
+				return [config, tag];
+			} else { // Legacy format https://shadowsocks.org/en/config/quick-guide.html
+				var plain = atob(uri.slice(scheme.length, hashPos)),
+					firstColonPos = plain.indexOf(':'),
+					lastColonPos = plain.lastIndexOf(':'),
+					atPos = plain.lastIndexOf('@', lastColonPos);
+				if (firstColonPos === -1 ||
+					lastColonPos === -1 ||
+					atPos === -1) return null;
+
+				var config = {
+					server: plain.slice(atPos + 1, lastColonPos),
+					server_port: plain.slice(lastColonPos + 1),
+					password: plain.slice(firstColonPos + 1, atPos),
+					method: plain.slice(0, firstColonPos)
+				};
+				return [config, tag];
+			}
+		}
+		return null;
 	}
 });
--- a/applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
+++ b/applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
@@ -1,21 +1,65 @@
 'use strict';
 'require form';
+'require uci';
+'require ui';
 'require shadowsocks-libev as ss';
 
-function startsWith(str, search) {
-	return str.substring(0, search.length) === search;
-}
+var conf = 'shadowsocks-libev';
 
 return L.view.extend({
 	render: function() {
 		var m, s, o;
 
-		m = new form.Map('shadowsocks-libev', _('Remote Servers'),
+		m = new form.Map(conf, _('Remote Servers'),
 			_('Definition of remote shadowsocks servers.  \
 				Disable any of them will also disable instances referring to it.'));
 
 		s = m.section(form.GridSection, 'server');
 		s.addremove = true;
+		s.handleLinkImport = function() {
+			var textarea = new ui.Textarea();
+			ui.showModal(_('Import Links'), [
+				textarea.render(),
+				E('div', { class: 'right' }, [
+					E('button', {
+						class: 'btn',
+						click: ui.hideModal
+					}, [ _('Cancel') ]),
+					' ',
+					E('button', {
+						class: 'btn cbi-button-action',
+						click: ui.createHandlerFn(this, function() {
+							textarea.getValue().split('\n').forEach(function(s) {
+								var config = ss.parse_uri(s);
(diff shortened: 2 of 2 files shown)


applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
```diff
@@ -16,6 +17,42 @@ return L.view.extend({
...
+											debugger;
+										});
+									}
+								}, [ _('Import') ])
+							])
+						]);
+					});
+				})
+			}, [ _('Import Links') ]));
+			return el;
```
jow-: You don't call `resolveFn` anywhere here, that is probably not intended.

ysc3839: It's working in progress.

jow-: Ok :)

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/shadowsocks-libev.js
```diff
@@ -238,5 +242,74 @@ return L.Class.extend({
 			window.open(L.url('admin/system/opkg') +
 				'?query=' + opkg_package, '_blank', 'noopener');
 		};
+	},
+	startsWith: function(str, search) {
+		return startsWith(str, search);
+	},
+	parse_uri: function(uri) {
+		var scheme = 'ss://';
+		if (!!uri && startsWith(uri, scheme)) {
```
jow-: `if (uri && uri.indexOf(scheme) === 0) {`

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
```diff
@@ -26,7 +64,7 @@ return L.view.extend({
 	},
 	addFooter: function() {
 		var p = '#edit=';
-		if (startsWith(location.hash, p)) {
+		if (ss.startsWith(location.hash, p)) {
```
jow-: `if (location.hash.indexOf(p) === 0) {`

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/shadowsocks-libev.js
```diff
@@ -67,6 +67,10 @@ function ucival_to_bool(val) {
 	return val === 'true' || val === '1' || val === 'yes' || val === 'on';
 }
 
+function startsWith(str, search) {
```
jow-: This could be dropped if you replace it with `str.indexOf(search) === 0` checks.

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/shadowsocks-libev.js
```diff
@@ -238,5 +242,74 @@ return L.Class.extend({
 			window.open(L.url('admin/system/opkg') +
 				'?query=' + opkg_package, '_blank', 'noopener');
 		};
+	},
+	startsWith: function(str, search) {
```
jow-: This could be dropped if you replace it with `str.indexOf(search) === 0` checks.

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
```diff
@@ -1,21 +1,59 @@
...
 				Disable any of them will also disable instances referring to it.'));
 
 		s = m.section(form.GridSection, 'server');
 		s.addremove = true;
+		s.renderSectionAdd = function(extra_class) {
+			var el = form.GridSection.prototype.renderSectionAdd.apply(this, arguments);
+			el.appendChild(E('button', {
+				'class': 'cbi-button cbi-button-add',
+				'title': _('Import Links'),
+				'click': function() {
```
jow-: This is a personal taste matter, but I usually move complex event handler functions into separate `handleXYZ` methods which I then invoke. To do that, you need to preserve the `this` context though.

```
s.renderSectionAdd = L.bind(function(extra_class) {
    ...
}, this);
```

And later within the `el.appendChild()`:

```
E('button', {
    ...
    'click': ui.createHandlerFn(this, 'handleLinkImport')
}, [ _('Import Links') ]);
```

Then move this anonymous function code to a `handleLinkImport` view class method.

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
```diff
@@ -1,21 +1,59 @@
 'use strict';
 'require form';
+'require uci';
```
jow-: Can you also explicitely require `ui` and use it instead of `L.ui` throughout the code, similar to https://github.com/openwrt/luci/commit/86f492173d1daab8b75e4d4bd70ba06872ea70fa? When testing LuCI I found various race conditions where `ui` might not yet be loaded when the view is processed.

ysc3839:
Ping @yousong

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
```diff
@@ -1,21 +1,61 @@
...
+									var tag = config[1];
+									if (tag && !tag.match(/^[a-zA-Z0-9_]+$/)) tag = null;
+									var sid = uci.add(conf, 'server', tag);
+									config = config[0];
+									Object.keys(config).forEach(function(k) {
+										uci.set(conf, sid, k, config[k]);
+									});
+								}
+							});
+							return uci.save().finally(ui.hideModal);
```
yousong: Can we reload the view after the addition succeeded?

ysc3839: OK.

yousong: Have you tried it?  It does not work here.  Only "name" and "method" were set, other options like server address, port, etc. were missing.

yousong:
Please either write a small summary about what has changed between each of your revisions, or add commits and squash them only when they are ready to be merged.  That should be easier to review what has happened.

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
ysc3839: Sorry for my mistake.

applications/luci-app-shadowsocks-libev/htdocs/luci-static/resources/view/shadowsocks-libev/servers.js
```diff
@@ -1,21 +1,61 @@
...
+									var tag = config[1];
+									if (tag && !tag.match(/^[a-zA-Z0-9_]+$/)) tag = null;
+									var sid = uci.add(conf, 'server', tag);
+									config = config[0];
+									Object.keys(config).forEach(function(k) {
+										uci.set(conf, sid, k, config[k]);
+									});
+								}
+							});
+							return this.handleModalSave(uci);
```
ysc3839: I found a tricky way. Because `handleModalSave` calls `.save()` on first argument, passing `uci` works.
https://github.com/openwrt/luci/blob/6496ed88a6dfa15c0f34f3865a559e1c405cdc4f/modules/luci-base/htdocs/luci-static/resources/form.js#L1313-L1319
Or should I instead copy paste the code of `handleModalSave` here?

yousong: It looks a hack to me ;)  I prefer we make it clear by duplicating part of the logic from `CBITableSection.handleModalSave()`.  Let's hear @jow- 's opinion first.

The other thing is that it seems to me height of the textarea element cannot be resized (Chrome 77.0.3865.120).  It would be great if there is a way around this in the context of LuCI framework.

jow-: What about `this.map.handleModalSave(this.map, ev);` ?

ysc3839: There's no `handleModalSave` function in `this.map`.

jow-: Okay, and `form.TableSection.prototype.handleModalSave.call(this, this.map, ev);` ?

ysc3839: Imported data were lost. I think `uci.save()` is essential.

ysc3839: Ping @jow-

jow-: Ah right, the explicit uci save is needed because you create the data outside of the map scope. In this please use `return uci.save().then(L.bind(this.handleModalSave, this));`

ysc3839: No working. Because `handleModalSave` takes a `modalMap` argument, result in calling `undefined.save()`. If using `L.bind(this.handleModalSave, this, this.map)`, the config get deleted.
image

yousong:
Thank you!

PR #3267, opened 2019-11-06
luci-app-ddns: convert to client side implementation

Ansuel:
This is an initial pr for reworked luci-app-ddns. I removed some very eccessive checks and custom scripts are still not supported. I'm posting this to ask some review of the code, some improvement and also actually test it to make sure it does works correctly. 

Signed-off-by: Ansuel Smith <ansuelsmth@gmail.com>

PR #3291, opened 2019-11-13
luci-app-nextdns: add luci integration for nextdns package

rs:
Signed-off-by: Olivier Poitrey <rs@nextdns.io>

Description:

Adding support for [NextDNS](https://nextdns.io), a DNS over HTTPS service that provides customized block-lists (ad-block, malware protection, kid safety) as well advanced analytics amongst other things.

Depends on: https://github.com/openwrt/packages/pull/10535

Screen Shot 2019-11-12 at 6 29 40 PM
Screen Shot 2019-11-12 at 6 29 49 PM

### DIFF
--- a/applications/luci-app-nextdns/Makefile
+++ b/applications/luci-app-nextdns/Makefile
@@ -1,0 +1,12 @@
+# Copyright 2019 Olivier Poitrey (rs@nextdns.io)
+# This is free software, licensed under the Apache License, Version 2.0
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for NextDNS
+LUCI_DEPENDS:=+luci-compat +nextdns
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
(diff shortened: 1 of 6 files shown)


rs:
FYI, main PR have been merged.

ysc3839:
I think you should use client-side implementation. If you don't know how to do that I can make a pull request.

rs:
@ysc3839: what do you mean by client-side implementation?

ysc3839:
@rs #3267

rs:
Got it. We can definitely move to client side, and we will. Could we do that in a second time tho? I'd love to this module make the cut for 19.07.0 together with the daemon package which is already merged.

hnyman:
Let's merge this first. You can then convert it later. 
(there is not that much computational/conditional logic in the app, so the client-side implementation would not decrease router's CPU load that much.)

rs:
Awesome, so you need another PR to put it in 19.07?

hnyman:
Yes, preferably.

But if quite identical code works for 19.07 (and you have tested it...), I can also cherry-pick your single commit.

PR #3305, opened 2019-11-13
luci-app-mwan3: fix uptime/downtime rendering

hesalx:


feckert:
I do not think this is a fix.
We have in the `ubus call mwan3 status` result per interface the values online, offline and uptime.

- The uptime time is the time that from netifd view the interface should up.
- The online time is the time that from mwan3 view the interface is recognized as online and so logical connected (tracking OK).
- The offline time is the time that from mwan3 view the interface is recognized as offline and so logical disconnected (tracking NOK)

So we have no downtime.

hesalx:
@feckert thanks for the feedback! 
But it does not line up with my experience.
Did I misunderstand something?

I had `NaNh:NaNm:NaNs` as uptime/downtime in the LuCI. My changes did fix that completely for up and down statuses. I fixed it based on http response and `ubus` response. I see no `online`/`offline` at all.

I have `mwan3 2.7.14-1` and `luci-app-mwan3 git-19.312.54336-a987e21-1` on `OpenWrt 19.07.0-rc1` (no updates available for the mentioned packages).

Maybe `luci-app-mwan3` assumes different version of `mwan3` or something? I though it's either a typo or it's using some older version. But might also be that `mwan3` does what you say but in some recent (master) version of it.

```root@openwrt:~# ubus call mwan3 status```
```json
{
	"interfaces": {
		"wan1": {
			"age": 1703,
			"uptime": 142107,
			"downtime": 0,
			"score": 12,
			"lost": 0,
			"turn": 40,
			"status": "online",
			"running": true,
			"track_ip": [
				{
					"ip": "8.8.4.4",
					"status": "skipped",
					"latency": 0,
					"packetloss": 0
				},
				{
					"ip": "8.8.8.8",
					"status": "up",
					"latency": 0,
					"packetloss": 0
				}
			]
		},
		"wan2": {
			"age": 4,
			"uptime": 0,
			"downtime": 92967,
			"score": 0,
			"lost": 26472,
			"turn": 6624,
			"status": "offline",
			"running": true,
			"track_ip": [
				{
					"ip": "8.8.4.4",
					"status": "down",
					"latency": 0,
					"packetloss": 0
				},
				{
					"ip": "8.8.8.8",
					"status": "down",
					"latency": 0,
					"packetloss": 0
				}
			]
		}
	},
	"connected": {
		"ipv4": [
			"127.0.0.0/8",
			"192.168.0.0/24"
		],
		"ipv6": [
			"fe80::/64"
		]
	},
	"policies": {
		"ipv4": {
			"failover_policy": [
				{
					"interface": "wan",
					"percent": 100
				}
			]
		},
		"ipv6": {
			"failover_policy": [

			]
		}
	}
}
```

feckert:
> Maybe `luci-app-mwan3` assumes different version of `mwan3` or something? 

Yes this is the case thank you very much for the hint and finding this out. I didn't think of it, but that was the trigger!

There was an ubus API change in the mwan3 package in the master branch. So in master of the luci feed I also update the luci-app-mwan3 to use the new values. This must be in sync.

Since the merge of luci from master branch into the openwrt-19.07 branch the change in the luci-app-mwan3 was also applied into the openwrt-19.07 branch. But the changes in the package feeds were not applied.

We have now two possibilities.

Add the following missing commits of mwan3 from the master branch to the openwrt-19.07 branch:
https://github.com/openwrt/packages/commit/3be40786f54b5d6534ddd6a58c090b3963180eed#diff-8754b1a131a2655cc86d2a97700d0c2e
https://github.com/openwrt/packages/commit/6c2cd78c6f5015fe24078b98a7455ae0cc88ff07#diff-8754b1a131a2655cc86d2a97700d0c2e
https://github.com/openwrt/packages/commit/bfa849c14e8c3404407930fc0d5db454514e513f#diff-8754b1a131a2655cc86d2a97700d0c2e

Or Revert the following commit form the openwrt-19.07 branch:
https://github.com/openwrt/luci/commit/1991b4ae64af79e3cd346d8383d3a586a16f6b80#diff-d3354696844c133874447586958ddcdc

@jow- What should I do now? I think that's a matter of principle.
Thanks for feedback and advice.

jow-:
@feckert - I'd say backport mwan3. I don't see much value in having it outdated even before the 19.07.0 final

feckert:
@jow Thanks for your advice.
 I will backport all missing commits that are not presented in the openwrt-19.07 branchso we are in sync again with master.

feckert:
I have pushed all missing commits from the package feed master branch of mwan3 into the openwrt-19.07 branch.
So please try the upcoming snapshot of openwrt-19.07
This should be fixed now and this pullrequest is not necessary.
Please reopen if you do not think it works.

@hesalx thanks for notification :+1:

hesalx:
@feckert Thank you!
I don't want to use a snapshot for now as it's a lot of manual work for a small bug. I will stick with the ad-hoc patch. At least you confirmed those two lines are enough for the fix.

Glad to know this is now addressed in the upcoming 19.07 stable.

PR #3322, opened 2019-11-17
luci-app-nextdns: add luci integration for nextdns package

rs:
Signed-off-by: Olivier Poitrey <rs@nextdns.io>

Tested on 19.07.0-rc1


rs:
@hnyman here it is

PR #3323, opened 2019-11-17
luci-app-ddns: use the correct widget Select

ptpt52:
ip_network should be NetworkSelect
ip_interface/interface should be DeviceSelect


hnyman:
cc @Ansuel

Ansuel:
@hnyman  LGTM

PR #3328, opened 2019-11-18
luci-app-nextdns: port to client side

ysc3839:
@rs


applications/luci-app-nextdns/root/usr/share/rpcd/acl.d/luci-app-nextdns.json
```diff
@@ -0,0 +1,10 @@
+{
+	"luci-app-nextdns": {
+		"description": "Grant logread access to LuCI app nextdns",
+		"write": {
```
rs: Do we need write access here (I'm guessing the meaning)?

ysc3839: This is for execute `/sbin/logread`.

rs: But it says "write", that's why I'm confused.

ysc3839: I'm confused too.
https://github.com/openwrt/luci/blob/b1cb0d6d59c84a16f084e8ec91714a3d247a2700/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json#L56-L77

jow-: ACL group definitions are split into two scopes, `read` for operations without side effects and `write` for ones with side effects.

The `read` and `write` labels there are mainly designations, they do not imply write permission to the underlying resource. Right now that distinction is not important because LuCI does not yet make use of it, but in a future version there will be support for readonly user accounts, at that point the split becomes important.

In this particular case, the `exec` permission for the `logread` command should be moved into the `read` scope as `logread` has no side-effects; it creates no files and does not alter the system state, so its eligible for the readonly permission scope.

ysc3839: Fixed.

jow-:
Merged, thanks @ysc3839 !

PR #3339, opened 2019-11-20
luci-app-cifsd: enable cifsd app [19.07]

Andy2244:
Tested on mips/lantiq with 19.07-master.
The matching backport PR is here: openwrt/packages#10603

PS: Thanks @ysc3839 for the updated .js code!

### DIFF
--- a/applications/luci-app-cifsd/Makefile
+++ b/applications/luci-app-cifsd/Makefile
@@ -3,7 +3,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=Network Shares - CIFSD CIFS/SMB kernel fileserver
-LUCI_DEPENDS:=+cifsd-tools @BROKEN
+LUCI_DEPENDS:=+cifsd-tools
 
 include ../../luci.mk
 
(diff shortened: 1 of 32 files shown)


applications/luci-app-cifsd/Makefile
```diff
@@ -3,7 +3,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=Network Shares - CIFSD CIFS/SMB kernel fileserver
```
neheb: I currently have this installed and it shows up as Network Shares. I imagine if I had luci-app-samba installed it would be the same. Is this a bug?

Andy2244: Yes you should have two "Network Shares" entries than, i was considering adding something like:
```
Network Shares (Cifsd)
Network Shares (Samba4)
```
Yet i think this is a special case if someone has both servers installed? We can add the app name, but i'm not sure if this is necessary?

Andy2244: Jow noted we can ignore this atm and i added naming/version infos to the actual network pages so its easy to spot for what server you are making changes.

PR #3345, opened 2019-11-22
luci-app-cifsd: add version info

Andy2244:
* add version info
* set guest_ok = yes as default
* regen translations

### DIFF
--- a/applications/luci-app-cifsd/htdocs/luci-static/resources/view/cifsd.js
+++ b/applications/luci-app-cifsd/htdocs/luci-static/resources/view/cifsd.js
@@ -8,14 +8,19 @@ return L.view.extend({
 		return Promise.all([
 			L.resolveDefault(fs.stat('/sbin/block'), null),
 			L.resolveDefault(fs.stat('/etc/config/fstab'), null),
+			L.resolveDefault(fs.exec("cifsd", ["-h"]), {}).then(function(res) { return L.toArray((res.stderr || '').match(/version : (\S+ \S+)/))[1] }),
 		]);
 	},
 	render: function(stats) {
-		var m, s, o;
+		var m, s, o, v;
+		v = '';
 
 		m = new form.Map('cifsd', _('Network Shares'));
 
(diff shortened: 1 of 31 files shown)


linbox1968:
would it be possible to have the writeable option also in the template for shares?

To get this example to a device. I had to manually change the template to add this.
Maybe I'm doing something else wrong, but this was the only way to get the shares writeable

````
[sda]`
        path = /mnt/sda1
        writeable = yes
```

Andy2244:
@linbox1968 Thanks its a bug in the new luci-module .js code for cifsd/samba4, it should actually put in `read only = no` for the share based on luci.
Will be fixed asap!

Andy2244:
@jow- Whats going on here i can't get luci .js to write `no` on my options unless i also set `o.rmempty = false;`, which was not needed on the lua version?  Yet the value is not empty its 'no' ?

I also tried `o.optional = false;` which i assumed will always force a write to the config, yet still gets ignored in the 'no' case?

So whats going on here?

jow-:
LuCI (the new JS code) will remove the flag value if it matches the default. If the implicit Samba default of this option is `yes`, you should mark it as default `yes` as well.

Andy2244:
> LuCI (the new JS code) will remove the flag value if it matches the default. If the implicit Samba default of this option is `yes`, you should mark it as default `yes` as well.

mhh i always assumed the `default` value is just that, a "default" value for the UCI config?
I would not have thought that changing a default there had any impact on whether its written to the config or not. 

On a other note, why is the value written if i add `o.rmempty = false;` ? If defaults are not written, how does this change this logic?

jow-:
The `rmempty` option controls whether options that would be cleared out are removed from the config or not. In any case, the proper fix for Flag options is to set the default value to the implicit default value of the underlying app. The defualt value of a flag is not written to the config (option is removed), the non default value is written.

linbox1968:
Works perfect!!!   Thanks for the fast fix and implementation.......

I love the cifsd It is sooooo much smaller than samba even 3.6 was a lot larger

PR #3348, opened 2019-11-22
luci-app-yggdrasil: Add application for Yggdrasil

zhoreeq:
Signed-off-by: George Iv <57254463+zhoreeq@users.noreply.github.com>

Tested on: snapshot with yggdrasil v0.3.11-4

### DIFF
--- a/applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/keys.js
+++ b/applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/keys.js
@@ -1,0 +1,22 @@
+'use strict';
+'require form';
+
+return L.view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('yggdrasil', 'Yggdrasil');
+
+		s = m.section(form.TypedSection, "yggdrasil", _("Encryption keys")); 
+		s.anonymous = true;
+
+		s.option(form.Value, "EncryptionPublicKey", _("Encryption public key"));
+		s.option(form.Value, "EncryptionPrivateKey", _("Encryption private key"),
+			_("Keep this private. When compromised, generate a new keypair and IPv6."));
+		s.option(form.Value, "SigningPublicKey", _("Signing public key"));
+		s.option(form.Value, "SigningPrivateKey", _("Signing private key"),
+			_("Keep this private. When compromised, generate a new keypair and IPv6."));
+
+		return m.render();
+	}
+});
--- a/applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/peers.js
+++ b/applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/peers.js
@@ -1,0 +1,31 @@
+'use strict';
+'require form';
+
+return L.view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('yggdrasil', 'Yggdrasil');
+
+		o = m.section(form.TableSection, "peer", _("Peers"),
+			_("List of connection strings for outbound peer connections in URI format, " +
+				"e.g. tcp://a.b.c.d:e or socks://a.b.c.d:e/f.g.h.i:j. These connections " +
+				"will obey the operating system routing table, therefore you should " +
+				"use this section when you may connect via different interfaces."));
+		o.option(form.Value, "uri", "URI");
+		o.anonymous = true;
+		o.addremove = true;
+
+		o = m.section(form.TableSection, "interface_peer", _("Interface peers"),
+			_("List of connection strings for outbound peer connections in URI format, " +
+				"arranged by source interface, e.g. { \"eth0\": [ tcp://a.b.c.d:e ] }. " +
+				"Note that SOCKS peerings will NOT be affected by this option and should " +
+				"go in the \"Peers\" section instead."));
+		o.option(form.Value, "interface", _("Interface"));
+		o.option(form.Value, "uri", "URI");
+		o.anonymous = true;
+		o.addremove = true;
+
+		return m.render();
+	}
+});
--- a/applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/session_firewall.js
+++ b/applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/session_firewall.js
@@ -1,0 +1,38 @@
+'use strict';
+'require form';
+
+return L.view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('yggdrasil', 'Yggdrasil');
+
+		s = m.section(form.TypedSection, "yggdrasil", _("Session firewall settings"));
+		s.anonymous = true;
+
+		s.option(form.Flag, "SessionFirewall_Enable", _("Enable session firewall"),
+			_("If disabled, network traffic from any node will be allowed. If enabled, the below rules apply"));
+		s.option(form.Flag, "SessionFirewall_AllowFromDirect", _("Allow from direct"), 
+			_("Allow network traffic from directly connected peers"));
+		s.option(form.Flag, "SessionFirewall_AllowFromRemote", _("Allow from remote"), 
+			_("Allow network traffic from remote nodes on the network that you are not directly peered with"));
+		s.option(form.Flag, "SessionFirewall_AlwaysAllowOutbound", 
+			_("Always allow outbound"), _("Allow outbound network traffic regardless of AllowFromDirect or AllowFromRemote"));
+
+		s = m.section(form.TableSection, "whitelisted_encryption_public_key", 
+			_("Whitelisted public keys"),
+			_("Network traffic is always accepted from those peers, regardless of AllowFromDirect or AllowFromRemote"));
+		s.option(form.Value, "key", _("Public key"));
+		s.anonymous = true;
+		s.addremove = true;
+
+		s = m.section(form.TableSection, "blacklisted_encryption_public_key", 
+			_("Blacklisted public keys"), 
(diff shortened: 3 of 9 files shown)


aparcar:
I think LuCI is moving on the JavaScript, from what I understand is the Lua to JavaScript conversion pretty straight forward. Would that be an option for you?

zhoreeq:
@aparcar sure, if there is API documentation. This app doesn't seem too complex to convert, but I can only see Lua API in the current docs and guidelines. 

If Lua apps are not totally deprecated yet, can it be converted after we add it?

aparcar:
I think the current documentation is [here](https://openwrt.github.io/luci/jsapi/) and [here](https://github.com/openwrt/luci/pull/3344/files) is a recent example of sombody moving from lua to js

zhoreeq:
@aparcar pushed a JS version. Review please.

wfleurant:
the transition to JS looks too good to be true

aparcar:
@wfleurant Thank you for looking into that. Could you please test the app and approve it? I'm then happy to merge.

aparcar:
@zhoreeq Ideally, not necessarily now, the Lua parts would be replaced by `rpcd` calls.

zhoreeq:
I've just removed luasocket dependency and controller is now using `yggdrasilctl -json` calls to get node stats. Unnecessary Lua code is also removed.

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/keys.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+'require fs';
+'require form';
+'require tools.widgets as widgets';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/keys.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+'require fs';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/peers.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+'require fs';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/peers.js
```diff
@@ -0,0 +1,33 @@
...
+'require form';
+'require tools.widgets as widgets';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/session_firewall.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+'require fs';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/session_firewall.js
```diff
@@ -0,0 +1,40 @@
...
+'require form';
+'require tools.widgets as widgets';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/settings.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+'require fs';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/settings.js
```diff
@@ -0,0 +1,59 @@
...
+'require form';
+'require tools.widgets as widgets';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/status.js
```diff
@@ -0,0 +1,113 @@
+'use strict';
+'require fs';
+'require form';
+'require tools.widgets as widgets';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/tunnel_routing.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+'require fs';
+'require form';
+'require tools.widgets as widgets';
```
jow-: You don't need this require

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/tunnel_routing.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+'require fs';
```
jow-: You don't need this require

jow-:
Apart from some redundant `require` statements, this looks quite good to me.

applications/luci-app-yggdrasil/root/usr/share/rpcd/acl.d/luci-app-yggdrasil.json
```diff
@@ -0,0 +1,10 @@
+{
+        "luci-app-yggdrasil": {
+                "description": "Grant access to LuCI app yggdrasil",
+                "read": {
```
jow-: Since `yggdrasilctl` provides commands to modify the system state (such as `addPeer`, `removePeer`, `setTunTap` etc.) we should change `read` to `write` here. There will be no functional change in the LuCI app due to this, but it'll be needed in the future for proper permission handling.

zhoreeq: Thanks for a review. Changed to write.

applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil/status.js
```diff
@@ -0,0 +1,96 @@
...
+	view.appendChild(info_title);
+	view.appendChild(self_info);
+	var peering_title = document.createElement("h3"); peering_title.innerText = _("Active peers");
+	view.appendChild(peering_title);
+
+	var peerings = document.createElement("table"); 
+	peerings.setAttribute("class", "table"); peerings.id = "yggdrasil-peerings";
+	var tr = document.createElement("tr");
+	tr.setAttribute("class", "tr table-titles");
+	["Endpoint", "Address", "Proto", "Uptime", "Rx / Tx"].forEach(function(t) {
```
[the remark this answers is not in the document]
jow-: Use `%1024mB'.format(bygtes)` instead.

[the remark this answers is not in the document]
jow-: Maybe use `'%t'.format(9052.876473808)` which would result in something like `2h 30m 52s`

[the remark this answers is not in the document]
wfleurant: i made these two format changes w/ @zhoreeq  https://github.com/zhoreeq/luci/pull/1

image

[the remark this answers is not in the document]
zhoreeq: It looks nice now :+1:

applications/luci-app-yggdrasil/luasrc/controller/yggdrasil.lua
```diff
@@ -0,0 +1,49 @@
...
+
+yggdrasil  = require("yggdrasil/admin")
+dkjson = require("dkjson")
+
+function index()
+	if not nixio.fs.access("/etc/config/yggdrasil") then
+		return
+	end
+
+	entry({"admin", "services", "yggdrasil"}, firstchild(), "yggdrasil").dependent = true
```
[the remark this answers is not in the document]
zhoreeq: :+1: That will make more sense. Should I change it to `{"admin", "network", "yggdrasil"}`? @wfleurant @aparcar @jow- @feckert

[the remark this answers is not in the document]
zhoreeq: Since there are no objections, I guess that means "yes"

aparcar:
I'm not to familiar with the ACL, looks like you allow "exec" but not "stat", but at some point you use the latter. Is that a problem?

Also, the only lua bits are for creating a menu entry, isn't there a proper way to do so in JS?

zhoreeq:
> I'm not to familiar with the ACL, looks like you allow "exec" but not "stat", but at some point you use the latter. Is that a problem?

Doesn't seem to. I copied it from that samba app example you've provided, and it works without problems.

> Also, the only lua bits are for creating a menu entry, isn't there a proper way to do so in JS?

No idea. Maybe @jow- can tell?

jow-:
The `stat` permissions are implicitly granted by `luci-base` which permits stat on the entire fs.

Menu declarations require the Lua controller yet. Support for JSON-only menu declarations will follow later this week hopefully.

aparcar:
Okay so I'd merge this now and add a patch later for the json menu?

jow-:
Merged, thanks!

Issue #465, opened 2015-09-04
luci-app-firewall missing "port range" / usability issue


plntyk:
(observed on IRC)

It is not "obvious" for new users that the 
"External port" field does support multiple ports or port ranges.

possible solution: 
expand field description to "External port, list of ports or port-range"

PS: the overview differentiates between "port" and "ports"

jow-:
The description states `Match incoming traffic directed at the given destination port or port range on this host` now

PR #3362, opened 2019-11-25
luci-app-firewall: update form field titles to indicate port ranges are ok

Strykar:
The Source | Internal | External `Port` form's field titles under Luci `Firewall` -> `Port Forwards` do not make it clear (_unlike in their description_) that port ranges (_instead of just a single port_) are accepted.
Closes https://github.com/openwrt/luci/issues/465

Signed-off-by: Avinash Duduskar <strykar@hotmail.com>

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/forwards.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/forwards.js
@@ -196,7 +196,7 @@ return L.view.extend({
 			));
 		});
 
-		o = s.taboption('advanced', form.Value, 'src_port', _('Source port'),
+		o = s.taboption('advanced', form.Value, 'src_port', _('Source port(s)'),
 			_('Only match incoming traffic originating from the given source port or port range on the client host'));
 		o.modalonly = true;
 		o.rmempty = true;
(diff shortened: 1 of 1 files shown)


jow-:
NACK. This change will invalidate existing translations.

Strykar:
> NACK. This change will invalidate existing translations.

Fair enough, I actually thought of that for a moment as I did some of the Hindi translations on Weblate.
@jow- Would `Ports` instead of `Port(s)` work?

feckert:
I think this change will also invalidate existing translations.

Strykar:
> I think this change will also invalidate existing translations.

@feckert So what's the right approach here? A (non-) solution would be to _not_ touch the (_misleading_) title at all because it will break the resulting (_also misleading_) translations.

@nijel @WeblateOrg @aparcar There ought to be a way to trigger an appropriate _notice to fix_ in the translations /  to the translators?

jow-:
Why is the title misleading? The field description clearly states that port ranges may be specified.

jow-:
I decided against merging this PR. Changing the option titles to plural would imply that multiple non-consecutive ports may be specified which isn't the case. The underlying `fw3` uci options accept exactly *one port* or one *consecutive port range*.

PR #3391, opened 2019-12-04
luci-mod-network: Add description field to VLANs (trivial)

TobleMiner:
Having a description attached to a VLAN makes identification a lot easier when
dealing with many VLANs

Signed-off-by: Tobias Schramm <tobleminer@gmail.com>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
@@ -334,6 +334,8 @@ return L.view.extend({
 				return (value || uci.get('network', section_id, 'vlan'));
 			};
 
+			s.option(form.Value, 'description', _('Description'));
+
 			for (var j = 0; Array.isArray(topology.ports) && j < topology.ports.length; j++) {
 				var portspec = topology.ports[j],
 				    portstate = Array.isArray(topology.portstate) ? topology.portstate[portspec.num] : null;


modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
```diff
@@ -358,6 +358,11 @@ return L.view.extend({
 				port_opts.push(o);
 			}
 
+			o = s.option(form.Value, 'description', 'Description');
```
jow-: Wouldn't that better be placed next to the VLAN ID and not after the port list?
Also the `Description` label is not marked as translatable.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
```diff
@@ -358,6 +358,11 @@ return L.view.extend({
...
+			o.write = function(section_id, value) {
```
jow-: Why is the `o.write()` override needed?

TobleMiner: Ah, ok. Didn't know it was not required. I'll remove it then

modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
TobleMiner: I'm somewhat torn regarding placement. Having it next to the VLAN ID makes it more visible without scrolling but takes away horizontal space from the actual config area.
Looking at other config interfaces it seems that the description/name is usually placed in the column after the VLAN ID so I'll do that, too.
Translation will be fixed, too

TobleMiner:
@jow- Changes implemented. Should be ok now

TobleMiner:
@jow- Hey, any news on this PR? Think it should be ready for merging

feckert:
From my side that´s LGTM.
All requested changes have been implemented.

PR #3388, opened 2019-12-03
[RFC] treewide: remove useless ucitrack definitions

ysc3839:
#3381

PR #3400, opened 2019-12-10
luci.mk: move indexcache delete into postinst

feckert:
Almost every uci-defaults script for ucitrack execute after the uci
commands the following line '/tmp/luci-indexcache'.
So that we don't always write the same thing, we now do this in
postinst. This was the result from discussion in #3388 .
We still need to remove this line from every uci-defualts script.


ysc3839:
It's #3388.

PR #3402, opened 2019-12-10
luci-app-https-dns-proxy: package name change, support for RFC8484 binary

stangri:
This luci app supports the new RFC8484-compatible package and more resolvers.

Depends on https://github.com/openwrt/packages/pull/10754.


tapper82:
luci-app-https-dns-proxy: Will not build in master. I think it's looking for the old package still.

stangri:
Logs or it didn't happen.

hnyman:
This has not been merged as there are conflicts.

applications/luci-app-https-dns-proxy/po/templates/https-dns-proxy.pot
```diff
@@ -0,0 +1,165 @@
...
+#: luci-app-https-dns-proxy/luasrc/view/https-dns-proxy/buttons.htm:45
+msgid "Stop"
+msgstr ""
+
+#: luci-app-https-dns-proxy/luasrc/model/cbi/https-dns-proxy.lua:37
+msgid "Stopped"
+msgstr ""
+
+#: luci-app-https-dns-proxy/luasrc/model/cbi/https-dns-proxy.lua:22
+msgid "Uknown Provider"
```
ldir-EDB0: typo - should be 'Unknown Provider"

stangri:
@hnyman how can I fix this conflicting files issue? I think they've been merged after I created this PR. The easiest way for me would be to create the new PR, please advise.

hnyman:
Easiest might be to rebase and force-push this branch. 

git checkout feature_branch
git rebase master
<fix conflicts>
<mark resolved>
git push -f 

I think that you are actually destroying/removing all existing translations (*.po), instead of moving/renaming the files to reflect the new package name. The error is due that you are trying to delete the old contents of those files instead of the current contents. (the contents have been changed in a translation sync, or by additional weblate translations in the meanwhile, I guess)

But a better (=more polite) approach would be to rename/move the .po files to be under the renamed package structure. Then the already done translations would not be destroyed. I guess that most of the strings are still roughly the same.

hnyman:
But naturally, also abandoning this PR and just creating a new PR is also simple, as you say.

stangri:
> But a better (=more polite) approach would be to rename/move the .po files to be under the renamed package structure. Then the already done translations would not be destroyed. I guess that most of the strings are still roughly the same.

Good call on renaming the existing translations! I will start anew.

PR #3406, opened 2019-12-12
luci-app-https_dns_proxy:add rubyfish.cn DOH servise in china

kkjoy:
luci-app-https_dns_proxy:add rubyfish.cn DOH servise in china


hnyman:
@stangri

stangri:
I don't get it, I thought I deleted the old Luci app in the last pr.

stangri:
This should be closed, I'm not sure how this even made it thru, as the `luci-app-https_dns_proxy` package was removed when I replaced it with `luci-app-https-dns-proxy` which uses RFC8484 instead of the JSON API (like in the old package with underscores).

kkjoy:
OK got it

PR #3412, opened 2019-12-14
19.07: ocserv: fix the public key ID calculation

nmav:
New versions of certtool change its output format and thus break
the previous parsing. This update works with certtool 3.5.x or later.
It also restricts the output to valid values for openconnect 7.x or
later for simplification.

Signed-off-by: Nikos Mavrogiannopoulos <nmav@gnutls.org>


jow-:
Merged in 391258f4f via cherry pick instead - thanks!

PR #3425, opened 2019-12-20
luci-app-fwknopd: fix typos

urbalazs:
Fixed typos:

- Base 64 -> Base64
- Use ANY for any source ip -> Use ANY for any source IP
- defaults to 120 seconds -> Defaults to 120 seconds.

### DIFF
--- a/applications/luci-app-fwknopd/luasrc/model/cbi/fwknopd.lua
+++ b/applications/luci-app-fwknopd/luasrc/model/cbi/fwknopd.lua
@@ -14,22 +14,22 @@ qr = s:option(DummyValue, "note0", "dummy")
 qr.tmp = tmp
 qr.template = "fwknopd-qr"
 qr:depends("uci_enabled", "1")
-s:option(Value, "SOURCE", "SOURCE", translate("Use ANY for any source ip"))
+s:option(Value, "SOURCE", "SOURCE", translate("Use ANY for any source IP"))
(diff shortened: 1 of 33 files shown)


applications/luci-app-fwknopd/luasrc/model/cbi/fwknopd.lua
```diff
@@ -14,22 +14,22 @@ qr = s:option(DummyValue, "note0", "dummy")
...
+l1:value("Base64 key", "Base64 key")
 k3 = s:option(Value, "HMAC_KEY", "HMAC_KEY", "The hmac key")
 k3:depends("hkeytype", "Normal Key")
 k4 = s:option(Value, "HMAC_KEY_BASE64", "HMAC_KEY_BASE64", translate("The base64 hmac key"))
-k4:depends("hkeytype", "Base 64 key")
+k4:depends("hkeytype", "Base64 key")
 l2 = s:option(ListValue, "hkeytype", "HMAC Key type")
 l2:value("Normal Key", "Normal Key")
-l2:value("Base 64 key", "Base 64 key")
+l2:value("Base64 key", "Base64 key")
```
feckert: I think this is wrong the first argument is the value oft the option and should not get changed. 
And the second argument is the translation.
It should look more like this

`l2:value("Base 64 key", translate("Base64 key"))`

urbalazs: I changed all keys, so this should be fine. I also checked other repositories (openwrt, packages, etc.) and I didn't find more occurrences.
But feel free to fix me, if I am wrong!

PR #3444, opened 2019-12-29
luci-mod-network: add certificate subj validation

swg0101:
This PR is tied to changes being made via openwrt/openwrt#2654, and it
allows users to provide a certificate constraint when connecting as a
STA to an AP using Enterprise modes. The constraints can either be
provided via the certificate's X509 subject using a substring (e.g.
/CN=wifi.mycompany.com) or subject alternate name values (e.g.
DNS:wifi.mycompany.com), although both can be combined for a stricter
validation criteria.


swg0101:
@feckert Dependency is now in master, thanks!

feckert:
Dependent pull request was merged into the master https://github.com/openwrt/openwrt/commit/22b07ff73e0b3429b36f75694a082a68a4fdb013
Thanks

PR #3457, opened 2020-01-03
i18n: fix and sync translations

musashino205:
- luci-app-ddns: fix some errors in pl/ddns.po
- luci-app-ocserv: fix errors in hu/ocserv.po
- luci-base: remove duplicated translations from Hungarian and Marathi
- i18n: sync translations

### DIFF
--- a/applications/luci-app-adblock/po/zh-tw/adblock.po
+++ b/applications/luci-app-adblock/po/zh-tw/adblock.po
@@ -93,7 +93,9 @@ msgstr "未揾到拉黑檔案清單!"
 msgid ""
 "Choose 'none' to disable automatic startups, 'timed' to use a classic "
 "timeout (default 30 sec.) or select another trigger interface."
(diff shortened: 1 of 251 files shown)


hnyman:
Thanks.
Some of those format errors have probably beed caused by me when I fixed the 50+ conflicts in the weblate commits two days ago :-(

weblate seems rather error-prone: the last translation oriented commits in LuCI have been fixes & weblate merges from @jow-  yesterday and still weblate has complained about conflicts since yesterday (although there have been nothing else than those weblate-oriented commits.

musashino205:
Oh, I see... Thank you for your big works...

hnyman:
I merged this. Likely that will again cause new breakage at weblate translation, but as weblate has not progressed in updating its sources since yesterday, it sounds stupid that that main project would stay hostage to the external translation engine.

Hopefully @aparcar or @jow- has energy to fix the old and the new breakage at weblate. I tried last time and apparently failed to do 100% correct work... 

It seems really strange that weblate has been at an error status since yesterday, already when there were no other commits that jow's translation fixes & merges of weblate's own commits.

PR #3458, opened 2020-01-03
luci-app-vnstat2: add application

janh:
This adds a LuCI application for vnStat 2.x.

Depends-on: https://github.com/openwrt/packages/pull/10935

### DIFF
--- a/applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/config.js
+++ b/applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/config.js
@@ -1,0 +1,109 @@
+// This is free software, licensed under the Apache License, Version 2.0
+
+'use strict';
+'require fs';
+'require ui';
+'require uci';
+'require form';
+'require tools.widgets as widgets';
+
+return L.view.extend({
+	handleDeleteModal: function(m, iface, ev) {
+		L.showModal(_('Delete interface <em>%h</em>').format(iface), [
+			E('p', _('The interface will be removed from the database permanently. This cannot be undone.')),
+			E('div', { 'class': 'right' }, [
+				E('div', {
+					'class': 'btn',
+					'click': L.hideModal
+				}, _('Cancel')),
+				' ',
+				E('div', {
+					'class': 'btn cbi-button-negative',
+					'click': ui.createHandlerFn(this, 'handleDelete', m, iface)
+				}, _('Delete'))
+			])
+		]);
+	},
+
+	handleDelete: function(m, iface, ev) {
+		return fs.exec('/usr/bin/vnstat', ['--remove', '-i', iface, '--force'])
+			.then(L.bind(m.render, m))
+			.catch(function(e) {
+				ui.addNotification(null, E('p', e.message));
+			})
+			.finally(L.hideModal);
+	},
+
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('vnstat', _('vnStat'), _('vnStat is a network traffic monitor for Linux that keeps a log of network traffic for the selected interface(s).'));
+
+		s = m.section(form.TypedSection, 'vnstat', _('Interfaces'));
+		s.anonymous = true;
+		s.addremove = false;
+
+		o = s.option(widgets.DeviceSelect, 'interface', _('Monitor interfaces'), _('The selected interfaces are automatically added to the vnStat database upon startup.'));
+		o.rmempty = true;
+		o.multiple = true;
+		o.noaliases = true;
+		o.nobridges = false;
+		o.noinactive = false;
+		o.nocreate = false;
+
+
+		o = s.option(form.DummyValue, '_database');
+
+		o.load = function(section_id) {
+			return fs.exec('/usr/bin/vnstat', ['--json', 'f', '1']).then(L.bind(function(result) {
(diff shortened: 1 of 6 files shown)


janh:
Removed unnecessary and wrong ucitrack definition.

applications/luci-app-vnstat2/luasrc/view/vnstat2/graphs.htm
```diff
@@ -0,0 +1,55 @@
...
+<%-
+
+local util = require "luci.util"
+local json = require "luci.jsonc"
+
+
+local ifaces = {}
+
+local data = util.exec("vnstat --json f 1 2>/dev/null")
+if #data > 0 and data:sub(1,1) == "{" then
```
feckert: I think this is not necessary you could check the return value of json.parse directly.
https://github.com/openwrt/luci/blob/33597c79ac5d7ee917f8452cc9c26fd546c2cf1f/applications/luci-app-attendedsysupgrade/luasrc/view/attendedsysupgrade.htm#L12-L15

applications/luci-app-vnstat2/root/usr/share/rpcd/acl.d/luci-app-vnstat2.json
```diff
@@ -0,0 +1,11 @@
+{
+	"luci-app-vnstat2": {
+		"description": "Grant access to LuCI app vnstat2",
+		"write": {
+			"file": {
+				"/usr/bin/vnstat": [ "exec" ]
+			}
```
feckert: Do we not need also this line to write the
```
		"write": {
			"uci": [ "vnstat" ]
```

janh:
I updated the check of vnstat output and the rpcd acl configuration.

I'm still unsure if the rpcd acl configuration is correct. Looking at other applications with client-side implementation, it is not really clear to me how it is supposed to be set up. For example, when uci files need to be specified (as right now any application seems to have access anyway), and whether items that already are in the "write" section should also be in the "read" section.

jow-:
> For example, when uci files need to be specified (as right now any application seems to have access anyway)

Ideally yes. The `luci-base` package implicitely grants full access to many files but ideally packages should list everything they need.

> whether items that already are in the "write" section should also be in the "read" section.

This is not needed. Write always implies read.

janh:
Thank you for the explanation. I updated the ACL configuration.

applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/config.js
```diff
@@ -0,0 +1,114 @@
...
+				E('div', {
+					'class': 'btn cbi-button-negative',
+					'click': ui.createHandlerFn(this, 'handleDelete', m, iface)
+				}, _('Delete'))
+			])
+		]);
+	},
+
+	handleDelete: function(m, iface, ev) {
+		console.log('test');
```
feckert: Please remove this statement

janh:
Removed `console.log` call. Also added myself as package maintainer.

applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/config.js
```diff
@@ -0,0 +1,113 @@
...
+'require fs';
+'require ui';
+'require uci';
+'require form';
+'require tools.widgets as widgets';
+
+return L.view.extend({
+	handleDeleteModal: function(m, iface, ev) {
+		L.showModal(_('Delete interface <em>%h</em>').format(iface), [
+			E('p', 'The interface will be removed from the database permanently. This cannot be undone.'),
```
jow-: Can you wrap this message in `_(...)` to make it translatable?

applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/config.js
```diff
@@ -0,0 +1,113 @@
...
+		o.noinactive = false;
+		o.nocreate = false;
+
+
+		o = s.option(form.DummyValue, '_database');
+
+		o.load = function(section_id) {
+			return Promise.all([
+				fs.exec('/usr/bin/vnstat', ['--json', 'f', '1']),
+				uci.load('vnstat')
```
jow-: The `uci.load('vnstat')` should be redundant since `form.Map('vnstat', ...)` already takes care of loading it.

applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/config.js
```diff
@@ -0,0 +1,113 @@
...
+						'click': ui.createHandlerFn(view, 'handleDeleteModal', m, this.interfaces[i])
+					}, [ _('Delete…') ])
+				]);
+			}
+
+			cbi_update_table(table, rows, E('em', _('No unconfigured interfaces found in database.')));
+
+			return E([], [
+				E('h3', _('Unconfigured interfaces')),
+				E('div', { class: 'cbi-section-descr' },
```
jow-: Can you put the `class:` field in single quotes? Iirc an unquoted "class" causes problems in some environment since it is a keyword in later ECMAScript versions.

jow-:
Merged, thank you!

Issue #3477, opened 2020-01-07
translation: zh_Han{s,t} and zh-{cn,tw} mixture


aparcar:
Hi, we partly have the *old* and new format for Simplified Chinese and Traditional Chinese. Somebody who speaks the languages, please merge them to zh-{cn,tw} as the zh_Han{s,t} format is not compatible with the browser language detection.

Current cases are:
* luci-app-nextdns
* luci-app-vpn-policy-routing

jow-:
There's also `nb_NO` and `pt_BR` which conflict with `no` and `pt-br` respectively.

ElonH:
the script 

https://github.com/openwrt/luci/blob/master/build/i18n-add-language.sh

wasn't updated.

PR #3480, opened 2020-01-08
luci-app-nextdns: remove duplicated zh_Hans language

urbalazs:
Fixed issue with duplicated zh_Hans language (https://github.com/openwrt/luci/issues/3477)


aparcar:
Are both files exact duplicates? I though of merging instead of losing any translations

hnyman:
I think that the file was created yesterdya by @jow- in the treewide sync.
```

applications/luci-app-nextdns/po/zh-cn/zh_Hans/nextdns.po 	[new file with mode: 0644] 	blob
applications/luci-app-nextdns/po/zh-tw/nextdns.po 	[new file with mode: 0644] 	blob
applications/luci-app-nextdns/po/zh_Hans/nextdns.po 	[deleted file] 	blob | history
applications/luci-app-nextdns/po/zh_Hant/nextdns.po 	[deleted file] 	blob | history
applications/luci-app-omcproxy/po/zh-tw/omcproxy.po 	[new file with mode: 0644] 	blob
applications/luci-app-omcproxy/po/zh_Hant/omcproxy.po 	[deleted file] 	blob | history
```

But I have no clue that which of the Chinese language codes are correct for the future, and why weblate wants to rename/add our already existing languages.

PR #3482, opened 2020-01-08
luci-app-olsr: fix typo

urbalazs:
Fixed typos:

- Interface Mode is -> Interface mode is
- valid Modes are -> Valid modes are


feckert:
Thanks

PR #3488, opened 2020-01-09
luci-app-advanced-reboot: pluggable device files, highlight current partition

stangri:
Switching to a model where device specifics are in pluggable files.
Highlight current (instead of first) partition in the list.

### DIFF
--- a/applications/luci-app-advanced-reboot/Makefile
+++ b/applications/luci-app-advanced-reboot/Makefile
@@ -13,7 +13,7 @@ LUCI_DESCRIPTION:=Provides Web UI (found under System/Advanced Reboot) to reboot
 
 LUCI_DEPENDS:=+luci-compat +luci-mod-admin-full
 LUCI_PKGARCH:=all
-PKG_RELEASE:=47
+PKG_RELEASE:=51
 
 include ../../luci.mk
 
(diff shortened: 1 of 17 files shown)


hnyman:
Your translation sync looks faulty: 
it removes the "applications/" directory from the path in the location annotation rows.

I assume that you have run sync from inside the app directory, causing this.
The next global sync would return that directory path, so the current action would cause unnecessary git fluctuation. Please run the i18n-sync.pl from the LuCI repo root.

stangri:
Hannu, the repo where I have hooks to generate the pot-files is flat: https://github.com/stangri/openwrt_packages.

I can either set the hooks on the temp repo from which I send PRs or omit the pot-files completely (the latter would be my preference), please advise me how to proceed.

Also, I've noticed "1" and "2" got into the pot-file, I need to fix the view file to get rid of them.

Finally, is there any point in sending PRs for 18.06 and 19.07 for these changes? I'm concerned that if I ever need to do a bugfix and the release is behind master it's going to be double the work.

stangri:
@hnyman fixed both issues, I'll create a new PR. Please advise if I should send it to released branches.

hnyman:
Might be best to leave the pot file away from PR if the annotation path is wrong.
(There seem to be two actual new strings, all other changes are either line numberchanges or wrong path omissions)

If your code works in 19.07, at least is should/could be updated. 
18.06 starts to be deprecated now that 19.07.0 has been officially released. You might retire the fixes in the 18.06 at some point

stangri:
I added sed to a hook in the local repo, the paths will be correct. I'll close existing PRs and create new ones for master and 19.07.

PR #3509, opened 2020-01-15
luci-app-vpn-policy-routing: styling bugfix

stangri:
Fix for the orphaned text in the view-file.


jow-:
Merged, thanks @stangri  - btw. if you want a commit to be merged to 19.07 as well and if the version of the package is the same there, it is enough if you tell me to cherry-pick it as well.

PR #3510, opened 2020-01-15
[19.07] luci-app-vpn-policy-routing: styling bugfix

stangri:
Fix for the orphaned text in the view-file.


jow-:
Cherry-picked 5baeb64b4 instead.

PR #3511, opened 2020-01-15
luci-base: remove hardcoded cgi-bin path

Ansuel:
Currently LuCI can be loaded only when placed in the root of the server as the cgi-bin paths are hardcoded. Change the index.html to load the cgi-bin path from the current level.
Also add a new entry in the env Object to make the cgi_base path easily accessible.
This variable will be based on the position of /cgi-bin/luci.

@jow- Was easier than I tought. No extra config needed. 
Should I also change the comments? 

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/fs.js
+++ b/modules/luci-base/htdocs/luci-static/resources/fs.js
@@ -342,7 +342,7 @@ var FileSystem = L.Class.extend(/** @lends LuCI.fs.prototype */ {
 		var postdata = 'sessionid=%s&path=%s'
 			.format(encodeURIComponent(L.env.sessionid), encodeURIComponent(path));
 
-		return L.Request.post('/cgi-bin/cgi-download', postdata, {
+		return L.Request.post(L.env.cgi_base + '/cgi-download', postdata, {
 			headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
 		}).then(handleCgiIoReply);
 	},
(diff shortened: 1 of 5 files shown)


modules/luci-base/htdocs/luci-static/resources/luci.js
```diff
@@ -1192,6 +1192,8 @@
 			if (env.base_url == null)
 				this.error('InternalError', 'Cannot find url of luci.js');
 
+			env.cgi_base = env.scriptname.substring(0, env.scriptname.lastIndexOf('luci'));
```
jow-: I think you better should use `L.env.scriptname.replace(/\/[^\/]+$/, '')` here, to take the current program path minus the last path segment. This should result in `/cgi-bin` (minus the trailing slash)

modules/luci-base/htdocs/luci-static/resources/fs.js
```diff
@@ -342,7 +342,7 @@ var FileSystem = L.Class.extend(/** @lends LuCI.fs.prototype */ {
 		var postdata = 'sessionid=%s&path=%s'
 			.format(encodeURIComponent(L.env.sessionid), encodeURIComponent(path));
 
-		return L.Request.post('/cgi-bin/cgi-download', postdata, {
+		return L.Request.post(L.env.cgi_base + 'cgi-download', postdata, {
```
jow-: Can you add an explicit trailing slash here and in the other places?

`L.env.cgi-base + '/cgi-download'`

Ansuel:
@jow- I missed the flash.js script. Ready to merge.

jow-:
Merged, thanks @Ansuel !

PR #3514, opened 2020-01-16
luci-app-mosquitto: translations + option list fixes

karlp:
Marks all the strings as translateable, which had been completely skipped in the past
Also includes a fix to the list of tls versions.

further details are in each commit


feckert:
- Could you please change the commit message to **luci-app-mosquitto: \<your message\>**
- I think we should also generate the po files so this could get translated by weblate users into other languages

karlp:
Sure, I felt the luci- prefix was superfluous, but will change.  Have pushed with an updated pot file too.

feckert:
Thanks merged

PR #1480, opened 2017-12-19
luci-app-ser2net: created

TimelessNL:
This changeset will add support for ser2net in luci-app-ser2net.

_Note: This depends on https://github.com/openwrt/packages/pull/5302 to be accepted first._

Note2: This changeset is based on https://github.com/JiapengLi/OpenWrt-luci-app-ser2net/blob/master/openwrt-packages-add-luci-app-ser2net-support.patch

Issue #3378, opened 2019-12-01
[RFC]Require JavaScript for new apps?


aparcar:
I'm wondering if all new LuCI apps should be written in JavaScript and LUA is from now on considered legacy and not accepted for new apps anymore.
The documentation seem straight forward enough to easily convert existing LUA code to JS.

Please share your thoughts.

hnyman:
I feel that it might be too strict to "require" JavaScript, but it might be "preferred".

Personally I think that it does not make much difference in simple uci config apps without much calculation, logic or interactive status screens. The goal of JS is to move the CPU intensive processing to the client side, but many apps are rather simple and just used for basic uci config.

aparcar:
Maybe I'm wrong but I though another advantage is to get rid of the lua
runtime and save some storage. Eventually lowering also the languages to
know for maintenance.

jow-:
Yes, the goal is to get rid of Lua completely for the LuCI base.

aparcar:
Should we pin this?

howard0su:
Where is migration guide to migrate from LUA-based to JS-based project?

oxr463:
So the JavaScript apps run on the client-side in the browser vs Lua running via cgi-bin on the device, right?

feckert:
Unfortunately, I saw this too late and did not think on this, but I have merged this https://github.com/openwrt/luci/pull/4664 and it is written in lua.
I have now another [P/R](https://github.com/openwrt/luci/pull/4665) which is also lua and which I also just wanted to merge.
So that I don't make the same mistake in the review and tell people right away that we want javascript, we should communicate this clearly and be in agreement.

I can not revert this https://github.com/openwrt/luci/pull/4665 but I can request to write this https://github.com/openwrt/luci/pull/4664 in javascript or should I merge them?

PolynomialDivision:
> write this #4664 in javascript or should I merge them?

I can rewrite that in javascript, but can u maybe link to a good example?
Maybe luci-mod-status is a good example?

aparcar:
@PolynomialDivision @howard0su 

I'm afraid at this point the best documentation for migrating from Lua is the [extensive LuCI JS documentation](http://openwrt.github.io/luci/jsapi/index.html) in combination with pull requests that did the migration in the past, there are multiple.

@PolynomialDivision if you go through the process and find the time it would be great to share the experience and write it down in a document.

PolynomialDivision:
[...]

PR #3541, opened 2020-01-20
luci-app-ser2net: new module supporting ser2net.

michyprima:
Sparked and loosely inspired by #1480
This implements full support to the new uci config format for ser2net in luci.
Signed-off-by: Michele Primavera <primavera@elmod.it>

### DIFF
--- a/applications/luci-app-ser2net/Makefile
+++ b/applications/luci-app-ser2net/Makefile
@@ -1,0 +1,16 @@
+#
+# Copyright (C) 2020 Michele Primavera
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI Support for ser2net
+LUCI_DEPENDS:=+ser2net
+
+PKG_LICENSE:=Apache-2.0
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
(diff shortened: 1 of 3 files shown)


feckert:
As discussed in https://github.com/openwrt/luci/issues/3378 I ask you to convert the implementation from LUA to Javascript? This shouldn't be too difficult as it is only use uci handling. There are already some application converted.
https://github.com/openwrt/luci/tree/master/applications/luci-app-yggdrasil/htdocs/luci-static/resources/view/yggdrasil

michyprima:
@feckert Sorry, didn't see the pinned issue. I converted the app and it's working, the only thing I'm missing is how do I trigger /etc/init.d/ser2net restart after saving?

michyprima:
it does not.
If I do:
> uci set ser2net.global.enabled='0' && uci commit
the process does not stop

jow-:
You also need to call `/sbin/reload_config` afterwards or `ubus call service event '{ "type": "config.change", "data": { "package": "ser2net" } }' (which is what LuCI does on Save & Apply.

jow-:
You also need to call `/sbin/reload_config` afterwards or `ubus call service event '{ "type": "config.change", "data": { "package": "ser2net" } }' (which is what LuCI does on Save & Apply).

michyprima:
It is still not working, I believe the init.d script is incomplete but I can't figure out what it is missing

PR #3544, opened 2020-01-21
Openwrt 19.07 with Neural Network WAF

semseo4you:
I find it very interesting to integrate with advanced firewall systems based on deep neural networks such as:
https://github.com/BBVA/waf-brain
How about?


feckert:
Please open an issue this is a pullrequest

PR #3559, opened 2020-01-24
Revert "Translations update from Weblate"

buicaocuong:
Reverts openwrt/luci#3556


hnyman:
Why do you propose to revert the translations of that PR ?

PR #3578, opened 2020-01-27
[19.07] luci-app-smbd: rename to luci-app-ksmbd

Andy2244:
* follow upstream name change to ksmbd
* remove old artifacts from "applications\luci-app-cifsd\po\"

Depends on: openwrt/packages#11143


neheb:
The other PRs have been merged.

PR #3595, opened 2020-02-01
luci-app-openvpn: add editing field for .up and .down scripts

walterav1984:
These changes allow a seperate 'ovpnprofile.up/ovpnprofile.down' file/script besides the existing 'ovpnprofile.auth' file and the existing 'ovpnprofile.ovpn'&'ovpnprofile.auth' edit fields. 

Review requested:
It creates 'ovpnprofile.up/ovpnprofile.down' which are executable, because of a quick/dirty hack without further understanding of the lua languange.

Please backport these simple changes to 19.07.x because its 'luci-app-openvpn' can than be configured in very advanced ways just from the webinterface.

Signed-off-by: Walter Sonius <walterav1984@gmail.com>

### DIFF
--- a/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-file.lua
+++ b/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-file.lua
@@ -6,6 +6,8 @@ local util      = require("luci.util")
 local uci       = require("luci.model.uci").cursor()
 local cfg_file  = uci:get("openvpn", arg[1], "config")
 local auth_file = cfg_file:match("(.+)%..+").. ".auth"
+local down_file = cfg_file:match("(.+)%..+").. ".down"
+local up_file   = cfg_file:match("(.+)%..+").. ".up"
 
 local m = Map("openvpn")
 
@@ -76,6 +78,48 @@ function file.remove(self, section, value)
 end
 
 function s.handle(self, state, data2)
+	return true
+end
+
+s = f:section(SimpleSection, nil, translatef("Section to add an optional 'up' file which typically add routes to the tunnel by script (%s)", up_file))
+file = s:option(TextValue, "data3")
+file.datatype = "string"
+file.rows = 20
+
+function file.cfgvalue()
+        return fs.readfile(up_file) or ""
+end
+
+function file.write(self, section, data3)
+        return fs.writefile(up_file, util.trim(data3:gsub("\r\n", "\n")) .. "\n")
+end
+
+function file.remove(self, section, value)
+        return fs.writefile(up_file, "")
+end
+
+function s.handle(self, state, data3)
+	return true
+end
+
+s = f:section(SimpleSection, nil, translatef("Section to add an optional 'down' file which typically removes routes to the tunnel by script (%s)", down_file))
+file = s:option(TextValue, "data4")
+file.datatype = "string"
+file.rows = 20
+
+function file.cfgvalue()
+	return fs.readfile(down_file) or ""
+end
+
+function file.write(self, section, data4)
+	return fs.writefile(down_file, util.trim(data4:gsub("\r\n", "\n")) .. "\n")
+end
+
+function file.remove(self, section, value)
+	return fs.writefile(down_file, "")
+end
+
+function s.handle(self, state, data4)
 	return true
 end
 
--- a/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-file.lua
+++ b/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-file.lua
@@ -91,7 +91,8 @@ function file.cfgvalue()
 end
 
 function file.write(self, section, data3)
-        return fs.writefile(up_file, util.trim(data3:gsub("\r\n", "\n")) .. "\n")
+        fs.writefile(up_file, util.trim(data3:gsub("\r\n", "\n")) .. "\n")
+        return fs.chmod( up_file, 755 )
 end
 
 function file.remove(self, section, value)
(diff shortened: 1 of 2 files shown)


feckert:
I have already submitted a similar patch functionality https://github.com/openwrt/openwrt/pull/1596
I think this is more generic.
This is pending upstream for a core developer reviewer.

walterav1984:
That https://github.com/openwrt/openwrt/pull/1596 PR is looking more advanced, but not 'luci' exposed?

Although this PR is incomplete when completed, it might add alot of functionality in a very similar existing way already done for .auth files. 

Does your hotplug mechanism support multiple/different up/down scripts for each seperate openvpn account/process at the same time? If yours will get merge, can it be exposed into the luci-app-openvpn for easy editing?

feckert:
> Does your hotplug mechanism support multiple/different up/down scripts for each seperate openvpn account/process at the same time? If yours will get merge, can it be exposed into the luci-app-openvpn for easy editing?

Yes that is possible but for now only **up** and **down** will use the OpenWrt hotplug in this pullrequest.
You could move your special script into **/etc/hotplug.d/openvpn** and this get called on openvpn up down events. That script place should be used for other package in OpenWrt that want to get informed on openvpn events.
For example:
- Send a notifification that some has connected
- Enable an LED if an connection is up or down

Or you could paste your script into **/etc/openvpn.user** this is also called but does not be related to other packages this file belongs to the openvpn package and should be edit by the user for special cases.

Another possibility is if you add your own script or command to the config option up/down, then the OpenWrt hotplug.d subsystem is not called only your configured script or command are executed.

walterav1984:
Hi @feckert your PR was recently added 8fe9940db66517679f09fa1d2f6f79229b6a8361 into openwrt instead of luci. however it's still unclear for me if you can have seperate up/down scripts for each vpn account and if the up/down scripts can be edited by l(uci) webinterface? 

If so I may close this PR?

feckert:
My solution is a generic solution and so was added to the openwrt main repository
You could add your up/down scripts into **/etc/hotplug.d/openvpn**.
On each up/down event the scripts are called.
In this callback scripts you could decide on which instance you want to react.

I have also added the possibility that on each up/down event the file [/etc/openvpn.user](https://github.com/openwrt/openwrt/blob/master/package/network/services/openvpn/files/etc/openvpn.user) is [called](https://github.com/openwrt/openwrt/blob/master/package/network/services/openvpn/files/etc/hotplug.d/openvpn/01-user). This file is also saved during firmware upgrade.

The next step is to add a possibility for editing this file in the LuCI.

jow-:
Note that the `01-user` script is currently called with `env -i` so it won't have access to the environment variables provided by the OpenVPN daemon which might or might not be an issue (@feckert).

feckert:
@jow- Due to the change in openvpn, we also have to remove the up and down from the option list. Since that doesn't make sense anymore if I configure openvpn via uci in the LuCI.

> Note that the 01-user script is currently called with env -i so it won't have access to the environment variables provided by the OpenVPN daemon which might or might not be an issue (@feckert).

If the hotplug scripts replace the up and down then we should not delete the env with `-i`. Then we get all call option from openvpn. This can be an advantage or a disadvantage when we think about security. Any script in openvpn can read the data and do something with it. But since openwrt is a single user system I think this is negligible.

Independently of this I would say we have to make it possible for the user to edit the file openvpn.user in the LuCI. This is also done by the mwan3.

walterav1984:
Now I understand this is a more generic way of making "openwrt" itself aware of the up/down status of a VPN connection and let the system respond to it instead of leaving it to "openvpn daemon" which may need extra user privileges on some OS to perform these tasks itself(script_security).

> ... environment variables provided by the OpenVPN daemon ...

These variables are essential in setup since every openvpn-profile may receive different settings from each openvpnserver which each different openvpnprofile.up/down scripts individually respond to.

I cannot say what is the "best/secure" way to setup the system but doing it by openvpn deamon is part of openvpn its feature set and relative easy to implement for me with (l)uci via this current PR.

The /etc/openvpn.user you suggest for these settings does it behave like /etc/openvpn/openvpn-profile-x.user and /etc/opevpn/openvpn-profile-y.user like the different .conf/.auth profiles right now?

feckert:
> The /etc/openvpn.user you suggest for these settings does it behave like /etc/openvpn/openvpn-profile-x.user and /etc/opevpn/openvpn-profile-y.user like the different .conf/.auth profiles right now and has to be made (yet) configurable by (l)uci which is still missing?

You could move this scripts **/etc/openvpn/openvpn-profile-x.user or /etc/openvpn/openvpn-profile-y.user** to /etc/hotplug.d/openvpn. On each up/down event alle files in this directory are called/executed. I think the /etc/openvpn.user is a quick and dirty possibility to test your script. But for now we get no openvpn env in the openvpn.user script. I test this with a logger command.

walterav1984:
> > But for now we get no openvpn env in the openvpn.user script. I test this with a logger command.

I guess this is fixed by: https://github.com/jollaman999/openwrt/commit/57664a1c9fd2dfd9045005b878c078a94889a83a

How to expose those user/up/down scripts via luci, should that be part of the openvpn package or some sort of "luci-hotplug-userscripts" example package?

feckert:
@walterav1984 From my point of view, the script should be stored under directory /etc/hotplug.d/openvpn.
All scripts that are stored there are called on an up/down event off an openvpn tunnel.
Maybe we should put a 00-user script that calls the editable /etc/openvpn.user script.
This can then be changed via the LuCI.
Have a look at this in mwan3.
https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/hotplug.d/iface/16-mwan3-user
and 
https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/mwan3.user

Issue #3542, opened 2020-01-21
luci-theme-material: small visual errors


castillofrancodamian:
**Giant buttons**

![3](https://user-images.githubusercontent.com/47670442/72770751-901f4100-3bdd-11ea-9b4e-0fa94b13d817.png)
![9](https://user-images.githubusercontent.com/47670442/72770759-97dee580-3bdd-11ea-8df4-f8482bed3b5a.png)

**Logout symbol attached to the text, not as in English that is separated**

![1](https://user-images.githubusercontent.com/47670442/72770851-dbd1ea80-3bdd-11ea-9b0d-ab1ecd39ed10.png)

**Misaligned buttons**

![2](https://user-images.githubusercontent.com/47670442/72770917-0e7be300-3bde-11ea-8aa9-5db44383887c.png)

**Text attached to the table**

![4](https://user-images.githubusercontent.com/47670442/72770957-2fdccf00-3bde-11ea-9790-fa840f716d0a.png)

**And a lot of white space**

![8](https://user-images.githubusercontent.com/47670442/72770992-497e1680-3bde-11ea-86e9-19d4553a4047.png)
![5](https://user-images.githubusercontent.com/47670442/72770993-4a16ad00-3bde-11ea-913d-63165445ed79.png)
![6](https://user-images.githubusercontent.com/47670442/72770995-4a16ad00-3bde-11ea-819d-6b7540df98da.png)
![7](https://user-images.githubusercontent.com/47670442/72770997-4a16ad00-3bde-11ea-9efe-814bec0345b5.png)

jow-:
@LuttyYang

castillofrancodamian:
I don't think it's going to respond. I see there is no activity on your GitHub profile.

dannil:
@stokito the only present issue seems to be "Text attached to the table", the rest is already fixed on the master branch. 

Regarding the white space in modals (= dialogs that gray out the background), I've assumed that's an intentional design decision of the theme since all modals is styled to always take up 90% of the width of the viewport. In the Bootstrap theme this isn't really noticeable since there the width is a fixed 900px. I assume that @LuttyYang is not active here anymore so getting an answer to that question will probably not happen, but my guess is that it's intentional since the Material theme is designed around the principle of taking up all available horizontal space.

stokito:
maybe then the issue can be closed, right?

dannil:
Yes, I would say so.

PR #3603, opened 2020-02-05
luci-app-battstatus

arrmo:
For battery powered routers (currently only one, but easy to add more!), to show battery status. Screen shots attached. Only visible if router is a supported model.

Capture-bootstrap
Capture-material
Capture-openwrt


feckert:
Thanks for the pullrequest.
It's a cool feature.
Never the less I would suggest to abstract the queries in an **ubus** call.
That has a lot of advantages.

-  The HTML page becomes more clearer, because no cryptic commands are executed in the template.

- And if other components in the system want to have this information, they would have to implement the same commands again `duplicate code`. If we have an ubus call, other components in the system could use this too for example prometheus/grafana monitoring stack.

An additional I am not sure if we want to have this polling in every page. Since only one system currently uses this feature the other 99.9% of the device does not have this. But maybe @jow could say more.

arrmo:
Thanks for the comments! A couple thoughts,
1) "I would suggest to abstract the queries in an ubus call" - likely not an issue, but not quite sure what you mean. LOL! Can you clarify?
2) "I am not sure if we want to have this polling in every page" - completely understand where you're coming from. I initially thought about making this a separate page, but as it's such a key item for a battery powered device, I concluded that it should be at the top of the / any page. But by all means disagree! Perhaps I could set the polling up to only happen for a supported device? Not sure about this, but I can dig more if desired.

Thanks again.

feckert:
>  1. "I would suggest to abstract the queries in an ubus call" - likely not an issue, but not quite sure what you mean. LOL! Can you clarify?

I mean that this whole block

https://github.com/openwrt/luci/pull/3603/files#diff-270c8e9816cf971df35214112e003e15R13-R18

should be implemented in a separate ubus call. Look at this, for example.

https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns

The name could be hwmon.bat and the function could be status.
If we do this that way, then other packages could also benefit from this information.
For example:
-Polling thei info with another script ands end a notification to a backend if we have not enough power anymore 
- Watch the battery power with prometheus grafana checkmk or collectd

only some examples

castillofrancodamian:
@arrmo Hi, could you correct [this](https://github.com/openwrt/luci/issues/3542)?

arrmo:
> > 1. "I would suggest to abstract the queries in an ubus call" - likely not an issue, but not quite sure what you mean. LOL! Can you clarify?
> 
> I mean that this block:
> 
> https://github.com/openwrt/luci/pull/3603/files#diff-270c8e9816cf971df35214112e003e15R13-R18
> 
> This should be implemented in a separate **ubus** call.
> 
> Look at this, for example:
> 
> https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns
> 
> The name could be **hwmon.bat** for example and the function could that get called should be named **status**. If we do this that way, then other packages could also benefit from this information.
> 
> For example:
> 
> * Polling the info with another script ands end a notification to a backend if we have not enough power anymore
> * Watch the battery power with prometheus grafana checkmk or collectd

I like this idea - thanks! Let me give it a go. BTW, where does the resulting executable show up ... is it in (in the case you note), /usr/libexec/rpcd/luci.ddns?

ysc3839:
I think this should be implemented in `luci-app-statistics` and then provide "show statistics on top bar" function.

ysc3839:
And it would be great to implement as `/sys/class/power_supply`, so that collectd is supported out-of-box.
https://collectd.org/wiki/index.php/Plugin:Battery

arrmo:
> I think this should be implemented in `luci-app-statistics` and then provide "show statistics on top bar" function. Because people may also want to see temperature of their devices.

Hi,

That could be an alternative of course, no issue here - but likely a bit more extensive than this ... agreed? And temperature - not currently supported, at least as far as I know.

Thanks!

arrmo:
> And it would be great to implement as `/sys/class/power_supply`, so that collectd is supported out-of-box.
> https://collectd.org/wiki/index.php/Plugin:Battery

Yes, agreed - but I don't know how to turn those i2c commands in to `/sys/class/power_supply` ... do you? Open to help, but that's past my pay grade :smile:

castillofrancodamian:
> > @arrmo Hi, could you correct [this](https://github.com/openwrt/luci/issues/3542)?
> 
> Sorry, to understand - pull in that PR to this one? Not sure I should be doing that, rather let that PR capture the updates for all apps ... right?

I ask if you could fix this error, not correct it in this Pull Request, because I see that you can modify `luci-theme-material`.

arrmo:
> I ask if you could fix this error, not correct it in this Pull Request, because I see that you can modify `luci-theme-material`.

OK, got your point now. Let me take care of this PR (and the other I have open), then I can try to take a look at that one. Thanks!

castillofrancodamian:
> > I ask if you could fix this error, not correct it in this Pull Request, because I see that you can modify `luci-theme-material`.
> 
> OK, got your point now. Let me take care of this PR (and the other I have open), then I can try to take a look at that one. Thanks!

I understand you and can wait since it is not urgent. Regards!

arrmo:
Hi,

Just following up - anything more to do here, or is this looking OK (to pull in)?

Thanks!

jow-:
Thanks for your work on this! To be honest, I don't like the implementation at all. The main problems I see are that a) themes need to be modified to work with it  b) a fork/exec happening on every page rendering.

I agree with @feckert that the actual battery query commands should be factored into an rpcd ubus plugin so that we can asynchronously query the data without blocking the render process.

I'll see if I can come up with a somewhat generic solution to put header notification badges.

arrmo:
> themes need to be modified to work with it
Yes, agreed! I didn't like that either, but was trying to work with the current structure. We can evolve, no issue here at all.

> come up with a somewhat generic solution to put header notification badges
That would be great - thanks! We can work through this together, no issue. Just yell, let me know what I can do to help.

arrmo:
Sorry, just checking in - any thoughts / updates on this? I'm more than happy to try to work on this, but not sure I'm the right person to try to define the (bigger) approach 🤣 

Thanks!

jow-:
I added a generic mechanism to add arbitrary notification labels, that is the first half of the solution. What is missing yet is a way to register code that should be executed with each view rendering, I'll hopefully get to it during the next days.

arrmo:
> I added a generic mechanism to add arbitrary notification labels, that is the first half of the solution. What is missing yet is a way to register code that should be executed with each view rendering, I'll hopefully get to it during the next days.

NP, really appreciate it - thanks!!!

arrmo:
BTW, to the comment above - while you are working on the display mechanism here ... should I be trying to convert my code into a function (i.e. command line, returns result)?

Thanks!

arrmo:
Hi,

Sorry, just not sure what to do here, to get this included. Do you want the status / information collected via a script? But also, how to add it to the header (for battery display)?

Thanks!!!

jow-:
HI @arrmo - sorry, I forgot about this open PR. So a while back I added a generic indicator mechanism which allows to add arbitrary badges to the top right notification area and implemented a proof of concept for GSM signal indication. I think you can adapt that for the battery level.

You need three parts for that: a client side JavaScript class which fetches raw values from ubus and renders them in an appropriate way, a server side ubus rpc applet to expose one or more procedures which provide the raw data, and an ACL specification which grants ubus access for the current user session.

After all three files are placed, you do need to log out of LuCI and issue an `/etc/init.d/rpcd restart` to pick up the new applet.

1. A preload class in `/www/luci-static/resources/preload/cellsignal.js` - any class in `/www/luci-static/resources/preload` is automatically instantiated for each page load.

```javascript

'use strict';
'require ui';
'require rpc';
'require poll';
'require baseclass';

var callSignalInfo = rpc.declare({
	object: 'luci.cellsignal',
	method: 'getSignalInfo',
	expect: { '': {} }
});

var devices = {};

return baseclass.extend({
	__init__: function() {
		this.updateIndicator();
		//poll.add(L.bind(this.updateIndicator, this), 5);
	},

	handleDetails: function(ev) {
		var dev = ev.currentTarget.getAttribute('data-indicator').replace(/^signal-/, ''),
		    info = devices[dev];

		if (!info)
			return;

		ui.showModal(_('Signal details'), [
			E('ul', [
				E('li', [ E('strong', [ _('Device: ') ]), dev ]),
				E('li', [ E('strong', [ _('Signal: ') ]), '%s (%d)'.format(info.bar, info.signal) ]),
				E('li', [ E('strong', [ _('Type: ') ]),   info.type.toUpperCase() ]),
				E('li', [ E('strong', [ _('IMEI: ') ]),   info.imei ])
			]),
			E('div', { 'class': 'right' }, [
				E('button', { 'click': ui.hideModal }, [ _('Close') ])
			])
		]);
	},

	updateIndicator: function() {
		return callSignalInfo().then(L.bind(function(devs) {
			for (var dev in devs) {
				var info = devs[dev],
				    type = info.type,
				    rssi = info.signal;

				if (rssi == 0 || rssi < -85)
					info.bar = '▯▯▯';
				else if (rssi < -75)
					info.bar = '▮▯▯';
				else if (rssi < -65)
					info.bar = '▮▮▯';
				else
					info.bar = '▮▮▮';

				ui.showIndicator('signal-%s'.format(dev),
					'%s %s'.format(type.toUpperCase(), info.bar),
					L.bind(this.handleDetails, this));

				devices[dev] = info;
			}

			for (var dev in devices) {
				if (!devs.hasOwnProperty(dev)) {
					ui.hideIndicator('signal-%s'.format(dev));
					delete devices[dev];
				}
			}
		}, this));
	}
});
```

2. An ubus rpc applet `/usr/libexec/rpcd/luci.cellsignal` providing the `luci.cellsignal` `getSignal` call:

```bash
#!/bin/sh

. /usr/share/libubox/jshn.sh

case "$1" in
	list)
		printf '{ "getSignalInfo": {} }'
	;;
	call)
		case "$2" in
			getSignalInfo)
				json_init

				for dev in /dev/cdc-wdm*; do
					eval $(/sbin/uqmi -d "$dev" --get-signal-info 2>/dev/null | jsonfilter -e 'SIGNAL=@["signal"||"rssi"]' -e 'TYPE=@.type')
					eval $(/sbin/uqmi -d "$dev" --get-imei 2>/dev/null | jsonfilter -e 'IMEI=@')

					if [ -n "$TYPE" ] && [ -n "$SIGNAL" ]; then
						json_add_object "$dev"
						json_add_int signal "$SIGNAL"
						json_add_string type "$TYPE"
						json_add_string imei "$IMEI"
						json_close_object
					fi
				done

				json_dump
			;;
		esac
	;;
esac
```

3. An ACL definition in `/usr/share/rpcd/acl.d/luci-mod-cellsignal.json` - allowing access to the signal info procedure

```json
{
	"luci-mod-cellsignal": {
		"description": "Grant access to cellular signal info",
		"read": {
			"ubus": {
				"luci.cellsignal": [ "getSignalInfo" ]
			}
		}
	}
}
```

arrmo:
No worries! Let me do some digging - my head is spinning a bit, but I'll figure it out ... LOL!

The example above seems to show where to put the files on the router, for development (i.e. to get it working). Once it's doing what it needs to => where do the files go in the LuCI source code (i.e. to push to the repository, so this app can then be made available in the builds)?

Thanks!

jow-:
On the router, these three files should go into `/www/luci-static/resources/preload/battstatus.js`, `/usr/libexec/rpcd/luci.battstatus` and `/usr/share/rpcd/acl.d/luci-mod-battstatus.json` respectively.

Once things are functioning as desired, create the following directory structure in the repo:

* `modules/luci-mod-battstatus/`
  * `Makefile`  (copy from e.g. `modules/luci-mod-network/Makefile`)
  * `htdocs/`
    * `luci-static/`
      * `resources/`
        * `preload/`
           * `battstatus.js`
  * `root/`
    * `usr/`
      * `libexec/`
        * `rpcd/`
          * `luci.battstatus` (make sure it is executable - chmod 0755)
      * `share/`
        * `rpcd/`
          * `acl.d/`
            * `luci-mod-network.json`

arrmo:
Will do, thanks! Once I have this working, will update this PR (or yell if you'd rather I close this one, open a new one - I can do it however you prefer).

arrmo:
Wow! That really works well, all up and running 😄. Just a couple questions,
1) I looking in ui.js, and the last parameter seems to be style, but only active or inactive? Is there a way to color the text I display (e.g. yellow or red, as the battery gets low)?
2) As the new method is really completely different - close this PR, open a new one? They are the same basic function, but really not all that similar in terms of the code or files.

Thanks!

jow-:
Right, there is no style apart from active / inactive. However, after enabling the indicator you can obtain its node and place arbitrary contents. Example:

    ui.showIndicator('foo', '');
    document.querySelector('[data-indicator="foo"]').innerHTML = '<span style="color:#f00">10%</span>';

arrmo:
Perfect, thanks! All working (on the router), now to build it. Put the files in the correct locations, but it's not showing up in make menuconfig ... I recall there was a trick to get it to show up?

jow-:
Did you run the following commands?

    ./scripts/feeds update luci
    ./scripts/feeds install luci-mod-battstatus

arrmo:
> Did you run the following commands?

I did! But I think I see the problem - because I'm on my branch, the update is failing => but if I stash my changes, to be able to go back to master, then the update won't detect of course. Catch 22 it seems 😞.

Hmmm, but I also see,
```
* (HEAD detached at origin/luci-app-battstatus) ac58fd8f5 Initial release, luci-app-battstatus
  master                                        2e2ab59ea [upstream/master] Merge pull request #4219 from rkkoszewski/rkkoszewski/feat/luci-proto-sstp
```

Not sure if detached is the issue (nor how I resolve that ... I'm a git-idiot 😆).

arrmo:
OK, I can't explain it, but re-did the git checkout, and now I get it working. Sorry for the hassle!

Building now 😄.

arrmo:
Woohoo! All working 😄. Pic below (and updated the commit),
image

Do you need anything else from me here? If so, just yell.

And BTW, the new approach ... ubus/rpcd in shell, JavaScript for updates - it's much nicer to debug, really easy in the browser debugger. Thanks!

castillofrancodamian:
> Woohoo! All working 😄. Pic below (and updated the commit),
> image
> 
> Do you need anything else from me here? If so, just yell.
> 
> And BTW, the new approach ... ubus/rpcd in shell, JavaScript for updates - it's much nicer to debug, really easy in the browser debugger. Thanks!

Out of curiosity, can you change where you get the information about the battery values?

arrmo:
> Out of curiosity, can you change where you get the information about the battery values?

Yep! The code is part of modules/luci-mod-battstatus/root/usr/libexec/rpcd/luci.battstatus ... it could be extended, to execute different commands for different HW. Pretty straightforward really to extend it. And just return the same information / JSON object, no changes need to other files to get it to display.

arrmo:
Hi,

Sorry, just not sure if this is waiting on me - which is entirely possible, and I may just not know it ... LOL.

Thanks!

jow-:
It looks almost perfect, just a  detail left - in the Makefile, please remove the copyright notice or set it to yourself

modules/luci-mod-battstatus/root/usr/libexec/rpcd/luci.battstatus
```diff
@@ -0,0 +1,39 @@
...
+
+				eval $(/bin/ubus call system board 2>/dev/null | /usr/bin/jsonfilter -e 'MODEL=@.model')
+				json_add_object "$MODEL"
+
+				if [ -f /usr/sbin/i2cset ] && [ -f /usr/sbin/i2cget ]; then
+                                        json_add_string valid "True"
+					if [ $(i2cset -y 0 0x0a 0x0a 0x01 && i2cget -y 0 0x0a 0x0a) = 0x40 ] ;  then
+						json_add_string charge "Charging"
+					else
+                                                json_add_string charge "Not Charging"
```
feckert, line 22: Please fix spaces use tabs

modules/luci-mod-battstatus/root/usr/libexec/rpcd/luci.battstatus
```diff
@@ -0,0 +1,39 @@
...
+					fi
+					json_add_int percentage $(i2cset -y 0 0x0a 0x0a 0x10 && i2cget -y 0 0x0a 0x0a | xargs printf %d)
+				else
+					json_add_string valid "False"
+					if [ ! -f /usr/sbin/i2cset ]; then
+	                                        json_add_string message "Need i2cset"
+					else
+                                                json_add_string message "Need i2cget"
```
feckert, line 30: Spaces vs. tabs

modules/luci-mod-battstatus/root/usr/libexec/rpcd/luci.battstatus
```diff
@@ -0,0 +1,39 @@
...
+					if [ $(i2cset -y 0 0x0a 0x0a 0x01 && i2cget -y 0 0x0a 0x0a) = 0x40 ] ;  then
+						json_add_string charge "Charging"
+					else
+                                                json_add_string charge "Not Charging"
+					fi
+					json_add_int percentage $(i2cset -y 0 0x0a 0x0a 0x10 && i2cget -y 0 0x0a 0x0a | xargs printf %d)
+				else
+					json_add_string valid "False"
+					if [ ! -f /usr/sbin/i2cset ]; then
+	                                        json_add_string message "Need i2cset"
```
feckert, line 28: Spaces vs. tabs

feckert:
Wy did you delete  the file themes/luci-theme-rosy/luasrc/view/themes/rosy/header.htm  in this pullrequest?

arrmo:
> It looks almost perfect, just a detail left - in the Makefile, please remove the copyright notice or set it to yourself

Done! Will push in a few minutes 😄.

modules/luci-mod-battstatus/root/usr/libexec/rpcd/luci.battstatus
arrmo: Done!

modules/luci-mod-battstatus/root/usr/libexec/rpcd/luci.battstatus
arrmo: Done.

modules/luci-mod-battstatus/root/usr/libexec/rpcd/luci.battstatus
arrmo: Done.

arrmo:
> Wy did you delete the file themes/luci-theme-rosy/luasrc/view/themes/rosy/header.htm in this pullrequest?

I wasn't necessarily meaning to, but when I tried to revert my changes, it removed this file. I did ask, and was told that this theme is no longer supported, and it doesn't show up in my menuconfig build. I can restore it if needed. Thanks!

jow-:
Yes, please restore `themes/luci-theme-rosy/luasrc/view/themes/rosy/header.htm`, it should not be touched by your PR.

arrmo:
> Yes, please restore `themes/luci-theme-rosy/luasrc/view/themes/rosy/header.htm`, it should not be touched by your PR.

Done - restored! Hopefully all good now 😄.

Thanks!

arrmo:
Hi,

Please just yell if I missed (or messed up) something - always possible, even probable 🤣.

Thanks!

jow-:
Merged with some adjustments in dcac704b3d6bc3c4ad2ccacb3d79fc67d900087f - thank you for your time and effort!

arrmo:
Thanks! BTW, I have some thoughts how to add more (different) HW in the future, storing needed commands in JSON. Just yell if that comes up 😄.

PR #3615, opened 2020-02-08
wireguard: add checkbox for `nohostroute` option

vfreex:
This change allows to configure `nohostroute` option for wireguard to explicitely prevent creation
of host routes to endpoints.

By default without `option nohostroute '1'`, an explicite route to the peer's endpoint will be created in the main routing table with the next hop to the gateway. However, it causes issues with some setup. Enabling this option will inhibit this behavior. See discussions at http://lists.openwrt.org/pipermail/openwrt-devel/2019-March/016329.html.


jow-:
Merged, thanks!

PR #3619, opened 2020-02-09
luci-app-samba4: add support for new UCI option

Andy2244:
* add support for new UCI option "allow_legacy_protocols"

### DIFF
--- a/applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js
+++ b/applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js
@@ -43,6 +43,9 @@ return L.view.extend({
 
 		o = s.taboption('general', form.Flag, 'macos', _('Enable macOS compatible shares'),
 			_('Enables Apple\'s AAPL extension globally and adds macOS compatibility options to all shares.'));
+			
+		o = s.taboption('general', form.Flag, 'allow_legacy_protocols', _('Allow legacy (insecure) protocols/authentication.'),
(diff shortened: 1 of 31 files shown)


applications/luci-app-samba4/Makefile
```diff
@@ -5,6 +5,6 @@ include $(TOPDIR)/rules.mk
 LUCI_TITLE:=Network Shares - Samba 4 SMB/CIFS fileserver
 LUCI_DEPENDS:=+samba4-server
 
-include ../../luci.mk
+include $(TOPDIR)/feeds/luci/luci.mk
```
hnyman: This change looks wrong to me.
You change from relative notation (from this package) to absolute path, making also auumptions about the build system.

Andy2244: luci-app-omcproxy was doing the same, so i used this style, since it fixes a issue i had if users wanted to build from my external feed. The problem was i had to maintain two versions of my luci app, since the relative notation did not work from other feeds, so i switched to this.
So far it did work for me, but if this is a issue i can change it back, its just inconvenient to have to remember to change the path if i update my apps for my feed + luci repo.

hnyman:
@jow- can maybe comment on this, but `include ../../luci.mk` is the standard form that should be used, I think.

applications/luci-app-samba4/Makefile
Andy2244: ok before this gets stuck any longer i just reverted this change.

jow-:
Merged, thanks @Andy2244

PR #3631, opened 2020-02-14
build: unify and clean up the  uci-defaults scripts

feckert:
- move luci-indexcache delete command to luci.mk postinst script
- move luci-modulecache delete command to luci.mk postinst script


ysc3839:
There's `luci-indexcache.xxx.json`, should it be deleted too?
https://github.com/openwrt/luci/blob/08eea3f93e9f2e9a2940df2e3a7afbe04c40f4b6/modules/luci-base/luasrc/dispatcher.lua#L1038

jow-:
It should not be needed to delete .json cache as LuCI is smart enough to auto-purge it.

PR #3639, opened 2020-02-17
luci-mod-system: add led plugin infrastructure

feckert:
This pullrequest is intended to create the possibility that not only kernel-led-triggers can be selected but also application-led-triggers from user space. This is done via a plugin mechanism.
Application-led-triggers are scripts that set LEDs on system events or services (for example the package rssileds service). Until now this has not been possible

The following new packages are added:

- luci-app-ledtrig-rssi (application-led-trigger) 
- luci-app-ledtrig-switch (kernel-led-trigger) not needed on every most devices
- luci-app-ledtrig-usport (kernel-led-trigger) optional trigger

Since we have now a plugin mechanism I have added the following triggers as a dependency. So this triggers are now installed per default on LuCI installation.

- kmod-ledtrig-default-on
- kmod-ledtrig-heartbeat
- kmod-ledtrig-netdev
- kome-ledtrig-timer

The kernel trigger `kmod-ledtrig-usbdev` was removed with the commit https://github.com/openwrt/openwrt/commit/d0b50c2770a0e2d54b37153f2801e2e7dc865fa6. So I have not ported the relevant code anymore.

If we do not want to get a warning on /etc/init.d/led restart if we select an application-led-trigger the following commit should also get commit in openwrt https://github.com/openwrt/openwrt/pull/2741

### DIFF
--- a/applications/luci-app-ledtrig-rssi/Makefile
+++ b/applications/luci-app-ledtrig-rssi/Makefile
@@ -1,0 +1,16 @@
+#
+# Copyright (C) 2020 TDT AG <development@tdt.de>
+#
+# This is free software, licensed under the Apache License Version 2.0.
+# See https://www.apache.org/licenses/LICENSE-2.0 for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:= LuCI Support for ledtrigger rssi
+LUCI_DEPENDS:=+rssileds
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
+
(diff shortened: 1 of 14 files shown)


jow-:
I didn't runtime-test the code but the approach is fine with me.

feckert:
Thanks for the positive feedback. I will change a few things, especially the switch<x>. And I will also update the translation.

feckert:
I would merge this in the next days.
Since this is a major change, could someone please review this?

PR #3641, opened 2020-02-17
luci-app-snmpd: Add snmpd as dependency

kbabioch:
luci-app-snmpd allows to configure snmpd via LuCI. Without making sure that
the snmpd package is installed, the daemon that is about to be configured via
luci-app-snmpd might be missing (unless explicitly installed otherwise).

This is not only counter-intuitive, but also inconsistent with other LuCI based
applications, as they always include the underlying package(s) in order to make
sure that the software being configured is being available in the first place.


karlp:
LGTM

PR #3649, opened 2020-02-19
luci-app-ddns: fix update interval unit values in combobox

namedun:
Value "hours" is used twice instead of "hours" and "days".


feckert:
Thanks

PR #3659, opened 2020-02-21
luci-app-yggdrasil: fix listen uri type and remove tap for v3.13

wfleurant:
Tested against yggdrasil v3.12-develop (v3.13) / APU2 x86_64 (next release expected tomorrow)
uri is required, using address was omitting listen address for: `ygguci get`   
tap interface support is removed in v3.13

do we need to update translation maps file now?


jow-:
Updating translation templates is only strictly necessary if you introduce new strings. For removals, changed locations etc. we do occasional tree-wide translation syncs which will take care of such changes.

PR #3676, opened 2020-02-24
luci-lib-nixio: allow building with Lua5.2/LuaJIT

Tarnyko:
Lua 5.1 defines a "luaL_Reg" alias for deprecated
"luaL_reg", but Lua >= 5.2 and LuaJIT do not.
Replace by "luaL_Reg", and define a "luaL_reg"
alias in case we build with old Lua 5.0.

Signed-off-by: Manuel BACHMANN <tarnyko@tarnyko.net>


jow-:
Merged, thank you!

PR #3679, opened 2020-02-25
luci-app-aria2: fix incorrect "Extra Settings" section option name

rwanyoike:
The "Extra Settings" section `extra_setting` option name does not match what the `net/aria2` package expects: `extra_settings`. 

See:

- https://github.com/openwrt/packages/blob/openwrt-19.07/net/aria2/files/aria2.init#L311
- https://github.com/openwrt/packages/blob/openwrt-19.07/net/aria2/files/aria2.conf#L31

Causing the luci UI "Extra Settings" option to have **no effect**. This fix renames `extra_setting` => `extra_settings`


feckert:
It makes sense good catch.

Thanks merged

moralrebuild:
Good to see the fix.

Issue #3644, opened 2020-02-18
RFC: rework LuCI theme to Bootstrap with new design guidelines


aparcar:
Hi all, now that OpenWrt got a new [design guide](https://openwrt.org/docs/guide-graphic-designer/openwrt-logo) we should consider reworking the current theme to use adequate colors and maybe update the current bootstrap file ([1.4.0](https://github.com/openwrt/luci/blob/master/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css) vs [4.1.3](https://getbootstrap.com/docs/4.1/getting-started/introduction/)).

This also include renaming a *lot* of elements, however is surely worth the work! From my understanding it would mostly mean to rename classes created by the new JavaScript code.

Please share your opinion on this!

tmo26:
Good idea!

I just stumbled myself over a place where the new logo needs to be implemented: LuCI favicon.

- IS currently: ![grafik](https://user-images.githubusercontent.com/12309103/74957055-4bb6d900-5407-11ea-8bd3-284d755eadf0.png)
- SHOULD BE in the future: ![grafik](https://user-images.githubusercontent.com/12309103/74957272-9b95a000-5407-11ea-8232-87f9f5608cb7.png)

Just as a reminder so it won't be forgotten.

ysc3839:
I found both bootstrap and material theme uses a highly modified css comparing to original bootstrap and mui. Is there any technological limit that we must modify css to make it works on luci? Can't we just use the original css? Because using original css makes authoring new theme easier.
@jow-

jow-:
> Can't we just use the original css?

The original CSS is too bloated. Unmodified Bootstrap 4.4.1 `bootstrap.min.css` is 159.5KB compared to 41.8 KB *unminified* in luci-theme-bootstrap.

ysc3839:
@jow- I'm not talking about size, those useless code can be deleted. The problem is, when I replaced `cascade.css` of bootstrap theme with the original 1.4.0 [bootstrap.min.css](https://getbootstrap.com/1.4.0/assets/css/bootstrap.min.css), the page become this:
![image](https://user-images.githubusercontent.com/12028138/75118254-86449e00-56b3-11ea-97f0-259b89cacf9c.png)
So AFAIK, we have to modify css to make a new theme works on luci, which makes authoring new theme more difficult. I'm talking about can we use those css without making out own changes?

jow-:
[...]

PR #3684, opened 2020-02-26
[RFC, WIP] luci-theme-bootstrap4: add new theme

ysc3839:
As discussed in #3644. I'm making a new Bootstrap 4 theme.
This theme contains two parts, one is in luci, another is in my [fork](https://github.com/ysc3839/bootstrap) of bootstrap.
The part in luci is based on current bootstrap theme. In order to show changes in git diff, I modify original code directly, and I will make a copy after it's done.
The part in my fork of bootstrap is based on bootstrap v4.4.1, you can see changes [here](https://github.com/twbs/bootstrap/compare/v4.4.1...ysc3839:luci).
I'm not familiar with Web front-end development. In order to simplify the work, I mainly use Sass' `@extend` to make it works with luci. However this will increase the size of final css. If you have any idea please tell me.

Currently it looks like this:
image

Ping @jow- and @aparcar


aparcar:
Hi, thank you very much for looking into this! I think even the early stage looks already pretty good!

From my understanding Bootstrap changed the names of various `classes` which you *map* with the scss files, right? Wouldn't a treewide replace of such classes also do the trick?

I think part of the legacy nature of the current cascade.css its differ so much from upstream....

jow-:
> Wouldn't a treewide replace of such classes also do the trick?

Not really, it would just push the burden of matching the markup to other themes. Keep in mind that there is not only bootstrap based themes.

ysc3839:
> Wouldn't a treewide replace of such classes also do the trick?
No, for example, `cbi-tabmenu` class maps to navs with tab `nav nav-tabs` in bootstrap.

aparcar:
> Not really, it would just push the burden of matching the markup to other themes. Keep in mind that there is not only bootstrap based themes.

Right, but if we also change their CSS files? I understand this is a bit of a mess once done future migration should be much simpler. I'd even claim that much more themes appear because boostrap CSS is so much more common.

>  No, for example, cbi-tabmenu class maps to navs with tab nav nav-tabs in bootstrap.

If you provide a map of such replacements I'll do the boring work of replace it within the code :)

ysc3839:
@aparcar I think you are talking about replace in bootstrap source.

aparcar:
> @aparcar I think you are talking about replace in bootstrap source.

No, replacing the LuCI bootstrap theme CSS and the other 3s. Keep the `boostrap.min.css` untouched.

jow-:
I don't want to peg LuCI markup to bootstrap though, it should remain agnostic. Besides bootstrap lacks a lot of component markup which we need so custom CSS is required anyway.

ysc3839:
> No, replacing the LuCI bootstrap theme CSS and the other 3s. Keep the `boostrap.min.css` untouched.

But other theme may use a different class name, this will result in modifying

aparcar:
> But other theme may use a different class name, this will result in modifying

Yes, so edit these other themes but not the original bootstrap file.

> Besides bootstrap lacks a lot of component markup which we need so custom CSS is required anyway.

Can you please elaborate a bit on this. What [compontents](https://getbootstrap.com/docs/4.4/components/alerts/) are missing in bootstrap?

ysc3839:
> Yes, so edit these other themes but not the original bootstrap file.

It's unfair to maintainers of other theme.

jow-:
> Can you please elaborate a bit on this. What compontents are missing in bootstrap?

- Comboboxes
- Rich dropdowns
- Combo buttons
- Rich tooltips

Another problem is bootstrap's reliance on JavaScript to implement more complex controls.

jow-:
One further thought about unmodified bootstrap: it contains extreme amounts of whitespace, we would need to modify that as well.

aparcar:
> It's unfair to maintainers of other theme.

I think this is an misunderstanding, I would replace things like `cbi-button` treewide to `btn` so no extra workload appear for the maintainers. As the class would be changed in their CSS files too, it'd still work.

> Another problem is bootstrap's reliance on JavaScript to implement more complex controls.

Thanks for the quick response. I wonder if these elements are trivial to implement in JavaScript and could be an additional layer of CSS/JS on top of Bootstrap. Also if using Bootstrap JavaScript code would safe LuCis JS at other points.

Please don't misunderstand my comments as Bootstrap fandom, I barely used it before. I'm just fan of lowering the obstacles for new people to join in. Maybe it's a dead end. 

After all Bootstrap seems to break compatibility between versions anyway [3](https://getbootstrap.com/docs/3.4/migration/) & [4](https://getbootstrap.com/docs/4.4/migration/) 

> One further thought about unmodified bootstrap: it contains extreme amounts of whitespace, we would need to modify that as well.

The `bootstrap.min.css` file is *minimized*

ysc3839:
> I would replace things like cbi-button treewide to btn so no extra workload appear for the maintainers.

But maybe there are some complex components.

jow-:
> Also if using Bootstrap JavaScript code would safe LuCis JS at other points.

From a quick glance, there is little overlap.

> The bootstrap.min.css file is minimized

I am talking about visual space as in margins and paddings, not white space in the CSS file.

ysc3839:
@jow- We need a luci app which contains all luci components. It can be used for theme development and testing, and demonstrate what components are available and how to use them.

ysc3839:
It's difficult for me to make the "Save & Apply" behave normal, maybe I have to change it's structure to make it work.
Current screenshot:
image

ysc3839:
Moved `#tabmenu` to top and use a new style:
image

Added border shadow to `.cbi-section`:
image

ysc3839:
Finished `#topmenu` and cbi section.
image
image

ysc3839:
@jow- Can we use standard table tags instead of div with .table class? Because doing this I don't need to write css for table classes.
image
image

messismore:
> One further thought about unmodified bootstrap: it contains extreme amounts of whitespace, we would need to modify that as well.

Just to chime in on this. I just updated 17.x → 20.x and find the denser UI harder to parse and navigate. The more generous amounts of white space in previous releases helped to easily discern visual elements and aided intelligibility and navigation.

Busy Luci

17.x on the left, 20.x on the right.  While the individual differences seem negligible, they do add up. Thus white space can be considered an accessibility feature, especially for users that do not know every nook and cranny of the UI.

aparcar:
@messismore This should be possible later on. I think for now it is mostly about making the current JavaScript framework work together with Bootstrap 4.

@ysc3839 Can you please share your modified bootstrap.css and add it to the current PR?

jow-:
> @jow- Can we use standard table tags instead of div with .table class? 

That is not doable and would require tree-wide changes to a lot of components. For now just add the following lines to your CSS:

```
.table { display: table; }
.tr { display: table-row; }
.th, .td { display: table-cell; }
```

ysc3839:
@aparcar Updated.

ysc3839:
Close this pr since it requires a lot of work, and I have no time doing this.
I'm looking for a CSS utility framework to build our own theme, instead of out-of-box CSS framework like bootstrap.
Recently I found [Tailwind CSS](https://tailwindcss.com/) which may suit for making a theme.

PR #3688, opened 2020-02-27
luci-app-adblock query form: Fix enter key submitting form but not showing any results

stefansundin:
Right now you can hit the enter key while the input field is focused, and it submits the form, but no result is displayed. Even though a POST request is made, it doesn't seem to actually be querying the domain.

It might be nice to fix that too, so it would work for users that have JavaScript disabled. But I am not sure how to make that work, so if someone can provide help with that, that would be great.

Thanks!

### DIFF
--- a/applications/luci-app-adblock/luasrc/view/adblock/query.htm
+++ b/applications/luci-app-adblock/luasrc/view/adblock/query.htm
@@ -43,7 +43,7 @@ This is free software, licensed under the Apache License, Version 2.0
 //]]>
 </script>
 
-<form method="post" action="<%=REQUEST_URI%>">
+<form method="post" action="<%=REQUEST_URI%>" onsubmit="update_status(this.input); return false;">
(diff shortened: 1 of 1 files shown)


dibdot:
Thanks for your contribution!
Please remove the CONTRIBUTING.md change from this PR, its unrelated to adblock and should go in a separate PR.

Thanks!

stefansundin:
Done. I cherry-picked that commit to this PR: https://github.com/openwrt/luci/pull/3708

dibdot:
Thanks!

dibdot:
cherry-picked to 19.07 as well.

stefansundin:
@dibdot Just wanted to note that it may be desirable to fix the form submission as well? I just didn't know how to do that. Thanks.

dibdot:
I don't really care, cause I've already migrated forthcoming adblock 4.0 to client side JS ...;-)

PR #3693, opened 2020-02-28
Remove luci-compat dependency

howard0su:
No LUA script in this application anymore


howard0su:
already commit with other changes.

PR #3704, opened 2020-03-01
luci-app-statistics: fix 'RRATimespan' option for rrdtool plugin

namedun:
Collectd plugin rrdtool uses option 'RRATimespan' not 'RRATimespans' (see [sources](https://github.com/collectd/collectd/blob/f43dc9256efce4bd53d451115a23c584b5cda2d6/src/rrdtool.c#L62)).


namedun:
I'm sorry, my mistake. The stat-genconfig script handles these options correctly.

PR #3713, opened 2020-03-03
luci-mod-network: extended interfaces.js for check if new interface i…

he-ma:
In some cases it could be needful to check whether a new interface can be created.
This extension implements this check and if the proto handler returns with an error the creation is aborted, the modal is closed and an error message is shown.

This PR refers to https://github.com/openwrt/luci/pull/3712

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -760,24 +760,35 @@ return L.view.extend({
 							'class': 'cbi-button cbi-button-positive important',
 							'click': ui.createHandlerFn(this, function(ev) {
 								var nameval = name.isValid('_new_') ? name.formvalue('_new_') : null,
-								    protoval = proto.isValid('_new_') ? proto.formvalue('_new_') : null;
+								    protoval = proto.isValid('_new_') ? proto.formvalue('_new_') : null,
+								    protoclass = protoval ? network.getProtocol(protoval) : null;
 
 								if (nameval == null || protoval == null || nameval == '' || protoval == '')
 									return;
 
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -765,19 +765,32 @@ return L.view.extend({
 								if (nameval == null || protoval == null || nameval == '' || protoval == '')
 									return;
 
-								return m.save(function() {
-									var section_id = uci.add('network', 'interface', nameval);
+								return network.getProtocol(protoval).isCreateable(nameval).then(L.bind(function(checkval) {
+									if (checkval != null) {
+										ui.addNotification(null,
+												E('p', _('New interface for') + ' "' +
```
jow-: Please change to:

`E('p', _('New interface for "%s" can not be created: %s').format(protoclass.getI18n(), checkval))`

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -765,19 +765,32 @@ return L.view.extend({
 								if (nameval == null || protoval == null || nameval == '' || protoval == '')
 									return;
 
-								return m.save(function() {
-									var section_id = uci.add('network', 'interface', nameval);
+								return network.getProtocol(protoval).isCreateable(nameval).then(L.bind(function(checkval) {
```
jow-: - Please cache the result of `network.getProtocol(protoval)` in a local variable (e.g. `protoclass`)
- No need to `bind()` the `then()` function since we do not use `this` anywhere

he-ma:
@jow- 
I made your requested changes

PR #3716, opened 2020-03-03
luci-app-wireguard: fix String.format is not a function

mival:
Because of missing JS import page JS is crashing and dos not work.


feckert:
I am not sure if this is necessary but please update the commit head with
`luci-app-wireguard: <your message> `

feckert:
Branch 19.07 is EOL

PR #3720, opened 2020-03-04
luci-proto-bonding: add protohandler configuration for channel bonding

he-ma:
This is a luci implementation to configure Link Aggregation (Channel Bonding) through the netifd proto handler.

The netifd proto handler is already integrated in the OpenWrt packages.

This PR is related with https://github.com/openwrt/luci/pull/3712, https://github.com/openwrt/luci/pull/3713

Although I have already done the German translations for this module, I don't know where to push them. Any advice would be appreciated


protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+'require network';
+
+function getSelectableSlaves(section_id) {
+	var rv = [];
+	var interfaces = uci.sections('network', 'interface');
+
+	return network.getDevices().then(function(devices) {
+		for (var i = 0; i < devices.length; i++) {
+			var in_use = false;
+			if (devices[i].ifname.match(/eth/)) {
```
jow-: Please use `devices[i]._devstate('type') == 1` to select ethernet interfaces only.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+	},
+
+	renderFormOptions: function(s) {
+		var o;
+
+		o = s.taboption('general', form.Value, 'ipaddr',
+				_('IPv4 address'),
+				_('The local IPv4 address'));
+		o.datatype = 'ip4addr';
+		o.validate = validateEmpty;
```
jow-: You can replace this and all other `o.validate = validateEmpty` lines with `o.rmempty = false` to achieve the same effect which allows you to get rid of `function validateEmpty() { ... }`

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+				var if_slaves = uci.get('network', section_id, 'slaves');
+
+				if (if_slaves != null) {
+					return if_slaves;
+				}
+
+				return;
+			}, this));
+		};
+		o.cfgvalue = function(section_id) {
```
jow-: This looks like the default implementation and can be dropped.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+				var primary = uci.get('network', section_id, 'primary');
+
+				if (primary != null) {
+					return primary;
+				}
+
+				return;
+			}, this));
+		};
+		o.cfgvalue = function(section_id) {
```
jow-: This looks like the default implementation and can be dropped.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+		o.default = 'balance-rr';
+		o.value('balance-rr', _('Round-Robin policy (balance-rr, 0)'));
+		o.value('active-backup', _('Active-Backup policy (active-backup, 1)'));
+		o.value('balance-xor', _('XOR policy (balance-xor, 2)'));
+		o.value('broadcast', _('Broadcast policy (broadcast, 3)'));
+		o.value('802.3ad', _('IEEE 802.3ad Dynamic link aggregation (802.3ad, 4)'));
+		o.value('balance-tlb', _('Adaptive transmit load balancing (balance-tlb, 5)'));
+		o.value('balance-alb', _('Adaptive load balancing (balance-alb, 6)'));
+
+		o = s.taboption('advanced', form.ListValue, 'primary',
```
jow-: Mabye consider using `luci.tools.widgets.DeviceSelect` here and use its `.filter = function(...) { ... }` callback to only present slave candidates.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: If I use your suggested change, I get more than only the eth interfaces returned (e.g. erspan0, wwan0, wwan1 and -indeed- former created bonding interfaces). That is not what I want. I really want only the eth interfaces. So, unfortunately, your suggestion doesn't work for me. Any other suggestions?

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: Oh, this is really a valuable tip I didn't know.
It works well, but it misbehaves a little if you add some o.value() without an o.default().
In this case yo get a dropdown with "-- Please choose --" per default but this is not interpreted as empty (leading e.g. to create a new interface nevertheless)

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: OK, will do

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: OK, will do

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: I think I have a misunderstanding here.
Exept for using the widgets instead the ListValue I see no more advantages.
If I change the slave candidates during the page is shown this will not update the primaries dependencies, isn't it? The primaries list will only be updated on loading the page. Right?
So, I will still need the validate function.
Is there a solution to dynamically update the primaries list?

he-ma:
@jow- Unfortunately I didn't hear anything from you.
At this times I hope you are well.
Could you give some comments on my annotations, please.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
jow-: You can update the primaries from the `validate()` callback of another function using something like this technique: https://github.com/openwrt/luci/blob/3c4bc228a1f7b5731cf464f3f407c9ed9ace3cd2/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js#L420

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
jow-: In addition to type `1`, try filtering for `devices[i]._devstate('flags', 'noarp') == false` too.

he-ma: Unfortunately Your suggestions didn't work as wanted, so I fixed with MAC address check.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: Well, I removed all o.validate and substituted them with o.rmempty exept the special case I mentioned above.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: I dropped it according your suggestion.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: I dropped it according your suggestion.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: I made it following your suggestion.

he-ma:
@jow- I completed the work as you suggested - only one implementation detail is different as I found no solution for this.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+	return network.getDevices().then(function(devices) {
+		for (var i = 0; i < devices.length; i++) {
+			var in_use = false;
+			var NotUsable = NonUsableMac.test(devices[i].getMAC());
+
+			// Only "real" interfaces for slaves needed
+			if (NotUsable == false) {
+				for (var j = 0; j < interfaces.length; j++) {
+					if (uci.get('network', interfaces[j]['.name'], 'proto') == 'bonding') {
+						var slaves = uci.get('network', interfaces[j]['.name'], 'slaves');
```
jow-, line 24: I suggest the following here:
`var slaves = L.toArray(uci.get('network', interfaces[j]['.name'], 'slaves'));`

The has a few benefits:
 - UCI lists are retained as-is
 - UCI options are converted to lists by breaking the value on spaces
 - Empty/unset values are converted to an empty list so you do not need an extra `if (slaves != null)` check

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+
+	var opt = this.map.lookupOption('slaves', section);
+	var selected_slaves = opt[0].formvalue(section);
+
+	var opt = this.map.lookupOption('primary', section);
+	var selected_primary = opt[0].formvalue(section);
+
+	if (selected_policy == 'active-backup' || selected_policy == 'balance-tlb' || selected_policy == 'balance-alb') {
+
+		for (var i = 0; i < selected_slaves.length; i++) {
```
jow-, line 118: I believe you can shorten this code to something like this:

```
if (selected_slaves.filter(function(slave) { return slave == selected_primary }).length == 0)
    return _('You must select a primary interface which is included in selected slave interfaces!');
```

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
```diff
@@ -0,0 +1,442 @@
...
+
+				if (devices.length == 0) {
+					this.placeholder = _('No more slaves available, can not save interface');
+					this.value('', '');
+					return '';
+				}
+
+				var if_slaves = uci.get('network', section_id, 'slaves');
+
+				if (if_slaves != null) {
```
jow-, line 205: This seems redundant, simply `return if_slaves;`

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: OK, I made it following your suggestion.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: OK, I made it following your suggestion.

protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
he-ma: OK, I made it following your suggestion.

he-ma:
Sorry, my last push went wrong anyway.
I'm going to repair it

he-ma:
@jow- Sorry something strange happened during my last push.
I will reopen a new PR with all the changes you requested tomorrow

PR #3723, opened 2020-03-05
luci-app-attendedsysupgrade: use new ASU API

aparcar:
The API of ASU changed as it now uses a static `versions` JSON file.
This moves the upgrade validation to the client site and simplifyies the
server code.

Build requests now use the unique `profile` withouth the need of sending
the devices target.

Also some XMLHttpRequests were replaced by the newer `fetch` function.


aparcar:
Tested on Netgear R6220 and works

PR #3749, opened 2020-03-12
luci-base: fix error 404 on missing relay protocol

Ansuel:
Currently relay.js is included in any case even if
the router doesn't have the needed package to use it.
Fix this by checking if the system has this feature.


feckert:
I wonder why we have the line relay in network.js. 
We have already a package luci-proto-relay?

Ansuel:
I think it's there to make it included (with the require) in the page that use the network module. 
(the luci-proto adds only relay.js but is actually loaded by network.js)

jow-:
Merged, thanks!

PR #3754, opened 2020-03-14
luci: fix exception in OpenWrt theme

Ansuel:
With the Firefox debugger is was reported that
with OpenWrt theme, 2 exception are reported.
One was caused in ui.js in case val is nil
The other were caused by invalid &nbsp and <br>
that are not accepted in XHTML.

@jow- is the use of textContext accepted ? innerHtml 
cause also an exception. 

Also it would be good to merge this in 19.07 too... 


jow-:
Thanks for this fix. I merged it with several changes in 03af89a24:
 - Converted `&nbsp` to `&#160;` which is both syntactically correct (semicolon present) and XHTML compliant
 - Dropped the `ui.js` change since it is unrelated and already wrapped by try/catch.

PR #3769, opened 2020-03-20
luci-base: accept alternative logread location

dibdot:
* minimal change to accept the usual logread location
  plus the alternative location (/usr/sbin/logread)
  used by syslog-ng (see openwrt/packages/issues/11535 for reference)


jow-:
Can you try to simply replace `/sbin/logread` with `logread` on the JS side (not in the ACL rule), this should auto-discover the executable in `$PATH`.

dibdot:
That's what I've tried first, but I always get an access denied.
grafik

jow-:
Merged, thanks!

Issue #3740, opened 2020-03-10
Error canceling load quickly


castillofrancodamian:
It appears throughout the LuCI interface when I do I do that action. I have snapshots.

![Sin título](https://user-images.githubusercontent.com/47670442/76365543-46fa8c00-6306-11ea-8719-3503d2a95603.png)

jow-:
Yes, don't do that then :) If you cancel page loading midway, related XHR requests will be aborted as well, no way to change that.

dibdot:
@jow- May I ask why do we report this "error" to the user at all? The popup comes up for a few seconds and then automatically disappears, cause the user already clicked another top level menu option. Maybe there are other use cases where this error popup makes more sense, I don't know.

Thanks!

jow-:
Good question, I guess we can simply hide this particular case.

castillofrancodamian:
That's the problem.  That appears all in time.  I forgot to clarify.

PR #3772, opened 2020-03-20
luci-base: remove needless warning message

dibdot:
* remove needless 'XHR request aborted by browser' warning message,
  see #3740 for reference


castillofrancodamian:
They don't want it to merge?

hnyman:
@jow- 
any opinion here?  
You know this best, and implicated in #3740 that this might be hidden.

jow-:
This PR cannot be merged like this. Since you removed the call of `rejectFn()` without replacing it with a `resolveFn()` or another `rejectFn()`, the promise for the XHR request will never complete in such a case, leaving ui code hanging forever.

PR #3774, opened 2020-03-21
Fix simpleform and dynslist in 18.06

lisaac:
- Add support default value in Dynlist
- Fix Simpleform redirect issue


feckert:
Closing openwrt-18.06 is EOL

PR #3778, opened 2020-03-22
ucitrack: fix bug for exec with params

jjm2473:
In file `/etc/config/ucitrack`
```
config fstab
        option exec '/sbin/block mount'
```
`/sbin/block mount` never be called after fstab changed.

Signed-off-by: Liangbin Lian <1129525450@qq.com>


jow-:
Merged, thank you.

PR #3786, opened 2020-03-26
This addresses the issue of openconnect.sh from openconnect package expecting a vpn- suffix, while the frontend didn't.

DasTestament:


nmav:
I think a better description is needed especially since this spans various projects. Something like:

This addresses the issue of openconnect.sh from openconnect package expecting a `-vpn` suffix, while the frontend didn't.

Addresses:
https://github.com/openwrt/packages/issues/11584

nmav:
Other than the description LGTM

DasTestament:
Well, I thought to make changes in openconnect.sh, but came to conclusion that fixing frontend will be much proper place rather than messing with a script. I've redacted the title.

nmav:
@jow- any concern in merging?

jow-:
Merged via adf0fb187 - thanks!

PR #3790, opened 2020-03-28
luci-app-dawn: add luci app for dawn

PolynomialDivision:
Dawn is a decentralized WiFi controller.
Just install dawn and the APs will find each other via umdns. They periodically exchange information about connected clients, wireless statistics and other needed information. With that, the daemon load balances clients between different APs through association control.

Luci-app-dawn is the graphical user interface.
It allows to:
- Configure dawn
- View Wireless Network Overview
- View Hearing Map

The hearing map is the list of all probe requests seen from a client from all APs that are running the controller.

networkoverview

hearingmapoverview

configurationdawn

https://github.com/openwrt/packages/pull/11648

### DIFF
--- a/applications/luci-app-dawn/Makefile
+++ b/applications/luci-app-dawn/Makefile
@@ -1,0 +1,10 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for DAWN
+LUCI_DEPENDS:=+dawn
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
+
--- a/applications/luci-app-dawn/luasrc/controller/dawn.lua
+++ b/applications/luci-app-dawn/luasrc/controller/dawn.lua
@@ -1,0 +1,8 @@
+module("luci.controller.dawn", package.seeall)
+
+function index()
+    entry({ "admin", "dawn" }, firstchild(), "DAWN", 60).dependent = false
+    entry({ "admin", "dawn", "configure_daemon" }, cbi("dawn/dawn_config"), "Configure DAWN", 1)
+    entry({ "admin", "dawn", "view_network" }, cbi("dawn/dawn_network"), "View Network Overview", 2)
+    entry({ "admin", "dawn", "view_hearing_map" }, cbi("dawn/dawn_hearing_map"), "View Hearing Map", 3)
+end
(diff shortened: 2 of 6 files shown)


PolynomialDivision:
Before merging, please wait 1 sec. :)

PolynomialDivision:
> Before merging, please wait 1 sec. :)

Finished. :) Please merge. :P

applications/luci-app-dawn/Makefile
```diff
@@ -0,0 +1,9 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for DAWN
+LUCI_DEPENDS:=+luci +dawn +luci-lib-json +luci-lib-httpclient +luci-compat
```
hnyman: Dependency on the "luci" meta-package sounds wrong.

PolynomialDivision: Oh, okay, I will delete the `+luci`. I had this in my own repository and I think this was needed. I will delete that.

PolynomialDivision:
@hnyman I removed all dependencies and tested again! Works! :) Could u merge?

hnyman:
I could merge it otherwise, but I doubt that your real name is Polynomialdivision 
Please fix the commit message to have the signed off with your real name.

PolynomialDivision:
> I could merge it otherwise, but I doubt that your real name is Polynomialdivision
> Please fix the commit message to have the signed off with your real name.

I did it. But now check fails. :/

hnyman:
I will merge it.
The next step would be to change it for client-side JS, instead of lua.

castillofrancodamian:
And create the template to translate.

PolynomialDivision:
The package is still not appearing on the web server? Is something wrong? I can not find any faillogs...

hnyman:
It is built for about 50 different targets, one by one. 
http://buildbot.openwrt.org/master/packages/grid
http://buildbot.openwrt.org/master/packages/one_line_per_build

image

PolynomialDivision:
Ah thanks! I thought every packet would be build around 0:00... :)

castillofrancodamian:
The translation template is missing in the application directory. I couldn't create it with i18n-scan.pl/i18n-update.pl (there are other apps that I can't get to work either).

PolynomialDivision:
> The translation template is missing in the application directory. I couldn't create it with i18n-scan.pl/i18n-update.pl (there are other apps that I can't get to work either).

I did not create any translation template? Could u post the error? I'm very thankful for PRs. :)

hnyman:
The template would make no difference, as most of strings in the app have not been marked as translatable.  (no "translate" wrappers on strings)

hoangminh88:
How install dawn & luci-app-dawn in Openwrt 19.07.3 stable?

Xtreme512:
> How install dawn & luci-app-dawn in Openwrt 19.07.3 stable?

exactly... I want to try it out for if it helps my 802.11k and .11v problems, posted also in OpenWrt forums. we need a stable release.

PR #3808, opened 2020-04-01
Fix typo in configSummary string

ce-4:
small one-char typo fix


jow-:
Merged via 4be05a5fb - thanks!

PR #3819, opened 2020-04-03
set mediaurlbase only when it is the first time to be installed

ptpt52:
If we build multi-themes into firmware, each of them set itself
to be the default theme, what theme should it be?

To make it clear, we only set mediaurlbase if the theme is the
first time to be installed/built-in.

This resolve the issue that theme always change to somewhat default
after upgrading the firmware even with a config-keep-upgrade

### DIFF
--- a/themes/luci-theme-bootstrap/root/etc/uci-defaults/30_luci-theme-bootstrap
+++ b/themes/luci-theme-bootstrap/root/etc/uci-defaults/30_luci-theme-bootstrap
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 if [ "$PKG_UPGRADE" != 1 ]; then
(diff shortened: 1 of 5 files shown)


feckert:
Please update your commit header.
`luci-theme-*: <your message>`

jow-:
@ptpt52 - I agree with this change, it makes sense. But can you change the checks to test `luci.main.mediaurlbase` instead of `luci.themes.XXX` ?

The `luci.themes.XXX` UCI key is just for populating the dropdown in Admin -> System -> Language & Style while `mediaurlbase` actually selects the active theme.

ptpt52:
@jow- 
I mean to check the `luci.themes.XXX`, if this not exist that means it is first time this theme install to system, and we set it as default theme.

jow-:
Merged, thanks!

PR #3832, opened 2020-04-03
luci-proto-gre: Protocol extension for GRE tunnels

mbargo23:
I'm running several GRE tunnels to different locations and
the option to see and to configure GRE tunnels in LuCI was not
crucial but nice to have.

My motivation was to simplify the management of GRE tunnels; and also to help other people asking on the internet for this functionality.

Debugged and tested on router Turris 1.1, powerpc 8540, OpenWrt 19.07.

GRE_add_new_interface
GRE_advanced_settings
GRE_advanced_settings_continue
GRE_general_settings

Signed-off-by: Jan Betik <jan.betik@svine.su>

### DIFF
--- a/protocols/luci-proto-gre/Makefile
+++ b/protocols/luci-proto-gre/Makefile
@@ -1,0 +1,21 @@
+#
+# Based on luci-proto-ipip.
+# Credited author of luci-proto-ipip is Roger Pueyo Centelles <roger.pueyo@guifi.net>
+# Copyright 2016 Roger Pueyo Centelles <roger.pueyo@guifi.net>
+#
+# Modified by Jan Betik <jan.betik@svine.su>
+# Copyright 2020 Jan Betik <jan.betik@svine.su>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Support for GRE tunnels (RFC2784)
+LUCI_DEPENDS:=+gre
+
+PKG_MAINTAINER:=Jan Betik <jan.betik@svine.su>
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/protocols/luci-proto-gre/htdocs/luci-static/resources/protocol/gre.js
+++ b/protocols/luci-proto-gre/htdocs/luci-static/resources/protocol/gre.js
@@ -1,0 +1,96 @@
+'use strict';
+'require form';
+'require network';
+'require tools.widgets as widgets';
+
+network.registerPatternVirtual(/^gre4-.+$/);
+
+return network.registerProtocol('gre', {
+	getI18n: function() {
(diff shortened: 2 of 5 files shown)


feckert:
Here are some remarks.

- We also have the option multicast https://github.com/openwrt/openwrt/blob/master/package/network/config/gre/files/gre.sh#L263

- The option weakif is only selectable if ip6addr is empty https://openwrt.org/docs/guide-user/network/tunneling_interface_protocols#common_options_for_gre_protocols
They are mutually exclusive

- On this line (on behalf of the others) https://github.com/openwrt/luci/pull/3832/files#diff-4d9c2123127a4fc22d9ad4be7cf12950R71 we could change the message that we await a valid tos hex value?

Otherwise :+1:

mbargo23:
Hello,
>     * We also have the option multicast https://github.com/openwrt/openwrt/blob/master/package/network/config/gre/files/gre.sh#L263

What does this option do?


>     * The option weakif is only selectable if ip6addr is empty https://openwrt.org/docs/guide-user/network/tunneling_interface_protocols#common_options_for_gre_protocols
>       They are mutually exclusive

So the option weakif should be visible/active only if no IPv6 wan address is available?

>     * On this line (on behalf of the others) https://github.com/openwrt/luci/pull/3832/files#diff-4d9c2123127a4fc22d9ad4be7cf12950R71 we could change the message that we await a valid tos hex value?

I agree. According the man page for ip-tunnel, the possible values are 
`TOS  := { STRING | 00..ff | inherit | inherit/STRING | inherit/00..ff }`
so I should rewrite the whole validate function to reflect the man page.

feckert:
> > ```
> > * We also have the option multicast https://github.com/openwrt/openwrt/blob/master/package/network/config/gre/files/gre.sh#L263
> > ```
> 
> What does this option do?

Please have a look https://www.packet.com/resources/guides/gre-mutlicast/

> > ```
> > * The option weakif is only selectable if ip6addr is empty https://openwrt.org/docs/guide-user/network/tunneling_interface_protocols#common_options_for_gre_protocols
> >   They are mutually exclusive
> > ```
> 
> So the option weakif should be visible/active only if no IPv6 wan address is available?

I would see it that way

> > ```
> > * On this line (on behalf of the others) https://github.com/openwrt/luci/pull/3832/files#diff-4d9c2123127a4fc22d9ad4be7cf12950R71 we could change the message that we await a valid tos hex value?
> > ```
> 
> I agree. According the man page for ip-tunnel, the possible values are
> `TOS := { STRING | 00..ff | inherit | inherit/STRING | inherit/00..ff }`
> so I should rewrite the whole validate function to reflect the man page.

In the openwrt page this option is a string. Not sure which is better. Depends on how sophisticated the validate function is to validate all cases.

jow-:
Merged via 2ced86048c7006f3637ed38ceeb3d21fddbb93e2 (duplicate Signed-off-by removed from commit). Thanks!

PR #3869, opened 2020-04-10
luci.mk: correct titles of zh_Hans and zh_Hant

dengqf6:
zh_Hans: 简体中文 (Chinese Simplified)
zh_Hant: 繁體中文 (Chinese Traditional)


hnyman:
cc @yousong

PR #3875, opened 2020-04-10
use translatef instead of translate

ce-4:
This hopefully fixes translation strings to allow different word order due to other foreign grammar rules (e.g. German).

### DIFF
--- a/applications/luci-app-simple-adblock/luasrc/model/cbi/simple-adblock.lua
+++ b/applications/luci-app-simple-adblock/luasrc/model/cbi/simple-adblock.lua
@@ -77,7 +77,7 @@ end
 
 local statusTable = {}
 local errorTable = {}
-statusTable["statusNoInstall"] = packageName .. translate("is not installed or not found")
+statusTable["statusNoInstall"] = translatef("%s is not installed or not found", packageName)
 statusTable["statusStopped"] = translate("Stopped")
 statusTable["statusStarting"] = translate("Starting")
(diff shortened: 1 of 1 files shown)


jow-:
Syntactically it looks fine. Could you runtime test it?

ce-4:
Yes, but that would have to wait til tomorrow, i have a box with 19.07.2 at hand.

stangri:
Runtime tested on 18.06.8/theme-OpenWrt and it runs just fine.

Since the signoff is missing, should we wait for @ce-4 to push the updated commit or should I create separate PRs for master/19.07/18.06 branches now?

ce-4:
Sorry for creating additional work - I have not uploaded a signing key yet, will do that now. If you don't mind you can create the PRs or wait a bit and i will sign it.

I have also checked it on a 19.07.2, looks good to me. I noticed that the value in the brackets of the tmpfsVersion value is empty with both versions of simple-adblock.lua on my box (but i just started using simple-adblock):
"Simple AdBlock Settings
Service Status []"

hnyman:
> I have not uploaded a signing key yet

No need for that.
Just add the required "Signed-off-by:" line to your commit message.
Amend the commit, and force push the PR branch.

hnyman:
Note that the commit title is also wrong. It should atrat with the app name:

`luci-app-simple-adblock: use translatef instead of translate`

stangri:
@hnyman should I submit the patches for 19.07 and 18.06 branches?

PR #3694, opened 2020-02-28
luci-base: Show used memory instead of Free

howard0su:
This makes sure solid/blue bar shows the used memory. Fix #3682 
 
Signed-off-by: Howard Su <howard0su@gmail.com>

PR #3906, opened 2020-04-13
luci-base: Show used memory instead of Free

Slenkis:
This makes sure solid/blue bar shows the used memory.
Copy from this master pull request:  https://github.com/openwrt/luci/pull/3694
Signed-off-by: Slenkis


Slenkis:
хуйло узкоглазое смогло пропихнуть эту залупу, а меня нахуй послали

PR #3915, opened 2020-04-14
luci-app-adblock: Update dnsreport.js

whitevision1:
fixes bug in report filtering when searching for host name that contains capital letters


hnyman:
No explanation, no signed-off-by
See https://github.com/openwrt/luci/blob/master/CONTRIBUTING.md

whitevision1:
fixed

PR #3927, opened 2020-04-18
luci-app-dawn: Rewrite of Dawn Hearing Map in Table Form

JonnyTischbein:
Doing a rewrite of the luci dawn hearing map in fancy table form.

Screenshot_2020-04-18_10-12-22-2

Thanks to @PolynomialDivision for the work

### DIFF
--- a/applications/luci-app-dawn/luasrc/model/cbi/dawn/dawn_hearing_map.lua
+++ b/applications/luci-app-dawn/luasrc/model/cbi/dawn/dawn_hearing_map.lua
@@ -6,7 +6,6 @@ s = m:section(NamedSection, "__hearingmap__")
 function s.render(self, sid)
     local tpl = require "luci.template"
     tpl.render_string([[
-        <style>.hearing_table tr{ border-bottom: 1px solid grey;} tr:nth-child(even){background-color: lightgrey;}</style>
         <%
         local utl = require "luci.util"
         local status = require "luci.tools.ieee80211"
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
Looks awesome! Thank you.

Please merge, since I do not have merge permissions.

dibdot:
@JonnyTischbein 
Has this been tested with all LuCI standard themes? 
Why do we still use deprecated table-syntax with hardcoded styles? 
I would suggest to better use the common div-structure with a separate css file, e.g.
```
<div class="cbi-section-node">
	<div class="table" id="...">
		<div class="tr table-titles">
			<div class="th"><%:...%></div>
			<div class="th"><%:...%></div>
			<div class="th"><%:...%></div>
[...]
```

JonnyTischbein:
Originally I wanted to use something like bootstrap grids but I didn't know that there are classes like that.

I'll to replace it

JonnyTischbein:
Works good so far.

@dibdot Do you know if there is an equivalent to rowspan ? I didn't found something like that in the css file. 
Is there documentation for the div table ?

JonnyTischbein:
Current changes with div tables
I don't know which other themes I should test. Where can I choose them ?

 Screenshot_2020-04-18_12-36-57-2

dibdot:
@JonnyTischbein that looks much better - many thanks!
The theme selection is placed in the system menu, at least you should test Bootstrap (most probably already done), OpenWrt & OpenWrt2020.

Thanks again for your efforts!

JonnyTischbein:
Thanks! In my build only bootstrap is availalble and was used for develpoping.

dibdot:
Just install the other themes with opkg (luci-theme-*)

PolynomialDivision:
looksawesome

Looks awesome! :)

dibdot:
OK then, let's merge it.

PolynomialDivision:
Let's see if I can convert the other page to "div-structure with a separate css file", too. ;)

PR #3929, opened 2020-04-18
luci-app-dawn: stop adding not used config files

PolynomialDivision:
Fixes https://github.com/berlin-open-wireless-lab/DAWN/issues/64.

The app adds config files under /etc/config/ for the Hearing Map and the Network Overview. Change to SimpleForm fixes the issue.

### DIFF
--- a/applications/luci-app-dawn/luasrc/model/cbi/dawn/dawn_hearing_map.lua
+++ b/applications/luci-app-dawn/luasrc/model/cbi/dawn/dawn_hearing_map.lua
@@ -1,4 +1,4 @@
-m = Map("Hearing Map", translate("Hearing Map"))
+m = Map("dawn", "Hearing Map", translate("Hearing Map"))
 m.pageaction = false
 
 s = m:section(NamedSection, "__hearingmap__")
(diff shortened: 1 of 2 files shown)


dibdot:
Sorry, IMHO this isn't a fix, just another quirk ... ;-))
Root cause is that your map syntax is wrong, the correct Map syntax consist of 
`class Map (config, title, description)`, e.g.:

Wrong:
`m = Map("Hearing Map", translate("Hearing Map"))`

Correct:
`m = Map("dawn", "Hearing Map", translate("Hearing Map"))`

Please start reading some basics like https://github.com/openwrt/luci/wiki/CBI or maybe it's better to migrate this directly to the new JS API.

I've also noticed that this package implements a new toplevel menu called "DAWN" - we should better place this under the Services menu!? Maybe others with more experience chime in here ...

PolynomialDivision:
> I've also noticed that this package implements a new toplevel menu called "DAWN" - we should better place this under the Services menu!? Maybe others with more experience chime in here ...

I started this App just for debugging purposes and now it seems that people start using the stuff. ^^
Yeah we can switch it somewhere else. Services menue sounds fine. But could you merge to prevent that config adding stuff?

dibdot:
OK, let's merge it. Please note the DCO check always fails (see above error Details), could you please doublecheck that with your next PR as well? Thanks!

PolynomialDivision:
> OK, let's merge it. Please note the DCO check always fails (see above error Details), could you please doublecheck that with your next PR as well? Thanks!

Someone said I should not use `Polynomialdivision <vincent@systemli.org>`, since this is not my real name? That is why the check fails?

PR #3934, opened 2020-04-19
Allow to disable descriptions row in TableSection + descriptions render fix

namedun:
Add 'nodescriptions' property to the TableSection class that allows to disable displaying table header row with descriptions.

Also fixed descriptions row rendering at actions column for some cases.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2421,7 +2421,7 @@ var CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection.p
 						(typeof(opt.width) == 'number') ? opt.width+'px' : opt.width;
 			}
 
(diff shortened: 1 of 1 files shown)


jow-:
Hm, why do we need this extra property? The code is supposed to simply not render the description row if there are no option descriptions.

namedun:
Some options may have quite long descriptions, while others may have no descriptions at all. In general, a table with a description row may not look pretty enough in some cases.

I think it would be a good idea to be able to disable the display of the description row in such cases.

jow-:
Why not simply removing the descriptions of the options then? Or are you talking about modal options / GridSections here?

jow-:
Sorry, hit the wrong button.

namedun:
Yes, I want to be able to remove descriptions only from the table and not from the modal window. Descriptions can be useful and needed in the modal window, but are not needed in the table view in some cases.

jow-:
Thanks, merged.

PR #3935, opened 2020-04-19
luci_app_statistics: add syslog plugin

tobiaswaldvogel:
By default the output of collectd is written to the syslog unfiltered,
which is spamming it with a lot of 'notice' entries,
especially when collecting data from other devices.
With the syslog plugin you can specify the log and notify level,
and usually warning is a suitable default setting.

Contents of this change:
- Added dependency to the syslog plugin to pull it in automatically
- Entry to the config file with default settings enabled and level set to warning
- Configuration dialog for luci

Compiled: ar71xx / archer c7 v2 / r13025
Tested: ar71xx / archer c7 v2 / r13025

Signed-off-by: Tobias Waldvogel <tobias.waldvogel@gmail.com>


hnyman:
In "normal" single router usage with typical stats plugins, collectd in luci_statistics produces almost nothing to the system log. There are a few lines at startup, but nothing since then.

What is your use case? What triggers the log spam that you complain about? collectd network stuff?
Do you have example you that log spam?

I don't see any immediate need to force the dependency to all users of luci_statistics.
An alternative might be to decrease the verbosity of the culprit component.

tobiaswaldvogel:
I'm getting many repeated entries of this kind:

Sun Apr 19 22:14:47 2020 daemon.err collectd[5915]: Filter subsystem: Built-in target `write': Dispatching value to all write plugins failed with status -1.
Sun Apr 19 22:14:47 2020 daemon.err collectd[5915]: Filter subsystem: Built-in target `write': Some write plugin is back to normal operation. `write' succeeded

I am collecting data from 8 different devices, including NAS and 2 Android TV boxes. The syslog plugin seems to me  the obvious plugin for writing to the syslog and it is just 3k size. But I also understand if you  prefer to make it optional rather than adding a dependecy.

feckert:
@hnyman I have the same problem. In the log all messages are errors because of this line https://github.com/openwrt/packages/blob/master/utils/collectd/files/collectd.init#L349 (I am not sure if this line will be used because luci-app-statistics does hijack /etc/init.d/collectd).

Can't we check if the syslog plugin is installed and configured? If so, collected will output the messages with the correct loglevel as defined by the collectd and not all with error.

tobiaswaldvogel:
As it seems that the preference not to add a dependency on an additional plugin I removed it from the change. The settings are still added to the config file, so if the plugin is installed it will be enabled with log and notify level set to warning.
I still think that it makes more sense to add the dependency because I'm afraid that others won't figure out that there is already an existing solution (= installing the syslog plugin).
But I'm also happy with just contributing the luci integration.

applications/luci-app-statistics/root/etc/config/luci_statistics
```diff
@@ -42,6 +42,11 @@ config statistics 'collectd_logfile'
 config statistics 'collectd_network'
 	option enable '0'
 
+config statistics 'collectd_syslog'
+   option enable '1'
```
hnyman, line 46: This should be 0 in the shipped default config, so that there will be no error if there is no plugin.

For that reason only the plugins that are included by default, have 1 as value.

applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/syslog.js
```diff
@@ -0,0 +1,35 @@
...
+
+return baseclass.extend({
+   title: _('Syslog Plugin Configuration'),
+   description: _('The SysLog plugin receives log messages from the daemon and dispatches them to syslog.'),
+
+   addFormOptions: function(s) {
+       var o, ss;
+
+       o = s.option(form.Flag, 'enable', _('Enable this plugin'));
+       o.default = '1';
```
hnyman, line 13: This should be 0 (and the line can then be omitted).

tobiaswaldvogel:
Thanks for the review: Absolutely right, when leaving enabled to '1' stat-genconfig would always include the plugin in the config file, which would break the startup of collectd if the plugin is not installed.
I adapted your changes.

feckert:
@hnyman I would like to take up again the problem with the syslog. We are not talking about luci-app-statistics but about the collectd package in the package feed.
I have the following output here.
```
Mon Apr 27 12:37:05 2020 daemon.err collectd[27584]: Exiting normally.
Mon Apr 27 12:37:05 2020 daemon.err collectd[27584]: collectd: Stopping 2 read threads.
Mon Apr 27 12:37:05 2020 daemon.err collectd[27584]: collectd: Stopping 5 write threads.
Mon Apr 27 12:37:05 2020 daemon.err collectd[27584]: rrdtool plugin: Shutting down the queue thread.
Mon Apr 27 12:37:05 2020 daemon.err collectd[29113]: plugin_load: plugin "threshold" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.err collectd[29113]: plugin_load: plugin "exec" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.err collectd[29113]: plugin_load: plugin "lua" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.err collectd[29113]: Lua plugin: File "/usr/lib/collectd-mod-lua/cellular.lua" loaded successfully
Mon Apr 27 12:37:05 2020 daemon.err collectd[29113]: plugin_load: plugin "df" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.info collectd[29113]: plugin_load: plugin "syslog" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.info collectd[29113]: plugin_load: plugin "rrdtool" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.notice collectd[29113]: rrdtool plugin: RRASingle = true: creating only AVERAGE RRAs
Mon Apr 27 12:37:05 2020 daemon.info collectd[29113]: plugin_load: plugin "load" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.info collectd[29113]: plugin_load: plugin "memory" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.info collectd[29113]: plugin_load: plugin "interface" successfully loaded.
Mon Apr 27 12:37:05 2020 daemon.info collectd[29113]: Initialization complete, entering read-loop.
```
The problem is the error level. This is because the output is redirected from the collectd by init `procd_set_param_stderr`. When I load the syslog plugin then it is back to the correct loglevel as defined in the collectd. Ich would like to change this.
If we have selected and installed the syslog then this should be loaded at first. Then we have a clean syslog with the right output. I think that the level error is misleading, especially since this is not an error on the startup process!

hnyman:
I am not sure if I catch your exact meaning of "We are not talking about luci-app-statistics but about the collectd package in the package feed". Do you just mean the obvious thing about the collectd package location, or do you mean the context "collectd without LuCI"?

> If we have selected and installed the syslog then this should be loaded at first. Then we have a clean syslog with the right output.

Here (in LuCI statistics side) the stat-genconfig script actually prioritises certain config sections & plugins for collectd.

https://github.com/openwrt/luci/blob/551d839f90853b58f81d124c77442009d4df0cb4/applications/luci-app-statistics/root/usr/bin/stat-genconfig#L315-L321

In practice, the general collectd section goes first to the native collectd conf file, and the possible logfile goes next. I think that adding `section(syslog)` there in between would enable it to be prioritised also, in the LuCI statistics context. 

But that would cover only LuCI statistics usage of collectd, not the generic "collectd for other purposes" usage.  If you want to fix also "collectd for other purposes without LuCI", it would require changes on the packages feed side, or possibly to the collectd sources directly.

feckert:
> I am not sure if I catch your exact meaning of "We are not talking about luci-app-statistics but about the collectd package in the package feed". Do you just mean the obvious thing about the collectd package location, or do you mean the context "collectd without LuCI"?

I mean collectd without LuCI so **no** luci-app-statistics

> possibly to the collectd sources directly.

I think this is not an option.

I think I will open a new issue (feature request) for the collectd package in the package feed.
I will share my thoughts there.

PR #3964, opened 2020-04-24
Correct link for brand / logo

arrmo:
In the Material theme, clicking on the logo takes you to the status page - as it should, makes sense. But in Bootstrap, it doesn't do anything => just pulling the same href from Material to Bootstrap, so the link works in Bootstrap also.

Thanks!


arrmo:
Arrgh! The exact commands in DCO do not seem to work. I keep getting,
```
To https://github.com/arrmo/luci.git
 ! [rejected]            bootstrapBrand -> bootstrapBrand (stale info)
error: failed to push some refs to 'https://github.com/arrmo/luci.git'
```
Help please?!?!

PR #3965, opened 2020-04-25
Bootstrap brand

arrmo:
In the Material theme, clicking on the logo takes you to the status page - as it should, makes sense. But in Bootstrap, it doesn't do anything => just pulling the same href from Material to Bootstrap, so the link works in Bootstrap also.

Thanks!

### DIFF
--- a/themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm
+++ b/themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm
@@ -41,7 +41,7 @@
 		<header>
 			<div class="fill">
 				<div class="container">
-					<a class="brand" href="#"><%=boardinfo.hostname or "?"%></a>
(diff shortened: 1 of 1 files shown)


hnyman:
Not sure if this will be accepted, but in any case you should squash your commits, so that there is just one commit left. (Now you have 5 commits, including 3 merges, for an one-line change.)

You can squash the commits with git rebase, and then you can force push to update the PR branch.

The commit message should give some explanation of the change, which explanation is not mainly in the PR text (that does not get stored to git).

arrmo:
> Not sure if this will be accepted, but in any case you should squash your commits, so that there is just one commit left. (Now you have 5 commits, including 3 merges, for an one-line change.)

Agreed! Was trying / going to do that ... or revert them here, clean it up. But until I can get past the "stale info" message I can't. Dang it, can't figure out how to fix that one (first).

> The commit message should give some explanation of the change, which explanation is not mainly in the PR text (that does not get stored to git).

OK, will add more there. I have seen complaints before about too much detail there, but can add it in - no issue at all!

arrmo:
OK, will close / delete this branch (on my repository), try to get it cleane up and re-submit.

Thanks!

PR #3967, opened 2020-04-25
Update href for Bootstrap "brand"

arrmo:
In the Material theme, clicking on the logo takes you to the status page.
This seems logical, and is very helpful. But in Bootstrap, it doesn't
do anything, so just pulling the same href from Material to Bootstrap,
so the link works in Bootstrap as well.

### DIFF
--- a/themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm
+++ b/themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm
@@ -41,7 +41,7 @@
 		<header>
 			<div class="fill">
 				<div class="container">
-					<a class="brand" href="#"><%=boardinfo.hostname or "?"%></a>
+					<a class="brand" href="<% if luci.dispatcher.context.authsession then %><%=url('admin/status/overview')%><% else %>#<% end %>"><%=boardinfo.hostname or "?"%></a>
 					<ul class="nav" id="topmenu" style="display:none"></ul>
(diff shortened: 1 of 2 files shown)


arrmo:
BTW, should this same "feature" be applied to the new openwrt-2020 theme? Thinking it would be nice to have this work (i.e. clicking on upper left corner, name or "icon" ... to the status page).

Thanks!

jow-:
This will not work for installations that don't use `luci-mod-status`. Can't we simply link to `/` unconditonally without checking the auth state?

arrmo:
Yep! Changed it on the router here, to check it out ... and yes, can just `href` to `/`. Want me to change it to that? I thought you wanted the rest of the `href`, but no issue here!

And do a similar thing (separately) in openwrt-2020?

Thanks!

jow-:
Yep, it would be good if you could change it to href `/`. Feel free to do the change to OpenWrt 2020 in this PR too, using a separate commit.

arrmo:
Hopefully all done right, and as you're wanting (2 separate commits). FYI, I did build each of these, tested them - and working 😄.

Thanks!

arrmo:
Hi,

Just did an update here, realized this is still pending ... 😆. Is something missing, that I need to take care of?

Thanks!

jow-:
Merged via 400ecb08f08be9045d83ce74534401a4ccae3aee and 27fc7a8b4e30d96228fe169e0266bd43564d5eb3 - thanks!

PR #3978, opened 2020-04-30
luci-proto-openconnect: add usergroup option

GuLinux:
Add UI configuration option for openconnect user group.
See https://github.com/openwrt/packages/pull/11798#issuecomment-621511896


GuLinux:
Hi,
Just noticing that the other PR was merged already.

Cheers,
Marco

feckert:
Thanks merged

PR #3981, opened 2020-04-30
Change the translation of frequency in ntpc

strysd:
Change the translation of frequency in ntpc to semantically appropriate translation


strysd:
once close

Issue #3183, opened 2019-10-14
Add weblate service integration


aparcar:
Hi, for testing I created an OpenSource account on [weblink](https://hosted.weblate.org/projects/openwrt/) which gives a nice interface to translate the PO files. To improve the process I'd install the weblate service integration (also available for GitLab an other services) which automatically a) add newly added strings to their translation web interface and b) creates pull request once every now and then when updates via weblate happend. 
If anyone is against this approach please let me know. Right now it looks like only 52% percent is translated.

<a href="https://hosted.weblate.org/engage/openwrt/?utm_source=widget">
<img src="https://hosted.weblate.org/widgets/openwrt/-/svg-badge.svg" alt="Translation status" />
</a>

feckert:
I did a registration with github and proposed a change on a translation.
I don't want to get too deeply involved for now.
But to get a feel for how it works, can you please explain to me how this change I have made then goes into the master branch on github?

aparcar:
I played a bit with the tool and it works surprisingly good!

From what I understand it is possible to manually trigger a commit (like https://github.com/openwrt/luci/pull/3190/) or set it to create a PR every x hours based on the recent changes. Instead of PRs we could also allow it to directly commit those changes, however I'd leave this off for now.

Cool thing is we can also host it on our own if weblate changes their policies, which I don't see coming anytime soon but still a requirement I'd say.

feckert:
Thanks for clarification.
I think directly commit is not an option.

I have done some translation in luci-app-statistics and saved this.
CPU-Frequenz and Adressfamilie
But nothing happens.
No creation of a new pullrequest in the openwrt/luci :-(
Do I miss something?

aparcar:
Currently there are the two options to auto commit every 24h or perform it manually - as I just did. This however ended up in a somewhat messy PR as people started to translate also Portuguese and Russian files.

I set it up that from now on PRs are squashed by author, meaning whatever you do in 24h will become a single commit.

aparcar:
It is possible to add the weblate git repository via 

    git remote add weblate https://hosted.weblate.org/git/openwrt/luci/

jow-:
Seems to work well. Please continue with this in case further changes are needed.

hnyman:
What is the mechanism here?
I have received ~100 email during the night from PR 3197...

```
Subject: Re: [openwrt/luci] Update from Weblate (#3197)
From: Weblate (bot)
To: openwrt/luci
Cc: Push
[...]

PR #3982, opened 2020-04-30
Change the translation of frequency in ntpc to semantically appropriate translation

strysd:
Different words are used in Japanese for the frequency of radio waves and  the frequency of repeated work.
And I add 1 translation. 

Signed-off-by: Satoru Yoshida <satoruyoshida@php.net>


feckert:
Please add/change the I18N translation with weblat
Please have a look at the pinned issue https://github.com/openwrt/luci/issues/3183

PR #4007, opened 2020-05-04
luci-mod-system: update LED i18N trigger translation

feckert:
@hnyman Is this OK from you?
With this pullrequest I18N translation should be adapted, so it is clear what this trigger does.


hnyman:
If that is just text, looks ok to me.

feckert:
@hnyman yes that are only translation changes.
thanks for watching

hnyman:
Please also run i18n-sync and then commit the necessary changes, so that the change actually gets propagated to the luci-base .pot file

feckert:
Sorry I am not on my workstation. My first point tomorrow is to update
this. Thanks for the hint.

Hannu Nyman <notifications@github.com> schrieb am Mo., 4. Mai 2020, 17:12:

hnyman:
PS. one practical advice:  to avoid merge errors in weblate, or at least decreae their probability, please merge the possible weblate PR first.

PR #4016, opened 2020-05-05
[RFT] Backports for 19.07

jow-:
This is a series of backports I intend to include in 19.07.3 - I'd appreciate if you could give it a try to check if anything breaks.

@hnyman @stangri @feckert


hnyman:
I compiled 19.07 with this PR included, and browsed through pretty much all my LuCI pages.

Otherwise everything seemed ok at the first glance, but nlbwmon did not show graphs. And only the first tab showed the table.

Firefox shows:
image

I did not try to debug that further, but it might be something that you might need to check.

jow-:
Thanks for testing @hnyman - will look into nlbwmon

PR #4018, opened 2020-05-06
luci-app-commands: show hint in status page if no entry defined yet

he-ma:
If no commands are defined yet, you only see an empty box, not knowing whether there happened an error not displaying anything, machine still busy, ...

This patch adds the standard message "This section contains no values yet" for better user guidance.


jow-:
See comments in #4019

jow-:
The markup is still not quite correct

PR #4019, opened 2020-05-06
luci-app-wireguard: show hint in status page if no entry defined yet

he-ma:
If nothing is configured yet for wireguard, you only see an empty box, not knowing whether there happened an error not displaying anything, machine still busy, ...

This patch adds the standard message "This section contains no values yet" for better user guidance.

### DIFF
--- a/applications/luci-app-wireguard/luasrc/view/wireguard.htm
+++ b/applications/luci-app-wireguard/luasrc/view/wireguard.htm
@@ -182,6 +182,17 @@
 <h2>WireGuard Status</h2>
 
 <div class="cbi-section">
+
+<% if next(data) == nil then %>
+	<div class="table cbi-section-table">
+		<div class="tr cbi-section-table-row">
+			<p>
+				<em><%:This section contains no values yet%></em>
+			</p>
+		</div>
+	</div>
+<% end %>
+
 <%-
 local ikey, iface
 for ikey, iface in pairs(data) do


castillofrancodamian:
Something missing from this status page is the translatable title.

he-ma:
> Something missing from this status page is the translatable title.

What do you mean?
It's just the original page only extended for the hint.

applications/luci-app-wireguard/luasrc/view/wireguard.htm
```diff
@@ -182,6 +182,18 @@
 <h2>WireGuard Status</h2>
```
feckert, line 182: @he-ma I think he means that

he-ma: Well, but this is not newly indroduced by me and has been in the original ever since.

applications/luci-app-wireguard/luasrc/view/wireguard.htm
```diff
@@ -182,6 +182,18 @@
...
 
 <div class="cbi-section">
+
+<% if next(data) == nil then %>
+	<div class="cbi-section">
+		<div class="table cbi-section-table">
+			<div class="tr cbi-section-table-row" align="center">
```
jow-, line 189: The `align` attribute does not make sense here, don't use it.

applications/luci-app-wireguard/luasrc/view/wireguard.htm
```diff
@@ -182,6 +182,18 @@
...
+				<br><em><%:This section contains no values yet%></em><br>
```
jow-, line 190: Please don't use redundant `<br>` tags for padding. This is also invalid XHTML.

applications/luci-app-wireguard/luasrc/view/wireguard.htm
```diff
@@ -182,6 +182,18 @@
...
+				<tr>&nbsp;</tr>
```
jow-, line 191: This markup makes no sense at all.

applications/luci-app-wireguard/luasrc/view/wireguard.htm
```diff
@@ -182,6 +182,18 @@
 <h2>WireGuard Status</h2>
 
 <div class="cbi-section">
+
+<% if next(data) == nil then %>
+	<div class="cbi-section">
```
jow-, line 187: Why do we nest yet another `<div class="cbi-section">` here?

jow-:
The HTML code looks a bit like cargo culting to me. Can't we simply add something like the code below?

```
<p>
  <em><%:This section contains no values yet%></em>
</p>
```

castillofrancodamian:
> > Something missing from this status page is the translatable title.
> 
> What do you mean?
> It's just the original page only extended for the hint.

Thats what I refer:

2020-05-07_17h16_54

hnyman:
> Thats what I refer:

Seems to be without translation tags:
https://github.com/openwrt/luci/blob/master/applications/luci-app-wireguard/luasrc/view/wireguard.htm#L182

castillofrancodamian:
> > Thats what I refer:
> 
> Seems to be without translation tags:
> https://github.com/openwrt/luci/blob/master/applications/luci-app-wireguard/luasrc/view/wireguard.htm#L182

Justly 😄

feckert:
I will merge this and fix the translation in a separate commit in a few minutes.

PR #4026, opened 2020-05-07
luci-app-transmission: add the read permissions for widgets

brvphoenix:
Currently, the `UserSelect` and `GroupSelect` widgets provided by widgets.js have been broken by the loss of read permissions.

Before:
before

After:
after

Signed-off-by: Van Waholtz <vanwaholtz@gmail.com>

PR #4028, opened 2020-05-07
 luci-app-frpc: add uci file read permissions

zhanhb:
Add required permission


feckert:
I could not see in the code that you need passwd or group permissions.
Can you show me this in frpc code?

zhanhb:
image

https://github.com/openwrt/luci/blob/e05c219b24645f27e8d0adc2a523d610392246a3/applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js#L11-L12

brvphoenix:
@feckert 
https://github.com/openwrt/luci/pull/4026
This is caused by the loss of read permissions in the `luci-base`. All packages using `UserSelect` and `GroupSelect` widgets have this issue.

feckert:
Got it
Thanks

PR #4036, opened 2020-05-08
luci-app-commands: do not escape pipe on cmd line

h4de5:
Signed-off-by: Andi Peh <h4de5@gmx.at>


feckert:
Thanks for your report.
Done with https://github.com/openwrt/luci/commit/6d9a0488d65413743c7bc1d0bff5fef6b9713b49

PR #4037, opened 2020-05-09
luci-app-hd-idle: convert to client side rendering

brvphoenix:
Signed-off-by: Van Waholtz <vanwaholtz@gmail.com>


applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
```diff
@@ -0,0 +1,44 @@
...
+		o = s.option(form.Value, 'disk', _('Disk'));
+		devs.forEach(function(dev) {
+			o.value(dev.name);
+		});
+
+		o = s.option(form.Value, 'idle_time_interval', _('Idle time'));
+		o.default = 10;
+
+		o = s.option(form.ListValue, 'idle_time_unit', _('Idle time unit'));
+		o.value('seconds', _('s'));
```
jow-, line 36: Since these translation labels are very short and ambiguous, could you add a translation context by passing a second argument? Something like `_('s', 'Abbreviation for seconds')` - the context will not be visible to the user but it'll be part of the `*.po` files to give translators a hint on what `s` is supposed to mean.

applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+'require form';
+'require fs';
+'require view';
+
+return view.extend({
+	load: function() {
+		return fs.list('/dev').then(function(devs) {
+			return devs.filter(function(dev) {
+				return dev.name.match(/^[sh]d[a-z]$/);
```
jow-, line 10: Maybe filter for `dev.type == "file"` too, just to be on the safe side.

jow-:
Good work! Just two little nitpicks as written above.

applications/luci-app-hd-idle/po/ca/hd_idle.po
```diff
@@ -15,52 +15,74 @@ msgstr ""
...
+msgid "s"
+msgstr ""
+
+# Hours
+#~ msgid "h"
+#~ msgstr "h"
+
+# Minutes (not minimum)
+#~ msgid "min"
+#~ msgstr "min"
```
brvphoenix, line 85: It seems like that the `i18n-update.pl` can't handle this properly after adding the "msgctxt".

jow-: Thats okay, the translations and source strings were the same anyway, so nothing lost.

brvphoenix:
Updated.

applications/luci-app-hd-idle/po/ca/hd_idle.po
brvphoenix: OK, thanks.

applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
brvphoenix: Seems the hdd devices should filter for `dev.type == 'block'`?

jow-:
Merged, thank you!

PR #4038, opened 2020-05-09
luci-app-acme: Update acme.sh URL, add support for challenge & domain alises.

4wrxb:
Signed-off-by: Will O'Neill <0100wrxb@gmail.com>

Maintainer: @tohojo 
Compile tested: Snapshot, default target (for makefile download of new URL)
Run tested: ramips_mt7621_DEVICE_ubnt_edgerouter-x (for shell scritp run)

Description:
1) The acme.sh project has moved on github to [acmesh-official](https://github.com/acmesh-official/acme.sh)
2) I have added some simple logic in the script to support challenge and domain aliases. These related functions allow using CNAME records from the main domain to point to a different domain where the script can modify text records.

I am preparing an associated packages commit with the script file changes.

### DIFF
--- a/applications/luci-app-acme/luasrc/model/cbi/acme.lua
+++ b/applications/luci-app-acme/luasrc/model/cbi/acme.lua
@@ -93,16 +93,25 @@ dom.datatype = "list(string)"
 
 dns = cs:option(Value, "dns", translate("DNS API"),
                 translate("To use DNS mode to issue certificates, set this to the name of a DNS API supported by acme.sh. " ..
-                          "See https://github.com/Neilpang/acme.sh/tree/master/dnsapi for the list of available APIs. " ..
+                          "See https://github.com/acmesh-official/acme.sh/wiki/dnsapi for the list of available APIs. " ..
                           "In DNS mode, the domain name does not have to resolve to the router IP. " ..
                           "DNS mode is also the only mode that supports wildcard certificates. " ..
                           "Using this mode requires the acme-dnsapi package to be installed."))
-dns.optional = true
 
 cred = cs:option(DynamicList, "credentials", translate("DNS API credentials"),
                  translate("The credentials for the DNS API mode selected above. " ..
-                           "See https://github.com/Neilpang/acme.sh/tree/master/dnsapi#how-to-use-dns-api for the format of credentials required by each API. " ..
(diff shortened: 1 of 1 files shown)


4wrxb:
Is there a build-test process for LUCI? I have only changed the lua file (and verified the interface on the running device"), but I don't see any other build checks to run.

4wrxb:
I'm not finding any details on a LUCI build... marking ready for review. Please let me know if I missed something.

applications/luci-app-acme/luasrc/model/cbi/acme.lua
```diff
@@ -93,16 +93,32 @@ dom.datatype = "list(string)"
...
+                "LUCI only supports one challenge alias per certificate."))
+
+calias.optional = true
+
+dalias = cs:option(Value, "dalias", translate("Domain Alias"),
+                translate("The domain alias to use for ALL domains. " ..
+                "See https://github.com/acmesh-official/acme.sh/wiki/DNS-alias-mode for the details of this process. " ..
+                "LUCI only supports one challenge domain per certificate."))
+
+dalias.optional = true
```
tohojo, line 122: Not sure I like the 'optional' markings; hiding fields has resulted in confusion before - see https://github.com/openwrt/packages/issues/11095 for an example...

4wrxb: The DNS checkbox itself is optional and all these options are unused unless that's checked. It felt inconsistent otherwise.

That said, it's up to you and how you feel the LUCI page should look. Feel free to make changes or let me know what you want and I'll push changes.

4wrxb: p.s. I am not sure how the internals work, but if it were possible to do something like luci-app-ddns where the configuration page changes based on a drop-down that may be clearer.

For ddns it's based on the providers, but in this case selecting http(s) vs dns challenge mode could make the relevant fields visible.

4wrxb: just saw your comments on the package PR, I'm pushing a removal of optional on all 4 (dns, creds I added, plus the two alias options I added).

This leaves only the webroot setting as hidden.

PR #4047, opened 2020-05-13
luci-proto-wireguard: fix value validation

feckert:
* add warning that `allowed_ips` must not be empty

* verify last base64 string symbol is an `=` sig
If this is not the case I get the following error during connection
```
Thu Apr 30 11:37:42 2020 daemon.notice netifd: wg1 (23647): Key is not the correct length or format: `3UW+WgCasdu1SEOngdwYZm2jc5ogAQx9GQAjCt520g01'
Thu Apr 30 11:37:42 2020 daemon.notice netifd: wg1 (23647): Configuration parsing error
```

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -124,7 +124,14 @@ return network.registerProtocol('wireguard', {
 
 		o = ss.option(form.DynamicList, 'allowed_ips', _('Allowed IPs'), _("Required. IP addresses and prefixes that this peer is allowed to use inside the tunnel. Usually the peer's tunnel IP addresses and the networks the peer routes through the tunnel."));
 		o.datatype = 'ipaddr';
-		o.rmempty = false;
+		o.validate = function(section, value) {
+			var opt = this.map.lookupOption('allowed_ips', section);
+			var ips = opt[0].formvalue(section);
+			if (ips.length == 0) {
+				return _('Value must not be empty');
+			}
+			return true;
+		};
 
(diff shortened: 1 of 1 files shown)


protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -10,6 +10,9 @@ function validateBase64(section_id, value) {
 	if (value.length != 44 || !value.match(/^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?$/))
 		return _('Invalid Base64 key string');
 
+	if (value['43'] != "=" )
+		return _('Invalid Base64 key string');
```
jow-, line 14: That looks wrong. `=` is a padding char, not every valid base64 string ends with a `=` char.

feckert: I did a `wg genkey` and I have not seen any other char then `=`  so I thought this is a must?
Maybe wireguard use this base64 diffrence? We could ask @zx2c4 if he could give us a hint.

jow-: Seems you're right, apparently a trailing `=` is mandatory:
https://github.com/WireGuard/wireguard-tools/blob/3bf1b64d44fdbe589ac145b61bd4b3aa62b625bf/src/encoding.c#L56

So fine with me then, but please use a numeric subscript (`value[43]` vs. `value['43']`)

feckert:
@jow- thanks for your review

zx2c4:
If you guys really want to nerd out, match against this regex:

```^[A-Za-z0-9+/]{42}[A|E|I|M|Q|U|Y|c|g|k|o|s|w|4|8|0]{1}=$```

This ensures that the key is actually 256-bits and not 257 or 258 bits.

PR #4051, opened 2020-05-14
luci-base: German translation: Use the uniform term "WLAN"

sch-m:
... instead of a mixture of "WLAN", "W-LAN", "Drahtlos" etc.


jow-:
Ack - thank you!

PR #4073, opened 2020-05-19
luci-app-dockerman: add package

feckert:
Since the last stable release of openwrt, we have now [docker](https://github.com/openwrt/packages/tree/master/utils/docker-ce) integrated in the master branch. That's great!! I use it successfully with my APU3. But I wanted to simplify the handling and felled over the following LuCI integration from [Lisaac](https://github.com/lisaac).

I wrote to him immediately to ask if he would be willing to integrate this feature into the LuCI master. And the answer was **yes**. To get this going I've opened this pullrequest.

I have pulled the sources from his repository and added 2 package.
* luci-lib-docker https://github.com/lisaac/luci-lib-docker
* luci-app-dockerman https://github.com/lisaac/luci-app-dockerman

I adapted the makefiles so that they work with the build system.
And I could load images from the dockerhub and start this docker containers.

Now the question is @jow- @hnyman how you feel about this. What I would like to point out is the application is still written in lua.


applications/luci-app-dockerman/Makefile
```diff
@@ -0,0 +1,12 @@
+include $(TOPDIR)/rules.mk
+
+PKG_LICENSE:=AGPL-3.0
+PKG_MAINTAINER:=lisaac <https://github.com/lisaac/luci-app-dockerman>
```
hnyman, line 4: Maintainer should be normal name & email  (so that e.g. messaging to maintainers can work).

applications/luci-app-dockerman/root/etc/config/dockerman
```diff
@@ -0,0 +1,10 @@
+config section 'local'
+        option socket_path '/var/run/docker.sock'
```
hnyman, line 2: Nitpicking, but spaces should be tabs here

applications/luci-app-dockerman/Makefile
```diff
@@ -0,0 +1,12 @@
...
+
+LUCI_TITLE:=LuCI Support for docker
+LUCI_DEPENDS:=+luci-lib-docker +docker-ce +ttyd
```
hnyman, line 7: If the required luci-lib-docker is x86_64 only, it might be good to add that restriction also here, so that casual browsers of source code will understand that easily.

Although I did not figure out why luci-lib-docker is marked x86_64 only... Not obvious to my one-minute browsing, especially as the docker-ce has been marked for (aarch64||arm||x86_64)

feckert: Thanks for the hint. I just remembered that in the beginning docker-ce had a dependecy on x86_64. 

 I will fix this

feckert:
@hnyman I will update your proposed change after my vacation.
@jow- Any objections, or can I merge it into the master after adjustment?

feckert:
@lisaac How does it look from your side now?
Please leave a comment.

lisaac:
@feckert 
Hi, I think luci-compat is required for LuCI 19.07+.
Also, we need PKG_VERSION.

ysc3839:
Who is maintaining `luci-app-dockerman` now? The code in this repo is not updated for a while compared to upstream repo. And I found a bug related to sub menu, I will investigate and open a new issue soon.

feckert:
I hoped that by integrating the app into the Luci feed, it would also be supported by @lisaac and the changes would also be incorporated here. Unfortunately, this was not the case ;-( and @lisaac continued alone in his own repo.

jow-:
Shall we drop it then, here?

feckert:
> Shall we drop it then, here?

I'm not sure about that yet.
I used this luci-app-dockerman from time to time.
Maybe we'll set it to @BROKEN for now?

Issue #4071, opened 2020-05-19
luci-base: show everything or show nothing


castillofrancodamian:
I see that some do not mind seeing the LuCI menus (not me either) when they are not logged in, but it does not show everything.
![Anotación 2020-05-18 234554](https://user-images.githubusercontent.com/47670442/82279075-0f413d80-9962-11ea-9471-f0ccf993c4d6.png)

castillofrancodamian:
![123](https://user-images.githubusercontent.com/47670442/82591154-46d40380-9b75-11ea-9f6a-1c1e993bafd2.png)

ptpt52:
this patch may fix this issue
```
diff --git a/applications/luci-app-wireguard/root/usr/share/luci/menu.d/luci-app-wireguard.json b/applications/luci-app-wireguard/root/usr/share/luci/menu.d/luci-app-wireguard
index 0cb87f3e2..a42677e1a 100644
--- a/applications/luci-app-wireguard/root/usr/share/luci/menu.d/luci-app-wireguard.json
+++ b/applications/luci-app-wireguard/root/usr/share/luci/menu.d/luci-app-wireguard.json
@@ -5,6 +5,9 @@
                "action": {
                        "type": "template",
                        "path": "wireguard"
+               },
+               "depends": {
+                       "acl": [ "luci-mod-status-index" ]
                }
        }
 }
```

castillofrancodamian:
Did you make a Pull Request?

castillofrancodamian:
👍👍👍

PR #4086, opened 2020-05-22
luci-app-advanced-reboot: bugfix: add ACL json

stangri:

### DIFF
--- a/applications/luci-app-advanced-reboot/README.md
+++ b/applications/luci-app-advanced-reboot/README.md
@@ -26,7 +26,7 @@ If your device is not in the list above, however it is a [dual-firmware device](
 
 ## Screenshot (luci-app-advanced-reboot)
 
-![screenshot](https://raw.githubusercontent.com/stangri/openwrt_packages/master/screenshots/luci-app-advanced-reboot/screenshot02.png "screenshot")
+![screenshot](https://cdn.jsdelivr.net/gh/stangri/openwrt_packages@master/screenshots/luci-app-advanced-reboot/screenshot02.png "screenshot")
 
 ## How to install
 
@@ -37,7 +37,7 @@ opkg update
 opkg install luci-app-advanced-reboot
 ```
 
-If the ```luci-app-advanced-reboot``` package is not found in the official feed/repo for your version of OpenWrt/LEDE Project, you will need to [add a custom repo to your router](https://github.com/stangri/openwrt_packages/blob/master/README.md#on-your-router) first.
+If the ```luci-app-advanced-reboot``` package is not found in the official feed/repo for your version of OpenWrt/LEDE Project, you will need to add a custom repo to your router following instructions on [GitHub](https://github.com/stangri/openwrt_packages/blob/master/README.md#on-your-router)/[jsDelivr](https://cdn.jsdelivr.net/gh/stangri/openwrt_packages@master/README.md#on-your-router) first.
 
 ## Notes/Known Issues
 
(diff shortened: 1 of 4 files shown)


stangri:
Do legacy apps even need the `.acl_depends` in the controller? If they do not, should I remove it from all my other legacy apps?

jow-:
@stangri - `acl_depends` annotations are strongly encouraged in master so LuCI knows which applications are available in which authentication state and to allow for fine grained multi user configuration in the long run.

Missing ACL annotations will lead to bugs like https://github.com/openwrt/luci/issues/4071

applications/luci-app-advanced-reboot/README.md
```diff
@@ -26,7 +26,7 @@ If your device is not in the list above, however it is a [dual-firmware device](
 
 ## Screenshot (luci-app-advanced-reboot)
 
-![screenshot](https://raw.githubusercontent.com/stangri/openwrt_packages/master/screenshots/luci-app-advanced-reboot/screenshot02.png "screenshot")
+![screenshot](https://cdn.jsdelivr.net/gh/stangri/openwrt_packages@master/screenshots/luci-app-advanced-reboot/screenshot02.png "screenshot")
```
dengqf6, line 29: `@master` can be removed if it is the default branch

stangri:
> * you should move all commands that allow write access to the system in any way (more general, any commands with side effects) into the `write` scope
Just to confirm, if I'm only using `dd` to read (`dd if=...` without `of=`) should it still be in the `write` section? Is there any way to limit the scope of `dd` in the ACL file to explicitly exclude `of=` (I mean without the helper executable)?

Also, I've asked in a PM on the forum, but this looks like a better place -- what do I need to run to append the template.pot with the localizable resources from the JSON file? And while we're on the subject, since it's way more than just UCI access, should the localizable text be different from `"Grant UCI access for luci-app-advanced-reboot"`?

stangri:
Where can I see the script running on build servers to confirm how the pot templates are generated from strings in acl.d json?

jow-:
> Just to confirm, if I'm only using dd to read (dd if=... without of=) should it still be in the write section? Is there any way to limit the scope of dd in the ACL file to explicitly exclude of= (I mean without the helper executable)?

Hm, nope - not with the currently supported fileglob syntax. I plan to eventually add support for full regex patterns, then you can specify something like `^dd if=[^ =]+` but right now it needs a helper executable to restrict things.

> Also, I've asked in a PM on the forum, but this looks like a better place -- what do I need to run to append the template.pot with the localizable resources from the JSON file?

You mean the descriptions? They should be covered since 0bc29706408c4cf8746917a6e4d4045397adca64 in master without any additional steps. I think this change isn't backported to 19.07 - however the only user facing code using these descriptions is `luci-app-acl` which is not available in 19.07 yet.

stangri:
> > Just to confirm, if I'm only using dd to read (dd if=... without of=) should it still be in the write section? Is there any way to limit the scope of dd in the ACL file to explicitly exclude of= (I mean without the helper executable)?
> 
> Hm, nope - not with the currently supported fileglob syntax. I plan to eventually add support for full regex patterns, then you can specify something like `^dd if=[^ =]+` but right now it needs a helper executable to restrict things.

I'll give it a go and ping you for review when ready, thanks.

> > Also, I've asked in a PM on the forum, but this looks like a better place -- what do I need to run to append the template.pot with the localizable resources from the JSON file?
> 
> You mean the descriptions? They should be covered since [0bc2970](https://github.com/openwrt/luci/commit/0bc29706408c4cf8746917a6e4d4045397adca64) in master without any additional steps. I think this change isn't backported to 19.07 - however the only user facing code using these descriptions is `luci-app-acl` which is not available in 19.07 yet.

Oh, bummer, I use the latest stable SDK to build my packages and I currently pull the `luci.mk` and `build` directory from the `openwrt-19.07` branch in luci before each build (which includes generating the pot template). Should I pull them from `master` branch of luci repo instead?

jow-:
Uhm, I'll backport the JSON translation stuff to 19.07, thats probably easier, even if nothing is using it yet.

PR #4087, opened 2020-05-22
generic: fix unit symbols and their formatting

getsnoopy:
This commit changes some of previously pending abbreviations to
their official unit symbols, and fixes some of the unit symbols
which had typographical errors. It also fixes the formatting of
some incorrectly formatted symbols to comply with SI, ISO, and IEC
standards.

Signed-off-by: Varun Varada <varuncvarada@gmail.com>

### DIFF
--- a/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
+++ b/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
@@ -294,8 +294,8 @@ function display(pattern)
 		currentDisplayRows.push([
 			name,
 			ver,
-			pkg.size ? '%.1024mB'.format(pkg.size)
-			         : (altsize ? '~%.1024mB'.format(altsize) : '-'),
+			pkg.size ? '%1024mB'.format(pkg.size)
+			         : (altsize ? '~%1024mB'.format(altsize) : '-'),
 			desc,
 			btn
 		]);
@@ -533,9 +533,9 @@ function renderDependencyItem(dep, info)
 		var text = pkg.name;
 
 		if (pkg.installsize)
-			text += ' (%.1024mB)'.format(pkg.installsize);
+			text += ' (%1024mB)'.format(pkg.installsize);
 		else if (pkg.size)
-			text += ' (~%.1024mB)'.format(pkg.size);
+			text += ' (~%1024mB)'.format(pkg.size);
 
 		li.appendChild(E('span', { 'data-tooltip': pkg.description },
 			[ text, ' ', pkgStatus(pkg, vop, ver, info) ]));
@@ -636,9 +636,9 @@ function handleInstall(ev)
 	    size;
 
 	if (pkg.installsize)
-		size = _('~%.1024mB installed').format(pkg.installsize);
+		size = _('~%1024mB installed').format(pkg.installsize);
 	else if (pkg.size)
-		size = _('~%.1024mB compressed').format(pkg.size);
+		size = _('~%1024mB compressed').format(pkg.size);
 	else
 		size = _('unknown');
 
@@ -662,7 +662,7 @@ function handleInstall(ev)
 		});
 
 	inst = E('p', {},
-		_('Require approx. %.1024mB size for %d package(s) to install.')
+		_('Require approx. %1024mB size for %d package(s) to install.')
 			.format(totalsize, totalpkgs));
 
 	if (deps) {
@@ -824,9 +824,9 @@ function handleRemove(ev)
 	    size, desc;
 
 	if (avail.installsize)
-		size = _('~%.1024mB installed').format(avail.installsize);
+		size = _('~%1024mB installed').format(avail.installsize);
 	else if (avail.size)
-		size = _('~%.1024mB compressed').format(avail.size);
+		size = _('~%1024mB compressed').format(avail.size);
 	else
 		size = _('unknown');
 
@@ -985,7 +985,7 @@ function updateLists(data)
 		    	.sort(function(a, b) { return a.mount > b.mount })[0] || { size: 0, free: 0 };
 
 		pg.firstElementChild.style.width = Math.floor(mount.size ? ((100 / mount.size) * mount.free) : 100) + '%';
-		pg.setAttribute('title', '%s (%.1024mB)'.format(pg.firstElementChild.style.width, mount.free));
+		pg.setAttribute('title', '%s (%1024mB)'.format(pg.firstElementChild.style.width, mount.free));
 
 		parseList(data[1], packages.available);
 		parseList(data[2], packages.installed);
--- a/applications/luci-app-wireguard/luasrc/view/wireguard.htm
+++ b/applications/luci-app-wireguard/luasrc/view/wireguard.htm
@@ -84,11 +84,11 @@
 		var seconds = (now.getTime() / 1000) - timestamp;
 		var ago = "";
 		if (seconds < 60) {
-			ago = parseInt(seconds) + '<%:s ago%>';
+			ago = parseInt(seconds) + '\u202f<%:s ago%>';
 		} else if (seconds < 3600) {
-			ago = parseInt(seconds / 60) + '<%:m ago%>';
+			ago = parseInt(seconds / 60) + '\u202f<%:min ago%>';
 		} else if (seconds < 86401) {
-			ago = parseInt(seconds / 3600) + '<%:h ago%>';
+			ago = parseInt(seconds / 3600) + '\u202f<%:h ago%>';
 		} else {
 			ago = '<%:over a day ago%>';
 		}
--- a/modules/luci-base/htdocs/luci-static/resources/cbi.js
+++ b/modules/luci-base/htdocs/luci-static/resources/cbi.js
@@ -630,8 +630,8 @@ String.prototype.format = function()
 						}
 
 						subst = (td > 0)
-							? String.format('%dd %dh %dm %ds', td, th, tm, ts)
-							: String.format('%dh %dm %ds', th, tm, ts);
+							? String.format('%d:%d:%d:%d', td, th, tm, ts)
+							: String.format('%d:%d:%d', th, tm, ts);
 
 						break;
 
@@ -641,7 +641,9 @@ String.prototype.format = function()
 
 						var i = 0;
 						var val = (+param || 0);
-						var units = [ ' ', ' K', ' M', ' G', ' T', ' P', ' E' ];
+						var units = (mf === 1024)
+							? [ ' ', ' Ki', ' Mi', ' Gi', ' Ti', ' Pi', ' Ei' ]
+							: [ ' ', ' k', ' M', ' G', ' T', ' P', ' E' ];
 
 						for (i = 0; (i < units.length) && (val > mf); i++)
 							val /= mf;
--- a/modules/luci-compat/luasrc/tools/webadmin.lua
+++ b/modules/luci-compat/luasrc/tools/webadmin.lua
@@ -9,42 +9,41 @@ local uci  = require "luci.model.uci"
 local ip   = require "luci.ip"
 
 function byte_format(byte)
-	local suff = {"B", "KB", "MB", "GB", "TB"}
+	local suff = {"B", "KiB", "MiB", "GiB", "TiB"}
 	for i=1, 5 do
 		if byte > 1024 and i < 5 then
 			byte = byte / 1024
 		else
-			return string.format("%.2f %s", byte, suff[i]) 
-		end 
+			return string.format("%.2f %s", byte, suff[i])
+		end
 	end
 end
 
 function date_format(secs)
-	local suff = {"min", "h", "d"}
 	local mins = 0
 	local hour = 0
 	local days = 0
-	
+
 	secs = math.floor(secs)
 	if secs > 60 then
 		mins = math.floor(secs / 60)
 		secs = secs % 60
 	end
-	
+
 	if mins > 60 then
 		hour = math.floor(mins / 60)
 		mins = mins % 60
 	end
-	
+
 	if hour > 24 then
 		days = math.floor(hour / 24)
 		hour = hour % 24
 	end
-	
+
 	if days > 0 then
-		return string.format("%.0fd %02.0fh %02.0fmin %02.0fs", days, hour, mins, secs)
+		return string.format("%.0f:%02.0f:%02.0f:%02.0f", days, hour, mins, secs)
 	else
-		return string.format("%02.0fh %02.0fmin %02.0fs", hour, mins, secs)
+		return string.format("%02.0f:%02.0f:%02.0f", hour, mins, secs)
 	end
 end
 
@@ -70,15 +69,15 @@ end
 
 function firewall_find_zone(name)
 	local find
-	
-	luci.model.uci.cursor():foreach("firewall", "zone", 
+
+	luci.model.uci.cursor():foreach("firewall", "zone",
 		function (section)
 			if section.name == name then
 				find = section[".name"]
 			end
 		end
 	)
-	
+
 	return find
 end
 
--- a/applications/luci-app-adblock/htdocs/luci-static/resources/view/adblock/overview.js
+++ b/applications/luci-app-adblock/htdocs/luci-static/resources/view/adblock/overview.js
@@ -273,8 +273,8 @@ return view.extend({
 		s.tab('sources', _('Blocklist Sources'), _('List of supported and fully pre-configured adblock sources, already active sources are pre-selected.<br /> \
 			<b><em>To avoid OOM errors, please do not select too many lists!</em></b><br /> \
 			List size information with the respective domain ranges as follows:<br /> \
-			&#8226;&#xa0;<b>S</b> (-10k), <b>M</b> (10k-30k) and <b>L</b> (30k-80k) should work for 128 MByte devices,<br /> \
-			&#8226;&#xa0;<b>XL</b> (80k-200k) should work for 256-512 MByte devices,<br /> \
+			&#8226;&#xa0;<b>S</b> (-10k), <b>M</b> (10k-30k) and <b>L</b> (30k-80k) should work for 128 MiB devices,<br /> \
+			&#8226;&#xa0;<b>XL</b> (80k-200k) should work for 256-512 MiB devices,<br /> \
 			&#8226;&#xa0;<b>XXL</b> (200k-) needs more RAM and Multicore support, e.g. x86 or raspberry devices.<br /> \
 			<p>&#xa0;</p>'));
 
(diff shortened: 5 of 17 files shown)


jow-:
Hi, thank you for your work on this. I'm fine with the spelling fixes but not the base conversations, this should be done in a separate commit / PR if at all.

hnyman:
I am not sure if I like the proposed changes.

Being an old-timer, megabyte has always been 1024x1024, and the proposed "mebibyte" is quite incomprehensible. Never heard about that...

Sure, there has always been some confusion, as the binary based units are naturally related to physical things like flash size (8 MB, not 8.4 MB or 8 MiB), while the some OEMs like hard disk manufacturers have still used the 10-based calc for advertised hard disk sizes.

Similarly, although the small "b" for bit and big "B" for byte has been the normal for ages, there are still people who do not know that. So the cosmetic changes in style of KByte-->kB feel like decreasing information content, although quite right be themselves.

getsnoopy:
@jow- I removed the base changes from this PR; I'll make them in a separate one.

@hnyman The prefixes "kilo-" and "mega-" have always meant powers of 1000, it's just that early computer scientists have used them incorrectly while the rest of the population has always known them to be what they actually mean, so the hard drive manufacturers have used the terms correctly as well. This confusion [has been corrected and is standard](https://en.wikipedia.org/wiki/Binary_prefix) now as of 1998. Same thing with _b_ instead of _bit_ for bit(s).

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -294,8 +294,8 @@ function display(pattern)
 		currentDisplayRows.push([
 			name,
 			ver,
-			pkg.size ? '%.1024mB'.format(pkg.size)
-			         : (altsize ? '~%.1024mB'.format(altsize) : '-'),
+			pkg.size ? '%1024mB'.format(pkg.size)
```
stangri, line 297: Should it not be `MiB`? Also would it not be better then to convert to Mb instead?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -294,8 +294,8 @@ function display(pattern)
...
+			         : (altsize ? '~%1024mB'.format(altsize) : '-'),
```
stangri, line 298: Should it not be `MiB`? Also would it not be better then to convert to Mb instead?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -533,9 +533,9 @@ function renderDependencyItem(dep, info)
 		var text = pkg.name;
 
 		if (pkg.installsize)
-			text += ' (%.1024mB)'.format(pkg.installsize);
+			text += ' (%1024mB)'.format(pkg.installsize);
```
stangri, line 536: Should it not be `MiB`? Also would it not be better then to convert to Mb instead?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -533,9 +533,9 @@ function renderDependencyItem(dep, info)
...
 		else if (pkg.size)
-			text += ' (~%.1024mB)'.format(pkg.size);
+			text += ' (~%1024mB)'.format(pkg.size);
```
stangri, line 538: Should it not be `MiB`? Also would it not be better then to convert to Mb instead?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -636,9 +636,9 @@ function handleInstall(ev)
 	    size;
 
 	if (pkg.installsize)
-		size = _('~%.1024mB installed').format(pkg.installsize);
+		size = _('~%1024mB installed').format(pkg.installsize);
```
stangri, line 639: Should it not be `MiB`? Also would it not be better then to convert to Mb instead?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -636,9 +636,9 @@ function handleInstall(ev)
...
 	else if (pkg.size)
-		size = _('~%.1024mB compressed').format(pkg.size);
+		size = _('~%1024mB compressed').format(pkg.size);
```
stangri, line 641: Should it not be `MiB`? Also would it not be better then to convert to Mb instead?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -662,7 +662,7 @@ function handleInstall(ev)
 		});
 
 	inst = E('p', {},
-		_('Require approx. %.1024mB size for %d package(s) to install.')
+		_('Require approx. %1024mB size for %d package(s) to install.')
```
stangri, line 665: Should it not be `MiB`? Also would it not be better then to convert to Mb instead?

applications/luci-app-wireguard/luasrc/view/wireguard.htm
```diff
@@ -84,11 +84,11 @@
...
 		var ago = "";
 		if (seconds < 60) {
-			ago = parseInt(seconds) + '<%:s ago%>';
+			ago = parseInt(seconds) + ' <%:s ago%>';
 		} else if (seconds < 3600) {
-			ago = parseInt(seconds / 60) + '<%:m ago%>';
+			ago = parseInt(seconds / 60) + ' <%:min ago%>';
 		} else if (seconds < 86401) {
-			ago = parseInt(seconds / 3600) + '<%:h ago%>';
+			ago = parseInt(seconds / 3600) + ' <%:h ago%>';
```
stangri, line 91: Is the extra space intentional?

modules/luci-base/htdocs/luci-static/resources/cbi.js
```diff
@@ -630,8 +630,8 @@ String.prototype.format = function()
 						}
 
 						subst = (td > 0)
-							? String.format('%dd %dh %dm %ds', td, th, tm, ts)
-							: String.format('%dh %dm %ds', th, tm, ts);
+							? String.format('%d d %d h %d min %d s', td, th, tm, ts)
+							: String.format('%d h %d min %d s', th, tm, ts);
```
stangri, line 634: What's the reason for extra spacing between the value and denominator?

modules/luci-compat/luasrc/tools/webadmin.lua
```diff
@@ -42,9 +42,9 @@ function date_format(secs)
 	end
 	
 	if days > 0 then
-		return string.format("%.0fd %02.0fh %02.0fmin %02.0fs", days, hour, mins, secs)
+		return string.format("%.0f d %02.0f h %02.0f min %02.0f s", days, hour, mins, secs)
 	else
-		return string.format("%02.0fh %02.0fmin %02.0fs", hour, mins, secs)
+		return string.format("%02.0f h %02.0f min %02.0f s", hour, mins, secs)
```
stangri, line 47: What's the reason for extra spacing between the value and denominator?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
getsnoopy: It took me a while to figure it out as well, but `%m` is defined as a formatting placeholder for unit prefixes, and the number determines the base. So `%1024mB` is replaced by `KiB` or `MiB`, for example, depending on the size of `pkg.size`. It would be better to convert, but as per our conversation above, I'll be making the base conversions in another commit to keep this one atomic.

applications/luci-app-wireguard/luasrc/view/wireguard.htm
getsnoopy: Yes. All units should have a space between the quantity and the unit symbol, both grammatically and as per the SI.

modules/luci-compat/luasrc/tools/webadmin.lua
getsnoopy: See my comment above.

modules/luci-base/htdocs/luci-static/resources/cbi.js
getsnoopy: See my comment above.

applications/luci-app-wireguard/luasrc/view/wireguard.htm
stangri: While obviously it's not that big of a deal on a greater scheme of things, if this results in `(5 s ago)` rather than `(5s ago)` I'm not sure if it's an improvement.

modules/luci-compat/luasrc/tools/webadmin.lua
stangri: I'm not sure if the `2 h 2 m 2 s` is better from the readability standpoint than `2h 2m 2s`.

hnyman: It isn't. It may be right grammatically, but for status info the short version is much more understandable

stangri:
@getsnoopy while I certainly appreciate your gigantic effort towards SI compliance, I personally would strongly prefer if you removed the extra spacing between numerical values and denominators in the 3 cases above.

jow-:
Another problem is that the additional space introduces potential word wrapping in unexpected/undesired spots. And typographically it indeed looks funny - probably U+202F (NARROW NO-BREAK SPACE) should be used instead.

modules/luci-compat/luasrc/tools/webadmin.lua
getsnoopy: Well it would be `2 h 2 min 2 s`, which is the correct way to print it. The symbol for minute(s) is `min`, not `m`. It adds readability because the space separates the numbers from the units.

getsnoopy:
@stangri I understand some people don't like the spaces, but that seems quite subjective. On the other hand, `3m` means "3 million", whereas `3 m` means "3 metres"; only `3 min` means "3 minutes". Similarly, `4s` means "fours", whereas `4 s` means "four seconds". The space is not cosmetic. We should not depend on context to write units properly.

@jow- Good point, I forgot about wrapping issues because I thought these are inserted into bounding boxes with enough space. I don't know if the NARROW NO-BREAK SPACE is appropriate though, since that is used to replace the comma as the thousands separator (e.g., "10 000" instead of "10,000"). I can change it to the U+00A0 (NO-BREAK SPACE).

applications/luci-app-wireguard/luasrc/view/wireguard.htm
getsnoopy: I understand it seems strange at first, but "5s ago" means "fives ago", whereas "5 s ago" means "5 seconds ago". I think it would make more sense if you look at it with minutes: "5 min ago" is the proper way while "5min ago" looks incorrect.

jow-:
U+00A0 is too wide and makes things like `2 h 2 min 2 s` look weird.

getsnoopy:
I think the code formatting is making it look weird because it uses a fixed-width font. It would look more like "2 h 2 min 2 s". Though, I could be convinced by "2 h 2 min 2 s" (with thin spaces) as well. I'll change them.

modules/luci-compat/luasrc/tools/webadmin.lua
```diff
@@ -42,9 +42,9 @@ function date_format(secs)
 	end
 	
 	if days > 0 then
-		return string.format("%.0fd %02.0fh %02.0fmin %02.0fs", days, hour, mins, secs)
+		return string.format("%.0f\u{202f}d %02.0f\u{202f}h %02.0f\u{202f}min %02.0f\u{202f}s", days, hour, mins, secs)
```
jow-, line 45: The `\u{....}` syntax is not supported by Lua 5.1. You need to paste the character literally.

hnyman: That sounds like quite complication, especially at the translations stage.

I suggest that it is left as it has been.

getsnoopy: @jow- Done.

@hnyman The strings should not be translated, as symbols are universal. That's one of the primary motivations for changing everything to symbols.

stangri: > @hnyman The strings should not be translated, as symbols are universal. 

Oh, they are absolutely not universal and will need to be translated if you look outside of the bubble of germanic group of languages.

If half-space is a violation of the standards system anyways, why implement it at all? This is no longer about standards, this is about an insistence that "2 h 2 m 2 s" has better readability than "2h 2m 2s". 

The half-space and no space are both violations, why not let things be as they used to be?

Until those strings are localizable and locale owners can change spacing and the order of things as most suitable for the locales, I strongly believe they should be left as is, as that had been universally accepted way of presenting information so far, based on the fact that no one has complained.

getsnoopy: @stangri Symbols [are universal](https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9-EN.pdf#page=10); that's why they're referred to as symbols instead of abbreviations (which are translated). There's akin to the per cent symbol (%): it's the same in all languages, doesn't change between singular/plural based on the quantity, and does not change based on upper-/lowercase. You can see it on this Japanese food package as an example:

81229561-a519b780-8fb5-11ea-8075-a4f7b780a68b

And the half-space is not a violation per se; the standard calls for [_some_ kind of space](https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9-EN.pdf#page=35), but doesn't specify the width. It is customary to use a thin space as the thousands separator and a full space to separate the quantity and the unit symbol.

The other main focus was about "m" being the symbol for the metre, not for minute. The symbol for the minute is "min", which is one of the main changes. Something does not become universally accepted just because no one has complained about it, it just means they have tolerated it or might not have noticed it. This is not to mention the fact that this PR is implicitly a complaint about the very issue.

stangri: > @stangri Symbols [are universal](https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9-EN.pdf#page=10); that's why they're referred to as symbols instead of abbreviations (which are translated). They're akin to the per cent symbol (%): it's the same in all languages, doesn't change between singular/plural based on the quantity, and does not change based on upper-/lowercase. 

No, it's definitely NOT similar to the % symbol and NOT the same in ALL languages, check out this wiki article about kilogramm in Russian: https://ru.wikipedia.org/wiki/%D0%9A%D0%B8%D0%BB%D0%BE%D0%B3%D1%80%D0%B0%D0%BC%D0%BC. It mentiones `kg` 3 times, the rest of the article uses the russian abbreviation `кг` to refer to kilogramm (14 times), not `kg`.

While some locales may be more tolerant to using SI symbols instead of their own abbreviations, but definitely not all will be.

> The other main focus was about "m" being the symbol for the metre, not for minute. The symbol for the minute is "min", which is one of the main changes. 

I believe that "main focus" wasn't implemented until after I brought up the spacing.

We're not talking about a scientific paper, we're talking about a device (in many cases consumer device) around the world, we should use what makes sense for (in this case, no spacing and m for minutes) and is in common use by the customer/user and not force something they are not accustomed to.

hnyman: > we should use what makes sense for (in this case, no spacing and m for minutes) and is in common use by the customer/user and not force something they are not accustomed to.

+1 for stangri's comment. The old way "2h 45m 32s" is better, although not strictly grammatically correct.

stangri: > Regardless, 96%+ of the world uses the SI as their sole system of measurement, and 100% of the world encounters the SI in some way or another on almost a daily basis.

Could you please provide reference to back that up. I'm not sure you're aware that a large portion of just the English speaking population is more accustomed to imperial units, not metric ones. Also part of that population doesn't even think in terms of 24 hours and thinks am/pm. How does that conform to SI as the sole system of measurement for 96%+?

> I don't understand the resistance here. 

Like I said before, IMHO "2 h 2 m 2 s" or even "2 h 2 min 2 s" are significantly harder to quick-read than "2h 2m 2s" and carry no benefit to the user.

getsnoopy: > Could you please provide reference to back that up. I'm not sure you're aware that a large portion of just the English speaking population is more accustomed to imperial units, not metric ones. Also part of that population doesn't even think in terms of 24 hours and thinks am/pm. How does that conform to SI as the sole system of measurement for 96%+?

[Sure](https://www.nist.gov/system/files/documents/pml/wmd/metric/1136a.pdf#page=3). The overwhelming majority of the [English-speaking population](https://en.wikipedia.org/wiki/English-speaking_world#English_as_a_global_language) is [accustomed to the metric system](https://www.cia.gov/library/publications/the-world-factbook/appendix/appendix-g.html), not imperial or US customary units. There might be some imperial/US customary units people still use, but those aren't the ones being discussed here. My main point is that being 96% of the population lives in a country where the metric system is the only legal measurement system or is legally the one to be used for the units we're talking about. This means that 96% of the population is accustomed to seeing "1 h 30 min", not "1h 30m", which is also not correct grammatically. This is not to mention that the rest of the 4% of the population encounters metric units almost on a daily basis, so claiming they're unaccustomed to seeing "1 h 30 min" is dubious. The 24-hour clock is not a part of the SI, and is only tangentially related to it (i.e., they usually go together, but are separate).

> Like I said before, IMHO "2 h 2 m 2 s" or even "2 h 2 min 2 s" are significantly harder to quick-read than "2h 2m 2s" and carry no benefit to the user.

I understand that you think it hinders readability, but this is purely subjective; I would disagree with that viewpoint, but that's besides the point. I don't think it's a convincing argument given at least 96% of the world disagrees with you. I sense the general argument here is that "this is how it always has been, so this is how it should be"; that argument, however, is never a strong one generally (let alone in this case), especially knowing that in this case the "way it has been done" is incorrect and not preferred by practically all of the population.

stangri:
Food for thought: https://forum.openwrt.org/t/poll-how-should-the-time-be-presented-in-webui/66585.

getsnoopy:
I have no problem changing it to the `hh:mm:ss` format, as this is a valid and international format to represent time durations as well.

getsnoopy:
I've updated the code as discussed.

modules/luci-base/htdocs/luci-static/resources/cbi.js
```diff
@@ -630,8 +630,8 @@ String.prototype.format = function()
 						}
 
 						subst = (td > 0)
-							? String.format('%dd %dh %dm %ds', td, th, tm, ts)
-							: String.format('%dh %dm %ds', th, tm, ts);
+							? String.format('%d:%d:%d:%d', td, th, tm, ts)
```
knarrff, line 633: I am not that fluent in JS, but this looks to be printf-style formatting. If it is, "%d:%d" results in "2:2" for 2 and 2 as input. For numbers representing hours, minutes or seconds, you would always expect two digits: potentially leading zeroes: "02:02".

modules/luci-base/htdocs/luci-static/resources/cbi.js
```diff
@@ -630,8 +630,8 @@ String.prototype.format = function()
...
+							: String.format('%d:%d:%d', th, tm, ts);
```
knarrff, line 634: same here (as in previous line 633)

systemcrash:
No response

knarrff:
Hi,

I am not quite sure which response would have been expected if not from a developer with merge rights. Was this PR superseeded by something else or fixed in another way?

thanks.

PR #4117, opened 2020-05-30
luci-app-https-dns-proxy: decrease reliance on shell commands; proper acl.d file; add cloudflare family/protect

stangri:


stangri:
I'm going to merge this as is and I'll submit another PR if I need to fix anything.

PR #4120, opened 2020-05-30
[19.07] luci-app-simple-adblock: needs curl as dep

casonadams:
Curl is needed to download any of the blocklists.


dibdot:
@stangri

stangri:
Curl is one of the options, it's not a dependency and definitely not for the Luci app.

casonadams:
Hmm okay.  When I install after a fresh install the blacklist urls can’t download because there is no curl.  I see you have a PR to maybe correct a bug as well.  The service is broken or if the box.  One of the urls is using is bad.  But am assuming that this the LuCi side of things.

PR #4123, opened 2020-05-31
[18.06] luci-app-wol: fix "Host to wake up" to select

fanck0605:
This removes two `undefined` option in the menu.

Signed-off-by: Chuck <fanck0605@qq.com>
(cherry-picked from commit coolsnowwolf/luci@c28e337b3e8ae261454f2a7980030825af084a78)


jow-:
NAK, as this removes the ability to enter custom MAC addresses. The underlying `undefined` should be fixed instead.

PR #4127, opened 2020-06-01
luci-mod-network: query WEP support

blocktrron:
This commit removes WEP from the list of supported encryption methods in case hostapd or wpa_supplicant were compiled without WEP support.

Currently WEP is always offered for STA as well as AP modes regardless of the support from the supplicant / hostapd.


jow-:
Merged, thank you!

PR #4130, opened 2020-06-03
[WIP]Add luci-mod-quick_setup and luci-mod-static_ip

biyun233:
luci-mod-quick_setup: combinations of basics configuration for Administrator, WAN, Wifi.
luci-mod-static_ip: addition of static leases.

Our objectif is to create a lite version of luci interface for users non experts, we have chosen some functionalities most popular.
these modules are parts of what we want to create, we explicate what we have done in the wiki of luci.
this is the lien of documentation: https://openwrt.org/docs/guide-user/luci/quick_setup.

We got a problem with the part of quick-setup, when the button "save & apply" is clicked, changes won't be applied, (changes are unsaved), it needs a seconde click on "save & apply".

for the part of static_ip, we just move it to a new module and change the mode of modification.


modules/luci-mod-quick_setup/htdocs/luci-static/resources/view/quick_setup/quick_setup.js
```diff
@@ -0,0 +1,337 @@
+'use strict';
+'require view';
+'require dom';
+'require ui';
+'require form';
+'require rpc';
+'require uci';
+'require network';
+
+var formData={
```
rmilecki, line 10: Please stick to coding style used in `luci` project and use spaces appropriately. E.g.
```
var foo = {
(...)
};
```

modules/luci-mod-quick_setup/Makefile
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (C) 2008-2014 The LuCI Team <luci@lists.subsignal.org>
```
karlp, line 2: this isn't right...

PR #4136, opened 2020-06-07
luci-app-opkg: update default opkg list dir

zhanhb:
The default opkg lists dir is /usr/lib/opkg/lists rather than /tmp/opkg-lists.
https://git.openwrt.org/?p=project/opkg-lede.git;a=blob;f=libopkg/opkg_conf.h;h=37f95a1a9935d673cbba5f786d58c150a74e2fa7;hb=HEAD#l34

### DIFF
--- a/applications/luci-app-opkg/root/usr/libexec/opkg-call
+++ b/applications/luci-app-opkg/root/usr/libexec/opkg-call
@@ -11,7 +11,7 @@ case "$action" in
 	;;
 	list-available)
 		lists_dir=$(sed -rne 's#^lists_dir \S+ (\S+)#\1#p' /etc/opkg.conf /etc/opkg/*.conf 2>/dev/null | tail -n 1)
-		find "${lists_dir:-/tmp/opkg-lists}" -type f '!' -name '*.sig' | xargs -r gzip -cd
+		find "${lists_dir:-/usr/lib/opkg/lists}" -type f '!' -name '*.sig' | xargs -r gzip -cd
 	;;
 	install|update|remove)
 		(


hnyman:
NAK from me.

When you run "opkg update"to update the lists, the new lists do get stored in "/tmp/opkg-lists":

```
root@router1:~# opkg update
Downloading http://downloads.openwrt.org/snapshots/targets/ipq806x/generic/packages/Packages.gz
Updated list of available packages in /var/opkg-lists/openwrt_core
Downloading http://downloads.openwrt.org/snapshots/targets/ipq806x/generic/packages/Packages.sig
Signature check passed.
...

root@router1:~# ls /tmp/opkg-lists/
openwrt_base           openwrt_luci           openwrt_routing
openwrt_base.sig       openwrt_luci.sig       openwrt_routing.sig
openwrt_core           openwrt_packages       openwrt_telephony
openwrt_core.sig       openwrt_packages.sig   openwrt_telephony.sig

root@router1:~# ls /usr/lib/opkg/lists/
```

We avoid to writing to the flash, so we write to the ramdisk in /tmp  (or actually to /var, but by default that is /tmp )
Configured in: 
https://github.com/openwrt/openwrt/blob/master/package/system/opkg/files/opkg.conf#L3


@jow- 
But should that be actually changed in LuCI from "tmp" to "var", as the opkg package config says /var ???  In most systems it makes no difference, but for those who have modifed /var location, there is a difference.

zhanhb:
I comment the line `lists_dir ext /var/opkg-lists` in /etc/opkg.conf, then execute result of command `opkp update` in directory `/usr/lib/opkg/lists`. Threre should not be any difference between devices.

PR #4138, opened 2020-06-08
luci-mod-network: Use validators from validator module

wt:
The current code has custom implementations of a couple validators
that are already available from the validator module in luci-base.

Signed-off-by: Wren Turkal <wt@penguintechs.org>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -303,7 +303,7 @@ return view.extend({
 
 		o.depends('rebind_protection', '1');
 		o.placeholder = 'ihost.netflix.com';
-		o.validate = validateAddressList;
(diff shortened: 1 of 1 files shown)


wt:
BTW, this has been tested via a qemu build.

jow-:
NAK, there's not the same. The address validator from the validation module does not support dnsmasq' forward slash notation (`/hostname/ipaddr`)

wt:
Oh, the rebind one is also fqdn as opposed to a hostname, so that makes sense. Would it be interesting if I added an fqdn validator and use that for the rebind domain if I could find a way to allow it to be surrounded with the /'s?

Is the hostname allowed in the static lease also an fqdn? That kinda doesn't make sense since the host wouldn't be able to use that name? But I guess is also kinda does make sense since it could simply be a dns name?

@jow-, It seems like having an fqdn validator would be useful. What do you think?

jow-:
Closing this as it won't be merged.

PR #4141, opened 2020-06-08
[WIP]Add luci-mod-quick_setup

biyun233:
luci-mod-quick_setup: combinations of basics configuration for Administrator, WAN, Wifi.

Our objectif is to create a lite version of luci interface for users non experts, we have chosen some functionalities most popular, these modules are parts of what we want to create, we explicate what we have done in the wiki of luci.
This is the lien of documentation: https://openwrt.org/docs/guide-user/luci/quick_setup.  

We got a problem with the part of quick-setup, when the button "save & apply" is clicked, changes won't be applied, (changes are unsaved), it needs a seconde click on "save & apply".


jonglezb:
@jow- @karlp @rmilecki the formal issues from #4130 should be fixed, could you do another round of reviews? Thanks!

luizluca:
That's a nice simplification for Luci. Another suggested common and complex usecase would be to turn the router into a repeater (WDS) or an access point (no routing or DHCP).

About the wifi setup, I'm not sure if "_2" is a nice suffix. I would use 2.GHz and 5Ghz, with a help text telling 5Ghz is faster but has shorter range.

I use the same SSID for both frequencies for years. If there is no major compatibility issues, I would use the same SSID for both as default. The suffix could be an option between "use the same name (for both radios, suffix each radio with the frequency or do both).

I would not suggest OpenWrt as default SSID. Imagine two neighbors using the same SSID and different password... Add some unique info (part of MAC address? Random number?) to build the default name. If random is used, avoid creating at boot process as the device might have a poor random seed and it might generate the same random for multiple devices. Or give a fixed example as a hint (not a prefilled value) like "Wireless Name" (It might be localized) but reject if the user use the example name.

Would it be possible to have it suggest a wireless password? And how about creating a guest network? It just need to ask for a guest password (use a fixed but localized " guest" suffix for wireless name).

biyun233:
Hello, thanks for your feedback，and sorry for my lated response.
For the complexe usecases we may improve it after merging our code, as for the name of WiFi, we will discuss with others and then make the modification.
Thanks!

EricLuehrsen:
- Default SSID could be like OEM do: `OpenWrt_hash` where hash is a partial snip of SHA of WiSOC MAC.
- SSID shared or unique per band should be a check box.
 - If checkbox common, example `OpenWrt_a1b2c3d4`
 - if checkbox unique, example `OpenWrt_a1b2c3d4_2ghz` and `OpenWrt_a1b2c3d4_5ghz`
- Default password might be SHA of both SHA of MAC and SHA of board identification
- Default disable WiFi may still be a good idea

aparcar:
> Default password might be SHA of both SHA of MAC and SHA of board unique identification (serial number) 

How would you figure out the serial number on a device?

EricLuehrsen:
Only one easy example, many devices have a flash area OpenWrt won't change. If you switch back to OEM firmware you will get the store-bought defaults back. You could try to parse this area for all the brands and even deliver the store bought defaults (lots of work) or you could just hash the whole mtdblock with that information per brand (less complex).
```
root@linksys-wrt3200acm:~# head -n 22 /dev/mtdblock3
admin1_passphrase=a1b2c3d4e5
admin2_passphrase=a1b2c3d4e5
cert_region=US
default_passphrase=a1b2c3d4e5
default_ssid=Linksys123456
deviceType=urn:home-linksys-com:device:Router:1
devinfo_version=20160121.1
hw_mac_addr=24:F5:A2:a1:b2:c3
hw_revision=1
hw_version=48SAM203.0GA
manufacturer=Linksys, LLC
manufacturerURL=http://www.linksys.com
manufacturer_date=1970/01/01
mfg_data_version=1
modelDescription=Simultaneous Dual-Band Wireless-AC Gigabit Router
modelNumber=WRT3200ACM
serial_number=1234567890
sw_version=v1.0.0
tc_passphrase=SomeHashStringBlob0123456789AbCdEf
tc_ssid=OtherHashStringBlob0123456789
uuid_key=12345678-DEAD-BEEF-CAFE-1234567890
wps_device_pin=12345678
```

aparcar:
Interesting. Seems like a cool feature to setup the same ssid/passphrase than written in the manual. However due to bad passphrase generation practices (e.g. admin/admin) of some vendors this is maybe a step backwards.

zorun:
This discussion (about enabling Wi-Fi by default and auto-configuring a suitable SSID and password) already happened some time ago on the mailing list. From what I remember, the conclusion was:

1) it's hard to do right for **all** devices supported by OpenWrt;
2) it's even harder to do it in a way that fits everybody's needs;
3) OpenWrt is a toolbox, it's easy to create your own OpenWrt images with this kind of features using [uci-defaults](https://openwrt.org/docs/guide-developer/uci-defaults). OEMs certainly have the ability to do that already.

In any case, this is unrelated to the quick setup, because here we are talking about manual configuration done by the device owner/user.

That being said, one thing we could do is to force the user to change the SSID if it's simply `OpenWrt`.

For the Wi-Fi password, the interviews have shown that people mostly want to choose the password instead of having an auto-generated password.

EricLuehrsen:
That all doesn't mean the discussion cannot come up for review again as times change. Whether supporting deployments as a professional or as a good friend and family member, you more often in 2020's will find you are stuck with a wireless device in a pinch (tablet, phone). If OpenWrt release build is still wired-only on install, then being helpful can be difficult. Some problems people get into often require reboot from defaults, or even reflash (what? you're still on chaos calmer?). Some OEM devices do have good defaults, unique defaults, and they are printed right on sticker with regulatory label and bar code. Some don't but usually device id, calibration, and other unique factors are stored permanently. If I could know what the defaults would be after a reboot defaults action (some info either in LuCI or /etc/file/something), leaving the WiFi enabled, then I could fix someones problem with a mere cell phone.

luizluca:
For enabling wireless with a not-so-secure password, we could either:
# enable wifi but avoid routing to internet, forcing the user to set a different password for admin and wifi before going online. It could redirect to admin page using the same strategy wifi with authentication portals do.
# enable wifi only immediately after the first boot and only if there is no pre-defined wireless config. The user must reset the router again in order to get wireless up.
# enable wifi with a weak password when WDS is pressed for a couple of seconds

zorun:
@EricLuehrsen Ok, I now better understand the use-case of initial configuration over Wi-Fi when you can't plug in with a cable. It makes sense.

However, this is not something that can be solved in LuCI, and even if LuCI would need adaptation for this new feature, I still believe it's unrelated to the quick setup. The SSID shown in the quick setup is simply taken from `/etc/config/wireless`. Currently it's `OpenWrt` when you first connect because that's the default SSID in OpenWrt; if the default SSID would change, the quick setup would automatically reflect this.

@luizluca good suggestions, please take them to the mailing list for discussion.

zorun:
@biyun233 here are the changes we discussed:

- rename `luci-mod-quick_setup` to `luci-mod-quick-setup` for consistency with other modules
- configure the same SSID for 2.4 GHz and 5 GHz (simpler user experience)
- don't allow saving if the SSID is "OpenWrt"
- don't allow saving if the Wi-Fi password is invalid (empty or less than 8 characters)

The checks on SSID and password should be done both directly in the form (nicer user experience) and also after the user validates (as a fallback).

By the way, if there is already a Wi-Fi password, is it shown in the form?  I think it should.

zorun:
I reviewed the code, here are a few comments.

In the render function, you are changing the configuration of wireless networks. This is really unexpected: simply loading a page should not change configuration. You should apply the changes only when saving.

You are doing many calls to UCI with this form:
```
uci.get('wireless', this.wifis[Ghz_2].getName(), 'key');
```

This is probably inefficient because (I think) it will make new calls to the backend every time. When you have a `WifiNetwork` or `WifiDevice` object, you should use the `.get()` method, see the documentation:

https://openwrt.github.io/luci/jsapi/LuCI.network.WifiNetwork.html
https://openwrt.github.io/luci/jsapi/LuCI.network.WifiDevice.html

zorun:
Also, when the user does not touch the wireless configuration, it should not be modified For instance, if I simply change the WAN mode and save, the wireless configuration should not be modified.

If this is too difficult to do with the current layout, maybe put each type of configuration in a separate section, each with its "save" button?

jow-:
> This is probably inefficient because (I think) it will make new calls to the backend every time.

The `uci.get()` calls operate on cached data, so this is fine.

zorun:
@jow- oh, good to know! So it's only a cosmetic difference then, is there a preferred way?

jow-:
In this particular case, mostly cosmetic indeed. 

 - `uci.get('wireless', this.wifis[Ghz_2].getName(), 'key');` makes it perfectly clear that we're operating on uci
 - `wifis[Ghz_2].get('key')` is way shorter though

tiagogaspar8:
Is this still in the works?

aparcar:
Please rebase this and squash your commits into one

aparcar:
I rebased and partly fixed this patch, e.g. `hwmode` is replaced by `band`.

It'll look something like this and is currently not functional

image

@jow- what do you think about this in general? Pretty much every vendor interface got a "simple" view like this, I think we should have something like this as well and ship it by default.

okibcn:
The page should allow, if possible, a fast config for different typical scenarios, as it is what most people are struggling with. The would be the different scenarios:
-Main router (attached to a modem) would be the default config with default parameters).
-Wired AP, as a wired dumb AP (DHCP client in LAN, removing WAN interfaces, disabling firewall and dnsmasq services, adding wan devices to LAN interface to make it easier for the user)
-Main router mesh enabled  (Main router of a mesh, things get starting to be more complex for scripted setup)
-Mesh AP (probably the most complex scenario, requiring data from the main router unless someone find an automatic protocol, a kind of what eero does).

stokito:
This is an extremely important feature and we must push this forward.
New inexperienced users just can't start using OpenWrt.
Please merge the PR even if it's not ideal.
The setup wizard was created by two students as a term project and here is their presentation:
https://www.youtube.com/watch?v=usFHV2WYvMw

Here is a link to GitLab with sources:
https://gricad-gitlab.univ-grenoble-alpes.fr/projet-ter-m1-wic-openwrt/documentation

There are three modules:
1. The [luci-mod-quick-setup](https://gricad-gitlab.univ-grenoble-alpes.fr/projet-ter-m1-wic-openwrt/luci/-/tree/quick-setup-v1/modules/luci-mod-quick-setup) i.e. this PR
2. [luci-mod-configuration](https://gricad-gitlab.univ-grenoble-alpes.fr/projet-ter-m1-wic-openwrt/luci/-/tree/simple_configuration/modules/luci-mod-configuration) with port forwards simplification
3. [luci-mod-dashboard](https://gricad-gitlab.univ-grenoble-alpes.fr/projet-ter-m1-wic-openwrt/luci/-/tree/dashboard/modules/luci-mod-dashboard) that provides some simpler dashboard

I think that OpenWrt team should grab those sources. If they are valuable then they'll start receive PRs. But please keep in mind that the students can't spent their time anymore. And this is a very complicated task to create a good UI.

But even now OEM vendors may take the sources and adopt for their devices.

I'll start working on the setup wizard and found two another existing solutions:

* [X-Wrt luci-app-wizard](https://github.com/x-wrt/com.x-wrt/tree/master/luci-app-wizard) which is in real life use. It's stores all files to a separate config but on reboot the settings are set to corresponding UCI settings.
* [kiddin9/luci-app-wizard](https://github.com/kiddin9/luci-app-wizard) was copied from the X-Wrt and has some additional features like Siderouter mode (e.g. "dumb switch") .

So I going to take the luci-mod-quick-setup and improve it. But this is a long process that needs for a lot of discussions and contributions.
We need for a starting point.

stokito (approved):

lu-zero:
Any updates on this?

stokito:
I can assist with fixing issues of the PR.

systemcrash:
Hi @stokito would you like to take this up and submit your PR? Just retain the original "quick-setup" name, and not "wizard". Also fix that other commit subject.

PR #4144, opened 2020-06-08
luci-app-bmx7: fix "nothing goes except topology"

FreifunkUFO:
nice to see luci-app-bmx7, but nothing goes except the topology sub-page.
here is a fix for all other status pages. modern webbrowsers do convert the $ to %24.
info-page is reloaded too much atm, so i delayed it for two seconds.
it would be nice to make the code even better in future, f.z. parameters in URLs should not contain $.
but i cant find the code where it is been called atm
so, but, this workaround will repair all non-working pages (except Topology) on luci-app-bmx7
Signed-off-by: Ufo <ufo@rund.freifunk.net>
p.s. do i have to pull-request this for openwrt-1907 repo too? would be a nice small test for opkg upgrade

### DIFF
--- a/applications/luci-app-bmx7/root/www/cgi-bin/bmx7-info
+++ b/applications/luci-app-bmx7/root/www/cgi-bin/bmx7-info
@@ -90,9 +90,11 @@ if [ "${QUERY##*/}" == "all" ]; then
 	QALL=1
 fi
 
-if [ "$QUERY" == '$info' ]; then
+if [ "$QUERY" == '%24info' ]; then
 	echo '{ "info": [ '
 	print_query status
+#sleep for bad code, see 
+sleep 2
 	echo -n ","
 	print_query interfaces && echo -n "," || echo -n '{ "interfaces": "" },'
 	print_query links && echo -n "," || echo -n '{ "links": "" },'
@@ -100,7 +102,7 @@ if [ "$QUERY" == '$info' ]; then
 	echo "] }"
 fi
 
-if [ "$QUERY" == '$neighbours' ]; then
+if [ "$QUERY" == '%24neighbours' ]; then
 	QALL=1
 	echo '{ "neighbours": [ '
 	echo '{ "originators": '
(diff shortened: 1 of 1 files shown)


FreifunkUFO:
ping @dangowrt for approving
screenshot with this fix.:
luci-app-bmx7

FreifunkUFO:
fixing-shell-script for running nodes:
`root@mr-loco-sued:/www/cgi-bin#  sed -i -e "s/== '$/== '%24/" /www/cgi-bin/bmx7-info
`
but doesnt work for me. maybe some baskslaveshes missing?

aparcar:
Sorry I don't have any access to a BMX7 setup.

dangowrt:
Seems to work, I deployed the patch on the gateways and the result looks good. I don't have the knowledge to judge whether this fix may break other things or if it is good practise to match against HTML-entities encoded chars, etc.

jow-:
@FreifunkUFO - thanks for the fix! Can you explain the `sleep 2` you added? It seems to be a workaround for a problem?

FreifunkUFO:
@jow- yes, its a workaround: the status-page is reloading these infos too fast, i think. i was not able to find the script which calls the reload so i made that sleep. 
screenshot without sleep:



luci-app-bmx7-ohne-sleep-2

aparcar:
The fast reloading problem only occurs for the "Status" tab, which code is [here](https://github.com/openwrt/luci/blob/6f57341f746f32e8222b46bd25416234f22b74a7/applications/luci-app-bmx7/root/usr/lib/lua/luci/view/bmx7/status_j.htm#L91). The view uses `TablePooler` to reload the interface and it's first argument is the reload interval, currently set to 1 second. Other views use [5 seconds.](https://github.com/openwrt/luci/blob/6f57341f746f32e8222b46bd25416234f22b74a7/applications/luci-app-bmx7/root/usr/lib/lua/luci/view/bmx7/nodes_j.htm#L95)

I suggest instead of adding the `sleep` to replace the TablePoller interval in `status_j.htm`.

FreifunkUFO:
fixing-shell-script for running nodes:
`sed -i -e "s/==\ \'\\\$/==\ \'%24/" /www/cgi-bin/bmx7-info`

see https://github.com/openwrt/luci/pull/4554 for a next try to fiy this bug (a bit better)

PR #4146, opened 2020-06-09
luci-app-dawn: add hostname and interface name

PolynomialDivision:
Requires: https://github.com/berlin-open-wireless-lab/DAWN/commit/ffa08dbccddb8c938e427a106c30351c9df6746b

https://github.com/berlin-open-wireless-lab/DAWN/issues/94

@dwmw2 Could u look into this and make the changes according to your needs and modify this? I just added both values.


dwmw2:
In https://github.com/dwmw2/luci/commits/dawn2 I cleaned that up a little to put the hostname/iface in parens and only when they exist. But I prefer the version I did which just puts them in separate columns on their own: https://github.com/openwrt/luci/pull/4147

PolynomialDivision:
Will close because of better PR: https://github.com/openwrt/luci/pull/4147

PR #4147, opened 2020-06-09
luci-app-dawn: Add hostname and interface to network overview

dwmw2:
My preferred alternative to @PolynomialDivision's https://github.com/openwrt/luci/pull/4146 putting the new fields in separate columns on their own.

Signed-off-by: David Woodhouse <dwmw2@infradead.org>


PolynomialDivision:
Please merge that! :)

PR #4162, opened 2020-06-12
ucitrack: fix bug for exec with params

jjm2473:
In file `/etc/config/ucitrack`
```
config fstab
        option exec '/sbin/block mount'
```
`/sbin/block mount` never be called after fstab changed.

Signed-off-by: 练亮斌 <1129525450@qq.com>

### DIFF
--- a/modules/luci-base/root/etc/init.d/ucitrack
+++ b/modules/luci-base/root/etc/init.d/ucitrack
@@ -35,7 +35,7 @@ register_trigger() {
 			;;
 			*)
 				logger -t "ucitrack" "Setting up non-init /etc/config/$config reload handler: $exec"
-				procd_add_config_trigger "config.change" "$config" "$exec"
+				procd_add_config_trigger "config.change" "$config" $exec
 			;;
 		esac
 	fi


modules/luci-base/root/etc/init.d/ucitrack
```diff
@@ -35,7 +35,7 @@ register_trigger() {
 			;;
 			*)
 				logger -t "ucitrack" "Setting up non-init /etc/config/$config reload handler: $exec"
-				procd_add_config_trigger "config.change" "$config" "$exec"
+				procd_add_config_trigger "config.change" "$config" $exec
```
jow-, line 38: @jjm2473 - Hm, if we unquote the exec value it will get evaluated by the shell which might be problematic if special chars are contained.

Can you try if the following works as well?

`procd_add_config_trigger "config.change" "$config" /bin/sh -c "$exec"`

jjm2473: I think they're equivalent

jow-:
Fixed via 23f30cf9cae62754e973f590828644a5894f84af

PR #4163, opened 2020-06-12
luci-mod-network: diagnostics.js: pass IP version flag to ping/ping6

wfleurant:
discovered when iputils-ping is installed diag cant ping the v4 of domains with both A and AAAA records.

tested on snapshots with ping utils:
- iputils-ping_20190709
- BusyBox v1.31.1 (busybox ping)


jow-:
Merged, thanks!

PR #4175, opened 2020-06-14
luci-proto-openfortivpn: add luci interface for openfortivpn

aaronjg:
Create a luci interface for the openfortivpn netifd protocol.

@jow- Is there anyway to exclude the current interface name in the `widgets.NetworkSelect` selector? I'd like to make it impossible for the user to accidentally try to use their new vpn interface as the bound interface.

cc: @lucize 
Signed-off-by: Aaron Goodman <aaronjg@stanford.edu>

### DIFF
--- a/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
+++ b/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
@@ -61,7 +61,11 @@ return network.registerProtocol('openfortivpn', {
 		o.datatype = 'and(hexstring,length(64))'
 		o.optional = true;
 
-		o = s.taboption('advanced', form.Flag, 'peerdns', _('Use DNS server provided by VPN'));
+		o = s.taboption('advanced', form.Flag, 'defaultroute', _('Use default gateway'), _('If unchecked, no default route is configured'));
+		o.default = o.enabled;
+		o.optional = true;
+
+		o = s.taboption('advanced', form.Flag, 'peerdns', _('Use DNS servers advertised by peer'), _('If unchecked, the advertised DNS server addresses are ignored'));
 		o.default = o.enabled;
 		o.optional = true;
 
--- a/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
+++ b/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
@@ -39,7 +39,7 @@ return network.registerProtocol('openfortivpn', {
 	renderFormOptions: function(s) {
 		var o;
 
-		o = s.taboption('general', form.Value, 'server', _('VPN Server'));
+		o = s.taboption('general', form.Value, 'peeraddr', _('VPN Server'));
 		o.datatype = 'host(0)';
 
 		o = s.taboption('general', form.Value, 'port', _('VPN Server port'));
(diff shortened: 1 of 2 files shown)


jow-:
> Is there anyway to exclude the current interface name in the widgets.NetworkSelect selector?

Try `o.exclude = s.section;`

protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
```diff
@@ -0,0 +1,85 @@
...
+
+		o = s.taboption('general', form.Value, 'password', _('Password'));
+		o.password = true;
+
+		o = s.taboption('advanced', widgets.NetworkSelect, 'iface_name', _('Bind interface'), _('Bind the tunnel to this interface (optional).'));
+		o.nocreate = true;
+		o.optional = true;
+
+		s.taboption('advanced', form.Value, 'trusted_cert', _("VPN Server's certificate SHA1 hash"));
+		o.nocreate = true;
```
jow-, line 63: Remove `o.nocreate` from this and the subsequent options

protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
```diff
@@ -0,0 +1,85 @@
...
+		o.optional = true;
+
+
+		o = s.taboption('advanced', form.Flag, 'set_dns', _('Set DNS'));
+		o.default = o.enabled;
+		o.nocreate = true;
+		o.optional = true;
+
+		o = s.taboption('advanced', form.Flag, 'pppd_use_peerdns', _('PPPD Use Peer DNS'));
```
jow-, line 72: This is not directly related to LuCI but to the underlying proto handler - is there any reason to not use the standard `peerdns` option name which is used by most other proto handlers?

protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
```diff
@@ -0,0 +1,85 @@
...
+	getDevices: function() {
+		return null;
+	},
+
+	containsDevice: function(ifname) {
+		return (network.getIfnameOf(ifname) == this.getIfname());
+	},
+
+	renderFormOptions: function(s) {
+		var dev = this.getDevice().getName(),
```
jow-, line 40: Remove the `dev = ...` assignment, it appears to be unused.

protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
```diff
@@ -0,0 +1,85 @@
...
+		s.taboption("general", form.Value, "username", _("Username"));
+
+		o = s.taboption('general', form.Value, 'password', _('Password'));
+		o.password = true;
+
+		o = s.taboption('advanced', widgets.NetworkSelect, 'iface_name', _('Bind interface'), _('Bind the tunnel to this interface (optional).'));
+		o.nocreate = true;
+		o.optional = true;
+
+		s.taboption('advanced', form.Value, 'trusted_cert', _("VPN Server's certificate SHA1 hash"));
```
jow-, line 62: Can we somehow tighten the validation here? What kind of input is required? A 40 byte hex string? If so, try using `o.datatype = 'and(hexstring,length(40))';`

protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
aaronjg: good point. I was mirroring the option names for the underlying openfortivpn program. I can make a PR to update the package to use the openwrt standard option.

aaronjg:
I updated the PR with your suggestions - this will also now require an update the the openfortivpn package to use the 'peerdns' configuration option. (PR12799)

aaronjg:
@jow- How does it look now? peerdns commit has been merged into master on openwrt/packages, so I think this is ready to land if it looks okay to you

jow-:
@aaronjg - looks almost perfect, with a little typo in `defaultroute` option (Use default*t* gateway)

aaronjg:
@jow-  Thank you for catching that. Fixed now.

feckert:
From my point of view all this should fit now.
@aaronjg thanks

PR #4178, opened 2020-06-16
luci-app-xinetd: add new package

he-ma:
This is a luci application to configure xinetd services.
Only the (probably) most used parameters are implemented for now.

This package depends on PR in another repo (https://github.com/openwrt/packages/pull/12389)

### DIFF
--- a/applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
+++ b/applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
@@ -1,0 +1,192 @@
+'use strict';
+'require uci';
+'require ui';
+'require form';
+'require view';
+'require fs';
+'require tools.widgets as widgets';
+
+function validateEmpty(section, value) {
+	if (value) {
+		return true;
+	}
+	else {
+		return _('Expecting: non-empty value');
+	}
+}
+
+return view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('xinetd', _('Xinetd Settings'), _('Here you can configure Xinetd services'));
+
+		s = m.section(form.GridSection, 'service');
+		s.modaltitle = _('Service definitions to be used by Xinetd');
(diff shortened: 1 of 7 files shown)


applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
```diff
@@ -0,0 +1,211 @@
...
+		o.default = 'tcp';
+		o.value('tcp', _('TCP'));
+		o.value('udp', _('UDP'));
+		o.rmempty = false;
+		o.modalonly = true;
+
+		o = s.taboption('basic', form.ListValue, 'user', _('User (UID)'), _('User ID for the server process for this service'));
+		o.rmempty = false;
+		o.modalonly = true;
+		o.load = function(section_id) {
```
jow-, line 117: I suggest to use `widgets.UserSelect` here. This way you can get rid of the custom `load()` and `getUsers()` logic.

he-ma: Thank you for your hint. I will modify it.
But, do you have a tipp how to clear the (existing) keylist if not clearing it in a custom load()?
The list get's longer every time as the user list is appended to it.
Thanks in advance.

jow-: That is probably a bug we should fix separately in `widgets.js`

he-ma: OK.
What do you mean, should I wait for the fix or maybe doing a workaround wit a load() function which resets the list by doing a `this.keylist = 0` only (as I already did)

he-ma:
@jow- Sorry for the delay in responding, but I had been very busy.
Now I made the change you requested (and some other improvements which probably wouldn't allow this app to get upstream).

jow-:
Merged, thank you!

PR #4179, opened 2020-06-16
luci-app-statistics: add ability to configure collectd-mod-sqm

jnahmias:
This patch makes it possible to enable and configure the SQM input plugin
for collectd from the Luci front-end. It also handles the generation of
the appropriate config for the SQM plugin using the Exec plugin.

Signed-off-by: Joseph Nahmias <joe@nahmias.net>


jnahmias:
will resubmit later

PR #4181, opened 2020-06-16
luci-app-statistics: integrate collectd-mod-sqm

jnahmias:
This PR integrates the SQM collectd plugin into luci-app-statistics. It provides for both the configuration of the module, and the generation & display of the graphs from the collected data.

cc: @ldir-EDB0

### DIFF
--- a/applications/luci-app-statistics/luasrc/model/cbi/luci_statistics/sqm.lua
+++ b/applications/luci-app-statistics/luasrc/model/cbi/luci_statistics/sqm.lua
@@ -1,0 +1,25 @@
+-- /usr/lib/lua/luci/model/cbi/luci_statistics/sqm.lua
+-- Copyright 2011 Jo-Philipp Wich <jow@openwrt.org>
+-- Copyright 2020 Joseph Nahmias <joe@nahmias.net>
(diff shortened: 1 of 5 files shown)


jnahmias:
One note for testing: you will have to create a dummy sqm module so that the template code doesn't skip this plugin. For example, you can run: `touch /usr/lib/collectd/sqm.so`.
Hopefully, this will be added to the collectd-mod-sqm package in the near future.

jnahmias:
Bump - do you have any feedback on this PR? is a review needed by anyone in particular? is there anything else I need to do in order for this to be merged?
cc: @hnyman @ldir-EDB0 @feckert

feckert:
The way I see it you want to add this to the openwrt-19.07 branch. But this branch is already released, so only bug fixes should be added here. Especially since the hack with the so file is not good.
This is already integrated in the master branch. The next release is in the pip so why not use the master?

hnyman:
The change required in collectd ("dummy sqm module") to make the LuCI stats menus work in 19.07 is the nasty part. That is the reason why I abandoned my own attempt earlier. 

I say the NAK for 19.07. Great that you got it for yourself, but lets leave the hack out of 19.07.

PR #4145, opened 2020-06-08
luci-app-attendedsysupgrade: Render on client

aparcar:
There are still multiple TODOs in the code, I'd be very happy if someone could help me with those.

Also I "stole" the `displayStatus` function from `ui.js`, maybe there is a better way of implementing status reports.

https://github.com/openwrt/luci/blob/edeb146eb0d22430625a77e8557b8d16eedad759/modules/luci-base/htdocs/luci-static/resources/ui.js#L4048

The current polling is naively implemented, maybe that could be improved.

Signed-off-by: Paul Spooren <mail@aparcar.org>

PR #4185, opened 2020-06-17
luci-mod-dashboard: initial checkin

ZoZhang:
Adding luci-mod-dashboard: a minimalist dashboard, including routers device information, wireless devices connected and dpch devices.

#### Responsive

 Signed-off-by: ZHANG Zhao <Zhao.Zhang2@etu.univ-grenoble-alpes.fr>


modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm
```diff
@@ -0,0 +1,287 @@
...
+
+<%+header%>
+
+<div id="view">
+	<div class="spinning"><%:Loading view…%></div>
+</div>
+
+<script type="text/javascript">//<![CDATA[
+function progressbar(query, value, max, byte)
+{
```
feckert, line 362: Please fix codding style

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
```diff
@@ -0,0 +1,175 @@
...
+				//E('div', { 'class': 'th' }, '%s / %s'.format(_('Down.'), _('Up.')))
+			])
+		]);
+		var container_deviceslist = E('div', { 'class': 'table assoclist devices-info' });
+
+		container_box.appendChild(E('div', { 'class': 'title'}, [
+			E('h3', 'DHCP Devices')
+		]));
+
+		// container_box.appendChild(E('hr'));
```
feckert: Please remove the comments

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
```diff
@@ -0,0 +1,175 @@
...
+			]),
+		]),
+
+		E('div', { 'class': 'td device-info'}, [
+			E('p', {}, [
+				E('span', { 'class': 'd-inline-block'}, [ lease.macaddr ]),
+			]),
+		]),
+
+		// E('div', { 'class': 'td device-info'  }, [
```
feckert: Please remove the comments line

feckert:
@ZoZhang Please stash your commits and change the header to `luci-mod-dashboard: initial checkin`

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
ZoZhang: it is done !

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
ZoZhang: it is done !

modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm
ZoZhang: it is done, thanks !

ZoZhang:
> @ZoZhang Please stash your commits and change the header to `luci-mod-dashboard: initial checkin`

Thank you for your reminder.

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
```diff
@@ -0,0 +1,331 @@
...
+			var ifc = data[i];
+
+			if (v6) {
+				this.params.internet.v6.ipprefixv6.value =  ifc.getIP6Prefix() || '-';
+				this.params.internet.v6.gatewayv6.value =  ifc.getGateway6Addr() || '-';
+				this.params.internet.v6.protocol.value=  ifc.getI18n() || E('em', _('Not connected'));
+				this.params.internet.v6.addrsv6.value = ifc.getIP6Addrs() || [ '-' ];
+				this.params.internet.v6.dnsv6.value = ifc.getDNS6Addrs() || [ '-' ];
+
+				if (ifc.getIPAddr()) {
```
zorun, line 156: To check that the interface is "connected", you should use https://openwrt.github.io/luci/jsapi/LuCI.network.Protocol.html#isUp

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
```diff
@@ -0,0 +1,331 @@
...
+					this.params.internet.v6.connected.value = true;
+				}
+
+			} else {
+				var uptime = ifc.getUptime();
+				this.params.internet.v4.uptime.value = (uptime > 0) ? '%t'.format(uptime) : '-';
+				this.params.internet.v4.protocol.value=  ifc.getI18n() || E('em', _('Not connected'));
+				this.params.internet.v4.gatewayv4.value =  ifc.getGatewayAddr() || '0.0.0.0';
+
+				if (ifc.getIPAddr()) {
```
zorun, line 166: Same here: https://openwrt.github.io/luci/jsapi/LuCI.network.Protocol.html#isUp

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
ZoZhang: Thank you for reminding.

Can I use ifc.isUp() || ifc.getIP6Addr() to verify v6 network connectivity?

ifc.isUp() || ifc.getIPAddr()　verify the v4 network.

Thank

zorun: Why can't you just use isUp()?

ZoZhang: Because I am not sure whether "running" means there is an existing network connection?

ZoZhang: Actually, I need to know whether the interface is up and whether it is connected to the public network.

zorun: The state (up/down) of a protocol is more than just about the state of the device. Each protocol such as DHCP or PPP sets the state to "up" only when it has reached a good state (for instance obtained an IP address, but it could be something else).  See [documentation](https://git.openwrt.org/?p=project/netifd.git;a=blob;f=DESIGN;h=4d58bcff5f8b67eef0b3cd253ccec4b08180e071;hb=HEAD#l94).

That means you must only use `isUp()` and trust it to do the right thing, instead of trying to check something else like IP addresses manually.

ZoZhang: Thank you for your reply

zorun: Can you fix the code?

aparcar:
Is it possible to make the icons smaller in storage size?

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
ZoZhang: I have changed the last commit

ZoZhang:
> Is it possible to make the icons smaller in storage size?
The icon uses svg code, the size is 1kb~6kb. I don't think it can be reduced. @aparcar

aparcar:
>  The icon uses svg code, the size is 1kb~6kb. I don't think it can be reduced. @aparcar

Okay, I just though we may recycle existing icons but that may introduce some other issues. https://icons.getbootstrap.com/icons/wifi/

@jow- can you imagine this as a default overview in some future?

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
zorun: I still see the old code which checks for IP address, did you push your changes?

ZoZhang: Of course! I just pushed the last version of the code.

modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm
```diff
@@ -0,0 +1,371 @@
...
+	<div class="spinning"><%:Loading view…%></div>
+</div>
+
+<script type="text/javascript">
+	L.require('ui').then(function(ui) {
+		ui.instantiateView('dashboard/index');
+	});
+</script>
+
+<%+footer%>
```
aparcar, line 371: Why is this code required? Would it be possible to remove all LUA dependencies?

modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm
```diff
@@ -0,0 +1,371 @@
...
+	padding: .2em .6em .3em;
+	font-size: 75%;
+	font-weight: 700;
+	line-height: 1;
+	color: #fff;
+	text-align: center;
+	white-space: nowrap;
+	vertical-align: baseline;
+	border-radius: .25em;
+	font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
```
aparcar, line 23: Text inside the label looks badly positioned.
image

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
```diff
@@ -0,0 +1,322 @@
...
+
+			system: {
+				title: _('Architecture'),
+				value: boardinfo.system
+			},
+
+			release: {
+				title: _('Firmware Version'),
+				value: (L.isObject(boardinfo.release) ? boardinfo.release.description : '')
+			}
```
aparcar, line 313: The value is badly shortened. Please use `<version>  (<revision>)` instead of allow the description to be longer.
image

```
root@gw:~# ubus call system board
{
	"kernel": "5.4.52",
	"hostname": "gw",
	"system": "ARMv7 Processor rev 1 (v7l)",
	"model": "Linksys WRT3200ACM",
	"board_name": "linksys,wrt3200acm",
	"release": {
		"distribution": "OpenWrt",
		"version": "SNAPSHOT",
		"revision": "r13881+3-bae4204e34",
		"target": "mvebu/cortexa9",
		"description": "OpenWrt SNAPSHOT r13881+3-bae4204e34"
	}
}
```

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
```diff
@@ -0,0 +1,150 @@
...
+			E('img', {
+				'src': L.resource('view/dashboard/icons/devices.svg'),
+				'width': 55,
+				'title': this.title,
+				'class': 'middle'
+			}),
+			E('h3', this.title)
+		]));
+
+		for(var idx in this.params.lan.devices) {
```
aparcar, line 48: Currently this list is neither sorted alphabetically on hostname or numerically on IP. I think for most users the primer is easiest to understand. 

Is the list currently ordered by age?

modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm
ZoZhang: Thank you for your feedback. please tell me the name and version of the browser you are using?

aparcar: Firefox 78

ZoZhang: Okay thanks :)

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
ZoZhang: Sorted by age, does it refer to the expires time of the device? 

I don't know if the following code can meet your suggestion.

`		leases.map(function(lease) {
			devices[lease.expires] = {
				hostname: lease.hostname || '?',
				ipv4: lease.ipaddr || '-',
				macaddr: lease.macaddr || '00:00:00:00:00:00',
			};
		});`

It will be sorted according to the devices lease expiration time from smallest to largest. That is, the order of the devices connected to the router.

ZoZhang: Sorted by age, does it refer to the expires time of the device? 

I don't know if the following code can meet your suggestion.

````
leases.map(function(lease) {
			devices[lease.expires] = {
				hostname: lease.hostname || '?',
				ipv4: lease.ipaddr || '-',
				macaddr: lease.macaddr || '00:00:00:00:00:00',
			};
		});
````

It will be sorted according to the devices lease expiration time from smallest to largest. That is, the order of the devices connected to the router.

File: modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js (hunk 8 above)
Comment (ZoZhang) [reply modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js:313]: Okay, I will update the code later.

File: modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm (hunk 6 above)
Comment (ZoZhang) [reply modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm:371]: Because I need to use template `htm` to load custom css code, no other effective method has been found for the time being.

Do you have any suggestions ?

Comment (jow-) [reply modules/luci-mod-dashboard/luasrc/view/dashboard/index.htm:371]: Did you try something like that?

```javascript
document.querySelector('head').appendChild(E('link', {
    'rel': 'stylesheet',
    'type': 'text/css',
    'href': L.resource('style.css')
});
```

ZoZhang: Thanks for reminding! Very good method :) @jow-

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
zorun: The current LuCI status page does not seem to have any particular ordering (from a very quick look)

I also suggest sorting by hostname so that it stays stable over time.

aparcar: It's sorted by lease and therefore newly added devices are on top. I think that good when people e.g. connect their smart TV and it appears I too of the list. I'd suggest to just keep it as is.

modules/luci-mod-dashboard/root/usr/share/rpcd/acl.d/luci-mod-dashboard.json
```diff
@@ -0,0 +1,79 @@
...
+		}
+	},
+
+	"luci-mod-dashboard-index": {
+		"description": "Grant access to main status display",
+		"read": {
+			"file": {
+				"/proc/sys/net/netfilter/nf_conntrack_count": [ "read" ],
+				"/proc/sys/net/netfilter/nf_conntrack_max": [ "read" ],
+				"/usr/lib/lua/luci/version.lua": [ "read" ],
```
aparcar, line 38: The LuCI version should be available somewhere else, @jow- can you help out?

modules/luci-mod-dashboard/Makefile
```diff
@@ -0,0 +1,18 @@
+#
+# Copyright 2019-2020 ZHANG Zhao <Zhao.Zhang2@etu.univ-grenoble-alpes.fr>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI Dashboard Pages
+LUCI_DEPENDS:=+luci-base +libiwinfo +libiwinfo-lua
```
aparcar, line 10: Why is libiwinfo-lua required?

aparcar:
@ZoZhang I think you're doing a great and important job here! Please don't be discouraged by the amount of comments and change requests!

modules/luci-mod-dashboard/Makefile
ZoZhang: I am not very familiar with these dependencies, are there any that I need to remove?

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
ZoZhang: So, should I change back to the previous version?

ZoZhang:
> @ZoZhang I think you're doing a great and important job here! Please don't be discouraged by the amount of comments and change requests!
thanks for your encouragement :) 
I am not angry, but I am very happy to get feedback from everyone. This helps to complete this small project.

ZoZhang:
> @ZoZhang I think you're doing a great and important job here! Please don't be discouraged by the amount of comments and change requests!

ZoZhang:
Thanks for your encouragement :) 
This is my first time submitting code to an open source project, and there are many things I am not familiar with.
I am very happy to get feedback from everyone. This helps to complete this package. 
@aparcar

ZoZhang:
My English is not very good, and some words may be inaccurate, which will cause some misunderstandings.

My original thought is: I first contacted LuCi, the first time to participate in open source code. I don't know these workflows, including the use of packages. So I am very happy and welcome everyone to guide my work.

But I have some work on my own, so I can't deal with the feedback immediately. There will be a lot more free time in the future, and feedback should be given as soon as possible.

Now I don’t know what changes to make, can I give some hints?
@aparcar

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/20_lan.js
aparcar: Yes please.

modules/luci-mod-dashboard/Makefile
aparcar: I'll look into that.

modules/luci-mod-dashboard/root/usr/share/rpcd/acl.d/luci-mod-dashboard.json
aparcar: All right, so @jow- the almighty wizard reads [the *raw* Lua file](https://github.com/openwrt/luci/blob/master/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js#L32), therefore not really requiring Lua

You either add under Firmware Version another Line containing the LuCI version or remove the ` 				"/usr/lib/lua/luci/version.lua": [ "read" ],` line.

modules/luci-mod-dashboard/Makefile
aparcar: I run tested it and it works without, too. Please remove it.
```suggestion
LUCI_DEPENDS:=+luci-base +libiwinfo
```

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
```diff
@@ -0,0 +1,322 @@
...
+			},
+
+			system: {
+				title: _('Architecture'),
+				value: boardinfo.system
+			},
+
+			release: {
+				title: _('Firmware Version'),
+				value: (L.isObject(boardinfo.release) ? boardinfo.release.version + ' ('+boardinfo.release.revision+')' : '')
```
aparcar, line 312: The test if `boardinfo.release` is an object doesn't work here. Confusingly it's always a release.

Here is a SNAPSHOT vs a 19.07.2 `ubus call system board` output:
```
{
	"kernel": "5.4.52",
	"hostname": "gw",
	"system": "ARMv7 Processor rev 1 (v7l)",
	"model": "Linksys WRT3200ACM",
	"board_name": "linksys,wrt3200acm",
	"release": {
		"distribution": "OpenWrt",
		"version": "SNAPSHOT",
		"revision": "r13936+7-739d282c2f",
		"target": "mvebu/cortexa9",
		"description": "OpenWrt SNAPSHOT r13936+7-739d282c2f"
	}
}
```
vs
```
{
	"kernel": "4.14.171",
	"hostname": "core1.reudnetz.org",
	"system": "AMD G-T40E Processor",
	"model": "PC Engines APU",
	"board_name": "pc-engines-apu",
	"release": {
		"distribution": "OpenWrt",
		"version": "19.07.2",
		"revision": "r10947-65030d81f3",
		"target": "x86/64",
		"description": "OpenWrt 19.07.2 r10947-65030d81f3"
	}
}
```

Either you check if `version == "SNAPSHOT"` or you just always use `description`.

aparcar:
> My English is not very good, and some words may be inaccurate, which will cause some misunderstandings.

Don't worry, the OpenWrt community is fairly diverse regarding peoples origin so English is often the second language (as for me). If you write something confusing people will let you know, no reason to be intimidated by that.

> My original thought is: I first contacted LuCi, the first time to participate in open source code. I don't know these workflows, including the use of packages. So I am very happy and welcome everyone to guide my work.

Yes and you do great work here. OpenWrt alone is complex, understanding the unique LuCI framework on-top is even harder. I'm [struggling with it](https://github.com/openwrt/luci/pull/4145) myself :) There are no expectation of you doing everything right on first try.

> But I have some work on my own, so I can't deal with the feedback immediately. There will be a lot more free time in the future, and feedback should be given as soon as possible.

Take your time. OpenWrt is for many people a hobby to work on and neither reviews nor fixes are expected to happen in a timely manner!

> Now I don’t know what changes to make, can I give some hints?

I'll try to fixup the remaining nitpicks.

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
aparcar, line 312:
```suggestion
				value: boardinfo.release.description)
```

aparcar:
I fixed the issues I mentioned earlier and everything work fine. Only issue I see is that a theme has to be installed manually. Adding `luci-theme-bootstrap` seems like a very dirty solution. Please advise what to do, I'd like to merge this package afterwards.

aparcar:
The dashboard doesn't work super well with other themes like `openwrt2020`.

image

I'd just add `luci-theme-bootstrap` as a dependency.

modules/luci-mod-dashboard/root/usr/share/rpcd/acl.d/luci-mod-dashboard.json
```diff
@@ -0,0 +1,78 @@
...
+	"luci-mod-dashboard-logs": {
+		"description": "Grant access to system logs",
+		"read": {
+			"cgi-io": [ "exec" ],
+			"file": {
+				"/bin/dmesg -r": [ "exec" ],
+				"/sbin/logread": [ "stat" ],
+				"/sbin/logread -e ^": [ "exec" ],
+				"/usr/sbin/logread": [ "stat" ],
+				"/usr/sbin/logread -e ^": [ "exec" ]
```
aparcar, line 11: I think these lines are not required.

feckert:
@aparcar Why do we need an theme dependency?
There are already dependency defined.
uhttpd:
https://github.com/openwrt/luci/blob/master/collections/luci/Makefile
nginx:
https://github.com/openwrt/luci/blob/master/collections/luci-ssl-nginx/Makefile

aparcar:
@feckert It only works with the specific bootstrap theme and should work without having the full LuCI collection installed.

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/css/custom.css
```diff
@@ -0,0 +1,353 @@
...
+    background-color: #dc3545!important;
+}
+
+/**
+ * Dashboard Principals Styles
+**/
+
+.Dashboard {
+    color: #212529!important;
+    font-family: -apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,"Noto Sans",sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol","Noto Color Emoji";
```
feckert, line 70: This line is very long. Can this line be broken in any way?
Do we need this a font-families?

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
```diff
@@ -0,0 +1,322 @@
...
+			 }
+		}
+	},
+
+	renderInternetBox: function(data) {
+
+		this.params.internet = {
+
+			v4: {
+				title: 'IPv4 Internet',
```
feckert, line 172: Doesn't that have to be translated

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
```diff
@@ -0,0 +1,322 @@
...
+
+				dnsv4: {
+					title: _('DNSv4'),
+					visible: true,
+					value: ['-']
+				}
+			},
+
+			v6: {
+				title: 'IPv6 Internet',
```
feckert, line 212: Doesn't that have to be translated

modules/luci-mod-dashboard/root/usr/share/rpcd/acl.d/luci-mod-dashboard.json
```diff
@@ -0,0 +1,78 @@
...
+	"luci-mod-dashboard-index-wifi": {
+		"description": "Grant access to wireless status display",
+		"read": {
+			"ubus": {
+				"iwinfo": [ "assoclist" ]
+			}
+		},
+		"write": {
+			"ubus": {
+				"hostapd.*": [ "del_client" ]
```
feckert, line 74: Do we need this? Didn't find the rpc call.

aparcar:
I removed some superfluous CSS and ACLs, it still works fine. I'm guessing there was initially some copy&paste involved.

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/css/custom.css
aparcar: It looks slightly different but still fine. I don't see a problem removing it.

feckert: I just noticed that there are already a lot of fonts. These do not have to be.

feckert:
> @feckert It only works with the specific bootstrap theme and should work without having the full LuCI collection installed.

That's not so good then!
For example, I use luci-theme-openwrt-2020, which means that this theme is not possible?
An exciting theme is always installed.
I would say that this app should work with every theme that could be installed from this repository at least luci-theme-openwrt-2020 ;-)
That would be the module that has a dependency on a certain theme.

feckert:
I have now played with the CSS.
And with two small changes it seems to work with the other theme now.

[0001-luci-mod-dashboard-fix-css-styling-for-luci-theme-op.txt](https://github.com/openwrt/luci/files/5167051/0001-luci-mod-dashboard-fix-css-styling-for-luci-theme-op.txt)


But I only tested it with the luci-theme-openwrt2020.
The theme luci-theme-openwrt does look the same as before so the view does not change for them.

**So we don't need a dependency on a special theme.**

@aparcar could you please verify this?
@ZoZhang If this works for you I will merge your P/R.
Please remove the dependency to the theme [luci-theme-openwrt](https://github.com/openwrt/luci/pull/4185/files#diff-6a5761e99293dfeb3161905c82c1c1f3R10) before merging.

After that I will add my small changes or you could apply my patch on this P/R. Thats is up to you.
Before merge please stash all commits so we have only one if you apply my change in this P/R.

Screenshot of luci-mod-dashborad with my change on theme luci-theme-openwrt2020:
Screenshot_2020-09-03 G3400-V - Übersicht - LuCI

stangri:
Is this supposed to be extendable by 3rd party services? In other words, if a third party service (ie: adblock/simple-adblock) wants to display their own includes, should they go to `modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard`?

feckert:
@stangri Should work from my point of view if you put your file under `htdocs/luci-static/resources/view/dashboard/include/` this should then be [included](https://github.com/openwrt/luci/pull/4185/files#diff-7636e273483e66866abb477e2358e020R15) . But I can't tell you how it is with the acls. You should try that.

aparcar:
@ZoZhang said he's busy for the next week(s) so I took the freedom to apply @feckert patch and update this PR. Thank you very much for the fixup @feckert that actually allows to make the whole thing modular!

I'd say this is now ready to merge.

aparcar (approved):

jonglezb:
@feckert thanks for your contribution!  It was a requirement that this module works with all themes, so these changes look good.

The dynamic include thing is a good idea. Does it mean that any luci app could add panels in the dashboard by adding a file in its own package? If this is the case, I wonder if we need more coordination (e.g. for ordering, filtering) or if we can just let any luci app include new panels in the dashboard without ending up with a mess.

aparcar:
> The dynamic include thing is a good idea. Does it mean that any luci app could add panels in the dashboard by adding a file in its own package? If this is the case, I wonder if we need more coordination (e.g. for ordering, filtering) or if we can just let any luci app include new panels in the dashboard without ending up with a mess.

Because this PR is already fairly old I'd appreciate if we could merge it with the current inclusion logic, however long term a configuration file could be used to set a list of activated modules, which ordering determines the appearance in the dashboard.

feckert:
@ZoZhang thank you for work :+1:
@aparcar thanks for testing and updating the P/R

modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/css/custom.css
ZoZhang: Yes, in theory, it is enough that we keep 2~3 common system fonts

ZoZhang:
@feckert  Thank you very much for your contribution! 
@aparcar Thank you very much for your follow-up! Sorry for not dealing with these issues in time. I see that this module has been merged into master. If there are other changes in the future, I will always be here.

erdoukki:
Héllo,
Is the luci-mod-dashboard merged in master ?
Is it available in snapshots ?

ZoZhang:
@erdoukki  hello, I saw that the module is available on the master branch and I don't know the snapshot situation, you can ask for the LuCi organization, thank you ~
https://github.com/openwrt/luci/tree/master/modules/luci-mod-dashboard

hnyman:
> Is it available in snapshots ?

How about looking directly from the download dirs????

e.g.
https://downloads.openwrt.org/snapshots/packages/x86_64/luci/
luci-mod-dashboard_git-21.020.38173-d3280e6_all.ipk | 7.4 KB | Wed Mar 31 09:05:17 2021

https://downloads.openwrt.org/snapshots/packages/arm_cortex-a15_neon-vfpv4/luci/
luci-mod-dashboard_git-21.020.38173-d3280e6_all.ipk | 7.4 KB | Wed Mar 31 01:34:41 2021

erdoukki:
Thanks, I didn’t find them at the first look...
My apologize...

aparcar:
@erdoukki Hi, the downloads section is a bit difficult to read since it's separation in architectures and package feeds, I understand you didn't find it straight away. In future you can check the [package overview](https://openwrt.org/packages/table/start) to find packages.

tiagogaspar8:
Hi everyone!

On the Material theme the icons yes and no and the tables seem to be incorrectly formatted, and the yes and no don't get the colours shown in the images above.

Thanks!

github535:
Hello all. I like the idea of a simplified-luci for end-user who only need to see and set the: Wifi (client or AP) ssid, password, and some basic info. The rest is accessible via luci.
Could someone guide me on how to install this simplified-luci please. I can WinSCP into openwrt via ssh.

PR #4186, opened 2020-06-17
Adding luci-mod-simple_configuration: combinations of two pages: stat…

biyun233:
Adding luci-mod-simple_configuration: combinations of two pages: static leases and port forwards.

Our objectif is to create a lite version of luci interface for users non experts, we have chosen some functionalities most popular, these modules are parts of what we want to create, we explicate what we have done in the wiki of luci.
This is the lien of documentation: https://openwrt.org/docs/guide-user/luci/port_forwards
https://openwrt.org/docs/guide-user/luci/static_ip

Signed-off-by: WANG biyun <biyun.wang@etu.univ-grenoble-alpes.fr>


systemcrash:
@stokito if you're feeling adventurous, this is a good complement or follow-up to the quick-setup.

PR #4196, opened 2020-06-22
move nixio__bin2hex to header and make static

neheb:
This fixes compilation under GCC10.

There's no reason for this to be in a C file anyway.

Also fixed a -Wmissing-prototypes warning with nixio__addr_prefix.


feckert:
LGTM.
But I didn't do a compile test.
I assume that it compiles, because **only** the definition was moved.

PR #4213, opened 2020-06-27
luci-app-samba4: add new UCI option "enable_extra_tuning"

Andy2244:
* add new UCI option "enable_extra_tuning"

Depends on openwrt/packages#12630

### DIFF
--- a/applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js
+++ b/applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js
@@ -39,13 +39,17 @@ return view.extend({
 		o = s.taboption('general', form.Value, 'description', _('Description'));
 		o.placeholder = 'Samba4 on OpenWrt';
 		
+		s.taboption('general', form.Flag, 'enable_extra_tuning', _('Enable extra Tuning'),
+			_('Enable some community driven tuning parameters, that may improve write speeds and better operation via WiFi.\
+			Not recommend if multiple clients write to the same files, at the same time!'));
+		
 		s.taboption('general', form.Flag, 'disable_async_io', _('Force synchronous  I/O'),
 			_('On lower-end devices may increase speeds, by forceing synchronous I/O instead of the default asynchronous.'));
 
-		o = s.taboption('general', form.Flag, 'macos', _('Enable macOS compatible shares'),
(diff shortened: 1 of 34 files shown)


applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js
```diff
@@ -39,13 +39,17 @@ return view.extend({
...
 		o.placeholder = 'Samba4 on OpenWrt';
 		
+		s.taboption('general', form.Flag, 'enable_extra_tuning', _('Enable extra Tuning'),
+			_('Enable some community driven tuning parameters, that may improve write speeds and better operation via WiFi.\
+			Not recommend if multiple clients write to the same files, at the same time!'));
+		
 		s.taboption('general', form.Flag, 'disable_async_io', _('Force synchronous  I/O'),
 			_('On lower-end devices may increase speeds, by forceing synchronous I/O instead of the default asynchronous.'));
 
-		o = s.taboption('general', form.Flag, 'macos', _('Enable macOS compatible shares'),
```
feckert, line 45: Why did you changed that as well?
This does not match your commit message

applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js
```diff
@@ -39,13 +39,17 @@ return view.extend({
...
+		s.taboption('general', form.Flag, 'macos', _('Enable macOS compatible shares'),
 			_('Enables Apple\'s AAPL extension globally and adds macOS compatibility options to all shares.'));
 			
-		o = s.taboption('general', form.Flag, 'allow_legacy_protocols', _('Allow legacy (insecure) protocols/authentication.'),
```
feckert, line 48: The same there

applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js
Andy2244: its just a minor change, the assignment is not needed, thats all

feckert:
**Note:**
Normally every change should be packed into a separate commit.
Then these changes could be better cherry-picked.
Please keep this in mind in the future when you create a pullrequest.
Any change that does something different and is not too concatenated should be put into a separate commit.

I will merge this now.
Thanks for your contribution

PR #3071, opened 2019-09-15
luci-proto-sstp: Add sstp protocol support for Luci

rkkoszewski:
This is a follow up PR of: https://github.com/openwrt/luci/pull/2768

This PR includes a single Commit based on the previous PR (https://github.com/openwrt/luci/pull/2768).

The content is based on the work of @harveyhu2012 (https://github.com/harveyhu2012/luci/commit/3854b5ac2647a90839f075b8c8af4a4ea33bdf89) and my own changes to adapt the scrips to the new changes in luci and other fix suggestions from previous PRs.

Signed-off-by: Robert Koszewski <rkkoszewski@gmail.com>

PR #4219, opened 2020-06-30
Add luci-proto-sstp implementation

rkkoszewski:
This is the "luci-proto-sstp" implementation to support SSTP via the sstp-client package from LUCI.

This is a follow up to this PR: https://github.com/openwrt/luci/pull/3071

This work is inspired by the work of @harveyhu2012 and has been ported to the new Javascript based LUCI API.

One thing to note is that this requires sstp-client with at least version 1.0.12.3. I noticed that the version included in the latest OpenWRT version 19.07.3 is still at the previous 1.0.12.2 version of sstp-client which is not compatible.

Signed-off-by: Robert Koszewski <rkkoszewski@gmail.com>

### DIFF
--- a/protocols/luci-proto-sstp/htdocs/luci-static/resources/protocol/sstp.js
+++ b/protocols/luci-proto-sstp/htdocs/luci-static/resources/protocol/sstp.js
@@ -1,0 +1,85 @@
+'use strict';
+'require form';
+'require network';
+
+network.registerPatternVirtual(/^sstp-.+$/);
+
+return network.registerProtocol('sstp', {
+	getI18n: function() {
+		return _('SSTP');
+	},
+
+	getIfname: function() {
+		return this._ubus('l3_device') || 'sstp-%s'.format(this.sid);
+	},
+
+	getOpkgPackage: function() {
+		return 'sstp-client';
+	},
+
+	isFloating: function() {
+		return true;
+	},
+
+	isVirtual: function() {
+		return true;
+	},
+
+	getDevices: function() {
+		return null;
+	},
+
+	containsDevice: function(ifname) {
+		return (network.getIfnameOf(ifname) == this.getIfname());
+	},
+
+	renderFormOptions: function(s) {
+		var dev = this.getL3Device() || this.getDevice(), o;
+
+		// -- general ---------------------------------------------------------------------
+
+		o = s.taboption('general', form.Value, 'server', _('SSTP Server'));
+		o.datatype = 'host';
+
+		o = s.taboption('general', form.Value, 'username', _('PAP/CHAP username'));
+
+		o = s.taboption('general', form.Value, 'password', _('PAP/CHAP password'));
+		o.password = true;
+
+		// -- advanced --------------------------------------------------------------------
+
+		o = s.taboption('advanced', form.Flag, 'ipv6', _('IPv6 support'), _('If checked, adds "+ipv6" to the pppd options'));
+
+		o = s.taboption('advanced', form.ListValue, 'log_level', _('sstpc Log-level'));
+		o.value('0', _('0', 'sstp log level value'));
+		o.value('1', _('1', 'sstp log level value'));
+		o.value('2', _('2', 'sstp log level value'));
+		o.value('3', _('3', 'sstp log level value'));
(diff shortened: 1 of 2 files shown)


rkkoszewski:
Is there any reason why sstp-client is not using the latest version that is available in the packages GitHub repository? How does it work to get the package version pushed to the OpenWRT repo? Are there any special requirements for this to happen?

rkkoszewski:
Tagging @feckert here as he had already a look at the previous PR.

feckert:
In the master branch we have already  the version [1.0.13](https://github.com/openwrt/packages/commit/b3f52f80adbd8cf445e0e3b12ff6c392c7b19833#diff-208f1ab83d74461a029af7341163f60c) so on next stable release in openwrt your pullrequest dows have the right version. Your pullrequest is also against the LuCI master.

protocols/luci-proto-sstp/htdocs/luci-static/resources/protocol/sstp.js
```diff
@@ -0,0 +1,82 @@
...
+		o = s.taboption('general', form.Value, 'username', _('PAP/CHAP username'));
+
+		o = s.taboption('general', form.Value, 'password', _('PAP/CHAP password'));
+		o.password = true;
+
+		// -- advanced --------------------------------------------------------------------
+
+		o = s.taboption('advanced', form.Flag, 'ipv6', _('IPv6 support'), _('If checked, adds "+ipv6" to the pppd options'));
+
+		o = s.taboption('advanced', form.Value, 'log_level', _('sstpc log-level'));
```
feckert, line 54: could we make a dropdown with loglevels that are available? So I do not have to remember which number is which loglevel?

protocols/luci-proto-sstp/htdocs/luci-static/resources/protocol/sstp.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+'require uci';
```
feckert, line 2: Why do we need uci?

rkkoszewski: Removed

protocols/luci-proto-sstp/htdocs/luci-static/resources/protocol/sstp.js
rkkoszewski: Added dropdown. Using numerical log level, as there is no specific description of each log-level as per documentation: https://www.mankier.com/8/sstpc#--log-level

feckert:
I did not a run test but the source LGTM.

protocols/luci-proto-sstp/htdocs/luci-static/resources/protocol/sstp.js
```diff
@@ -0,0 +1,85 @@
...
+		o = s.taboption('general', form.Value, 'username', _('PAP/CHAP username'));
+
+		o = s.taboption('general', form.Value, 'password', _('PAP/CHAP password'));
+		o.password = true;
+
+		// -- advanced --------------------------------------------------------------------
+
+		o = s.taboption('advanced', form.Flag, 'ipv6', _('IPv6 support'), _('If checked, adds "+ipv6" to the pppd options'));
+
+		o = s.taboption('advanced', form.ListValue, 'log_level', _('sstpc log-level'));
```
jow-, line 53: Two small nit-picks:
 - Can you change the label to just `Log-level` (w/ capitalized L for consistency with other labels)
 - Add translation context to the number labels: `_('0', 'sstp log level value')` - the second argument won't be visible to the user but serve as hint for translators

rkkoszewski: Does this need to be translated at all? Having something like "o.value('0');" also seems to work, but I haven't seen any explicit mention of this approach in the official documentation, so I'm avoiding it in case this would break things in the future.

feckert:
Thanks for your contribution :+1:

rkkoszewski:
Cool! Thank you all for having a look at it!

openwrt-diy:
Can it support adaptation to openwrt-19.07 luci ?

PR #4227, opened 2020-07-03
luci-base: add back the help.gif

brvphoenix:
The help.gif has been moved to luci-compat after 102124115d4c2e204bf4cf1ffafe1d8b304fd0df. But it is still used by the luci-theme-bootstrap and luci-theme-openwrt as the following codes. So we should add it back or also remove it form the themes.

https://github.com/openwrt/luci/blob/558525783cf33a986efa33d23f5283b33a6c936e/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L1928

Signed-off-by: brvphoenix <vanwaholtz@gmail.com>


jow-:
Solved by 69608199973651baf585d24a095edf8c0c42e21f

PR #4239, opened 2020-07-07
luci-base: dispatcher: error404: flow message into template

karlp:
message was only being shown in the plain text case when the render
failed.


feckert:
I don't know if this overshoots the mark, but shouldn't we translate the message as well?
https://github.com/openwrt/luci/blob/master/modules/luci-base/luasrc/dispatcher.lua#L1032

karlp:
I have no objection to that at least :) but yeah, I feel it's out of scope of this at least.  All the asserts() in dispatcher.lua would/could/should all be translated then too

PR #4257, opened 2020-07-12
luci-app-uhttpd: fix translation for key option

### LINKED ISSUE #4248: Change label that also DEM certificate format is permitted
On the luci configuration page _Services / uhttpd_ the fields for choosing the certificate and private key are labeled **HTTPS Certificate (DER Encoded)** and **HTTPS Private Key (DER Encoded)**. This is misleading because uhttpd also accepts the PEM format.

The problem is that the DER format cannot contain a certificate chain. This is no issue for self-signed certificates, but a big problem for CA-signed ones. That's why the PEM format is needed.

I found this ticket where the PEM support has been added: https://dev.archive.openwrt.org/ticket/12457
and this ticket where someone had a similar problem: https://github.com/openwrt/packages/issues/2797

I suggest to simply change the labels on the web page to something like: "HTTPS Certificate (DER or PEM format)"

atiensivu:
Resolves #4248

### DIFF
--- a/applications/luci-app-acme/po/ar/acme.po
+++ b/applications/luci-app-acme/po/ar/acme.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
-"PO-Revision-Date: 2020-07-02 23:42+0000\n"
-"Last-Translator: Marwan Amireh <amirehmarwan@gmail.com>\n"
+"PO-Revision-Date: 2020-07-09 06:17+0000\n"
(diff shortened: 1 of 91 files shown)


feckert:
Please update your commit header to `luci-app-uhttpd: fix translation for key option`

feckert:
@atiensivu Something went fundamentally wrong with your last force push.
Please fix this PR or open a new one with your commits and my requested changes.

stokito:
The PR can be closed. The keys were already fixed "HTTPS Certificate (DER or PEM format)"

PR #4258, opened 2020-07-12
luci-app-uhttpd: add missing options

atiensivu:
The request: 
"maintainer: @cshoredaniel
Since UCI supports option key_type (rsa | ec) it would be helpful if the setting would be exposed through luci-app-uhttpd
Perhaps also option ec_curve"

Adds 'key_type' and 'ec_curve' fields to menu - syntax/validity of values not checked


feckert:
Please edit your commit header to` luci-app-uhttpd: add new options key_type and ec_curve`.

Also remove the following lines from the commit message

> The request:
> "maintainer: @cshoredaniel

feckert:
No reaction from the creator

PR #4264, opened 2020-07-13
luci-app-nlbwmon: move into menu entry into service section

feckert:
I'm referring to the discussion with @jow- where we once discussed whether we should rename the service section to app.
Unfortunately I cannot find the thread anymore.

To avoid extending the menu tree the application luci-app-nlbwmon should also be moved to the service section and not get an extra menu entry.


feckert:
@jow- Any objections, or can I merge it.

feckert:
I have just released that I forgot something. The links in display.js and config.js needs to be adjusted as well!

PR #4265, opened 2020-07-13
luci-app-dockerman: improvements

feckert:
1. Move uci config calls into the start function
2. Add btn css class to all button inputs. Fix button view.

@lisaac  Please have a look so we could also move this changes to https://github.com/lisaac/luci-app-dockerman


lisaac:
Thanks.

feckert:
I have merged this.
Please cherry-pick this changes to your Repository so we have the same working base.

PR #4266, opened 2020-07-13
luci-theme-openwrt-2020: move definitions into custom.css file

feckert:
With this change the css variables can be adjusted independently. It is
not necessary to change the css file directly in cascade.css.
The changes are stored in a custom.css file.


feckert:
@jow- Any objections, or can I merge it.

jow-:
Didn't get around to test it yet

evelikov:
The commit is a simple code motion and seems perfectly reasonable IMHO.

A couple of thing worth considering as follow-up:
 - rename "secondary" to something more obvious - where is it used, etc
 - factorise the 9 instances of rgba() in cascade.css into variables in custom.css

feckert:
I am now using my old luci-theme-material again thanks to @Ansuel for the change to get this working again.
Therefore, this change is no longer relevant for me.

PR #4270, opened 2020-07-16
luci-app-unbound: add rate_limit and dns_assist option support

### LINKED ISSUE #4174: luci: option domain in unbound.conf is lost on "apply changes"
Foreword: originally reported under [packages](https://github.com/openwrt/packages/issues/12444) but moved here as [recommended.](https://github.com/openwrt/packages/issues/12444#issuecomment-643671131)

Environment: Fritzbox 4040, OpenWRT 19.07.3

Description:
1. configure unbound in parallel mode
2. insert [relevant "domain" keyword](https://github.com/openwrt/packages/commit/b6caafef46829922dade557d6b3082360053bf8f) for custom domain in /etc/config/unbound
`option domain 'yourdomain'`
3. in the GUI make a minor change, say TTL from 120 to 121, and apply changes
4. verify that the keyword domain has disappeared from the config
5. again in the GUI make another minor chage, say TTL back to 120, and apply changes
6. verify that resolv.conf now has a "search lan." line instead of "search yourdomain."

Basically, after two restarts, the keyword is gone from resolv.conf
Issue first discussed in the [forum](https://forum.openwrt.org/t/unbound-option-domain-is-lost/64987/12?u=aboaboit)

EricLuehrsen:
maintainer: me
tested: Linksys WRT3200ACM
description:
resolves #4174 and picks up options from https://github.com/openwrt/packages/pull/12798

### DIFF
--- a/applications/luci-app-unbound/luasrc/model/cbi/unbound/configure.lua
+++ b/applications/luci-app-unbound/luasrc/model/cbi/unbound/configure.lua
@@ -138,7 +138,7 @@ if (valman == "0") then
     dlk = s1:taboption("DHCP", ListValue, "dhcp_link",
         translate("DHCP Link"),
         translate("Link to supported programs to load DHCP into DNS"))
-    dlk:value("none", translate("No Link"))
+    dlk:value("none", translate("(none)"))
     dlk:value("dnsmasq", "dnsmasq")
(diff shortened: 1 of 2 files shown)


jow-:
Merged, thanks!

PR #4275, opened 2020-07-20
luci-compat: add btn class to all button inputs

feckert:
This fixes the view of the buttons, that are still written in lua and therefore use luci-compat.
The buttons do not look nice when using the new luci-theme-openwrt-2020 theme.
Therefore I have added the btn css class to every button input in this pullrequest

Before the change:
Screenshot_2020-07-20 st-dev-07 - Schnittstellen - LuCI
After the change:
Screenshot_2020-07-20 st-dev-07 - Richtlinien - LuCI


jow-:
Great - thank you!

Issue #4234, opened 2020-07-06
[luci-app-uhttpd | feature suggestion] expose certificate signature algorithms option


n8v8R:
maintainer: @cshoredaniel 

Since UCI supports `option key_type` (rsa | ec) it would be helpful if the setting would be exposed through _luci-app-uhttpd_

n8v8R:
since this is pending adding `tls_key` option (https://patchwork.ozlabs.org/project/openwrt/patch/20200215225551.1652646-1-jo@mein.io/) to LuCI would make sense.

stokito:
I think we can avoid of such option. The Luci should use a default generated snake-oil self signed cert, same as in Debian. Ideally it should be automatically renewed. Since all browsers supports ECC keys then just generate the ECC cert by default. 
The uhttpd config has options to configure the cert fields but this is useless because anyway they fields can't be trusted so we better to just remove this options. Only the CN field is important but it must be same as the hostname.

So the whole section in the [luci-app-uhttpd](https://github.com/openwrt/luci/blob/master/applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js#L225) should be removed too.
But once the `acme` package is installed it should automatically issue and sign a certificate for the hostname and replace the self-signed certificate.

PR #4276, opened 2020-07-20
luci-app-uhttpd: add missing options

atiensivu:
Adds 'key_type' and 'ec_curve' fields to menu.
Syntax and validity of values not checked.

maintainer: @cshoredaniel

Issue: #4234


applications/luci-app-uhttpd/luasrc/model/cbi/uhttpd/uhttpd.lua
```diff
@@ -211,10 +211,16 @@ o = s:option(Value, "days", translate("Valid for # of Days"))
...
+o = s:option(Value, "key_type", translate("Key type: RSA or EC"))
+o.default = "RSA"
+
 o = s:option(Value, "bits", translate("Length of key in bits"))
 o.default = 2048
 o.datatype = "min(1024)"
 
+o = s:option(Value, "ec_curve", translate("Elliptic curve (EC) to use"))
+o.default = "P-256"
+
```
feckert, line 223: I think we should add a drop down so the use could only select one of theme.

applications/luci-app-uhttpd/luasrc/model/cbi/uhttpd/uhttpd.lua
```diff
@@ -211,10 +211,16 @@ o = s:option(Value, "days", translate("Valid for # of Days"))
 o.default = 730
 o.datatype = "uinteger"
 
+o = s:option(Value, "key_type", translate("Key type: RSA or EC"))
+o.default = "RSA"
+
```
feckert, line 216: I think we should add a drop down so the use could only select one of theme.

systemcrash:
@atiensivu could you address these fixes, please?

PR #4277, opened 2020-07-20
luci-app-upnp: Add STUN options

atiensivu:
/etc/init.d/miniupnpd already has variable support for these additions to the miniupnpd package.

This adds configuration support for those items.

### DIFF
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
@@ -167,6 +167,15 @@ return view.extend({
 		o = s.taboption('advanced', form.Value, 'upnp_lease_file', _('UPnP lease file'))
 		o.placeholder = '/var/run/miniupnpd.leases'
 
+        	s.taboption('advanced', form.Flag, 'use_stun', _('Use STUN')).default = '0'
+
+		o = s.taboption('advanced', form.Value, 'stun_host', _('STUN host'))
+		o.placeholder = 'stun.stunprotocol.org'
+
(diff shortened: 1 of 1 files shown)


applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -167,6 +167,15 @@ return view.extend({
 		o = s.taboption('advanced', form.Value, 'upnp_lease_file', _('UPnP lease file'))
 		o.placeholder = '/var/run/miniupnpd.leases'
 
+        	s.taboption('advanced', form.Flag, 'use_stun', _('Use STUN')).default = '0'
```
feckert, line 170: Please change the following:
- Change `default` to `o.default = '0'` into a new line this is more common
- Add a dependency to `use_stun` for `stun_host` and `stun_port` so that this option are only shown if the flag `use_stun` is selected
- Update the commit message header to `luci-app-upnp: add STUN options`

atiensivu: Fixing in an upcoming commit - thanks for the feedback

feckert:
@atiensivu Could you please stash the last two commits into the first one?
Then it is ready to merge.

systemcrash:
@atiensivu Can you squash these commits as @feckert describes?

Neustradamus:
A repost of @systemcrash (2021-12-03):
@atiensivu Can you squash these commits as @feckert describes?

feckert:
Sorry no reaction from creator
If this is still needed please rebase.

PR #4283, opened 2020-07-22
luci-app-dockerman: refactoring and update coding style

feckert:
@lisaac This is maybe the first step to update the coding style of the whole luci-app-dockerman application.

The code is very difficult to edit and to maintain, because it does not correspond to the coding style I am used to in LuCI.

I see following advantage:

- Code blocks can be recognized more easily
- The code can be refactored better because it becomes clearer
- The changes could be tracked better, because the individual lines of code are not so long anymore
- The whole module becomes smaller, because not every time several spaces are used to indent (but that is a minor matter)

I have also removed the **ac** and **uci** handling and so also the init script.
This is not needed anymore since we added uci handling to docker-ce.
The ac handling for [docker0](https://github.com/openwrt/packages/pull/12897) is pushed.


lisaac:
Ok, it's good for me.
And I think we should modify docker-ce first, once it's done, then adjust the dockerman.

feckert:
I have now merged the changes in the luci-app-dockerman so that I have a stable starting point again.
The [P/R](https://github.com/openwrt/packages/pull/12897) of docker-ce could also be merged but I am still waiting for the OK of @G-M0N3Y-2503 . By merging luci-app-dockerman the setting of the [bip](https://github.com/openwrt/packages/pull/12897/files#diff-d2f590ffe4d65867a6968c82a4a2ed5aR126) and the [hosts](https://github.com/openwrt/packages/pull/12897/files#diff-d2f590ffe4d65867a6968c82a4a2ed5aR130) is not possible till the docker-ce P/R is merged. But everything else should work.

PR #4285, opened 2020-07-23
[RFC] support reverse proxy with pathname prefix

nunojpg:
This is a basic "works for me" sketch to allow to run luci with a pathname prefix under a reverse proxy, as to allow for example to have several routers accessible from a single domain. Eg:

(cgi-bin/luci/ appended for clarity)

https://example.com/router1/cgi-bin/luci/
https://example.com/router2/cgi-bin/luci/
https://example.com/router3/cgi-bin/luci/

There is a choice of 3 implementations strategies to achieve this:
- HTTP header from the proxy server specifying the base pathname
- Fix all the code to use relative paths
- Create a cookie when index.html loads with it's pathname

As a proof of concept I've tried the last.

The variable "resource" is reused to either refer to HTTP paths or to Filesystem paths (see change at modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js). Since I have it now using the path including the prefix it fails for filesystem paths. I've hardcoded it manually at that file just to get it to work. It is likely broken in other places. A eventual solution is to split this variable into a resource_fs and resource_http.

I now set the cookie regardless if the prefix is empty (direct access case) but this can naturally be avoided.


valm13:
I also need to access multiple OpenWrt Luci interfaces via a reverse proxy.
Is it supposed to be merged with the master branch?

nunojpg:
> I also need to access multiple OpenWrt Luci interfaces via a reverse proxy.
> Is it supposed to be merged with the master branch?

Hi Valm, this is only a proof of concept and would definitly need long term considerations from @jow-.

Btw, not all can be in mainline, maybe you will really need to keep your patches in top. Not desirable and not always easy, but that's how work works :).

valm13:
Alright, that's a bit sad, it would have been much more powerful if we could set a basepath in an easier way.😃
Thanks for your quick reply and have a nice day. ☀

jempatel:
@nunojpg , this works perfectly and I have been testing these changes. I have noticed one issue when any static URL is accessed directly, index.html does not come into the picture and it does not set the pathname variable and at that time, it shows an error

`/usr/lib/lua/luci/dispatcher.lua:373: attempt to concatenate a nil value`

systemcrash:
@nunojpg Could you just rebase your original commit(s) onto current master if this PR is still relevant? There's about 3 years worth of accretion and I can't even see what the original PR is...

nunojpg:
Abandoned

systemcrash:
OK, thanks.

PR #4286, opened 2020-07-24
luci-base: fix widgets CBIUserSelect appending list on load

he-ma:
By using `widgets.UserSelect` the user list is appended to the dropdown again each time the page is loaded.
Although I'm not sure whether this PR is the correct solution to the problem I experienced it is working.

I came across this problem during the implementation of the LuCI xinetd module (https://github.com/openwrt/luci/pull/4178)
when @jow- gave me the hint to use the `widgets.UserSelect` instead of the local realization.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/tools/widgets.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/widgets.js
@@ -580,6 +580,8 @@ var CBIUserSelect = form.ListValue.extend({
 
 	load: function(section_id) {
 		return getUsers().then(L.bind(function(users) {
+			delete this.keylist;
+			delete this.vallist;
 			for (var i = 0; i < users.length; i++) {
 				this.value(users[i]);
 			}


modules/luci-base/htdocs/luci-static/resources/tools/widgets.js
```diff
@@ -580,6 +580,7 @@ var CBIUserSelect = form.ListValue.extend({
 
 	load: function(section_id) {
 		return getUsers().then(L.bind(function(users) {
+			this.keylist = 0;
```
jow-, line 583: Don't forget about `vallist` - it needs to be cleared as well. Also use `delete`, it is more explicit and feels less hackish:

```
delete this.keylist;
delete this.vallist;
```

he-ma: @jow- Updated according your suggestion

PR #4307, opened 2020-07-28
Introduce support for managing `config device` and `config bridge-vlan` sections

jow-:
This PR adds initial support for managing `config device` sections which allows for fine grained control over low level device details in uci. It requires the latest netifd Git HEAD to function.

Additionally, it implements support for bridge VLAN filtering as defined in http://lists.openwrt.org/pipermail/openwrt-devel/2020-July/030151.html

grafik
grafik
grafik
grafik
grafik
grafik

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
@@ -1,0 +1,653 @@
+'use strict';
+'require uci';
+'require form';
+'require network';
+'require baseclass';
+'require validation';
+'require tools.widgets as widgets';
+
+function validateAddr(section_id, value) {
+	if (value == '')
+		return true;
+
+	var ipv6 = /6$/.test(this.section.formvalue(section_id, 'mode')),
+	    addr = ipv6 ? validation.parseIPv6(value) : validation.parseIPv4(value);
+
+	return addr ? true : (ipv6 ? _('Expecting a valid IPv6 address') : _('Expecting a valid IPv4 address'));
+}
+
+function setIfActive(section_id, value) {
+	if (this.isActive(section_id)) {
+		uci.set('network', section_id, this.ucioption, value);
+
+		/* Requires http://lists.openwrt.org/pipermail/openwrt-devel/2020-July/030397.html */
+		if (false && this.option == 'ifname_multi') {
+			var devname = this.section.formvalue(section_id, 'name_complex'),
+			    m = devname ? devname.match(/^br-([A-Za-z0-9_]+)$/) : null;
+
+			if (m && uci.get('network', m[1], 'type') == 'bridge') {
+				uci.set('network', m[1], 'ifname', devname);
+				uci.unset('network', m[1], 'type');
+			}
+		}
+	}
+}
+
+function validateQoSMap(section_id, value) {
+	if (value == '')
+		return true;
+
+	var m = value.match(/^(\d+):(\d+)$/);
+
+	if (!m || +m[1] > 0xFFFFFFFF || +m[2] > 0xFFFFFFFF)
+		return _('Expecting two priority values separated by a colon');
+
+	return true;
+}
+
+function deviceSectionExists(section_id, devname) {
+	var exists = false;
+
+	uci.sections('network', 'device', function(ss) {
+		exists = exists || (ss['.name'] != section_id && ss.name == devname /* && !ss.type*/);
+	});
+
+	/* Until http://lists.openwrt.org/pipermail/openwrt-devel/2020-July/030397.html lands,
+	   prevent redeclaring interface bridges */
+	if (!exists) {
+		var m = devname.match(/^br-([A-Za-z0-9_]+)$/),
+		    s = m ? uci.get('network', m[1]) : null;
+
+		if (s && s['.type'] == 'interface' && s.type == 'bridge')
+			exists = true;
+	}
+
+	return exists;
+}
+
+function isBridgePort(dev) {
+	if (!dev)
+		return false;
+
+	if (dev.isBridgePort())
+		return true;
+
+	var isPort = false;
+
+	uci.sections('network', null, function(s) {
+		if (s['.type'] != 'interface' && s['.type'] != 'device')
+			return;
+
+		if (s.type == 'bridge' && L.toArray(s.ifname).indexOf(dev.getName()) > -1)
+			isPort = true;
+	});
+
+	return isPort;
+}
+
+function lookupDevName(s, section_id) {
+	var typeui = s.getUIElement(section_id, 'type'),
+	    typeval = typeui ? typeui.getValue() : s.cfgvalue(section_id, 'type'),
+	    ifnameui = s.getUIElement(section_id, 'ifname_single'),
+	    ifnameval = ifnameui ? ifnameui.getValue() : s.cfgvalue(section_id, 'ifname_single');
+
+	return (typeval == 'bridge') ? 'br-%s'.format(section_id) : ifnameval;
+}
+
+function lookupDevSection(s, section_id, autocreate) {
+	var devname = lookupDevName(s, section_id),
+	    devsection = null;
+
+	uci.sections('network', 'device', function(ds) {
+		if (ds.name == devname)
+			devsection = ds['.name'];
+	});
+
+	if (autocreate && !devsection) {
+		devsection = uci.add('network', 'device');
+		uci.set('network', devsection, 'name', devname);
+	}
+
+	return devsection;
+}
+
+function getDeviceValue(dev, method) {
+	if (dev && dev.getL3Device)
+		dev = dev.getL3Device();
+
+	if (dev && typeof(dev[method]) == 'function')
+		return dev[method].apply(dev);
+
+	return '';
+}
+
+function deviceCfgValue(section_id) {
+	if (arguments.length == 2)
+		return;
+
+	var ds = lookupDevSection(this.section, section_id, false);
+
+	return (ds ? uci.get('network', ds, this.option) : null) ||
+		uci.get('network', section_id, this.option) ||
+		this.default;
+}
+
+function deviceWrite(section_id, formvalue) {
+	var ds = lookupDevSection(this.section, section_id, true);
+
+	uci.set('network', ds, this.option, formvalue);
+	uci.unset('network', section_id, this.option);
+}
+
+function deviceRemove(section_id) {
+	var ds = lookupDevSection(this.section, section_id, false),
+	    sv = ds ? uci.get('network', ds) : null;
+
+	if (sv) {
+		var empty = true;
+
+		for (var opt in sv) {
+			if (opt.charAt(0) == '.' || opt == 'name' || opt == this.option)
+				continue;
+
+			empty = false;
+		}
+
+		if (empty)
+			uci.remove('network', ds);
+	}
+
+	uci.unset('network', section_id, this.option);
+}
+
+function deviceRefresh(section_id) {
+	var dev = network.instantiateDevice(lookupDevName(this.section, section_id)),
+	    uielem = this.getUIElement(section_id);
+
+	if (uielem) {
+		switch (this.option) {
+		case 'mtu':
+		case 'mtu6':
+			uielem.setPlaceholder(dev.getMTU());
+			break;
+
+		case 'macaddr':
+			uielem.setPlaceholder(dev.getMAC());
+			break;
+		}
+
+		uielem.setValue(this.cfgvalue(section_id));
+	}
+}
+
+return baseclass.extend({
+	replaceOption: function(s, tabName, optionClass, optionName, optionTitle, optionDescription) {
+		var o = s.getOption(optionName);
+
+		if (o) {
+			if (o.tab) {
+				s.tabs[o.tab].children = s.tabs[o.tab].children.filter(function(opt) {
+					return opt.option != optionName;
+				});
+			}
+
+			s.children = s.children.filter(function(opt) {
+				return opt.option != optionName;
+			});
+		}
+
+		return s.taboption(tabName, optionClass, optionName, optionTitle, optionDescription);
+	},
+
+	addOption: function(s, tabName, optionClass, optionName, optionTitle, optionDescription) {
+		var o = this.replaceOption(s, tabName, optionClass, optionName, optionTitle, optionDescription);
+
+		if (s.sectiontype == 'interface' && optionName != 'type' && optionName != 'vlan_filtering') {
+			o.cfgvalue = deviceCfgValue;
+			o.write = deviceWrite;
+			o.remove = deviceRemove;
+			o.refresh = deviceRefresh;
+		}
+
+		return o;
+	},
+
+	addDeviceOptions: function(s, dev, isNew) {
+		var isIface = (s.sectiontype == 'interface'),
+		    ifc = isIface ? network.instantiateNetwork(s.section) : null,
+		    gensection = ifc ? 'physical' : 'devgeneral',
+		    advsection = ifc ? 'physical' : 'devadvanced',
+		    simpledep = ifc ? { type: '', ifname_single: /^[^@]/ } : { type: '' },
+		    o, ss;
+
+		if (isIface) {
+			var type;
+
+			type = this.addOption(s, gensection, form.Flag, 'type', _('Bridge interfaces'), _('Creates a bridge over specified interface(s)'));
+			type.modalonly = true;
+			type.disabled = '';
+			type.enabled = 'bridge';
+			type.write = type.remove = function(section_id, value) {
+				var protoname = this.section.formvalue(section_id, 'proto'),
+				    protocol = network.getProtocol(protoname),
+				    new_ifnames = this.isActive(section_id) ? L.toArray(this.section.formvalue(section_id, value ? 'ifname_multi' : 'ifname_single')) : [];
+
+				if (!protocol.isVirtual() && !this.isActive(section_id))
+					return;
+
+				var old_ifnames = [],
+				    devs = ifc.getDevices() || L.toArray(ifc.getDevice());
+
+				for (var i = 0; i < devs.length; i++)
+					old_ifnames.push(devs[i].getName());
+
+				if (!value)
+					new_ifnames.length = Math.max(new_ifnames.length, 1);
+
+				old_ifnames.sort();
+				new_ifnames.sort();
+
+				for (var i = 0; i < Math.max(old_ifnames.length, new_ifnames.length); i++) {
+					if (old_ifnames[i] != new_ifnames[i]) {
+						// backup_ifnames()
+						for (var j = 0; j < old_ifnames.length; j++)
+							ifc.deleteDevice(old_ifnames[j]);
+
+						for (var j = 0; j < new_ifnames.length; j++)
+							ifc.addDevice(new_ifnames[j]);
+
+						break;
+					}
+				}
+
+				if (value)
+					uci.set('network', section_id, 'type', 'bridge');
+				else
+					uci.unset('network', section_id, 'type');
+			};
+		}
+		else {
+			s.tab('devgeneral', _('General device options'));
+			s.tab('devadvanced', _('Advanced device options'));
+			s.tab('brport', _('Bridge port specific options'));
+			s.tab('bridgevlan', _('Bridge VLAN filtering'));
+
+			o = this.addOption(s, gensection, form.ListValue, 'type', _('Device type'));
+			o.readonly = !isNew;
+			o.value('', _('Network device'));
+			o.value('bridge', _('Bridge device'));
+			o.value('8021q', _('VLAN (802.1q)'));
+			o.value('8021ad', _('VLAN (802.1ad)'));
+			o.value('macvlan', _('MAC VLAN'));
+			o.value('veth', _('Virtual Ethernet'));
+
+			o = this.addOption(s, gensection, widgets.DeviceSelect, 'name_simple', _('Existing device'));
+			o.readonly = !isNew;
+			o.rmempty = false;
+			o.noaliases = true;
+			o.default = (dev ? dev.getName() : '');
+			o.ucioption = 'name';
+			o.write = o.remove = setIfActive;
+			o.filter = function(section_id, value) {
+				return !deviceSectionExists(section_id, value);
+			};
+			o.validate = function(section_id, value) {
+				return deviceSectionExists(section_id, value) ? _('A configuration for the device "%s" already exists').format(value) : true;
+			};
+			o.depends('type', '');
+		}
+
+		o = this.addOption(s, gensection, widgets.DeviceSelect, 'ifname_single', isIface ? _('Interface') : _('Base device'));
+		o.readonly = !isNew;
+		o.rmempty = false;
+		o.noaliases = !isIface;
+		o.default = (dev ? dev.getName() : '').match(/^.+\.\d+$/) ? dev.getName().replace(/\.\d+$/, '') : '';
+		o.ucioption = 'ifname';
+		o.validate = function(section_id, value) {
+			var type = this.section.formvalue(section_id, 'type'),
+			    name = this.section.getUIElement(section_id, 'name_complex');
+
+			if (type == 'macvlan' && value && name && !name.isChanged()) {
+				var i = 0;
+
+				while (deviceSectionExists(section_id, '%smac%d'.format(value, i)))
+					i++;
+
+				name.setValue('%smac%d'.format(value, i));
+				name.triggerValidation();
+			}
+
+			return true;
+		};
+		if (isIface) {
+			o.write = o.remove = function() {};
+			o.cfgvalue = function(section_id) {
+				return (ifc.getDevices() || L.toArray(ifc.getDevice())).map(function(dev) {
+					return dev.getName();
+				});
+			};
+			o.onchange = function(ev, section_id, values) {
+				for (var i = 0, co; (co = this.section.children[i]) != null; i++)
+					if (co !== this && co.refresh)
+						co.refresh(section_id);
+
+			};
+			o.depends('type', '');
+		}
+		else {
+			o.write = o.remove = setIfActive;
+			o.depends('type', '8021q');
+			o.depends('type', '8021ad');
+			o.depends('type', 'macvlan');
+		}
+
+		o = this.addOption(s, gensection, form.Value, 'vid', _('VLAN ID'));
+		o.readonly = !isNew;
+		o.datatype = 'range(1, 4094)';
+		o.rmempty = false;
+		o.default = (dev ? dev.getName() : '').match(/^.+\.\d+$/) ? dev.getName().replace(/^.+\./, '') : '';
+		o.validate = function(section_id, value) {
+			var base = this.section.formvalue(section_id, 'ifname_single'),
+			    vid = this.section.formvalue(section_id, 'vid'),
+			    name = this.section.getUIElement(section_id, 'name_complex');
+
+			if (base && vid && name && !name.isChanged()) {
+				name.setValue('%s.%d'.format(base, vid));
+				name.triggerValidation();
+			}
+
+			return true;
+		};
+		o.depends('type', '8021q');
+		o.depends('type', '8021ad');
+
+		o = this.addOption(s, gensection, form.ListValue, 'mode', _('Mode'));
+		o.value('vepa', _('VEPA (Virtual Ethernet Port Aggregator)', 'MACVLAN mode'));
+		o.value('private', _('Private (Prevent communication between MAC VLANs)', 'MACVLAN mode'));
+		o.value('bridge', _('Bridge (Support direct communication between MAC VLANs)', 'MACVLAN mode'));
+		o.value('passthru', _('Pass-through (Mirror physical device to single MAC VLAN)', 'MACVLAN mode'));
+		o.depends('type', 'macvlan');
+
+		if (!isIface) {
+			o = this.addOption(s, gensection, form.Value, 'name_complex', _('Device name'));
+			o.rmempty = false;
+			o.datatype = 'maxlength(15)';
+			o.readonly = !isNew;
+			o.ucioption = 'name';
+			o.write = o.remove = setIfActive;
+			o.validate = function(section_id, value) {
+				return deviceSectionExists(section_id, value) ? _('The device name "%s" is already taken').format(value) : true;
+			};
+			o.depends({ type: '', '!reverse': true });
+		}
+
+		o = this.addOption(s, advsection, form.DynamicList, 'ingress_qos_mapping', _('Ingress QoS mapping'), _('Defines a mapping of VLAN header priority to the Linux internal packet priority on incoming frames'));
+		o.rmempty = true;
+		o.validate = validateQoSMap;
+		o.depends('type', '8021q');
+		o.depends('type', '8021ad');
+
+		o = this.addOption(s, advsection, form.DynamicList, 'egress_qos_mapping', _('Egress QoS mapping'), _('Defines a mapping of Linux internal packet priority to VLAN header priority but for outgoing frames'));
+		o.rmempty = true;
+		o.validate = validateQoSMap;
+		o.depends('type', '8021q');
+		o.depends('type', '8021ad');
+
+		o = this.addOption(s, gensection, widgets.DeviceSelect, 'ifname_multi', _('Bridge ports'));
+		o.size = 10;
+		o.rmempty = true;
+		o.multiple = true;
+		o.noaliases = true;
+		o.nobridges = true;
+		o.ucioption = 'ifname';
+		if (isIface) {
+			o.write = o.remove = function() {};
+			o.cfgvalue = function(section_id) {
+				return (ifc.getDevices() || L.toArray(ifc.getDevice())).map(function(dev) { return dev.getName() });
+			};
+		}
+		else {
+			o.write = o.remove = setIfActive;
+			o.default = L.toArray(dev ? dev.getPorts() : null).filter(function(p) { return p.getType() != 'wifi' || p.isUp() }).map(function(p) { return p.getName() });
+			o.filter = function(section_id, device_name) {
+				var d = network.instantiateDevice(device_name);
+				return d.getType() != 'wifi' || d.isUp();
+			};
+		}
+		o.onchange = function(ev, section_id, values) {
+			ss.updatePorts(values);
+
+			return ss.parse().then(function() {
+				ss.redraw();
+			});
+		};
+		o.depends('type', 'bridge');
+
+		o = this.addOption(s, gensection, form.Flag, 'bridge_empty', _('Bring up empty bridge'), _('Bring up the bridge interface even if no ports are attached'));
+		o.default = o.disabled;
+		o.depends('type', 'bridge');
+
+		o = this.addOption(s, advsection, form.Value, 'priority', _('Priority'));
+		o.placeholder = '32767';
+		o.datatype = 'range(0, 65535)';
+		o.depends('type', 'bridge');
+
+		o = this.addOption(s, advsection, form.Value, 'ageing_time', _('Ageing time'), _('Timeout in seconds for learned MAC addresses in the forwarding database'));
+		o.placeholder = '30';
+		o.datatype = 'uinteger';
+		o.depends('type', 'bridge');
+
+		o = this.addOption(s, advsection, form.Flag, 'stp', _('Enable <abbr title="Spanning Tree Protocol">STP</abbr>'), _('Enables the Spanning Tree Protocol on this bridge'));
+		o.default = o.disabled;
+		o.depends('type', 'bridge');
+
+		o = this.addOption(s, advsection, form.Value, 'hello_time', _('Hello interval'), _('Interval in seconds for STP hello packets'));
+		o.placeholder = '2';
+		o.datatype = 'range(1, 10)';
+		o.depends({ type: 'bridge', stp: '1' });
+
+		o = this.addOption(s, advsection, form.Value, 'forward_delay', _('Forward delay'), _('Time in seconds to spend in listening and learning states'));
+		o.placeholder = '15';
+		o.datatype = 'range(2, 30)';
+		o.depends({ type: 'bridge', stp: '1' });
+
+		o = this.addOption(s, advsection, form.Value, 'max_age', _('Maximum age'), _('Timeout in seconds until topology updates on link loss'));
+		o.placeholder = '20';
+		o.datatype = 'range(6, 40)';
+		o.depends({ type: 'bridge', stp: '1' });
+
+
+		o = this.addOption(s, advsection, form.Flag, 'igmp_snooping', _('Enable <abbr title="Internet Group Management Protocol">IGMP</abbr> snooping'), _('Enables IGMP snooping on this bridge'));
+		o.default = o.disabled;
+		o.depends('type', 'bridge');
+
+		o = this.addOption(s, advsection, form.Value, 'hash_max', _('Maximum snooping table size'));
+		o.placeholder = '512';
+		o.datatype = 'uinteger';
+		o.depends({ type: 'bridge', igmp_snooping: '1' });
+
+		o = this.addOption(s, advsection, form.Flag, 'multicast_querier', _('Enable multicast querier'));
+		o.defaults = { '1': [{'igmp_snooping': '1'}], '0': [{'igmp_snooping': '0'}] };
+		o.depends('type', 'bridge');
+
+		o = this.addOption(s, advsection, form.Value, 'robustness', _('Robustness'), _('The robustness value allows tuning for the expected packet loss on the network. If a network is expected to be lossy, the robustness value may be increased. IGMP is robust to (Robustness-1) packet losses'));
+		o.placeholder = '2';
+		o.datatype = 'min(1)';
+		o.depends({ type: 'bridge', multicast_querier: '1' });
+
+		o = this.addOption(s, advsection, form.Value, 'query_interval', _('Query interval'), _('Interval in centiseconds between multicast general queries. By varying the value, an administrator may tune the number of IGMP messages on the subnet; larger values cause IGMP Queries to be sent less often'));
+		o.placeholder = '12500';
+		o.datatype = 'uinteger';
+		o.depends({ type: 'bridge', multicast_querier: '1' });
+
+		o = this.addOption(s, advsection, form.Value, 'query_response_interval', _('Query response interval'), _('The max response time in centiseconds inserted into the periodic general queries. By varying the value, an administrator may tune the burstiness of IGMP messages on the subnet; larger values make the traffic less bursty, as host responses are spread out over a larger interval'));
+		o.placeholder = '1000';
+		o.datatype = 'uinteger';
+		o.validate = function(section_id, value) {
+			var qiopt = L.toArray(this.map.lookupOption('query_interval', section_id))[0],
+			    qival = qiopt ? (qiopt.formvalue(section_id) || qiopt.placeholder) : '';
+
+			if (value != '' && qival != '' && +value >= +qival)
+				return _('The query response interval must be lower than the query interval value');
+
+			return true;
+		};
+		o.depends({ type: 'bridge', multicast_querier: '1' });
+
+		o = this.addOption(s, advsection, form.Value, 'last_member_interval', _('Last member interval'), _('The max response time in centiseconds inserted into group-specific queries sent in response to leave group messages. It is also the amount of time between group-specific query messages. This value may be tuned to modify the "leave latency" of the network. A reduced value results in reduced time to detect the loss of the last member of a group'));
+		o.placeholder = '100';
+		o.datatype = 'uinteger';
+		o.depends({ type: 'bridge', multicast_querier: '1' });
+
+		o = this.addOption(s, gensection, form.Value, 'mtu', _('MTU'));
+		o.placeholder = getDeviceValue(ifc || dev, 'getMTU');
+		o.datatype = 'max(9200)';
+		o.depends(simpledep);
+
+		o = this.addOption(s, gensection, form.Value, 'macaddr', _('MAC address'));
+		o.placeholder = getDeviceValue(ifc || dev, 'getMAC');
+		o.datatype = 'macaddr';
+		o.depends(simpledep);
+		o.depends('type', 'macvlan');
+		o.depends('type', 'veth');
+
+		o = this.addOption(s, gensection, form.Value, 'peer_name', _('Peer device name'));
+		o.rmempty = true;
+		o.datatype = 'maxlength(15)';
+		o.depends('type', 'veth');
+		o.load = function(section_id) {
+			var sections = uci.sections('network', 'device'),
+			    idx = 0;
+
+			for (var i = 0; i < sections.length; i++)
+				if (sections[i]['.name'] == section_id)
+					break;
+				else if (sections[i].type == 'veth')
+					idx++;
+
+			this.placeholder = 'veth%d'.format(idx);
+
+			return form.Value.prototype.load.apply(this, arguments);
+		};
+
+		o = this.addOption(s, gensection, form.Value, 'peer_macaddr', _('Peer MAC address'));
+		o.rmempty = true;
+		o.datatype = 'macaddr';
+		o.depends('type', 'veth');
+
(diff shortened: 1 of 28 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -0,0 +1,767 @@
...
+				nameinp.triggerValidation();
+			}
+
+			return true;
+		};
+		o.depends('type', '8021q');
+		o.depends('type', '8021ad');
+
+		o = s.taboption('general', form.ListValue, 'mode', _('Mode'));
+		o.value('vepa', _('VEPA (Virtual Ethernet Port Aggregator)', 'MACVLAN mode'));
```
feckert, line 345: Way is there a third argument 'MACVLAN mode' ?

Borromini:
@jow- I should be running the latest netifd git (checked out yesterday, latest commit 645ceed0) but I'm not seeing the LAN bridge (MT7621, DSA). Or is this to be expected with DSA setups?

Schermafdruk van 2020-08-24 21-38-00

jow-:
@Borromini - yes this is expected and prevented atm since it leads to undefined results in netifd. Once we have http://lists.openwrt.org/pipermail/openwrt-devel/2020-July/030397.html *and* a migrated default config, the switch0 / br-lan bridge (or however name we settle on) will appear.

ivanich:
Any progress with this? It seems that discussion on http://lists.openwrt.org/pipermail/openwrt-devel/2020-July/030397.html has been stopped for a while.

Sam-arch99:
I hope this will be implemented soon. I'm still using an older build with swconfig for my WRT3200ACM to avoid DSA as I couldn't find a solid guide on howto convert my current swconfig into a DSA one.

erdoukki:
+1

marcin1j:
I've tested this code and it looks impressive. Unfortunately it's still not production-ready.

LuCI properly handles DSA wired networks but **will mess up wireless configuration** and what's worse, trying to fix wireless you will mess wired config so double check what changes are to be commited!

There seem to be four issues:
1. LuCI don't allow WiFi network referring non-bridge interface. So when you convert network interface type to non-bridge, LuCI wipes its associated WiFi devices' `network` parameter,
2. Conversely, when you try to restore WiFi `network` parameter LuCI converts chosen interface's type to `bridge`, which will break your network (don't even try to commit that!),
3. Wireless network devices are ignored in _Bridge VLAN filtering_ page so you can't assign them from LuCI. Even if you add them manually to `bridge-vlan` section, these entries will disappear once you edit VLAN configuration from UI,
4. Fooling UI by adding wireless devices as bridge members (which is unnecessary for sure; I can't remember for sure but it may also break wireless device VID) kinda works: wireless network devices are shown in VLAN page. However in my case only the first one was properly displayed as untagged. All the others were incorrectly displayed as tagged.

Having said that, for the time being you'd be probably better off converting network configuration using plain old `uci`. It's actually quite simple and should just work with snapshot builds:

1. Define new switch bridge device (let's name it `switch0`) and assign all wired ports (`lan1`, ..., `lan4`, `wan`) as members:
   ```
   uci add network device
   uci set network.@device[-1].type=bridge
   uci set network.@device[-1].name=switch0
   uci add_list network.@device[-1].ifname='lan1'
   uci add_list network.@device[-1].ifname='lan2'
   uci add_list network.@device[-1].ifname='lan3'
   uci add_list network.@device[-1].ifname='lan4'
   uci add_list network.@device[-1].ifname='wan'
   ```
   which adds this entry to `/etc/config/network`:
   ```
   config device
       option type 'bridge'
       option name 'switch0'
       list ifname 'lan1'
       list ifname 'lan2'
       list ifname 'lan3'
       list ifname 'lan4'
       list ifname 'wan'
   ```

2. Define VLAN port mapping by adding bridge-vlan section. Here I'm using VLAN 1 with untagged ports `lan1`,..., `lan3` and tagged `lan4`:
   ```
   uci add network bridge-vlan
   uci set network.@bridge-vlan[-1].device=switch0
   uci set network.@bridge-vlan[-1].vlan=1
   uci add_list network.@bridge-vlan[-1].ports=lan1
   uci add_list network.@bridge-vlan[-1].ports=lan2
   uci add_list network.@bridge-vlan[-1].ports=lan3
   uci add_list network.@bridge-vlan[-1].ports=lan4:t
   
   ```
   which add this entry to `/etc/config/network`:
   ```
   config bridge-vlan
       option device 'switch0'
       option vlan '1'
       list ports 'lan1'
       list ports 'lan2'
       list ports 'lan3'
       list ports 'lan4:t'
   ```

3. Repeat step 2 for each VLAN.

4. Unbridge `lan` network interface:
   ```
   uci del network.lan.type
   ```

5. Set `lan` network `ifname` to `switch0.1` (regardless of how many ports and wireless network are actually assigned; here 1 is VLAN):
   ```
   uci set network.lan.ifname=switch0.1
   ```

6. Repeat steps 4-5 for each network. Remember to change VLAN accordingly.

After this setup, the only bridge should be `switch0`. No network should be of type `bridge`.

Now you can either try to reload network and commit changes if everything works:
```
/etc/init.d/network reload
# Wait and check if network is working properly
uci commit
```
or backup configuration, commit changes and reboot router.
If your device stops responding after `/etc/init.d/network reload`, it's not necessarily a configuration error. In some cases reboot might be needed.


Next step is to fix wireless network setup:

1. I'd recommend assigning fixed wireless network interface names eg. for a `wifi-iface` section named `default_radio0` network device name is fixed as `wl0-lan`:
   ```
   uci set wireless.default_radio0.ifname='wl0-lan'
   ```
   or if you have anonymous `wifi-iface` sections:
   ```
   uci set wireless.@wifi-iface[0].ifname='wl0-lan'
   ```
   You can use LuCI for that but only before "unbridging" networks.

2. Remember to keep wireless `network` parameter. For example `wireless.@wifi-iface` entry named `default_radio0`, being a member of `lan` network should look like this:
   ```
   # uci show wireless.default_radio0
   wireless.default_radio0=wifi-iface
   wireless.default_radio0.device='radio0'
   ...
   wireless.default_radio0.network='lan'
   wireless.default_radio0.ifname='wl0-lan'
   ...
   ```

3. Don't add wireless `ifname` as `switch0` member. Netifd will take care of that once WiFi is brought up!

4. Add wireless network interface name (in this example: `wl0-lan`) as a port in right `bridge-vlan` section.
   For example if you want to add `wl0-lan` network interface as untagged VLAN 1 member and VLAN 1 is defined in first `bridge-vlan` section:
   ```
   uci add_list network.@bridge-vlan[0].ports=wl0-lan
   ```
   which will be commited to `/etc/config/network` as:
   ```
   config bridge-vlan
       option device 'switch0'
       option vlan '1'
       list ports 'lan1'
       list ports 'lan2'
       list ports 'lan3'
       list ports 'lan4:t'
       list ports 'wl0-lan'
   ```

5. Repeat steps 1-4 for each WiFi network.

Once again: don't change `switch0` bridge members when converting WiFi configuration! Only right `bridge-vlan` entries needs to be changed.

You can then reload WiFi networks:
```
wifi
```
and commit changes if everything is working properly.

jow-:
> I've tested this code and it looks impressive. Unfortunately it's still not production-ready.
> 
> LuCI properly handles DSA wired networks but **will mess up wireless configuration** and what's worse, trying to fix wireless you will mess wired config so double check what changes are to be commited!

The scope of this PR has intentionally been to cover the wired part only, wireless interface are explicitly filtered out even in various places. The idea was that you either build bridges containing both switch0.1 and the wireless ifaces, much like it is being done with swconfig today, or that we implement support for `option network-vlan` (see http://lists.openwrt.org/pipermail/openwrt-devel/2020-July/030397.html) one it is implemented. This PR was also designed with a revised default config in mind.

> There seem to be four issues:
> 
>     1. LuCI don't allow WiFi network referring non-bridge interface. So when you convert network interface type to non-bridge, LuCI wipes its associated WiFi devices' `network` parameter,

If you convert an interface from bridge to non-bridge, LuCI will simply remove all but the first ifaces from it, this is why the `option network` configs are removed. Attaching a wifi-iface to a non-bridge network is supported just fine.

>     2. Conversely, when you try to restore WiFi `network` parameter LuCI converts chosen interface's type to `bridge`, which will break your network (don't even try to commit that!),

Likely because the target network is not empty (as in has one iface configured already). So LuCI will turn it into a bridge and add the subsequent ifaces to it.

>     3. Wireless network devices are ignored in _Bridge VLAN filtering_ page so you can't assign them from LuCI. Even if you add them manually to `bridge-vlan` section, these entries will disappear once you edit VLAN configuration from UI,

As written above, this is by design. In OpenWrt's configuration model, we never configure wifi iface names in /etc/config/network since the names are unpredicatable and in some cases there's not even a 1:1 relation of wifi-ifaces to interfaces. For example, one AP/VLAN wifi-iface can result in 1+N different wifi netdevs (e.g. `wlan0`, `wlan0.sta1`, `wlan0.sta2`, ...) that all need to be added to the bridge.

Due to that, wifi-ifaces need to be "attached" to networks through `option network`. Right now that requires an additional bridge but once the pending wifi bridge VLAN patches (http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030734.html, http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030732.html, http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030733.html) are merged, we can think about displaying wifi-ifaces in the VLAN filter table but translate those to `wireless.wifi-iface.network` + `wireless.wifi-iface.vlan` instead of `network.bridge-vlan.ports` options.

>     4. Fooling UI by adding wireless devices as bridge members (which is unnecessary for sure; I can't remember for sure but it may also break wireless device VID) kinda works: wireless network devices are shown in VLAN page. However in my case only the first one was properly displayed as untagged. All the others were incorrectly displayed as tagged.

As explained above, it is highly discouraged to directly add wifi-netdevs as bridge members on OpenWrt, and LuCI does very little to support it.

So to recap; right now a configuration should look like this (non-relevant parts omitted):

```
# /etc/config/network
# [...]
config device
  option type bridge
  option name switch0

config bridge-vlan
    option vlan '1'
    list ports 'lan1'
    list ports 'lan2'
    list ports 'lan3'
    list ports 'lan4:t'

config interface lan
  # For the time being we do need the redundant bridge
  option type bridge
  option ifname switch0.1

# /etc/config/wireless
# [...]
config wifi-iface
  option device radio0
  option ssid Example
  option network lan
```

This should result in a bridge setup like `br-lan(switch0.1, wlanX)` + `switch0(lan1, lan2, lan3, lan4)` with
the layer 3 IP config being on `br-lan`.

Once the RFC patch series mentioned above is merged (and LuCI support has been implemented),
the config can be simplified to:

```
# /etc/config/network
# [...]
config device
  option type bridge
  option name switch0

config bridge-vlan
    option vlan '1'
    list ports 'lan1'
    list ports 'lan2'
    list ports 'lan3'
    list ports 'lan4:t'

config interface lan
  # Note the missing "option type bridge" - netifd now understands that "switch0.1" refers to a
  # bridge VLAN, so it will add the resulting wifi netdevs to the bridge and set "1" as PVID.
  option ifname switch0.1

# /etc/config/wireless
# [...]
config wifi-iface
  option device radio0
  option ssid Example
  # Netifd now knows that lan refers to switch0.1 which is a bridge-vlan entry, so it'll add the
  # resulting wifi netdevs to the switch0 bridge
  option network lan
  
  # This is just added for clarity and actually inherited by default from the target bridge-vlan
  option vlan '1'
```

This should result in a bridge setup like `switch0(lan1, lan2, lan3, lan4, wlanX)` with
the layer 3 IP config being on `switch0.1`.


While your proposed manual config conversion works which manually adds wifi ifaces as ports to bridge-vlans it will fail for AP/VLAN interfaces which result in more than one netdev per wifi-iface.

marcin1j:
Hi

Look at the config that I've posted. It's precisely your example except for the wireless `bridge-vlan` member and works just fine (both wired and wireless). No complaints here (although it's a shame there's still no doc and one needs to read network daemon source code).

BTW Could you elaborate on `wireless.@wifi-iface[].vlan` parameter? Is it already commited (then I must have missed something) or pending or just a proposal? How do we set whether it's tagged or untagged?

My point was that it's not possible - at this stage - to convert and edit configuration from UI and it's behavior can even soft-brick the device.

> > 1. LuCI don't allow WiFi network referring non-bridge interface. So when you convert network interface type to non-bridge, LuCI wipes its associated WiFi devices' `network` parameter,
>
> If you convert an interface from bridge to non-bridge, LuCI will simply remove all but the first ifaces from it, this is why the `option network` configs are removed. 

Sorry, I don't buy this. Purely for reference let's take your example `/etc/config/network`. I want to convert default network/wireless config to your example using UI.

If `lan` network type is `bridge` (which is or at least was the case in default network configuration for DSA enabled devices) then removing that parameter is a mandatory action. Why does UI wipe `wireless.@wifi-iface.network` parameter if - according to your example - that needs to be set to `lan`? At best, it forces user to do an extra, tedious step.

Which leads us to this:

> Attaching a wifi-iface to a non-bridge network is supported just fine.

Sorry, it's not. I admit, it's attached but non-bridge network is forcibly changed to bridge. That's the second issue:

> > 2. Conversely, when you try to restore WiFi `network` parameter LuCI converts chosen interface's type to `bridge`, which will break your network (don't even try to commit that!),
>
> Likely because the target network is not empty (as in has one iface configured already). So LuCI will turn it into a bridge and add the subsequent ifaces to it.

At that point user can either commit the changes (which breaks `lan` networks and soft-bricks the device) or... go back to interfaces config and unbridge `lan` interface again which wipes out `wireless.@wifi-iface.network` parameter he's just set :)

See my point? Having one of these issues is annoying but with extra steps you can convert (or further edit) the configuration. With both of them it's impossible. 

> > 3. Wireless network devices are ignored in _Bridge VLAN filtering_ page so you can't assign them from LuCI. Even if you add them manually to `bridge-vlan` section, these entries will disappear once you edit VLAN configuration from UI,
>
> As written above, this is by design. In OpenWrt's configuration model, we never configure wifi iface names in /etc/config/network since the names are unpredicatable and in some cases there's not even a 1:1 relation of wifi-ifaces to interfaces. For example, one AP/VLAN wifi-iface can result in 1+N different wifi netdevs (e.g. `wlan0`, `wlan0.sta1`, `wlan0.sta2`, ...) that all need to be added to the bridge.

Where did I set wifi interface name in `/etc/config/network`? Of course it's one network interface per ap and fixed interface name can be set in `/etc/config/wireless`. In my example it was set for reference purposes and referenced in network `bridge-vlan` section. I don't like it either and would gladly use another *working* method.

Now for the `wireless.@wifi-iface.vlan` parameter mentioned in your example. If it's supported then the lack of wireless interfaces in _Bridge VLAN filtering_ is not a bug but usability issue (which I believe would be addressed in the future).
But if it's supported, how do we set `wireless.@wifi-iface.vlan` from UI? I couldn't and still can't find relevant field in wireless page.
 
> Due to that, wifi-ifaces need to be "attached" to networks through `option network`. Right now that requires an additional bridge but once the pending wifi bridge VLAN patches (http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030734.html, http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030732.html, http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030733.html) are merged, we can think about displaying wifi-ifaces in the VLAN filter table but translate those to `wireless.wifi-iface.network` + `wireless.wifi-iface.vlan` instead of `network.bridge-vlan.ports` options.

I don't understand this part. Previously there was a single bridge per network. Right now, with vlan filtering it's sufficient to have a single bridge per device regardless of how many networks you have. Is this subject to change?

> > 4. Fooling UI by adding wireless devices as bridge members (which is unnecessary for sure; I can't remember for sure but it may also break wireless device VID) kinda works: wireless network devices are shown in VLAN page. However in my case only the first one was properly displayed as untagged. All the others were incorrectly displayed as tagged.
>
> As explained above, it is highly discouraged to directly add wifi-netdevs as bridge members on OpenWrt, and LuCI does very little to support it.
This was for UI testing purposes. Never mind.

Please tell us more about wireless `vlan` parameter. Is it supported, is it possible to set it with UI?

Please consider the first two issues again. If wireless part of configuration cannot be edited yet, that's understandable. But it's completely different story if you can neither enter nor edit entirely valid network configuration and it gets broken every time you touch it from UI.

jow-:
> Look at the config that I've posted. It's precisely your example except for the wireless `bridge-vlan` member and works just fine (both wired and wireless). No complaints here (although it's a shame there's still no doc and one needs to read network daemon source code).

I didn't say that it won't work, I tried to explain that adding wifi ifnames to the network config is discouraged, precisely due to the 1:n problem of wifi opmodes that result in more than one netdev per wifi-iface. Granted, in a naive WPA-PSK AP setup this case does not apply, but it will with AP/VLAN interfaces. 

> BTW Could you elaborate on `wireless.@wifi-iface[].vlan` parameter? Is it already commited (then I must have missed something) or pending or just a proposal? How do we set whether it's tagged or untagged?

No it is not committed, atm it is a pending proposal. The value format is `<vid>:<suffix>` where suffix may be either omitted or be a specification in the form `:[ut*]+`, e.g. `option vlan '10:t*'` to specify VLAN 10 as egress tagged, PVID.

A `wifi-iface` could theoretically specify something along these lines to attach all resulting wifi netdevs to the corresponding bridge-vlans:

```
config wifi-iface
  option network lan
  list vlan 1:*   # Add to bridge vlan 1, egress untag, PVID
  list vlan 10:t # Add to bridge vlan 10, egress tag
  list vlan 20:t # Add to bridge vlan 20, egress tag
```
If `vlan` is omitted, the wifi netdevs will be treated as egress untag and the VID is inherited from the target network netdev's bridge-vlan membership, so  `option network lan` -> `option ifname switch0.1` -> bridge `switch0`, bridge-vlan ID 1 -> `list vlan 1:u`.

> My point was that it's not possible - at this stage - to convert and edit configuration from UI and it's behavior can even soft-brick the device.

Correct, since neither LuCI nor netifd support all the necessary config semantics yet. Primarily, the support for attaching a `config wifi-iface` to a non-bridge `config interface` referencing a bridge `config device` is lacking in netifd and due to that prohibited in LuCI yet. Due to that, all wifi things are currently filtered away in the ui bridge-vlan config, instead of being implicitly translated to `wireless.wifi-iface.network` like it was done for legacy bridges in the past. This is not a bug however but an intentional omission in this particular implementation stage.

Since OpenWrt is going to revise its default DSA network configuration, and since no config migration is going to happen on the OpenWrt side either, I did not take the ability of a full config rebuild via ui into consideration and instead require a more or less clean state on top of revised default config.

Since the revised OpenWrt default network config will also declare bridge devices explicitly (through `config device; option type bridge; option name switch0`) rather than implicitly (through `config interface lan; option type bridge` -> br-lan) the entire `network.interface.type` option is rendered obsolete and I might consider removing support for it from the ui and instead offer the user some automatic kind of config conversion.

> > > 1. LuCI don't allow WiFi network referring non-bridge interface. So when you convert network interface type to non-bridge, LuCI wipes its associated WiFi devices' `network` parameter,
> > 
> > 
> > If you convert an interface from bridge to non-bridge, LuCI will simply remove all but the first ifaces from it, this is why the `option network` configs are removed.
> 
> Sorry, I don't buy this. Purely for reference let's take your example `/etc/config/network`. I want to convert default network/wireless config to your example using UI.
> 
> If `lan` network type is `bridge` (which is or at least was the case in default network configuration for DSA enabled devices) then removing that parameter is a mandatory action.

In my first example you need to keep the br-lan bridge to bind switch0.1 to wifi, there is no need to remove the bridge config at all. That other example where `config interface lan` is not a bridge is not supported by netifd just yet and prohibited by LuCI due to that, see above.

I was commenting on your general assertion that `LuCI don't allow WiFi network referring [to a] non-bridge interface.`.

> Why does UI wipe `wireless.@wifi-iface.network` parameter if - according to your example - that needs to be set to `lan`? At best, it forces user to do an extra, tedious step.

If you untick the `Bridge interfaces` option on a network that has multiple ifaces attached, then LuCI will remove all but the first one, which usually happens to be a wired one. This causes the removal `option network` from `config wifi-iface`. If you attach a `config wifi-iface` to `config interface` that already has an `option ifname`, then LuCI will force it back to a bridge. If you attach a `config wifi-iface` to a `config interface` that has no `option ifname` and no `option type bridge` then it will work and the IP config is simply inherited.

Having both an `option ifname` in `config interface` and an `option network` in `config wifi-iface` referring to it *is not supported by LuCI* atm, since aforementioned netifd support is not merged yet. Once that support is present, such a configuration will be allowed. Right now it is not within the scope of this PR.

> Which leads us to this:
> 
> > Attaching a wifi-iface to a non-bridge network is supported just fine.
> 
> Sorry, it's not. I admit, it's attached but non-bridge network is forcibly changed to bridge. That's the second issue:

As explained above, it seems you're thinking about a different use-case here (`config interface lan; option ifname switch0.1` plus `config wifi-iface; option network lan`) - this is *not supported yet and not within the scope of this PR* as it depends on unmerged patches [1](http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030734.html), [2](http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030732.html) and [3](http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030733.html)

> > > 1. Conversely, when you try to restore WiFi `network` parameter LuCI converts chosen interface's type to `bridge`, which will break your network (don't even try to commit that!),

I was trying to point out that attaching `wifi-iface` sections to *empty* `config interface` ones will not enforce a bridge. Bridging is only enforced if the logical target interface already has one netdev assigned. This is deliberate at the moment because netifd does not properly handle this case (with `config interface foo; option ifname eth0` + `config wifi-iface; option network foo` it is undefined whether eth0 or wlanX will inherit the IP config, in any case they won't be bridged unless patches [1]-[3] above are merged, only then we can allow this kind of configuration from LuCI.

> At that point user can either commit the changes (which breaks `lan` networks and soft-bricks the device) or... go back to interfaces config and unbridge `lan` interface again which wipes out `wireless.@wifi-iface.network` parameter he's just set :)
> 
> See my point? Having one of these issues is annoying but with extra steps you can convert (or further edit) the configuration. With both of them it's impossible.

Sure I see it, as I said it is a limitation of netifd atm, such a configuration would be semantically invalid with current netifd Git HEAD.

> Where did I set wifi interface name in `/etc/config/network`? 

Here: 
`list ports 'wl0-lan'`

> Of course it's one network interface per ap and fixed interface name can be set in `/etc/config/wireless`.

Not with e.g. `option mode ap; option wds 1` and multiple STAs associated, in this case there will be `wl0-lan`, `wl0-lan.sta1`, `wl0-lan.sta2` and so on. Similar case for various RADIUS configurations.

> In my example it was set for reference purposes and referenced in network `bridge-vlan` section. I don't like it either and would gladly use another _working_ method.

Right now there is no working method without using an additional bridge as an interim solution, until patches [1]-[3] land in netifd.

> Now for the `wireless.@wifi-iface.vlan` parameter mentioned in your example. If it's supported then the lack of wireless interfaces in _Bridge VLAN filtering_ is not a bug but usability issue (which I believe would be addressed in the future).
> But if it's supported, how do we set `wireless.@wifi-iface.vlan` from UI? I couldn't and still can't find relevant field in wireless page.

It is not supported by LuCI yet because it is not supported by netifd yet.

> I don't understand this part. Previously there was a single bridge per network. Right now, with vlan filtering it's sufficient to have a single bridge per device regardless of how many networks you have. Is this subject to change?

It is not subject to change, the end goal is to have one bridge with VLAN filtering (my second example, which is not fully implemented yet). The current implementation level in netifd does not fully support that yet though, so at the moment the only choices are:
 - encode wifi netdev names directly as `list port` option in `bridge-vlan` (discouraged, prevented by LuCI)
 - have an additional bridge for bridging switch0.X with wlanX as interim solution (my first example using two layered bridges)

Once netifd changes are fully implemented, LuCI support will be amended to support this configuration.

> Please tell us more about wireless `vlan` parameter. Is it supported, is it possible to set it with UI?

It is not yet supported yet but once it lands in netifd, I'll do more changes on top of this one to support it. One will both be able to implicitly manage it by adding wifi-netdevs to the bridge filter configuration in the ui or to explicitly set it in the wireless wifi-iface configuration dialog.

marcin1j:
> I didn't say that it won't work, I tried to explain that adding wifi ifnames to the network config is discouraged, precisely due to the 1:n problem of wifi opmodes that result in more than one netdev per wifi-iface. Granted, in a naive WPA-PSK AP setup this case does not apply, but it will with AP/VLAN interfaces.

As I said, I don't like it either but that turned out to be the only option that works.


> No it is not committed, atm it is a pending proposal. The value format is `<vid>:<suffix>` where suffix may be either omitted or be a specification in the form `:[ut*]+`, e.g. `option vlan '10:t*'` to specify VLAN 10 as egress tagged, PVID.

Thanks for explaining. That's a bummer. I'm looking forward to it but it looks like until it's commited we're forced to adding wireless device names to `bridge-vlan` or is there another way?

> > My point was that it's not possible - at this stage - to convert and edit configuration from UI and it's behavior can even soft-brick the device.
> 
> Correct, since neither LuCI nor netifd support all the necessary config semantics yet. Primarily, the support for attaching a `config wifi-iface` to a non-bridge `config interface` referencing a bridge `config device` is lacking in netifd and due to that prohibited in LuCI yet.

I still don't get this. It's possible and works correctly in git master i.e.  adds wireless devices to bridge `switch0` without listing them as `ports` and sets proper vlan id. Or is it working only accidentally?


> Since OpenWrt is going to revise its default DSA network configuration, and since no config migration is going to happen on the OpenWrt side either, I did not take the ability of a full config rebuild via ui into consideration and instead require a more or less clean state on top of revised default config.

So there's going to be yet another forced migration? Nice.

The problem with Luci is that it currently handles neither migration nor modyfing already migrated config. Simply changing network assigned to wireless device break things. 

> Since the revised OpenWrt default network config will also declare bridge devices explicitly (through `config device; option type bridge; option name switch0`) rather than implicitly (through `config interface lan; option type bridge` -> br-lan) the entire `network.interface.type` option is rendered obsolete and I might consider removing support for it from the ui and instead offer the user some automatic kind of config conversion.

Of course it's up to you but IMO something has to be done. LuCI enforces rules that's been correct for a long time but now they're out of sync with netifd. Maybe it should distinguish (regardless of implicit or explicit notation) between old-style bridge per network and new style single bridge per device since both are currently valid even on DSA-enabled devices? If that's not worth the effort simply removing some of the rules would be probably better than enforcing changes that break network connectivity.

> I was commenting on your general assertion that `LuCI don't allow WiFi network referring [to a] non-bridge interface.

Luci won't allow but netifd will (purposely or not). That's my point. Even LuCI with bridge VLAN filtering is out of sync with netifd already commited to master.

> 
> > Why does UI wipe `wireless.@wifi-iface.network` parameter if - according to your example - that needs to be set to `lan`? At best, it forces user to do an extra, tedious step.
> 
> If you untick the `Bridge interfaces` option on a network that has multiple ifaces attached, then LuCI will remove all but the first one, which usually happens to be a wired one. This causes the removal `option network` from `config wifi-iface`. If you attach a `config wifi-iface` to `config interface` that already has an `option ifname`, then LuCI will force it back to a bridge. If you attach a `config wifi-iface` to a `config interface` that has no `option ifname` and no `option type bridge` then it will work and the IP config is simply inherited.
> 
> Having both an `option ifname` in `config interface` and an `option network` in `config wifi-iface` referring to it _is not supported by LuCI_ atm, since aforementioned netifd support is not merged yet. Once that support is present, such a configuration will be allowed. Right now it is not within the scope of this PR.
> ...
> As explained above, it seems you're thinking about a different use-case here (`config interface lan; option ifname switch0.1` plus `config wifi-iface; option network lan`) - this is _not supported yet and not within the scope of this PR_ as it depends on unmerged patches [1](http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030734.html), [2](http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030732.html) and [3](http://lists.openwrt.org/pipermail/openwrt-devel/2020-August/030733.html)
> ...
> I was trying to point out that attaching `wifi-iface` sections to _empty_ `config interface` ones will not enforce a bridge. Bridging is only enforced if the logical target interface already has one netdev assigned. This is deliberate at the moment because netifd does not properly handle this case (with `config interface foo; option ifname eth0` + `config wifi-iface; option network foo` it is undefined whether eth0 or wlanX will inherit the IP config, in any case they won't be bridged unless patches [1]-[3] above are merged, only then we can allow this kind of configuration from LuCI.
> ...
> such a configuration would be semantically invalid with current netifd Git HEAD.

Believe me that - purposely or not - `option network lan` works right now. The only unsupported feature which needs these patches is `wireless.@wifi-iface.vlan` option. All of above implement wireless vlan config. Unless there is another patch not mentioned here - that's not an accident that wireless network option is already supported. Probably the old code handles it just fine.

If it's invalid then netifd does poor job of validating network config but it's good after all - that way it's possible to configure WiFi networks even it's abusing abusing network subsystem a bit.

> > Where did I set wifi interface name in `/etc/config/network`?
> 
> Here:
> `list ports 'wl0-lan'`

Purely semantical: that's a *reference* to interface name. It's *set* in wireless config.

> > Of course it's one network interface per ap and fixed interface name can be set in `/etc/config/wireless`.
> 
> Not with e.g. `option mode ap; option wds 1` and multiple STAs associated, in this case there will be `wl0-lan`, `wl0-lan.sta1`, `wl0-lan.sta2` and so on. Similar case for various RADIUS configurations.

My bad. For clients it's a single AP (yes, I know the scheme, I've used it before :)


> Right now there is no working method without using an additional bridge as an interim solution, until patches [1]-[3] land in netifd.

No. There's working method (intentionally or not) without additional bridge. The only downside is the need to reference wireless device name in VLAN ports sections. When these patches are merged, VLANs are going to be specified in wireless section and refering to wireless devices in `bridge-vlan` network section would be no longer needed. Other than that, nothing changes.

---

Thank you again for explaining everything and excuse my curiosity. There's still no doc. The only piece of information is somewhat unfinished mailing list discussion and network daemon source code.

I'm going to keep network device names referenced in `bridge-vlan` until these patches are merged and then migrate wireless vlan settings away from network sections.

jow-:
> Thanks for explaining. That's a bummer. I'm looking forward to it but it looks like until it's commited we're forced to adding wireless device names to `bridge-vlan` or is there another way?

Apart from an extraneous bridge, not at the moment.

> I still don't get this. It's possible and works correctly in git master i.e. adds wireless devices to bridge `switch0` without listing them as `ports` and sets proper vlan id. Or is it working only accidentally?

I don't get it either because such a config is definitely not working for me, the following uci yields an result where the wifi interfaces are not added to switch0 which is expected since netifd does not handle that yet:

```
root@OpenWrt:/# cat /etc/config/network ; echo "###"; cat /etc/config/wireless
config interface 'loopback'
	option ifname 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'

config globals 'globals'
	option ula_prefix 'fdc2:dccd:22fa::/48'

config device
	option name switch0
	option type bridge

config bridge-vlan
	option device switch0
	option vlan 1
	list ports eth0

config interface 'lan'
	option ifname 'switch0.1'
	option proto 'static'
	option ipaddr '192.168.1.1'
	option netmask '255.255.255.0'
	option ip6assign '60'
	option gateway '192.168.1.2'
	option dns '8.8.8.8'

###

config wifi-device 'radio0'
	option type 'mac80211'
	option channel '36'
	option hwmode '11a'
	option path 'virtual/mac80211_hwsim/hwsim0'
	option htmode 'VHT80'
	option disabled '0'

config wifi-iface 'default_radio0'
	option device 'radio0'
	option network 'lan'
	option mode 'ap'
	option ssid 'OpenWrt'
	option encryption 'none'

config wifi-device 'radio1'
	option type 'mac80211'
	option channel '36'
	option hwmode '11a'
	option path 'virtual/mac80211_hwsim/hwsim1'
	option htmode 'VHT80'
	option disabled '0'

config wifi-iface 'default_radio1'
	option device 'radio1'
	option network 'lan'
	option mode 'ap'
	option ssid 'OpenWrt'
	option encryption 'none'
```

Result: 
```
root@OpenWrt:/# brctl show
bridge name	bridge id		STP enabled	interfaces
switch0		7fff.525400123456	no		eth0
root@OpenWrt:/# 
```

Turning "lan" back into an additional bridge itself will work:

```
root@OpenWrt:/# uci set network.lan.type=bridge
root@OpenWrt:/# uci commit network
root@OpenWrt:/# ifup -a
[  373.663060] switch0: port 1(eth0) entered disabled state
[  373.684620] switch0: port 1(eth0) entered disabled state
root@OpenWrt:/# [  374.914504] 8021q: adding VLAN 0 to HW filter on device eth0
[  374.921890] e1000: eth0 NIC Link is Up 1000 Mbps Full Duplex, Flow Control: RX
[  374.938506] switch0: port 1(eth0) entered blocking state
[  374.941008] switch0: port 1(eth0) entered disabled state
[  374.971839] br-lan: port 1(switch0.1) entered blocking state
[  374.974131] br-lan: port 1(switch0.1) entered disabled state
[  374.976468] device switch0.1 entered promiscuous mode
[  374.978455] device switch0 entered promiscuous mode
[  374.980407] device eth0 entered promiscuous mode
[  375.695347] switch0: port 1(eth0) entered blocking state
[  375.697660] switch0: port 1(eth0) entered forwarding state
[  375.728240] IPv6: ADDRCONF(NETDEV_CHANGE): switch0: link becomes ready
[  375.765229] br-lan: port 1(switch0.1) entered blocking state
[  375.767274] br-lan: port 1(switch0.1) entered forwarding state
[  375.818876] IPv6: ADDRCONF(NETDEV_CHANGE): br-lan: link becomes ready
[  379.973417] br-lan: port 2(wlan1) entered blocking state
[  379.975908] br-lan: port 2(wlan1) entered disabled state
[  379.978300] device wlan1 entered promiscuous mode
[  379.998418] br-lan: port 3(wlan0) entered blocking state
[  380.000775] br-lan: port 3(wlan0) entered disabled state
[  380.002984] device wlan0 entered promiscuous mode
[  380.005100] br-lan: port 3(wlan0) entered blocking state
[  380.007038] br-lan: port 3(wlan0) entered forwarding state
[  380.231608] IPv6: ADDRCONF(NETDEV_CHANGE): wlan1: link becomes ready
[  380.234001] br-lan: port 2(wlan1) entered blocking state
[  380.235972] br-lan: port 2(wlan1) entered forwarding state
[  380.261147] IPv6: ADDRCONF(NETDEV_CHANGE): wlan0: link becomes ready

root@OpenWrt:/# brctl show
bridge name	bridge id		STP enabled	interfaces
br-lan		7fff.525400123456	no		switch0.1
							wlan0
							wlan1
switch0		7fff.525400123456	no		eth0
root@OpenWrt:/# 
```

> The problem with Luci is that it currently handles neither migration nor modyfing already migrated config. Simply changing network assigned to wireless device break things.

This is what I do not understand, I was able to reconfigure the system into the br-lan + switch0 config without any soft-brick or connection loss, I recorded the ui steps here: http://mein.io/rebuild-config-2020-09-29_11.04.25.mp4

Of course this results in the interim solution of using two bridges and I also did some extraneous steps (like disabling bridge on br-lan instead of simply replacing eth0 with switch0.1 and implicitely re-enabling the br-lan bridge by resetting the network association in the wifi config). In any case there was no soft brick. But maybe actual DSA capable hardware behaves differently, which would hint at a bug in OpenWrt though if the same steps result in a brick there.

> Of course it's up to you but IMO something has to be done. LuCI enforces rules that's been correct for a long time but now they're out of sync with netifd. Maybe it should distinguish (regardless of implicit or explicit notation) between old-style bridge per network and new style single bridge per device since both are currently valid even on DSA-enabled devices? 

Unless I totally misunderstood your supposedly working config there aren't any out of sync rules right now.

> Luci won't allow but netifd will (purposely or not). That's my point. Even LuCI with bridge VLAN filtering is out of sync with netifd already commited to master.

As written above, it did not work for me, neither on actual hardware when I developed these changes a few weeks ago nor just now when re-testing on Qemu x86/64.

> Believe me that - purposely or not - `option network lan` works right now. 

Maybe it would make sense to provide the complete /etc/config/network and /etc/config/wireless that is working. I get a hunch that we're talking about different things. Or maybe it is due to your `list ports 'wl0-lan'` in the bridge-vlan section - if yes then *that* is making it work, not `option network` which will be ignored and/or result in undefined behavior in this setup.

> No. There's working method (intentionally or not) without additional bridge. The only downside is the need to reference wireless device name in VLAN ports sections.

Sure, this method is working in some cases but it is discouraged and will not be supported by LuCI. As written earlier, it is not compatible to all operation modes and requires setting fixed wifi netdev names.

marcin1j:
Jo, thank you once more for testing. This looks very confusing but unfortunately I lack the time needed to double check everything tried in the past and this would serve no practical purpose.

As far as I remember, using old approach (bridge per network without `switch0`) used to work but mixing `switch0` with "extraneous" bridges (as forced by LuCI) didn't.

What I'm sure is that configuration with reference to wlan device name in `bridge-vlan` works, whether it's supposed to or not. It also looks the most convenient configuration to migrate once wireless vlan-related netifd code is commited so I'll thick to that.
Having fixed wireless device names is not a concern to me since I've had them fixed for a long time for monitoring purposes. 

Since you've already explained that attaching wireless devices to `switch0` is not supposed to work with `netifd` and that's the reason LuCI forces "the old way" that's supposed to work, let's stop here.
The whole discussion (and misunderstanding) could be avoided if someone's written as little as a short note (even as mailing list post) concluding merging a bunch of commits: _this_ configuration is supposed to work from now on and _that_ configuration is supposed to work once we commit _XYZ_.

Forgive me if you find any part of this or previous comments rude or offensive, of course that was not my intention.

---

For the record: the firmware tested was compiled from https://github.com/openwrt/openwrt/commit/2a90d308c73f3f254e93fe3c55d4e40eaabc06e3 but that shouldn't be of much difference to snapshots binaries. The hardware was Netgear R6220 (ramips; two devices supposed to be the same revision). Handles both tagged and untagged traffic (even mixed on some ports) and most of the networks have attached one or two WiFi APs.



> I don't get it either because such a config is definitely not working for me, the following uci yields an result where the wifi interfaces are not added to switch0 which is expected since netifd does not handle that yet:
> 
> ```
> root@OpenWrt:/# cat /etc/config/network ; echo "###"; cat /etc/config/wireless
> config interface 'loopback'
> 	option ifname 'lo'
> 	option proto 'static'
> 	option ipaddr '127.0.0.1'
> 	option netmask '255.0.0.0'
> 
> config globals 'globals'
> 	option ula_prefix 'fdc2:dccd:22fa::/48'
> 
> config device
> 	option name switch0
> 	option type bridge
> 
> config bridge-vlan
> 	option device switch0
> 	option vlan 1
> 	list ports eth0
> ```

The reference to wlan device is missing here. I don't recall if that made any difference though. Never mind.

> Turning "lan" back into an additional bridge itself will work:

I'm almost certain (too late now to test this again) that it didn't work on my hardware. You seem to be testing this on a virtual environment. There's a couple of factors that could make the difference, including Linux drivers and even hardware implementation.

aparcar:
What's the status of this? It appears to be a release blocker for the next OpenWrt releaes.

Ansuel:
Also some type of documentation for the netifd support of dsa would be very good. Can't find anything on the wiki

jow-:
> What's the status of this? It appears to be a release blocker for the next OpenWrt releaes.

Wireless bridge-vlan support is missing yet.

kubrickfr:
@jow- Thanks for all your contributions, is the Wireless bridge-vlan support tracked somewhere else?

LGA1150:
If wlan interfaces don't work on a VLAN-aware bridge, you can create a separate VLAN-unaware bridge to connect them.

bobafetthotmail:
> > What's the status of this? It appears to be a release blocker for the next OpenWrt releaes.
> 
> Wireless bridge-vlan support is missing yet.

Can it still be merged in its current state? Having a working GUI for ethernet switch, even without wifi support, would be still much better than having no GUI for this at all.

As a workaround the wifi can be placed on a different bridge

jow-:
>     2. Conversely, when you try to restore WiFi `network` parameter LuCI converts chosen interface's type to `bridge`, which will break your network (don't even try to commit that!),

This has been fixed with https://github.com/openwrt/luci/commit/3b4c161e01e2bd19105c123fbec463bc069d637b

>     3. Wireless network devices are ignored in _Bridge VLAN filtering_ page so you can't assign them from LuCI. Even if you add them manually to `bridge-vlan` section, these entries will disappear once you edit VLAN configuration from UI,

LuCI will still filter wireless devices but due to the commit linked above, you can now attach them indirectly to specific bridge-vlans through the existing "option network" setting in the wireless configuration.

Considering the following example network configuration:

```
config device
  option type bridge
  option name switch0

config bridge-vlan
  option device switch0
  option vlan 200
  list ports lan2
  list ports lan3

config interface vlan200
  option ifname switch0.200
  option proto none
```

... you can now set a wifi-iface "option network" to "vlan200" and it'll cause the related `wlan*` netdev to get added as untagged port to the bridge vlan 200 with 200 being the PVID on the wifi iface (so equivalent to a hypothetical `list ports wlan0:u*`).

I would've loved to offer an ability to manage wireless bridge membership directly from the bridge VLAN configuration matrix but the logic turned out to be too complex and brittle so I decided against it. So far upstream insists on keeping `wireless.wifi-iface.network` as sole possibility to tie wifi netdevs into bridges, which unfortunately also means that whenever you want to attach a wireless interface to a bridge-vlan, you'll need a corresponding `config interface` using the bridge vlan layer 2 device as ifname, so that you can target this interface section using `option network` in wireless.

At a later state I'll take a stab at implementing a new `wireless.wifi-iface.bridge` option which will allow specifying the target bridge (vlan) directly, instead of indirectly via a `config interface foo; option ifname targetbridge[.vlan]` section.

hathlife:
when will this PR be backported to 21.02 branch? it's nice if it could be shipped in first 21.02 release.

jow-:
Yes, once it received some more testing.

kubrickfr:
I managed to get my LAN with tagging working on the WRT3200ACM, so that's great. Thanks a ton.
However, if I add wan to switch0 and set it on its own vlan (as it was by default with swconfig), and change the iface of WAN to switch0.2, it doesn't seem the bring the interface up and dhcp client requests fail. Any idea?

ivanich:
it seems that it broke luci interfaces as well, here is my lan bridge interface, it shouldn't look this
изображение

jow-:
@ivanich - it is supposed to look this way if a `config device` with `option name br-lan` exists, in this case, interface level bridging is disabled. Your lan interface should simply specify `Bridge "br-lan"` as interface then.

ivanich:
@jow- the interesting part is that I did full config reset and the defaults looks the same as in the screenshot above. That's all I have after config reset
```
config interface 'lan'
        option type 'bridge'
        option proto 'dhcp'
        option ifname 'lan1 lan2'
```

jow-:
@ivanich - interesting. Can you provide the complete `/etc/config/network` for reference? I'll take a look.

ivanich:
@jow-  here you go, after a full reset
```
root@OpenWrt:~# cat /etc/config/network 

config interface 'loopback'
	option ifname 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'

config globals 'globals'
	option packet_steering '1'
	option ula_prefix 'fdb3:0a1e:9446::/48'

config interface 'lan'
	option type 'bridge'
	option ifname 'lan1 lan2'
	option proto 'static'
	option ipaddr '192.168.1.1'
	option netmask '255.255.255.0'
	option ip6assign '60'

config interface 'wan'
	option ifname 'wan'
	option proto 'dhcp'

config interface 'wan6'
	option ifname 'wan'
	option proto 'dhcpv6'
```

tiagogaspar8:
I had the same happen to me both the bridge issue and an issue in that the br-lan was conflicting with the new bridge in that it was using the same vlan (which it wasn't), also, after setting up the bridge with vlan filtering, I had to go to the device tab and add each interface manually ( by means of "add interface configuration") for them to work properly with the new switch, it was scary 😅

jow-:
@ivanich - that issue should be fixed now.

ivanich:
@jow- confirmed, thanks.

heri16:
I realised luci cannot be used to configure "self": https://unix.stackexchange.com/questions/484644/what-are-self-and-master-options-for-in-bridge-vlan-add

Is this intentional?

jow-:
@heri16 - the `Local` checkbox in the Bridge VLAN config should enable `self`:
image

Mushoz:
Hey, I have been following this PR for a while now, and was wondering if someone could explain to me what exactly this allows us to do inside Luci that isn´t already possible? I receive my internet on my WAN port tagged with VLAN 6 and IPTV tagged with VLAN 4. Furthermore, I have two different subnets (on their own untagged VLANs) and they are correctly separated. This was all configured in 21.02 snapshot inside the Luci interface, no command line needed whatsoever. This is how my /etc/config/network looks like:

```
config interface 'loopback'
	option ifname 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'

config globals 'globals'
	option packet_steering '1'
	option ula_prefix 'redacted'

config interface 'lan'
	option type 'bridge'
	option proto 'static'
	option ipaddr '192.168.1.1'
	option netmask '255.255.255.0'
	option igmp_snooping '1'
	option ip6assign '64'
	option ifname 'lan1 lan2'

config interface 'wan'
	option proto 'pppoe'
	option ifname 'wan.6'
	option username 'redacted'
	option password 'redacted'
	option ipv6 'auto'

config interface 'iptv'
	option proto 'dhcp'
	option ifname 'wan.4'
	option vendorid 'IPTV_RG'
	option defaultroute '0'
	option peerdns '0'

config interface 'untrusted'
	option proto 'static'
	option type 'bridge'
	option ifname 'lan3'
	option igmp_snooping '1'
	option ipaddr '192.168.2.1'
	option netmask '255.255.255.0'
	option ip6assign '64'
```

Given that this was already possible from the GUI, what exactly does this PR allow us to do that isn´t already possible? I am trying to get a better understanding of DSA in the coming 21.02 release and how this PR influences the way we configure this, and was hoping to get a better understanding of it all. Thank you very much in advance!

jow-:
@Mushoz 
- It introduces support for managing `config device` section which is needed for proper mac address overrides, extended physical interface and bridge settings
- It introduces support for bridge-vlan filtering which is needed if you want to run VLAN trunks over bridges and have different bridge ports behave differently wrt. egress tagging, ingress PVID etc.

For your setup (spawning a bunch of VLAN interfaces on top of a standalone `wan`) nothing changes and the features introduces in this PR are probably not relevant.

joaophi:
Hey @jow- , when I save my wireless configuration Luci tries to add the wlan to the lan interface. Maybe it's because I have a custom ifname in my wifi config?

image

hnyman:
@jow- 
Do you feel that there has been enough testing, so that you could backport this to 21.02 ?

tiagogaspar8:
On my setup, when I enable igmp snooping it breaks ipv6

LGA1150:
> On my setup, when I enable igmp snooping it breaks ipv6

Is multicast querier enabled? Or is it an Android device?

tiagogaspar8:
Sorry for the delay, I just got home to make sure of what I'm saying.

So, When I enable The IGMP snooping option in the bridge devices at first get an IPv6 address, but then lose it, after that they never get it again.
Some more important info:

1. When I enable IGMP snooping the multicast querier option is enabled automatically. I have also tried to enable IGMP snooping with the multicast querier option disabled and it still didn't work.
2. I am currently using omcproxy, this package at boot messes with some flags in the kernel, but I don't believe this is the issue since I have tried to boot the router with omcproxy disabled and I experience the same issue.
3. when I enable the IGMP snooping option (and consequently the multicast query one) I notice on wireshark the multicast queries come in IGMPv2 and MLDv1 versions, I have tried to force the version with all the options available in Luci (which some are in places that don't make sense) but it makes no difference.
Anything just ask!

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -649,5 +818,167 @@ return baseclass.extend({
 			o.default = o.disabled;
 			o.depends(Object.assign({ multicast: '1' }, simpledep));
 		}
+
+		o = this.addOption(s, 'bridgevlan', form.Flag, 'vlan_filtering', _('Enable VLAN filterering'));
```
clayface, line 822: filtering not filterering

clayface:
Having issues with bridges > 8 ports in the UI. The MX65 I am porting has 12 ports. Unable to click save.
Screenshot from 2021-04-13 20-24-38_

Issue #4303, opened 2020-07-28
rpcd-mod-luci: only reports active leases from one dnsmasq server


aaronjg:
If multiple dnsmasq servers are running, only the first lease file is opened by `lease_open` in `libs/rpcd-mod-luci/src/luci.c` will only open the first one that it finds, rather than opening all the lease files from all instances of dnsmasq to find all the dhcp leases that have been issued.

socketbox:
Is this related to https://github.com/openwrt/luci/issues/911 ?

aaronjg:
Wow. This is a really old issue. Yes, this issue is a duplicate of #911. It looks like it was never patched. 

How does P/R #4308 look to fix this?

socketbox:
> How does P/R #4308 look to fix this?

Sorry, I haven't a clue. You might consider linking to the PR from that issue, or communicating with one of the three users in the associated thread.

jow-:
Fixed via 0c7a33523

Issue #911, opened 2017-01-01
Active leases not working with mutliple dnsmasq instances


kbabioch:
With the recent inclusion of a patch with support for multiple dnsmasq instances (see [this](https://github.com/lede-project/source/pull/408)), the `Active DHCP Leases` table on the `DHCP and DNS` (`admin/network/dhcp`) no longer works correclty. When defining leasefiles (i.e. `option leasefile /tmp/dnsmasq1.leases` vs `option leasefile /tmp/dnsmasq2.leases`), only leases from the first file are displayed. When using the same file for both dnsmasq instances, there seems to be a race condition and only leases of the last instance that writes to this file are displayed.

Hence LUCI needs to be patched, so it supports multiple leasefiles by iterating over all available leasefiles.

cshoredaniel:
@jow- patch for this backported to 17.01 after testing in trunk?

aaronjg:
Fixed by https://github.com/openwrt/luci/pull/4308

PR #4308, opened 2020-07-29
rpcd-mod-luci: pull dhcp leases from all lease files

aaronjg:
Fixes: https://github.com/openwrt/luci/issues/4303

rpcd mod currently only opens one lease file for each of dnsmasq/odhcpd. With this patch, it will open all the files that uci shows.

maintainer: @jow- 
compiled with: openwrt-sdk-19.07.3-ipq806x-generic_gcc-7.5.0_musl_eabi.Linux-x86_64 

Signed-off-by: Aaron Goodman <aaronjg@stanford.edu>

### DIFF
--- a/libs/rpcd-mod-luci/Makefile
+++ b/libs/rpcd-mod-luci/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rpcd-mod-luci
-PKG_VERSION:=20191114
+PKG_VERSION:=20200729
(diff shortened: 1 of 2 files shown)


aaronjg:
Also fixes #911 (of which #4303 is a dup)

aaronjg:
@jow- Could you take a look at this? I just tested it again on 19.07.4, and it is still working.

aaronjg:
@jow- Could you take a look or assign a reviewer please?

jow-:
Fixed via 0c7a33523

PR #4314, opened 2020-07-30
typo in CloudFlare Security IPs

robrich:


robrich:
See https://blog.cloudflare.com/introducing-1-1-1-1-for-families/

hnyman:
cc @stangri

Sounds like it is not a plain typo, but 1.1.1.2 has different functionality than 1.1.1.1 as .2 blocks malware and .3 would block also adutl material. Not suite sure what stangri originally targeted.

So this commit is likely not about correcting a typo, but about changing the provided functionality.
Additionally, the commit title should start with the app name "luci-app-https-dns-proxy:"

stangri:
Afaik, the bootstrap DNS is only used to resolve the address in resolver URL, not users traffic, so it shouldn't affect the functionality.

I'm not saying we shouldn't use the IP address for bootstrap DNS which is in line with the type of resolver, but it's not a mission critical issue.

stangri:
Proper PRs for this fix for all 3 branches: https://github.com/openwrt/luci/pull/4317 https://github.com/openwrt/luci/pull/4318 https://github.com/openwrt/luci/pull/4319, @hnyman given the nature of this fix, I'm not sure if it should be merged into stable builds, please make this call.

hnyman:
Feel free to merge also into the stable builds.

robrich:
CloudFlare has 3 endpoints all represented in https://github.com/openwrt/luci/tree/master/applications/luci-app-https-dns-proxy/luasrc/https-dns-proxy/providers.  This is the only one with incorrect data.  See https://blog.cloudflare.com/introducing-1-1-1-1-for-families/ and https://1.1.1.1/dns/ for the correct settings.

PR #4315, opened 2020-07-31
luci-app-statistics: support ReportNumCPU

mrjoel:
Signed-off-by: Joel Johnson <mrjoel@lixil.net>

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/cpu.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/cpu.js
@@ -23,6 +23,12 @@ return baseclass.extend({
 		o.rmempty = false;
 		o.depends('enable', '1');
 
+		o = s.option(form.Flag, 'ReportNumCpu', _('Report number of CPUs'),
+			_('When set to true, reports number of available CPUs'));
+		o.default = '0';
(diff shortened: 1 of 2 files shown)


hnyman:
Number of CPUs should be static in a router, so I wonder about the usefulness of this (especially in the LuCI stats)

mrjoel:
> Number of CPUs should be static in a router, so I wonder about the usefulness of this (especially in the LuCI stats)

Sure, it's definitely not interesting in isolation of a single router. It does end up being useful for me in a few different contexts though, including a) single collectd data repo maintained across multiple installs/configurations of virtual machines, b) similarly, shared data history across single "role" of router positions with circulating underlying hardware, and c) sending network reporting to a central log server in order to have a reminder and comparison basis for statistics across multiple hosts.

systemcrash:
I see no harm in having this included. @jow- @hnyman ?

systemcrash:
@mrjoel Could you rebase on current master please?

systemcrash:
In my tests, I could not get this to do anything. Possibly as a result of what plugins area available in cpu.so. After install, the CPU graph is stuck at loading. Restarting collectd didn't seem to fix it, so it's a problem stuck in-between collectd and the GUI. 

Can you get this to work on current master @mrjoel ?

PR #4316, opened 2020-07-31
luci-app-statistics: add users plugin support

mrjoel:
The users plugin is functional and useful, but requires usage of custom
built packages of busybox, dropbear, openssh, or other console/shell
access mechanisms.

Signed-off-by: Joel Johnson <mrjoel@lixil.net>

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/users.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/users.js
@@ -1,0 +1,16 @@
+'use strict';
+'require baseclass';
+'require form';
+
+return baseclass.extend({
+	title: _('Users Plugin Configuration'),
+	description: _('The users plugin collects statistics about users logged in locally via shell. NOTE: Local shell (wtmp) tracking is NOT enabled in default builds. Additional setup is required to get non-zero counts.'),
+
+	addFormOptions: function(s) {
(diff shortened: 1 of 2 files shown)


applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/users.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+'require baseclass';
+'require form';
+
+return baseclass.extend({
+	title: _('Users Plugin Configuration'),
+	description: _('The users plugin collects statistics about users logged in locally via shell. NOTE: Local shell (wtmp) tracking is NOT enabled in default builds. Additional setup is required to get non-zero counts.'),
```
feckert, line 7: Can you please break the line? It's just too long.

feckert:
Openwrt is a single user system
What do I need the plugin for?
Don't I also have to turn on this collectd module or is that built?

mrjoel:
The collectd-mod-users package is already available and built.

My use case is with a rebuilt openssh package doing SSH relaying since it allows tracking SSH connections in use across time.

applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/users.js
systemcrash: @mrjoel

PR #4327, opened 2020-08-03
luci-app-aria2: convert to client side

brvphoenix:


brvphoenix:
@jow- @kuoruan

I have ported the luci-app-aria2 using JavaScript APIs. All the features have been kept except that the function to show the RPC URL has been reworked.

applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js
```diff
@@ -0,0 +1,54 @@
...
+
+return view.extend({
+	load: function() {
+		return uci.load('aria2').then(function() {
+			var config_dir = uci.get_first('aria2', 'aria2', 'config_dir') || '/var/etc/aria2';
+			var config_file = '%s/aria2.conf.main'.format(config_dir);
+			var session_file = '%s/aria2.session.main'.format(config_dir);
+
+			return Promise.all([
+				fs.exec_direct('/bin/cat', [ config_file ]).then(function(res) { return { 'file': config_file, 'rows': res.trim().split(/\n/).length, 'content': res.trim() } }),
```
feckert, line 14: This is a matter of taste, but I think the lines are too long. Can you please reformat them?

applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/log.js
```diff
@@ -0,0 +1,72 @@
...
+	background-color: #33ccff;	\
+}					\
+';
+
+function pollLog(e) {
+	return uci.load('aria2')
+		.then(function() {
+			var logFile = uci.get_first('aria2', 'aria2', 'log') || '/var/log/aria2.log';
+			return Promise.all([
+				fs.exec_direct('/usr/bin/tail', [ '-n', '50', logFile ]).then(function(res) { return res.trim().split(/\n/).reverse().join('\n') }),
```
feckert, line 28: This is a matter of taste, but I think the lines are too long. Can you please reformat them?

applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js
brvphoenix: Is this OK now?

applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/log.js
brvphoenix: Done.

applications/luci-app-aria2/root/usr/share/rpcd/acl.d/luci-app-aria2.json
```diff
@@ -2,6 +2,18 @@
...
+			},
+			"file": {
+				"/etc/group": [ "read" ],
+				"/etc/passwd": [ "read" ],
+				"/bin/cat": [ "exec" ],
+				"/bin/grep": [ "exec" ],
+				"/sbin/logread": [ "exec" ],
+				"/usr/bin/aria2c": [ "exec" ],
+				"/usr/bin/tail": [ "exec" ]
+			},
```
feckert, line 16: Does we need this commands. I have not seen the usage in your code?

- "/etc/group": [ "read" ]
- "/etc/passwd": [ "read" ]

jow-: They’re needed if the UserSelect or GroupSelect ui widgets are used

applications/luci-app-aria2/root/usr/share/rpcd/acl.d/luci-app-aria2.json
```diff
@@ -2,6 +2,18 @@
 	"luci-app-aria2": {
 		"description": "Grant UCI access for luci-app-aria2",
 		"read": {
+			"ubus": {
+				"service": [ "list" ]
+			},
+			"file": {
+				"/etc/group": [ "read" ],
+				"/etc/passwd": [ "read" ],
+				"/bin/cat": [ "exec" ],
```
jow-, line 11: Granting full access to cat and grep will allow read access to any file on the system, this is an undesired privilege escalation.

You shouldn’t need cat, can be replaced with fs.read() and corresponding file read ACLs. As for the grep, I believe you only use it to check for installed webui files. What about probing well known html or js files with fs.stat() instead? This way you can eliminate the need for grep as well.

If you absolutely need this grep functionality, you need to move it into a helper executable which only allows grepping inside the intended target file and not, say, /etc/config/wireless by injecting further command arguments

jow- (changes requested):

applications/luci-app-aria2/root/usr/share/rpcd/acl.d/luci-app-aria2.json
brvphoenix: > Granting full access to cat and grep will allow read access to any file on the system, this is an undesired privilege escalation.

It seems that the privilege escalation is inevitable unless we drop the features of reading log from the file and displaying the current config file. Because these files can be configurable. We can't grant LuCI the read access for every possible files.


> As for the grep, I believe you only use it to check for installed webui files. What about probing well known html or js files with fs.stat() instead? This way you can eliminate the need for grep as well.

OK.

applications/luci-app-aria2/root/usr/share/rpcd/acl.d/luci-app-aria2.json
brvphoenix: Only `UserSelect` is used and I have removed the read access to `/etc/group`.

applications/luci-app-aria2/root/usr/libexec/aria2-call
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+action=$1
+shift
+
+[ -r "$1" ] || exit 1
+
+case "$action" in
+	cat)
+		cat $1
```
jow-, line 9: Instead of allowing arbitrary files here you should pull these values out of the uci configuration. This way a read only user may not be able to view resources other than the ones configured.

applications/luci-app-aria2/root/usr/libexec/aria2-call
```diff
@@ -0,0 +1,14 @@
...
+	;;
+	tail)
+		tail -n 50 $1
```
jow-, line 12: Instead of allowing arbitrary files here you should pull these values out of the uci configuration.

Example:
```bash
log_file=$(uci -q get aria2.@aria2[0].log)
tail -n 50 "${log_file:-/var/log/aria2.log}"
```

brvphoenix: Done

applications/luci-app-aria2/root/usr/libexec/aria2-call
brvphoenix: Done

applications/luci-app-aria2/root/usr/libexec/aria2-call
```diff
@@ -0,0 +1,21 @@
...
+
+. "$IPKG_INSTROOT/usr/share/libubox/jshn.sh"
+
+action=$1
+shift
+
+case "$action" in
+	cat)
+		config_dir="$(uci -q get aria2.main.config_dir)"
+		list_file="${config_dir:-/var/etc/aria2}/aria2.$1.main"
```
jow-, line 11: This is almost golden! :grinning:  But we need to validate `$1` yet in order to prevent path traversal attempts (which likely fail due to the forced `aria2.` prefix and `.main` suffix, but still.

Wrap the above into something like:

```bash
cat)
	case "$1" in
	conf|session)
		config_dir="$(uci -q get aria2.main.config_dir)"
		list_file="${config_dir:-/var/etc/aria2}/aria2.$1.main"
		json_init
		json_add_string file "$list_file"
		json_add_string content "$(cat "$list_file")"
		json_dump
	;;
	*)
		# maybe emit an error here or simply do nothing (in this case, drop the entire "*) ... ;;" case)
	;;
;;
```

brvphoenix: > Wrap the above into something like:

OK

The CI build has failed for some reasons which seems have nothing to do with this LuCI app.

jow-:
Merged, thanks a lot for your patience! 👍🙏

PR #4328, opened 2020-08-03
luci-proto-modemmanger: add missing proto options

valexi7:
Added options to luci proto to match Pull request #13009

https://github.com/openwrt/packages/pull/13009

### DIFF
--- a/protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
+++ b/protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
@@ -117,11 +117,22 @@ return network.registerProtocol('modemmanager', {
 		o.value('ipv6', _('IPv6 only'));
 		o.default = 'ipv4v6';
 
+		s.taboption('general', form.Value, 'signalrate', _('Signal refresh rate'));
+
 		o = s.taboption('advanced', form.Value, 'mtu', _('Override MTU'));
 		o.placeholder = dev ? (dev.getMTU() || '1500') : '1500';
 		o.datatype    = 'max(9200)';
 		
 		s.taboption('general', form.Value, 'metric', _('Gateway metric'));
+		
+		s.taboption('advanced', form.Flag, 'debugmode', _('Enable Debugmode'));
 
+		o = s.taboption('advanced', form.ListValue, 'loglevel', _('Log output level'));
+		o.value('ERR', _('Error'))
+		o.value('WARN', _('Warning'));
+		o.value('INFO', _('Info'));
+		o.value('DEBUG', _('Debug'));
+		o.default = 'ERR';
+		
 	}
 });


feckert:
Please squash the two commits and also add a valid commit header for example `luci-proto-modemmanger: add missing options`

valexi7:
> Please squash the two commits and also add a valid commit header for example `luci-proto-modemmanger: add missing options`

Thanks, done.

feckert:
I probably didn't make myself clear.
You have now changed the heading from P/R.
The [commit](https://github.com/openwrt/luci/pull/4328/commits/8132d94b41f2d26e71fe6d4f3e3c9b3cdb12264f) heading is the same.
Can you please change that too?
The commit heading will appear in the repo and not the P/R heading.

We already have log translation. Could you use this [translation](https://github.com/openwrt/luci/blob/master/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js#L165) please?

feckert:
And also please update your commit message.
We only need one *Signed-off-by*.
Also the commit message should be more meaningful what you have done in this commit.
In the LuCI it is usually sufficient if it is obvious that only the heading of the commit message is entered.
As you have already done. So you need only on *Signed-off-by*. Please have a look at the other commits in this reposotory

protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
```diff
@@ -117,11 +117,22 @@ return network.registerProtocol('modemmanager', {
 		o.value('ipv6', _('IPv6 only'));
 		o.default = 'ipv4v6';
 
+		s.taboption('general', form.Value, 'signalrate', _('Signal refresh rate'));
```
nickberry17, line 120: Hey @valexi7 , can you change it to something like this?  You should validate the input where possible.

```
o = s.taboption('general', form.Value, 'signalrate', _('Signal refresh rate'), _('How often to get the signal strength metrics from the modem, in seconds.'));
o.datatype = "uinteger";
```

protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
```diff
@@ -117,11 +117,22 @@ return network.registerProtocol('modemmanager', {
...
+
 		o = s.taboption('advanced', form.Value, 'mtu', _('Override MTU'));
 		o.placeholder = dev ? (dev.getMTU() || '1500') : '1500';
 		o.datatype    = 'max(9200)';
 		
 		s.taboption('general', form.Value, 'metric', _('Gateway metric'));
-
+		
+		s.taboption('advanced', form.Flag, 'debugmode', _('Enable Debugmode'));
+
```
nickberry17, line 129: Please remove this blank line

nickberry17 (changes requested):

nickberry17:
ping @valexi7

PR #4329, opened 2020-08-03
luci-app-ttyd: add debug value list

feckert:
@ysc3839 Can you see if the combinations match.

Add only allowed debug value as a drop down.

This depends on [pullrequest](https://github.com/openwrt/packages/pull/13011).

### DIFF
--- a/applications/luci-app-ttyd/htdocs/luci-static/resources/view/ttyd/config.js
+++ b/applications/luci-app-ttyd/htdocs/luci-static/resources/view/ttyd/config.js
@@ -70,8 +70,12 @@ return view.extend({
 		o = s.option(form.Value, 'ssl_ca', _('SSL ca'), _('SSL CA file path for client certificate verification'));
 		o.depends('ssl', '1');
 
-		o = s.option(form.Value, 'debug', _('Debug'), _('Set log level (default: 7)'));
-		o.placeholder = '7';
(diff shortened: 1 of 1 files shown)


ysc3839:
I have not run tested, but I think it's right.

https://github.com/tsl0922/ttyd/blob/9edb245aafebf3ed90d136a5afb652d8b0a652d3/src/server.c#L448
ttyd finally passes `debug_level` to `lws_set_log_level`.

https://libwebsockets.org/lws-api-doc-master/html/group__log.html#ga244647f9e1bf0097ccdde66d74f41e26
> level OR together the LLL_ debug contexts you want output from
#define LLL_ERR   (1 << 0)
#define LLL_WARN   (1 << 1)
#define LLL_NOTICE   (1 << 2)
#define LLL_INFO   (1 << 3)
#define LLL_DEBUG   (1 << 4)

And you missed "Debug".

feckert:
> And you missed "Debug".

@ysc3839 I have read that this is only included if we compile the tool with the debug flag enabled?

ysc3839:
@feckert
https://github.com/warmcat/libwebsockets/blob/0e705d410c3f67df282701ae55c22d14adab5798/READMEs/README.coding.md
Confirmed.
> If you are not building with _DEBUG defined, ie, without this
then log levels below notice do not actually get compiled in.

PR #4330, opened 2020-08-03
luci-app-https-dns-proxy: add AliDNS DoH servers

wevsty:
It's difficult to use Cloudflare or the DNS provided by Google in China, which means that users with https-dns-proxy enabled will have trouble accessing the site properly.

Alibaba Group now offers DoH services to regular users, so I added AliDNS servers to give users located in China the option to use them.

Signed-off-by: Yuan Tao ty@wevs.org


stangri:
LGTM, however all commits need to be signed with real names, not aliases. Can you fix that?

wevsty:
> 
> 
> LGTM, however all commits need to be signed with real names, not aliases. Can you fix that?

I'm sorry I'm not familiar with DCO checks, so I tried a lot of forece push.
But it's fixed now.

PR #4342, opened 2020-08-09
treewide: replace `which` with `command -v`

aparcar:
Fix shellcheck SC2230
> which is non-standard. Use builtin 'command -v' instead.

Once applied to everything concerning OpenWrt we can disable the busybox
feature `which` and save 3.8kB.


aparcar:
Patches like these where accepted both in openwrt.git and packages.git, I'll merge this by the end of the week if nobody nacks it.

feckert:
@aparcar :+1: for luci-app-mwan3 and luci-app-dockerman

PR #4347, opened 2020-08-11
[WIP] luci-app-mwan3: interface family dual stack (ipv4+ipv6) support

ptpt52:
interface family accept value "", which indicate ipv4 + ipv6

### DIFF
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/interfaceconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/interfaceconfig.lua
@@ -30,7 +30,8 @@ initial_state:value("online", translate("Online"))
 initial_state:value("offline", translate("Offline"))
 
 family = mwan_interface:option(ListValue, "family", translate("Internet Protocol"))
-family.default = "ipv4"
+family.default = ""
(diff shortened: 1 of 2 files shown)


aaronjg:
I believe @feckert is already working on this:
https://github.com/TDT-AG/luci/commits/pr/20200804-luci-app-mwan3

feckert:
@aaronjg my [pullrequest](https://github.com/openwrt/luci/pull/4349) updates the ruleconfig page.
But the way I see it, this option change proposed by @ptpt52 makes no sense.
I have to choose a family for the interface to work with.
The **empty** option with the translation **Any** does not work in this context.
Or am I wrong?

ptpt52:
hi, this is different from https://github.com/openwrt/luci/pull/4349
here is to make the interface family empty(indicate ipv4+ipv6) possible

Why I add this? yep, for dual stack interface support, it is handle after this PR https://github.com/openwrt/packages/pull/13048

ptpt52:
update: now support dual stack interface and display on status page
this depends on https://github.com/openwrt/packages/pull/13048

feckert:
This pullrequest is no longer applicable, as I have changed the app from lua to javascript in the master.

PR #4349, opened 2020-08-12
luci-app-mwan3: rulconfig variable name refactoring and dependency fixes

feckert:
- Use standard LuCI variable name for map, section and option as m, s and o
- Add missing family selection on rulconfig (any | IPv4 | IPv6) if any is selected the rule is applied to IPv4 and IPv6
- Add dependency to port option. This is only valid if proto is tcp or udp

### DIFF
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
@@ -33,17 +33,6 @@ o:value("info", translate("Info"))
 o:value("debug", translate("Debug"))
 o:depends("logging", "1")
 
-o = s:option(Value, "rtmon_interval",
-	translate("Update interval"),
-	translate("How often should rtmon update the interface routing table"))
-o.datatype = "integer"
-o.default = "5"
-o:value("1", translatef("%d second", 1))
-o:value("3", translatef("%d seconds", 3))
-o:value("5", translatef("%d seconds", 5))
-o:value("7", translatef("%d seconds", 7))
-o:value("10", translatef("%d seconds", 10))
-
 o = s:option(DynamicList, "rt_table_lookup",
 	translate("Routing table lookup"),
 	translate("Also scan this Routing table for connected networks"))
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
@@ -37,8 +37,6 @@ o = s:option(DynamicList, "rt_table_lookup",
 	translate("Routing table lookup"),
 	translate("Also scan this Routing table for connected networks"))
 o.datatype = "integer"
-o:value("1", translatef("Routing table %d", 1))
-o:value("2", translatef("Routing table %d", 2))
 o:value("220", translatef("Routing table %d", 220))
 
 return m
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
@@ -13,7 +13,7 @@ o = s:option(Value, "mmx_mask",
 	translate("Firewall mask"),
 	translate("Enter value in hex, starting with <code>0x</code>"))
 o.datatype = "hex(4)"
-o.default = "0xff00"
+o.default = "0x3F00"
 
 o = s:option(Flag, "logging",
 	translate("Logging"),
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
@@ -3,61 +3,53 @@
 
 local net = require "luci.model.network".init()
 
-local s, m, mask, rtmon, rtlookup, logging, loglevel
+local s, m, o
 
 m = Map("mwan3", translate("MWAN - Globals"))
 
 s = m:section(NamedSection, "globals", "globals", nil)
 
-mask = s:option(
-	Value,
-	"mmx_mask",
+o = s:option(Value, "mmx_mask",
 	translate("Firewall mask"),
 	translate("Enter value in hex, starting with <code>0x</code>"))
-mask.datatype = "hex(4)"
-mask.default = "0xff00"
+o.datatype = "hex(4)"
+o.default = "0xff00"
 
-logging = s:option(Flag,
-	"logging",
+o = s:option(Flag, "logging",
 	translate("Logging"),
 	translate("Enables global firewall logging"))
 
-loglevel = s:option(
-	ListValue,
-	"loglevel",
+o = s:option(ListValue, "loglevel",
 	translate("Loglevel"),
 	translate("Firewall loglevel"))
-loglevel.default = "notice"
-loglevel:value("emerg", translate("Emergency"))
-loglevel:value("alert", translate("Alert"))
-loglevel:value("crit", translate("Critical"))
-loglevel:value("error", translate("Error"))
-loglevel:value("warning", translate("Warning"))
-loglevel:value("notice", translate("Notice"))
-loglevel:value("info", translate("Info"))
-loglevel:value("debug", translate("Debug"))
-loglevel:depends("logging", "1")
+o.default = "notice"
+o:value("emerg", translate("Emergency"))
+o:value("alert", translate("Alert"))
+o:value("crit", translate("Critical"))
+o:value("error", translate("Error"))
+o:value("warning", translate("Warning"))
+o:value("notice", translate("Notice"))
+o:value("info", translate("Info"))
+o:value("debug", translate("Debug"))
+o:depends("logging", "1")
 
-rtmon = s:option(
-	Value,
-	"rtmon_interval",
+o = s:option(Value, "rtmon_interval",
 	translate("Update interval"),
 	translate("How often should rtmon update the interface routing table"))
-rtmon.datatype = "integer"
-rtmon.default = "5"
-rtmon:value("1", translatef("%d second", 1))
-rtmon:value("3", translatef("%d seconds", 3))
-rtmon:value("5", translatef("%d seconds", 5))
-rtmon:value("7", translatef("%d seconds", 7))
-rtmon:value("10", translatef("%d seconds", 10))
+o.datatype = "integer"
+o.default = "5"
+o:value("1", translatef("%d second", 1))
+o:value("3", translatef("%d seconds", 3))
+o:value("5", translatef("%d seconds", 5))
+o:value("7", translatef("%d seconds", 7))
+o:value("10", translatef("%d seconds", 10))
 
-rtlookup = s:option(DynamicList,
-	"rt_table_lookup",
+o = s:option(DynamicList, "rt_table_lookup",
 	translate("Routing table lookup"),
 	translate("Also scan this Routing table for connected networks"))
-rtlookup.datatype = "integer"
-rtlookup:value("1", translatef("Routing table %d", 1))
-rtlookup:value("2", translatef("Routing table %d", 2))
-rtlookup:value("220", translatef("Routing table %d", 220))
+o.datatype = "integer"
+o:value("1", translatef("Routing table %d", 1))
+o:value("2", translatef("Routing table %d", 2))
+o:value("220", translatef("Routing table %d", 220))
 
 return m
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
@@ -30,6 +30,8 @@ o.datatype = ipaddr
 
 o = s:option(Value, "src_port", translate("Source port"),
 	translate("May be entered as a single or multiple port(s) (eg \"22\" or \"80,443\") or as a portrange (eg \"1024:2048\") without quotes"))
+o:depends("proto", "tcp")
+o:depends("proto", "udp")
 
 o = s:option(Value, "dest_ip", translate("Destination address"),
 	translate("Supports CIDR notation (eg \"192.168.100.0/24\") without quotes"))
@@ -37,6 +39,8 @@ o.datatype = ipaddr
 
 o = s:option(Value, "dest_port", translate("Destination port"),
 	translate("May be entered as a single or multiple port(s) (eg \"22\" or \"80,443\") or as a portrange (eg \"1024:2048\") without quotes"))
+o:depends("proto", "tcp")
+o:depends("proto", "udp")
 
 o = s:option(Value, "proto", translate("Protocol"),
 	translate("View the content of /etc/protocols for protocol description"))
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
@@ -18,6 +18,12 @@ s = m:section(NamedSection, arg[1], "rule", "")
 s.addremove = false
 s.dynamic = false
 
+o = s:option(ListValue, "family", translate("Internet Protocol"))
+o.default = ""
+o:value("", translate("IPv4 and IPv6"))
+o:value("ipv4", translate("only IPv4"))
+o:value("ipv6", translate("only IPv6"))
+
 o = s:option(Value, "src_ip", translate("Source address"),
 	translate("Supports CIDR notation (eg \"192.168.100.0/24\") without quotes"))
 o.datatype = ipaddr
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
@@ -5,8 +5,7 @@
 local dsp = require "luci.dispatcher"
 local util   = require("luci.util")
 
-local m, mwan_rule, src_ip, src_port, dest_ip, dest_port, proto, sticky
-local timeout, ipset, logging, policy
+local m, s, o
 
 arg[1] = arg[1] or ""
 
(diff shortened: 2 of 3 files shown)


applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
```diff
@@ -15,62 +14,72 @@ local ipsets = util.split(util.trim(util.exec("ipset -n -L 2>/dev/null | grep -v
...
-mwan_rule = m:section(NamedSection, arg[1], "rule", "")
-mwan_rule.addremove = false
-mwan_rule.dynamic = false
+s = m:section(NamedSection, arg[1], "rule", "")
+s.addremove = false
+s.dynamic = false
 
-src_ip = mwan_rule:option(Value, "src_ip", translate("Source address"),
+o = s:option(ListValue, "family", translate("Internet Protocol"))
+o.default = "ipv4"
```
aaronjg, line 22: Default should be "any" since that is the current default, and this could start switching rules to ipv4 if no option is explicitly set.

aaronjg:
From #4347 to keep the discussion in one place:

> @aaronjg my pullrequest updates the ruleconfig page.
But the way I see it, this option change proposed by @ptpt52 makes no sense.
I have to choose a family for the interface to work with.
The empty option with the translation Any does not work in this context.
Or am I wrong?

Leaving family as empty is fine, since when it is used in the source we use "any" if no default value is set:
https://github.com/openwrt/packages/blob/master/net/mwan3/files/lib/mwan3/mwan3.sh#L881

It's perhaps a bit cleaner to leave it blank, since that is what the luci-firewall package does:
https://github.com/openwrt/luci/blob/master/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js#L217

What do you mean "I have to choose a family for the interface to work with."?

applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
ptpt52: yes default should be "ipv4 + ipv6", but please never use "any" as value and save to config file, we never use "any"
we just leave family to `""` do indicate "any"

ptpt52:
>What do you mean "I have to choose a family for the interface to work with."?

yes, the current code in mwan3 do not handle the interface with family empty
for now we must set family to be `"ipv4"` or `"ipv6"` to make it work with.
and it assume ipv4 if empty family

I may/would be handled by this PR https://github.com/openwrt/packages/pull/13048
where, it assume ipv4+ipv6 if empty family

feckert:
@aaronjg @ptpt52 Let me summarize with thoughts.
It remains to be seen whether it works like this now.

In the interface section of mwan3 the family can be set to IPv4 **or** IPv6 and **not** to any.
So we must decide which address family the interface should track.
So we need at least one interface config for each family so every thing is setup for each address family.
This already exists in the [configuration](https://github.com/openwrt/packages/blob/master/net/mwan3/files/etc/config/mwan3).

And in the rule configuration we can choose on which family the rule gets applied too.
We heave 3 option **IPv4** or **IPv6** or **any**.

As far as I know it is so that if we set in the interface section the option to IPv4 then IPv6 is also setup.
So we could handle both IPv4 and IPv6. This also applies to IPv6 so that we can handle IPv4.

Is that how we want to do it?

Would it be not better to separate them.
If we don't have IPv6 tracking then mwan3 shouldn't install their IPv6 stuff.
The same is true the other way around for IPv4.
If we only have interface section that uses IPv6 then IPv4 should not be handled.

aaronjg:
Yes, that's exactly right - interface must be either ipv4 or ipv6, but a rule can be ipv4, ipv6 or 'any' ('any' is the current default if no value is set).

>As far as I know it is so that if we set in the interface section the option to IPv4 then IPv6 is also setup.
So we could handle both IPv4 and IPv6. This also applies to IPv6 so that we can handle IPv4.

I think it is a bit more complicated than that - you can have policies that have a mixture of IPv4 and IPv6 members, which is when you would want to use a rule that handles both IPv4 and IPv6.

> Would it be not better to separate them.

If we separate them and force policies to have only members of one type, it is going to cause configurations to be even more verbose, so I think it is better to keep as is.

> If we don't have IPv6 tracking then mwan3 shouldn't install their IPv6 stuff.

Personally, I prefer to have mwan3 handle the dual stack by default rather than try to infer which of ipv4 or ipv6 it should manage. I find it a bit cleaner to have it take over all my WANs. I also worry that someone may install mwan3 and set it up on WANA and WANB and forget about WANA6 and WANB6 since traffic will still be going out of WANA6 and they may not realize that IPv6 stack was silently disabled. Perhaps this wouldn't happen. I'd prefer to have an explicit global config option to toggle ipv4 or ipv6. 

I'll defer to you, as maintainer, on this decision though.

ptpt52:
It is currently ok.

the current situation is:
interface must be ipv4 only or ipv6 only, no dual stack

I prefer making mwan3 works with
1. ipv4 only interface
2. ipv6 only interface
3. dual stack interface

All cases can be covered

ptpt52:
by the way
There is such a situation:
wana is ipv6
wana_4 is ipv4

I see it on some ISP network

aaronjg:
Can you describe your dual stack setup a bit more?

With wana ipv6, and wana_4 ipv4, each interface is currently one protocol. ("interface" is the OpenWRT interface. OpenWRT uses "device" to refer to the physical or virtual device e.g. eth0.2, tun0, etc).  With that setup each interface still has only one stack, (even though the device is dual stack). For openwrt configurations, the device is abstracted away and everything should be done on the interface level (though it is true some packages, such as SQM deviate from this, but that tends to cause problems). 

As far as I can tell, the only way to have dual stack on one **interface** is with static ipv4 and ipv6 addresses. Are there other situations when you have a single **interface** with both ipv4 and ipv6?

ptpt52:
>I'd prefer to have an explicit global config option to toggle ipv4 or ipv6.

maybe we don't need it, it is very clear if we do it like this:

1. if all interface is ipv4 only, don't touch on ipv6, mwan3 do work for ipv4 only.

2. if all interface is ipv6 only, don't touch on ipv4, mwan3 do work for ipv4 only.

3. if both ipv4 and ipv6 interface exist, mwan3 work for dual stack.

What do you think? does cover all cases?

Let's think about it together

aaronjg:
also paging @jamesmacwhite to weigh in on the dual stack configurations, since he seems to have the most advanced IPv6 setup.

ptpt52:
> Can you describe your dual stack setup a bit more?
> 
> With wana ipv6, and wana_4 ipv4, each interface is currently one protocol. ("interface" is the OpenWRT interface. OpenWRT uses "device" to refer to the physical or virtual device e.g. eth0.2, tun0, etc). With that setup each interface still has only one stack, (even though the device is dual stack). For openwrt configurations, the device is abstracted away and everything should be done on the interface level (though it is true some packages, such as SQM deviate from this, but that tends to cause problems).
> 
> As far as I can tell, the only way to have dual stack on one **interface** is with static ipv4 and ipv6 addresses. Are there other situations when you have a single **interface** with both ipv4 and ipv6?


dual stack situations:
1. wan is ipv4, virtual wan_6 is ipv6
2. wan is ipv6, virtual wan_4 is ipv4
3. dhcp and dhcpv6 on wan+wan6 (maybe this is not)
4. static ipv4 and ipv6 on wan

aaronjg:
> What do you think? does cover all cases?

I understand the point. The thing I don't like about this is that if you end up disabling devices in mwan3 one by one, you end up with very different behavior when you disable the last device on the stack. For each device that you enable before the last one, you are stopping traffic on that device. Then when you disable the last one, you end up sending everything to the default routing table.  So you are executing a command that appears to have an impact just on the device (uci set mwan3.wanb.enabled=0) but it has an mwan3-wide global impact. 

> wan is ipv4, virtual wan_6 is ipv6

Can you explain this a bit more? Which protocols give you the virtual wan_6/virtual wan_4? How are /etc/config/firewall established for these interfaces?

feckert:
> I'll defer to you, as maintainer, on this decision though.

I would like to finish the [pullrequest](https://github.com/openwrt/packages/pull/13041) of @aaronjg, and then we should fix the ping problem for kernel on version 5.4. After that we could discuss and fix/update/improve the interface family config option (IPv4 | IPv6 | dualstack )

My setup:
VDSL (xdsl over pppoe):
One interface xdsl I get an IPv4 address and an IPv6 address on the same interface.
In mwan3 I only track IPv4 with the interface section xdsl

Cellular (wwan over qmi):
One Interface wwan I get an IPv4 for my new virtual interface wwan_4 and IPv6 for my virtual interface wwan_6.
So I have two additional interface but I do not use theme.
In mwan3 I only track IPv4 with the interface section wwan

ptpt52:
>So you are executing a command that appears to have an impact just on the device (uci set mwan3.wanb.enabled=0) but it has an mwan3-wide global impact.

It seems cause none of problem. All results are user's own request.

>Can you explain this a bit more? Which protocols give you the virtual wan_6/virtual wan_4? How are /etc/config/firewall established for these interfaces?

for most ISP, we setup pppoe on wan, virtual wan_6 is up auto

ptpt52:
> > Can you describe your dual stack setup a bit more?
> > With wana ipv6, and wana_4 ipv4, each interface is currently one protocol. ("interface" is the OpenWRT interface. OpenWRT uses "device" to refer to the physical or virtual device e.g. eth0.2, tun0, etc). With that setup each interface still has only one stack, (even though the device is dual stack). For openwrt configurations, the device is abstracted away and everything should be done on the interface level (though it is true some packages, such as SQM deviate from this, but that tends to cause problems).
> > As far as I can tell, the only way to have dual stack on one **interface** is with static ipv4 and ipv6 addresses. Are there other situations when you have a single **interface** with both ipv4 and ipv6?
> 
> dual stack situations:
> 
> 1. wan is ipv4, virtual wan_6 is ipv6
> 2. wan is ipv6, virtual wan_4 is ipv4
> 3. dhcp and dhcpv6 on wan+wan6 (maybe this is not)
> 4. static ipv4 and ipv6 on wan


@feckert @aaronjg 
So according to @feckert 's  qmi case,

5. qmi setup on wan, virtual wan_4 and wan_6 is up

This is another the dual stack situation

aaronjg:
> One interface xdsl I get an IPv4 address and an IPv6 address on the same interface.
In mwan3 I only track IPv4 with the interface section xdsl

What does that look like? If you call ifstatus xdsl, does it report both the ipv4 and ipv6 addresses? It is the same 'OpenWRT' interface? 

What is the netifd protocol for your cellular device? Do you get the status with ifstatus 'wwan' or 'wwan_4'? Or do both work?

On my setup I get ipv4 and ipv6 addresses via DHCP on the same device (eth0.2), but they are separate OpenWRT interfaces. They are called by default, 'wan' and 'wan6' so, unfortunately, it seems there is no consistency in how things are named in the dual stack setting.

jamesmacwhite:
I know I was pinged on this, but I think a fairly comprehensive discussion has taken place on various elements around this on other PRs, so not sure that I can add too much to it. There will be a lot of different setups in play, so I guess it's just trying to consider all the potential options without forcing a specific "way" of doing something, having mwan3 remain adaptable to different network setups. IPv4 only, dual stack etc.

I think @feckert is right about tackling the key issues and then moving forward on some more potentially breaking changes. I'm happy to help test and help restructure/updating the mwan3 docs: https://openwrt.org/docs/guide-user/network/wan/multiwan/mwan3 if more breaking changes start to be implemented, given the documentation could do with an overhaul and given the increasing use of IPv6 with mwan3 to be more embedded rather than just a small footnote as it is currently, but carefully considering how to approach it without breaking existing configurations unless absolutely required is key. 

There may be a lot of mwan3 users that are using it from an IPv4 only point of view and if IPv6 requires breaking changes for something they weren't routing in the first place, I guess you'd want to avoid something like that.

I have a full dual stack environment for 2x physical WANs and then two Wireguard interfaces, so like @aaronjg said, classed as the more advanced setup side of the scale. If you ever need a test case for IPv6/dual stack. Always ping me and I'll happily test it out. The work that's been happening on mwan3 is great, more than happy to help verify some of the new code and improvements to make it even better when big changes are potentially happening.

aaronjg:
> I have a full dual stack environment for 2x physical WANs and then two Wireguard interfaces, so like @aaronjg said, classed as the more advanced setup side of the scale. If you ever need a test case for IPv6/dual stack

Thanks. I'm trying to learn more about the configuration and how you have this all set up on the /etc/config/networks side.

Do each of these have an ipv4 and an ipv6 entry in /etc/config/networks? So you would have a total of 8 entries in the networks config? Or do some of them use the same interface or create a new interface without an /etc/config/networks entry?

For the wireguard interfaces does it use the same device name for the ipv4 and ipv6 stack, or does it have separate device names for these?

jamesmacwhite:
@aaronjg Hopefully not too off topic and long, but here's some more detail.

>Do each of these have an ipv4 and an ipv6 entry in /etc/config/networks? So you would have a total of 8 entries in the networks config? Or do some of them use the same interface or create a new interface without an /etc/config/networks entry?

Yes. I have followed the approach of splitting network interfaces when doing dual stack as that's what I had read and learnt from other guidance and usage of mwan3 from an IPv6 point of view, whether or not this is an absolute requirement for future use I guess is part of the discussion, but it was my understanding that you should split IPv4 and IPv6 with WAN interfaces for it to work properly with mwan3. 

Because of being able to create alias interfaces in OpenWrt like you can, it will ultimately mean the network interface config from `ifconfig` or whatever has the IPv4 and IPv6 configuration together but for mwan3 purposes allows you to route the traffic either IPv4 or IPv6 conditionally. You are correct I have 8 interface definitions in `/etc/config/network` which then is configured in mwan3 with 8 interface configurations matching with the family of either IPv4 or IPv6 as appropriate. I guess one advantage to this is having very bespoke control of the routing of both protocols for each WAN. i.e. you could mix and match IPv4 over one WAN, while IPv6 over another. Why you'd do that in most circumstances is probably questionable, but you have that level of routing control with the alias interface approach.

Naming wise I follow the interface label then lettering approach.

WAN & WAN6
WANB & WANB6

I could rename the default WAN to WANA and WANA6, but I've just never changed it from the default, because I don't feel the need to, it's clear enough for me. It is also the default naming convention in the standard configuration that mwan3 ships with.

Wireguard I use:

WG & WG6
WGB & WGB6

>For the wireguard interfaces does it use the same device name for the ipv4 and ipv6 stack, or does it have separate device names for these?

It is separate following the alias interface approach as above, in my case for the IPv6 portion it is simply an alias interface with the static IPv6 of the /128 ULA provided by Mullvad. Because of Wireguard and it's static IP requirement currently, it probably works better for mwan3 as the interface configuration is pretty fixed for both protocols.

I did the same with OpenVPN, having an alias interface for IPv6 which worked as well, although the interface configuration was all DHCP, because unlike Wireguard nothing was static and most VPN providers don't provide this, where as Wireguard was designed this way from the beginning.

aaronjg:
> There may be a lot of mwan3 users that are using it from an IPv4 only point of view and if IPv6 requires breaking changes for something they weren't routing in the first place, I guess you'd want to avoid something like that.

Interesting point - if someone is using mwan3 on an IPv4 with VPN, and they don't have IPv6 setup, then mwan3 is currently blocking IPv6 and stopping IPv6 IP address leakage. That is not really a good way to be blocking IPv6 traffic, but it would work well enough for many use cases I suppose, and I'd be surprised if there are no setups like that. 

> Hopefully not too off topic and long, but here's some more detail.

Hah, yes this has gotten quite off topic, but it is helpful discussion.

> Wireguard I use:
WG & WG6
WGB & WGB6

So for these are they set up with protocol `static` or are you using `unmanaged` and you essentially have two interfaces that are doing the same thing?

> I did the same with OpenVPN, having an alias interface for IPv6 which worked as well, although the interface configuration was all DHCP,

Interesting. You had it set up with the protocol in /etc/config/networks as DHCP?  

> I guess one advantage to this is having very bespoke control of the routing of both protocols for each WAN. i.e. you could mix and match IPv4 over one WAN, while IPv6 over another. Why you'd do that in most circumstances is probably questionable, but you have that level of routing control with the alias interface approach.

I think it can be helpful to track these separately. Sometimes my IPv6 link on my WAN goes down while my IPv4 link appears more reliable.

jamesmacwhite:
@aaronjg 

> Hopefully not too off topic and long, but here's some more detail.

Yes my apologies to all on the notifications list for this. Happy to move this to private discussion to avoid further noise!

> So for these are they set up with protocol static or are you using unmanaged and you essentially have two interfaces that are doing the same thing?

Static. The main interface is configured as a Wireguard interface normally with the IPv4 address then having a @wg and @wgb alias with just the static ULA IPv6 address configured for the wg6 and wgb6 as "Alias Interface (Static address)"

> Interesting. You had it set up with the protocol in /etc/config/networks as DHCP?

Yes for OpenVPN, seems to work OK when I tested it with Mullvad, but then moved to Wireguard due to the CPU bottleneck with OpenVPN. I had avoided Wireguard for a while due to the reported issues with mwan3 but then I tested it myself and at least on the 4.14 kernel it works OK like this. As mentioned in that other discussion, it appears there is a problem with the 5.x kernel branch with Wireguard and mwan3 that's not present on 19.07 but is in master branch currently. So that would be something I'd be aware of for the next major release, for now I would say on 19.07 until that was reviewed further.

> I think it can be helpful to track these separately. Sometimes my IPv6 link on my WAN goes down while my IPv4 link appears more reliable.

That's what I have understood to be the recommended approach with mwan3 and dual stack.

applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
feckert: Yes that is true with any and it should not be only ipv4. Thanks.
But I think the default should not be **""** it would be better to set the value to **any**.
In mwan3 the option family is also set to [any](https://github.com/openwrt/packages/blob/master/net/mwan3/files/lib/mwan3/mwan3.sh#L950) if this option is not in the configured

applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
```diff
@@ -3,61 +3,42 @@
...
-mask = s:option(
-	Value,
-	"mmx_mask",
+o = s:option(Value, "mmx_mask",
 	translate("Firewall mask"),
 	translate("Enter value in hex, starting with <code>0x</code>"))
-mask.datatype = "hex(4)"
-mask.default = "0xff00"
+o.datatype = "hex(4)"
+o.default = "0xff00"
```
aaronjg, line 16: While you are fixing default values:

etc/config/mwan3 has a default of 0x3F00, and the docs reflect that ("The firewall mask (default 0x3F00) which is used to mark the outgoing package can be configured in the /etc/config/mwan3 globals section"), so this should be changed as well. Though, the docs are a bit inconsistent here, since there are many references to 250 interfaces being supported, when in reality the number of supported devices is FWMASK - 4 for the reserved marks (0 for unmarked, FWMASK for default, FWMASK - 1 for unreachable and FWMASK - 2 for blackhole).

applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
```diff
@@ -3,61 +3,42 @@
...
+o = s:option(DynamicList, "rt_table_lookup",
 	translate("Routing table lookup"),
 	translate("Also scan this Routing table for connected networks"))
-rtlookup.datatype = "integer"
-rtlookup:value("1", translatef("Routing table %d", 1))
-rtlookup:value("2", translatef("Routing table %d", 2))
-rtlookup:value("220", translatef("Routing table %d", 220))
+o.datatype = "integer"
+o:value("1", translatef("Routing table %d", 1))
+o:value("2", translatef("Routing table %d", 2))
```
aaronjg, line 41: I know that this was broken before your fixes, but "1" and "2" should not be default options here since those routing tables are very likely going to be used by mwan3

feckert: The main issue here was to enter the routes for IPsec in table 220. And mwan3 should have a look at this too. This was a workaround with IPsec. The question is, of course, do we still need it? For that i have to look at IPsec with all the changes in mwan3 again we have done recently. I would leave it like this for now

applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
feckert: You're right, of course. We should change the default here.

In the past this was fixed to 250 and could not be changed.

@jamesmacwhite if you have time could you also update the docs for mmx_mask as mentioned by @aaronjg?

feckert: Default value update for **mmx_mask** done with commit https://github.com/openwrt/luci/pull/4349/commits/9c99b42f034e0b14363c28f993db59ac9809c845

jamesmacwhite: @feckert The globals configuration was entirely absent from the docs until recently, but that should be in place already here with the correct `mmx_mask`: https://openwrt.org/docs/guide-user/network/wan/multiwan/mwan3#globals_configuration.

Realistically, to avoid the wrong number, should the exact value just be dropped in the docs and simply mention mwan3 supports a large amount of both physical and logical network interfaces. Who's actually going to be able to configure 250+ interfaces in a single deployment?!

There is a reference to this statement in LuCI btw (252 interfaces):

image

feckert: Thanks max interface hint fixed with commit https://github.com/openwrt/luci/pull/4349/commits/a7036258f9d8d89ae03014c527a4406c8fcd80f4

applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
aaronjg: I'm not sure if this feature is still needed. If it is, table "220" seems fine to leave as a default. However, there is no reason for a user to add routing table 1 or 2 here, since tables 1 and 2 are very likely going to be used by mwan3 for the first and second interfaces.

applications/luci-app-mwan3/luasrc/model/cbi/mwan/globalsconfig.lua
aaronjg: I think the 252 is correct for a mask of 0xFF00. You can have fwmarks from 0x0100 to 0xFC00. Though, I'm not sure that hint is correct. Conceivably if someone wanted more interfaces for whatever reason, they could change the mask to 0XFFFF and add even more interfaces.

jamesmacwhite: @aaronjg @feckert I think the interface count seems pretty ambiguous, maybe it shouldn't be promoted with an exact number. The amount of interfaces the average use case of mwan3, surely wouldn't be an issue?

aaronjg: Yeah, I think that line could safely be removed. The top line ("There are currently X of Y supported interfaces configured") has the correct number of interfaces, so the "MWAN supports up to..." is not really needed.

feckert:
@aaronjg @ptpt52 @jamesmacwhite this are my fixes for luci to new mwan3 version in the master branch

applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
```diff
@@ -15,62 +14,72 @@ local ipsets = util.split(util.trim(util.exec("ipset -n -L 2>/dev/null | grep -v
...
-mwan_rule.addremove = false
-mwan_rule.dynamic = false
+s = m:section(NamedSection, arg[1], "rule", "")
+s.addremove = false
+s.dynamic = false
 
-src_ip = mwan_rule:option(Value, "src_ip", translate("Source address"),
+o = s:option(ListValue, "family", translate("Internet Protocol"))
+o.default = "any"
+o:value("any", translate("IPv4 and IPv6"))
```
ptpt52, line 23: I prefer use empty as "IPv4 and IPv6"
we treat “” as `IPv4 and IPv6`
and treat "any" as `IPv4 and IPv6`

ptpt52: This is very easy to understand. If a rule does not specify a family, we think it is a dual stack
and In actual programming, this is very practical and effective

applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
```diff
@@ -15,62 +14,72 @@ local ipsets = util.split(util.trim(util.exec("ipset -n -L 2>/dev/null | grep -v
...
+s = m:section(NamedSection, arg[1], "rule", "")
+s.addremove = false
+s.dynamic = false
 
-src_ip = mwan_rule:option(Value, "src_ip", translate("Source address"),
+o = s:option(ListValue, "family", translate("Internet Protocol"))
+o.default = ""
+o:value("", translate("IPv4 and IPv6"))
+o:value("ipv4", translate("only IPv4"))
+o:value("ipv6", translate("only IPv6"))
```
ptpt52, line 25: could we just change the `only` prefix?
I notice the same code in firewall.js
```
        o.value('', _('IPv4 and IPv6'));
        o.value('ipv4', _('IPv4 only'));
```

ptpt52 (commented):
@feckert maybe need to change some text

applications/luci-app-mwan3/luasrc/model/cbi/mwan/ruleconfig.lua
feckert: Done with commit https://github.com/openwrt/luci/commit/1c482dde3d33f79fd7c4e8f131f210b60bc4223c

Nowaker:
I can confirm this MR fixed these problems:
```
Try `ip6tables -h' or 'ip6tables --help' for more information.
ip6tables v1.8.3 (legacy): host/network `xxx' not found
```

The last remaining error that I encounter on `/etc/init.d/mwan3 restart` is:
```
ip6tables v1.8.3 (legacy): The protocol family of set novpn is IPv4, which is not applicable.

Try `ip6tables -h' or 'ip6tables --help' for more information.
```

Its entry in `/etc/config/mwan3`:
```
config rule 'novpn'
        option proto 'all'
        option sticky '0'
        option ipset 'novpn'
        option use_policy 'p_main_novpn'
```

I'm trying to modify the entry via Luci app. I click Edit on this:
image

And get this:
image

I see the UI claims `IPv4 and IPv6` is selected but `/etc/init.d/mwan3` claims it's IPv4. Not sure what the correct setting is, but most certainly seeing any errors on `/etc/init.d/mwan3` is not expected when the config was created solely with the Luci app and `/etc/config/mwan3` was never modified.

aaronjg:
That error is telling you that there is an issue with the novpn ipset. What is the output of `ipset list novpn`? Maybe it is an ipv4 ipset?

Nowaker:
> That error is telling you that there is an issue with the novpn ipset. What is the output of `ipset list novpn`? Maybe it is an ipv4 ipset?

Oh, you're absolutely right. I totally forgot mwan3 luci app does not create ipsets, and I created it myself a long time ago.

I've changed the family to IPv4 and the error is no longer there. Thanks for a tip.

PR #4355, opened 2020-08-16
protocols: add legacymap option for map

RemiNV:
legacymap causes map to use the legacy IPv6 Interface Identifier format
that was described in draft-ietf-softwire-map-00, but was eventually
changed in RFC7597. It is however still used by some major ISPs,
including in Japan.

Signed-off-by: Remi NGUYEN VAN <remi.nguyenvan+openwrt@gmail.com>


feckert:
Sounds plausible. I do not use this option, but I found it in the [protocol handler](https://github.com/openwrt/openwrt/blob/bcf0704bd2ca852b1cbfa481095afb73e78287f9/package/network/ipv6/map/files/map.sh#L234).

Thanks.

PR #4360, opened 2020-08-17
luci-proto-mbim: Add MBIM support

hyc:
Based on luci-proto-qmi
Tested with Sierra Wireless MC7455

Signed-off-by: Howard Chu <hyc@symas.com>

### DIFF
--- a/protocols/luci-proto-mbim/htdocs/luci-static/resources/protocol/mbim.js
+++ b/protocols/luci-proto-mbim/htdocs/luci-static/resources/protocol/mbim.js
@@ -1,0 +1,100 @@
+'use strict';
+'require rpc';
+'require form';
+'require network';
+
+var callFileList = rpc.declare({
+	object: 'file',
+	method: 'list',
+	params: [ 'path' ],
+	expect: { entries: [] },
+	filter: function(list, params) {
+		var rv = [];
+		for (var i = 0; i < list.length; i++)
+			if (list[i].name.match(/^cdc-wdm/))
+				rv.push(params.path + list[i].name);
+		return rv.sort();
+	}
+});
+
+network.registerPatternVirtual(/^mbim-.+$/);
+
+return network.registerProtocol('mbim', {
+	getI18n: function() {
+		return _('MBIM Cellular');
+	},
+
+	getIfname: function() {
+		return this._ubus('l3_device') || 'mbim-%s'.format(this.sid);
+	},
+
+	getOpkgPackage: function() {
+		return 'umbim';
+	},
+
+	isFloating: function() {
+		return true;
+	},
+
+	isVirtual: function() {
+		return true;
+	},
+
+	getDevices: function() {
+		return null;
+	},
+
+	containsDevice: function(ifname) {
+		return (network.getIfnameOf(ifname) == this.getIfname());
+	},
+
+	renderFormOptions: function(s) {
+		var dev = this.getL3Device() || this.getDevice(), o;
+
+		o = s.taboption('general', form.Value, 'device', _('Modem device'));
+		o.rmempty = false;
+		o.load = function(section_id) {
+			return callFileList('/dev/').then(L.bind(function(devices) {
+				for (var i = 0; i < devices.length; i++)
+					this.value(devices[i]);
+				return form.Value.prototype.load.apply(this, [section_id]);
+			}, this));
+		};
+
+		s.taboption('general', form.Value, 'apn', _('APN'));
+		s.taboption('general', form.Value, 'pincode', _('PIN'));
+
+		o = s.taboption('general', form.ListValue, 'auth', _('Authentication Type'));
+		o.value('both', 'PAP/CHAP');
+		o.value('pap', 'PAP');
+		o.value('chap', 'CHAP');
+		o.value('none', 'NONE');
+		o.default = 'none';
+
+		o = s.taboption('general', form.Value, 'username', _('PAP/CHAP username'));
+		o.depends('auth', 'pap');
+		o.depends('auth', 'chap');
+		o.depends('auth', 'both');
+
+		o = s.taboption('general', form.Value, 'password', _('PAP/CHAP password'));
+		o.depends('auth', 'pap');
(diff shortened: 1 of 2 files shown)


hyc:
Should be used in conjunction with https://github.com/openwrt/openwrt/pull/3317

Leo-PL:
Tested-By: Lech Perczak <lech.perczak@gmail.com>
Together with https://github.com/openwrt/openwrt/pull/2759
One suggestion: please add selection of static/DHCP configuration for IP settings, as supported in the mentioned PR - a lot if not the most of MBIM modems use static configuration acquired through MBIM control interface itself - like my Sierra EM7455.

I'd really like to see this merged.

hyc:
From what I can see, DHCP is never needed. My referenced PR` only reads the IP config from MBIM. https://github.com/openwrt/openwrt/pull/3317

Leo-PL:
Agreed, IIRC ModemManager also does not use DHCP when setting up MBIM connection. I have yet to see modem not providing configuration through MBIM natively. 
CC @sch-m

hyc:
Fwiw, I used to work on the modem/radio drivers for Android, and we never used DHCP. https://github.com/xdandroid/hardware_xdandroid-ril/commits/gingerbread (DHCP was only used for wifi.)

hyc:
And yeah, should use https://github.com/openwrt/openwrt/pull/2759 instead of 3317.

Leo-PL:
On my previous build, this plugin was working 100% fine. After I rebased to current master, LuCI keeps deleting device name with every configration edit, and also consistently deletes and replaces "accept_ra" flag for IPv6. You might like to check this, and look at luci-proto-qmi if something changed recently.

hyc:
Seems related to 96ee6dc8d6c9796ab67de6f313a068a4bd3bb20f I guess I have to rebase...

hyc:
@Leo-PL please try again with these updates.

Leo-PL:
Issue with device overwriting is fixed, however, every update I still get this applied:
```
uci del dhcp.wan.ra_flags
uci add_list dhcp.wan.ra_flags='none'
```

hyc:
That doesn't appear to have anything to do with this PR at all.

````
luci> git grep ra_flags
modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:                                       so = ss.taboption('ipv6-ra', cbiRichListValue, 'ra_flags', _('<abbr title="Router Advertisement">RA</abbr> Flags'),
modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:                                               var flags = L.toArray(uci.get('dhcp', section_id, 'ra_flags'));
modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:                                               uci.set('dhcp', section_id, 'ra_flags', [ 'none' ]);
luci> 
````

**Comment by Leo-PL** [comment]
Acknowledged. I'll try to test with other protocols as well.

**Comment by Leo-PL** [comment]
I found one thing that might come in handy, and which `luci-proto-qmi` also lacks, at least as of 19.07.x. If you uncheck "Use peer DNS" option, more "generic" protocols show a list to enter custom DNS servers. It's handy to have those if using MBIM as sole WAN interface. MBIM protocol handler itself understands standard `list dns` entries in /etc/config/network, hidden in 'standard options'.

**Comment by ptpt52** [comment]
mark

File: protocols/luci-proto-mbim/htdocs/luci-static/resources/protocol/mbim.js [hunk 1]
```diff
@@ -0,0 +1,130 @@
...
+		o = s.taboption('advanced', form.Value, 'metric',
+			_('Use gateway metric'));
+			o.placeholder = '0';
+		o.datatype = 'uinteger';
+		o.depends('defaultroute', '1');
+
+		o = s.taboption('advanced', form.Flag, 'peerdns',
+			_('Use DNS servers advertised by peer'),
+			_('If unchecked, the advertised DNS server addresses are ignored'));
+		o.default = o.enabled;
```
ptpt52, line 127: these options are handle by generic, it is likely no need to handle here.

ptpt52: maybe we have to handle `mtu` option (Override MTU)

hyc: and yet it was necessary to add these to the other protocol modules, e.g. fa702c0387dd09a43b1d447e627d118f782ac767

For the most part, this file should be a close copy of the luci-proto-qmi handler.

Leo-PL:
@hyc since openwrt#4405 was merged, maybe it is good idea to resurrect this. I've made some improvements myself, pushed them to https://github.com/Leo-PL/luci/tree/proto-mbim. Minor adjustment will be needed to configure DHCP and DHCPv6 separately, as this was split before merging upstream.

ptpt52:
@Leo-PL great! Maybe opening a new PR can push this change faster.

Leo-PL:
Yeah, could be, although I need to focus on one thing at a time, and I'd like to be fair with @hyc as well. Struggling with ZTE MF28D Ethernet support right now.

hyc:
@Leo-PL don't worry about me, as long as something gets merged to make this all work, that's all that matters.

Leo-PL:
I think it'll be easier if I open new one. I'll likely squash the commits, keeping you as main author, and add needed fixes - this is suitable for adding new package, after all.

Leo-PL:
Ah, during this whole refactoring I forgot about one more thing: restore a way to enable DHCP-only mode. I made some adjustments to protocol handler itself and will open another PR soon. Update for luci-proto-mbim will follow.

hyc:
Sounds good.

Leo-PL:
Created: https://github.com/openwrt/openwrt/pull/12521
Now time for LuCI module.

systemcrash:
@hyc can we close this?

systemcrash:
Let me know if you want to re-open this.

PR #4361, opened 2020-08-18
luci-app-mwan3: fix interface grid

### LINKED ISSUE #4352: luci-app-mwan3: Interfaces grid can become mis-aligned on MWAN3 interface status page
Depending on the content within each interface block, it can cause the alignment to be broken. I believe this could be due to the variable height on each grid block when the interface uptime changes every few seconds.

image

Potentially a CSS tweak is needed, possibly increasing the width slightly or to set a fixed height to avoid it pushing a row further down.

aaronjg:
Interface grid can become misaligned due to varying number of
lines. Use non-breaking spaces to ensure all interface boxes are three
lines long.

Fixes #4352

### DIFF
--- a/applications/luci-app-mwan3/luasrc/view/mwan/overview_status_interface.htm
+++ b/applications/luci-app-mwan3/luasrc/view/mwan/overview_status_interface.htm
@@ -33,7 +33,7 @@ XHR.poll(-1, '<%=luci.dispatcher.build_url("admin", "status", "mwan", "interface
 						case 'online':
 							state = '<%:Online%>';
 							time = String.format(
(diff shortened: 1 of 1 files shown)


feckert:
@jamesmacwhite could you please verify this if this fix your issue?
From my point of view this should fit for a quick fix.
In the long run I want to get rid of all the additional CSS stuff that luci-app-mwan3 introduce, because there are always problems with the different themes.

jamesmacwhite:
@aaronjg Thanks for this. Good idea!

@feckert Yes. This would work as a quick fix to force a consistent number of lines, rather than modifying the width or height of alert message DOM elements.

PR #4369, opened 2020-08-19
luci-proto-wireguard: Add generate key button

wjowsa:
Allow user to generate private key by clicking a button in the wireguard protocol configuration form.


protocols/luci-proto-wireguard/root/usr/libexec/rpcd/luci.wireguard
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+. /usr/share/libubox/jshn.sh
+
+case "$1" in
+	list)
+		printf '{ "getWgKeys": {} }'
```
feckert, line 7: I think we should use this
```
json_init
json_add_object "getWgKeys"
json_close_object
json_dump
```

wjowsa: Thanks, changed

feckert:
@wjowsa Please fixup the commits to one.

jow-:
A minor nitpick... since we're not actually getting (as in reading) keys but generating them, it should be reflected in the function name. Also since generating a key is consuming entropy it could be considered a function with side effects.

I'd propose to rename the function from `getWgKeys` to `generateKeyPair` and to change the `"read": {` scope to `"write": {` in the ACL JSON.

wjowsa:
@feckert done

feckert:
> I'd propose to rename the function from `getWgKeys` to `generateKeyPair` and to change the `"read": {` scope to `"write": {` in the ACL JSON.

@jow- can you please tell me more about why it sould be a write access? Is it because of entropy.
I don't understand what you're trying to do.
Just for the future. So that I can understand it better and pay attention to it in the next review

wjowsa:
@jow-  I have the same question as @feckert . Just to be better in the future :)

jow-:
> can you please tell me more about why it sould be a write access?
Because it uses up entropy. So it is affecting the system state in contrast to e.g. a pure read operation on a file.

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -1,8 +1,14 @@
 'use strict';
 'require uci';
+'require rpc';
 'require form';
 'require network';
 
+var generateKey = rpc.declare({
+	object: 'luci.wireguard',
+	method: 'generateKeyPair'
```
jow-, line 9: I also suggest to add an `expect: { keys: {} }` here...

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -55,6 +61,18 @@ return network.registerProtocol('wireguard', {
 		o.validate = validateBase64;
 		o.rmempty = false;
 
+		o = s.taboption('general', form.Button, 'generate_key', _('Generate Key'), _(''));
+		o.inputstyle = 'apply';
+		o.onclick = L.bind(function(ev) {
+			return  generateKey().then(L.bind(function(result) {
+				var keyInput = document.getElementById("widget.cbid.network." + s.section + ".private_key" );
+				keyInput.value = result.keys.priv;
```
jow-, line 69: ... and changing this to `result.priv || ''` (the `expect` above will take care of only returning the sub-object). You should also add the `|| ''` to avoid setting the input value to `null` in case the call failed.

wjowsa: Thanks for the tip. Changed.

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -55,6 +62,18 @@ return network.registerProtocol('wireguard', {
 		o.validate = validateBase64;
 		o.rmempty = false;
 
+		o = s.taboption('general', form.Button, 'generate_key', _('Generate Key'), _(''));
+		o.inputstyle = 'apply';
+		o.onclick = L.bind(function(ev) {
```
jow-, line 67: Please use the following code (and add an additional `'require ui';` to the view):

```javascript
ui.createHandlerFn(this, function(ev) {
    return generateKey().then(function(keypair) {
        var keyInput = document.getElementById("widget.cbid.network." + s.section + ".private_key"),
           changeEvent = new Event('change');

        keyInput.value = keypair.priv || '';
        keyInput.dispatchEvent(changeEvent);
    });
});
```

This will take care of disabling the button and displaying a loading indicator while the promise is resolving.

wjowsa: Thanks for the tip. Changed.

PR #4373, opened 2020-08-19
luci-base: Add possiblity to set path and timeout of ubus connection

wjowsa:
It might happen (e.g. when downloading a large file with low speed connection) that the ubus connection will time out.
The ubus_connect function allows to specify ubus socket path and ubus connection timeout so it should allowed to set these arguments from luci.


jow-:
Merged, thanks!

PR #4374, opened 2020-08-20
protocols: rename type to maptype for map

RemiNV:
"type" is already used as a common option for all protocols. This makes
the configuration ambiguous, and Luci sees JS errors when trying to save
a MAP configuration.

Switch to "maptype" instead to avoid the conflict. MAP currently uses
"maptype" and falls back to "type" when not specified.

Signed-off-by: Remi NGUYEN VAN <remi.nguyenvan+openwrt@gmail.com>

### DIFF
--- a/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js
+++ b/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js
@@ -40,7 +40,7 @@ return network.registerProtocol('map', {
 	renderFormOptions: function(s) {
 		var o;
 
-		o = s.taboption('general', form.ListValue, 'type', _('Type'));
+		o = s.taboption('general', form.ListValue, 'maptype', _('Type'));
 		o.value('map-e', 'MAP-E');
 		o.value('map-t', 'MAP-T');
 		o.value('lw4o6', 'LW4over6');


feckert:
Unfortunately I am not sure if *type* is really a general option.
Can you show me this?

RemiNV:
Sure, _type_ is documented in the "Options valid for all protocol types" here:
https://openwrt.org/docs/guide-user/base-system/basic-networking

>If set to “bridge”, a bridge containing the given ifnames, and any wireless networks assigned using the network option in UCI wireless configuration, is created

It's also available as an option in protocols without being declared as a protocol-specific parameter; for example in [map.sh](https://github.com/openwrt/openwrt/blob/master/package/network/ipv6/map/files/map.sh#L30) it's available via json_get_vars, even though it's not declared in proto_map_init_config.
map.sh used to repurpose "type" to read it as the map type, but that actually makes configs ambiguous: the doc says it's whether the interfaces must be bridged, but map.sh uses it completely differently. I got a PR merged to map.sh so that it now uses "maptype" first, and "type" only if "maptype" is not set (effectively deprecating "type"), [here](https://github.com/openwrt/openwrt/pull/3314).

Most importantly, Luci gets confused and generates UI controls for the common "type" bridge option, and the map "type", both with the same name; then it gets javascript errors on save because it mixes up the two controls. So currently Luci is just unable to save any "map" config without this patch. See the bug: 
https://bugs.openwrt.org/index.php?do=details&task_id=3287&order=dateopened&sort=desc

I hope this clarifies it !

feckert:
Thanks

Issue #3909, opened 2020-04-14
luci-theme-openwrt-2020: button rendering (wireguard/ advanced-reboot)


pkgadd:
Hi

Using current OpenWrt/ master on ipq806x/ nbg6817 (also seen on ipq40xx/ map-ac2200)

```
master/r12964-35f208da3c
src-git packages https://git.openwrt.org/feed/packages.git^a4d19d8
src-git luci https://git.openwrt.org/project/luci.git^d0daa2c
src-git routing https://git.openwrt.org/feed/routing.git^f71d733
src-git telephony https://git.openwrt.org/feed/telephony.git^bf49bc5
```
I notice a strange rendering for the buttons used in `luci-app-wireguard` and `luci-app-advanced-reboot`, all other buttons I've seen so far seem to work fine - and it's working nicely using `luci-theme-bootstrap`.

![Screenshot_2020-04-14 nbg6817 - Advanced Reboot - LuCI](https://user-images.githubusercontent.com/6666853/79180410-2b553a80-7e0a-11ea-966d-0240da19d9b6.png)
```

<div class="cbi-section"><h3>Interface wg0</h3>
	<div class="cbi-value" id="button" style="padding: 5px">
		<input class="cbi-button cbi-button-apply" type="button" name="qrcode_wg0" value="Show/Hide QR-Code" onclick="toggle_qrcode(this)" />
	</div><div class="cbi-section-node">
```

![Screenshot_2020-04-14 nbg6817 - WireGuard Status - LuCI](https://user-images.githubusercontent.com/6666853/79180414-3019ee80-7e0a-11ea-97f9-d8813eb6c976.png)

```
			<div class="td"><form method="post" action="/cgi-bin/luci/admin/system/advanced_reboot/alternative_reboot">
				<input type="hidden" name="token" value="XXX" />
				<input id="altreboot-button" type="submit" class="cbi-button cbi-button-apply important" value="Reboot to alternative partition..." />
			</form></div>
```
```
			<div class="td"><form method="post" action="/cgi-bin/luci/admin/system/advanced_reboot/reboot">
						<input type="hidden" name="token" value="XXX" />
						<input id="reboot-button" type="submit" class="cbi-button cbi-button-apply important" value="Reboot to current partition" />
					</form></div>
```
```
</fieldset><hr /><form method="post" action="/cgi-bin/luci/admin/system/advanced_reboot/power_off">
	<input type="hidden" name="token" value="XXX" />
[...]

PR #4381, opened 2020-08-23
luci-app-advanced-reboot: Correct board name for linksys-ea3500 (to match updating naming approach)

arrmo:
Tested - corrects failure in advanced-reboot, due to updated board naming (i.e. link below),
https://forum.openwrt.org/t/advanced-reboot-warning-unable-to-obtain-device-information/72276/2

### DIFF
--- a/applications/luci-app-advanced-reboot/luasrc/advanced-reboot/devices/linksys-ea3500.lua
+++ b/applications/luci-app-advanced-reboot/luasrc/advanced-reboot/devices/linksys-ea3500.lua
@@ -1,7 +1,7 @@
 return {
 	vendorName = "Linksys",
 	deviceName = "EA3500",
-	boardNames = { "linksys-audi", "linksys,audi" },
+	boardNames = { "linksys-ea3500", "linksys,ea3500" },
 	partition1MTD = "mtd3",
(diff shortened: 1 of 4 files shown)


stangri:
The new board names should be added to the existing list, not overwrite it.

@hnyman can you comment if I need to add the `btn` class to buttons in the html files (see code changes in this commit)?

arrmo:
> The new board names should be added to the existing list, not overwrite it.

That makes sense, thanks!



> add the `btn` class to buttons in the html files (see code changes in this commit)?

FYI, took these changes from this open issue,
https://github.com/openwrt/luci/issues/3909#issuecomment-675275713

Thanks!

arrmo:
> The new board names should be added to the existing list, not overwrite it.

Actually, thinking about this a bit more - the code changes in master have replaced the names as well - so why not delete the old names (as this is only for master, not older versions)? Thanks!

stangri:
Fixed via #4384, thanks!

PS. Trying to maintain the same code for 18/19/snapshots.

arrmo:
> Trying to maintain the same code for 18/19/snapshots

That makes sense to me - and really appreciate this app. Thanks!!!

PR #4386, opened 2020-08-24
Update https-dns-proxy.po

libinanshu:
Improve simplified Chinese translation


feckert:
Translate for LuCI is done with [Weblate](https://weblate.org/de/).

PR #4394, opened 2020-08-28
luci-mod-admin-full: fix fstab get device by uuid

jjm2473:
UUID is uppercased in NTFS, fstab page always say "not present" because it looks up by lowercased one


feckert:
There must have been some reason why we made this *o *lower*.
Otherwise it would never have worked?
I think we need a different solution.
Only remove this would break some others usage.

jjm2473:
Maybe the author know why. Maybe the author just avoid somebody use uppercased UUID in other filesystems.
Is UUID of NTFS uppercase in normal OpenWRT?

feckert:
It  is for the branch openwrt-18.06 so only @cshoredaniel and  @jow- could help us with this topic.

hnyman:
18.06 is mostly deprecated, so let's forget this for now.

PR #4400, opened 2020-08-30
luci-app-travelmate: fix uci wireless permission

lazywalker:
add uci read/write permissions to make luci-app-travelmate work with luci master, since travelmate needs to write /etc/config/wireless

### DIFF
--- a/applications/luci-app-travelmate/root/usr/share/rpcd/acl.d/luci-app-travelmate.json
+++ b/applications/luci-app-travelmate/root/usr/share/rpcd/acl.d/luci-app-travelmate.json
@@ -2,10 +2,10 @@
 	"luci-app-travelmate": {
 		"description": "Grant UCI access for luci-app-travelmate",
 		"read": {
-			"uci": [ "travelmate" ]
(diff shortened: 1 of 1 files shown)


hnyman:
cc @dibdot

dibdot:
I can't reproduce your issue - what error did you receive? Did you use a special build? Usually all legacy/lua apps have full access to all uci files (luci-compat.json).

lazywalker:
Yes i use a special build, i'm pretty sure that i have luci-compat.ipk pre installed and luci-compat.json is already in there. no error produced, luci-app-travelmate can read data from wireless uci, but the save & commit button just won't work. when i patched luci-app-travelmate.json, everything went OK.

dibdot:
Still a local issue with your build - "Save & Apply" works fine for me (plain vanilla current master). Maybe you need a /etc/init.d/rpcd restart in between to refresh your local ACLs.

dibdot:
@lazywalker I'll close this because it's not reproducible with current master and in the next couple of days/weeks I'll release travelmate 2.0, which has been migrated to client side JS (with a complete new ACL set).

PR #4401, opened 2020-08-30
luci-ssl: switch to WolfSSL

aparcar:
OpenWrt installs WolfSSL per default to offer WPA3. To save some space
switch to a WolfSSL based px5g implementation for self-signed
certificate generation instead of a MbedTLS version.

### DIFF
--- a/collections/luci-ssl/Makefile
+++ b/collections/luci-ssl/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2008-2016 The LuCI Team
+# Copyright (C) 2008-2020 The LuCI Team
 #
 # This is free software, licensed under the Apache License, Version 2.0 .
 #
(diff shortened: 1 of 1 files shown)


aparcar:
Blocked by https://github.com/openwrt/openwrt/pull/3363

collections/luci-ssl/Makefile
```diff
@@ -9,8 +9,8 @@ include $(TOPDIR)/rules.mk
 LUCI_TYPE:=col
 LUCI_BASENAME:=ssl
 
-LUCI_TITLE:=LuCI with HTTPS support (mbedTLS as SSL backend)
-LUCI_DEPENDS:=+luci +libustream-mbedtls +px5g
+LUCI_TITLE:=LuCI with HTTPS support (WolfSSL as SSL backend)
+LUCI_DEPENDS:=+luci +libustream-mbedtls +px5g-wolfssl
```
hnyman, line 13: You are still requiring both mbedtls and wolfssl to be included, as you did NOT change the SSL backend (libustream) itself.

``` 
+libustream-wolfssl +px5g-wolfssl
```

hnyman (changes requested):

collections/luci-ssl/Makefile
hnyman: You could also remove the px5g part from the commit title, as this should handle both the SSL backend and the px5g side at the same time.

hnyman:
A cleanup question for @jow- :

Could we remove libs/luci-lib-px5g/ at the same time?
To my knowledge it is not actually used by anything.

The package's description says that it is needed for "LuCId", which has been removed in 2014 by jow. 
I removed most remaining references to lucid in 2015, but apparently here is still an obscure mention.

https://github.com/openwrt/luci/tree/master/libs/luci-lib-px5g
"RSA/X.509 Key Generator (required for LuCId SSL support)"

collections/luci-ssl/Makefile
aparcar: Done, thanks.

hnyman (approved):

Issue #4077, opened 2020-05-20
19.07.3 regression: installing an app doesn't show in the menu unless logging out/in again


jonglezb:
In 19.07.2, after installing a LuCI app, it started appearing in the menu as soon as you go to a new page.

In 19.07.3, the new app doesn't appear in the menu even after changing page. A logout and login is needed.

I tried the bootstrap theme and the following apps:

- luci-app-openvpn
- luci-app-nlbwmon

To reproduce:

- log in
- go to System / Software
- update lists
- install `luci-app-nlbwmon`
- go to (for instance) Status / Overview
- the "Bandwidth Monitor" menu still does not appear
- log out / log in : the menu now appears as expected

jow-:
Yep, its a known issue which we cannot fix for 19.07.3 anymore. I added a note to the release notes. A fix will entail an update to luci-app-opkg which will flush the SessionStorage menu cache after a package install operation.

singalen:
Is there a workaround? I wouldn't like to reboot.
I tried 
```
/etc/init.d/uhttpd stop
rm -rf /tmp/luci-*
/etc/init.d/uhttpd start
```
but it didn't help.

zorun:
Would it be enough to backport the (secondary) fix from dc57e4bc6c83d6c869c318256057e3636bfc5d66 ?

jow-:
@zorun - yes, most likely. Not entirely sure if we'd need further backports in the dispatcher code wrt. server side cache invalidation

zorun:
It turns out that simply backporting the fix is enough, see #4402

zorun:
The fix is merged, thanks jow!

PR #4402, opened 2020-08-30
[19.07] luci-app-opkg: flush menu cache after opkg actions

zorun:
This is a partial backport of dc57e4bc6c83d6c869c318256057e3636bfc5d66

Fixes: #4077

Signed-off-by: Baptiste Jonglez <git@bitsofnetworks.org>

### DIFF
--- a/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
+++ b/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
@@ -775,6 +775,9 @@ function handleOpkg(ev)
 				E('div', {
 					'class': 'btn',
 					'click': L.bind(function(res) {
+						if (L.ui.menu && L.ui.menu.flushCache)
+							L.ui.menu.flushCache();
+
 						L.hideModal();
 						updateLists();
 


aparcar:
I faced this multiple times, I'll give it a try

zorun:
@aparcar did you managed to test this?

I had tested it by installing luci-app-openvpn and it worked: the menu entry appeared after the installation (I just needed to change page)

Btw, this is considered a release blocker for 19.07.4.

hnyman:
Making it absolutely perfect in 19.07 would likely take too much time compared to the benefit. Even the improvement of immediate action after new package is great.

All in all, the whole OpenWrt with LuCI universe does not expect desktop-level seamlessness. A reboot required here and there is quite ok.

zorun:
> Making it absolutely perfect in 19.07 would likely take too much time compared to the benefit. Even the improvement of immediate action after new package is great.

Agreed that making it perfect is impossible, but backporting a fix for a regression should be fine.

> All in all, the whole OpenWrt with LuCI universe does not expect desktop-level seamlessness. A reboot required here and there is quite ok.
> 
> Ps.
> 
> > this is considered a release blocker for 19.07.4.
> 
> By whom? Quite minor issue in any case.

By me :)

> Quite minor issue in any case.

It's a usability issue that causes confusion.

> (as the issue has been present in all 19.07.x releases, I find it hard to believe that it would really be consider a show-stopper...)

No, it's a regression introduced in 19.07.3, see #4077 . I haven't checked which exact commit caused the regression, but that would probably be 94f504f17864fcdf3447fe40160c59a3e93bbbdf .

Regressions should be fixed, especially when it's easy to do.

aparcar:
But brackets around it and I'm fine with a merge :)

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -775,6 +775,9 @@ function handleOpkg(ev)
 				E('div', {
 					'class': 'btn',
 					'click': L.bind(function(res) {
+						if (L.ui.menu && L.ui.menu.flushCache)
+							L.ui.menu.flushCache();
```
aparcar, line 779:
```suggestion
						if (L.ui.menu && L.ui.menu.flushCache) { L.ui.menu.flushCache(); }
```

aparcar (changes requested):

jow-:
No brackets around single statement conditions please.

aparcar:
> No brackets around single statement conditions please.

I just tested and it works as expected, I wasn't ware of that JavaScript behaviour.

```javascript
if(false)
console.log("foo"); console.log("bar")
```

Only print `'bar`, therefore it's fine.

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
aparcar: Nevermind

zorun:
Yeah, and let's not start adding style changes to backports... Why make this so complicated?

zorun:
Thanks for merging!

IngwiePhoenix:
Came here after installing a few LuCi apps and realizing that the menu did not refresh.

Is there a command-line method to manually refresh the menu cache?

```
root@FriendlyWrt:/etc/tvheadend# opkg info luci
Package: luci
Version: git-20.074.84698-ead5e81
Depends: libc, uhttpd, uhttpd-mod-ubus, luci-mod-admin-full, luci-theme-bootstrap, luci-app-firewall, luci-app-opkg, luci-proto-ppp, libiwinfo-lua, luci-proto-ipv6, rpcd-mod-rrdns
Status: install user installed
Section: luci
Architecture: all
Size: 1045
Filename: luci_git-20.074.84698-ead5e81_all.ipk
Description: Standard OpenWrt set including full admin with ppp support and the default Bootstrap theme
Installed-Time: 1665787481

root@FriendlyWrt:/etc/tvheadend# cat /etc/openwrt_release
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.2'
DISTRIB_REVISION='r19803-9a599fee93'
DISTRIB_TARGET='rockchip/armv8'
DISTRIB_ARCH='aarch64_generic'
DISTRIB_DESCRIPTION='OpenWrt 22.03.2 r19803-9a599fee93'
DISTRIB_TAINTS='busybox'
```

Thanks :)

jow-:
No CLI way since you have to trigger a flush of the session storage in your browser.

PR #4408, opened 2020-09-02
luci-app-statistics: add noavg option

feckert:
It is sometimes useful not to display avg values int the graph.
This option is used to prevent this from appearing in the graph.
In a graph definition this option must be set to *true*,
 if the avg values should not be displayed.

```
options = {
	gauge = {
		title = "Status",
		color = "0000ff",
		noarea = true,
		noavg = true
	}
```


hnyman:
I noticed after the merge that the comment above should have edited, too.
`always include AVERAGE`

Fixed with b95ac83ffd

PR #4414, opened 2020-09-04
luci-app-ddns: add improvments

feckert:
* Update config view, so that the individual sections (info, global) are divided into different tabs. 
* Cleaned up source

@Ansuel could you please have a look if this breaks anything?
You did the intial checkin of this package in javascript.

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -150,7 +150,7 @@ return view.extend({
 
 		m = new form.Map('ddns', _('Dynamic DNS'));
 
-		s = m.section(form.NamedSection, 'global', 'ddns', _('Information'));
+		s = m.section(form.NamedSection, 'global', 'ddns',);
 
 		s.tab('info', _('Information'));
 		s.tab('global', _('Global Settings'));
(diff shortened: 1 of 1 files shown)


Ansuel:
Tested and looks good... Do you have any other suggestion to improve this mess? (a better start/restart for the ddns rules)

feckert:
>  Do you have any other suggestion to improve this mess? (a better start/restart for the ddns rules)

From my point of view, this can certainly be improved. But I have to take a closer look at it again. After merging this P/R.

feckert:
>  Do you have any other suggestion to improve this mess? (a better start/restart for the ddns rules)

@Ansuel I am in the process of refactoring the ddns-scripts package. So it might be that we have to rebuild the luci-app-ddns a bit. Would you be willing to test the changes in the ddns-scripts package when I am ready? I will open a P/R [WIP] in the next days.

Ansuel:
Sure tag me when you open the pr... and thx a lot for the work

feckert:
@Ansuel I pushed a  [P/R](https://github.com/openwrt/packages/pull/13402) in WIP for ddsn-scripts.
Can you have a look at it? And try it if this works for your setup and give me some feedback?

I have seen that when I outsource this to individual JSON files, the size of the installed package changes significantly. I don't see this as a problem, because I would improve other things that works against it. For example _maintainability_. It is also not always necessary to install all providers. Which is not possible now.

As it is, there are some hacks that are not easy to maintain and therefore not necessary if we outsource it to JSON.
Also the question is if we really need all providers.

PR #4421, opened 2020-09-08
master: mosquitto: support notifications_local_only flag

karlp:
Supported since mosquitto 1.5 released in May 2018, and has long
been supported in the init scripts.


jow-:
Merged, thanks!

PR #4429, opened 2020-09-11
luci-app-mwan3: Update default values for down, up, timeout and interval on interfaces

jamesmacwhite:
@feckert LuCI currently sets the following defaults if they are not directly set on interface configurations.

 Option | LuCI default  | mwan3track default  |
| :------------- |-------------| -----|
| `up`  | `3` | `5` |
| `down`  | `3` |  `5` |
| `interval` | `5` | `10` |
| `timeout` | `2` | `4` |

These are different to the defaults specified by mwan3track, so this updates them to match.

### DIFF
--- a/applications/luci-app-mwan3/luasrc/model/cbi/mwan/interfaceconfig.lua
+++ b/applications/luci-app-mwan3/luasrc/model/cbi/mwan/interfaceconfig.lua
@@ -148,7 +148,7 @@ recovery_loss:value("20")
 recovery_loss:value("25")
 
 timeout = mwan_interface:option(ListValue, "timeout", translate("Ping timeout"))
-timeout.default = "2"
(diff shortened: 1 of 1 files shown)


jamesmacwhite:
@feckert Did you happen to see this one, fixes some defaults that are currently mismatched with mwan3track.

feckert:
@jamesmacwhite We can merge this independently of the[ P/R](https://github.com/openwrt/packages/pull/13169).
Could you please update the header prefix of you commits? So we have `luci-app-mwan3: xxxx`?

jamesmacwhite:
@feckert Thanks. I've amended the commits to have `luc-app-mwan3:` on each commit message.

feckert:
:+1: thanks

PR #4432, opened 2020-09-11
luci-lib-nixio: update TLS support

wevsty:
luci-lib-nixio: Update TLS support

Changelog:
Modify makefile.
Replace cyassl to wolfSSL.
Remove axTLS support.
Replace the deprecated function:
    TLSv1_client_method()
    TLSv1_server_method()

Signed-off-by: Yuan Tao <ty@wevs.org>


aparcar:
Sorry the patch does not apply on my master branch of LuCI, am I missing something?

wevsty:
@aparcar 
Since the code repository integrates the complete axTLS code with some line feeds and non-UTF-8 characters, using git am to merge patches directly will cause some errors.


Here I recommend just cloning from my branch to generate the patch and then using git am to open the merge.
```bash
git clone -b update_nixio https://github.com/wevsty/luci.git
cd luci
git format-patch --binary -1
# This should generate 0001-luci-lib-nixio-update-TLS-support.patch, copy the file.
# cd to the need patch luci path
# Finally, use the git am merge patch (need to add --ignore-whitespace)
git am --ignore-whitespace 0001-luci-lib-nixio-update-TLS-support.patch
```
If there are still questions you can post the error log.

seuros:
Can we have this patch merged ?

aparcar:
@jow- comments?

Neustradamus:
It will be possible to have in 22.03?
This PR is very old and good...

jow-:
The regression potential is too high, also there's no TLS users for nixio left in LuCI, so I likely won't touch it anymore in the future.

Neustradamus:
@jow-: I think it is better to use current system that an old and unsecure...
- https://www.cvedetails.com/vulnerability-list/vendor_id-17579/product_id-43200/Axtls-Project-Axtls.html

libs/luci-lib-nixio/src/tls-context.c
```diff
@@ -54,20 +54,20 @@ static int nixio_tls_ctx(lua_State * L) {
...
 
 	if (!(*ctx)) {
 		return luaL_error(L, "unable to create TLS context");
 	}
-
-#ifdef WITH_CYASSL
+	/* Setting the minimum version of SSL */
+	SSL_CTX_set_min_proto_version(*ctx,TLS1_2_VERSION);
+	/* Setting do not verify certificate */
 	SSL_CTX_set_verify(*ctx, SSL_VERIFY_NONE, NULL);
```
cotequeiroz, line 70: Here's one point we should be looking to change.  There's not much of a point in using a TLS connection if we are not verifying the server certificate.

cotequeiroz (changes requested):
I've reviewed your changes.  They are all good, and most of my suggestions are nitpicks.  

However, an old problem--not introduced by this PR--remains in the code.  There is no peer validation being done.  This undermines the efforts to truly justify TLS support.

Neustradamus:
@cotequeiroz: Thanks for your review, I hope that @wevsty will see your comments and will change it ^^

wevsty:
> @cotequeiroz: Thanks for your review, I hope that @wevsty will see your comments and will change it ^^

I saw that, if I need to change the PR, I need to retest it, please give me a few days to work on it.

cotequeiroz:
One more suggestion:  split this into different commits:
- Remove axTLS support from nixio proper (i.e. changes outside of `ax*/`).  This will make the changes to nixio code easier to see and audit.
- Remove axTLS source code

At this point we should not have any changes to the OpenWrt package, or to anyone building the library with OpenSSL, so I would definitely merge it (new PR, maybe?).  Then:

- Move from cyaSSL to wolfSSL
- Enhancements to the nixio code (TLS version changes)
- Reenable the TLS options in OpenWrt.

libs/luci-lib-nixio/src/tls-socket.c
```diff
@@ -202,20 +101,17 @@ static int nixio_tls_sock_send(lua_State *L) {
...
 	const int stat = SSL_connect(sock);
-	nixio_tls__set_connected(L, stat == 1);
 	return nixio__tls_sock_pstatus(L, sock, stat);
 }
 
 static int nixio_tls_sock_shutdown(lua_State *L) {
 	SSL *sock = nixio__checktlssock(L);
-	nixio_tls__set_connected(L, 0);
 	return nixio__tls_sock_pstatus(L, sock, SSL_shutdown(sock));
 }
```
cotequeiroz, line 117: There's another `#ifdef WITH_AXTLS` block in lines 123-125 (orig: 227-229) below that needs to be taken care of.

wevsty:
@cotequeiroz  @Neustradamus
All the questions and suggestions I have raised so far have been modified as much as possible.
I split the PR into 3 commits and now it looks a bit easier.

I have done basic compilation tests and found no problems so far.
I need more time for the complete test, I will update this message when I finish.

Please contact me if you still have questions.

libs/luci-lib-nixio/Makefile
```diff
@@ -10,6 +10,38 @@ LUCI_TITLE:=NIXIO POSIX library
 LUCI_DEPENDS:=+liblua
 
 PKG_LICENSE:=Apache-2.0
+PKG_CONFIG_DEPENDS:= \
+	CONFIG_LUCI_NIXIO_NOTLS \
+	CONFIG_LUCI_NIXIO_WOLFSSL \
+	CONFIG_LUCI_NIXIO_OPENSSL
+
```
cotequeiroz, line 17:
```suggestion

```
Extra whitespace here.

libs/luci-lib-nixio/Makefile
```diff
@@ -10,6 +10,38 @@ LUCI_TITLE:=NIXIO POSIX library
...
+		config LUCI_NIXIO_OPENSSL
+			bool "OpenSSL"
+	endchoice
+endef
+
+export NIXIO_TLS :=
+
+ifneq ($(CONFIG_LUCI_NIXIO_WOLFSSL),)
+	NIXIO_TLS:=wolfssl
+	LUCI_DEPENDS += +libwolfssl
```
cotequeiroz, line 38:
```suggestion
```
The package dependencies are written to `tmp/.packagedeps` before config is read, so this won't work.  Use  conditional dependencies above.

libs/luci-lib-nixio/Makefile
```diff
@@ -10,6 +10,38 @@ LUCI_TITLE:=NIXIO POSIX library
...
+endif
+
+ifneq ($(CONFIG_LUCI_NIXIO_OPENSSL),)
+	NIXIO_TLS:=openssl
+	LUCI_DEPENDS += +libopenssl
```
cotequeiroz, line 43:
```suggestion
```

libs/luci-lib-nixio/Makefile
```diff
@@ -10,6 +10,38 @@ LUCI_TITLE:=NIXIO POSIX library
 LUCI_DEPENDS:=+liblua
```
cotequeiroz, line 10:
```suggestion
LUCI_DEPENDS:=+liblua +LUCI_NIXIO_OPENSSL:libopenssl +LUCI_NIXIO_WOLFSSL:libwolfssl
```

cotequeiroz (changes requested):
The conditional dependencies in the package Makefile need to follow the "OpenWrt syntax" to be processed correctly.  `scripts/package-metadata.pl` does not read `.config` to write the dependencies, so you can't have `make` add them based on the `.config` contents.

libs/luci-lib-nixio/src/tls-context.c
```diff
@@ -65,9 +65,8 @@ static int nixio_tls_ctx(lua_State * L) {
 		return luaL_error(L, "unable to create TLS context");
 	}
 
-#ifdef WITH_CYASSL
+	/* Setting do not verify certificate */
 	SSL_CTX_set_verify(*ctx, SSL_VERIFY_NONE, NULL);
```
cotequeiroz, line 69: You should set the minimum version here, like you had done before.

wevsty: I'm not sure if I should drop support for protocols lower than TLS 1.2 outright, but I've added it back now.

wevsty:
> The conditional dependencies in the package Makefile need to follow the "OpenWrt syntax" to be processed correctly. `scripts/package-metadata.pl` does not read `.config` to write the dependencies, so you can't have `make` add them based on the `.config` contents.

I have made changes to these sections.

cotequeiroz:
@jow- 

I've reviewed this, and I believe it is good enough to be accepted.

- Removal of axTLS from tree is long overdue.  Removal of its support from code was done cleanly, only removing code that was either guarded by `WITH_AXTLS` or calls to `nixio_tls__check_connected` and `nixio_tls__set_connected` macros that were empty without `WITH_AXTLS`.
- The update to wolfSSL was done by using 100% the same code as OpenSSL.  The only difference when building them is the location of the headers.  The difference in certificate validation behavior between libraries had already been taken care of before this change.  It is off by default (OpenSSL behavior), and can be changed with `TLSContext.set_verify`.  So if the OpenSSL code is OK, then I don't see why wolfSSL would not be.
- When built with OpenSSL, the one visible change brought by this is the update from `TLSv1_*_method`  to `TLS_*_method`, along with the call to `SSL_CTX_set_min_proto_version()`. This changes supported TLS versions from 1.0-only to 1.2 and above.

What do you think?

systemcrash:
No comments. But this looks like a good PR if TLS is required in the future here.

Issue #4370, opened 2020-08-19
Insufficient permissions to read UCI configuration.


eliaspizarro:
Steps:
1
![image](https://user-images.githubusercontent.com/8431845/90614233-66acaf00-e1d8-11ea-8e8a-3e2e417ea5c7.png)

2
![image](https://user-images.githubusercontent.com/8431845/90614163-4ed52b00-e1d8-11ea-9b7e-7bce7176d0ae.png)

from fresh install 
![image](https://user-images.githubusercontent.com/8431845/90614345-8cd24f00-e1d8-11ea-9771-5d1c472602e6.png)

noyidoit:
Is there any tutor or docs explained how to write `rpcd/acl.d/xxx.json` ?
My only workaround is revert the `map.html` to a old version, that's so uncivilized.

n3gwg:
Has there been any movement on squashing this bug at all?
What ostensible workarounds have been tested thus far?

I read in a forum a patch was submitted but it was rejected and no explanation given as to why.

stangri:
@n3gwg you may want to test @dibdot's fix: https://github.com/dibdot/luci/commit/d915425c01651f13ab7ba74a72ccbd229433a87b

PR #4438, opened 2020-09-16
luci-compat: fix wrong error msg for "non-uci" maps

dibdot:
fixes issue #4370

### DIFF
--- a/modules/luci-compat/luasrc/view/cbi/map.htm
+++ b/modules/luci-compat/luasrc/view/cbi/map.htm
@@ -2,7 +2,7 @@
 	<div class="alert-message warning"><%=pcdata(msg)%></div>
 <%- end end -%>
 
-<% if readable then %>
+<% if firstmap and readable then %>
 <div class="cbi-map" id="cbi-<%=self.config%>"<%=ifattr(not writable, "style", "opacity:.6; pointer-events:none")%>>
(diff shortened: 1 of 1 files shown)


jow-:
@dibdot - thanks for looking into it. I believe your fix is incomplete and wouldn't work in all cases, e.g. when the first returned map happens to be a SimpleForm instance.

Can you try the following change instead?

```diff
diff --git a/modules/luci-base/luasrc/dispatcher.lua b/modules/luci-base/luasrc/dispatcher.lua
index 44c17c85f..eca305367 100644
--- a/modules/luci-base/luasrc/dispatcher.lua
+++ b/modules/luci-base/luasrc/dispatcher.lua
@@ -1443,8 +1443,8 @@ function _cbi(self, ...)
        end
 
        for i, res in ipairs(maps) do
-               local is_readable_map = has_uci_access(res.config, "read")
-               local is_writable_map = has_uci_access(res.config, "write")
+               local is_readable_map = util.instanceof(res, cbi.SimpleForm) or has_uci_access(res.config, "read")
+               local is_writable_map = util.instanceof(res, cbi.SimpleForm) or has_uci_access(res.config, "write")
 
                writable = writable or is_writable_map
 
```

dibdot:
@jow: I've reverted my change and applied yours ... does not work.

image

PR #4446, opened 2020-09-19
luci-base: make translatable default values for dhcp hostname

namedun:
We must be able to translate these values.

Fixes: fd75c2b7c ("luci-base: add default value options dhcp hostname")


feckert:
@namedun Sorry my fault. I did it too fast and did not think about the translation

PR #4471, opened 2020-09-25
luci-app-ddns: rework with new ddns changes

Ansuel:
This commit rework the app with the new ddns changes. DDns services are now stored in a dedicated list and the service specific data is stored in a dedicated json. This json can both preinstalled with a companion package or be downloaded on demand. The new app now check if the script is present and give a button to install it if not present in the system.
The app now reset the ddns rule settings on service change.
Also rework the app to drop any global function and rework the function to use more default way to get data.

@feckert this should be the changes needed. Can you check and give some suggestion about variable naming (it's a mess) and if i miss something of the new feature?

(in short, I added
- the global ca_cert and url option
- in status a way to get last modified services list
- button to update the list by calling your script
1. A button with switch service will show up when you change service name or ip addres
2. When the button is click the app check if the json is present
3. If it's not present a button with install service show up (and call your script if pressed)
4. If the downloaded script doesn't support ipv6 and ipv6 is selected, a message is shown and nothing can be set
)

If i miss something tell me. 
Changes to do would be drop the services cacert and use directly the global option and test the new app.
Also this lacks of the custom_list file that we still need to introduce in the makefile in some way. (adding this feature will be straightforward since it will be the same function used the standard list)

This depends on P/R openwrt/packages#13509

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -1,4 +1,5 @@
 'use strict';
+'require ui';
 'require view';
 'require dom';
 'require poll';
@@ -8,43 +9,40 @@
 'require form';
 'require tools.widgets as widgets';
 
-var callGetLogServices, callInitAction, callDDnsGetStatus;
-
-var NextUpdateStrings = {};
-
-NextUpdateStrings = {
-	'Verify' : _("Verify"),
-	'Run once' : _("Run once"),
-	'Disabled' : _("Disabled"),
-	'Stopped' : _("Stopped")
-}
-
-var time_res = {};
-time_res['seconds'] = 1;
-time_res['minutes'] = 60;
-time_res['hours'] = 3600;
-
-callGetLogServices = rpc.declare({
-	object: 'luci.ddns',
-	method: 'get_services_log',
-	params: [ 'service_name' ],
-	expect: {  },
-});
-
-callInitAction = rpc.declare({
-	object: 'luci',
-	method: 'setInitAction',
-	params: [ 'name', 'action' ],
-	expect: { result: false }
-});
-
-callDDnsGetStatus = rpc.declare({
-	object: 'luci.ddns',
-	method: 'get_ddns_state',
-	expect: {  }
-});
-
 return view.extend({
+
+	NextUpdateStrings : {
+		'Verify' : _("Verify"),
+		'Run once' : _("Run once"),
+		'Disabled' : _("Disabled"),
+		'Stopped' : _("Stopped")
+	},
+
+	time_res : {
+		seconds : 1,
+		minutes : 60,
+		hours : 3600,
+	},
+
+	callGetLogServices: rpc.declare({
+		object: 'luci.ddns',
+		method: 'get_services_log',
+		params: [ 'service_name' ],
+		expect: {  },
+	}),
+
+	callInitAction: rpc.declare({
+		object: 'luci',
+		method: 'setInitAction',
+		params: [ 'name', 'action' ],
+		expect: { result: false }
+	}),
+
+	callDDnsGetStatus: rpc.declare({
+		object: 'luci.ddns',
+		method: 'get_ddns_state',
+		expect: {  }
+	}),
 
 	callDDnsGetEnv: rpc.declare({
 		object: 'luci.ddns',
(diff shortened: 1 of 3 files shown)


feckert:
> This commit rework the app with the new ddns changes. DDns services are now stored in a dedicated list and the service specific data is stored in a dedicated json. This json can both preinstalled with a companion package or be downloaded on demand. The new app now check if the script is present and give a button to install it if not present in the system.

I would suggest that we check in the new ddns shell script if we have installed the ddns-scripts_service package with opkg and if so we show a message that we could not install services on demand. If you want to install new services you have to remove with opkg the ddns-scripts_service first. (By the way as [mentioned](https://github.com/openwrt/packages/pull/13402#issuecomment-699025971) by @jow- we should rename the package to ddns-scripts-services)

>     * the global ca_cert and url option

Yes this are the new option I have added to the globals section to download the service list with the new ddns script.

>     * in status a way to get last modified services list

I think this is not needed.

>     * button to update the list by calling your script

Only available if we have not installed the ddns-scripts_service with opkg.

>     1. A button with switch service will show up when you change service name or ip addres

I think we should change the /etc/init.d/ddns to procd, so the services gets restarted if the config has changed. So we do not need this handling. Or do I get something wrong.

>     2. When the button is click the app check if the json is present

Can you please describe exactly what you mean. If this is a about installing custom services, then this button should only visible if we have not installed the ddns-scripts_service package. If the package is not installed then we could download the list file from github.
 
>     3. If it's not present a button with install service show up (and call your script if pressed)

Can you please describe exactly what you mean.


>     4. If the downloaded script doesn't support ipv6 and ipv6 is selected, a message is shown and nothing can be set

As I can see in the current LuCI source of luci-app-ddns we have one drop-down for ipv4 and one ipv6 services. So we should parse the json files on load and add the service to the individual service list for [ipv4](https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L139) and [ipv6](https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L145)

> If i miss something tell me.
> Changes to do would be drop the services cacert and use directly the global option and test the new app.

I am not sure this is applicable for the whole ddns service the ca_cert and url option are only for the custom ddns service load.
The individual uci service section need there own ca_cert and url options I think.

> Also this lacks of the custom_list file that we still need to introduce in the makefile in some way. (adding this feature will be straightforward since it will be the same function used the standard list)

If we check that the ipkg package ddns-scripts_service is installed we do not need the custom_list.

Now the only question is what to do with the special services that come with a script. Either we delete the packages and move the files into the ddns-sciprts_service package, so that we only have ddns-scripts and ddns-script-service or we change the list file in a preinst hook to add them to this list, so they could be selected.

Ansuel:
> I would suggest that we check in the new ddns shell script if we have installed the ddns-scripts_service package with opkg and if so we show a message that we could not install services on demand. If you want to install new services you have to remove with opkg the ddns-scripts_service first. (By the way as [mentioned](https://github.com/openwrt/packages/pull/13402#issuecomment-699025971) by @jow- we should rename the package to ddns-scripts-services)

Mh yes I should add some check about the "all service" script. But do we really need it? If all it does is preinstall all the script for the luci-app it won't change anything since it will notice that the script is already available.

> I think this is not needed.

Just a debug info...  ok will remove (was usefull to check how old was the script)

> Only available if we have not installed the ddns-scripts_service with opkg.

I can add this check but again is it a problem if the list gets updated with the service package installed?

> I think we should change the /etc/init.d/ddns to procd, so the services gets restarted if the config has changed. So we do not need this handling. Or do I get something wrong.

Was talking to something else... think a gif is better to answer also the other question.
captured
(the dropdown list is not showed...)

> As I can see in the current LuCI source of luci-app-ddns we have one drop-down for ipv4 and one ipv6 services. So we should parse the json files on load and add the service to the individual service list for [ipv4](https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L139) and [ipv6](https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L145)

I removed all the complexity (by adding others ....) and the app now have one list and check if the service supports ipv6.

> I am not sure this is applicable for the whole ddns service the ca_cert and url option are only for the custom ddns service load.
> The individual uci service section need there own ca_cert and url options I think.

I think it's redundant to have one global option for cacerts and every service section has their own. But yes can be useful if someone for some reason wants to skip cert validation for a specific service

> If we check that the ipkg package ddns-scripts_service is installed we do not need the custom_list.

Think the custom_list is a must. Since we have scripts that can replace the normal list, a simple post-inst script is not enough since the changes can be erased by the ddns script. 

> Now the only question is what to do with the special services that come with a script. Either we delete the packages and move the files into the ddns-sciprts_service package, so that we only have ddns-scripts and ddns-script-service or we change the list file in a preinst hook to add them to this list, so they could be selected.

I still think that they should be separated (since they can't be downloaded on demand since an additional script is needed). This is why we need a custom_list to handle this special case. (The special pkg will execute a postinst on the custom_list that will not be touched by any script)

feckert:
@Ansuel There were already some comments and complaints on the mailing list!

-  https://lists.openwrt.org/pipermail/openwrt-devel/2020-September/031507.html
- https://github.com/openwrt/packages/issues/13495

We should fix the ddns-scripts package for the next stable release.
I would suggest that I rename the package ddns-scripts_service to ddns-scripts-services as suggested by jow- .
For now we should **not** concentrate on loading providers on demand.
We should concentrate to make the luci-app-ddns working again so that if the ddns-scripts-service package is installed then the services could selected again.

The way I see it many people use the master branch!!
And it is broken now.
Next time I think I should not merge anything before we have updated the LuCI as well.
But that was my mistake!

Ansuel:
@feckert give me a sec

Ansuel:
@feckert Was thinking can we keep the list file? I know it will be redundant, since we will install all the services but having a list will be very useful when we will enable the "service on demand" feature in the future.

Ansuel:
I changed the app to not depend on the presence of the service list.
(the app will enumerate the services based on the json present in the services directory and then optionally on the services in the list, if present).
The button to download the list is still present but will complain about not find the list. (clicking on it will simply download the list and add the missing services if any)
I think for the future we will need to add a feature to autodownload the json if is not present and a config exist for it.

feckert:
This depends on P/R https://github.com/openwrt/packages/pull/13509

Ansuel:
@feckert I updated the app with the new directory scheme. Should I squash the 3 commits to one?

feckert:
Looks pretty good already.
But if we create a new service and select ipv4 or ipv6, can we not offer only those drop-down services that support it.
Now I can choose IPv6 and then select a service that does not support ipv6?

Ansuel:
Problem is that since we merged ipv4 and ipv6 service in one json file, for optimization reason and to not scan 50+ json file by the webui, the service support for both ipv4 and ipv6 is evaluated only when is selected and saved. I can try to find a workaround for this, but I find the use of the 'switch service' button cleaner and easier to implement.
Anyway as I showed in the gif, selecting a ipv6 + a service not supported, results in an alert and the hide of all the required option to configure the ddns rule.

(to sum up. the service list is generated only on the json it does find, it doesn't actually opens it and check the contents)

feckert:
> Problem is that since we merged ipv4 and ipv6 service in one json file, for optimization reason and to not scan 50+ json file by the webui, the service support for both ipv4 and ipv6 is evaluated only when is selected and saved. I can try to find a workaround for this, but I find the use of the 'switch service' button cleaner and easier to implement.

Have you ever looked to see if this really has such a big impact on the performance?

> Anyway as I showed in the gif, selecting a ipv6 + a service not supported, results in an alert and the hide of all the required option to configure the ddns rule.

I use the new openwrt-theme-2020 and I have to look twice to see the hint.

The other option would be if this could not be validated at runtime, we could check it during creation and then issue an error message that this service does not support this proto. The advantage would be that the service is not created.

Ansuel:
> Have you ever looked to see if this really has such a big impact on the performance?

Nope but I see it as a waste.

> I use the new openwrt-theme-2020 and I have to look twice to see the hint.

Strange. What should happen is, select a new service, click switch service, then the ui tells you if the service is not supported or show more option.

> The other option would be if this could not be validated at runtime, we could check it during creation and then issue an error message that this service does not support this proto. The advantage would be that the service is not created.

Yes for the creation I should be able to add this check and stop the save of the rule.

Ansuel:
@feckert ok... done some magic... I added some type of promise validation...
Anyway when the user change the ip or the service, the service is set invalid. When the service check is done a new validation is fired with the new data.
The result is that if the service doesn't support ipv6, the ddns rule can't be created. (If the service is not installed, the service is valid by default). The same logic applies for ddns rule already created, if someone try to change the service and ipv6 is not supported, the service button is disabled.

feckert:
Very nice :+1:
But I found out that if no default service is installed and I press the download button the services are not displayed. Only when I clear my browser cache. @Ansuel could you also seed this on your setup?

Ansuel:
@feckert think I undestrand what you mean... And I already notice that.
If no default service is installed and no list is downloaded, pressing the 'update services list' doesn't change anything and a reload is needed to show the new services. Is this what you mean?
Anyway I changes how the app gets services list and now it's updated when you press the 'update services list' and when the service is installed...
Tell me if I got this wrong.

feckert:
@Ansuel I have merged the depended pullrequest.
@jow- I am not sure what the policy is. Should with stash the commit to one? Because it is a big change or leave it as it is.

Ansuel:
Personally I'm from the stash since the other commits are here just to show progress since they can't be applied individually...

feckert:
@Ansuel could you please fixup your commits?
I have already merged the ddns changes.
To get back to a working state for LuCI also we should merge your changes as fast as possible.

Ansuel:
What changes should I do?

feckert:
> What changes should I do?

I don't want to do anything wrong when I click the webbutton to `squash and merge`.
I have not done that yet.
Which commit message does he use then if all commits are squashed into one if I use the button?
So I thought you could squash them.

Ansuel:
If i remember correctly, It will tell what commits message to set. Give me a sec and I squash them and force push.

Ansuel:
@feckert done

PR #4485, opened 2020-10-01
luci-app-firewall: rules: add ICMPv6 Packet Too Big (Type 2)

Robby-:
The "Match ICMP type" drop-down menu was missing this ICMPv6 type. According to RFC 4890 section 4.3.1 it is essential for communications and must not be dropped. This patch allows for doing this through LuCI.


jow-:
Merged, thanks!

PR #4490, opened 2020-10-05
luci-app-fwknopd: Client-side rendering and improvements/fixes.

oldium:
* Use client-side rendering.
 * Asynchronous QR code generation and on-the-fly updates.
 * Added button to generate keys.
 * Added keys validation (empty, \_\_CHANGEME__ and CHANGEME values are not valid).
 * Added SOURCE option non-empty validation (SOURCE is mandatory field).
 * Added network and DESTINATION options.
 * Use checkboxes for Y/N options instead of text fields.
 * Fix the qrencode tool, which currently has only `--inline` option, not `-I`. Use `--inline` instead.
 * Improve configuration handling by removing the key type from the config file. The key type is deduced from the actual values (normal/base64).
 * Provide migration script (clean-up of stale keytype/hkeytype values in config).
 * Fixed usage of translations.
 * Added /etc/fwknop/access.conf parsing to show QR codes.

Remove setting values in `uci-defaults`, which can be changed from LuCI directly:

 * Keys can be generated from LuCI, so it is not necessary to set them.
 * ENABLE_IPT_FORWARDING and ENABLE_NAT_DNS options can be set in LuCI.


oldium:
I will update translations later today.

oldium:
Ready for review.

oldium:
Changed title, because I have the client-side rendering version almost ready (not yet committed). Client-side form works, QR code generating works. I want to additionally update QR code on-the-fly when keys change and add a button to generate the keys – before I commit it.

oldium:
Finished, please review.

oldium:
One last update - simplification of key options names logic.

oldium:
Further ideas:

1. Remove key generating from `uci-defaults` script, because the keys can be generated from LuCI by clicling the Generate Keys button. Also setting sensible defaults like `ENABLE_IPT_FORWARDING=y` and `ENABLE_NAT_DNS=y` should be done in fwknopd OpenWrt package, not in the LuCI package.
2. Offer parsing of custom /etc/fwknop/access.conf (when `uci_enabled=0`) to show the QR code(s) for custom configuration.

I think that at least idea 1. should be implemented. I will prepare CR for the OpenWrt part.

oldium:
Implemented idea 1 - removed uci-defaults manipulation (kept migration clean-up only) and enhanced fwknop package in openwrt/packages#13678

oldium:
Implemented idea 2 - show QR codes for custom configuration (`/etc/fwknop/access.conf`).

Now there is time to fix some UI glitches (in separate PRs):

 * `onclick` option handler is called before `map.checkDepends()`, so `isActive()` of shown options does not reflect up-to-date state (#4516).
 * Material theme not playing well with read-only/disabled forms – even and odd lines are not colored correctly. The best would be to have support in form style and explicitly mark even/odd lines with special CSS class.
 * Bootstrap theme does not allow tooltips (show validation errors) to be shown for read-only/disabled forms, it explicitly disables pointer-events CSS on those read-only/disabled fields.

oldium:
Small bugfix and rebased to latest master.

hnyman:
cc @jp-bennett

jp-bennett:
It looks like @oldium understands Luci better than I do. I hope to do some testing of this tonight, but so far it looks good to me.

oldium:
I just discovered access.conf parsing logic error, I will fix it asap.

oldium:
Done. Enjoy 😊

oldium:
Removed `map.checkDepends()` workaround (due to #4531), rebased to master.

hnyman:
I will merge this so that you will get feedback (hopefully)

Issue #3681, opened 2020-02-26
feature: one click to create a static lease


bjlockie:
Allow creating a static lease by one click on an active lease.

Ansuel:
@feckert this can be closed

PR #4499, opened 2020-10-06
luci-mod-status: add support for one-click static lease

Ansuel:
Introduce a button in the status lease table to quickly add an host as static.

This would address #3681 . I notice it would be not that hard to implement and I'm here to propose a version of it. (the quickest)
Another version would be show a modal with the text autocompleted.
For ipv4 this creates a host rule named with the mac address (without the : ) and with ip mac and hostname (or '-') or the lease.
For ipv6 the rule is named with the mac address and with only the duid and the mac address.

@jow- sorry for the tag but this would remove an issue from UTOPIA :D

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
@@ -1,7 +1,9 @@
 'use strict';
 'require baseclass';
 'require rpc';
+'require uci';
 'require network';
+'require validation';
 
 var callLuciDHCPLeases = rpc.declare({
 	object: 'luci-rpc',
@@ -12,29 +14,85 @@ var callLuciDHCPLeases = rpc.declare({
 return baseclass.extend({
 	title: '',
 
+	isMACStatic: {},
+	isDUIDStatic: {},
+
 	load: function() {
 		return Promise.all([
 			callLuciDHCPLeases(),
-			network.getHostHints()
+			network.getHostHints(),
+			uci.load('dhcp')
 		]);
+	},
+
+	handleCreateStaticLease: function(lease, ev) {
+		ev.currentTarget.classList.add('spinning');
+		ev.currentTarget.disabled = true;
+		ev.currentTarget.blur();
+
+		var cfg = uci.add('dhcp', 'host');
+		uci.set('dhcp', cfg, 'name', lease.hostname || '-');
+		uci.set('dhcp', cfg, 'ip', lease.ipaddr);
+		uci.set('dhcp', cfg, 'mac', lease.macaddr.toUpperCase());
+
+		return uci.save()
+			.then(L.bind(L.ui.changes.init, L.ui.changes))
+			.then(L.bind(L.ui.changes.displayChanges, L.ui.changes));
+	},
+
+	handleCreateStaticLease6: function(lease, ev) {
+		ev.currentTarget.classList.add('spinning');
+		ev.currentTarget.disabled = true;
+		ev.currentTarget.blur();
+
+		var cfg = uci.add('dhcp', 'host'),
+		    ip6arr = lease.ip6addrs[0] ? validation.parseIPv6(lease.ip6addrs[0]) : null;
+
+		uci.set('dhcp', cfg, 'name', lease.hostname || '-');
+		uci.set('dhcp', cfg, 'duid', lease.duid.toUpperCase());
+		uci.set('dhcp', cfg, 'mac', lease.macaddr);
+		if (ip6arr)
+			uci.set('dhcp', cfg, 'hostid', (ip6arr[6] * 0xFFFF + ip6arr[7]).toString(16));
+
+		return uci.save()
+			.then(L.bind(L.ui.changes.init, L.ui.changes))
+			.then(L.bind(L.ui.changes.displayChanges, L.ui.changes));
 	},
 
 	renderLeases: function(data) {
 		var leases = Array.isArray(data[0].dhcp_leases) ? data[0].dhcp_leases : [],
 		    leases6 = Array.isArray(data[0].dhcp6_leases) ? data[0].dhcp6_leases : [],
-		    machints = data[1].getMACHints(false);
+		    machints = data[1].getMACHints(false),
+		    hosts = uci.sections('dhcp', 'host');
+
+		for (var i = 0; i < hosts.length; i++) {
+			var host = hosts[i];
+
+			if (host.mac) {
+				var macs = L.toArray(host.mac);
+				for (var j = 0; j < macs.length; j++) {
+					var mac = macs[j].toUpperCase();
+					this.isMACStatic[mac] = true;
+				}
+			}
+			if (host.duid) {
+				var duid = host.duid.toUpperCase();
+				this.isDUIDStatic[duid] = true;
+			}
+		};
 
 		var table = E('div', { 'class': 'table lases' }, [
 			E('div', { 'class': 'tr table-titles' }, [
 				E('div', { 'class': 'th' }, _('Hostname')),
 				E('div', { 'class': 'th' }, _('IPv4-Address')),
 				E('div', { 'class': 'th' }, _('MAC-Address')),
-				E('div', { 'class': 'th' }, _('Lease time remaining'))
+				E('div', { 'class': 'th' }, _('Lease time remaining')),
+				E('div', { 'class': 'th cbi-section-actions' }, _('Static Lease'))
 			])
 		]);
 
-		cbi_update_table(table, leases.map(function(lease) {
-			var exp;
+		cbi_update_table(table, leases.map(L.bind(function(lease) {
+			var exp, rows;
 
 			if (lease.expires === false)
 				exp = E('em', _('unlimited'));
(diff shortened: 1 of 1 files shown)


jow-:
Great work @Ansuel - unfortunately it fails to render for me with 

```
TypeError: lease.macaddr is undefined
    renderLeases https://jj/luci-static/resources/view/status/include/40_dhcp.js?v=git-19.292.31773-cc35220:107
    renderLeases https://jj/luci-static/resources/view/status/include/40_dhcp.js?v=git-19.292.31773-cc35220:106
    render https://jj/luci-static/resources/view/status/include/40_dhcp.js?v=git-19.292.31773-cc35220:154
    step https://jj/luci-static/resources/view/status/index.js?v=git-19.292.31773-cc35220:39
```

Ansuel:
Will check this in a sec... Anyway I have some question... With the current implementation the click of the button force the save of all the pending settings... I think this is wrong but I can't really find another way of doing this. Is it wrong or we can leave it as is?

jow-:
Hmm, yes, I think committing all changes without reconfirmation is not ideal. You should save the config and directly open the pending changes dialog. That might be a good interim solution.

jow-:
So I think the logic should work like this:

- initialize empty object `isDUIDStatic`
- initialize empty object `isMACStatic`
- foreach `dhcp.host` section
  - foreach `mac` value (`L.toArray(section.mac)`)
    - set `isMACStatic[mac]` to `true`
  - set `isDUIDStatic[section.duid]` to `true`

Then, while rendering the DHCPv4 table:

- test `isMACStatic[lease.macaddr]`
  - true => display "Make static" button
  - false => display "Forget host" button (which deletes the lease) - alternatively simply render the "Make static" button as disabled

While rendering the DHCPv6 table:

- test `isDUIDStatic[lease.duid]`
  - true => display "Make static" button
  - false => display "Forget host" button (which deletes the lease) - alternatively simply render the "Make static" button as disabled

Also keep in mind that:
 - MACs in the uci config may appear in upper- or lowercase, so you need to normalize them
 - DUIDs in the uci config may appear in upper- or lowercase
 - a single host section may cover both a DUID and one or multiple MACs, so in case you implement a "forget" logic, you must not remove the entire section if it covers more than only DHCPv4 or only DHCPv6

Ansuel:
I think the forget option would add too much complexity to a simple status script. If someone wants to remove the lease he can easily go to dhcp and removes it.
I address all the request... I don't know if it's correct the way I used to get the dhcp config...

Only problem... It looks like uci.save() doesn't respect promise queue.

Using
```
return uci.save().then(function () {
			return L.ui.changes.displayChanges();
		});
```

Make the changes modal appear before the changes are saved.

Ansuel:
I fixed this by calling changes.init before displayChanges...
@jow- think it's ready

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -12,29 +13,82 @@ var callLuciDHCPLeases = rpc.declare({
...
+	},
+
 	renderLeases: function(data) {
 		var leases = Array.isArray(data[0].dhcp_leases) ? data[0].dhcp_leases : [],
 		    leases6 = Array.isArray(data[0].dhcp6_leases) ? data[0].dhcp6_leases : [],
-		    machints = data[1].getMACHints(false);
+		    machints = data[1].getMACHints(false),
+		    dhcp_conf = this.uci.state.values.dhcp;
+
+		for (var element in dhcp_conf) {
```
jow-, line 63: You're touching uci class internals here, please use the public [`uci.sections()`](https://openwrt.github.io/luci/jsapi/LuCI.uci.html#sections) api instead:

```javascript
var leases = ...,
    ...,
    hosts = uci.sections('dhcp', 'host');

for (var i = 0; i < hosts.length; i++) {
    var host = hosts[i];
    if (host.mac) {
         ...
    }
}
```

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -12,29 +13,82 @@ var callLuciDHCPLeases = rpc.declare({
...
+
+			if (dhcp_conf[element]['.type'] == 'host') {
+				if (dhcp_conf[element].mac) {
+					var macs = dhcp_conf[element].mac.split(',')
```
jow-, line 67: As I wrote earlier, please use `L.toArray(host.mac)` since the value may be either a whitespace separated string or an array in case uci list notation was used. Using `split()` is not correct in either case.

Also please add a semicolon to the end of the statement.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -12,29 +13,82 @@ var callLuciDHCPLeases = rpc.declare({
...
+					for (var mac in macs) {
+						var mac = macs[mac].replace(/\s/g, "").toUpperCase()
```
jow-, line 69: That whitespace trimming is redundant when you use `L.toArray()`. Also please add a semicolon to the end of the statement.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -12,29 +13,82 @@ var callLuciDHCPLeases = rpc.declare({
...
+						this.isMACStatic[mac] = true;
+					}
+				}
+				if (dhcp_conf[element].duid) {
+					var duid = dhcp_conf[element].duid.toUpperCase()
```
jow-, line 74: Please add a semicolon here.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -102,4 +179,4 @@ return baseclass.extend({
 
 		return E([]);
 	}
-});
+});
```
jow-, line 182: Please ensure that the file ends with a newline.

jow- (changes requested):
That looks generally good :) Some more remarks below.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
Ansuel: Wasted 20 minutes find that api and it was that easy...

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -43,25 +96,37 @@ return baseclass.extend({
...
 				lease.macaddr,
 				exp
 			];
-		}), E('em', _('There are no active leases')));
+
+			if (lease.macaddr != null) {
+				var mac = lease.macaddr.toUpperCase();
+				rows.push(E('button', {
+					'class': 'cbi-button cbi-button-apply',
+					'click': L.bind(this.handleCreateStaticLease, this, lease),
```
jow-, line 110: Please use `ui.createHandlerFn(this, 'handleCreateStaticLease', lease)` here (see [`ui.createHandlerFn()`](https://openwrt.github.io/luci/jsapi/LuCI.ui.html#createHandlerFn))

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -80,13 +145,24 @@ return baseclass.extend({
...
 				lease.duid,
 				exp
 			];
-		}), E('em', _('There are no active leases')));
+
+			if (lease.duid != null) {
+				var duid = lease.duid.toUpperCase();
+				rows.push(E('button', {
+					'class': 'cbi-button cbi-button-apply',
+					'click': L.bind(this.handleCreateStaticLease6, this, lease),
```
jow-, line 159: Please use `ui.createHandlerFn(this, 'handleCreateStaticLease6', lease)` here

jow- (changes requested):

jow-:
Found two more little improvements, apart from that it looks very good!

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -12,29 +13,81 @@ var callLuciDHCPLeases = rpc.declare({
...
-		    machints = data[1].getMACHints(false);
+		    machints = data[1].getMACHints(false),
+		    hosts = uci.sections('dhcp', 'host');
+
+		for (var i = 0; i < hosts.length; i++) {
+			var host = hosts[i];
+
+			if (host.mac) {
+				var macs = L.toArray(host.mac);
+				for (var i = 0; i < macs.length; i++) {
```
jow-, line 68: One more thing... you need to change `i` to `j` or something else here to avoid an infinite loop since the outer `for(...)` already uses `i`.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
Ansuel: I use `bind` instead of create `createHandler` since with createHandler the 'set Static' button is enabled again after the function is completed. Looks bad that you click the Button. Then it became enabled and when the pool function refresh the data the button became disabled.

jow-: I see, it is fine then.

jow-:
Unfortunately there's one remaining thing I noticed while testing this... in DHCPv6 leases, when we add a lease through the button click, we add a `config host` entry with a `duid`, `hostname` and `mac` option - but to actually achieve a static IPv6 lease we would also need to set a `hostid` which is the last 32 bits of the IPv6 address as hex string.

You should be able to derive the hostid from the lease information like this:

```javascript
'require validation';
...
handleCreateStaticLease6: function(lease, ev) {
    ...
    var ip6arr = lease.ip6addrs[0] ? validation.parseIPv6(lease.ip6addrs[0]) : null;
    if (ip6arr)
        uci.set('dhcp', cfg, 'hostid', (ip6arr[6] * 0xFFFF + ip6arr[7]).toString(16));
    ...
},
...
```

Ansuel:
Should the rule creation fail if the hostid can't be derived for some reason?

jow-:
Good question, but I'd say we can ignore this hypothetical corner case for now since a config change dialog is displayed right after so the user would have the chance to cancel if things look odd.

Ansuel:
Should be ready

jow-:
Merged, thanks!

Issue #1072, opened 2017-03-16
WPS PBC softbutton


coelner:
Hey,
it would we useful if a WPS PBC softbutton could be placed into the wireless overview. it is not useful to push the real button, because it activate WPS for all ssids on all radios. (But that confuses other devices, because the can't differ between those wps active wlan's)

Info: [https://forum.openwrt.org/viewtopic.php?id=56833](url)

PR #4508, opened 2020-10-10
luci-mod-status: add WPS control for wifi info

Ansuel:
Add a way to trigger and get the WPS Push Button status from the webui if supported.

Not in UTOPIA but I think it should. If the system supports this feature any type of check or trigger is absent in the webui. This feature can only be manage with the hardware button. Since the button call the exact same command, we can easily simulate this.

If this gets approved, I would ask some improvement for this... 
Is the button in the badge acceptable? How can I change this to make things better?
Also I would make the check for hostapd_cli as and env variable like we do with dnsmasq...
To clean things a bit, is it necessary to have global function instead of adding them in the view struct?

Fixes: #1072

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
@@ -2,89 +2,43 @@
 'require baseclass';
 'require dom';
 'require network';
+'require uci';
+'require fs';
 'require rpc';
-
-var callSessionAccess = rpc.declare({
-	object: 'session',
-	method: 'access',
-	params: [ 'scope', 'object', 'function' ],
-	expect: { 'access': false }
-});
-
-function renderbox(radio, networks) {
-	var chan = null,
-	    freq = null,
-	    rate = null,
-	    badges = [];
-
-	for (var i = 0; i < networks.length; i++) {
-		var net = networks[i],
-		    is_assoc = (net.getBSSID() != '00:00:00:00:00:00' && net.getChannel() && !net.isDisabled()),
-		    quality = net.getSignalPercent();
-
-		var icon;
-		if (net.isDisabled())
-			icon = L.resource('icons/signal-none.png');
-		else if (quality <= 0)
-			icon = L.resource('icons/signal-0.png');
-		else if (quality < 25)
-			icon = L.resource('icons/signal-0-25.png');
-		else if (quality < 50)
-			icon = L.resource('icons/signal-25-50.png');
-		else if (quality < 75)
-			icon = L.resource('icons/signal-50-75.png');
-		else
-			icon = L.resource('icons/signal-75-100.png');
-
-		var badge = renderBadge(
-			icon,
-			'%s: %d dBm / %s: %d%%'.format(_('Signal'), net.getSignal(), _('Quality'), quality),
-			_('SSID'), net.getActiveSSID() || '?',
-			_('Mode'), net.getActiveMode(),
-			_('BSSID'), is_assoc ? (net.getActiveBSSID() || '-') : null,
-			_('Encryption'), is_assoc ? net.getActiveEncryption() : null,
-			_('Associations'), is_assoc ? (net.assoclist.length || '-') : null,
-			null, is_assoc ? null : E('em', net.isDisabled() ? _('Wireless is disabled') : _('Wireless is not associated')));
-
-		badges.push(badge);
-
-		chan = (chan != null) ? chan : net.getChannel();
-		freq = (freq != null) ? freq : net.getFrequency();
-		rate = (rate != null) ? rate : net.getBitRate();
-	}
-
-	return E('div', { class: 'ifacebox' }, [
-		E('div', { class: 'ifacebox-head center ' + (radio.isUp() ? 'active' : '') },
-			E('strong', radio.getName())),
-		E('div', { class: 'ifacebox-body left' }, [
-			L.itemlist(E('span'), [
-				_('Type'), radio.getI18n().replace(/^Generic | Wireless Controller .+$/g, ''),
-				_('Channel'), chan ? '%d (%.3f %s)'.format(chan, freq, _('GHz')) : '-',
-				_('Bitrate'), rate ? '%d %s'.format(rate, _('Mbit/s')) : '-'
-			]),
-			E('div', {}, badges)
-		])
-	]);
-}
-
-function wifirate(rt) {
-	var s = '%.1f\xa0%s, %d\xa0%s'.format(rt.rate / 1000, _('Mbit/s'), rt.mhz, _('MHz')),
-	    ht = rt.ht, vht = rt.vht,
-		mhz = rt.mhz, nss = rt.nss,
-		mcs = rt.mcs, sgi = rt.short_gi;
-
-	if (ht || vht) {
-		if (vht) s += ', VHT-MCS\xa0%d'.format(mcs);
-		if (nss) s += ', VHT-NSS\xa0%d'.format(nss);
-		if (ht)  s += ', MCS\xa0%s'.format(mcs);
-		if (sgi) s += ', ' + _('Short GI').replace(/ /g, '\xa0');
-	}
-
-	return s;
-}
 
 return baseclass.extend({
 	title: _('Wireless'),
+
+	WPSTranslateTbl: {
+		Disabled: _('Disabled'),
+		Active: _('Active'),
+		'Timed-out': _('Timed-out'),
+		Overlap: _('Overlap'),
+		Unknown: _('Unknown')
+	},
+
+	callSessionAccess: rpc.declare({
+		object: 'session',
+		method: 'access',
+		params: [ 'scope', 'object', 'function' ],
+		expect: { 'access': false }
+	}),
+
+	wifirate: function(rt) {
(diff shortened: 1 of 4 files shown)


Ansuel:
Also I notice that adding acl.d to the wifi index doesn't work...
Exec permission works even without acl.d permission

jow-:
What do you mean by that? Can you provide a specific example?

Ansuel:
When I was adding fs.stat for hostapd_cli for the include/60_wifi.js script, adding the permission to the acl.d section `luci-mod-status-index-wifi`, the fs.stat failed with permission denied. 
Adding it to `luci-mod-status-index` section, fs.stat correctly gets the data. (same thing for `exec`)

Also about the exec permission, as you can see there is no `exec` permission in 
```
"file": [ "list", "read", "stat" ],
```

But still 
```
				"/usr/sbin/hostapd_cli -i wlan[0-9]* wps_pbc": [ "exec" ],
				"/usr/sbin/hostapd_cli -i wlan[0-9]* wps_get_status": [ "exec" ],
				"/usr/sbin/hostapd_cli -i wlan[0-9]* wps_cancel": [ "exec" ]
```
works.

jow-:
Likely because some other acl enables file/exec, e.g. https://github.com/openwrt/luci/blob/master/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.jsonL54

You can review the effective ACL using `ubus call session list` on the cli.

Ansuel:
@jow- thx for the command. Anyway are we interested in this feature or should we drop this?

castillofrancodamian:
> @jow- thx for the command. Anyway are we interested in this feature or should we drop this?

It is an interesting function for a guest Wi-Fi network and to connect easily.

Ansuel:
As I said I have some question about this...

1. Is the button allowed in the wifi badge? 
2. Should I add the hostapd_cli to the SystemFeature instead of calling the stat at every pool?
3. Can I clean all the global function of 60_wifi and make them part of the view Object?
4. Is the acl rule correct? (the exec command accept any wlan[0-9]* but can the wifi interface have different naming in openwrt?)

jow-:
1. Not sure, can you provide a screenshot to visualize what you mean?
2. Yes, I think that would make sense
3. Sure, but the function is not global, it is local to the `60_wifi.js` view class. Why do you want to move it?
4. We probably need to wrap this call in a `/usr/libexec/wps_action` utility script which takes care of enforcing the parameter. The current pattern `"/usr/sbin/hostapd_cli -i wlan[0-9]* wps_pbc"` is too broad since it would allow an arbitrary amount of parameters after `wlan...`.

The utility script could look like this:
```bash
#!/bin/sh

case "$1" in
  pbc|get_status|cancel)
    if [ -e "/sys/class/net/$2" ]; them
      exec /usr/sbin/hostapd_cli -i "$2" "wps_$1"
    else
      echo "No such interface: $2" >&2
      exit 1
    fi
  ;;
  *)
    echo "Usage: $0 {pbc,get_status,cancel} ifname" >&2
    exit 1
  ;;
esac
```
And a corresponding ACL rule:
```
"/usr/libexec/wps_action": [ "exec" ]
```

Ansuel:
1. Screen
Immagine 2020-10-10 223234
Will address the other suggestion thanks.

jow-:
Hmmm, looks a bit out of place. I think it would look better if you omit the label, so that it looks like this:

grafik

Ansuel:
The best thing would be have a small button aligned to the size of the text but i think it's not available in luci css.

dangowrt:
We actually got a ubus interface for this, so instead you could `ubus call hostapd.wlan0 wps_start` or, even better, use `rpcd` for that.

Ansuel:
@dangowrt just as an info... where this was documented? Don't know if it's me but it's hard to find what ubus expose :(

Anyway this means that wps can be supported even without hostapd-utils using ubus... Can you confirm this?

dangowrt:
> @dangowrt just as an info... where this was documented? Don't know if it's me but it's hard to find what ubus expose :(
Use `ubus -v list` to explore this kinda a self-documenting API ;)

> 
> Anyway this means that wps can be supported even without hostapd-utils using ubus... Can you confirm this?
Yes. And it works for wpa_supplicant just the same way.

Ansuel:
Didn't know -v list could be used also with params (`hostapd.wlan0`) nice to know

Well I have to make a patch to add support for get status ahahah 
This needs to wait :(

dangowrt:
> Also no wps_get_status :(

I could add this easily if needed. Is it?

Ansuel:
@dangowrt if you want and can implement this quickly it would be very good...
```
root@No-Lag-Router:~# hostapd_cli -i wlan0 wps_get_status
PBC Status: Timed-out
Last WPS result: None
```

Best thing would be convert this to a json output

Ansuel:
@dangowrt I implemented a test version anyway...

dangowrt:
https://git.openwrt.org/?p=openwrt/staging/dangole.git;a=commit;h=4c3d9f113d86ecfebbd4c8ef0531be45e5ed3a3b

Ansuel:
Thx a lot. Is CONFIG_WPS enabled by default?

dangowrt:
> Thx a lot. Is CONFIG_WPS enabled by default?

Currently we got `CONFIG_WPS` enabled only for the `-full` and `-p2p` variants, but not in the `-basic` and `-mini` variants which are installed by default.

Ansuel:
@dangowrt sorry if I bother you again... Since CONFIG_WPS can be disabled... Can we make a patch to add wps to the `has_feature`
function?  (src/utils/build_features.h)

Ansuel:
@jow- I updated this... this depends on 2 patch
1. https://git.openwrt.org/?p=openwrt/staging/dangole.git;a=commit;h=4c3d9f113d86ecfebbd4c8ef0531be45e5ed3a3b
2. The `CONFIG_WPS` in`hostapd -v` that still needs to be created

Since now wps is supported without hostapd_cli, I fixed the wireless page to show the pushbutton option even without hostapd_cli.

dangowrt:
> @dangowrt sorry if I bother you again... Since CONFIG_WPS can be disabled... Can we make a patch to add wps to the `has_feature`
> function? (src/utils/build_features.h)
> Can't find any other way to check this....

Can't you just check if the ubus methods are present? That'd be the most obvious way to do it imho.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
```diff
@@ -2,90 +2,45 @@
...
-
-	return s;
-}
-
 return baseclass.extend({
 	title: _('Wireless'),
 
+	WPSTranslateTbl: {
+		WPS_PBC_STATUS_DISABLE: _('Disabled'),
+		WPS_PBC_STATUS_ACTIVE: _('Active'),
```
dangowrt, line 14: Probably it'll be easier then I'd just output those shorted strings right away then...

Ansuel:
Currently luci uci the presence of hostapd_cli to check if wps is supported. I think that calling ubus to check if there are wps entries in any wlan interface is not good... (imho check if a command doesn't fail to make sure a feature is supported is wrong...)
Also luci should call hostapd.(random present interface), so every time it should load all the wifi interface, take one and call ubus (and check if the command fail)

IMHO it's much cleaner to add support to the -v command as it's already done to check if hostapd supports 11ac 11r and other feature

dangowrt:
@Ansuel I've updated the status API to make things easier, see my staging tree:
https://git.openwrt.org/?p=openwrt/staging/dangole.git;a=commit;h=b51e22fa2b75d0c05dfbb02d0f120cbccbd65395

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
Ansuel: I actually like your implementation that prints the symbol... That way we can enum all the different strings and can be translated by the webui... (in short I had to do the same exact thing even if ubus reported the text directly)

Ansuel:
Anyway the change would be mainly this in `src/utils/build_features.h`
```
#ifdef CONFIG_WPS
	if (!strcmp(feat, "wps"))
		return 1;
#endif
```

dangowrt:
> Anyway the change would be mainly this in `src/utils/build_features.h`
> 
> ```
> #ifdef CONFIG_WPS
> 	if (!strcmp(feat, "wps"))
> 		return 1;
> #endif
> ```

Updated in https://git.openwrt.org/?p=openwrt/staging/dangole.git;a=commit;h=f1704f2fc8c03c544ba3b314121e7a848457979a

Ansuel:
Again thx a log for the patch. Will update this while I wait for a review from jow

Ansuel:
The patch has been merged and this is ready.

modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
```diff
@@ -120,12 +120,12 @@
 		"description": "Grant access to wireless status display",
 		"read": {
 			"ubus": {
-				"iwinfo": [ "assoclist" ]
+				"iwinfo": [ "assoclist", "wps_get_status" ]
```
jow-, line 123: Shouldn't that be `"hostapd.*": [ "wps_status" ]` ?

Ansuel: Yes you are right.

jow-:
Merged, thanks!

PR #4519, opened 2020-10-16
luci-app-sqm: convert to client side implementation

Ansuel:
Convert sqm app to client side implementation. Only a small rework in the logic to get qos scripts help text. The rest of the script is a porting 1:1.


hnyman:
cc also @tohojo @moeller0

applications/luci-app-sqm/root/usr/share/luci/menu.d/luci-app-sqm.json
```diff
@@ -0,0 +1,13 @@
+{
+	"admin/network/sqm2": {
```
tohojo, line 2: nit: why sqm2?

tohojo:
Thanks for doing this! Need to spin up an up-to-date openwrt to test it, will try to do so one of the coming days...

applications/luci-app-sqm/root/usr/share/luci/menu.d/luci-app-sqm.json
Ansuel: (while I was testing this I renamed it to sqm2 to have both the old and the new implementation) will fix.
Any other comment?

tohojo: Thanks! And no, not yet, will need to take it for a spin first...

tohojo (changes requested):
Right, took this for a spin, looks good!

You'll need to also update the Makefile, though: I would actually suggest just dropping the `PKG_VERSION` and `PKG_RELEASE` and just switching to the automatic date-based versioning used for the other Luci apps. Also, I believe this means you can drop the `luci-compat` dependency.

Ansuel:
@hnyman think this can be merged

tohojo (approved):
LGTM

PR #4521, opened 2020-10-17
luci-app-acme: Add option for custom ACME server

jannispinter:
This allows the configuration of `acme_server` from the LuCI application.

Signed-off-by: Jannis Pinter <jannis+openwrt@pinterjann.is>


jow-:
Merged, thanks!

PR #4527, opened 2020-10-19
luci-app-acme: Use tabs for certificate section

jannispinter:
This changes the layout of the certificate section. The configuration is now **organized in tabs** and mutually exclusive options are hidden, which makes it **more accessible for the end user**.

For a standard configuration (fetch a certificate from Let's Encrypt using standalone mode), only the `General Settings` tab needs to be configured/adapted. Anything else can be found in either the `Services`, `Challenge Validation` or the `Advanced Settings` tab.

This change also adds the `Days until renewal` option, which can be used later on by the `acme` package to set the `--days` option in `acme.sh`.

**Note:** @tohojo and others: This is a proposal, please give me some feedback on what you like/dislike with this new tabbed UI. You can find a demo of the new UI below.

acme

Signed-off-by: Jannis Pinter <jannis+openwrt@pinterjann.is>

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
@@ -24,7 +24,7 @@ return view.extend({
 				"Check the logs for progress and any errors."));
 
 		s = m.section(form.TypedSection, "acme", _("ACME global config"));
-		s.anonymous = true
+		s.anonymous = true;
 
 		o = s.option(form.Value, "state_dir", _("State directory"),
 			_("Where certs and other state files are kept."));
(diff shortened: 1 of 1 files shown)


tohojo:
I like it! A few comments:

- I wonder if the 'services' should not just be part of 'general settings'. I imagine it will be fairly common to want to enable one of those, and it's just a couple of check boxes.

- Should we add a bit of help text to the 'standalone' option while we're at it?

Also, very cool with the demo .gif  :)

jannispinter:
Thanks! I have removed the `Services` tab and moved it to `General Settings`. Additionally, I have added a short descriptive text for the individual validation methods.

grafik

PR #4528, opened 2020-10-19
Set the fallback default of rollback timeout to 90s

hnyman:
cc @jow- @feckert 

luci-base & luci-app-dockerman: set the fallback default of rollback timeout to 90s

Set the fallback value of the config change rollback timeout to 90 seconds to match the change in /etc/config/luci by commit 81cf99a50.

That commit changed the default value in the config file, but did not change the underlying fallback values that do get applied when there is no proper config item in etc/config/luci.

Users sysupgrading from old systems may have carried an ancient /etc/config/luci (without rollback config) with them, so this change should help them to see the intended user experience of 90 seconds instead of 30 seconds that may be too short e.g. for wifi changes.

Related to forum discussion at https://forum.openwrt.org/t/luci-changing-wifi-config-results-in-timeout/77104/9

Ps.
The luci-base part should also be backported to 19.07.

### DIFF
--- a/applications/luci-app-dockerman/luasrc/view/dockerman/apply_widget.htm
+++ b/applications/luci-app-dockerman/luasrc/view/dockerman/apply_widget.htm
@@ -44,7 +44,7 @@
 
 <script type="text/javascript">//<![CDATA[
 	var xhr = new XHR(),
-	uci_apply_rollback = <%=math.max(luci.config and luci.config.apply and luci.config.apply.rollback or 30, 30)%>,
+	uci_apply_rollback = <%=math.max(luci.config and luci.config.apply and luci.config.apply.rollback or 90, 90)%>,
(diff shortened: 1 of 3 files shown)


applications/luci-app-dockerman/luasrc/view/dockerman/apply_widget.htm
```diff
@@ -44,7 +44,7 @@
 
 <script type="text/javascript">//<![CDATA[
 	var xhr = new XHR(),
-	uci_apply_rollback = <%=math.max(luci.config and luci.config.apply and luci.config.apply.rollback or 30, 30)%>,
+	uci_apply_rollback = <%=math.max(luci.config and luci.config.apply and luci.config.apply.rollback or 90, 90)%>,
```
feckert, line 47: LGTM

Issue #4516, opened 2020-10-14
luci-base: Cannot check `isActive` from `onchange` notification


oldium:
When in `onchange` notification (`AbstractValue` and descendants in `form.js`), it is not possible to get the up-to-date `isActive`. The problem is that `checkDepends`, which updates the `hidden` class of elements (and is checked by `isActive`), is called after the `onchange` notification, because of the order of event handler registration:

https://github.com/openwrt/luci/blob/386cd4f4dd2c37e9c081e8f9db36ef03c01a13a1/modules/luci-base/htdocs/luci-static/resources/form.js#L3353-L3357

Maybe this could simply be swapped?

Another thing – would it be possible to make `checkDepends` _public_ (currently it is marked as _private_ API) in order to allow dependency refresh after programmatical updates of option values?

PR #4531, opened 2020-10-20
luci-base: Fix using isActive in widget-change notification.

oldium:
The `onchange` notification handler is called too early to be able to evaluate other widget's `isActive()` status. Solve this by changing order of event handling - first register/execute `map.checkDepends` and then `onchange`.

Fixes: openwrt/luci#4516.


oldium:
I am open to discussion, issue #4516 can be fixed simply by switching the order of events registration. Feel free to suggest even different approach 😀

jow-:
I think we should simply swap the event handlers as you suggested. Can you update the PR accordingly?

oldium:
> I think we should simply swap the event handlers as you suggested. Can you update the PR accordingly?

Done.

jow-:
> Done.

Awesome! Merged.

PR #4538, opened 2020-10-24
luci-base/luci-compat: Replace legend element with h3.

### LINKED ISSUE #3149: luci-theme-material: missing panel title in with client-side view
When using client-side view, panel title is missing:
image
Server-side view is normal:
image

I think it's because the code that creates it excutes on load, but the view is not loaded at that time.
https://github.com/openwrt/luci/blob/e459683822925bcdeb552673a4811aa3b4322640/themes/luci-theme-material/htdocs/luci-static/material/js/script.js#L174-L180

oldium:
The `legend` HTML element is allowed only as first element in `fieldset`, so use `h3` instead, which is rendered the same within themes.

Fixes #3149.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2219,7 +2219,7 @@ var CBITypedSection = CBIAbstractSection.extend(/** @lends LuCI.form.TypedSectio
 			});
 
 		if (this.title != null && this.title != '')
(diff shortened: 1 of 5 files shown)


oldium:
CC @jow-

oldium:
I will also have a look at the possibility to clean-up legend handling in themes.

oldium:
Fixed material handling of new H3 theme, all others look fine.

jow-:
Awesome! Merged, thanks :+1:

Issue #4555, opened 2020-10-31
Luci errors on master : RPC call to luci/getFeatures failed with error -32000: Object not found


curtdept:
Fresh master branch build, Luci system rpc seems broken

![Screenshot_20201031-004827_Chrome](https://user-images.githubusercontent.com/23644717/97773140-ef1dc900-1b12-11eb-9ba9-5fcbe10b13c0.jpg)

hnyman:
I see the same with LuCI Master (git-20.304.73778-5a485f8) / OpenWrt SNAPSHOT r14809-37ef2197e6 


```
RPCError

RPC call to luci/getFeatures failed with error -32000: Object not found
  at handleCallReply (https://192.168.1.1/luci-static/resources/rpc.js?v=git-20.304.73778-5a485f8:11:3)
```

or

```
RPCError
RPC call to luci/getFeatures failed with error -32000: Object not found
at handleCallReply (https://192.168.1.1/luci-static/resources/rpc.js?v=git-20.304.73778-5a485f8:11:3)
at promise callback*parseCallReply (https://192.168.1.1/luci-static/resources/rpc.js?v=git-20.304.73778-5a485f8:9:70)
at promise callback*call (https://192.168.1.1/luci-static/resources/rpc.js?v=git-20.304.73778-5a485f8:5:110)
at declare/
```

And Firefox shows the exception/breakpoint as


> 
> Paused on exception
[...]

PR #4560, opened 2020-10-31
 luci-base: fix syntax error in hostapd features

MartB:
Signed-off-by: Martin Böh <mart.b@outlook.de>

### DIFF
--- a/modules/luci-base/root/usr/libexec/rpcd/luci
+++ b/modules/luci-base/root/usr/libexec/rpcd/luci
@@ -219,7 +219,7 @@ local methods = {
 			rv.cabundle      = fs.access("/etc/ssl/certs/ca-certificates.crt")
 			rv.relayd        = fs.access("/usr/sbin/relayd")
 
-			local wifi_features = { "eap", "11n", "11ac", "11r", "11w", "acs", "sae", "owe", "suiteb192", "wep", "wps }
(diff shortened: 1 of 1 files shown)


MartB:
Sorry for the DCO spam but it didnt like that i tried to use github directly for this minor edit and then it didnt like the mismatch between my realname (which is required for signing off) and my github username.

jow-:
Merged, thanks!

lynxthecat:
Error is back. 

https://github.com/openwrt/luci/issues/4555#issuecomment-927104087

clonin216:
I have the same error message, but I can't find what's the problem?
image
This is our "luci" file
[luci.txt](https://github.com/openwrt/luci/files/8813244/luci.txt)

lynxthecat:
@clonin216 can you try hard reset of browser cache?

clonin216:
@lynxthecat Yes, I did, but the problem persists

clonin216:
@jow- 
I am using the latest version of luci
image
I use the same source code, one computer compiles without this error, but another computer compiles with the error.
When we add these codes, there will be errors, please help me to confirm if there is a problem.
image
image
image
image

clonin216:
I use the same source code to compile on two different computers, one succeeds and the other fails, what's the difference before?

image
image

PR #4561, opened 2020-11-01
Fix one display bug for ap join and #4524

### LINKED ISSUE #4524: Luci chooses WPA3
When I try to connect to a network when I use the network browser even though it does not use WPA3, it always chooses it.
Use snapshots.

Ansuel:
Extensive description the commits log. I think i wrote too much there... ahahah
@jow- 

Fixes #4524


jow-:
Great work! Thanks @Ansuel

Issue #4544, opened 2020-10-26
luci-app-upnp: missing config option to use STUN protocol


dianlujitao:
miniupnpd's init script added STUN related configs but they're not implemented in luci https://github.com/openwrt/packages/blob/99e77a716b346384c776247c3196efeb44a5cb28/net/miniupnpd/files/miniupnpd.init#L83-L85

The possible STUN server candidates can be found in miniupnp's official config file:
https://github.com/miniupnp/miniupnp/blob/85f81235047c5684cbb2ad200e9b61dadb783aaf/miniupnpd/miniupnpd.conf#L12-L30

FYI they can be manually configured in `/etc/config/upnpd` like this:

```
config upnpd 'config'
        ......
        option use_stun '1'
        option stun_host 'stun.stunprotocol.org'
        option stun_port '3478'
```

Ansuel:
Easy to implement ;) will take a shot

PR #4562, opened 2020-11-01
luci-app-upnp: add stun support

Ansuel:
STUN support has been added to upnp init but luci lacks support of it.

Fixes: #4544


jow-:
Merged, thanks @Ansuel !

Issue #4543, opened 2020-10-26
DUID length exceeds 36 chars when adding a static DHCP lease


moetayuko:
My OpenWRT version is OpenWrt 19.07.4 r11208-ce6496d796

Here's a screenshot when adding a static DHCPv6 lease, obviously either the reported DUID from the pull-down menu or the internal length limit is wrong

![image](https://user-images.githubusercontent.com/5774651/97144618-bcd13c00-179f-11eb-91f5-86fc8cd33a30.png)

hnyman:
Looks like 36 hex digits + colons, which apparently are displayed contrary to the validation rule.

This is from master, where the rule expects 20-36 hex chars without colons.  The length fails first, but the colons would cause failure too (as "hexstring" does not allow colons)

https://github.com/openwrt/luci/blob/a805a3178f902fe609d3d4a6c7b6b5b1dad88838/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js#L496

Ansuel:
@dianlujitao can we get more info? Are you using a special service to provide ipv6 dhcp?

dianlujitao:
> @dianlujitao can we get more info? Are you using a special service to provide ipv6 dhcp?

I use `dnsmasq-full` as DHCPv6 server.

dianlujitao:
bump, any updates?

Ansuel:
Do you still have this with latest master?

dianlujitao:
> Do you still have this with latest master?

Seems not with the latest 19.07.4, the colons are gone from the pull-down menu. But I reinstalled `odhcpd-ipv6only` because there are some minor issues when using `dnsmasq-full` alone, dunno if this happens to be a workaround.

hmoffatt:
This is still an issue. If you select a hex value from the drop-down list, they are not accepted as too long. Nor can you type in a full hex string value in the custom entry.

Also, the colons should not be stripped - dnsmasq needs the hex values with colons. Otherwise the values would be considered a string. The dnsmasq manual says

>  It is allowed to use client identifiers (called client DUID in IPv6-land) rather than hardware addresses to identify hosts by prefixing with 'id:'. Thus: --dhcp-host=id:01:02:03:04,..... refers to the host with client identifier 01:02:03:04. It is also allowed to specify the client ID as text, like this: --dhcp-host=id:clientidastext,..... 

I have tried stripping the colons and then the client IDs don't match. I put the hex string with colons in directly and then they match.

[...]

PR #4563, opened 2020-11-01
luci-base: add new duid datatype

Ansuel:
Useful to validate duid that can contains colon and should be checked without.

@jow-  should we fix the generated dropdown or investigate why we can find some value with colon from the start instead of a new type?

Fixes: #4543


jow-:
Yeah, I think we should figure out where the DUID with colons comes from. For now I am reluctant to add a special type for it.

Ansuel:
Could be that the user use a different dhcp service? Will ask and check that... 
Anyway we can also use a dedicated validation function for this. (hacky but practical... don't know if having colon in the duid config would cause any problem)

Ansuel:
Suppressed by #4642

Issue #2085, opened 2018-08-19
luci-mod-network: may need a "Please fill in the SSID" prompt


fengchen-github:
On the wireless configuration page, you may need a "Please fill in the SSID" prompt. When the wireless mode is the client and the AP hides the SSID.Wireless connection cannot be successful without SSID.

PR #4565, opened 2020-11-01
luci-mod-network: require ssid when join hidden wifi

Ansuel:
WiFi join won't work if is tried to connect to a Hidden Wifi without specifying the WiFi SSID.

Fixes: #2085

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1790,11 +1790,13 @@ return view.extend({
 		s.handleJoinConfirm = function(radioDev, bss, form, ev) {
 			var nameopt = L.toArray(form.lookupOption('name', '_new_'))[0],
 			    passopt = L.toArray(form.lookupOption('password', '_new_'))[0],
+			    ssidopt = L.toArray(form.lookupOption('ssid', '_new_'))[0],
 			    bssidopt = L.toArray(form.lookupOption('bssid', '_new_'))[0],
 			    zoneopt = L.toArray(form.lookupOption('zone', '_new_'))[0],
 			    replopt = L.toArray(form.lookupOption('replace', '_new_'))[0],
 			    nameval = (nameopt && nameopt.isValid('_new_')) ? nameopt.formvalue('_new_') : null,
 			    passval = (passopt && passopt.isValid('_new_')) ? passopt.formvalue('_new_') : null,
+			    ssidval = (ssidopt && ssidopt.isValid('_new_')) ? ssidopt.formvalue('_new_') : null,
 			    bssidval = (bssidopt && bssidopt.isValid('_new_')) ? bssidopt.formvalue('_new_') : null,
 			    zoneval = zoneopt ? zoneopt.formvalue('_new_') : null,
 			    enc = L.isObject(bss.encryption) ? bss.encryption : null,
@@ -1840,6 +1842,9 @@ return view.extend({
 				else if (bss.bssid != null) {
 					uci.set('wireless', section_id, 'bssid', bss.bssid);
 				}
+
+				if (ssidval != null)
+					uci.set('wireless', section_id, 'ssid', ssidval);
 
 				if (is_sae) {
 					uci.set('wireless', section_id, 'encryption', 'sae');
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1911,6 +1911,16 @@ return view.extend({
 				]).then(this.renderContents.bind(this));
 			};
 
+			if (bss.ssid == null) {
+				name = s2.option(form.Value, 'ssid', _('WiFi SSID'), _('SSID must be manually set when a Hidden WiFi is used.'));
```
jow-, line 1915: Please change to:
`_('Network SSID')` and `_('The correct SSID must be manually specified when joining a hidden wireless network')`

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1911,6 +1911,16 @@ return view.extend({
...
+				name.validate = function(section_id, value) {
```
jow-, line 1916: I think you can simply replace the entire `name.validate` function with `name.optional = false` or `name.rmempty = false`

jow- (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1911,6 +1911,11 @@ return view.extend({
 				]).then(this.renderContents.bind(this));
 			};
 
+			if (bss.ssid == null) {
+				name = s2.option(form.Value, 'ssid', _('Network SSID'), _('The correct SSID must be manually specified when joining a hidden wireless network'));
+				name.optional = false
```
jow-, line 1916: Can you add a semicolon here (sorry to be pedandic about that :) )

Ansuel: np i'm stupid that i miss think kind of things...

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1911,6 +1911,11 @@ return view.extend({
 				]).then(this.renderContents.bind(this));
 			};
 
+			if (bss.ssid == null) {
```
jow-, line 1914: I think you also need to actually handle the new `ssid` option somehow around [line 1791](https://github.com/openwrt/luci/pull/4565/files#diff-adf712ab47ca2e5c052d0ad45de5737b2be6860525cd51ec8f424c7e3a43b78aR1791)

jow- (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1790,11 +1790,13 @@ return view.extend({
 		s.handleJoinConfirm = function(radioDev, bss, form, ev) {
 			var nameopt = L.toArray(form.lookupOption('name', '_new_'))[0],
 			    passopt = L.toArray(form.lookupOption('password', '_new_'))[0],
+			    ssidopt = L.toArray(form.lookupOption('ssid', '_new_'))[0],
```
Ansuel, line 1793: ssid or essid ?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1837,8 +1839,12 @@ return view.extend({
 					if (bssidval == '1')
 						uci.set('wireless', section_id, 'bssid', bss.bssid);
 				}
-				else if (bss.bssid != null) {
-					uci.set('wireless', section_id, 'bssid', bss.bssid);
+				else {
+					if (ssidval != null)
```
Ansuel, line 1843: I'm a little confused now... bssid should always available right?
Should I put this if in the bssid check or leave as is? 

Or remove all the check as bssid should be always available and ssidval is mandatory?
Or check both at the same time with an and? (think this should be correct)

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
Ansuel: Didn't notice that.. Tought the data were saved automagically.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
jow-: However you named the option - `'ssid'` in your case.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
jow-: Leave the BSSID logic as-is, do the SSID stuff in parallel.

jow-:
Merged, thanks @Ansuel !

PR #4570, opened 2020-11-03
luci-proto-wireguard: change allowed range of mtu

### LINKED ISSUE #4282: wireguard MTU cannot exceed 1420
https://github.com/openwrt/luci/blob/b8ff46c6d6f403b5881f812d91611b6b5b2167ed/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js#L77-L79

```
ethernet_mtu = 1500
sizeof_ip4_hdr = 20
sizeof_udp_hdr = 8
wireguard_overhead = 32
ip4_wireguard_mtu = ethernet_mtu - sizeof_ip4_hdr - sizeof_udp_hdr - wireguard_overhead = 1440
```

Per my understanding, a MTU of 1440 should be acceptable.

zhangyoufu:
`range(1280,1420)` => `range(0,8940)`
default value 1420 keeps unchanged
fix #4282

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -95,7 +95,7 @@ return network.registerProtocol('wireguard', {
 		o.optional = true;
 
 		o = s.taboption('advanced', form.Value, 'mtu', _('MTU'), _('Optional. Maximum Transmission Unit of tunnel interface.'));
-		o.datatype = 'range(1280,1420)';
+		o.datatype = 'range(0,8940)';
 		o.placeholder = '1420';
 		o.optional = true;
 


zhangyoufu:
@hnyman Please take a look at this pull request. Thank you!

hnyman:
cc @zx2c4 who might know better the value of the proposition...

zx2c4:
Overhead for wg-over-v6 is 80 bytes. Overhead for wg-over-v4 is 60 bytes. 1420 is thus a sensible default assuming 1500 byte Ethernet.

But as far as a _maximum_... Shouldn't it be closer to 2**16? Don't people use OpenWRT with jumbo frames and such?

zhangyoufu:
According to https://github.com/torvalds/linux/commit/175f1ca9a9ed8689d2028da1a7c624bb4fb4ff7e, it seems that MTU=0 is "quasi-valid". So I changed the lower bound to 0.

According to [wikipedia](https://en.wikipedia.org/wiki/Jumbo_frame#Super_jumbo_frames), super jumbo frame (MTU>9000) is not widely adopted, and has big memory footprint which does not suit most OpenWRT supported devices. So I changed the upper bound to 9000(Ethernet)-20(IPv4)-8(UDP)-32(Wireguard)=8940.

PTAL

feckert:
I pushed the change directly with https://github.com/openwrt/luci/commit/7206d8fd6aac0708dda6ac17c96709e066f6328b

simondeziel:
> According to [wikipedia](https://en.wikipedia.org/wiki/Jumbo_frame#Super_jumbo_frames), super jumbo frame (MTU>9000) is not widely adopted, and has big memory footprint which does not suit most OpenWRT supported devices. So I changed the upper bound to 9000(Ethernet)-20(IPv4)-8(UDP)-32(Wireguard)=8940.

8940 might be a sensible upper limit for most devices but ultimately, the admin/user should be able to ignore the *recommendation* if the target hardware can take it. Even that wikipedia article mentions 64k jumbo frames being the next step ;)

That said, I appreciate the higher upper limit now so thanks!

Issue #4523, opened 2020-10-17
luci-mod-network: It is possible to create a invalid network in the "General Setup" panel.


krieghof:
The Interface name can't include special characters like "-". In the "General Setup" however you can specify such a interface name.

For example:
![Screenshot from 2020-10-17 23-00-39](https://user-images.githubusercontent.com/17674464/96353469-b4587180-10cc-11eb-81ee-c4c68066ec10.png)

The "Add new interface ... " function under Network -> Interfaces, is correctly implemented. 
![Screenshot from 2020-10-17 23-07-46](https://user-images.githubusercontent.com/17674464/96353590-b1aa4c00-10cd-11eb-828b-b06e852fed75.png)It is not possible to click on "Create Interface"

I'm not quite sure, because I don't know how the backend works, but this bug could potentially ruin the config. UCI could parse, a part of the interface name as an option.    

Firmware Version: OpenWrt 19.07.4 r11208-ce6496d796 / LuCI openwrt-19.07 branch git-20.247.75781-0d0ab01

Ansuel:
Can confirm this bug! So no validation in the network selector

PR #4573, opened 2020-11-04
luci-base: deny creation of interface with - for NetworkSelect

Ansuel:
widgets.NetworkSelect permit to create an interface with '-'. This would brake the entire network config. Add a global validation for this that is evaluated before the custom validation if provided.

@jow-

Fixes: #4523


jow-:
Fixed via 2b99473f2

PR #4579, opened 2020-11-06
luci-proto-qmi: add missing proto default options

### LINKED ISSUE #4577: QMI protocol: introduce option "Use DNS servers advertised by peer"
It's not possible to request that "QMI Cellular" interface does not modify /etc/resolv.conf when it gets an IP:

image

Other interfaces have a way to enforce that. For example, a generic interface "DHCP" has this option:

image

feckert:
This change add the following missing default options.
- defaulroute
- peerdns
- metric
This should fixes #4577 

@Nowaker **Since I do not use the qmi I cannot test it**.
Can you please check if it meets your requirements and give a feedback?

### DIFF
--- a/protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
+++ b/protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
@@ -103,5 +103,22 @@ return network.registerProtocol('qmi', {
 		o.value('ipv4', 'IPv4');
 		o.value('ipv6', 'IPv6');
 		o.default = 'ipv4v6';
+
+		o = s.taboption('advanced', form.Flag, 'defaultroute',
+			_('Use default gateway'),
+			_('If unchecked, no default route is configured'));
+		o.default = o.enabled;
+
+		o = s.taboption('advanced', form.Value, 'metric',
+			_('Use gateway metric'));
+			o.placeholder = '0';
+		o.datatype = 'uinteger';
+		o.depends('defaultroute', '1');
+
+		o = s.taboption('advanced', form.Flag, 'peerdns',
+			_('Use DNS servers advertised by peer'),
+			_('If unchecked, the advertised DNS server addresses are ignored'));
+		o.default = o.enabled;
+
 	}
 });


Nowaker:
@feckert Thanks, I will.

I'm just wondering how where $peerdns gets its value from. https://github.com/openwrt/openwrt/blob/dd651e54cc5eadba480a56a7d2c18471e560f491/package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh#L330 It doesn't look like it's defined in `proto_qmi_init_config`. To me, though I've never developed anything in OpenWRT/Luci, it looks like `$peerdns` was copied from somewhere but never got its value mapped from the config.

Nowaker:
@feckert Thanks a lot for this. The fix seems to work, as restarting the interface in Luci did not result in a modified `/etc/resolv.conf`.

Nowaker:
@feckert That said - the behavior I historically observed was very irregular. Sometimes my `/etc/resolv.conf` would stay unmodified for weeks. Sometimes, it keeps changing a lot. So I don't know for sure. Please reconsider my comment above about `qmi.sh` that likely needs to fetch the value of `peerdns`.

feckert:
@Nowaker I think the change you are looking for  *peerdns* for ipv4 was send to the [mailinglist](https://lists.openwrt.org/pipermail/openwrt-devel/2020-November/032135.html) resently

Nowaker:
> @Nowaker I think the change you are looking for _peerdns_ for ipv4 was send to the [mailinglist](https://lists.openwrt.org/pipermail/openwrt-devel/2020-November/032135.html) resently

I doubt this submission is going to change anything. It still does not fetch the value of peerdns from UCI using json_get_vars so $peerdns is always empty.

> I think if the change in uqmi is merged we should then also add the following code to luci-proto-qmi
> https://github.com/openwrt/luci/blob/fd40efdf4da6322fef9e3b02e713c914bc8ad530/protocols/luci-proto-ppp/htdocs/luci-static/resources/protocol/l2tp.js#L62-L65

This makes sense, sure. I won't use since I have a static reaolv.conf, but it makes sense.

feckert:
@Nowaker Independent of the custom dns uci config option, does this change in the luci-proto-qmi solve your issue?
So you could set the defaultroute, metric and peerdns uci Options in the qmi proto?
In short, can I merge this?

Nowaker:
Yes, this is a merging material for sure.

Whether this actually fixes my problem is yet to be seen, because: 

> @feckert That said - the behavior I historically observed was very irregular. Sometimes my `/etc/resolv.conf` would stay unmodified for weeks. Sometimes, it keeps changing a lot. So I don't know for sure.

I think qmi.sh needs some changes but since it's not part of luci app, the luci part is complete and can be merged. Thanks!

PR #4581, opened 2020-11-06
luci-app-firewall: force user to add a meaningful name

feckert:
The uci name option is set as a comment by fw3 into iptables.
If no name es set, then the name in the comment of the rule is
for exmaple set to @rule[1].

You will surely wonder why I need this!
Explanation:
I have the requirement that I want to combine several firewall rules into one group. These should then be switched on together. For example the firewall rules you have to set for IPsec. I have decided to define an additional firewall section with the name group where the names of the single rules are listed. When I call up the group, all rules with the defined names are switched on and the firewall is reloaded. But this is currently not possible because the name is optional and not clear.
I can remember the rule id @rule[1], but it is not unique and can move under circumstances, because the position of the uci section in the firewall configuration is important.
This is only a suggestion or does anyone else or @jow- have something else in mind?

Feedback welcome ;-)


jow-:
NAK. There is no requirement to set a name.

feckert:
@jow- do you have then another hint for a solution to set grouped firewall rules?

PR #4595, opened 2020-11-15
luci-app-packet-capture: add tcpdump application

wjowsa:
In some troubleshooting situations (e.g. catching logs by unprivileged user) is useful to run network troubleshooting tools like tcpdump or iperf3 directly from the Luci interface. This PR introduces a Troubleshooting menu and a first tool - Packet Capture. 
Packet Capture application is a fronted for tcpdump. With this application, a user is able to configure the most important and commonly used options of tcpdump. The output can be lived streamed to the browser or saved to a file.
The application uses the new [uhttpd API](https://git.openwrt.org/?p=project/uhttpd.git;a=commit;h=11723570af9cb7bd87842e79c85ee99530be9902). The EventSource interface is used in order to handle server-sent events. However, in order to make the API work properly, there are two patches for uhttpd that have to be applied.
 * Because the EventSource does not allow to set custom header fields, the uhttpd has patched to support passing rpcd session id in the subscription url. [patch 1](https://patchwork.ozlabs.org/project/openwrt/patch/20201110080209.6368-1-wojciech.jowsa@gmail.com/)
* The fix for uhttpd crash.  [patch  2](https://patchwork.ozlabs.org/project/openwrt/patch/20201115091917.7798-1-wojciech.jowsa@gmail.com/)
I have tested this application on TP-Link WDR4300 and PCEngines APU3 using the latest openwrt master.

packet_capture

Signed-off-by: Wojciech Jowsa <wojciech.jowsa@gmail.com>


applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+	fs.exec("/usr/sbin/packet_capture_stop").then(function(replay) {
+		if(eventSource) { 
+			eventSource.close();
+		}
+	}.bind(this)).catch(function(error) {
+		console.log(error);
+	});
+}
+
+window.onbeforeunload = stopTcpdump;
```
jow-, line 24: Please use `window.addEventListener('beforeunload', stopTcpdump);`

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+'require network';
+'require tools.widgets as widgets';
+
+var eventSource,
+	captureFilePoll,
+	hostName;
+
+function stopTcpdump() {
+	fs.exec("/usr/sbin/packet_capture_stop").then(function(replay) {
+		if(eventSource) {
```
jow-, line 16: Please put a space between the keyword (`if`) and the left paren (`if (eventSource)`). Also please omit the curly braces for single-statement loops or conditionals:

```
if (eventSource)
    eventSource.close();
```

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+
+var callInitAction = rpc.declare({
+	object: 'luci',
+	method: 'setInitAction',
+	params: [ 'name', 'action' ],
+	expect: { result: false }
+});
+
+function findButton(title) {
+	var buttons = document.querySelectorAll('button');
```
jow-, line 40: This will not work once `Download` is translated into other languages. I suggest to add a dedicated attribute to the button, e.g. `data-action="download"` which you can match. This would also simplify this function into `return document.querySelector('button[data-action="..."]');`

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+	for (var i=0; i < buttons.length; i++) {
+		if (buttons[i].firstChild.nodeValue == title) {
+			return buttons[i];
+		}
+	}
+}
+
+function addOutput() {
+	var tcpdumpOut = document.querySelectorAll('[id$="tcpdump_out"]')[0];
+	if(tcpdumpOut) {
```
jow-, line 50: Space after `if` and no curly braces:
```
if (tcpdumpOut)
    return;
```

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+			'class': 'cbi-input-textarea',
+			'readonly': '',
+			'style': 'width:100%',
+			'rows': 30,
+	}));
+
+	frameEl.firstElementChild.style.fontFamily = 'monospace';
+
+	var downloadBtn = document.querySelectorAll('[id$="download_file"]')[0];
+	if(downloadBtn) {
```
jow-, line 67: Space after `if` and no curly braces

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+}
+
+var downloadCaptureFile = function(ev) {
+	var form = E('form', {
+		method: 'post',
+		action: '/cgi-bin/cgi-download',
+		enctype: 'application/x-www-form-urlencoded'
+	}, E('input', { type: 'hidden', name: 'sessionid', value: rpc.getSessionID()},
+	   E('input', { type: 'hidden', name: 'path', value: "/tmp/capture.pcap"},
+	   E('input', { type: 'hidden', name: 'filename', value: hostName + "-" + Date.now() + ".pcap"}
```
jow-, line 79: Suggest to add `E('input', { type: 'hidden', name: 'mimetype', value: 'application/vnd.tcpdump.pcap'})`

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+		downloadBtn.parentNode.insertBefore(frameEl, downloadBtn.nextSibling);
+	}
+}
+
+var downloadCaptureFile = function(ev) {
+	var form = E('form', {
+		method: 'post',
+		action: '/cgi-bin/cgi-download',
+		enctype: 'application/x-www-form-urlencoded'
+	}, E('input', { type: 'hidden', name: 'sessionid', value: rpc.getSessionID()},
```
jow-, line 77: The inputs here are improperly nested:
```javascript
E('input', {...},
  E('input', {...},
    E('input', {...},
      ...)))
```

Should be:

```javascript
E('input', {...}),
E('input', {...}),
E('input', {...}),
...
```

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
```diff
@@ -0,0 +1,293 @@
...
+	eventSource.addEventListener("tcpdump.data", function(event) {
+		textOut.value = textOut.value + "\n" + JSON.parse(event.data).data;
+	});
+}
+
+function updateButtons() {
+	var tasks = [];
+	tasks.push(callLuciProcessList().then(L.bind(function(list) {
+		for (var i = 0; i < list.length; i++) {
+			if(list[i].COMMAND.includes("/usr/sbin/packet_capture")) {
```
jow-, line 106: For this and the subsequent `if` conditions:
 - add space after `if`
 - omit curly braces for single-statement conditionals

jow-, line 106: Hm, so you're doing the equivalent of `ps auxww ·| grep /usr/sbin/packet_capture` here? This does not look very robust imho. Can't we extend `/usr/sbin/packet_capture` to provide some kind of status subcommand? E.g. something like `/usr/sbin/packet_capture is_running` which either exits with `0` (yes) or `1` (no) - or which outputs some JSON status info to stdout.

applications/luci-app-packet-capture/root/usr/sbin/packet_capture
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env lua
```
jow-, line 1: I believe this and the subsequent helper executables or not meant to be directly usable on the shell as normal commands, so please move them into `/usr/libexec/` to not pollute `$PATH`. Also I think all these scripts should be merged into one single executable.

applications/luci-app-packet-capture/root/usr/sbin/packet_capture_start
```diff
@@ -0,0 +1,42 @@
...
+json_get_var file file
+
+args="-n"
+
+if [ "$domains" == "1" ];then
+	args=""
+fi
+
+if  [ -n "$interface" ];then
+	args="$args  -i  $interface"
```
jow-, line 21: Check for `-e "/sys/class/net/$interface"` to prevent shell injection via crafted `interface` JSON field value

applications/luci-app-packet-capture/root/usr/sbin/packet_capture_start
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+if [ -n "$packets" ];then
```
jow-, line 24: Use something like  `if echo "$packets" | grep -qE '^[0-9]+$'; then` to prevent shell injection via crafted `packets` JSON field value

applications/luci-app-packet-capture/root/usr/sbin/packet_capture_start
```diff
@@ -0,0 +1,42 @@
...
+if [ "$verbose" == "1"  ];then
+	args="$args  -e"
+fi
+
+if [ "$file" == "1" ];then
+	mem=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
+	args="$args -W 2 -C $((mem/(1024 * 10))) -w /tmp/capture.pcap -z /usr/sbin/packet_capture_stop"
+fi
+
+if [ -n "$filter" ];then
```
jow-, line 37: I suggest to use something like `if [ -n "$filter" ] && tcpdump -i lo -d "$filter" >/dev/null 2>/dev/null; then` to test the filter expression for validity.

applications/luci-app-packet-capture/root/usr/sbin/packet_capture_start
```diff
@@ -0,0 +1,42 @@
...
+    echo "$filter" >> /tmp/tcpdump_filter
+fi
+
+(/usr/sbin/packet_capture "$args" "$duration")&
```
jow-, line 41: I suggest to write the resulting pid (`$!`) into a pid file...

applications/luci-app-packet-capture/root/usr/sbin/packet_capture_stop
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+pgrep tcpdump | xargs kill -9
```
jow-, line 3: ... and use that pid file here to avoid killing unrelated processes.

```bash
pid=$(cat /var/run/packet_capture.pid)
if [ -n "$pid" ] && grep -sq tcpdump "/proc/$pid/cmdline"; then
    kill -TERM $pid
fi
rm -f /var/run/packet_capture.pid
```

applications/luci-app-packet-capture/root/usr/share/luci/menu.d/luci-app-packet-capture.json
```diff
@@ -0,0 +1,25 @@
...
+	},
+	"admin/troubleshooting/packet_capture": {
+		"title": "Packet Capture",
+		"order": 1,
+		"action": {
+			"type": "view",
+			"path": "packet_capture/tcpdump"
+		},
+		"depends" : {
+            "acl": [ "luci-app-packet-capture" ],
```
jow-, line 17: The indentation here appears to be inconsistent

applications/luci-app-packet-capture/root/usr/share/rpcd/acl.d/luci-app-packet-capture.json
```diff
@@ -0,0 +1,24 @@
...
+		"description": "Grant access to tcpdump ubus object",
+		"read": {
+			"cgi-io": [ "download", "exec" ],
+			"ubus": {
+				"tcpdump": [ "*" ],
+				"luci": [ "getProcessList" ]
+			},
+			"uci": [ "packet_capture", "system" ],
+			"file": {
+				"/tmp/capture.pcap": [ "read", "write" ]
```
jow-, line 12: Please grant only `read` permission here...

applications/luci-app-packet-capture/root/usr/share/rpcd/acl.d/luci-app-packet-capture.json
```diff
@@ -0,0 +1,24 @@
...
+			}
+		},
+		"write": {
+			"uci": [ "packet_capture" ],
+			"file": {
```
jow-, line 17: ... and `"/tmp/capture.pcap": [ "write" ]` here.

jow- (changes requested):

jow-:
Thanks a lot for your contribution @wjowsa - I added a couple of review comments.

wjowsa:
Thanks for the review @jow- , I will apply your comments soon.
The credit goes also to @kristrev who helped me with the design and development of this application.  I hope it will make debugging of networking issues more user-friendly :)

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
wjowsa: Ok

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
wjowsa: Ok.

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
wjowsa: Ok

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
wjowsa: If I change nesting then only the first attribute (sessionid) and e.g. path is not passed.

applications/luci-app-packet-capture/root/usr/sbin/packet_capture_stop
wjowsa: Actually tcpdump process has to be terminated to that is does not become zombie

applications/luci-app-packet-capture/root/usr/sbin/packet_capture
wjowsa: I moved scripts to libexec. One executable is a good idea but if you agree then I would like to keep separate scripts for now. Merging scripts sh and lua into one robust script might require some additional refactoring here and there so if possible then I would like to let users try this application. I promise that I will merge it into one exec in the nearest feature. More, I am going to push iperf3 application that is similar to the packet capture application. Iperf3 application will have a single executable. After that, I will copy the same structure to the packet capture application.

applications/luci-app-packet-capture/htdocs/luci-static/resources/view/packet_capture/tcpdump.js
wjowsa: I think that we can check whether /var/run/packet_capure.pid file exists or not? It works fine together with the changes you proposed below.

wjowsa:
@jow- I think I have addressed all your comments. Unfortunately, tcpdump does allow to set maximum capture file size yet so I had to use a workaround to add this functionality. However, there is a PR https://github.com/the-tcpdump-group/tcpdump/pull/781 which adds the file max-size option. It works fine so potentially it could be added to tcpdump package. Then we could remove the file max-size workaround.

systemcrash:
@wjowsa please address all comments, or you can close this PR. ( It might also help to drive PR you mentioned above for max sizes )

PR #4597, opened 2020-11-16
luci-mod-status: routes.js: link-local neighbours

mikma:
It would be nice if at least some link-local neighbours were shown on the Routes status page. I propose showing all except those containing the MAC/EUI-48 address.


jow-:
Can you please elaborate a bit on the rationale? Why do you filter out IPv6 addresses with embedded MACs?

Issue #4572, opened 2020-11-04
feature: add channel scan for wifi


Ansuel:
I notice that with the wifi scanning feature we have all the data to create a graph with that show the wifi channel utilization. (channel and wifi rssi)
So i'm asking @jow- 
Do you think this feature would be useful? I would try to use the same code we use for the traffic graph so no external libs needed.

evelikov:
Fwiw I would greatly appreciate such a feature.

At the moment, one has to the wlan to monitor mode (via aircrack-ng) and use `iwlist scan` to get said data. Something which I had to do just a few moments ago :-)

Ansuel:
Actually this is already possibile using the WiFi join feature. I'm
searching a way to get channel width.

Il Gio 5 Nov 2020, 22:40 Emil Velikov <notifications@github.com> ha scritto:

Ansuel:
Some WIP
![Immagine 2020-11-09 202627](https://user-images.githubusercontent.com/20289090/98587141-0787a380-22ca-11eb-8f7e-1aa3ceb13d34.png)

Ansuel:
The ""Final"" Result.
A patch is needed for this to work. Now i just need to test this and polish the code. (improve readability since it's very compact 50-60 line)
![Immagine 2020-11-09 202627](https://user-images.githubusercontent.com/20289090/99209677-0079fd00-27c4-11eb-893a-a17d716e728a.png)
![Immagine 2020-11-16 042626](https://user-images.githubusercontent.com/20289090/99209681-01ab2a00-27c4-11eb-9d20-c95f67da3df2.png)

jow-:
Very nice, great work!

Ansuel:
@jow- Where should i put this? Currently is integrated in the wifi scan... Is that right? Was thinking of moving it to the stats page

hnyman:
Statistics sounds wrong,
"scan" is better. 

Maybe create a "channel analysis" button next to "scan" in network/wireless?

Ansuel:
With statistics i mean the normal stats page, the one that show traffic in the interfacae...

Do you think a new button with Channel Analysis that show the graph + wifi (with info related to channel is more appropriate?)

hnyman:
>  With statistics i mean the normal stats page, the one that show traffic in the interfacae...

I got lost with that reference.
Which page you mean?

I see traffic count only on https://192.168.1.1/cgi-bin/luci/admin/network/network that does not have wireless.

Neither 
[...]

PR #4598, opened 2020-11-17
luci-mod-status: add new channel analysis page

Ansuel:
Add channel analysis support. This can be very useful as a user can directly use the webui to check wifi channel utilization without using external tool. This use data already provided by iwinfo.

@jow- This is the """final""" implementation. If you are fine with this (and after some feedback with some testing), we still need to decide where to put this.  @hnyman advised to put this in a separate status menu. Currently this is added to the join wifi modal.

(some details about how the page works, a tab with all the radios is created. The pool function is unique and use the custom event 'tab-change' to detect the tab displayed and based on the tab name, updates the related graph and table. Multiple pool function were not acceptable since this would overload the wifi driver and also slowdown the wifi performance. Tab data is not cleared on tab switch and data is updated on the next pool update. A dummy wifi is added that rappresents the Local Interface.)

**Also this depends on the 4 patch I posted on the mailing list**
https://patchwork.ozlabs.org/project/openwrt/list/?series=218829 (v4)
They are rfc but they are already ready... just some concern in the comments.

Without the 4 patch, any wifi will be wrongly shown with 20mhz width.

Fixes: #4572

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -3597,9 +3597,11 @@ var UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 				this.setActiveTabId(panes[selected], selected);
 			}
 
-			panes[selected].dispatchEvent(new CustomEvent('cbi-tab-active', {
-				detail: { tab: panes[selected].getAttribute('data-tab') }
-			}));
+			requestAnimationFrame(L.bind(function(pane) {
+				pane.dispatchEvent(new CustomEvent('cbi-tab-active', {
+					detail: { tab: pane.getAttribute('data-tab') }
+				}));
+			}, this, panes[selected]));
 
 			this.updateTabs(group);
 		},
--- a/modules/luci-base/htdocs/luci-static/resources/firewall.js
+++ b/modules/luci-base/htdocs/luci-static/resources/firewall.js
@@ -57,21 +57,7 @@ function getColorForName(forName) {
 	else if (forName == 'wan')
 		return '#f09090';
 
-	random.seed(parseInt(sfh(forName), 16));
-
-	var r = random.get(128),
-	    g = random.get(128),
-	    min = 0,
-	    max = 128;
-
-	if ((r + g) < 128)
-		min = 128 - r - g;
-	else
-		max = 255 - r - g;
-
-	var b = min + Math.floor(random.get() * (max - min));
-
-	return '#%02x%02x%02x'.format(0xff - r, 0xff - g, 0xff - b);
+	return random.derive_color(forName);
 }
 
 
--- a/modules/luci-base/htdocs/luci-static/resources/tools/prng.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/prng.js
@@ -89,5 +89,23 @@ return L.Class.extend({
 		}
 
 		return Math.floor(r * (u - l + 1)) + l;
+	},
+
+	derive_color: function(string) {
+		this.seed(parseInt(sfh(string), 16));
+
+		var r = this.get(128),
+		    g = this.get(128),
+		    min = 0,
+		    max = 128;
+
+		if ((r + g) < 128)
+			min = 128 - r - g;
+		else
+			max = 255 - r - g;
+
+		var b = min + Math.floor(this.get() * (max - min));
+
+		return '#%02x%02x%02x'.format(0xff - r, 0xff - g, 0xff - b);
 	}
 });
--- a/modules/luci-mod-status/htdocs/luci-static/resources/svg/channel_analysis.svg
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/svg/channel_analysis.svg
@@ -1,0 +1,19 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+
+<svg width="100%" height="100%" version="1.1" xmlns="http://www.w3.org/2000/svg">
+	<polyline id="rx" points="" style="fill:blue;fill-opacity:0.4;stroke:blue;stroke-width:1" />
+	<polyline id="tx" points="" style="fill:green;fill-opacity:0.4;stroke:green;stroke-width:1" />
+
+	<line x1="0" y1="25%" x2="100%" y2="25%" style="stroke:black;stroke-width:0.1" />
+	<text id="label_75" x="10" y="24%" style="fill:#eee; font-size:9pt; font-family:sans-serif; text-shadow:1px 1px 1px #000">-25 dbm</text>
+
+	<line x1="0" y1="50%" x2="100%" y2="50%" style="stroke:black;stroke-width:0.1" />
+	<text id="label_50" x="10" y="49%" style="fill:#eee; font-size:9pt; font-family:sans-serif; text-shadow:1px 1px 1px #000">-50 dbm</text>
+
+	<line x1="0" y1="75%" x2="100%" y2="75%" style="stroke:black;stroke-width:0.1" />
+	<text id="label_25" x="10" y="74%" style="fill:#eee; font-size:9pt; font-family:sans-serif; text-shadow:1px 1px 1px #000">-75 dbm</text>
+
+	<line x1="0" y1="90%" x2="100%" y2="90%" style="stroke:black;stroke-width:0.1" />
+	<text id="label_10" x="10" y="89%" style="fill:#eee; font-size:9pt; font-family:sans-serif; text-shadow:1px 1px 1px #000">-90 dbm</text>
+</svg>
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
@@ -1,0 +1,396 @@
+'use strict';
+'require view';
+'require poll';
+'require request';
+'require network';
+'require ui';
+'require rpc';
+'require tools.prng as random';
+
+return view.extend({
+	callFrequencyList : rpc.declare({
+		object: 'iwinfo',
+		method: 'freqlist',
+		params: [ 'device' ],
+		expect: { results: [] }
+	}),
+
+	callInfo : rpc.declare({
+		object: 'iwinfo',
+		method: 'info',
+		params: [ 'device' ],
+		expect: { }
+	}),
+
+	render_signal_badge: function(signalPercent, signalValue) {
+		var icon, title, value;
+
+		if (signalPercent < 0)
+			icon = L.resource('icons/signal-none.png');
+		else if (signalPercent == 0)
+			icon = L.resource('icons/signal-0.png');
+		else if (signalPercent < 25)
+			icon = L.resource('icons/signal-0-25.png');
+		else if (signalPercent < 50)
+			icon = L.resource('icons/signal-25-50.png');
+		else if (signalPercent < 75)
+			icon = L.resource('icons/signal-50-75.png');
+		else
+			icon = L.resource('icons/signal-75-100.png');
+
+		value = '%d\xa0%s'.format(signalValue, _('dBm'));
+		title = '%s: %d %s'.format(_('Signal'), signalValue, _('dBm'));
+
+		return E('div', {
+			'class': 'ifacebadge',
+			'title': title,
+			'data-signal': signalValue
+		}, [
+			E('img', { 'src': icon }),
+			value
+		]);
+	},
+
+	add_wifi_to_graph: function(chan_analysis, res, scanCache, channels, channel_width) {
+		var offset_tbl = chan_analysis.offset_tbl,
+			height = chan_analysis.graph.offsetHeight - 2,
+			step = chan_analysis.col_width,
+			height_diff = (height-(height-(res.signal*-4)));
+
+		if (scanCache[res.bssid].color == null)
+			scanCache[res.bssid].color = random.derive_color(res.bssid);
+
+		if (scanCache[res.bssid].graph == null)
+			scanCache[res.bssid].graph = [];
+
+		for (var i=0; i < channels.length; i++) {
+			var chan_offset = offset_tbl[channels[i]],
+				points = [
+				(chan_offset-(step*channel_width))+','+height,
+				(chan_offset-(step*(channel_width-1)))+','+height_diff,
+				(chan_offset+(step*(channel_width-1)))+','+height_diff,
+				(chan_offset+(step*(channel_width)))+','+height
+			];
+
+			if (scanCache[res.bssid].graph[i] == null) {
+				var group = document.createElementNS('http://www.w3.org/2000/svg', 'g'),
+					line = document.createElementNS('http://www.w3.org/2000/svg', 'polyline'),
+					text = document.createElementNS('http://www.w3.org/2000/svg', 'text'),
+					color = scanCache[res.bssid].color;
+
+				line.setAttribute('style', 'fill:'+color+'4f'+';stroke:'+color+';stroke-width:0.5');
+				text.setAttribute('style', 'fill:'+color+';font-size:9pt; font-family:sans-serif; text-shadow:1px 1px 1px #000');
+				text.appendChild(document.createTextNode(res.ssid || res.bssid));
+
+				group.appendChild(line)
+				group.appendChild(text)
+
+				chan_analysis.graph.firstElementChild.appendChild(group);
+				scanCache[res.bssid].graph[i] = { group : group, line : line, text : text };
+			}
+
+			scanCache[res.bssid].graph[i].text.setAttribute('x', chan_offset-step);
+			scanCache[res.bssid].graph[i].text.setAttribute('y', height_diff - 2);
+			scanCache[res.bssid].graph[i].line.setAttribute('points', points);
+			scanCache[res.bssid].graph[i].group.style.zIndex = res.signal*-1;
+			scanCache[res.bssid].graph[i].group.style.opacity = res.stale ? '0.5' : null;
+		}
+	},
+
+	create_channel_graph: function(chan_analysis, freq_tbl, is5GHz) {
+		var columns = is5GHz ? freq_tbl.length * 4 : freq_tbl.length + 3,
+		    chan_graph = chan_analysis.graph,
+		    G = chan_graph.firstElementChild,
+		    step = (chan_graph.offsetWidth - 2) / columns,
+		    curr_offset = step;
+
+		function createGraphHLine(graph, pos) {
+			var elem = document.createElementNS('http://www.w3.org/2000/svg', 'line');
+			elem.setAttribute('x1', pos);
+			elem.setAttribute('y1', 0);
+			elem.setAttribute('x2', pos);
+			elem.setAttribute('y2', '100%');
+			elem.setAttribute('style', 'stroke:black;stroke-width:0.1');
+			graph.appendChild(elem);
+		}
+
+		function createGraphText(graph, pos, text) {
+			var elem = document.createElementNS('http://www.w3.org/2000/svg', 'text');
+			elem.setAttribute('y', 15);
+			elem.setAttribute('style', 'fill:#eee; font-size:9pt; font-family:sans-serif; text-shadow:1px 1px 1px #000');
+			elem.setAttribute('x', pos + 5);
+			elem.appendChild(document.createTextNode(text));
+			graph.appendChild(elem);
+		}
+
+		chan_analysis.col_width = step;
+
+		createGraphHLine(G,curr_offset);
+		for (var i=0; i< freq_tbl.length;i++) {
+			var channel = freq_tbl[i].channel
+			chan_analysis.offset_tbl[channel] = curr_offset+step;
+
+			createGraphHLine(G,curr_offset+step);
+			createGraphText(G,curr_offset+step, channel);
+			curr_offset += step;
+
+			if (is5GHz && freq_tbl[i+1]) {
+				var next_channel = freq_tbl[i+1].channel;
+				/* Check if we are transitioning to another 5Ghz band range */
+				if ((next_channel - channel) == 4) {
+					for (var j=1; j < 4; j++) {
+						chan_analysis.offset_tbl[channel+j] = curr_offset+step;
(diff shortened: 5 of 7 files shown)


hnyman:
For some reason I can't get this to work in LuCI. The scan from LuCI hangs with spinning wheel on scan button, and Firefox shows a JS error:
image

But the channel widths are shown in the patches iwinfo itself.
 
```
root@router1:~# iwinfo wlan0 scan
Cell 01 - Address: 58:90:4xxxxx
          ESSID: "DNA-Wxxxx"
          Mode: Master  Channel: 36
          Signal: -90 dBm  Quality: 20/70
          Encryption: WPA2 PSK (CCMP)
          HT Operation:
                    Primary Channel: 36
                    Secondary Channel Offset: above
                    Channel Width: 20 MHz
          VHT Operation:
                    Channel Width: 80 MHz
                    Center Frequency 1: 42
                    Center Frequency 2: 0
```

(Not sure if I patched rpcd wrongly, or if that needs to be in the compiled image itself)

Ansuel:
@hnyman yes the rpcd patch is needed to add the new data to ubus
the ui use this command to get the data
ubus call iwinfo scan '{"device":"wlan0"}'

Let me check the error with the function

Ansuel:
Fixed the problem... (my fault working on old source....) pls give another test (or just change `callFrequencyList `to `this.callFrequencyList`)

hnyman:
I recompiled everything and now it gives a slightly different error:
  "TypeError: this.callFrequencyList is not a function"

image

hnyman:
But the scan itself seems to work:

```
root@router1:~# ubus call iwinfo scan '{"device":"wlan1"}'
{
        "results": [
                {
                        "ssid": "Koti_xxx",
                        "bssid": "FE:F5xxxx",
                        "mode": "Master",
                        "channel": 2,
                        "signal": -85,
                        "quality": 25,
                        "quality_max": 70,
                        "ht_operation": {
                                "primary_channel": 2,
                                "secondary_channel_offset": "above",
                                "channel_width": "any"
                        },
                        "encryption": {
                                "enabled": true,
                                "wpa": [
                                        1,
                                        2
                                ],
                                "authentication": [
                                        "psk"
                                ],
                                "ciphers": [
                                        "tkip",
                                        "ccmp"
                                ]
                        }
                },
                {
                        "ssid": "ASxxx",
                        "bssid": "30:85xxxx",
                        "mode": "Master",
                        "channel": 1,
                        "signal": -91,
                        "quality": 19,
                        "quality_max": 70,
                        "ht_operation": {
                                "primary_channel": 1,
                                "secondary_channel_offset": "no secondary",
                                "channel_width": "20 MHz"
 

```

Ansuel:
Ok good @hnyman sorry again... Anyway to quickly test this... (if you don't want to recompile) Replace the js in the firmware with the one in this pr.

hnyman:
Well, a different error...

image

Does the PR have the same code that actually works in your live router ????

Ansuel:
Now yes... Ok... I placed the svg in the wrong path... Move channel_analysis.svg from` www/luci-static/resources/view` to `www/luci-static/resources`

hnyman:
Well, now it actually worked ;-)

image

Ansuel:
What about 5ghz?

hnyman:
also 5 GHz works:

image

Ansuel:
Thx a lot for the testing

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -289,18 +290,157 @@ function add_dependency_permutations(o, deps) {
...
+				curr_offset += step;
+			}
+		}
+	}
+	createGraphHLine(G,curr_offset+step);
+}
+
+function add_wifi_to_graph(chan_analysis, res, scanCache) {
+	if (scanCache[res.bssid].color == null)
+		scanCache[res.bssid].color = '#'+('00000'+(Math.random()*(1<<24)|0).toString(16)).slice(-6);
```
jow-, line 360: I suggest to use the logic from [`firewall.getColorForName()`](https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/firewall.js#L52) here, this ensures that the same network names always get the same pseudorandom colors.

Ansuel: To not have duplicated code, I generalized part of the `firewall.getColorForName()` and added to prng.

Ansuel:
Added some small modification. 
- line and text are now in a group 
- added support for stale wifi (opacity for wifi detected as stale)
- added a point before the wifi to quickly identify the wifi by the color (useful with wifi with the same name)

Immagine 2020-11-18 024410

pkgadd:
I've successfully tested the current version (b3505b7) on ipq8065/ qca9984 (nbg6817) and lantiq/ AR9227+QCA9880 (bthub5a), it's a nice improvement. 

For the graphical representation it might be even better to inject the local AP interfaces (of the device running the test) into the channel utilization as well.

hnyman:
Interesting, the modified color randomiser now causes an error for me:

image

Apparently it worked for @pkgadd so I wonder what happens here.

Ansuel:
@hnyman think you missed the other commit

hnyman:
I didn't.
It was a browser cache problem.

Apparently the .js file did not get marked as new, as I did not commit the changes, so the version string remained static "prng.js?v=git-20.311.85590-5c5b134". 

Clearing browser cache and restarting the browser helped ;-)

Ansuel:
(shift+refresh = load the page + purge cache for that site)

hnyman:
That didn't work.  Tried it first...

But in any case, the enw code seems to work.

Ansuel:
Do you have any way to test if 5 ghz 160 mhz or 80+80 band works?

hnyman:
Not really.

pkgadd:
Sorry, I don't have any 160 MHz or 80+80 MHz channels in my environment either.

Ansuel:
I tested 80+80 and 160 by manually setting the data. Looks to work so i think this is ready... We still need to decide the place of this. @jow- any hint on that? Is it good? (graph in the wifi scan button?)

hnyman:
That works ok, and it could be also there in any case.
But people rarely need to do "scan for joining", so the nice feature may remain hidden there.

I would also add it to the status menu

Ansuel:
Ok so i will change this to
1. move the graph in a separate page
2. parse the wifi data differently and show more data related to channel width
3. add a button to the wifi settings that redirect to the channel analysis page

Ansuel:
@hnyman I moved the feature to a new page. What do you think?

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -3597,9 +3597,11 @@ var UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 				this.setActiveTabId(panes[selected], selected);
 			}
 
-			panes[selected].dispatchEvent(new CustomEvent('cbi-tab-active', {
-				detail: { tab: panes[selected].getAttribute('data-tab') }
-			}));
+			requestAnimationFrame(L.bind(function(pane) {
```
hnyman, line 3600: nitpicking, but there is a space at the end of the line.
"whitespace error"

hnyman:
Thanks. I like the new "Channel Analysis" page in the status menu.

Merging this requires that @jow- first commits the needed patches for iwinfo and rpcd (and the needed version bumps in OpenWrt repo):
https://patchwork.ozlabs.org/project/openwrt/list/?series=214580

Ps. note that your second patch there patches "rpcd" while the commit title says "iwinfo"...

Ansuel:
Yes I'm following the naming scheme. If you notice in the rpcd repo,
changes related to iwinfo rpcd module have iwinfo as title.

Il Sab 21 Nov 2020, 09:12 Hannu Nyman <notifications@github.com> ha scritto:

Ansuel:
@hnyman i'm thinking of adding to the graph but i need to confirm something...
Do you have a wifi that has set 40mhz width but is actually working to 20mhz? 

I need to check if ht_mode in `ubus call iwinfo info '{"device":"radio1"}'` is what the wifi is configured or what the wifi is actually using. (40 mhz set but due to reg and other wifi is using 20 mhz)

hnyman:
> check if ht_mode in ubus call iwinfo info '{"device":"radio1"}' is what the wifi is configured or what the wifi is actually using. (40 mhz set but due to reg and other wifi is using 20 mhz)

You mean, the width for the radio in the device itself ???

I think that is the configured one, not the really running channel width. Sadly.
Not quite sure, but might be so.

I tried to test that with VHT160 by setting it to 160 and then doing a scan from my other router:
* device itself says: htmode 160
* scan from other device says:  80

Device itself:
```
root@router1:/etc/config# ubus call iwinfo info '{"device":"radio0"}'
{
        "phy": "phy0",
        "bssid": "DC:....",
        "country": "FI",
        "mode": "Master",
        "channel": 36,
        "frequency": 5180,
        "frequency_offset": 0,
        "txpower": 23,
        "txpower_offset": 0,
        "quality_max": 70,
        "noise": 0,
        "htmodes": [
                "HT20",
                "HT40",
                "VHT20",
                "VHT40",
                "VHT80",
                "VHT80+80",
                "VHT160"
        ],
        "hwmodes": [
                "ac",
                "n"
        ],
        "hwmode": "ac",
        "htmode": "VHT160",
        "hardware": {
                "id": [
                        5772,
                        70,
                        5772,
                        51966
                ],
                "name": "Qualcomm Atheros QCA9984"
        }
}
```

Seen from other router:
image

Testing with VHT160 is maybe not quite reliable, but still hints that it may be misleading.

Ansuel:
Will check if I can extract the info from iwinfo... (the center_chan entry should be what we actually use)

Ansuel:
Any I'm stupid and I never check my spelling... My fault I do this kind of thing in the night. Anyway
The new patch
https://patchwork.ozlabs.org/project/openwrt/list/?series=216286

@hnyman would really appreciate some comments about the 'Local Interface' wifi...
- I set it to a dummy -10 signal (doesn't make sense to have a signal of a station)
- The new patchset is needed to know the right channel to parse

pkgadd:
I really love those changes, works nicely on my nbg6817 and bthub5:

2.4 GHz
2g

5 GHz
5g

Tested c89eed8d1e5ba2dc9369caf86d6332e474a64712 and v2 of your rpcd/ iwinfo patches on top of r15021-b0ecae504b and kernel v5.4.79.

pkgadd:
It works also nicely on a tri-radio device (ipq4019/ map-ac2200) with segmented channel coverage (radio0 ch36-ch64, radio2 ch100-ch165).

openwrt-diy:
This feature is great, can it be applied to OpenWrt-19.07?

hnyman:
@jow- 
I think that this works pretty nicely.

Can you look at evaluating/committing the necessary patches to iwinfo and rpcd, which ansuel submitted to mailing list adn patchwork.

image

@Ansuel 
This is past WIP quality, I think.  ;-)

Ansuel:
Still dubious about the Local Interface data and naming!
Also thx to @pkgadd i need to fix the `20 or 40mhz` bug

Ansuel:
Added a way to check when a wifi has the Intolerant Bit set (for 40MHz channel) and fixed the problem :D

pkgadd:
Just to confirm, I've successfully tested this version as well (ipq8065, ipq4019, lantiq; all ath9k/ ath10k wireless).
`Tested-by: Stefan Lippers-Hollmann <s.l-h@gmx.de>`

hnyman:
Works for me, too.

This LuCI PR could be merged once @jow- (or somebody else with write access to OpenWrt main repo) merges the needed changes to rpcd and iwinfo.

jow-:
Unfortunately I can't merge the rpcd and iwinfo patches as is. The iwinfo changes in particular will break the existing ABI. I also do not the like the way the JSON information is structured (e.g. using "40 Mhz" as field value instead of something like `{ "mhz": 40 }`.

I agree that this feature is nice, but the iwinfo side needs more polishing first.

Ansuel:
Sure I can make all the needed change. About the ABI there is no way to resolve the problem?

Ansuel:
@jow- sorry for the tag. Is there a way to add the iwinfo changes with the ABI problem? I could be very wrong but the patches add data and add entries at the end of the struct so in theory bin compiled on the old library should still work.

hnyman:
I think that you need to properly bump the ABI version in Makefile.

https://github.com/openwrt/openwrt/blob/4dc5b647602dc33be9bc940e91e5b849fdc98015/package/network/utils/iwinfo/Makefile#L36

Ansuel:
Think that bump should be done by jow on iwinfo update on the openwrt repo. Think that jow was referring to the fact that he wanted to avoid in any way possibile an ABI change.

jow-:
There's multiple things that I'd like to see done on the iwinfo side:
 - change the libiwinfo packaging to ship a versioned shared library (libiwinfo.so.$ABI_VERSION)
 - bump the ABI version
 - change the JSON output format, to produce something like `"channel_width": 40` or `"channel_width": 8080` instead of `"40 Mhz" or "80 + 80 MHz" respectively
 - add appropriate stubs for `struct iwinfo_ops.center_chan1` and `struct iwinfo_ops.center_chan2` to the other backends *or* check whether `ops->center_chan1` is != NULL before calling it. Right now, the code will segfault with non-nl80211 backends.

Ansuel:
> There's multiple things that I'd like to see done on the iwinfo side:
> 
> * change the libiwinfo packaging to ship a versioned shared library (libiwinfo.so.$ABI_VERSION)

My statement could look a bit rude but I think it would be better if you or someone more skilled than me do this change. I don't want to create some disaster.

> * bump the ABI version

Same thing. (since the iwinfo makefile in the openwrt repo has a specific commit, we can work on fixing the proposed patch and then work on the changes needed to the makefile.

> * change the JSON output format, to produce something like `"channel_width": 40` or `"channel_width": 8080` instead of `"40 Mhz"` or `"80 + 80 MHz"` respectively - didn't look in detail but if multiple things are possible then an array would work, e.g. `"channel_width": [ 20, 40, ... ]`

The width is unique. So an enum that store these value should be sufficent

> * add appropriate stubs for `struct iwinfo_ops.center_chan1` and `struct iwinfo_ops.center_chan2` to the other backends _or_ check whether `ops->center_chan1` is != NULL before calling it. Right now, the code will segfault with non-nl80211 backends.

Think the check is better than a stub that would return 0 or -1. What do you think?

jow-:
> Think the check is better than a stub that would return 0 or -1. What do you think?

Yes, a check is fine with me.

Ansuel:
I pushed v4 https://patchwork.ozlabs.org/project/openwrt/list/?series=218829 and updated the pr to reflect the changes.
Can someone test?

hnyman:
> Can someone test?

v4 patch series may have a whitespace error.  Spaces instead of tab on one original line.

image

(similar spaces are apparently also on the lines below, but those spaces on added lines do not break patching.)


Fixing that tab in the local patch helped me:
```
--- a/package/network/utils/iwinfo/patches/patch1.patch
+++ b/package/network/utils/iwinfo/patches/patch1.patch
@@ -81,7 +81,7 @@ index 0332bc2..6bfe0ab 100644
                        format_quality_max(e->quality_max));
 -              printf("          Encryption: %s\n\n",
 +              printf("          Encryption: %s\n",
-                      format_encryption(&e->crypto));
+                       format_encryption(&e->crypto));
 +              printf("          HT Operation:\n");
 +              printf("                    Primary Channel: %d\n",
 +                     e->ht_chan_info.primary_chan);

```

I will come back to feedback on the functionality itself, once I get it compiled ;-)

hnyman:
After fixing that whitespace error in the v4 iwinfo patch, the channel analysis functionality itself works ok ;-)

image

hnyman:
Hmm, I noticed one minor issue.
I am not quite sure of the exact reason but this is my speculative guess:

If there is a weak station that disappears during a periodic refresh scan, it is taken away from the list, but it may stay on the graph.

See the "kotiboksi..." here. It was shown both on the graph and the list on the first scan, but disappeared from the list in an automatic refresh, maybe 5 seconds later (or is it 10 seconds, I haven't looked). But it stayed visible in the graph.

image

Not a major issue,  but might be something to be fixed later.

Ansuel:
@hnyman 'It's not a bug, it's a feature' cit.
In theory the page should show the channel utilization so any station that comes up and disappear after a while should be still displayed as it actually exists and cause some interference. Do you think this is wrong?
(about the whitespace errors, in theory they should be there to correctly align the args of the function at the next line. Does the patch complains about them? Could be that I messed up that 2 line)

hnyman:
> Does the patch complains about them? Could be that I messed up that 2 line)

Yes it does. The build breaks.
Thatswhy I flagged the issue.

hnyman:
The iwinfo & rpcd part were committed yesterday, so I think that this LuCI part could also be merged.

Ps.
I think that @jow- 's comments yesterday on mailing list regarding iwinfo and rpcd changes were not yet noted when @dangowrt committed your earlier changes, so I suspect that a new modification round for rpcd and iwinfo might be needed. The old version is now live in OpenWrt master.

Ansuel:
Mhhh yep but I think I have to push a new patch to address jow changes.

Il Mer 6 Gen 2021, 08:58 Hannu Nyman <notifications@github.com> ha scritto:

Ansuel:
Anyway I'm really happy that this feature finally will be added. Honestly it's a very useful feature as the very base of the OpenWRT project is Wireless Freedom.

Ansuel:
@hnyman the required commits are now upstream. We just need to wait a review from jow and this can be merged.

hnyman:
>  the required commits are now upstream. We just need to wait a review from jow

Jow made additions/changes to the iwinfo Makefile and merged the bump there, so I think that this LuCI part can now be also merged.

The current LuCI commit version has been in my own builds for the last month and has worked ok in several routers. So lets just merge this and wait for possible feedback (and/or change tweaks).

dangowrt:
@hnyman @Ansuel have you tried LuCI after this got merged? I'm kinda stuck with non-working iwinfo, symptoms feel like libiwinfo-lua is missing (it's not, ldd shows all ok) but since everything switched to JavaScript I have absolutely no clue at all what's going on and how it is supposed to work.

Ansuel:
Honestly in past test... I encounter some problem with these patch if the center chan were at the center of the struct instead of putting them at the bottom. But it should be caused by versioning and not using ABI. 
anyway this use js and ubus so lua is completely skipped.

hnyman:
I have used the channel analysis page since November. The last LuCI patch in this PR stayed the same since 6 Dec and has been included in all my builds.  It has worked ok, and works even currently.

I have currently it in R7800 with  iwinfo/libiwinfo/-lua 2021-01-06-23d2722b-1 with LuCI luci-mod-status git-20.340.80217-277ffef  manually patched with an identical patch as this merged LuCI PR commit.
* The channel analysis works.
* But the other iwinfo related stuff seems semi-dead. (wireless status info in overview shows up almost empty, and in network/wireless is shows as as not associated. But associated devices are still shown with sensible info...)

As this LuCI channel analysis patch has not changed and has no impact on those other LuCI pages, it is likely related to the last minute non-LuCI changes to the iwinfo & rpcd stuff. 

The other wireless packages have worked ok until 5 January (with the patches then available in the mailing list).

My first guess is that the last minute change on January 6th have somehow changed the iwinfo output, so that the other LuCI components need now a refresh.

I have actually in my R7800 archive several firmware versions from 6th and 7th, e.g. one with r15437-b4c0d377f6-20210106 meaning right after the first version of iwinfo & rpcd were merged. I will re-flash those and investigate a bit, when the breakage happens.

jow-:
Fixed by bc121a5a456b564618423b67a6aef859c0ba6a87

BashfulBladder:
@hnyman can you elaborate on this:

> But the other iwinfo related stuff seems semi-dead. (wireless status info in overview shows up almost empty, and in network/wireless is shows as as not associated. But associated devices are still shown with sensible info...)

I have 2 R700 (RC3 & another some NSS build) that exhibit:

- days of only localhost in 5GHz, 2.4GHz had a 100 cells (both in iw & iwinfo)
- noise absent on 2.4GHz (sort of expected as it doesn't have a network configured)
- noise present only in 2 channels in 5GHz (both in iw & iwinfo using this patch below)

```
--- a/iwinfo_cli.c	2021-01-31 09:05:58.000000000 -0500
+++ b/iwinfo_cli.c	2021-07-31 11:45:25.000000000 -0400
@@ -677,6 +677,32 @@
 }
 
 
+static void print_channel_survey(const struct iwinfo_ops *iw, const char *ifname)
+{
+	int i, len;
+	char buf[IWINFO_BUFSIZE];
+	struct iwinfo_survey_entry *e;
+
+	if (iw->survey(ifname, buf, &len) || len <= 0)
+	{
+		printf("No channel survey information available\n");
+		return;
+	}
+
+	for (i = 0; i < len; i += sizeof(struct iwinfo_survey_entry))
+	{
+		e = (struct iwinfo_survey_entry *) &buf[i];
+		
+		printf("Survey data from %s frequency %d MHz: %s noise (active time: %llu ms, busy time: %llu ms)\n",
+			ifname,
+			e->mhz,
+			format_noise((int8_t)e->noise),
+			e->active_time,
+			e->busy_time);			
+	}
+}
+
+
 static void print_txpwrlist(const struct iwinfo_ops *iw, const char *ifname)
 {
 	int len, pwr, off, i;
@@ -878,6 +904,7 @@
 			"	iwinfo <device> assoclist\n"
 			"	iwinfo <device> countrylist\n"
 			"	iwinfo <device> htmodelist\n"
+			"	iwinfo <network> noiselist\n"
 			"	iwinfo <backend> phyname <section>\n"
 		);
 
@@ -973,6 +1000,10 @@
 				case 'h':
 					print_htmodelist(iw, argv[1]);
 					break;
+				
+				case 'n':
+					print_channel_survey(iw, argv[1]);
+					break;
 
 				default:
 					fprintf(stderr, "Unknown command: %s\n", argv[i]);
```
I can get a richer result if I employ:

```
ifconfig wlan0 down
iw phy phy0 interface add scan0 type station
ifconfig scan0 up
iwinfo scan0 noiselist		(< nope, not yet)
iwinfo scan0 scan		(<- this yields 30 cells AND noise across all channels)
iwinfo scan0 noiselist
iw dev scan0 del
ifconfig wlan0 up
killall -HUP hostapd
```
And sure enough it remains that way until the next reboot. Still, that doesn't explain why it it needs to happen at all after a reboot.

Neustradamus:
@hnyman, @abajk: Have you seen the @BashfulBladder comment?

PR #4606, opened 2020-11-20
[RFC] luci-base: introduce afterRender function

Ansuel:
Add new afterRender function. This function can optionally used to execute function after the view rendering. This can be really handy when it's needed the width and height of some element and this kind of data are available only AFTER the page render.

Inspecting the realtime stats scripts I notice that every graph base they width and height value based on the `#view` element. This looks wrong and can cause some problem (I could be wrong and be a problem specific to the material skin) for example the width is wrongly calculated in the wireless graph where the tab bar adds padding on the left and on the right and the graph doesn't count this missing space).
 
Another solution to this problem would be create a one time pool that parse the width and height after some time but this would add extra time to the page generation and looks hacky.
Append some extra function after the DOM generation looks cleaner and quicker. 

I know this is very specific and we really need this just for the graph that needs this kind of parsing so I understand a NAK on this.

@jow- (sorry... i know I'm bothering you A LOT)

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/luci.js
+++ b/modules/luci-base/htdocs/luci-static/resources/luci.js
@@ -1954,7 +1954,9 @@
 
 					DOM.content(vp, nodes);
 					DOM.append(vp, this.addFooter());
-				}, this)).catch(LuCI.prototype.error);
+				}, this))
+				.then(LuCI.prototype.bind(this.afterRender, this))
+				.catch(LuCI.prototype.error);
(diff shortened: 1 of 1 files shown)


jow-:
I don't think this is needed. You can try using `requestAnimationFrame()` in your render code to execute a function after the next repaint, at which point the dimensions should be final.

It also does not seem to affect non-material themes.

Ansuel:
Thx for the suggestion!

PR #4607, opened 2020-11-20
fix the link on clicking hostname

roeidalm:


jamesmacwhite:
Nice spot. Shouldn't this just be `/` the hash wouldn't be needed as an absolute root URL.

roeidalm:
> Nice spot. Shouldn't this just be `/` the hash wouldn't be needed as an absolute root URL.

yes you right, I will change it and pr again

PR #4612, opened 2020-11-21
fix the link on clicking hostname change it to '/'

roeidalm:


hnyman:
No sign-off, no explanation :-(

PR #4617, opened 2020-11-23
Overhaul bootstrap theme (cascade.css)

vituperative:
* Separate dropdown menu items with dividers and enhance hover states for better legibility
* Responsive layout - increase maximum width of content to 1600px
* Treat table divs as discrete tables and style accordingly
* Add custom css for radio and checkbox icons to ensure better theme integration
* Various layout modifications to make better use of available screen space

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -2389,6 +2389,20 @@ label {
   background: linear-gradient(to bottom, #f6f6f6, #eaeaea) !important;
 }
 
+div[id^="cbi-firewall"].tr.cbi-section-table-row {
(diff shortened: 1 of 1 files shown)


tapper82:
Hi post up some screen shots. Have you tested this on master? There has bin some changes in master.

vituperative:
Not yet tested on the master branch. Here's a screenshot of the software section. Let me know if you need more. Thanks.
software

Ansuel:
(if you really want this be accepted with so much change, I advice you to actually describe the changes and what you fixed/improved. Also no commit description and no signed-off tag. Also i would suggest to set this a draft or add some type of WIP tag.)

jow-:
Thank you for your effort. I quickly tested this on my gateway and overall it appears to work. The changes are rather extensive though so I am somewhat opposed to add them to the existing bootstrap theme.

One thing I dislike is that the fonts appear to be mostly bold now, another is the gradient backgrounds in various boxes which makes the entire appearance rather ... blocky. It also seems as if there's even more white space now - some people complained that the current bootstrap theme already has too much of it.

We could consider merging this as new theme (e.g. bootstrap2020) if you're willing to maintain it for the forseeable future.

I do like the help icon and the checkbox styling.

Things I noticed:

1. A lot of bold fonts: 
image

2. Text-shadows tied to hardcoded colors: 
image \
The colors are pseudorandomly generated from the interface names, so a generic CSS styling is needed here which works with all colors

3. Hardcoded indicator icons: 
image \
There might be more indicators than just the poll stop/start toggle, so the CSS selector needs to be more specific, e.g. match on `[data-indicator="poll-status"]`

4. Dynamic list buttons are not aligned: 
image

5. Rich dropdown arrow misplaced / wrong color:
image

6. Active (spinning) buttons larger than default ones:
image
 vs. 
image

7. Missing space between checkbox and caption in rich dropdowns:
image

8. Truncated heading in modal overlays:

tapper82:
If you make a new theme then I suggest the name rebootstrap.

vituperative:
Note to testers: the theme relies on having a multi-weight font installed (open sans, ubuntu sans, or segoe ui currently supported), otherwise most text will appear bold. Install the Open Sans font family, either from your OS repo, or via google, for best results: https://fonts.google.com/specimen/Open+Sans

hnyman:
I find the normal alternating shades of gray in table rows much more clear and easy to read than these "similar gradients in each row" here. Much easier to keep eye on the correct row on wide tables.

vituperative:
hnyman wrote:
> I find the normal alternating shades of gray in table rows much more clear and easy to read than these "similar gradients in each row" here. Much easier to keep eye on the correct row on wide tables.

I agree that the wider tables need more work to differentiate the table rows, and the presentation of the content could also use some work to make more efficient use of available screen space. To address the table row issue, I'm thinking of adding support for an optional override.css file which, if present, allows the user to add their own css for persistent customizations. On my todo list is an assessment of how to improve the layout of content that doesn't necessarily merit the full table treatment.

systemcrash:
@vituperative would you like to address the comments? Perhaps rename the theme to make a new one?

PR #4621, opened 2020-11-27
luci-mod-system: limit the password charset range

ptpt52:


jow-:
Why?

ptpt52:
somewhat ime input code as password，they cannot be type in directly via
keyboard


Jo-Philipp Wich <notifications@github.com> 于 2020年11月27日周五 21:26写道：

PR #4623, opened 2020-11-27
luci-app-adblock: update dns report page

dibdot:
* add search anchor to all listed domains on dns report page
* small wording changes & cosmetics
* update translations


applications/luci-app-adblock/htdocs/luci-static/resources/view/adblock/dnsreport.js
```diff
@@ -227,13 +228,13 @@ return view.extend({
 				b_cnt = content.data.top_domains[i].count;
 			}
 			if (content.data.top_domains[i]) {
-				b_addr = content.data.top_domains[i].address;
+				b_addr = '<a href="https://duckduckgo.com/?q=' + content.data.top_domains[i].address + '&k1=-1&km=l&kh=1" target="_blank" title="Search this domain">' + content.data.top_domains[i].address + '</a>';
```
jow-, line 231: Hmm, I suggest to wrap the `content.data...` value into `encodeURIComponent()` to be on the safe side, even if we don't expect those fields to contain special characters.

Same for the other two instances below.

dibdot: Thanks, done. Although I am quite sure that there will be no more special characters in this handmade json ... ;-))

PR #4629, opened 2020-11-30
luci-app-openvpn: update encryption values

swg0101:
Update the list of encryption values to reflect sensible/available values
in OpenVPN 2.5.0. Certain values are long deprecated and/or insecure and
should not be made available in LuCI.


feckert:
Are these insecure **ciphers** and **tls_cipher** already switched off in openvpn?
So that they cannot be used anymore?

I would suggest that we mark them with `strong|medium|weak` security here in the luci-app-openvpn if we could still use this values.
In the WLAN area we also can select and use WEP but mark it with `weak security`.
https://github.com/openwrt/luci/blob/openwrt-19.07/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L1302

swg0101:
Those values are pulled directly from openvpn when it is compiled against mbedTLS. Other values can be specified, but they won't work since the underlying library that it is compiled against will not take these values. Using OpenSSL allows more encryption modes to be used, although only AES/CHACHA20-POLY1305 is on the official supported list.

Support for all other legacy ciphers are to be removed completely from the code base starting in 2.6:
`Using BF-CBC is no longer recommended, because of its 64-bit block size. This small block size allows attacks based on collisions, as demonstrated by SWEET32. See https://community.openvpn.net/openvpn/wiki/SWEET32 for details. Due to this, support for BF-CBC, DES, CAST5, IDEA and RC2 ciphers will be removed in OpenVPN 2.6.`

One thing that is different about this particular list is that a user can type in a custom value in the field, even if the value is not in the default list, whereas typing a custom value under the Wireless Security page is not possible. I think this is a good compromise that gives the users safe defaults while allowing them to use something like BF-CBC if they typed in the value (or "null" to disable encryption, which is not currently on the list).

You can see here that the patches also are security-focused, disabling ciphers like DES:
https://github.com/openwrt/openwrt/tree/openwrt-19.07/package/network/services/openvpn/patches

feckert:
@swg0101 As I see it, there are also other adjustments to be made in order to[ support all new options](https://github.com/openwrt/packages/commit/2e55fc8b2d42682cd1c26e9827b7b6f47fb51398#diff-74200d1e2891d8f8d5e5810628c985ce9c8e0316dc52b8e296b90de76356cbc4) that are supported in the LuCI since version 2.5.0.

Could we also add this new option with this pullrequest, If then we should do it right?

swg0101:
Yeah, I noticed today that as I was playing around with it today a lot of the options are either out of date or not even in LuCI to begin with.. this will probably need a bit of time to sort out. I will take a stab at it when I have some time...

systemcrash:
@swg0101 did you get around to that?

> Yeah, I noticed today that as I was playing around with it today a lot of the options are either out of date or not even in LuCI to begin with.. this will probably need a bit of time to sort out. I will take a stab at it when I have some time...

Or can we close this PR?

swg0101:
Hi,

Unfortunately, I didn't really have the time and ended up doing a workaround instead. When I get the chance I will try to get to this again, but will close this for now.

PR #4637, opened 2020-12-04
luci-mod-network: wireless.js Disable 802.11b by default

nicklowe:
The time has long come where 802.11b DSSS/CCK data rates should be disabled
by default in OpenWRT. Users in need of 802.11b client support can reasonably
enable these where they are needed.

The balance of equities has significantly, and for a long time, tipped
such that dropping backwards compatibility by default with 802.11b
devices is appropriate, proportionate and justified. By doing so,
management and control traffic is moved by default to a 20
MHz wide 6 Mb/s OFDM data rate instead of a 22 MHz wide 1 Mb/s data rate.
This is significantly more airtime efficient.

Signed-off-by: Nick Lowe nick.lowe@gmail.com

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -883,6 +883,8 @@ return view.extend({
 				o.ucisection = s.section;
 
 				if (hwtype == 'mac80211') {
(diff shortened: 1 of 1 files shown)


nicklowe:
Taking in to account David Bauer's legitimate concerns in the PR that this depends on, I have updated the proposed LuCI patch this morning for discoverability:

Moving the option from the Advanced Settings to the General Setup tab.
Adding a description explaining potential compatibility implications. This states:

**Poorly behaving 802.11g or 802.11n devices may require legacy 802.11b rates to interoperate. Airtime efficiency may be significantly reduced where these are used. It is recommended to disable this option where possible.**

aparcar (approved):

nicklowe:
Received feedback suggesting to change in the wording "poorly behaving" to "badly behaving" to be clearer for those with a foreign language to English. Change made.

PR #4646, opened 2020-12-07
rpcd-mod-*: improve postinst script

Cynerd:
Usage of killall is replaced with init script. This is cleaner solution
as it does not consider some implementation detail but rather passes
that on to init script implementation.

IPKG_INSTROOT was added to prevent execution when not running in current
root. It is invalid to request reload if install-root is not current
root. In this case it can be considered harmless but it is invalid
nonetheless.

Last change is removal of `exit 0`. This caused skip of default
postinst. Execution of default postinst does no harm and is more
standard considering possible future expansion of it.

### DIFF
--- a/libs/rpcd-mod-luci/Makefile
+++ b/libs/rpcd-mod-luci/Makefile
@@ -48,8 +48,7 @@ endef
 
 define Package/rpcd-mod-luci/postinst
 #!/bin/sh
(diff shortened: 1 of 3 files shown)


jow-:
Merged, thanks.

BKPepe:
@jow- Would it be possible to include it in OpenWrt 19.07 branch?

PR #4649, opened 2020-12-09
luci-mod-network Add description: Hidden SSID, WMM

nicklowe:
luci-mod-network Describe issues: Hidden SSID, WMM

The performance and reliability implications of hiding SSIDs or disabling
WMM are not commonly known and these settings often end up being
misconfigured to harmful effect.

To seek to mitigate this, add descriptions explaining that:

Where the ESSID is hidden, clients may fail to roam and airtime efficiency
may be significantly reduced.

Where Wi-Fi Multimedia (WMM) Mode QoS is disabled, clients may be limited
to 802.11a/802.11g rates.

Signed-off-by: Nick Lowe <nick.lowe@gmail.com>


feckert:
Thanks

PR #4651, opened 2020-12-11
luci-app-throughput-measurements: add iperf3 app

wjowsa:
In some troubleshooting situations (e.g. testing network performance) it is useful to run network troubleshooting tools like iperf3, directly from the Luci interface. This PR introduces a Throughput Measurements application. 
Throughput Measurements application is a fronted for iperf3. With this application, a user is able to configure iperrf3 as a server or client with the most important and commonly used configuration options. The output of the iperf3 is lived streamed to the browser view. 
~~The application uses the new uhttpd API. The EventSource interface is used to handle server-sent events. However, in order to make the API work properly, there is one uhttpd patch be applied~~ 

 ~~* Because the EventSource does not allow to set custom header fields, the uhttpd has to be patched to support passing rpcd session id in the subscription url. [patch 1](https://patchwork.ozlabs.org/project/openwrt/patch/20201110080209.6368-1-wojciech.jowsa@gmail.com/)~~

I have tested this application on TP-Link WDR4300 and PCEngines APU3 using the latest openwrt master.

client
server

### DIFF
--- a/applications/luci-app-throughput-measurements/Makefile
+++ b/applications/luci-app-throughput-measurements/Makefile
@@ -1,0 +1,13 @@
+# Copyright 2020 Wojciech Jowsa (wojciech.jowsa@gmail.com)
+# This is free software, licensed under the Apache License, Version 2.0
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Throughput measurements application
+LUCI_DEPENDS:=+luci-mod-admin-full +iperf3
+PKG_MAINTAINER:=Wojciech Jowsa <wojciech.jowsa@gmail.com>
+PKG_LICENSE:=Apache-2.0
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/client.js
+++ b/applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/client.js
@@ -1,0 +1,240 @@
+'use strict';
+'require rpc';
+'require uci';
+'require ui';
+'require fs';
+'require form';
+'require network';
+'require tools.widgets as widgets';
+
+var eventSource,
+	hideStart = false;
+
+var callIperf3Start = rpc.declare({
+	object: 'luci.tputmeas',
+	method: 'iperf3Start',
+	params:  [ "mode", "server_address", "interface", "interface_address", "port",
+			   "duration", "bytes", "udp", "bitrate", "reverse", "bidir", "connections", "interval" ],
+	expect: { '': {} }
+});
+
+var callIperf3Stop =  rpc.declare({
+	object: 'luci.tputmeas',
+	method: 'iperf3Stop',
+	expect: { '': {} }
+});
+
+function stopIperf3() {
+	callIperf3Stop().then(function(replay) {
+		if (eventSource)
+			eventSource.close();
+	}.bind(this)).catch(function(error) {
+	});
+}
+
+window.addEventListener('beforeunload', stopIperf3);
+
+function addOutput() {
+	var iperf3Out = document.querySelectorAll('[id$="iperf3_out"]')[0];
+	if (iperf3Out)
+		return;
+
+	var frameEl = E('div', {'class': 'cbi-value'});
+
+	frameEl.appendChild(E('textarea', {
+			'id': 'iperf3_out',
+			'class': 'cbi-input-textarea',
+			'readonly': '',
+			'style': 'width:100%',
+			'rows': 30,
+	}));
+
+	frameEl.firstElementChild.style.fontFamily = 'monospace';
+
+	var stopBtn = document.querySelectorAll('[id$="stop_iperf3"]')[0];
+	if (stopBtn)
+		stopBtn.parentNode.insertBefore(frameEl, stopBtn.nextSibling);
+}
+
+function subscribeIperf3() {
+	if (eventSource)
+		eventSource.close();
+
+	eventSource = new EventSource('/ubus/subscribe/luci.tputmeas.notify')
+	eventSource.onerror = function(event) {
+		console.log(event);
+		eventSource.close();
+	};
+
+	addOutput();
+	var textOut = document.querySelectorAll('[id$="iperf3_out"]')[0];
+	textOut.value = ""
+	eventSource.addEventListener("luci.tputmeas.notify.data", function(event) {
+		textOut.value = textOut.value + "\n" + JSON.parse(event.data).data;
+	});
+};
+
+function updateButtons() {
+	var tasks = [];
+	tasks.push(fs.stat("/var/run/luci-iperf3.pid").then(L.bind(function(list) {
+			if (!eventSource || eventSource.readyState == 2)
+				subscribeIperf3();
+			var textOut = document.querySelectorAll('[id$="iperf3_out"]')[0];
+			if (textOut)
+				textOut.style.borderColor = "green";
+			var startBtn = document.querySelectorAll('[id$="start_iperf3"]')[0];
+			if (startBtn)
+				startBtn.hidden = true;
+			var stopBtn = document.querySelectorAll('[id$="stop_iperf3"]')[0];
+			if (stopBtn)
+				stopBtn.hidden = false;
+			return;
+		})).catch(function(error) {
+			var textOut = document.querySelectorAll('[id$="iperf3_out"]')[0];
+			if (textOut)
+				textOut.style.borderColor = "red";
+			var startBtn = document.querySelectorAll('[id$="start_iperf3"]')[0];
+			if (startBtn)
+				startBtn.hidden = hideStart;
+			var stopBtn = document.querySelectorAll('[id$="stop_iperf3"]')[0];
+			if (stopBtn)
+				stopBtn.hidden = true;
+			if (eventSource)
+				eventSource.close();
+	}));
+
+	return  Promise.all(tasks);
+}
+
+return L.view.extend({
+
+	render: function(processes) {
+		var m, s, o;
+
+		m = new form.Map('luci_tputmeas', _('Throughput Measurements - Iperf3 Client'), _('Iperf3 is a tool for performing network throughput measurements. It can test TCP, UDP, or SCTP throughput. To perform an iperf3 test the user must establish both a server and a client.'));
+		s = m.section(form.TypedSection, 'iperf3')
+		s.anonymous = 1;
+
+		o = s.option(form.Value, 'server_address', _('Server address'), _('Address of the iperf3 server'));
+		o.optional = false;
+		o.datatype = 'ip4addr("nomask")';
+
+		o = s.option(widgets.DeviceSelect, 'interface', _('Interface'), _('Bind to the specific interface associated with address host.'))
+		o.noaliases = true;
+
+		o = s.option(form.Value, 'interface_address', _('Interface address'), _('Bind to the specific interface address associated with address host.'));
+		o.optional = false;
+		o.datatype = 'ip4addr("nomask")';
+		o.depends('interface', undefined);
+
+		o = s.option(form.Value, 'port', _('Port'), _('Set server port to listen on/connect to to n (default 5201)'));
+		o.modalonly = false;
+		o.datatype = 'list(neg(port))';
+		o.default = '5201'
+
+		o = s.option(form.Value, 'duration', _('Duration'), _('Time in seconds to transmit for (default 10 secs)'));
+		o.modalonly = false;
+		o.datatype = 'range(1,4294967296)'
+		o.validate =function(section_id, value) {
+			var bytes_input = document.querySelectorAll('[id$="bytes"]'),
+				duration_input = document.querySelectorAll('[id$="duration"]');
+
+			if (bytes_input[2] && bytes_input[2].value && duration_input[2].value){
+				hideStart = true;
+				return _("Only one test end condition (duration or bytes) may be specified")
+			}
+			hideStart = false;
+
+			return true;
+		}
+
+		o = s.option(form.Value, 'bytes', _('Bytes'), _('Number of bytes to transmit (instead of duration)'));
+		o.modalonly = false;
+		o.validate = function(section_id, value) {
+			var bytes_input = document.querySelectorAll('[id$="bytes"]'),
+				duration_input = document.querySelectorAll('[id$="duration"]');
+
+			if (duration_input[2] && duration_input[2].value && bytes_input[2].value) {
+				hideStart = true;
+				return _("Only one test end condition (duration or bytes) may be specified")
+			}
+			hideStart = false;
+
+			return true;
+		}
+
+		o = s.option(form.Value, 'bitrate', _('Bitrate'), _('Set target bitrate to n bits/sec (default 1 Mbit/sec for UDP, unlimited for TCP/SCTP).'));
+		o.modalonly = false;
+
+		o = s.option(form.Value, 'connections', _('Connections'), _('Number of parallel client streams to run. '));
+		o.modalonly = false;
+		o.datatype = 'range(1,4294967296)';
+
+		o = s.option(form.Value, 'interval', _('Interval'), _('Pause n seconds between periodic throughput reports; default is 1, use 0 to disable'));
+		o.modalonly = false;
+		o.datatype = 'range(1,4294967296)';
+
+		o = s.option(form.Flag, 'udp', _('Udp'), _("Use UDP rather than TCP"));
+
+		o = s.option(form.Flag, 'reverse', _('Reverse'), _("Reverse the direction of a test, so that the server sends data to the client"));
+
+		o = s.option(form.Flag, 'bidir', _('Bidirectional'), _("Run in bidirectional mode"))
+
+		o = s.option(form.Button, 'start_iperf3', _('Start iperf3'), _(''));
+		o.inputstyle = 'apply';
+		o.onclick = ui.createHandlerFn(this, function(section_id, ev) {
+			var iface = document.querySelectorAll('[id$="interface"]')[1].value ? document.querySelectorAll('[id$="interface"]')[1].value : null,
+				port = document.querySelectorAll('[id$="port"]')[2].value ?  document.querySelectorAll('[id$="port"]')[2].value : null,
+				server_address = document.querySelectorAll('[id$="server_address"]')[2].value ? document.querySelectorAll('[id$="server_address"]')[2].value : null,
+				interface_address = document.querySelectorAll('[id$="interface_address"]')[2].value ? document.querySelectorAll('[id$="interface_address"]')[2].value : null,
+				duration = document.querySelectorAll('[id$="duration"]')[2].value ? document.querySelectorAll('[id$="duration"]')[2].value : null,
+				bytes = document.querySelectorAll('[id$="bytes"]')[2].value ? document.querySelectorAll('[id$="bytes"]')[2].value : null,
+				bitrate = document.querySelectorAll('[id$="bitrate"]')[2].value ? document.querySelectorAll('[id$="bitrate"]')[2].value : null,
+				connections = document.querySelectorAll('[id$="connections"]')[2].value ? document.querySelectorAll('[id$="connections"]')[2].value : null,
+				interval = document.querySelectorAll('[id$="interval"]')[2].value ? document.querySelectorAll('[id$="interval"]')[2].value : null,
+				udp = document.querySelectorAll('[data-widget-id$="udp"]')[0].checked ? true : false,
+				reverse = document.querySelectorAll('[data-widget-id$="reverse"]')[0].checked ? true : false,
+				bidir = document.querySelectorAll('[data-widget-id$="bidir"]')[0].checked ? true : false;
+
+			return callIperf3Start("-c", server_address, iface, interface_address, port,
+									duration, bytes, udp, bitrate, reverse, bidir, connections, interval).then(function(replay) {
+				if (replay.error){
+					ui.showModal(_(replay.error), [
+						E('div', { 'class': 'right' }, [
+							E('button', {
+								'class': 'cbi-button cbi-button-negative important',
+								'click': function(ev) {
+									ui.hideModal();
+								}
+							}, _('Close')),
+						])
+					]);
+					return;
+				}
+				rpc.list.apply(rpc).then(function(res) {
+					for (var k in res) {
+						if (res[k] == "luci.tputmeas.notify")
+							subscribeIperf3()
+					}
+				}.bind(this))}.bind(this)).catch(function(error) {
+					console.log(error);
+				});
+		});
+
+		o = s.option(form.Button, 'stop_iperf3', _('Stop iperf3'), _(''));
+		o.inputstyle = 'apply';
+		o.onclick =  ui.createHandlerFn(this, function(section_id, ev) {
+			if (!eventSource)
+				return;
+			return callIperf3Stop().then(function(replay) {
+				eventSource.close();
+			}.bind(this)).catch(function(error) {
+				console.log(error);
+			});
+		});
+
+		L.Poll.add(L.bind(updateButtons, m),1);
+
+		return m.render();
+	},
+});
--- a/applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/server.js
+++ b/applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/server.js
@@ -1,0 +1,177 @@
+'use strict';
+'require rpc';
+'require uci';
+'require ui';
+'require fs';
+'require form';
+'require network';
+'require tools.widgets as widgets';
+
+var eventSource;
+
+var callIperf3Start = rpc.declare({
+	object: 'luci.tputmeas',
+	method: 'iperf3Start',
+	params:  [ "mode", "server_address", "interface", "interface_address", "port",
+			   "duration", "bytes", "udp", "bitrate", "reverse", "connections", "interval" ],
+	expect: { '': {} }
+});
+var callIperf3Stop =  rpc.declare({
+	object: 'luci.tputmeas',
+	method: 'iperf3Stop',
+	expect: { '': {} }
+});
+
+function stopIperf3() {
+	callIperf3Stop().then(function(replay) {
+		if (eventSource)
+			eventSource.close();
+	}.bind(this)).catch(function(error) {
+		console.log(error);
+	});
+}
+
+window.addEventListener('beforeunload', stopIperf3);
+
+function addOutput() {
+	var iperf3Out = document.querySelectorAll('[id$="iperf3_out"]')[0];
+	if (iperf3Out)
+		return;
+
+	var frameEl = E('div', {'class': 'cbi-value'});
+
+	frameEl.appendChild(E('textarea', {
+			'id': 'iperf3_out',
+			'class': 'cbi-input-textarea',
+			'readonly': '',
+			'style': 'width:100%',
+			'rows': 30,
+	}));
+
+	frameEl.firstElementChild.style.fontFamily = 'monospace';
+	var stopBtn = document.querySelectorAll('[id$="stop_iperf3"]')[0];
+	if (stopBtn)
+		stopBtn.parentNode.insertBefore(frameEl, stopBtn.nextSibling);
+}
+
+function subscribeIperf3() {
+	if (eventSource)
+		eventSource.close();
+
+	eventSource = new EventSource('/ubus/subscribe/luci.tputmeas.notify');
+	eventSource.onerror = function(event) {
+		eventSource.close();
+	};
+
+	addOutput();
+	var textOut = document.querySelectorAll('[id$="iperf3_out"]')[0];
+	textOut.value = "";
+	eventSource.addEventListener("luci.tputmeas.notify.data", function(event) {
+		textOut.value = textOut.value + "\n" + JSON.parse(event.data).data;
+	});
+};
+
+function updateButtons() {
+	var tasks = [];
+	tasks.push(fs.stat("/var/run/luci-iperf3.pid").then(L.bind(function(list) {
+			if (!eventSource || eventSource.readyState == 2)
+				subscribeIperf3();
+			var textOut = document.querySelectorAll('[id$="iperf3_out"]')[0];
+			if (textOut)
+				textOut.style.borderColor = "green";
+			var startBtn = document.querySelectorAll('[id$="start_iperf3"]')[0];
+			if (startBtn)
+				startBtn.hidden = true;
+			var stopBtn = document.querySelectorAll('[id$="stop_iperf3"]')[0];
+			if (stopBtn)
+				stopBtn.hidden = false;
+			return;
+		})).catch(function(error) {
+			var textOut = document.querySelectorAll('[id$="iperf3_out"]')[0];
+			if (textOut)
+				textOut.style.borderColor = "red";
+			var startBtn = document.querySelectorAll('[id$="start_iperf3"]')[0];
+			if (startBtn)
+				startBtn.hidden = false;
+			var stopBtn = document.querySelectorAll('[id$="stop_iperf3"]')[0];
+			if (stopBtn)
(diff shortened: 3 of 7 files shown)


openwrt-diy:
@wjowsa This plug-in is enabled at boot time? Can you consider adding a switch option

wjowsa:
@openwrt-diy It works like a regular Luci application. If you add it to your custom image or download a package then it will be available in the Luci UI. In order to start measurements you have to press start button. There is not auto-start on boot option, if this is what you meant?

feckert:
Since this is a ubus for the luci, I would suggest from my point of view that we also provide it with **luci** prefix.
Just like other luci apps do. This groups the objects better.
Or is this so general, that we can then move it into the iperf3 package?

wjowsa:
Sure, it is luci related so I will add a luci prefix.

feckert:
I do not think I have expressed myself correctly. I rather meant that you name the [file](https://github.com/openwrt/luci/pull/4651/files#diff-6f8682c8db8402f5aefd78758326461ddd4fd93b1d3358ec05c495828a169884) like this `luci.tputmeas`

Like the file in https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns and move this into rpcd. I haven't looked closely now but you seem to be creating a native ubus call with lua, right?
Wouldn't it be better with the rpcd?

wjowsa:
Yes, you are right. The current implementation is a bit "old style" i.e. lua script is called directly. It is because of my laziness ;) 
I agree that the preferred way is with rpcd so I have no arguments for not using it.

wjowsa:
@feckert done.

applications/luci-app-throughput-measurements/root/usr/share/rpcd/acl.d/luci-app-throughput-measurements.json
```diff
@@ -0,0 +1,17 @@
...
+			"cgi-io": [ "exec" ],
+			"uci": [ "tputmeas" ],
+			"ubus": {
+				"luci.tputmeas": [ "iperf3Start", "iperf3Stop" ],
+				"luci.tputmeas.notify": [ "*" ]
+			}
+		},
+		"write": {
+			"uci": [ "tputmeas" ],
+			"file": {"/usr/libexec/luci.tputmeas": ["exec"]}
```
feckert, line 14: I think this must be a write ubus? or not

applications/luci-app-throughput-measurements/root/usr/share/luci/menu.d/luci-app-throughput-measurements.json
```diff
@@ -0,0 +1,41 @@
...
+		"title": "Throughput Measurements",
+		"order": 1,
+		"action": {
+			"type": "alias",
+			"path": "admin/troubleshooting/tputmeas/client"
+		},
+		"depends" : {
+			"acl": [ "luci-app-throughput-measurements" ],
+			"uci": { "tputmeas": true },
+			"fs": { "/usr/libexec/luci.tputmeas": "executable" }
```
feckert, line 20: I can not find a rpcd in the directory

applications/luci-app-throughput-measurements/root/usr/share/rpcd/acl.d/luci-app-throughput-measurements.json
```diff
@@ -0,0 +1,17 @@
...
+		}
+	}
+}
```
feckert, line 17: please add a new line

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,225 @@
...
+	local _, method, rv = nil, nil, {}
+	for _, method in pairs(methods) do rv[_] = method.args or {} end
+	print((json.stringify(rv):gsub(":%[%]", ":{}")))
+elseif arg[1] == "call" then
+	local args = parseInput()
+	local method = validateArgs(arg[2], args)
+	local result, code = method.call(args)
+	print((json.stringify(result):gsub("^%[%]$", "{}")))
+	os.exit(code or 0)
+end
```
feckert, line 225: please add a new line

applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/server.js
```diff
@@ -0,0 +1,180 @@
...
+		stopBtn.parentNode.insertBefore(frameEl, stopBtn.nextSibling);
+}
+
+function subscribeIperf3() {
+	if (eventSource)
+		eventSource.close();
+
+	eventSource = new EventSource('/ubus/subscribe/luci.tputmeas.notify' + '?' + rpc.getSessionID());
+	eventSource.onerror = function(event) {
+		console.log(event);
```
feckert, line 63: Can you please remove this log?

applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/server.js
```diff
@@ -0,0 +1,180 @@
...
+
+		o = s.option(form.Button, 'stop_iperf3', _('Stop iperf3'), _(''));
+		o.inputstyle = 'apply';
+		o.onclick =  ui.createHandlerFn(this, function(section_id, ev) {
+			if (!eventSource)
+				return;
+			return callIperf3Stop().then(function(replay) {
+				eventSource.close();
+			}.bind(this)).catch(function(error) {
+				console.log(error);
```
feckert, line 172: Do we need this log

applications/luci-app-throughput-measurements/root/etc/config/tputmeas
```diff
@@ -0,0 +1 @@
+config iperf3
```
feckert, line 1: The configuration belongs to the luci, doesn't it? Wouldn't it then be better to identify it? The configuration only works with the luci. Can you then also list all the configurations.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,225 @@
...
+
+local ubus = require "ubus"
+local uci = require "luci.model.uci"
+local fs = require "nixio.fs"
+local json = require "luci.jsonc"
+local sys = require "luci.sys"
+local posix = require "posix"
+
+
+local methods = {
```
feckert, line 11: I am not sure whether it is right to always use the shell like grep, egrep and ip.
Can't we replace some things with native lua?

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,225 @@
...
+				end
+			end
+
+			iperf3_cmd = iperf3_cmd .. " 2>&1"
+
+			local pid = posix.fork()
+			if pid == 0 then
+				local cpid = posix.getpid()
+				fs.writefile("/var/run/iperf3.pid",cpid)
+				fs.writefile("/var/run/iperf3.pid1",iperf3_cmd)
```
feckert, line 122: The pid1 is also never cleaned up or

applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/client.js
```diff
@@ -0,0 +1,241 @@
...
+	method: 'iperf3Stop',
+	expect: { '': {} }
+});
+
+function stopIperf3() {
+	callIperf3Stop().then(function(replay) {
+		if (eventSource)
+			eventSource.close();
+	}.bind(this)).catch(function(error) {
+		console.log(error);
```
feckert, line 32: console.log?

applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/client.js
```diff
@@ -0,0 +1,241 @@
...
+			if (eventSource)
+				eventSource.close();
+	}));
+
+	return  Promise.all(tasks);
+}
+
+return L.view.extend({
+
+	load: function() {},
```
feckert, line 112: I think we could remove this line?

applications/luci-app-throughput-measurements/root/usr/share/rpcd/acl.d/luci-app-throughput-measurements.json
wjowsa: It is not needed, I think

applications/luci-app-throughput-measurements/root/usr/share/luci/menu.d/luci-app-throughput-measurements.json
wjowsa: Forgot to update path with rpcd directory

applications/luci-app-throughput-measurements/root/usr/share/rpcd/acl.d/luci-app-throughput-measurements.json
wjowsa: ok

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
wjowsa: ok

applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/server.js
wjowsa: ok

applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/server.js
wjowsa: No, exception will be visible in console anyway

applications/luci-app-throughput-measurements/root/etc/config/tputmeas
wjowsa: ok

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
wjowsa: removed

applications/luci-app-throughput-measurements/htdocs/luci-static/resources/view/tputmeas/client.js
wjowsa: removed

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
wjowsa: ok

feckert:
I have now taken a closer look at your things, how this works together.
@wjowsa I'm not entirely happy with whole.
But that is just my personal opinion.
I would implement the whole thing differently.

Can't we move the server stuff into the iperf3 package?
The way I see it, these configurations are supposed to be stored.
We could then start the server at bootup.
That would simplify the handling in the rpcd so we do only handle the client stuff.
We could show the ouput of the enabled server like we do this with the log and the dmesg.
I have staged my procd for iperf3 [here](https://github.com/TDT-AG/openwrt/tree/pr/20201215-iperf3).

Client things do not have to be stored in the uci.
This does not have to be persistent i think.
It is only needed and executed once per run. 

Of course, I am not sure if this is what the upstream maintainer in openwrt of iperf3 packages wants.
Lets ping him to get his @pprindeville  opinion.
I am only speaking from my own point of view.

wjowsa:
@feckert  My intention was to create an application that works on demand so that users can easily start/stop iperf3 when it is needed.
In my opinion, this design works better in e.g. dynamic environments or with devices with multiple listening interfaces (modems, wifis etc). Users can select or change interface/address from one place and see the live output from the iperf3 without the need for changing or refreshing the view. The uci configuration file can be used for storing configuration options of iperf3 so that users don't have to fill configuration every time there is a need for repeating the test.  This might be also a good starting point for implementing the "on boot" functionality (+ procd script). 
As I understand you propose something else i.e. when iperf3 server is enabled then it is started on boot. The output of the server is redirected to file? Luci reads the content of the file and displays it on demand (refresh button)? The client works more or less like with the current implementation.

feckert:
> @feckert My intention was to create an application that works on demand so that users can easily start/stop iperf3 when it is needed.

We can do the same with the procd.

> As I understand you propose something else i.e. when iperf3 server is enabled then it is started on boot. The output of the server is redirected to file? Luci reads the content of the file and displays it on demand (refresh button)? The client works more or less like with the current implementation.

Yes that is what I meant
I think the handling in the procd is much easier and cleaner than if we do that in the ubus rpcd with a fork.
The code must also be maintained.

We can configure things in the LuCI and then start them dynamically as well.
This is applicable for the server as well as for the client.
Server mode could also be started on demand with a button. This is also done in the luci-app-openvpn.
The procd init script could be smart, so that it also dynamically uses addresses that are currently being used on the interface in client mode.

I need to take a much closer look.
I will now implement a client call in the procd, then you can look at it in my branch. 

It may also be that I am wrong and your approach is the right one.

Give me some more time to evaluate it.

pprindeville:
> Can't we move the server stuff into the iperf3 package?

@feckert Not following.  Which part is that?

If you want to do UCI support as a subordinate package for `iperf3` I'm okay with that.  Submit a PR and I'll look at it.

feckert:
> > Can't we move the server stuff into the iperf3 package?
> 
> @feckert Not following. Which part is that?
> 
> If you want to do UCI support as a subordinate package for `iperf3` I'm okay with that. Submit a PR and I'll look at it.

@pprindeville Actually, this is about the new luci-app-throughput-measurements that @wjowsa has staged her with this pullrequest.

But I am not completely satisfied with the fact that this is implemented via an extra ubus call was an fork.
I would have suggested that we outsource the handling of iperf3 to the procd and this app only handels the uci configuration show the results and could start and stop the iperf3 command.

I wanted to hear your opinion on whether you think it's OK for you, that I add uci handling to the iperf3 for the client and server, because you are the maintainer of this package.

pprindeville:
> I would have suggested that we outsource the handling of iperf3 to the procd and this app only handels the uci configuration show the results and could start and stop the iperf3 command.

@feckert I concur: I don't want to assume that the only way that this will be used is via Luci.  I prefer having command-line support, so one could (for instance) run regularly scheduled tests via `crond` if desired, etc.

feckert:
@pprindeville If I implement the client side as an additional procd [command](ttps://github.com/TDT-AG/openwrt/commit/64ed589674fc0a1690c1a35db95ab2861a1f2ca1) for iperf3, this should not be a problem to call this also with crond.

pprindeville:
> If I implement the client side as an additional procd command for iperf3, this should not be a problem to call this also with crond.

@feckert Works for me.

stangri:
This looks amazing, great job! A small question -- would it not be better to call this luci-app-iperf3 if it's just the frontend for the iperf3?

wjowsa:
@stangri Thanks. The idea behind the name is that it is more informative for users. It also gives some space for future extensions that are not iperf3 related (statistics, charts, etc). However, I am not really bound to this name so let's see what others will say.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+local methods = {
+	iperf3Start = {
+		args = { mode = "mode", server_address = "server_address", interface = "interface",
+				interface_address = "interface_address", port = "port", duration = "duration",
+				bytes = "bytes", udp = "udp",  bitrate = "bitrate", reverse = "reverse",
+				connections = "connections", interval = "interval" },
+		call = function(args)
+			local iperf3_cmd = "iperf3 --forceflush "
+
+			if fs.stat("/var/run/iperf3.pid") then
```
jow-, line 21: I suggest to drop this pid file check and execute iper3 under [flock(1)](https://linux.die.net/man/1/flock) instead. This is a more robust solution that is not prone to race conditions.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+				return { result = "error: Iperf3 is running" }
```
jow-, line 22: Instead of replaying `{ "result": "error: ..." }` I suggest to reply with `{ "error": "..." }`

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+			else
+				if not args then
+					return { result = "error: Lack of arguments" }
+				end
+
+				if (not args.mode or (args.mode ~= "-c" and args.mode ~= "-s")) then
+					return { result = "error: mode has not been specified " }
```
jow-, line 29: Trailing space.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+				end
+
+				iperf3_cmd = iperf3_cmd .. args.mode
+
+				if args.server_address and args.server_address ~= '' then
+					local ipparts = {args.server_address:match("^(%d+)%.(%d+)%.(%d+)%.(%d+)$")}
```
jow-, line 35: You could use [`luci.ip.new()`](http://openwrt.github.io/luci/api/modules/luci.ip.html#new), which would also allow IPv6. If you indeed need to enforce IPv4, use [`luci.ip.IPv4()`](http://openwrt.github.io/luci/api/modules/luci.ip.html#IPv4) instead.

```lua
local addr = luci.ip.new(args.server_address)
if addr then
    iperf3_cmd = iperf3_cmd .. " " .. addr:host()
else
    return { error = "Invalid server address" }
end
```

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+							ok = true
+						end
+					end
+					if not ok then
+						return { result = "error: Incorrect format of an interface address" }
+					end
+				end
+
+				if args.port and args.port ~= '' then
+					if tonumber(args.port) then
```
jow-, line 71: You need to check for > 0 and <= 65535 as well

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+						iperf3_cmd = iperf3_cmd .. " -p " .. args.port
+					else
+						return { result = "error: Incorrect port argument" }
+					end
+				end
+
+				if args.duration and args.duration ~= '' then
+					if tonumber(args.duration) then
```
jow-, line 79: You need to check for > 0

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+
+				if args.bytes and args.bytes ~= '' then
+					if args.bytes:match("^[0-9]*$") or args.bytes:match("^[0-9]*[kKmM]$") then
+						iperf3_cmd = iperf3_cmd .. " -n " .. args.bytes
+					else
+						return { result = "error: Incorrect bytes argument" }
+					end
+				end
+
+				if args.udp == "1" and args.udp ~= '' then
```
jow-, line 94: The checks are redundant. But why not using a boolean? Would simplify this into `if args.udp then`

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+					iperf3_cmd = iperf3_cmd .. " -u"
+				end
+
+				if args.bitrate and args.bitrate ~= '' then
```
jow-, line 98: The checks are redundant - either one is sufficient

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+					if args.bitrate:match("^[0-9]*$") or args.bitrate:match("^[0-9]*[kKmM]$") then
+						iperf3_cmd = iperf3_cmd .. " -b " .. args.bitrate
+					else
+						return { result = "error: Incorrect bitrate argument" }
+					end
+				end
+
+				if args.reverse == "1" and args.reverse ~= nil then
```
jow-, line 106: The checks are redundant

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+					iperf3_cmd = iperf3_cmd .. " -R"
+				end
+
+				if args.connections and args.connections ~= ''  then
```
jow-, line 110: The checks are redundant

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+					if tonumber(args.connections) then
+						iperf3_cmd = iperf3_cmd .. " -P " .. args.connections
+					else
+						return { result = "error: Incorrect connections argument" }
+					end
+				end
+
+				if args.interval and args.interval ~= '' then
```
jow-, line 118: The checks are redundant

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+						iperf3_cmd = iperf3_cmd .. " -i " .. args.interval
+					else
+						return { result = "error: Incorrect interval argument" }
+					end
+				end
+			end
+
+			iperf3_cmd = iperf3_cmd .. " 2>&1"
+
+			local pid = posix.fork()
```
jow-, line 129: You can use [`nixio.fork()`](https://neopallium.github.io/nixio/modules/nixio.html#nixio.fork)

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+			if pid == 0 then
+				local cpid = posix.getpid()
```
jow-, line 131: As written above, I suggest to drop the entire PID stuff since it is prone to race conditions.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+					os.exit(1)
+				end
+
+				local ubus_objects = {
+					["luci.tputmeas.notify"] = {
+					}
+				}
+				conn:add( ubus_objects )
+
+				sys.call("sleep 1")
```
jow-, line 146: You can use `nixio.poll({}, 1000)` instead to avoid executing a shell / external sleep here.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+
+				local pipe = io.popen(iperf3_cmd)
```
jow-, line 148: I suggest to execute the command as:
`io.popen('flock -n -x /var/lock/luci-iperf3.lock ' .. iperf3_cmd .. ' --pidfile /var/run/luci-iperf3.pid')`

If no instance is running, it will succeed. If another process exists, it will fail immediately with code 1.
Also let iperf3 itself write its pid file to avoid race conditions.

This will turn the code into:

```lua
local pipe, err = io.popen('flock -n -x /var/lock/luci-iperf3.lock ' .. iperf3_cmd)
if pipe then
    -- do the ubus stuff
else
    return { error = "Failed to execute iperf3 client: " .. err }
end
```

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+				pipe:close()
+				fs.remove("/var/run/iperf3.pid")
+			else
+				return { result = "success" }
+			end
+		end
+	},
+	iperf3Stop = {
+		call = function(args)
+			local pid = fs.readfile("/var/run/iperf3.pid")
```
jow-, line 166: The pid file might not exist, e.g. if the process was not started before.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
+#!/usr/bin/env lua
+
+local ubus = require "ubus"
+local uci = require "luci.model.uci"
+local nixio = require "nixio"
+local fs = require "nixio.fs"
+local json = require "luci.jsonc"
+local sys = require "luci.sys"
+local posix = require "posix"
```
jow-, line 9: Lua POSIX is not needed, see remarks below.

applications/luci-app-throughput-measurements/root/etc/config/tputmeas
```diff
@@ -0,0 +1,13 @@
...
+    option interface ''
+    option interface_address ''
+    option port ''
+    option duration ''
+    option bytes ''
+    option udp ''
+    option bitrate ''
+    option reverse ''
+    option connections ''
+    option interval ''
```
jow-, line 13: Please ensure that files end with a newline

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
```diff
@@ -0,0 +1,238 @@
...
+				fs.remove("/var/run/iperf3.pid")
+			else
+				return { result = "success" }
+			end
+		end
+	},
+	iperf3Stop = {
+		call = function(args)
+			local pid = fs.readfile("/var/run/iperf3.pid")
+			local plist = sys.process.list()
```
jow-, line 167: That operation is quite inefficient and expensive. You can use [`fs.readlink()`](http://openwrt.github.io/luci/api/modules/nixio.fs.html#nixio.fs.readlink) to sanity check that the process currently having this pid is indeed iperf3:

```lua
if fs.readlink("/proc/%d/exe" % pid) == "/usr/bin/iperf3" then
    nixio.kill(pid, 15)
else
    -- an unrelated process, iperf3 most likely ended itself
end
```

(Compare with with `ls -lh /proc/$(pidof iperf3)/exe` on the shell)

jow- (changes requested):

wjowsa:
@jow- I have implemented most of your review remarks. However  --pidfile is server mode only. More, when used in server mode then for some reason the first output "Server listening .. " of the iperf3 is lost. Due to that ,stopping iperf3 has to be done by checking subprocesses PIDs. I have used sys.exec and grep because I could not find better solution. Besides that everything seems to work fine.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
jow-: Note that it should be <*=* 65535

jow-:
Hm, I feared that. Then I suggest this alternative:

Change 
```lua
local pipe, err = io.popen(iperf3_cmd)
```

to

```lua
local pipe, err = io.popen("echo $$ > /var/run/luci-iperf3.pid && exec %s" % iperf3_cmd)
```

It works like this:
 - The `popen()` function passes its argument to `sh -c`, so we can use shell constructs
 - First we echo the own pid of the shell process into a pid file
 - Then we exec the shell process into the actual iperf command, keeping the current pid

Normally code like the one above would be a textbook example of a potential shell injection vulnerability but since we validate all parameters ending up in `iperf3_cmd` we can rule out that special characters like `$`, `` ` `` or `;` appear.

After you implemented this change, you can use the pid file value directly and do not need to resolve child process IDs anymore, avoiding the `pgrep` dependency.

applications/luci-app-throughput-measurements/root/usr/libexec/rpcd/luci.tputmeas
wjowsa: ok

wjowsa:
@jow- works like a charm, thanks.

pprindeville:
@jow- I'm also an upstream contributor, and Bruce Mah is very open to enhancements.  I can add this functionality, but it won't be in a release for another 2-3 months if you're okay with that...  What's the exact requirement?

jow-:
> @jow- I'm also an upstream contributor, and Bruce Mah is very open to enhancements. I can add this functionality, but it won't be in a release for another 2-3 months if you're okay with that... What's the exact requirement?

You mean for iperf3? I think making `--pidfile` available in non-server mode would already help a lot.

pprindeville:
The question [has been raised](https://groups.google.com/g/iperf-dev/c/2PhbY-ok7SM/m/ALpemC1wCAAJ).

applications/luci-app-throughput-measurements/Makefile
```diff
@@ -0,0 +1,13 @@
...
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Throughput measurements application
+LUCI_DEPENDS:=+luci-mod-admin-full +iperf3
+PKG_MAINTAINER:=Wojciech Jowsa <wojciech.jowsa@gmail.com>
+PKG_LICENSE:=Apache-2.0
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signatureet
```
Vladdrako, line 13: A little typo mistake :)

wjowsa: Yes, indeed :) Fixed, thanks.

wjowsa:
My iperf3 [patch](https://github.com/esnet/iperf/pull/1110) is merged. I will change the application. @jow- please let know how do we proceed with the iperf3 patch i.e. bump up iperf3 package to the master branch version or add a single patch to the the current version of the iperf3 package?

openwrt-diy:
No new progress?

wjowsa:
@openwrt-diy As far as I am concerned then I think I added all requested improvements/fixes. Now waiting for @jow- @feckert  opinions.

stangri:
Just stumbled upon this accidentally. @wjowsa do you want to mark changes requested as resolved to see if this will get things moving?

wjowsa:
@stangri good idea. Just have to rebase it, and verify if it still works fine.

castillofrancodamian:
This is forgotten?

feckert:
@castillofrancodamian Unfortunately, I cannot say whether this is technically better from a safety point of view.
I don't have the expertise, sorry.

From my point of view, I would like to move the whole Iperf3 handling and configuration (uci) into the iperf3 package.
I already made a pullrequest for this https://github.com/openwrt/packages/pull/16282
I have also wrapped this ubus notificatation https://github.com/openwrt/packages/pull/16282/files#diff-bf955449eaba4a541395ec00eb092b7531cc16d0c9ff11131f73e827e12e812bR27

I would prefer to be able to use the whole thing without the LuCI.
I have no objection to configuring and starting it in the LuCI and then pushing it to the browser via ubus `luci.iperf3.notify`
I personally do not like the suggested lua rpcd in this pullrequest.

systemcrash:
Close this for now. Since the underlying PRs don't seem to be going anywhere.

Issue #3001, opened 2019-08-25
Security! Certificate problem for HTTPS of LuCI


Neustradamus:
By default no secure LuCI!
- https://fabianlee.org/2016/09/19/openwrt-enabling-https-for-the-luci-web-admin-interface/

After install of needed part, there is an automaticaly generation of the certificate!

Bad values:
- https://github.com/openwrt/openwrt/blob/master/package/network/services/uhttpd/files/uhttpd.init#L34

```
C=ZZ
ST=Somewhere
L=Unknown
O=OpenWrt+ID
CN=OpenWrt
```

Can you change it?
Add a possibility to admin to choice values?

hnyman:
???

/etc/config/uhttpd contains the config items for those data fields

And why do you claim that those values make it more unsecure? 
In any case the certificate will be a self-signed cert without proper FQDN, so most browsers will flag that cert somewhat unsecure in any case.

Neustradamus:
@hnyman: By default no HTTPS://openwrt/ only HTTP://openwrt/

After adding package in the router, uhttpd creates a certificate but the value is not good.

I have noted previously:
- https://github.com/openwrt/openwrt/blob/master/package/network/services/uhttpd/files/uhttpd.init#L34

```
C=ZZ
ST=Somewhere
L=Unknown
O=OpenWrt+ID
CN=OpenWrt
```

There is no option for change values by mines in LuCI or at generation no request to choose...

Neustradamus:
Please reopen this ticket, it is not solved.

hnyman:
Due to the aim for small footprint, no SSL support is added by default. HTTP is the default.

You can change the SSL keygen values in /etc/config/uhttpd if you like, and you can naturally re-generate the keys.  But for self-signed certs, the values do not have much value in any case.

You can also install letsencrypt support (acme) and use that if you have a full domain. 

And of course you can manually import your own certs if they are in a suitable format.

Neustradamus:
@hnyman:
- Where we can add HTTPS support in LuCI?
- Where we can select values in LuCI?
- Where we can add certs in LuCI?

LuCI is here to be easily for end users ;)

hnyman:
> Where we can enable HTTPS support in LuCI?

* system / software --> install luci-ssl

> Where we can select values of the default certificate in LuCI?

* install luci-app-uhttpd and use that to edit values. 
* or edit /etc/config/uhttpd directly

> Where we can add Lets Encrypt support in LuCI?

[...]

PR #4659, opened 2020-12-14
Allow configuration of HTTPS redirection

ynezz:
Add a possibility to control uHTTPd's `redirect_https` config option from LuCI. While at it, add some missing abbreviations found in SSH keys and add missing styling of abbreviations as well.

image

References: https://lists.infradead.org/pipermail/openwrt-devel/2020-December/032718.html

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
@@ -1,0 +1,22 @@
+'use strict';
+'require view';
+'require form';
+
+return view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('uhttpd', _('HTTP(S) Access'), _('uHTTPd offers <abbr title="Hypertext Transfer Protocol">HTTP</abbr> or <abbr title="Hypertext Transfer Protocol Secure">HTTPS</abbr> network access.'));
+
+		s = m.section(form.NamedSection, 'main', 'uhttpd', _('Settings'));
+		s.addremove = false;
+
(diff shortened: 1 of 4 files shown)


ynezz:
image

hnyman:
> Add a possibility to control uHTTPd's redirect_https config option from LuCI. 

One sidenote: we already have luci-app-uhttpd that covers the uhttpd options more extensively.

ynezz:
> One sidenote: we already have luci-app-uhttpd that covers the uhttpd options more extensively.

Thanks, I'm aware about that, but I didn't wanted to drag it in as a whole, since this PR (and #4660 ) are only about HTTPS feature, which is now hopefully more clear now with #4660

hnyman:
Why the added abbreviations? E.g. RSA inventor names that add no actual value?  Similarly SSH and HTTP/S look pretty unnecessary, as "hypertext transfer protocol" does not tell much more to the novice user than HTTP.

I do not think that explaining all acronyms is necessary.

Making those changes to the translated source strings will invalidate the current translations as the source string changes.

I would leave those abbreviation change away from a commit about adding that HTTPS toggle.

ynezz:
> Why the added abbreviations?

Consistency? I've simply noticed, that they're missing so I've added them.

> Similarly SSH and HTTP/S look pretty unnecessary, as "hypertext transfer protocol" does not tell much more to the novice user than HTTP.

Yeah, like following:

<abbr title="Internet Protocol Version 4">IPv4</abbr>
<abbr title="Domain Name System">DNS</abbr>
<abbr title="Media Access Control">MAC</abbr>
<abbr title="Dynamic Host Configuration Protocol">DHCP</abbr>
<abbr title="Light Emitting Diode">LED</abbr>
...

> I would leave those abbreviation change away from a commit about adding that HTTPS toggle.

Ok.

Neustradamus:
Thanks @ynezz, it is linked to:
- https://github.com/openwrt/luci/issues/3001

aparcar:
Tested and works, sorry for the delay.

Neustradamus:
@aparcar: Thanks!

PR #4661, opened 2020-12-15
luci-app-frp: add the dynamiclist for additional parameters

brvphoenix:
Maintainer: @ysc3839 
Add a dynamiclist which can be used to define some additional parameters. The init.d scripts have already supported this option.


ysc3839:
LGTM, thanks!

PR #4662, opened 2020-12-16
luci-app-wol: fix some untranslated strings

brvphoenix:


jow-:
Merged, thanks @brvphoenix !

PR #4664, opened 2020-12-16
luci-app-babeld: add babeld overview

PolynomialDivision:
Add overview of babeld status, xroutes and routes.

### DIFF
--- a/applications/luci-app-babeld/Makefile
+++ b/applications/luci-app-babeld/Makefile
@@ -1,0 +1,10 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for babeld
+LUCI_DEPENDS:=+babeld +luci-compat
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
+
--- a/applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
+++ b/applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
@@ -1,0 +1,127 @@
+m = Map("babeld", translate("Babeld Overview"))
+m.pageaction = false
+
+s = m:section(NamedSection, "__babeldoverview__")
+
+function s.render(self, sid)
+	local tpl = require "luci.template"
+
+	tpl.render_string([[
+		<%
+		local utl = require "luci.util"
+		local v6_port = 33123
+		-- FIXME: check if call is correct
+		local dump = utl.exec("(echo dump | nc ::1 %d)" % v6_port)
+		local t = {}
+
+		for strline in string.gmatch(dump, "[^\n]+") do
+			t[#t+1] = strline;
+		end
+
+		local babel_version = t[1]
+		local babeld_version = t[2]
+		local babeld_host = string.split(t[3], " ")[2]
+		local babeld_id = string.split(t[4], " ")[2]
+		local babeld_status = t[5]
+
+		%>
+		<div class="cbi-section-node">
+			<h3>Babel-Info</h3>
+			<div class="table" id=babel_status_main">
+				<div class="tr table-titles">
+					<div class="th">Version</div>
+					<div class="th">Version-Daemon</div>
+					<div class="th">Host</div>
+					<div class="th">ID</div>
+					<div class="th">Status</div>
+				</div>
+				<div class="tr">
+					<div class="td"><%= babel_version %></div>
+					<div class="td"><%= babeld_version %></div>
+					<div class="td"><%= babeld_host %></div>
+					<div class="td"><%= babeld_id %></div>
+					<div class="td"><%= babeld_status %></div>
+				</div>
+			</div>
+		</div>
+		
+		<%
+		local xroutes = {}
+		for key,value in ipairs(t) do
+			if string.match(value, "xroute") then
+				xroutes[#xroutes+1] = value
+			end
+		end
+		%>
+
+		<div class="cbi-section-node">
+			<h3>X-Routes</h3>
+			<div class="table" id=babel_overview_xroute">
+				<div class="tr table-titles">
+					<div class="th">Route</div>
+					<div class="th">Prefix</div>
+					<div class="th">From</div>
+					<div class="th">Metric</div>
+				</div>
+				<%
+				for key,route in ipairs(xroutes) do
+					local route_sep = string.split(route," ")
+				%>
+					<div class="tr">
+						<div class="td"><%= route_sep[3] %></div>
+						<div class="td"><%= route_sep[5] %></div>
+						<div class="td"><%= route_sep[7] %></div>
+						<div class="td"><%= route_sep[9] %></div>
+					</div>
+				<%
+				end
+				%>
+			</div>
+		</div>
+		
+		<%
+		local routes = {}
+		for key,value in ipairs(t) do
+			if string.match(value, "add route") then
+					routes[#routes+1] = value
+			end
+		end
+		%>
+
+		<div class="cbi-section-node">
+			<h3>Routes</h3>
+			<div class="table" id=babel_overview_route">
+				<div class="tr table-titles">
+					<div class="th">Prefix</div>
+					<div class="th">From</div>
+					<div class="th">Installed</div>
+					<div class="th">ID</div>
+					<div class="th">Metric</div>
+					<div class="th">Ref-Metric</div>
+					<div class="th">Via</div>
+					<div class="th">Interface</div>
+				</div>
+				<%
+				for key,route in ipairs(routes) do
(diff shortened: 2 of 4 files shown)


PolynomialDivision:
babeldapp

PolynomialDivision:
Fixed Typo

feckert (changes requested):
Thanks for your contribution. See my comments-

* space vs. tabs in controller babeld.lua or change controller to menu.d json that would be better. Please have a look at https://github.com/openwrt/luci/blob/master/applications/luci-app-clamav/root/usr/share/luci/menu.d/luci-app-clamav.json
* Missing new line for babeld_network.lua

This is a status page. So why not move them to the status menu entry?
I could also imagine to add a polling to update this page?
But this can also be done afterwards.

I am not familiar with babld.
So what kind of command is this and where does the data come from?
https://github.com/openwrt/luci/pull/4664/files#diff-1dde564372ff1363073d8c79269ca9f57f5bf1b7527ec17e545d834cb2b5cf36R13

PolynomialDivision:
@feckert Thanks for your comments. Currently, I'm adding ubus bindings to babeld.
 That is why I maybe will refactor that again and will include your changes. ;)

PolynomialDivision:
>     * space vs. tabs in controller babeld.lua or change controller to menu.d json that would be better. Please have a look at https://github.com/openwrt/luci/blob/master/applications/luci-app-clamav/root/usr/share/luci/menu.d/luci-app-clamav.json

Changed.

>     * Missing new line for babeld_network.lua

Changed.

> This is a status page. So why not move them to the status menu entry?

What do u mean with that?

>  So what kind of command is this and where does the data come from?

From the local socket from babeld. Actually, I made a PR addding ubus bindings:
https://github.com/openwrt-routing/packages/pull/630

Since, this is still under discussion I would appreciate, if we could merge this app without ubus support, and then if the ubus bindings are there, I can rewrite. :)

PolynomialDivision:
@feckert Sry, just saw that code formating is not correct. I will change that.

PolynomialDivision:
Okay. Tested everything and fixed code format. I'm finished. :)

applications/luci-app-babeld/Makefile
```diff
@@ -0,0 +1,10 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for babeld
+LUCI_DEPENDS:=+babeld +luci-compat +luci-lib-json
```
feckert, line 4: Do we need luci-lib-json?

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
```diff
@@ -0,0 +1,126 @@
+m = Map("babeld", "Babeld Overview", translate("Babeld Overview"))
```
feckert, line 1: The translation is missing from the second argument in the function. This is the heading of the section. The question is then only do you need the third argument in the function. This is usually used to describe the section more precisely. Just have a look at the other map functions in lua in the repo.

applications/luci-app-babeld/root/usr/share/luci/menu.d/luci-app-babeld.json
```diff
@@ -0,0 +1,13 @@
+{
+	"admin/services/babeld": {
```
feckert, line 2: I have mentioned this before but I don't think you understood me ( I remember that you asked about). I think the overview is better kept in the status menu tree for bable than in the configuration. You only show status information here, it is not configuration page. See this for example https://github.com/openwrt/luci/blob/be49059e4d4ce156cfdf1b0240b05072b495d882/applications/luci-app-wireguard/root/usr/share/luci/menu.d/luci-app-wireguard.json#L2

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
```diff
@@ -0,0 +1,126 @@
...
+s = m:section(NamedSection, "__babeldoverview__")
+
+function s.render(self, sid)
+	local tpl = require "luci.template"
+
+	tpl.render_string([[
+		<%
+		local utl = require "luci.util"
+		local v6_port = 33123
+			local dump = utl.exec("(echo dump | nc ::1 %d)" % v6_port)
```
feckert, line 13: There the indentation does not fit.
I believe that the echo before dump is not necessary.
Is the command really called dump?

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
```diff
@@ -0,0 +1,126 @@
...
+					<div class="td"><%= babeld_version %></div>
+					<div class="td"><%= babeld_host %></div>
+					<div class="td"><%= babeld_id %></div>
+					<div class="td"><%= babeld_status %></div>
+				</div>
+			</div>
+		</div>
+		
+		<%
+		xroutes = {}
```
feckert, line 48: local missing

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
```diff
@@ -0,0 +1,126 @@
...
+						<div class="td"><%= route_sep[9] %></div>
+					</div>
+				<%
+				end
+				%>
+			</div>
+		</div>
+		
+		<%
+		routes = {}
```
feckert, line 82: local missing

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
```diff
@@ -0,0 +1,126 @@
...
+					<div class="th">ID</div>
+					<div class="th">Metric</div>
+					<div class="th">Ref-Metric</div>
+					<div class="th">Via</div>
+					<div class="th">Interface</div>
+				</div>
+				<%
+				for key,route in ipairs(routes) do
+					local route_sep = string.split(route," ")
+				%>
```
feckert, line 106: If the variable is not defined and has no content, do we not want to display `-`?
I am not familiar with bable, so I don't know if this can happen when the information gets dumped

feckert (changes requested):

feckert:
> Okay. Tested everything and fixed code format. I'm finished. :)

Looks good so far. Except for my comments.

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
PolynomialDivision: Is the command really called dump?
 
 Yes. ;)

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
PolynomialDivision: that should not happen

PolynomialDivision:
Compile again and tested! :D Thanks a lot. I have to update my other luci apps accordginly to this. ;)

PolynomialDivision:
statusbabeld cleaned

applications/luci-app-babeld/luasrc/model/cbi/babeld.lua
```diff
@@ -0,0 +1,126 @@
...
+s = m:section(NamedSection, "__babeldoverview__")
+
+function s.render(self, sid)
+	local tpl = require "luci.template"
+
+	tpl.render_string([[
+		<%
+		local utl = require "luci.util"
+		local v6_port = 33123
+		local dump = utl.exec("(echo dump | nc ::1 %d)" % v6_port)
```
feckert, line 13: You still owe me an answer. Is the `echo` necessary?

PolynomialDivision: Yes. If I remove it, the output is not correct? Or is there another way? I copied that from luci-app-olsr ;)

PolynomialDivision: So I tested it. And it did not work.

feckert: I think you need to redirect stdout or stderr.
I think the echo is not necessary here.
But as I said, I can't try it out.
You can write the difference in a file.

But that should not bother use. It's just an extra fork that I think is not neccessary.
Maybe we should write a FIXME comment above it.
If someone has the time and the inclination, they can take another look at it.

PolynomialDivision: Added fixme comment. I will rewrite that when we have IPC via ubus merged. :)

PolynomialDivision:
I can not klick on request re-view. ;)
So I mention you again @feckert .

Thanks for the quick responses and the nice code review. :) This will not be the last commit to the babeld app. :)

feckert (approved):

feckert:
Thanks

PR #4666, opened 2020-12-18
luci-app-advanced-reboot: drop luci-compat dependency

brvphoenix:
This package has converted to client side and doesn't need `luci-compat` any more.

### DIFF
--- a/applications/luci-app-advanced-reboot/Makefile
+++ b/applications/luci-app-advanced-reboot/Makefile
@@ -10,7 +10,7 @@ LUCI_TITLE:=Advanced Linksys Reboot Web UI
 LUCI_DESCRIPTION:=Provides Web UI (found under System/Advanced Reboot) to reboot supported Linksys and ZyXEL routers to\
 	an alternative partition. Also provides Web UI to shut down (power off) your device. 	Supported dual-partition\
 	routers are listed at https://github.com/openwrt/luci/blob/master/applications/luci-app-advanced-reboot/README.md
-LUCI_DEPENDS:=+luci-compat +luci-mod-admin-full
+LUCI_DEPENDS:=+luci-mod-admin-full
 LUCI_PKGARCH:=all
 PKG_RELEASE:=55
 


stangri:
Hello Van,

I don't have access to a dual-boot router at the moment, which model/OpenWrt version have you tested the updated IPK on? Was the `luci-compat` manually removed prior to testing?

Thanks!

brvphoenix:
> Hello Van,
> 
> I don't have access to a dual-boot router at the moment, which model/OpenWrt version have you tested the updated IPK on? Was the `luci-compat` manually removed prior to testing?
> 
> Thanks!

I run the updated IPK on platform x86_64 but have not tested on dual-boot router.

I have checked the codes and doesn't find they use the `luci-compat`. Maybe I'm wrong.

Ping @Ansuel

hnyman:
> have not tested on dual-boot router.

So this is completely untested on the main target, dual partition routers ?????
You really should have mentioned that in the PR description.

(testing needs a build for dual-partition routers, made so that other packages do not pull luci-compat into build. I will see if I can made a really-stripped-down test build for my wrt3200acm)

hnyman:
I tested with a stripped-down build with this change for WRT3200ACM.
Both "reboot to current" and "reboot to alternative" work ok.

(As the automatic main LuCI version changes with the git commit, I do not see an immediate need for PKG_RELEASE bump)

Ansuel:
Any dependency with Luci compat is not needed. The PKG is full js. LGTM

Il Ven 18 Dic 2020, 10:00 Hannu Nyman <notifications@github.com> ha scritto:

stangri:
The package still depends on lua being present in the system. @Ansuel -- I can take a crack at re-writing luci.advanced_reboot in shell script, but I have a few questions, do you mind if I email you directly? 

For now tho, I think it would be wise to add dependency on lua to the Makefile. Would it be the principal lua package then, not `liblua` nor `libubus-lua`, right?

Ansuel:
Sure feel free to email me

Il Sab 26 Dic 2020, 21:36 Stan <notifications@github.com> ha scritto:

hnyman:
> For now tho, I think it would be wise to add dependency on lua to the Makefile. 

There is already an indirect dependency to lua via luci-mod-admin-full. 
Not sure why this app would require more.

(Typically extra dependencies to low-level libraries already referenced in the dependency chain, tend to cause more trouble than benefits.)

stangri:
> There is already an indirect dependency to lua via luci-mod-admin-full.

Oh, didn't realize it has dependency on lua. I was reading that lua is to be dropped from default packages eventually and was thinking of that eventuality.

PR #4667, opened 2020-12-18
luci-base: add generic tooltip handling

feckert:
@jow- I just have a use case where I need a generic tooltip handling.
Please have a look at this [commit](https://github.com/openwrt/luci/commit/c2f4e917cd099e0516bd7b80fa63482d6e17482c)

I want to output a hint in the LuCI that only restricted masquerading is enabled.
This has led to problems several times.

To map this correctly I have integrated into the Luci framwork.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -3602,13 +3602,35 @@ var CBIFlagValue = CBIValue.extend(/** @lends LuCI.form.FlagValue.prototype */ {
 	 * @default 0
 	 */
 
+	/**
+	 * Set a tooltip for the flag option.
+	 *
+	 * If set to a string, it will be used as-is as a tooltip.
+	 *
+	 * If set to a function, the function will be invoked and the return
+	 * value will be shown as a tooltip. If the return value of the function
+	 * is `null` no tooltip will be set.
+	 *
+	 * @name LuCI.form.TypedSection.prototype#tooltip
+	 * @type string|function
+	 * @default null
+	 */
+
 	/** @private */
 	renderWidget: function(section_id, option_index, cfgvalue) {
+		var tooltip = null;
+
+		if (typeof(this.tooltip) == 'function')
+			tooltip = this.tooltip.apply(this, [section_id]);
+		else if (typeof(this.tooltip) == 'string')
+			tooltip = (arguments.length > 1) ? ''.format.apply(this.tooltip, this.varargs(arguments, 1)) : this.tooltip;
+
 		var widget = new ui.Checkbox((cfgvalue != null) ? cfgvalue : this.default, {
 			id: this.cbid(section_id),
 			value_enabled: this.enabled,
 			value_disabled: this.disabled,
 			validate: L.bind(this.validate, this, section_id),
+			tooltip: tooltip,
 			disabled: (this.readonly != null) ? this.readonly : this.map.readonly
 		});
 
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -3616,6 +3616,17 @@ var CBIFlagValue = CBIValue.extend(/** @lends LuCI.form.FlagValue.prototype */ {
 	 * @default null
 	 */
 
+	/**
+	 * Set a tooltip icon.
+	 *
+	 * If set, this icon will be shown for the default one.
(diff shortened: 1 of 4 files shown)


modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -610,6 +610,18 @@ var UICheckbox = UIElement.extend(/** @lends LuCI.ui.Checkbox.prototype */ {
 
 		frameEl.appendChild(E('label', { 'for': id }));
 
+		var tooltip = this.options.tooltip(this.options.name);
```
jow-, line 613: This will raise an exception if `this.options.tooltip` is null.

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -1892,6 +1892,26 @@ var CBIAbstractValue = CBIAbstractElement.extend(/** @lends LuCI.form.AbstractVa
...
+	 * @abstract
+	 * @param {string} section_id
+	 * The configuration section ID
+	 *
+	 * @returns {*}
+	 * The method shall return a string. This is show as an tooltip.
+	 */
+	tooltip: function(section_id) {
+		return false;
+	},
```
jow-, line 1913: I think we should allow this property to be either a string (in which case it'll be filtered through `String.format()` with section_id as argument) or a function (which behaves like you implemented it) - similar to e.g. [`TypedSection.addbtntitle`](http://openwrt.github.io/luci/jsapi/LuCI.form.TypedSection.html#addbtntitle)

The vast majority of tooltip using code likely only provides a single static value.

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -3609,6 +3629,7 @@ var CBIFlagValue = CBIValue.extend(/** @lends LuCI.form.FlagValue.prototype */ {
 			value_enabled: this.enabled,
 			value_disabled: this.disabled,
 			validate: L.bind(this.validate, this, section_id),
+			tooltip: this.tooltip ? L.bind(this.tooltip, this, section_id) : null,
```
jow-, line 3632: Since you introduce the new `tooltip` property in the `LuCI.form.AbstractElement` properties, you should add support for it to all widget renderers and not just flag values. Alternatively move the default declaration & JSDoc comment above this function to the other `LuCI.form.FlagValue` specific properties.

jow-:
Nice addition - I like it. Please refer to my other review comments for details I'd like to change.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -610,6 +610,18 @@ var UICheckbox = UIElement.extend(/** @lends LuCI.ui.Checkbox.prototype */ {
...
+		if (typeof(tooltip) === 'string') {
+			frameEl.appendChild(
+				E('label', { 'class': 'cbi-tooltip-container' },[
+					"⚠️",
```
jow-, line 617: Not sure if we should hardcode this emoji as tooltip element / text. Maybe introduce another property `tooltiptitle` or similar which allows overriding it.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -610,6 +610,18 @@ var UICheckbox = UIElement.extend(/** @lends LuCI.ui.Checkbox.prototype */ {
 
 		frameEl.appendChild(E('label', { 'for': id }));
 
+		var tooltip = this.options.tooltip(this.options.name);
+		if (typeof(tooltip) === 'string') {
```
jow-, line 614: I'd rather test for `!== false` here. The `E()` function allows other, non-string values as content too, e.g. an array of child nodes or a DOM Node element which could've been returned by a user defined `tooltip()` implementation.

modules/luci-base/htdocs/luci-static/resources/ui.js
feckert: Made changeable with commit https://github.com/openwrt/luci/pull/4667/commits/8a3fbcb18db5ae99a35e0de289bd969bb3891a97

modules/luci-base/htdocs/luci-static/resources/ui.js
feckert: I'm not sure yet if this works with the changes I've made.

feckert:
I decided to do it only for CBIFlag for the time being.
@jow- Can you do another review for this?

PR #4670, opened 2020-12-19
luci-app-mwan3: Add UDP rule hint for QUIC

Strykar:
@feckert QUIC uses UDP and while new, is widespread enough that users may not even be aware it's not over TCP. Adding a rule hint seems less invasive than creating a whole template. Most users who are aware of what QUIC is are probably aware they need to swap the rule to UDP. Perhaps at some point in the future, this would default to UDP.
Closes https://github.com/openwrt/packages/issues/11571

Avinash H. Duduskar 2946372+Strykar@users.noreply.github.com


Strykar:
Forgot to sign-off correctly and `git push --force-with-lease origin patch-3` failed too.

hnyman:
I do not like it.

All other advice items in that mwan3 text are generic, and this would be only app/protocol specific. And the contents are quite unrelated to mwan3.

If you want to use QUIC and need advice about basic TCP/IP protocols, you should search advice from QUIC pages, not from mwan3 documentation.

Other thing is that this should be to the development master first, not to the old stable 19.07

So, NAK from me.

PR #4676, opened 2020-12-21
luci-app-ddns: whitespace fixes in overview.js

### LINKED ISSUE #4675: ddns check interval minimum value (seemingly) too high
Currently the smallest check interval is set to 5 minutes that is hard-coded in the code [1], [2] of `luci-app-ddns`.
I thought it should be possible to check for the IP change every second or even faster than that for these reasons:
1. It's a low cost operation for the router to check if it's WAN IP has changed or not
2. The point of DDNS (for me) is to eliminate downtime in case of a forced IP reallocation from my ISP so in the event of an IP change, I would hope to have the cut-over time as small as possible and 5 minutes seems way too long when I can configure DNS TTLs of 2 minutes in my DNS provider for example.

[1]
https://github.com/openwrt/luci/blob/879a71dc2dacf4ae4ff8ba738d000562ed42d728/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L970

[2]
```javascript
o.validate = function(section_id, formvalue) {
  var unit = this.section.children.filter(function(o) { return o.option == 'check_unit' })[0].formvalue(section_id),
	  time_to_sec = _this.time_res[unit || 'minutes'] * formvalue;
  
  if (formvalue && time_to_sec < 300)
	  return _('Values below 5 minutes == 300 seconds are not supported');
  
  return true;
};

```

petermetz:
```sh
commit 05328b2746602a7dd35e3b9e89448986438e7492
Author: Peter Somogyvari <peter.metz@unarin.com>
Date:   Mon Dec 21 11:39:31 2020 -0800

    luci-app-ddns: whitespace fixes in overview.js
    
    No change in logic, just the whitespace fixes
    as picked up by VSCode's auto-formatter.
    
    Signed-off-by: Peter Somogyvari <peter.metz@unarin.com>

```

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -66,7 +66,7 @@ return view.extend({
 	 *    downloaded)
 	 * 3. /usr/share/ddns/list contains all the service that can be
 	 *    downloaded by using the ddns script ('service on demand' feature)
-	 * 
+	 *
 	 * (Special services that requires a dedicated package ARE NOT
 	 * supported by the 'service on demand' feature)
 	 */
@@ -98,7 +98,7 @@ return view.extend({
 		}, this))
 	},
 
-	/* 
+	/*
 	* Check if the service is supported.
 	* If the script doesn't find any json assume a 'service on demand' install.
 	* If a json is found check if the ip type is supported.
@@ -469,10 +469,10 @@ return view.extend({
 			var section_id = name.isValid('_new_') ? name.formvalue('_new_') : null,
 				service_value = service_name.isValid('_new_') ? service_name.formvalue('_new_') : null,
 				ipv6_value = ipv6.isValid('_new_') ? ipv6.formvalue('_new_') : null;
-	
+
 			if (section_id == null || section_id == '' || service_value == null || section_id == '' || ipv6_value == null || ipv6_value == '')
 				return;
-	
+
 			return m.save(function() {
 				uci.add('ddns', 'service', section_id);
 				uci.set('ddns', section_id, 'service_name', service_value);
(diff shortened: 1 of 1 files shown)


feckert:
I don't think that will [work](https://github.com/openwrt/packages/blob/014e02ab075333bccd6ac5f3c8eaceba6ab41157/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh#L154). I have not looked at the code. but according to the comment, the kleinset here is 1 minute.

Could you also update your commit header to `luci-app-ddns: <message>` and also split your commit?
Whitespace and your time fix.

petermetz:
> I don't think that will [work](https://github.com/openwrt/packages/blob/014e02ab075333bccd6ac5f3c8eaceba6ab41157/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh#L154). I have not looked at the code. but according to the comment, the smallest here is 1 minute.

@feckert Maybe I'm misunderstanding what the `check_interval` actually does: That [explanation](https://github.com/openwrt/packages/blob/014e02ab075333bccd6ac5f3c8eaceba6ab41157/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh#L154) doesn't make sense to me because my router would be aware of it's WAN IP changing as soon as it happens, right?
My assumption here is that the DDNS package is checking if the router's WAN IP changed and if it has, then it sends a DNS update immediately (and then that DNS update takes however long it takes for "the internet" to start using the new IP, but regardless of how long that takes, the check itself for WAN IP changes should run as often as possible so that the mentioned DNS update can be triggered as soon as possible simply because the sooner you start it the sooner it will finish).
 
> Could you also update your commit header to `luci-app-ddns: <message>` and also split your commit?
> Whitespace and your time fix.

@feckert: Done.

feckert:
thanks anyway

kocoten1992:
I believe your original the 5 minutes check was 11 years ago, nowaday you can configure TTL of DNS record to be 1 minute (at least cloudflare allowed that), maybe we should allow 1 minute check and update too?

PR #4685, opened 2020-12-22
luci-app-openvpn: add hint for comp_lzo and compress issue

feckert:
The compress and comp_lzo are not recommended.
See https://community.openvpn.net/openvpn/wiki/VORACLE

So add a hint to the LuCI for this issue.
Also add the option `stub-v2` .

Since commit https://github.com/openwrt/packages/commit/e4376793b4e093089543cb1bad64eef34ed25eca lzo was set to default no on openwrt builds. The question is now should we also remove this from LuCI?

### DIFF
--- a/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-advanced.lua
+++ b/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-advanced.lua
@@ -162,7 +162,7 @@ local knownParams = {
 		{ ListValue,
 			"compress",
 			{ "lzo", "lz4", "stub-v2"},
-			translate("Enable a compression algorithm") },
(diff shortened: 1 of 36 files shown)


nickberry17:
>Since commit openwrt/packages@e437679 lzo was set to default no on openwrt builds. The question is now should we also remove this from LuCI?

FWIW, I am in favour of keeping the front and back ends synchronised for the sake of clarity, as well as promoting best practices where practical.  Don't you need to regenerate the i18n files before this can get merged?

nickberry17 (changes requested):
>Since commit openwrt/packages@e437679 lzo was set to default no on openwrt builds. The question is now should we also remove this from LuCI?

FWIW, I am in favour of keeping the front and back ends synchronised for the sake of clarity, as well as promoting best practices where practical.  Don't you need to regenerate the i18n files before this can get merged?

PR #4686, opened 2020-12-22
luci-base: update placeholder for IPv4 gateway on proto static

feckert:
The problem is that when switching from the dhcp proto to the static
proto, the IPv4 gateway is already filled in as a placeholder but not
preselected. With some themes (luci-theme-openwrt-2020) it is not well
visible that this is only a suggestion and is not a preselection.

By adding 'Add IPv4 gateway...' to the placeholder, it is more cleaner
that it is only a suggestion.


jow-:
That change does not really make sense to me. The placeholder displays the current state, it is not about "adding" any gateway.

 I think we should rather fix the themes in question to display placeholder values in a more muted manner.

PR #4687, opened 2020-12-23
luci-app-attendedsysupgrade: remove depends uhttpd-mod-ubus

zhanhb:
Remove depenedency uhttpd-mod-ubus which is not required.
User may want to work with nginx rather uhttpd.


aparcar:
Thanks

PR #4689, opened 2020-12-24
luci-mod-network: skip check for 802.11w feature

dobo90:
OpenWrt commit 1a9b896d ("treewide: nuke DRIVER_11W_SUPPORT") enables
802.11w feature for all wpad/hostapd configurations. The feature flag
was removed at all but for the compatibility reasons 11w is still
advertised (but there's a plan to nuke it also) [1].

Remove conditional 802.11w LuCI support to match current behavior.

[1]: https://github.com/openwrt/openwrt/pull/3347

Signed-off-by: Dobroslaw Kijowski <dobo90@gmail.com>


hnyman:
Thanks. Merged.

PR #4690, opened 2020-12-25
Add Luci for Eoip package

bogdik:
For https://github.com/openwrt/openwrt/pull/3721

### DIFF
--- a/applications/luci-app-eoip/Makefile
+++ b/applications/luci-app-eoip/Makefile
@@ -1,0 +1,16 @@
+# Copyright 2020 Bogdan Shatik (bogdikxxx@mail.ru)
+# This is free software, licensed under the Apache License, Version 2.0
+
+include $(TOPDIR)/rules.mk
+
+PKG_MAINTAINER:=Bogdan Shatik <bogdikxxx@mail.ru>
+PKG_NAME:=luci-app-eoip
+PKG_LICENSE:=Apache-2.0
+
+LUCI_TITLE:=LuCI Support for EoIP
+LUCI_DEPENDS:=+eoip +luci-base
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-eoip/htdocs/luci-static/resources/view/eoip/eoip.js
+++ b/applications/luci-app-eoip/htdocs/luci-static/resources/view/eoip/eoip.js
@@ -1,0 +1,50 @@
+'use strict';
+'require form';
+'require view';
+'require uci';
+
+return view.extend({
+	render: function() {
+		var m, s, n, id, d, o, v;
+		
+		m = new form.Map("eoip", _("EoIP - Tunneling"), _("Here you can configure EoIP tunnel. At current moment it is easiest way to create stateless tunnel with Mikrotik."));
+
+		s = m.section(form.TypedSection, "eoip", _("Settings"));
+		s.addremove = true;
+		s.anonymous = true;
+		
+		o = s.option(form.Flag, "enabled", _("Enable tunnel"));
+
+		n = s.option(form.Value, "name", _("Name interface [zeoip"), _("If you input 0 interface name zeoip0"));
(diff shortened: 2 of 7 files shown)


stangri:
Hey, thanks for the contribution. There was a discussion if the new lua-based luci packages will be accepted or not (since the transition to the javascript-based WebUI), I hope @hnyman or @Ansuel  or @feckert will chime in on that.

There are still a few things which need to be fixed to improve chances that this will be merged:
1. Please add "luci-app-eoip: " prefix to the commit message.
2. Please squash all commits into one commit and force-push the changes.
3. Please sign the commit with your real name.

hnyman:
Reference to packages PR https://github.com/openwrt/packages/pull/14334

Regarding LuCI, 
* there should be just one commit
* the commit needs to be properly titled, and signed off by you
* .pot file is empty although you have created
* preferably the new packages should use javascript instead of lua. This is so small GUI package, that it should be easy to convert.

bogdik:
Done

bogdik:
On last add info for dybamic option

bogdik:
> * there should be just one commit
> * the commit needs to be properly titled, and signed off by you
> * .pot file is empty although you have created
> * preferably the new packages should use javascript instead of lua. This is so small GUI package, that it should be easy to convert.

Check please

applications/luci-app-eoip/Makefile
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Bogdan Shatik (bogdikxxx@mail.ru)
+# This is free software, licensed under the Apache License, Version 2.0
+
+include $(TOPDIR)/rules.mk
+
+PKG_MAINTAINER:=Bogdan Shatik <bogdikxxx@mail.ru>
+PKG_NAME:=luci-app-eoip
+PKG_VERSION:=1.0.1
```
hnyman, line 8: There is no need for versioning in the LuCI packages as there is automatic versioning based on commit dates.
You can remove the PKG_VERSION line.

applications/luci-app-eoip/htdocs/luci-static/resources/view/eoip/eoip.js
```diff
@@ -0,0 +1,39 @@
...
+		
+		d = s.option(form.Value, "dst", _("Destination"), _("Destination IP address for connection EoIP."));
+		d.nocreate = true;
+		d.datatype = "ipaddr";
+		d.placeholder = "0.0.0.0";
+		
+		o = s.option(form.Flag, "dynamic", _("Dynamic"), _("If you use dynamic option, take attention that there is no authorization, and it is not secure. It is not good idea to use this feature with public ip or insecure(not completely under your control, each host) network."));
+		
+		v = s.option(form.DynamicList,"vlan", _("VLAN"), _("VLAN TAG on this interface"));
+		v.datatype = "and(and(max(4094), integer),and(min(1), integer))";
```
hnyman, line 35: "range" datatype would be much easier and more clear.
`range(1,4094)`

applications/luci-app-eoip/Makefile
```diff
@@ -0,0 +1,15 @@
...
+
```
hnyman, line 9: You might add the target-indenpendent line:
`LUCI_PKGARCH:=all`

hnyman, line 9: You should add the license database line. Your copyright sets is as Apache 2

`PKG_LICENSE:=Apache-2.0`

hnyman (changes requested):
Some small comments

hnyman:
I have one more comment:

I think that there are still invalid datatypes:   `"min(0), integer"` looks faulty, used in two places.
It should be `"and(min(0), integer)"`  or likely it could just be the unsigned integer type `"uinteger", if the only restriction is that it needs to be non-negative integer. that "uinteger" pretty much defined that by itself.

Have you tested the app?

bogdik:
> Have you tested the app?

Yes, tested fine

bogdik:
Compiled, tested, everything is fine, everything works

Issue #4475, opened 2020-09-27
WG intercace public key is not created, and the one of peer is displayed.


nerdoc:
openwrt-19.07 branch (git-20.247.75781-0d0ab01) / OpenWrt 19.07.4 r11208-ce6496d796 

I created a new interface named `wg`, entered a private key, and filled out the basic needed forms, and clicked on save, even rebooted the router. 
![grafik](https://user-images.githubusercontent.com/2955584/94366838-30612a00-00ca-11eb-8380-8f3873b5d361.png)

I did not yet add a peer, as this is not needed by now - I want to add an android device by displaying the QR code.
But at the wireguard status page, no QR code is displayed, and it tells me that "there is no public key available" for this interface.

IMHO this is clearly wrong, as the pubkey should be calculated from the private key.

When I add a peer (any IP, test pubkey 9YzLmC73/vH5sHTFDEXK3RXknM8RUrqs51F2/sui4Gc=) and click on save + reboot, the QR code appears and displays the pubkey of the peer:
![grafik](https://user-images.githubusercontent.com/2955584/94366944-dad94d00-00ca-11eb-8242-1ef39a02bd27.png)

BUT: this is not the pubkey of the interface itself, but the one of the peer it should connect to.

Even if I delete the peer agian, this pubkey stays in the wireguard status.

----------------------
What I expect:
1. add a wireguard device in Luci, let Luci (by calling `wg privatekey`) create a new private key. Even without adding a peer, the pubkey should be created and displayed readonly.
2. when displaying the QR code (which I can scan from e.g. the android app), display the public key as well in cleartext, or even better: add a button next to it to save this newly created peer as peer in the wireguard config under `/luci/admin/network/network -> <wg interface>`. So I can scan and setup this peer on my phone, and with a few clicks, on the router.

But the current behaviour is definitely wrong (the displayed pubkey is not the one of the wg interface) - and could be done much easier.

nerdoc:
This is funny, after deleting peer a few times and rebooting, it works as expected. Maybe this is just the case after first time setup?

nerdoc:
Can anyone confirm this? If not, please close it, and I'll file a separate bug for the "one-click-experience" enhancement to create a peer.

castillofrancodamian:
It seems to me that it is incorrect that the public key is generated from the private one.
https://forum.openwrt.org/t/duckdns-openwrt-on-wireguard/65354/30?u=castillofrancodamian

n-thumann:
[...]

PR #4692, opened 2020-12-27
luci-app-wireguard: select qrencode optional dependency

nickberry17:

Partial fix for https://github.com/openwrt/luci/issues/4475


dibdot:
What kind of error will be fixed by this additional package dependency? This adds ~60 KB to every luci-app-wg user although the QRCode display only optional.

hnyman:
Where did that normal dependency convert to "optional dependency" ????

NAK from me.

PR #4702, opened 2020-12-31
luci-app-sqm: Use CAKE as default qdisc

tohojo:
The sqm-scripts package depends on kmod-sched-cake, meaning CAKE is always
available in normal installs. So we might as well just make CAKE the
default config in SQM configs.

### DIFF
--- a/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
+++ b/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
@@ -95,7 +95,7 @@ return view.extend({
 		for (var i=0; i < qdiscs.length; i++) {
 			o.value(qdiscs[i].name);
 		}
-		o.default = "fq_codel";
+		o.default = "cake";
 		o.rmempty = false;
 
 		var qos_desc = "";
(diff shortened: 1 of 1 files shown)


hnyman:
Like I commented in https://github.com/tohojo/sqm-scripts/issues/130#issuecomment-752964460
the actual deafult in SQM in OpenWrt is still simple/fq_codel

This PR needs to wait until you have changed the SQM itself to have cake as default in OpenWrt.

tohojo:
Updated sqm-scripts with https://github.com/openwrt/packages/commit/2c652128c52e37954b500d575d4251047278cbbb - merging this as well.

And happy new year! :)

hnyman:
@tohojo 
You forgot to update PKG_MIRROR_HASH in https://github.com/openwrt/packages/commit/2c652128c52e37954b500d575d4251047278cbbb

It still has the hash git .tar.xz of git clone of the previous version on line 17

(you did right the previous time, with https://github.com/openwrt/packages/commit/1126368e0e855107be91ecb5b206505237b9fee2 )

tohojo:
Hannu Nyman <notifications@github.com> writes:

> @tohojo 
> You forgot to update PKG_MIRROR_HASH in
> https://github.com/openwrt/packages/commit/2c652128c52e37954b500d575d4251047278cbbb

Ah, bugger, sorry about that - fixed! As always, thanks for keeping me
honest ;)

PR #4706, opened 2021-01-02
luci.mk: remove findrev function

aparcar:
The function moved over to openwrt.git.

In that process the PO logic was inverted to be "parameter less" per
default for packages within openwrt.git.

Also the "git-" prefix was removed as that information is redundant to
the suffixed revision.

### DIFF
--- a/luci.mk
+++ b/luci.mk
@@ -68,32 +68,6 @@ HTDOCS = /www
 LUA_LIBRARYDIR = /usr/lib/lua
 LUCI_LIBRARYDIR = $(LUA_LIBRARYDIR)/luci
 
-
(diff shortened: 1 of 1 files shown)


SvenRoederer:
relating thread on openwrt-devel ML: http://lists.openwrt.org/pipermail/openwrt-devel/2021-January/033037.html

hnyman:
>  function moved over to openwrt.git.

I do not yet find it there...

(Apparently so far it is only in your staging tree, so I marked this as depending on something )

aparcar:
Yes sorry for the confusion.

jow-:
NACK, I'd like to keep a private copy here to remain independent from unintended OpenWrt changes.

PR #4710, opened 2021-01-05
[19.07] let `passwd` using `busybox` to avoid failure when `shadow-utils` is installed

KaneGreen:
If the package `shadow-utils` is installed, `passwd` will not be provided by `busybox`. In this case, changing the password on the web page will fail.
The PR fixed this problem by calling the `busybox passwd` subcommand.

This has been successfully tested on the `openwrt-19.07` branch. In theory, this also applies to the `master` branch.


jow-:
Would be interesting to know why it fails. Is `(echo secret; sleep 1; echo secret) | passwd` not working with shadow-passwd? What's the error when running this command?

KaneGreen:
@jow- As far as I know, `shadow-passwd` receive information from stdin.
(But only some old version of `passwd` can. For example CentOS 6, `passwd --stdin` can receive passwords from stdin)

egorenar:
I'm also hitting this problem now with current master. And `busybox passwd` helps.
Is there a problem merging this PR ?

KaneGreen:
@egorenar 
On master, there are some problems on shadow's hash algorithm. See comments of this commit https://github.com/openwrt/packages/commit/adf9c249498fe526e2d6afd19bf7985f116c8531.

PR #4712, opened 2021-01-05
[master] let `passwd` using `busybox` to avoid failure when `shadow-utils` is installed

KaneGreen:
This is the master version of PR #4710

> If the package shadow-utils is installed, passwd will not be provided by busybox. In this case, changing the password on the web page will fail.
> The PR fixed this problem by calling the busybox passwd subcommand.
> 
> This has been successfully tested on the openwrt-19.07 branch. In theory, this also applies to the master branch.


KaneGreen:
@jow- I've noticed this commit https://github.com/openwrt/luci/commit/22e2bfb09667e167a93b5eca4487ac41669126d9
And I think maybe this file `modules/luci-base/luasrc/sys.lua` should be update, too.

I've updated this PR.

Besides, I hope these changes can be cherry-picked to branch 21.02 and 19.07, also see https://github.com/openwrt/luci/pull/4710

Thank you.

PR #4713, opened 2021-01-05
luci-mod-network: wireless.js Add/fix descriptions

nicklowe:
Add and improve the existing descriptions for the options available.
Improve naming.
Use consistent upper camel casing for names in the UI.

Signed-off-by: Nick Lowe <nick.lowe@gmail.com>


feckert:
In principle, this fits.
But I would have split the commits.
* a commit that changes the translation
* a commit that changes the datatype
* a commit that changes the **depends** to **add_dependency_permutations**

Now everything is in **one** commit.
This makes the review more difficult with the size of the change.
That would be an improvement :-)

feckert (commented):
In principle, this fits.
But I would have split the commits.

* one for the translation
* one for the datatype
* one for the depends to add_dependency_permutations change

Now everything is in one commit.
This makes the review more difficult with the size of the change.
That would be an improvement :-)

nicklowe:
> In principle, this fits.
> But I would have split the commits.
> 
> * one for the translation
> * one for the datatype
> * one for the depends to add_dependency_permutations change
> 
> Now everything is in one commit.
> This makes the review more difficult with the size of the change.
> That would be an improvement :-)

I have made this proposed commit narrowly only touch names and descriptions.

hnyman:
To be honest, changing that many translated strings (and mainly for cosmetic reasons) looks excessive, as it would invalidate all those for all translated languages (as you change the translated source string, the .po files get invalid).

You make 100-200 changes, (I didn't count exactly), that multiplied by 40+ translated languages is about 4000-8000 translated strings invalidated. Maybe 1/2 of those are actually translated, so that this commit only destroys maybe 2000-4000 strings. Pretty much all translation content for wireless.js get invalid...

E.g. if you just change capitalisation or marginally clarify the word, it makes no sense to remove all translations.

I assume that you would correspondingly edit the source strings (with sed or similar) in all .po files appropriately in order to preserve the translations.

(Just as reference, I changed yesterday one string here: e2ac2e92  )

nicklowe:
Thanks, I am aware of the translation impacts.

If we could first get review and agreement on both the revised strings and the new strings, I can then go through and automate a way to generate appropriate updates to .po files to link to the existing translations.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1322,61 +1324,61 @@ return view.extend({
 				crypto_modes.sort(function(a, b) { return b[2] - a[2] });
 
 				for (var i = 0; i < crypto_modes.length; i++) {
-					var security_level = (crypto_modes[i][2] >= 30) ? _('strong security')
-						: (crypto_modes[i][2] >= 20) ? _('medium security')
-							: (crypto_modes[i][2] >= 10) ? _('weak security') : _('open network');
+					var security_level = (crypto_modes[i][2] >= 30) ? _('High')
```
hnyman, line 1327: No.
These explanations were added a while ago.  High/Medium/Low is much more vague the the current explanations.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1438,111 +1440,111 @@ return view.extend({
 					// Probe 802.11r support (and EAP support as a proxy for Openwrt)
 					var has_80211r = L.hasSystemFeature('hostapd', '11r') || L.hasSystemFeature('hostapd', 'eap');
 
-					o = ss.taboption('encryption', form.Flag, 'ieee80211r', _('802.11r Fast Transition'), _('Enables fast roaming among access points that belong to the same Mobility Domain'));
+					o = ss.taboption('encryption', form.Flag, 'ieee80211r', _('Fast transition roaming'), _('Enable 802.11r <abbr title="Fast Transition">FT</abbr> roaming between APs that are part of the same mobility domain.'));
```
hnyman, line 1443: I dislike moving 802.11r like of explanations to only comment text

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
nicklowe: This change was to stop values being too large for the dropdown menu with after being selected and being truncated.

hnyman:
Good to hear. 
I think that automated translation string manipulation may succeed to some extent for the complex strings, where wireless.js is the only user, but some simple strings like "disabled" and "Disabled" are used by various other places (6 and 11), so neither should be removed/invalidated.  base.po offers reference to the places where each string is used.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
nicklowe: I humbly disagree, 802.11k, 802.11r, 802.11v and 802.11w are all just  historical amendments to 802.11 to originally defined these that has long since been folded in to later revisions of the standard, 802.11-2016.
In much the same way we do not commonly refer to WMM as "802.11e WMM" these days, we just call it WMM.
Fast Transition (FT) is the name of the feature defined in the 802.11 standard.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
nicklowe: Switched back to previous design which may truncate after being selected.

amer988:
I am a beginner person but I only need a full version of the AC5loco of your mods and hassle-free can you help me

nickberry17:
> I am a beginner person but I only need a full version of the AC5loco of your mods and hassle-free can you help me

Email me

aparcar:
@jow- is there a script for migrating the "source" strings in all translations?
@nicklowe please rebase

PR #4719, opened 2021-01-08
[RFC] luci-mod-system: drop kmod-ledtrig-* dependency

mans0n:
~The kmod packages were dropped from OpenWrt,~
~and the led triggers are now kernel built-in.~
...Not yet, but it will be.

This is the counterpart of openwrt/openwrt#3754.


jow-:
The change makes sense to me. Merged to clear the way for the OpenWrt side.

PR #4723, opened 2021-01-10
luci-mod-status: fix bugs in channel analysis

Ansuel:
- Add check for missing ht_operation data
- Add support for radio that both support 5 and 2.4 band
- Hide wireless station that doesn't belong to the graph band

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
@@ -63,8 +63,8 @@ return view.extend({
 		if (scanCache[res.bssid].graph == null)
 			scanCache[res.bssid].graph = [];
 
-		for (var i=0; i < channels.length; i++) {
-			var chan_offset = offset_tbl[channels[i]],
+		channels.forEach(function(channel) {
+			var chan_offset = offset_tbl[channel],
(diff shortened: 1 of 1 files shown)


hnyman:
I copied it to a live router, and it seemed to work ok.

Do you want to wait for additional feedback? (or just merge this?)

Ansuel:
I would wait for additional feedback

gw826943555:
radio0
radio1
good job! 
But the channel width of GFW(5GHz) and nicaicai_5G(5GHz) are both 80MHz.

hnyman:
I have noticed that some radios (at least 5GHz) are first displayed as 20 GHz but if the scan page remains open, they may widen on further scans.  That happened already with the December version of code.

Ansuel:
@hnyman the current code by default set 20mhz wide if the ht or vht info are not received. Could be that this is the problem. Could be that our wifi while scanning sometimes doesn't receive the vht info. In theory the graph should fix itself and make the wifi wider. Does this feature works?

hnyman:
> In theory the graph should fix itself and make the wifi wider. 

Yeah, it does. I have noticed especially mwlwifi in WRT3200ACM to first show up ar 20 MHz, and then later correct itself to wider.

hnyman:
I will merge it, so that possible feedback targets the new code.

PR #4733, opened 2021-01-16
Clean up trailing whitespace in 'modules'

thg2k:
Whitespace and indent cleanup of 'modules' folder.

- Remove trailing whitespace
- Ensure one and only one final newline at the end of file
- Fix indentation

### DIFF
--- a/modules/luci-base/luasrc/ccache.lua
+++ b/modules/luci-base/luasrc/ccache.lua
@@ -61,7 +61,7 @@ function cache_enable(cachepath, mode)
 	package.loaders[2] = function(mod)
 		local encoded = cachepath .. "/" .. _encode_filename(mod)
 		local modcons = _load_sane(encoded)
-		
+
 		if modcons then
 			return modcons
 		end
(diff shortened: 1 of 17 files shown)


feckert:
From my point of view, I think fixing white-space is good.
It always annoys me when it doesn't fit..
@hnyman what do you think?
I'm just thinking of the history when you want to check with git blame.
That could make it difficult to find the last time who changed the area.

hnyman:
I feel that it is nice tidying, but a bit unnecessary. 
git history (blame etc.) gets more cryptic, like you said.

thg2k:
git blame has option -w to skip whitespace-only changes, in that case it will display directly the commit prior to that. But yes, strictly speaking all the touched lines would alter the basic git blame.

Still, I personally hate trailing whitespace, and would hope you consider accepting the PR. I only restricted the cleanup to "modules" but it should be done project-wise. This is more like a probe to see maintainers reaction.

thg2k:
So I rebased and now this PR is only related to whitespace and extra newlines at the end of files. I don't think it actually makes much sense as it is, i.e. restricted to the "modules" folder, but applying this change to the whole source tree is a much bigger change.

I will leave the decision to you, it you think it's unnecessary just close it. Thanks!

PR #4737, opened 2021-01-18
Fix case for 'Bit/s' to 'bit/s'

thg2k:
The correct letter case for bits-per-second is "bit/s"

https://en.wikipedia.org/wiki/Data-rate_units

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/iwinfo.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/iwinfo.js
@@ -57,8 +57,8 @@ return baseclass.extend({
 		var bitrate = {
 			title: "%H: Average phy rate on %pi",
 			detail: true,
-			vlabel: "MBit/s",
-			number_format: "%5.1lf%sBit/s",
+			vlabel: "Mbit/s",
+			number_format: "%5.1lf%sbit/s",
 			data: {
 				types: [ "bitrate" ],
 				options: {
--- a/modules/luci-base/po/ar/base.po
+++ b/modules/luci-base/po/ar/base.po
@@ -3653,12 +3653,6 @@ msgstr ""
 msgid "MAP rule is invalid"
 msgstr ""
 
-#: modules/luci-mod-status/htdocs/luci-static/resources/view/status/wireless.js:321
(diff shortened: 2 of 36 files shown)


feckert:
Could you also please change the msgid in the po files as well?
So that weblate is satisfied and the translators don't have to translate everything again?

thg2k:
> Could you also please change the msgid in the po files as well?
> So that weblate is satisfied and the translators don't have to translate everything again?

To be fair in the translation files it appears both as "MBit/s" and "Mbit/s", in fact I noticed while translating. So it is not additional work for the translators, it is in fact a removal and I understand it should happen automatically by running the xgettext scripts?

If needed I can also add it to the pull request. Should I remove the sentence from templates or also all the translations?

feckert:
The following commit changes the luci-i18n-base-* for all languages 
modules/luci-mod-status/htdocs/luci-static/resources/view/status/wireless.js

The following commit changes the luci-18n-statistics-* for all languages
applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/iwinfo.js luci-i18m-statistics

I just wanted to say that if we make this change, then we should also adjust all msgid in the languages bo files. This should be added to this pullrquest. Then not all translations will be invalid and the translators don't have to start all over again.
If there is another place that does not fit, you can make another pullrequest. I am a fan of keeping the changes per pullrequest small. Experience has shown that this is then quickly reviewed and excepted.

hnyman:
> So it is not additional work for the translators, it is in fact a removal and I understand it should happen automatically by running the xgettext scripts?

Feckert means that you should correspondingly fix in .po files (and the .pot) the translation English source strings ("msgid"), which serve as the keys for the actual translations ("msgstr"). 

see example a a recent string change that I did.
https://github.com/openwrt/luci/commit/e2ac2e92b0de9f4be78b090ad7fc62a8df222b37

thg2k:
@hnyman I rebased and squashed the change with the language updates. As I tried to say, in the base.pot there were both `MBit/s` *and* `Mbit/s`, so I just unified the two strings.

Also please note that you should run an `./build/mkbasepot.sh && ./build/i18n-update.pl` as there is some pending updates to the language lines and translations.

hnyman:
There is a slight conflict as I merged the weblate changes, including a MBit/s Italian translation from you...

I will manually edit your commit to fix that.

hnyman:
Applied as dfa82b4494

PR #4742, opened 2021-01-19
luci-base: luci.js: fix `LuCI.request.poll.add()` exception handling

thg2k:
The try/catch is meant for the `res.json()` call and should only apply to that. As it was before, an exception inside the poll callback would cause the callback to be reinvoked without the JSON parameter, which is an odd behaviour. Moreover, it makes it hard to debug because it is completely hidden from the browser console. We now differentiate between exceptions thrown due to bad JSON in `responseText` from exceptions generated inside the callback itself, which are let through for browser console logging.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/luci.js
+++ b/modules/luci-base/htdocs/luci-static/resources/luci.js
@@ -982,12 +982,13 @@
 						if (!Poll.active())
 							return;
 
+						var res_json = null;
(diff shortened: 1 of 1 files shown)


thg2k:
@jow- what do you think about this?

jow-:
Yep, makes sense.

jow-:
Merged, thanks!

PR #4748, opened 2021-01-26
luci-mod-status: refactor code for processes

GermanAizek:
Signed-off-by: Herman Semenov GermanAizek@yandex.ru


feckert:
Why is this necessary. The pattern is like this and is also used in other modules.

jow-:
The change is also incorrect since the constructed RPC method is not actually invoked.

PR #4749, opened 2021-01-26
luci-mod-status: use the new ubus dsl metrics

dhewg:
This is significantly faster.

This is the luci feed part of: http://patchwork.ozlabs.org/project/openwrt/list/?series=220557
That needs to be merged first before this can go in


hauke:
Thank you for the work, the related changes in the main repository were also merged.

PR #4753, opened 2021-01-29
luci-app-statistics: add ipstatistics plugin

PolynomialDivision:


PolynomialDivision:
openwrtipstatsticsnew

PolynomialDivision:
@hnyman Can we merge this, too?

hnyman:
You are not going to format label display units first?

I would assume that something like `number_format: "%5.1lf%SB/s"` might make sense.
(or %s instead of %S if the values magnitudes differ much)

E.g.
https://github.com/openwrt/luci/blob/e46c96750aa8b5a4f9d3ed1cd7909485a26c1353/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/iptables.js

Docs at:
https://oss.oetiker.ch/rrdtool/doc/rrdgraph_graph.en.html

hnyman:
@PolynomialDivision 

Hmm. I didn't test this before merging, as it seemed to run for you. 

But when I try to enter the LuCI statistics config section with that collectd plugin installed, I get now an error, likely due to the missing configSummary section.

Does the config part really work for you in LuCI?

PolynomialDivision:
No it did not work.
Sry, I did not try on a fresh install. I wrote the collectd config manually, so I did not notice and deleted the configSummary Part since I thought it was useless. Next time I will be more careful.

Issue #4160, opened 2020-06-11
Please add missing sysupgrade options


zorxd:
Hi

sysupgrade has some interesting options, including the following:

-k : include in backup a list of current installed packages
-u : skip from backup files that are equal to those in /rom

It would be nice if they were added to LuCI as well. Personally, I would always choose these two options when upgrading. If you think adding these options would add to much bloat to the GUI, I think at least -k could be considered safe to add by default, perhaps -u as well. If a user didn't touch a file, it's likely the default one will be good after the upgrade. Maybe -u could even help prevent bugs when keeping configuration files and upgrading to a new major release.

shoulders:
hi

for newbies can someone explain what the following settings do?

-k : include in backup a list of current installed packages
-u : skip from backup files that are equal to those in /rom

![image](https://user-images.githubusercontent.com/319997/142461901-0df83eac-1433-4b78-9edb-38600e726dfc.png)

And just confirm that `Keep settings and retain the current configuration` just preserves the config settings during the upgrade but not any of the binaries/packages installed by users.

My internet searches just bring me back here and these options are not explained anywhere.

thanks

zorxd:
-k creates a file with all packages installed before the update which will be available in installed_packages.txt after the update. So no more "Oh, which package am I missing?" since yes, software update doesn't preserve installed packages.

-u saves some space by not including unmodified configuration files in the backup. It also has the benefit of getting new default options if they changed with the new update, instead of getting stuck with legacy configuration files forever.

PR #4754, opened 2021-01-29
luci-mod-system: improve sysupgrade page

Ansuel:
- Add missing -k and -u option from sysupgrade page.
- Fix missing check for image verification exit code
- Provide the actual reason of the image verification

Fixes: #4160

@jow- 

Here the screen of the fixes for the theme and the new option. (in the current state, with the material theme, the long text are never wrapped and goes outside the modal...)

Immagine 2021-01-29 190242


tapper82:
Can you make the modle grab focus for screenreaders. As it is now there is know way to know if the upgrade page has poped up with out arrowing all the way down to the botem of the page to see if a mesage has popped up.

Ansuel:
> Can you make the modle grab focus for screenreaders. As it is now there is know way to know if the upgrade page has poped up with out arrowing all the way down to the botem of the page to see if a mesage has popped up.

Can you better explain what you are saying ?

tapper82:
Hi yeah I will try and explain what I am on about. When using a screen reader and you are on the sysupgrade page in luci. After clicking on Flash image there is no way to know that the modle is asking for a response from the user. To the screen reader it seems like the button for Flash image does nothing. To find a way of interacting with the modle you have to jump all the way to the bottom of the page were you will here your screen reader say dialogue. When you here that you have to press the enter key and it will let the screen reader get focus on the actions for flashing your device. After clicking on the browse button to find the file to flash and then clicking on the upload button it then jumps you back up the page and you have to jump back down to the bottom of the page find where it says dialogue.  press enter. Then you can see the file name, size and things like that. After clicking on continue It then jumps you back on the page and there is no way for a screen reader to know that the router is flashing.

Ansuel:
Can you post some screen of the problem?

Il Sab 30 Gen 2021, 19:42 tapper82 <notifications@github.com> ha scritto:

tapper82:
The modle  thing can be fixt at a later date.

feckert:
What about the other themes? Don't they have to be adapted the css as well?

Ansuel:
Honestly I didn't check but they should be good... Will check

Il Gio 11 Feb 2021, 08:22 Florian Eckert <notifications@github.com> ha
scritto:

nickberry17:
I hope this gets merged soon

wulfy23:
did you test '-u' handling on ext4 systems?

Ansuel:
@feckert i finally had time to test this on every supported theme. I fixed all the minor problem

hnyman:
At the first glance it looks ok to me. 
I tested sysupgrading an invalid image for E8450 and after your changes it fails nicely and visibly.
Also the compat-version error is well visible.

image


Ps.
At that time I noticed that there is no line break between two lines of error messages, which makes the dialog a bit unclear, as the second timestamp starts right after the previous word without any punctuation. But looking at the source code (lines 240-244 that you did not touch), I think that it may come directly from stderr, so that a line break can't be easily inserted.

image

Ansuel:
Now that I notice I think it's redundant to have 2 times the same error?

Il Mer 4 Ago 2021, 21:16 Hannu Nyman ***@***.***> ha scritto:

hnyman:
hmmm.
some more testing with E8450 makes me to think that something is somehow broken.

* Normal sysupgrading of  r17254-edb6bc1990 with settings seems to cause the browser to hang with the spinning wheel, so that the new firmware does not load after the sysupgrdae. (Sysupgrade is still performed ok)
* If the option to "write installed packages" is selected, the sysupgrade does not even start. Not even the SSH console session is broken. (LuCI just remains with the spinning wheel)

Ansuel:
Could be that something is messed up? I refreshed and fixed the theme
issue. Will check what is wrong.
Anyway did you notice any missing feature? Or improvement?

Il Mer 4 Ago 2021, 21:41 Hannu Nyman ***@***.***> ha scritto:

hnyman:
I liked the new features in principle.

And the error condition checking seemed more complete than in the current LuCI.
 (I think that the compat-version handling at least may currently fail invisibly, so your changes are toward a better direction.)

Ansuel:
Anyway the not reloading bug was simple... investigating the problem you reported about sysupgrade not working with the '-k' option

Ansuel:
@hnyman there was a permission problem and i also disabled the button since it doesn't make sense to use the advanced backup config when is asked to reset the router on upgrade.

hnyman:
Thanks, tested with R7800 and now also the -k seems to work (after you fixed the ACL permission definitions).

hnyman:
Seems to work also in E8450
No comments from others in the meanwhile, so lets merge this and then see if there is any fallout.

PR #4757, opened 2021-02-01
[RFC] luci-proto-ipv6: split IPv6 proto into individual packages

feckert:
This change splits the ipv6 protocols into individual packages with the associated dependencies.
The package luci-proto-ipv6 contains all possible IPv6 protocol, which are not always necessary.
Additional it overloads the selection of available protocols.
Users who know what they want can also install them via the package management.

For compatibility reasons, I have left the luci-proto-ipv6 package as the meta package, so that nothing changes for the others.
The only difference is that now the netifd proto dependencies are also installed.
I am not sure if the meta package is necessary?


hmh:
This is about IPv6 *transitioning* protocols (as opposed to "native" IPv6), really.

If the overhead of multiple packages is too annoying, it could also be a KConfig option to enable/disable inclusion of each transition "proto" (464xlat, 6rd...) into the main luci-proto-ipv6 package...

feckert:
@hmh Interesting comment, I'll have to take another look at that.

@jow- @aparcar any comments from your side.

feckert:
@hnyman @jow- Could you please give your opinion on this? The RFC pullrequest is currently rotting.
I personally think the selection of protocols when inserting an interface confusing.
Especially since a normal user does not need the whole IPv6 protocol.
Therefore I would do the same as with IPv4 and only show dhcp and static as default.

Someone who needs others can install them later.
He must also be familiar with them.

So that users who know the luci-proto-ipv6 package do not have to change, I have left it as a meta package.
This meta package installs all protocols as before.

systemcrash:
Does this mean that if I pull in (more specific packages) `6in4`, it pulls in the whole suite, or same as before, only what I need?

feckert:
@aparcar For IPv4 we have static and dhcp the same I would have seen for IPv6. 

But currently other IPv6 related protocols are additional installed with the package luci-proto-ipv6.
I, for example, use none of them and most protocols do not tell me anything about what is installed with luci-proto-ipv6.
To get a better user experience and not to overload the user with protocols, this P/R suggest to split the luci-proto-ipv6 into individual package.

systemcrash:
I'm OK with this if any other reviewers are.

systemcrash:
No other comments. Closing. Re-open if this is interesting.

PR #4762, opened 2021-02-04
luci-app-banip: sync with banIP 0.7.0

dibdot:
* sync with current banIP release
* move app to client side JS
* tested with all standard themes


dibdot:
See https://github.com/openwrt/packages/pull/14649 for reference

PR #4763, opened 2021-02-04
luci-theme-openwrt-2000: Add doctype to header

bryanforbes:
Without a doctype, the browser is put into quirks mode. In quirks mode, Webkit (Safari on macOS and iOS) ignores `display: block;` for `<td>` elements which causes the firewall page (among others) to display improperly.

**Without doctype:**
image

**With doctype:**
image


jow-:
Noticed the same, thank you!

PR #4768, opened 2021-02-05
luci-app-statistics: fix config of ipstatistics

PolynomialDivision:
The config view did not return a configSummary resulting in an error:
  plugin.form.configSummary is not a function

This commit fixes the error.


hnyman:
Also, the config does not actually get applied into collectd, as the config generation definition is also missing from /usr/share/luci/statistics/plugins/ipstatistics.json

I am currently authoring and testing my own fix.

(I think that I will also remove the unnecessary config submenu, as there are no options to be selected)

hnyman:
Fix for the issue applied as 5da8a0d2

PR #4771, opened 2021-02-06
luci-base: add validation for APN datatype

nickberry17:

Adds a new datatype/validation for **APN** ([Access Point Name](https://en.wikipedia.org/wiki/Access_Point_Name))

bad apn

Use like this:
```
o = s.taboption('general', form.Value, 'apn', 
			_('APN'),
			_("Setting the appropriate APN for your SIM and data subscription is recommended."));
o.datatype = 'apn';
```

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/validation.js
+++ b/modules/luci-base/htdocs/luci-static/resources/validation.js
@@ -569,6 +569,13 @@ var ValidatorFactory = baseclass.extend({
 				_('hexadecimal encoded value'));
 		},
 
+		apn: function() {
+			if (this.value.length <= 63)
+				return this.assert(this.value.match(/^[a-zA-Z0-9\-.]*[a-zA-Z0-9]$/), _('valid APN'));
+
+			return this.assert(false, _('valid APN'));
+		},
+
 		string: function() {
 			return true;
 		}


hnyman:
Not sure about JS performance optimization logic, but I am not sure if it is wise to define this new rarely used validation function as the first one. 

I would move the definition further down the possible evaluation queue.

nickberry17:
> Not sure about JS performance optimization logic, but I am not sure if it is wise to define this new rarely used validation function as the first one.
> 
> I would move the definition further down the possible evaluation queue.

Excellent point, I will move it down to the bottom, or near the bottom of the list.

nickberry17:
> Hmm, I think an APN datatype is a little bit too specific to go into the validation module. I'm slightly in favor of doing a custom validator function instead:
> 
> ```js
> o.validate = function(section_id, value) { 
>     if (!/^[a-zA-Z0-9\-.]*[a-zA-Z0-9]$/.test(value))
>         return _('Invalid APN provided');
> 
>     return true;
> };
> ```

I guess that is doable, but I thought it would fit here because `phonedigit` is included.  What is your reservation for expanding the scope of the built-in validators based on?  Are we trying to keep the scope limited to IP-based networking and system administration tasks?  If so, then should we remove `phonedigit`?  If this gets included then luci-proto-modemmanager, luci-proto-qmi, and luci-proto-3g can use it straight away.

PR #4790, opened 2021-02-09
luci-mod-network: add dynamic vlan configurability

thomasthep:
Add the possibility to configure dynamic vlan when using radius.


hnyman:
New capabilities should be introduced to master, while you are proposing it to be added to the old stable 19.07 branch

thomasthep:
My build is based on 19.07
I'll move it over.

thomasthep:
@hnyman there you go.

Neustradamus:
@thomasthep: Have you progressed on it?

thomasthep:
@Neustradamus unfortunately I haven't had a chance to make it fancy. My access points are currently running the patch and all work as expected. The configuration might not be as straight forward as one would hope. 😅

What's your use case?

systemcrash:
This already exists in master (via another PR I think).

PR #4799, opened 2021-02-10
lua-base: dispatcher lua not ignore action type "call" on looking for subpages

MuteSpirit:
Function "find_subnode" in dispatcher.lua script of "luci-base" module recognizes all LuCI handler actions besides "call" type.

As a result path rendered by handler with "call" action
    cannot be opened just after login on WebUI,
      even if such path has high order (1) among others
         in /usr/share/luci/menu.d/*.json files

Add check "call" action type in logic of "find_subnode" function similar
to other standard LuCI handlers: "view", "cbi", "form", etc

### DIFF
--- a/modules/luci-base/luasrc/dispatcher.lua
+++ b/modules/luci-base/luasrc/dispatcher.lua
@@ -636,6 +636,7 @@ local function find_subnode(root, prefix, recurse, descended)
 
 	if descended then
 		if not recurse or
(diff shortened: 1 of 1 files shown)


jow-:
This was actually done on purpose since `call()` targets usually output non-human readable data such as JSON feeds. We want to avoid a situation where `firstchild()` or similar actions direct the user to a non-HTML view.

PR #4802, opened 2021-02-10
nextdns: remove deprecated hardened privacy option

rs:


RouterCFW:
Sadness I loved this option

PR #4814, opened 2021-02-16
[Bug-Fix] luci-app-firewall: add wan6 support

prateekrastogi:
Firewall zones Forwarding was not working properly if `wan` was disabled, and `wan6` is enabled with both IPv4 and IPv6 Active Upstreams.


jow-:
This change makes no sense at all. It simply changes the default value for the source zone of new port forwards to "wan6", which is unsuitable for the vast majority of users.

PR #4819, opened 2021-02-18
luci-{app,proto}-wireguard: remove kmod-wireguard

lipnitsk:
Prepares for 5.10 migration. wireguard-tools will bring in the correct
wireguard kernel module dependency - either kmod-wireguard or
kmod-wireguard-oot.

Depends on openwrt/openwrt#3885

### DIFF
--- a/applications/luci-app-wireguard/Makefile
+++ b/applications/luci-app-wireguard/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=WireGuard Status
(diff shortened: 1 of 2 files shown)


lipnitsk:
I'm going to close this for now, there has been quite a bit of discussion on how to proceed with WireGuard going forward, but I think the idea is to limit disruption to user space packages that depend on kmod-wireguard. So, no changes should be needed here.

hnyman:
I wonder if this is PR again relevant now when https://github.com/openwrt/openwrt/commit/cbcddc9f318607881799e329b327a68c4e76d5cb has been merged

lipnitsk:
Yeah I think it's worth removing the unnecessary dependency.

PR #4833, opened 2021-02-24
[WIP] luci-app-firewall: add ipsets handling

feckert:
The ipset handling is supported by netifd but is not yet integrated in the LuCI. 
This adds the subpage ipsets to luci-app-firewall, this was the result from the [discussion](https://github.com/openwrt/packages/issues/14718#issuecomment-782953570). The validation handling for the options is still missing.


feckert:
This was replaced by fw4 and is not needed anymore

PR #4837, opened 2021-02-25
rpcd-mod-luci: Fix parsing of DUID-LLT's in duid2ea

nwidger:
Fix parsing of DUID-LLT's in duid2ea.  Unlike DUID-LL's, DUID-LLT's have a 32-bit time field between the hardware type and link-layer address fields, see RFC 3315 Sections 9.2 and 9.4:

https://tools.ietf.org/html/rfc3315#section-9.2
https://tools.ietf.org/html/rfc3315#section-9.4

therefore, the link-layer address starts at offset 16 instead of 8.

Signed-off-by: Niels Widger <niels@qacafe.com>


jow-:
Merged, thank you!

PR #4846, opened 2021-02-28
luci-static/luci.js: Fix a handful of eslint errors

nikolas:
Fix a handful of potential javascript problems in luci.js, found with eslint.

Signed-off-by: Nikolas Nyby <nikolas@gnu.org>


aparcar:
I think this makes more sense if we introduce a `.eslintrc.json` file as well, e.g.

```json
{
    "env": {
        "browser": true,
        "commonjs": true,
        "es2021": true
    },
    "extends": "eslint:recommended",
    "parserOptions": {
        "ecmaVersion": 12
    },
    "rules": {
    }
}
```

@jow- thoughts on linting?

Issue #4839, opened 2021-02-26
Adding system description (short) and system notes (long) fields to UCI system section


hmh:
I'd like to propose adding two new fields to UCI's system config.   I am also offering to do all the work required to implement the required changes in LuCI (for all branches) as well as updating the openwrt wiki.

This is actually filling an usability niche we detected, and which would be much better addressed in openwrt itself than in a downstream fork.

If the proposal below is acceptable, I'll submit the PRs.  Otherwise, I'd like to know what should be changed so that we can have equivalent functionality in OpenWRT upstream.  If this should be handled through the mailing lists instead, please say so.  I sent it here because it is 99,9% a LuCI / UI matter.

**system.@system[0].description = 'an optional, short description for this device, one line, some reasonable char limit'**

The "description" option is purposefully short, so that it can be used e.g. on UIs to help describe this device in a selector.  The  hostname is often not available for such purposes, or not descriptive enough, and it is better not not overload it with too much functionality.

It would use the LuCI "Value" option type, so embedded newlines would be discarded/removed.

**system.@system[0].notes = 'an optional text for notes about this device in free-form,
such as install notes, location notes, serial number of the unit,
maybe inventory numbers, or telephone number of tech support'**

The "notes" option is multi-line, and can be used to store general free-form metadata *for human consumption*.  It would use the LuCI "TextValue" option type.

These two new options are very simple and very safe to implement, as they have no side-effects on the rest of the system (well, it adds a few i18n strings, but that's it) and require no specific handling.

The intended usage is to just show them in the LuCI web interface, and to get/set them over ubus uci from remote applications (such as mobile apps, etc).  It would be handy for a future "remote LuCI" selector that has a list of known devices, etc.

Besides implementing them and submitting them through a PR for LuCI (for all branches that are still accepting PRs), I would do the work to document them (should they be accepted) officially in the main wiki in pages where UCI system is described, so that one can actually extend packages to use system.@system[0].description where it make sense to do so, and to avoid duplicate work.

Edit: proposed option "annotations" renamed to "notes".

jow-:
Fine with me.

hmh:
Apparently, "notes" is a more appropriate word choice than "annotations" for the intended usage of the field I proposed, so I am renaming it from "annotations" to "notes".

I will submit PRs for master/trunk, 21.02, 19.07 and 18.06 -- we will also have one for 15.05, but I doubt it makes sense to submit it as a PR.

hmh:
[...]

PR #4859, opened 2021-03-01
[18.06] luci-mod-admin-full: implement system.description, system.notes

hmh:
Implement two new text "options" for UCI system config, intended to
help humans describe the device.

"system.description" is a short, single-line description suitable for
selector UIs in remote administration applications, or remote UCI (over
ubus RPC), etc.  It would also be suitable as a default for LLDP/SNMP
"system description".

"system.notes" is a multi-line, free-form text field that can be used in
any way the user wishes, e.g. to hold installation notes, or unit serial
number and inventory number, location, etc.

Signed-off-by: Henrique de Moraes Holschuh <henrique@nic.br>

### DIFF
--- a/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/system.lua
+++ b/modules/luci-mod-admin-full/luasrc/model/cbi/admin_system/system.lua
@@ -40,6 +40,14 @@ function o.write(self, section, value)
 end
 
 
+-- could be used also as a default for LLDP, SNMP "system description" in the future
+o = s:taboption("general", Value, "description", translate("Description"), translate("An optional, short description for this device"))
+o.optional = true
+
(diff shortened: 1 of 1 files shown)


hmh:
Related to #4839

hnyman:
As 18.06 is already deprecated, I am not sure that this should be merged there, any more.

hmh:
The luci branch still gets updates, so why not?  It is an 100% safe change, and it actually makes things easier by keeping the divergence against 19.07 and 21.02 the same if you merge it into 18.06...

hnyman:
> The luci branch still gets updates,

Well, the only changes since December (when 18.06 was officially deprecated) are @stangri's updates for his apps.

And nothing i18n  has been backported since then.

To my knowledge also the 18.06 buildbot has been turned off, so the downloadable packages are not updated any more.

hmh:
Ok, makes sense.

Thank you!

PR #4860, opened 2021-03-01
rpcd-mod-luci: Return array of addresses in getHostHints

### LINKED ISSUE #4838: rpcd-mod-luci: Return array of IPv4/IPv6 addresses in getHostHints
The getHostHints method in rpc-mod-luci seems to go to a lot of trouble to find all IP addresses associated with hosts connected to OpenWRT, but the response only allows returning one IPv4 and one IPv6 address per host.  There doesn't seem to be any prioritization of which addresses are returned if a host has multiple IPv4 or (more likely) multiple IPv6 addresses, instead it seems to be the case that the first address found is the one returned.  This can lead to inconsistent results.  For instance if a host has both a SLAAC and DHCPv6 IPv6 address, getHostHints may return a different IPv6 address for the host depending on which address appears first in the ND table.

Was it a conscious decision to only return one address per address family?  Would it be possible to add new i.e. `ipv4_addrs` and `ipv6_addrs` array fields which return _all_ IPv4/IPv6 addresses found for a host?  If so I'd be happy to take a stab at this.

nwidger:
Update luci-rpc's getHostHints method to return two string arrays for
each host, `ipaddrs` and `ip6addrs`, each containing the host's IPv4
and IPv6 addresses, respectively.  Each array is sorted by a priority
derived from the source from which the address was discovered.  The
current address sources and their priority is as follows (a
higher (larger) priority is listed first):

    - neighbor table entries: 10
    - /etc/ethers entries:    50
    - DHCP leasefile:        100
    - RRDNS queries:         100
    - getifaddrs():          200
    - UCI static leases:     200

The existing `ipv4` and `ipv6` string fields for each host in
`getHostHints` has been removed.  Downstream users of getHostHints
still need to be updated.

Fixes #4838.

Signed-off-by: Niels Widger <niels@qacafe.com>

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js
@@ -395,22 +395,30 @@ return baseclass.extend({
 		var choice_values = [], choice_labels = {};
 
 		if (!family || family == 'ipv4') {
-			L.sortedKeys(hosts, 'ipv4', 'addr').forEach(function(mac) {
-				var val = hosts[mac].ipv4,
-				    txt = hosts[mac].name || mac;
+			L.sortedKeys(hosts).forEach(function(mac) {
+				if (hosts[mac].ipaddrs && hosts[mac].ipaddrs.length) {
+					hosts[mac].ipaddrs.forEach(function(ip) {
+						var val = ip,
+							txt = hosts[mac].name || mac;
 
-				choice_values.push(val);
-				choice_labels[val] = E([], [ val, ' (', E('strong', {}, [txt]), ')' ]);
+						choice_values.push(val);
+						choice_labels[val] = E([], [ val, ' (', E('strong', {}, [txt]), ')' ]);
+					});
+				}
 			});
 		}
 
 		if (!family || family == 'ipv6') {
-			L.sortedKeys(hosts, 'ipv6', 'addr').forEach(function(mac) {
-				var val = hosts[mac].ipv6,
-				    txt = hosts[mac].name || mac;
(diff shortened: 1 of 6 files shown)


nwidger:
@jow- I updated all the frontend code I could fine that uses luci-rpc getHostHints.  I wasn't sure in some cases whether the code should be updated to just use the first address in the list or whether to use them all (for in instance code that is generate dropdown choices).  I tried to make an educated guess but please let me know if the behavior should be changed.  I also tried to respect the indentation and coding style as much as possible, but please let me know if I'm doing anything I shouldn't.

Finally, I'm not sure if there's any documentation that needs to be updated for this change.  I found `docs/jsapi/LuCI.network.Hosts.html`, but it doesn't seem to require any updates.  I also found `docs/jsapi/network.js.html`, which seems to be machine generated (perhaps by `build/makedocs.sh`?).  Please let me know if that something I need to do to get this merged.

nwidger:
@jow- Just a ping to see if you've had a chance to look over these changes or if there's anything else that should be in this PR.

jow-:
Hi @nwidger - I noticed your updates already but didn't get around to check them out yet. Will keep you posted.

nwidger:
Hey @jow- it's been a while since I heard from you.  Just wondering if there's still any hope for getting this PR merged at some point.

jow-:
@nwidger - I merged your PR after some further rework in a5195e7825cbbc3942ca6e571ae0020c1cf080c4 and c7b7b42cd3840cfd67f412191578a8659eb63c87. Thanks a lot for your work on this.

nwidger:
@jow- Fantastic!  Thank you so much for fixing it up and getting it in. :) Do you think this will make it into the 21.02 release, or is it too late for that?

Issue #4506, opened 2020-10-08
luci-app-ddns: Time error


small-5:
The update time displayed by luci-app-ddns of the master branch is incorrect, which is inconsistent with the time displayed in the log file.
Version:git-20.280.43255-6483e80

feckert:
You have to give us more information

small-5:
The system time of my screenshot is 2020-10-08 20:13, but Last Update shows 2020-10-08 20:22.
![1](https://user-images.githubusercontent.com/61473216/95458187-69d84600-09a4-11eb-858d-23d60549ba8f.jpg)

feckert:
That means the time on your windows is different then the time on your router.
Have you set an ntp server?

small-5:
The router time is consistent with the windows time. The problem is not here, the time displayed by luci is not consistent with the updated time in the log file.

castillofrancodamian:
I can attest that the same thing happens here. The time is wrong.

Ansuel:
@feckert in theory the time from luci is taken directly from the script... If there is a bug it should be there. Did you found a way to reproduce this?

jow-:
Maybe a numeric overflow?

lorthirk:
I think that the timezone is calculated twice, somehow:

![Screenshot 2021-05-18 at 09 34 05](https://user-images.githubusercontent.com/8140139/118610656-4be58600-b7bc-11eb-9bdf-7372e7af97a8.png)

According to logs, last check happened at 09:25 and next one is due in 72 minutes, so at 10:37. I am GMT+1 (although with DST now, so it should be GMT+2).

PR #4864, opened 2021-03-02
luci-app-ddns: Fix time error

small-5:
Fix https://github.com/openwrt/luci/issues/4506

### DIFF
--- a/applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns
+++ b/applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns
@@ -109,12 +109,8 @@ local methods = {
 					tonumber(s["force_interval"]) or 72,
 					s["force_unit"] or "hours" )
 
-				-- process running but update needs to happen
-				-- problems if force_seconds > uptime
-				force_seconds = (force_seconds > uptime) and uptime or force_seconds
-
 				if last_update > 0 then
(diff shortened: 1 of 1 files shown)


small-5:
Check the dynamic_dns_updater.sh file and find that FORCE_SECONDS did not update the IP after the next update time, so line 114 should be retained.

small-5:
Check the dynamic_dns_updater.sh and find that FORCE_SECONDS did not update the IP after the next update time, so line 114 should be retained.

small-5:
Line 114 should be retained.

feckert:
Can your please sign your commit with sign-off-by?
And in the commit message explain in more detail why we don't need this.
My problem is that I didn't write the code and don't know what the original developer was thinking.
If we change it now, then later you wonder why you changed it.

Lets ping @Ansuel maybe he could give us more info

small-5:
I can't with sign-off-by now, because fork has been deleted. The removal of "+ force_seconds" is to compare the 18.06 version of ddns.lua (https://github.com/openwrt/luci/blob/openwrt-18.06/applications/luci-app-ddns/luasrc/tools/ddns.lua), this can be fixed The bug of the last time overflow. The 114 line is removed because this will cause the bug that the next update time is wrong when it is just started. You can try to modify this file in my way, and then service rpcd restart to verify.

PR #4873, opened 2021-03-04
luci-app-ddns: Fix time error

small-5:
Fix https://github.com/openwrt/luci/issues/4506
The removal of "+ force_seconds" is to compare the 18.06 version of ddns.lua (https://github.com/openwrt/luci/blob/openwrt-18.06/applications/luci-app-ddns/luasrc/tools/ddns.lua), this can be fixed The bug of the last time overflow. 
The 114 line is removed because this will cause the bug that the next update time is wrong when the router is just started. 
In addition,I checked dynamic_dns_updater.sh and found that next_update should be "epoch + force_seconds + check_seconds", so I added the variable definition of check_seconds. Through the test, the time displayed by luci matches the time of the log file.

### DIFF
--- a/applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns
+++ b/applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns
@@ -109,15 +109,15 @@ local methods = {
 					tonumber(s["force_interval"]) or 72,
 					s["force_unit"] or "hours" )
 
-				-- process running but update needs to happen
-				-- problems if force_seconds > uptime
-				force_seconds = (force_seconds > uptime) and uptime or force_seconds
+				local check_seconds = calc_seconds(
+					tonumber(s["check_interval"]) or 10,
+					s["check_unit"] or "minutes" )
 
 				if last_update > 0 then
-					local epoch = os.time() - uptime + last_update + force_seconds
+					local epoch = os.time() - uptime + last_update
 					-- use linux date to convert epoch
 					converted_last_update = epoch2date(epoch,date_format)
-					next_update = epoch2date(epoch + force_seconds)
+					next_update = epoch2date(epoch + force_seconds + check_seconds)
 				end
 
 				if pid > 0 and ( last_update + force_seconds - uptime ) <= 0 then


feckert:
Now it looks better.
Can you please update the commit header `luci-app-ddns: fix update time calculation` and write in the [commit message](https://github.com/openwrt/luci/pull/4873/commits/82bf7fc5e7ed27ac7ed228ad1ca739f0d121d164) what you wrote in pullrequest description but in shortened form?

small-5:
@feckert Can you please merge this pull request?

feckert:
@small-5 sorry but the `Signed-off-by` does not fit
Real name and a valid email must be set.

applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns
```diff
@@ -109,15 +109,15 @@ local methods = {
...
+					tonumber(s["check_interval"]) or 10,
+					s["check_unit"] or "minutes" )
 
 				if last_update > 0 then
-					local epoch = os.time() - uptime + last_update + force_seconds
+					local epoch = os.time() - uptime + last_update
 					-- use linux date to convert epoch
 					converted_last_update = epoch2date(epoch,date_format)
-					next_update = epoch2date(epoch + force_seconds)
+					next_update = epoch2date(epoch + force_seconds + check_seconds)
```
feckert, line 120: @small-5 I would say it fits.
@Ansuel I just noticed, why didn't we also specify date_format as the second argument here?

apocalypse1233:
I have same problem. Last Update is after the last/current date about 30 minutes and increases with every minute that passes. 

Next Update time is also wrong (it's 58 minutes higher than last update) and is increasing by minute.

Running latest snapshot. I will try this patch.

applications/luci-app-ddns/root/usr/libexec/rpcd/luci.ddns
Ansuel: I need to check the script

feckert:
@small-5 thanks for the fix
@apocalypse1233 thanks for testing.

I will add the change now, since it fixed the problem with the wrong display.

PR #4895, opened 2021-03-10
luci-app-vpnbypass: transition to client-side rendering

stangri:
This is the first luci app I'm transitioning, I would appreciate more eyes on this (@Ansuel @dibdot).


applications/luci-app-vpnbypass/root/usr/libexec/rpcd/luci.vpnbypass
```diff
@@ -0,0 +1,100 @@
...
+	json_add_object "$name"
+	json_add_boolean 'enabled' "$(is_enabled "$name")"
+	json_add_boolean 'running' "$(is_running "$name")"
+	json_close_object
+	json_dump
+	json_cleanup
+}
+
+set_init_action() {
+	local name="$1" action="$2" cmd
```
jow-, line 30: The `$name` variable here allows for arbitrary shell command injection. Please validate it through something like `[ -f "/etc/init.d/$name" ]` and return something like `{ "error": "Init script not found" }` if the condition is not satisfied.

applications/luci-app-vpnbypass/root/usr/share/rpcd/acl.d/luci-app-vpnbypass.json
```diff
@@ -2,38 +2,16 @@
...
-				],
-				"/usr/sbin/grep *": [
-					"exec"
-				],
-				"/usr/sbin/iptables *": [
-					"exec"
+			"ubus": {
+				"luci.vpnbypass": [
+					"getInitList",
+					"setInitAction",
```
jow-, line 8: The `setInitAction` method has side-effects (allows altering system state), please move it from the `"read": { ... }` to the `"write": { ... }` scope.

jow- (changes requested):
Good work! Ive two remarks we should address before merging.

PR #4897, opened 2021-03-10
luci-app-advanced-reboot: explicit package version

stangri:


jow-:
Why is this needed?

stangri:
> Why is this needed?

To ease installation of newer (test) packages from my own repo. Ever since the luci package versioning switched to the git commit stamp, the usual `opkg update; opkg upgrade <package-name>;` stopped working.

PR #4912, opened 2021-03-11
luci-theme-openwrt-2020: add missing css success color

### LINKED ISSUE #4734: luci-app-mwan3 with theme openwrt2020 missing green box background in status for active links
On a local build of master branch with luci-theme-bootstrap the overview page shows green background boxes for active links, if I switch to theme openwrt2020 then both active and offline interfaces have the same brown background

feckert:
This fixes #4734

I would also cherry-pick this to openwrt-21.02 if this works for @ajayramaswamy you.


hnyman:
Nitpicking, but the commit title (and the PR title) talks about `luci-theme-openwrt`, while it actually concerns another theme, `luci-theme-openwrt-2020`

ajayramaswamy:
Yes I confirm that this works for me. I put the cascade.css on the router directly and now the color is correct. Thank you.

PR #4914, opened 2021-03-12
[WIP] luci-base: prevent empty field for adding new entry

he-ma:
"Old" LUA modules using "cbi/tblsection"  or "cbi/tsection" as template show an "Add" Button with a preceeding input field for a name if the `addremove` attribute is set. This -in a different manner- is the same for the "new" JavaScript modules, too.

This sometimes leads to misunderstandings if the input field is empty and the button is pressed (as nothing happens exept for reloading the page in browser, but no new section is added, or, for the JavaScript by opening a modal which doesn't save then).

This patch visually marks the field red (like other fields which have the `optional = false` attribute set) and shows an alert window (in LUA modules) when submitting an empty value. In JavaScript just nothing will happen.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/cbi.js
+++ b/modules/luci-base/htdocs/luci-static/resources/cbi.js
@@ -350,7 +350,7 @@ function cbi_init() {
 function cbi_validate_form(form, errmsg)
 {
 	/* if triggered by a section removal or addition, don't validate */
-	if (form.cbi_state == 'add-section' || form.cbi_state == 'del-section')
+	if (form.cbi_state == 'del-section')
 		return true;
 
 	if (form.cbi_validators) {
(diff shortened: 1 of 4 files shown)


he-ma:
@jow- Could you have a look at my patches and tell me whether it looks OK for you.

Thanks
Helge

he-ma:
@jow- 
I just experienced that with my patch the "Add" and "Save & Apply" work well, but only applying the "Save" always results in the "Some fields are invalid, cannot save values" alert.

I came up to the result, that the function `cbi_validate_form()` is called but without a `form.cbi_state` and lastly this results in the error.

Could you give me an advice how I could better or more general handle this issue?

Thank you.

feckert:
This P/R for the javascript is superseded by [P/R](https://github.com/openwrt/luci/pull/5137/commits/8298936b38e2103cd4870cc4b110424855c6aa43)

Issue #4877, opened 2021-03-05
RFC: LuCI failed/successful logging to syslog


ja-pa:
Hi,
I'm not using uhttpd on my router and probably because of that failed/succesfull login for LuCI are not in syslog. (Similar issue is  https://forum.openwrt.org/t/luci-code-that-handles-failed-and-success-login-results-to-syslog/83776)

I fixed that with this patch but I'm not sure if this is would be accepted in upstream because it will result in double error login in syslog.

I'm looking for idea how to improve it and will be glad for any suggestions.

```
--- a/modules/luci-base/luasrc/dispatcher.lua
+++ b/modules/luci-base/luasrc/dispatcher.lua
@@ -565,12 +565,16 @@ local function session_setup(user, pass)
 
 		io.stderr:write("luci: accepted login on /%s for %s from %s\n"
 			%{ rp, user or "?", http.getenv("REMOTE_ADDR") or "?" })
+		nixio.syslog("info","luci: accepted login on /%s for %s from %s\n"
+			%{ rp, user or "?", http.getenv("REMOTE_ADDR") or "?" })
 
 		return session_retrieve(login.ubus_rpc_session)
 	end
 
 	io.stderr:write("luci: failed login on /%s for %s from %s\n"
 		%{ rp, user or "?", http.getenv("REMOTE_ADDR") or "?" })
+	nixio.syslog("info","luci: failed login on /%s for %s from %s\n"
+		%{ rp, user or "?", http.getenv("REMOTE_ADDR") or "?" })
 end

```

feckert:
We could replace the `io.stderr:write` in alle source files with [perror](https://github.com/openwrt/luci/blob/bd40525a69fe591b5ab487df25238401196cb77a/libs/luci-lib-base/luasrc/util.lua#L136) and decide with a config option if we want to log to stderr or syslog or check if we use nginx or uhttpd

PR #4925, opened 2021-03-16
luci-base: allow direct syslog login for perror

ja-pa:
This PR replaces io.stderr:write with perror in dispatcher.lua and modifies perror in such way, that in can send error messages to stderr or syslog based on value stored in /etc/config/luci

```
luci.main.log='syslog'
or
luci.main.log='stderr'
```
Fixes https://github.com/openwrt/luci/issues/4877

Signed-off-by: Jan Pavlinec <jan.pavlinec@nic.cz>

### DIFF
--- a/modules/luci-base/luasrc/dispatcher.lua
+++ b/modules/luci-base/luasrc/dispatcher.lua
@@ -563,15 +563,13 @@ local function session_setup(user, pass)
 			ubus_rpc_session = login.ubus_rpc_session,
 			values = { token = sys.uniqueid(16) }
 		})
-
-		io.stderr:write("luci: accepted login on /%s for %s from %s\n"
-			%{ rp, user or "?", http.getenv("REMOTE_ADDR") or "?" })
+		nixio.syslog("info", tostring("luci: accepted login on /%s for %s from %s\n"
+			%{ rp, user or "?", http.getenv("REMOTE_ADDR") or "?" }))
 
 		return session_retrieve(login.ubus_rpc_session)
 	end
-
-	io.stderr:write("luci: failed login on /%s for %s from %s\n"
-		%{ rp, user or "?", http.getenv("REMOTE_ADDR") or "?" })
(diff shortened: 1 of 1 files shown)


jow-:
Why is this needed?

ja-pa:
This change is useful mainly in case that someone is running luci without uhttpd. Usually in that case LuCI error messages are not sent to syslog.(see https://forum.openwrt.org/t/luci-code-that-handles-failed-and-success-login-results-to-syslog/83776 ) 
This change adds the possibility to send error messages directly to syslog but still preserves previous behavior.

ja-pa:
Rebased on master
@feckert @jow- What do you think about this change?

feckert:
From my point of view, that's OK.
If there are no objections @jow- then I will merge this the next few days.

I can't find it now but can we document the new option somewhere?
I'm not sure if we already have something in the openwrt wiki aboutt LuCI

jow-:
tbh I am not overly happy with the added complexity just to send two messages to syslog. Feel free to replace the stderr writes with nixio.syslog() invocations but do not introduce new options and alternative code paths for this.

modules/luci-base/luasrc/dispatcher.lua
```diff
@@ -564,13 +564,13 @@ local function session_setup(user, pass)
 			values = { token = sys.uniqueid(16) }
 		})
 
-		io.stderr:write("luci: accepted login on /%s for %s from %s\n"
```
jow-, line 567: Please replace with `nixio.syslog()` and get rid of the extra logic.

jow- (changes requested):

modules/luci-base/luasrc/dispatcher.lua
ja-pa: Replaced with syslog call.

ja-pa:
I removed additional logic and replaced it with nixio.syslog. 
Run tested on my Omnia

feckert:
Thanks

PR #4930, opened 2021-03-17
[Backport] luci-mod-network: issue with breakdown of dnsmasq after duplication o…

opastushkov:
…f static IP was fixed

Before this commit, assigning the same static IP address to two
different hosts disabled dnsmasq.

Logic of adding a new static lease was modified. If user try to assign a
new MAC address to already reserved IP, old lease will be modified (list
of MAC addresses will be extended by new MAC) instead of creation a new lease with the same IP.

Signed-off-by: Oleksandr Pastushkov <oleks.pastushkov@gmail.com>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -469,6 +469,25 @@ return view.extend({
 			so.value(mac, hint ? '%s (%s)'.format(mac, hint) : mac);
 		});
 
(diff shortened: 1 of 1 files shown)


BKPepe:
This seems to be a backport and it is missing in the commit description that it was cherry-picked from the commit. Can you please add it to the commit or use `git cherry-pick -x hash`?

feckert:
@opastushkov This was reverted in master with commit https://github.com/openwrt/luci/commit/91f46d765b36825fcfe63e2b7d4863de00cdc69d
Since it breaks something else. Can you look at it again?

PR #4954, opened 2021-04-01
i18n: enable translation sync for battstatus/dashboard modules

musashino205:
This patch series add 2x modules to build/mkbasepot.sh and sync translations.

BTW, i18n-scan.pl reports a warning ```standard input:56: warning: RegExp literal terminated too early``` on luci-mod-battstatus

```
$ ./build/i18n-scan.pl modules/luci-mod-battstatus/
standard input:56: warning: RegExp literal terminated too early
msgid ""
msgstr "Content-Type: text/plain; charset=UTF-8"

...
```

### DIFF
--- a/modules/luci-mod-battstatus/po/ar/battstatus.po
+++ b/modules/luci-mod-battstatus/po/ar/battstatus.po
@@ -1,0 +1,21 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+"Project-Id-Version: \n"
+"Last-Translator: Automatically generated\n"
+"Language-Team: none\n"
+"Language: ar\n"
+"MIME-Version: 1.0\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: modules/luci-mod-battstatus/htdocs/luci-static/resources/preload/battstatus.js:26
+msgid "Charging"
+msgstr ""
+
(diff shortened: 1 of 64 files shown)


jow-:
Hmm, I'm not sure I want to pull strings from mod-dashboard and mod-battstatus into the base translation, since these components are usually not installed by default.

Is there anything preventing us from letting these components have their own translations? Afair it should be enough to initialize a `po/templates/xy.pot` file for them once to have them recognized. Even touching an empty `.pot` file and running `./build/i18n-sync.sh` should be enough.

musashino205:
Thank you for your comment,

> Is there anything preventing us from letting these components have their own translations?

indeed, I think that is better too...

musashino205:
@jow- I prepared [another version](https://github.com/musashino205/luci/commits/i18n/batt-dash_ownpo) for own translations of battstatus/dashboard, should I update this PR to it?

hnyman:
> should I update this PR to it?

Please do that. Much more sensible to have the translations separate from the "base" itself.

musashino205:
>>should I update this PR to it?
>
>Please do that. Much more sensible to have the translations separate from the "base" itself.

I undarstand, thank you. I updated this PR.

hnyman:
I had to manually add the new "translation component" in weblate for it to see battstatus strings. 
Probably due to these being "modules", not "applications", so it is not detected automatically.

I will do the same for dashboard.

PR #4956, opened 2021-04-01
luci-mod-network: Add more comprehensive help text for odhcpd IPv6 parameters, plus introduced RA + NDP options to GUI. 

systemcrash:
For discussion, refer to [forum thread](https://forum.openwrt.org/t/clarifying-ipv6-luci-terminology-ndp-proxy-ula-prefix-server-mode-relay-mode-hybrid-mode/18743)

Also added some RA parameters for increased GUI based configuration flexibility and help-text.

Until now, some of the IPv6 settings are mysterious or cryptic. They are now
clarified, having referred to odhcpd README and source. 
The information is accessible to IPv6 beginners.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -565,6 +565,50 @@ return view.extend({
 					so.value('relay', _('relay mode'));
 					so.value('hybrid', _('hybrid mode'));
 
+					so = ss.taboption('ipv6', form.Value, 'ra_maxinterval', _('Max <abbr title="Router Advertisement">RA</abbr> interval'), _('Maximum time allowed \
+						between sending unsolicited <abbr title="Router Advertisement, ICMPv6 Type 134">RA</abbr>. Default is 600 seconds (<code>600</code>).'));
+					so.optional = true;
+					so.default = '600';
+					so.depends('ra', 'server');
+					so.depends('ra', 'hybrid');
+					so.depends('ra', 'relay');
+
+
+					so = ss.taboption('ipv6', form.Value, 'ra_mininterval', _('Min <abbr title="Router Advertisement">RA</abbr> interval'), _('Minimum time allowed \
+						between sending unsolicited <abbr title="Router Advertisement, ICMPv6 Type 134">RA</abbr>. Default is 200 seconds (<code>200</code>).'));
+					so.optional = true;
+					so.default = '200';
+					so.depends('ra', 'server');
+					so.depends('ra', 'hybrid');
+					so.depends('ra', 'relay');
+
+					so = ss.taboption('ipv6', form.Value, 'ra_lifetime', _('<abbr title="Router Advertisement">RA</abbr> Lifetime'), _('Router Lifetime published \
+						in <abbr title="Router Advertisement, ICMPv6 Type 134">RA</abbr> messages. Default is 1800 seconds (<code>1800</code>). \
+						Max 9000 seconds.'));
+					so.optional = true;
+					so.default = '1800';
+					so.depends('ra', 'server');
+					so.depends('ra', 'hybrid');
+					so.depends('ra', 'relay');
+
+					so = ss.taboption('ipv6', form.Value, 'ra_mtu', _('<abbr title="Router Advertisement">RA</abbr> MTU'), _('The <abbr title="Maximum Transmission Unit">MTU</abbr> \
+						to be published in <abbr title="Router Advertisement, ICMPv6 Type 134">RA</abbr> messages. Default is 0 (<code>0</code>).\
+						Min 1280.'));
+					so.optional = true;
+					so.default = '0';
+					so.depends('ra', 'server');
+					so.depends('ra', 'hybrid');
+					so.depends('ra', 'relay');
+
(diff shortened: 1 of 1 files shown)


systemcrash:
Core changes in file:
`modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js`

Hope I got everything right. I think I ran these correctly after edits, according to the instructions:
```
./build/mkbasepot.sh
./build/i18n-update.pl
```

hnyman:
> Hope I got everything right.

Some of the translation refreshes seem strange.
For example,  thousands of double-// are added for some reason, like here:
```
- #: modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontab.js:29
+ #: modules/luci-mod-system//htdocs/luci-static/resources/view/system/crontab.js:29
```


> Core changes in file:

Might be better, if 
* there is one commit about your actual changes, and 
* you do not do the translations updates at all, if you are only trying to do the automatic refresh/sync without correcting the existing translations that you break.

The PR and commit title is wrong, as you are actually 
1) adding/exposing lots of IPv6 related options to LuCI
2) changing existing options' explanations

It might be even more clear, if you do two separate commits about those two things.  (and do not touch the translations at all)

systemcrash:
> > Hope I got everything right.
> 
> Some of the translation refreshes seem strange.
> For example, thousands of double-// are added for some reason, like here:
> 
> ```
> - #: modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontab.js:29
> + #: modules/luci-mod-system//htdocs/luci-static/resources/view/system/crontab.js:29
> ```
> 

I agree, I thought the path thing strange also - I ran the scripts on macOS - and with or without parameters, it adds the double `/`

hnyman:
Those erroneous strings will be removed in the update. 

Better if you leave the translations to .pot and .po  aside.

systemcrash:
OK - cleaned it up - but I don't think it is worth it to make separate PRs. Does this look better?

systemcrash:
Ah, sorry, commits - I read that as PR 😆 I'll see what I can wrestle out of git....

jow-:
Please change all `<br>` to `<br />` to make the translation strings XHTML compliant.

systemcrash:
> Please change all `<br>` to `<br />` to make the translation strings XHTML compliant.
> 
> Another remark:
> Instead of formatting the text with `<br>` Elements, it might make sense to use actual `<ul>` Lists for enumerations of select box choice descriptions.

Good tip. 

OK - all clear, I think.

systemcrash:
Also split into 2 commits - one adding help-text to present options, one adding new options with help-text.

systemcrash:
For the record, and those interested - looks roughly like this:

Screenshot 2021-04-01 at 16 31 42

Screenshot 2021-04-01 at 16 31 51

hnyman:
It is ok to have those new detailed options there, but the page looks somewhat cluttered. The important main options are now scattered among the minor options that the user usually does not need to touch.

It might be better to group the (old) main options to the top, and then place the RA detail options and NDP detail options etc. to the bottom of the page.

Ps. you should also edit the commit titles. Now there is only the "luci-mod-network:" on both commits without any other content except the module name.
Commits might be "luci-mod-network: add new IPv6 options"  and "luci-mod-network: clarify IPv6 option explanations", or something like that.

systemcrash:
> It is ok to have those new detailed options there, but the page looks somewhat cluttered. The important main options are now scattered among the minor options that the user usually does not need to touch.
> 
> It might be better to group the (old) main options to the top, and then place the RA detail options and NDP detail options etc. to the bottom of the page.

I tried it that way - and the 'grouping' disappears. Grouping helps provide strong cohesion and context. Scattering options lower down which appear only when an option higher up is selected makes for confusing options. 

The beginner will find this grouping less intimidating, and the advanced IPv6 user will understand what changed. 

> Ps. you should also edit the commit titles. Now there is only the "luci-mod-network:" on both commits without any other content except the module name.
> Commits might be "luci-mod-network: add new IPv6 options" and "luci-mod-network: clarify IPv6 option explanations", or something like that.

Agreed - I'll try and fix those.

systemcrash:
Slightly re-ordered. See updated screens above.

systemcrash:
Will we merge in time for 21.02 ?

hnyman:
21.02 has been branched off from master already in mid-February...

(it might be backported, but by default, no it is not going to hit 21.02)

hnyman:
I still think that the not-usual RA details are too prominent.

But we can merge this now to master, and tweak it later.

systemcrash:
OK! 

> (it might be backported, but by default, no it is not going to hit 21.02)

I think it'd be helpful for many were it back-ported/picked, and not miss another generation of releases. 😢 

> I still think that the not-usual RA details are too prominent.
> 
> But we can merge this now to master, and tweak it later.


Perhaps for your case - but RA details were added to odhcpd because they were deemed necessary. I guess it's a matter of opinion and necessity for those who need to tweak them, so when they are grouped, they are relevant. 

Well, it's been a learning curve 😄

systemcrash:
> Is hybrid mode really server+relay? Code looks like a smart switch between relay and server mode for dhcpv6 and ra.
> 
(and NDP). 

It does. How would you describe in plain English for a GUI what those parameters resolve to (i.e. how)?

Interface defers to something depending on what `master` is set to.

systemcrash:
DHCPv6 hybrid could probably be written as:

hybrid mode: Relay mode, if the interface boolean `master` is set to 1/True, otherwise, Server mode.

PR #4962, opened 2021-04-04
luci-app-transmission: fix '&nbsp' exception in OpenWrt theme

egorenar:
Convert `&nbsp;` to `&#160;` which is both syntactically correct (semicolon present) and XHTML compliant.

Signed-off-by: Alexander Egorenkov <egorenar-dev@posteo.net>


egorenar:
bump ? maintainers are to busy :(

PR #4965, opened 2021-04-06
luci-mod-network: Support for disabling routes

namedun:
Support for disabling routes in netifd was added in [this] commit. This commit
adds support for disabling routes through the LuCI interface.

Additionally hidden the route options descriptions in the header of the
routes table (they are still available in the modal view).

[this]: https://git.openwrt.org/?p=project/netifd.git;a=commit;h=327da9895327bc56b23413ee91a6e6b6e0e4329d


feckert:
Thanks for this :+1:

PR #4968, opened 2021-04-07
[21.02] luci-app-advanced-reboot: support 5.10 kernel on Zyxel

stangri:
As discovered by @Ansuel and @pkgadd, with the 5.10 kernel the partition name which used to be `0:DUAL_FLAG` is now in lowercase (`0:dual_flag`). This patch is based on the code originally developed by @pkgadd to address this and also includes an additional error case handling. Original discussion/code: https://github.com/openwrt/openwrt/pull/3954.

If the 5.10 kernel is going to be backported to 21.02 this needs to be merged.

Tested-by: Stefan Lippers-Hollmann <s.l-h@gmx.de>


hnyman:
>  If the 5.10 kernel is going to be backported to 21.02

Kernel 5.10 will not backpoted to 21.02.

PR #4985, opened 2021-04-13
luci-theme-openwrt-2020: remove submenu margin-bottom when submenu not visible

j123b567:
When mainmenu submenu is not visible, it still has visible `margin-bottom: .5em` so mixing main menu items with submenu (e.g. System) and without (e.g. Dashboard) causes uneven spaces between menu items.

This little tweek just makes all collapsed main menu items evenly distributed and does not change look of activated/selected items.


jow-:
Merged, thank you!

PR #5000, opened 2021-04-21
luci-theme-material: fix wrong active state on common prefix node

jjm2473:
Before fixed, if we have two nodes: 'services/ddns' and 'services/ddnsto',
click any one of they, will show they all actived.


jjm2473:
ping @jow- @feckert

jow-:
Merged, thanks

PR #5001, opened 2021-04-21
luci-theme-material: fix wrong active state on common prefix node

jjm2473:
Before fixed, if we have two nodes: 'services/ddns' and 'services/ddnsto',
click any one of they, will show they all actived.

(cherry picked from commit 97d50d2c6b80805cd7e513eeafc8b62fef4ab1b6)


hnyman:
17.01 is deprecated and unsupported. EOL

PR #5008, opened 2021-04-22
luci-mod-network: do not remove enable_vlan from config

mpratt14:
There are many switch drivers where VLAN functionality is enabled by default.
In this situation, LuCI cannot be used to disable VLAN functionality,
because removing the line from configuration results in VLAN remaining enabled
by the driver.

When enable_vlan is set to 0 by default
by using functions in board.d,
and an unrelated change is made on the Switch webpage,
then the enable_vlan option gets deleted when changes are saved.

Therefore, the option needs to be preserved, whether set to 0 or 1.

Signed-off-by: Michael Pratt <mcpratt@pm.me>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
@@ -180,8 +180,10 @@ return view.extend({
 			s = m.section(form.NamedSection, sid, 'switch', switch_title);
 			s.addremove = false;
 
-			if (feat.vlan_option)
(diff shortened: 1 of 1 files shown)


mpratt14:
this is good to backport for 21.02 as well

mpratt14:
additional reference

this is needed for any board whose network config includes:

`ucidef_add_switch_attr "switch0" "enable" "false"`
or
`ucidef_add_switch_attr "switch1" "enable" "false"`

mpratt14:
@aparcar this is a pretty simple fix...can we get it into both branches?

feckert:
From my point of view, it can't hurt to leave this in the configuration if the problem exists.
Thanks merged!

PR #5011, opened 2021-04-24
luci-app-attendedsysupgrade: add target in reuqest

aparcar:
The updated ASU api requires `target` within reuqests since the
`profile` or `board_name` is never truely unique. This is especially a
problem for x86 devices which mostly use `generic` as profile name.


freeesty:
changed the overview.js on a mt7621 router manually, but still getting:

`Error building the sysupgrade

Missing target

Please report the error message and request
Request to server:

{
    "board_name": "linksys,ea7500-v2",
    "target": "ramips/mt7621",
    "version": "SNAPSHOT",
    "packages": [`

aparcar:
Thanks you very much @freeesty , during refactoring I missed a single line which should have been added. I tested it again and it works again. Please report on issues!

PR #5020, opened 2021-04-29
luci-app-https-dns-proxy: add IPv6 addresses for cz.nic odvr

ja-pa:
This PR adds IPv6 addresses for CZ.NIC ODVR.
It is based on info from https://www.nic.cz/odvr/

cc @stangri


stangri:
Thank you!

PR #5024, opened 2021-04-30
luci-base: rename Dismiss button to Close in UCI-apply changes overview

SvenRoederer:
Change the ~german translation for "Dismiss" to "Abbrechen"~  "Dismiss" button to "Close" to differentiate with "Revert". Currently both are translated to "Verwerfen". This is confusing at least in the uci-apply dialog, which shows "Verwerfen" - "Speichern & Anwenden" - "Verwerfen".

Screenshot_20210501_003914

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -4063,7 +4063,7 @@ var UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 						E('button', {
 							'class': 'btn',
 							'click': UI.prototype.hideModal
-						}, [ _('Dismiss') ]), ' ',
+						}, [ _('Close') ]), ' ',
 						E('button', {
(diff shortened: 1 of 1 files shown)


feckert:
The problem is that the translation is done in weblate.
So I think if we change it here now it will retranslatet from a weblate translation people back to "Verwerfen".
We need to change the msgid.
I think we have to change the left one from **dismiss** to **close** because this closes the overlay.

SvenRoederer:
@feckert I had this in mind also shortly, but did not follow the weblate implementation really. When it's automatically, my change would be only a short success.

As your suggestion with "close" seems even the better wording, I updated my change accordingly.

feckert:
Thanks merged :-)

feckert:
Also pushed this to stable branch openwrt-21.02 with commit https://github.com/openwrt/luci/commit/1a5958416747a0cbb137e35c056c6f4161783c28

PR #5035, opened 2021-05-05
luci-mod-status: check if center_chan1 is defined

wjowsa:
While using channel analysis I noticed an exception. It turned out that center_chan1 is not returned in scan results for a "local_wifi" I use dual band card Compex WLE900VX. This causes that channels chart is not generated. 
Fix it by checking if center_chan1 is  defined. 

exc


hnyman:
cc @Ansuel

Ansuel:
sure it's a fix but it's strange that the center_chan is nor reported...

PR #5057, opened 2021-05-18
[luci-app-ddns] Fix Domain Tooltip in ddns form

### LINKED ISSUE #5056: [luci-app-ddns] Wrong string for "Domain" tooltip
In https://github.com/openwrt/luci/blob/94d35bbaa1e10c579d30125bf13a9c784253ec85/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L734-L736 the "Domain" field is declared, but the tooltip is the one for the "Username" fields. I know there's a Weblate project already, but I think that since the string is actually duplicated from the Username field, it's not currently possible to translate it there in Weblate as the system treats those as the same string. Should this be changed in the source code?

lorthirk:
Fix for the "Domain" tooltip in the ddns form

Fixes #5056

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -733,7 +733,7 @@ return view.extend({
 
 					o = s.taboption('basic', form.Value, 'domain',
 						_("Domain"),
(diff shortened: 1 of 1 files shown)


feckert:
Please prefix the commit head with `luci-app-ddns: <your message>`
So we see which app this commit belongs to.

lorthirk:
Just to let you know the commit message should be ok now

feckert:
Thanks merged

PR #5064, opened 2021-05-24
luci-app-ddns: fix time error

small-5:
luci-app-ddns:Fix the error that next_update sometimes display Verify

Fix time display error like this:

Running time should display 1h 0m 36s instead of 0h 60m 36s


hnyman:
cc @Ansuel

PR #5070, opened 2021-05-25
luci-mod-network: add 802.11ax HE support

blocktrron:
This commit adds the ability to configure HE-modes for radios
(HE20 / HE40 / HE80 / HE160) as well as HE rate information in the
assiciation view.

Tested-on: Ubiquiti UniFi 6 Lite / LR

Signed-off-by: David Bauer <mail@david-bauer.net>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -199,13 +199,22 @@ function format_wifirate(rate) {
 	var s = '%.1f\xa0%s, %d\xa0%s'.format(rate.rate / 1000, _('Mbit/s'), rate.mhz, _('MHz')),
 	    ht = rate.ht, vht = rate.vht,
 	    mhz = rate.mhz, nss = rate.nss,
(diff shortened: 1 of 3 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -351,7 +367,11 @@ var CBIWifiFrequencyValue = form.Value.extend({
 				],
 				'ac': [
 					'11a', '5 GHz', true
-				]
+				],
+				'ax': [
+					'11g', '2.4 GHz', this.channels['11g'].length > 3,
+					'11a', '5 GHz', this.channels['11a'].length > 3
+				],
```
hnyman, line 374: The extra comma at the end looks unnecessary, as there are no further items. 
`],`

andyrichardson:
Looks good! HE80+80 support would be dope too but I'm not sure how much of the stack supports this

blocktrron:
If VHT80+80 is supported, HE80+80 should also be supported.

I've implemented the required capability parsing in `iwinfo`, however i didn't test this specific mode.

PR #5074, opened 2021-05-26
luci-app-advanced-reboot: add board config for Linksys EA7500-V1

djadair:
EA7500-v1 works the same as the other Linksys models except
the name is different. Verified with 21-02 and master.

Signed-off-by: David Adair <djabhead@aol.com>


stangri:
Hello David, great job, thanks for this PR! You may want to send a similar PR for the [openwrt-21.02](https://github.com/openwrt/luci/tree/openwrt-21.02) branch.

PR #5086, opened 2021-05-30
luci-mod-network: hint that the network-upgrade might triggered multiple times

SvenRoederer:
Based on the 2-step migration of the bridge- and interface-configuration (e7c9c63c657963daa1a0d57dcb0848d88e909c25) add a
hint that the "Network configuration migration" might get triggered multiple
times. This will not confuse the user with some thought of endless loop.


oldium:
It would be good to merge this PR. I just did an upgrade and without knowing this pull request beforehand, it looked really confusing/weird.

oldium (approved):

SvenRoederer:
@jow- @rmilecki what do you think of this?

SvenRoederer:
@jow- @rmilecki ping :-)

Issue #2998, opened 2019-08-23
[feature suggestion] luci-app-ipv6: Add NDP-Proxy option to LuCI when protocol is DHCPv6 Client


ffrediani:
Sometimes when the ISP doesn't doesn't provide a IPv6 Prefix Delegation or when it only provides a /64 to be used by the LAN interface it may be necessary to enable NDP-Proxy in order that other networks can have IPv6 connectivity.

In order to archive that the LAN interface 'Router Advertisement-Service', 'DHCPv6-Service and NDP-Proxy must be all changed to 'Relay Mode'. That can all be done fine in LuCI under DHCP Server options.
However according to OpenWrt documentation it is necessary to also add the same options (plus a fourth one - 'option master 1') to the wan interface in the /etc/config/dhcp file (https://oldwiki.archive.openwrt.org/doc/uci/network6#router_advertisement_dhcpv6).

This can only be done via CLI and cannot be done via LuCI for the WAN6 interface when the selected Protocol is 'DHCPv6 Client'.

Please add an option that allows the user to easily enable this and select which WAN interface is the master for DHCPv6, RA and NDP when in Relay mode.

This is useful because many people inadvertently connects the LAN port of a home router to a WAN port of a second router in order to extend signal and by using this technique allows that users connected to the secondary routers keep IPv6 connectivity.

feckert:
Please provide the uci network config options which have do be set by the luci

Strykar:
Seems like the checkbox option available under `LAN` for `Master`, (`Set this interface as master for the dhcpv6 relay.`)
But for the `wan6` interface when configured as a DHCPv6 client:
```
# uci show dhcp.wan6 | grep master
dhcp.wan6.master='1'
```

ailin-nemui:
the problem is that the DHCP/RA/NDP/ relay options are hidden from the UI when configured as DHCPv6 client (wan6), they are only made visible in the "Static" configuration. Thus, the fix should be rather simple -- show the "DHCP IPv6" tab also on the DHCPv6 client mode interface

ailin-nemui:
although logically it could also make sense to add a dedicated "Configure relay master to be on interface [wan6]" option to the LAN config instead

ffrediani:
Exactlly @ailin-nemui 
It's pretty much that.

PR #5095, opened 2021-06-01
luci-mod-network: interfaces: restructure DHCPv6 and IPv6 RA options

jow-:
- Condense overly large IPv6 RA/DHCPv6 description texts and get rid of most embedded markup

 - Switch ra/ndp/dhcpv6 mode selections to rich dropdown lists and move extended choice
   descriptions next to the selection options

 - Drop ndproxy_static option which has been removed from odhcpd long ago

 - Add format validations to all text input fields

 - Add ability to configure master/relay modes for non-static interfaces (#2998)

 - Move extended RA configuration options into a new tab

 - Prevent enabling master mode on multiple interfaces

 - Prevent enabling ra/dhcpv6 server mode on non-static or master interfaces

 - Drop ra_management in favor to ra_flags option (#5083)

 - Add support for dns_service option

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -228,6 +228,39 @@ function get_netmask(s, use_cfgvalue) {
 	return subnetmask;
 }
 
+var cbiRichListValue = form.ListValue.extend({
+	renderWidget: function(section_id, option_index, cfgvalue) {
+		var choices = this.transformChoices();
+		var widget = new ui.Dropdown((cfgvalue != null) ? cfgvalue : this.default, choices, {
+			id: this.cbid(section_id),
+			sort: this.keylist,
+			optional: true,
+			select_placeholder: this.select_placeholder || this.placeholder,
+			custom_placeholder: this.custom_placeholder || this.placeholder,
+			validate: L.bind(this.validate, this, section_id),
+			disabled: (this.readonly != null) ? this.readonly : this.map.readonly
+		});
+
+		return widget.render();
+	},
+
+	value: function(value, title, description) {
+		if (description) {
+			form.ListValue.prototype.value.call(this, value, E([], [
+				E('span', { 'class': 'hide-open' }, [ title ]),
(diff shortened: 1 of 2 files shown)


jow-:
@hnyman - I haven't been overly happy with the changes merged in 988c98bdc6b30a1d9fb17e1ba4444d0beb236b8a, 504bdb23fbaff28cc31cc1dd2408a9f1cd5eddde - the extreme amount of text as well as the embedded markup looked quite out of place. It also introduced a bunch of regressions.

Due to that, I decided to rework that entire area. I would appreciate if you could give it a test.

### Configuration for static (downstream) interfaces

image

image

image

### Configuration for non-static (upstream) interfaces

image
image

hnyman:
I will test it. 
The help texts have been extensive, and I wondered about the amount of the less-used options already in the original PR, as they may confuse the casual user.

I thought to do some minor cleanup myself , but you have done a much better job in organising it  ;-)

You might also remove the wrong default option text and placeholder from ra_mtu at the same time:
` Default is 0 (<code>0</code>).` is invalid.  
https://github.com/openwrt/luci/pull/5095/files#diff-56df355dcbbd3eaeb1ed044f96b8e956fa4cb13b5202987f6622057823c08616R827-R831

(I removed the originally set default values with placeholder a while ago, but did not edit the texts at that time as it was unclear why that 0 was wrong. That has been clarified since then, and the README in pdhcpd fixed)


Reference to 
* https://git.openwrt.org/?p=project/odhcpd.git;a=commitdiff;h=a12fcb3cee2d489b8648a2398812d7bed2f25faa;hp=ebd4297fc8317a426a6c0faea1cbba2e523270d4
* https://bugs.openwrt.org/index.php?do=details&task_id=3789

hnyman:
@jow- 
I applied this to both master and 21.02, and it seems ok to me.
Looks really nice ;-)

Like said earlier, I would remove the wrong ra_mtu default value 0 claim
Additionally I would maybe clean a bit further by removing the unnecessary "code" numbers inside parenthesis:
image

jow-:
Thanks for the suggestions @hnyman - I further cleaned up the descriptions and removed the `<code>` tags as suggested. The effective default values for IPv6 MTU and hop limit are now read from proc fs as well.

image

ErnyTech:
It looks very good to me, very well organized

PR #5098, opened 2021-06-02
Fix fwmark validation to allow 32 bit values

differentblue:
The iptables mark field is 32 bits wide, which is 4 bytes and so 8 hex characters.  Fix the fwmark validation to allow 8 characters in the hex string.


jow-:
Merged via bc1015f - thank you!

PR #1635, opened 2018-02-20
freifunk-policyrouting: do not run init-script in imagebuilder

SvenRoederer:
When creating final images the build-scripts running the package init-scripts. As of including files like on a real system (line 4, 5), with absolute path, this will fail. 

> Activating init scripts
build_dir/target-mips_24kc_musl/root-ar71xx/etc/init.d/freifunk-policyrouting: line 4: /lib/functions/network.sh: No such file or directory
build_dir/target-mips_24kc_musl/root-ar71xx/etc/init.d/freifunk-policyrouting: line 5: /lib/functions.sh: No such file or directory

Running this init-script outside of a real-system don't seem to make sense at all, so just exit gracefully.

PR #5100, opened 2021-06-04
luci.mk: don't run uci-defaults scripts ourself

SvenRoederer:
Do not run the uci-defaults scripts of a package in the package-postinst, but leave this for the default_postinst() of the OpenWrt base-files package.
Based on https://github.com/openwrt/luci/pull/1635#issuecomment-368266512 and a short test it worked as expected. 

Looks to me like this was added back in 2014 when OpenWrt did not start the uci-defaults automtically.


jow-:
Merged, thanks.

PR #5101, opened 2021-06-05
luci-app-sqm: fix JS error if service is disabled

dibdot:
* fix a 'resource not found error' if the sqm service has been disabled
  and the dir '/var/run/sqm/available_qdiscs' not exist
* inform the user about this issue (and how to fix this)

### DIFF
--- a/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
+++ b/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
@@ -1,5 +1,6 @@
 'use strict';
 'require fs';
+'require ui';
 'require rpc';
 'require uci';
 'require view';
@@ -23,8 +24,8 @@ return view.extend({
 
 	load: function() {
 		return Promise.all([
-			fs.list("/var/run/sqm/available_qdiscs"),
-			fs.list("/usr/lib/sqm").then(L.bind(function(scripts) {
+			L.resolveDefault(fs.list('/var/run/sqm/available_qdiscs'), []),
+			L.resolveDefault(fs.list('/usr/lib/sqm'), []).then(L.bind(function(scripts) {
 				var tasks = [], scriptHelpTbl = {};
 
 				for (var i = 0; i < scripts.length; i++)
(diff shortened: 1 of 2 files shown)


dibdot:
@tohojo 
SQM before:
sqm_before

SQM after:
sqm_after

tohojo:
Hmm, ideally we would just trigger the probing function if the available_qdiscs dir has not been populated...

dibdot:
What did you mean with "probing function" - the simple check for an empty array? I don't know a better way.

tohojo:
/usr/lib/sqm/update_available_qdiscs is the script that populates that directory. Would be good if we could have a server-side hook in luci to trigger that script if the directory is empty instead of just listing the dir contents...

dibdot:
Ah OK, I'm not using SQM ... this comes up during random pre 21.02 testing. It's only meant as a minimal bugfix.

hnyman:
The whole sqm service might be disabled intentionally, so the probing script should not be triggered just because the user goers to the LuCI page (without changing anything there).

tohojo:
Hannu Nyman ***@***.***> writes:

> The whole sqm service might be disabled intentionally, so the probing
> script should not be triggered just because the user goers to the LuCI
> page (without changing anything there).

Why not? The probing doesn't actually interfere with network traffic,
and it needs to run before the user can actually use the configuration
page for anything...

The problem with the proposed error message here is that it's not
sufficient to enable sqm-scripts in the init script: it needs to
actually run to populate the list of qdiscs. So it's a bit of chicken
and egg here; we really need a way for the user to be able to trigger
the probing from luci, and just doing it automatically when opening the
sqm config page seems like the simplest way?

dibdot:
> The problem with the proposed error message here is that it's not sufficient to enable sqm-scripts in the init script: it needs to actually run to populate the list of qdiscs.

Nope, the message clearly states "Enable" and "Restart" to create the required sqm dirs/files.
If a service is disabled I don't expect that this service creates any service related files - just my 2 cents.

tohojo:
Jo-Philipp Wich ***@***.***> writes:

> What about offering to enable/start/probe the service with the click
> of a button or a link? Something like "Setup SQM now..."

Yeah, that would be better, definitely! "SQM is currently disabled,
click here to enable before configuring" - and then automatically
refreshing the UI after that has been done (and autodetection has
run)...

dibdot:
I've added the requested "Enable SQM" button to enable & start the service, e.g.:
sqm_service

tohojo (approved):
Yeah, that's better. LGTM, @jow- WDYT?

jow-:
@dibdot - I had to apply some fixes to ensure that enable/start and reload do not race and that the loading indicator remains active while the operations run. I'd appreciate if you could double-check 005c09a293

dibdot:
@jow yeah, looks much cleaner now - thanks!

PR #5109, opened 2021-06-08
themes: Call striptags() on hostname to prevent XSS

hauke:
This calls striptags() on the hostname to prevent any XSS over the
hostname. This should fix CVE-2021-33425 as far as I understood it.

If someone adds some Javascript into `system.@system[0].hostname` it would
have been directly added to the page, this prevents the problem.

This can only be exploited by someone being able to modify the uci
configuration, normally a user with such privileges could also just
modify the webpage.

Signed-off-by: Hauke Mehrtens <hauke@hauke-m.de>

### DIFF
--- a/themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm
+++ b/themes/luci-theme-bootstrap/luasrc/view/themes/bootstrap/header.htm
@@ -41,7 +41,7 @@
 		<header>
 			<div class="fill">
 				<div class="container">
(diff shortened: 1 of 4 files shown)


aparcar:
I'm not sure if this is a real thing, if you can modify settings, you might as well do pretty much anything else.

jow-:
Doesn't hurt though, so applying it.

hauke:
@aparcar I agree with you, but I think we should still fix it.
I will also backport this to 21.02 and 19.07 later.

Issue #5106, opened 2021-06-08
Channel option not appearing in 5ghz radio


CaboBeach:
There is no channel option in Luci in the 5ghz radio when selecting AC mode, although it works fine when setting in cli, channel option is still there in N and Legacy modes. Netgear R7800/OpenWrt SNAPSHOT r16906-f36990eae7 / LuCI Master git-21.158.53317-da97288

![Screen Shot 2021-06-07 at 8 01 13 PM](https://user-images.githubusercontent.com/48701904/121106546-d6257680-c7cb-11eb-8f64-b60fcc0cc952.png)

zxlhhyccc:
me too

CaboBeach:
Thanks @zxlhhyccc, didn't realize this was already closed, i'll go ahead and close this as well.

zxlhhyccc:
> Thanks @zxlhhyccc, didn't realize this was already closed, i'll go ahead and close this as well.

Hello, problem not solved why close it?

CaboBeach:
Just assuming the devs were aware and working on it if it had been closed, I’ll go ahead and reopen.

hnyman:
Works for me.
R7800 with master r16914-1aa80ce393  with LuCI  git-21.156.61451-516c58e


![image](https://user-images.githubusercontent.com/7926856/121173757-d96b4180-c861-11eb-9ff5-4a7f7247a76c.png)

zxlhhyccc:
> Works for me.
> R7800 with master r16914-1aa80ce393 with LuCI git-21.156.61451-516c58e
> 
> ![image](https://user-images.githubusercontent.com/7926856/121173757-d96b4180-c861-11eb-9ff5-4a7f7247a76c.png)

You are not compiling firmware using the master branch

zxlhhyccc:
Can be flexible solution, namely: first set N mode to save and apply and then select AC mode can set the channel

> There is no channel option in Luci in the 5ghz radio when selecting AC mode, although it works fine when setting in cli, channel option is still there in N and Legacy modes. Netgear R7800/OpenWrt SNAPSHOT r16906-f36990eae7 / LuCI Master git-21.158.53317-da97288
> 
> ![Screen Shot 2021-06-07 at 8 01 13 PM](https://user-images.githubusercontent.com/48701904/121106546-d6257680-c7cb-11eb-8f64-b60fcc0cc952.png)

https://github.com/openwrt/luci/issues/5106#issue-914125730

[...]

PR #5110, opened 2021-06-09
luci-mod-network: support newer wireless.wifi-device.band option

jow-:
OpenWrt switched from "option hwmode" to "option band" in order to select
the frequency band to use for the radio phy.

Extend the channel selector to recognize and use an existing "option band"
to select the appropriate channel list. When operating upon a wireless
configuration still using "option hwmode", then translate it to a band
value internally and translate it back to "option hwmode" on save.

This should provide forward- and backwards compatibility with both current
OpenWrt master and older versions still using hwmode.

Fixes: #5106

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -312,16 +312,32 @@ var CBIWifiFrequencyValue = form.Value.extend({
 			this.callFrequencyList(section_id)
 		]).then(L.bind(function(data) {
 			this.channels = {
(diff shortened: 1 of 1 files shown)


castiel652:
Works as expected on R7800 running master

image

zxlhhyccc:
> OpenWrt switched from "option hwmode" to "option band" in order to select
> the frequency band to use for the radio phy.
> 
> Extend the channel selector to recognize and use an existing "option band"
> to select the appropriate channel list. When operating upon a wireless
> configuration still using "option hwmode", then translate it to a band
> value internally and translate it back to "option hwmode" on save.
> 
> This should provide forward- and backwards compatibility with both current
> OpenWrt master and older versions still using hwmode.
> 
> Fixes: #5106
> Signed-off-by: Jo-Philipp Wich [jo@mein.io](mailto:jo@mein.io)

Problem solved!

PR #5116, opened 2021-06-10
lucihttp: fix compilation under 32-bit

neheb:
Also fixed compilation with Ninja by adding + to CMAKE_OPTIONS.


jow-:
Thanks for the patch @neheb - I decided to push the fix directly upstream instead of adding it as patch file.

PR #5129, opened 2021-06-16
luci-app-nlbwmon: fix url and clean up the ui buttons

feckert:
When moving this application to the service section, the links were forgotten to be changed.

@jow- Since I messed up, it's up to me to deal with it on my own. 
But in the process I noticed that the change in display.js wasn't right in the first place, was it?

If I had not moved the application, then it should have looked like this.
```

- E('a', { 'href': L.url('admin/nlbw/commit') }, _('Force reload…'))
+ E('a', { 'href': L.url('admin/nlbw/update') }, _('Force reload…'))

```

### DIFF
--- a/applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/backup.js
+++ b/applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/backup.js
@@ -52,12 +52,14 @@ return view.extend({
 			E('h5', [ _('Restore Database Backup') ]),
 			E('p', [
 				E('button', {
+					'class': 'cbi-button',
 					'click': ui.createHandlerFn(this, 'handleArchiveUpload')
 				}, [ _('Restore') ])
 			]),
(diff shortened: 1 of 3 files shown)


jow-:
Force reload is supposed to trigger a commit to make data available for display. The change to make it simply reload the page is wrong.

feckert:
@jow- The handleCommit function already exists. I have now simply used this one. It seems to work now.

I still don't like the buttons in the luci-theme-openwrt-2020, so I will change this with this [P/R](https://github.com/openwrt/luci/pull/5130).

feckert:
The export tab on the display page did not look good with the themes openwrt2020 and bootstrap.
So I cleaned that up as well while I'm at it.
**luci-theme-openwrt-2020:**
Screenshot_2021-06-17 st-dev-07 - Display - LuCI(1)
**luci-theme-bootstrap:**
Screenshot_2021-06-17 st-dev-07 - Display - LuCI

Issue #4824, opened 2021-02-20
luci-theme-openwrt-2020: loading animation is not centered


castillofrancodamian:
I see that it doesn't happen on all LuCI sites where animation is. I could see in the scanner and opkg.

It is a new compilation of OpenWrt, without configuration files and clean browser of cookies.

![1](https://user-images.githubusercontent.com/47670442/108610139-1b0ad600-73b2-11eb-8061-08eaa82229f0.png)
![2](https://user-images.githubusercontent.com/47670442/108610141-1e05c680-73b2-11eb-94b9-980e824db601.png)
![3](https://user-images.githubusercontent.com/47670442/108610185-62916200-73b2-11eb-81b4-e8a387da64c9.png)
![4](https://user-images.githubusercontent.com/47670442/108610186-64f3bc00-73b2-11eb-817f-eea214343da7.png)

feckert:
@castillofrancodamian The way I see it, the spinner is spinning around the wrong center. Unfortunately, I am not a CSS expert, I do not know where to start here. Of course, it may be that this is intentional

PR #5130, opened 2021-06-16
modules: minor styling fixes

feckert:
During testing I notice some styling issue with the luci-theme-openwrt-2020.
Due missing of the btn class in form.js the button in xintd looks strange. I have added to all `cbi-button` css class the `btn` css class in from.js. We have already discussed the topic in other issues. I don't know the btn css class. But it looks like it's necessary.

### Before change:
Screenshot_2021-06-16 st-dev-07 - Xinetd - LuCI

### After the change:
Screenshot_2021-06-16 st-dev-07 - Xinetd - LuCI

### Before change:
Due using div for instead of table element the wlan channel analysis looks strange.
Screenshot_2021-06-16 st-dev-07 - Channel Analysis - LuCI

### After the change:
Screenshot_2021-06-16 st-dev-07 - Channel Analysis - LuCI

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2203,9 +2203,8 @@ var CBITypedSection = CBIAbstractSection.extend(/** @lends LuCI.form.TypedSectio
 
 			dom.append(createEl, [
 				E('div', {}, nameEl),
-				E('input', {
+				E('button', {
 					'class': 'cbi-button cbi-button-add',
-					'type': 'submit',
 					'value': btn_title || _('Add'),
 					'title': btn_title || _('Add'),
 					'click': ui.createHandlerFn(this, function(ev) {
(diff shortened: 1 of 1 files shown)


castillofrancodamian:
The loading animation [here](https://github.com/openwrt/luci/issues/4824) too.

feckert:
@jow By changing the html element from input to button in form.js all buttons are fixed. Please try this out.

feckert:
@jow- I have now called a module that is not rendered by javascript but by lua.
There we have everywhere when adding anonymous section the html input field element with the css class btn set.
I wonder what is right now!

The Javascript doesn't care if it is a button or an input. So that the CSS does not become even more complicated, we should make it uniform. For lua we have to take the input otherwise the form doesn't work anymore?

feckert:
I have pushed the channel-analysis changes to [openwrt-21.02](https://github.com/openwrt/luci/commit/95b3633055c17c7b396bde1c41053006a82992ee) and [master](https://github.com/openwrt/luci/commit/8f8ce8613fb6e41954aa5bcbb60330e3b034bd3a) because this was ACK by @jow- 

For the other issue I will open a separate P/R

PR #5137, opened 2021-06-21
luci-base: minor fixes and improvments

feckert:
- Change html tag `input` to `button` for adding a named section. Button is then rendert correct for luci-theme-openwrt-2020
- Call validation, because named section is always empty on page load. User must added a name for named sections.
- Remove `btn` css class and also rename them to `button`

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2203,10 +2203,8 @@ var CBITypedSection = CBIAbstractSection.extend(/** @lends LuCI.form.TypedSectio
 
 			dom.append(createEl, [
 				E('div', {}, nameEl),
-				E('input', {
+				E('button', {
 					'class': 'cbi-button cbi-button-add',
-					'type': 'submit',
-					'value': btn_title || _('Add'),
 					'title': btn_title || _('Add'),
 					'click': ui.createHandlerFn(this, function(ev) {
 						if (nameEl.classList.contains('cbi-input-invalid'))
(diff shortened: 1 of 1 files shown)


feckert:
@jow- @hnyman Can I merge this improvments? Or is there something against it? I would also cherry-pick this to the upcoming openwrt-21.02 stable release.

hnyman:
Sorry, goes out of my knowledge scope  :-(

jow-:
Rendering a form with some fields having an initial invalid state is not good UX imho. I'd prefer the add named section input to be "valid" and "empty" by default. If the user provides no value and clicks add then simply do nothing or even better, disable the add button and only unblock it once the entered section name value is valid.

feckert:
@jow- Thank you for your advice.
I have now adjusted the handling to get a good UX.
That the **Add** button at the beginning (on load) is disabled.
If someone enters a string into the input field then the button gets enabled and could be pressed to add this named section to the configuration.

feckert:
@jow- ready to go or do you have any objections?

jow-:
LGTM

feckert:
@jow- should we also cherry-pick this to openwrt-21.02?

jow-:
There’s a bug I overlooked. If the map is readonly you must not reenable the add button. You removed that condition without replacement

feckert:
@jow- Sorry, I didn't see that either. Should I change it like this? 

```
diff --git a/modules/luci-base/htdocs/luci-static/resources/form.js b/modules/luci-base/htdocs/luci-static/resources/form.js
index e9c8d2eb00..ddc59222bd 100644
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2212,7 +2212,7 @@ var CBITypedSection = CBIAbstractSection.extend(/** @lends LuCI.form.TypedSectio

                                                return this.handleAdd(ev, nameEl.value);
                                        }),
-                                       'disabled': true
+                                       'disabled': this.map.readonly || true
                                }, [ btn_title || _('Add') ])
                        ]);
```

feckert:
@jow- I have now verified this again. If I set the map in acm.js to `m.readonly = ture`, then I cannot enter anything in the input field. As a result, the validation is never triggered and the button is never switched on. So that works. But I agree with you that it is not clean.

jow-:
I think you can simply omit the validator if m.readonly is true, this way the button will always remain statically disabled.

PR #5138, opened 2021-06-21
luci-app-acme: change form view to GridSection

feckert:
@tohojo If multiple certificates are added, then the page is a bit confusing with
all the information's. Therefore the GridSection view would be better
suited here.

**Before:**
Screenshot_2021-06-21 st-dev-07 - ACME Zertifikate - LuCI

**After:**
Screenshot_2021-06-21 st-dev-07 - ACME Zertifikate - LuCI(2)

Screenshot_2021-06-21 st-dev-07 - ACME Zertifikate - LuCI(1)


tohojo (approved):
Sure, fine with me :)

PR #5141, opened 2021-06-22
luci-app-frpc: allow hostname in server address

msylgj:
Sometimes, our frp server cannot use a static ip address. 
So perhaps we should allow people to use hostname in server address.

Signed-off-by: SiYao Mo <msylgj@vip.qq.com>

PR #5142, opened 2021-06-22
luci-app-frpc: allow hostname in server address

msylgj:
Sometimes, our frp server cannot use a static ip address.
So perhaps we should allow people to use hostname in server address.


feckert:
Duplicate https://github.com/openwrt/luci/pull/5141

1715173329:
This one is for openwrt-21.02 branch however...

feckert:
@msylgj Sorry my fault. Please also change the Signed-off-by here as well.

msylgj:
> @msylgj Sorry my fault. Please also change the Signed-off-by here as well.

It's ok now~thx

PR #5154, opened 2021-06-28
luci-base: prevent empty field for adding new entry

feckert:
Currently it is not checked whether a valid string was entered in the input field when adding a named section.
This commit validates if the input field is empty or not.
If the input field is not empty, then the Add button is displayed.

refs https://github.com/openwrt/luci/pull/5154


feckert:
@jow- ready to go or do you have any objections?

jow-:
@feckert - I took a look at it again and I think there‘s still some issue. If there‘s multiple maps on the page then the query selectorvwill only ever find the first button from the first map.

I think you should find the button relative to the input node, something like input.parentNode.querySelector(...) or input.nextElementSibling

PR #5156, opened 2021-06-29
luci-base: batch DOM updates to prevent slowdown

0x0aa:
Currently, if there are a lot of rows in the Realtime Connections table, the page slows down to a crawl.

image

A major actor here is the _cbi_update_table_ function, which upsets the browser by making a lot of DOM updates with _slight pause in between_ when updating the table rows. Because of that pause between the updates, the browser tries to keep the view updated and does tons of useless style recalculations and this is what freezes the page. 

Before:

image

Luckily this behavior can be fixed by batching the DOM updates.

After:

image

After this optimization the page can handle thousands of rows without breaking a sweat, whereas before it would choke with some hundreds of rows.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/cbi.js
+++ b/modules/luci-base/htdocs/luci-static/resources/cbi.js
@@ -752,7 +752,7 @@ function cbi_update_table(table, data, placeholder) {
 		});
 
 		if (Array.isArray(data)) {
-			var n = 0, rows = target.querySelectorAll('tr, .tr');
+			var n = 0, rows = target.querySelectorAll('tr, .tr'), trows = [];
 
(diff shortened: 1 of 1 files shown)


0x0aa:
A simple way to experience the slowdown is to enable DNS lookups (on the Connections page) and run a DNS leak test (e.g. www.dnsleaktest.com) in 5 browser tabs =)

nickberry17:
Tested on Chromium, it works (of course it does) 
LGTM

before
before js opti


after
after js opti

jow-:
Merged via f919635310286b3b905aedb716d707da1ea2c811 (including minor white space fix). Thanks!

PR #5158, opened 2021-06-30
luci-app-acl: some fixes

feckert:
This P/R does make two changes

1. luci-app-acl: fix translation function call. The second argument in the translation function makes no sense.
2. luci-app-acl: unset read and write before acl set.  If the setting in the view is set to `denied`, only the read list option is deleted. This is not correct. The write list option must also be deleted. To ensure that the correct configuration is saved, the write and read list options are always deleted beforehand and then rewritten.

**Configuration that is being changed:**
```
config login
        option username 'test'
        option timeout '300'
        option password '$1$$whuMjZj.HMFoaTaZRRtkO0'
        list read 'luci-app-acme'
        list read 'luci-app-commands'
        list read 'luci-app-ddns'
        list read 'luci-base'
        list read 'unauthenticated'
        list write 'luci-app-acme'
        list write 'luci-app-commands'
        list write 'luci-app-ddns'
```
**Before acl.js change to denied (wrong):**
Screenshot_2021-06-30 st-dev-07 - ACL Settings - LuCI(1)
```
config login
        option username 'test'
        option timeout '300'
        option password '$1$$whuMjZj.HMFoaTaZRRtkO0'
        list write 'luci-app-acme'
        list write 'luci-app-commands'
        list write 'luci-app-ddns'
        list read 'luci-base'
        list read 'unauthenticated'
```
**After acl.js change to denied (correct):**
Screenshot_2021-06-30 st-dev-07 - ACL Settings - LuCI
```
config login
        option username 'test'
        option timeout '300'
        option password '$1$$whuMjZj.HMFoaTaZRRtkO0'
        list read 'luci-base'
        list read 'unauthenticated'
```

### DIFF
--- a/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
+++ b/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
@@ -146,6 +146,9 @@ var cbiACLSelect = form.Value.extend({
 	},
 
 	write: function(section_id, value) {
+		uci.unset('rpcd', section_id, 'read');
+		uci.unset('rpcd', section_id, 'write');
+
 		if (L.isObject(value) && Array.isArray(value.read))
 			uci.set('rpcd', section_id, 'read', value.read);
 


jow-:
> luci-app-acl: fix translation function call. The second argument in the translation function makes no sense.

NACK. The second argument to the translation call is the message context (`msgctxt` in gettext, see https://www.gnu.org/software/gettext/manual/html_node/Contexts.html)

feckert:
> > luci-app-acl: fix translation function call. The second argument in the translation function makes no sense.
> 
> NACK. The second argument to the translation call is the message context (`msgctxt` in gettext, see https://www.gnu.org/software/gettext/manual/html_node/Contexts.html)

got it thanks, I didn't know that, thanks for pointing it out.

PR #5162, opened 2021-07-02
luci-app-nlbwmon: Add translation context for Download/Upload terms

namedun:
Added a translation context for the terms 'Download' and 'Upload'
used in this application. The reason is that these terms are
translated differently in Russian depending on the context.

If the terms are used in the sense of an action (download file
or upload file) then the translations should be as follows:
- Download = скачать
- Upload   = загрузить

But if they are used in the context of a traffic counter (like
in this application) the correct translations are:
- Download = скачано
- Upload   = загружено

A translation context has been added to all affected PO files
to avoid breaking existing translations. The Russian translation
in PO file is fixed to comply with the used context.


jow-:
Merged, thanks!

PR #5171, opened 2021-07-06
luci-app-ddns: fix next_update error

small-5:
* Fix the error that next_update sometimes display Verify


small-5:
@jow- Cloud you please merge this PR? I've been testing it for a month, and it shows that verify has been fixed, and it doesn't happen again.

PR #5172, opened 2021-07-06
luci-base:Fix time display error

small-5:
Fix time display error like this:
119333137-1bba4d80-bcbc-11eb-8d6a-a2cd664061de
Running time should display 1h 0m 36s instead of 0h 60m 36s


jow-:
Heh, that was an embarrassing bug 😳

Thanks for fixing! 👍

kuanyili:
Instead of changing the value to compare, maybe we should just change the comparator from `>` to `>=`.
Or even further, omit these `if`s altogether and run the code inside anyway?

Also, are we going to have the fix in other release branches?

PR #5173, opened 2021-07-06
luci-lib-jsonc: stringify int use json_object_new_int64 instead

lisaac:


jow-:
This patch looks wrong. It will convert any double value that is between INT64_MIN and INT64_MAX into an 64bit integer.

Issue #4603, opened 2020-11-18
luci-app-mwan3: Duplicate metric warning on separate family interfaces and other potential false warnings


jamesmacwhite:
@feckert @aaronjg 

I've kind of ignored most of the warnings from luci-app-mwan3 as in my opinion some of them might be a bit false positive because it doesn't really handle IPv6 interfaces well from what I can see.

![image](https://user-images.githubusercontent.com/8067792/99581620-f0873680-29d8-11eb-8b2d-5b2630c27345.png)

I have a couple of questions though.

1. Is having the same metric value applied to both the WAN and WAN6 interface really invalid? Because one is IPv4 and the other IPv6 surely this is fine, because it's two different routing tables for IPv4 and IPv6? I'm assuming it's perhaps just looking at the interface entry itself and thinks it's duplicated.

2. Any IPv6 interface seems to throw the `Interface xxxx has no default route in the main routing table`, again I think this because it's not specifically looking at the IPv6 configuration here.

All interfaces work fine, I have just been ignoring the errors but I've seen other users in the forums get put off by the above errors from LuCI, when I don't think they are that useful because I don't even think from an IPv6 perspective they are properly reporting an issue.

I'm not an expert however, so I wanted to ask for further clarification. Similar points around the diagnostics functionality, for any IPv6 interface most of the diagnostic options don't work. I personally don't use it that much and would always use SSH and test with various tools like traceroute, iperf3, ping etc.

Thank you!

jamesmacwhite:
I'm assuming my first point it's because the condition is looking at an IPv4 route of 0.0.0.0 present on the interface which won't be valid for an IPv6 interface.

https://github.com/openwrt/luci/blob/master/applications/luci-app-mwan3/luasrc/model/cbi/mwan/interface.lua#L85-L94

jamesmacwhite:
And the duplicate metric warning looks to be just looping through interfaces and seeing if the assigned metric value is present, not taking into account the interface being a different family. In fact the interface config itself doesn't appear to actually expose the interface family set in this case.

https://github.com/openwrt/luci/blob/master/applications/luci-app-mwan3/luasrc/model/cbi/mwan/interface.lua#L73-L83

aaronjg:
That all sounds correct - the diagnostics appear to be quite off for IPv6, and a P/R fixing this would be welcomed.

[...]

PR #5174, opened 2021-07-07
luci-app-mwan3: convert to JS

### LINKED ISSUE #5022: luci-app-mwan3: Renaming the current Load balancing label in LuCI navigation
@feckert @aaronjg 

This is something minor but currently the mwan3 navigation menu label is "Load balancing" in LuCI for both the Status and Network menu. Because of the combination of both load balancing and failover functionality, should the label be renamed to something that is more encompassing of it's feature set e.g. "Multi WAN".

**Examples could include:**

For the status menu:

* Multi WAN Status
* mwan3 Status

For the network menu:

* Multi WAN
* mwan3 Interfaces
* Multi WAN Configuration

Having it labelled as load balancing doesn't seem accurate given it's capabilities and uses. I'd be open to other suggests but I think having it labelled as Load Balancing isn't really accurate.

feckert:
I have finally found the time to convert the luci-app-mwan3 to java script.

Todo after merge:

- Currently, the plausibility checks are only done for IPv4. But that is not correct see #4603. The interface can now also be IPv6 only. This still needs to be worked on.
- Update I18N

Dependency:

- In order for the troubleshoot output to work again, this [P/R](https://github.com/openwrt/packages/pull/16066) must be merged before. The query is better off there.

Info:

- @jamesmacwhite I have also unified the naming from `Mwan3` to `MultiWAN Manager. fixes #5022


applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/mwan3.js
```diff
@@ -0,0 +1,95 @@
+function secondsToString(time) {
```
jow-, line 1: You should be able to replace every `secondsToString(val)`  call with `'%t'.format(val)` and subsequently drop this function

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/mwan3.js
```diff
@@ -0,0 +1,95 @@
...
+	var hrs = Math.floor(seconds / 3600);
+	seconds -= hrs*3600;
+	var mnts = Math.floor(seconds / 60);
+	seconds -= mnts*60;
+	return String.format("%sh:%sm:%ss", hrs, mnts, seconds);
+}
+
+function renderMwan3Status(status) {
+
+	if (status.interfaces) {
```
jow-, line 13: This is just a stylistic remark: Personally I prefer to use an "early return" pattern in functions which reduces the amount of indentation later on, like this:

```javascript
function renderMwan3Status(status) {
    if (!status.interface)
        return '<strong><%:No MWAN interfaces found%></strong>';

    var statusview = '';
    ...
}
```

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/mwan3.js
```diff
@@ -0,0 +1,95 @@
...
+		var statusview = '';
+		for ( var iface in status.interfaces) {
+			var state = '';
+			var css = '';
+			var time = '';
+			switch (status.interfaces[iface].status) {
+				case 'online':
+					state = _('Online');
+					time = String.format(
+						'<div><strong>%s:&nbsp;</strong>%s</div>',
```
jow-, line 23: I suggest to replace every `%s` with `%h` which takes care of HTML escaping any special characters in the input. Granted, it is unlikely that the backend status reply values contain `<`, `>`, `&` etc. but you can't fully rule it out (invalid user config, ...).

Since we already had "stored XSS" reports for these kinds of invalid config value issues in the past, it is better to avoid it right away.

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/mwan3.js
```diff
@@ -0,0 +1,95 @@
...
+						_('Uptime'),
+						secondsToString(status.interfaces[iface].online)
+					);
+					css = 'success';
+					break;
+				case 'offline':
+					state = _('Offline');
+					time = String.format(
```
jow-, line 31: Also a matter of taste: you can replace every `String.format('...fmt...', ...args...)` with `'...fmt...'.format(...args...)`

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/diagnostics.js
```diff
@@ -0,0 +1,131 @@
...
+	},
+
+	handleAction: function(ev) {
+		var iface = document.getElementById('iface');
+		var task = document.getElementById('task');
+
+		switch (task.value) {
+			case 'gateway':
+				return this.handleCommand('/usr/libexec/luci-mwan3', [ 'diag', 'gateway', iface.value ]);
+				break;
```
jow-, line 35: You can omit the `break;` statements after the returns.

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/overview.js
```diff
@@ -0,0 +1,45 @@
...
+
+var callMwan3Status = rpc.declare({
+	object: 'mwan3',
+	method: 'status',
+	expect: {  },
+});
+
+document.querySelector('head').appendChild(E('script', {
+	'type': 'text/javascript',
+	'src': L.resource('view/mwan3/mwan3.js')
```
jow-, line 14: You should consider moving the parts of `'view/mwan3/mwan3.js'` which you use in multiple views into a separate helper class, like [`tools/firewall.js`](https://github.com/openwrt/luci/blob/master/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js) or [`tools/network.js`](https://github.com/openwrt/luci/blob/master/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js).

Once you did that, you can simply do `'require tools.mwan3 as mwan3';` and use its `mwan3`'s methods instead of manually injecting a script here.

applications/luci-app-mwan3/root/usr/libexec/luci-mwan3
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	table="$(ip route list table $ID)"
+	table_result="$?"
+
+	if [ "$table_result" = 0 ]; then
+		echo "Routing table \"$ID\" for interface \"$iface\" found"
+		echo "$table"
+	else
+		"Routing table \"$ID\" for interface \"$iface\" not found"
```
jow-, line 146: I guess this is missing an `echo`

applications/luci-app-mwan3/root/usr/share/luci/menu.d/luci-app-mwan3.json
```diff
@@ -0,0 +1,105 @@
...
+		"action": {
+			"type": "view",
+			"path": "mwan3/status/troubleshooting"
+		}
+	},
+
+	"admin/network/mwan3": {
+		"title": "MultiWAN Manager",
+		"order": "600",
+		"action": {
```
jow-, line 49: You could use action type "firstchild" here to avoid the need to repeat the `admin/network/mwan3/globals` path.

jow- (commented):
Great stuff, this looks like it has been a lot of work! A couple of remarks below...

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/overview.js
feckert: Thanks for pointing that out, I didn't know.
However, I can now manage without this require file.
But good to know for next time

feckert:
> Great stuff, this looks like it has been a lot of work! A couple of remarks below...

Yes, that took me a few hours.
But now the app looks better.
No more lua, javascript and html mixed up.

@jow- I have implemented your comments. If that fits, then I would merge it.

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/network/interface.js
```diff
@@ -0,0 +1,285 @@
...
+		o.value('2');
+		o.value('5');
+		o.value('10');
+		o.value('20');
+		o.value('25');
+		o.modalonly = true;
+
+		o = s.option(form.ListValue, "timeout", _("Ping timeout"));
+		o.default = '4';
+		o.value('1', _('%d second').format('1'));
```
jow-, line 170: Use `N_(123, '%d second', '%d seconds')` here to allow translators to use the correct plural form for their respective languages

PR #5186, opened 2021-07-13
luci-app-firewall: Fix translation context for Routing/NAT Offloading…

RikudouPatrickstar:
… in zh_Hans


feckert:
The translation for LuCI is done in [Weblate](https://hosted.weblate.org/projects/openwrt/)

PR #5190, opened 2021-07-15
luci-proto-bonding: fix member because of API change

he-ma:
Signed-off-by: Helge Mader <ma@dev.tdt.de>

This change is necessary as the result of `network.getDevices()` now returns `device` instead of `ifname`,
so the previous version no more works.

### DIFF
--- a/protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
+++ b/protocols/luci-proto-bonding/htdocs/luci-static/resources/protocol/bonding.js
@@ -24,14 +24,16 @@ function getSelectableSlaves(section_id) {
 						var slaves = L.toArray(uci.get('network', interfaces[j]['.name'], 'slaves'));
 
 						for (var k = 0; k < slaves.length; k++) {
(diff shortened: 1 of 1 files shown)


jow-:
You would actually need to support both ifname and device

he-ma:
@jow-
Just for interest, could you explain me, why support for ifname is still required? Did I miss something?
But, OK, I will change this.

he-ma:
@jow- 
I made the change you requested

jow-:
> Just for interest, could you explain me, why support for ifname is still required?

Netifd still supports option ifname and maps it to option device internally. Also not migrated user network configs might contain option ifname yet. I usually try to be backwards compatible in LuCI since people might be restoring old configs from backups etc. and being lenient wrt. supported configuration helps to reduce churn on the user side.

he-ma:
> 
> Netifd still supports option ifname and maps it to option device internally. Also not migrated user network configs might contain option ifname yet. I usually try to be backwards compatible in LuCI since people might be restoring old configs from backups etc. and being lenient wrt. supported configuration helps to reduce churn on the user side.


Thank you very much for your explanation.

he-ma:
@jow- 
Thank you for merging.

Issue #5092, opened 2021-06-01
Access Multiple Luci behind nginx reverse proxy


valm13:
Hi,

I'm trying to create access to multiple OpenWrt Luci through Nginx reverse proxy.

I want to access them with a specif URL, providing the IP of the remote machine: /target-ip/[IP].

Here is the related question I've asked on StackOverflow: https://stackoverflow.com/questions/67787122/multiple-openwrt-luci-behind-nginx-reverse-proxy 

PS: I'm gonna recreate the question on SuperUser

----------Edit----------

I've found a pull request to the Luci repository to support reverse proxy with pathname prefix from a cookie: [Here](https://github.com/openwrt/luci/pull/4285)

Unfortunately, it has not been merged into the official repo, and I can only use official releases.

jow-:
You could try modifying the `lua_prefix` and `ubus_prefix` options of `/etc/config/uhttpd` but I suppose it is easier to do use nginx' `sub_filter` facility to rewrite embedded resource links as needed. See https://serverfault.com/questions/713148/modify-html-pages-returned-by-nginx-reverse-proxy for examples.

valm13:
Thanks for the tip ! Unfortunately, I've tried it, and it didn't work : I was only able to reach the login page
I've found another way to bypass this problem, simply by creating a subdomain for each Luci web app.

PR #5194, opened 2021-07-15
luci-mod-network: clarify device settings "Delete" button

SvenRoederer:
Use proper label text "Unconfigure" for button which "Remove related device settings". This seems to describe best what it's doing. Current "Reset" is misleading and "Delete" (#5090, #5092) are technically not correct.

Also this one should be cherry-picked to openwrt-21.02


SvenRoederer:
@jow-, @pmelange can you review / comment? getting this into 21.02 via master might solve some confusion (see #5090, #5092)

feckert:
@SvenRoederer I cherry-picked your [commit](https://github.com/openwrt/luci/commit/1b58decdabbbf631c89bdd56b63429b6ccc7fb8d) and also updated the i18n for luci-base

PR #5196, opened 2021-07-17
luci-mod-ucentral: initial simplified ui for ucentral firmware

jow-:
The luci-mod-ucentral module implements a heavily stirpped down GUI only
offering essential settings to bootstrap an ucentral client firmware.


feckert:
Does we not need also a backend dependency for the client in **LUCI_DEPENDS:=+ucentral**?

hnyman:
out of curiosity, what is ucentral ?

aparcar:
Check out blogics GitHub account

blogic:
@hnyman I am working on a cloud mgmt stack for OpenWrt, Jow helped me build a theme to allow easy device on boarding. this will all hopefully land in an openwrt feed in 2-3 months

feckert:
If I add a user `ucentral` that only has access to the luci-mod-ucentral, then I get a white page after login. I have to restart the rpcd service, so that I can log in again.

Does not work:
```
config login
        option username 'ucentral'
        option password '$p$root'
        option timeout '300'
        list read 'luci-base'
        list read 'luci-mod-ucentral'
        list read 'unauthenticated'
        list write 'luci-mod-ucentral'
```
If I add `luci-mod-status-index` then it works again. But I have now the additional `Administration` configuration pages.
```
config login
        option username 'ucentral'
        option password '$p$root'
        option timeout '300'
        list read 'luci-base'
        list read 'luci-mod-status-index'
        list read 'luci-mod-ucentral'
        list read 'unauthenticated'
        list write 'luci-mod-status-index'
        list write 'luci-mod-ucentral'
```

jow-:
@feckert - you need the changes from commit https://github.com/openwrt/luci/pull/5196/commits/8faccdaaa8144a5fb420f2dca63e0bdafbe0c560 to avoid the blank page. Without these changes, the restricted ucentral user gets dispatched to the Admin/Logout action which is allowed by luci-base. After the changes linked above (plus flushing session storage to get rid of the cached menu, local cache to update ui.js as well as purging remote caches by rm -r /tmp/luci-*), the blank page issue should be gone.

I didn't yet push the dispatcher rework to master as it requires more testing and since it is a somewhat security sensitive change.

feckert:
@jow- I have installed and replaced all files except `po` from your branch to my system.
I also rebooted my system.
My configuration rpcd configuration is as follows

```
config login
        option username 'test'
        option password '$p$root'
        option timeout '300'
        list read 'luci-base'
        list read 'unauthenticated'
        list read 'luci-mod-ucentral'
        list write 'luci-mod-ucentral'
```

But I still get a white page after login.

If I add `list read 'luci-mod-status-index' `and ` list write 'luci-mod-status-lindex'` it works again.

I have also used your uCentral theme.
Can you confirm this?

feckert:
@jow- I have think the problem is that I have also installed an additional own written app that also adds a submenu.
I will open an issue to address this.

feckert:
I did everything you said.
But unfortunately I still have the problem.
That I get a white page after the login, when the user is only allowed to see or use the app `luci-mod-ucentral.`

```
config login
        option username 'central'
        option password '$p$root'
        option timeout '300'
        list read 'luci-base'
        list read 'luci-mod-ucentral'
        list read 'unauthenticated'
        list write 'luci-mod-ucentral'
```

Interesting, is that after login the Url in the browser looks like this `https://192.168.0.51/cgi-bin/luci`

If I now write behind the URL `ucentral` so that it looks like this  `https://192.168.0.51/cgi-bin/luci/ucentral`, then I can use the app and only have access to this app.

PR #5198, opened 2021-07-18
luci-mod-network: depend on luci-proto-ipv6 (fixes "wan6")

SvenRoederer:
Maintainer: @jow- @hnyman

OpenWrt sets up a wan6 interface by default. Without the "proto-ipv6" this interface is reported "broken / unsupported protocol" in LuCI and can't be edited therefore.
Adding this dependency, fixes the Interface-listing. This dependency will (as in a708378) only be relevant when IPV6-feature is enabled at all.

Tested on: ramips-mt7621
This affects also openwrt-21.02

### DIFF
--- a/modules/luci-mod-network/Makefile
+++ b/modules/luci-mod-network/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=LuCI Network Administration
-LUCI_DEPENDS:=+luci-base +libiwinfo-lua +rpcd-mod-iwinfo
+LUCI_DEPENDS:=+luci-base +libiwinfo-lua +rpcd-mod-iwinfo +IPV6:luci-proto-ipv6
 
 PKG_LICENSE:=Apache-2.0
 


jow-:
Please add the dependency to the luci meta package instead. I would like the subcomponents to have as few dependencies as possible.

SvenRoederer:
Just seen that there is already such dependency in 'luci'.
As I don't include this meta-package (to avoid ppp and firewall dependencies) the proto is not pulled automatically.

Keeping the dependency at "mod-network" seems the most correct choice, as that's where the dependency exists.

jow-:
I don’t agree that this is a required dependency. The luci-mod-network module das not require it to function. It is merely a convenience selection to accommodate the default configuration. That is what the meta packages are for. Subcomponents should only contain hard dependencies that are required for the code to function, not to pull in features.

jow-:
Maybe an alternative solution: I think we should move DHCPv6 support to the base protocol package and only keep the tunnel related proto support (6in4, 6to4, etc.) in proto-ipv6

PR #5200, opened 2021-07-19
luci-theme-material: fix JS error from ES6 method in IE

nating:
The JS `String.prototype.endsWith()` method is not supported in Internet Explorer, which results in the interface not loading after logging in.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#browser_compatibility

Signed-off-by: geoff <geoff@invizbox.com>


nating:
That's looks right @jow-, thanks. This PR can be closed if you're going with that 👍.

PR #5224, opened 2021-08-04
[RFC] luci-mod-status: make index page configurable

Ansuel:
Add support for show/hide index cards.
Card's state is saved using the browser localstorage and is
restored on page reload/relogin.
Rework the pool function to check and skip loading of hidden
cards.
Rework themes to address new button position.

@jow- I don't remember if i ever proposed this. The idea here is that the index page can become heavy and currently we don't have a way to configure them. The old implementation saved the index page preference in the luci config and the index page but i think it wasn't that efficient. This new implementation store the value in the browser localstorage and the pool function is reworked to read a local variable and decide to skip the card loading.

image

When the button is pressed and the card is hidden, a dummy value is displayed until the pool periodic function reload the data in the background
image


modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -171,9 +171,9 @@ return baseclass.extend({
 		}, this)), E('em', _('There are no active leases')));
 
 		return E([
-			E('h3', _('Active DHCP Leases')),
+			E('h3', _('Active IPv4 DHCP Leases')),
 			table,
-			E('h3', _('Active DHCPv6 Leases')),
+			E('h3', _('Active IPv6 DHCP Leases')),
```
systemcrash, line 176: It's my understanding there is no IPv6 DHCP - only DHCPv6... Still, I guess this works.

Ansuel:
@jow- any chance for a respin of this?

systemcrash:
Purely cosmetically hideable panels would be useful (to hide unique or sensitive data) for those who prefer screenshots.

Ansuel:
Notice this was very broken and outdated...

@jow- Working on a better solution... what do you think of this design?

image

(saddly some additional entry in css are still required, is it really a problem?)

jow-:
Frankly... I don't like it.

If additional clicks are required to show that stuff we can as well move it to a different page...

Ansuel:
@jow- no that was just an example by default everything is showed. It's really to add the possibility to add stuff

systemcrash:
I like it. Although I think it has life, it's usefulness is perhaps
limited. The time I spend on that landing page is limited. But for
screenshot and perhaps other use cases it's definitely useful.

Ansuel:
@jow- I updated the first post with some screen for each theme

systemcrash:
Closed by 788e47034becc7806485b40dbef9774a14f8ea9d

PR #5233, opened 2021-08-07
luci-mod-network: merge hosts with dns settings

vgaetera:
The current location for "Network > Hostnames" is confusing.
It is provided by Dnsmasq and partly overlaps with static leases.

Merge "Hostnames" with "Network > DHCP and DNS" as an extra tab.
Also add a simple explanation of the use case.


jow-:
Merged via 96ef7beb66de322411ad9a784953d7ff5a0ddfc1 - thanks!

PR #5239, opened 2021-08-09
luci-base: fix request timeout docstring

aparcar:
The `timeout` is defined in milliseconds, not seconds.

TIL: My server API does not respond within 8ms.


jow-:
Merged, thank you!

PR #5241, opened 2021-08-10
luci-theme-material: standardize menu and drop jquery

Ansuel:
- Drop jquery lib
- Move to the new menu.js implementation
- Rework css to mimic old js transition
- Rework footer and header to new implementation

I need some testing about this. Especially I need someone to tell me the difference between this and the old material theme. Working on this made me forget how the old theme worked. The idea here is to drop jquery and use pure js. We are mature enough now to make this transition since we use flex and all so we don't need ie hacks and stuff anymore (finally)


feckert:
@Ansuel Cool I wanted to tackle that too. I'll have a look at it right away tomorrow.

Have you also fixed the problem with the ACL in this theme?
The menu is always displayed, regardless of whether the user has access or not.

If you add an user to the `/etc/config/rpcd` who has only access to luci-mod-status-index. Then you will see the whole menu tree.
With the other LuCI themes, the menu item is hidden if the user has no access.
```
config login
        option username 'test'
        option password '$p$root'
        option timeout '300'
        list read 'luci-base'
        list read 'luci-mod-status-index'
        list read 'unauthenticated'
        list write 'luci-mod-status-index'
```

Ansuel:
I need to check the acl didn't know this problem was present

Currently while testing this i notice that with the login screen nothing is displayed in the menu. So i think using the standard way to handle the menu autofixed the acl issue?

jow-:
I believe `menu-material.js` is missing yet from your commit.

Ansuel:
Oh no i didn't include that....

Il Mar 10 Ago 2021, 16:24 Jo-Philipp Wich ***@***.***> ha
scritto:

Ansuel:
@feckert i pushed the missing js file

feckert:
Very good job :+1:.I have tested different application and base modules and it works so far. 
There are still one issues with this theme that we should also work on right away? This issue has been around for a while, and is independent of your change.
If, for example, an app/mod is installed that opens its own menu tree, e.g. [ucentral](https://github.com/openwrt/luci/pull/5196), then the selection does not integrate properly into the theme.
luci-theme-material (looks ugly)
Screenshot_2021-08-11 st-dev-07 - Status - LuCI
luci-theme-bootstrap (looks fine)
Screenshot_2021-08-11 st-dev-07 - Status - LuCI(1)

Ansuel:
I suspect that is not supported in the theme and some tweak to css/footer are needed... will compare with openwrt2020 with the theme

Ansuel:
@feckert I'm currently fixing it this way...
What do you think?

image

feckert:
It certainly looks better than before :+1:. Thanks for that.
I also think it's good that you don't have it at the bottom like with the luci-app-bootstrap.
But I wouldn't put it in the line of the `indicators`.
Since there is not only the refresh here.

I would put it in its own line, like the luci-theme-openwrt-2020.
Screenshot_2021-08-11 st-dev-07 - Status - LuCI(2)

Ansuel:
Oh ok so an additional line right under the header. Ok will implement that.

Ansuel:
@feckert something like that?

image

feckert:
Looks good. Maybe you can push that as an additional commit. Then those who use the theme will know that the issue is now also fixed.

buicaocuong9595:
Vào 19:11, Th 3, 10 thg 8, 2021 Ansuel Smith ***@***.***> đã
viết:

feckert:
@Ansuel I have now integrated this into my own build.
And it looks very good so far.
I haven't noticed any problems so far.

The only thing I have seen is that the administration is displayed although I don't have a new configuration tree installied.
Screenshot_2021-08-17 VR2-106149 - Übersicht - LuCI

This should still be changed.
The menu should only be displayed if I have more than one configurations tree.

Ansuel:
@feckert i made a logic error and that triggered the display (the separator is added as a child so it was wrongly detected as a button and the bar was displayed)
Anyway should be fixed. Did you notice any problem with the menu or the use of the theme with a mobile?

feckert:
I have now tested this for a while.
From my point of view, we could include it in the master.
@jow- Or is there something against it?

hnyman:
If it is works, just merge.
Material has been unmaintained/semi-broken for a long time...

(and maybe also for 21.02?   maybe a bit later?)

feckert:
@Ansuel Thanks for this greadt work :+1:
@hnyman I also think that we will first test this in the master and get feedback from the community.
After that it we could then backported this during next maintenance release of the stable branch openwrt-21.02.

hnyman:
Packages are constantly rebuilt also on 21.02, so there is no real hurry.

(So, it will later be available for those who build 21.02 from the 21.02 branch HEAD, but not for those who love to build from the release tag 21.02.0 with fixed feed points)

feckert:
@Ansuel I have now noticed that when I install luci-mod-dashboard, it does not look right. The following things do not fit.
**Menu entry for the overview:**
Screenshot_2021-08-31 st-dev-07 - Übersicht - LuCI
**And as described in https://github.com/openwrt/luci/pull/4185#issuecomment-903845723 the labels**
Screenshot_2021-08-31 st-dev-07 - Übersicht - LuCI(1)

Can you take a look at this?

Ansuel:
@feckert can you test ? https://github.com/openwrt/luci/pull/5331

PR #5242, opened 2021-08-10
luci-mod-network: add acs_exclude_dfs option

feckert:
This option `acs_exclude_dfs` was added in the openwrt core package mac80211
with commit [a5ec41b0e598f73cd1f878a107c480f472dfb15a](https://github.com/openwrt/openwrt/commit/a5ec41b0e598f73cd1f878a107c480f472dfb15a).
This should also be configurable via LuCI.

- This option is only valid when we are in the 5Ghz band. (done)
- Also, the option should only be displayed if the channel selection is set to `auto` on the 5Ghz band. Unfortunately, I was not able to add a depends here. @jow I have already tried everything I could think of. But I couldn't do it because it's its own [CBI form.Value](https://github.com/openwrt/luci/blob/master/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L301). I don't understand what the **_freq** option looks like. (failed)


BashfulBladder:
The frequency list in that rpc call might fail if you don't have the rpcd-mod-iwinfo & iwinfo & rpc installed.

You can obtain the data returned in that call using:
`ubus call iwinfo freqlist '{ "device": "radio0" }'`
get the list of devices from:
`ubus call network.wireless status`

A single channel from my results:

```
ubus call iwinfo freqlist '{ "device": "radio0" }'
{
	"results": [
		{
			"channel": 36,
			"mhz": 5180,
			"restricted": false,
			"active": false
		}
	]
}
```

feckert:
@BashfulBladder Thank you for your advice.
 But that is not what I am looking for. Or I do not understand your suggestion.

My problem is that the option `acs_exclude_dfs` is only valid when we are in the 5GHz range and the frequency selection option is set to auto.
The problem now is that the frequency and band selection is a separate cbi form where I can't access the selection directly what the user has done.
I can only use the virtual option **_freq**.

BashfulBladder:
Ah.

Maybe something like `o.depends(o.formvalue(radioNet.getWifiDeviceName(), "auto"));`

BashfulBladder:
```
o = ss.taboption('general', form.Flag, 'acs_exclude_dfs', _('Disable DFS'), _("Do not use DFS channels on auto selection"));
o.depends({'_freq': '11a', '!contains': true}); //this works for me; it was changed later (to what works for you apparently)
o.depends({'_freq': 'channel', '!contains': 'auto'});
o.default = o.disabled;
```

```
uci show wireless
wireless.radio0=wifi-device
wireless.radio0.type='mac80211'
wireless.radio0.hwmode='11a'
wireless.radio0.path='soc/1b500000.pci/pci0000:00/0000:00:00.0/0000:01:00.0'
wireless.radio0.country='US'
wireless.radio0.cell_density='0'
wireless.radio0.htmode='VHT80'
wireless.radio0.log_level='1'
wireless.radio0.channel='auto'
wireless.radio0.acs_exclude_dfs='1'
```

feckert:
@BashfulBladder After a bit of back and forth, I think I've found a solution that works.
Can you verify whether it works for you too?

```
o = ss.taboption('general', form.Flag, 'acs_exclude_dfs', _('Disable DFS'), _("Do not use DFS channels on auto selection"));
o.depends({'_freq': '5g', '_freq': 'auto', '!contains': true});
```
If you know how to do it, it's not that difficult :joy: 

I would update the P/R

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -944,6 +944,9 @@ return view.extend({
 					o = ss.taboption('general', form.Flag, 'legacy_rates', _('Allow legacy 802.11b rates'), _('Legacy or badly behaving devices may require legacy 802.11b rates to interoperate. Airtime efficiency may be significantly reduced where these are used. It is recommended to not allow 802.11b rates where possible.'));
 					o.depends({'_freq': '2g', '!contains': true});
 
+					o = ss.taboption('general', form.Flag, 'acs_exclude_dfs', _('Disable DFS'), _("Do not use DFS channels on auto selection"));
+					o.depends({'_freq': '5g', '_freq': 'auto', '!contains': true});
```
jow-, line 948: This code makes no sense, the second `_freq` will overwrite the first one.

feckert: @jow Thanks for your replay.
 I just noticed that when I selected 2.4 GHz and the frequency selection isset to `auto`. The `acs_exclude_dfs `option was still displayed. That is not what I want.

Is there a solution at all for this case?
I have not found anything in the entire LuCI repostiory.
Couldn't the validation be updated so that such a query works?
I am not quite sure what the data structure of _freq looks like.
It is obviously due to the special cbi view for _freq so I could not directly select the uci values.

BashfulBladder:
While searching luci.js (trying to figure out why I can't disable a button after rendering), I stumbled upon  the dom (DOM) class with these 2 methods: findClassInstance & callClassMethod. It might be able to get the selection via dom.findClassInstance('channel') & then if dom.callClassMethod('channel','getValue') == 'auto'.

I'm just guessing - this could be useless info.

Also, I imagine you could set a breakpoint in your browser somewhere & inspect the values in the function(s).

PR #5246, opened 2021-08-11
Statistics

selanf:
a11a81c268 statistics: allow combobox theming
4a13836c6c statistics: make interface optional
02449e4e87 statistics: add missing dns types
940664242d statistics: move to status menu


hnyman:
I do not understand the reasoning for removing the interface module from the default selection. For the basic user, it is the main traffic counter / speed indicator in the LuCI stats. (Sure, there is nlbwmon, but that is a separate package.)

Your other changes are stylistic or adding missing types like DNS variants, but this is removal of a core feature ???

selanf:
This pr in another feet belongs to this: https://github.com/openwrt/packages/pull/16319
@hnxman: With the other commit the package is still selected by default, but it could be deselected.
Current state is that there is now way to disable it, and after every sysupgrade it has to be removed again and again

selanf:
Screenshot of "move to status menu"
image

hnyman:
I do not get your reasoning there. Moving the LuCI stats module selection partially to the packages side collectd config, and changing the modules selection logic there, causing changes to the case where somebody uses collectd independently of the LuCI statistics. So far NAK from me regarding that change.

selanf:
As https://github.com/openwrt/packages/pull/16319 was closed i have to remove 1 commit

PR #5247, opened 2021-08-11
Xinetd

selanf:
6df97b75af xinetd: add INTERNAL support
4b8bd6eee9 xinetd: ipv6 support for host fields
5b92e74a80 xinetd: add ipv6 support

Depends:
https://github.com/openwrt/packages/pull/16318

### DIFF
--- a/applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
+++ b/applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
@@ -40,6 +40,12 @@ return view.extend({
 		o = s.option(form.DummyValue, 'server', _('Server'));
 		o.modalonly = false;
 
+		o = s.option(form.DummyValue, 'flags', _('IPv6'));
+		o.cfgvalue = function(section) {
+			return (uci.get('xinetd', section, 'flags') == "IPv6") ? _("yes") : _("no");
+		};
+		o.modalonly = false;
+
 		o = s.option(form.DummyValue, 'disable', _('Enabled'));
 		o.cfgvalue = function(section) {
 			return (uci.get('xinetd', section, 'disable') == "no") ? _("yes") : _("no");
@@ -56,6 +62,13 @@ return view.extend({
 		o.enabled  = 'no';
 		o.disabled = 'yes';
 		o.default  = o.enabled;
+		o.rmempty = false;
+		o.modalonly = true;
+
+		o = s.taboption('basic', form.Flag, 'flags', _('IPv6'), _('Listen on IPv6 additional'));
+		o.enabled  = 'IPv6';
+		o.disabled = 'IPv4';
+		o.default  = o.disabled;
 		o.rmempty = false;
 		o.modalonly = true;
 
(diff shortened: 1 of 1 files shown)


applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
```diff
@@ -71,26 +97,29 @@ return view.extend({
...
+//			var sections = uci.sections('xinetd', 'service');
+//
+//			for (var i = 0; i < sections.length; i++) {
+//				if (uci.get('xinetd', sections[i]['.name'], 'port') == value && section_id != sections[i]['.name'])
+//					return _('Port already in use by service "%s"'.format(sections[i]['.name']));
+//			}
+//
+//		return true;
+//
+//		};
```
feckert, line 111: If this is not working any more, then we should delete the whole validate function or we should fix them.

feckert (changes requested):
I would suggest following changes:
Could you please also remove this line https://github.com/openwrt/luci/blob/master/applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js#L28 So we have the default value `Add`.

And because of your change we have now two options for `type` (UNLISTED and INTERNAL). If we change this `type` option in the advanced tab, it will change the `port` option in the basic tap. I would suggest that we move the option `type` also to the basic tab. So the user will see the change.

I would also make the option `name` not optional so the user have to add a value for this.

applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
selanf: This never worked, even before these commits if there is a tcp port used and the same port as udp (could be different daemons, does not matter).
And now additional it does not work for a service name read from /etc/services. So it is possible to fix it.
"Impossible" is it to know if some other daemon uses it. Maybe during configuration, but not if some additional packages are later installed

On the other side xinetd tell in syslog if a port could not be used - if misconfigured or some other daemon uses it already, and does not start this single service

selanf:
> remove addbtntitle 

I clicked through mounts/leds/dnsmasq/ddns/xinetd/samba/. Some have only "add" and some an "add + what to add". So i dont know what the default should be and just keep it as it is.

>  option name not optional

Yes, good idea. If no port is given (not possible wiht luci, but okay) the port is read by the name from services fil

> port option in the basic tap

Its the easier, but i did not want to change to much first...
`Socket type` sould also be moved, as if you change tcp/udp you want to change in 99% also this.
Imho `Server arguments` should be below `Server` , and not on 2 different tabs
forbidden + allowed hosts whould also not be on 2 tabs, i think both on "advanced" as there are also conn-limit, access-time, number-of-instances
This could be a separate commit

applications/luci-app-xinetd/htdocs/luci-static/resources/view/xinetd/xinetd.js
feckert: Well then we delete the whole thing

selanf:
Now the 1st tab has many items, but all should be set on configuring a service.
The items on the other 2 are not mandatory

Screenshots:

feckert:
I made a few adjustments to the commit message and pushed the changes directly.
Thanks for your contribution :+1:
https://github.com/openwrt/luci/commit/0402de12c4d4b7bf630fc342a62266a046b06f57
https://github.com/openwrt/luci/commit/f4c3b012dbe63013072afd9063580b4e185bddd4
https://github.com/openwrt/luci/commit/9af5a594dcccae2d6d0bb91c82d8154eeab46267
https://github.com/openwrt/luci/commit/2f1f46e51be16192f070898f4ed610ec72c4838d

selanf:
Thx! Its great when i can reduce count of local patches.
I've noted: Use full package name of package at comment begin :)

PR #5249, opened 2021-08-11
theme openwrt-2020: cut fqdn in titles

selanf:


jow-:
Something like `hostonly = hostname:match("^[^%.]+")` should be simpler and more efficient.

Apart from that please don't hardcode `OpenWrt @ ` in the title.

selanf:
Thanks, im not an expoer in javascript!
Why should "OpenWrt" not be hardcoded, and replaced by what? It should not be translated and OpenWrt is not renamed so often :)

jow-:
Yep, `%` is the escape character in Lus regular expression patterns, see https://www.lua.org/pil/20.2.html

As for the OpenWrt remark - OpenWrt is often used by downstream projects that wish to rebrand it to something else, for that there exists a configuration option (https://github.com/openwrt/openwrt/blob/master/package/base-files/image-config.in line 158) which we should respect. Since parsing that from /etc/os-release or /etc/openwrt_version is tedious and introduces further overhead it might be easier to not show any distro name to begin with.

selanf:
Thx!
But it is okay to have "luci" in the title?
Using the name of the web site which opened as the 1st in the title and after that the subpage is common. Just open some webiste. This helps to find the correct browser tab quickly. When you have opened many tabs, only the start of the title is displayed.
What about to add `VERSION_DIST` to a place where it could accessed with less overhead? Maybe ubus?

Btw, you could attach `#L158` to github links

selanf:
As there is yet existing because of the hostname
```
local boardinfo = util.ubus("system", "board") or { }
```
it is enough to add
```
local distrib = boardinfo.release.distribution or "?"
```

jow-:
Problem I see now is that when there's multiple tabs open with different OpenWrt devices, they all start with `OpenWrt - ...`

selanf:
Title hah to set for for every skin
This commit is only a "backport" form dark skin, which im using.
So i not really need this, and as you dont like it i just close it

PR #5263, opened 2021-08-12
ksmbd: fix the layout of textbox

selanf:
without this, only 50% of screen where used
also show version only 1 time if Ksmbd/Kmod
versions match


jow-:
Merged, thank you!

PR #5266, opened 2021-08-14
status: show RECOVERY on overview

selanf:
to make it easy to see if initramfs is running
see https://github.com/openwrt/openwrt/commit/5181af5585c8aab0a04d2bd95c41165d40decd48

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
@@ -52,7 +52,7 @@ return baseclass.extend({
 
 		var fields = [
 			_('Hostname'),         boardinfo.hostname,
-			_('Model'),            boardinfo.model,
+			_('Model'),            boardinfo.model + (boardinfo.rootfs_type == 'initramfs' ? ' - ' + _('RECOVERY') : ''),
 			_('Architecture'),     boardinfo.system,
 			_('Firmware Version'), (L.isObject(boardinfo.release) ? boardinfo.release.description + ' / ' : '') + (luciversion || ''),
 			_('Kernel Version'),   boardinfo.kernel,


hnyman:
There are also other purposes for initramfs than "recovery" (like in E8450/RT3200), so labeling initramfs as "RECOVERY" sounds dubious.  E8450 is a singular exception on that front...

It would be better to show also the "initramfs" there

selanf:
If it should be "initramfs" there could be added another line to status page which shows the "rootfs_type". But imho this help not.
I started last month with openwrt on 5 differen devices, and the initram/sysupgrade thin was very confusing. Nearly the same size files in different formats packed. The wiki explained how to flash, for some even no initram image is needed. It took some time to recognize initamfs is most time only volatile in ram - its some sort of "recovery" system to write a full system into flash
So if you dont like "recovery" it could be replaced by "volatile"?

hnyman:
> and the initram/sysupgrade thin was very confusing. Nearly the same size files in different formats packed. The wiki explained how to flash, for some even no initram image is needed.  It took some time to recognize initamfs is most time only volatile in ram - its some sort of "recovery" system to write a full system into flash

Needing to use the initramfs a rarity. I would guess that with >98% of devices you never use initramfs. It has mostly been used for devices that due to OEM bootloader quircks require extra steps, debugging, development, or something similar.  @dangowrt introducing the recovery tool to E8450/RT3200 is a new thing, which has brought initramfs to increased usage (and as far as I know, that is the only case where a flashed initramfs image is used for permanent recovery partition purposes.)  I claim that >99% of OpenWrt user populace has not heard about initramfs and is not using it, and out of that <1% again less than 1% thinks that initramfs=recovery.

selanf:
For avm7490 is no sysupgrade working currently, so only initramfs
Flash avm1200 is only working by first start initram by adam bootloader, then do in the (reycovery) flash the real system

PR #5269, opened 2021-08-14
luci-app-statistics: add support for MaxMissed

jtkohl:
* Add support for MaxMissed, documented in collectd: https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_ping
* Correct the type of hosts entries, they can be hostnames as well as IP addresses

Tested in concert with a change to packages/collectd-mod-ping to manage the collectd configuration for `MaxMissed`, viewable here: https://github.com/openwrt/packages/pull/16362

I tested setting the value to -1 does not re-resolve hosts after pings fail.
Watched DNS traffic with tcpdump to see that setting the value to positive integer results in a DNS re-resolve after that many ping failures. I swapped my phone between two networks reachable from the same router, and saw that after switching networks, collectd switched to pinging the new address after the desired number of failures.

Signed-off-by: John Kohl <jtk.git@bostonpog.org>

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/ping.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/ping.js
@@ -13,7 +13,7 @@ return baseclass.extend({
 
 		o = s.option(form.DynamicList, 'Hosts', _('Monitor hosts'));
 		o.default = '127.0.0.1';
-		o.datatype = 'ipaddr("nomask")';
+		o.datatype = 'host';
 		o.depends('enable', '1');
 
 		o = s.option(form.ListValue, 'AddressFamily', _('Address family'));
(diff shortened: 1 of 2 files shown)


hnyman:
There should be no default, instead the new option should be defined "optional".  Otherwise LuCI would always set the value, although most users might not want it.

There might instead be a "placeholder" hint value like 10. That would be visible in LuCI but would not get actually set.

jtkohl:
> There should be no default, instead the new option should be defined "optional". Otherwise LuCI would always set the value, although most users might not want it.
> 
> There might instead be a "placeholder" hint value like 10. That would be visible in LuCI but would not get actually set.

OK, getting rid of the default is simple.  The placeholder or hint, I'm a luci novice and haven't yet found an example to follow there.  Can you suggest a package that has one I can read?

hnyman:
>  The placeholder or hint, I'm a luci novice and haven't yet found an example to follow there. Can you suggest a package that has one I can read?

Github search is your friend. Examples for you on "placeholder" usage:
https://github.com/openwrt/luci/search?q=placeholder

Good example from luci-app-statistics:
https://github.com/openwrt/luci/blob/70a749020198367e188369646da9cf5adb767071/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/unixsock.js#L25-L27

applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/ping.js
```diff
@@ -32,6 +32,12 @@ return baseclass.extend({
 		o.default = '30';
 		o.datatype = 'ufloat';
 		o.depends('enable', '1');
+
+		o=s.option(form.Value,'MaxMissed',_('Trigger a DNS resolve after the host has not replied to this number of packets'));
```
jtkohl, line 36: I took this text from the manual page...if that's not OK I can rewrite it

jtkohl: Looking at other pages, I found a better way to do this with more descriptive text, so I pushed a new change.
Looks like this now:
image

PR #5270, opened 2021-08-15
luci-base: support loopback by widgets.NetworkSelect

vgaetera:
Support the "loopback" interface by "widgets.NetworkSelect".
Enable the "loopback" option to include this interface.
Can be used to configure interface binding for services.


jow-:
Merged, thank you!

PR #5271, opened 2021-08-15
module system: startup page enhancements

selanf:
* d0f514fbbd module system: add stoplevel to startup
* dd7ff453ff module system: fix sorting in startup
* 8571727553 module system: add warning to startup
* bfafae724d base: add stoplevel + rename index in luci.sys.init

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js
@@ -74,25 +74,26 @@ return view.extend({
 		var table = E('table', { 'class': 'table' }, [
 			E('tr', { 'class': 'tr table-titles' }, [
 				E('th', { 'class': 'th' }, _('Start priority')),
+				E('th', { 'class': 'th' }, _('Stop priority')),
 				E('th', { 'class': 'th' }, _('Initscript')),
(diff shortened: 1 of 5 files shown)


selanf:
s

modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js
```diff
@@ -74,25 +74,26 @@ return view.extend({
...
+				return a.startlevel - b.startlevel
 
-			return a.name > b.name;
+			return ('' + a.name).localeCompare(b.name);
 		});
 
 		for (var i = 0; i < list.length; i++) {
 			rows.push([
-				'%02d'.format(list[i].index),
+				list[i].startlevel == '255' ? '--' : '%02d'.format(list[i].startlevel),
```
jow-, line 95: Please do a numerical comparison here and move the success case first:
```javascript
list[i].startlevel != 255 ? '%02d'.format(list[i].startlevel) : '---'
```

selanf: I think the condition was read+understandable better before
I've not checked what happens when someone sets in script `START='veryfirst'`

PR #5272, opened 2021-08-15
module system: add memory usage in mb to processes

selanf:
.. as on a 500mb ram system processes below 7.5 mb use 0 or 1 %
also format cpu% + mem% columns a little


selanf:
p

jow-:
Not sure I like adding such redundant information. Maybe always show the actual numbers and drop the percentage or show it in a CBI tooltip.

selanf:
But as long as there is enough space, i think it should stay. On a "smaller" display the MEM% col could be hidden, but i dont know hot to do shits dynamically

PR #5275, opened 2021-08-15
module status: show buffered + cached rows on 0 bytes used too

selanf:
and hide them only on invalid values,
eg missing line in /proc/meminfo


jow-:
Please add spaces around `!=`

PR #5287, opened 2021-08-18
luci-mod-system: hide unsupported option on flash.js

Ansuel:
Hide unsupported option when /rom is not used in flash.js


wulfy23:
thanks for your quick action on this... successfully hides -u ( skip equal to /rom ) on ext4 system.

Tested-by: Imran Khan <gururug@gmail.com>

PR #5301, opened 2021-08-26
luci-mod-network: fix option ip6table in network/interfaces

vgaetera:
Fix the incorrectly displayed value for the option "ip6table".


jow-:
Merged, thank you!

PR #5303, opened 2021-08-26
luci-mod-network: fix route6 source autocomplete

vgaetera:
Fix the wrong value for the "source" option of the "route6" section.


feckert:
Thanks for the fix

Issue #502, opened 2015-10-05
PXE


parasense:
Let's talk about enabling a "PXE" dialog under the Network area some place, probably DHCP/DNS options (aka dnsmasq settings). 

What we want to do is wrap around the '/etc/config/dhcp' configuration parameters for PXE boot.
Like so:

config boot
        option servername 'hostname-of-tftpd-goes here'
        option serveraddress 'ipv4-of-tftpd-goes-here'
        option filename 'pxelinux.o'

That configured OpenWRT to point at a TFTPD server, and it could be a remote system on the LAN, or even the OpenWRT router itself. There is a built-in tftpd service that goes along with dnsmasq, and it really orthogonal to these parameters. One should be able to have a tftpd on OpenWRT, but have PXE point elsewhere on the network should that be desired, or just point of the local dnsmaq built-in tftpd. The point being one could have a bazzillion tftpd services setup around their WAN, but here we configure where the DHCP PXE system points to for network booting. Obviously these parameters only make sense if luci is operating a dhcp service that can do PXE stuff.

LipkeGu:
It seems he means the bootmenue something like this?

![bildschirmfoto vom 2016-12-22 20-37-31](https://cloud.githubusercontent.com/assets/2057932/21437857/835e86ec-c886-11e6-99b6-377d004ebf52.png)

LipkeGu:
I'm thinking about to put this again up when im get it right up again i can fill a pull request with this changes but this will spread in about 3 or more PRs.

jow-:
Closing this as its not supported by OpenWrt and never gained traction. PRs welcome.

vKnmnn:
There is one feature request for this feature. See #3356 
I would also like to see this feature in OpenWRT.

PR #5296, opened 2021-08-24
luci-mod-network: update dhcp and dns labels and descriptions

vgaetera:
Update labels and descriptions for DHCP and DNS settings.
Add missing descriptions and clarify the existing ones.
Group settings by tab and avoid redundant abbreviations.
Consolidate terminology according to the dnsmasq documentation:
https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html
Fix spelling according to the previously adopted policy:
https://github.com/openwrt/luci/commit/88b9d843882cf52a6acf4d08a878fd005120edd4

Signed-off-by: Vladislav Grigoryev <vg.aetera@gmail.com>

PR #5304, opened 2021-08-27
luci-mod-network: add PXE Boot options to TFTP/PXE tab under DHCP & DNS

systemcrash:
manages 'boot' entries in /etc/config/dhcp

Should go into master and 21.02. 
Tested on 21.02-rc4

I think it can also go into 19.07

If I'm not mistaken, I believe this resolves issue https://github.com/openwrt/luci/issues/502 

See: https://openwrt.org/docs/guide-user/base-system/dhcp#booting_options

Screenshot 2021-08-27 at 04 07 17

Screenshot 2021-08-27 at 03 45 06

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -7,6 +7,7 @@
 'require form';
 'require network';
 'require validation';
+'require tools.widgets as widgets';
 
 var callHostHints, callDUIDHints, callDHCPLeases, CBILeaseStatus, CBILease6Status;
 
@@ -250,7 +251,7 @@ return view.extend({
 
 		s.tab('general', _('General Settings'));
 		s.tab('files', _('Resolv and Hosts Files'));
-		s.tab('tftp', _('TFTP Settings'));
+		s.tab('pxe_tftp', _('PXE/TFTP Settings'));
 		s.tab('advanced', _('Advanced Settings'));
 		s.tab('leases', _('Static Leases'));
 		s.tab('hosts', _('Hostnames'));
@@ -467,23 +468,76 @@ return view.extend({
 		o.datatype = 'range(0,10000)';
 		o.placeholder = 150;
 
-		o = s.taboption('tftp', form.Flag, 'enable_tftp',
-			_('Enable TFTP server'));
+		o = s.taboption('pxe_tftp', form.Flag, 'enable_tftp',
+			_('Enable TFTP server'),
+			_('Enable the built-in single-instance TFTP server.'));
 		o.optional = true;
 
-		o = s.taboption('tftp', form.Value, 'tftp_root',
+		o = s.taboption('pxe_tftp', form.Value, 'tftp_root',
 			_('TFTP server root'),
-			_('Root directory for files served via TFTP.'));
+			_('Root directory for files served via TFTP.' +
+			'<br><code>Enable TFTP server</code> and <code>TFTP server root</code> turn on the TFTP server and serve files from <code>TFTP server root</code>.'));
 		o.depends('enable_tftp', '1');
 		o.optional = true;
 		o.placeholder = '/';
 
-		o = s.taboption('tftp', form.Value, 'dhcp_boot',
+		o = s.taboption('pxe_tftp', form.Value, 'dhcp_boot',
 			_('Network boot image'),
 			_('Filename of the boot image advertised to clients.'));
 		o.depends('enable_tftp', '1');
 		o.optional = true;
 		o.placeholder = 'pxelinux.0';
+
+		/* PXE - https://openwrt.org/docs/guide-user/base-system/dhcp#booting_options */
+		o = s.taboption('pxe_tftp', form.SectionValue, '__pxe__', form.GridSection, 'boot', null,
+			_('Special <abbr title="Preboot eXecution Environment">PXE</abbr> boot options for Dnsmasq.'));
+		ss = o.subsection;
+		ss.addremove = true;
+		ss.anonymous = true;
+		ss.nodescriptions = true;
+
+		so = ss.option(form.Value, 'filename',
+			_('Filename'),
+			_('Host requests this filename from the boot server.'));
+		so.optional = false;
+		so.placeholder = 'pxelinux.0';
+
+		so = ss.option(form.Value, 'servername',
+			_('Server name'),
+			_('The hostname of the boot server'));
+		so.optional = false;
+		so.placeholder = 'myNAS';
+
+		so = ss.option(form.Value, 'serveraddress',
+			_('Server address'),
+			_('The IP address of the boot server'));
+		so.optional = false;
+		so.placeholder = '192.168.1.2';
+
+		so = ss.option(form.DynamicList, 'dhcp_option',
+			_('DHCP Options'),
+			_('Options for the Network-ID. (Note: needs also Network-ID.) E.g. "<code>42,192.168.1.4</code>" for NTP server, "<code>3,192.168.4.4</code>" for default route. <code>0.0.0.0</code> means "the address of the system running dnsmasq".'));
+		so.optional = true;
+		so.placeholder = '42,192.168.1.4';
+
+		so = ss.option(widgets.DeviceSelect, 'networkid',
+			_('Network-ID'),
+			_('Apply DHCP Options to this net. (Empty = all clients).'));
+		so.optional = true;
+
+		so = ss.option(form.Flag, 'force',
+			_('Force'),
+			_('Always send DHCP Options. Sometimes needed, with e.g. PXELinux.'));
+		so.optional = true;
+
+		so = ss.option(form.Value, 'instance',
+			_('Instance'),
+			_('Dnsmasq instance to which this boot section is bound. If unspecified, the section is valid for all dnsmasq instances.'));
+		so.optional = true;
+
+		Object.values(L.uci.sections('dhcp', 'dnsmasq')).forEach(function(val, index) {
+			so.value(index, '%s (Domain: %s, Local: %s)'.format(index, val.domain || '?', val.local || '?'));
+		});
 
 		o = s.taboption('hosts', form.SectionValue, '__hosts__', form.GridSection, 'domain', null,
 			_('Hostnames are used to bind a domain name to an IP address. This setting is redundant for hostnames already configured with static leases, but it can be useful to rebind an FQDN.'));


modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -548,7 +603,9 @@ return view.extend({
 			uci.unset('dhcp', section, 'dns');
 		};
 
-		so = ss.option(form.Value, 'mac', _('<abbr title="Media Access Control">MAC</abbr>-Address'));
+		so = ss.option(form.Value, 'mac', 
+			_('<abbr title="Media Access Control">MAC</abbr>-Address(es)'),
+			_('The hardware address(es) of this host, separated by spaces. '));
```
feckert, line 608: Remove space after dot

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -474,6 +476,55 @@ return view.extend({
...
+			_('DHCP Options'),
+			_('Options for the Network-ID. (Note: needs also Network-ID.) E.g. "<code>42,192.168.1.4</code>" for NTP server, "<code>3,192.168.4.4</code>" for default route. <code>0.0.0.0</code> means "the address of the system running dnsmasq".'));
+		so.optional = true;
+		so.placeholder = '42,192.168.1.4';
+
+		so = ss.option(widgets.DeviceSelect, 'networkid',
+			_('Network-ID'),
+			_('Apply DHCP Options to this net. (Empty = all clients).'));
+		so.optional = true;
+		//so.multiple = true; // Unsure whether boot options handles multiple net-ids
```
feckert, line 515: remove comment

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -474,6 +476,55 @@ return view.extend({
...
+		so = ss.option(form.Flag, 'force',
+			_('Force'),
+			_('Always send DHCP Options. Sometimes needed, with e.g. PXELinux.'));
+		so.optional = true;
+
+		so = ss.option(form.Value, 'instance',
+			_('Instance'),
+			_('Dnsmasq instance to which this boot section is bound. If unspecified, the section is valid for all dnsmasq instances.'));
+		so.optional = true;
+		//End PXE
```
feckert, line 526: remove comment

feckert (changes requested):
Please update the commit header with the prefix `luci-mod-network: <your message>`

systemcrash:
OK done. Should I make a separate PR for 21.02?

feckert:
@systemcrash 
I have first merged this [pullrequest](https://github.com/openwrt/luci/pull/5296), as it cleans up the dhcp.js.
Could you please rebase your changes?

jow-:
Please don't add yet another tab. I would prefer to merge this ftp and pxe into one as they're closely related and tftp currently is mostly empty.

systemcrash:
With some extra hinting for TFTP as at https://openwrt.org/docs/guide-user/base-system/dhcp#common_options

Screenshot 2021-08-31 at 14 03 38

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -614,16 +671,34 @@ return view.extend({
...
+		so = ss.option(form.Value, 'hostid',
+			_('IPv6-Suffix (hex)'),
+			_('The IPv6 interface identifier (address suffix) as hexadecimal number (max. 8 chars).'));
+
+		so = ss.option(form.Value, 'tag',
+			_('Tag'),
+			_('Set the given tag for matching hosts.'));
+
+		so = ss.option(form.Flag, 'broadccast',
+			_('Tag'),
```
jow-, line 696: Both options use the `Tag` caption, is this intentional? Shouldn't the latter use something like `_('Broadcast')` as caption? Also the `broadccast` option name looks like a type.

systemcrash: Yeah, held hostage by my keyboard..... needs fixing 🤦 

Fixed.

systemcrash:
OK

systemcrash:
Anything more from my side required?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -467,24 +468,73 @@ return view.extend({
...
+
+		so = ss.option(form.Flag, 'force',
+			_('Force'),
+			_('Always send DHCP Options. Sometimes needed, with e.g. PXELinux.'));
+		so.optional = true;
+
+		so = ss.option(form.Value, 'instance',
+			_('Instance'),
+			_('Dnsmasq instance to which this boot section is bound. If unspecified, the section is valid for all dnsmasq instances.'));
+		so.optional = true;
```
feckert, line 536: We could add a dropdown of dnsmasq section name. So the user does have a selection?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -614,16 +671,34 @@ return view.extend({
...
 		});
 
-		so = ss.option(form.Value, 'hostid', _('IPv6 suffix (hex)'));
+		so = ss.option(form.Value, 'hostid',
+			_('IPv6-Suffix (hex)'),
+			_('The IPv6 interface identifier (address suffix) as hexadecimal number (max. 8 chars).'));
+
+		so = ss.option(form.Value, 'tag',
+			_('Tag'),
+			_('Set the given tag for matching hosts.'));
```
feckert, line 693: We could extend the help text with an example of what this option looks like

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -614,16 +671,34 @@ return view.extend({
...
+			_('The DHCPv6-DUID of this host.'));
 		so.datatype = 'and(rangelength(20,36),hexstring)';
 		Object.keys(duids).forEach(function(duid) {
 			so.value(duid, '%s (%s)'.format(duid, duids[duid].hostname || duids[duid].macaddr || duids[duid].ip6addr || '?'));
 		});
 
-		so = ss.option(form.Value, 'hostid', _('IPv6 suffix (hex)'));
+		so = ss.option(form.Value, 'hostid',
+			_('IPv6-Suffix (hex)'),
+			_('The IPv6 interface identifier (address suffix) as hexadecimal number (max. 8 chars).'));
```
feckert, line 689: We could add her the following line to check the datatype?
`o.datatype = 'hex(8)';`

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -614,16 +671,34 @@ return view.extend({
 			so.value(ipv4, ipaddrs[ipv4] ? '%s (%s)'.format(ipv4, ipaddrs[ipv4]) : ipv4);
 		});
 
-		so = ss.option(form.Value, 'leasetime', _('Lease time'));
+		so = ss.option(form.Value, 'leasetime', 
+			_('Lease time'),
+			_('Host-specific lease time, e.g. <code>2m</code>, <code>3h</code>, <code>5d</code>.'));
```
feckert, line 676: We could add here the following line to give the user a Drop down with prefilled values.
```
so.value('2m', _('2 minutes'));
so.value('3h', _('3 houres'));
```

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Not a section name, but dnsmasq instance. I guess few people use this within their config, so having a special GUI element for it is only extra points. Not sure how to go about finding all of the other instances.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Please extend the wiki with an example of what this looks like.....

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Tried this. When should the error checking occur? Nothing happened in the GUI with this. Only see this in one other place in the code (and I don't use mwan3).  Browser console stackdumps with `SyntaxError: Unhandled token "hex"`. Whatever you were inspired by is broken. Doc refers to non-existent: `LuCI.validation`, which should be linked, but isn't(?).

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Done.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Manpage has:

 The set:<tag> construct sets the tag whenever this --dhcp-host directive is in use. This can be used to selectively send DHCP options just for this host. More than one tag can be set in a --dhcp-host directive (but not in other places where "set:<tag>" is allowed). When a host matches any --dhcp-host directive (or one implied by /etc/ethers) then the special tag "known" is set. This allows dnsmasq to be configured to ignore requests from unknown machines using --dhcp-ignore=tag:!known If the host matches only a --dhcp-host directive which cannot be used because it specifies an address on different subnet, the tag "known-othernet" is set.

The tag:<tag> construct filters which dhcp-host directives are used. Tagged directives are used in preference to untagged ones.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Although the above might not fully relate to how dnsmasq works in openwrt...

So if we really need to limit to a static 16 bit IPv6 suffix (suffixes can be longer...) then e.g.

`		so.datatype = 'and(rangelength(0,8),hexstring)';`

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Tag is basically any arbitrary string, with a few special ones, e.g. "known"

systemcrash:
Added `macwildcard` - MACs can contain wildcards e.g. `*:*:*:*:*:*` as specified in example `dnsmasq.conf` files, which enables e.g. MAC range matching.

systemcrash:
e.g. the entry `dhcp-host=11:22:33:44:55:66,set:red` would be represented by

MAC: 11:22:33:44:55:66
Tag: red

while the entry `dhcp-host=11:22:33:*:*:*,set:red` would be:

MAC: 11:22:33:*:*:*
Tag: red

(If I've interpreted everything correctly)

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Done.

modules/luci-base/htdocs/luci-static/resources/validation.js
```diff
@@ -344,6 +344,12 @@ var ValidatorFactory = baseclass.extend({
 				multicast ? _('valid multicast MAC address') : _('valid MAC address'));
 		},
 
+		macwildcard: function(multicast) {
+			var m = this.value.match(/^([a-fA-F0-9]{2}|\*):([a-fA-F0-9]{2}:|\*:){4}(?:[a-fA-F0-9]{2}|\*)$/);
+			return this.assert(m != null && !(+m[1] & 1) == !multicast,
+				multicast ? _('valid multicast MAC address with wildcard (*)') : _('valid MAC address with wildcard (*)'));
+		},
+
```
feckert, line 352: From my side, that should fit. But I remember that @jow- doesn't like to see new types being added here.

jow-: Right, I don't like highly specific one-off types to be added here. This kind of validation above can go to a `.validate()` function directly in the form.

systemcrash:
Moved it closer to its usage. Should it need to return to `validation.js` one day, it's just a copy/paste away.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -536,19 +594,24 @@ return view.extend({
...
+		var macwildcard = function(multicast) {
+			var m = this.value.match(/^([a-fA-F0-9]{2}|\*):([a-fA-F0-9]{2}:|\*:){4}(?:[a-fA-F0-9]{2}|\*)$/);
+			return this.assert(m != null && !(+m[1] & 1) == !multicast,
+				multicast ? _('valid multicast MAC address with wildcard (*)') : _('valid MAC address with wildcard (*)'));
+		}
+
+		so = ss.option(form.Value, 'mac',
+			_('MAC address(es)'),
+			_('The hardware address(es) of this entry/host, separated by spaces.'));
+		so.datatype = 'list(macwildcard)';
```
jow-, line 606: This won't work. It needs to be something like `so.validate = function(section_id, value) { ... }`

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -536,19 +594,24 @@ return view.extend({
 			uci.unset('dhcp', section, 'dns');
 		};
 
-		so = ss.option(form.Value, 'mac', _('MAC address'));
-		so.datatype = 'list(macaddr)';
+		var macwildcard = function(multicast) {
+			var m = this.value.match(/^([a-fA-F0-9]{2}|\*):([a-fA-F0-9]{2}:|\*:){4}(?:[a-fA-F0-9]{2}|\*)$/);
+			return this.assert(m != null && !(+m[1] & 1) == !multicast,
```
jow-, line 599: `this.assert(...)` cannot be used in this context, neither is the `multicast` parameter available. 

Should be something like:

```javascript
so.validate = function(section_id, value) {
    var macaddrs = L.toArray(value);

    for (var i = 0; i < macaddrs.length; i++)
        if (!macaddrs[i].match(/^([a-fA-F0-9]{2}|\*):([a-fA-F0-9]{2}:|\*:){4}(?:[a-fA-F0-9]{2}|\*)$/))
            return _('Expecting a valid MAC address, optionally including wildcards');;

    return true;
};
```

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -536,19 +594,24 @@ return view.extend({
...
 			    result = [];
 
 			for (var i = 0, mac; (mac = macs[i]) != null; i++)
-				if (/^([0-9a-fA-F]{1,2}):([0-9a-fA-F]{1,2}):([0-9a-fA-F]{1,2}):([0-9a-fA-F]{1,2}):([0-9a-fA-F]{1,2}):([0-9a-fA-F]{1,2})$/.test(mac))
-					result.push('%02X:%02X:%02X:%02X:%02X:%02X'.format(
-						parseInt(RegExp.$1, 16), parseInt(RegExp.$2, 16),
-						parseInt(RegExp.$3, 16), parseInt(RegExp.$4, 16),
-						parseInt(RegExp.$5, 16), parseInt(RegExp.$6, 16)));
+				if (/^([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*)$/.test(mac))
+					result.push('%s:%s:%s:%s:%s:%s'.format(RegExp.$1, RegExp.$2, RegExp.$3, RegExp.$4, RegExp.$5, RegExp.$6));
```
jow-, line 614: The changed code will not normalize MACs anymore, e.g.
`0:1:2:3:4:5` -> `00:01:02:03:04:05`
`a:b:11:FF:00:3` -> `AA:BB:11:FF:00:30`
`0:11:a:*:*:*` -> `00:11:0A:*:*:*`

Maybe consider the following:
```javascript
if (/^([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*):([0-9a-fA-F]{1,2}|\*)$/.test(mac)) {
    var m = [
        parseInt(RegExp.$1, 16), parseInt(RegExp.$2, 16),
        parseInt(RegExp.$3, 16), parseInt(RegExp.$4, 16),
        parseInt(RegExp.$5, 16), parseInt(RegExp.$6, 16)
    ];

    result.push(m.map(function(n) { return isNaN(n) ? '*' : '%02X'.format(n) }).join(':'));
}
```

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -614,16 +678,53 @@ return view.extend({
...
+			_('Lease time'),
+			_('Host-specific lease time, e.g. <code>5m</code>, <code>3h</code>, <code>7d</code>.'));
 		so.rmempty = true;
-
-		so = ss.option(form.Value, 'duid', _('DUID'));
+		so.value('5m', _('5m (5 minutes)'));
+		so.value('3h', _('3h (3 hours)'));
+		so.value('12h', _('12h (12 hours - default)'));
+		so.value('7d', _('7d (7 days)'));
+		so.value('infinite');
```
jow-, line 689: A translation would make sense here, e.g. `_('infinite (lease does not expire)')`

jow- (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Ah yes, thank you, the cache probably hid this between test deploys. Fixed as per recommendations.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: 👍

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: This normalization functions properly.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Good thinking...

systemcrash:
I split the static lease stuff into a separate PR #5401

systemcrash:
ping - according to review this is ready.

Issue #5307, opened 2021-08-27
[wireguard] AllowedIPs should be optional, not required


keith24:
In wireguard, `AllowedIPs` is optional.  Other implementations, like desktop and phone versions of wg do _not_ require any `AllowedIPs`.  All the wireguard tutorials on google fill in this field, so maybe luci devs thought it was required?  Sometimes it _must_ actually be left blank to allow for routing.  

Consider the following configuration, with clients 1 and 2 behind NAT: 

**Myserver**
```
[Interface]
Address = 10.0.0.1/24

[peer]
PublicKey = <client1's pubkey>
AllowedIPs = 10.0.0.2/32

[peer]
PublicKey = <client2's pubkey>
AllowedIPs = 10.0.0.3/32
```

**Client 1**
```
[Interface]
Address = 10.0.0.2/24

[peer]
PublicKey = <myserver's pubkey>
AllowedIPs = 10.0.0.0/24
Endpoint= <myserver's endpoint>

[peer]
PublicKey = <client2's pubkey>
```

**Client 2**
```
[Interface]
Address = 10.0.0.3/24

[peer]
PublicKey = <myserver's pubkey>
AllowedIPs = 10.0.0.0/24
Endpoint= <myserver's endpoint>

[peer]
PublicKey = <client1's pubkey>
```

The two client machines can still ping each other with this setup, because each client will route all traffic for the wg subnet to myserver.  (10.0.0.0/24 as myserver's allowedIPs).  This works if myserver has forwarding wg0 > wg0 enabled in the firewall.  

Now, say client 1 is running OpenWrt, so we are required to give all its peers AllowedIPs: 

**Client 1 (openwrt)**
```
[Interface]
Address = 10.0.0.2/24

[peer]
PublicKey = <myserver's pubkey>
AllowedIPs = 10.0.0.0/24
Endpoint= <myserver's endpoint>

[peer]
PublicKey = <client2's pubkey>
AllowedIPs = 10.0.0.3/32
```

Since the AllowedIPs for client 2 is more specific than for myserver (`/32` vs `/24`), the openwrt client will try to send data to `10.0.0.3` without attempting to route through `10.0.0.1`.  When pinging, this manifests as a "Destination address required" because no endpoint has been set for client 2.  (It wouldn't get through NAT anyway).  

[...]

PR #852, opened 2016-11-15
luci-proto-wireguard: WireGuard VPN Protocol (New)

danrl:
WireGuard is a novel VPN that runs inside the Linux Kernel and utilizes
state-of-the-art cryptography. It aims to be faster, simpler, leaner, and
more useful than IPSec, while avoiding the massive headache. It intends to
be considerably more performant than OpenVPN.  WireGuard is designed as a
general purpose VPN for running on embedded interfaces and super computers
alike, fit for many different circumstances.
It runs over UDP.

Signed-off-by: Dan Lüdtke mail@danrl.com

PR #5310, opened 2021-08-29
luci-proto-wireguard: make AllowedIPs optional

keith24:
From #5307

Honestly, I haven't tested this because I'm not sure how to build openwrt.  In theory it should work because I have tried removing the `AllowedIPs` lines from `/etc/config/network` and the wireguard interface worked fine.  

So I think this commit should fix it without any side effects... but I would appreciate if someone would test it or help me test it.

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -140,16 +140,9 @@ return network.registerProtocol('wireguard', {
 		o.validate = validateBase64;
 		o.optional = true;
 
-		o = ss.option(form.DynamicList, 'allowed_ips', _('Allowed IPs'), _("Required. IP addresses and prefixes that this peer is allowed to use inside the tunnel. Usually the peer's tunnel IP addresses and the networks the peer routes through the tunnel."));
+		o = ss.option(form.DynamicList, 'allowed_ips', _('Allowed IPs'), _("Optional. IP addresses and prefixes that this peer is allowed to use inside the tunnel. Usually the peer's tunnel IP addresses and the networks the peer routes through the tunnel."));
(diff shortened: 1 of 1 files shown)


keith24:
Ok, I tested this change by editing making the edits in `/www/luci-static/resources/protocol/wireguard.js` on a working  openwrt device.  

I was able to delete all `AllowedIPs` for one peer in the GUI.  After saving and applying, I checked`/etc/config/network` and confirmed that the `list allowed_ips` lines were gone.

hnyman:
cc @danrl  @zx2c4 @feckert 

AllowedIPs has been a required field since 2016 when danrl introduced the wireguard protocol support into LuCI with #852.  
Discussion about it has also been in #4047
Not sure if it should be required or optional, so I am CCing more knowledgeable people...

More explanation from the author in issue #5307

danrl:
LGTM.

Indeed, I was under the impression this is a required field back when I implemented the LuCI part of the stack. Thanks for the fix and providing the detailed scenario in he other thread, Kevin.

danrl (approved):
LGTM

feckert:
@keith24 From my point of view, there is nothing to be said against it.
This option is also optional for the proto handler https://github.com/openwrt/openwrt/blob/master/package/network/utils/wireguard-tools/files/wireguard.sh

Therefore I am merging this now

hnyman:
Should this be backported to 21.02?  (the 21.0.20 release tagging will likely happen tomorrow)

Ps. If it is backported, the commit title might be sanitised.

feckert:
> Should this be backported to 21.02? (the 21.0.20 release tagging will likely happen tomorrow)

done https://github.com/openwrt/luci/commit/71fe35ce69b9dbbd3e4ae2d6e1cae12809c01e10

> Ps. If it is backported, the commit title might be sanitised.

@hnyman damn, I overlooked that, sorry :-) you're absolutely right.

PR #5311, opened 2021-08-29
luci-app-attendedsysupgrade: support revision checks

aparcar:
SNAPSHOTS are not real releases and therefore the app always offers an
upgrade, even if running the latest build. To prevent that all SNAPSHOTS
now check for the running revision and if a newer one is available.

Also do a bunch of refactoring based on JavaScript I learned over the
last week.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -9,382 +9,412 @@
 'require dom';
 
 var callPackagelist = rpc.declare({
-	object: 'rpc-sys',
-	method: 'packagelist',
+        object: 'rpc-sys',
+        method: 'packagelist',
 });
 
 var callSystemBoard = rpc.declare({
-	object: 'system',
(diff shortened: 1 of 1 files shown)


feckert:
@aparcar Why did you changed the whole file from tab to spaces?
Ans I also get an error if I do an I18n update for this app.
I think the following line is wrong https://github.com/openwrt/luci/pull/5311/files#diff-6f92da1ee8b945f03c89ae863115e3689eda060789bf3d66401a0f9acde32eb1R43 it does have a wrong quotes for the translation 
also I think the `console.log` is not necessary one line https://github.com/openwrt/luci/pull/5311/files#diff-6f92da1ee8b945f03c89ae863115e3689eda060789bf3d66401a0f9acde32eb1R41

aparcar:
Hiding away the whitespace changes makes things much more enjoyable. I started to use `js-beautify` for linting and it decided this is the way to go... I'm not used to javascript so I just obey the framework of the week.

image

Thanks I didn't realize there is still a console print in there, will fix that in an upcoming release.

Thanks for the review!

feckert:
@aparcar Do you then undo this with the white spaces?
I don't know of any app that has spaces.
We shouldn't start mixing up spaces and tabs now.

PR #5312, opened 2021-08-29
luci-app-ledtrig-usbport: fix allow multiple selection

boretom:
The kernel module `kmod-usb-ledtrig-usbport` allows for multiple selections of
USB ports to trigger a single LED.
See https://www.kernel.org/doc/html/latest/leds/ledtrig-usbport.html

This simplest of fixes allows for selection one or more USB ports to trigger a LED.

The ASUS RT-AC88U router which is getting ported to OpenWrt by arinc9 has got an
USB 3.0 port at the front which will use different internal USB ports, depending
on if you connect an USB 2.0 device or an USB 3.0 device.

This fix/patch was tested on the mentioned router.

Signed-off-by: Thomas Kupper <thomas.kupper@gmail.com>


feckert:
Thanks merged

PR #5314, opened 2021-08-31
[21.02] luci-app-attendedsysupgrade: support revision checks

aparcar:
SNAPSHOTS are not real releases and therefore the app always offers an
upgrade, even if running the latest build. To prevent that all SNAPSHOTS
now check for the running revision and if a newer one is available.

Also do a bunch of refactoring based on JavaScript I learned over the
last week.

(cherry picked from commit f799d550b6389baee5b388eefe638ba9a690f4c5)


aparcar:
Oops, duplicate of https://github.com/openwrt/luci/pull/5360

PR #5315, opened 2021-08-31
luci-app-rosy-fileserver: mark as broken

dibdot:
The LuCI application is unmaintained and broken.


aparcar:
I'd rather drop it or isn't that a think at luci.git?

aparcar:
I'm merging this, long term it should be removed. Makefile does not follow the formal requirements like containing a Maintainer anyway.

PR #5319, opened 2021-08-31
luci-mod-wireguard: fix h/m/s ago display

systemcrash:
luci-i18n-wireguard-*: fix h/m/s ago display

translator note: remember: %d in the output string


jow-:
The missing translations should be empty, and not `%d` otherwise "12s ago" would be simply translated as "12". Also `over a day ago` must be written as `_('over a day ago')` otherwise it will not get translated.

systemcrash:
That's the idea: that translators will understand that `%d` is needed in the translated string. Anyway, this is already an improvement. 

Fixed the `_('over a day ago')`

jow-:
But it will render untranslated strings broken. Either leave the msgstr empty or provide a complete translation string, not just `%d`

systemcrash:
Ah, sucks. Fixed.

PR #5321, opened 2021-08-31
luci-app-attendedsysupgrade: use correct linter settings

aparcar:
The previous settings would use spaces which messed up things since LuCI
uses tabs everywhere.

Also removed a stray console.log

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -9,412 +9,444 @@
 'require dom';
 
 var callPackagelist = rpc.declare({
-        object: 'rpc-sys',
-        method: 'packagelist',
+	object: 'rpc-sys',
+	method: 'packagelist',
 });
 
 var callSystemBoard = rpc.declare({
-        object: 'system',
-        method: 'board'
+	object: 'system',
+	method: 'board',
 });
 
 var callUpgradeStart = rpc.declare({
-        object: 'rpc-sys',
-        method: 'upgrade_start',
-        params: ["keep"]
+	object: 'rpc-sys',
+	method: 'upgrade_start',
+	params: [ 'keep' ],
 });
 
 function get_branch(version) {
-        // determine branch of a version
-        // SNAPSHOT -> SNAPSHOT
-        // 21.02-SNAPSHOT -> 21.02
-        // 21.02.0-rc1 -> 21.02
-        // 19.07.8 -> 19.07
-        return version.replace("-SNAPSHOT", "").split(".").slice(0, 2).join(".");
+	// determine branch of a version
+	// SNAPSHOT -> SNAPSHOT
+	// 21.02-SNAPSHOT -> 21.02
+	// 21.02.0-rc1 -> 21.02
+	// 19.07.8 -> 19.07
+	return version.replace('-SNAPSHOT', '').split('.').slice(0, 2).join('.');
 }
 
 function get_revision_count(revision) {
-        return parseInt(revision.substring(1).split("-")[0])
+	return parseInt(revision.substring(1).split('-')[0]);
 }
 
 function error_api_connect(response) {
-        console.log(response)
-        ui.showModal(_('Error connecting to upgrade server'), [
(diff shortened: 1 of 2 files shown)


aparcar:
@feckert thoughts?

feckert:
@aparcar :+1:
But as far as I know, LuCI (and I do it that way) doesn't use `"` it uses `'` for strings.
But it's up to you if you want to change that.
And the line https://github.com/openwrt/luci/pull/5321/files#diff-6f92da1ee8b945f03c89ae863115e3689eda060789bf3d66401a0f9acde32eb1R44 has also the wrong quotations.
Which linter do you use?

aparcar:
> But as far as I know, LuCI (and I do it that way) doesn't use " it uses ' for strings.

Reworked to use '

>  And the line https://github.com/openwrt/luci/pull/5321/files#diff-6f92da1ee8b945f03c89ae863115e3689eda060789bf3d66401a0f9acde32eb1R44 has also the wrong quotations.

I don't understand what's wrong with the line you mention, it's using " in a print, that seems fine?

> Which linter do you use?

js-beautify https://github.com/beautify-web/js-beautify but I don't really care

feckert:
This line has the wrong quotation its not a **``** this must be changed to `'`.
I would also change this line to 

`_('Could not reach API at %s. Please try again later.').format(response.url)`

but I have not tested it

aparcar:
So using these special ` quotes is like f-strings in python, you can directly paste variables in there. Is there a problem with that?

feckert:
I have seen that there are problems with updating the translation. 
`Updating applications/luci-app-attendedsysupgrade/po/templates/attendedsysupgrade.pot ... standard input:44: warning: unterminated string`

aparcar:
Okay seems like translations shouldn't use these, I'll fix that.

aparcar:
The translation strings should be fixed now.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -9,412 +9,530 @@
...
+			case 200:
+				res = response.json();
+				var image;
+				for (image of res.images) {
+					if (image.type == 'sysupgrade') {
+						break;
+					}
+				}
+				if (image.name != undefined) {
+					var sysupgrade_url = `${server_url}/store/${res.bin_dir}/${image.name}`;
```
feckert, line 149: I was just about to merge and went back over the pullrequest and saw the wrong comma again. Sorry
I would suggest that we use `'` over all.
If you are building a string then I would always use the `format` function.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -9,412 +9,530 @@
...
+
+					var keep = E('input', {
+						type: 'checkbox',
+					});
+					keep.checked = true;
+
+					var fields = [
+						_('Version'),
+						`${res.version_number} ${res.version_code}`,
```
feckert, line 158: The same as above

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -9,412 +9,530 @@
...
+		'notice spinning',
+		E(
+			'p',
+			_('Searching for an available sysupgrade of %s - %s').format(version, revision)
+		)
+	);
+
+	if (version.endsWith('SNAPSHOT')) {
+		response = await request.get(
+			`${server_url}/api/v1/revision/${version}/${target}`
```
feckert, line 342: Also wrong high commas?

feckert (changes requested):
I am not a javascript expert, maybe @jow- can help us, but I don't know your pattern with the `${`.  Is this valid javascript? I haven't seen it in the LuCI yet.
I would suggest we should use the function `format` and always use these `'`. Then we are on the safe side. Otherwise we break the translation?

aparcar:
Please see more about the template strings here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

aparcar:
I've updated the script slightly to work with the openwrt-2020 theme as well.

PR #5323, opened 2021-09-01
luci-app-wireguard: convert to JavaScript

lvoegl:
### luci-app-wireguard
- luci-app-wireguard has been rewritten in JavaScript in a similar style to the old wireguard status app
- `luci.wireguard` rpcd has been rewritten in lua and now also includes the qr code generation and data backend from the old lua status app
- moved `luci.wireguard` rpcd from luci-proto-wireguard into luci-app-wireguard

app-status

### luci-proto-wireguard
- moved the qr code generation from the app into proto-wireguard
- added dependency to luci-app-wireguard

generate qr
qr code


stangri:
Thanks for this tremendous effort!

> ### luci-proto-wireguard
> * moved the qr code generation from the app into proto-wireguard

What was the reason for moving qr code generation from luci app into luci-proto-wireguard?

applications/luci-app-wireguard/htdocs/luci-static/resources/view/wireguard/status.js
```diff
@@ -0,0 +1,223 @@
...
+'require poll';
+
+
+var callGetWgInstances = rpc.declare({
+	object: 'luci.wireguard',
+	method: 'getWgInstances'
+});
+
+
+function bytesToStr(bytes) {
```
jow-, line 14: This entire function can be replaced with `'%1024mB'.format(bytes)`

applications/luci-app-wireguard/htdocs/luci-static/resources/view/wireguard/status.js
```diff
@@ -0,0 +1,223 @@
...
+function generatePeerTable(options, iconSrc) {
+	return E('div', { 'class': 'table cbi-section-table' }, [
+		E('div', { 'class': 'td' },
+			E('img', { 'src': iconSrc, 'class': 'tunnel-icon' })
+		),
+		E('div', { 'class': 'td peer-options' },
+			options.filter(function (option) {
+				return option[2] != null;
+			}).map(function (option) {
+				return generatePeerOption(...option);
```
jow-, line 66: LuCI JS should stay at ES5 yet, can you replace the spread operator with function.apply? `generatePeerOption.apply(null, option)`

applications/luci-app-wireguard/htdocs/luci-static/resources/view/wireguard/status.js
```diff
@@ -0,0 +1,223 @@
...
+return view.extend({
+	load: function () {
+		return callGetWgInstances();
+	},
+	poll_status: function (nodes, ifaces) {
+		Object.keys(ifaces).forEach(function (ifaceName) {
+			var iface = ifaces[ifaceName];
+
+			var section = nodes.querySelector(
+				'[data-section-id="' + ifaceName + '"]'
```
jow-, line 141: This and similar string concat expressions could be replaced with `nodes.querySelector('[data-section-id="%q"]'.format(ifaceName))` for slightly better readability.

jow- (changes requested):
Great work! Thanks a lot for your effort. I added a couple of minor remarks below.

lvoegl:
> Thanks for this tremendous effort!
> 
> > ### luci-proto-wireguard
> > 
> > * moved the qr code generation from the app into proto-wireguard
> 
> What was the reason for moving qr code generation from luci app into luci-proto-wireguard?

Thank you. When you add a new peer you can generate a qr code for it without having to go to the status first. Also I think it rather belongs to the configuration than the status

systemcrash:
Well, if this conversion takes precedence, ensure that it accounts for what should happen in https://github.com/openwrt/luci/pull/5319 (or merge that first and rebase)

feckert:
@systemcrash You are right, but the `wireguard.htm` file is completely repllaced by status.js.
I am more in favour of merging this pull request and then your fix.
I think you need to rebase your pullrequest once this one has been merged.

feckert:
I have tested this now. I noticed the following issue.

I just have the situation on my system that qrencode is not installed and  the button to generate the QR is still displayed but does not work.

Either we include qrencode package as a dependency to luci-app-wireguard.
Or we have to check whether qrencode is installed.

The easiest way is to add qrencode package as a dependency to luci-app-wireguard from my point of view.

systemcrash:
> Either we include qrencode package as a dependency to luci-app-wireguard.
> Or we have to check whether qrencode is installed.
> 
> The easiest way is to add qrencode package as a dependency to luci-app-wireguard from my point of view.

There is no easy way >:D

I would do both - there's always going to be someone or some other package one day that removes qrencode or breaks this, or pkg is missing from a repo or someone downloads only one pkg file etc etc. So both dep and check for qrencode is safe.

lvoegl:
I have now added a message that appears when trying to generate a qr code if qrencode is not installed.

Also with the latest commit the qr code carries more information as shown in the description in the image below

qr option

feckert:
Thanks

After the merge, this function should be update.
https://github.com/openwrt/luci/pull/5323/files#diff-6763953b63f260140aa13661899e49e1b06283df44fb6be93df63d30446a8313R13-R31

There is already a pull request that has to be rebased.
https://github.com/openwrt/luci/pull/5319

Ping @systemcrash

systemcrash:
@lvoegl are the translation strings what the compiler system put into the po files? Or something you manually cooked up?

Because every single `ago` message string lacks the %d (which is partially what #5319 should fix).

hnyman:
@feckert  @lvoegl 

I think that this PR made a strange change to the dependencies, deviating from the usual LuCI practice.

We had here the normal LuCI way: GUI app being on top, pulling in the underlying support packages
luci-app-wireguard --> luci-proto-wireguard + wireguard-tools

Now we strangely have the protocol on top, pulling in the GUI app.  That sounds abnormal to me.
luci-proto-wireguard depends on the GUI app.
Is there any real reasoning for that? 
We still need both in practice, so why the change?

I would like to revert that change.

(I noticed this as I lost the wireguard in my build, as my recipe contains the normal top-level items, usually the luci-app-xxx)

feckert:
luci-app-wireguard was coverted to Javascript.
So it uses now the ubus rpc to print the status. Also the qr code
generation moved from app to proto. So the luci-app-wireguard and
luci-proto-wirguard uses now a ubus rpcd luci.wireguard. Each of them have
a dependency to luci.wireguard. I think we have to fix the dependency or
split luci.wireguard to work independent. I would prefer the dependency fix

Hannu Nyman ***@***.***> schrieb am So., 26. Sept. 2021,
09:09:

PR #5330, opened 2021-09-03
luci-mod-status: move index acl into own file

feckert:
The ACL file luci-mod-status.json is already quite long.
It is also a separate menu item in LuCI, so in my view it makes sense to
move it to a separate file. An additional positive effect is that the
file then becomes clearer.


jow-:
Merged, thank you!

PR #5335, opened 2021-09-04
luci-mod-network: fix route and rule options

vgaetera:
Allow adding routes with unspecified interface.
This is required for prohibit, blackhole and unreachable routes.

Allow adding routes with loopback interface.
This can work as a blackhole route.

Fix netmask to prefix conversion for the routes.
Identify the IP family by the UCI section type.

Avoid setting the routing table textvalue for the rules.
The rules should target none table unless specified otherwise.


jow-:
Merged, thank you!

PR #5348, opened 2021-09-08
luci-base: show default forward in firewall zone

feckert:
The overview for the zone setting does not show the zone forwarding as it really is.
In reality the general forward always applies, since the sub chains do not have a default.

**Before Change:**
master-Krusty-63-FW-Zonen-IST

**After Change:**
master-Krusty-63-FW-Zonen-SOLL


jow-:
For me the change is confusing, I read it as "forward to xxx is rejected", therefore I tend to NAK it in this particular form.

Maybe a completely different visualization should be used.

feckert:
How about this?
Image Pasted at 2021-9-10 14-46

I just want to make it clear that the default forward action always applies in the end on zone to zone forwarding

feckert:
@jow- How about this?
Screenshot_2021-09-30 dev - General Settings - LuCI

jow-:
Maybe we should pull the zone forwarding visualization and configuration out of the zone declarations into a separate table / list widget. So the table would only contain the colored zone name badge, policy nat and action buttons, then there would be another table/gridview/whatever. displaying and configuring the forwarding policies

Neustradamus:
@feckert: Have you progressed on it?

Neustradamus:
@feckert: Good job, thanks and thank to @systemcrash for merging!

PR #5355, opened 2021-09-11
fix wrong ed25519 information

neheb:
ECDH is not used for the ed25519. The scheme is called EdDSA.

Curve25519 is a Montgomery curve. ed25519 uses an Edwards curve. The
author recommends X25519 as a replacement for Curve25519 in any form.


aparcar:
Could you please point me to the source? I see that ECDH should be replacet but I'm not sure about Curve25519.

PR #5358, opened 2021-09-12
luci-app-attendedsysupgrade: add README

aparcar:
The app should be explained.


aparcar:
Added

PR #5378, opened 2021-09-20
luci-mod-network: support network.interface.sourcefilter

vgaetera:
Allow to modify the sourcefilter option in the network.interface config.
This is required for custom policy routing with IPv6 MWAN and/or NAT6.


feckert:
Since I have never seen the option before and have not used it, I looked in the [documentation](https://openwrt.org/docs/guide-user/network/ipv6/start?#protocol_dhcpv6) and found that it is only available for dhcpv6.
Should the option only be added for this proto?

feckert:
Thanks

PR #5388, opened 2021-09-24
luci-app-https-dns-proxy: add ControlD as provider

ByteAndNibble:
luci-app-https-dns-proxy: add ControlD as provider

Signed-off-by: Aleem Hossain <aleem.hossain@tectonicinteractive.com>

### DIFF
--- a/applications/luci-app-acl/po/bs/acl.po
+++ b/applications/luci-app-acl/po/bs/acl.po
@@ -1,0 +1,144 @@
+msgid ""
+msgstr ""
+"PO-Revision-Date: 2021-09-22 10:48+0000\n"
+"Last-Translator: Dino Mesic <undeene@gmail.com>\n"
+"Language-Team: Bosnian <https://hosted.weblate.org/projects/openwrt/"
+"luciapplicationsacl/bs/>\n"
+"Language: bs\n"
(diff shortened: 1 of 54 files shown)


stangri:
@ByteAndNibble Thanks! Could you please undo change to `.gitignore`?

@hnyman I don't usually include changes to translated `.po`-files in my PRs, is there a downside to that?

hnyman:
Not much downside, assuming they are not broken. (They will be regenerated/fixed/formatted at the next i18n-sync run in any case)

ByteAndNibble:
> @ByteAndNibble Thanks! Could you please undo change to `.gitignore`?

@stangri I can; but may I ask why? `.idea` is a standard inclusion in my `.gitignore` for my workspace; and for others too, I'm sure. I'm not saying 'no'; I'd just like to understand 'why', given it doesn't hurt.

hnyman:
Well, the first thing is that it has absolutely nothing to do with luci-app-https-dns-proxy

Slipping it it semi-hidden without any mention in the PR explanation or the commit message is bad practice.
It should have its own commit and PR, so that it might be discussed...
(And I have never seen .idea so far, no idea to what it is related)

ByteAndNibble:
understood. Screwed up the squash commit, so I'm going to close this PR and raise a new one.

PR #5391, opened 2021-09-26
luci-app-wireguard: fix dependency

feckert:
This is a untested fix for https://github.com/openwrt/luci/pull/5323#issuecomment-927247499
@hnyman  I don't have a chance to test this right now (Maybe next week).
But I wanted to provide a fix as soon as possible, since I'm the one who merged this change :-)


hnyman:
It would still be nice to understand why @lvoegl originally switched the dependency. Was there any real reason?

lvoegl:
> It would still be nice to understand why @lvoegl originally switched the dependency. Was there any real (technical) reason?

I thought it was most common to put the rpcd into the app with most of them being there while the app and proto belong to each other anyways.

PR #5399, opened 2021-09-30
luci-proto-wireguard: fix luci.wireguard rpcd dependency

feckert:
should fixes #5393

PR #5400, opened 2021-09-30
luci-proto-wireguard: display interface public key

systemcrash:

We have private keys, but ***the thing we need to share***, the public key, is completely absent from the `Interface` GUI.

FIXED.

Tested on 21.02.0
Cooperates with the "Generate Key" function below it. 

This PR also implicitly needs #5399

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -11,6 +11,13 @@ var generateKey = rpc.declare({
 	expect: { keys: {} }
 });
 
+var getPublicAndPrivateKeyFromPrivate = rpc.declare({
+	object: 'luci.wireguard',
+	method: 'getPublicAndPrivateKeyFromPrivate',
+	params: ['privkey'],
+	expect: { keys: {} }
+});
+
 var generateQrCode = rpc.declare({
 	object: 'luci.wireguard',
 	method: 'generateQrCode',
@@ -88,14 +95,34 @@ return network.registerProtocol('wireguard', {
 		o.validate = validateBase64;
 		o.rmempty = false;
 
+		var sections = uci.sections('network');
+		var serverName = this.getIfname();
+		var server = findSection(sections, serverName);
+
+		o = s.taboption('general', form.Value, 'public_key', _('Public Key'), _('Base64-encoded public key of this interface for sharing.'));
+		o.rmempty = false;
+		o.write = function() {/* write nothing */};
+
+		o.load = function(s) {
+			return getPublicAndPrivateKeyFromPrivate(server.private_key).then(
+				function(keypair) {
+					return keypair.pub || '';
+				}, 
+				function(error){
+					return _('Error getting PublicKey');
+			}, this)
+		};
+
 		o = s.taboption('general', form.Button, 'generate_key', _('Generate Key'));
 		o.inputstyle = 'apply';
 		o.onclick = ui.createHandlerFn(this, function(section_id, ev) {
 			 return generateKey().then(function(keypair) {
 				var keyInput = document.getElementById('widget.cbid.network.%s.private_key'.format(section_id)),
-				    changeEvent = new Event('change');
+					changeEvent = new Event('change'),
+					pubKeyInput = document.getElementById('widget.cbid.network.%s.public_key'.format(section_id));
 
 				keyInput.value = keypair.priv || '';
+				pubKeyInput.value = keypair.pub || '';
 				keyInput.dispatchEvent(changeEvent);
 			});
 		}, s.section);
--- a/protocols/luci-proto-wireguard/root/usr/libexec/rpcd/luci.wireguard
+++ b/protocols/luci-proto-wireguard/root/usr/libexec/rpcd/luci.wireguard
@@ -16,6 +16,14 @@ local methods = {
 			return {keys = {priv = prv, pub = pub}}
 		end
 	},
+	getPublicAndPrivateKeyFromPrivate = {
+		args = {privkey = "privkey"},
+		call = function(args)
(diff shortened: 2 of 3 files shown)


protocols/luci-proto-wireguard/root/usr/libexec/rpcd/luci.wireguard
```diff
@@ -15,6 +15,14 @@ local methods = {
 			return {keys = {priv = prv, pub = pub}}
 		end
 	},
+	getPublicKeyFromPrivate = {
+		args = {privkey = "privkey"},
+		call = function(args)
+			local pubkey = sys.exec("echo '" .. args.privkey .. "' | wg pubkey 2>/dev/null"):sub(1, -2)
+
+			return {keys = {priv = args.privkey, pub = pubkey}}
```
lvoegl, line 23: I don't think you need the private key to be returned here. Wouldn't it be better to only return the public key?

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -88,14 +95,33 @@ return network.registerProtocol('wireguard', {
...
+		o = s.taboption('general', form.Value, 'public_key', _('Public Key'), _('Base64-encoded public key of this interface for sharing.'));
+		o.rmempty = false;
+
+		o.load = function(s) {
+			return getPublicKeyFromPrivate(server.private_key).then(
+				function(keypair) {
+					return keypair.pub || '';
+				}, 
+				function(error){
+					return 'Error getting PublicKey';
```
lvoegl, line 111: error message is not translated

```suggestion
					return _('Error getting PublicKey');
```

protocols/luci-proto-wireguard/root/usr/libexec/rpcd/luci.wireguard
systemcrash: This function is extendable, and can be co-opted for other purposes later, when the private key is (also) desirable. This also means we can perform a sanity check whereby we verify what we sent to the lower layers. The very creation of this new function is ample reason why both should be included.

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
systemcrash: Fair point...

protocols/luci-proto-wireguard/root/usr/libexec/rpcd/luci.wireguard
```diff
@@ -15,6 +15,14 @@ local methods = {
 			return {keys = {priv = prv, pub = pub}}
 		end
 	},
+	getPublicKeyFromPrivate = {
+		args = {privkey = "privkey"},
+		call = function(args)
+			local pubkey = sys.exec("echo '" .. args.privkey .. "' | wg pubkey 2>/dev/null"):sub(1, -2)
```
jow-, line 21: This is a shell injection with root privileges. You should use https://openwrt.github.io/luci/api/modules/luci.util.html#shellquote as it has been done in 44445a8097d05dbcc807c95c5b2c016f1a49a350

jow- (changes requested):

protocols/luci-proto-wireguard/root/usr/libexec/rpcd/luci.wireguard
systemcrash: Sure

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -88,14 +95,33 @@ return network.registerProtocol('wireguard', {
 		o.validate = validateBase64;
 		o.rmempty = false;
 
+		var sections = uci.sections('network');
+		var serverName = this.getIfname();
+		var server = findSection(sections, serverName);
+
+		o = s.taboption('general', form.Value, 'public_key', _('Public Key'), _('Base64-encoded public key of this interface for sharing.'));
+		o.rmempty = false;
```
jow-, line 103: Just for understanding, is this new option supposed to be editable by the user or is this just an informational display thing? In the latter case consider either using [`form.DummyValue`](https://openwrt.github.io/luci/jsapi/LuCI.form.DummyValue.html) or [`o.readonly = true`](https://openwrt.github.io/luci/jsapi/LuCI.form.Value.html#readonly)

systemcrash: Also no clear way to update the `public` key HTML when using `form.DummyValue`, and one generates a new key(pair). I thought I came up with a reason why not to `o.write` but seem to have forgotten what that was.

systemcrash:
ping @jow- @feckert

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -88,14 +95,34 @@ return network.registerProtocol('wireguard', {
...
+			}, this)
+		};
+
 		o = s.taboption('general', form.Button, 'generate_key', _('Generate Key'));
 		o.inputstyle = 'apply';
 		o.onclick = ui.createHandlerFn(this, function(section_id, ev) {
 			 return generateKey().then(function(keypair) {
 				var keyInput = document.getElementById('widget.cbid.network.%s.private_key'.format(section_id)),
 				    changeEvent = new Event('change');
+				var pubKeyInput = document.getElementById('widget.cbid.network.%s.public_key'.format(section_id));
```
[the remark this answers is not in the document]
feckert: I agree, but this is how it is done.
I also find the other one clearer

feckert:
I have pushed the change to my system.
Works as expected
Could you please update your `Signed-off-by`? That is not a valid mail address
`Signed-off-by: Paul Dee <systemcrash@users.noreply.github.com>`

feckert:
Sorry for the delay. Thanks for this feature :+1:

PR #5401, opened 2021-09-30
luci-mod-network: Static Leases improvements

systemcrash:
Under DHCP and DNS, Static Leases tab:
Add extra fields and booleans
Add 'ignore' value for ip field
Add description text for fields
Enable wildcards in mac addresses


systemcrash:
Static Leases:

Screenshot 2021-08-27 at 05 14 19

alexeys85:
Multiple "Tag" fields on the screenshot.

systemcrash:
> Multiple "Tag" fields on the screenshot.

Already fixed via parent PR

systemcrash:
ping @jow- @feckert

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -615,20 +619,33 @@ return view.extend({
...
+				if (!macaddrs[i].match(/^([a-fA-F0-9]{2}|\*):([a-fA-F0-9]{2}:|\*:){4}(?:[a-fA-F0-9]{2}|\*)$/))
+				return _('Expecting a valid MAC address, optionally including wildcards');
+
+			return true;
+		};
 		so.rmempty  = true;
 		so.cfgvalue = function(section) {
 			var macs = L.toArray(uci.get('dhcp', section, 'mac')),
-			    result = [];
+				result = [];
```
feckert, line 637: Incorrect indentation @jow- prefers spaces for the variable declaration.
See https://github.com/openwrt/luci/blob/0640e487b5db524a29dff4b325bd458de0f0be7b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js#L653

feckert, line 637: Incorrect indentation @jow- prefers spaces for the variable declaration.
https://github.com/openwrt/luci/blob/0640e487b5db524a29dff4b325bd458de0f0be7b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js#L651-L653

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -615,20 +619,33 @@ return view.extend({
...
 
 			for (var i = 0, mac; (mac = macs[i]) != null; i++)
```
feckert, line 639: Why do we not use the forEach pattern https://www.w3schools.com/jsref/jsref_forEach.asp?
Or
Use the iteration like in https://github.com/openwrt/luci/pull/5401/files#diff-90caf036d2316d858be5df6b382c44d4130a66c52cc7b6fda8ee715971a4f9bdR628

feckert (commented):
Everything else looks fine so far

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Not sure - it's ES5 - is that what OWRT 21.02 aims for?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Good catch. Fixed.

systemcrash:
@jow- any other insights here?

systemcrash:
Ready for merge....?

aetha:
Bump! This looks very handy, would really like to see it mainlined. 🙏

systemcrash:
@jow- ?

BKPepe:
Do we accept contribution where is used GitHub's noreply address? :thinking:

shodanx2:
Hello

This modification might fulfill the following feature request
https://forum.openwrt.org/t/add-dns-setting-to-static-lease-configuration/151646

See also these other requests which somewhat relate

https://forum.openwrt.org/t/add-default-gateway-setting-to-static-lease-configuration/151644

I initially believe that a static DNS or Gateway could be assigned per device, but then I learned about the "tag" system

In this new change, it is great to see valid tags listed in the DHCP static lease dialog, however is there a way for the user to create new tags from the user interface or is the only way to manually edit /etc/config/dhcp ?

I think it would also help to suggest pre-made tags for example

deadend gateway (just don't provide internet access to a specific device)
Gateway through a VPN
Gateway through secondary WAN or multiwan
DNS third party high security DNS (rather than ISP provided DNS)
DNS through local server such as a pi-hole
Gateway to Fakeinternet, for hostile devices which need to be contained
Gateway to IP filtered  connection (for instance windows devices which one would want to prevent to talking to any microsoft server via filter rules)
Gateway to IP filtered ban list (block certain sites at certain time of day for only certain devices, similar to parental controls but enabled by DHCP/mac address filtering)

I think this covers most possible scenarios of common users. Having these as ready to go suggestions would empower users by making them realize what is possible and then helping them implementing those solutions quickly.

systemcrash:
@jow- please merge :)

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -841,16 +859,53 @@ return view.extend({
...
+		so = ss.option(form.Value, 'tag',
+			_('Tag'),
+			_('Set the given tag (used to match hosts).'));
+		so.value('known');
+		so.value('!known', _('!known (not known)'));
+		so.value('known-othernet', _('known-othernet (on different subnet)'));
+
+		so = ss.option(form.Value, 'instance',
+			_('Instance'),
+			_('Dnsmasq instance to which this boot section is bound. If unspecified, the section is valid for all dnsmasq instances.'));
```
jow-, line 894: I suppose this should read "DHCP section" not "boot section"

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -841,16 +859,53 @@ return view.extend({
...
+		so = ss.option(form.Value, 'hostid',
+			_('IPv6-Suffix (hex)'),
+			_('The IPv6 interface identifier (address suffix) as hexadecimal number (max. 8 chars).'));
+		so.datatype = 'and(rangelength(0,8),hexstring)';
+
+		so = ss.option(form.Value, 'tag',
+			_('Tag'),
+			_('Set the given tag (used to match hosts).'));
+		so.value('known');
+		so.value('!known', _('!known (not known)'));
```
jow-, line 889: That looks more like a tag to match, not a tag name to set. Is this `!known` choice really correct?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -763,20 +767,33 @@ return view.extend({
...
-		so.datatype = 'list(macaddr)';
+		so = ss.option(form.Value, 'mac',
+			_('MAC address(es)'),
+			_('The hardware address(es) of this entry/host, separated by spaces.'));
+		so.validate = function(section_id, value) {
+			var macaddrs = L.toArray(value);
+
+			for (var i = 0; i < macaddrs.length; i++)
+				if (!macaddrs[i].match(/^([a-fA-F0-9]{2}|\*):([a-fA-F0-9]{2}:|\*:){4}(?:[a-fA-F0-9]{2}|\*)$/))
+				return _('Expecting a valid MAC address, optionally including wildcards');
```
jow-, line 778: Indentation looks slightly off here

jow- (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Taken from: https://openwrt.org/docs/guide-user/base-system/dhcp 

Actually, it reads `host` in the relevant context - good catch.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: As used in the dnsmasq man page:
https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html


The set:<tag> construct sets the tag whenever this --dhcp-host directive is in use. This can be used to selectively send DHCP options just for this host. More than one tag can be set in a --dhcp-host directive (but not in other places where "set:<tag>" is allowed). When a host matches any --dhcp-host directive (or one implied by /etc/ethers) then the special tag "known" is set. This allows dnsmasq to be configured to ignore requests from unknown machines using --dhcp-ignore=tag:!known If the host matches only a --dhcp-host directive which cannot be used because it specifies an address on different subnet, the tag "known-othernet" is set.

jow-: Yeah, so this option is about *setting* a tag for the host, right? At least the description says so. I understand the man page you quoted in a way that "known", "!known", "other-network" are tag values set by dnsmasq itself which you can use to match but the way I understand it, it makes little sense to set those values manually, so they shouldn't be in the default value suggestions.

Unless the uci "set" option is for *matching* tags - in this case the description would be wrong

systemcrash: You're right - split it into the two available fields.

systemcrash:
Added some helpful text from the man page.

jow-:
Merged, thanks!

shodanx2:
Looking forward to give this a try

Another thing I did not list in my  above comment

However it is not part of the static lease configuration per say

Would be dhcp "white list mode"

Which would be a state where new, unknown devices are given a restricted or no internet access 

In whitelist mode only whitelisted devices get full network access

When an unknown device connects to the network, there could be a status page notification to router admin about what to do about the device (full access, restricted internet, fakeinternet, segregate to quarantine subnet, segregate to hostile subnet)

This is to accomodate unknown and potentially hostile devices such as smart TVs, apple, android phones which will call home, export data and then download and execute unauthorized payloads when given free internet access.  

Likewise, fakeinternet is to disable their own network capabilities when refused internet access.

I know that feels like an advanced feature that few people will use but, I think it all can be done with a few scripts and no extra software. 

I think a default openwrt router should be able to handle hostile devices and work in hostile environement out of the box by just pressing a button

systemcrash:
> Is there a way to assign a default tag to unknown devices ?

`!known`

Read the dnsmasq man page. You'll be enlightened.

shodanx2:
Apparently, it is not possible to assign a default tag to unknown new and/or unknown old clients 

The --dhcp-range apparently allows assigning a tag to a range

--dhcp-range=set:unknown,192.168.0.50,192.168.0.150,255.255.255.0,12h

but then, do known clients also get tagged as unknown ?

Does this means unknown clients cannot be on the same subnet and can never communicate (except through the router perhaps) ?

Also, does the openwrt user interface allow the user to apply this tag to the default range in an intuitive way ? (I don't have access to openwrt device at the moment but I don't recall seeing an option to this effect)

systemcrash:
You can to do what you want today already:

Set a MAC, (or a wildcard MAC (or whatever tags) if you are using this code already), then set the `leasetime` to the keyword `ignore`.

With this code, or by manually editing the dhcp config, you can enter `!known` in the `match_tag` field.

PR #5406, opened 2021-10-04
luci.mk: respect `version` file in package folders

aparcar:
LuCI generates package versions either by the last git commit or the
content of the `SOURCE_DATE_EPOCH` env variable. This commit allows to
use the content of a `version` file if present.

This should allow a better reproducibility when creating releases. The
`version` file per folder can be created by the (extended) maintenance
script which is used to generate OpenWrt releases.

### DIFF
--- a/luci.mk
+++ b/luci.mk
@@ -74,7 +74,9 @@ LUCI_LIBRARYDIR = $(LUA_LIBRARYDIR)/luci
 # 1: everything expect po subdir or only po subdir
 define findrev
   $(shell \
-    if git log -1 >/dev/null 2>/dev/null; then \
+    if [ -f ./version ]; then \
+      cat ./version; \
+    elif git log -1 >/dev/null 2>/dev/null; then \
       set -- $$(git log -1 --format="%ct %h" --abbrev=7 -- $(if $(1),. ':(exclude)po',po)); \
       if [ -n "$$1" ]; then
         secs="$$(($$1 % 86400))"; \


jow-:
Change looks fine, please explain the use case in more detail.

aparcar:
I'm thinking that we should offer archives for every point release which result in a reproducible build of the buildbot. Right now this requires full git clones some people doesn't seem to use, which corrupts their $(AUTORELEASE) variables. Using a `version` file and additionally a `version.date` for every package included in a release should make things reproducible.

I'd extend the `makerelease.sh` script at some point to spit out a archive containing all those files. It should be stored next to the build targets.

hnyman:
> Right now this requires full git clones some people doesn't seem to use, which corrupts their $(AUTORELEASE) variables. 

"some people doesn't seem to use"???  Like it is a choice made by them?

The default behaviour of the main OpenWrt is to make shallow copies of all feeds.  Most people do not even realize that.

The default of shallow copies makes AUTORELEASE dubious, as was pointed out to you in the AUTORELEASE commit discussion.

I think that due to AUTORELEASE, it might be better clone the feeds with full git history by default. That would decrease unnecessary confusion.  Trying to mitigate that "shallow clone" with adding even more complexity sounds a bit unncessary.

aparcar:
> "some people doesn't seem to use"??? Like it is a choice made by them?

Yes, I think they download the release archive provided by GitHub.

> The default behaviour of the main OpenWrt is to make shallow copies of all feeds. Most people do not even realize that.

We could update the default feeds to use `git-src-full`? https://github.com/openwrt/openwrt/blob/master/feeds.conf.default

> The default of shallow copies makes AUTORELEASE dubious, as was pointed out to you in the AUTORELEASE commit discussion.

AUTORELEASE is as *dubious* as LUCI_VERSION since they rely on `git log`. 

I'm wondering how companies approach our files. If they do so by cloning our code and adding commits, fine. If they use a release archive and start working on that, using their own version control system, AUTORELEASE and LUCI_VERSION will always be broken.

hnyman:
> >     "some people doesn't seem to use"??? Like it is a choice made by them?
>
> Yes, I think they download the release archive provided by GitHub.

I think that it is much more simple: they simply clone the Openwrt main source, and let the feed repos to be created automatically with ./scripts/feeds, and  - simsalabim - they got a shallow copy.  Getting a full feed repo into the buildroot requires some effort.

> AUTORELEASE is as dubious as LUCI_VERSION 

Sure, LUCI_VERSION has similar trouble. I have pointed out that earlier in 
https://github.com/openwrt/luci/pull/2257#issuecomment-611910608

> >  The default behaviour of the main OpenWrt is to make shallow copies of all feeds. Most people do not even realize that.
>
> We could update the default feeds to use git-src-full? https://github.com/openwrt/openwrt/blob/master/feeds.conf.default

Dropping the shallow feeds would definitely help with AUTORELEASE. Full git clones might cause a bit more web traffic and somewhat more disk space usage, but that would make versioning more uniform.

aparcar:
@hnyman so maybe opt for this? https://github.com/openwrt/openwrt/pull/4669

PR #5418, opened 2021-10-07
luci-theme-bootstrap: fix overlay display failures

### LINKED ISSUE #5405: luci needs some layout love for mobile
Tried fixing a few things on my mobile recently - which is a nice capability. Can this be fixed?

Make the display-layout as big as the text, even if it's bigger than the view port. 

IMG_0400

systemcrash:
this fixes #5405 - you can verify these fixes by "inspect element" for the overlay elements, and modifying their CSS properties.

Fixed overlay display failures for users on mobile, or users who booted up in safe mode XD. Noticeable when vertical or horizontal real-estate is too small for the displayed overlay.

This failure is most evident when you "edit" an interface, or display your unsaved changes.

When content is wider than the screen view, the table hugs the (widest) content maximally, and the overlay window becomes scrollable.

Disabled `-webkit-overflow-scrolling`, `left`, and `right`, because both FF and Safari disable them as invalid anyway.

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -2199,12 +2199,12 @@ div.cbi-value var.cbi-tooltip-container,
 	position: fixed;
 	top: 0;
 	bottom: 0;
-	left: -10000px;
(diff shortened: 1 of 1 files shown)


jow-:
NAK, this change causes scrolling and overflow issues on iOS.

PR #5423, opened 2021-10-08
luci-theme-bootstrap: various tweaks and cleanups

jow-:
- drop unused navigation style rules and selectors
 - use sticky positioning for header bar and drop top margin hacks
 - use flex box layout for header bar contents
 - use uniform line-height for button elements instead of pixel paddings
 - fix too bright .cbi-value bottom border in mobile dark mode
 - avoid rendering duplicate .cbi-value bottom borders for nested sections
 - simplify header markup, get rid of unused container elements
 - use non-wrapping flex box layout for page actions, prioritize primary
   action button in automatic width calculation


themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -31,19 +31,24 @@
...
-	--text-color-medium-delta-l: calc(64 / 255 * 100%);
-	--text-color-medium-l: calc(var(--text-color-high-l) + var(--text-color-delta-l-sign) * var(--text-color-medium-delta-l));
+	--text-color-highest-h: 0;
+	--text-color-highest-s: 0%;
+	--text-color-highest-l: 0%;
+	--text-color-highest: hsl(var(--text-color-highest-h), var(--text-color-highest-s), var(--text-color-highest-l));
+	--text-color-high-h: var(--text-color-highest-h);
+	--text-color-high-s: var(--text-color-highest-s);
+	--text-color-high-delta-l: calc(64 / 255 * 100%);
+	--text-color-high-l: calc(var(--text-color-highest-l) + var(--text-color-delta-l-sign) * var(--text-color-highest-delta-l));
```
hnyman, line 41: @jow- 
There is no `text-color-highest-delta-l`. It should be high-delta here, I think

I got also quite strange menu colors. Looks like the background definition for the top bar is missing?, and the main title also gets appended to right from the menu

image

jow-: I forgot to include markup changes to header.htm snd footer.htm in a previous iteration of this PR, did you update those files as well in your test?

hnyman: Yes, I took the whole current PR as a patch, including changes to all 5 files.

"From 86fac0337530956443c8e290ed8dad3a07d609a4 Mon Sep 17 00:00:00 2001"
should be the current one.

jow-: Okay, I'll do some more thorough testing later. I guess cssmin broke some rules causing the broken header you observed.

teohhanhui: What's the actual benefit of minifying when the vast majority of use cases is LAN-local? 🙈

hnyman: > What's the actual benefit of minifying 

Flash space. The saving is approx. 19%, but naturally that is only 9 kB (of 49 kB). 
(Not sure if that is actually worthwhile, as the readbility/editability decreases a lot.)

Source vs. minified:
```
root@router1:~# wc /tmp/cascade.css
     2449      5372     49518 /tmp/cascade.css
root@router1:~# wc  /rom/www/luci-static/bootstrap/cascade.css
        0      1172     40185 /rom/www/luci-static/bootstrap/cascade.css
```

jow-:
@hnyman - thanks for confirming. I udpated the rules again to fix csstidy related breakage and pushed the changes to master.

PR #5424, opened 2021-10-09
luci-app-dawn: remove page for steering options

PolynomialDivision:
The luci-app-dawn is blocking the PR to be merged:
https://github.com/openwrt/packages/pull/16299

Remove the configuration page and set appropriate standard values.


hnyman:
You mean that the whole dawn config ability is to be removed from LuCI, and the LuCI app remains only as a status viewing tool???

PolynomialDivision:
Yes?

PR #5425, opened 2021-10-09
CI: add formal tests

aparcar:
Migrate the same formal tests used in packages.git

### DIFF
--- a/.github/workflows/formal.yml
+++ b/.github/workflows/formal.yml
@@ -1,0 +1,63 @@
+name: Test Formalities
+
+on:
+  pull_request:
+
+jobs:
+  build:
+    name: Test Formalities
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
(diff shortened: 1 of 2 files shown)


.github/workflows/formal.yml
```diff
@@ -0,0 +1,63 @@
...
+          fetch-depth: 0
+
+      - name: Determine branch name
+        run: |
+          BRANCH="${GITHUB_BASE_REF#refs/heads/}"
+          echo "Building for $BRANCH"
+          echo "BRANCH=$BRANCH" >> $GITHUB_ENV
+
+      - name: Test formalities
+        run: |
```
feckert, line 26: I just saw that you checked in about CI.
Can't we outsource the run commands?
So that when we do the ci steps, we only call a wrapper script that abstracts the CI handling.

```
- name: Determine branch name
- run: ci --get-branch
```

The advantage would be that I can also run the ci commands without always triggering the CI.
In my view, this makes CI development simpler to implement and to test.
Or is there another possibility?

aparcar: Please have a look at https://gitlab.com/ynezz/openwrt-ci/ as well as my additional changes https://gitlab.com/aparcar/openwrt-ci/-/commits/staging

Overall this approach is much nicer than using GitHub directly since it allows local testing.

PR #5426, opened 2021-10-09
CI: test build LuCI packages

aparcar:
While all LuCI packages should always build fine, this allows others to
directly install those packages within test setups.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,71 @@
+name: Test Build
+
+on:
+  pull_request:
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


aparcar:
@jow- I tested again without other packages and it doesn't work because e.g. [lua.h](https://github.com/openwrt/luci/runs/3848010958?check_suite_focus=true#step:5:2899) is missing.

aparcar:
I changed the default behavior to compile only one package which doesn't have as many dependencies. Most other packages shouldn't pull in too many dependencies neither. Let's see how this goes.

PR #5366, opened 2021-09-15
luci-theme-bootstrap: add dark theme variant

teohhanhui:
Add a dark theme variant which is enabled by default when
prefers-color-scheme is set to dark.

Signed-off-by: Teoh Han Hui <teohhanhui@gmail.com>

PR #5428, opened 2021-10-09
luci-theme-bootstrap: fix log text legibility

teohhanhui:
Increase the contrast of read-only text areas.

Signed-off-by: Teoh Han Hui <teohhanhui@gmail.com>

fixes https://github.com/openwrt/luci/pull/5366#issuecomment-938762102

p/s: [The `readonly` attribute is not valid on the `button` and `select` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly), so I've removed those.


jow-:
Hmm, in case a CBI map is readonly due to insufficient permissions or ACL settings, all child widgets are marked `readonly`, even buttons and the like. If we remove the selectors without changing form.js accordingly, those widgets will lose their visual disabled indication.

Can you update your commit to not remove the selectors for now.

teohhanhui:
Closing as I believe this is superseded by #5423

PR #5437, opened 2021-10-11
luci-mod-admin-full: hide arp-scan if not installed

aparcar:
This should be less confusing for users.


aparcar:
image

PR #5438, opened 2021-10-12
[WIP] luci-app-vnstat2: refactoring

he-ma:
@selanf
@janh 

This is a huge change to the already existing vnstat2 LuCI module with some
improvements and some new features.

"Graphs Page":

  - Implementation of refreshing the graphs.

  - Only in the UCI configuration existing interfaces will be shown in any
    case. Before this change, all interfaces existing in the database were
    shown.

  - Introduced a button for each interface to clear the statistics (in fact
    this is removing and adding again the interface from/to the database).

    Attention: This breaks the recently introduced "2x3 Graphs display"

General:

  - Updated translations, added missing translations for ./po/de

  - Renamed the menu entry from "vnStat Traffic Monitor" to "Traffic Monitor" only

luci-Screenshot_20211012_110937

Signed-off-by: Helge Mader <ma@dev.tdt.de>

### DIFF
--- a/applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/graphs.js
+++ b/applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/graphs.js
@@ -1,9 +1,65 @@
 // This is free software, licensed under the Apache License, Version 2.0
 
 'use strict';
+'require poll';
 'require view';
 'require fs';
 'require ui';
+'require uci';
+'require rpc';
+
+var RefreshIfaces = "";
+var RefreshTabs = ['s', 't', '5', 'h', 'd', 'm', 'y'];
+
+var callServiceList = rpc.declare({
+	object: 'service',
+	method: 'list',
+	params: [ 'name' ],
+	expect: { '': {} }
+});
+
+var isReadonlyView = !L.hasViewPermission() || null;
+
+function RefreshGraphs() {
+	RefreshTabs.forEach(function (id) {
+		RefreshIfaces.forEach(function (iface) {
+			fs.exec_direct('/usr/bin/vnstati', [ '-' + id, '-i', iface, '-o', '-' ], 'blob').then(function(res) {
+				document.getElementById('graph_' + id + '_' + iface).src = URL.createObjectURL(res);
+			});
+		});
+	});
+}
+
+function IfacesResetData(ev) {
+	ui.showModal(_('Delete data for ALL interfaces'), [
+		E('p', _('The data will be removed from the database permanently. This cannot be undone.')),
+		E('div', { 'class': 'right' }, [
+			E('div', {
+				'class': 'btn',
+				'click': L.hideModal
+			}, _('Cancel')),
+			' ',
+			E('div', {
+				'class': 'btn cbi-button-negative',
+				'click': function(ev) {
+					var if_count = 0;
+
+					RefreshIfaces.forEach(function (iface) {
+						fs.exec_direct('/usr/bin/vnstat', [ '--remove', '-i', iface, '--force' ], 'blob').then(function() {
+							fs.exec_direct('/usr/bin/vnstat', [ '--add', '-i', iface ], 'blob').then(function() {
+								if_count++;
+								if (if_count == RefreshIfaces.length) {
+									RefreshGraphs();
+								}
+							});
+						});
+					});
+					ui.hideModal();
+				}
+			}, _('Delete'))
+		])
+	]);
+}	
 
 return view.extend({
 	renderTab: function(ifaces, style, title) {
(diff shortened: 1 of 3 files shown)


applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/graphs.js
```diff
@@ -16,41 +65,46 @@ return view.extend({
...
-
-			return (L.isObject(json) ? L.toArray(json.interfaces) : []).map(function(iface) {
-				return iface.name;
-			}).sort();
-		}).catch(function(err) {
-			ui.addNotification(null, E('p', { 'style': 'white-space:pre' }, [err]));
-			return [];
-		});
+		return Promise.all([
+			L.resolveDefault(callServiceList('vnstat'), {}),
```
janh, line 83: This fails with a "Access denied" response. You would need to add something like the following to the ACL file:
```
"ubus": {
	"service": [ "list" ]
},
```
(This probably works for your, because you installed another package that allows this call.)

applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/graphs.js
```diff
@@ -1,9 +1,58 @@
...
+				'class': 'btn',
+				'click': L.hideModal
+			}, _('Cancel')),
+			' ',
+			E('div', {
+				'class': 'btn cbi-button-negative',
+				'click': function(ev) {
+					RefreshIfaces.forEach(function (iface) {
+						fs.exec_direct('/usr/bin/vnstat', [ '--remove', '-i', iface, '--force' ], 'blob').then(function() {
+							fs.exec_direct('/usr/bin/vnstat', [ '--add', '-i', iface ], 'blob').then(function() {
```
janh, line 45: This will silently fail if the user only has read-only permission. The ACL only allows calls to `vnstat --remove` and `vnstat --add` if the user has full access. It probably makes sense to disable this button when there is only read-only access (similar to the delete button in the config page).

applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/graphs.js
```diff
@@ -16,41 +65,46 @@ return view.extend({
...
+
+			(isRunning == false) ? E('p', { 'class': 'alert-message warning' }, _('Warning: The service is not running, graphs will not be updated!')):E('p'),
+	
+			E('div', { 'class': 'right', 'style': 'margin-bottom:-1.5em' }, [
+				E('button', {
+					'class': 'cbi-button',
+					'click': IfacesResetData
+				}, [ _('Clear data for all interfaces') ]),
+				' ',
+			]),
```
janh, line 107: The button is hidden behind the tabs on narrow screens.

janh (changes requested):
I tested your code and noticed some issues (mostly ACL-related).

janh:
> > * I noticed that the graphs seem to be unnecessarily loaded twice after navigating to the graphs page.
> 
> I couldn't see this behaviour. Having a debug console open in browser (Firefox) the first `RefreshGraphs` is called through the poll callback after 60 seconds. Can you tell how you came up to this problem?

I am also using Firefox and reproducing this doesn't take any special actions here. The network tab in the developer tools shows all graphs being loaded twice when navigating to the graphs page. Setting a breakpoint in `RefreshGraphs` also shows that it is first called directly after navigation to the graphs page (and then the next time a minute later).

Another small (and mostly cosmetic) issue I noticed is that there is no handling for graph images that fail to load. Now that the graphs page displays all configured interface (instead of the ones in the database), this can happen when vnStat is not running, or when a monitored interface is not up yet. I think an easy way to solve this could be to hide any images that failed to load, and display a message about this?

applications/luci-app-vnstat2/htdocs/luci-static/resources/view/vnstat2/graphs.js
```diff
@@ -1,9 +1,60 @@
...
+	expect: { '': {} }
+});
+
+var isReadonlyView = !L.hasViewPermission() || null;
+
+function RefreshGraphs() {
+	RefreshTabs.forEach(function (id) {
+		RefreshIfaces.forEach(function (iface) {
+			fs.exec_direct('/usr/bin/vnstati', [ '-' + id, '-i', iface, '-o', '-' ], 'blob').then(function(res) {
+				document.getElementById('graph_' + id + '_' + iface).src = URL.createObjectURL(res);
```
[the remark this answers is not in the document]
he-ma: Should be corrected with new version of PR

he-ma:
@janh 

> > > * I noticed that the graphs seem to be unnecessarily loaded twice after navigating to the graphs page.
> > 
> > 
> > I couldn't see this behaviour. Having a debug console open in browser (Firefox) the first `RefreshGraphs` is called through the poll callback after 60 seconds. Can you tell how you came up to this problem?
> 
> I am also using Firefox and reproducing this doesn't take any special actions here. The network tab in the developer tools shows all graphs being loaded twice when navigating to the graphs page. Setting a breakpoint in `RefreshGraphs` also shows that it is first called directly after navigation to the graphs page (and then the next time a minute later).
> 

Could you please try the code of the updated PR.
Sorry, I don't see this behaviour. Without offending you, could it be that you missed the different parameters when the graphs are created, as the calls to vnstati look very very similiar?

> Another small (and mostly cosmetic) issue I noticed is that there is no handling for graph images that fail to load. Now that the graphs page displays all configured interface (instead of the ones in the database), this can happen when vnStat is not running, or when a monitored interface is not up yet. I think an easy way to solve this could be to hide any images that failed to load, and display a message about this?

Well, I added a hint if a graph is not available.

he-ma:
@janh 
As it is some time ago I posted a notice to you, maybe you missed my last comment?
Please have a look at it and give some feedback.

Thanks in advance.

janh:
Sorry for the late response, I entirely forgot about this.

> Sorry, I don't see this behaviour. Without offending you, could it be that you missed the different parameters when the graphs are created, as the calls to vnstati look very very similiar?

I am still seeing the same issue using a freshly built OpenWrt master image with your changes, i.e. `RefreshGraphs` is called immediately after navigation to the graphs page, and thus all graphs are loaded twice (there are 14 graph requests with only a single configured interface). This is the same in either Firefox or Chrome.

A few other notes:

As the clear button is now at the bottom of the page, the negative margin should be removed (currently it can overlap with the footer).

It looks like the `RefreshGraphs` is called for every interface in `IfacesResetData`. Is this actually necessary? It seems that is should be enough to do that once after all interfaces were handled.

he-ma:
@janh 

Sorry for the delay for my response but I had been very busy over the last two weeks

> > Sorry, I don't see this behaviour. Without offending you, could it be that you missed the different parameters when the graphs are created, as the calls to vnstati look very very similiar?
> 
> I am still seeing the same issue using a freshly built OpenWrt master image with your changes, i.e. `RefreshGraphs` is called immediately after navigation to the graphs page, and thus all graphs are loaded twice (there are 14 graph requests with only a single configured interface). This is the same in either Firefox or Chrome.

I still can't reproduce this behaviour (see attached image). Having only one interface configured I only get seven requests, which seems OK to me. I really don't know how we should proceed.

 Screenshot_20220203_130818

> As the clear button is now at the bottom of the page, the negative margin should be removed (currently it can overlap with the footer).

I fixed this.

> It looks like the `RefreshGraphs` is called for every interface in `IfacesResetData`. Is this actually necessary? It seems that is should be enough to do that once after all interfaces were handled.

OK, I modified the code. Now refreshing of the graphs is only done once if all interfaces are processed.

I will upload the updated  PR soon.

he-ma:
@janh 
Hello Jan,
just to keep you informed.
Now I dived deeper, threw away our modified distribution and installed (the vanilla) OpenWrt SNAPSHOT r18779-64125ed1d0 and LuCI Master (git-22.025.79016-22e2bfb). Additionally installed the vnstat2, vnstat2 package and my  luci-app-vnstat2.

Now thankfully I can reproduce the problem - but up to now I have no idea where the 'magic' happens in contrast to our distro.
I'll let you know if I have any news.

janh:
> So, to be more general, **every** view which does a `poll.add()` ought to have a similar problem

Yes, looking at a few other LuCI apps/modules it seems that some rely on `poll.add` immediately calling the function, while others don't. Maybe it would be a good idea to open a separate issue about this, so a proper fix can be implemented?

A few examples that rely on `poll.add` calling the function immediately, resulting in data only being loaded after the interval has first elapsed:
https://github.com/openwrt/luci/blob/c6a5c809ed1fb050d257cf6273b8bc832babfa9b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/load.js#L287 https://github.com/openwrt/luci/blob/d24e3295f004c87722ad59927738f208a2a84ac5/modules/luci-mod-status/htdocs/luci-static/resources/view/status/iptables.js#L363 https://github.com/openwrt/luci/blob/c6a5c809ed1fb050d257cf6273b8bc832babfa9b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js#L420

A few examples that seem to always work, and also don't load data twice:
https://github.com/openwrt/luci/blob/70a749020198367e188369646da9cf5adb767071/modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js#L60 https://github.com/openwrt/luci/blob/35df2adaf8a2c5b4fa61f58049f409ca087c0547/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L1574

he-ma:
So,
I'd like to 'trigger' jow to merge this PR. Is this OK for you?

stangri:
> So, I'd like to 'trigger' jow to merge this PR. Is this OK for you?

You probably meant to tag @jow-.

PR #5442, opened 2021-10-13
luci-base: show validation message in save error

lvoegl:
When saving your changes with an invalid input field the message is always generic. I have added the validation message to the validation-failure event and now you get the precise message also in the save error popup.

### Generic Error
generic error message

### Custom Error
custom validation message


jow-:
Merged, thank you!

PR #5443, opened 2021-10-13
luci-app-netdata: add application for netdata

shenpuvip:
This luci app supports netdata.

Depends on [package/netdata](https://github.com/openwrt/packages/tree/master/admin/netdata) .

Signed-off-by: Pu Shen <shenpuvip@gmail.com>

### DIFF
--- a/applications/luci-app-netdata/Makefile
+++ b/applications/luci-app-netdata/Makefile
@@ -1,0 +1,11 @@
+# This is free software, licensed under the Apache License, Version 2.0 .
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for Netdata
+LUCI_DEPENDS:=+netdata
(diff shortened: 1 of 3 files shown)


BKPepe (changes requested):
I am not sure if I find this useful. You are trying to use an iframe to embed the standalone web UI into LuCI. This could lead to unwanted issues (responsibility, etc.). On the other hand, I don't think this is desired. Because you can embed all the Web UI into LuCI, would you guarantee its compatibility and that it will work with each theme? What if someone configured to run Netdata on a different port. How it is handled?

Also, this is the first application that does this.

From my side - NAK.

BKPepe:
Can we close this?

BKPepe:
@hnyman Please close this pull request. Thanks! :) 
My reasoning to close this is above and OP no longer replied.

PR #5444, opened 2021-10-13
luci-app-hd-idle: Improve content to display

psiegl:
1) Show transport (SATA, USB, ...), vendor and model of disk
2) Offer and show only disks that rotate (Linux is a bit conservative, has e.g. issues with USB drives, but from my perspective ok for now).
3) Condense the menu so that it looks similar to System/Mount Points/Mount Points.

Screenshot_20211013_210925
Screenshot_20211013_213520

### DIFF
--- a/applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
+++ b/applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
@@ -2,42 +2,94 @@
 'require form';
 'require fs';
 'require view';
+'require uci';
+
+function disk(devs, options, section_id) {
+	var v = uci.get('hd-idle', section_id, 'disk') || '';
+	var disk = devs.find(function(itm){ return itm.name == v; });
+	var out = '';
+	if(disk != undefined){
+		out = options.map(function(opt){ return disk[opt].trim(); });
+		out = out.filter(function(o){ return o != ''; });
+		out = out.join(' ');
+	}
+	return E('span', out);
+}
+
+function prettytime(section_id) {
+	return E('span', (uci.get('hd-idle', section_id, 'idle_time_interval') || '')
+			 + ' '
+			 + (uci.get('hd-idle', section_id, 'idle_time_unit') || ''));
+}
 
 return view.extend({
 	load: function() {
-		return fs.list('/dev').then(function(devs) {
-			return devs.filter(function(dev) {
-				return dev.type == 'block' ? dev.name.match(/^[sh]d[a-z]$/) : false;
-			});
+		return fs.exec("/usr/bin/lsblk", ["-n", "-J", "-do", "NAME,TRAN,ROTA,RM,VENDOR,MODEL"]).then(function(res) {
+			if( res.code )
+				return [];
+			var json = JSON.parse(res.stdout);
+			return ( 'blockdevices' in json ) ? json['blockdevices'] : [];
 		});
 	},
 
 	render: function(devs) {
 		var m, s, o;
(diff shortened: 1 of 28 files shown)


applications/luci-app-hd-idle/root/usr/share/rpcd/acl.d/luci-app-hd-idle.json
```diff
@@ -3,7 +3,8 @@
 		"description": "Grant UCI access for luci-app-hd-idle",
 		"read": {
 			"file": {
-				"/dev": [ "list" ]
+				"/dev": [ "list" ],
```
jow-, line 6: Since you removed the `fs.list('/dev')` call you can most likely drop the `"/dev": [ "list" ],` permission as well.

applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
```diff
@@ -2,43 +2,89 @@
 'require form';
 'require fs';
 'require view';
+'require uci';
+
+function disk(devs, option, section_id) {
+	var v = uci.get('hd-idle', section_id, 'disk') || '';
+	var e = devs.findIndex(function(itm) { return itm.name == v; } );
```
jow-, line 9: Maybe use `devs.find()` instead of `devs.findIndex()` ? Saves the extra `devs[e]` and `e != -1` checks later on.

applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
```diff
@@ -2,43 +2,89 @@
...
+	return E('span', e != -1 ? option.map(o => devs[e][o]).filter(o => o).join(' ').trim() : '');
```
jow-, line 10: Please don't use arrow functions (yet), we're still aiming for ES5 syntax compatibility atm.

applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
psiegl: I changed it to `.find()`, however I still have a `!= undefined`.

applications/luci-app-hd-idle/htdocs/luci-static/resources/view/hd_idle.js
```diff
@@ -2,43 +2,89 @@
...
+			if( res.code )
+				return [];
+			var json = JSON.parse(res.stdout);
+			return ( 'blockdevices' in json ) ? json['blockdevices'] : [];
 		});
 	},
 
 	render: function(devs) {
 		var m, s, o;
-		m = new form.Map('hd-idle', _('HDD Idle'), _('HDD Idle is a utility program for spinning-down external disks after a period of idle time.'));
```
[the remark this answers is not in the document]
jow-: Oh, I didn't mean to ask you to change the foreign language strings, just the english original, something like `sed -i -e 's#spinning-down external disks#spinning-down disks#g' applications/luci-app-hd-idle/po/*/*.po`

zdz2019:
It's broken.

Snipaste_2021-10-17_10-06-06

That's it after saving.

psiegl:
> It's broken.
> 
> Snipaste_2021-10-17_10-06-06
> 
> That's it after saving.

Yep, you are correct. There was a typo in the function declaration. Now it is fixed.

zdz2019:
@psiegl  I think your changes are okay, just waiting to be merged, but I don’t know when it will merge.

chunfengyao:
When I open `/cgi-bin/luci/admin/services/hd_idle`, HDD will be activated every time. possibly reading vendor's info and model name of disk will access the disk. so, is there any way to get vendor and model of disk without access disk, or just did the essential jobs like previous version?

PR #5471, opened 2021-10-30
luci-app-openwisp: add luci-app-openwisp

nickberry17:
LuCI support for configuring OpenWISP agent

Screenshots of UI
openwisp1

openwisp 2


nickberry17:
ping @feckert

applications/luci-app-openwisp/htdocs/luci-static/resources/view/openwisp.js
```diff
@@ -0,0 +1,101 @@
...
+        o.value('disabled', _('Disabled'));
+        o.value('monitoring', _('Enabled'));
+
+        o = s.taboption('general', form.Value, 'url', _('Server URL'), _('The URL to the OpenWISP server. Example: https://openwisp2.mynetwork.com'))
+        o.placeholder = 'https://openwisp2.mynetwork.com'
+
+        o = s.taboption('general', form.Value, 'shared_secret', _('Shared Secret'), _('The organization shared secret for auto enrollment.'))
+        o.password = true
+
+        o = s.taboption('general', form.Value, 'interval', _('Update Interval'), 'How often to check in with the OpenWISP server. Expressed in seconds.')
```
systemcrash, line 38: Shouldn't that be `_('How often...')`?

nickberry17: Fixed up that and the others

nickberry17:
> 'others' not fixed. Also, please ensure the po files get the same strings.

Not sure what you mean.  I had missed wrapping a couple other strings in the translation, but that has been done.  I also updated the .pot using the script.  Should be ready

applications/luci-app-openwisp/root/etc/uci-defaults/40_luci-openwisp
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+uci -q batch <<-EOF >/dev/null
+	delete ucitrack.@openwisp[-1]
+	add ucitrack openwisp
+	set ucitrack.@openwisp[-1].init=openwisp_config
+	commit ucitrack
+EOF
+
+exit 0
```
feckert, line 10: I'm not quite sure now, but I don't think we need this file any more, as we track the configuration file for changes with the following [line](https://github.com/openwisp/openwisp-config/blob/master/openwisp-config/files/openwisp.init#L74). In the init Script

applications/luci-app-openwisp/htdocs/luci-static/resources/view/openwisp.js
```diff
@@ -0,0 +1,101 @@
...
+        o.placeholder = '/usr/sbin/my_pre_reload_hook'
+        o.datatype = "file"
+
+        o = s.taboption('advanced', form.Value, 'post_reload_hook', _('Post-reload Hook'), _('Path to post reload hook script. The hook is not called if the path does not point to an executable script file. This hook is called each time openwisp-config applies a configuration, but after services are reloaded.'))
+        o.placeholder = '/usr/sbin/my_post_reload_hook'
+        o.datatype = "file"
+
+        return m.render();
+    }
+});
```
feckert, line 101: Please use `tabs` in this file

applications/luci-app-openwisp/htdocs/luci-static/resources/view/openwisp.js
```diff
@@ -0,0 +1,101 @@
...
+        o.rmempty = false
+
+        o = s.taboption('advanced', form.Flag, 'consistent_key', _('Consistent Key'), _('When using Automatic registration, this feature allows devices to keep the same configuration even if reset or reflashed.  This feature is enabled by default, but must be enabled also in the controller application in order to work.'))
+        o.rmempty = false
+
+        o = s.taboption('advanced', widgets.DeviceSelect, 'mac_interface', _('MAC Interface'), _('The interface to use for getting the MAC for this device.'))
+        o.rmempty = false
+
+        o = s.taboption('advanced', widgets.NetworkSelect, 'mac_interface', _('Management Interface'), _('The interface to use for management and control.'))
+        o.rmempty = false
```
feckert, line 53: mac_interface is also used in the config option above?

applications/luci-app-openwisp/htdocs/luci-static/resources/view/openwisp.js
```diff
@@ -0,0 +1,101 @@
...
+
+        o = s.taboption('advanced', form.Value, 'max_time', _('Max Time'), _('Value passed to curl --max-time argument, defaults to 30. Expressed in seconds.'))
+        o.placeholder = '30'
+        o.datatype = 'uinteger';
+
+        o = s.taboption('advanced', form.Value, 'capath', _('CA Path'), _('Path passed to curl --capath argument, by default is empty.'))
+        o.datatype = "file"
+
+        o = s.taboption('advanced', form.Value, 'capath', _('CA Cert'), _('Path passed to curl --cacert argument, by default is empty.'))
+        o.datatype = "file"
```
feckert, line 89: Do we need to mention in the `description` that we will use these options in Curl?
for example `Path passed to curl --cacert argument, by default is empty.`
That is not relevant isn`t it?

feckert (changes requested):
I miss the dependency on the openwisp service!
 I know there is a openwisp package in the package feed
https://github.com/openwrt/packages/tree/master/admin/openwisp-config
But I think its outdated!
I think we have to think about how to handle the agent in the package feed in the future.

nickberry17:
> I miss the dependency on the openwisp service! I know there is a openwisp package in the package feed https://github.com/openwrt/packages/tree/master/admin/openwisp-config But I think its outdated! I think we have to think about how to handle the agent in the package feed in the future.

Yes, I did not add OpenWISP as a dependency yet because there are several packages, one for each SSL library, and one without SSL.  I agree, it needs some work.

applications/luci-app-openwisp/htdocs/luci-static/resources/view/openwisp.js
nickberry17: You're right, probably an irrelevant detail.  I have changed it to what is in the curl documentation, since that is more to the point.

applications/luci-app-openwisp/htdocs/luci-static/resources/view/openwisp.js
nickberry17: My mistake, fixed.

feckert:
> > I miss the dependency on the openwisp service! I know there is a openwisp package in the package feed https://github.com/openwrt/packages/tree/master/admin/openwisp-config But I think its outdated! I think we have to think about how to handle the agent in the package feed in the future.
> 
> Yes, I did not add OpenWISP as a dependency yet because there are several packages, one for each SSL library, and one without SSL. I agree, it needs some thought.

I have always wondered why we need 4 variants. Can't we make it one?

nickberry17:
> > > I miss the dependency on the openwisp service! I know there is a openwisp package in the package feed https://github.com/openwrt/packages/tree/master/admin/openwisp-config But I think its outdated! I think we have to think about how to handle the agent in the package feed in the future.
> > 
> > 
> > Yes, I did not add OpenWISP as a dependency yet because there are several packages, one for each SSL library, and one without SSL. I agree, it needs some thought.
> 
> I have always wondered why we need 4 variants. Can't we make it one? Edit: We do not compile openwisp config. Openwisp config uses curl and uses mix of shell and lua. Curl, does have a default ssl backend. Currently we are using [wolfssl](https://github.com/openwrt/packages/blob/master/net/curl/Config.in#L7)

I've added openwisp-config-wolfssl as a dep

feckert:
The upstream changes, that we have now only one openwisp-config package, has been merged.
https://github.com/openwisp/openwisp-config/commit/08df34db29180a0d13c2e357723e8c546523014a

With the following pullrequest, the package in the package feed must now be adapted so that we can update and merge this LuCI integration.
https://github.com/openwrt/packages/pull/17282

badboyjoo:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

nickberry17:
Github messed up my repo when I merged online, so I'm moving this PR to a new branch
https://github.com/openwrt/luci/pull/5597

PR #5472, opened 2021-10-30
luci-app-unbound: fix error on empty leasetrigger

### LINKED ISSUE #5473: luci-app-unbound: Unbound interface errors out when using DHCP Link odhcpd
When using `odhcpd` by default `dhcp.odhcpd.leasetrigger` is not defined. While `luci-app-unbound` handles such situation, it crashes out in this case because it attempts to concatenate `nil` into error message string and the web UI throws following message on the screen:

```
/usr/lib/lua/luci/model/cbi/unbound/configure.lua:38: attempt to concatenate local 'lstrig' (a nil value)
stack traceback:
	/usr/lib/lua/luci/model/cbi/unbound/configure.lua:38: in function 'func'
	/usr/lib/lua/luci/cbi.lua:66: in function 'load'
	/usr/lib/lua/luci/dispatcher.lua:1353: in function '_cbi'
	/usr/lib/lua/luci/dispatcher.lua:1024: in function 'dispatch'
	/usr/lib/lua/luci/dispatcher.lua:985: in function 'dispatch'
	/usr/lib/lua/luci/dispatcher.lua:479: in function </usr/lib/lua/luci/dispatcher.lua:478>
```

This message is useful but it never gets displayed on the screen,
https://github.com/openwrt/luci/blob/0740023a3c44ebfb47126b8c9bb4e9a20985aa55/applications/luci-app-unbound/luasrc/model/cbi/unbound/configure.lua#L36

I have opened #5472 to fix this

avin3sh:
maintainer: @EricLuehrsen 
Description: When `odhcpd` `leasetrigger` is not defined - which is by default - the `Unbound` interface crashes because of attempting to concatenate nil value. This fixes it.

Closes #5473

### DIFF
--- a/applications/luci-app-unbound/luasrc/model/cbi/unbound/configure.lua
+++ b/applications/luci-app-unbound/luasrc/model/cbi/unbound/configure.lua
@@ -19,7 +19,7 @@ local ds = require "luci.dispatcher"
 local ucl = luci.model.uci.cursor()
 local valman = ucl:get_first("unbound", "unbound", "manual_conf")
 local dhcplk = ucl:get_first("unbound", "unbound", "dhcp_link")
-local lstrig = ucl:get_first("dhcp", "odhcpd", "leasetrigger")
+local lstrig = ucl:get_first("dhcp", "odhcpd", "leasetrigger") or "undefined"
 
 m1 = Map("unbound")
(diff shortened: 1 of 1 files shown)


jow-:
Merged, thank you!

PR #5478, opened 2021-10-31
luci-app-rosy-file-server: drop

BKPepe:
This package was marked as broken [1], so it's time to remove it.
As package rosy-file-server will be dropped soon [2].

Reasons:
1. it's broken (and package requires LuCI, which was marked as broken,
   thus it is useless) see https://github.com/openwrt/luci/pull/5315
2. maintainer is inactive
3. maintainer's company does not seem to work.

[1] https://github.com/openwrt/luci/commit/34b682afac310859f0d4696110d8a1af60f16c04
[2] https://github.com/openwrt/packages/pull/17031

Signed-off-by: Josef Schlehofer <pepe.schlehofer@gmail.com>


jow-:
Merged, thank you!

PR #5480, opened 2021-11-02
luci-proto-6in4: Add some HE.net pre-defined values for MTU.

systemcrash:


protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/6in4.js
```diff
@@ -87,7 +87,16 @@ return network.registerProtocol('6in4', {
 		o.datatype    = 'range(1,255)';
 
 		o = s.taboption('advanced', form.Value, 'mtu', _('Use MTU on tunnel interface'));
-		o.placeholder = '1280';
+		o.value('1480', _('1480 (HE default)'));
```
jow-, line 90: Can you change `'1480'` into an empty string here? This will cause `option mtu` to get removed from the configuration if it matches the default value. (Background: We generally try to avoid writing default values into the configuration)

systemcrash: It's my understanding that (based on https://openwrt.org/docs/guide-user/network/ipv6_ipv4_transitioning )
1) 6in4 assumes a default of `1280`
2) HT.net assumes a default of `1480`

So what you're saying above is that 6in4 assumes a default of 1480, and that, when the user 'sets' 1480, we simply assume the default and it sets nothing, correct? Assuming the link above is correct, it's actually 1280, and I should delete the 1280 value. 

Ja?

systemcrash: I suppose that, in any case, the real default is determined in the users control panel at HE when they log in:

henetcpl

jow-: Yeah, my intent was to match the effective default value as shown by `ip link show dev 6in4-xxx` if nothing is explicitly configured. Could be that it indeed is just 1280. Im not near any computer atm so I can't check myself.

systemcrash: I hit a snag: If I leave 1280 empty, it works as intended **the first time**. Then if I go back in again, it chooses 1480 as default, so that if I happen to click save, 1480 is now written. Result: value changes when it should not. Any ideas how to fix? Options [here](https://openwrt.github.io/luci/jsapi/LuCI.form.Value.html) aren't much help

systemcrash: ( One could re-order the list 1280 increasing, but this just hides the actual problem, so I consider it a workaround )

systemcrash: I think I'm turfing out other bugs here. I run on my local host:
`ping6 2001:470:20::2 -s 1400`
and run on openwrt
`tcpdump -i 6in4-HENET icmp6`

I see:

```
13:53:28.090983 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (0|1232) ICMP6, echo request, seq 203, length 1232
13:53:28.091076 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (1232|176)
13:53:28.092347 IP6 ordns.he.net > 2001:470:XX:YY:ZZ::3: ICMP6, echo reply, seq 203, length 1408
13:53:28.092600 IP6 tunnelabcxyz.tunnel.tserv24.abcxyz.ipv6.he.net > ordns.he.net: ICMP6, packet too big, mtu 1280, length 1240
13:53:29.096090 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (0|1232) ICMP6, echo request, seq 204, length 1232
13:53:29.096185 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (1232|176)
13:53:29.097395 IP6 ordns.he.net > 2001:470:XX:YY:ZZ::3: ICMP6, echo reply, seq 204, length 1408
13:53:29.097650 IP6 tunnelabcxyz.tunnel.tserv24.abcxyz.ipv6.he.net > ordns.he.net: ICMP6, packet too big, mtu 1280, length 1240
13:53:30.099653 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (0|1232) ICMP6, echo request, seq 205, length 1232
13:53:30.099746 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (1232|176)
13:53:30.101005 IP6 ordns.he.net > 2001:470:XX:YY:ZZ::3: ICMP6, echo reply, seq 205, length 1408
13:53:30.101256 IP6 tunnelabcxyz.tunnel.tserv24.abcxyz.ipv6.he.net > ordns.he.net: ICMP6, packet too big, mtu 1280, length 1240
13:53:31.104088 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (0|1232) ICMP6, echo request, seq 206, length 1232
13:53:31.104181 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (1232|176)
13:53:31.105415 IP6 ordns.he.net > 2001:470:XX:YY:ZZ::3: ICMP6, echo reply, seq 206, length 1408
13:53:31.105665 IP6 tunnelabcxyz.tunnel.tserv24.abcxyz.ipv6.he.net > ordns.he.net: ICMP6, packet too big, mtu 1280, length 1240
13:53:32.100471 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (0|1232) ICMP6, echo request, seq 207, length 1232
13:53:32.100566 IP6 2001:470:XX:YY:ZZ::3 > ordns.he.net: frag (1232|176)
13:53:32.101858 IP6 ordns.he.net > 2001:470:XX:YY:ZZ::3: ICMP6, echo reply, seq 207, length 1408
13:53:32.102109 IP6 tunnelabcxyz.tunnel.tserv24.abcxyz.ipv6.he.net > ordns.he.net: ICMP6, packet too big, mtu 1280, length 1240
```

This is with an MTU of 1480 set locally and in tunnelbroker.net. I'm not sure which layer responds here. It seems the max I can send is 1232... WTH. It seems as though the MTU setting is never applied somewhere...? Altho I have:

`6in4-HENET@NONE: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1480 qdisc noqueue state UNKNOWN qlen 1000`

`ip -6 route show` does not show any MTU or MSS set anywhere in the ipv6 or 4 route tables. 

Do you see similar behaviour?

jow-:
Looks like normal path MTU discovery at work? My understanding was that the local (and he.net configured remote) interface MTU is basically just an upper limit for packets we can emit/forward to this interface. ICMPv6 might reduce the actual in-transit MTU even further if any intermediate hop does not support our chosen MTU. IPv6 requires 1280 at least, so there might be routers in the internet that cannot handle more than that.

systemcrash:
> Looks like normal path MTU discovery at work? 

I'll wait for an answer from HE. Just want to make sure that everything works as it should.

PR #5488, opened 2021-11-05
trivial fix of wrong varible name

mash-graz:
This trivial patch will fix some issues affecting the display of collected olsrd statistics resp. graphs.


hnyman:
Merged with a corrected title as b0279f474e98

PR #5490, opened 2021-11-06
luci-app-statistics: protect nested formating litteral

mash-graz:
Signed-off-by: Martin Schitter <ms+git@mur.at>


jow-:
Merged, thank you!

PR #5496, opened 2021-11-07
luci-theme-bootstrap: fix bug in light theme

Slenkis:
if a dark theme was selected in the OS, and a light one was forced or automatic in LuCI, then some elements (checkboxes, scrolls) were from the dark theme

**Selected BootrstrapLight in LuCI and dark theme in OS:**
>before:
image
after:
image


Slenkis (commented):
sd

PR #5500, opened 2021-11-08
luci-mod-network: Disambiguate DNS config message

systemcrash:
luci-mod-network: Disambiguate DNS config message

Responses themselves are not RFC1918, but the address types that they
contain may be.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -299,7 +299,7 @@ return view.extend({
 
 		o = s.taboption('general', form.Flag, 'rebind_protection',
 			_('Rebind protection'),
-			_('Discard upstream RFC1918 responses.'));
+			_('Discard upstream responses containing <%s>RFC1918<%s> addresses.').format('a href="https://datatracker.ietf.org/doc/html/rfc1918" target="_blank" rel="noreferrer noopener"', '/a'));
 		o.rmempty = false;
 
 		o = s.taboption('general', form.Flag, 'rebind_localhost',
--- a/modules/luci-base/po/ar/base.po
+++ b/modules/luci-base/po/ar/base.po
@@ -1883,7 +1883,7 @@ msgid "Disassociate On Low Acknowledgement"
 msgstr "إلغاء الارتباط عند الإقرار القليل"
 
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js:302
-msgid "Discard upstream RFC1918 responses."
+msgid "Discard upstream responses containing <%s>RFC1918<%s> addresses."
 msgstr "تجاهل استجابات المنبع RFC1918"
 
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:198
--- a/modules/luci-base/po/bg/base.po
+++ b/modules/luci-base/po/bg/base.po
@@ -1855,7 +1855,7 @@ msgid "Disassociate On Low Acknowledgement"
 msgstr "Дисоцииране при ниско потвърждение"
 
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js:302
-msgid "Discard upstream RFC1918 responses."
+msgid "Discard upstream responses containing <%s>RFC1918<%s> addresses."
 msgstr ""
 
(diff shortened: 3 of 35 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -299,7 +299,7 @@ return view.extend({
 
 		o = s.taboption('general', form.Flag, 'rebind_protection',
 			_('Rebind protection'),
-			_('Discard upstream RFC1918 responses.'));
+			_('Discard upstream responses containing <a href="https://datatracker.ietf.org/doc/html/rfc1918" target="_blank" rel="noreferrer noopener">RFC1918</a> addresses.'));
```
feckert, line 302: Can't we take the whole html stuff thing out of translation and just write the following?

```suggestion
                        _('Discard upstream <%s>RFC1918<%s> responses.').format('a href="https://datatracker.ietf.org/doc/html/rfc1918" target="_blank" rel="noreferrer noopener"', '/a'));
```

Everything else is very error-prone from my point of view

systemcrash: Absolutely right. Great suggestion. Thanks for the formatting idea.

feckert:
In order for this to work, we must of insert <%s> into the already translated msgstr.

systemcrash:
> In order for this to work, we must insert <%s>RFC1918<%s> into the already translated msgstr.

Correct - but I cannot simply put those tags into translated strings, since the string translation has effectively changed, in order to disambiguate from:

> Discard upstream RFC1918 responses

and
> Discard upstream responses containing <%s>RFC1918<%s> addresses

Those should get done at the next translation pass. I advise this route, since the meanings now differ.

jow-:
Uhm, please revert the `<%s>` stuff. It is invalid HTML and will confuse translation software like poedit as well as human translators. It is neither clear that `<%s>...<%s>` is supposed to interpolate into both a start *and* end tag and `<%s>` might be mistaken as a quoted form of `%s` like `«%s»` or `〈%s〉`

hnyman:
@jow- @feckert 
I think that we should discourage extensive abbreviation markings in general.  They often add more complexity than provide clarity for the end-user.

feckert:
@jow- @hnyman Sorry for the inconvenience.
I will take better care next time so that this doesn't happen to me again. 
I found the original suggestion of adding the full url to the translation even worse.

hnyman:
The final mess this time is probably because I merged the open Weblate PR while jow was already editing the fix for this PR's changes.  

I have noticed that weblate is picky about simultaneous changes. 
(I have learned to make sure that, for example, the open weblate PRs are merged before any i18n sync is performed. Weblate doesn't like if strings included in its own commits/PRs are changed in the main repo.)

feckert:
@hnyman You're right about that. I always merge the weblat pullrequest beforehand and then I resync i18n.

I meant before that next time I will pay more attention to the html in the translation.
I didn't realise that the i18n tools would get confused.

hnyman:
That tool confusion was actually a new bit for us all, I think. Just highlights that the translation string is partially a programming language like thing, not just an arbitrary string.

(But in general, too many abbreviations with links can add a lot of complexity.)

jow-:
@feckert @hnyman - thanks for dealing with this. Sorry if my remark sounded rude, I didn't mean to criticize the effort - I do agree with Hannu though that we should try to minimize the amount of markup in translations, it proved to be a problem various times.

systemcrash:
For future reference, what do we (hitherto) think would be a good way forward? Avoiding HTML (tags) is likely best, but sometimes it's unavoidable. This instance is not harmed by lacking an explanatory link, though.

jow-:
I was toying with the idea of having some kind of auto-glossary which applies either `<abbr title="Long description">...</abbr>` or `<a href="https://even/longer/referfence">...</a>` markup to known keywords in translation strings in a context-dependent manner.

The idea is then that we maintain a central glossary `.js` or `.json` file where we maintain keyword => description and keyword => url mappings. The file will be translatable like any other source file. Then `form.js` or `ui.js` could apply those auto-glossary markup to places where it makes sense, e.g. in option captions, descriptions, tab descriptions, section titles and descriptions and map descriptions.

systemcrash:
Having everything in one place will make management a bit easier, I think.

PR #5505, opened 2021-11-09
luci-theme-bootstrap: fix scrollbar in modal window

Slenkis:
### Before:
owerflow_scroll

### After:
owerflow_auto


Slenkis:
@jow- qq

PR #5511, opened 2021-11-12
luci-proto-wireguard: add option to disable peer-section

rwalli:
Add luci option to [c4e9940](https://github.com/openwrt/openwrt/commit/c4e994011f56d30e031705c16c6b5c498c530852) which allows disabling a wireguard peer

Signed-off-by: Robert Walli <12079858+rwalli@users.noreply.github.com>

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -151,6 +151,9 @@ return network.registerProtocol('wireguard', {
 			]);
 		};
 
+		o = ss.option(form.Flag, 'disabled', _('Peer disabled'), _('Enable / Disable peer. Restart wireguard interface to apply changes.'));
+		o.optional = true;
+
 		o = ss.option(form.Value, 'description', _('Description'), _('Optional. Description of peer.'));
 		o.placeholder = 'My Peer';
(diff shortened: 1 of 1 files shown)


feckert:
Please use a valid Signed-off-by email address.

systemcrash:
> Please use a valid Signed-off-by email address.

Just curious - but what's the mandate to use a valid SOB email? The address presented *is* a valid address - it just hides the user's real one.

feckert:
> > Please use a valid Signed-off-by email address.
> 
> Just curious - but what's the mandate to use a valid SOB email? The address presented _is_ a valid address - it just hides the user's real one.

I am not a github expert. I had to look it up first. That's why it wasn't clear to me. I find the mail address strange. But if that is the case, then it fits for me.

jow-:
The S-o-b requirements come from https://openwrt.org/submitting-patches which we generally try to adhere to. For trivial changes I usually don't mind much but for extensive original work I prefer proper attribution and a working contact mail address. If the submitter does not want to / can't provide either I usually add a Submitted-by or Suggested-by annotation and push the commit with me ad author and my own S-o-b.

This requires extra effort though and means that such PRs often stall since they cannot be merged with the big green button.

feckert:
Merged with https://github.com/openwrt/luci/commit/aaa2b3dd2b7124d7d5f23ed31f670e89dcde81c6
Thanks

PR #5512, opened 2021-11-13
ci: Only run github pages on the official repository

thg2k:
I noticed this undesired actions running on the forked repository.

As I understand this, generating the 'gh-pages' branch is only meaningful in the official repository. I think this should fix it.


jow-:
@aparcar ?

aparcar:
Please rename the commit to `CI: only run GitHub pages on the official repository`

PR #5520, opened 2021-11-16
luci-lib-docker: fix chunked decode

acooler15:
cherry-pick form [lisaac/luci-lib-docker/commit/98a6634](https://github.com/lisaac/luci-lib-docker/commit/98a663449208b65e68702ad0f3ad61c536bfa40c)

@feckert
Signed-off-by: Ren Zongjia <acooler15@foxmail.com>


feckert:
Can you please write an explanation? Why we need this now. Unfortunately, there is nothing in the upstream commit.

PR #5525, opened 2021-11-17
Add support for SK-ECDSA-SHA2 keys to be uploaded in web interface

cortices:
Allow adding sk-ecdsa-sha2 keys, which are regular ECDSA-SHA2 keys with authentication by FIDO/U2F token.
Mine start with the following identifier: `sk-ecdsa-sha2-nistp256@openssh.com`.


cortices:
Whoops, never mind, I see that dropbear doesn't have elliptic curve support at all on openwrt, this wasn't the problem.

PR #5550, opened 2021-11-25
https-dns-proxy support option use_http1

robinZhao:
https-dns-proxy support option use_http1

Signed-off-by: Robin <zhaoruibin@aliyun.com>


stangri:
Duplicate of #5553

Issue #5565, opened 2021-12-01
IPv6: RA MTU and RA Hop Limit always show default values


dorsiflexion:
### Steps to reproduce
- go to: _Network_ → _Interfaces_ → _LAN_ (an interface with a static address) → _DHCP server_ tab → _IPv6 RA settings_ tab
- set _RA MTU_ and _RA Hop Limit_ to arbitrary non-default value
- Save & Apply
- go to _IPv6 RA settings_ tab again

### Actual behavior
- `option ra_mtu '1500'` is set correctly in `/etc/config/dhcp`
- In LuCI though, _RA MTU_ and _RA Hop Limit_ are still displaying the default values in light gray font color.

### Expected behavior
- _RA MTU_ and _RA Hop Limit_ are displaying the correct values found in `/etc/config/dhcp` in black font color.

### Additional Information
- OpenWrt 21.02.1 r16325-88151b8303 / LuCI openwrt-21.02 branch git-21.295.67054-13df80d
- TP-Link TL-WR1043ND v2

dorsiflexion:
I upgraded to 21.02.2 and it is working perfectly.
Thanks a lot!

Issue #5564, opened 2021-12-01
IPv6: Advertised DNS server not configurable for RA only


dorsiflexion:
### Steps to reproduce
- go to: _Network_ → _Interfaces_ → _LAN_ (an interface with a static address) → _DHCP server_ tab → _IPv6 settings_ tab
- set _RA-Service_ to _server mode_ or _hybrid mode_
- set _DHCPv6-Service_ to _server mode_ or _hybrid mode_
- _Announced IPv6 DNS servers_ is now configurable. This will set e. g.
```
config dhcp 'lan'
	list dns 'fd00::abcd'
```
in `/etc/config/dhcp`. This _odhcpd_ setting determines which DNS server is advertised via DHCP **and** RA.
- set _DHCPv6-Service_ to _disabled_, save & apply

### Actual behavior
- _Announced IPv6 DNS servers_ setting is deleted, i. e. `list dns 'fd00::abcd'` in `/etc/config/dhcp` is deleted.
- DNS server advertised via RA cannot be set anywhere, none is advertised.

### Expected behavior
- _Announced IPv6 DNS servers_ setting is configurable and kept when any of _RA-Service_ or _DHCPv6-Service_ is in _server mode_ or _hybrid mode_.

### Additional Information
- OpenWrt 21.02.1 r16325-88151b8303 / LuCI openwrt-21.02 branch git-21.295.67054-13df80d
- TP-Link TL-WR1043ND v2

dorsiflexion:
I upgraded to 21.02.2 and it is working perfectly.
Thanks a lot!

PR #5567, opened 2021-12-02
[RFC] github: add issue templates file

feckert:
@jow-  I like the issue templates of #5565 and #5564.
I would nice if we had something like that.

### DIFF
--- a/.github/issue_template
+++ b/.github/issue_template
@@ -1,0 +1,40 @@
+# Issue template:
+1. Please make sure that the issue subject starts with `<package-name>:`
+2. Issues related to end of life (EOL) releases are not supported or maintained and will be closed
+3. Remove lines from top till here and fill in the following questions
+***
+
+## Steps to reproduce:
+**Example: Replace the following lines and remove this**
+
+1. go to: Network → Interfaces → LAN (an interface with a static address) → DHCP server tab → IPv6 RA settings tab
+2. set RA MTU and RA Hop Limit to arbitrary non-default value
+3. Save & Apply
+4. go to: IPv6 RA settings tab again
+
+## Actual behavior:
+**Example: Replace the following lines and remove this**
+
+1. option ra_mtu '1500' is set correctly in /etc/config/dhcp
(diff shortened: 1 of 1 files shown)


.github/issue_template
```diff
@@ -0,0 +1,40 @@
...
+
+## Expected behavior:
+**Example: Replace the following lines and remove this**
+
+```
+RA MTU and RA Hop Limit are displaying the correct values found in /etc/config/dhcp in black font color.
+```
+
+## Additional Information:
+Openwrt version information from system `/etc/openwrt_release`
```
BKPepe, line 29:
```suggestion
OpenWrt version information from system `/etc/openwrt_release`
```

.github/issue_template
```diff
@@ -0,0 +1,40 @@
+# Issue template:
+1. Please make sure that the issue subject starts with `<package-name>:`
+2. Issues related to releases **below 18.06** and forks are not supported or maintained and will be closed
+3. Remove lines from top till here and fill in the follwing questions
```
BKPepe, line 4:
```suggestion
3. Remove lines from top till here and fill in the following questions
```

.github/issue_template
```diff
@@ -0,0 +1,40 @@
+# Issue template:
+1. Please make sure that the issue subject starts with `<package-name>:`
+2. Issues related to releases **below 18.06** and forks are not supported or maintained and will be closed
```
BKPepe, line 3: Nitpick: Well.. there was recent RFC in openwrt-devel that we should rather use [downstream distributions](https://lists.openwrt.org/pipermail/openwrt-devel/2021-December/037338.html). So, I'd say it is a question about forks.  There are forks, which contributes back to upstream or they don't and playing on their playground.

feckert: Thanks for the review, I have adjusted the text.
@BKPepe I have now chosen EOL so that we don't have to adjust the text every time a new release comes out.
That is usually forgotten.

I have also removed the `forks` note, then we do not need to refer to it.

PR #5571, opened 2021-12-02
luci-mod-network: Expose 802.11k settings.

systemcrash:
Tested on: 21.02.1

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1151,6 +1151,24 @@ return view.extend({
 					o = ss.taboption('advanced', form.Flag, 'multicast_to_unicast', _('Multi To Unicast'), _('ARP, IPv4 and IPv6 (even 802.1Q) with multicast destination MACs are unicast to the STA MAC address. Note: This is not Directed Multicast Service (DMS) in 802.11v. Note: might break receiver STA multicast expectations.'));
 					o.rmempty = true;
 
+					/* 802.11k settings start */
+					// Probe 802.11k support via EAP support (full hostapd has EAP)
+					if (L.hasSystemFeature('hostapd', 'eap')) {
+						o = ss.taboption('advanced', form.Flag, 'ieee80211k', _('802.11k RRM'), _('Radio Resource Measurement - Sends beacons to assist roaming. Not all clients support this.'));
+						// add_dependency_permutations(o, { mode: ['ap', 'ap-wds'], encryption: ['psk', 'psk2', 'psk-mixed', 'sae', 'sae-mixed'] });
+						o.depends('mode', 'ap');
+						o.depends('mode', 'ap-wds');
+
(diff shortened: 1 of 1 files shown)


systemcrash:
Could this PR get some love, @jow- @hnyman

Feerip:
@hnyman @jow- Any chance we can get this PR merged? It would be very helpful :)

hnyman:
Does this actually work as intended?

I applied the patch to my system, and I think that it has got the default value wrong (0) for the two sub-options.
The default in the OpenWrt hostapd init script is actually 1 if 80211k is enabled.

See
https://github.com/openwrt/openwrt/blob/master/package/network/services/hostapd/files/hostapd.sh#L859-L867
```
	json_get_vars ieee80211k rrm_neighbor_report rrm_beacon_report
	set_default ieee80211k 0
	if [ "$ieee80211k" -eq "1" ]; then
		set_default rrm_neighbor_report 1
		set_default rrm_beacon_report 1
	else
		set_default rrm_neighbor_report 0
		set_default rrm_beacon_report 0
	fi
```

I enabled 80211k and then checked that both "enabled" and "disabled" for the two rrm suboptions actually enabled the two rrm suboptions.  
* "enabled" --> there is a config item 1 that is harmless, as the underlying default is enabled in any case
```
root@router1:/www# grep rrm /etc/config/wireless
        option rrm_neighbor_report '1'
        option rrm_beacon_report '1'
root@router1:/www# grep rrm /var/run/hostapd-phy0.conf
rrm_neighbor_report=1
rrm_beacon_report=1
```
  
* "disabled" --> causes LuCI to remove the option from etc/config/wireless as LuCI thinks that is now reverted to default disabled and should be removed as unnecessary.
```
root@router1:/www# grep rrm /etc/config/wireless
root@router1:/www# grep rrm /var/run/hostapd-phy0.conf
rrm_neighbor_report=1
rrm_beacon_report=1
```

I think that both suboptions should have statements defining default as "enabled".

McGiverGim:
@systemcrash I know this is a little forgotten, but I hope this will be merged sometime... I think this needs a rebase, and verify what @hnyman says. Thanks!

systemcrash:
> Does this actually work as intended?
> ...
> I think that both suboptions should have statements defining default as "enabled".

That they can be configured separately, is as intended. hostapd might do well to allow granularity for the settings. 

802.11k is a standard, and for a device to be operating 'compliantly', both settings should be on. But there may exist scenarios when disabling one is desirable. 

But neither of the two suboptions are active without the main k option. 

Unfortunately, I do not remember the design decisions that were based on testing, as I did this a year ago. I added the default on options as suggested.

hnyman:
Thanks for fixing the defaults to be in line with OpenWrt behavior.

hnyman:
@systemcrash  @jow- 

There are now quite a lot of WLAN roaming related settings. Mainly in the Advanced tab, but 802.11r related stuff in "Encryption".

I wonder if we should define a new tab "WLAN roaming" and move these 802.11r, k and v settings there.

systemcrash:
I'm okay with that.

As long as some organization system is logical for people looking for the
settings and an expert alike.

Apple support lumps k r and v together. Their implementations are good for
client devices.

PR #5576, opened 2021-12-03
[21.02] luci-base: send failed/successfull login to syslog

BKPepe:
Cherry-picked from https://github.com/openwrt/luci/pull/4925


feckert:
Works as expected:
Before change
`daemon.err uhttpd[7662]: luci: failed login on / for root from 192.168.0.2`
After change:
`user.info : luci: failed login on / for root from 192.168.0.2`

PR #5611, opened 2021-12-19
luci-mod-network: add dnsmasq log-facility option

clayface:
Linked OpenWrt PR: [#4864](https://github.com/openwrt/openwrt/pull/4864)
Add dnsmasq's log-facility option to luci-mod-network, which already
exists as a uci option. This allows the user to specify a custom file
location for queries to be written to, rather than the default syslog
when enabled via the log-queries option.

Signed-off-by: Matthew Hagan <mnhagan88@gmail.com>


dave14305:
Would the log-facility path need to be added to the dnsmasq jail?

systemcrash:
Closed by 6666d32e1d871cf28d693d71f0e13922f7ab8ae5

PR #5613, opened 2021-12-22
luci-app-statistics: ping module, Update MaxMissed data type

tew42:
MaxMissed was implemented with positive integers only, whereas I think it should also allow (and default to) the option of "-1" (disabled state) as per collectd documentation

### DIFF
--- a/applications/luci-app-ddns/po/fi/ddns.po
+++ b/applications/luci-app-ddns/po/fi/ddns.po
@@ -1,23 +1,23 @@
 msgid ""
 msgstr ""
-"PO-Revision-Date: 2021-06-18 19:32+0000\n"
-"Last-Translator: Demian Wright <wright.demian+weblate@gmail.com>\n"
+"PO-Revision-Date: 2021-12-21 16:49+0000\n"
+"Last-Translator: Hannu Nyman <hannu.nyman@iki.fi>\n"
 "Language-Team: Finnish <https://hosted.weblate.org/projects/openwrt/"
 "luciapplicationsddns/fi/>\n"
 "Language: fi\n"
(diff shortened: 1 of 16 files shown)


jow-:
Just `min(-1)` should be enough, it will imply integer.

hnyman:
Please also change the commit title to start with "luci-app-statistics: ping module" like you did for the PR.

And please remember to sign off your commit (in the commit message).

tew42:
> Just `min(-1)` should be enough, it will imply integer.

Just tested, unfortunately that doesn't appear to be the case - just `min(-1)` gave a text field where I could enter 0.5 or -0.5



> Please also change the commit title to start with "luci-app-statistics: ping module" like you did for the PR.
> And please remember to sign off your commit (in the commit message).

Looks like this isn't possible in the web interface? Guess I'll have to install a client, will try to fix in the next few hours.

tew42:
> Please also change the commit title to start with "luci-app-statistics: ping module" like you did for the PR.
> 
> And please remember to sign off your commit (in the commit message).

I think this should be good now?

hnyman:
I looked at the collectd source code, this proposed change seems a bit unnecessary, as the default value in the upstream sources is -1 (=disabled), and the option field is optional in LuCI.  The current '10' is just a placeholder text, not an option value. So, at the default situation there is no value in LuCI options, and the effect is the same as with a real value -1.

https://github.com/collectd/collectd/blob/collectd-5.13/src/ping.c#L83

I guess that merging this (without the unrelated second commit) is still ok, but the net impact will be small. This mainly makes it possible to explicitly disable it after trying some value there.

hnyman:
Merged with slightly edited text as 8572ae032f3

PR #5615, opened 2021-12-24
luci-app-frpc: Set the disable_log_color option to false by default

acooler15:
In [fatedier/frp](https://github.com/fatedier/frp/blob/0fb6aeef583c59e2d97fa0cf8e02e39868a0fae7/conf/frpc_full.ini#L24), the default value of  disable_log_color is false.


ysc3839:
LGTM, thanks!

PR #5616, opened 2021-12-24
sysauth: add autocomplete and id attributes to login inputs

Tsvetkovv:
this let browser automatically fill according to HTML spec for input elements.
luci-theme-bootstrap and luci-base are affected.

image


jow-:
Merged via 3e9d9a9dbb045c24eb93643838f8e8e3b9074e4b, thanks!

PR #5627, opened 2022-01-03
[21.02] luci-app-https-dns-proxy: massive providers update

stangri:
* introduce 31 new providers
* uniform names (FQDNs) in provider files
* other minor fixes

(cherry picked from commit fae45215aa3a42b944270e0e75bfa1ba23e5702f)

### DIFF
--- a/applications/luci-app-https-dns-proxy/Makefile
+++ b/applications/luci-app-https-dns-proxy/Makefile
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_LICENSE:=GPL-3.0-or-later
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>
-PKG_VERSION:=2021-09-27-2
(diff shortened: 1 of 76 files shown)


systemcrash:
Hi Stan. Given that this list is surely likely to grow as time goes on (and manually providing these might become untenable), how about an extra 'custom' field where users can provide their own at config time?

Possibly an idea for future iteration.

stangri:
> Hi Stan. Given that this list is surely likely to grow as time goes on (and manually providing these might become untenable), how about an extra 'custom' field where users can provide their own at config time?
> 
> Possibly an idea for future iteration.

Paul, this is a frequently requested feature and it's totally planned for after transition to javascript. I really don't want to implement it in Lua now since the transition is inevitable (although not immenent).

I'd welcome a PR for either (transition or a custom provider implementation in lua).

systemcrash:
I hear you on the lua thing...

PR #5633, opened 2022-01-07
luci-mod-status: display full HE rate information

dangowrt:
Associated Stations on the status page should also show full HE rate
information which was added by
commit f35e877dc6 ("luci-mod-network: add 802.11ax HE support")

However, it was only added to the Wireless page and not to the summary
on the Status page. Add it there as well.

(It would be nice to move this function to a common include instead of duplicating it, I suppose. I'm not so much into LuCI, so I'd need some advise how that should be done, if at all)


jow-:
Merged, thanks!

blocktrron:
Saw it just now - looks good!

PR #5634, opened 2022-01-07
luci-app-nft-qos: move limit_mac_enabled

qiuzi:
fix luci error

### DIFF
--- a/applications/luci-app-nft-qos/luasrc/model/cbi/nft-qos/nft-qos.lua
+++ b/applications/luci-app-nft-qos/luasrc/model/cbi/nft-qos/nft-qos.lua
@@ -100,6 +100,13 @@ o.datatype = "ipaddr"
 o:depends("limit_enable","1")
 
 --
+-- limit speed by mac address
+--
+o = s:taboption("limitmac", Flag, "limit_mac_enable", translate("Limit Enable"), translate("Enable Limit Rate Feature"))
+o.default = limit_mac_enable or o.enabled
+o.rmempty = false
+
+--
 -- Priority
 --
 o = s:taboption("priority", Flag, "priority_enable", translate("Enable Traffic Priority"), translate("Enable this feature"))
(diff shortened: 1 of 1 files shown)


feckert:
Can you please describe what is actually the problem here? I can't do much with `update`.

qiuzi:
@feckert 
using such a configuration results in an error
 `config default 'default'
	option limit_enable '1'
	option limit_type 'dynamic'
	option dynamic_bw_down '100'
	option dynamic_bw_up '100'
	option limit_mac_enable '1'
	option priority_enable '1'
	option priority_netdev 'lan'

`
image

feckert:
Interesting, I don't understand the problem. Why it should work after this change. It should also work before the change.

jow-:
> Interesting, I don't understand the problem. Why it should work after this change. It should also work before the change.

It is because of this code:
https://github.com/openwrt/luci/blob/ff8e6a3772a01cd87059c6bedf77d7ca40d9bed2/applications/luci-app-nft-qos/luasrc/model/cbi/nft-qos/nft-qos.lua#L197-L199

The `s` (section) variable is overwritten with another CBI section definition only if `option limit_enable '1'` is set, afterwards the `o = s:taboption("limitmac", Flag, "limit_mac_enable", translate("Limit Enable"), translate("Enable Limit Rate Feature"))` declaration fails because `s` now refers to another CBI section, not the one with the declared tabs. Moving the option declaration further up indeed fixes this.

feckert:
@jow- thanks for your explanation

@qiuzi Please update your commit message with `luci-app-nft-qos: fix lua trace by moving limit_mac_enabled option`.
Also add a proper Signed-off-by. Have a look at the other commits.

qiuzi:
@feckert Submit it, I'm familiar with these

qiuzi (commented):
Move limit_mac_enabled to fix location

feckert:
Please read my comments in my previous post!

- Update your commit message with `luci-app-nft-qos: fix lua trace by moving limit_mac_enabled option`.
- Also add a proper Signed-off-by. Have a look at the other commits.
- Also squash your commits to one, so we have only one change commit.

qiuzi:
@feckert I use the web version of github to submit and can't use Signed-off-by

PR #5642, opened 2022-01-11
luci-proto-xfrm: add luci xfrm support

hgl:
I'm not really sure about MTU range, should it always match the underlying interface's MTU?

### DIFF
--- a/protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
+++ b/protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
@@ -1,0 +1,51 @@
+'use strict';
+'require uci';
+'require form';
+'require network';
+'require tools.widgets as widgets';
+
+return network.registerProtocol('xfrm', {
+	getI18n: function() {
+		return _('IPsec XFRM');
+	},
+
+	getIfname: function() {
+		return this._ubus('l3_device') || this.sid;
+	},
+
+	getOpkgPackage: function() {
+		return 'xfrm';
+	},
+
+	isFloating: function() {
+		return true;
+	},
+
+	isVirtual: function() {
+		return true;
+	},
+
+	getDevice: function() {
+		return null;
+	},
+
+	containsDevice: function(ifname) {
+		return (network.getIfnameOf(ifname) == this.getIfname());
+	},
+
+	renderFormOptions: function(s) {
+		var o, ss;
+
+		o = s.taboption('general', form.Value, 'ifid', _('Interface ID'), _('Required. XFRM interface ID to be used for SA.'));
+		o.datatype = 'integer';
+
+		o = s.taboption('general', widgets.NetworkSelect, 'tunlink', _('Required. Underlying interface.'));
+		o.exclude = s.section;
+		o.nocreate = true;
+
+		o = s.taboption('general', form.Value, 'mtu', _('MTU'), _('Optional. Maximum Transmission Unit of the XFRM interface.'));
+		o.datatype = 'range(68,65535)';
+		o.placeholder = '1280';
+		o.optional = true;
+	}
+});
--- a/modules/luci-base/po/ar/base.po
+++ b/modules/luci-base/po/ar/base.po
@@ -11,7 +11,7 @@ msgstr ""
 "&& n%100<=10 ? 3 : n%100>=11 ? 4 : 5;\n"
 "X-Generator: Weblate 4.10-dev\n"
 
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1513
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1516
 msgid "%.1f dB"
 msgstr "%.1f ديسيبل"
 
@@ -229,19 +229,19 @@ msgid ""
 "the router"
 msgstr ""
 
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1493
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1496
 msgid "A43C + J43 + A43"
 msgstr "A43C + J43 + A43"
 
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1494
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1497
 msgid "A43C + J43 + A43 + V43"
 msgstr "A43C + J43 + A43 + V43"
 
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1506
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1509
 msgid "ADSL"
 msgstr "ADSL"
 
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1482
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1485
 msgid "ANSI T1.413"
 msgstr "ANSI T1.413"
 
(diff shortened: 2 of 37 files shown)


protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
```diff
@@ -0,0 +1,51 @@
...
+		return 'xfrm';
+	},
+
+	isFloating: function() {
+		return true;
+	},
+
+	isVirtual: function() {
+		return true;
+    },
```
feckert, line 26: wrong indentation

protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
```diff
@@ -0,0 +1,51 @@
...
+
+    getDevice: function() {
+        return null;
+    },
```
feckert, line 30: wrong indentation

protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
```diff
@@ -0,0 +1,51 @@
...
+        o.datatype = 'integer';
+
+        o = s.taboption('general', widgets.NetworkSelect, 'tunlink', _('Required. Underlying interface.'));
+        o.exclude = s.section;
+        o.nocreate = true;
+
+        o = s.taboption('general', form.Value, 'mtu', _('MTU'), _('Required. Maximum Transmission Unit of the XFRM interface.'));
+		o.datatype = 'range(1280,1500)';
+		o.placeholder = '1492';
+		o.optional = true;
```
feckert, line 49: wrong and missing indentation for function renderFormOptions

protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+
+	containsDevice: function(ifname) {
+		return (network.getIfnameOf(ifname) == this.getIfname());
+	},
+
+	renderFormOptions: function(s) {
+        var o, ss;
+
+        o = s.taboption('general', form.Value, 'ifid', _('Interface ID'), _('Required.'));
```
feckert, line 39: Can you formulate the detail help text a little more?

feckert (changes requested):

hgl:
Done.

protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
```diff
@@ -0,0 +1,51 @@
...
+		o.datatype = 'integer';
+
+		o = s.taboption('general', widgets.NetworkSelect, 'tunlink', _('Required. Underlying interface.'));
+		o.exclude = s.section;
+		o.nocreate = true;
+
+		o = s.taboption('general', form.Value, 'mtu', _('MTU'), _('Required. Maximum Transmission Unit of the XFRM interface.'));
+		o.datatype = 'range(1280,1500)';
+		o.placeholder = '1492';
+		o.optional = true;
```
feckert, line 49: Why is this required in your help message, if do you set this optional?

protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+'require uci';
+'require form';
+'require network';
+'require tools.widgets as widgets';
+
+return network.registerProtocol('xfrm', {
+	getI18n: function() {
+		return _('XFRM');
```
feckert, line 9: I think this text is not very meaningful if you do not know exactly what XFRM means.
Maybe` IPsec Tunnel Interface (XFRM)`?
https://github.com/openwrt/openwrt/blob/ec6293febc244d187e71a6e54f44920be679cde4/package/network/config/xfrm/Makefile#L19

jow-: >  Maybe IPsec Tunnel Interface (XFRM)?

Keep in mind that the description should not be too long to fit into the various UI elements. Maybe `IPsec XFRM` provides enough context. After all, people installing XFRM support and adding XFRM interfaces likely now what it is about.

protocols/luci-proto-xfrm/htdocs/luci-static/resources/protocol/xfrm.js
hgl: Good catch, it was unintentional. From the xfrm package, it seems it should be optional.

> Is the range also correct here?

As I mentioned in the PR, I really have no idea what the range should be, would appreciate some guidance.

hgl:
The xfrm proto [supports a multicast flag](https://github.com/openwrt/openwrt/blob/7f7705923a8299323057e78ccae4473b5bb05807/package/network/config/xfrm/files/xfrm.sh#L39). However, I can't find its support in either [netifd's tunnel_attrs or xfrm_data_attrs](https://git.openwrt.org/?p=project/netifd.git;a=blob;f=system.c;h=32597c14e29145a5187d0642769f6eb5e9458aa7;hb=5ca5e0b4d058a47d72ba4102acdcec826e203c41#l18), that flag seems to be supported only by a device.

I wonder if this flag should be added to this luci package?

hgl:
@feckert @jow- Other than the multicast flag I'm not sure about, all other issues should be addressed.

feckert:
I'm not sure about the mtu range either. But we can start with it. If that doesn't fit, then hopefully this one will get in touch. So far it looks good.
TODO:
* Can you please combine the commits into one with the commit message `luci-proto-xfrm: add xfrm proto support`
* Can you also please update the base translation?

jow-:
> I'm not sure about the mtu range either.

hgl:
@jow- The page you [previously referred me to](https://elixir.bootlin.com/linux/v5.0.21/source/net/xfrm/xfrm_interface.c#L626) says the max value is 1500, I wonder which one is correct?

@feckert sure. I'm new to LuCI's i18n system, still trying out the generating commands from wiki.

I still need help for the multicast flag (not be confused with mtu).

jow-:
@hgl - seems the limit was raised in a later version https://github.com/torvalds/linux/commit/f042365dbffea98fb8148c98c700402e8d099f02 - afair Kenrel versions 5.5 onwards allow up to 65535 bytes.

hgl:
@jow- I see, thanks for locating the commit, and I presume LuCI's master branch is for OpenWrt with Kernel 5.5+ as well?

jow-:
Yeah, let's raise the limit to 65535. As for the multicast option; it is a generic option supported by all device types. Netifd will allow these options: https://lxr.openwrt.org/source/netifd/device.c#L32 in all `config interface` sections, regardless of the chosen proto. Additionally, these options can be set in a separate `config device` section which is the preferred way by LuCI. 

So omit the multicast option from the proto handler, if users wish to override it, they can set it via Network -> Interfaces -> Devices in LuCI.

hgl:
Makes sense.

Sorry, just one more final question: [xfrm.sh uses 1280](https://github.com/openwrt/openwrt/blob/7f7705923a8299323057e78ccae4473b5bb05807/package/network/config/xfrm/files/xfrm.sh#L35) as the default value, but [Kernel uses 1500](https://lxr.missinglinkelectronics.com/linux/net/xfrm/xfrm_interface.c#L561). I currently use 1500 in this package, but maybe I should match it with xfrm.sh?

hgl:
I changed the default to 1280, and i18n is updated too. Ready for review.

hgl:
Any maintainer is free to review? The base i18n change has a high probability of resulting in merge conflicts if it's not merged soone.

feckert:
Thank you for your contribution.
Let´s try it :+1:

Issue #4970, opened 2021-04-09
force upgrade markup (sysupgrade/luci-21.02 - luci-theme-bootstrap)


iuncuim:
The option "Force upgrade" is outside the container (Bootstrap theme). Tested Firefox/Chrome browsers with reset caches. 
![Снимок экрана от 2021-04-09 12-05-14](https://user-images.githubusercontent.com/15281352/114157603-79712100-992c-11eb-8340-b8620a685668.png)
LuCI openwrt-21.02 branch (git-21.088.35541-d627abe) / OpenWrt 21.02-SNAPSHOT r15952+1-6090337

iuncuim:
@dibdot the reason is this [commit](https://github.com/openwrt/luci/commit/6ffd1592ce574afe5493d8d7d11f65bec1916e86)

dibdot:
I doubt that, rather I wonder why this label was defined as a button!? E.g.
```
[...]
body.push(E('p', {}, E('label', { 'class': 'btn alert-message danger' }, [
[...]
```
remove the "btn" part should fix this IMHO.

iuncuim:
> remove the "btn" part should fix this IMHO.
Don't know, maybe this will break other themes.

I made this new rule for bootstrap-theme, now the alert looks ok:
`
.modal .btn.danger {
	white-space: normal;
}
`

iuncuim:
@jow- now other buttons are affected by the problem:
![изображение](https://user-images.githubusercontent.com/15281352/135233233-283512f1-9e75-46fd-9a75-8c6bdd555a79.png)
With new css rule on cascade.css it looks like:
`.modal .btn.danger,
.modal .btn {
	white-space: normal;
}`
![изображение](https://user-images.githubusercontent.com/15281352/135233688-3741fd78-11ed-4df6-b544-4761270a0858.png)
Probably worth reopening the issue.

iuncuim:
Patch missed on the 21.02 branch.
https://github.com/openwrt/luci/blob/b39d9bf4bb88acd3120098c3a087e47331d1d757/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js#L280

PR #5650, opened 2022-01-19
luci-proto-ipip: add nohostroute configurable

clayface:
Add nohostroute configurable to luci-proto-ipip.

Also explicitly set this option, and the previous df configurable, to optional.

linked pr (merged) [openwrt #4970](https://github.com/openwrt/openwrt/pull/4970)


feckert:
Thanks

PR #5652, opened 2022-01-21
luci-mod-status: add nftables status page

jow-:
The PR introduces an nftables status page similar to the existing one for iptables.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js
@@ -1,0 +1,663 @@
+'use strict';
+'require view';
+'require poll';
+'require fs';
+'require ui';
+'require dom';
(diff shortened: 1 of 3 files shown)


hnyman:
Thanks Jow,
Seems to work ok with my simple (pretty default) firewall.

hnyman:
@jow- 
I noticed a new branch.
Did you accidentally push an update of this PR's commit to a branch at the main site, instead of your own fork?

bittin:
@jow- @hnyman noticed that now too, when i updated

hnyman:
@jow- 
Any blockers for merging this?
The nftables status screen would help users a lot.

hnyman:
@jow- 
I feel that we should merge this, as the snapshot users have been the last two weeks without a firewall status page.
The functionality may not be complete, yet, but to me it looks  good enough.

Any objections?

PR #5654, opened 2022-01-23
luci-mod-network: add support for UID iprules, add descriptions to routing/rules sections

clayface:
Please ensure [netifd](https://github.com/openwrt/openwrt/blob/master/package/network/config/netifd/Makefile) is updated beyond 14/01/22 (3043206e) to include [ed718768](https://git.openwrt.org/?p=project/netifd.git;a=commit;h=ed7187684685430ee6de49e551775badbee39761) before merging.

Allow user-specific routing policies to be defined in luci. A single UID
or range of UIDs may be defined.

Signed-off-by: Matthew Hagan <mnhagan88@gmail.com>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/routes.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/routes.js
@@ -38,11 +38,11 @@ return view.extend({
 			s.tab('general', _('General Settings'));
 			s.tab('advanced', _('Advanced Settings'));
 
-			o = s.taboption('general', widgets.NetworkSelect, 'interface', _('Interface'));
+			o = s.taboption('general', widgets.NetworkSelect, 'interface', _('Interface'), _('Specifies the logical interface name of the parent (or master) interface this route belongs to'));
 			o.loopback = true;
 			o.nocreate = true;
 
-			o = s.taboption('general', form.ListValue, 'type', _('Route type'));
+			o = s.taboption('general', form.ListValue, 'type', _('Route type'), _('Specifies the route type to be created'));
 			o.modalonly = true;
 			o.value('', 'unicast');
 			o.value('local');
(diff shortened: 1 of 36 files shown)


clayface:
netifd [updated](https://github.com/openwrt/openwrt/commit/7edd10f9dfd9010ac899ed837ce1b7929e8b68f9) so this is now mergeable, subject to review.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/routes.js
```diff
@@ -183,6 +183,11 @@ return view.extend({
 			o.datatype = 'uinteger';
 			o.placeholder = 10;
 
+			o = s.taboption('advanced', form.Value, 'uidrange', _('User identifier (UID)'));
```
hnyman, line 186: I think that the new (undocumented?) option might need a bit more explanation, and also an example of the allowed values (in the help text).

modules/luci-mod-network/htdocs/luci-static/resources/view/network/routes.js
```diff
@@ -183,6 +183,11 @@ return view.extend({
...
+			o.modalonly = true;
+			o.datatype = 'string';
+			o.placeholder = '1000/1000-1005';
```
hnyman, line 189: The placeholder mangles two alternative values together, right?  (a single value and a range)
Might be more clear just to put here one legitimate value, instead of an illegal value.
Better to have the allowed option values explained in the help text.

hnyman (commented):
I think that the new option might be clarified a bit. See my comments below.

Additionally/alternatively, I wonder if a help sentence should be added to the Advanced tab header, so that a newcomer might easier understand the values' role.  The items either restrict the applicability of the rule to certain conditions, or exclude by inverting conditions.

@dedeckeh , you committed the netifd change. Any ideas/comments here?

clayface:
@hnyman have added a second commit to deal with lack of explanation for these options, mostly taken from the documentation. Have also added `uidrange` to the [documentation](https://openwrt.org/docs/guide-user/network/routing/ip_rules#options_for_ipv4_and_ipv6_rules).

Can you check these and let me know if they are too long. I'll update the translation files once you are happy.

clayface:
Have fixed the js issues and updates the strings in the po files. Please see the images for example of descriptions.

Screenshot 2022-02-10 at 15 48 26
Screenshot 2022-02-10 at 15 48 34
Screenshot 2022-02-10 at 15 48 48
Screenshot 2022-02-10 at 15 48 56

PR #5663, opened 2022-01-26
luci-app-nft-qos: fix lua trace by moving limit_mac_enabled option.

qiuzi:
Signed-off-by: Lopot li <gxfclql@gmail.com>


feckert:
Ok thanks for your fix :+1:

cybermind2:
moving limit mac enabled option? How?

PR #5665, opened 2022-01-28
luci-app-frpc: use anonymous sections

brvphoenix:
1. Use anonymous sections and add name options to specify the proxy name
2. Add server name field for visitor mode

Depend on: https://github.com/openwrt/packages/pull/17712

Signed-off-by: Van Waholtz <vanwaholtz@gmail.com>


brvphoenix:
ping @feckert @ysc3839

PR #5679, opened 2022-02-08
add minimal APK compatibility

aparcar:
This PR is just to track my local changes required to run OpenWrt and LuCI with APK.

The main work happens over [here](https://github.com/openwrt/openwrt/pull/4294).


kaka1365 (approved):
Good

kaka1365:
Okay

aparcar:
I patched APK to accept LuCI versions so this is no longer required.

PR #5687, opened 2022-02-16
luci-app-socatg: add app

big-tooth:


feckert (commented):
- Please change the commit message to `luci-app-socat: add app`
- Also add a `Signed-off-by:` to your commit please have a look to the other commits
- Please add a new line in each file
- Use the socat https://github.com/openwrt/packages/blob/master/net/socat/Makefile backend from the packages feed

applications/luci-app-socatg/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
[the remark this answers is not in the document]
big-tooth: maybe,or not

applications/luci-app-socatg/README.md
```diff
@@ -0,0 +1,14 @@
+# luci-app-socatg
```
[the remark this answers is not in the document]
big-tooth: English is default version,now

applications/luci-app-socatg/root/etc/init.d/socatg
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh /etc/rc.common
```
feckert, line 1: If you integrate the new uci options into the socat package, then this should not be necessary.

https://github.com/openwrt/packages/blob/master/net/socat/files/socat.init

applications/luci-app-socatg/root/etc/uci-defaults/luci-socatg
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
feckert, line 1: If we use the socat backend from the package feed, then we do not need this, as it is reload/restart automatically by openwrt

applications/luci-app-socatg/luasrc/controller/socatg.lua
```diff
@@ -0,0 +1,9 @@
+module("luci.controller.socatg", package.seeall)
```
feckert, line 1: If you port this to javascript, then we need an ACL instead of this file.

applications/luci-app-socatg/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+PKG_NAME:=luci-app-socatg
+PKG_VERSION=1.2
+PKG_RELEASE:=1
+PKG_LICENSE:=MIT
+PKG_MAINTAINER:=drophair <https://github.com/big-tooth>
+LUCI_TITLE:=LuCI Support for socatg
+LUCI_PKGARCH:=all
+LUCI_DEPENDS:=+socat
+# PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
```
feckert, line 11: We do not need that

applications/luci-app-socatg/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+include $(INCLUDE_DIR)/package.mk
```
feckert, line 13: Please also replace the this lines with:

```
include ../../luci.mk

# call BuildPackage - OpenWrt buildroot signature
```
Have a look at the other application Makefiles in the repository

feckert (changes requested):
I am also in favour of not needing the readme, because there is no added value in using the socat backend from the package feed.

BKPepe:
@big-tooth Would you mind taking a look at your pull request and addressing the review? If not, there is a possibility that your pull request will be closed due to inactivity.

PR #5696, opened 2022-02-23
Introduce sortable status tables

jow-:
This PR introduces sortable status tables in LuCI and replaces the old `cbi_update_table()` functon with wrapper logic for backwards compatibility.

In a later step, form `TableSection` and `GridSection` implementations will be rebased on top of the new table facility to provide column sorting for them as well.

image

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/cbi.js
+++ b/modules/luci-base/htdocs/luci-static/resources/cbi.js
@@ -764,72 +764,14 @@ function cbi_update_table(table, data, placeholder) {
 	if (!isElem(target))
 		return;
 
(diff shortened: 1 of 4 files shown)


hnyman:
I applied this to a live router, and it mainly works ok. 
Nice.

I noticed some peculiarities:
For some reason the DHCP static leases MAC addresses sort a few MACs into a wrong place.
Other fields like IP address, lease time and name work ok.

image

jow-:
Thanks for testing, that quirks should be gone now in the latest iteration

castillofrancodamian:
I have noticed that after a few seconds the order of the list is reset.

PR #5704, opened 2022-03-02
luci-app-attendedsysupgrade: LuCIfy codebase

aparcar:
This should make the code a bit more readable and LuCI like instead of
using vanially JavaScript.


kaka1365 (commented):
Good

kaka1365:
Okay

PR #5710, opened 2022-03-08
luci-mod-status,mod-network: Added fqdn-name to DHCPv4 lease table 

MaxSKash:
(as in DHCPv6)

Signed-off-by: Max S Kash <asukms@ya.ru>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -726,8 +726,17 @@ return view.extend({
 							else
 								exp = '%t'.format(lease.expires);
 
+							var hint = lease.macaddr ? hosts[lease.macaddr] : null,
+								name = hint ? hint.name : null,
+								host = null;
+
+							if (name && lease.hostname && lease.hostname != name)
+								host = '%s (%s)'.format(lease.hostname, name);
+							else if (lease.hostname)
+								host = lease.hostname;
+
 							return [
-								lease.hostname || '?',
+								host || '-',
 								lease.ipaddr,
 								lease.macaddr,
 								exp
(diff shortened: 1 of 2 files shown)


feckert:
Please stash the two commits to one or update the second wit a valid commit message.

MaxSKash:
@feckert done.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -726,8 +726,17 @@ return view.extend({
 							else
 								exp = '%t'.format(lease.expires);
 
+							var hint = lease.macaddr ? hosts[lease.macaddr] : null,
+							name = hint ? hint.name : null,
```
feckert, line 730: Just one more small comment:
Either indent or `var` at the beginning and with `;` at the end

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -726,8 +726,17 @@ return view.extend({
...
+							host = null;
```
feckert, line 731: Same here

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -102,8 +102,16 @@ return baseclass.extend({
 			else
 				exp = '%t'.format(lease.expires);
 
+			var hint = lease.macaddr ? machints.filter(function(h) { return h[0] == lease.macaddr })[0] : null,
+				host = null;
```
feckert, line 106: Since you have done it this way here, I would suggest that you do the same for the other two notes.

feckert (changes requested):
I have tested it and it works so far. Please change the styling
Screenshot_20220324_110740

MaxSKash:
@feckert done.

feckert:
Fixed indentation and so pushed directly https://github.com/openwrt/luci/commit/8a46648e5bfb4b8f4d2452c51b5181631d260b6f

Thanks

PR #5712, opened 2022-03-09
luci-app-openvpn: Implement autocomplete Name field based on the uploaded file name

digitaltechconsult:
luci-app-openvpn: Implement autocomplete Name field based on the uploaded file name.
Signed-off-by: Bogdan Coticopol

### DIFF
--- a/applications/luci-app-openvpn/luasrc/view/openvpn/cbi-select-input-add.htm
+++ b/applications/luci-app-openvpn/luasrc/view/openvpn/cbi-select-input-add.htm
@@ -68,6 +68,18 @@
 			setTimeout(function(){ output.innerHTML=""}, timeout);
 		}
 	}
+		var vpn_file = document.getElementById("ovpn_file").value;
+
+		if(!vpn_name && !vpn_name.length) {
+			let vpn_file_name = vpn_file.replace('C:\\fakepath\\','').replace('.ovpn','').replace(/[^\x00-\x7F]|[\s\.!@#$%^&*()\-+=\[\]{};':"\\|,<>\/?]/g,'');
+			if (vpn_file_name && vpn_file_name.length) {
+				vpn_name = vpn_file_name;
+				document.getElementById("instance_name2").value = vpn_name;
+			}
+		}
+
+
+	}
 //]]>
 </script>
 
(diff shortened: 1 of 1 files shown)


applications/luci-app-openvpn/luasrc/view/openvpn/cbi-select-input-add.htm
```diff
@@ -67,6 +67,18 @@
 		{
 			setTimeout(function(){ output.innerHTML=""}, timeout);
 		}
+	}
```
feckert, line 70: Are you sure that the indentation is correct? The one with the bracket looks strange

applications/luci-app-openvpn/luasrc/view/openvpn/cbi-select-input-add.htm
```diff
@@ -67,6 +67,18 @@
...
+		var vpn_file = document.getElementById("ovpn_file").value;
+
+		if(!vpn_name && !vpn_name.length) {
+			let vpn_file_name = vpn_file.replace('C:\\fakepath\\','').replace('.ovpn','').replace(/[^\x00-\x7F]|[\s\.!@#$%^&*()\-+=\[\]{};':"\\|,<>\/?]/g,'');
```
feckert, line 74: I find the regex difficult to understand. Isn't there something simple here?
Maybe a path split see https://www.geeksforgeeks.org/how-to-get-the-file-name-from-full-path-using-javascript/

I also think that the assumption that it is about **windows** is not correct either.

eraser2021999: The regex was already in place and it checks the name for illegal characters. You can check the code where the name is setup manually.

systemcrash:
ping @eraser2021999

Neustradamus:
I have sent an email to Bogdan Coticopol about this PR.

applications/luci-app-openvpn/luasrc/view/openvpn/cbi-select-input-add.htm
systemcrash: https://developer.mozilla.org/en-US/docs/Web/API/FileList

Then avoid all the regex and just get the pure file name.

Issue #5714, opened 2022-03-09
switch.js: switch page add vlan failed if 2 or more switch exist


ptpt52:
/etc/board.json
```
...
        "switch": {
                "switch0": {
                        "enable": true,
                        "reset": true,
                        "ports": [
                                {
                                        "num": 0,
                                        "role": "lan"
                                },
                                {
                                        "num": 1,
                                        "role": "lan"
                                },
                                {
                                        "num": 2,
                                        "role": "lan"
                                },
                                {
                                        "num": 3,
                                        "role": "lan"
                                },
                                {
                                        "num": 4,
                                        "role": "wan"
                                },
                                {
                                        "num": 6,
                                        "device": "eth0",
                                        "need_tag": false,
                                        "want_untag": false
                                }
                        ],
                        "roles": [
                                {
                                        "role": "lan",
                                        "ports": "0 1 2 3 6t",
                                        "device": "eth0.1"
                                },
                                {
                                        "role": "wan",
                                        "ports": "4 6t",
                                        "device": "eth0.2"
                                }
                        ]
                },
                "switch1": {
                        "enable": true,
                        "reset": true,
                        "ports": [
                                {
                                        "num": 0,
                                        "role": "lan"
                                },
                                {
                                        "num": 1,
                                        "role": "lan"
                                },
                                {
                                        "num": 2,
                                        "role": "lan"
                                },
                                {
                                        "num": 3,
                                        "role": "lan"
                                },
                                {
                                        "num": 5,
                                        "device": "eth1",
                                        "need_tag": false,
                                        "want_untag": false
                                }
                        ],
                        "roles": [
                                {
                                        "role": "lan",
                                        "ports": "0 1 2 3 5t",
                                        "device": "eth1.1"
                                }
                        ]
                }
        },

...
```

switch.js not work property if 2 or more switch dev exist

ptpt52:
<img width="1088" alt="xx" src="https://user-images.githubusercontent.com/5331426/157489813-d9117487-46df-41c8-8f51-e53710fd2623.png">

ptpt52:
@jow-  hi
could you help fix this issue?

ptpt52:
switch display wrong:
config/network:
```
config switch
        option name 'switch0'
        option reset '1'
        option enable_vlan '1'

config switch_vlan
        option device 'switch0'
        option vlan '1'
        option ports '0 1 2 3 6t'

config switch_vlan
        option device 'switch0'
        option vlan '2'
[...]

PR #5718, opened 2022-03-11
luci-mod-network: switch: Fix multi instances handling

ptpt52:
This fix issue https://github.com/openwrt/luci/issues/5714

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
@@ -224,7 +224,7 @@ return view.extend({
 
 			s.filter = function(section_id) {
 				var device = uci.get('network', section_id, 'device');
-				return (device == switch_name);
+				return (device == this.device);
 			};
 
 			s.cfgsections = function() {
@@ -248,7 +248,7 @@ return view.extend({
 				    max_vid = 0;
 
 				for (var j = 0; j < sections.length; j++) {
-					if (sections[j].device != s.device)
+					if (sections[j].device != this.device)
 						continue;
 
 					var vlan = +sections[j].vlan,
@@ -261,7 +261,7 @@ return view.extend({
 						max_vid = vid;
 				}
 
-				uci.set('network', section_id, 'device', s.device);
+				uci.set('network', section_id, 'device', this.device);
 				uci.set('network', section_id, 'vlan', max_vlan + 1);
 
 				if (feat.vid_option)
@@ -269,8 +269,6 @@ return view.extend({
 
 				return this.map.save(null, true);
 			};
-
-			var port_opts = [];
 
 			o = s.option(form.Value, feat.vid_option || 'vlan', 'VLAN ID');
 			o.rmempty = false;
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
```diff
@@ -299,7 +297,11 @@ return view.extend({
 				return true;
 			};
 
+			o.port_opts = [];
+			var port_opts = o.port_opts;
```
jow-, line 301: @ptpt52 - this line (301) seems redundant. Can you please confirm that your fix still works after removing it?

ptpt52: no, it is not redundant.

jow-: The the code should be refactored. Having a `var port_opts` in both the outer and inner scope is confusing.

ptpt52: https://github.com/openwrt/luci/blob/c7cae787e18a4eea647978bb7ac6dae5fbad7d8b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js#L367
https://github.com/openwrt/luci/blob/c7cae787e18a4eea647978bb7ac6dae5fbad7d8b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js#L370

ptpt52: If convenient you can help rename it

jow-: Yep seen it now, could you please change it like that:
Merge line 300 & 301 into a chained assignment
```javascript
var port_opts = o.port_opts = [];
```

modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
```diff
@@ -299,7 +297,11 @@ return view.extend({
...
+
 			o.write = function(section_id, value) {
+				var port_opts = this.port_opts;
```
jow-, line 304: Drop this line...

modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
```diff
@@ -299,7 +297,11 @@ return view.extend({
...
 				var topology = this.section.topology,
 				    values = [];
```
jow-, line 307: ... and change all subsequent uses of `port_opts` within the `o.write()` function scope to `this.port_opts`

jow- (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch.js
ptpt52: done

jow-: Perfect, thank you!

PR #5729, opened 2022-03-18
luci-proto-wireguard: add generate psk button

lvoegl:
Adds a "Generate Key" button to be able to generate a preshared key within LuCI.

generate key button


feckert:
Thanks :-)

Issue #3293, opened 2019-11-13
luci-lib-jsonc: INT_MAX issue with 


lisaac:
Form luci-lib-json switch to luci-lib-jsonc, I found that if number_in_json > INT_MAX, it will return the INT_MAX instead the real value.
```lua
json = require "luci.jsonc"
test ='{"big_number":2147483649}'
-- 2147483649 is INT_MAX + 2
res = json.parse(test)
print(res.big_number)
-- got: 2147483647
```

feckert:
Which branch?
What shall he give back according to your opinion?

lisaac:
im using 18.06.4, and it should return the real value (2147483649), not the INT_MAX(2147483647)

lisaac:
@jow- 
Is there any possible to support int64 in luci.jsonc.stringify ?

```
print(luci.jsonc.stringify({int64=3000000000}))
-- got: {"int64":3000000000.0}
-- This is not an INT
```
I'm using luci-lib-json it works well:
```
print(luci.json.encode({int64=3000000000}))
-- {"int64":3000000000}
```

Thanks

jck112:
Can someone take a look at https://github.com/openwrt/luci/pull/5730? I'm trying to fix some of the issues upstream for aarch64 devices that manufacturers (like FriendlyElec) are currently patching in their builds.

PR #5730, opened 2022-03-19
luci-lib-jsonc: stringify int use json_object_new_int64 instead

jck112:
Fixes: #3293
Signed-off-by: Justin Klaassen <jck112@users.noreply.github.com>


jck112:
This should cover all int64 integers that can be represented as doubles. Per the [json-c documentation](https://json-c.github.io/json-c/json-c-0.12/doc/html/json__object_8h.html#ae6cab7d4c7c707b9053035a6405d3891) calling `json_object_new_int` simply invokes `json_object_new_int64` internally.

PR #5742, opened 2022-03-23
luci-theme-material: Use description field as page title if set

majik-sheff:
The page title defaults to the hostname.  If the "description" option is set in the system config, use it instead.

Signed-off-by: Tim Thorpe <timfthorpe@gmail.com>


feckert:
I am not entirely convinced. The description can be quite long and consist of several words.

feckert:
Then why isn't the hostname enough for you?

majik-sheff:
I have a bunch of devices that get moved around to different sites/jobs.  It's nice to be able to change the displayed title without changing the hostname.

feckert:
I just see the problem that the option description can be quite long.
And your change always adds this to the title when the description option is set.

Maybe we should add a config option to also show this in the title?

PR #5752, opened 2022-03-31
luci-mod-system: Add field to allow user-customizable branding

majik-sheff:
Add a field in the "Language and Style" tab of System to allow users to override
the title/branding of Luci (even emoji).  If left empty, behavior reverts to using the hostname.

Signed-off-by: Tim Thorpe <tim@tfthorpe.net>

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
@@ -242,6 +242,13 @@ return view.extend({
 			if (k[i].charAt(0) != '.')
 				o.value(uci.get('luci', 'themes', k[i]), k[i]);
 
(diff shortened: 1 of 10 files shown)


hnyman:
I don't like hiding the hostname.  That can easily create confusion if there are multiple routers and a customized "branded" build is used.

I would prefer adding an optional branding but keeping the hostname always visible.

majik-sheff:
> I don't like hiding the hostname. That can easily create confusion if there are multiple routers and a customized "branded" build is used.
> 
> I would prefer adding an optional branding but keeping the hostname always visible.

The idea behind making it optional was to give the end-user a choice.  Until they change it, it defaults to using the hostname.

stokito:
As a regular user I'm not sure why would I need this

feckert:
This pull request has been lying around for a while now. It can also no longer be used, since the LuCI is using ucode. If this is of further interest, please rebase and open it again.

PR #5765, opened 2022-04-12
luci-mod-system: mount sections are always mounted unless they are explicitly not mounted

### LINKED ISSUE #4862: luci-mod-system: fstab entry enabled by default
The underlying service and UCI configuration is such that an entry in /etc/config/fstab is enabled by default. However when showing the mountpoint entries in System -> Mount Points, the listing in Luci shows the checkbox as unselected if the enabled setting is not present in the config.

feckert:
fixes #4862


kaka1365 (approved):
Good

kaka1365:
Good

PR #5770, opened 2022-04-20
luci-app-rp-pppoe-server: use netlist template

dacrossman:
pppoe-server init script is expecting an OpenWRT interface name. The physical device name is looked up by the init script. 

So config 'interface' should be equal to an interface name e.g LAN not a physical device e.g eth0

This PR fixes the luci application to select Interfaces, not physical devices.


jow-:
Merged, thank you!

PR #5776, opened 2022-04-23
luci-app-https-dns-proxy: cbi-value-field div is not needed and adds unneeded margin

arbolitoloco1:
Signed-off-by: Santiago Kozak arbolitoloco1@protonmail.com


feckert:
Please use a proper commit header for example `luci-app-https-dns-proxy: cbi-value-field div is not needed and adds unneeded margin` and also sign off your work.

jow-:
Merged via 6b3bba36d1381d0cd72e77f096b7f6892fd3314f

PR #5785, opened 2022-04-28
luci-mod-simple: initial commit

jow-:
This commits imports luci-mod-simple, a vastly simplified shell only
exposing common operations instead of the whole range of available
uci settings.

The LuCI simple module has been originally developed for prplWrt but
was intentionally kept compatible with mainline OpenWrt.


BKPepe:
This looks interesting, @jow-! ;)
I wonder why it is a draft. Is there something how it is possible to help you with on this one?

modules/luci-mod-simple/root/usr/libexec/rpcd/topology
```diff
@@ -0,0 +1,239 @@
...
+	end
+
+	if not brforward[br] then
+		brforward[br] = {}
+
+		local f = io.open("/sys/class/net/" .. br .. "/brforward", "rb")
+		if f then
+			local ports = {}
+			for port in (fs.dir("/sys/class/net/" .. br .. "/brif") or function() end) do
+				local port_no = tonumber(readfile("/sys/class/net/" .. dev .. "/brif/" .. port .. "/port_no"), 16)
```
openwrt-ai, line 48: The `brif/` directory belongs to the bridge `br`, but this reads `port_no` from `/sys/class/net/<dev>/brif/...` using the function argument `dev`. In the normal call path `dev` is a port name (e.g. from the neigh loop, `to_port(to_master(rec.dev), ...)` actually passes the master, but `to_port` itself was also called as `to_port(dev, mac)` from `to_port`'s own neighbor caller where `dev` is a port). Whenever `dev != br`, this path does not exist, `port_no` is `nil`, the `ports` map stays empty, and every FDB entry resolves to no port — so `to_port()` silently returns `nil`. The loop iterates `brif` of `br`, so the read should use `br` too:

```suggestion
				local port_no = tonumber(readfile("/sys/class/net/" .. br .. "/brif/" .. port .. "/port_no"), 16)
```

---

modules/luci-mod-simple/root/usr/share/luci/menu.d/luci-mod-simple.json
```diff
@@ -0,0 +1,79 @@
...
+		"order": 5,
+                "action": {
+                        "type": "firstchild",
+                        "preferred": "status",
+                        "recurse": true
+                },
+                "auth": {
+                        "methods": [ "cookie:sysauth" ],
+                        "login": true
+                }
```
openwrt-ai, line 13: nit: these `action`/`auth` blocks of the `simple` node are indented with spaces, while the rest of the file (and the sibling nodes below) use tabs. Re-indent with tabs for consistency:

```suggestion
		"action": {
			"type": "firstchild",
			"preferred": "status",
			"recurse": true
		},
		"auth": {
			"methods": [ "cookie:sysauth" ],
			"login": true
		}
```

---

openwrt-ai (commented):
Reviewed the new `luci-mod-simple` module (fresh review, 1 commit). Two findings inline: a bridge-port resolution bug in the `topology` rpcd script, and an indentation nit in the menu JSON. CI reports no checks on the head commit.

---

PR #5786, opened 2022-04-29
luci-app-frpc/frps: fix service status bar style

msylgj:
mod the service status bar to use normaly 'cbi-section' class style.
Testing on openwrt master SNAPSHOT, also can be merge to openwrt-21.02/openwrt-22.03

before:
style-before
after:
style-fixed

### DIFF
--- a/applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
+++ b/applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
@@ -135,7 +135,7 @@ function getServiceStatus() {
 
 function renderStatus(isRunning) {
 	var renderHTML = "";
-	var spanTemp = "<span style=\"color:%s;font-weight:bold;margin-left:15px\">%s - %s</span>";
+	var spanTemp = '<em><span style="color:%s"><strong>%s %s</strong></span></em>';
 
 	if (isRunning) {
 		renderHTML += String.format(spanTemp, 'green', _("frp Client"), _("RUNNING"));
@@ -163,8 +163,8 @@ return view.extend({
 			});
 
 			return E('div', { class: 'cbi-map' },
-				E('div', { class: 'cbi-section'}, [
-					E('div', { id: 'service_status' },
+				E('fieldset', { class: 'cbi-section'}, [
+					E('p', { id: 'service_status' },
 						_('Collecting data ...'))
 				])
 			);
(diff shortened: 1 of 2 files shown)


applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
```diff
@@ -135,7 +135,7 @@ function getServiceStatus() {
 
 function renderStatus(isRunning) {
 	var renderHTML = "";
-	var spanTemp = "<span style=\"color:%s;font-weight:bold;margin-left:15px\">%s - %s</span>";
+	var spanTemp = "<em><b><font color=%s>%s %s</font></b></em>";
```
jow-, line 138: Hm, `<font>` tags are long deprecated, could you replace this with `<span style="color:%s">%s %s</span>` please? Also use `<strong>` instead of `<b>`.

applications/luci-app-frps/htdocs/luci-static/resources/view/frps.js
```diff
@@ -110,7 +110,7 @@ function getServiceStatus() {
 
 function renderStatus(isRunning) {
 	var renderHTML = "";
-	var spanTemp = "<span style=\"color:%s;font-weight:bold;margin-left:15px\">%s - %s</span>";
+	var spanTemp = "<em><b><font color=%s>%s %s</font></b></em>";
```
jow-, line 113: Same as above, please use `<span style...>` instead of `<font>`.

jow- (changes requested):

applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
msylgj: I see. It must be like this, right?
```
"<em><span style=\"color:%s;\"><strong>%s %s</strong></span></em>"
```

jow-: You can also enclose the string in single quotes, this way you do not need to escape the double quotes. Also the final semicolon in the `style="..."` attribute can be omitted:

```javascript
'<em><span style="color:%s"><strong>%s %s</strong></span></em>'
```

msylgj: oh my fault. Done it.

jow-:
Merged via fbd3a40c3e364c91a9effc1193456c8b3303b880

PR #5787, opened 2022-04-29
luci-theme-material: fix size of progressbar text

McGiverGim:
The text of the progressbar is bigger than the progressbar height. This removes the font-size element that produced that.

Signed-off-by: Miguel Angel Mulero <mcgivergim@gmail.com>

This is a screen capture of the before:
image

and after:
image

I've tested it under Chrome in Windows and Android.


jow-:
Merged, thank you!

PR #5788, opened 2022-04-29
luci-material-theme: reduce size of logout icon

McGiverGim:
The icon of the logout option overflows the text. This reduces the size of the icon to not overflow.

Signed-off-by: Miguel Angel Mulero <mcgivergim@gmail.com>

This is before:
image

And this is after:
image

I've tested it under Chrome in Windows and Android.


jow-:
Merged, thank you!

castillofrancodamian:
The icon is very close to the text and it is not like that in English.

McGiverGim:
It seems you are right. Maybe it depends on the size of the text. I will try to find a better solution and push a new PR.

PR #5789, opened 2022-04-29
luci-app-dockerman: Update to play nice with recent changes

Entropy512:
https://github.com/openwrt/luci/commit/180d39dcd2427e3c32c0ec7ecc3c7bfb48c0d0ab broke dockerman's appearance in the menus

Make a small change to dockerman's index entry to make it play nice with the new architecture, firstchild() was obtained using DAWN's controller/dawn.lua as an example

signed-off


Entropy512:
Commit message updated/reworded to address the failures of the previous formalities check, and force-pushed.

PR #5801, opened 2022-05-04
luci-mod-system: add interface selection for NTPD

odmdas:
UCI option `interface` for sysntpd server was introduced in
OpenWrt commit 4da60500ebd2. NTP server binds to the specified interface,
or if unspecified, to all.

This patch adds selection widget to LuCI.

Signed-off-by: Alexey Dobrovolsky <dobrovolskiy.alexey@gmail.com>

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
@@ -5,6 +5,7 @@
 'require uci';
 'require rpc';
 'require form';
+'require tools.widgets as widgets';
 
 var callInitList, callInitAction, callTimezone,
     callGetLocaltime, callSetLocaltime, CBILocalTime;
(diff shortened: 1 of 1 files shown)


modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
```diff
@@ -282,6 +283,14 @@ return view.extend({
 			o.ucisection = 'ntp';
 			o.depends('enabled', '1');
 
+			o = s.taboption('timesync', widgets.NetworkSelect, 'interface',
+				_('Bind NTP server to the specified interface or, if unspecified, to all'));
```
feckert, line 287: Please make the following changes so we have a short and a detail description before.
```suggestion
				_('Bind NTP server'),
				_('Provide the NTP server to the selected interface or, if unspecified, to all interfaces'));
```

odmdas: Corrected. Thanks for the advice!

feckert:
Thanks for your contribution

PR #5815, opened 2022-05-17
kmod-batman-adv compile from packge batman-adv

meferspb:


feckert:
This is correct, but here we are talking about runtime dependency and the package is called kmod-batman-adv.
https://github.com/openwrt/routing/blob/master/batman-adv/Makefile#L101

PR #5834, opened 2022-06-07
[RFC] Add luci-app-attendedsysupgrade as default package

aparcar:
This is more of a RFC, I'd like to see the app installed by default so more people use it and eventually use the latest (and greatest/safest/fastest) OpenWrt release possible. Telling someone to *click this button and wait 15 seconds, then click again* is much easier than any other way to upgrade a running system.

This pulls in packages like `ucert` and `attendedsysupgrade-common`, I'll look into the total space in a bit.

Another approach could be to integrate the code into the **Flash operations** view of LuCI.

### DIFF
--- a/collections/luci/Makefile
+++ b/collections/luci/Makefile
@@ -12,9 +12,16 @@ LUCI_BASENAME:=luci
 LUCI_TITLE:=LuCI interface with Uhttpd as Webserver (default)
 LUCI_DESCRIPTION:=Standard OpenWrt set including full admin with ppp support and the default Bootstrap theme
 LUCI_DEPENDS:= \
-	+uhttpd +uhttpd-mod-ubus +luci-mod-admin-full +luci-theme-bootstrap \
-	+luci-app-firewall +luci-app-opkg +luci-proto-ppp +libiwinfo-lua +IPV6:luci-proto-ipv6 \
-	+rpcd-mod-rrdns
+	+IPV6:luci-proto-ipv6 \
+	+libiwinfo-lua \
+	+luci-app-firewall \
+	+luci-app-opkg \
+	+luci-mod-admin-full \
+	+luci-proto-ppp \
+	+luci-theme-bootstrap \
+	+rpcd-mod-rrdns \
+	+uhttpd \
+	+uhttpd-mod-ubus
 
 PKG_LICENSE:=Apache-2.0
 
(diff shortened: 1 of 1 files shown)


aparcar:
@jow- thoughts on this?

josteink:
Right now, for most non-hardcore techies, installing OpenWRT is quite a daunting tasks, but some people may still be up for it and manage to get it done, over a weekend or something, with some concentrated effort and reading.

For those users, trying to upgrade their devices once they hear a new release is out, will most likely put them back where they started, requiring them to put in an almost equal amount of effort to look up the right image, upload the right file, and what about installed packages?

One may argue what should be minimally expected from users of OpenWRT firmware, but I think the real world picture is that this process is not reasonable or near trivial-enough for most users. The result of that is probably many people are running on obsolete/insecure OpenWRT-versions, because upgrading is too much effort.

Having attended-upgrade installed by default seems like a no-brainer to me and would eliminate most of those issues for most people. All thumbs up!

hnyman:
From the perspective of community builds for popular routers (e.g. several ipq806x devices), having the attendedsysupgrade LuCI feature built-in by default may cause confusion to people, as the users upgrading via attendedsysupgrade will not get the possible source modifications that are included in the original community build. They will get the same packages but not the source changes.  (And e.g. the NSS builds from @ACwifidude will be unupgradable.)

I wonder if I should de-select this package in my own build or just add a warning/infomessage into LuCI that using this to sysupgrade will lead you back to "normal" OpenWrt. Possibly just add a warning.

josteink:
I agree that from the perspective of (a user of) a community build, when that build makes customizations which deviate from the "standard" OpenWRT images available for the same devices, it is in the interest of the community to ensure that upgrades happen based on the "right" images or not at all.

Perhaps the atttended-upgrade package could be modified to have a config option for URL to poll/check for updates? And then the community-provided builds could override the default URL to a community controlled endpoint, where they can publish their own updates?

This could possibly also support the use-case of incremental builds from the community outside the main OpenWRT releases as well.

That leaves the community builds to options:

* explicitly remove the attended-upgrade package (lowest, near-zero effort)
* override the default URL, and publish their own update manifests (higher effort, but a much better user experience, even compared to community builds today)

What do you think @hnyman  ?

hnyman:
People compiling & publishing community builds have so varying build styles, build frequencies and download sites,  that I feel that there is no realistic way to have standard auc-accepted manifests & package download repos. 

I feel that the most realistic option is removing the package from the build (re-patching the LuCI dependency).

Other option for builds with relatively few changes would be to add a warning to LuCI.
E.g. for my own R7800 build, the users would mainly lose a few busybox customizations, a few wifi tweaks (802.11r defaults), see a change in reset button haviour and lose a few helper scripts. But nothing major (as most of my earlier changes since 2016 have by now been implemented into OpenWrt itself). So, I might opt for a warning text.

PR #5839, opened 2022-06-13
luci-app-upnp: Fix rpc handler for nftables

kode54:
Rewrite great portions of the rpcd handler script to utilize the
newer nftables implementation for firewall4-using versions. This
is required because miniupnpd-nftables creates tables that are
incompatible with iptables-nft. There may be some race conditions
with deleting the lease entries, since it opens the file to
locate the line number to delete. There could be a better way to
do this.

Signed-off-by: Christopher Snowhill <kode54@gmail.com>


hnyman:
I wonder if we would need separate luci-app-upnp-iptables and luci-app-upnp-nftables, as miniupnpd has both variants (and there are die-hard fw3 users)

Any thoughts, @jow-  ?

jow-:
Superseded by 82eec98944714c2b61f6e54291bc8f22983bb122

PR #5844, opened 2022-06-19
luci-app-statistics: Add initial support for DSL status.

bluebrother:
Handle data for lantiq dsl metrics.

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/dsl.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/dsl.js
@@ -7,8 +7,9 @@ return L.Class.extend({
 
 	rrdargs:  function(graph, host, plugin, plugin_instance, dtype) {
 		var g = [];
+		var dtypes = graph.dataTypes(host, plugin, plugin_instance);
 
-		g.push({
+		const d_snr = {
 			title: "DSL Signal",
 			vlabel: "dB",
 			data: {
@@ -46,8 +47,8 @@ return L.Class.extend({
 					},
 				}
 			}
-		});
-		g.push({
+		};
+		const d_uptime = {
 			title: "DSL Line Uptime",
 			vlabel: "seconds",
 			data: {
@@ -59,8 +60,8 @@ return L.Class.extend({
 					}
 				}
 			}
-		});
-		g.push({
+		};
+		const d_flags = {
 			title: "DSL Flags",
 			data: {
 				instances: {
@@ -94,8 +95,8 @@ return L.Class.extend({
 					},
 				}
 			}
-		});
-		g.push({
+		};
+		const d_bitrate = {
 			title: "Bitrate",
 			vlabel: "b/s",
 			data: {
@@ -130,8 +131,8 @@ return L.Class.extend({
 					}
 				}
 			}
-		});
-		g.push({
+		};
+			const d_count = {
 			title: "Errors",
 			vlabel: "count",
 			data: {
@@ -169,7 +170,24 @@ return L.Class.extend({
 					},
 				}
 			}
-		});
+		};
+
+		if (dtypes.includes("snr")) {
+			g.push(d_snr);
+		}
+		if (dtypes.includes("uptime")) {
+			g.push(d_uptime);
+		}
+		if (dtypes.includes("bool")) {
+			g.push(d_flags);
+		}
+		if (dtypes.includes("bitrate")) {
+			g.push(d_bitrate);
+		}
+		if (dtypes.includes("count")) {
+			g.push(d_count);
+		}
+
 		return g;
 	}
 });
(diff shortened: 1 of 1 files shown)


applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/dsl.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
```
feckert, line 1: Can you please follow the syntax of the other modules?
If it's all so compact, then you can't really track changes.
So please have a look at how the other modules have done it.

feckert (changes requested):

applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/dsl.js
```diff
@@ -0,0 +1,173 @@
...
+						title: "Data Rate Up",
+						noarea: true,
+						overlay: true
+					},
+					bitrate_data_rate_down: {
+						title: "Data Rate Down",
+						noarea: true,
+						overlay: true
+					}
+				}}});
```
feckert, line 131: I'm confused, can you indent the brackets correctly with new line?

feckert (changes requested):

feckert:
I have now had a look at it.
I must say I don't have dsl plugged into my system.
That means there are only `state_num`, `power_state_num` and `uptime` and I get the following image.

Screenshot 2022-07-12 at 15-58-12 VR2-106149 - Diagramme - LuCI

This is because I Only have the following rrd files on my system.
```
root@VR2-106149 /tmp/rrd/VR2-106149/dsl # ls -la
drwxr-xr-x    2 root     root           100 Jul 12 10:47 .
drwxr-xr-x    9 root     root           180 Jul 12 10:47 ..
-rw-r--r--    1 root     root          7096 Jul 12 15:59 gauge-power_state_num.rrd
-rw-r--r--    1 root     root          7096 Jul 12 15:59 gauge-state_num.rrd
-rw-r--r--    1 root     root          7096 Jul 12 15:59 uptime-uptime.rrd
```

The 3 other needed files are missing and so no rrd values could be read for image rendering.

The question now is, how can we solve this so that these artifacts do not appear with a non-existent rrd file?

bluebrother:
Finally managed to get back to this. PR updated, now only shows the graphs for which collectd has data -- if DSL was never connected it only shows the "DSL Line Uptime" graph, since that's the only one it has data. Once DSL gets connected and collectd starts to gather data for the other values (i.e. creates the other rrd files) the other graphs will show up.

Another approach would be to always show all graphs and instead change the collectd script to always report all data, and use zero values for the ones that are currently not available. Would that be preferrable?

feckert:
Thanks for this. And sorry for the delay I overlooked this. I pushed it directly to master via
https://github.com/openwrt/luci/commit/a1741bd13b3747f7ef9dd05d0b8dd29d4da46f86 and merged the two commits to one.
Should we also backport this openwrt-23.05?

bluebrother:
Added #6453 for enabling translations, and address the title issue that was mentioned in the code comments on in a1741bd13b3747f7ef9dd05d0b8dd29d4da46f86#r120400575

I think it would be nice to have this in 23.05. Should I create a PR for that or do you simply cherry-pick?

feckert:
@bluebrother I cherry-picked the requested changes to openwrt-23.05 with the following commits
https://github.com/openwrt/luci/commit/43b8c23e51c2262a98e6c73a6643c20760165f33
https://github.com/openwrt/luci/commit/2d23a3674939a78646f5e44eaa0522fabc859201
https://github.com/openwrt/luci/commit/2c7b34bd7b80023c2833bc7e007616840d1f20cc

cmhrky:
Hello,

DSL status currently available in luci-app-statistics?

I have lantiq/xrx200 OpenWrt SNAPSHOT r24256-1d6f9bdfc0 but I didn't see anything in the settings.

feckert:
Please have a look at https://github.com/openwrt/packages/blob/master/utils/collectd/Makefile#L535

cmhrky:
Thank you.

I tried to do something manually but I guess there is nothing I can do.
collectd-mod-lua, libubus-lua installed. I know the location of dsl.lua but the lua plugin does not appear in "Statistics > Setup ".

feckert:
As far as I can remember, the plugin cannot be configured, only the source is available.
You have to update the collectd.init to process this plugin as `https://github.com/openwrt/packages/blob/master/utils/collectd/files/collectd.init#L486` does. Or you could add the missing config  lines to a separate config file for [collectd](https://www.collectd.org/wiki/index.php/Plugin:Lua), so the file gets loaded. This is a missing feature. Pullreqeusts are welcome.

cmhrky:
I entered the lines related to the lua plugin in the collectd.conf file but without success. Also, the lua settings in collectd.con disappear every time the device restarts.

I hope soon dsl status will be available with collectd.

Thank you for reply.

PR #5852, opened 2022-06-24
Optimized PNG filesizes

y04:
Reduced PNG-images filesizes. All modified images except BMX7 logo and OpenWrt logo are visually identical. These two logos are visually the same.


jow-:
Merged, thanks!

PR #5863, opened 2022-06-30
luci-app-frpc: some improvement

jjm2473:
1. fix local_ip data type
2. add plugin options
3. show remote port on overview page
4. some translations fix


feckert:
Please update your commits with a valid Signed-off-by with your real name and a valid mail address.
See the other commits.

jjm2473:
ok, I will check it

jjm2473:
pushed to master https://github.com/openwrt/luci/pull/5935

PR #5864, opened 2022-06-30
[WIP] i18n: add missing translation files

feckert:
@hnyman @jow- 
I would like to add and sync the missing po files for languages that are already add to luci.mk
I have noticed this by adding the po files for new app **luci-app-xfrpc**.
I have also seen that we have languages in luci-app-acl ( and mybe in some other applications) that do not have an entry in luci.mk!

### DIFF
--- a/applications/luci-app-acme/po/nl/acme.po
+++ b/applications/luci-app-acme/po/nl/acme.po
@@ -1,0 +1,250 @@
+msgid ""
+msgstr ""
+"Content-Type: text/plain; charset=UTF-8\n"
+"Project-Id-Version: PACKAGE VERSION\n"
+"Last-Translator: Automatically generated\n"
+"Language-Team: none\n"
+"Language: nl\n"
+"MIME-Version: 1.0\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#: applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js:17
+msgid "ACME certificates"
+msgstr ""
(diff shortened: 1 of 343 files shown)


hnyman:
Why?
Weblate initializes missing languages automatically when somebody actually starts translation.
What is the benefit from adding files without any actual translation done?

feckert:
@hnyman I was not aware that weblate does that but that is create!

But there is still a problem.
At first I was wondering why there are translation files even there is no entry in luci.mk.

As an example, I would like to point out the translation bo (Bosnian) https://github.com/openwrt/luci/tree/master/applications/luci-app-acl/po does exist but could not be build as a package because I cannot select this under `make menuconfig Luci -> Modules -> Translation`

If that is the case, that weblate does all this, then we should also adapt the luci.mk and add the missing LUCI_LANG.* entries.

hnyman:
Some of those languages are started but only a few strings have actually translated. (There aren't that many enthusiasts for small languages and the initial person may tire quickly.)

E.g. Hindi, Malay, Sinhala, Swahili, yue 
https://hosted.weblate.org/projects/openwrt/#languages

There was some forum(?) discussion some time ago about adding new languages, and the general feeling was that at base.po should be first translated to some extent before the language has been added to luci.mk.

  I have selectively added new languages to luci.mk when I have noticed serious translation activity about a language.

There is no automation, as weblate allow starting translation for any language in its catalog, and we have a static list in luci.mk

If you feel that there are half-translated languages that should be enabled, just add them to luci.mk.

feckert:
Thank you for the explanation. Then everything should fit so far

PR #5867, opened 2022-07-03
Fix package name for wpa-supplicant

Toomoch:
No wpa_supplicant exists, it's actually wpa-supplicant

Signed-off-by: macaquinyho <vallsfustearnau@gmail.com>


Neustradamus:
@feckert, @jow-: What do you think of this little change to have the good name?

feckert:
Should fit. Unfortunately, the headline is not correct.
@Toomoch  can you please change it like this` luci-mod-network: fix package naming`. And also please add your realname to your Signed-off-by

feckert:
It is not the package name. It is binary. Therefore, that is correct.

PR #5879, opened 2022-07-11
Luci-proto-ncm: Renamed 'service' to 'mode'

### LINKED ISSUE #5877: Luci-proto-ncm sets wrong option to control radiomode for interface of type NCM
## Steps to reproduce:
**Example: Replace the following lines and remove this**

1. go to: Network → Interfaces → WWAN0 (the interface used for NCM)
2. Click edit

## Actual behavior:

1. Service type is displayed with pulldown showing lte/preferlte/umts/etc
2. This gets added as 'option service '<servicetype>' to /etc/config/network

## Expected behavior:

1. Network mode is displayed with pulldown showing lte/preferlte/umts/etc
2. This gets added as 'option mode '<servicetype>' to /etc/config/network

## Additional Information:
Package luci-proto-ncm depends on gcom-ncm. Gcom-ncm uses the option called 'mode' to set the radiomode of the modem. There is no option 'service' in the gcom-ncm scripts.

breenstorm:
fixes #5877

### DIFF
--- a/protocols/luci-proto-ncm/htdocs/luci-static/resources/protocol/ncm.js
+++ b/protocols/luci-proto-ncm/htdocs/luci-static/resources/protocol/ncm.js
@@ -69,7 +69,7 @@ return network.registerProtocol('ncm', {
 			}, this));
 		};
 
-		o = s.taboption('general', form.Value, 'service', _('Service Type'));
+		o = s.taboption('general', form.Value, 'mode', _('Network Mode'));
 		o.value('', _('Modem default'));
 		o.value('preferlte', _('Prefer LTE'));
 		o.value('preferumts', _('Prefer UMTS'));


breenstorm:
Good suggestion. Although this would be easier to configure, I do see problems with this. As the only way for modeminfo to know what brand of modem ncm has detected is to query the modem over serial, this would mean the setuppage for modeminfo would become somewhat dependant on a properly functioning and correctly configured tty. Personally, I would advice against making the possibility to configure dependant on a working situation, as in most cases where people start is a non-configured and this non-working setup.

Adding checks to circumvent these restrictions would be possible of course, but are well outside the scope of this PR. You are welcome and encouraged to submit your own PR for more realistic options in the pulldown for the networkmode.

feckert:
All right then, let's do it that way. When I feel like it, I will restrict the dropdown.
In order for me to merge the pullrequest, you need to add a valid `Signed-off-by` with correct **full** name,
Also you have to adjust the commit heading in your commit. Please use the prefix `luci-proto-ncm:` so we have the heading for this change `luci-proto-ncm: rename service to mode` as in your pullrequest heading.

jow-:
Merged via 4d6642b636a9feaa173c568c99f4129c440ed8b8

PR #5880, opened 2022-07-11
Update nft-qos.po

canmengxian:
添加了一些翻译

Signed-off-by:  残梦 <37858783+canmengxian@users.noreply.github.com>


feckert:
Please use [weblate](https://hosted.weblate.org/projects/openwrt/) to add your translation.
The reason why this string are not get translated is that this is not wrapped by the function `translate("<string>")`.
I pushed the following fixes https://github.com/openwrt/luci/commit/d3a6ba9d55ccfb83e30abc350735fca1e308c6e3 https://github.com/openwrt/luci/commit/233b83d6e56bb7b01717be186aa6a7acd896d90b

PR #5887, opened 2022-07-16
luci-app-apinger: Add LuCI for Apinger

jempatel:
A new app for apinger with LuCI Support. This new LuCI app is added on top of the recently merged apinger PR
https://github.com/openwrt/packages/pull/18913, It covers backend UCI configuration information.

**We also need https://github.com/openwrt/packages/pull/19073**

Screenshots from the new Luci App for apinger:

Overview:
Apinger_Overview

Interfaces:
Apinger_Interfaces

Down Alarm:
Apinger_Alarm_Down

Delay Alarm:
Apinger_Alarm_Delay

Loss Alarm:
Apinger_Alarm_Loss

Targets:
Apinger_Targets

@feckert

### DIFF
--- a/applications/luci-app-apinger/Makefile
+++ b/applications/luci-app-apinger/Makefile
@@ -1,0 +1,19 @@
+#
+# Copyright (C) 2022 Jaymin Patel <jem.patel@gmail.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for the Apinger
+LUCI_DEPENDS:=+apinger +apinger-rrd
+LUCI_PKGARCH:=all
+PKG_LICENSE:=GPL-2.0
+
+PKG_MAINTAINER:=Jaymin Patel <jem.patel@gmail.com>
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
+
--- a/applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/alarm_delay.js
+++ b/applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/alarm_delay.js
@@ -1,0 +1,30 @@
+'use strict';
+'require view';
+'require form';
+
+return view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('apinger', _('Apinger - Delay Alarms'), 
+			 ('This alarm will be fired when target responses are delayed more than "Delay High"') + '<br />' +
+			_('This alarm will be canceled, when the delay drops below "Delay Low"') + '<br />');
+
+		s = m.section(form.GridSection, 'alarm_delay');
+		s.anonymous = false;
+		s.addremove = true;
+		s.addbtntitle = _('Add Delay/Latency Alarm');
+
+		o = s.option(form.Value, 'delay_low', _('Delay Low (ms)'));
+		o.datatype = 'range(1-500)';
+		o.default = '30';
+		o.placeholder = '30';
+
+		o = s.option(form.Value, 'delay_high', _('Delay High (ms)'));
+		o.datatype = 'range(1-500)';
+		o.default = '50';
+		o.placeholder = '50';
+
+		return m.render();
+	},
+});
--- a/applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/targets.js
+++ b/applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/targets.js
@@ -1,0 +1,80 @@
+'use strict';
+'require view';
+'require form';
+'require uci';
+
+return view.extend({
+	load: function() {
+		return Promise.all([
+			uci.load('apinger'),
+		])
+	},
+
+	render: function(data) {
+		var m, s, o;
+		var a_ifaces, a_down, a_delay, a_loss;
+
+		a_ifaces = uci.sections('apinger', 'interface');
+		a_down = uci.sections('apinger', 'alarm_down');
+		a_delay = uci.sections('apinger', 'alarm_delay');
+		a_loss = uci.sections('apinger', 'alarm_loss');
+
+		m = new form.Map('apinger', _('Apinger - Targets'),
+			_('Interface: Interface to use to track target') + '<br />' +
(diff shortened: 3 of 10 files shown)


applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/alarm_delay.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+'require view';
+'require form';
+
+return view.extend({
+	load: function() {
+	},
+
+	renderDelayAlarms: function(m) {
+		var alarm_delay, alarm_delay_dl, alarm_delay_dh;
```
feckert, line 10: Please use the common `s m` pattern and not for every option and section an new variable name.

applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/targets.js
```diff
@@ -0,0 +1,86 @@
...
+		return m;
+	},
+
+	render: function(data) {
+		var m;
+
+		a_ifaces = uci.sections('apinger', 'interface');
+		a_down = uci.sections('apinger', 'alarm_down');
+		a_delay = uci.sections('apinger', 'alarm_delay');
+		a_loss = uci.sections('apinger', 'alarm_loss');
```
feckert, line 72: Why are these variables global can't we have them local?

applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/targets.js
```diff
@@ -0,0 +1,86 @@
...
+
+		m = new form.Map('apinger', _('Apinger - Targets'),
+			_('Interface: Interface to use to track target') + '<br />' +
+			_('Address: Target address to be tracked') + '<br />' +
+			_('Ping Interval: How often the probe should be sent') + '<br />' +
+			_('Average Delay: How many replies should be used to compute average delay') + '<br />' +
+			_('Average Loss: How many probes should be used to compute average loss') + '<br />' +
+			_('Average Delay and Loss: The delay (in samples) after which loss is computed, without this delays larger than interval would be treated as loss') +
+			'<br />');
+		m = this.renderTargets(m);
```
feckert, line 82: Why do you have an extra function in all views? I would prefer the content of the called function to be inserted here. Instead of calling this. This would also allow you to keep the variables local.

feckert (changes requested):

applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/alarm_delay.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+'require view';
+'require form';
+
+return view.extend({
+	load: function() {
```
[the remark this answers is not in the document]
jempatel: Updated

applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/alarm_delay.js
jempatel: Updated

applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/targets.js
jempatel: Updated

applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/targets.js
jempatel: Updated

applications/luci-app-apinger/root/usr/share/rpcd/acl.d/luci-app-apinger.json
```diff
@@ -0,0 +1,15 @@
+{
+	"luci-app-apinger" : {
+		"description" : "Grant access to LuCI app Apinger",
+		"read" : {
+			"ubus" : {
+				"apinger" : [ "*" ]
```
feckert, line 6: I think we also need a `read` and `write` uci acl for arpinger uci file

feckert (changes requested):

applications/luci-app-apinger/root/usr/share/rpcd/acl.d/luci-app-apinger.json
jempatel: Is it really needed? From LuCI, it's not reading/writing to /etc/config/apinger explicitly, instead, Its using RPC to get apinger overview.

feckert: As I see it, you are using the [uci handling](https://github.com/openwrt/luci/pull/5887/files#diff-43eb48c6a064931b875ac7df92656e5fef64b0d46101ec2d94b68b7984728768R4).
You can create a user and give him only the authorisation for the luci-app-apinger, then we will see if it also works without the uci acl.

jempatel: Yes, You are right. Updated rpcd ACL to handle apinger config.

applications/luci-app-apinger/Makefile
```diff
@@ -0,0 +1,19 @@
...
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for the Apinger
+LUCI_DEPENDS:=+apinger
+LUCI_PKGARCH:=all
+PKG_LICENSE:=GPLv2
```
feckert, line 12: Sorry, but can you use a valid `[SPDX](https://spdx.org/licenses/)` here?
maybe `PKG_LICENSE:=GPL-2.0` because `PKD_LICENSE:=GPLv2` is not valid

feckert (changes requested):
Except for the license, it looks good :+1:

jempatel:
Depends on PR
https://github.com/openwrt/packages/pull/19073

applications/luci-app-apinger/Makefile
jempatel: Updated.

feckert:
Lets try it. Thanks for your pullrequest

PR #5894, opened 2022-07-23
luci-proto-wireguard backports

jow-:
This PR backports the WireGuard protocol handler improvements from master and all related required changes to luci-base.


f00b4r0:
Quick update after some testing:
- there is a bug in the main import system: if I dump a conf with several peers, only the last one is imported. If the last peer has no endpoint configured, the system mangles the endpoint of the penultimate peer with the pubkey and allowed IPs of the last.
- a small nit for the peer import: the popup says "import as peer", which is explicit, but the button shows "Import peer configuration..." where I would expect to be able to dump a lone [Peer] section and have it imported, which is not possible (although would be nice). I would suggest renaming the button "Import configuration as peer..." to match the popup.

HTH

jow-:
All changes have been backported.

PR #5895, opened 2022-07-24
luci-mod-dsl: initial commit

turboproc:
This commit imports the Luci DSL Spectrum graph into LUCI. The spectrum graph shows up as an option in the toplevel
status menu. Code works completely standalone, no dependency on other packages (e.g. chart.js)

Signed-off-by: Roland Barenbrug <roland@treslong.com>

### DIFF
--- a/modules/luci-mod-dsl/Makefile
+++ b/modules/luci-mod-dsl/Makefile
@@ -1,0 +1,16 @@
+#
+# Copyright (C) 2022 Roland Barenbrug <roland@treslong.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI DSL spectrum graph
+LUCI_DEPENDS:=+luci-base
+
+PKG_LICENSE:=Apache-2.0
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/modules/luci-mod-dsl/Makefile
+++ b/modules/luci-mod-dsl/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=LuCI DSL spectrum graph
-LUCI_DEPENDS:=+luci-base
+LUCI_DEPENDS:=@TARGET_lantiq_xrx200 +luci-base +ltq-dsl-base
 
 PKG_LICENSE:=Apache-2.0
 
--- a/modules/luci-mod-dsl/Makefile
+++ b/modules/luci-mod-dsl/Makefile
@@ -6,7 +6,7 @@
 
 include $(TOPDIR)/rules.mk
 
-LUCI_TITLE:=LuCI DSL spectrum graph
+LUCI_TITLE:=LUCI DSL spectrum graph
 LUCI_DEPENDS:=@TARGET_lantiq_xrx200 +luci-base +ltq-dsl-base
 
 PKG_LICENSE:=Apache-2.0
--- a/modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js
+++ b/modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js
@@ -1,0 +1,388 @@
+//
+// Rendering of DSL spectrum graphs  showing
+// US/DS SNR and US/DS bits/tone
+//
+// This version does not depend on chart.js or
+// any other package
+
+ class DataSet {
+    constructor (inputString, extractFunction) {
+       this.groupSize = this.getGroupSize(inputString);
+       this.dataPos = this.getDataPos(inputString);
+       this.numData = this.getNumData(inputString);
+       if (!isNaN(this.groupSize)) {                   // this is the tricky part as output of all commands is not consistant
+          this.maxX = this.numData * this.groupSize;   // needs to be validated with various input
+       } else {
+          this.maxX = this.numData;
+          this.groupSize = 1
+       }
+       this.data = inputString.substr(this.dataPos).slice(0, -5).split(" ").map(extractFunction,{groupSize: this.groupSize});
+   }
+
+   getGroupSize(data) {
+      let groupSizePos = data.search('nGroupSize=') + 11
+      return (parseInt(data.substr(groupSizePos,2),10))
+   }
+
+   getDataPos(data) {
+      let dataPos = data.search('nData=') + 7
+      return (dataPos)
+   }
+
+   getNumData(data) {
+      let nNumDataPos = data.search('nNumData=') + 9
+      return (parseInt(data.substr(nNumDataPos,4),10))
+   }
+
+ }
+ 
+ const usSnrData   = new DataSet(document.getElementById("usdB").innerText, mySnrFunction);
+ const dsSnrData   = new DataSet(document.getElementById("dsdB").innerText, mySnrFunction); 
+ const usBitsData = new DataSet(document.getElementById("usBits").innerText, myBitsFunction);
+ const dsBitsData = new DataSet(document.getElementById("dsBits").innerText, myBitsFunction);
+ const usQLNData  = new DataSet(document.getElementById("usQln").innerText, myQLNFunction);
+ const dsQLNData  = new DataSet(document.getElementById("dsQln").innerText, myQLNFunction);
+ const usHLOGData = new DataSet(document.getElementById("usHLog").innerText, myHLOGFunction);
+ const dsHLOGData = new DataSet(document.getElementById("dsHLog").innerText, myHLOGFunction);
+
+
+ function myBitsFunction(value, index, array) {
+    return({x: parseInt(index), y: parseInt(value, 16), error: false});
+ }
+
+ function mySnrFunction(value, index, array) {
+    let t = value.replace(/[\(\)\n]/g,'').split(',');
+    let result;
+
+    let v = t[1];
+    let i = t[0];
+    if (v == 255) {                                                                      // this is the only special value
+       result = {x: parseInt(i) * this.groupSize, y: -40 , error: true}
+    } else {
+       result = {x: parseInt(i) * this.groupSize, y: parseInt(- 32 + v / 2), error: false} // rec ITU-T G.933.2, section 11.4.1.1.3, SNR(k × G × Δf) = −32 + (snr(k)/2) dB
+    };
+    return(result);
+ }
+
+ function myQLNFunction(value, index, array) {
+    let t = value.replace(/[\(\)\n]/g,'').split(',');
+    let result;
+
+    let v = t[1];
+    let i = t[0];
+    if (v == 255) {                                                                 // this is the only  special value defined
+       result = { x: parseInt(i) * this.groupSize, y:  - 150, error: true}
+    } else {
+       result = { x: parseInt(i) * this.groupSize, y:  - 23 - v / 2, error: false}  // ref ITU-T G.933.2, section 11.4.1.1.2, QLN(k × G × Δf) = −23 − (n(k)/2) dBm/Hz
+    };
+    return(result);
+ }
+
+ function myHLOGFunction(value, index, array) {
+    let t = value.replace(/[\(\)\n]/g,'').split(',');
+
+    let v = t[1];
+    let i = t[0];
+    let result;
+    if (v == 1023) {                                                            // this is the only special value defined
+       result = {x: parseInt(i) * this.groupSize, y: -100, error: true}
+    } else {
+       result = {x: parseInt(i) * this.groupSize, y: 6 - v / 10, error: false}  // ref ITU-T G.933.2,  section 11.4.1.1.1, Hlog(k × G × Δf) = 6 − (m(k)/10)
+    };
+    return(result);
+ }
+
+
+const marginX = 50;
+const marginY = 80;
+let darkMode = document.getElementsByTagName("body")[0].parentNode.dataset.darkmode;
+
+
+bitsChart = {
+   "config": {
+       "canvas": document.getElementById("bitsChart"),
+       "ctx" : document.getElementById("bitsChart").getContext("2d"),
+       "minX" : 0,
+       "maxX" : Math.max(dsBitsData.maxX, usBitsData.maxX),
+       "stepX": Math.max(dsBitsData.maxX, usBitsData.maxX) / 16,
+       "graphWidth" : document.getElementById("bitsChart").width - 2 * marginX,
+       "lineWidth" : 1,
+       "titleX" : "Sub-carrier",
+       "minY" : 0,
+       "maxY" : 16,
+       "stepY": 2,
+       "graphHeight" : document.getElementById("bitsChart").height - 2 * marginY,
+       "titleY" : "bits",
+   },
+   "dataSet" : [
+      {
+         "data" :usBitsData.data,
+         "color":"YellowGreen",
+         "title": "Upstream bits allocation"
+      },
+      {
+         "data" : dsBitsData.data,
+         "color": "navy",
+         "title": "Downstream bits allocation"
+      }
+   ] 
+};
+
+dBChart = {
+   "config": {
+       "canvas": document.getElementById("dbChart"),
+       "ctx" : document.getElementById("dbChart").getContext("2d"),
+       "minX" : 0,
+       "maxX" : Math.max(dsSnrData.maxX, usSnrData.maxX),
+       "stepX": Math.max(dsSnrData.maxX, usSnrData.maxX) / 16,
+       "graphWidth" : document.getElementById("dbChart").width - 2 * marginX,
+       "lineWidth": 4,
+       "titleX" : "Sub-carrier",
+       "minY" : -40,
+       "maxY" : 100,
+       "stepY": 10,
+       "graphHeight" : document.getElementById("dbChart").height - 2 * marginY,
+       "titleY" : "dB"
+   },
+   "dataSet" : [
+      {
+         "data" :usSnrData.data,
+         "color":"Turquoise",
+         "title": "Upstream SNR"
+      },
+      {
+         "data" : dsSnrData.data,
+         "color": "Coral",
+         "title" : "Downstream SNR"
+      }
+   ]
+
+};
+
+
+qLNChart = {
+   "config": {
+       "canvas": document.getElementById("qlnChart"),
+       "ctx" : document.getElementById("qlnChart").getContext("2d"),
+       "minX" : 0,
+       "maxX" : Math.max(dsQLNData.maxX, usQLNData.maxX),
+       "stepX": Math.max(dsQLNData.maxX, usQLNData.maxX) / 16,
+       "graphWidth" : document.getElementById("qlnChart").width - 2 * marginX,
+       "lineWidth": 4,
+       "titleX" : "Sub-carrier",
+       "minY" : -150,
+       "maxY" : -20,
+       "stepY": 10,
+       "graphHeight" : document.getElementById("qlnChart").height - 2 * marginY,
+       "titleY" : "dBm/Hz"
+   },
+   "dataSet" : [
+      {
+         "data" :usQLNData.data,
+         "color":"brown",
+         "title": "Upstream QLN"
+      },
+      {
+         "data" : dsQLNData.data,
+         "color": "teal",
+         "title" : "Downstream QLN"
+      }
+   ]
+
+};
+
+
+hLogChart = {
+   "config": {
+       "canvas": document.getElementById("hlogChart"),
+       "ctx" : document.getElementById("hlogChart").getContext("2d"),
+       "minX" : 0,
+       "maxX" : Math.max(dsHLOGData.maxX, usHLOGData.maxX),
+       "stepX": Math.max(dsHLOGData.maxX, usHLOGData.maxX) / 16,
+       "graphWidth" : document.getElementById("hlogChart").width - 2 * marginX,
+       "lineWidth": 4,
+       "titleX" : "Sub-carrier",
+       "minY" : -100,
+       "maxY" : 14,
+       "stepY": 10,
+       "graphHeight" : document.getElementById("hlogChart").height - 2 * marginY,
+       "titleY" : "dB"
+   },
+   "dataSet" : [
+      {
+         "data" :usHLOGData.data,
+         "color":"#E8E800",
+         "title": "Upstream HLOG"
+      },
+      {
+         "data" : dsHLOGData.data,
+         "color": "darkmagenta",
+         "title" : "Downstream HLOG"
+      }
+   ]
+
+};
+
+
+function drawChart (info) {
+   drawAxisX(info.config, info.config.minX, info.config.maxX, info.config.stepX, info.config.titleX);
+   drawAxisY(info.config, info.config.minY, info.config.maxY, info.config.stepY, info.config.titleY);
+
+   drawLegend(info.config, info.dataSet);
+
+   drawData(info.config, info.dataSet[0].data, info.dataSet[0].color);
+   drawData(info.config, info.dataSet[1].data, info.dataSet[1].color);
+}
+
+
+function drawBlocks(config, dataPoints, color, borders) {
+
+   borders.map(drawBlock,{config, dataPoints, color, borders});
+}
+
+
+function drawData(config, dataPoints, color) {
+   let ctx = config.ctx;
+   let len = dataPoints.length;
+   let minX =config.minX;
+   let maxX = config.maxX;
+   let minY = config.minY;
+   let maxY = config.maxY;
+   let startX = (dataPoints[0].x  - config.minX) / (config.maxX - config.minX)
+   let startY = (config.minY - config.minY) / (config.maxY - config.minY)
+
+   ctx.fillStyle = color;
+   ctx.beginPath();
+   ctx.moveTo(startX * config.graphWidth + marginX, marginY + config.graphHeight - startY * config.graphHeight);
+
+   for (let i  = 1 ;  i < len  ; i++) {
+      let relX = (dataPoints[i].x - minX) / (maxX - minX);
+      let relY = (dataPoints[i].y - minY) / (maxY - minY);
+      ctx.lineTo(relX * config.graphWidth + marginX, marginY + config.graphHeight - relY * config.graphHeight);
+   }
+
+   let endX = (dataPoints[len-1].x - minX) / (maxX - minX)
+   let endY = (config.minY - minY) / (maxY - minY)
+
+   ctx.lineTo(endX * config.graphWidth + marginX, marginY + config.graphHeight - endY * config.graphHeight);
+   ctx.lineTo(startX * config.graphWidth + marginX, marginY + config.graphHeight - startY * config.graphHeight);
+   ctx.closePath();
+   ctx.fill();
+
+}
+
+
+function drawLegend(config, dataSet){
+   let ctx = config.ctx;
+
+   let graphWidth = config.graphWidth;
+   let graphHeight = config.graphHeight;
+
+   ctx.font = "12px Arial";
+   ctx.fillStyle = dataSet[0].color;
+   ctx.fillRect(                  0.5 * graphWidth + marginX - ctx.measureText(dataSet[0].title).width - 50, config.canvas.height - marginY*1/4 - 8, 30, 10);
+   ctx.strokeStyle = "#C0C0C0";
+   ctx.strokeRect(                  0.5 * graphWidth + marginX - ctx.measureText(dataSet[0].title).width - 50, config.canvas.height - marginY*1/4 - 8, 30, 10);
+
+   if (darkMode == "true") {
+      ctx.strokeStyle = "#505050";
+      ctx.fillStyle = "#A0A0A0";
+   } else {
+      ctx.strokeStyle = "#303030";
+      ctx.fillStyle = "#303030";
+   }
+
+   ctx.textAlign = "right"
+   ctx.fillText(dataSet[0].title, 0.5 * graphWidth + marginX - 10, config.canvas.height - marginY*1/4);
+
+   ctx.fillStyle = dataSet[1].color;
+   ctx.fillRect(      0.5 * graphWidth + marginX,                  config.canvas.height - marginY*1/4 - 8, 30, 10);
+   ctx.strokeStyle = "#C0C0C0";
+   ctx.strokeRect(      0.5 * graphWidth + marginX,                  config.canvas.height - marginY*1/4 - 8, 30, 10);
+
+   if (darkMode == "true") {
+      ctx.fillStyle = "#A0A0A0";
+   } else {
+      ctx.fillStyle = "#303030";
+   }
+
+   ctx.textAlign = "left"
+   ctx.fillText(dataSet[1].title, 0.5 * graphWidth + marginX + 40, config.canvas.height - marginY*1/4);
+}
+
+function drawAxisX(config, minValue, maxValue, step, title) {
+   let ctx = config.ctx;
+
+   let graphWidth = config.graphWidth;
+   let graphHeight = config.graphHeight;
+
+   ctx.font = "12px Arial";
+   ctx.textAlign = "center";
+
+   if (darkMode == "true") {
+      ctx.strokeStyle = "#505050";
+      ctx.fillStyle = "#A0A0A0";
+   } else {
+      ctx.strokeStyle = "#E0E0E0";
+      ctx.fillStyle = "#303030";
+   }
+
+   for (let x = minValue ; x <= maxValue ; x=x+step) {
+      let relX = (x - config.minX) / (config.maxX - config.minX);
+      ctx.fillText(x , relX * graphWidth + marginX,  config.canvas.height - marginY*3/4);
+
+      ctx.beginPath();
+      ctx.moveTo(relX * graphWidth + marginX, marginY);
+      ctx.lineTo(relX * graphWidth + marginX, config.canvas.height - marginY);
+      ctx.stroke();
+
+   }
+   ctx.font = "12px Arial";
+   ctx.textAlign = "center";
+   ctx.fillText(title, config.canvas.width/2, config.canvas.height - marginY*2/4);
+}
+
+function drawAxisY(config, minValue, maxValue, step, title) {
+   let ctx = config.ctx
+
+   let graphWidth = config.graphWidth;
+   let graphHeight = config.graphHeight;
+
+   ctx.font = "12px Arial";
+   ctx.textAlign = "center";
+
+   if (darkMode == "true") {
+      ctx.strokeStyle = "#505050";
+      ctx.fillStyle = "#A0A0A0"; 
+  } else {
+      ctx.strokeStyle = "#E0E0E0";
+      ctx.fillStyle = "#303030";
+   }
+
+   for (let y = minValue ; y <= maxValue ; y=y+step) {
+      let relY = (y - config.minY) / (config.maxY - config.minY);
+      ctx.fillText(y , marginX *2 / 3,  marginY + graphHeight - relY * graphHeight + 4);
+
+      ctx.beginPath();
+      ctx.moveTo(marginX, marginY + graphHeight - relY * graphHeight );
+      ctx.lineTo(config.canvas.width - marginX, marginY + graphHeight - relY * graphHeight);
+      ctx.stroke();
+
+   };
+
+   ctx.font = "12px Arial";
+   ctx.textAlign = "center";
+   ctx.translate(marginX/3, marginY + graphHeight / 2);
+   ctx.rotate(-3.14 /2);
+   ctx.fillText(title, 0, 0);
+   ctx.rotate(3.14 /2)
+   ctx.translate(-marginX/3,-(marginY + graphHeight / 2));
+}
+
+
+//console.time();
+drawChart(dBChart);
+drawChart(bitsChart);
+drawChart(qLNChart);
+drawChart(hLogChart);
+//console.timeEnd();
--- a/modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js
+++ b/modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js
@@ -2,103 +2,71 @@
 // Rendering of DSL spectrum graphs  showing
 // US/DS SNR and US/DS bits/tone
 //
-// This version does not depend on chart.js or
-// any other package
+// This version does depend on an ubus version that support DSL line stattiscis  but 
+// does not depend on chart.js or any other package
 
- class DataSet {
-    constructor (inputString, extractFunction) {
-       this.groupSize = this.getGroupSize(inputString);
-       this.dataPos = this.getDataPos(inputString);
-       this.numData = this.getNumData(inputString);
-       if (!isNaN(this.groupSize)) {                   // this is the tricky part as output of all commands is not consistant
-          this.maxX = this.numData * this.groupSize;   // needs to be validated with various input
-       } else {
-          this.maxX = this.numData;
-          this.groupSize = 1
-       }
-       this.data = inputString.substr(this.dataPos).slice(0, -5).split(" ").map(extractFunction,{groupSize: this.groupSize});
+class DataSet {
+   constructor (input, extractFunction) {
+      this.groupSize = input.groupsize;
+      this.numData = input.groups;
+      this.maxX = this.numData * this.groupSize;   // needs to be validated with various input
+      this.data = input.data.map(extractFunction,{groupSize: this.groupSize});
    }
+}
 
-   getGroupSize(data) {
-      let groupSizePos = data.search('nGroupSize=') + 11
-      return (parseInt(data.substr(groupSizePos,2),10))
-   }
+function myBitsFunction(value, index, array) {
+   return({x: index, y: value, error: false});
+}
 
-   getDataPos(data) {
-      let dataPos = data.search('nData=') + 7
-      return (dataPos)
-   }
+function mySnrFunction(value, index, array) {
+   let result;
 
-   getNumData(data) {
-      let nNumDataPos = data.search('nNumData=') + 9
-      return (parseInt(data.substr(nNumDataPos,4),10))
-   }
+   if (value == 'NaN') {                                                                      // means error value
+      result = {x: index * this.groupSize, y: -40 , error: true}
+   } else {
+      result = {x: index * this.groupSize, y: value, error: false}
+   };
 
- }
- 
- const usSnrData   = new DataSet(document.getElementById("usdB").innerText, mySnrFunction);
- const dsSnrData   = new DataSet(document.getElementById("dsdB").innerText, mySnrFunction); 
- const usBitsData = new DataSet(document.getElementById("usBits").innerText, myBitsFunction);
- const dsBitsData = new DataSet(document.getElementById("dsBits").innerText, myBitsFunction);
- const usQLNData  = new DataSet(document.getElementById("usQln").innerText, myQLNFunction);
- const dsQLNData  = new DataSet(document.getElementById("dsQln").innerText, myQLNFunction);
- const usHLOGData = new DataSet(document.getElementById("usHLog").innerText, myHLOGFunction);
- const dsHLOGData = new DataSet(document.getElementById("dsHLog").innerText, myHLOGFunction);
+   return(result);
+}
 
+function myQLNFunction(value, index, array) {
+   let result;
 
- function myBitsFunction(value, index, array) {
-    return({x: parseInt(index), y: parseInt(value, 16), error: false});
- }
+   if (value == 'NaN') {                                                                 // means error value
+      result = { x: index * this.groupSize, y:  - 150, error: true}
+   } else {
+       result = { x: index * this.groupSize, y:  value, error: false}
+   };
 
- function mySnrFunction(value, index, array) {
-    let t = value.replace(/[\(\)\n]/g,'').split(',');
-    let result;
+   return(result);
+}
 
-    let v = t[1];
-    let i = t[0];
-    if (v == 255) {                                                                      // this is the only special value
-       result = {x: parseInt(i) * this.groupSize, y: -40 , error: true}
-    } else {
-       result = {x: parseInt(i) * this.groupSize, y: parseInt(- 32 + v / 2), error: false} // rec ITU-T G.933.2, section 11.4.1.1.3, SNR(k × G × Δf) = −32 + (snr(k)/2) dB
-    };
-    return(result);
- }
+function myHLOGFunction(value, index, array) {
+   let result;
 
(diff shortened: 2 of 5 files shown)


feckert:
I've had a look at it now. I think the feature is great. But I noticed that it doesn't look good with different themes. I have the darktheme from bootstrap installed.  But apart from that, I have always missed such a feature in OpenWrt.

Screenshot 2022-07-25 at 10-13-27 VR2-103954 - DSL line spectrum - LuCI

turboproc:
Below a screendump of how it looks like without any theme (or the default theme). Let me have a look how the graph can anticipate the theme being used. Valid comment, never considered this as I only use the defulat one.

image

turboproc:
@feckert would this be better ?

image

janh:
You need to properly handle scaling for the SNR data, as the group size isn't always 8. There is a maximum of 512 subcarrier groups, and the group size depends on how many subcarriers are actually in use. For ADSL the group size is always 1 (as there is only a maximum of 512 subcarriers), but for VDSL2 the group size can even be different for up- and downstream (for example on a line with profile 35b, the downstream group size is typically 16, while it is 8 for upstream data).

As `g997sansg` does not report the group size, it makes sense to use `g997dsnrg` instead (the first parameter of that command is the direction, the second is the data type and needs to be 1 for showtime data). If you implement support for the output of that command, it should also be easy to add support for QLN and Hlog using `g997dqlng` and `g997dhlogg`.

Here is data from a VDSL2 profile 35b line (from my own [standalone tool](https://github.com/janh/go-dsl)): [dsl_20220520_195501_raw.txt](https://github.com/openwrt/luci/files/9184282/dsl_20220520_195501_raw.txt)
If it would be helpful for testing, I can also provide data for other VDSL2 profiles (8a/b/c/d, 12a/b, 30a).

turboproc:
> You need to properly handle scaling for the SNR data, as the group size isn't always 8. There is a maximum of 512 subcarrier groups, and the group size depends on how many subcarriers are actually in use. For ADSL the group size is always 1 (as there is only a maximum of 512 subcarriers), but for VDSL2 the group size can even be different for up- and downstream (for example on a line with profile 35b, the downstream group size is typically 16, while it is 8 for upstream data).
> 
> As `g997sansg` does not report the group size, it makes sense to use `g997dsnrg` instead (the first parameter of that command is the direction, the second is the data type and needs to be 1 for showtime data). If you implement support for the output of that command, it should also be easy to add support for QLN and Hlog using `g997dqlng` and `g997dhlogg`.
> 
> Here is data from a VDSL2 profile 35b line (from my own [standalone tool](https://github.com/janh/go-dsl)): [dsl_20220520_195501_raw.txt](https://github.com/openwrt/luci/files/9184282/dsl_20220520_195501_raw.txt) If it would be helpful for testing, I can also provide data for other VDSL2 profiles (8a/b/c/d, 12a/b, 30a).

Hi @janh , thnx for your feedback. Let me cook on this for some days and I'll come back.

turboproc:
@janh , @feckert updated the commit to take care of

- better colors in case of dark-mode themes
- handling sub-carriers depending on their number (scaling)
- added graphs covering QLN and HLOG

Please check and share your results.

janh:
Scaling in x-direction looks good now: [vdsl2-35b.png](https://user-images.githubusercontent.com/949763/181818207-76d8b96c-86eb-4ae7-8016-964dc13cbe67.png)
In some cases, the x value range may not be the same for each graph and the graphs don't align (but it isn't wrong): [vdsl2-8b.png](https://user-images.githubusercontent.com/949763/181819040-5ccfaa8b-ecba-4d6a-ba75-817f5065c70d.png)

The scaling in y-direction for QLN and Hlog is broken, though. It should be `-23 - v/2` for QLN, and `6 - v/10` for Hlog.

Another thing I noticed is that small details may be hidden due to the way the graphs are drawn. In the profile 8b example, there are actually quite a few invalid SNR values around the range of subcarriers 128-256. However, the graph fails to show that. Here is the raw data: [dsl_20220729_185621_raw.txt](https://github.com/openwrt/luci/files/9222404/dsl_20220729_185621_raw.txt)

turboproc:
> Scaling in x-direction looks good now: [vdsl2-35b.png](https://user-images.githubusercontent.com/949763/181818207-76d8b96c-86eb-4ae7-8016-964dc13cbe67.png) In some cases, the x value range may not be the same for each graph and the graphs don't align (but it isn't wrong): [vdsl2-8b.png](https://user-images.githubusercontent.com/949763/181819040-5ccfaa8b-ecba-4d6a-ba75-817f5065c70d.png)
> 
> The scaling in y-direction for QLN and Hlog is broken, though. It should be `-23 - v/2` for QLN, and `6 - v/10` for Hlog.
> 
> Another thing I noticed is that small details may be hidden due to the way the graphs are drawn. In the profile 8b example, there are actually quite a few invalid SNR values around the range of subcarriers 128-256. However, the graph fails to show that. Here is the raw data: [dsl_20220729_185621_raw.txt](https://github.com/openwrt/luci/files/9222404/dsl_20220729_185621_raw.txt)

@janh, are you sure regarding the value translation? Reading the ITU recommendation:
```
7.10.9.3 Downstream QLN(f) (QLNpsds)
Description: Reports the downstream QLN(f) as one value per sub-carrier group. A first special value
(255) indicates that the QLN(f) for this sub-carrier group is undetermined. A second special value
(254) indicates that no measurement could be done for this sub-carrier group because it is out of the
downstream MEDLEY set. A third special value (0) indicates that the QLN(f) for this sub-carrier
group is greater than or equal to −35dBm/Hz. A fourth special value (251) indicates that the QLN(f)
for this sub-carrier group is less than or equal to −160.5 dBm/Hz. The object is updated only in L0
link state.
Type: Array of QLNdescriptor[0…N–1], where N = index of highest sub-carrier index in the
downstream MEDLEY set, divided by QLNGds, rounded to the higher integer.
QLNdescriptor:
Type: 8-bit unsigned integer.
Unit: −0.5 dBm/Hz.
Offset: −35 dBm/Hz.
Valid values: 1..250 (−35.5 to –160 dBm/Hz), 0, 251, 254, 255 (special values).
```

HLOG is indeed wrong, had to remove the minus sign:
```
7.10.7.3 Downstream HLOG(f) (HLOGpsds)
Description: Reports the downstream Hlog(f) as one value per sub-carrier group. A first special value
(1023) indicates that the Hlog(f) for this sub-carrier group is undetermined. A second special value
(1022) indicates that no measurement could be done for this sub-carrier group because it is out of the
MEDLEY set. A third special value (0) indicates that the Hlog(f) for this sub-carrier group is greater
than or equal to +6.0 dB. A fourth special value (1020) indicates that the Hlog(f) for this sub-carrier
group is less than or equal to −96.0 dB. The object is updated only in L0 link state.
Type: Array of HLOGdescriptor[0…N–1], where N = index of highest sub-carrier index in the
MEDLEY set, divided by HLOGG, rounded to the higher integer.
HLOGdescriptor:
Type: 16-bit unsigned integer.
Rec. ITU-T G.997.2 (03/2019) 57
Unit: −0.1 dB.
Offset: 6 dB.
Valid values: 1…1019 (+5.9 to –95.9 dB) and 0,1020,1022,1023 (special values).
Reference: Clause 11.4.1.2.1 of [ITU-T G.9701].
```

janh:
> @janh, are you sure regarding the value translation? Reading the ITU recommendation:
> 
> ```
> 7.10.9.3 Downstream QLN(f) (QLNpsds)
> ```

This seems to be from G.997.2, which applies to G.fast. For xDSL, the QLN value is encoded differently. If you look in [G.993.2 (VDSL2)](https://www.itu.int/rec/T-REC-G.993.2-201902-I/en) or [G.997.1 (Physical layer management for DSL transceivers)](https://www.itu.int/rec/T-REC-G.997.1-201902-I), you'll find that -23 is used as offset (the reserved values are also different for xDSL).

turboproc:
> > @janh, are you sure regarding the value translation? Reading the ITU recommendation:
> > ```
> > 7.10.9.3 Downstream QLN(f) (QLNpsds)
> > ```
> 
> This seems to be from G.997.2, which applies to G.fast. For xDSL, the QLN value is encoded differently. If you look in [G.993.2 (VDSL2)](https://www.itu.int/rec/T-REC-G.993.2-201902-I/en) or [G.997.1 (Physical layer management for DSL transceivers)](https://www.itu.int/rec/T-REC-G.997.1-201902-I), you'll find that -23 is used as offset (the reserved values are also different for xDSL).

Clear, wrong recommendation. Fixed.

Just need to decide on a way to represent the out of range values.

turboproc:
One way to represent the error values is by plotting a little red cross as in the example below. The cross is positioned are the defined minimum or maximum value depending on the error. Would that a sensible way forward ?
image

janh:
In my opinion there is no need to draw missing values in a special way. For xDSL there is only a single special value anyway, which doesn't tell why the value is not valid (and being outside the supported/used carrier set is the most likely reason).

I think fixing the issue I mentioned in my previous comment instead would be enough. It looks to me like the cause is that each data point is drawn using separate overlapping rectangles. Drawing the entire data as a single path should avoid this issse (the drawing code in my own tool does it that way).

turboproc:
Created a  new commit with rewritten drawing routing. Rather than drawing bar by bar, it first draws the outline and then fills the graph. This is done block by block .

janh:
Something went wrong with your latest commit, it contains conflict markers.

turboproc:
Mmh, guess these are the red lines in the json files. Really need some help how to solve this.

turboproc:
@janh , should all be fixed now.

nicefile:
Tested OK @ FB7362SL 
image

modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-spectrum.js
```diff
@@ -0,0 +1,85 @@
...
+'require view';
+'require fs';
+'require ui';
+
+
+return view.extend({
+        load: function() {
+
+                  return Promise.all([
+                      fs.exec_direct('/sbin/dsl_cpe_pipe.sh', ['g997dsnrg', '0 1']),
```
jow-, line 16: Just to make sure since I'm not familiar with `dsl_cpe_pipe.sh` - are the `0 1` really meant to be passed as *one* single argument or should it have been `['g997dsnrg', '0', '1']` ?

janh:
It doesn't work properly here, there is an exception when drawing downstream SNR for VDSL2 profile 35b (looks like the code assumes a group size of 8 there):
```
Looking for  65 859
Found  4 -1
Uncaught TypeError: this.dataPoints[endIndex] is undefined
    drawBlock …/luci-static/resources/view/status/dsl/dsl-graph-new.js:22
    drawBlocks …/luci-static/resources/view/status/dsl/dsl-graph-new.js:18
    drawChart …luci-static/resources/view/status/dsl/dsl-graph-new.js:17
    <anonymous> …/luci-static/resources/view/status/dsl/dsl-graph-new.js:37
```

I am not sure if drawing each band separately is a good idea at all. During showtime state, the `bbsg` command returns values for the MEDLEY set. However the QLN and Hlog data is available for the SUPPRTEDCARRIERS set, which is a superset of the MEDLEY set. So, a part of the data may actually be cut off in some cases. Even for SNR, you would need to use the limit values instead of the border values (the former is based subcarriers that are allowed to be used, the latter is based only on subcarriers that are currently used).

And in any case, the drawing code also needs to be able to handle invalid/zero values within a band, so I don't think drawing bands separately actually adds any value. If performance is a concern, it may help to omit redundant points from the path (i.e. when subsequent values are identical).

Other than that, I can confirm that invalid values are now drawn correctly. (It looks like the code still handles special values as defined for G.fast, though).

modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-spectrum.js
turboproc: @jow- new never seen an issue with this. `dsl_cpe_pipe.sh` would  throw an error in case of missing arguments.

janh:
All four graphs are now working here.

turboproc:
> All four graphs are now working here.

That's great news. Let's wait for some more review results. Won't touch this for a while until there is some feedback.

turboproc:
For those curious, this is how the charts look like:
image

nicefile:
after updating *.js graph look like below

image

Borromini:
@turboproc I'm not sure what's going wrong here, but git seems to suggest there's duplicate files in the PR somehow? I grabbed the plain patch (done by appending `.patch` to the PR URL). These are the errors git is throwing:

```
$ git am ../../../patches/luci/0001-luci-mod-dsl.patch
Applying: luci-mod-dsl: initial commit
.git/rebase-apply/patch:57: trailing whitespace.
 } 
.git/rebase-apply/patch:99: trailing whitespace.
   ] 
warning: 2 lines add whitespace errors.
Applying: luci-mod-dsl: initial commit
.git/rebase-apply/patch:81: trailing whitespace.
//console.time(); 
.git/rebase-apply/patch:83: trailing whitespace.
 const dsSnrData   = new DataSet(document.getElementById("dsdB").innerText, mySnrFunction); 
.git/rebase-apply/patch:174: trailing whitespace.
   ] 
.git/rebase-apply/patch:367: trailing whitespace.
      ctx.fillStyle = "#A0A0A0"; 
error: modules/luci-mod-dsl/Makefile: already exists in index
error: modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js: already exists in index
error: modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-spectrum.js: already exists in index
error: modules/luci-mod-dsl/root/usr/share/luci/menu.d/luci-mod-status-dsl.json: already exists in index
error: modules/luci-mod-dsl/root/usr/share/rpcd/acl.d/luci-mod-status-dsl.json: already exists in index
Patch failed at 0002 luci-mod-dsl: initial commit
hint: Use 'git am --show-current-patch=diff' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
```

When I inspect the plain patch, I see it creating `modules/luci-mod-dsl/Makefile` twice, but also other files (the JavaScript files differ in size between patch 1/7 and 2/7 in your patch set though). E.g. `dsl-graph-new.js` is being added once (patch 1/7, 201 lines of code) and then a second time (patch 2/7, 373 lines of code). I've uploaded the patch [here](https://paste.debian.net/1249117/). Result of `wget https://patch-diff.githubusercontent.com/raw/openwrt/luci/pull/5895.patch -O 0001-luci-mod-dsl.patch`.

turboproc:
@Borromini , not sure what happened. When checking the files in the `changed files` tab all files look OK.

modules/luci-mod-dsl/Makefile
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (C) 2022 Roland Barenbrug <roland@treslong.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI DSL spectrum graph
+LUCI_DEPENDS:=+luci-base
```
feckert, line 10: - We do not need the luci-base dependency here, we only use Javascript?
- We should add `@TARGET_lantiq` or `@TARGET_lantiq_xrx200` dependency, because this only works with this targets?
- We should consider to add dependency to the packages [ltq-vdsl-app](https://github.com/openwrt/openwrt/blob/master/package/network/config/ltq-vdsl-app/Makefile#L79) which is installing `dsl_cpe_pipe.sh`
- I have also seen that [ltq-adsl-app](https://github.com/openwrt/openwrt/blob/master/package/network/config/ltq-adsl-app/Makefile#L88) is installing a similar command. I wonder if this LuCI app does also works with this one.

feckert:
I have seen in the history, that you have a merge commit.
To fix your git issue you could do the following:

- `git reset HEAD~8` remove your last commits with out removing our changes
- `git add .` add your changes
- `git push --force` Make a push force to your pullrequest branch

feckert (changes requested):
It's a bit late, but shouldn't we consider using the Ubus to get the information we need? I am thinking of `checkmk` or `grafana` (without time series input) to push this information to these backends as well. Also the advantage would be that we have the whole thing as json. But I am not sure if this works? I have seen that the output of the command is large.

turboproc:
> It's a bit late, but shouldn't we consider using the Ubus to get the information we need? I am thinking of `checkmk` or `grafana` (without time series input) to push this information to these backends as well. Also the advantage would be that we have the whole thing as json. But I am not sure if this works? I have seen that the output of the command is large.

Have been looking into this when starting this topic. Concluded that in that case the `ubus` connector would require extension as the information required is not exposed yet by any ubus call. For now it works, indeed the data provided is not in JSON format but with some effort it was possible to parse all data.

modules/luci-mod-dsl/Makefile
turboproc: > * We do not need the luci-base dependency here, we only use Javascript?
> * We should add `@TARGET_lantiq` or `@TARGET_lantiq_xrx200` dependency, because this only works with this targets?
> * We should consider to add dependency to the packages [ltq-vdsl-app](https://github.com/openwrt/openwrt/blob/master/package/network/config/ltq-vdsl-app/Makefile#L79) which is installing `dsl_cpe_pipe.sh`
> * I have also seen that [ltq-adsl-app](https://github.com/openwrt/openwrt/blob/master/package/network/config/ltq-adsl-app/Makefile#L88) is installing a similar command. I wonder if this LuCI app does also works with this one.

Per item:

- not sure. the DSL charts do depend on a luci foundation being available. Without Luci being build it is not going the work
- on Lantiq, this make sense. It is all related to xDSL and that's only supported on Lantiq based platforms.
- correct
- should work as well. Personally not having access to ADSL hence  bit difficult to test.

On dependencies, just need to figure out how to put this in but should be able to manage this.

turboproc:
> I have seen in the history, that you have a merge commit. To fix your git issue you could do the following:
> 
> * `git reset HEAD~8` remove your last commits with out removing our changes
> * `git add .` add your changes
> * `git push --force` Make a push force to your pullrequest branch

Thnx for the suggestion. Will give it a try (just afraid of losing information with this kind of steps).

turboproc:
> @turboproc I'm not sure what's going wrong here, but git seems to suggest there's duplicate files in the PR somehow? I grabbed the plain patch (done by appending `.patch` to the PR URL). These are the errors git is throwing:
> 
> ```
> $ git am ../../../patches/luci/0001-luci-mod-dsl.patch
> Applying: luci-mod-dsl: initial commit
> .git/rebase-apply/patch:57: trailing whitespace.
>  } 
> .git/rebase-apply/patch:99: trailing whitespace.
>    ] 
> warning: 2 lines add whitespace errors.
> Applying: luci-mod-dsl: initial commit
> .git/rebase-apply/patch:81: trailing whitespace.
> //console.time(); 
> .git/rebase-apply/patch:83: trailing whitespace.
>  const dsSnrData   = new DataSet(document.getElementById("dsdB").innerText, mySnrFunction); 
> .git/rebase-apply/patch:174: trailing whitespace.
>    ] 
> .git/rebase-apply/patch:367: trailing whitespace.
>       ctx.fillStyle = "#A0A0A0"; 
> error: modules/luci-mod-dsl/Makefile: already exists in index
> error: modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js: already exists in index
> error: modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-spectrum.js: already exists in index
> error: modules/luci-mod-dsl/root/usr/share/luci/menu.d/luci-mod-status-dsl.json: already exists in index
> error: modules/luci-mod-dsl/root/usr/share/rpcd/acl.d/luci-mod-status-dsl.json: already exists in index
> Patch failed at 0002 luci-mod-dsl: initial commit
> hint: Use 'git am --show-current-patch=diff' to see the failed patch
> When you have resolved this problem, run "git am --continue".
> If you prefer to skip this patch, run "git am --skip" instead.
> ```
> 
> When I inspect the plain patch, I see it creating `modules/luci-mod-dsl/Makefile` twice, but also other files (the JavaScript files differ in size between patch 1/7 and 2/7 in your patch set though). E.g. `dsl-graph-new.js` is being added once (patch 1/7, 201 lines of code) and then a second time (patch 2/7, 373 lines of code). I've uploaded the patch [here](https://paste.debian.net/1249117/). Result of `wget https://patch-diff.githubusercontent.com/raw/openwrt/luci/pull/5895.patch -O 0001-luci-mod-dsl.patch`.

@Borromini, @feckert , cleaned the chain of commits and did a new commit. Should look better now. Please confirm.

modules/luci-mod-dsl/Makefile
turboproc: Makefile comment has been soved with https://github.com/openwrt/luci/pull/5895/commits/def03e6216f7e9727eacfdb7c5c64aa94a686f74

turboproc:
@feckert, on the `ubus` topic something like
```
root@dsl-02:/tmp# ubus -v list dsl
'dsl' @88f73b6b
        "metrics":{}
        "statistics":{}
```
and
```
root@dsl-02:/tmp# ubus call dsl statistics
{
        "bits": {
                "downstream": {
                        "direction": 1,
                        "groups": 4096,
                        "data": [
                                0,
                                0,
                                0,
                                0,
                                0,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                15,
                                0,
                                0,
                                0,
                                0
                        ]
                },
                "upstream": {
                        "direction": 0,
                        "groups": 4096,
                        "data": [
                                0,
                                0,
                                0,
                                0,
                                0,
                                0,
                                0,
                                6,
                                8,
                                9,
                                10,
                                12,
                                12,
                                0,
                                0,
                                0,
                                0
                        ]
                }
        },
        "snr": {
                "downstream": {
                        "direction": 1,
                        "groupsize": 8,
                        "groups": 512,
                        "data": [
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                149,
                                153,
                                155,
                                157,
.....
                                255,
                                255,
                                255,
                                255,
                                255
                        ]
                },
                "upstream": {
                        "direction": 0,
                        "groupsize": 8,
                        "groups": 512,
                        "data": [
                                64,
                                130,
                                163,
                                165,
                                163,
                                167,
                                165,
                                147,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                64,
                                164,
                                165,
                                168,
                                168,
                                166,
                                167,
                                166,
                                166,
                                166,
                                167,
                                166,
                                165,
                                165,
                                165,
                                164,
                                166,
                                166,
                                164,
                                165,
                                165,
                                164,
                                164,
                                165,
                                164,
                                163,
                                163,
                                164,
                                163,
                                164,
                                163,
                                163,
                                163,
                                162,
                                163,
                                162,
                                162,
                                161,
                                161,
                                161,
                                160,
                                64,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                64,
                                149,
                                150,
                                149,
                                149,
                                149,
                                149,
                                148,
                                149,
                                140,
                                148,
                                147,
                                148,
                                147,
                                147,
                                147,
                                147,
                                147,
                                146,
                                146,
                                137,
                                142,
                                146,
                                146,
                                145,
                                145,
                                145,
                                145,
                                143,
                                144,
                                144,
                                143,
                                143,
                                143,
                                143,
                                143,
                                141,
                                142,
                                142,
                                142,
                                142,
                                141,
                                141,
                                141,
                                141,
                                140,
                                140,
                                140,
                                139,
                                139,
                                140,
                                138,
                                139,
                                139,
                                138,
                                138,
                                138,
                                137,
                                137,
                                138,
                                138,
                                138,
                                137,
                                137,
                                137,
                                136,
                                136,
                                136,
                                136,
                                136,
                                136,
                                136,
                                136,
                                135,
                                135,
                                135,
                                134,
                                134,
                                134,
                                133,
                                134,
                                133,
                                133,
                                132,
                                132,
                                132,
                                131,
                                132,
                                132,
                                131,
                                131,
                                130,
                                130,
                                130,
                                130,
                                129,
                                130,
                                129,
                                130,
                                129,
                                64,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255
                        ]
                }
        },
        "qln": {
                "downstream": {
                        "direction": 1,
                        "groupsize": 8,
                        "groups": 512,
                        "data": [
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
....
                                255,
                                255,
                                255,
                                255,
                                255,
                                255,
                                255
                        ]
                },
                "upstream": {
                        "direction": 0,
                        "groupsize": 8,
                        "groups": 512,
                        "data": [
                                255,
                                184,
                                184,
                                179,
                                171,
                                180,
                                185,
                                187,
                                255,
                                255,
....
                                255,
                                255,
                                255,
                                255
                        ]
                }
        },
        "hlog": {
                "downstream": {
                        "direction": 1,
                        "groupsize": 8,
                        "groups": 512,
                        "data": [
                                1023,
                                1023,
                                1023,
....
                                1023,
                                1023,
                                1023
                        ]
                },
                "upstream": {
                        "direction": 0,
                        "groupsize": 8,
                        "groups": 512,
                        "data": [
                                1023,
                                156,
                                90,
                                87,
                                91,
                                93,
......
                                1023,
                                1023,
                                1023
                        ]
                }
        }
}
root@dsl-02:/tmp#
```

Still requires some tweaking but the foundation should be OK.

Borromini:
@turboproc Thanks, applies like expected now. Very interested myself, but I have no time to swap out my main router (which is on 22.03) for one with a master build atm :(.

turboproc:
> It's a bit late, but shouldn't we consider using the Ubus to get the information we need? I am thinking of `checkmk` or `grafana` (without time series input) to push this information to these backends as well. Also the advantage would be that we have the whole thing as json. But I am not sure if this works? I have seen that the output of the command is large.

Did some work on this. Pleasr check out https://github.com/openwrt/openwrt/pull/10420. The ubus part has been done. The luci part using ubus is almost ready for commit.

feckert:
@turboproc Thank four your great work on this. I will test the `ubus` integrration in one go when you are ready with the LuCI.

I have now tested the last LuCI status. The look and feel for the dark them looks good now.
But somehow I think something is wrong. The graphics look strange.
I will test the `ubus` integration with my setup when you have updated the LuCI.

turboproc:
> @turboproc Thank for your great work on this. I will test the `ubus` integration in one go when you are ready with the LuCI.
> 
> I have now tested the last LuCI status. The look and feel for the dark them looks good now. But somehow I think something is wrong. The graphics look strange. I will test the `ubus` integration with my setup when you have updated the LuCI. Screenshot 2022-08-09 at 08-46-23 VR2-103954 - DSL line spectrum - LuCI

Indeed a kind of weird. Will commit the Luci DSL module that uses dbus a.s.a.p. Curious whether that gives the same results (hopefully not).

turboproc:
@feckert, any better updates by using the new `ubus` commit combined with the latest `luci-mod-dsl` commit ?

feckert:
I have now tested this again with the ubus backend.
The ubus works as it should.
Only now the graph looks the same as before!
I have uploaded my dsl json to https://pastebin.com/jfcRD7fu
That is the graph I get with this json data.
Screenshot 2022-08-19 at 12-54-57 VR2-103954 - DSL line spectrum - LuCI
.q

turboproc:
> I have now tested this again with the ubus backend. The ubus works as it should. Only now the graph looks the same as before! I have uploaded my dsl json to https://pastebin.com/jfcRD7fu That is the graph I get with this json data. Screenshot 2022-08-19 at 12-54-57 VR2-103954 - DSL line spectrum - LuCI .q

OK, observed already one issue: in your case US and DS numbr of groups is different (512 versus 64). Would be helpful to understand the profile your DSL modem is using (ADSL/VDSL, etc, ...)

janh:
> OK, observed already one issue: in your case US and DS numbr of groups is different (512 versus 64).

I don't think that's actually an issue. Only the first 64 sub-carriers are used for upstream (looks like ADSL2+), so any additional values should be invalid anyway.

To me it looks like there are a few other issues though.

- For downstream Hlog, only the first 256 groups seem to contain valid values, the rest looks like garbage. The actual group size seems to be 2.
- Downstream QLN extends to the upstream range and contains weird values there.
- For Hlog and QLN it looks like a raw value of 0 is reported for invalid value instead of the proper 255/1024.
- SNR seems to reports 0 (raw value 64) for some invalid values (start of upstream band).

These all look to me like the data returned by the firmware/driver is already wrong. Maybe the raw data would be useful here to confirm that.

For the first two issues it probably isn't worth it doing anything about that (except for making sure the drawing code doesn't draw outside the graph boundaries). The Hlog issue may be fixable by some heuristic, but it is not clear if this would break more than it fixes (this could just be a bug in a specific firmware version).

The third issue could be worked around by also treating 0 as an invalid value (unlikely to break anything, as values on the edge of the value range are probably extremely rare anyway).

For the fourth issue, I think the best options would be either just leaving it like that, or reducing the value range of the graph to start at 0 (I guess the question is how useful negative SNR values are, although there are actually a few in the data here).

turboproc:
> > OK, observed already one issue: in your case US and DS numbr of groups is different (512 versus 64).
> 
> I don't think that's actually an issue. Only the first 64 sub-carriers are used for upstream (looks like ADSL2+), so any additional values should be invalid anyway.
> 
> To me it looks like there are a few other issues though.
> 
> * For downstream Hlog, only the first 256 groups seem to contain valid values, the rest looks like garbage. The actual group size seems to be 2.
> * Downstream QLN extends to the upstream range and contains weird values there.
> * For Hlog and QLN it looks like a raw value of 0 is reported for invalid value instead of the proper 255/1024.
> * SNR seems to reports 0 (raw value 64) for some invalid values (start of upstream band).
> 
> These all look to me like the data returned by the firmware/driver is already wrong. Maybe the raw data would be useful here to confirm that.
> 
> For the first two issues it probably isn't worth it doing anything about that (except for making sure the drawing code doesn't draw outside the graph boundaries). The Hlog issue may be fixable by some heuristic, but it is not clear if this would break more than it fixes (this could just be a bug in a specific firmware version).
> 
> The third issue could be worked around by also treating 0 as an invalid value (unlikely to break anything, as values on the edge of the value range are probably extremely rare anyway).
> 
> For the fourth issue, I think the best options would be either just leaving it like that, or reducing the value range of the graph to start at 0 (I guess the question is how useful negative SNR values are, although there are actually a few in the data here).

Hi @janh, thnx for stepping in. As you might have seen, changed my conclusion already that the 64 / 256 carrier groups are probably OK in the case of a ADSL2+ profile.

Regarding the US part of HLOG and QLN, could it be the case that these measurements do not exist at all in combination with ADSL/ADSL2+, i.e. VDSL only? In that case it might be better to bring some logic that filters that data based on the profile or standard actually used.

modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js
```diff
@@ -0,0 +1,347 @@
...
+}
+
+function myBitsFunction(value, index, array) {
+   return({x: index, y: value, error: false});
+}
+
+function mySnrFunction(value, index, array) {
+   let result;
+
+   if (value == 'NaN') {                                                                      // means error value
```
dhewg, line 24: As commented on the other PR, I meant NAN from math.h, not the string representation. That hopefully makes this redundant?

janh:
> Regarding the US part of HLOG and QLN, could it be the case that these measurements do not exist at all in combination with ADSL/ADSL2+, i.e. VDSL only? In that case it might be better to bring some logic that filters that data based on the profile or standard actually used.

Support for SNR, Hlog and QLN was added in ADSL2 and VDSL2, and it was always designed to allow access to both up- and downstream data from both ends of the line. In the original ADSL and VDSL standards these three test parameters weren't available at all (but both should be practically irrelevant today). Grouping of data was added only in VDSL2, so for ADSL2 and ADSL2+ the data should always be reported for all subcarriers.

This means that special casing for different xDSL standards doesn't make sense (at least not because of the standards itself).

I have seen similar issues (i.e. missing upstream or downstream data) before on non-OpenWrt Lantiq-based modems, with actual results depending on the used DSL firmware version. So it is possible these are just driver/firmware/interoperability issues, and we probably have to accept that the data returned from the modem may not always be perfect.

turboproc:
> > Regarding the US part of HLOG and QLN, could it be the case that these measurements do not exist at all in combination with ADSL/ADSL2+, i.e. VDSL only? In that case it might be better to bring some logic that filters that data based on the profile or standard actually used.
> 
> Support for SNR, Hlog and QLN was added in ADSL2 and VDSL2, and it was always designed to allow access to both up- and downstream data from both ends of the line. In the original ADSL and VDSL standards these three test parameters weren't available at all (but both should be practically irrelevant today). Grouping of data was added only in VDSL2, so for ADSL2 and ADSL2+ the data should always be reported for all subcarriers.
> 
> This means that special casing for different xDSL standards doesn't make sense (at least not because of the standards itself).
> 
> I have seen similar issues (i.e. missing upstream or downstream data) before on non-OpenWrt Lantiq-based modems, with actual results depending on the used DSL firmware version. So it is likely these are just driver/firmware/interoperability issues, and we probably have to accept that the data returned from the modem may not always be perfect.

Clear explanation. Let me just introduce the additional exceptional values as mentioned above.

modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/dsl-graph-new.js
turboproc: Sorry, missed this one. Let me look into this.

turboproc:
> I have now tested this again with the ubus backend. The ubus works as it should. Only now the graph looks the same as before! I have uploaded my dsl json to https://pastebin.com/jfcRD7fu That is the graph I get with this json data. Screenshot 2022-08-19 at 12-54-57 VR2-103954 - DSL line spectrum - LuCI .q

@feckert , with the latest commit of ltq-vdsl-app (https://github.com/openwrt/openwrt/pull/10420/commits/c47cfa11a7c7fb461bf72e551a583fb14c121e92) your graphs should look a lot more decent.

BTW: could you share the Lantiq firmware version that you run? Would be helpful to understand how well various versions behave with respect to correctly providing statistics data.

feckert:
Thats my output from `ubus call dsl metrics`
```
root@VR2-103954 ~ # ubus call dsl metrics
{
        "api_version": "4.17.18.7",
        "firmware_version": "5.7.4.12.0.2",
        "chipset": "Lantiq-VRX200",
        "driver_version": "1.5.17.7",
        "state": "Showtime with TC-Layer sync",
        "state_num": 7,
        "up": true,
        "uptime": 152,
        "atu_c": {
                "vendor_id": [
                        181,
                        0,
                        66,
                        68,
                        67,
                        77,
                        164,
                        96
                ],
                "vendor": "Broadcom 164.96",
                "system_vendor_id": [
                        0,
                        0,
                        48,
                        48,
                        48,
                        48,
                        0,
                        0
                ],
                "version": [
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0,
                        0
                ],
                "serial": [
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48,
                        48
                ]
        },
        "power_state": "L0 - Synchronized",
        "power_state_num": 0,
        "xtse": [
                0,
                0,
                0,
                0,
                0,
                0,
                1,
                0
        ],
        "annex": "J",
        "standard": "G.992.5",
        "mode": "G.992.5 (ADSL2+)",
        "upstream": {
                "trellis": true,
                "bitswap": true,
                "retx": false,
                "virtual_noise": false,
                "interleave_delay": 4000,
                "data_rate": 2260000,
                "latn": 1.700000,
                "satn": 0.800000,
                "snr": 7.700000,
                "actps": -41.000000,
                "actatp": 12.800000,
                "attndr": 2251000
        },
        "downstream": {
                "trellis": true,
                "bitswap": false,
                "retx": false,
                "virtual_noise": false,
                "interleave_delay": 7000,
                "data_rate": 5330000,
                "latn": 1.300000,
                "satn": 0.800000,
                "snr": 6.200000,
                "actps": -57.300000,
                "actatp": 4.500000,
                "attndr": 5316000
        },
        "errors": {
                "near": {
                        "es": 0,
                        "ses": 0,
                        "loss": 0,
                        "uas": 50,
                        "lofs": 0,
                        "fecs": 0,
                        "hec": 0,
                        "ibe": 362840426,
                        "crc_p": 0,
                        "crcp_p": 0,
                        "cv_p": 0,
                        "cvp_p": 0
                },
                "far": {
                        "es": 1320,
                        "ses": 1163,
                        "loss": 1120,
                        "uas": 50,
                        "lofs": 0,
                        "fecs": 697409,
                        "hec": 4857,
                        "ibe": 340761,
                        "crc_p": 0,
                        "crcp_p": 0,
                        "cv_p": 0,
                        "cvp_p": 0,
                        "rx_corrupted": 0,
                        "rx_uncorrected_protected": 0,
                        "rx_retransmitted": 0,
                        "rx_corrected": 0,
                        "tx_retransmitted": 0
                }
        }
}
```
And this is spectrum for this setup:
Screenshot 2022-08-31 at 16-52-30 VR2-103954 - DSL line spectrum - LuCI

turboproc:
@feckert this starts to look better already (confirming the ADSL2+ profile in the output) now running into the limitations of the firmware version (5.7.4.12.0.2). Would you be able to change to a different firmware version (5.9.something) and check whether the results are the same? We could improve the HLOG graphs but that would essentially mean we're fixing some broken output and it might uninentially impact proper data. Preference would be to use simply better data. That might be provided by another firmware version.

feckert:
> @feckert this starts to look better already (confirming the ADSL2+ profile in the output) now running into the limitations of the firmware version (5.7.4.12.0.2).

I seem to remember that this two data types are only valid to VDSL. If you also see it that way with ADSL2/ADSL2+ for HLOG and QLN data, then we can only display the data or output it via the `ubus` if we are in VDSL mode. 

> Would you be able to change to a different firmware version (5.9.something) and check whether the results are the same?

I'll have to see how I can update it. Unfortunately, I can't do it that quickly.

> We could improve the HLOG graphs but that would essentially mean we're fixing some broken output and it might uninentially impact proper data. Preference would be to use simply better data. That might be provided by another firmware version.

I agree that we should not fix the DSL firmware expenditure because we do not know what is right or wrong. That must be done by the DSL firmware.

modules/luci-mod-dsl/Makefile
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (C) 2022 Roland Barenbrug <roland@treslong.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LUCI DSL spectrum graph
+LUCI_DEPENDS:=@TARGET_lantiq_xrx200 +luci-base +ltq-dsl-base
```
dhewg, line 10: In the meantime there's https://github.com/openwrt/openwrt/commit/1afd0fefd22992b808e98fcf726a8e4d3af91604 , so please use:
```suggestion
LUCI_DEPENDS:=+luci-base ltq-dsl-app
```

modules/luci-mod-dsl/root/usr/share/luci/menu.d/luci-mod-status-dsl.json
```diff
@@ -0,0 +1,14 @@
...
+         "admin/status/dsl-spectrum": {
+                 "title": "DSL line spectrum",
+                 "order": 7,
+                 "action": {
+                         "type": "view",
+                         "path": "status/dsl/dsl-spectrum"
+                 },
+                 "depends": {
+                         "acl": [ "luci-mod-status-dsl" ],
+                         "uci": { "network": { "@dsl": true } }
```
dhewg, line 11: You probably meant:
```suggestion
                         "uci": { "network": { "dsl": true } }
```
At least this fixes it for me on the unmerged VR11 support

modules/luci-mod-dsl/root/usr/share/rpcd/acl.d/luci-mod-status-dsl.json
```diff
@@ -0,0 +1,30 @@
...
+                                "/sbin/dsl_cpe_pipe.sh g997dqlng 0 1": [ "exec" ],
+                                "/sbin/dsl_cpe_pipe.sh g997dqlng 1 1": [ "exec" ],
+                                "/sbin/dsl_cpe_pipe.sh g997dhlogg 0 1": [ "exec" ],
+                                "/sbin/dsl_cpe_pipe.sh g997dhlogg 1 1": [ "exec" ],
+                                "/sbin/dsl_cpe_pipe.sh bbsg 0": [ "exec" ],
+                                "/sbin/dsl_cpe_pipe.sh bbsg 1": [ "exec" ]
+                        }
+                }
+        },
+        "luci-mod-status-index-dsl": {
```
dhewg, line 20: Leftover
```suggestion
```

dhewg (changes requested):
With the suggested changes above I get this on ipq40xx/v11: 
dsl
Does that look sane?

systemcrash:
This is absolutely bangin'. Get it in. With the fixes landing in openwrt for VRX300 series DSL chipsets, platforms like Fritzbox 7520/30 are becoming available. These devices have a large penetration in the UK, and DE (and potentially other) markets.

jow-:
@turboproc - Is this PR ready to merge? Are any required changes in OpenWrt base / OpenWrt packages merged?

turboproc:
> @turboproc - Is this PR ready to merge? Are any required changes in OpenWrt base / OpenWrt packages merged?

Hi @jow, thought this has been comitted already. The only action to take care of is that the relevant Luci package needs to be committed as well, otherwise nothing will happen.

dhewg:
To get this over the finishing line:
[This commit](https://github.com/dhewg/luci/commit/e800a947f637178fe2ecb33cc8932893afd17839) contains the cleanups I mentioned above. Works for me on ipq40xx/vr11.
If this works on lantiq/vr9 can we agree to merge that, please? I don't see a reason why this should bitrot here...

@feckert do you mind giving that commit a test drive, please - combined with [this branch](https://github.com/dhewg/openwrt/commits/dsl-ubus-stats) as mentioned in the base PR?

nicefile:
@dhewg I've build and run your branches on FB7412 vr9 and graph works . Black one below is to compare with  1st edition commented above 
new
old

asdas127816:
I'm new to openwrt. I want these charts but couldn't see any instructions on how to use them. Please explain.

feckert:
Thanks for your work.
Pushed with https://github.com/openwrt/luci/commit/cd48d766add74df60b6c6a882defa859b311ec97 directly. Made some changes see commit Message

Issue #5845, opened 2022-06-21
Allow manually setting MAC/BSSID on each wireless interface (client, AP, etc) [Enhancement]


sirfail:
I'm not sure why this isn't implemented in OpenWrt yet since there is demand for it and it has been available for years in Linux Desktops. A quick search shows OpenWrt users are struggling while trying to change the MAC/BSSID address on wireless interfaces, for example in https://www.reddit.com/r/openwrt/comments/9q0ezo/mac_address_spoofing_with_openwrt/ you see the recommendation to set "option macaddr" in file /etc/config/wireless, but the documentation in https://openwrt.org/docs/guide-user/network/wifi/basic says this is not used to change the device mac but to identify the underlying interface.

There is also a script to set a random MAC in https://github.com/krabelize/openwrt-random-mac-changer that uses ifconfig but it does not work. Seems interest to set and change MAC/BSSID addresses is there, but there is no easy and reliable way to do it.

In Issue https://github.com/openwrt/luci/issues/2613 "A wireless device cannot have it's mac address set from luci (option macaddr)" the comment from 7. March 2019 says:

 "OpenWrt eventually introduced "config device" sections to set things like macaddr, queue lengths etc. per netdev. LuCI currently does not support adding or modifying "config device" sections at all. Support for this needs to be implemented first."

The Issue 2613 is closed but as far as I can see this was never implemented.

Issue https://github.com/openwrt/luci/issues/5517 "Button to generate MAC" is still open and would be similar to this one, likely the first step is to allow entering a MAC/BSSID manually and then generating a random one would be the next step. And an option to change automatically to a random MAC/BSSID every day or n hours would be great too. 

But to get started it would be already great if the MAC/BSSID could be set manually on every wireless interface (client, AP mode etc) same you can set a static IP address. 

PS. The only "reliable" way to change the MAC/BSSID so far is using "uci -q set wireless.@wifi-iface[-1].macaddr=$mac_addr" as explained in https://forum.archive.openwrt.org/viewtopic.php?id=65751 comment 20 Jun 2016, 23:30. It's not persistent, I have to figure out if I can run it from rc.local or a cron job. Changing it with ifconfig or in  /etc/config/wireless didn't work.

Neustradamus:
Linked to:
- https://github.com/openwrt/luci/issues/5517
- https://github.com/openwrt/luci/issues/5819
- https://github.com/openwrt/luci/issues/5845
- https://github.com/openwrt/luci/pull/5896
[...]

Issue #5819, opened 2022-05-20
Feature Request: enable random mac address function in LuCI


Che15ea:
Feature Request:
I think we need a random/stable/specify Mac address spoofing function in LuCI as wifi/ethernet server/client to protect against mac address tracking. The feature is already built in most linux distros.

jow-:
Duplicate of #5819

jow-:
Correction. Duplicate of #5517

Neustradamus:
Linked to:
- https://github.com/openwrt/luci/issues/5517
- https://github.com/openwrt/luci/issues/5819
- https://github.com/openwrt/luci/issues/5845
- https://github.com/openwrt/luci/pull/5896
- https://github.com/openwrt/luci/issues/7131

PR #5896, opened 2022-07-25
luci-mod-network: wireless.js: add wifi-iface macaddr random support

msvamp:
Add ability to choose `random` or custom `macaddr` for wifi-iface

wifi-iface macaddr option - BSSID visible
wifi-iface macaddr option - BSSID invisible
(**Updated** screenshot as per latest commit: https://github.com/openwrt/luci/pull/5896/commits/ab73d459ca6a57ba0e5b4898c8eb3dbdc46f252a)

Choosing the new *randomly generated* MAC address option will cause a new MAC address to be generated whenever the WiFi interface is (re-)configured or the system is rebooted.

This will work only if wifi-iface supports `option macaddr 'random'`.
Depends on https://github.com/openwrt/openwrt/pull/10307

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1157,10 +1157,11 @@ return view.extend({
 					if (/^radio\d+\.network/.test(o.placeholder))
 						o.placeholder = '';
 
+					var macaddr = uci.get('wireless', radioNet.getName(), 'macaddr');
 					o = ss.taboption('advanced', form.Value, 'macaddr', _('MAC address'), _('Override default MAC address - the range of usable addresses might be limited by the driver'));
-					o.optional = true;
-					o.placeholder = radioNet.getActiveBSSID();
-					o.datatype = 'macaddr';
+					o.value('', _('driver default (%s)').format(!macaddr ? radioNet.getActiveBSSID() : _('no override')));
+					o.value('random', _('randomly generated'));
+					o.datatype = "or('random',macaddr)";
 
 					o = ss.taboption('advanced', form.Flag, 'short_preamble', _('Short Preamble'));
 					o.default = o.enabled;


Ansuel:
Why not add a button to generate a random Mac? Instead of a select?

Il Lun 25 Lug 2022, 08:00 Manas Sambhus ***@***.***> ha
scritto:

msvamp:
> Why not add a button to generate a random Mac?

@Ansuel The purpose of adding a `macaddr random` option to `wifi-iface` was to ensure a **new** random MAC address is generated everytime the interface (or the router itself) is restarted. Please have a look at https://github.com/openwrt/openwrt/pull/10307.

Ansuel:
Oh ok sorry for the noise then.

Il Lun 25 Lug 2022, 08:16 Manas Sambhus ***@***.***> ha
scritto:

Ansuel:
Wonder if you should add some reference about this in the commit description. Anyway good work.

msvamp:
@Ansuel Updated commit and PR description

Che15ea:
Is it possible to allow random/custom mac enabled by default? I mean to make the Openwrt router forget about the original factory MAC address at all, and never expose the original factory MAC address anytime anywhere, in order to prevent accidental leakage(during first time setup for example). Broadcasting real MAC address is just too...big brother.
I suggest to allow to choose between Permanent/Temporary/Factory MAC and manual regenerate for the button design.

jow-:
No, it will most likely not become the default as it would break existing setups relying on specific MACs for DHCP lease reservations etc.

msvamp:
@Che15ea I agree with @jow-, most people would prefer to use the router's default MAC address wherever possible, and opt in to using a randomised MAC only for some specific interfaces. Some have a network setup relying on MAC address for address assignment, access control etc. and also, there are some ISPs that identify a customer's device by MAC address and block access otherwise - hence using a randomized MAC by default would be troublesome for them.

---

> Is it possible to allow random/custom mac enabled by default? I mean to make the Openwrt router forget about the original factory MAC address at all

If you need such a firmware, you can easily build one for yourself with the [OpenWrt Imagebuilder](https://openwrt.org/docs/guide-user/additional-software/imagebuilder). On most platforms, the `/etc/board.d/xx_network` script is responsible for configuring the correct MAC address for the router by reading it from some flash partition or firmware folders in the kernel. You can replace this file with your own, that calls the [`macaddr_random()` function](https://github.com/openwrt/openwrt/blob/6e13794/package/base-files/files/lib/functions/system.sh#L192:L196) in place of the correct procedure defined for your device. Note that in most cases, this would also alter your ethernet MAC address, so be careful.

> in order to prevent accidental leakage(during first time setup for example)

If referring to just wireless, I don't think that the original MAC address is leaked during the initial setup, because the wireless interfaces are disabled by default. You should be able to customise the settings and enable the interfaces only once done.

> I suggest to allow to choose between Permanent/Temporary/Factory MAC and manual regenerate for the button design.

Thank you for the suggestion. I'll consider adding a button to generate a new random MAC address. Currently, with the new _randomly generated_ option that I've added, the MAC address is randomised whenever the WiFi interface is (re-)configured, so I believe that using the restart button in LuCI for that interface, should also create a new MAC address for it.

aparcar:
I merged https://github.com/openwrt/openwrt/pull/10307 and think this PR is fine, @jow- ?

msvamp:
I hope the 2 options in the select list are clear enough and are not misinterpreted 🙃

Also, do we need a custom validation message?

HubKing:
Is this "Network -> Interfaces -> Devices -> Configure -> Mac address", and going to be applied to all device types, not just Wi-Fi? I want to randomise the MAC address of my "WAN" device, so that my ISP would give a random IP, because the ISP always gives the same IP for the same MAC address. 

If I have set "randomly generated" to the Mac address of my WAN device, and I click "Network -> Interfaces -> WAN-> Restart", my WAN device's Mac address will be changed to a random address without restarting the whole router, right?

msvamp:
@HubKing The change that I implemented was only for MAC randomisation for Wi-Fi interfaces, not Ethernet or other kinds of interfaces.

If your WAN connection is over a wireless interface, then MAC randomisation will surely work for it. You can change the MAC address by restarting the WiFi interface, without having to restart the router.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1157,9 +1157,9 @@ return view.extend({
 						o.placeholder = '';
 
 					o = ss.taboption('advanced', form.Value, 'macaddr', _('MAC address'), _('Override default MAC address - the range of usable addresses might be limited by the driver'));
-					o.optional = true;
-					o.placeholder = radioNet.getActiveBSSID();
-					o.datatype = 'macaddr';
+					o.value('', _('automatic (no override)'));
```
aparcar, line 1160:
```suggestion
					o.value('', _('automatic (%s)').format(radioNet.getActiveBSSID()));
```

I'd prefer that you still see the address.

aparcar (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
msvamp: I think this will display the randomly generated MAC address there if the current setting is set to random? It will be misleading.

msvamp: Do you know of a way to read the next available/assign-able MAC address for the actual WiFi interface from `sysfs` or some other place as appropriate?

HubKing:
> @HubKing The change that I implemented was only for MAC randomisation for Wi-Fi interfaces, not Ethernet or other kinds of interfaces.
> 
> If your WAN connection is over a wireless interface, then MAC randomisation will surely work for it. You can change the MAC address by restarting the WiFi interface, without having to restart the router.

No, my WAN is not Wi-Fi; like most people, it is wired connection to a broadband modem from my ISP. Can't the same random feature applies to Ethernet (wired) MAC address, too?

msvamp:
> Can't the same random feature applies to Ethernet (wired) MAC address, too?

@HubKing For this, one will have to look at the corresponding scripts for ethernet device setup and add the change there (just like I did for [wireless interfaces](https://github.com/openwrt/openwrt/pull/10307)).

This is a good idea though, I'll try to implement this when I get some time (probably in late December).

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
msvamp: Implemented in https://github.com/openwrt/luci/pull/5896/commits/8cb87cbbf7dd4ff48a09d2e3d248cd7db6c21266

msvamp:
@aparcar I have implemented your suggestion about keeping the MAC address visible, but only if it is not randomised.
@jow- Are we good to merge now?

msvamp:
Ping. 🙂
@Ansuel @aparcar @jow-

Ansuel:
@msvamp can you update the screen with new implementation?

msvamp:
> @msvamp can you update the screen with new implementation?

@Ansuel Edited my [first comment](https://github.com/openwrt/luci/pull/5896#issue-1316311436) to add new screenshots

Neustradamus:
Linked to:
- https://github.com/openwrt/luci/issues/5517
- https://github.com/openwrt/luci/issues/5819
- https://github.com/openwrt/luci/issues/5845
- https://github.com/openwrt/luci/pull/5896
- https://github.com/openwrt/luci/issues/7131

PR #5903, opened 2022-07-28
luci-proto-ncm: adds missing error messages, removes 'dialnumber' opt…

Jalakas:
…ion, changes 'delay' according to real behavior

This commit:
Adds missing error messages and harmonizes message texts with comgt-ncm script. [1]
Removes option 'dialnumber' not used in comgt-ncm. [1][2]
Fixes option 'delay' according to comgt-ncm script real behavior. (No default value, if parameter is set, then there's always sleep) [1]

[1] https://github.com/openwrt/openwrt/blob/master/package/network/utils/comgt/files/ncm.sh
[2] https://github.com/openwrt/openwrt/blob/master/package/network/utils/comgt/files/ncm.json

Signed-off-by: Anari Jalakas <anari.jalakas@gmail.com>


jow-:
Merged, thanks!

PR #5905, opened 2022-07-29
luci-mod-dashboard: show IPv6 connection uptime

ldir-EDB0:
Show connection uptime for both IPv4 & IPv6 interfaces.
Rename display field from 'Connected since' to 'Uptime' as this makes
more sense to me.  You could say 'Connected for' or 'Connection
duration'.  After a lot of thought decided that 'Uptime' is what I'm
most comfortable saying.

Signed-off-by: Kevin Darbyshire-Bryant <ldir@darbyshire-bryant.me.uk>


jow-:
Merged, thank you!

PR #5912, opened 2022-07-31
luci-app-example: improve description

stokito:
I spent for some time while figured out how to start the example app. Please accept the PR that improves README


stokito:
@andibraeu could you please review and merge this? The small change may save a lot of time for newcomers

andibraeu:
I don't have rights to merge PRs here.

One note: you don't need to execute the file in `uci-defaults`. I think after a reboot it should be fine, too.

jow-:
That extra ssh chmod command is not needed since `uci-defaults` scripts are sourced, not executed.

stokito:
ok, can you merge and remove it? because I don't have a time now to fix this

jow-:
Unfortunately I also only had a few minutes to spare to glance over PRs. I am not on a machine right now where I can do Git operations.

stokito:
Ok I cloned and replaced that line with:

    ssh root@192.168.1.1 "sh /etc/uci-defaults/80_example"

E.g. the script will be executed with shell but not marked as executable.
I checked and it works

jow-:
Great, thanks!

PR #5914, opened 2022-07-31
luci-app-email: configure emailrelay

stokito:
A new Luci App to configure [EmailRelay package](https://openwrt.org/docs/guide-user/services/email/emailrelay). With this inexperienced users may set up their own SMTP and POP server just on a router. Sending of emails is still a pain because you need to configure SPF but 90% of a typical usage is just receiving email.

I decided to call it just luci-app-email because anyway only the emailrelay has a UCI config. Also I think to add a basic view of a mailbox.

This is a first version and far from ideal but at least this will be a good start.
All options from the PR https://github.com/openwrt/packages/pull/18536 are supported.
It still lacks of configuration of authorization and a user still need to edit the /etc/emailrelay.auth file manually over scp. But I'll fix that too.

### DIFF
--- a/applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
+++ b/applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
@@ -1,0 +1,297 @@
+'use strict';
+'require view';
+'require form';
+
+return view.extend({
+	render: function () {
+		var docsRefAttrs = 'target="_blank" rel="noreferrer" href="https://emailrelay.sourceforge.net/';
+		var m, s, o;
+
+		m = new form.Map('emailrelay', _('Email Server Configuration'),
+			_('E-MailRelay Server Configuration.') + '<br />' +
+			_('For further information <a %s>check the documentation</a>')
+				.format('href="https://openwrt.org/docs/guide-user/services/email/emailrelay" target="_blank" rel="noreferrer"')
+		);
+
+		s = m.section(form.GridSection, 'emailrelay', _('Instance config'));
+		s.anonymous = false;
+		s.addremove = true;
+		s.nodescriptions = true;
+
+
+		o = s.tab('smtp_server', _('SMTP Server'));
+		o = s.tab('smtp_client', _('SMTP Client'));
+		o = s.tab('pop_server', _('POP3'));
+		o = s.tab('advanced', _('Advanced Settings'));
+
+		o = s.taboption('smtp_server', form.Flag, 'enabled', _('Enabled'));
+		o.rmempty = false;
+
+		o = s.taboption('smtp_server', form.ListValue, 'mode', _('Mode'),
+			_('See <a %s>Running E-MailRelay</a>')
+				.format(docsRefAttrs + '#userguide_md_Running_E_MailRelay"')
+		);
+		o.value('server', _('Server: receive incoming mail'));
+		o.value('proxy', _('Proxy: submission outgoing mail, store and forward to smarthost SMTP server'));
+		o.value('cmdline', _('Manual command line options (deprecated)'));
+		o.default = 'server';
+		o.rmempty = false;
+
+		o = s.taboption('smtp_server', form.Value, 'port', _('SMTP Port'),
+			_('SMTP Port to listen for incoming emails.') + '<br />' +
+			_('Incoming mail by default received on <em>25</em> port.') + '<br />' +
+			_('Outcoming mail by usually received on <em>587</em> or <em>465</em> (TLS only) ' +
+				'but the <em>25</em> is also used often.')
+		);
+		o.datatype = 'port';
+		o.default = '25';
+
+		o = s.taboption('smtp_server', form.Flag, 'remote_clients', _('Allow remote clients'),
+			_('Allow connections from the public internet.') + '<br />' +
+			_('<b>You may receive spam so be careful</b>.') + '<br />' +
+			_('Enable ports in firewall.') + '<br />' +
+			_('See <a %s>--remote-clients</a>')
+				.format(docsRefAttrs + '#__remote_clients"')
+		);
+		o.modalonly = true;
+		o.default = '0';
+
+		o = s.taboption('smtp_server', form.Flag, 'anonymous', _('Anonymous'),
+			_('Reduce the amount of information leaked to remote clients.') + '<br />' +
+			_('See <a %s>--anonymous</a>')
+				.format(docsRefAttrs + '#__anonymous"')
+		);
+		o.modalonly = true;
+		o.default = '0';
+
+		o = s.taboption('smtp_server', form.Value, 'domain', _('Domain'),
+			_(
+				'Specifies the server\'s domain name that is used in SMTP EHLO.<br />' +
+				'By default, the local hostname is used.<br />' +
+				'See <a %s>--domain</a>'
+			).format(docsRefAttrs + '#__domain"')
+		);
+		o.datatype = 'hostname';
+		o.optional = true;
+		o.rmempty = false;
+		o.depends('anonymous', '0');
+
+		o = s.taboption('smtp_server', form.FileUpload, 'server_auth', _('Auth file'),
+			_(
+				'Server/proxy authorization file.<br />' +
+				'See <a %s>Authentication</a>'
+			).format(docsRefAttrs + '#reference_md_Authentication"')
+		);
+		o.datatype = 'file';
+		o.root_directory = '/';
+		o.default = '/etc/emailrelay.auth';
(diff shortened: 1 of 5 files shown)


stokito:
Hey @fededim you are a maintainer of the emailrelay. Please have a look

feckert (changes requested):
Could you use please `<tabs>` for the indentation?

stokito:
@feckert on a next PR with more functionality with authorization I may change to tabs if this is necessary. If this is a strict requirement please add some linter check.
Even now it does work and may turn attention of other who may decide to contribute

Neustradamus:
@stokito: Have you looked for the change request?

stokito:
I found a time and changed them to tabs. Please merge

stokito:
hi, please merge

applications/luci-app-email/root/usr/share/rpcd/acl.d/luci-app-email.json
```diff
@@ -0,0 +1,24 @@
+{
+	"luci-app-email": {
+		"read": {
+			"ubus": {
+				"uci": [
+					"get"
+				]
+			},
```
feckert, line 8: I think we do not need this explicit ubus acl for uci there. This acl is already granted within the next line.

applications/luci-app-email/root/usr/share/rpcd/acl.d/luci-app-email.json
```diff
@@ -0,0 +1,24 @@
...
+			"uci": [
+				"emailrelay"
+			]
+		},
+		"write": {
+			"ubus": {
+				"uci": [
+					"set", "commit"
+				]
+			},
```
feckert, line 18: dito

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
```diff
@@ -0,0 +1,218 @@
...
+			_('Server/proxy authorization file. See --server-auth'));
+		o.datatype = 'file';
+		o.default = '/etc/emailrelay.auth';
+		o.optional = true;
+		o.rmempty = true;
+		o.modalonly = true;
+
+		o = s.taboption('smtp_server', form.Value, 'dnsbl', _('DNSBL'),
+			_('Comma separated config of DNS Block List. Starts with the transport DNS server:port, a timeout, threshold and servers.\n' +
+				'See --dnsbl and Connection blocking in reference'));
```
feckert, line 65: Please split the translation in two ofter the '\n'

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
```diff
@@ -0,0 +1,218 @@
...
+		o.datatype = 'file';
+		o.optional = true;
+		o.rmempty = false;
+		o.modalonly = true;
+		o.depends('server_tls', '1');
+
+		o = s.taboption('smtp_server', form.Value, 'server_tls_verify', _('CA certificate'),
+			_('Enables verification of remote SMTP and POP clients\' certificates against any of the trusted ' +
+				'CA certificates in the specified file or directory. In many use cases this should be a file containing ' +
+				'just your self-signed root certificate. ' +
```
feckert, line 98: Please split the translation in two after '.'
And please fix the alignment oft translation.

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
```diff
@@ -0,0 +1,218 @@
...
+		o.optional = true;
+		o.rmempty = false;
+		o.modalonly = true;
+		o.depends('client_tls', '1');
+
+		o = s.taboption('smtp_client', form.Value, 'client_tls_verify', _('CA certificate'),
+			_('Verify the remote SMTP server\'s certificate against any of the trusted CA certificates ' +
+				'in the specified file or directory. ' +
+				'The <code>&lt;default&gt;</code> means <code>/etc/ssl/certs/ca-certificates.crt</code>')
+		);
```
feckert, line 153: Please split the translation in two after '.'
And please fix the alignment oft translation.

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
```diff
@@ -0,0 +1,218 @@
...
+
+		o = s.taboption('advanced', form.Value, 'filter', _('Filter'),
+			_('Runs the specified external filter program whenever a mail message is stored.'));
+		o.optional = true;
+		o.rmempty = false;
+		o.modalonly = true;
+
+		o = s.taboption('advanced', form.Value, 'extra_cmdline', _('Extra command line options'),
+			_('Specify all arguments that should be passed to EmailRelay. ' +
+				'See <a href=\'http://emailrelay.sourceforge.net/#reference_md_Reference\'>command line reference</a>'));
```
feckert, line 211: Please split the translation in two after '.'
And please fix the alignment oft translation.

stokito:
Hi @feckert, I fixed your comments and extended descriptions.

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
```diff
@@ -0,0 +1,306 @@
...
+		o.rmempty = false;
+		o.modalonly = true;
+		o.depends('server_tls', '1');
+
+		o = s.taboption('smtp_server', form.Value, 'server_tls_verify', _('CA certificate'),
+			_(
+				'Verify an SMTP and POP client\'s certificates against any of the trusted ' +
+				'CA certificates in the specified file or directory.<br/>' +
+				'In many use cases this should be your self-signed root certificate.<br/>' +
+				'Use <code>&lt;default&gt;</code> to use the system trusted CAs'
```
feckert, line 132: Can you move all the sentences separated by `</br>` into its own translation?
So we have:
```
_('Verify an SMTP and POP client\'s certificates against any of the trusted CA certificates in the specified file or directory.) + '<br/>' +
_('In many use cases this should be your self-signed root certificate.) + '<br/>' +
_('Use <code>&lt;default&gt;</code> to use the system trusted CAs.')
```
Could you please update this also for the other Sections in this pullrequest.

This facilitates the translation and the translators do not have to deal with the HTML tags.

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
```diff
@@ -0,0 +1,306 @@
...
+		o.optional = true;
+		o.rmempty = false;
+		o.modalonly = true;
+		o.depends('client_tls', '1');
+
+
+		o = s.taboption('pop_server', form.Flag, 'pop', _('Enable POP3'),
+			_(
+				'The POP3 used to fetch a mail.<br/>' +
+				'See <a href="%s#userguide_md_Running_as_a_POP_server" target="_blank" rel="noreferrer">Running as a POP server</a>'
```
feckert, line 225: We should also move the the HTML link tags into the format so the translater doe not need to deal with HTML tags.
So we have:
```
_('The POP3 used to fetch a mail.) + <br/>' +
_('See <a "%s">Running as a POP server</a>')
.format('href="https://emailrelay.sourceforge.net/#userguide_md_Running_as_a_POP_server" target="_blank" rel="noreferrer"')
```

feckert (changes requested):
The two comments also apply to the other translation blocks. These changes facilitate the work of the Weblate translators.

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
stokito: I checked the Weblate interface to see how it feels for translators.
It shows fields separately but also don't show a place in code so if we have three messages then a translator may not have a clue that they are all related and have a same context.
Short messages would be more likely translated. Bigger messages may be better for performance because less keys. Not sure about this.

The html tags looks like not a problem for the weblate.
Anyway I'll apply the requested changes.

applications/luci-app-email/htdocs/luci-static/resources/view/email/emailrelay.js
stokito: fixed

stokito:
@feckert I fixed you comments.
CC @stangri

stokito:
I made emailrelay.auth, pop.auth and tls cert files with form.FileUpload so a user can directly change them

feckert:
I corrected some translations that you forgot and pushed directly.
See https://github.com/openwrt/luci/commit/a06ca479115d5924eed785c7435ba37fbef422ee

Thanks for your work!

PR #5915, opened 2022-07-31
luci-app-opkg: move pager to bottom

### LINKED ISSUE #5671: luci-app-opkg: add pager at the bottom of listed packages
While looking through my installed packages I noticed that the [pager](https://github.com/openwrt/luci/blob/df07e6b833e97fde2718a5f74feeaa04ad47c099/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js#L1059) for the package list is placed at the top of the list only. After reaching the bottom of a page, navigating to the next page requires scrolling all the way back up. An additional pager at the bottom would make navigating between pages more convenient.

Current:
image

Proposed change:
image

Making this second pager dependent on the length of the package list would be a nice touch if it doesn't add too much code debt.

I'm open to submitting a PR, although I can't make any promises as to when I'll have some free time to work on this.

Additional information:
- OpenWrt 21.02.1 r16325-88151b8303
- LuCI openwrt-21.02 branch (git-22.025.79177-4ec18ec)
- luci-app-opkg git-21.312.69848-4745991

stokito:
Fix #5671

### DIFF
--- a/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
+++ b/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
@@ -1056,14 +1056,6 @@ return view.extend({
 				E('li', { 'data-mode': 'updates', 'class': 'installed cbi-tab-disabled', 'click': handleMode }, E('a', { 'href': '#' }, [ _('Updates') ]))
 			]),
 
-			E('div', { 'class': 'controls', 'style': 'display:none' }, [
-				E('div', { 'id': 'pager', 'class': 'center' }, [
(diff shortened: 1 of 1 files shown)


jow-:
Shouldn't we also automatically scroll at the top then? Otherwise users would be forced to navigate all the way back to the top every time they switch pages.

stokito:
Makes sense. Looks like we just need to add the `window.scrollTo(0, 0);`

instag1b:
Wouldn't it make sense to have a pager at the top _and_ bottom? This way it doesn't matter which way a user navigates through the list. With a pager at the bottom only, navigating forward is easier, but going the other way round now requires additional scrolling.

jow-:
Superseded by ac47bc9dd7bff9f415ffa13a1b02892ca3f4453f

stokito:
tested, works fine. Thank you

PR #5916, opened 2022-08-01
luci-theme-openwrt-2020: disabled button text wrap

nuintun:
luci-theme-openwrt-2020: disabled button text wrap


jow-:
This was added intentionally to allow long button labels to wrap on narrow screens.

nuintun:
@jow- On the software page, the screen width is enough, but the button text also wraps.

PR #5934, opened 2022-08-16
luci-app-attendedsysupgrade: Fixed a typo

ManuLinares:
Fixed a typo in: applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/configuration.js

Signed-off-by: Manu Linares <mbarriolinares@gmail.com>


ManuLinares:
mmm, a 1 letter PR, not implemented yet.

hnyman:
If you merge it now, the translations will be invalidated as the string changes. But the .pot would be regenerated at the translation sync. And translations could be re-home then

You could also do a second commit that updates also the .pot and all .po files that have an actual translation for that string. That  would make it seamless.

ManuLinares:
> If you merge it now, the translations will be invalidated as the string changes. But the .pot would be regenerated at the translation sync. And translations could be re-home then
> 
> You could also do a second commit that updates also the .pot and all .po files that have an actual translation for that string. That would make it seamless.

I made a second PR that encompasses all .po .pot and the original files.

https://github.com/openwrt/luci/pull/6019

PR #5935, opened 2022-08-16
luci-app-frpc: some improvement

jjm2473:
1. fix local_ip data type
2. add plugin options
3. show remote port on overview page
4. some translations fix

### DIFF
--- a/applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
+++ b/applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
@@ -72,6 +72,16 @@ var stcpProxyConf = [
 	[form.Value, 'sk', _('Sk')],
 ];
 
+var pluginConf = [
+	[form.ListValue, 'plugin', _('Plugin'), undefined, {values: ['', 'http_proxy', 'socks5', 'unix_domain_socket'], rmempty: true}],
+	[form.Value, 'plugin_http_user', _('HTTP user'), undefined, {depends: {plugin: 'http_proxy'}}],
+	[form.Value, 'plugin_http_passwd', _('HTTP password'), undefined, {depends: {plugin: 'http_proxy'}}],
+	[form.Value, 'plugin_user', _('SOCKS5 user'), undefined, {depends: {plugin: 'socks5'}}],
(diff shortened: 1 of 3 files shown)


applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
```diff
@@ -212,6 +212,16 @@ return view.extend({
...
 		s.option(form.Value, 'local_ip', _('Local IP')).modalonly = false;
 		s.option(form.Value, 'local_port', _('Local port')).modalonly = false;
+		o = s.option(form.Value, 'remote_port', _('Remote port'));
+		o.modalonly = false;
+		o.depends('type', 'tcp');
+		o.depends('type', 'udp');
+		o.cfgvalue = (function(cfgvalue, _this){
+			return function() {
+				var v = cfgvalue.apply(_this, arguments);
+				return v&&v!='0'?v:'#';
```
feckert, line 233: ? Can't it be simpler? No one understands that

feckert (changes requested):

applications/luci-app-frpc/htdocs/luci-static/resources/view/frpc.js
jjm2473: Just a javascript closure, for reducing local variables.
This is equals to:
```
var oldcfgvalue = o.cfgvalue;
o.cfgvalue = function() {
  var v = oldcfgvalue.apply(o, arguments);
  return v&&v!='0'?v:'#';
};
```

jjm2473: Is there any way to wrap `cfgvalue` function?

feckert: I'm sorry, but every time I look at this line, I have to think what this is doing.
Since I read a lot of code with different language, I prefer it to make this expression easier. 
Regardless of whether we reduce local variables, I think the readability is more important from my point of view!

jjm2473: I found that `this.super` fit my situation, it's simple now

cambrian07 (approved):

PR #5936, opened 2022-08-17
luci.mk: execute install recipe in subshell

mpratt14:
(CI failing because it assumes I'm working on a package)

The use of bash if statements has extra syntax requirements
compared to logical operators && and ||
for example, that each line must end with a semicolon.
Also, each line will execute regardless of the exit status of the last.

The default definition of Build/Install/Default
is not managed within the luci repo,
so a syntax requirement is not ideal.
A missing or extra semicolon for whatever reason
causes build failure.

By placing the Build/Install/Default recipe in a subshell
and connecting it to the next line with logical operator &&
the syntax of Build/Install/Default doesn't matter anymore,
and the copying from PKG_INSTALL_DIR is dependent
on Build/Install/Default being successful,
where the copy would still execute if Build/Install/Default fails
and the lines are separated.

Signed-off-by: Michael Pratt <mcpratt@pm.me>

### DIFF
--- a/luci.mk
+++ b/luci.mk
@@ -202,7 +202,7 @@ define Package/$(PKG_NAME)/install
 	  cp -pR $(PKG_BUILD_DIR)/root/* $(1)/; \
 	else true; fi
 	if [ -d $(PKG_BUILD_DIR)/src ]; then \
-	  $(call Build/Install/Default) \
+	  ( $(call Build/Install/Default) ) && \
 	  $(CP) $(PKG_INSTALL_DIR)/* $(1)/; \
 	else true; fi
 endef


jow-:
Hmm, the `Install/Default` macro as it is right now will execute each line in a new shell according to the usual Make semantics - although there currently is just one line/command in it, wrapping the default macro in a subshell here will violate that assumption and if it ever gains more lines it might break or behave differently in the context of `luci.mk` which likely is not desired.

Can you give an example of an issue this PR fixes?

mpratt14:
> each line in a new shell according to the usual Make semantics

AFAIK that only applies to target rules. the way it is called here, with the newlines at the end of each line `\` means it is only part of a single command

example of breakage is if you were to remove the semicolon in `Build/Install/Default` in the core repo

jow-:
> AFAIK that only applies to target rules. the way it is called here, with the newlines at the end of each line \ means it is only part of a single command

Yeah sure, that's why I said that there is only command in there right now. But it's legally allowed to add another, unrelated command not joined by `&&`, `;` or `\`.

mpratt14:
all this does is make the syntax requirements for adding something require a change in luci.mk without any possibility of issues caused by the syntax in the main repo

`if []; then ; else ; fi` requires the semicolons regardless of whether it comes from the call or not

do you think the `&&` should just be a semicolon instead?
the only reason I did it this way is because it doesn't make sense to me for the second line to execute if the first fails

jow-:
I noticed that the existing code had the same issue and also would break if the default install recipe ever gains another line. I decided to fix the presented issue differnetly, relying on make logic to not generate uneeded install commands in the first place.

mpratt14:
thanks for your time, I like your alternative 👍🏼

PR #5938, opened 2022-08-18
luci-app-lorawan-basicstation: Add luci-app-lorawan-basicstation

mars642:
LuCi support for Basicstation

Depends on https://github.com/openwrt/packages/pull/19188

Signed-off-by: Marcus Schref <mschref@web.de>

### DIFF
--- a/applications/luci-app-lorawan-basicstation/Makefile
+++ b/applications/luci-app-lorawan-basicstation/Makefile
@@ -1,0 +1,19 @@
+#
+# Copyright (C) 2022 TDT AG <development@tdt.de>
+#
+# This is free software, licensed under the Apache License Version 2.0.
+# See https://www.apache.org/licenses/LICENSE-2.0 for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI Support for LoRaWAN basicstation
+LUCI_DEPENDS:=+basicstation
+LUCI_PKGARCH:=all
+
+PKG_MAINTAINER:=Marcus Schref <mschref@tdt.de>
+PKG_LICENSE:=APACHE-2.0
+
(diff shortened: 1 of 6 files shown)


mars642:
ping @aparcar

aparcar:
Please post some screenshots

mars642:
Here are some screenshots of the luci app
GeneralSettings1
GeneralSettings2
AdvancedSettings
LogMessages

applications/luci-app-basicstation/Makefile
```diff
@@ -0,0 +1,22 @@
...
+# Copyright (C) 2022 TDT AG <development@tdt.de>
+#
+# This is free software, licensed under the Apache License Version 2.0.
+# See https://www.apache.org/licenses/LICENSE-2.0 for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-app-basicstation
+PKG_VERSION:=2.0.6
```
BKPepe, line 11: Why such versioning? Couldn't it be 1.0.0 as this is the first initial version, which could gets into this repository?

hnyman:
I would rename this to something like luci-app-lorawan-basicstation
And also add LoRaWAN to the title seen in menuconfig:
`LUCI_TITLE:=LuCI Support for basicstation`
--> `LUCI_TITLE:=LuCI Support for LoRaWAN basicstation`

Otherwise the "basic station" may sound too much like a simplified setup tool for a router.  (there is already a PR for luci-mod-simple, etc.)

applications/luci-app-basicstation/Makefile
hnyman: Removing versioning is fine, as there is automatic versioning in LuCI. (most packages do not have manual versions at all)

PR #5958, opened 2022-09-06
luci-app-v2raya: add basic support for v2rayA

1715173329:
This adds the basic LuCI support for v2rayA which is already in [packages feed](https://github.com/openwrt/packages/tree/master/net/v2raya).


1715173329:
Screenshots:
image
image

PR #5961, opened 2022-09-06
luci-app-upnp: make nftables compatible

jow-:
Make luci.upnp rpcd backend plugin compatible with miniupnpd-nftables.

Depends on PR https://github.com/openwrt/packages/pull/19320


fuqiang03:
Use fw3
luci-app-upnp
xxx
xxxx
There is no problem with using use firewall4

PR #5964, opened 2022-09-07
luci-base: fix empty config not satisfy uci depends

jjm2473:


jow-:
What is the use case for it? The existing logic is intended, we usually don't want empty uci files to satisfy the dependency as those are rarely meaningful.

Maybe you should consider a filesystem dependency on /etc/config/... instead?

jjm2473:
@jow- I have install `nut`, `luci-app-nut`,`nut-server`, `nut-upsmon`, and `nut-web-cgi`, but luci shows 404 page on `admin/services/nut`

jow-:
Superseded by bfc0fcfe6e97b1b39317f42d96debf3286c42696

PR #5977, opened 2022-09-17
proper cache control in firefox

danielkucera:
Added more meta headers to ensure proper cache control in Firefox. Before this patch, the `/` url hasn't even been requested when opened in Firefox.

Signed-off-by: Daniel Kucera <daniel.kucera@gmail.com>


jow-:
Merged, thanks!

PR #5980, opened 2022-09-18
luci-app-smartdns: fix redirect error message issue

pymumu:
Maintainer: me
Compile tested: TL-WR703N/OpenWrt 22.03 stable

screenshot


jow-:
Merged, thanks!

PR #5989, opened 2022-09-26
Update DNSPod DOH ip

ycloud:
https://www.dnspod.cn/Products/publicdns

Signed-off-by: Jang Jun <abc123@foxmail.com>


stangri:
The `bootstrap_dns` setting is not for the IP addresses of the DoH servers, it's an IP address for the fallback, non-DoH, resolver, which in this case is still the same as in the current file.
image

ycloud:
Sorry,
I understand wrong.

PR #5992, opened 2022-09-27
luci-app-dawn: return 0 in case of weird frequency

PolynomialDivision:
Sometimes 0 is passed to the frequency. Return 0 in that case.


loopy321:
Thanks.  Tested and works.  
Tried something similar yesterday, but missed the `rm -r /tmp/luci-*` step after editing ieee80211.lua.

PolynomialDivision:
@hnyman Plesse merge

PR #5996, opened 2022-09-29
luci-mod-network: improve description for filterwin2k option

janh:
Make it clear that this option also filters some not-so-useless queries.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -385,7 +385,8 @@ return view.extend({
 
 		s.taboption('advanced', form.Flag, 'filterwin2k',
 			_('Filter useless'),
-			_('Do not forward queries that cannot be answered by public resolvers.'));
+			_('Avoid uselessly triggering dial-on-demand links (filters SRV/SOA records and names with underscores).') + '<br />' +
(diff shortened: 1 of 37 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -385,7 +385,7 @@ return view.extend({
 
 		s.taboption('advanced', form.Flag, 'filterwin2k',
 			_('Filter useless'),
-			_('Do not forward queries that cannot be answered by public resolvers.'));
+			_('Avoid uselessly triggering dial-on-demand links by filtering SRV and SOA records, as well as names containing underscores. May prevent VoIP or other services from working.'));
```
feckert, line 388: Sorry but can you please put this into two separate translations.
The whole sentence is already quite long.
```
_('Avoid uselessly triggering dial-on-demand links by filtering SRV and SOA records, as well as names containing underscores') 
+ _( 'May prevent VoIP or other services from working.');
``

janh: Yeah, I'm not really happy with that long description myself, but I feel like the behaviour of this config option is really unexpected without an explanation (and it actually breaks things).

I shortened it a little bit and split it into two parts (and as some kind of whitespace is needed anyway I added a linebreak that makes the the warning about VoIP stand out a bit). Is that better?

Issue #5987, opened 2022-09-23
luci-app-ddns: error when trying to create a service


castillofrancodamian:
# Issue template:
1. Please make sure that the issue subject starts with `<package-name>:`
2. Issues related to end of life (EOL) releases are not supported or maintained and will be closed
3. Remove lines from top till here and fill in the following questions
***

## Steps to reproduce:
**Example: Replace the following lines and remove this**

1. go to: Services → Dynamic DNS→ Add new services...
2. In the pop-up enter any name and choose any provider.
3. Click on "Create service"

Actual behavior:
Error that prevents the creation of a service.

Expected behavior:
Create a DDNS service and make the corresponding configuration.

Additional Information:
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='SNAPSHOT'
DISTRIB_REVISION='r20728-a24e07700d'
DISTRIB_TARGET='ath79/generic'
DISTRIB_ARCH='mips_24kc'
DISTRIB_DESCRIPTION='OpenWrt SNAPSHOT r20728-a24e07700d'
DISTRIB_TAINTS='no-all'

csjtl:
me too

IP don't Update

 214812       : ************ ************** ************** **************
 214812  note : PID '3375' started at 2022-10-01 21:48
 214812       : ddns version  : 2.8.2-27
 214812       : uci configuration:
ddns.proxypoolCheck.check_unit='minutes'
ddns.proxypoolCheck.domain='ddns.xxxxxx.com'
ddns.proxypoolCheck.enabled='1'
ddns.proxypoolCheck.force_unit='minutes'
ddns.proxypoolCheck.interface='wan'
ddns.proxypoolCheck.ip_network='wan'
ddns.proxypoolCheck.ip_source='network'
ddns.proxypoolCheck.lookup_host='ddns.xxxxxx.com'
ddns.proxypoolCheck.password='***PW***'
ddns.proxypoolCheck.retry_unit='seconds'
ddns.proxypoolCheck.service_name='google.com'
ddns.proxypoolCheck.use_ipv6='0'
ddns.proxypoolCheck.use_syslog='2'
ddns.proxypoolCheck.username='xxxxxxxxxxxxxx'
ddns.proxypoolCheck=service
 214813       : verbose mode  : 0 - run normal, NO console output
 214813       : check interval: 600 seconds
 214813       : force interval: 4320 seconds
 214813       : retry interval: 60 seconds
 214813       : retry max count : 0 times
[...]

PR #5999, opened 2022-10-02
luci-app-ddns: fix creation of custom service (fix #4926)

### LINKED ISSUE #4926: luci-ddns writes incorrect service_name when "--custom--" service is chosen
luci-ddns write incorrect service into /etc/config/ddns when "--custom--" service is chosen
`option service_name '-'` where this line should be deleted or commented ( https://openwrt.org/docs/guide-user/services/ddns/client#custom_service1 )

This would result in "No update_url found/defined or no update_script found/defined! - TERMINATE" error when it try to update

jcassette:
The `service_name` option must be empty for custom services, otherwise an error is raised in `dynamic_dns_updater.sh`.

Signed-off-by: Julien Cassette <julien.cassette@gmail.com>


castillofrancodamian:
Would [this](https://github.com/openwrt/luci/issues/5987) be fixed?

PR #6001, opened 2022-10-03
luci-base: add alphanumeric validators in validation.js

Akira25:
This adds two validators in the validation class. One validator for alphanumeric strings [A-z0-9] and one for alpha numeric strings with spaces [A-z0-9 ]. These come in very handy for validating input strings which aren't allowed to contain any characters that could be used in valid code statements.

Signed-off-by: Martin Hübner <martin.hubner@web.de>

We need theses validators for a App, we are currently working on at Freifunk Berlin. Description of Freifunk-Services must not contain any other characters than alphanumeric and whitespace.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/validation.js
+++ b/modules/luci-base/htdocs/luci-static/resources/validation.js
@@ -236,6 +236,14 @@ var ValidatorFactory = baseclass.extend({
 	},
 
 	types: {
+		alphanumeric: function() {
+			return this.assert(this.value.match(/^[a-zA-Z0-9]+$/), _('string with characters and numbers only.'));
+		},
+
+		alphanumericspace: function() {
+			return this.assert(this.value.match(/^[a-zA-Z0-9 ]+$/), _('string with characters, numbers and spaces only.'));
+		},
+
 		integer: function() {
 			return this.assert(!isNaN(this.factory.parseInteger(this.value)), _('valid integer value'));
 		},


jow-:
This looks rather arbitrary, at least I didn't see a use case for this elsewhere. Is using custom validators not an option in your case? 

```js
o = s.option(...);
o.validate = function(section_id, value) {
    return String(value).match(/^[a-zA-Z0-9]+$/) ? true : _('Require string with characters and numbers only.'));
};
```

Akira25:
> This looks rather arbitrary, at least I didn't see a use case for this elsewhere. Is using custom validators not an option in your case?
> 
> ```js
> o = s.option(...);
> o.validate = function(section_id, value) {
>     return String(value).match(/^[a-zA-Z0-9]+$/) ? true : _('Require string with characters and numbers only.'));
> };
> ```

Using a cusotm validator is indeed an option, but I've found no documentation nor any example for this. Thank you for providing one.

This can be closed then.

jow-:
> but I've found no documentation nor any example for this

Just for reference, the documentation bit is here: https://openwrt.github.io/luci/jsapi/LuCI.form.AbstractValue.html#validate

You can find various examples by issuing a `grep -r 'validate = function` in the LuCI tree.
Also keep in mind that the validator function will receive an empty string as input in case the option object is marked as optional (`.rmempty = true`) so make sure your function handles this, e.g. by using `…*` instead of `…+` in regular expressions.

PR #6015, opened 2022-10-11
luci-app-firewall: only allow one flow offloading option

arinc9:
Either software or hardware flow offloading can be used at a time. Reflect this on the firewall section of LuCI.

Signed-off-by: Arınç ÜNAL <arinc.unal@arinc9.com>

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
@@ -82,12 +82,13 @@ return view.extend({
 				_('Software flow offloading'),
 				_('Software based offloading for routing/NAT'));
 			o.optional = true;
+			o.depends({flow_offloading_hw: '1', '!reverse': true});
 
 			o = s.option(form.Flag, 'flow_offloading_hw',
 				_('Hardware flow offloading'),
 				_('Requires hardware NAT support. Implemented at least for mt7621'));
 			o.optional = true;
-			o.depends('flow_offloading', '1');
+			o.depends({flow_offloading: '1', '!reverse': true});
 		}
 
 


jow-:
This contradicts the firewall3 implementation:
https://git.openwrt.org/?p=project/firewall3.git;a=blob;f=defaults.c;h=8a9a9295f00500ca7ba655de111acc643d41b214;hb=4cd7d4f36bea731bf901cb067456f1d460294926#l258

And the firewall4 implementation:
https://git.openwrt.org/?p=project/firewall4.git;a=blob;f=root/usr/share/ucode/fw4.uc;h=2a1e397c2e84c9437319232760c528c36dd3e148;hb=3db474135c62d1eeea7a80dd151d53d94005f1d5#l460

On either variant, the `flow_offloading_hw` option has no effect without `flow_offloading` being enabled.

arinc9:
The implementation on firewall3 and firewall4 is different than what I originally thought it would be (set an nft/iptables rule, disable flow_offloading if flow_offloading_hw was enabled). Instead the check for flow_offloading_hw works after enabling flow_offloading.

> Furthermore your change introduces a circular option dependency.

I initially wanted to turn this into a drop list but figured we'd need to keep the section IDs. It doesn't matter now anyway.

Thanks for pointing this out Jo.

arinc9:
I've read a bit the LuCI JS APIs, I don't think I can pull this off. It gets more complex than what I can currently manage.

If you find time to do this yourself something like this should fit it, with slight changes on the strings for better clarification.

### Flow Offloading
Experimental feature. Not fully compatible with QoS/SQM.

		Offloading type <Dropdown List>
				Disabled
				Software offloading
					Software based offloading for routing with/without NAT.
				Hardware offloading
					Requires the hardware to support it. Implemented at least for MT7621.

Like here:

Screenshot from 2022-10-17 00-55-06

arinc9:
@jow- I'm taking another shot at this now that ChatGPT is here. I don't think I can reopen the pull request so could you review this?

https://github.com/arinc9/luci/commit/87114333daadfc33fd985eb2ecd56b6bc77c8fd6

I just copy pasted the cbiRichListValue class from interfaces.js. Let me know how would I properly import it.

Choosing the offloading type actually sets the correct options via uci.set. But it resets to "None" after pressing save or reloading the page. I did try some things but it won't work on its current form.

It currently looks like this, there's an empty option there I couldn't get rid of:

Screenshot from 2023-02-13 01-36-30

arinc9:
Ok everything works with https://github.com/arinc9/luci/commit/3e6cc13445170cd382546b1ce10a4c0141da7571.

I still couldn't figure out how to stop the empty o.value('') from appearing. If I set the value of None to '', choosing None won't do anything.

@jow- looking forward to your suggestions.

jow-:
When you instantiate the `ui.Dropdown()` widget, you pass `optional: true`. Since the choice list you supply does not contain any option with value `""`, the dropdown implementation adds one automatically, that's the empty one you see.

It should be gone once you remove the `optional: true` in your code.

arinc9:
Got it, thanks a lot. I'll open a new pull request.

PR #6026, opened 2022-10-14
luci-theme-openwrt-2020: quantize png image

neheb:
Small size reduction.

From: 5908
To:   3896


neheb:
33%. Not bad.

PR #6035, opened 2022-10-18
luci-app-dockerman: change default to nil if data.blkio_weight is not defined

myrlyn:
Signed-off-by: Jonathon Walker <jonathon.l.walker@gmail.com>


myrlyn:
Submitted with the minimum changes necessary to get luci-app-dockerman working with the default kernel.

PR #6043, opened 2022-10-19
luci-mod-network: 'mxhosts' tab added

systemcrash:
sub area: luci-app-dhcp

DHCP Help does not document mxhost, but they are available in dnsmasq.

This effectively implements MX records within DNS.

Tested on 22.03.2

mx


jow-:
Please use `luci-mod-network: ` as commit subject prefix. Also I think the additional record type tabs should be added after `Hostnames` as they'll be less commonly used.

jow-:
Merged via 3240714ed3

PR #6047, opened 2022-10-20
luci-mod-network: 'relay' tab added

systemcrash:
Adds LuCI GUI for dnsmasq DHCP 'relays'.

The helper functions in dnsmasq/files/dnsmasq.init limit behaviour whereby it is not possible to omit server_addr, where to omit it (dnsmasq supported) would result in a broad- or multicast further on.

Tested on 22.03.2

relaynew

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -231,7 +231,8 @@ return view.extend({
 		return Promise.all([
 			callHostHints(),
 			callDUIDHints(),
-			getDHCPPools()
+			getDHCPPools(),
+			network.getDevices()
 		]);
 	},
 
@@ -240,6 +241,7 @@ return view.extend({
 		    hosts = hosts_duids_pools[0],
 		    duids = hosts_duids_pools[1],
 		    pools = hosts_duids_pools[2],
+		    ndevs = hosts_duids_pools[3],
 		    m, s, o, ss, so;
 
 		m = new form.Map('dhcp', _('DHCP and DNS'),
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -340,6 +343,66 @@ return view.extend({
...
+
+		for (var family = 4; family <= 6; family += 2) {
+			for (var i = 0; i < ndevs.length; i++) {
+				var addrs = (family == 6) ? ndevs[i].getIP6Addrs() : ndevs[i].getIPAddrs();
+				for (var j = 0; j < addrs.length; j++)
+					so.value(addrs[j].split('/')[0]);
+			}
+		}
+
+		so = ss.option(form.Value, 'server_addr', _('Relay To'));
```
feckert, line 380: Could we not change the translation for the uci option `local_addr` to `Listen address` and for the uci option `server_addr` to `Relay address`? 

That is more common.
See also the other translations in this pullrequest where `RelayTo` and `Listen addr` are used.
https://github.com/openwrt/luci/pull/6047/files#diff-90caf036d2316d858be5df6b382c44d4130a66c52cc7b6fda8ee715971a4f9bdR368
https://github.com/openwrt/luci/pull/6047/files#diff-90caf036d2316d858be5df6b382c44d4130a66c52cc7b6fda8ee715971a4f9bdR403
https://github.com/openwrt/luci/pull/6047/files#diff-90caf036d2316d858be5df6b382c44d4130a66c52cc7b6fda8ee715971a4f9bdR397

systemcrash: Relay address seems more ambiguous in this context as a relay. "Is this the from or to?" "Which side is it?"
I can certainly update to "Listen address", however.

feckert:
Thanks :+1:

PR #6055, opened 2022-10-26
luci-mod-network: remove uppercase for interface name on overview page

feckert:
In the overview page, the name of the interface is converted to uppercase. However, this is not the name in the configuration. From my point of view, this makes no sense. The name displayed should correspond exactly to the name in the configuration.


jow-:
Merged, thanks @feckert

PR #6061, opened 2022-10-27
luci-mod-network: rename button reset to delete on device page

feckert:
If a device is created that did not exist before, the description
deconfigure is not correct. When the 'Deconfigure' button is pressed,
the interface is deleted and removed from the list. This is not what one
would expect. Therefore, it is more appropriate to rename the button to
delete. This is rather the function that is executed when the button is
pressed. Devices that cannot be created dynamically cannot be
deconfigured and thus deleted.
Screenshot 2022-10-27 at 14-21-31 VR2-103954 - Interfaces - LuCI


feckert:
@jow- Thanks for the hint.
I must have completely missed the topic!

PR #6063, opened 2022-10-29
luci-mod-network: fix address validation

Zeranoe:
Fixes `--server` validation for:
- `/domain[/domain]`
- `//`
- `/#/`

Fixes hostname validation for:
- `*.domain`
- `*domain`

Changes `..domain` to be an invalid hostname

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -126,7 +126,7 @@ function validateHostname(sid, s) {
 	if (s.length > 256)
 		return _('Expecting: %s').format(_('valid hostname'));
 
-	var labels = s.replace(/^\.+|\.$/g, '').split(/\./);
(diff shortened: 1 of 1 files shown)


jow-:
Good catch! Unfortunately the `validateAddressList()` function is also directly used as validator by the `rebind_domain` option which *does* require entries with leading and trailing slash (ref https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html, ctrl-f `--rebind-domain-ok`).

So unfortunately I can't apply as-is. Would you be willing to rework it so that both cases work? Maybe take the lazy way out and simply pass `m[1]` as `'/' + m[1] + '/'` within `validateServerSpec()`...

Zeranoe:
@jow-, I ended up fixing multiple issues in de9341648aba82ebbf4af1843854dffb9e63322e, what you mentioned should also be resolved.

I noticed `address` isn't running any validation and it has a similar set of rules, perhaps it makes sense to add `validateServerSpec` to it or drop validation for `server` to match.

jow-:
Merged, thanks!

PR #6079, opened 2022-11-06
luci-proto-wireguard: fix generated AllowedIPs

### LINKED ISSUE #5956: luci-proto-wireguard: Wrong generated AllowedIPs
## Steps to reproduce:
The generated peer configuration in luci-proto-wireguard will put the *peer* IPs in `AllowedIPs`. As this is meant to be the configuration for the peer, those IPs should be configured with `Address` in the `[Interface]` section instead. The `AllowedIPs` configuration should instead include the configured host IP addresses of the interface.

1. go to: Network → Interfaces → Wireguard interface → Peers tab → Edit peer → Generate configuration

## Actual behavior:
1. The peers `AllowedIPs` are set to the peer IPs, not the host IPs
2. The peer `Address` in the `[Interface]` section is not set at all

## Expected behavior:
1. `AllowedIPs` should contain the host (openwrt) IPs set in the interface configuration
2. `Address` should be set to the peer IPs

## Additional Information:

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.0-rc4'
DISTRIB_REVISION='r19426-2b1941e47d'
DISTRIB_TARGET='ramips/mt7621'
DISTRIB_ARCH='mipsel_24kc'
DISTRIB_DESCRIPTION='OpenWrt 22.03.0-rc4 r19426-2b1941e47d'
DISTRIB_TAINTS=''
```

jcassette:
In the generated client configuration, the `AllowedIPs` option should contain the host IPs, instead of the client IPs.

Fix #5956

Signed-off-by: Julien Cassette <julien.cassette@gmail.com>

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -694,12 +694,11 @@ return network.registerProtocol('wireguard', {
 
 		o.modalonly = true;
 
-		o.createPeerConfig = function(section_id, endpoint) {
+		o.createPeerConfig = function(section_id, endpoint, ips) {
 			var pub = s.formvalue(s.section, 'public_key'),
 			    port = s.formvalue(s.section, 'listen_port') || '51820',
 			    prv = this.section.formvalue(section_id, 'private_key'),
 			    psk = this.section.formvalue(section_id, 'preshared_key'),
-			    ips = L.toArray(this.section.formvalue(section_id, 'allowed_ips')),
 			    eport = this.section.formvalue(section_id, 'endpoint_port'),
 			    keep = this.section.formvalue(section_id, 'persistent_keepalive');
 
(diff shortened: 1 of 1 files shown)


jow-:
This fix looks wrong - simply stripping the network prefix and replacing it with `/32` or `/128` will not work in all cases, e.g. when the CIDR address was a network and not a host address. Also, shouldn't a typical WireGuard roadwarrior client route all IPs through the tunnel (`::/0, 0.0.0.0/0`)? Wouldn't setting `AllowedIPs` to only the router side WireGuard interface addresses prevent traffic towards the internet?

For now, this is a NACK from me.

jow-:
> Why would one specify a network address as the interface address?

You're right, I overlooked this - so simply stripping the prefix (a little bit less verbose than appending `/32` or `/128`) should be fine.

> Should we consider roadwarrior as the default

I think road warrior is the most common case for which the users export client configs for, so that should work nicely.

> always generate AllowedIPs = 0.0.0.0/0, ::/0

We could extend the export dialog to present an additional combo box dropdown to configure the embedded peer `AllowedIPs`. It would default to `0.0.0.0/0, ::/0` and offer an additional `custom...` option which, if selected, morphs into a input value prefilled with your suggested values (router's wg `Address` values converted to host addresses).

jow-:
Perfect - merged via 73aca68b9e35857ec155f33ca1b1fb5358d69e52, thanks!

PR #6085, opened 2022-11-08
lua-runtime: template: fix "tpl" refactoring

karlp:
This fixes crashes on "tpl" being undefined.


jow-:
Merged, thanks!

PR #6109, opened 2022-11-20
luci-mod-network: improve description for filterwin2k option again

lukastribus:
@janh improved this somewhat in #5996.

This change goes one step further:

Stop using the word useless both in the name and the description and call the feature what it actually is (Filter SRV/SOA service discovery).


jow-:
Merged via 8607aa7a6712d6a78e4c8cc1be391fe60b85908c

PR #6126, opened 2022-12-03
luci-app-pbr: initial commit

stangri:
Depends on https://github.com/openwrt/packages/pull/19763


jow-:
Great work! Is there any particular reason why you dropped all translations? Is the new implementation using completely different texts?

stangri:
> Great work! Is there any particular reason why you dropped all translations? Is the new implementation using completely different texts?

Texts are mostly the same with the vpn-policy-routing, not carrying over translations was an oversight on my part. 

If I keep the translations and some of them are outdated, it's better than not having any translations, right? 

Would the translations I include in the PR be imported to the translations web-site or would anything else need to be done there to carry over translations from `vpn-policy-routing`?

hnyman:
> If I keep the translations and some of them are outdated, it's better than not having any translations, right?

Yes.

> Would the translations I include in the PR be imported to the translations web-site

Weblate translation site reads the already done stuff from our git, so nothing else needs to be done by you.
Additionally, one weblate admin (jow, aparcar, me) needs to initialize the new module.

> what's the best way of bringing in translations from vpn-policy-routing? Can I just delete all the comments with the origin file/line number in the translated fil, so  this one  will become this?

You can  do that.
But even that should not be necessary:
After copying (and renaming) the files, you need to run the translation sync, which will adjust the line number context comments in the .pot and .po files and remove all non-existing lines.
`./build/i18n-sync.sh`
(does it for all apps, but is easiest.  Just include the files regarding  your new app in the PR and forget about the other apps).

stangri:
> You can do that. But even that should not be necessary: After copying (and renaming) the files, you need to run the translation sync, which will adjust the line number context comments in the .pot and .po files and remove all non-existing lines. `./build/i18n-sync.sh` (does it for all apps, but is easiest. Just include the files regarding your new app in the PR and forget about the other apps).

Thank you Hannu! Do I need to run sed inside the .po files on [this line](https://github.com/openwrt/luci/blob/master/applications/luci-app-vpn-policy-routing/po/ar/vpn-policy-routing.po#L6) and change `vpn-policy-routing` to `pbr` in it?

hnyman:
would be good, but likely that line  has no actual role.

stangri:
@hnyman do I need to push this to the release branch as well, or will it get changes picked when the weblate artefacts as a whole are merged there?

hnyman:
Nicest would be if you commit the relevant changes to the release branch, too.

Weblate only handles master, and all translations are manually backported to 22.03.

You could also try the backport in 22.03:
checkout 22.03
build/i18n-sync, build/i18n-merge-master, and finally once more i18n-sync just in case.

PR #6129, opened 2022-12-03
luci-app-frpc：add enable options

zxlhhyccc:
without the enable option, there is no control over running or stopping the plug-in. After the enable option is added, you can control whether the plug-in is enabled or not.adaptation frps

Signed-off-by: zxlhhyccc <45259624+zxlhhyccc@users.noreply.github.com>


hnyman:
Is this about frps or frpc???
Title and commit title do not match commit contents.

PR #6132, opened 2022-12-03
luci-app-frpc：add enable options

zxlhhyccc:
without the enable option, there is no control over running or stopping the plug-in. After the enable option is added, you can control whether the plug-in is enabled or not. adaptation frpc.


BKPepe:
There are two PRs opened for the same thing. I think this PR can be closed and we can continue in https://github.com/openwrt/luci/pull/6133

PR #6133, opened 2022-12-03
luci-app-frps：add enable options

zxlhhyccc:
without the enable option, there is no control over running or stopping the plug-in. After the enable option is added, you can control whether the plug-in is enabled or not.adaptation frps.


systemcrash:
Tweak your commit and force push 
```
: Author name (zxlhhyccc) need to be your real name 'firstname lastname'
: Commit subject line MUST start with '<package name>: ' (luci-app-frps：add enable options)
```

Maybe just do `zxlhh yccc`

BKPepe:
@zxlhhyccc If you want to somehow proceed with this pull request, you should address the failing ``Test Formalities`` check.

PR #6134, opened 2022-12-04
Openwrt 22.03

ushmucan:
Modem not detected. Fibocom L860GL.


hnyman:
Invalid (and not related to LuCI GUI)

PR #6146, opened 2022-12-12
luci-base: fix button handling for named sections

feckert:
If more than one named section is added to the page, it is currently the case, that the first button is always switched on or off during input validation of the uci seciton name.
This is because the current usage of the 'document.querySelector' function uses the CSS classes. However, this is not unique if several named sections are used. By setting a unique ID for the button, the function 'document.querySelector' can now use the ID and thus the correct button is switched on or off during validation of uci section name.


jow-:
I think an easier fix would be replacing `document.querySelector(...)` with `createEl.querySelector(...)` or even `createEl.lastChild`

feckert:
Pushed changes directly with your suggestion. https://github.com/openwrt/luci/commit/440a2e79a762b4152ccb7e41c90a2826d9fb3538

Issue #6008, opened 2022-10-08
luci-app-firewall: inconsistent length checking of firewall zone names


evlli:
## Steps to reproduce:

1. go to: Network → Interfaces → edit $interface → Firewall Settings tab → Assign to new zone with a very long name >11
2. Save & Apply
3. go to: Network → Firewall → edit $new zone name
4. change some setting

## Actual behavior:

1. save button does not close the window
2. discard does not discard the changes
3. save & apply will apply the changes anyway

## Expected behavior:

**no part of the UI complains about long zone names (preferred)**
or
no part of the UI lets you set long zone names

adelton:
This seems like a duplicate of https://github.com/openwrt/luci/issues/7522, addressed in https://github.com/openwrt/luci/pull/7549.

PR #6158, opened 2022-12-18
luci-app-firewall: remove zone name length limitiation

evlli:
It was already possible to exceed the 11 character limit by creating the new zone in the interface edit view. #6008 
I could not find a test case where a longer name causes functional problems, therefor this patch removes the limit in the firewall view as well, in order to allow for more descriptive zone names. 

Signed-off-by: Evelyn Alicke <dev@evl.li>

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
@@ -122,7 +122,7 @@ return view.extend({
 		o.placeholder = _('Unnamed zone');
 		o.modalonly = true;
 		o.rmempty = false;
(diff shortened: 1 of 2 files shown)


jow-:
This change is only legal for firewall4, the iptables based firewall3 still has a length limitation.

evlli:
@systemcrash  I've thought about this a few times and I'm honestly not very familiar with the codebase. I guess replacing the htm check with more js is the only way? that feels like a much bigger change than I originally planned for but I'm fine with it.

systemcrash:
The JS part seems trivial enough. But the HTML part, I guess we can drop? All new stuff is JS only, though this does not say what we must do with the Lua only parts. @jow- ? 

If you think you can crack the HTML part also, give it a go!

PR #6164, opened 2022-12-21
[RFC] luci-mod-network: place wlan roaming related options to a tab

hnyman:
Collectd all WLAN roaming related options (802.11r, 802.11k, 802.11v) to a new dedicated tab.

The options are currently in the encryption and advanced tabs, and the options are mixed with other options. The long options list can be pretty confusing . It would be more clear to have the options in a dedicated tab.

I am not sure if "WLAN roaming" is the prefect name for the tab, but that was the best that I figured out.


jow-:
Can you supply a screenshot, just for completeness? Apart from that, good idea. Please go ahead.

hnyman:
Sure:

image

hnyman:
With 802.11r enabled the .11r options make it to a so long page that it doesn't even fit to the screen...
image

hnyman:
Funny, from the screenshot I actually spotted now a major typo ;-)

image

jow-:
@hnyman - LGTM, please merge.

castillofrancodamian:
WLAN seems redundant to me.

hnyman:
I added "wlan" to the tab title as I thought that some people might confuse a plain "Roaming" with 3G/cellular/phone roaming.

castillofrancodamian:
But in "Settings for assisting..." it already clarifies what it is for.

andybjackson:
While busy on this, please consider also the assumption that 802.11k (RRM) requires EAP, which in turn requires more than `wpad-basic`.

Yet I'm fairly sure this isn't a strict requirement. 802.11k can work in a Pre-Shared Key (PSK) context.

I agree that the title Roaming is sufficient and clear.

PR #6186, opened 2023-01-05
luci-mod-network: disambiguate DHCP IP set help text

systemcrash:
Despite this string existing for a while now, it rarely got translated, and where it did, it was equally obscure. A hint that the source string was unclear. 

Disambiguated.


jow-:
Merged, thanks!

PR #6191, opened 2023-01-09
luci-theme-: add postrm script

feckert:
Remove theme configuration uci section on theme uninstall.

### DIFF
--- a/themes/luci-theme-bootstrap/Makefile
+++ b/themes/luci-theme-bootstrap/Makefile
@@ -11,6 +11,16 @@ LUCI_DEPENDS:=
 
 PKG_LICENSE:=Apache-2.0
 
+define Package/luci-theme-bootstrap/postrm
+#!/bin/sh
+[ -n "$${IPKG_INSTROOT}" ] || {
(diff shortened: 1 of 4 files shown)


themes/luci-theme-bootstrap/Makefile
```diff
@@ -11,6 +11,24 @@ LUCI_DEPENDS:=
...
+	}
+	uci get luci.themes.BootstrapDark >/dev/null 2>&1 && {
+		uci set luci.themes.BootstrapDark=""
+		uci commit luci
+	}
+	uci get luci.themes.BootstrapLight >/dev/null 2>&1 && {
+		uci set luci.themes.BootstrapLight=""
+		uci commit luci
+	}
+}
```
jow-, line 29: I think you can simplify that into:

```bash
uci -q delete luci.themes.Bootstrap
uci -q delete luci.themes.BootstrapDark
uci -q delete luci.themes.BootstrapLight
uci commt luci
```

If the delete operation didn't change anything because the referenced option was not present initially, the `uci commit` command should become a no-op.

PR #6193, opened 2023-01-10
Fix crash issue when viewing hearing map

Brain2000:
When viewing the hearing map, sometimes a crash will happen when one of the routers suddenly cannot view a client (see screenshot). This PR simply checks that the radio frequency in data2 is not 0, and skips the entry if it is a 0.

I've verified this works on my installation of OpenWRT 22.03.2 on the Archer C7 and Belkin RT3200 running the latest version of DAWN. Unless there is a use for having a radio frequency of 0 in the hearing map, this should be fairly safe to commit.

image


jow-:
Merged via 2221110c17c2799948209463590618190cce8808 - thanks!

PR #6204, opened 2023-01-20
luci-app-firewall: Add clarification to masquerading option

mastaal:
I wasted a little to much time tinkering around with my configuration before realizing I just needed to set the "Masquerading" option. In an attempt to save everyone else this wasted time, I want to suggest to add a tooltip clarifying this option.

I am not sure how the gettext internationalization files are to be edited/generated; If anyone has pointers for this I will gladly make sure that this all works fine with my newly added interface string.


aparcar:
Please rebase your PR to lose the merge commit. Apart from that looks fine to me.

mastaal:
@aparcar Thanks, should be fixed now

PR #6209, opened 2023-01-22
rpcd-mod-luci: bump PKG_RELEASE for libiwinfo ABI change

Ansuel:
Bump PKG_RELEASE for libiwinfo ABI change for rpcd-mod-luci.


dhewg:
There's also modules/luci-mod-status which uses libiwinfo

Ansuel:
It's just a dependency but the module is just javascript files so it should not be affected

hnyman:
Agree with Ansuel.

To my knowledge ,  luci-mod-status (and bwc) do not link at build time, but seek and load the library at run time in a dynamic way

dhewg:
It's a bit hidden: https://github.com/openwrt/luci/blob/master/modules/luci-mod-status/src/luci-bwc.c#L260
Strictly speaking, we didn't change the ABI of the functions it's using, but I guess we still want to bump it so the dependencies are updated

jow-:
Please go ahead

Ansuel:
@jow- is it ok now?

jow-:
@Ansuel - yes looks fine

PR #6211, opened 2023-01-24
luci-light: rework and reenable

dhewg:
The `luci` collection contains package management and attended sysupgrades support, which are not that useful if not even undesired for self-compiled images.

Rework the `luci-light` collection to exclude the two above mentioned features, and make `luci` instead depend on the light collection in additon to those features.

The `luci-ssl` and `luci-ssl-openssl` collections then only need to depend on `luci-light`.

We now have three variants with won't pull in `luci-app-opkg` or `luci-app-attendedsysupgrade`, making everybody happy.

### DIFF
--- a/collections/luci-light/Makefile
+++ b/collections/luci-light/Makefile
@@ -9,8 +9,19 @@ include $(TOPDIR)/rules.mk
 LUCI_TYPE:=col
 LUCI_BASENAME:=light
 
-LUCI_TITLE:=Minimum package set using only admin mini and the standard theme
-LUCI_DEPENDS:=+uhttpd +luci-mod-admin-mini +luci-theme-openwrt @BROKEN
+LUCI_TITLE:=LuCI interface with Uhttpd as Webserver (light)
+LUCI_DESCRIPTION:=Light OpenWrt set including admin support and the default Bootstrap theme
+LUCI_DEPENDS:= \
+	+IPV6:luci-proto-ipv6 \
+	+luci-app-firewall \
+	+luci-mod-admin-full \
+	+luci-proto-ppp \
+	+luci-theme-bootstrap \
+	+rpcd-mod-rrdns \
+	+uhttpd \
+	+uhttpd-mod-ubus
+
+PKG_LICENSE:=Apache-2.0
 
 include ../../luci.mk
 
--- a/collections/luci-ssl-openssl/Makefile
+++ b/collections/luci-ssl-openssl/Makefile
@@ -14,7 +14,7 @@ LUCI_DESCRIPTION:=LuCI with OpenSSL as the SSL backend (libustream-openssl). \
  OpenSSL cmd tools (openssl-util) are used by uhttpd for SSL key generation \
  instead of the default px5g. (If px5g is installed, uhttpd will prefer that.)
 
-LUCI_DEPENDS:=+luci +libustream-openssl +openssl-util
+LUCI_DEPENDS:=+luci-light +libustream-openssl +openssl-util
 
 include ../../luci.mk
 
--- a/collections/luci-ssl/Makefile
+++ b/collections/luci-ssl/Makefile
@@ -10,7 +10,7 @@ LUCI_TYPE:=col
 LUCI_BASENAME:=ssl
 
 LUCI_TITLE:=LuCI with HTTPS support (WolfSSL as SSL backend)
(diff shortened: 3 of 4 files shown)


castillofrancodamian:
An even lighter version that does not include ppp or IPv6 would be much better.

dhewg:
Thanks, jow!

@castillofrancodamian What I first didn't realize is that all collections are just meta-packages. They don't offer anything themselves, they're just depending on a specific set of other packages.

We could make up a few other variants, but I think with luci and luci-light that covers the most used cases. What you can always do is deselect luci/luci-light and then deselect e.g. ppp.

castillofrancodamian:
@dhewg Anyway my solution is not to install any metapackage, but to install `uhttpd-mod-ubus`, `luci-mod-admin-full`, `luci-app-firewall`, `luci-theme-openwrt-2020` (or `luci-theme-bootstrap`) and remove `odhcp6c`, `odhcpd-ipv6only`, `opkg`, `ppp`, and `ppp-mod-pppoe`.

f00b4r0:
> I do not quite understand why also luci-app-opkg was removed also from the luci-ssl-openssl. 

I can probably answer that

> opkg can well be used for most normal packages. 

> The attendedsysupgrade change is ok, but the opkg removal feels strange. 

In the context of minimal space usage (which is implied here) it doesn't. Consider that the devices targeted by this metapackage are typically unable to install–let alone upgrade–any package and work only by flashing a complete custom-built image. Avoiding pulling opkg deps while still allowing a functional luci is a win here. And again let's not forget we're talking about a metapackage.

dhewg:
> Not quite.

Hehe, okay, sorry for that ;)

I still believe luci is a nice fit for downloaded (IB-)snapshots, and luci-light for the compiling crowd. The latter are cmdline-capable people anyway, a missing luci app doesn't stop you from using opkg. But generally those will just include whatever they need in the image anyway I guess.

hnyman:
> luci is a nice fit for downloaded (IB-)snapshots, and luci-light for the compiling crowd. 

There is also a third group: users of popular community builds...

From the community build perspective it is a bit different. 
E.g. my R7800 build (since year 2016 about 1300 different builds so far, and about 293.000 forum thread views, 21.500 clicks to the download site link) is being used by lots of people, who may want to install additional packages in addition to the selection that I have made.

Well, I have to add the luci-app-opkg seprately. No big deal, but strange (especially regarding the openssl luci).

Ps.
> In the context of minimal space usage (which is implied here) it doesn't. 

Talking about luci-ssl-openssl in the context of minimal space usage????    ;-) 
OpenSSL adds a megabyte the build while the removal of luci-app-opkg reduces it by a few kilobytes (as the underlying opkg is still installed in any case).

dhewg:
Well, some ppl may want to remove opkg altogether. Just like I want to remove asu from my builds.
Adding something on top of luci-light is easy, but removing something is not without removing the collection itself.
So I'd still keep opkg out of it, but for the very same reasons ppp could be moved to the full collections.

This is just what I thought satisfies the most used cases, it's not set in stone nor does it stop new collections being added.

f00b4r0:
> Talking about luci-ssl-openssl in the context of minimal space usage????    ;-) 
> 
> OpenSSL adds a megabyte the build while the removal of luci-app-opkg reduces it by a few kilobytes (as the underlying opkg is still installed in any case).

You have a point: i missed that part and indeed I would agree luci-ssl-openssl doesn't belong in luci-light. As for the underlying opkg, it can be opted out without breaking luci-light, which AIUI was the goal.

f00b4r0:
> I still believe luci is a nice fit for downloaded (IB-)snapshots, and luci-light for the compiling crowd.

In fact i think luci-light is also good for IB users :)

dhewg:
> You have a point: i missed that part and indeed I would agree luci-ssl-openssl doesn't belong in luci-light.

It's the other way around: luci-ssl and luci-ssl-openssl only depend on luci-light now, so they don't pull in something a user may want to avoid.

The idea is that you can mix it as you wish:
luci-light with no ssl or wolfssl or openssl.
Or luci with no ssl, wolfssl or openssl.

In fact, I use luci-light with luci-ssl-openssl. Not because of the size, which, as pointed out, don't make sense, but because I require openssl anyway and can have ssl-enabled luci without opkg and asu.

IB/asu users have the same choice.

PR #6212, opened 2023-01-24
luci-lib-httpprotoutils: add airplay mime types

komenski:
Airplay of a statically hosted video file from one Apple device to another fails due to unrecognized content-type. Adding Airplay supported mime types fixes the issue.

Signed-off-by: Sasha Andonov <s.andonnov@gmail.com>


l3-coord:
Related to openwrt/openwrt#11871

komenski:
Related to openwrt/openwrt/pull/11871

PR #6215, opened 2023-01-29
luci: remove installation of attendedsysupgrade

aparcar:
The package is only useful when running an official build. The ImageBuilders used for this service can not contain any special compiler flags etc. used by custom builds.

Instead of installing the package whenever `luci` is installed, it should be installed via the buildbots itself, like it's currently done for `luci`.

@ynezz do you mind adding the `loci-app-attendedsysupgrade` package via the buildbots?


dhewg:
Sounds like you want luci-light? See https://github.com/openwrt/luci/pull/6211

aparcar:
Well I don't want to confuse people using a custom build

PR #6217, opened 2023-01-30
luci-mod-system: use new "rc" ubus object for init.d scripts

Ansuel:
Convert startup.js and system.js to use the generic ubus rc method to
handle /etc/init.d/ scripts for enable/disable/start/restart/reload/stop
operation.

Signed-off-by: Rafał Miłecki <rafal@milecki.pl>
[ reword commit description, convert system.js ]

---

This comes from an old patch here [1]. I just tested and looks to now work. Also I modified to add some missing part and a cleanup in the acl.d file.

@jow- I'm not really sure why we had the extra permission for uhttpd and ssh... I guess a leftover from old js pages?
I think I will add a change to rc to add support to optionally pass a name so we can return info of just one specific service... This is currently the only difference from the luci implementation... 

Also I notice that for some reason rc list is 1 second slower and lack info about the stop index... I will also investigate why it's slow... (probably the running check?) No idea if we need the stop index?

[1] https://patchwork.ozlabs.org/project/openwrt/patch/20200911112645.20101-1-zajec5@gmail.com/

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js
@@ -7,29 +7,28 @@
 var isReadonlyView = !L.hasViewPermission() || null;
 
 return view.extend({
-	callInitList: rpc.declare({
-		object: 'luci',
-		method: 'getInitList',
+	callRcList: rpc.declare({
+		object: 'rc',
+		method: 'list',
 		expect: { '': {} }
 	}),
 
-	callInitAction: rpc.declare({
-		object: 'luci',
-		method: 'setInitAction',
+	callRcInit: rpc.declare({
+		object: 'rc',
+		method: 'init',
 		params: [ 'name', 'action' ],
-		expect: { result: false }
 	}),
 
 	load: function() {
 		return Promise.all([
 			L.resolveDefault(fs.read('/etc/rc.local'), ''),
-			this.callInitList()
+			this.callRcList()
 		]);
 	},
 
 	handleAction: function(name, action, ev) {
-		return this.callInitAction(name, action).then(function(success) {
-			if (success != true)
+		return this.callRcInit(name, action).then(function(ret) {
+			if (ret)
 				throw _('Command failed');
 
 			return true;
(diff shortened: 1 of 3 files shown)


Ansuel:
Ok I found what is slowing down the rc ubus method list. The running check slow down the entire process since it seems to be a slow check... Also it's supported to only by procd init.d scripts.
To fix this:
- Add a check if script USE_PROCD and check running only in this case
- Add option to get list without checking for running state

Modification done to rc ubus method
- Added 2 new option to list method (skip_running_check and name)
- List method can now return the status of one requested name
- Fixed the list method output to also parse stop (currently it's bugged and stop at the first START or STOP)
- Limit scanning to the first 10 lines

With the following changes and luci using these option the performance are similar to luci implementation with a 0.02s penalty.

Will send these patches to patchwork once I have prepared everything.

modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json
```diff
@@ -60,17 +59,17 @@
...
 			}
 		},
 		"write": {
 			"file": {
 				"/etc/rc.local": [ "write" ]
 			},
 			"ubus": {
-				"file": [ "write" ],
-				"luci": [ "setInitAction" ]
+				"file": [ "write" ]
```
jow-, line 71: This is a syntax error (trailing comma, missing comma in the line before)

modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json
```diff
@@ -60,17 +59,17 @@
 				"/etc/rc.local": [ "read" ]
 			},
 			"ubus": {
-				"file": [ "read" ],
-				"luci": [ "getInitList" ]
+				"file": [ "read" ]
```
jow-, line 62: This is a syntax error (trailing comma, missing comma in the line before)

modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js
```diff
@@ -7,29 +7,29 @@
...
 
-	callInitAction: rpc.declare({
-		object: 'luci',
-		method: 'setInitAction',
+	callRcInit: rpc.declare({
+		object: 'rc',
+		method: 'init',
 		params: [ 'name', 'action' ],
-		expect: { result: false }
+		expect: -1
```
jow-, line 20: Remove the `expect` specification entirely.

Ansuel:
@jow- i sent the required patches for rpcd. Would be good if you take a look at them if the changes are ok for you.

Ansuel:
@jow- the rpcd dependency is now merged, this is ready... Wonder if you want to give a last check.

Ansuel:
@systemcrash a year should be enough waiting to sync all the thing. @jow- ok for you?

systemcrash:
I'm okay with it.

PR #6223, opened 2023-02-03
luci-mod-dsl: add pilot tones

janh:
This adds pilot tones to the bit allocation graph (the data is available via ubus since https://github.com/openwrt/openwrt/commit/b91d7d9d78ea12097de4c518c8cd2215ff286f7a).


janh:
Here is what the updated graph looks like:
bit-allocation-pilot-tones

jow-:
Did you test the behaviour after this change when the OpenWrt side `ltq-*-app` has not been updated yet? It's not uncommon for people to upgrade their LuCI packages

janh:
Yes, it falls back gracefully if the data doesn't contain the pilot tones (`window.json` is the data returned from the ubus call):
https://github.com/openwrt/luci/blob/3ed20aac9d086be6cec6291d0e33f19ef2d58355/modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/graph.js#L92

dhewg:
LGTM, tested with luci updated but without ltq-app, which didn't break, and then a full sysupgrade, where the lines start to show up

jow-:
Thanks for the clarification!

PR #6224, opened 2023-02-03
luci-mod-status: extend DSL metrics

janh:
This restructures the DSL metrics to make them easier to read, and adds several additional metrics, making use of https://github.com/openwrt/openwrt/commit/b91d7d9d78ea12097de4c518c8cd2215ff286f7a.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/50_dsl.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/50_dsl.js
@@ -33,23 +33,35 @@ function renderbox(dsl) {
 				_('Line State'), dsl.state || '-',
 				_('Line Mode'), dsl.mode || '-',
 				_('Line Uptime'), '%t'.format(dsl.uptime),
-				_('Annex'), dsl.annex || '-',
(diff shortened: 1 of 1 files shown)


dhewg:
Nice, I still need to try this!

This looks all good at first glance, but that DSL status box is getting rather huge.

Maybe we can move some stuff elsewhere? Network/Interfaces/DSL might be too hidden though.
Or group stats by topic and make it child boxes, like the wifi networks are on the phy box? So they float/arrange themselves based on the browser window width, that'll make it at least shrink in height.

jow-:
I'd say make a separate DSL status page and only keep the most relevant indicators (SNR & line status?) in the general overview page.

janh:
I assume it would make sense to add the separate status page to `luci-mod-dsl` instead of this one then. Should I create a separate pull request for that once it is done?

dhewg:
We could also integrate it into the dsl spectrum status and rename it accordingly.
Tested as-is, apart from the big box it works as described :)

jow-:
I merged the change for now. Would be great if you could follow up with moving extended status info into a separate view and only keeping most important indicators on the main status screen.

PR #6226, opened 2023-02-04
luci-mod-system: add support for led brightness

vtremblay:
This PR goes in pair with another one in OpenWRT (See [here](https://github.com/openwrt/openwrt/pull/11811) - reviews are welcome). 
It adds the support to configure LEDs' brightness through uci/luci.

Signed-off-by: Vincent Tremblay <vincent@vtremblay.dev>

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/leds.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/leds.js
@@ -72,6 +72,11 @@ return view.extend({
 			o.value(name)
 		});
 
+		o = s.option(form.Value, 'brightness', _('Brightness'),
+		_('The brightness is based on a virtual scale of 0 to 1000.<br />The result depends on your hardware specifications.'));
+		o.default = 1000;
+		o.datatype = 'range(0,1000)';
+
 		o = s.option(form.ListValue, 'trigger', _('Trigger'));
 		for (var i = 0; i < plugins.length; i++) {
 			var plugin = plugins[i];


openwrt-ai (commented):
Reviewed 1 new commit; the commit message matches the change.

One thing worth raising beyond the inline nits: this branch appears to be based on an old revision of `master`. The current `master` version of `view/system/leds.js` has been restructured to ES6 (`const`/`let`, array destructuring in `load`/`render`) and now also adds an `interval` option at the end of the section. The patched hunk here still reflects the older `var`-based form, so it will likely need a rebase onto current `master` to apply cleanly and to stay consistent with the surrounding (now modernized) code. This is also still a draft and is gated on the companion openwrt/openwrt change (per the description and the "depends on PR in other repo" label) that actually applies `brightness` to sysfs, so the UI option is inert without it — noted, not flagged.

Note: I was unable to load the CI status tool in this environment, so I could not ground against check results; the above is based on source inspection only.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/leds.js
```diff
@@ -72,6 +72,11 @@ return view.extend({
 			o.value(name)
 		});
 
+		o = s.option(form.Value, 'brightness', _('Brightness'),
+		_('The brightness is based on a virtual scale of 0 to 1000.<br />The result depends on your hardware specifications.'));
+		o.default = 1000;
```
openwrt-ai, line 77: nit: UCI option values are strings, and the in-tree convention is to quote default values accordingly (e.g. `o.placeholder = '50'` on the `interval` option in current master). A bare number works via coercion, but quoting keeps the default consistent with what uci returns on read.

```suggestion
		o.default = '1000';
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/leds.js
```diff
@@ -72,6 +72,11 @@ return view.extend({
 			o.value(name)
 		});
 
+		o = s.option(form.Value, 'brightness', _('Brightness'),
+		_('The brightness is based on a virtual scale of 0 to 1000.<br />The result depends on your hardware specifications.'));
```
openwrt-ai, line 76: nit: the wrapped help-text argument is indented with the same 2 tabs as the statement, whereas in-tree multi-line calls indent continuation lines one extra level (the `form.Map(...)` call earlier in this file uses 3 tabs for its continuation lines). Add one tab so the continuation lines up under the opening call.

```suggestion
			_('The brightness is based on a virtual scale of 0 to 1000.<br />The result depends on your hardware specifications.'));
```

---

openwrt-ai (commented):
Reviewed 1 new commit; the commit message matches the change.

One thing worth raising beyond the inline nits: this branch appears to be based on an old revision of `master`. The current `master` version of `view/system/leds.js` has been restructured to ES6 (`const`/`let`, array destructuring in `load`/`render`) and now also adds an `interval` option at the end of the section. The patched hunk here still reflects the older `var`-based form, so it will likely need a rebase onto current `master` to apply cleanly and to stay consistent with the surrounding (now modernized) code.

This is also still a draft and is gated on the companion openwrt/openwrt change (per the description and the "depends on PR in other repo" label) that actually wires `brightness` through to sysfs, so the UI option is inert without it — noted, not flagged.

(No CI statuses are reported on the head commit, so there was nothing to ground against on that front.)

---

PR #6227, opened 2023-02-07
[22.03] xfrpc : add xfrpc to 22.03 branch

liudf0716:
add xfrpc to 22.03 branch


BKPepe:
Related package was merged in https://github.com/openwrt/packages/pull/20430 to OpenWrt 22.03

PR #6235, opened 2023-02-09
Move DSL status to new page in luci-mod-dsl

janh:
As discussed in #6224, create a new page for DSL information in luci-mod-dsl, and change luci-mod-status to only show the most important data on the overview page.

### DIFF
--- a/modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/stats.js
+++ b/modules/luci-mod-dsl/htdocs/luci-static/resources/view/status/dsl/stats.js
@@ -1,0 +1,188 @@
+'use strict';
+'require view';
+'require dom';
+'require poll';
+'require rpc';
+
+var callDSLMetrics = rpc.declare({
+	object: 'dsl',
+	method: 'metrics',
+	expect: { '': {} }
+});
+
+function format_on_off(val) {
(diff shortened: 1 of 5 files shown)


jow-:
Code-wise LGTM. Can you attach a screen shot of the main status page portion and the full DSL status page after this change?

janh:
It looks like this:
dsl-status-overview
dsl-status-full

jow-:
Perfect, thanks!

abajk:
@janh I think the latency has the wrong units. In your example it should be 9ms instead of 0.09ms.

janh:
@abajk: The [original value in the `DSL_G997_ChannelStatusData_t` struct is in 1/100 ms](https://gitlab.com/prpl-foundation/intel/dsl_cpe_api/-/blob/ugw_8.5.2.10/src/include/drv_dsl_cpe_api_g997.h#L1317), and [the ubus code in the DSL control daemon multiplies it by 10](https://github.com/openwrt/openwrt/blob/b91d7d9d78ea12097de4c518c8cd2215ff286f7a/package/network/config/ltq-vdsl-vr9-app/src/src/dsl_cpe_ubus.c#L725). So, the value as reported via ubus is in microseconds, and dividing by 1000 as done here gives the value in milliseconds.

PR #1826, opened 2018-05-26
luci-app-firewall: expose flow offloading options

hnyman:
@nbd168 @jow- 
comments, please.

Is it sensible to expose the offloading options in LuCI?
 
I noticed on forum discussion that some people have even tried to make over-complex special LuCI apps for controlling the offloading, but now that the feature is in the normal firewall uci config in /etc/config/firewall, I see no need for a special LuCI app just for that.

Please give feedback if there should be more explanations or the proposed ones should be changed.

I did not try to craft kernel version (or offloading support) detection logic, but simply exposed the two options while mentioning that kernel 4.14 is required.

image

PR #6247, opened 2023-02-14
luci-app-firewall: make a dropdown list for flow offloading options

arinc9:
Either software or hardware offloading can be used at a time. Make a dropdown list for them to reflect this on the firewall section of LuCI.

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
@@ -72,22 +72,75 @@ return view.extend({
 		/* Netfilter flow offload support */
 
 		if (L.hasSystemFeature('offloading')) {
+			var cbiRichListValue = form.ListValue.extend({
+				renderWidget: function(section_id, option_index, cfgvalue) {
+					var choices = this.transformChoices();
+					var widget = new ui.Dropdown((cfgvalue != null) ? cfgvalue : this.default, choices, {
+						id: this.cbid(section_id),
+						sort: this.keylist,
+						optional: false,
+						select_placeholder: this.select_placeholder || this.placeholder,
+						custom_placeholder: this.custom_placeholder || this.placeholder,
+						validate: L.bind(this.validate, this, section_id),
+						disabled: (this.readonly != null) ? this.readonly : this.map.readonly
+					});
+
+					return widget.render();
+				},
+
+				value: function(value, title, description) {
+					if (description) {
+						form.ListValue.prototype.value.call(this, value, E([], [
+							E('span', { 'class': 'hide-open' }, [ title ]),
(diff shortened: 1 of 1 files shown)


arinc9:
I wanted to import the cbiRichListValue variable from interfaces.js' like this:

import { cbiRichListValue } from '../network/interfaces.js';

cbiRichListValue.prototype.renderWidget = function(section_id, option_index, cfgvalue) {
	var choices = this.transformChoices();
	var widget = new ui.Dropdown((cfgvalue != null) ? cfgvalue : this.default, choices, {
		id: this.cbid(section_id),
		sort: this.keylist,
		optional: false,
		select_placeholder: this.select_placeholder || this.placeholder,
		custom_placeholder: this.custom_placeholder || this.placeholder,
		validate: L.bind(this.validate, this, section_id),
		disabled: (this.readonly != null) ? this.readonly : this.map.readonly
	});

	return widget.render();
};

But I kept getting the "import declarations may only appear at top level of a module" error. So I currently define the variable again on zones.js but change "optional: true" to "optional: false".

arinc9:
@jow- looking forward to your review.

arinc9:
@jow- another ping regarding this

systemcrash:
I don't see any inherent problems in doing so. But does it need to be redefined just to change the boolean? Won't it inherit and override?

systemcrash:
So I tested it on v22. Works fine, but for the missing `'require ui';`. 

It does the same thing in the same amount of clicks. Is there any reason this must go in?

arinc9:
Because the current representation of the flow offloading options is wrong. As I described on the commit log, only one flow offloading option can be used at a time.

arinc9:
Thanks for picking this up @systemcrash! I've been tying up the loose ends in my life so I'm very happy to see this happen.

jow-:
Note that this invalidated the translations for the hw offloading option.

systemcrash:
I didn’t see them immediately compatible but I’ll look at updating those next. Thanks @jow-

systemcrash:
Fixed.

systemcrash:
Point to consider: some newer chipsets it's an open and shut issue. For older ones, it's (probably permanently) experimental. What is your platform @PalebloodSky ?

tiagogaspar8:
I don't have a c7 v2 but a c7 V4, and it does work well with offloading, no hardware support though.

hnyman:
Well, it was still pretty experimental 6.5 years ago, when I added those options to LuCI and added the text in #1826 commit 177224c when the then newest kernel 4.14 had brought the support for it  ;-)

No objections to removal of the experimental text.
But I would leave something about the QoS/SQM compatibility, as many people do not realize that the offloading means bypassing lots of normal packet handling.  (But I am not quite sure how much there is currently incompatibility with SQM)

systemcrash:
See 98662b0834f12bc218b339609f4f7da3088673fc

PR #6040, opened 2022-10-19
luci-app-firewall: add support for ipv6 setup

ptpt52:
Allow setup ipv6 for Port Forwards and NAT Rules

PR #6249, opened 2023-02-17
luci-app-firewall: make SNAT IPv6 capable

systemcrash:
Re-worked addLocalIPOption to return also IPv6 device addrs.

There are use cases when it makes sense for IPv6 SNAT.


systemcrash:
Ah, just saw #6040 which I think also covers SNAT.

PR #6256, opened 2023-02-21
luci-proto-modemmanager: fix acl regex to support more then 9 modem objects

feckert:
@mips171 If a modem is restarted often, the modem manager object number is increased by one. Currently, the regex is set so that the modem can no longer be selected if the object number is greater than 9. This change fixes that.


jow-:
NACK. This would allow passing arbitrary options within the `read` permission scope, including destructive ones like `-d` (disable modem) or `--pin` to mess with the SIM PIN.

Better duplicate the `"/usr/bin/mmcli -m [0-9]": [ "exec" ]` rule into `"/usr/bin/mmcli -m [0-9][0-9]": [ "exec" ]` and maybe even `"/usr/bin/mmcli -m [0-9][0-9][0-9]": [ "exec" ]`. This is not pretty but it is the safer choice.

jow-:
Merged with changes via 71b7ecb7770e11168f02e53b00195c94cee9b28e

feckert:
Thanks for looking into it

Issue #6101, opened 2022-11-16
luci-mod-network: Wireless interface name field not validated


crass:
## Steps to reproduce:

1. go to: Network → Wireless
2. Click edit button on existing wireless interface (not radio).
3. In modal dialog, go to: Advanced Settings tab in Interface Configuration section.
4. Set Interface name field to "iface678901234567890".
5. Click Save button.
6. Click Save & Apply.

## Actual behavior:
**Example: Replace the following lines and remove this**

1. Save is allowed with validation even though kernel only allows interface names of up to 15 characters.
2. All interfaces on the radio with the invalid interface name fail to come up.
3. Search in logread messages for "iface67890", there should be messages indicating that an interface name "iface6789012345" is being used.
4. There are log messages indicating that the truncated interface name fails to be created.

Example:
```
Tue Nov 15 06:45:25 2022 daemon.err hostapd: nl80211: kernel reports: Attribute failed policy validation
Tue Nov 15 06:45:25 2022 daemon.err hostapd: Failed to create interface iface67890123456: -22 (Invalid argument)
Tue Nov 15 06:45:25 2022 daemon.err hostapd: Failed to add BSS (BSSID=00:23:5c:fc:4d:c0)
Tue Nov 15 06:45:25 2022 daemon.err hostapd: Interface initialization failed
```

## Expected behavior:

LuCI should validate this field and only allow a maximum of 15 characters to be entered.

## Additional Information:
The maximum length of an interface name is 15 characters.
```
root@OpenWRT:~# iw phy phy0 interface add iface67890123456 type monitor
kernel reports: Attribute failed policy validation
command failed: Invalid argument (-22)
root@OpenWRT:~# iw phy phy0 interface add iface6789012345 type monitor
root@OpenWRT:~# iw dev iface6789012345 del
root@OpenWRT:~# 
```

OpenWrt version information from system `/etc/openwrt_release`

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.2'
DISTRIB_REVISION='r19803-9a599fee93'
DISTRIB_TARGET='mvebu/cortexa9'
DISTRIB_ARCH='arm_cortex-a9_vfpv3-d16'
DISTRIB_DESCRIPTION='OpenWrt 22.03.2 r19803-9a599fee93'
DISTRIB_TAINTS=''
```

hnyman:
Should be fixed in master with 0c1be9ae5
(possible backport to 22.03 later)

[...]

PR #6270, opened 2023-03-02
luci-mod-network: add stricter wireless interface name validation

crass:
@hnyman 
Linux wireless interface names have the following restrictions:

 * It must not be an empty string
 * It must not be '.' or '..'
 * It must not contain any /, : or space character ( , \t, \n, ...)
 * It must be less than 16 chars
 * It likely must not contain any % either

Fixes: 8673aef8db luci-mod-network: remove uciname validation from wireless interface

This is discussed in #6101 and @jow- dug up the [restrictions on the interface name](https://elixir.bootlin.com/linux/latest/source/net/core/dev.c#L995). He suggests that `%` probably can't be used either, but provides no documentation of this. I've opted to disallow `%` to err on the side of caution. The validation is done as a datatype to make it reuseable. This has not been tested as I'm currently not setup to test this. I think its important to have the validation be stricter because allowing names which the kernel will not accept is a bad user experience. Everything looks like it succeeded when changing the interface name, but the interface will not come up and the user is left wondering why, which is not obvious.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/validation.js
+++ b/modules/luci-base/htdocs/luci-static/resources/validation.js
@@ -426,6 +426,15 @@ var ValidatorFactory = baseclass.extend({
 			return this.assert(this.value.match(/^[a-zA-Z0-9_]+$/), _('valid UCI identifier'));
 		},
 
+		netdevname: function() {
+			var v = this.value;
+
+			if (v == '.' || v == '..')
+				return this.assert(false, _('Must be a valid network device name'));
+
+			return this.assert(v.match(/^[^:/%\s]{1,15}$/), _('Must be a valid network device name'));
+		},
+
 		range: function(min, max) {
 			var val = this.factory.parseDecimal(this.value);
 			return this.assert(val >= +min && val <= +max, _('value between %f and %f').format(min, max));
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1150,9 +1150,9 @@ return view.extend({
 					o.depends('mode', 'ap');
 					o.depends('mode', 'ap-wds');
 
-					o = ss.taboption('advanced', form.Value, 'ifname', _('Interface name'), _('Override default interface name'));
+					o = ss.taboption('advanced', form.Value, 'ifname', _('Interface name'), _('Override default interface name - the name can not be ".", "..", nor contain the characters :, /, %, or whitespace and must contain from 1 to 15 characters'));
 					o.optional = true;
-					o.datatype = 'maxlength(15)';
+					o.datatype = 'netdevname';
 					o.placeholder = radioNet.getIfname();
 					if (/^radio\d+\.network/.test(o.placeholder))
 						o.placeholder = '';


jow-:
Thank you for your effort @crass - unfortunately the change is incorrect due to the following reasons:

  - The wireless page uses JavaScript datatype validators in https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/validation.js while you patched the legacy Lua validator functions
  - The `!=` operator is invalid syntax in Lua, it must be `~=`
  - The `%` character in Lua match patterns is special and must be escaped using `%%`
  - The "all space characters" character class must be written as `%s` instead of `\s`
  - The match pattern is not anchored to the string end (`...$`) so it would match interface names starting with one or more legal chars but followed by an illegal one
  
My assumption about disallowed `%` came from the fact that some drivers/places in the kernel use interface name patterns such as `eth%d` upon netdev registration, leaving it to the kernel to pick the next free ordinal to form a device name. I did some tests with `ip link` now and indeed it seems that it is legal to request an ifname containing `%d` where the kernel will replace `%d` with the next free number. I don't think that this behaviour is useful though as it would result in a (non-deterministic) device name different from the one requested.

```
$ sudo ip link add name br% type bridge
RTNETLINK answers: Invalid argument
$ sudo ip link add name br%d type bridge
$ ip link
...
24: br0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether a6:b0:b4:6d:b8:54 brd ff:ff:ff:ff:ff:ff
```

Note how a sole `%` is invalid though, it *must* be followed by `d`, any other combination seems to be disallowed:

```
$ sudo ip link add name br%a type bridge
RTNETLINK answers: Invalid argument
$ sudo ip link add name br%i type bridge
RTNETLINK answers: Invalid argument
$ sudo ip link add name br%s type bridge
RTNETLINK answers: Invalid argument
$ sudo ip link add name br% type bridge
RTNETLINK answers: Invalid argument
$ sudo ip link add name br%% type bridge
RTNETLINK answers: Invalid argument
$
```

crass:
> Thank you for your effort @crass - unfortunately the change is incorrect due to the following reasons:

And thank you for looking at this. My unfamiliarity with both the source and Lua are obvious :)

>     * The wireless page uses JavaScript datatype validators in https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/validation.js while you patched the legacy Lua validator functions

Yep, did a simple search and that seemed at a glance to be where validators were defined.  Since its legacy, I should drop that change and make a change to `validation.js`, correct?

>     * The `!=` operator is invalid syntax in Lua, it must be `~=`
> 
>     * The `%` character in Lua match patterns is special and must be escaped using `%%`
> 
>     * The "all space characters" character class must be written as `%s` instead of `\s`

The three points should not apply is I switch to using a javascript validator, correct?

>     * The match pattern is not anchored to the string end (`...$`) so it would match interface names starting with one or more legal chars but followed by an illegal one

Indeed.

> 
> My assumption about disallowed `%` came from the fact that some drivers/places in the kernel use interface name patterns such as `eth%d` upon netdev registration, leaving it to the kernel to pick the next free ordinal to form a device name. I did some tests with `ip link` now and indeed it seems that it is legal to request an ifname containing `%d` where the kernel will replace `%d` with the next free number. I don't think that this behaviour is useful though as it would result in a (non-deterministic) device name different from the one requested.
> 
> ```
> $ sudo ip link add name br% type bridge
> RTNETLINK answers: Invalid argument
> $ sudo ip link add name br%d type bridge
> $ ip link
> ...
> 24: br0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
>     link/ether a6:b0:b4:6d:b8:54 brd ff:ff:ff:ff:ff:ff
> ```
> 
> Note how a sole `%` is invalid though, it _must_ be followed by `d`, any other combination seems to be disallowed:
> 
> ```
> $ sudo ip link add name br%a type bridge
> RTNETLINK answers: Invalid argument
> $ sudo ip link add name br%i type bridge
> RTNETLINK answers: Invalid argument
> $ sudo ip link add name br%s type bridge
> RTNETLINK answers: Invalid argument
> $ sudo ip link add name br% type bridge
> RTNETLINK answers: Invalid argument
> $ sudo ip link add name br%% type bridge
> RTNETLINK answers: Invalid argument
> $
> ```

Ok, so then disallowing `%` seems justified.

crass:
I've changed to using the newer javascript validation framework. Also, I've added a description of the constrains in the placeholder text for the field.

jow-:
Merged via 6ad6a241c32ebd4d62f6021a2e3cbe4e1bc192c2 - thanks!

PR #6278, opened 2023-03-05
luci-app-natmap: Update chinese translations

muink:


McGiverGim:
The translations are managed at weblate https://forum.openwrt.org/t/help-translating-openwrts-web-interface/47774

PR #6288, opened 2023-03-12
luci-proto-cni: add package

oskarirauta:
simple protocol support for *simple protocol* - cni protocol can be used with container networking, but even if it's name is cni protocol, it also works great with netavark (cni replacement for Podman)

cni-protocol is extremely useful when you want to create a firewall zone and manage firewalling of containers with openwrt's own firewall configuration, as it makes it possible to define a zone that doesn't exist yet (container not running yet?). There's not much to do with it at Luci, but with proto package, it at least won't say unsupported protocol or what ever it was that it says on occasions like that.

### DIFF
--- a/protocols/luci-proto-cni/Makefile
+++ b/protocols/luci-proto-cni/Makefile
@@ -1,0 +1,8 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Support for CNI protocol
+LUCI_DEPENDS:=+cni-protocol
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/protocols/luci-proto-cni/htdocs/luci-static/resources/protocol/cni.js
+++ b/protocols/luci-proto-cni/htdocs/luci-static/resources/protocol/cni.js
@@ -1,0 +1,8 @@
+'use strict';
+'require network';
+
+return network.registerProtocol('cni', {
+	getI18n: function () {
+		return _('CNI (Externally managed interface)');
+	}
+});


jow-:
Interesting. I think `cni-protocol` is not an ideal name for what this protocol handler actually does, it does not setup any kind of CNI or integrate with it, it merely scrapes L3 address info off an externally managed device and makes it known to netifd. This is an actually useful generic facility which is needed for any kind of scenario where a user wishes to make externally managed devices known to netifd. In theory the `none` protocol would be used for that but since it clears out any IP address info it is unsuitable for this task.

So a better name for this protocol handler would be something like `extdev-protocol` with a human readable description like "Externally managed device".

Regarding the actual LuCI change:
You don't actually need to implement all those procedures you did in your PR as device selection is handled in a generic manner, along with the usual generic interface options unrelated to the specific protocol handler. You should be able to condense your proto support file down into just that:

```javascript
'use strict';
'require network';

return network.registerProtocol('cni', {
        getI18n: function() {
                return _('CNI (Externally managed interface)');
        }
});
```

oskarirauta:
@jow-
Huge improvement to "protocol" file.
I'll first change the PR, from rest, not sure yet.

jow-:
Merged, thanks @oskarirauta !

oskarirauta:
@jow- No, thanks to you. Now I just need to find some time to write a wiki entry to ease up future container users... There's few tricks to get networking to work ideally and these packages just give it what it needs (possibility to define a firewall zone without network being present at the moment, if containers are not running..)

PR #6294, opened 2023-03-16
[22.03] luci-app-openwisp: improved schema for usability

nemesisdesign:
Without these changes, saving the form of luci-app-openwisp without doing any change generates a configuration which will not work in most cases.

In this patch I took care of providing the right settings to show the default values, avoid saving default values and show the interface names properly. With these changes it will be possible for a user to edit just 1 field without introducing breaking changes.

Signed-off-by: Federico Capoano <f.capoano@openwisp.io>
(cherry picked from commit 8727b1cbc8542e06b0b9db7e41366bbbe19d42f5)


188644:
من ایران هستم فارسی بنویس با زبان فارسی یا پرشین prsin

در تاریخ جمعه ۱۷ مارس ۲۰۲۳،‏ ۱:۲۰ Federico Capoano ***@***.***>
نوشت:

PR #6295, opened 2023-03-17
luci-mod-status: fix iptables jump on argon theme

jjm2473:


jow-:
I think we can change to `scrollIntoView()` completely without keeping the old `scrollTop` setting code.

jjm2473:
@jow- just for some old browsers may not support scrollIntoView.

jow-:
@jjm2473 - browsers not supporting scrollIntoView() will fail with other parts of the LuCI code as well. I think we can gradually fade out support for really old browsers.

jjm2473:
@jow-  I saw that luci-theme-openwrt still using `XHTML 1.0 Strict` mode html

https://github.com/openwrt/luci/blob/master/themes/luci-theme-openwrt/luasrc/view/themes/openwrt.org/header.htm

PR #6303, opened 2023-03-22
luci-app-tinyproxy: support socks5 upstream

autoexpect:
Use socks5 upstream.


jow-:
Removing the input validation is not ideal, so what is the format supposed to look like? Basically `socks5 ` followed by the usual `host:port` tuple?

autoexpect:
The socks5 protocol is similar to the following:
```
config upstream
	option type 'proxy'
	option via 'socks5 127.0.0.1:8080'
```

PR #6305, opened 2023-03-22
Disable LTO on some LuCI packages

timocapa:
Ran into linking errors on these, haven't tested with all packages enabled


timocapa:
Apology for the current bad title, I didn't have time to properly think of a good one.

dhewg:
All of those build fine here on ipq40xx with CONFIG_USE_LTO last I checked.
That arch are you building and what's the exact problem?

PR #6310, opened 2023-03-26
luci-theme-bootstrap: File input background color

arbolitoloco1:
Remove the white background color for the file input elements, this way it'll use the default color defined for input elements, matching the light/dark theme.


jow-:
Merged, thanks!

PR #6313, opened 2023-03-27
[luci-app-dockerman] Misspell fix in UI

AHOHNMYC:


feckert:
@AHOHNMYC please add valid **Signed-off-by** to your commit with your real name and a valid email address.
Also update the commit message to `luci-app-dockerman: fix spelling` , we do not use `[`luci-app-dockerman`]`.
Please have a look to the other commit messages in this repository

stokito:
The dockerman is mainly developed in the repo https://github.com/lisaac/luci-app-dockerman. But some commits from the openwrt repo missing there and verse versa. It looks like the original repo is more active.
So it would be probably better to send a PR there. And then backport commits here.

Also please rename the keys in translations *.po files. The Weblate will pull the changes.

feckert:
@AHOHNMYC Sorry, I overlooked that one. The change was already made by another commit.
https://github.com/openwrt/luci/commit/6afe7d91e5ae22008314de46e65310c9cf660edc

PR #6315, opened 2023-03-27
luci-app-transmission: update web interface path

csharper2005:
~~Depends on https://github.com/openwrt/packages/pull/20739/commits~~

---------------

Transmission 4.0 web interface files changed from /web to /public_html.

This is necessary to make "Open Web interface" button visible again.

### DIFF
--- a/applications/luci-app-transmission/htdocs/luci-static/resources/view/transmission.js
+++ b/applications/luci-app-transmission/htdocs/luci-static/resources/view/transmission.js
@@ -22,7 +22,7 @@ return view.extend({
 	load: function() {
 		return Promise.all([
 			L.resolveDefault(callServiceList('transmission')),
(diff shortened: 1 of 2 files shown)


1715173329:
I guess this should be updated too ;)
https://github.com/openwrt/luci/blob/e5a97eb551ce0fdc50d640eb7f8a1b117da1ae1a/applications/luci-app-transmission/root/usr/share/rpcd/acl.d/luci-app-transmission.json#L8

csharper2005:
> I guess this should be updated too ;)

@1715173329 thanks! fixed in https://github.com/openwrt/luci/commit/b699851ac2734a9cd0d0226e7c91973c2487f526

1715173329:
Cc @hnyman

csharper2005:
There is no interest from committers side. So I close this.

jjm2473:
reopen, please

PR #6317, opened 2023-03-28
luci-app-firewall: add masq6 option for zones

1715173329:
Allow configure Masquerading6 via LuCI interface.

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
@@ -33,6 +33,7 @@ return view.extend({
 		var ctHelpers = data[0],
 		    fwDefaults = data[1],
 		    m, s, o, inp, out;
+		var fw4 = L.hasSystemFeature('firewall4');
 
 		m = new form.Map('firewall', _('Firewall - Zone Settings'),
 			_('The firewall creates zones over your network interfaces to control network traffic flow.'));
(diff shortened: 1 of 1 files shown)


applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
```diff
@@ -157,18 +157,31 @@ return view.extend({
...
-			_('Enable network address and port translation (NAT or NAPT) for outbound traffic on this zone. This is typically enabled on the <em>wan</em> zone.'));
+		o = s.taboption('general', form.Flag, 'masq', _('IPv4 Masquerading'),
+			_('Enable network address and port translation IPv4 (NAT4 or NAPT4) for outbound traffic on this zone. This is typically enabled on the <em>wan</em> zone.'));
 		o.editable = true;
 		o.tooltip = function(section_id) {
+			var family = uci.get('firewall', section_id, 'family')
 			var masq_src = uci.get('firewall', section_id, 'masq_src')
 			var masq_dest = uci.get('firewall', section_id, 'masq_dest')
-			if (masq_src || masq_dest)
+			if ((!family || family === 'ipv4') && (masq_src || masq_dest))
```
jow-, line 167: The check is overly strict, `family` may be `inet`, `inet6`, `ipv4`, `ipv6`, `4`, `6`. I suggest to do `family.indexOf('6') == -1`

1715173329: Got it, thanks!

applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
```diff
@@ -157,18 +157,31 @@ return view.extend({
...
 			var masq_src = uci.get('firewall', section_id, 'masq_src')
 			var masq_dest = uci.get('firewall', section_id, 'masq_dest')
-			if (masq_src || masq_dest)
+			if ((!family || family.indexOf('6') == -1) && (masq_src || masq_dest))
 				return _('Limited masquerading enabled');
 
 			return null;
 		};
 
+		o = s.taboption('general', form.Flag, 'masq6', _('IPv6 Masquerading'),
```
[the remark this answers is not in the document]
1715173329: Sorry I don't even notice there's no masq6 support in fw3.. Should I also check firewall version for masq_src / masq_dest like this?

```js
if (!fw4) {
	o.depends('family', '');
	o.depends('family', 'ipv4');
	o.datatype = 'list(neg(or(uciname,hostname,ipmask4)))';
} else
	o.datatype = 'list(neg(or(uciname,hostname,ipmask)))';
```

jow-:
And one final remark; I am not sure if we should expose the `masq6` option this prominently. At the very least it should be hidden from the tabular zone overview (by removing `o.editable = true;`) and maybe it should be even moved into the advanced section.

IPv6 masquerading on a zone should be the exception (as last resort after PD and relay mode) rather than the norm.

1715173329:
> And one final remark; I am not sure if we should expose the masq6 option this prominently. At the very least it should be hidden from the tabular zone overview (by removing o.editable = true;) and maybe it should be even moved into the advanced section.

Updated and moved it into the advanced section.

jow-:
Merged, thanks!

PR #6322, opened 2023-03-30
[WIP] luci-app-macfilter: add macfilter

PolynomialDivision:


jow-:
Hmm, I would like to avoid introducing new Lua based applications. This one should be easy to convert. Please test the following code and update the PR with it if it works for you:

1) Drop the Lua controller file entirely and replace it with a static JSON menu definition:
`applications/luci-app-macfilter/root/usr/share/luci/menu.d/luci-app-macfilter.json`
```json
{
        "admin/network/macfilter": {
                "title": "MAC Filter",
                "action": {
                        "type": "view",
                        "path": "macfilter"
                },
                "depends": {
                        "uci": { "firewall": true }
                }
        }
}
```

2) Replace the Lua CBI model with a JavaScript view:
`applications/luci-app-macfilter/htdocs/luci-static/resources/view/macfilter.js`
```javascript
'use strict';
'require form';
'require view';
'require network';

return view.extend({
	load: function() {
		return network.getHostHints();
	},

	render: function(hints) {
		var m, s, o;

		m = new form.Map('firewall');

		s = m.section(form.TableSection, 'rule', _('MAC Addresses'));
		s.anonymous = true;
		s.addremove = true;
		s.filter = function(section_id) {
			var sectionObject = this.map.data.get('firewall', section_id);

			return (sectionObject.src == '*' &&
			        sectionObject.dest == 'wan' &&
			        sectionObject.proto == 'any' &&
			        sectionObject.target == 'ACCEPT');
		};
		s.handleAdd = function(ev) {
			var firstRuleSection = this.map.data.get_first('firewall', 'rule');
			var addedRuleID = this.map.data.add('firewall', 'rule');

			this.map.data.set('firewall', addedRuleID, 'enabled', '1');
			this.map.data.set('firewall', addedRuleID, 'src', '*');
			this.map.data.set('firewall', addedRuleID, 'dest', 'wan');
			this.map.data.set('firewall', addedRuleID, 'target', 'ACCEPT');
			this.map.data.set('firewall', addedRuleID, 'proto', 'any');

			// if a previous rule section existed, move the new one before it
			if (firstRuleSection)
				this.map.data.move('firewall', addedRuleID, firstRuleSection['.name']);

			return this.map.save(null, true);
		};

		o = s.option(form.Flag, 'enabled', _('Enabled'));
		o.default = o.enabled;

		o = s.option(form.Value, 'name', _('Description'));

		o = s.option(form.Value, 'src_mac', _('MAC address'));
		hints.getMACHints(false).forEach(function(hint) { 
			o.value(hint[0], hint[1]); 
		});

		return m.render();
	}
});
```

PolynomialDivision:
I know first realize that I did this PR against the official luci repository and not my own fork. O.o
Sorry. I just wanted to play around in my own luci-fork. I did not want to upstream it.

PR #6331, opened 2023-04-05
luci-app-firewall: fix the IPv6 forwards/snats view

dibdot:
* corrects the view as IPv4 and IPv6 for rules where the family is 'any' and the IP not set (this fixes #9c55500), e.g. a forward rule like that:

```
config redirect 'adblock_lan53'
	option name 'Adblock DNS (lan, 53)'
	option src 'lan'
	option proto 'tcp udp'
	option src_dport '53'
	option dest_port '53'
	option target 'DNAT'
	option family 'any'
```
**Old behavior**
Forward:
forwards_old

SNAT:
snats_old

**New behavior:**
Forward:
forwards_new

SNAT:
snats_new


efahl:
Changes tested on SNAPSHOT r22493-5a8741aee9.  Looks good to me.

PR #6335, opened 2023-04-08
luci-mod-network: add support for creating cname records

arbolitoloco1:
I used the MX and SRV tabs as templates to create a new tab where users can create cname records in dnsmasq, using the cname and target options.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -262,6 +262,7 @@ return view.extend({
 		s.tab('hosts', _('Hostnames'));
 		s.tab('srvhosts', _('SRV'));
 		s.tab('mxhosts', _('MX'));
+		s.tab('cnamehosts', _('CNAME'));
 		s.tab('ipsets', _('IP Sets'));
 
 		s.taboption('general', form.Flag, 'domainneeded',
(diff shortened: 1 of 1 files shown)


ErwanMAS:
@hnyman @jow- @Zeranoe @feckert   hey can someone take a look on this , this is very useful when you want to rewrite/shadow  public DNS . 

from DNSMASQ man pages : 

> `addresses in /etc/hosts will "shadow" different addresses for the same names in the upstream DNS, so "mycompany.com 1.2.3.4" in /etc/hosts will ensure that queries for "mycompany.com" always return 1.2.3.4 even if queries in the upstream DNS would otherwise return a different address. There is one exception to this: if the upstream DNS contains a CNAME which points to a shadowed name, then looking up the CNAME through dnsmasq will result in the unshadowed address associated with the target of the CNAME. To work around this, add the CNAME to /etc/hosts so that the CNAME is shadowed too.
`

erwan-mas-okcupid:
@dibdot @jcassette if you have time to commit this PR , i tested this on my router , and this working .

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -694,6 +695,27 @@ return view.extend({
 		so.datatype = 'range(0,65535)';
 		so.placeholder = '0';
 
+		o = s.taboption('cnamehosts', form.SectionValue, '__cname__', form.TableSection, 'cname', null, 
+			('Set an alias for a hostname.'));
```
systemcrash, line 699: Shouldn't that be like so?:
```
			_('Set an alias for a hostname.'));
```

arbolitoloco1:
Sorry yeah, I think so, just fixed it!

arbolitoloco1:
This is how it looks on my router currently, with the underscore added.
image

arbolitoloco1:
I messed up in a horrendous way but I think this is good now (unless the test build fails). Sorry for all the mess

systemcrash:
Just for my own edification, how does one push zero commits? 🤔

arbolitoloco1:
I did:
`git reset --soft HEAD~1`, where 1 means remove 1 commit
`git push origin +patch-4`

found it on [https://stackoverflow.com/questions/448919/how-can-i-remove-a-commit-on-github](url) in the middle of my desperation lol

systemcrash:
Oh, now I know 😁

Issue #4472, opened 2020-09-26
luci-mod-dashboard: request - remove wifi info panel when no wifi


Grommish:
Currently, the dashboard has a section for WiFi information, even on devices without WiFi.

Probably cosmetic, but could lead to confusion for some less-than-technical users.

jow-:
Closed by #6341

PR #6341, opened 2023-04-13
luci-mod-dashboard: hide wireless pane when there are no radios

raenye:
Implements feature request #4472.


raenye:
Thank you.

raenye:
@jow- could you also kindly close #4472?

PR #6347, opened 2023-04-16
luci-base: ui.js: avoid polling in a tight loop on confirm/rollback

erikrk:
The call function was based on the legacy XHR interface and it has not been updated to comply with the new request interface which passes the duration as part of the response object rather than as a separate argument. This resulted in polling in a tight loop in certain cases since the duration was undefined.

In addition there is no need to adjust apply_holdoff based on elapsed time as no significant amount of time has elapsed at this point.


systemcrash:
@jow- do you have any opinion on this?

systemcrash:
Thanks @erikrk - it took a while.

PR #6348, opened 2023-04-17
luci-app-openvpn-server: add this plugin

8MiYile:


feckert:
Sorry, this pullrequest is totaly messed up! Please have a look again and open a pullrequest either for luci-app-openclash or luci-app-openvpn-server. Additional you should use the new Javascript API an not the deprectated Lua API

PR #6351, opened 2023-04-19
luci-proto-wireguard: adding a DNS option to the wireguard peer config generator

nevumx:
Some clients like iOS require this explicitly, and so this change adds the appropriate config with some sensible defaults. I am happy to change the defaults if anyone thinks they should be different, or just absent.

Screen Shot 2023-09-21 at 11 52 45 PM

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -692,7 +692,7 @@ return network.registerProtocol('wireguard', {
 
 		o.modalonly = true;
 
-		o.createPeerConfig = function(section_id, endpoint, ips) {
+		o.createPeerConfig = function(section_id, endpoint, ips, dns) {
 			var pub = s.formvalue(s.section, 'public_key'),
 			    port = s.formvalue(s.section, 'listen_port') || '51820',
 			    prv = this.section.formvalue(section_id, 'private_key'),
@@ -709,6 +709,7 @@ return network.registerProtocol('wireguard', {
 				'[Interface]',
 				'PrivateKey = ' + prv,
 				eport ? 'ListenPort = ' + eport : '# ListenPort not defined',
+				dns && dns.length ? 'DNS = ' + dns.join(', ') : '# DNS not defined',
 				'',
 				'[Peer]',
 				'PublicKey = ' + pub,
@@ -728,6 +729,7 @@ return network.registerProtocol('wireguard', {
 			return Promise.all([
 				network.getWANNetworks(),
 				network.getWAN6Networks(),
+				network.getNetwork('lan'),
 				L.resolveDefault(uci.load('ddns')),
 				L.resolveDefault(uci.load('system')),
 				parent.save(null, true)
@@ -752,9 +754,19 @@ return network.registerProtocol('wireguard', {
 
 				var ips = [ '0.0.0.0/0', '::/0' ];
 
+				var dns = [];
+
+				var lan = data[2];
+				if (lan) {
+					var lanIp = lan.getIPAddr();
+					if (lanIp) {
+						dns.unshift(lanIp)
+					}
+				}
+
 				var qrm, qrs, qro;
 
-				qrm = new form.JSONMap({ config: { endpoint: hostnames[0], allowed_ips: ips } }, null, _('The generated configuration can be imported into a WireGuard client application to set up a connection towards this device.'));
+				qrm = new form.JSONMap({ config: { endpoint: hostnames[0], allowed_ips: ips, dns_servers: dns } }, null, _('The generated configuration can be imported into a WireGuard client application to set up a connection towards this device.'));
 				qrm.parent = parent;
 
 				qrs = qrm.section(form.NamedSection, 'config');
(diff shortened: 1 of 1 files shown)


protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -747,9 +749,19 @@ return network.registerProtocol('wireguard', {
 
 				var ips = [ '0.0.0.0/0', '::/0' ];
 
+				var dns = [ '1.1.1.1', '1.0.0.1' ];
```
feckert, line 752: I would not set a default value for the DNS server.

castillofrancodamian:
It would be necessary for the generator to also add the client interface IP(s).

systemcrash:
Which DNS you use is just one of those things - if you can set it to the openwrt LAN IP - I think this is a reasonable default. Put all traffic thru the VPN anyway, avoid leaks.

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -747,9 +749,19 @@ return network.registerProtocol('wireguard', {
...
+
+				var lan = data[2];
```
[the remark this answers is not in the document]
nevumx: As @systemcrash mentioned in https://github.com/openwrt/luci/pull/6351#issuecomment-1654784349, this is the most sensible default, the cloudflare DNS servers are just sensible backups.

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
nevumx: Done.

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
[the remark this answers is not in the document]
nevumx: If you would like me to remove it, I can do that, let me know.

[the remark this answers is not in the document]
systemcrash: I like that today we can choose which IP are allowed under the config export. I guess something like a check box to add the default LAN IP, I.e. The IP of dnsmasq.

Some instances might have a specific IP for the tunnel end on openwrt so these users might just uncheck a box for this suggested dns. 

Having a drop down for which IP the remote client will have would be welcome, taken from the peer config.

systemcrash:
@nevumx can you include a screenshot in the first post of how the changes currently look please?
Can the DNS server be optional, behind a boolean checkbox (if not already)?

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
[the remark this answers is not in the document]
nevumx: @systemcrash I... Think I already have that? I attached a screenshot for you to take a look at, let me know if I am missing something...

nevumx:
> @nevumx can you include a screenshot in the first post of how the changes currently look please?
> Can the DNS server be optional, behind a boolean checkbox (if not already)?

@systemcrash done.

systemcrash:
I hope to see a bit more code which fetches the e.g. LAN IP of the device and present that as a default selection for DNS server. That's the most sane default I can think of that satisfies the silent majority. 

But if we have at least a field there, that's OK. 

Some text which disambiguates this DNS field from the one higher up in the wireguard interface goes a long way also. Taking your text and building on it:

`DNS servers for the remote clients using this tunnel to your openwrt device. Some wireguard clients require this to be set.`

nevumx:
> I hope to see a bit more code which fetches the e.g. LAN IP of the device and present that as a default selection for DNS server. That's the most sane default I can think of that satisfies the silent majority.
> 
> But if we have at least a field there, that's OK.
> 
> Some text which disambiguates this DNS field from the one higher up in the wireguard interface goes a long way also. Taking your text and building on it:
> 
> `DNS servers for the remote clients using this tunnel to your openwrt device. Some wireguard clients require this to be set.`

@systemcrash I believe the code you are referring to is in this PR at lines 760-765... Am I missing something here? And I replaced the description with yours.

nevumx:
@feckert / @jow- Does this look good for merging?

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -709,6 +709,7 @@ return network.registerProtocol('wireguard', {
 				'[Interface]',
 				'PrivateKey = ' + prv,
 				eport ? 'ListenPort = ' + eport : '# ListenPort not defined',
+				dns && dns.length ? 'DNS = ' + dns.join(', ') : '# DNS not defined',
```
systemcrash, line 712: According to my tests, wireguard clients do not like unfinished lines that end with a comma `,`, so one must test for the array size and only if it is `>= 1` should one `join`.

systemcrash (changes requested):

akallabeth:
just wanted to add a small thing: this field can also contain dns domains for split DNS setups (e.g. a DNS ip and a (list of) domain name that is routed through the interface)

systemcrash:
> just wanted to add a small thing: this field can also contain dns domains for split DNS setups (e.g. a DNS ip and a (list of) domain name that is routed through the interface)

https://manpages.debian.org/unstable/wireguard-tools/wg-quick.8.en.html

`...or non-IP hostnames to be set as the interface's DNS search domains. May be specified multiple times.`

I forgot about this. But given that this is a different behaviour to what this PR is adding, I think this can be left to a separate PR. I think in such a PR, it makes sense to get domain suffixes from the interface configuration (e.g. the bridge where this is typically configured). Would it then be `config dnsmasq` or `config dhcp`?

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
systemcrash: Or as the man page details, add one `DNS = ` line per dns entry. 

The man page says `May be specified multiple times.`

systemcrash: @nevumx if you can address this, we can get this merged. 

If necessary, use nested ternaries to test for `.length > 1`

systemcrash (changes requested):

systemcrash:
Never mind, fixed.

Issue #6050, opened 2022-10-21
luci-app-wireguard: Missing option to configure peer address


manuelsteiner:
The Wireguard peer configuration section is missing the address field. I am not aware of configurations omitting the address parameter. It would be handy to set that in the peer configuration so the exported configuration as well as QR code will automatically include the parameter.

At least on Android, the Wireguard app will complain if no address is set and one has to be entered manually after scanning the QR code from the OpenWrt peer config generation page.

In case this is desired behaviour (i.e. to export configs to peers without an address and manually set it on the respective devices), please consider this issue void.
 
## Steps to reproduce:
1. Go to: Network → Interfaces → Wireguard → Edit → Peers tab → Edit a peer

## Actual behavior:
1. The option to set the `address` parameter is not present
2. As a result, the generated peer config does not contain any `address` parameter

## Expected behavior:
1. An option to set the `address` parameter for a peer in the peer configuration settings
2. The `address` parameter is included in the generated peer configs (text and QR code)

## Additional Information:
```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.1'
DISTRIB_REVISION='r19777-2853b6d652'
DISTRIB_TARGET='rockchip/armv8'
DISTRIB_ARCH='aarch64_generic'
DISTRIB_DESCRIPTION='OpenWrt 22.03.1 r19777-2853b6d652'
DISTRIB_TAINTS=''
```

castillofrancodamian:
I think the option is "Allowed IPs" in the peer settings.

manuelsteiner:
Hm but that generates the parameter `AllowedIps` for the `Peer` section and not `Address` for the `Interface section`. So that setting would indicate which packet sources are accepted from the connected endpoint. Which is how it is supposed to be used and apparently mapped correctly by the Luci interface.

castillofrancodamian:
You have to enter an IP in the `Allowed IPs` section. For example `192.168.34.2/32`.

manuelsteiner:
Well but what would I do if I want to use `Allowed IPs` for its intended purpose? Say I want to configure the peer with `Address 10.0.0.2/32` in the `Interface` section and want to set `AllowedIps 192.168.0.0/24` so the device can receive packets from that specific range?

[...]

PR #6352, opened 2023-04-19
luci-proto-wireguard: adding an Addresses option to the wireguard peer config generator

nevumx:
Some clients like iOS require this explicitly, and so this change adds the appropriate config with some sensible defaults. If the intended behavior is at all different from what was agreed upon in the issue, let me know and I will se what I can do.
Addresses issue #6050
Signed-off-by: Nicholaos Mouzourakis <nevumx@gmail.com>

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -692,7 +692,7 @@ return network.registerProtocol('wireguard', {
 
 		o.modalonly = true;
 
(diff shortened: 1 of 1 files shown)


feckert:
Please remove the merge commits and rebase your branch, so that there is only one commit with the change in your branch.

venzislav:
Please put this fix in work because it is major problem for many people !
IP field must be generated to have correct confing !

nevumx:
@feckert Like that?

systemcrash:
OK - tested on master.

venzislav:
Now from the QR or copy paste VPN works as it should do ! 
Before I must manually put the IP part that was annoying and time waste and can make mistake !
THANKS FOR THAT FIX !!!

PR #6371, opened 2023-05-03
luci-statistics: rework cpu and memory graph definitions

hippi-viking:
Simplify CPU and memory definitions, make them FreeBSD compatible

Signed-off-by: Istvan S Sz <hippi-viking@protonmail.org>

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/memory.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/memory.js
@@ -35,7 +35,7 @@ return baseclass.extend({
 						memory_buffered: { color: "0000ff", title: "Buffered" },
 						memory_cached: { color: "ff00ff", title: "Cached" },
 						memory_used: { color: "ff0000", title: "Used" },
-						memory_active: { color: "880000", title: "Active" },
+						memory_active: { color: "e00000", title: "Active" },
 						memory_cache: { color: "ff00ff", title: "Cache" },
 						memory_inactive: { color: "0000ff", title: "Inactive" },
 						memory_wired: { color: "ff0000", title: "Wired" }
@@ -58,7 +58,7 @@ return baseclass.extend({
 						percent_buffered: { color: "0000ff", title: "Buffered" },
 						percent_cached: { color: "ff00ff", title: "Cached" },
 						percent_used: { color: "ff0000", title: "Used" },
-						percent_active: { color: "880000", title: "Active" },
+						percent_active: { color: "e00000", title: "Active" },
 						percent_cache: { color: "ff00ff", title: "Cache" },
 						percent_inactive: { color: "0000ff", title: "Inactive" },
 						percent_wired: { color: "ff0000", title: "Wired" }
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/memory.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/memory.js
@@ -10,89 +10,63 @@ return baseclass.extend({
 	title: _('Memory'),
 
 	rrdargs: function(graph, host, plugin, plugin_instance, dtype) {
-		var p = [];
+		var definitions = [];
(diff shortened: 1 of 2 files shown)


hnyman:
What do you mean by "FreeBSD compatible" ?  
LuCI is just for OpenWrt, which is not FreeBSD.

hippi-viking:
I am using luci-statistics to collect and display telemetry of very different devices - some running Linux, some running FreeBSD. CPU and memory graphs from FreeBSD devices are not parsed by luci-statistics without this patch because of different nomenclature used by the OS.

systemcrash:
Is openwrt still parsed after this patch? If so I think it would be okay to merge.

systemcrash:
Welp, in my naïve view, this looks acceptable. 

I like the code reduction. Just needs verifying on openwrt :)

hnyman:
@jow- 
Any opinion on this?

jow-:
If it is confirmed to still work normally with/for OpenWrt then I'm fine with merging it.

raenye:
I tested this.
Works fine but the order of fields has changed (probably due to the use of a dictionary or however hash-maps are called in JS)
But I like the code reduction too.

Before:
image
image

After:
image
image

systemcrash:
OK, I'm satisfied with that.

hnyman:
Based on the screenshots, the display order changes are really bad.

The current orders are based on "system level interrupts" / "used" at bottom, and then having more userland/optional/cached items gradually , and "free" at the top.

The proposed change completely destroys that logic, as it appears to show them in alphabetically sorted :-(    E.g. that memory graph is crazy, as free mem is in the middle.

systemcrash:
Does not bother me one bit. Everything is colour coded. Alphabetical order is logical for even a school child.

Have you a suggestion to fix ordering in JS?

hnyman:
> Have you a suggestion to fix ordering in JS?

The current status ;-)

applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/memory.js
```diff
@@ -10,89 +10,63 @@ return baseclass.extend({
...
-						title: "Cached"
-					},
-					memory_used: {
-						color: "ff0000",
-						title: "Used"
-					},
-					memory_free: {
-						color: "00ff00",
-						title: "Free"
+			return matching.length ? { [dtype]: matching } : null;
```
raenye, line 21: I think this needs to be changed to a list, and then it'll maintain the order.

raenye: Here's a working solution (still [quadratic](https://accidentallyquadratic.tumblr.com/)).

```js
function find_instances(dtype, wanted) {
  var di = graph.dataInstances(host, plugin, plugin_instance, dtype);
  var matching = wanted.filter(w => di.includes(w));
  return matching.length ? { [dtype]: matching } : null;
}
````
Should be placed in both `cpu.js` and `memory.js`.

systemcrash: @hippi-viking could you implement the fix?

hippi-viking: @raenye thanks for the solution proposal.
@systemcrash I'll have a go at it, but TBH I'm more of a backend-guy not really fluent in javascript. (And TBH it's good enough for my use as well the way it is, but when I'll have some time I'll definitely try to make it more aesthetically pleaseing.)

raenye: (while you're at it, you should squash the commits and write a full name in the signoff line)

feckert:
@systemcrash Please write the next time why you decided to close this.

systemcrash:
Sure: @hnyman (I think) already committed changes which break this update without further changes to the PR and the original committer does not show signs of updating it, where the original intention was incidental ("add this if it's useful"). We can reopen this @hippi-viking should you decide to update this.

PR #6374, opened 2023-05-05
luci-app-pbr: update to 1.1.1-5

stangri:
* Artificial version bump as WebUI compares version number to the principal package

Principal package PR: https://github.com/openwrt/packages/pull/20930

### DIFF
--- a/applications/luci-app-pbr/Makefile
+++ b/applications/luci-app-pbr/Makefile
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_LICENSE:=GPL-3.0-or-later
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>
-PKG_VERSION:=1.1.1-1
+PKG_VERSION:=1.1.1-5
 
 LUCI_TITLE:=Policy Based Routing Service Web UI
 LUCI_DESCRIPTION:=Provides Web UI for Policy Based Routing Service.


Arunmor1510:
https://github.com/openwrt/luci/blob/10a9047c62ac71ad7804fab85a48ac429a6c1424/applications/luci-app-pbr/root/usr/libexec/rpcd/luci.pbr#L39

It seems with translation package installed (`luci-i18n-pbr-xxx`), it might appears before `luci-app-pbr` in `/usr/lib/opkg/status` so the function cannot retrieve the version of `luci-app-pbr`. Should be better if use `opkg status luci-app-pbr` directly.

stangri:
> https://github.com/openwrt/luci/blob/10a9047c62ac71ad7804fab85a48ac429a6c1424/applications/luci-app-pbr/root/usr/libexec/rpcd/luci.pbr#L39
> 
> It seems with translation package installed (`luci-i18n-pbr-xxx`), it might appears before `luci-app-pbr` in `/usr/lib/opkg/status` so the function cannot retrieve the version of `luci-app-pbr`. Should be better if use `opkg status luci-app-pbr` directly.

Thank you for the report. There were issues with checking status thru opkg command with the principal package, but maybe it would not happen with the luci package as the RPCD script is not ran immediately post-install. I'll try to fix it asap either way.

stangri:
> https://github.com/openwrt/luci/blob/10a9047c62ac71ad7804fab85a48ac429a6c1424/applications/luci-app-pbr/root/usr/libexec/rpcd/luci.pbr#L39
> 
> It seems with translation package installed (`luci-i18n-pbr-xxx`), it might appears before `luci-app-pbr` in `/usr/lib/opkg/status` so the function cannot retrieve the version of `luci-app-pbr`. Should be better if use `opkg status luci-app-pbr` directly.

Suggested fix: https://github.com/stangri/source.openwrt.melmac.net/commit/2fe548df50e766fa93086e3d770a75f116b6415e

Arunmor1510:
> Suggested fix: [stangri/source.openwrt.melmac.net@2fe548d](https://github.com/stangri/source.openwrt.melmac.net/commit/2fe548df50e766fa93086e3d770a75f116b6415e)

It's not working for me. I don't know much about grep, but it seems that `$` means match end-of-line so it's may still go wrong, like 
```
Package: luci-i18n-pbr-xxx
Version: git-23.114.65014-ea9e3ad
Depends: libc, luci-app-pbr
```
If `opkg`is not an option, maybe `grep -m1 -A1 "Package: $1$" '/usr/lib/opkg/status'`?

stangri:
> > Suggested fix: [stangri/source.openwrt.melmac.net@2fe548d](https://github.com/stangri/source.openwrt.melmac.net/commit/2fe548df50e766fa93086e3d770a75f116b6415e)
> 
> It's not working for me. I don't know much about grep, but it seems that `$` means match end-of-line so it's may still go wrong, like
> 
> ```
> Package: luci-i18n-pbr-xxx
> Version: git-23.114.65014-ea9e3ad
> Depends: libc, luci-app-pbr
> ```
> 
> If `opkg`is not an option, maybe `grep -m1 -A1 "Package: $1$" '/usr/lib/opkg/status'`?

Does this help: https://github.com/stangri/source.openwrt.melmac.net/commit/b58e30bc0dc72e181583fe74f77c6ca092e1a2d9 ?

Arunmor1510:
> Does this help: [stangri/source.openwrt.melmac.net@b58e30b](https://github.com/stangri/source.openwrt.melmac.net/commit/b58e30bc0dc72e181583fe74f77c6ca092e1a2d9) ?

Yes.

stangri:
> > Does this help: [stangri/source.openwrt.melmac.net@b58e30b](https://github.com/stangri/source.openwrt.melmac.net/commit/b58e30bc0dc72e181583fe74f77c6ca092e1a2d9) ?
> 
> Yes.

Thanks, will be fixed in an update.

PR #6390, opened 2023-05-15
luci-app-firewall: add missing ICMPv6 MLD types

jonathanunderwood:
This adds entries for ICMPv6 MLD types. This fixes the ICMPv6 MLD types to be consistent with fw4. These types were added to fw4 in this [commit](https://github.com/openwrt/firewall4/commit/e6e82a55206cf7017f26b92f7097f779161b5cac) but were omitted from the corresponding luci-app-firewall [commit](https://github.com/openwrt/luci/commit/88a016cbff7eacf3a8248bc4949904abacef6685).

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/rules.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/rules.js
@@ -325,6 +325,9 @@ return view.extend({
 		o.value('mobile-prefix-advertisement'); /* icmpv6 147 */
 		o.value('mobile-prefix-solicitation'); /* icmpv6 146 */
 		o.value('mpl-control-message'); /* icmpv6 159 */
+		o.value('multicast-listener-query'); /* icmpv6 130 */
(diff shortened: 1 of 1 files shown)


jonathanunderwood:
@systemcrash FYA - this attempts to fix what I think is an omission from your earlier PR.

jonathanunderwood:
Note that the corresponding functionality in firewall4 actually seems to be broken: https://github.com/openwrt/openwrt/issues/12623

jonathanunderwood:
> Note that the corresponding functionality in firewall4 actually seems to be broken: [openwrt/openwrt#12623](https://github.com/openwrt/openwrt/issues/12623)

Wasn't broken, just not released in 22.03 (which is a shame, would be great to get that into the next point release).

jonathanunderwood:
@dibdot @hnyman @jow- any chance we can get this merged for the next 23.05 RC? Without this change users cannot create a fully compliant IPv6 network/interface directly in luci and have to manually edit config files.

PR #5436, opened 2021-10-11
luci-mod-network: support dhcp.ipset as section

vgaetera:
Allow configuring ipset sections in the dhcp config:
https://openwrt.org/docs/guide-user/base-system/dhcp#ip_sets
This can also be combined with instance-specific IP sets:
https://github.com/openwrt/luci/pull/5402
However, it makes easier to assign domains to IP sets.

Signed-off-by: Vladislav Grigoryev <vg.aetera@gmail.com>

PR #6395, opened 2023-05-21
luci-mod-network: support dhcp.nftset section

rpigott:
This is similar to the ipset section, but uses the --nftset dnsmasq option to build nft sets instead.

---

This is pretty much a blind copy of #5436. I'm not even sure a good way to test it, but nftset is a supported key now openwrt/openwrt@d7f378796f98


dave14305:
Why is this necessary? The current LuCI “IP Sets” tab will properly create dnsmasq configuration for ipsets and/or nftsets, depending on which options are compiled in dnsmasq (currently only nftset in master/23.05 snapshots).

rpigott:
Ah alright then. I'll just close this.

PR #6398, opened 2023-05-22
Fixed minor typo

sree-86:


hnyman:
NAK.

Genuine typo, but you are not changing the original string, just the unused/untranslated source string in an unused translation. ("en" is pretty much dummy).

The original typo is in
https://github.com/openwrt/luci/blob/master/applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js#L62

PR #6399, opened 2023-05-22
Fixed minor typo: "forceing"

sree-86:


hnyman:
Thanks. 
Your commit has wrong title, package name missing, so it would have required some changes to be right.

I fixed the typo with my own commit, and also fixed the translations at the same time.
See c0478f0a45b

PR #6418, opened 2023-06-06
luci-mod-status: dmesg offsets as date/time

midaranik:
Provide date/time in dmesg output instead of offsets. Resolution goes to millis. If that is not good enough then will look to extend with a toggle: offset/date-time.


hnyman:
What is your goal here?  (You offer no commit message, no signed-off, and no explanation)

Displaying date/time instead of the "uptime from boot" seconds in the dmesg output?

Personally I feel that especially during the early boot process the current practice of "uptime from boot" is more useful in debugging.

PR #6425, opened 2023-06-08
luci-app-shadowsocks-libev: don't repeat fields

zhanhb:
The field `local_address` shows twice on column `Overview` for client components


jow-:
Merged, thanks!

PR #6430, opened 2023-06-15
luci-app-attendedsysupgrade: handle image names for Arm EFI (armsr)

mcbridematt:
This fixes an issue where luci-app-attendedsysupgrade failed to present an image to download, after the asu server had built the image.

The cause was `selectImage` returning null, because it was looking for a `combined-efi` image.

`armsr` has no legacy image types (e.g x86 'BIOS'), so all images are named `combined` only.

There is also a fix to the asu server so it will select the `generic` profile rather than by board_name, as it does for x86.
(asu pr will be linked here)


hnyman:
cc @aparcar

aparcar:
Honestly I'd rather rename the upstream images instead of introducing special cases here. I'll think about it for a second

mcbridematt:
> Honestly I'd rather rename the upstream images instead of introducing special cases here. I'll think about it for a second

I am happy to create a pull request to change the armsr image names from `combined` to `combined-efi`, if desired.

aparcar:
Do you mind creating it and we see how the discussion goes?

PR #6359, opened 2023-04-25
luci-app-uhttpd: migrate to JavaScript-based implementation

Ayushmanwebdeveloper:
Dear maintainers of OpenWrt,

I am submitting this pull request to migrate luci-app-uhttpd to a JavaScript-based implementation. The old system involves rendering pages on the router and delivering them as HTML to the browser, which results in a higher load on the embedded devices. To address this issue, the LuCI applications must be migrated to a JavaScript-based framework, which will reduce the load on embedded devices and improve performance.
https://forum.openwrt.org/t/luci-migrate-to-js-based-framework-gsoc-2023-project/155719

[2404Screencast.webm](https://user-images.githubusercontent.com/91346165/234263345-c6d080a8-f9cb-43e8-be17-325d138f9b29.webm)

Actually I can't upload the ipk file that I created using the OpenWrt build system using the same implementation as in the pull request. As shown in the attached screencast, the ipk file was successfully tested on a live OpenWrt system.

PR #6438, opened 2023-06-21
luci-app-uhttpd: fix access control list permissions

Ayushmanwebdeveloper:


feckert:
Seems to work so far.
Files can only be saved under `/etc/luci-uploads`
All other files can only be read

Maybe we can add a note that the files can only be saved under `/etc/luci-uploads`

I also have seen,  that you have changed this line from [uhttpd/uhttpd](
https://github.com/openwrt/luci/pull/6359/files#diff-935989b21cf5f26a96d4f3cbb6d0963176677c4aa9e7ceab98f8b47f47968c92L6) to uhttpd. But the `uhttpd.js` file is in the subdirectory `/www/luci-static/resources/view/uhttpd/`, so you have to revert your change so we still have to set the view to `uhttpd/uhttpd`. I could only view the page if I revert your change.

feckert:
Unfortunately, I noticed another problem! The option [enable_delete](https://github.com/search?q=repo%3Aopenwrt%2Fluci%20enable_delete&type=code) for your ported `luci-app-uhttpd` and the already checked` luci-app-firewall/ipset`, does not exist. it must be named to `enable_remove`, otherwise it will not work and the button deletes are always shown.

feckert (changes requested):
Unfortunately, I was too hasty to merge it. I'll have to take a closer look next time, @jow- sorry. @Ayushmanwebdeveloper Can you please fix the points that have been noted?

Ayushmanwebdeveloper:
> Unfortunately, I was too hasty to merge it. I'll have to take a closer look next time, @jow- sorry. @Ayushmanwebdeveloper Can you please fix the points that have been noted?

Okay! I will try to fix it by today itself, thanks for pointing the issues

Ayushmanwebdeveloper:
> Unfortunately, I noticed another problem! The option [enable_delete](https://github.com/search?q=repo%3Aopenwrt%2Fluci%20enable_delete&type=code) for your ported `luci-app-uhttpd` and the already checked` luci-app-firewall/ipset`, does not exist. it must be named to `enable_remove`, otherwise it will not work and the button `Delete` are always shown.

I fixed the issues, I added the note "Files can only be uploaded and saved to the /etc/luci-uploads directory." and fixed enable_remove.

Also I fixed "path": "uhttpd/uhttpd" Actually this was caused by a slight difference in my openwrt dev environment package feed & repo. But I'll surely take more care from the next time.

feckert:
Thanks for your work and the fixes.
I have update the commit message, so i pushed directly to master https://github.com/openwrt/luci/commit/5865d02bfd9dad34e2d3106d57675045d8eb235c

jow-:
> I added the note "Files can only be uploaded and saved to the /etc/luci-uploads directory."

Why don't you simply set the `o.root_directory` property to `/etc/luci-uploads` ? It is made exactly for this situation and should restrict  the file browser that directory. See also https://openwrt.github.io/luci/jsapi/LuCI.form.FileUpload.html#root_directory

feckert:
@jow- I don't know the background, but I think I tested this to verify the pullrequest and the files where still uploaded to `/etc` and gave me the warning that LuCI will replace the file under `/etc/uhttpd.*`. Just to understand, if I upload the files uhttpd.* to /etc/luci-upload, how are they stored and used under /etc/?

Ayushmanwebdeveloper:
> > I added the note "Files can only be uploaded and saved to the /etc/luci-uploads directory."
> 
> Why don't you simply set the `o.root_directory` property to `/etc/luci-uploads` ? It is made exactly for this situation and should restrict the file browser that directory. See also https://openwrt.github.io/luci/jsapi/LuCI.form.FileUpload.html#root_directory
> 
> I'm not really a fan of adding textual descriptions of what not to do to the ui.
Yes, the o.root_directory was set to /etc/luci-uploads by default, but I modified it to root_directory = '/' to maintain similarity with the old app. Actually, I wanted to implement it in the same way as the old app. 
But I can fix it you think it will be better.

Ayushmanwebdeveloper:
Actually, I'm trying to keep most things the same while migrating old luci apps because, if I change something in all apps, there's a good chance it might cause issues in the future. However, if any issues arise in the apps that I migrated , or if similar issues arise in other apps Or if any change is recommended. I'll always try to fix it very soon.

PR #6439, opened 2023-06-21
luci-mod-network: add PREF64 support

oskar456:
This should be merged together with
https://github.com/openwrt/odhcpd/pull/186

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -832,6 +832,13 @@ return view.extend({
 						}
 					};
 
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -832,6 +832,14 @@ return view.extend({
 						}
 					};
 
+					so = ss.taboption('ipv6-ra', form.Value, 'ra_pref64', _('NAT64 prefix'), _('Announce NAT64 prefix in <abbr title="Router Advertisement">RA</abbr> messages.'));
```
Ansuel, line 835: @jow considering these are translated string is it ok to put html stuff here?

jow-: Okay but not ideal. Embedded HTML tends to get broken by translators. It also requires extra care to be XHTML compliance.

PR #6440, opened 2023-06-21
luci-app-tinyproxy: add DisableViaHeader option

Dystopian:
Requires openwrt/packages#21424.

I don't like inverted sense of this option, especially near `X-Tinyproxy header` option. But this is the simplest way.

I also added dependency to `Via hostname` option so activating `Disable Via header` option disables it.


systemcrash:
Closing in tandem with https://github.com/openwrt/packages/pull/21424

PR #6443, opened 2023-06-25
luci-mod-network: interfaces.js: add dhcpv6_pd_min_len support

jtkohl:
Add support for setting dhcpv6_pd_min_len, which enforces a floor on the IPv6 prefix delegation length, so the administrator can conserve IPv6 address space on an interface.

(Note that I tested this manually by dumping interfaces.js into /www/luci-static/resources/view/network/
Yup, it's a shortcut...if you request it, I'll resubmit after I reset my build environment to make a true package for testing.)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -909,6 +909,11 @@ return view.extend({
 						_('Forward DHCPv6 messages between the designated master interface and downstream interfaces.'));
 					so.value('hybrid', _('hybrid mode'), ' ');
 
+					so = ss.taboption('ipv6', form.Value, 'dhcpv6_pd_min_len', _('<abbr title="Prefix Delegation">PD</abbr> minimum length'),
+						_('Configures the minimum delegated prefix length assigned to a requesting downstream router, potentially overriding a requested prefix length. If left unspecified, the device will assign the smallest available prefix greater than or equal to the requested prefix.'));
+					so.optional = true;
```
Ansuel, line 914: Think you can drop this as it's true by default?

Ansuel:
> (Note that I tested this manually by dumping interfaces.js into /www/luci-static/resources/view/network/
> Yup, it's a shortcut...if you request it, I'll resubmit after I reset my build environment to make a true package for testing.)

it's ok it's the same thing

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
jtkohl: Sure.  (I was following the other examples in the file, such as line 687 for DHCP range `start` value.)

Ansuel: Just to be sure, remove the optional line and try to save a conf from luci with that value empty

jtkohl: Yup, saving with the form empty, the written config section omits the setting.  Saving with it filled in, it writes the setting with value.  saving again with the form field empty, the written config again omits the setting.

PR #6451, opened 2023-06-30
luci-proto-modemmanager: add status page

lvoegl:
Adds a ModemManager status page to proto-modemmanager. It gets **general information**, **location** and **SIMs** for all available modems from ModemManager via `mmcli`.

Screenshot 2023-06-30 at 12-36-14 LuCI

Also uses a new helper for getting available devices as JSON output in the protocol instead of parsing them with Regex.


feckert:
Thanks!

PR #6452, opened 2023-07-02
luci-mod-network: improve descriptions for DSL annex option values

janh:
Spell out the names of the DSL standards to make the options easier to understand. For modems supporting VDSL, also mention clearly which of the options actually set the bits for VDSL2.

As the items now start with the name of an ADSL standard, use that for ordering. The generic variants including VDSL2 support stay at the top, as they should be a good choice in almost all cases.

(I think ideally it would also be nice to highlight the first few options (especially the A/L/M and the B/J option) in some way, as one of those should be the right choice in almost all cases. But I'm not sure if there is good way to do that.)

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -1509,21 +1509,27 @@ return view.extend({
 			s.anonymous = true;
 
 			o = s.option(form.ListValue, 'annex', _('Annex'));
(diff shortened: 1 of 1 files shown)


systemcrash:
See https://forum.openwrt.org/t/vdsl2-firmware-for-fritz-7530/164539/7

feckert:
Thanks!

feckert:
Also cherry-picked this to openwrt-23.05 https://github.com/openwrt/luci/commit/8343c526b4cdde6e878dc2c86a2280d1186c4cf3

PR #6458, opened 2023-07-05
luci-app-adblock: drop luci-lib-jsonc dependency

dhewg:
Those are the lua bindings, which isn't used by this package anymore.

Fixes: 0f18b873 "luci-app-adblock: release 4.0.0"


dibdot:
Superseded by e6ff6826884ecd6912bb488c21eebd500030dd47, also fix this for banIP and travelmate as well.

Issue #6419, opened 2023-06-06
luci-mod-status: channel_analysis: error in the graph


castillofrancodamian:
I understand that there are no overlapping channels on 5GHz. Here it is seen that yes.

![Screenshot_20230606_150724](https://github.com/openwrt/luci/assets/47670442/559de50c-e577-4a2f-b089-1eae045166d7)

root@archer_c60:~# cat /etc/openwrt_release
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='SNAPSHOT'
DISTRIB_REVISION='r23288-7a0ec001ff'
DISTRIB_TARGET='ath79/generic'
DISTRIB_ARCH='mips_24kc'
DISTRIB_DESCRIPTION='OpenWrt SNAPSHOT r23288-7a0ec001ff'
DISTRIB_TAINTS=''

castillofrancodamian:
@Ansuel It is my understanding that the channels do not overlap each other in 5GHz. The graph shows that it does happen with the "Fibertel WiFi950 5.8GHz" networks on channel 64 and a hidden network on channel 60.

jow-:
The "Fibertel WiFi500 5.8GHz" networks utilize multiple channels. 5GHz channels are just 20MHz wide, those networks utilize 80Mhz.

castillofrancodamian:
@jow- But they are both 20 MHz, not 80 MHz. That's why I find it weird that they appear as 20 MHz and also overlap.

jow-:
Note that only the tops of those bars are meaningful. The bottom is wider and supposed to overlap, in order symbolize neighboring channel interference.

castillofrancodamian:
@jow- So why the "TEST" network does not overlap the hidden network using channel 64 like "Fibertel WiFi500 5.8GHz" which in the graph shows that they overlap?

![Screenshot_20230607_000006](https://github.com/openwrt/luci/assets/47670442/cd1519e4-03e7-4c57-b91f-413d8750d0fb)

jow-:
Looks shifted indeed. @Ansuel

Ansuel:
@jow- actually funny enough the graph is correct... it's the channel text that is wrong... I made some improvement but I need some hint on what is better...

![image](https://github.com/openwrt/luci/assets/20289090/4b7abdbf-2c7c-47fe-8cda-b3ddb55f1d1c)

The change I did:
- Fixed the wrong offset
- Added extra blue graph to identify the channel space
[...]

PR #6461, opened 2023-07-07
luci-mod-status: channel_analysis.js: fix stale scan result.

mmohtashamirad:
When the new result is empty, including when another scan is running in the background, the table was emptied, but the graph persisted. With this change, the list will persist and missing APs will be shown faded.


Ansuel:
@mmohtashamirad change looks ok to me!

Can you by chance add some 2-3 screenshoot with a before and after?

castillofrancodamian:
[Since we are](https://github.com/openwrt/luci/issues/6419)...

mmohtashamirad:
@Ansuel , sure. 
Here you are.
this is before, that the table was not persistent. 
image
This is after:
image

jow-:
Merged, thanks!

mmohtashamirad:
My pleasure. ;)

Issue #1560, opened 2018-01-18
Firewall - Port forward: NAT Loopback not working for other than LAN Zone


bbogey:
I have multiple LAN zones. NAT loopback works only from LAN, not from other zones.
Please add an option to select from which zones NAT loopback should be working.

feckert:
@bbogey could you explain more where to add this?

jow-:
That would need to be implemented in OpenWrt/LEDE's firewall3 first before such an option could be added to LuCI.

micw:
Hello,
it's 2 years ago since this was reported. Has the state changed? I have the same issue with 19.07.1.

GrimKriegor:
@micw Here is [the link](https://bugs.openwrt.org/index.php?do=details&task_id=1645&dev=4) for this issue in the OpenWRT bugtracker.

heapxor:
so everyone ignores such a important issue?

jmou:
For anyone else searching for a solution, this appears to have been implemented with the `reflection_zone` option in `/etc/config/firewall`. I don't think it is configurable through LuCI yet. For an example see https://github.com/openwrt/openwrt/issues/8868#issuecomment-1229441623

PR #6464, opened 2023-07-12
luci-app-firewall: add reflection_zone field to forwards

jcassette:
This allows to to define multiple zones for NAT reflection rules.

Fixes: #1560


jow-:
Merged, thanks!

PR #6484, opened 2023-07-26
luci-app-mosquitto: fixes for notification options

dos1:
1. Fix notifications_local_only flag -  the option wasn't being applied at all because of a typo.
2. Fix invalid XML in option description - trying to add "notifications" option resulted in malformed XML.


jow-:
Merged, thanks!

PR #6491, opened 2023-07-31
luci-app-firewall: cleaning up outdated instructions on hardware NAT

wordlesswind:
The MT7621 is no longer the only SoC that supports hardware NAT, so the description has been removed.

BTW, is there a detailed list of hardware NAT support?


jow-:
Merged, thanks!

PR #6497, opened 2023-08-03
luci-app-uhttpd: Remove LUA_TARGET

stokito:
The app was migrated to JS so the LUA_TARGET is not needed. Also remove LUCI_PKGARCH:=all which is anyway a default.

Same for luci-proto-vpnc


stokito:
done

PR #6498, opened 2023-08-04
luci-mod-status: add support for the 6g band

dhewg:
Currently, results from the 2.4 and 6 ghz bands get mixed together, because internally the channel number is used for some checks.

This adds some checks to prevent that, and then adds a new 6g band tab.

This works on my end, but my AP is the only 6g one nearby, so I don't have others in my scan results. Can someone test that please?

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
@@ -205,36 +205,38 @@ return view.extend({
 				scanCache[results[i].bssid].data.stale = false;
 			}
 
-			if (scanCache[local_wifi.bssid] == null)
-				scanCache[local_wifi.bssid] = {};
+			if (band + 'g' == radio.dev.get('band')) {
+				if (scanCache[local_wifi.bssid] == null)
+					scanCache[local_wifi.bssid] = {};
 
-			scanCache[local_wifi.bssid].data = local_wifi;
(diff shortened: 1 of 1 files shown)


mj22226:
@dhewg I tested, scan works, 6 ghz band is less crowded.

Screenshot_20230805

dhewg:
Nice, thanks!
The channel width looks wrong though, could you please run `ubus call iwinfo scan '{"device":"radioX"}'` (where X=your 6g radio), fish out the 6g nets and paste them here?

dhewg:
Well, maybe those widths are okay. At least my local 80mhz wide network shows up correctly.
This is a nice improvement, so let's merge and fix anything up if issues pop up?

Ansuel:
@dhewg we should think of a solution for those channel number... Any idea?

dhewg:
The shot above may be an extreme example. It depends on the country code and the available channels. It looks like this for me:
6g

Ansuel:
@dhewg mhhh ok but we also need to handle that case... I think a 2 line implementation is the simplest one and would solve the thing... and maybe guard it and enable this only when then chan number is > of a specific value?

I can do some test if you don't have time with handling this

dhewg:
That would be nice, js isn't my strong suit...

Ansuel:
@mj22226 can you create an issue about this adding your image and tagging me so we can track this and I won't forget to improve this?

Ansuel:
@dhewg nit last commit can you improve the commit description? Something like "Enable support for 6GHz band now that code supports it"

PR #6500, opened 2023-08-04
Move static leases 'ignore' to leasetime

systemcrash:
The underlying dnsmasq init script conflates ignore into leasetime better.

Ignore fits better into the chronological field than the IP field.


jow-:
So in order to ignore a host I now have to fill in the MAC, leave the IP empty and set the lease time to "ignore"? Feels uninituitive to me

systemcrash:
It's my understanding that ignore, lease time, must be the last argument in
the dnsmasq init script. If that is not the case, then perhaps things can
remain as they are.

PR #6502, opened 2023-08-05
luci-mod-network: Adjust DHCP/DNS tab order

hnyman:
Adjust the order of tabs in the DNS/DHCP page.
Based on estimated importance of tabs
* move common (old) tabs to be earlier
* move the less frequently used (new) tabs later

Before:
image

New:
image

@jow- 
I actually think that we might prefer to reorganise/group some tabs e.g the new DNS fields (MX, CNAME, SRV) to be subtabs. Possible?

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -254,16 +254,16 @@ return view.extend({
 		s.addremove = false;
 
 		s.tab('general', _('General Settings'));
-		s.tab('relay', _('Relay'));
-		s.tab('files', _('Resolv and Hosts Files'));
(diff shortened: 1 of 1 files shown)


systemcrash:
Call me human but... Alphabetical order would be easiest and just plain practical. I never really know the exact name of what I'm looking for anyway, so I just go spatially and find what I'm looking for. So order isn't so important. 

But do NOT bury tabs within tabs. More steps to reach destination. Ideally max 3 clicks to destination.

jow-:
It's far too many tabs, also some really fringe use cases are overrepresented by having their own tab, like MX or SRV records. MX, SRV, CNAME and Hostnames (which is essentially A/AAAA) should be combined into one common DNS tab

jow-:
Merged as an iterim solution, thanks!

systemcrash:
Was also thinking of having several of the dns type tabs merged into a
single tab. An acceptable solution.

PR #6504, opened 2023-08-06
Remove old BROKEN luci-mod-admin-mini and luci-lib-rpcc

hnyman:
Clean the repo by removing ancient BROKEN stuff:

**luci-mod-admin-mini:**
Remove the old mini version that has been marked BROKEN at least since 2015 when LuCI build system was reorgnised in a major way. The module is not compatible with the current stuff.

**luci-lib-rpcc:**
Remove the old library that has been BROKEN since 2015.

@jow- 
Your thoughts about removing these two outdated apps that have been unusable for at least 8 years?

(I couldn't even fully figure out when compiled packages have been downloadable for the last time. Backfire 10.03?)


jow-:
LGTM, thanks!

PR #6513, opened 2023-08-10
docs: Synchronize with Wiki

### LINKED ISSUE #2360: documentation: Migrate documentation/*.md to wiki
In the /documentation folder are present some markdown files with HowTos. The same information is stored on luci wiki. There is some divergences like something is converted from Wiki format to Markdown and something isn't. Also in some places information seems more old that in other source.
So I tried to synchronize them and now Wiki contains latest version.
Please review those articles and remove the documentation from sources to avoid collisions.

stokito:
The Wiki on GitHub has newer version of documentation. Still we may have references to the docs folder.
Update all pages but also add a notice:

> See [online wiki](https://github.com/openwrt/luci/wiki/) for latest version.

The LAR page is removed because there no any mentions of it anywhere in Luci.

This closes #2360


jow-:
Merged, thanks!

PR #6518, opened 2023-08-13
luci-app-mjpg-streamer: fix and migrate to JavaScript-based implementation

Ayushmanwebdeveloper:
https://github.com/openwrt/luci/assets/91346165/e90ff3b5-b622-4cfa-84e3-eae62e5ed778

Hi! I've completed the migration of luci-app-mjpg-streamer to JavaScript. All the features are functioning. When I tried the old app, it was not working as it can't get the port. So, I fixed it too & now everything works.


andibraeu:
@jow- this app looks good to me, and it's now working, as the port is fixed now

jow-:
Merged via 1b8e6e01b7225ea7a43487a79bc881d8b58cc639

PR #6537, opened 2023-08-23
luci-mod-firewall: Expand on naming of forwarding rule inside the zone

jonas2515:
Apparently the "Forward" entry of the individual firewall zones controls forwarding within the zone (between the individual interfaces) only, and not the forwarding of packets from the zone to other zones. This is quite confusing, as the meaning is different from the global "Forward" option above, which does control forwarding between zones.

Quote from user jow on the forum:
> The per-zone forward controls forwarding traffic among the ifaces of this
> zone. Traffic from/to other zones is handled by the global forward policy,
> or individual forwardings or rules.

See https://forum.openwrt.org/t/likely-bug-in-openwrt-firewall-rule-generation/18152

Let's try to be a bit more concise with the naming here and rename this entry to "Intra-Zone Forward", which hopefully makes the difference clear.

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
@@ -144,7 +144,7 @@ return view.extend({
 		var p = [
 			s.taboption('general', form.ListValue, 'input', _('Input')),
 			s.taboption('general', form.ListValue, 'output', _('Output')),
-			s.taboption('general', form.ListValue, 'forward', _('Forward'))
+			s.taboption('general', form.ListValue, 'forward', _('Intra zone forward'))
 		];
 
 		for (var i = 0; i < p.length; i++) {


jonas2515:
Marked as draft because I haven't tested this...

Also I'm not too happy with the new name for the entry, better suggestions are welcome!

Maybe it would make more sense to rename the global option to "Forwarding between zones" and this one to "Forwarding within zone"?

systemcrash:
Intra and Inter have worked fine for aeons. Please rebase your changes onto current master.

jow-:
Please spell it `Intra zone forwarding` or `Intra zone forward` to keep casing somewhat consistent with the other option labels.

jonas2515:
Thanks for the reviews, rebased and renamed to "Intra zone forward".

systemcrash:
@jonas2515 Please fix your signed-off-by in the commit.

jonas2515:
Ah sure, added the signed-off-by!

PR #6544, opened 2023-08-28
luci-app-adblock-fast: initial commit

stangri:
* Depends on https://github.com/openwrt/packages/pull/21943


applications/luci-app-adblock-fast/htdocs/luci-static/resources/view/adblock-fast/overview.js
```diff
@@ -0,0 +1,406 @@
...
+	},
+	// https://stackoverflow.com/a/14919494
+	humanFileSize: function (bytes, si = false, dp = 1) {
+		const thresh = si ? 1000 : 1024;
+
+		if (Math.abs(bytes) < thresh) {
+			return bytes + " B";
+		}
+
+		const units = si
```
jow-, line 27: You should be able to reimplement this function as:

```
humanFileSize: function (bytes, si = false, dp = 1) {
    return `${si ? 1000 : 1024}.${dp ?? 0}mB`.format(bytes);
}
```

stangri: @jow- Thank you for the suggestion, but I'd like to keep the option of displaying kB as well as MB, as some allow-lists can be just a few kB big and might look weird when size is rounded up even to 2 decimal points in MB. Also, I've just learnt about the string interpolation in js, but I feel that the division is missing from your example. I'm all for code compactness, so I'll try to play with doing this thru string interpolation in the future release.

jow-: The `(string).format()` function is LuCI specific and a roughly C-sprintf() like implementation. The above expression will dynamically construct a format string and apply the bytes argument.

Here's a few examples from the debug console:
image

PR #6546, opened 2023-08-29
luci-app-https-dns-proxy: rewrite in javascript

stangri:
While I'm waiting for the feedback from testers I wanted to create this PR to get more eyes on the new javascript code.

@jow- if you can update the i18n scan tool to parse json files in `applications/luci-app-https-dns-proxy/root/usr/share/https-dns-proxy/providers/` I'd be grateful. The localizable entries are inside the `title` and `description` properties like you've advised.

PS. The version is set to match the version of principal package.


stokito:
The list of DOH providers would be better to extract to a separate package. Then it would be easier to upgrade and less distraction for Luci team. Also the package with DNS DoH registry may be used by other software.
Here @dibdot already made one https://github.com/dibdot/DoH-IP-blocklists
Maybe we can ask to adopt this json files.

Some lists of public DNS services that support DoT/DoH:
* https://github.com/curl/curl/wiki/DNS-over-HTTPS
* https://dnsprivacy.org/wiki/display/DP/DNS+Privacy+Public+Resolvers
* https://kb.adguard.com/en/general/dns-providers
* https://github.com/wranders/doh-list

@stangri the Keenetic routers also have [UI for DOH](https://help.keenetic.com/hc/en-us/articles/360007687159-DNS-over-TLS-and-DNS-over-HTTPS-proxy-servers-for-DNS-requests-encryption). Maybe it can be interesting or inspiring for you.

stangri:
> Some lists of public DNS services that support DoT/DoH:

Thanks, I'll try to go over that list and add missing providers to the package after the current changes have been merged.

> The list of DOH providers would be better to extract to a separate package. Then it would be easier to upgrade and less distraction for Luci team.

**If** there's an interest from other packages maintainers to share the list of DoH providers, I may consider separating them into a different package. For now, I'd prefer to keep everything in the single package and I can quickly merge providers updates myself.

castillofrancodamian:
My ISP doesn't allow other DNS servers to be used, but I can use one of the NextDNS addresses on port 5353. How can I use it as a "Bootstrap DNS"? I have tried the option in red by writing the IP of the server with the # character and then : with port 5353 and none of it works.

Screenshot_20230904_164641

stangri:
> My ISP doesn't allow other DNS servers to be used, but I can use one of the NextDNS addresses on port 5353. How can I use it as a "Bootstrap DNS"? I have tried the option in red by writing the IP of the server with the # character and then : with port 5353 and none of it works.

This is not in any way related to the PR you've commented on. If you believe you need to change the bootstrap DNS for any provider you could use your ISP DNS. If you want to discuss it further, please post in the support thread at OpenWrt forums.

zhanhb:
The provider `dns.tuna.tsinghua.edu.cn` is removed. Can I know the reason?

PR #6547, opened 2023-08-29
luci-mod-network: Update wireless.js to clarify Distance Optimization

### LINKED ISSUE #2784: luci-mod-network: Clarify distance parameter is only needed when using a long-distance connection
As per https://forum.openwrt.org/t/openwrt-18-06-2-caused-havoc-with-5ghz-on-my-archer-c7-v2/32838/10?u=felagund the distance parameter in advanced wireless settings shouldn't be set unless the distance in question is measured in kilometres instead of metres. I think the information text below the parameter 'Distance to farthest network member in meters. ' should be set to something conveying this information like for example: 'Distance to farthest network member in meters. Set only for distances above one kilometer.'

Screenshot from 2019-06-23 09-01-28

felagund:
Fixes #2784

Right now, it is not clear that most users in indoor settings should stay clear off this setting and people struggle with it as evidenced in the linked issue.


systemcrash:
@jow- @hnyman I think this should go in - just a string change.

systemcrash:
Can you fix (remove one) of your `Signed-off-by: ` lines in the commit message please?

felagund:
I did not figure out how to change the commit message (tried to use --amend but then it wanted me to rebase and then it started to throw conflicts) so I made a new request: https://github.com/openwrt/luci/pull/6734

PR #6549, opened 2023-08-30
luci-network-interfaces: Add IPv6 lifetime options

systemcrash:
to the bottom of interface IPv6 settings tab

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -956,6 +956,18 @@ return view.extend({
 					so = ss.taboption('ipv6', form.Flag, 'ndproxy_slave', _('NDP-Proxy slave'), _('Set interface as NDP-Proxy external slave. Default is off.'));
 					so.depends({ ndp: 'relay', master: '0' });
 					so.depends({ ndp: 'hybrid', master: '0' });
+
+					so = ss.taboption('ipv6', form.Value, 'preferred_lifetime', _('IPv6 Prefix Lifetime'), _('Preferred lifetime for a prefix.'));
+					so.optional = true;
+					so.placeholder = '12h';
+					so.value('5m', _('5m (5 minutes)'));
+					so.value('3h', _('3h (3 hours)'));
+					so.value('12h', _('12h (12 hours - default)'));
+					so.value('7d', _('7d (7 days)'));
+
+					//This is a ra_* setting, but its placement is more logical/findable under IPv6 settings.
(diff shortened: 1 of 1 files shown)


systemcrash:
ping @jow- @hnyman 

See [forum topic](https://forum.openwrt.org/t/ipv6-lease-time-only-work-with-12h/170085)

hnyman:
With the experience from 504bdb2 (from you), that was fixed with b7dd10d,  I don't like the "optional" settings to have numeric default values. They may get accidentally written to the config file even if the user has not modified them (as LuCI finds a value there).  Discussion in https://forum.openwrt.org/t/ipv6-works-on-clients-with-default-config-not-with-default-config/94460

You might define "placeholder" value instead.

And you did not define "optional" for ra_useleasetime ?
Based on odhcpd sources, the default is actually "disabled", unlike this PR claims?
https://github.com/openwrt/odhcpd/blob/master/README#L135-L137

systemcrash:
> You might define "placeholder" value instead.
> 
> And you did not define "optional" for ra_useleasetime ? Based on odhcpd sources, the default is actually "disabled", unlike this PR claims? https://github.com/openwrt/odhcpd/blob/master/README#L135-L137

Good catches - I recall the instance. I've instead added some suggestions to choose. 

Fixed.

timur-davletshin:
Shouldn't this be backported to current 23.05 and 22.03 branches too? I see this question about DHCPv6 lease time asked several times.

Neustradamus:
@systemcrash, @hnyman: Have you seen the @timur-davletshin comment?

feckert:
Backported to openwrt-23.05 via https://github.com/openwrt/luci/commit/257f54cb8bcd493d9be0a45a3c316668b793e8ae

Neustradamus:
@feckert: Thanks for 23.05! :)
About 22.03 branch, it is possible to add too?

feckert:
@hnyman @Neustradamus With backporting this change to openwrt-22.03, I have now read through the thread in the forum again. If I have understood correctly, the option `ra_useleasetime `must be set, so that `preferred_lifetime` works. This means we need a dependency for `preferred_lifetime` to` ra_useleasetime`?
But that is not the case! Please clarify.

systemcrash:
> Please clarify.

I thought through this carefully when I made the PR:

`ra_useleasetime` =  When set, the configured *DHCPv4* leasetime is used *both as limit* for the *preferred* and *valid lifetime* of an IPv6 prefix. 

BUT

`preferred_lifetime` = Value for the preferred lifetime for a prefix. i.e. IPv6 Prefix preferred lifetime. 

Summary: Check `ra_useleasetime` to use the DHCPv4 field `leasetime` - UNCHECK it to use the DHCPv6 `preferred_lifetime`.


In this case, this PR has revealed two mutually exclusive options.

systemcrash:
See also [this odhcpd commit](https://git.openwrt.org/?p=project/odhcpd.git;a=commitdiff;h=3bda90079ec5574ef469e2a7804808302f17769d;hp=b75bcad7bd5fd03f64011a532b9960d78e4aac22).

PR #6565, opened 2023-09-05
luci-mod-status: fix iptables not showing comments

jjm2473:
fix  comments disappeared after this commit https://github.com/openwrt/luci/commit/b3ef7a17eacc6ce2f2385bbb517c67906dd2a753

@jow-


jow-:
Merged, thanks!

PR #6570, opened 2023-09-07
luci-mod-status: Hide Active DHCP Leases table(s) if there are no active leases

CharlesMengCA:
If the DHCP Server is disabled, showing empty lease table on status page will be redundant.


CharlesMengCA:
After this changes with no DHCP leases. 

image

jow-:
Please only hide the table if the server is disabled, not if there's no leases yet.

PR #6579, opened 2023-09-19
Fix 'instances' fields for dnsmasq (dhcp) config entries.

systemcrash:
Fix an error wherein luci erroneously saved the iterator integer of the current dnsmasq config object to a host (and boot/PXE) config entry 'instance' field, instead of correctly referring to its name.

Now we use the correct ".name" field of the dnsmasq config entry. Anonymous entries have e.g. "cfg01411c". The ".name" field corresponds to 'myName' in /etc/config/dhcp entries of:

config dnsmasq 'myName'
  ...

In this way, host and other entry types are bound correctly to specific dnsmasq instances.

Looks like:

Screenshot 2023-09-19 at 17 34 48

Tested on 22.03.4 ( on f73abffa2aa64a91e6c16ddae7966a5c35e73fe1 )

See forum discussion [here](https://forum.openwrt.org/t/dnsmasq-might-be-broken/171255)

FAO: @jow- @hnyman


jow-:
Please make the caption translatable and provide a translation context hint:
`_('%s (Name: %s, Domain: %s, Local: %s)', 'dnsmasq instance description value').format(...)`

systemcrash:
Yessir. Done.

systemcrash:
LGTM

systemcrash:
ping @jow- @stangri @hnyman

stangri:
@systemcrash My 2 cents: while it's certainly an improvement and the code looks very neat, would it not be better to instead do something like: `Instance: dnsmasq[0]` or `Instance: schaft` and if domain and local are unset do not show them at all, rather than: `Domain: unset` and `Local: unset`?

systemcrash:
> @systemcrash My 2 cents: while it's certainly an improvement and the code looks very neat, would it not be better to instead do something like: `Instance: dnsmasq[0]` (for unnamed instances) or `Instance: schaft` (for named ones) and if domain and local are unset do not show them at all, rather than: `Domain: unset` and `Local: unset`? The `cfg01411c` doesn't mean much to users and I don't even know how would one translate it to the dnsmasq unnamed instances indexes.

Valid points! The first one I might steal. Although I dint think it's a bad idea that users see what's going on also under the hood. I'll give those a try and see how they look. Thanks for the input.

jow-:
You also need to make `noname` and `unset` translatable (with some added meaningful translation context)

systemcrash:
That should satisfy everything, I think.

PR #6582, opened 2023-09-21
luci-app-attendedsysupgrade: add x86 efi/bios case

aparcar:
x86 is the only target that allows you to install either EFI or BIOS images, thereby add an extra check for that.


aparcar:
This supersedes both https://github.com/openwrt/luci/pull/6430 and https://github.com/openwrt/openwrt/pull/12963

Hey @mcbridematt  I'm sorry for the delay, could you please test this?

mcbridematt:
Tested OK here.

PR #6584, opened 2023-09-22
luci-app-statistics: memory: make hiding 'free' configurable

hnyman:
Make hiding/showing the 'free' memory configurable.
Set default as disabled, not hidden.

* If enabled, the graph scales to actually used memory, enabling a more detailed view to memory usage.
* If disabled, the graph always shows the whole memory / 100%.

Example:
Enabled:
image

Disabled, default (and the only choice previously):
image

Maintainer: me

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/memory.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/statistics/rrdtool/definitions/memory.js
@@ -5,12 +5,14 @@
 
 'use strict';
 'require baseclass';
+'require uci';
 
 return baseclass.extend({
 	title: _('Memory'),
 
 	rrdargs: function(graph, host, plugin, plugin_instance, dtype) {
 		var p = [];
+		var hide_free = uci.get("luci_statistics", "collectd_memory", "HideFree") == "1" ? true : false;
 
 		var memory = {
 			title: "%H: Memory usage",
@@ -21,7 +23,7 @@ return baseclass.extend({
 			data: {
 				instances: {
 					memory: [
-						"free",
+						...(hide_free ? [] : ["free"]),
 						"buffered",
 						"cached",
 						"used"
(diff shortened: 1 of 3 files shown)


hnyman:
The option could also be "HideFree" so that default (disabled) would be the old view with free RAM showing. And the user could then hide the free mem.

It is also questionable, which alternative is better view as default, as there are two main uses:

* Do I have lots of free mem or is the router RAM constrained?  --> Showing also free RAM to get the whole RAM is better.
* Do I have memory leak?  --> Hiding RAM is better, as the actually used memory scale gets amplified in graph.

Best would be to not use any uci option in config file at all, but just have a live checkbox on the page to toggle showing free RAM. But that exceeds my javascript skills :-(

raenye:
`HideFree` sounds better to me personally.
But why use uci at all in `rrdtool/definitions/memory.js`? can't it be done similar to the percent/absolute distinction?

hnyman:
> similar to the percent/absolute distinction?

The percent / absolute is actually a really heavy choice, as it is about data collection and database generation phase itself in collectd.

> why use uci at all in rrdtool/definitions/memory.js?

Well, like I wrote above, I would like to avoid that and have just display toggling...

> > Best would be to not use any uci option in config file at all, but just have a live checkbox on the page to toggle showing free RAM. But that exceeds my javascript skills :-(

raenye:
> Well, like I wrote above, I would like to avoid that and have just display toggling...

I guess a hacky but reasonable fix would be to add support for click events (per graph, or perhaps per tab).
Something not too far from this perhaps? https://github.com/openwrt/luci/pull/6619

Would it work for you if every time it starts with free mem, but on click it changes to hiding free mem?

hnyman:
I changed the option to HideFree.

> Would it work for you if every time it starts with free mem, but on click it changes

I would prefer a clickbox to enable clear toggling. Toggling by just clicking the graph might also be viable, but would require a clear visible explanation.

PR #6588, opened 2023-09-23
luci-app-minidlna: move album covers to advanced tab

stokito:
Also remove o.default = o.enabled

I have no idea why anyone may want to change the cover file name. The big list looks confusing so it's better to move it to advanced tab.


jow-:
Merged, thanks!

PR #6591, opened 2023-09-24
Luci dependencies  (luci-collection-xxx reorganisation)

stokito:
Currently the luci package depends on uhttpd.
But Turris and Gl.inet routers using Lighttpd.
In the TurrisOS there is a patch to [Drop uhttpd dependency](https://gitlab.nic.cz/turris/os/build/-/blob/f28b2b9d76a80a7ba185707faa2908af6ffc285b/patches/luci/branding/0001-Drop-uhttpd-dependency.patch) from the Luci.

So the PR removes dependency of the luci to uhttpd.
But if you want to install luci and uhttp then you still can do that by installing luci-light package.
See commit messages for details.

Another one problem is that the luci package installs /www/index.html that redirects to /cgi-bin/luci/.
But firmware manufactures usually have their own admin panel.
Because of this they have to use Lighttpd alias to override path for the index.html.
So my proposition is to create a separate package luci-www that installs the index.html.
It should be installed by default but manufactures can disable the package.

After this PR the next will be to add [luci-lighttp](https://gitlab.nic.cz/turris/os/build/-/issues/406)

CC @BKPepe 
CC @dhewg you were last who reworked the luci and luci-light packages. Please have a look


BKPepe:
Thanks for mention, but since I am no longer working with Turris, you should rather reach them. Thus ping @turris-cz and especially @miska . Because in the past once I mentioned them, they are not responding, you should contact them by e-mail packaging@turris.cz or support/Twitter.

stokito:
I see that Turris team added the issue to backlog https://gitlab.nic.cz/turris/os/build/-/issues/406

Also we may ping GL.iNet team @gl-inet and maybe @linmq @gakki-kareshi @gl-zhaojianhui

jow-:
The `luci` meta package is an empty convenience meta package to select everything needed to install a standard LuCI on vanilla OpenWrt. If you need a different component selection, don't select `luci` but pick desired components manually.

Issue #6592, opened 2023-09-24
luci-proto-wireguard: Peer "Generate configuration" incorrectly formats IPv6 Endpoint


Kasoo:
## Steps to reproduce:

1. Go to Interfaces » vpn » Edit peer » Generate configuration
2. Select IPv6 address on "Connection endpoint"
3. Note generated config uses incorrect form of `Endpoint = fc00::1:51820` instead of `Endpoint = [fc00::1]:51820`
4. Attempt to scan generated QR code using Wireguard app and receive an "Invalid configuration" error.

## Additional Information:
`luci-proto-wireguard` version: `git-23.073.81399-59a6f06`

jow-:
Fixed by #6593

PR #6593, opened 2023-09-24
luci-proto-wireguard: Escape IPv6 endpoints with [] in generated wireguard config

Kasoo:
If the chosen wireguard Endpoint is an IPv6 address, we currently generate invalid format, as IPv6 addresses with a port appended need escaping with `[]`

Fixes issue #6592


jow-:
Merged, thanks!

Issue #734, opened 2016-05-31
LuCI issues with lighttpd


bilik:
Router model: TP-Link TL-WR1043ND. 
After change uhttpd => lighttpd.
Every time I use LUCI crash with mesage.
thanks in advance
Everything another is OK (php, html, scripts, rrd, ...)

/usr/lib/lua/luci/util.lua:587: Unable to establish ubus connection
stack traceback:
    [C]: in function 'assert'
    /usr/lib/lua/luci/util.lua:587: in function 'ubus'
    /usr/lib/lua/luci/dispatcher.lua:321: in function 'dispatch'
    /usr/lib/lua/luci/dispatcher.lua:168: in function </usr/lib/lua/luci/dispatcher.lua:167>

bilik:
/usr/lib/lua/luci/dispatcher.lua:433: Failed to execute function dispatcher target for entry '/'.
The called action terminated with an exception:
/usr/lib/lua/luci/util.lua:587: Unable to establish ubus connection
stack traceback:
	[C]: in function 'assert'
	/usr/lib/lua/luci/dispatcher.lua:433: in function 'dispatch'
	/usr/lib/lua/luci/dispatcher.lua:168: in function </usr/lib/lua/luci/dispatcher.lua:167>

jow-:
Make sure luci runs as root.

bilik:
But I can not run web server (lighttpd) as root. Is any another choice?

Rom4nB:
I have the same issue with the same router.

Is there anything new about this?

jow-:
What same issue? That LuCI is not running as root?

Rom4nB:
No, that luci won't start but show the message
/usr/lib/lua/luci/util.lua:587: Unable to establish ubus connection

I tried installing OwnCloud using this howto
https://wiki.openwrt.org/doc/howto/owncloud

It works fine until the part with "disable httpd and use luci on lighttpd". I could not manage to get luci working again, I only got the same error message as blllk above.
/usr/lib/lua/luci/dispatcher.lua:433: Failed to execute function dispatcher target for entry '/'.
The called action terminated with an exception:
/usr/lib/lua/luci/util.lua:587: Unable to establish ubus connection


Finally I used 
firstboot

to regain the default settings and luci works again.

hurenkam:
The following (workaround) fixes this issue for me on lighttpd:
root@OpenWrt:~# chmod +s /usr/bin/lua

[...]

PR #6596, opened 2023-09-25
Luci collections add luci-lighttpd

stokito:
The PR is based on #6591 so check only the last commit.

Turris Omnia and Gl.Inet use the Lighttpd webserver.
But for the Luci we need to add some configuration. In order to help users who also want to use the Lighttpd the new package luci-lighttpd would be helpful.

For the TurrisOS there is a task [Add package luci-lighttpd](https://gitlab.nic.cz/turris/os/build/-/issues/406)


feckert:
I haven't looked into the whole thing in detail now. But while we are at it, that we want to run the LuCI on other web servers. Then we should have a package naming convention.

luci-www: Contains all packages that are common to all LuCI installations independent of the use webserver. 

luci-light -> luci-uhttpd: Dependencies (at least luci-www) and configuration files to run LuCI with uhttpd.
luci-nginx: dependencies (at least luci-www) and configuration files to run LuCI with nginx
luci-lighttpd: dependencies (at least luci-www) and configuration files to run LuCI with lighttpd
luci-ssl-openssl -> luci-uhttpd-openssl: dependencie to luci-uhttpd
luci-ssl -> luci-uhttpd-mbedtls: dependencie to luci-uhtpd
luci-ssl-nginx -> luci-nginx-openssl: dependencie to luci-nginx

stokito:
I agree with you and also thought about this. But here some thoughts that stopped me:
0. Just to clarify: The luci-www package that I proposed contains a single file /www/index.html. The package needs to be installed by default but firmware manufactures may choice to not use it and use their own admin panel instead (e.g. Turris reForris)
1. The luci-light is already existing package and some firmwares may already have it configured in scripts. We may rename it as well but that's probably better to do in next releases.

And finally:
Basically there shouldn't be any luci-SOME_WEBSERVER packages. In ideal world the Luci  should work out of the box with any webserver.
The purpose of the luci-light and luci-nginx is to add a dependency to mod-ubus and ucode.
Instead we can simplify this by just adding the mod-ubus just to uhttpd and nginx packages. They are anyway small.

The only difference is luci-lighttpd that adds a [config file to Lighttpd](https://github.com/openwrt/luci/pull/6596/commits/e43792468df6cc852dc2b2069c8a5bee0baaf502#diff-e9084521faa671dae8e756b546adb639859895ff2167dacf49ae8f0a47ad52e4).
The config file is small and basically it just makes the Lighttpd to be started as root.
It's a difficult topic about how to make this properly and was discussed before
https://github.com/openwrt/luci/issues/734#issuecomment-283769760
In fact I think that the luci should be started on own port and additionally a proxy should be placed before.
But just for today to simplify things the luci-lighttpd should be very helpful.

feckert:
>     1. Just to clarify: The luci-www package that I proposed contains a single file /www/index.html. The package needs to be installed by default but firmware manufactures may choice to not use it and use their own admin panel instead (e.g. Turris reForris)

This is a problem of downstream. Regardless, they can build their own package that has `luci-www` as a dependency and then delete the file in a `pre-install` script of their new package and replace it with their own. 

>     2. The luci-light is already existing package and some firmwares may already have it configured in scripts. We may rename it as well but that's probably better to do in next releases.

There we can use the [PROVIDE:=](https://github.com/search?q=repo%3Aopenwrt%2Fopenwrt+PROVIDES%3A&type=code) feature of openwrt to set an alternative name so that we can still use `luci-light` to install the package.


> Basically there shouldn't be any luci-SOME_WEBSERVER packages. In ideal world the Luci should work out of the box with any webserver. The purpose of the luci-light and luci-nginx is to add a dependency to mod-ubus and ucode. Instead we can simplify this by just adding the mod-ubus just to uhttpd and nginx packages. They are anyway small.

LuCI should work with any web server, but you need additional packages and configuration to make it work on them. That's why I think it's not bad if I can simply type `luci-ngnix` and the LuCI runs.

stokito:
> they can build their own package that has luci-www as a dependency and then delete the file in a pre-install script

If this can work then I probably can drop the commit with extracting of the luci-www.
Still this looks confusing to me when one package removes files from another.

collections/luci-light/Makefile
```diff
@@ -9,15 +9,10 @@ include $(TOPDIR)/rules.mk
 LUCI_TYPE:=col
 LUCI_BASENAME:=light
 
-LUCI_TITLE:=LuCI interface with Uhttpd as Webserver (light)
-LUCI_DESCRIPTION:=Light OpenWrt set including admin support and the default Bootstrap theme
+LUCI_TITLE:=LuCI Configuration Interface with light Uhttpd as webserver
```
raenye, line 12: _Uhttpd_, _uhttpd_ or _uHTTPd_?
It seems [official docs](https://openwrt.org/docs/guide-user/services/webserver/uhttpd) favour the latter.

stokito: it was already written here as Uhttpd so I kept it as is. It's a good idea to change to be the same as in docs

feckert:
> > they can build their own package that has luci-www as a dependency and then delete the file in a pre-install script

> If this can work then I probably can drop the commit with extracting of the luci-www. Still this looks confusing to me when one package removes files from another.

We don't have to delete it. We can move it with a postfix.
For example `index.html.luci-base`. And if we delete the package that moved this file, we could move `index.html.lubi-base` to the correct location again.

stokito:
Nice trick. That's just for me the luci-www seems more easier to implement (it's already), understand and use.

collections/luci-lighttpd/root/etc/lighttpd/conf.d/80-luci.conf
```diff
@@ -0,0 +1,12 @@
+# LuCI require root to manage router
+server.username  := ""
+server.groupname := ""
+
+alias.url += (
+  "/cgi-bin/" => server_root + "cgi-bin/"
```
openwrt-ai, line 6: lighttpd array elements must be separated by commas. With no comma between the two `=>` entries, lighttpd fails to parse this block (syntax error at the second key) and the server will not start.

```suggestion
  "/cgi-bin/" => server_root + "cgi-bin/",
```

---

collections/luci-lighttpd/Makefile
```diff
@@ -0,0 +1,40 @@
...
+	+lighttpd \
+	+lighttpd-mod-alias \
+	+lighttpd-mod-cgi
+
+PKG_LICENSE:=Apache-2.0
+
+define Package/luci-lighttpd/postinst
+#!/bin/sh
+if [ -z "$$IPKG_INSTROOT" ]; then
+  rm /var/log/lighttpd/error.log
```
openwrt-ai, line 25: On a fresh install `/var/log/lighttpd/error.log` may not exist yet (lighttpd has never started, or the file was already removed), so a bare `rm` exits non-zero and prints an error during package configuration. Use `rm -f`. Same applies to the `postrm` at line 33.

```suggestion
  rm -f /var/log/lighttpd/error.log
```

---

openwrt-ai (commented):
Reviewed the last commit (`collections: add luci-lighttpd`) per the PR description; the other diffs belong to the #6591 dependency. Two inline issues on the new files.

---

PR #6597, opened 2023-09-26
Convert LuCI JS code to ES6 standard

jow-:
This PR contains commits dropping polyfills not needed anymore on modern browsers and refactoring code to utilize ES6 syntax elements and features.


systemcrash:
With this PR, is it your opinion that we can start implementing ES6 for future JS code?

stokito (approved):

stokito:
The luci already have many places where the ES6 syntax is used.
Keep sources simple and small is complicated so we better to move to ES6.
Also please consider to use the [JavaScript Standard Style](https://standardjs.com/) e.g indent with spaces and no semicolons.

I can convert to ES6 other apps.

jow-:
I dislike semicolon omission and space indentation, so personally I am opposed to standard style.

PR #6616, opened 2023-10-03
Ramon00 patch 2

Ramon00:


hnyman:
Looks badly incomplete. Please first actually create the LuCI content, test it, and then create the PR.
(and no merge commits allowed)

PR #6617, opened 2023-10-03
First version of luci-app-usteer

Ramon00:
This is my initial version of the luci-app-usteer, also my 1st contribution to OpenWRT and my 1st pull request. Not really sure if code quality is good enough. Feel free to edit.


Ramon00:
ah i see, no lua, just as i was getting the hang of it. Let me see if i can find some time to convert to js.

PR #6619, opened 2023-10-03
luci-app-statistics: regenerate graphs on window resize

raenye:
Currently graphs are redrawn only based on the refresh interval (_never_ or _every 5/30/60 seconds_).
Since the image size is calculated based on the window size, redraw graphs (once) also after resizing the window. This also captures window resize due to orientation change (e.g., for mobile).
Since multiple resize events are fired when dragging the window border, there is a 250ms delay for debouncing.


systemcrash:
Merged via 8fcd52f5130acd1a0738ff47f9bba2c9cc42e662

PR #6625, opened 2023-10-08
luci-app-dawn: show hostnames of clients if known

micw:
This commits adds a "Hostname" column that shows the name of a host from host-hints if known.

### DIFF
--- a/applications/luci-app-dawn/luasrc/model/cbi/dawn/dawn_network.lua
+++ b/applications/luci-app-dawn/luasrc/model/cbi/dawn/dawn_network.lua
@@ -57,6 +57,7 @@ function s.render(self, sid)
 							<td class="td" style="vertical-align: top;">
 								<table class="table" id="clients-<%= mac %>">
 									<tr class="tr table-titles">
+										<th class="th">Hostname</th>
 										<th class="th">MAC</th>
 										<th class="th">HT</th>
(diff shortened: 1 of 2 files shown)


applications/luci-app-dawn/luasrc/model/cbi/dawn/dawn_network.lua
```diff
@@ -68,6 +69,7 @@ function s.render(self, sid)
 											if (type(clientvals) == "table") then
 									%>
 										<tr class="tr">
+											<td class="td"><% if (hosts[clientmac] and hosts[clientmac].name) then print(hosts[clientmac].name) end %></td>
```
jow-, line 72: This looks prone to XSS if someone manages to inject an odd hostname into the network. Please use the following expression:

```html
<td class="td"><%= hosts[clientmac] and hosts[clientmac].name and pcdata(hosts[clientmac].name) or "-" %></td>
```

micw: Thank you for the very good feedback. It's my first step with lua, now I learned the difference between print and <= ...
I have changed the line.

micw:
Closing this since luci-app-dawn has been re-implemented in JS meanwhile

Issue #6620, opened 2023-10-04
luci-app-ksmbd: multiple interfaces


AndroKev:
## Steps to reproduce:
1. go to: Services → Network Shares
2. try to set multiple interfaces

## Actual behavior:
only one interface can be selected

## Expected behavior:

allow to select multiple interfaces, in samba4 its allowed(not tested)

its working when you edit your `/etc/config/ksmbd` manually [reference](https://forum.openwrt.org/t/configure-ksmbd-samba-on-two-interfaces/106893)
```
config globals
  option interface 'lan wg0'
```

https://github.com/openwrt/luci/blob/18eed14abae33ed523480e2ffbe6dd3710b2d5bd/applications/luci-app-ksmbd/htdocs/luci-static/resources/view/ksmbd.js#L35-L37

https://github.com/openwrt/luci/blob/127ca64a672ba4b748ea230cce79b0c4cc1c8207/applications/luci-app-samba4/htdocs/luci-static/resources/view/samba4.js#L34-L50

## Additional Information:
OpenWrt version information from system `/etc/openwrt_release`

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.4'
DISTRIB_REVISION='r20123-38ccc47687'
DISTRIB_TARGET='ramips/mt76x8'
DISTRIB_ARCH='mipsel_24kc'
DISTRIB_DESCRIPTION='OpenWrt 22.03.4 r20123-38ccc47687'
DISTRIB_TAINTS=''
```
version of luci-app-ksmbd

`luci-app-ksmbd | git-21.224.17289-a624647`

PR #6631, opened 2023-10-10
luci-app-ksmbd: add UI for multiple interfaces

raenye:
`/etc/config/ksmbd` supports multiple interfaces (in `config globals`->`option interface`) but the current UI only supports selecting a single one. This patch adds support for selecting multiple interfaces, based on how `luci-app-samba4` does it.

Closes: #6620

Suggested-by: @AndroKev (let me know if you want your name in the commit)


jow-:
Merged, thanks!

PR #6632, opened 2023-10-11
Opkg: search all dest installed packages instead of root only

ghost:
Search all dest in `/etc/opkg.conf` listed, and output all dest installed packages


feckert:
Some more information would be nice why we need this?

feckert:
> Install software to external storage without using extroot

The answer should go to the commit message and you must also add a Signed-off-by with your real name and a valid email address. Not with a mail from github `users.noreply.github.com`

stokito (approved):

systemcrash:
> > > Install software to external storage without using extroot
> > 
> > 
> > The answer should go to the commit message and you must also add a Signed-off-by with your real name and a valid email address. Not with a mail from github `users.noreply.github.com`
> 
> Install software to external storage without using extroot

In the git commit message, not this PR thread :)

applications/luci-app-opkg/root/usr/libexec/opkg-call
```diff
@@ -7,7 +7,12 @@ shift
 
 case "$action" in
 	list-installed)
-		cat /usr/lib/opkg/status
+		dest_dirs=$(sed -rne 's#^dest \S+ (\S+)#\1#p' /etc/opkg.conf /etc/opkg/*.conf 2>/dev/null)
```
jow-, line 10: All of keyword, name and value words in opkg.conf may be double quoted (ref: https://git.openwrt.org/?p=project/opkg-lede.git;a=blob;f=libopkg/opkg_conf.c;h=0cbd1ccab73fb1c8051ffd811d8e2c40f010e969;hb=d038e5b6d155784575f62a66a8bb7e874173e92e#l212

Additionally, arbitrary amount of white space is allowed before, after and between directive words. For robustness reasons, the code should also fall back to /usr/lib/opkg/status in case destinations cannot be properly parsed.

PR #6638, opened 2023-10-19
luci-app-attendedsysupgrade: fix search for x86 target

McGiverGim:
The conditional is wrong. IndexOf returns a number and not true/false, so it does not work.


McGiverGim:
Modified the commit title, it was wrong.

Issue #5994, opened 2022-09-29
nlbwmon: sort is broken -- sorts by the number, not respecting the units


mikeage:
## Steps to reproduce:

1. Install luci-app-nlbwmon
2. Click on a column to sort by bytes or packets

## Actual behavior:

99 KiB comes before 10 MiB when sorting by descending bytes

## Expected behavior:

99 KiB should come after 10 MiB 

It should be sorting these as 10000000 vs 99000, instead of 10 vs 99

## Additional Information:
OpenWrt version information from system `/etc/openwrt_release`

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.0'
DISTRIB_REVISION='r19685-512e76967f'
DISTRIB_TARGET='rockchip/armv8'
DISTRIB_ARCH='aarch64_generic'
DISTRIB_DESCRIPTION='OpenWrt 22.03.0 r19685-512e76967f'
DISTRIB_TAINTS=''
```

jow-:
This has been fixed by 4f4ad9bff5482211695bf564543e4d34de4808d8

Issue #6120, opened 2022-11-28
admin/system/opkg: Sorting by size sorts alphabetically


tvogel:
## Steps to reproduce:

1. go to: System → Software → Installed tab
2. Click on "Size (.ipk)" column header

## Actual behavior:
Sorts by alphabet, i.e. you'll find sections like this:

Package name | Version | Size (.ipk) | Description | &nbsp;
-- | -- | -- | -- | --
...
kmod-nf-conntrack6 | 5.10.146-1 | ~846 B | Netfilter IPv6 connection tracking | Remove…
jsonfilter | 2018-02-04-c7e938d6-1 | ~9.0 KB | OpenWrt JSON filter utility | Remove…
kmod-crypto-rng | 5.10.146-1 | ~9.3 KB | CryptoAPI random number generation | Remove…
kmod-usb2 | 5.10.146-1 | ~9.6 KB | Kernel support for USB2 (EHCI) controllers | Remove…
kmod-crypto-null | 5.10.146-1 | ~940 B | Null CryptoAPI module | Remove…
...

## Expected behavior:
The unit suffix is taken into account and sorting is done by the real size.

## Additional Information:
OpenWrt version information from system `/etc/openwrt_release`

**Example: Replace the following lines and remove this**
```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.2'
DISTRIB_REVISION='r19803-9a599fee93'
DISTRIB_TARGET='ath79/generic'
DISTRIB_ARCH='mips_24kc'
DISTRIB_DESCRIPTION='OpenWrt 22.03.2 r19803-9a599fee93'
DISTRIB_TAINTS=''
```

jow-:
Duplicate of #5994 - same root cause/functionality

PR #6643, opened 2023-10-21
luci-base: fix sort order for size columns

raenye:
Currently elements of package size, packet size, etc. columns are interpreted as text strings, so we have _"800 B" > "7.54 KiB" > "5.11 MiB"_. To fix this, a new regexp is added to `UITable.deriveSortKey`.

It is still not a perfect fix for System->Software->Installed, since pagination is done based on package names, so within each page the packages are sorted correctly, but the smallest (resp., largest) package is not necessarily on the first (resp., last) page.

Closes: #5994 (see also #6120)

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -3412,6 +3412,20 @@ var UITable = baseclass.extend(/** @lends LuCI.ui.table.prototype */ {
 			if (m)
 				return '%010d%s'.format(+m[1], m[2]);
 
+			m = /^~?((?:\d+)(?:[.]\d+)?)\s(?:(K|M|G)i)?B$/.exec(value);
+
+			if (m)
+				switch (m[2]) {
+					default:
+						return +m[1];
+					case 'K':
+						return +m[1] * 1024;
+					case 'M':
+						return +m[1] * 1024 * 1024;
+					case 'G':
+						return +m[1] * 1024 * 1024 * 1024;
+				}
+
 			return String(value);
 
 		case 'ignorecase':


raenye:
@jow- Would you please take a look?

jow-:
I think a better solution is adding a `data-value` attribute to the affected cells which contains the raw, unformatted byte value.

raenye:
Maybe this should be the only value, and let css magic do the text formatting for rendering?

jow-:
CSS magic can't do that as far as I know.

systemcrash:
> I think a better solution is adding a `data-value` attribute to the affected cells which contains the raw, unformatted byte value.

I came to suggest this. Exclude any units from the numeric value, and append those at a later stage, so sorting is done on raw numbers.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -3412,6 +3412,20 @@ var UITable = baseclass.extend(/** @lends LuCI.ui.table.prototype */ {
 			if (m)
 				return '%010d%s'.format(+m[1], m[2]);
 
+			m = /^~?((?:\d+)(?:[.]\d+)?)\s(?:(K|M|G)i)?B$/.exec(value);
+
+			if (m)
+				switch (m[2]) {
+					default:
+						return +m[1];
+					case 'K':
```
knarrff, line 3421: Are you sure about the 'K' here? You seem to mean 'kilo', but 'kilo' is 'k' in short (lower case), not 'K' (upper case).

knarrff:
> This looks related to #4087

It is - plus as #4087 correctly points out, 'k' 'M' and 'G' are for 10^3, 10^6 and 10^9. However, the idea of the report is still an important one: columns should sort by real number, and even across pages.

modules/luci-base/htdocs/luci-static/resources/ui.js
raenye: Note the `i` in the regexp: `(?:(K|M|G)i)?B`; it's IEC, not SI.
https://en.wikipedia.org/wiki/Binary_prefix

knarrff: I did miss the 'i' and stand corrected.

raenye:
@systemcrash my JS skills are not sufficient to make the recommended changes. Should I close the PR?

systemcrash:
Have a crack at it. This is a great learning opportunity.

jow-:
Are there currently any tables left requiring this change? While I agree that it might be nice to smartly detect formatted size values and sort them accordingly, the proper fix should be adjusting affected tables to contain a `data-value` attribute. This can be achieved by changing the existing value expression to a two-element array where the first element should be the raw numeric value and the second value the formatted display text.

See for example commit https://github.com/openwrt/luci/commit/4f4ad9bff5482211695bf564543e4d34de4808d8 which fixed some tables shown in the nlbwmon application.

systemcrash:
Does the sorting logic then depend in the browser?

jow-:
> Does the sorting logic then depend in the browser?

I don't understand that question. Why should it depend on the browser?

systemcrash:
Because the data is at the client, and the client is the users web browser. So column sorting is performed by the browser. Or?

knarrff:
> Because the data is at the client, and the client is the users web browser. So column sorting is performed by the browser. Or?

It should not be done at the browser, because at most the data of the current page is available to the browser and sorting should span pages.

jow-:
> It should not be done at the browser, because at most the data of the current page is available to the browser and sorting should span pages.

There is no server side rendering in LuCI anymore, all data is rendered locally in the browser by building the table DOM from a multi dimensional array (or, in some cases by reading JS generated HTML table DOM and translating it into an internally table representation which is then eventually re-rendered) so sorting happens in the browser as well, naturally.

systemcrash:
So pagination of data should also happen at the browser.

systemcrash:
@knarrff was opkg page your only concern here?

systemcrash:
> @systemcrash my JS skills are not sufficient to make the recommended changes. Should I close the PR?

Would you like to try and add the necessary fields using this PR that @jow- highlighted earlier here?

raenye:
I'll try this, thanks.

systemcrash:
@raenye Do you have an update?

systemcrash:
See @jow-'s latest e7f2e6bc8ee2bbbfd3d1f7caa570a6529cc03eee

raenye:
> See @jow-'s latest [e7f2e6b](https://github.com/openwrt/luci/commit/e7f2e6bc8ee2bbbfd3d1f7caa570a6529cc03eee)

Thanks, but 63e5d38db02fa1 already did that part for `opkg.js`.
Nevertheless, it doesn't solve the pagination issue, just the per-page sorting.

In any case, this PR is no longer relevant, so I'm closing this. If I find the time to fix pagination, I'll create a new PR.

PR #6652, opened 2023-10-26
luci-app-dockerman: fix stats page always blank

jjm2473:
I have never saw stats page has any 'stats', maybe dockerd call `ps -aux` or `ps -ef` on openwrt host, which not supported.

`CPU usage` and `Memory usage` does not depends on docker:top, always show on container running (docker:top status code is not 409).


jow-:
Merged, thanks!

PR #6656, opened 2023-10-26
luci-app-dawn: Implement in JavaScript

danielvijge:
This pull request re-implements the app `luci-app-dawn` in JavaScript, removing the lua implementation.

Besides a re-implementation the following things are changed:
- In both the network overview and the hearing map MAC addresses are replaced by host name if known.
- In the hearing map, the table is sortable. If the same client is connected to multiple access points/frequencies the MAC/host name is listed twice, where in the lua implementation the second MAC address was empty to show it was referring to the same client.
- The view in the network overview is a bit different. This table is not sortable, because LuCi doesn't seem to like a table inside a table for sorting.
- Aligning the column names between the network overview and the hearing table.
- Add tooltips for abbreviations.

Still to test:
- ~building package~ (works); ~install package on a fresh install~ (works); ~upgrade package from lua to javascript~ (can lead to a problem, see https://github.com/openwrt/luci/pull/6656#issuecomment-1784198904)


danielvijge:
I also tried to compile it and install it. On a  install where the lua version of `luci-app-dawn` is installed, there is (or could be) an error after the installation. I believe this is not really related to the changed in this merge request, but just how the previous package was configured in combination with `opkg`.

If `luci-app-dawn` (lua version) is the only installed package that requires `luci-compat` this package will be uninstalled when installing the JavaScript version of `luci-app-dawn` over it. As a consequence, there is an error after login because it cannot find the lua compatibility layer, even though it is not really needed any more.

To fix this, either:
- Reinstalled the compatibility layer with `opkg install luci-compat`. This is the safest option, and won't break any other packages.
- Remove lua completely by running `opkg remove luci-compat luci-lua-runtime ucode-mod-lua luci-lib-base lua`. This might break other packages if lua is needed!

Is there a recommended way to deal with this? Or is it OK because an upgrade of a package, without a reinstall/update of openwrt, isn't recommended anyway?

PolynomialDivision:
Wow thanks a lot! Please squash commits together and prefix them with `luci-app-dawn: ...`.

danielvijge:
@PolynomialDivision Done. I've rebased on the latest commit from `master` and squashed everything into one commit.

danielvijge:
@PolynomialDivision I've added the full list of changes back to the commit message. It's basically the same as in this pull request, but you're right that it is helpful to have it in the commit message as well.

PolynomialDivision:
I think your signed-off is missing a valid email that we require here in this repository. Could you maybe add a valid mail? Sorry, for doing nitpicking.

danielvijge:
No problem. I understand the need for a good review process, and good to see it is applied. Had to change some settings because GitHub didn't like when my address is exposed in the commit, but it should be good now. The git author now also matched the one in `Signed-off-by`, as does the GPG signature. And as a bonus I reformatted the commit description to be less than 75 characters per line.

PolynomialDivision:
Thanks a lot. Great work. :) I will merge.

PolynomialDivision:
I will also cherry-pick that to 23.05.

micw:
@danielvijge This is great, thank you. I have a few comments and would be happy if you would create a follow-up PR.

1. the hearing map is not showing more than one AP per client anymore.
this is the data:
image

this is the result:
image

2. replacing the MAC with the names is a bit sub-optimal. It would be better to show the names in addition to the MAC

3. The nmame of the APs could be shjown in the hearing map. To do so, you can fetch the network and extract it from there. Here's how I did this in my now obsolete PR: https://github.com/openwrt/luci/pull/6625/files#diff-a17ac5c6613bbbfcb01f6a442e12dd594644d7778f580e2be7d5e29e11cf0667

4. you could flatten the network overview to one row per AP and client and add SSID as a column. This way you have only one "denormalized" table. If that table allows sorting and filtering, it would be _way_ better than the old table. E.g. one could filter by SSID or sort by client.

Kind regards,
Michael.

danielvijge:
@micw Thanks for testing it and submitting the feedback. For the first one, I remember explicitly testing for this, but it appears in the final, cleaned-up version, the `flat()` statement is applied to the wrong array.

Fixes for 1,2, and 3 are available in https://github.com/danielvijge/luci/commits/luci-app-dawn-js-fixes-and-improvements/ each in their own commit.

For 4 I have to look at it later. It can be done, but as I understand it now, it would remove the H3 headings for the SSID, and it would just be one table? Then it might make sense to apply the same logic to the hearing map as well. Or maybe I'm misunderstanding it. I was thinking of a better structure for this, but couldn't really come up with one. It might help if you could comment the table structure you envision.

Could you already test these changes and see if they all work?

(disclaimer: I tested them directly on my router, then applied the changes directly as an edit in GitHub. There is a chance I copied something wrong. I didn't test copying things back yet, or even making a package out of it. I'll later squash everything in one commit, do a proper change log, sign-off, pull request, etc.)

micw:
Hey @danielvijge ,

yes, for 4 I imagined just one large table. At the moment there is an arbitrary structure (SSID -> AP -> Clients). This might be meaningful for some uses but not in general. A large table with everything in cloumns would be sortable/filterable by each aspect and be much more usefull. E.g. one could filter by SSID, by AP or by a particular client. Same applies to the hearing map.

How can I test the new changes? I just updated today and was able to test the last PR since it was already cherry-picked to 23.x

danielvijge:
You could directly edit the files on OpenWrt with the new versions. The files are in `/www/luci-static/resources/dawn` and `/www/luci-static/resources/view/dawn`. The installed files are minimised, but you can jus paste the full source version in there. Then just reload the page.

danielvijge:
Thinking a bit more about the table. I think this is what you proposed:

| SSID | Access Point | Interface | MAC | Utilization | Frequency | HT | VHT | Client | HT | VHT | Signal
|--------|--------|--------|--------|--------|--------|--------|--------|--------|--------|--------|--------|
| MyWiFi | ap1 | phy0-ap0 | 00:01:02:03:04:05 | 25.49% | 5.200 GHz (Channel: 40) | Available | Available | Laptop | Available | Available | -74 |
| MyWiFi | ap1 | phy0-ap0 | 00:01:02:03:04:05 | 25.49% | 5.200 GHz (Channel: 40) | Available | Available | Phone | Available | Not available | -53 |
| MyWiFi | ap2 | phy0-ap0 | FF:09:08:07:06:AA | 2.83% | 5.200 GHz (Channel: 40) |  Available | Available | *No clients connected* | | | |

There might be several problems with such a table:
- The first 8 columns are repeated many time, especially when there are many client connected to an AP. This might make things harder to read.
- Some columns refer to AP properties, while other refer to client properties. This might be difficult to communicate.
- Aggregate properties, like number of clients connected don't really make that much sense.

In a way, the table would be almost the same as the hearing map, with just an additional column to indicate if a client is actually connected/authenticated to the network, or only seen by an AP.

micw:
It's not the same as the hearing map since it would contain only connected clients, not all visible clients (so it's much smaller). But now than you write this, Seing which clients is connected would be a great extension to the hearing map ;-)

My idea was that if everything in one table, it could be made much more interactive. Want to see how many clients are connected to one AP - filter by that AP. Want to see which client are on a particular SSID - filter by that SSID etc.

danielvijge:
The first 3 bugs are fixed in #6674. For an optional view of the table I opened #6675 to have a discussion.

PR #6661, opened 2023-10-29
luci-app-olsr-services: change menu order

andibraeu:
we do not want olsr services as first menu entry, if there are more.


feckert:
Can you please use a valid mail address in your `Signed-off-by`
This is not correct:
`Signed-off-by: Andi Bräu <andibraeu@users.noreply.github.com>`

andibraeu:
thank you @feckert ✅

jow-:
Merged, thanks! In the future please just open a PR against master and state in the description that you would like to see a cherry pick to 23.05. Thanks!

PR #6664, opened 2023-10-30
[23.05] luci-app-dawn: Implement in JavaScript

PolynomialDivision:
This commit re-implements luci-app-dawn in JavaScript, removing the older lua implementation. Besides a 1-to-1 port, there are some changes/improvements:

* In both the network overview and the hearing map, replace MAC addresses by host name if known.
* In the hearing map, the table is sortable. If the same client is connected to multiple access points/frequencies the MAC/host name is listed twice, whereas in the lua implementation the second MAC address was empty to show it was referring to the same client. This means the table can be sorted on any column, and the information remains correct.
* The view in the network overview is a bit different. This table is not sortable, because LuCi doesn't seem to like a table inside a table for sorting.
* Align the column names between the network overview and the hearing table.
* Add tooltips for abbreviations in column names.

(cherry picked from commit ea8c0aa2a1a1ff9f29ffd452049a7b749298ed17)
[apply latest Makefile changes that did not get backported to 23.05]


micw:
Hello,

thank you. I just closed my own PR since this adds the feature of showing hostnames too. Would you also add the AP hostname to the hearing map instead of the MAC? Ths AP Hostname can be fetched from the network overview. Here's how I did it:
https://github.com/openwrt/luci/pull/6625/files

Kind regards,
Michael.

PR #6658, opened 2023-10-28
luci-app-usteer: First version of luci-app-usteer

Ramon00:
First version of luci-app-usteer. 2 pages, 1 for settings, 1 for status.

PR #6671, opened 2023-11-02
luci-app-usteer: First version of luci-app-usteer

Ramon00:
This implements the 1st version of luci-app-usteer. 3 pages, 1 for settings 1 for status of connected clients and 1 for the hearing map.

Note this is a resubmit (see https://github.com/openwrt/luci/pull/6658 for review) .

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -1,0 +1,480 @@
+'use strict';
+'require view';
+'require rpc';
+'require poll';
+'require dom';
+'require ui';
+'require form';
+'require uci';
+
+
+var Hosts,Remotehosts,Remoteinfo,Localinfo,Clients;
+var HearingMap = form.DummyValue.extend({
+	renderWidget: function() {
+		var body = E([
+			E('h3', _('Hearing map'))
+		]);
+		for(var mac in Clients) {
+			var maciphost='';
+			maciphost=mac;
+			var macUp=String(mac).toUpperCase()
+			if (typeof Hosts[macUp] !== 'undefined') {
+				if ((String(Hosts[macUp]['ipaddrs'][0]).length>0) && (typeof Hosts[macUp]['ipaddrs'][0] !== 'undefined'))
+					maciphost+='\u2003'+Hosts[macUp]['ipaddrs'];
+				if ((String(Hosts[macUp]['name']).length>0) && (typeof Hosts[macUp]['name'] !== 'undefined')) 
+					maciphost+='\u2003%h'.format(Hosts[macUp]['name']);
+				}
+			body.appendChild(
+				E('h4', maciphost)
+			);
+			var client_table = E('table', { 'class': 'table cbi-section-table' }, [
+				E('tr', { 'class': 'tr table-titles' }, [
+					E('th', { 'class': 'th', 'style': 'width:35%' }, _('IP & Interface','Combination of IP and interface name in usteer overview')),
+					E('th', { 'class': 'th', 'style': 'width:25%' }, _('SSID')),
+					E('th', { 'class': 'th', 'style': 'width:15%' }, _('Frequency', 'BSS operating frequency in usteer overview')),
+					E('th', { 'class': 'th', 'style': 'width:15%' }, _('Connected', 'Connection state in usteer overview')),
+					E('th', { 'class': 'th', 'style': 'width:15%' }, _('Signal','Signal strength reported by wireless station in usteer overview'))
+				])
+			]);
+			var client_table_entries =[];
+			for(var wlanc in Clients[mac]) {
+				var SSID='';
+				var freq=0;
+				if (typeof Localinfo[wlanc] !== 'undefined') { 
+					SSID=Localinfo[wlanc]['ssid'];
+					freq=Localinfo[wlanc]['freq'];
+				} 
+				if (typeof Remoteinfo[wlanc] !== 'undefined') { 
+					SSID=Remoteinfo[wlanc]['ssid'];
+					freq=Remoteinfo[wlanc]['freq'];
+				}
+				client_table_entries.push([
+					'<nobr>'+wlanc+'</nobr>', 
+					SSID,
+					freq,
+					Clients[mac][wlanc]['connected'] === true ? 'Yes' : 'No',
+					Clients[mac][wlanc]['signal']
+				]);
+			}
+			cbi_update_table(client_table, client_table_entries, E('em', _('No data')));
+			body.appendChild(client_table);
+		}
+		return E('div', { 'class': 'cbi-section cbi-tblsection' }, [body]);
+	}
+});
+var Clientinfooverview = form.DummyValue.extend({
+	renderWidget: function() {
+		var body = E([
+			E('h3', 'Remotehosts')
+		]);
+		var remotehost_table = E('table', { 'class': 'table cbi-section-table' }, [
+			E('tr', { 'class': 'tr table-titles' }, [
+				E('th', { 'class': 'th' }, _('IP address')),
+				E('th', { 'class': 'th' }, _('Identifier'))
+			])
+		]);
+		var remotehosttableentries =[];
+		for(var IPaddr in Remotehosts) {
+			remotehosttableentries.push([IPaddr, Remotehosts[IPaddr]['id']]);
+		}
+		cbi_update_table(remotehost_table, remotehosttableentries, E('em', _('No data')));
+		body.appendChild(remotehost_table);
+		body.appendChild(
+			E('h3', 'Client list')
+		);
+		var connectioninfo_table = E('table', { 'class': 'table cbi-section-table' }, [
+			E('tr', { 'class': 'tr table-titles' }, [
+				E('th', { 'class': 'th' }, _('IP & Interface name','Combination of IP and interface name in usteer overview')),
+				E('th', { 'class': 'th' }, _('BSSID')),
+				E('th', { 'class': 'th' }, _('SSID')),
+				E('th', { 'class': 'th' }, _('Frequency', 'BSS operating frequency in usteer overview')),
+				E('th', { 'class': 'th' }, _('N', 'Number of associated clients in usteer overview')),
+				E('th', { 'class': 'th' }, _('Noise', 'Channel noise in usteer overview')),
+				E('th', { 'class': 'th' }, _('Load', 'Channel load in usteer overview')),
+				E('th', { 'class': 'th' }, _('Max assoc','Max associated clients in usteer overview')),
+				E('th', { 'class': 'th' }, _('Roam src','Roam source in usteer overview')),
+				E('th', { 'class': 'th' }, _('Roam tgt','Roam target in usteer overview'))
+			])
+		]);
+		var connectioninfo_table_entries =[];
+		for(var wlan in Localinfo) {
+			connectioninfo_table_entries.push([
+							'<nobr>'+wlan+'</nobr>', 
+							Localinfo[wlan]['bssid'],
+							Localinfo[wlan]['ssid'],
+							Localinfo[wlan]['freq'],
+							Localinfo[wlan]['n_assoc'],
+							Localinfo[wlan]['noise'],
+							Localinfo[wlan]['load'],
+							Localinfo[wlan]['max_assoc'],
+							typeof Localinfo[wlan]['roam_events']['source'] !== 'undefined' ?  Localinfo[wlan]['roam_events']['source'] : '',
+							typeof Localinfo[wlan]['roam_events']['target'] !== 'undefined' ?  Localinfo[wlan]['roam_events']['target'] : ''
+			]);
+		}
+		for(var wlan in Remoteinfo) {
+			connectioninfo_table_entries.push([
+							'<nobr>'+wlan+'</nobr>',                                                                             
+							Remoteinfo[wlan]['bssid'],
+							Remoteinfo[wlan]['ssid'],
+							Remoteinfo[wlan]['freq'],
+							Remoteinfo[wlan]['n_assoc'],
+							Remoteinfo[wlan]['noise'],
+							Remoteinfo[wlan]['load'],
+							Remoteinfo[wlan]['max_assoc'],
+							typeof Remoteinfo[wlan]['roam_events']['source'] !== 'undefined' ?  Remoteinfo[wlan]['roam_events']['source'] : '',
+							typeof Remoteinfo[wlan]['roam_events']['target'] !== 'undefined' ?  Remoteinfo[wlan]['roam_events']['target'] : ''
+			]);
+		}
+		cbi_update_table(connectioninfo_table, connectioninfo_table_entries, E('em', _('No data')));
+		body.appendChild(connectioninfo_table);
+		var compactconnectioninfo_table = E('table', { 'class': 'table cbi-section-table' }, [
+				E('tr', { 'class': 'tr table-titles' }, [
+					E('th', { 'class': 'th' }, _('IP & Interface name','Combination of IP and interface name in usteer overview')),
+					E('th', { 'class': 'th' }, _('SSID')),
+					E('th', { 'class': 'th' }, _('Frequency', 'BSS operating frequency in usteer overview')),
+					E('th', { 'class': 'th' }, _('Load', 'Channel load in usteer overview')),
+					E('th', { 'class': 'th' }, _('N', 'Number of associated clients in usteer overview')),
+					E('th', { 'class': 'th' }, _('Host','host hint in usteer overview'))
+					
+				])
+		]);
+		var compactconnectioninfo_table_entries =[];
+		for(var wlan in Localinfo) {
+			var hostl=''
+			for(var mac in Clients) {
+				if (typeof Clients[mac] !== 'undefined') 
+					if (typeof Clients[mac][wlan] !== 'undefined') 
+						if (String(Clients[mac][wlan]['connected']).valueOf()==String("true").valueOf()) {
+							var foundname=mac;
+							var macUp=String(mac).toUpperCase()
+							if (typeof Hosts[macUp] !== 'undefined') {
+								if ((String(Hosts[macUp]['ipaddrs'][0]).length>0) && (typeof Hosts[macUp]['ipaddrs'][0] !== 'undefined')) {
+									foundname=Hosts[macUp]['ipaddrs'][0];
+								}
+								if ((String(Hosts[macUp]['name']).length>0) && (typeof Hosts[macUp]['name'] !== 'undefined')) {
+									foundname=Hosts[macUp]['name'];
+								}
+							}
+							hostl += '%h\u2003'.format(foundname);
+				}
+			}
+			compactconnectioninfo_table_entries.push([
+								'<nobr>'+wlan+'</nobr>', 
+								Localinfo[wlan]['ssid'],
+								Localinfo[wlan]['freq'],
+								Localinfo[wlan]['load'],
+								Localinfo[wlan]['n_assoc'],
+								hostl
+							]);
+		}
+		for(var wlan in Remoteinfo) {
+			var hostl=''
+			for(var mac in Clients) {
+				if (typeof Clients[mac] !== 'undefined') 
+					if (typeof Clients[mac][wlan] !== 'undefined') 
+						if (String(Clients[mac][wlan]['connected']).valueOf()==String("true").valueOf()) {
+							var foundname=mac;
+							var macUp=String(mac).toUpperCase()
+							if (typeof Hosts[macUp] !== 'undefined') {
+								if ((String(Hosts[macUp]['ipaddrs'][0]).length>0) && (typeof Hosts[macUp]['ipaddrs'][0] !== 'undefined')) {
+									foundname=Hosts[macUp]['ipaddrs'][0]; 
+								}
+								if ((String(Hosts[macUp]['name']).length>0) &&	(typeof Hosts[macUp]['name'] !== 'undefined')) {
+									foundname=Hosts[macUp]['name'];
+								}
+							}
+							hostl += '%h\u2003'.format(foundname);
+				}
+			}
+			compactconnectioninfo_table_entries.push([
+								'<nobr>'+wlan+'</nobr>', 
+								Remoteinfo[wlan]['ssid'],
+								Remoteinfo[wlan]['freq'],
+								Remoteinfo[wlan]['load'],
+								Remoteinfo[wlan]['n_assoc'],
+								hostl
+			]);
+		}
+		cbi_update_table(compactconnectioninfo_table, compactconnectioninfo_table_entries, E('em', _('No data')));
+		body.appendChild(compactconnectioninfo_table);
+		return E('div', { 'class': 'cbi-section cbi-tblsection' }, [body]);
+	}
+});
+
+var Settingstitle = form.DummyValue.extend({
+	renderWidget: function() {
+		var body = E([
+			E('h3', 'Settings'), 
+			E('body','First four options are mandatory. Also be sure to enable rrm reports, 80211kv, etc see: https://openwrt.org/docs/guide-user/network/wifi/usteer')
+		]);
+		return E('div', [body]);
+	}
+});
+
+return view.extend({
+	callHostHints: rpc.declare({
+		object: 'luci-rpc',
+		method: 'getHostHints',
+		expect: { '': {} }
+	}),
+	callGetRemotehosts: rpc.declare({
+		object: 'usteer',
+		method: 'remote_hosts',
+		expect: {  '': {}}
+	}),
+	callGetRemoteinfo: rpc.declare({
+		object: 'usteer',
+		method: 'remote_info',
+		expect: { '': {} }
+	}),
+	callGetLocalinfo: rpc.declare({
+		object: 'usteer',
+		method: 'local_info',
+		expect: { '': {} }
+	}),
+	callGetClients: rpc.declare({
+		object: 'usteer',
+		method: 'get_clients',
+		expect: { '': {} }
+	}),
+	load: function() {
+		return Promise.all([
+			this.callHostHints(),
+			this.callGetRemotehosts(),
+			this.callGetRemoteinfo(),
+			this.callGetLocalinfo(),
+			this.callGetClients()
+		]);
+	},
+	render: function(data) {
+		var m, s, o;
+		Hosts = data[0];
+		Remotehosts=data[1];
+		Remoteinfo=data[2];
+		Localinfo=data[3];
+		Clients=data[4];
+
+		m = new form.Map('usteer', _('Usteer'));
+
+		s = m.section(form.TypedSection);
+		s.anonymous = true;
+		s.tab('status', _("Status"));
+		s.tab('hearingmap', _("Hearing map"));
+		s.tab('settings', _("Settings"));
+		
+		o = s.taboption('status',Clientinfooverview);
+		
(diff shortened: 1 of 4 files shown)


ptpt52:
nice.

McGiverGim:
The comment of the commit is wrong, it seems it contains comments from all the commits from the first version. Apart of that, some strange indentations but it's ok to me.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+				E('th', { 'class': 'th' }, _('rrm_nr mac')),
+				E('th', { 'class': 'th' }, _('rrm_nr hex'))
+				
+			])
+		]);
+	
+		var connectioninfo_table_entries =[];
+		for(var wlan in Localinfo) {
+			connectioninfo_table_entries.push([
+							'<nobr>'+wlan+'</nobr>',
```
jow-, line 112: Is `wlan` the network SSID? If so, please change the code into `'<nobr>%h</nobr>'.format(wlan),` to HTML escape the value in order to avoid XSS.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+							Localinfo[wlan]['max_assoc'],
+							Localinfo[wlan]['roam_events']['source'],
+							Localinfo[wlan]['roam_events']['target'],
+							Localinfo[wlan]['rrm_nr'][0],
+							Localinfo[wlan]['rrm_nr'][2]
+						]);
+				}	
+		for(var wlan in Remoteinfo) {
+			connectioninfo_table_entries.push([
+							'<nobr>'+wlan+'</nobr>',
```
jow-, line 128: ditto

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+								}
+								if ((String(hosts[macUp]['name']).length>0) && (typeof hosts[macUp]['name'] !== 'undefined')) {
+									foundname=hosts[macUp]['name'];
+								}
+							}
+							hostl=hostl+ foundname+'&emsp;';
+				}
+			}		
+			compactconnectioninfo_table_entries.push([
+								'<nobr>'+wlan+'</nobr>',
```
jow-, line 180: ditto

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+								}
+								if ((String(hosts[macUp]['name']).length>0) &&	(typeof hosts[macUp]['name'] !== 'undefined')) {
+									foundname=hosts[macUp]['name'];
+								}
+							}
+							hostl=hostl+ foundname+'&emsp;';
+				}
+			}	
+			compactconnectioninfo_table_entries.push([
+								'<nobr>'+wlan+'</nobr>',
```
jow-, line 208: ditto

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+							var macUp=String(mac).toUpperCase()
+							if (typeof hosts[macUp] !== 'undefined') {
+								if ((String(hosts[macUp]['ipaddrs'][0]).length>0) && (typeof hosts[macUp]['ipaddrs'][0] !== 'undefined')) {
+									foundname=hosts[macUp]['ipaddrs'][0]; 
+								}
+								if ((String(hosts[macUp]['name']).length>0) &&	(typeof hosts[macUp]['name'] !== 'undefined')) {
+									foundname=hosts[macUp]['name'];
+								}
+							}
+							hostl=hostl+ foundname+'&emsp;';
```
jow-, line 204: There's a chance that the hostname is bogus/malicious. Please change the expression to `hostl += '%h\u2003'.format(foundname);`

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+		);
+		for(var mac in Clients) {
+			var maciphost='';
+			maciphost='Mac: '+mac;
+			var macUp=String(mac).toUpperCase()
+				if (typeof hosts[macUp] !== 'undefined') {
+					if (typeof hosts[macUp]['ipaddrs'] !== 'undefined') 
+						maciphost=maciphost+'&emsp;IP: '+hosts[macUp]['ipaddrs'];
+					if (typeof hosts[macUp]['name'] !== 'undefined') 
+						maciphost=maciphost+'&emsp;Host: '+hosts[macUp]['name'];
```
jow-, line 233: There's a chance that the hostname is bogus/malicious. Please change the expression to `maciphost += '\u2003Host: %h'.format(hosts[macUp]['name']);`

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+					E('th', { 'class': 'th' }, _('connected')),
+					E('th', { 'class': 'th' }, _('signal'))
+				])
+			]);
+		
+			var client_table_entries =[];
+			for(var wlanc in Clients[mac]) {
+				client_table_entries.push([
+					'<nobr>'+wlanc+'</nobr>', 
+					(String(Clients[mac][wlanc]['connected']).valueOf()==String("true").valueOf()) ? "True" : "",
```
jow-, line 250: Should be just
`Clients[mac][wlanc]['connected'] === true ? 'True' : ''`

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+		body.appendChild(
+			E('h3', 'Hearing map')
+		);
+		for(var mac in Clients) {
+			var maciphost='';
+			maciphost='Mac: '+mac;
+			var macUp=String(mac).toUpperCase()
+				if (typeof hosts[macUp] !== 'undefined') {
+					if (typeof hosts[macUp]['ipaddrs'] !== 'undefined') 
+						maciphost=maciphost+'&emsp;IP: '+hosts[macUp]['ipaddrs'];
```
jow-, line 231: Please replace `&emsp;` with `\u2003`

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/settings.js
```diff
@@ -0,0 +1,206 @@
...
+	o.value('0','0 Fatal');
+	o.value('1','1 info');
+	o.value('2','2 Verbose');
+	o.value('3','3 Some debug');
+	o.value('4','4 network packet info');
+	o.value('5','5 all debug messages');
+	o.rmempty = false;
+	o.editable = true;
+
+	o = s.option(form.Value, 'max_neighbour_reports', _('max neighbour reports'), _('Maximum number of neighbor reports set for a node'));
```
jow-, line 34: Please capitalize option titles, like it is done elsewhere in LuCI.

jow- (changes requested):
Great work! I added a number of remarks. Please don't open a new PR but keep updating this one.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: No, its the IP+interface name so something like "192.168.64.1#hostapd.phy0-ap0". Note this table only contains information about the users own network, so XSS should not be an issue. I can change it to that suggestion, but I do not think it is really needed. Let me know what you prefer

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: yes let me do this.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: agree

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: ok

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: ok

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+							var macUp=String(mac).toUpperCase()
+							if (typeof hosts[macUp] !== 'undefined') {
+								if ((String(hosts[macUp]['ipaddrs'][0]).length>0) && (typeof hosts[macUp]['ipaddrs'][0] !== 'undefined')) {
+									foundname=hosts[macUp]['ipaddrs'][0];
+								}
+								if ((String(hosts[macUp]['name']).length>0) && (typeof hosts[macUp]['name'] !== 'undefined')) {
+									foundname=hosts[macUp]['name'];
+								}
+							}
+							hostl=hostl+ foundname+'&emsp;';
```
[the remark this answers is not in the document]
Ramon00: agree

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+							Remoteinfo[wlan]['rrm_nr'][2]
+
+						]);
+				}	
+		cbi_update_table(connectioninfo_table, connectioninfo_table_entries, E('em', _('No data')));
+		body.appendChild(connectioninfo_table);
+/////////////////////////////////////////
+
+
+		var compactconnectioninfo_table = E('table', { 'class': 'table cbi-section-table' }, [
```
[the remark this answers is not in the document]
Ramon00: That changes the layout completely. That is not an improvement...
image
image

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/settings.js
Ramon00: done

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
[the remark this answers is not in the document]
jow-: What theme is this?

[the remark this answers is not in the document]
Ramon00: just default

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: ditto

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: ditto

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: ditto

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+		var body = E([
+			E('h2', _('Usteer status'))
+		]);
+
+
+/////////////////////////
+		body.appendChild(
+			E('h3', 'Remotehosts')
+		);
+		var remotehost_table = E('table', { 'class': 'table cbi-section-table' }, [
```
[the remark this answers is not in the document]
Ramon00: see above

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+				}
+
+		cbi_update_table(remotehost_table, remotehosttableentries, E('em', _('No data')));
+		body.appendChild(remotehost_table);
+		
+/////////////////////////
+		body.appendChild(
+			E('h3', 'Client list')
+		);
+		var connectioninfo_table = E('table', { 'class': 'table cbi-section-table' }, [
```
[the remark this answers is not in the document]
Ramon00: see above

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
```diff
@@ -0,0 +1,264 @@
...
+							Remoteinfo[wlan]['ssid'],
+							Remoteinfo[wlan]['freq'],
+							Remoteinfo[wlan]['n_assoc'],
+							Remoteinfo[wlan]['noise'],
+							Remoteinfo[wlan]['load'],
+							Remoteinfo[wlan]['max_assoc'],
+							Remoteinfo[wlan]['roam_events']['source'],
+							Remoteinfo[wlan]['roam_events']['target'],
+							Remoteinfo[wlan]['rrm_nr'][0],
+							Remoteinfo[wlan]['rrm_nr'][2]
```
McGiverGim, line 139: This is producing an error in my system. It seems `Remoteinfo[wlan]['rrm_nr']` does not exist. I don't know why. It exists in the `Localinfo` but not in the `Remoteinfo`.

image

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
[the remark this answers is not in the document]
McGiverGim: As curiosity I've tested this. With bootstrap theme (default) it seems the design is broken. If I use argon, for example, the design is correct.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
Ramon00: ok weird, on my setup it gets filled in, is the "network" correctly filled in on all APs? Regardless let me add some error checking on that, if its "undefined" then i will skip printing

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/status.js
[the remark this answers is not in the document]
Ramon00: Maybe make an issue for this, such that somebody can look at it?

Ramon00:
im having trouble squashing the commits, if i do that then they dubplicate...

Ramon00:
ah nevermind, i got it

rany2:
I'm having an issue with some items in hearing map showing `IP: ` without any value. I checked the JSON and it seems in some cases ipaddr is an empty array which is object not "undefined". Code seems to check if `hosts..["ipaddr"]` is `undefined` when I think it should be `typeof ...["ipaddr"] !== undefined && ...length > 0`.

image

rany2:
Notes about settings page:

- Missing probe steering, defaults to 0.
- Load balancing threshold defaults to 0, not 5.

See the OpenWRT wiki about this, I recently updated the information on it.

Ramon00:
> 'm having an issue with some items in hearing map showing `IP: ` without any value. I checked the JSON and it seems in some cases ipaddr is an empty array which is object not "undefined". Code seems to check if `hosts..["ipaddr"]` is not `undefined` when I think it should be `typeof ...["ipaddr"] !== undefined && ...length > 0`.
Ok I added a check. Try it again please :)

Ramon00:
> Notes about settings page:
> 
> * Missing probe steering, defaults to 0.
> * Load balancing threshold defaults to 0, not 5.
> 
> See the OpenWRT wiki about this, I recently updated the information on it.

Ok updated. Thx

rany2:
Getting `hosts[macUp].name is undefined` now.

Ramon00:
> @Ramon00
> 
> ```
> 				if ((String(hosts[macUp]['ipaddrs'][0]).length>0) && (typeof hosts[macUp]['ipaddrs'] !== 'undefined'))
> 					maciphost+='\u2003IP: '+hosts[macUp]['ipaddrs'];
> 				if ((String(hosts[macUp]['name'][0]).length>0) && (typeof hosts[macUp]['name'] !== 'undefined')) 
> 					maciphost+='\u2003Host: %h'.format(hosts[macUp]['name']);
> 				}
> ```
> 
> should be
> 
> ```
> 				if ((typeof hosts[macUp]['ipaddrs'] !== 'undefined') && (String(hosts[macUp]['ipaddrs'][0]).length>0))
> 					maciphost+='\u2003IP: '+hosts[macUp]['ipaddrs'];
> 				if ((typeof hosts[macUp]['name'] !== 'undefined') && (String(hosts[macUp]['name'][0]).length>0)) 
> 					maciphost+='\u2003Host: %h'.format(hosts[macUp]['name']);
> 				}
> ```

The [0] with the "name" was a copy/paste mistake, there should be only 1 name in the list, so the index zero is not needed. For the ipadders I think it makes sense to  use 
if ((String(hosts[macUp]['ipaddrs'][0]).length>0) && (typeof hosts[macUp]['ipaddrs'][0] !== 'undefined'))
not sure if its better to rempve the p0[ for the "undefined" check?

Ramon00:
One question to all, does it make sense to break down the "RRM NR hex string"? There should be info in there to indicate capabilities of APs. Not sure if it is interesting information or not.

rany2:
> The [0] with the "name" was a copy/paste mistake, there should be only 1 name in the list, so the index zero is not needed. For the ipadders I think it makes sense to use
> if ((String(hosts[macUp]['ipaddrs'][0]).length>0) && (typeof hosts[macUp]['ipaddrs'][0] !== 'undefined'))
> not sure if its better to rempve the [0] for the "undefined" check?

I think you could just check if the # of array elements instead. I think for most people, only ip6 might have multiple addresses.

> One question to all, does it make sense to break down the "RRM NR hex string"? There should be info in there to indicate capabilities of APs. Not sure if it is interesting information or not.

I would remove it from the status page or hide it behind some debug checkbox. I feel like if someone needs this info, they'd use the CLI but just my opinion.

rany2:
Also I think `rrm nr mac` is redundant as it always be the BSSID.

rany2:
This is a nitpick, but could we instead of what we currently have for client headers:

 `Mac: 22:59:9c:82:3c:6e IP: 192.168.1.2 Host: laptop.home.arpa`

Have this: `laptop.home.arpa (22:59:9c:82:3c:6e, 192.168.1.2, fd00:1234:1234::1)` (similar to the Wireless tab's "Associated Stations" table.

Ramon00:
> I would remove it from the status page or hide it behind some debug checkbox. I feel like if someone needs this info, they'd use the CLI but just my opinion.



> Also I think `rrm nr mac` is redundant as it will always be the BSSID.

Yeah I was thinking it is a bit useless both. Hence the reason to ask. Let me see if I can maybe replace it with some of the info contained in it, like HT,  VHT 
image

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/settings.js
```diff
@@ -0,0 +1,209 @@
...
+	s = m.section(form.TypedSection, 'usteer', _('Options'), _('First four options are mandatory. Also be sure to enable rrm reports, 80211kvr, etc see: https://openwrt.org/docs/guide-user/network/wifi/usteer'));
+	s.anonymous = true;
+
+	s.option(form.Value, 'network', _('Network'), _('The network interface for inter-AP communication)'));
+
+	o = s.option(form.Flag, 'syslog', _('Syslog'), _('Log messages to syslog (0/1)'));
+	o.default = '1';
+	o.rmempty = false;
+
+	o = s.option(form.Flag, 'ipv6', _('Ipv6 mode'), _('Use IPv6 for remote exchange'));
```
rany2, line 20: Rename Ipv6 => IPv6

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/settings.js
```diff
@@ -0,0 +1,209 @@
...
+'require view';
+'require form';
+
+return view.extend({
+	render: function() {
+	var m, s, o;
+
+	m = new form.Map('usteer', _('Usteer'), _('Usteer configuration.'));
+
+	s = m.section(form.TypedSection, 'usteer', _('Options'), _('First four options are mandatory. Also be sure to enable rrm reports, 80211kvr, etc see: https://openwrt.org/docs/guide-user/network/wifi/usteer'));
```
rany2, line 11: 802.11r is not required for usteer.

rany2:
> > I would remove it from the status page or hide it behind some debug checkbox. I feel like if someone needs this info, they'd use the CLI but just my opinion.
> 
> > Also I think `rrm nr mac` is redundant as it will always be the BSSID.
> 
> Yeah I was thinking it is a bit useless both. Hence the reason to ask. Let me see if I can maybe replace it with some of the info contained in it, like HT, VHT image

I think you can get away with just exposing `multi-band-operation` from usteer ubus, for 99% of cases that should be enough.

Ramon00:
> 
> I think you can get away with just exposing `multi-band-operation` from ubus usteer connected_clients, for 99% of cases that should be enough.

That info is only available on the AP with the clients, not from e.g. my router.

rany2:
@Ramon00 I think you could get rid of it because the frequency field is already exposed, i think the amount of APs with 5ghz and no VHT is close to zero anyway. Same with HT.

Ramon00:
> h 5ghz and no VHT is close to zero anyway. Same with HT.

Yeah I agree, i will remove. If anybody has some suggestion of something useful to extract from it, let me know.

rany2:
BTW great job just some last comments:

- Something like this would be neat: https://github.com/openwrt/luci/pull/6671#issuecomment-1813214434
- For the connected field for the client table, it would be nice if you also stated "False" for when it's not connected. On mobile, it is hard to follow because there is no text on that field when it's not connected. Just looks odd.
- Splitting the status page into separate pages would be good. Right now it's is a bit overwhelming IMO.

Ramon00:
> * Something like this would be neat: [luci-app-usteer: First version of luci-app-usteer #6671 (comment)](https://github.com/openwrt/luci/pull/6671#issuecomment-1813214434)

well that is possible i guess, but the wireless status page has  "mac, host (IPv4, IPv6), if no host then : "mac, IPv4 IPv6."  if no IP address then  "mac, ?" depending on what info is available. I do think that actually the formatting on the wireless status page is a bit inconsistent as it is. So I am not sure.

> * For the connected field for the client table, it would be nice if you also stated "False" for when it's not connected. On mobile, it is hard to follow because there is no text on that field when it's not connected. Just looks odd.

I had a quick look and made it "yes" "no", as that is more to the point and easier to spot the difference if you scroll through the list quickly. Have a look.

> * Splitting the status page into separate pages would be good. Right now it's is a bit overwhelming IMO.

That would be possible of course. I guess the hearing map would then be the candidate to be moved to another page. But what do other people think? Better to have everything in 1 go, or split? I am actually fine either way

rany2:
> That would be possible of course. I guess the hearing map would then be the candidate to be moved to another page. But what do other people think? Better to have everything in 1 go, or split? I am actually fine either way

I believe DAWN luci app already splits those into separate pages, so you might want to look at that for inspiration.

rany2:
Sorry for another comment, but I was configuring a new node using your new luci app and I noticed that:

- band_steering_interval: Attempting to steer clients to a higher frequency-band every n ms. A value of 0 disabled band-steering. (Default: 120000)
- band_steering_min_snr: Minimal SNR or absolute signal a device has to maintain over band_steering_interval to be steered to a higher frequency band. (Default: -60)

were missing from the settings page.

Ramon00:
> Sorry for another comment,

Ah no problem :) Those must have been added after I made the page initially. Check it now!

Ramon00:
@hnyman  Can you have a look at this PR, and approve it if ok?

danpawlik:
LGTM
```
Tested-by: Daniel Pawlik <dduuch@gmail.com>
```

hnyman:
Some comments. I do not normally use usteer, so a bit hard to test and evaluate ;-)
But I installed usteer to my routers, and this PR seems to work. At least it shows the status, hearing maps etc. 

However, a few things regarding the PR:
* Fails formalities test in CI: The commit title should start with "luci-app-usteer:" (like the PR does).
* There are lots of trailing spaces/tabs after content. At least 40 occurences, I think. They should be removed, just for tidying up.  (git commit should actually warn you about that)

For longer development: 
I do not like that this creates a new top-menu item. It should be part of Services or Network menu.

In general, it looks useful for usteer users.

Ramon00:
@hnyman:
> * Fails formalities test in CI: The commit title should start with "luci-app-usteer:" (like the PR does).
That should be easy to fix

> * There are lots of trailing spaces/tabs after content. At least 40 occurences, I think. They should be removed, just for tidying up.  (git commit should actually warn you about that)
I can do that np.

> For longer development:
> I do not like that this creates a new top-menu item. It should be part of Services or Network menu.
Shall i make it into 1 page with 3 tabs instead?

hnyman:
> Shall i make it into 1 page with 3 tabs instead?

That would be better

Ramon00:
oh hmmm i seem to have broken the settings part. Can anybody spot the mistake?

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -0,0 +1,472 @@
...
+		
+		s = m.section(form.NamedSection, 'hearingmap', 'Hearing map');
+		s.anonymous = true;
+
+		o = s.option(HearingMap);
+
+		s = m.section(form.TypedSection, 'settings','Settings',  _('First four options are mandatory. Also be sure to enable rrm reports, 80211kv, etc see: https://openwrt.org/docs/guide-user/network/wifi/usteer'));
+		s.anonymous = true;
+
+		s.option(form.Value, 'network', _('Network'), _('The network interface for inter-AP communication)'));
```
hnyman, line 262: One ) too much?
 But it is inside a string, so it shouldn't hurt, right?

Ramon00: Let me try, but I dont think an extra ) inside a string can be the issue. I think it is related that i switched to` m = new form.JSONMap(mapdata, 'Usteer'); ` instead of `form.Map`, i think that changed the way the uci settings are mapped. If i dont then the different tabs do not work correctly. But I am not sure how to fix it.

hnyman: luci-app-nlbwmon might offer you an example of pretty similar structure. Three pages shown as tabs, one item in menu.

Ramon00: I think its fixed. I moved back to a form.Map and tabbed the section instead of the map.  Please have another look

Ramon00:
@hnyman yes, that would be the next thing to fix. I probably will also add refresh of the status when I get a chance. Thx for the help

Issue #6448, opened 2023-06-27
luci-app-dawn: 500 server error when dawn service is not running


raenye:
## Steps to reproduce:

1. stop dawn service either via CLI (`service dawn stop`) or via LuCI (System → Startup → dawn → Stop).
2. visit DAWN → View Network Overview or visit DAWN → View Hearing Map.

## Actual behavior:

A 500 server error page is displayed (same for both webpages):

<details><summary>Searchable text of error page</summary>

```
Runtime error
Unhandled exception during request dispatching
/usr/lib/lua/luci/ucodebridge.lua:23: /usr/lib/lua/luci/template.lua:181: Failed to execute template 'cbi/map'.
A runtime error occurred: /usr/lib/lua/luci/template.lua:181: Failed to execute template '[string]'.
A runtime error occurred: [string "[string]"]:9: bad argument #1 to 'pairs' (table expected, got nil)
stack traceback:
[C]: in function 'pairs'
[string "[string]"]:9: in main chunk
stack traceback:
[C]: in function 'error'
/usr/lib/lua/luci/template.lua:181: in function
(tail call): ?
/usr/lib/lua/luci/model/cbi/dawn/dawn_network.lua:10: in function 'render'
/usr/lib/lua/luci/cbi.lua:266: in function 'render_children'
[string "/usr/lib/lua/luci/view/cbi/map.htm"]:6: in main chunk

In error(), file [C]
called from function [anonymous function] (/usr/lib/lua/luci/ucodebridge.lua:23)
called from function ((tail call))
In [anonymous function](), file /usr/share/ucode/luci/runtime.uc, line 148, byte 45:
  called from function [arrow function] (/usr/share/ucode/luci/dispatcher.uc:794:4)
  called from function render ([C])
  called from function render_action (/usr/share/ucode/luci/dispatcher.uc:768:24)
  called from function run_action (/usr/share/ucode/luci/dispatcher.uc:795:4)
  called from function [anonymous function] (/usr/share/ucode/luci/dispatcher.uc:998:48)
  called from anonymous function (/www/cgi-bin/luci:39:12)

 `        return lcall.call(modname, method, ...args);`
  Near here ----------------------------------------^
```
</details>

![image](https://github.com/openwrt/luci/assets/59210462/c17aaa83-03f7-493b-be0e-4f47281c270b)

## Expected behavior:
Webpages load OK, and display something like "DAWN is not running".
Possible feature: have a "Start DAWN" button, or add a link to System → Startup.

[...]

PR #6674, opened 2023-11-05
luci-app-dawn: Bug fixes for JavaScript implementation

danielvijge:
Some bug fixes and a small improvements that were discovered after the initial implementation of the JavaScript version of luci-app-dawn:

* Correctly show multiple APs per client in the hearing map
* Display correct name of all APs in the hearing map
* Show if client is connected to the network in the hearing map. This replaces the column for Stations Connected, which is not a property of a client but of an AP, and is available still in the network overview.
* Display both hostname and MAC address for clients/APs
* Convert spaces to tabs in dawn-common.js for consistency

### DIFF
--- a/applications/luci-app-dawn/htdocs/luci-static/resources/dawn/dawn-common.js
+++ b/applications/luci-app-dawn/htdocs/luci-static/resources/dawn/dawn-common.js
@@ -5,65 +5,70 @@
 let callDawnGetNetwork, callDawnGetHearingMap, callHostHints;
 
 callDawnGetNetwork = rpc.declare({
-  object: 'dawn',
-  method: 'get_network',
-  expect: {  }
+	object: 'dawn',
+	method: 'get_network',
+	expect: {  }
 });
 
 callDawnGetHearingMap = rpc.declare({
-  object: 'dawn',
-  method: 'get_hearing_map',
-  expect: {  }
+	object: 'dawn',
+	method: 'get_hearing_map',
+	expect: {  }
 });
 
 callHostHints = rpc.declare({
-  object: 'luci-rpc',
-  method: 'getHostHints',
-  expect: { }
+	object: 'luci-rpc',
+	method: 'getHostHints',
+	expect: { }
 });
 
 function getAvailableText(available) {
-  return ( available ? _('Available') : _('Not available') );
+	return ( available ? _('Available') : _('Not available') );
+}
+
+function getYesText(yes) {
+	return ( yes ? _('Yes') : _('No') );
 }
 
 function getChannelFromFrequency(freq) {
-  if (freq <= 2400) {
-    return 0;
-  }
-  else if (freq == 2484) {
-    return 14;
-  }
-  else if (freq < 2484) {
-    return (freq - 2407) / 5;
-  }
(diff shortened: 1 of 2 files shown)


danielvijge:
@PolynomialDivision As mentioned in https://github.com/openwrt/luci/pull/6656#issuecomment-1793549325 some bug are present in the current implementation. This new pull request fixes this.

The request to (also) display MAC address next to host names is new functionality. From my perspective a hostname means more than a MAC address, so it's now implements as `hostname (MA:CA:DD:RE:SS)`. But is could easily be swapped if, for many use cases, the MAC address is more meaningful, and sorting on MAC address makes more sense.

raenye:
Perhaps you could also address the situation where the dawn service is not currently running? (please see #6448 for ideas).

danielvijge:
The same issue was reported in https://github.com/openwrt/luci/commit/ea8c0aa2a1a1ff9f29ffd452049a7b749298ed17#commitcomment-131699709 The JavaScript version presents the error as:
280452077-9afde8a4-ff20-4591-84b7-14add47ffa9d
Which I think is already way better than the 500 server error previously.

But the same question as there: is there a use case where DAWN is regularly not running when accessing a DAWN page? Or is it OK to display such an error message, and the user should go an investigate why it is not running.

Do you know of any LuCi applications that handle such an error correctly? Then I could copy the code from there. As far as I can tell now, the RPC implementation in LuCi just displays a [fixed error message](https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/rpc.js#L82) without any way to customise it.

jow-:
You can use the `LuCi.rpc.list()` method to check if a given object or procedure exists in ubus, prior to calling it.

Something like this:

```javascript
function isDawnRPCAvailable() {
    return rpc.list("dawn").then(function(signatures) {
        return 'dawn' in signatures && 'get_network' in signatures.dawn && 'get_hearing_map' in signatures.dawn;
    });
}

return view.extend({
	handleSaveApply: null,
	handleSave: null,
	handleReset: null,

	load: function() {
		return Promise.all([
			checkDawnRPCAvailable().then(dawn.callDawnGetHearingMap),
			dawn.callHostHints()
		]);
	},

	render: function(data) {
		const dawnHearingMapData = data[0];
		const hostHintsData = data[1];

		if (!dawnHearingMapData) {
			return E([
				E('h2', _('Dawn service unavailable')),
				E('p', _('Unable to query the dawn service via ubus, the service appears to be stopped.'))
			]);
		}

		...
```

jow-:
An even simpler approach is wrapping the call to `dawn.callDawnGetHearingMap()` into [`L.resolveDefault()`](https://openwrt.github.io/luci/jsapi/LuCI.html#resolveDefault), like this:

```javascript

	load: function() {
		return Promise.all([
			L.resolveDefault(dawn.callDawnGetHearingMap(), null),
			dawn.callHostHints()
		]);
	},
```

Only downside is that there will still be an error logged to the console.

danielvijge:
Thanks @jow- for providing this detailed information. I'll implement this in the current pull request.

PolynomialDivision:
Nice thanks a lot! Would be nice if you could implement @jow- feedback in upcoming PR.

PR #6682, opened 2023-11-07
Firewall: Add masq6 option

schuettecarsten:
Currently, only masquerading for IPv4 can be configured from luci. With this commit, it is also possible to set masq6 option from luci.


jow-:
This already is implemented, the option is placed in the advanced section.

PR #6687, opened 2023-11-11
luci-app-statistics: cleanups of bad spacing

jtkohl:
fixes the output of `/etc/init.d/luci_statistics help`


jtkohl:
@hnyman I just noticed this glitch in the original changes, I botched the help spacing at some point during one of the edits

PR #6688, opened 2023-11-11
luci-app-statistics: harden sysupgrade_backup

jtkohl:
Improve the help for sysupgrade_backup and have it check parameters


jtkohl:
@hnyman have a look.

Given that sysupgrade_backup is really for internal use, perhaps I should remove it from the help and only document the 'backup' option?

hnyman:
I simultaneously said in https://github.com/openwrt/luci/pull/6646#issuecomment-1806858306  ;-)

> sysupgrade_backup is really for internal use, perhaps I should remove it from the help

Yes, please remove it.

PR #6689, opened 2023-11-11
luci-proto-yggdrasil: yggdrasil now supported by netifd

wfleurant:
- this package replaces luci-app-yggdrasil


protocols/luci-proto-yggdrasil/root/usr/libexec/rpcd/luci.yggdrasil
```diff
@@ -0,0 +1,36 @@
...
+				json_add_string "priv" "$PrivateKey"
+				json_add_string "pub" "$PublicKey"
+				json_close_object
+				json_dump
+			;;
+			getPeers)
+				read -r input
+				json_load "$input"
+				json_get_vars interface
+				echo "$(yggdrasilctl -endpoint=unix:///tmp/yggdrasil/${interface}.sock -json getPeers)"
```
jow-, line 32: The `interface` parameter is a root shell injection, please quote the URL string where the variable is interpolated in.

jow- (changes requested):

protocols/luci-proto-yggdrasil/root/usr/libexec/rpcd/luci.yggdrasil
wfleurant: ok thanks good eye
tested with quoted url and removed the call to /bin/echo

wfleurant:
added @turretkeeper support and validation for 0.5.2 features

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+				cell.textContent = '%.2mB'.format(peer.bytes_sent);
+
+				cell = row.insertCell(-1)
+				cell.className = "td"
+				cell.textContent = peer.priority;
+
+				cell = row.insertCell(-1)
+				cell.className = "td"
+				if (!peer.up) {
+					cell.innerHTML = "<u style='cursor: default'>%t ago</u>".format(peer.last_error_time)
```
systemcrash, line 114: Does this produce weird strings? Last error time sounds like an absolute reference, whereas "ago" suffixes an amount (a relative reference). What does ygg output?

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+			o.password=true;
+			o.validate=validatePrivateKey;
+
+			o=s.taboption('general',form.Value,'public_key',_('Public key'),_('The public key for your Yggdrasil node'));
+			o.optional=false;
+			o.validate=validatePublicKey;
+
+			s.taboption('general',cbiKeyPairGenerate,'_gen_server_keypair',' ');
+
+			o=s.taboption('advanced',form.Value,'mtu',_('MTU'),_('Specify an MTU (Maximum Transmission Unit) for your local TUN interface. Default is the largest supported size for your platform. The lowest possible value is 1280.'));
```
systemcrash, line 201: Can MTU explanation be shortened? Does ygg's MTU meaning differ to how MTU is described elsewhere on the platform?

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+			o.optional=true;
+			o.placeholder=1280;
+			o.datatype='range(1280, 65535)';
+
+			o=s.taboption('general',form.TextValue,'node_info',_('Node info'),_('Optional node info. This must be a { "key": "value", ... } map or set as null. This is entirely optional but, if set, is visible to the whole network on request.'));
+			o.optional=true;
+			o.placeholder="{}";
+
+			o=s.taboption('general',form.Flag,'node_info_privacy',_('Node info privacy'),_('By default, node info contains some defaults including the platform, architecture and Yggdrasil version. These can help when surveying the network and diagnosing network routing problems. Enabling node info privacy prevents this, so that only items specified in "Node info" are sent back if specified.'));
```
systemcrash, line 210: Perhaps something like
```
_('Enable node info privacy so that only items specified in "Node info" are sent back.')
```

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+				s.tab('peers',_('Peers'));
+			} catch(e) {};
+			o=s.taboption('peers', form.SectionValue, '_active', form.NamedSection, this.sid, "interface", _("Active peers"))
+			ss=o.subsection;
+			ss.option(cbiActivePeers, this.sid);
+
+			o=s.taboption('peers', form.SectionValue, '_listen', form.NamedSection, this.sid, "interface", _("Listen for peers"))
+			ss=o.subsection;
+
+			o=ss.option(form.DynamicList,'listen_address',_('Listen addresses'),_('Listen addresses for incoming connections. You will need to add listeners in order to accept incoming peerings from non-local nodes. Multicast peer discovery will work regardless of any listeners set here. Each listener should be specified in URI format, e.g.tls://0.0.0.0:0 or tls://[::]:0 to listen on all interfaces.'));
```
systemcrash, line 223: `_('Add listeners in order to accept incoming peerings from non-local nodes. Multicast peer discovery works regardless of listeners set here. URI Format: <code>tls://0.0.0.0:0</code> or <code>tls://[::]:0<code> to listen on all interfaces.')`

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+			o.placeholder="tls://0.0.0.0:0"
+			o.validate=validateYggdrasilListenUri;
+
+			o=s.taboption('peers',form.DynamicList,'allowed_public_key',_('Allowed public keys'),_('List of peer public keys to allow incoming peering connections from. If left empty then all connections will be allowed by default. This does not affect outgoing peerings, nor does it affect link-local peers discovered via multicast.'));
```
systemcrash, line 227:
```js
o=s.taboption('peers',form.DynamicList,'allowed_public_key',_('Accept from public keys'),_('If empty, all connections will be allowed (default). This does not affect outgoing peerings, nor does it affect link-local peers discovered via multicast.'));
```

systemcrash (changes requested):

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+		return _('Invalid public key string');
+	return true;
+};
+
+function validateYggdrasilListenUri(section_id,value) {
+	if (value.length == 0) {
+		return true;
+	};
+	if (!value.match(/^(tls|tcp|unix|quic):\/\//))
+		return _('URI scheme not supported');
```
systemcrash, line 37: Quote `value` in the return string so we know which URI was problematic: multiple URIs are possible.

systemcrash: e.g. 
```js
 return _('unsupported URI scheme in ' + value);
```

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+	}
+	return true;
+};
+
+function validatePublicKey(section_id,value) {
+	if (value.length == 0) {
+		return true;
+	};
+	if (!value.match(/^([0-9a-fA-F]){64}$/))
+		return _('Invalid public key string');
```
systemcrash, line 28: e.g. 
```js
 return _('Invalid public key string %s').format(value);
```

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
systemcrash: or...
e.g. 
```js
 return _('Unsupported URI scheme in %s').format(value);
```

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+		return true;
+	};
+	if (!value.match(/^(tls|tcp|unix|quic):\/\//))
+		return _('URI scheme not supported');
+	return true;
+};
+
+function validateYggdrasilPeerUri(section_id,value) {
+	if (!value.match(/^(tls|tcp|unix|quic|socks|socktls):\/\//))
+		return _('URI scheme not supported');
```
systemcrash, line 43: e.g. 
```js
 return _('Unsupported URI scheme in %s').format(value);
```

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
```diff
@@ -0,0 +1,272 @@
...
+
+var cbiKeyPairGenerate = form.DummyValue.extend({
+	cfgvalue: function(section_id, value) {
+		return E('button', {
+			'class':'btn',
+			'click':ui.createHandlerFn(this, function(section_id,ev) {
+				var prv = this.section.getUIElement(section_id,'private_key'),
+					pub = this.section.getUIElement(section_id,'public_key'),
+					map = this.map;
+				if ((prv.getValue()||pub.getValue()) && !confirm(_('Do you want to replace the current keys?')))
```
systemcrash, line 55: Remove this confirmation: the user has the opportunity to review what changes will be applied as a whole when they go to apply. Make changing the config friction free.

systemcrash (changes requested):

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: Yes by default (and as a feature) the MTU is maxed to `65535`

- [ ] We should reduce the text to quickly explain the default 65535 is optimal for yggdrasil

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: ok i will have to test with new examples, but prior to this i swear i saw something like:  

:small_red_triangle: URI scheme not supported  

it was immediatly under any typo'd url field... i'll doublecheck

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: thanks, i also spot a typo there i'll fix

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: ok and ill add for private key

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: - [ ] add format

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: changed placeholder to `65535`
added `A default MTU of 65535 is set by Yggdrasil. It is recomended to utilize the default.`

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: taken as-is, thanks this reads well.

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: - added the wording and added compatible URIs 
image

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: aight, thanks again

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: took the suggested format method -- much appreciated.

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: done, looks good on mouseover

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: fixed the typo `socktls` is now `sockstls` .. and added format suggestion thanks again

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: bit the bullet, removed public_key option field since it cannot be obtained during generation. its also ignored by the daemon, but.. it is sort-of important and we will have to raise a feature request to yggdrasil devs. until then, its removed.

removed the confirm popup, works fine.

protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
wfleurant: nothing else to do here beyond making a merge request in yggdrasil for this time edgecase... i will try to reproduce this later.

the 'last error' description text is useful and shouldnt be hard to find. this will need more testing. I do not believe this should block us going forward.

wfleurant:
thanks a ton for reviewing @jow- and @systemcrash a lot was fixed.

protocols/luci-proto-yggdrasil/root/usr/libexec/rpcd/luci.yggdrasil
wfleurant: the code now reads:
```bash
...
getPeers)
  read -r input
  json_load "$input"
  json_get_vars interface
  yggdrasilctl -endpoint="unix:///tmp/yggdrasil/${interface}.sock" -json getPeers
;;
```

this should suffice? interface is eg; yggdrasil-tun2 or what is defined in uci/network  
ie: `yggdrasil-<interface>`

systemcrash:
If this package replaces app, go ahead and add a commit which deletes the app version.

turretkeeper:
@systemcrash any way to restart the underlying service on config change? We don't know if there is a better way than a busy loop, but the daemon needs a restart to work under the new uci configuration.

wfleurant:
> If this package replaces app, go ahead and add a commit which deletes the app version.

done in e6a1f119e6ec2eed48ea583d8176c21bf658a499

systemcrash:
> @systemcrash any way to restart the underlying service on config change? We don't know if there is a better way than a busy loop, but the daemon needs a restart to work under the new uci configuration.

@turretkeeper Not looked into this, though one might do well to see how other apps do it, e.g. luci-app-ddns or luci-app-https-dns-proxy. I know that wireguard still requires a (manual) restart when changes have been made.

protocols/luci-proto-yggdrasil/root/usr/libexec/rpcd/luci.yggdrasil
wfleurant: is this change trash? sorry, please advise.

protocols/luci-proto-yggdrasil/root/usr/libexec/rpcd/luci.yggdrasil
```diff
@@ -16,12 +16,12 @@ case "$1" in
...
 			generateKeyPair)
 				json_load "$(yggdrasil -genconf -json)"
-				json_get_vars PublicKey PrivateKey
+				json_get_vars PrivateKey
 				json_cleanup
 				json_init
 				json_add_object "keys"
 				json_add_string "priv" "$PrivateKey"
-				json_add_string "pub" "$PublicKey"
+				json_add_string "pub" "${PrivateKey:64}"
```
systemcrash, line 24: So my eyes are naïve here but the public half is the private key truncated at 64 bytes? If that's the way things are, that's fine. 

But this isn't the more cpu intensive genkeys variation is it? https://yggdrasil-network.github.io/configuration.html

My concern here is whether the GUI would time out if the operation takes a long time.

wfleurant: this is whats going on.
we would need to build and package genkeys. i would like to see a 'generate keys' button with possibly some input (a checkbox or some sort of input/dropdown) to try and find a key within say a 1 to 3 (or whatever) timeout duration.

the current -genkeys is (i think) first find first use.

we would need to make some sort of non-blocking operation to poll for the existence of /tmp/genkeys.txt (or something).. when the end of the duration is hit. i last checked and we do not have a `timeout` with busybox.. so... genkeys will max out all logical cpu cores when executed. i think we could patch the genkeys.go file with some router-friendly parameters eg; -timeout and -save-last-key-to-file

fair to say this is out of scope of LuCI for now.. i'll see if i cant patch/upstream something worth using.

if we get this going, how do we do it in LuCI? is there a non-blocking way to go about this?

as a user, i wouldnt care if i did 'generate key' and 'key generation timeout=10min' ... if i could set it and come back in 10 min to see available keys (or last key it found) .. sure i'll take that.

turretkeeper: > this is whats going on. we would need to build and package genkeys. i would like to see a 'generate keys' button with possibly some input (a checkbox or some sort of input/dropdown) to try and find a key within say a 1 to 3 (or whatever) timeout duration.
Most router CPUs would probably not be good at iterating through many keypairs. If you were to mine it in Javascript, though...

systemcrash: In terms of protocol design in yggdrasil, it seems... terrible. I'm just making a drive-by comment here, rather than attempting to appreciate the reasoning behind this choice. 

On the surface, this seems like a valid suggestion, but... Does there exist an Ed25519 JS implementation to even go this route? Waiting for 10 minutes until a valid configuration can be built does not seem very user-friendly. Websites are dead in the water if they don't respond within 7 seconds.

wfleurant: key generation is performed only once when a configuration is generated.

dvn0: > the current yggdrasil -genconf is (i think) first find and first use.

my 2 cents as an observer (as well as longtime yggdrasil user and occasional openwrt contributor) is that it would be beneficial to get this PR merged with just the "first find and first use" key generation, as that is what most people should/would need.

jow- (approved):

PR #6690, opened 2023-11-12
luci-app-yggdrasil: remove package in favor of luci-proto-yggdrasil

wfleurant:


wfleurant:
sorry was out of town, left it WIP after opening it

PR #6695, opened 2023-11-13
luci-app-yggdrasil status. Fix "cannot convert undefined or null to object" Error

Tygernaut:
Fix "cannot convert undefined or null to object" Error on status.js


wfleurant:
i am not able to reproduce this. which versions of yggdrasil and luci-app-yggdrasil?

Tygernaut:
> Member
yggdrasil | 0.4.7-1
luci-app-yggdrasil | git-21.202.59029-3953b1a

systemcrash:
Closed in favour of #6690 

@wfleurant wonder whether you could cook up a new status (luci-mod-status) page now that we have proto?

wfleurant:
> @wfleurant wonder whether you could cook up a new status (luci-mod-status) page now that we have proto?

yes! started to look and rip something together...

PR #6697, opened 2023-11-14
luci-proto-wireguard: Remove confirmation dialogue to generate keys

systemcrash:
Quality of life improvements. Reduce click amounts. LuCI batches all changes for user-review anyway.

Tested on 23.05.0


jow-:
I don't know. A lot of people simply smash save & apply without ever reviewing changes. Also that crypto material is a bit more important than generic settings which can be reverted to defaults or restored from remembered settings...

I tend to NAK this one.

systemcrash:
OK, I understand your concern. Although the private key can be reverted just like any other setting.

PR #6698, opened 2023-11-14
luci-mod-network: fix route defaults

vgaetera:
Routes inherit metric and table from the relevant interface.
Display route metric and table as `auto` when unspecified.
Consolidate modal view for unspecified options.


jow-:
Merged, thanks!

PR #5932, opened 2022-08-14
luci-app-libreswan: Add LuCI for Libreswan

jempatel:
LuCI Support for IPSec VPN (Libreswan)

Signed-off-by: Jaymin Patel <jem.patel@gmail.com>

Depends On PR
https://github.com/openwrt/packages/pull/19079

PR #6700, opened 2023-11-15
luci-app-strongswan-swanctl: initial commit

lvoegl:
**Adds strongSwan status and configuration page based on #6566.**
The configuration matches the latest [UCI strongSwan configuration](https://openwrt.org/docs/guide-user/services/vpn/strongswan/configuration) with `strongswan-swanctl` backend.
Remote Configuration
Tunnel Configuration
Crypto Proposal Configuration

The status page relies on openwrt/packages#22354 for querying strongSwan data as JSON.
Status Page


systemcrash:
So... I'm going to repeat my comment from the other PR here: I think it's worthwhile to figure out whether the [recent libreswan addition](https://github.com/openwrt/luci/pull/5932) can be co-opted for use with strongswan also. Libreswan derive from the strongswan code-base and they both provide largely identical functionality and configurability. 

Abstractly, that the libreswan GUI recognizes the presence of strongswan, possibly via simple symlinks. I experienced that strongswan is the less buggy and more mature of the two, having used both. But here we are.

The best way to go about that may require input from those who know the system best: @jow- and @hnyman I think. 

The above swanmon might be necessary: I think this functionality on other platforms is provided by pluto.

systemcrash:
This is all to say: both systems and GUI look good. But if they're largely identical, why have two? Can we have one?

If we can't then this effort looks like a worthy inclusion.

feckert:
@systemcrash There are various IPSec implementations in OpenWrt that are not compatible with each other. They all have their own parameter set. The libreswan and strongswan (via swanctl) may now have the same parameters, but this does not always have to be the case. In addition, we always have to take two things into account during the review if something changes in this App. For example, I don't have a libreswan setup and so could net check this. I would say we need to have one app for each IPsec service implementation.

tiagogaspar8:
Hi @lvoegl,

If it's not too much to ask, when you update this, can you also include the new replay window option merged today?

Thanks!

systemcrash:
Please put changes in the GUI which alert users to _broken_ and _dangerous_ cryptos. 

https://www.rfc-editor.org/rfc/rfc8247#section-2.3 this document is already six years old, and computing resources have improved since then.

All cryptos can remain in an attempt to be useful or compatible, but should be clearly marked as broken or dangerous, e.g. MD5, SHA1, 3DES, as well as all MODP groups less than 2048 bit. 

And also https://www.rfc-editor.org/rfc/rfc9395#name-iana-considerations
IKEv1 is *deprecated*

Mark deprecated and dangerous cryptos with e.g. `*` as done in #5932 - then translation strings might also utilize the same strings.

lvoegl:
> Please put changes in the GUI which alert users to _broken_ and _dangerous_ cryptos.
> 
> https://www.rfc-editor.org/rfc/rfc8247#section-2.3 this document is already six years old, and computing resources have improved since then.
> 
> All cryptos can remain in an attempt to be useful or compatible, but should be clearly marked as broken or dangerous, e.g. MD5, SHA1, 3DES, as well as all MODP groups less than 2048 bit.
> 
> And also https://www.rfc-editor.org/rfc/rfc9395#name-iana-considerations IKEv1 is _deprecated_
> 
> Mark deprecated and dangerous cryptos with e.g. `*` as done in #5932 - then translation strings might also utilize the same strings.

Insecure algorithms (see https://docs.strongswan.org/docs/5.9/config/IKEv2CipherSuites.html) are now marked with a *
Crypto Proposal

systemcrash:
Sorry, just had a thought: if 'both' contains IKEv1 which is deprecated, then both should also be marked deprecated, I think. IKEv2 alone is fine. Nice if this goes on, otherwise I am satisfied here.

ptpt52:
mark

zcracker:
> mark
，
Chinese ISP blocks ipsec port，Unicom，Telecom

zcracker:
> Ok, so we're just waiting for [openwrt/packages#22354](https://github.com/openwrt/packages/pull/22354) ?


Swanmon has been merged, please merge it

systemcrash:
@lvoegl https://github.com/openwrt/luci/commit/828716ad73be81079ffdaa1592f637efa1747f31#commitcomment-135220721

Issue #6702, opened 2023-11-17
luci-app-vnstat2: XHR request timed out accessing config


jkugler:
## Steps to reproduce:
I'm using vnstat2 to track bandwidth usage, and it's working great.

Going to /cgi-bin/luci/admin/status/vnstat2/graphs shows all my graphs, and everything looks good.

However, if I go to '/cgi-bin/luci/admin/status/vnstat2/config, I get an error: "XHR request timed out" and it will not display the config.

I've looked through the logs, and haven't seen any messages from vnstat2 or Luci. 

I can upload the DB file somewhere if you wish.

## Actual behavior:
XHR timeout

## Expected behavior:
Displays configuration screen

## Additional Information:
OpenWrt version information from system `/etc/openwrt_release`

```
# cat /etc/openwrt_release 
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.5'
DISTRIB_REVISION='r20134-5f15225c1e'
DISTRIB_TARGET='ipq40xx/generic'
DISTRIB_ARCH='arm_cortex-a7_neon-vfpv4'
DISTRIB_DESCRIPTION='OpenWrt 22.03.5 r20134-5f15225c1e'
DISTRIB_TAINTS='
```
vnstat2 config:
```
# cat /etc/config/vnstat

config vnstat
        list interface '6in4-wan6'
        list interface 'pppoe-wan'
```
and the conf file:
```
# cat /etc/vnstat.conf
# vnStat 2.9 configuration file
#
# lines starting with # or ; are comments, everything has default
# values, remove ; before each option to change its value


# default interface (leave empty for automatic selection)
;Interface ""

# location of the database directory
DatabaseDir "/mnt/sda1/vnstat"

# locale (LC_ALL) ("-" = use system locale)
;Locale "-"

# date output formats for -d, -m, -t and -w
;DayFormat    "%Y-%m-%d"
;MonthFormat  "%Y-%m"
;TopFormat    "%Y-%m-%d"

# characters used for visuals
;RXCharacter       "%"
;TXCharacter       ":"
;RXHourCharacter   "r"
;TXHourCharacter   "t"

# how units are prefixed when traffic is shown
# 0 = IEC standard prefixes (KiB/MiB/GiB...)
# 1 = old style binary prefixes (KB/MB/GB...)
# 2 = SI decimal prefixes (kB/MB/GB...)
;UnitMode 0

# used rate unit (0 = bytes, 1 = bits)
[...]

PR #6704, opened 2023-11-18
luci-app-vnstat2: use more efficient way to get database interfaces

janh:
The currently used command "vnstat --json f 1" can be very slow when the database contains many entries. This can result in an XHR timeout error on the configuration page.

Use the "--dbiflist" option instead, which was added in vnStat 2.7.

Fixes: #6702

---

This should also be cherry-picked for 23.05 and 22.03.


jkugler:
Thanks much!

PR #6706, opened 2023-11-18
luci-app-firewall: allow redirects using ip family any

vgaetera:
Allow creating redirects using IP family `any`.
This helps redirect both IPv4 and IPv6 traffic.
It is used to intercept traffic on the router.

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/forwards.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/forwards.js
@@ -9,7 +9,7 @@
 'require tools.widgets as widgets';
 
 function rule_proto_txt(s, ctHelpers) {
-	var family = (uci.get('firewall', s, 'family') || '').toLowerCase().replace(/^(?:any|\*)$/, '');
+	var family = (uci.get('firewall', s, 'family') || '').toLowerCase().replace(/^(?:all|\*)$/, 'any');
 	var dip = uci.get('firewall', s, 'dest_ip') || '';
 	var proto = L.toArray(uci.get('firewall', s, 'proto')).filter(function(p) {
 		return (p != '*' && p != 'any' && p != 'all');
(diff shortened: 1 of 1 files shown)


systemcrash:
Isn't `any` just a synonym for automatic? The wiki says:

"Protocol family (ipv4, ipv6 or any) to generate iptables rules for. "

( OK the wiki needs updating since fw4 does not use iptables any longer )

But the point is still valid - `any` just means automatic. Wouldn't it be better simply to Rename all automatic to e.g. `Either` or `Both`?

vgaetera:
An unspecified address family makes fw4 create an IPv4-only redirect.
This was also mentioned by @jow- here:
https://forum.openwrt.org/t/current-state-of-nftables-end-of-2020/80056/16

systemcrash:
Unspecified is the empty value, yes? Didn't know that.

So both the wiki need updating and this merging then.

vgaetera:
> Unspecified is the empty value, yes?

To be clear, an empty `family` option works like this for redirects:
* It generates an IPv6 redirect when an IPv6 source or destination address is specified.
* It generates an IPv4 redirect when an IPv4 source or destination address is specified.
* It generates an IPv4 redirect when no source or destination address is specified.

So, a dual-stack redirect requires to specify the `family=any` which is the goal of this PR.

> the wiki need updating

Done.

jow-:
Hmm, I think there's further changes required. We probably should disable the source and destination address fields if "any" is selected.

Can you provide an example uci rule requiring family any? Is this for port redirection?

vgaetera:
> Can you provide an example uci rule requiring family any?

https://openwrt.org/docs/guide-user/firewall/fw3_configurations/intercept_dns

systemcrash:
> We probably should disable the source and destination address fields if "any" is selected.

@vgaetera is that something you'd be comfortable doing also in this PR?

vgaetera:
The current validation blocks the user from saving the rule if the combination of the IP family and address do not match.
I tested the patched web interface and this already works well, so adding more restrictions would be overkill.

PR #6707, opened 2023-11-19
luci-proto-vxlan: Enable IPv6 address and port settings

ElbertMai:
The VXLAN6 settings form accepts only CIDR notation for IP addresses, but VXLAN6 does not accept such addresses. Attempting to use an address ending with /128 does not bring up the interface. Also, the form is missing the destination port option, unlike its IPv4 counterpart. This commit fixes both these issues.


jow-:
Merged, thank you!

PR #6711, opened 2023-11-20
luci-mod-network: nitpick of CNAME, MX, SRV syntax formatting

systemcrash:


feckert (commented):
Isn't that misleading? Do you even need the dot at the end?

systemcrash:
Not strictly, but it *is* correct and disambiguates.

When you specify a domain name without the ".", it's considered a relative domain name and is appended with the current domain name or the local domain, depending on the context. Including the "." at the end makes it an absolute domain name, ensuring that it's explicitly pointing to the root of the DNS hierarchy.

https://en.wikipedia.org/wiki/SRV_record:
- `target: the canonical hostname of the machine providing the service, ending in a dot.`

See also:
https://en.wikipedia.org/wiki/CNAME_record
https://en.wikipedia.org/wiki/MX_record
https://en.wikipedia.org/wiki/Root_name_server

PR #6717, opened 2023-11-23
re-draw favicon in svg and re-render it

532910:
1. manually vectorized favicon in svg
2. re-render 16px from svg
3. add 32px
4. "shortcut icon" -> "icon", see mdn: https://developer.mozilla.org/docs/Web/HTML/Attributes/rel#icon


stokito:
The repo already contains the SVG logo here:
https://github.com/openwrt/luci/blob/master/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/logo.svg
Also please check the page:
https://openwrt.org/docs/guide-graphic-designer/openwrt-logo
On the forum there was a lot of log related discussions:
https://forum.openwrt.org/

I personally not sure if the logo needs for an improvement: this adds some complexity and additional size on flash disk. Not so critical to have the icon in a bigger resolution since users visits a router not so often. But maybe for the OpenWrt official website that would be a great addition.

532910:
1. openwrt icon differs from this one
2. this pr is not about icon change, but improvement
3. not sure if to have the same icon for luci instance and all other openwrt websites is a good idea.

stokito:
It looks like we have some mess here. As far I understood you found that the themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/favicon.png is poorly looking. And you created the 32px version that looks better.
It looks like all themes have their own favicon:
* themes/luci-theme-material/htdocs/luci-static/material/favicon.ico
* themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/logo.svg

The openwrt2020 theme additionally has a PNG for `apple-touch-icon` themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/logo.png

I think that we should just remove the logo.png and the `apple-touch-icon` because previously this was needed because Safari didn't supported the SVG icons but now it does https://stackoverflow.com/a/25079463/1049542

Today almost all browsers support the SVG icons https://caniuse.com/link-icon-svg

The openwrt2020 uses it and it's looks fine:
luci-logo

I think that all the themes should use exactly the same SVG icon like in the openwrt2020: 

    <link rel="icon" href="{{ media }}/logo.svg" type="image/svg+xml" />

The `type="image/svg+xml"` may be required if a web server (e.g. uhttpd) doesn't return a valid mime type (but it does). So we can remove it.

Even if some browser doesn't support for now it won't be critical and eventually all will be settled. Code similicity and is more important.

Are you fine with the plan to use SVG icon?

systemcrash:
@532910 can you give the other themes the same treatment, and also e.g.

```html
<link rel="icon" href="/icon.png" type="image/png">
<link rel="icon" href="/icon.svg" type="image/svg+xml">
```

stokito:
The 2020 theme had an incorrect icon.
According to the Logo Usage Guide it must have a light blue background https://openwrt.org/_media/docs/guide-graphic-designer/openwrt-logo-usage-guidelines.pdf
This is how it looks fixed:
https://github.com/openwrt/luci/blob/7fc816ac6903bb1036800aefb39fd42f62ea1d7f/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/logo.svg

The OpenWrt.org uses a different logo: 

Blue & Dark Blue with white background.

So we'll have different logos for routers.

stokito:
@532910 
I like the icon that you vectorized to SVG. Still the OpenWrt logo should be used according to a design guide that specifies four variants.
Also in the version that you made the resulting SVG is really big: 4kb. It can be minimized significantly.

I created a separate PR #6738 that uses the existing logo.svg from the 2020 theme.
Please check it an leave feedback. You can approve the PR if you think it's ok. Your approval is not needed for a merge but it will show that the PR was reviewed.
Thank you for your contribution and rising the issue.

532910:
> Also in the version that you made the resulting SVG is really big: 4kb. It can be minimized significantly.

Sure, but it was attached as a source and not intended for web.

PR #6709, opened 2023-11-19
luci-app-advanced-reboot: add support for Mercusys MR90X v1

Moondarker:
screenshot

You might want to see [UBI volume structure](https://openwrt.org/toh/mercusys/mr90x_v1#ubi_sctructure) of this device to understand some changes

I didn't want to intrude into the bash part too much, but I think other devices might benefit from certain changes in the future. It's completely backwards compatible and won't affect previously added devices

~~I will mark it as draft for now for more tests, but pretty sure it's more or less final unless reviewers request any changes :)~~

Tested, works perfectly. To reboot back from OEM firmware one would need to:
1. Connect using ssh to [router ip]:20001
2. Edit `/etc/hotplug.d/iface/65-iptv`, adding `telnetd -l /bin/login.sh` on a newline after `#!/bin/sh`
3. Toggle on `IPTV/VLAN Enable` in the web interface
4. Telnet into the router
5. Check `fw_printenv | grep tp_boot_idx`
6. Do `fw_setenv tp_boot_idx 0` or `fw_setenv tp_boot_idx 1`, whichever it wasn't set to in step 5

P.S. I don't understand the meaning behind [this +1 offset](https://github.com/openwrt/luci/blob/41cee29f3bc832bbadc24786f082fddfb78c009d/applications/luci-app-advanced-reboot/root/usr/libexec/rpcd/luci.advanced_reboot#L189), it always shifted the partition off by one for me, must be some device-specific thing?

```
root@OpenWrt:~# ubus call system board
...
        "system": "ARMv8 Processor rev 4",
        "model": "MERCUSYS MR90X v1",
        "board_name": "mercusys,mr90x-v1",
        "rootfs_type": "squashfs",
...
```
```
root@OpenWrt:~# cat /tmp/sysinfo/board_name
mercusys,mr90x-v1
```
```
root@OpenWrt:~# cat /proc/mtd
dev:    size   erasesize  name
mtd0: 00200000 00020000 "boot"
mtd1: 00100000 00020000 "u-boot-env"
mtd2: 03200000 00020000 "ubi0"
mtd3: 03200000 00020000 "ubi1"
mtd4: 00800000 00020000 "userconfig"
mtd5: 00400000 00020000 "tp_data"
```
```
root@OpenWrt:~# fw_printenv
...
loadaddr=0x46000000
mtdids=spi-nand0=spi-nand0
[...]

PR #6719, opened 2023-11-24
luci-app-advanced-reboot: bump version number due to changes in code

Moondarker:
Bumps version of `luci-app-advanced-reboot` package in response to changes introduced in #6709, @stangri


hnyman:
@stangri 
How about reverting back to the default automatic incremental commit based LuCI versioning?

stangri:
@hnyman I remember there was an issue with that versioning and builds in my private repo always being behind builds in the package release, so they could not be easily updated thru `opkg update`.

PR #6723, opened 2023-11-27
luci-mod-status: highlight primary 20 MHz channel

knarrff:
before
after
(The reason the strongest station from image 1 is missing in image 2 is simple: it was one of mine, which simply is not there anymore.)

As of now, channel analysis highlights the complete frequency spectrum of a base station (all channels used by the station, which might be more or fewer depending on channel bonding). This means in particular, that in the plot, a station at one main channel might look identical to another station at a different channel, because both might use the same bonded channel spectrum.

One such example is two 80 MHz stations on channels 100 and 108 respectively: both will be plotted from channel 100 up to 115, as both will use those those 80 MHz. This is not incorrect, but it makes it more difficult to see which station is really where, as they might as well switch back to 20 or 40 MHz at times, or for different clients, and where they are then is currently not visible in the plot.

This patch adds another trapez to the plot at the main 20 MHz channel of a station, with width 20 MHz, in case the station uses a wider general width. This is visible, because those polygons are partially transparent. It also moves the x-position of the label to the main channel of the station (which now is marked: this would have looked odd before).

This does not highlight the primary 40 MHz channel for 80 MHz channels and wider. It should be possible be compute this from the knowledge in luci of the primary 20 MHz channel, the center channel and the bandwidth, assuming ac logic, but I decided against adding it, to keep the plot less cluttered.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/svg/channel_analysis.svg
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/svg/channel_analysis.svg
@@ -16,4 +16,21 @@
 
 	<line x1="0" y1="90%" x2="100%" y2="90%" style="stroke:black;stroke-width:0.1" />
 	<text id="label_10" x="10" y="89%" style="fill:#eee; font-size:9pt; font-family:sans-serif; text-shadow:1px 1px 1px #000">-90 dbm</text>
+	<defs>
+		<linearGradient id="GradientVerticalCenteredBlack">
+			<stop class="stopinvis" offset="0%" />
+			<stop class="stopblack" offset="50%" />
+			<stop class="stopinvis" offset="100%" />
+		</linearGradient>
+	</defs>
+	<style>
+		.stopinvis {
+			stop-color: "transparent";
+			stop-opacity: 0;
+		}
+		.stopblack {
+			stop-color: "black";
+			stop-opacity: .1;
+		}
+	</style>
 </svg>
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
@@ -63,38 +63,50 @@ return view.extend({
 		if (scanCache[res.bssid].graph == null)
 			scanCache[res.bssid].graph = [];
 
-		channels.forEach(function(channel) {
+		function add_channel_to_graph(chan_analysis, res, scanCache, i, channel, channel_width, label) {
 			var chan_offset = offset_tbl[channel],
 				points = [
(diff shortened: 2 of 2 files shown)


castillofrancodamian:
Could you review [this](https://github.com/openwrt/luci/issues/6419)?

knarrff:
I did have a look and commented there. The PR here should be independent though, as #6419 is about a shifted 20 MHz station in the plot, and this PR here only adds something for channels wider than 20 MHz, so it leaves 20 MHz channels as before.

systemcrash:
> This does not highlight the primary 40 MHz channel for 80 MHz channels and wider. It should be possible be compute this from the knowledge in luci of the primary 20 MHz channel, the center channel and the bandwidth, assuming ac logic, but I decided against adding it, to keep the plot less cluttered.

@knarrff Could you show us just how cluttered it looks for 40 and 80 MHz please? ("why treat only 20 when we can treat all of them...")

knarrff:
@systemcrash: I can't show it at the moment, without implementing it in the first place. I do not think it would be particularly difficult to implement, just tedious.

Ideally, if you want to go that route, some more logic would be nice to easily get to all necessary values (like the center frequency of 40 MHz channels if the overall bandwidth is larger). This could come in handy also someplace else than channel_analysis.js, and as such should be provided to channel_analysis.js instead of living inside it. iwinfo comes to mind, but then this becomes an entirely different matter.

systemcrash:
Is it possible to provide some extra hinting in the SVG which helps anchor the fact that this is a primary channel within this SSID, rather than how it looks now: possibly two overlapping SSIDs?

systemcrash:
That looks pretty sweet.

Is that gradient or blur an expensive function?

knarrff:
I have adapted to the gradient highlight. It is not expensive, as this is a very standard svg gradient and nothing that would be implemented in js. Also, because this way, the drawing of the main channel (20 MHz) does something different than the drawing of the complete channel (>20 MHz), I reworked the original patch to make it simpler in the end and also smaller compared with current master. The change to the svg only adds the definition of the gradient itself (thus, no visible change), which is then later referenced in the js code.

systemcrash:
@knarrff Could you update the header graphics demonstrating the PR changes please?

knarrff:
> @knarrff Could you update the header graphics demonstrating the PR changes please?

I updated the image in the main description with the one from https://github.com/openwrt/luci/pull/6723#issuecomment-1843815093

systemcrash:
@knarrff OK, are you happy with how things are now?

@jow- any reservations on this cosmetic change?

knarrff:
> @knarrff OK, are you happy with how things are now?

Yes.

PR #6724, opened 2023-11-27
luci-mod-status: fix ACL for channel analysis

erikrk:
Without this fix channel analysis will not work unless write access is also granted to luci-mod-network-config

### DIFF
--- a/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
+++ b/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
@@ -61,7 +61,7 @@
 		"description": "Grant access to wireless channel status",
 		"read": {
 			"ubus": {
-				"iwinfo": [ "info", "freqlist" ]
+				"iwinfo": [ "info", "freqlist", "scan" ]
 			}
 		}
 	},


jow-:
Hm, not sure about this. Scan is a disruptive operation, potentially impacting existing connections, so not sure if it should be allowed in a read only context.

erikrk:
The page does not work unless scan is allowed. Is has only been working "by accident" because of overlap with luci-mod-network-config

Ansuel:
I think the page should only be accessible by ""admin"" user as on some driver the scan may cause DenialOfService problem...

Maybe a workaround might be lock the refresh button under some kind of permission??? (should limit the abuse in some context)

erikrk:
Can't whoever who is worried about abuse just refrain from granting read access to luci-mod-status-channel_analysis? As a comparison, luci-mod-network-diagnostics could also in principle be abused and this is also read access.

The problem with having scan as write access is then if you only grant read access, you end up with a broken page. Note that with the current set of permissions the page does not work, i.e. this is a bug fix. It only happens to work on a lot of occasions because write access to luci-mod-network-config is being granted simultaneously.

Regarding locking out or rate limiting the use of the refresh button, this would not really guard against DoS attacks since the logic for this would realistically be in the frontend.

PR #6737, opened 2023-12-04
luci-mod-network: DHCP tab IPsets ( add NFTsets properties )

systemcrash:
See also:

https://github.com/openwrt/openwrt/commit/d7f378796f985c902ff7906767c275c40a15347f

@jow- Is a straight rename safe? Since fw4 is default now.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -822,21 +822,40 @@ return view.extend({
 		});
 
 		o = s.taboption('ipsets', form.SectionValue, '__ipsets__', form.GridSection, 'ipset', null,
-			_('List of IP sets to populate with the IPs of DNS lookup results of the FQDNs also specified here.'));
+			_('List of IP sets to populate with the IPs of DNS lookup results of the FQDNs also specified here.') + '<br />' +
+			_('The netfilter components below are only regarded when running fw4.'));
 
 		ss = o.subsection;
 
 		ss.addremove = true;
 		ss.anonymous = true;
 		ss.sortable  = true;
+		ss.rowcolors = true;
+		ss.nodescriptions = true;
+		ss.modaltitle = _('Edit IP set');
 
-		so = ss.option(form.DynamicList, 'name', _('IP set'));
+		so = ss.option(form.DynamicList, 'name', _('Name of the set'));
 		so.rmempty = false;
+		so.editable = true;
 		so.datatype = 'string';
 
-		so = ss.option(form.DynamicList, 'domain', _('Domain'));
+		so = ss.option(form.DynamicList, 'domain', _('FQDN'));
 		so.rmempty = false;
+		so.editable = true;
 		so.datatype = 'hostname';
+
+		so = ss.option(form.Value, 'table', _('Netfilter table name'), _('Defaults to fw4.'));
+		so.editable = true;
+		so.placeholder = 'fw4';
+		so.rmempty = true;
+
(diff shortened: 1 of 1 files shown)


dave14305:
Why not just call them “Sets” since the same config can be used to populate fw3 ipsets or fw4 nftables sets, depending how dnsmasq is compiled.

Nftset isn’t really a term except within dnsmasq’s configuration.

rpigott:
fwiw systemd uses the term NFTSet as well: https://github.com/search?q=repo%3Asystemd%2Fsystemd%20NFTSet&type=code

jow-:
This will break compatibility with firewall3, which still can be installed.

jow-:
I think fw3/fw4 actually is unrelated here. The code you're about to touch relates to dhcp (dnsmasq) uci configuration and upstream dnsmasq.init currently only supports `config ipset` sections (which are transparently translated to --nftset settings if nftset is compiled in). I think this change (changing the `ipset` section type to `nftset`) will render the functionality broken.

systemcrash:
> what about "Address set"?

Could work. I suspect it would take a bit for those users who need it to figure this out. 
"Remotely plausible menu here called 'sets' exists, but ipsets/nftsets is gone... hmm, click here."

How about handling ipsets and nftsets with two tables both under the same tab (I generally dislike conflating similar concepts), until dnsmasq.init has caught up?

jow-:
Why not just leave it as-is? Maybe call the labels `nftset/ipset`. Right now dnsmasq init accepts "ipset", translates it to both `--ipset` and `--nftset` simultaneously and then throws away what is not supported. So in essence it is agnostic and should be imho.

Having a configuration abstraction in the form of uci makes little sense if every little implementation detail and native config idiosyncrasy is exposed as-is to uci. The uci configuration should be intent-driven, means "please build a set of resolved IPs for domain X" and not concern itself with details such as "is ipset compiled in or nftset? is it firewall3 or firewall4? do domain names need to be separated with slash, or comma, or space?"

Unfortunately the existing dnsmasq (dhcp) uci support is quite bad in this regard, organically grown over years and exposing weird dnsmasq design choices such as the slash separated address format as-is instead of doing proper structured lists. But that does not mean that we should needlessly continue this approach. When some hypothetical kind of eBPF based filtering comes along in a few years we'll be dealing with ipsets, nftsets and bpfsets.

The term "ipset" predates "nftsets" but is imho still a valid generic term for "set of IPs". Apart from these points, a `config ipset` is backwards compatible to old systems and upwards compatible to new, nft based ones while a `config nftset` will only work on new ones. We'd also need to duplicate the ui code and support both ipsets and nftsets with mostly overlapping functionality, leading to even further bloat.

systemcrash:
> Having a configuration abstraction in the form of uci makes little sense if every little implementation detail and native config idiosyncrasy is exposed as-is to uci. The uci configuration should be intent-driven, means "please build a set of resolved IPs for domain X" and not concern itself with details such as "is ipset compiled in or nftset? is it firewall3 or firewall4? do domain names need to be separated with slash, or comma, or space?"
> 
You've said something very important here. 

> Unfortunately the existing dnsmasq (dhcp) uci support is quite bad in this regard, organically grown over years and exposing weird dnsmasq design choices such as the slash separated address format as-is instead of doing proper structured lists. But that does not mean that we should needlessly continue this approach. When some hypothetical kind of eBPF based filtering comes along in a few years we'll be dealing with ipsets, nftsets and bpfsets.

Ideally bpfsets (should such a thing come into existence) get their own proper handling. 

I'll rework the text on the page and just add the nftsets bits, marked appropriately as so.

systemcrash:
OK, that's roughly done.

Issue #6716, opened 2023-11-23
luci-theme-bootstrap: 32px favicon


532910:
Please add bigger favicon for HiDPI screens. I believe 32px will be enough. Right now only 16x16 px is exists and it looks very blurry:
https://github.com/openwrt/luci/blob/master/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/favicon.png

532910:
https://github.com/openwrt/luci/pull/6717

systemcrash:
Pixels? Blech! Vector graphixx 😄

532910:
no safari support: https://caniuse.com/link-icon-svg

PR #6738, opened 2023-12-04
SVG favicon

stokito:
Replacement for #6717 

* Fix OpenWrt 2020 logo colors and background
* Use the SVG logo as a favicon
* Migrate all themes to the same logo


stokito:
Needs for testing and feedback but generally it's should be ready to be merged. It will close #6716 and #6717

systemcrash:
One thing #6717 does correctly is retain the png for those browsers that do not support them (safari...). 

https://caniuse.com/link-icon-svg

systemcrash:
Looks mostly good. You retained the 180x180 png ya?

stokito:
Yes, sorry, forgot to mention this: the original logo.png from the 2020 theme remains untouched. I was afraid to make a regression there.
For the theme the logo is also shown on the front page as a home logo.
Maybe there the 48px would be not enough.
But the material and the bootstrap will use the 48px version.

stokito:
> For the theme the logo is also shown on the front page as a home logo.

Sorry, I was wrong, there is used the logo.svg, not png.
It should be safe to downscale that png too. It must be approved by the theme author.

hnyman:
I tested with Firefox and Chrome, both bootstrap and openwrt2020 themes.
Looks ok to me.

Any objections, @jow-  ?

stokito:
Thank you!

jow-:
Looks quite ugly in dark mode:
grafik

I think the white curved bars at the top should be light blue, the circle should remain dark blue and the background should become transparent. This way it works well in both dark and light backgrounds.

stokito:
According to [logo usage guidelines](https://openwrt.org/_media/docs/guide-graphic-designer/openwrt-logo-usage-guidelines.pdf) there are a few schemes:

1. White & Dark Blue that is used here
2. White & Blue that should look better on a dark background
3. White that also may be used in a dark mode.

The problem that I see is that here we have a dark theme of OS but the site itself may be in a light mode.

stokito:
FYI: the SVG can be adopted to a dark mode by using internal styles https://rigtorp.se/notes/dark-mode-svg/

jow-:
Here's a quick mockup of the changed color scheme:
grafik
grafik
grafik
grafik

PR #6740, opened 2023-12-04
Prevent odhcp6c from asking for Softwire46 OROs by default, as `map` package isn't installed by default

### LINKED ISSUE #6739: luci-proto-ipv6: Update dhcpv6.js to allow ORO selection and not request Softwire46 opts by default
## Steps to reproduce:
1. go to: Network → Interfaces
2. Create new WAN interface with DHCPv6 protocol
3. Save & Apply

## Actual behavior:
By default the DHCPv6 client, odhcp6c, is spawned without the `-R` option that prevents it from asking for a lot of Option Request Options, including the Softwire46 (RFC 7598) options, which require additional packages that are not installed by default.

ISPs are now starting to support these Softwire46 protocols such as MAP-T and MAP-E. Clients that ask for these IPv4aaS protocols but don't actually support them, will at best be wasting IPv4 addresses, and at worst, will have broken IPv4 connectivity if the ISP refuses to then give them a native DHCPv4 lease because of it.

## Expected behavior:
odhcp6c should not ask include Options 94, 95, or 96, in the Option Request Option, without the `map` package being installed.

## Additional Information:

See also: https://github.com/openwrt/openwrt/issues/14056

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='23.05.2'
DISTRIB_REVISION='r23630-842932a63d'
DISTRIB_TARGET='x86/64'
DISTRIB_ARCH='x86_64'
DISTRIB_DESCRIPTION='OpenWrt 23.05.2 r23630-842932a63d'
DISTRIB_TAINTS=''
```

detobate:
Add defaultreqopts toggle that's off by default
Add MultiValue list to select Option Request options, leaving non-softwire46 OROs enabled by default.

Closes #6739

Note: requires a [PR14136](https://github.com/openwrt/openwrt/pull/14136) in to OpenWRT to be accepted

### DIFF
--- a/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/dhcpv6.js
+++ b/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/dhcpv6.js
@@ -30,7 +30,31 @@ return network.registerProtocol('dhcpv6', {
 		o.value('64');
 		o.default = 'auto';
 
+		o = s.taboption('general', form.Flag, 'defaultreqopts', _('Request all known DHCPv6 options'));
+		o.rmempty = false;
+		o.default = '0';
+
+		o = s.taboption('general', form.MultiValue, 'reqopts', _('Request Specific DHCPv6 Options'));
+		o.depends('defaultreqopts', '0');
+		o.widget = 'checkbox';
+		o.multiple = true;
+		o.value('21', 'SIP Server Domain Name List (21)');
+		o.value('22', 'SIP Servers IPv6 Address List (22)');
+		o.value('23', 'DNS Recursive Name Server (23)');
+		o.value('24', 'Domain Search List (24)');
+		o.value('31', 'Simple NTP Server (31)');
+		o.value('56', 'NTP Server (56)');
+		o.value('64', 'DS-Lite AFTR Name (64)');
+		o.value('94', 'MAP-E (94)');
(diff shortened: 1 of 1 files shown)


detobate:
I left the other non-S46 OROs as requested-by-default in the list, so as to not disturb the status quo too much.

Downside is that now a user who install the `map` package will also have to explicitly tell the DHCPv6 client to ask for MAP-T/E options.     Maybe that's something the `map` package could auto-configure upon install?

systemcrash:
Tested on 23.05. Looks good. 

Could you add the other options found here?: https://openwrt.org/docs/guide-user/network/ipv6/dhcp6c

jow-:
I am against this change. It introduces regressions for existing MAP users who will be unaware of the new requirement. The hardcoding of available option number choices is not ideal either. 

Imho, the most robust way forward is simply patching odhcpd to simply omit MAP related request options if no `/sys/modules/nat46` directory is present. The odhcpd daemon is OpenWrt specific software after all, so we can freely change it to accommodate our needs.

detobate:
I am all for fixing it in `odhcp6c` and would prefer that over this proposal, I just didn't think that that suggestion would gain any traction (and is probably beyond my skillset to submit a PR)

detobate:
Closing as the main issue is fixed by new PR [odhcp6c#81](https://github.com/openwrt/odhcp6c/pull/81)

PR #6742, opened 2023-12-05
luci-app-adblock-fast: update to 1.1.0-1

stangri:
Depends on https://github.com/openwrt/packages/pull/22820


systemcrash:
@stangri I recently discovered [these](https://lists.sr.ht/~aparcar/openwrt-devel-test/%3C20210119211217.3983617-1-mail%40aparcar.org%3E). Have you had an opportunity to utilize them anywhere? This could just as easily be a question to anyone, but here you are :)

stangri:
@systemcrash if you mean AUTORELEASE it's been phased out recently. Not sure about COMMITCOUNT.

PR #6746, opened 2023-12-05
luci-app-vnstat2: fix configuration page when database is empty

janh:
The output of the command "vnstat --dbiflist 1" is empty when there are no interfaces in the database. Add a check to avoid a "result.stdout is undefined" error in that case.

Ref: https://forum.openwrt.org/t/error-in-luci-vnstat2-configuration-menu/179700
Fixes: 3ac4f567dfdf ("luci-app-vnstat2: use more efficient way to get database interfaces")

---

This fix is also needed in the openwrt-23.05 and openwrt-22.03 branches.


systemcrash:
that's in master, 23 and 22.

PR #6747, opened 2023-12-06
luci-proto-ipv6: dhcpv6.js fix translation

try496:
This is an option that hasn't been translated in many years


systemcrash:
You also need to put translatable strings in the source JS file.

PR #6755, opened 2023-12-08
luci-base: Display 'Access Point' in the wireless page instead of 'Master'

EvanTheB:
The config dialogue uses "Access Point". 

Fix up a missed translation and some other minor issues. 

Screenshot 2023-12-08 134325

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/network.js
+++ b/modules/luci-base/htdocs/luci-static/resources/network.js
@@ -3837,7 +3837,7 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */ 
 	 *  - `Unknown`
 	 */
 	getActiveMode: function() {
-		var mode = this.ubus('net', 'iwinfo', 'mode') || this.ubus('net', 'config', 'mode') || this.get('mode') || 'ap';
+		var mode = this.ubus('net', 'iwinfo', 'mode') || this.getMode();
 
 		switch (mode) {
 		case 'ap':      return 'Master';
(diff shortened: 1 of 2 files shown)


EvanTheB:
Thanks jow, I have changed to modify only the I18n variant. 

I could not work out what the getActiveMode function is used for, all the usages seem to be user facing strings, so I'm not sure why the I18n variant is not used everywhere. Would you accept a patch changing all usage to the I18n variant?

Thanks

modules/luci-base/htdocs/luci-static/resources/network.js
```diff
@@ -3861,12 +3861,17 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */
 		var mode = this.getActiveMode();
 
 		switch (mode) {
-		case 'Master':  return _('Master');
-		case 'Client':  return _('Client');
-		case 'Ad-Hoc':  return _('Ad-Hoc');
-		case 'Mash':    return _('Mesh');
-		case 'Monitor': return _('Monitor');
-		default:        return mode;
+		case "Master":       return _('Access Point');
```
systemcrash, line 3864: nitpick, but do the `'` really need to change to `"` in the JS?

systemcrash (changes requested):

stokito (approved):

modules/luci-base/htdocs/luci-static/resources/network.js
EvanTheB: yep, slip of the tongue. fixed

systemcrash (approved):

PR #6756, opened 2023-12-09
luci-app-adblock-fast: bugfix: L.uci.load

stangri:
* fix app not working on systems without smartdns
* fix smartdns instances selection dependencies

Thanks @Ansuel !


Ansuel:
I would split the 2 commits? the title might be confusing and unrelated to the dependency fix

PR #6764, opened 2023-12-11
luci-proto-ipv6: Add DHCPv6 norelease menu option as default on

detobate:
As recommended by [RFC 9096: WPD-9](https://datatracker.ietf.org/doc/html/rfc9096#name-automatic-dhcpv6-releases), to reduce the risk of prefix changes

### DIFF
--- a/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
+++ b/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
@@ -519,6 +519,26 @@ return view.extend({
 		o.optional = true;
 		o.rmempty = true;
 
+		o = s.taboption('adv_log', form.Flag, 'ban_remotelog', _('Enable Remote Logging'), _('Enable the cgi interface to receive remote logging events.'));
+		o.default = 0
+		o.optional = true;
+		o.rmempty = true;
+
+		o = s.taboption('adv_log', form.Value, 'ban_remotetoken', _('Remote Token'), _('Token to communicate with the cgi interface.'));
+		o.depends('ban_remotelog', '1');
+		o.datatype = 'and(minlength(3),maxlength(20))';
+		o.validate = function (section_id, value) {
+			if (!value) {
+				return _('Empty field not allowed');
+			}
+			if (!value.match(/^[A-Za-z0-9\.\:]+$/)) {
+				return _('Invalid characters');
+			}
+			return true;
+		}
+		o.optional = true;
+		o.rmempty = true;
+
 		/*
 			advanced email settings tab
 		*/
--- a/applications/luci-app-banip/po/ar/banip.po
+++ b/applications/luci-app-banip/po/ar/banip.po
@@ -95,7 +95,7 @@ msgstr ""
 msgid "512"
 msgstr ""
 
-#: applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js:581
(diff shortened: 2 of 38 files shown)


systemcrash:
If I take your other luci changes for the options and have it emit a string (so as not to break backwards compat), would that still work with the underlying odhcp6c with your PR there?

detobate:
Yeah that should be fine.    

I'd change the default value of the `defaultreqopts` flag to 1, to avoid changing the current default behaviour, as it's no longer required to close the original issue.  

And maybe move both options in the advanced tab?

systemcrash:
Please fix these (might require changing github settings too):
```
: Author name (Det) need to be your real name 'firstname lastname'
: Commit subject line MUST start with '<package name>: ' (Add DHCPv6 norelease menu option as default on, as recommended by RFC 9096: WPD-9)
: Signed-off-by is missing or doesn't match author (should be 'Signed-off-by: Det <detonate@helix.net.nz>')
```

detobate:
Ahh, giving it a try, but am probably messing it up.  Might be easier if I recreate it with fresh commits?

protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/dhcpv6.js
```diff
@@ -30,6 +30,10 @@ return network.registerProtocol('dhcpv6', {
 		o.value('64');
 		o.default = 'auto';
 
+		o = s.taboption('general', form.Flag, 'norelease', _('Do not send a Release when restarting'),
+						_('Enable to reduce the risk of prefix changes following a restart'));
+		o.default = '1';
```
jow-, line 35: The uci option defaults to off, not on. Please remove the `o.default = 1` line.

systemcrash:
> Ahh, giving it a try, but am probably messing it up. Might be easier if I recreate it with fresh commits?

However you do it. It shouldn't take more than a single commit.

protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/dhcpv6.js
detobate: Changing the default was the intent of this PR (as well as expose the config var via Luci), given the new advice from RFC 9096.    Not sure where to change the uci option default, but if you point me in the right direction I could create another PR there.  Presumably starting a _"should we change this default?"_ discussion officially.

jow-: The option default would need to be changed here: https://github.com/openwrt/openwrt/blob/main/package/network/ipv6/odhcp6c/files/dhcpv6.sh

detobate: Is that by changing [Line18](https://github.com/openwrt/openwrt/blob/main/package/network/ipv6/odhcp6c/files/dhcpv6.sh#L18) from

`proto_config_add_string 'norelease:bool'`
to
`proto_config_add_string 'norelease:bool:1'`

correct?

detobate:
Ahh whoops, messing around with commit history auto-closed this PR.  Should I open another one now, or wait for the other PR to change the UCI default first?

systemcrash:
Just push the (one) commit without the `default` line.

detobate:
> Just push the (one) commit without the `default` line.

Done #6775

PR #6775, opened 2023-12-20
luci-proto-ipv6: Add DHCPv6 norelease menu option

detobate:
Resubmitting #6764 without the change of default behaviour.

### DIFF
--- a/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/dhcpv6.js
+++ b/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/dhcpv6.js
@@ -29,6 +29,8 @@ return network.registerProtocol('dhcpv6', {
 		o.value('60');
 		o.value('64');
 		o.default = 'auto';
+		o = s.taboption('general', form.Flag, 'norelease', _('Do not send a Release when restarting'),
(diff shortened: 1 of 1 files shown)


systemcrash:
Merged, thanks :)

try496:
Why is there no translation file?

hnyman:
The possible translation strings get synced/populated in batch in random intervals

hnyman:
>  build based on the latest 23.05 encountered a large number of dhcp content translation failures. I don’t know the reason.

Well, all translations are made by volunteers like you and me.
New strings are not automatically translated by some paid organisation, but require manual work. So, it may take a while after changes until you (or somebody else) does the translation and the new strings translations get merged (into main/master).  And then later they get backported into stable 23.05.

Feel free to contribute at https://hosted.weblate.org/projects/openwrt/

try496:
Screenshot_2023-12-22-20-57-04-82

try496:
IMG_20231222_210732
But this website cannot solve this problem

systemcrash:
Correct. I added some new strings recently because the older ones were
ambiguous or did not explain clearly, and as such will require translating,
because the old ones were subtly wrong.

PR #6777, opened 2023-12-21
add luci-theme-kucat 2.3.9

sirpdboy:
add luci-theme-kucat 2.3.9


hnyman:
You propose to introduce a theme, but offer no reasoning and no screenshots in PR?

NAK from me.

In general, we have bad experiences from enthusiast themes that tend to get unmaintained pretty soon after the initial commit.
The theme also has a distorted OpenWrt logo.
Why would the theme need curl?

systemcrash:
Do you commit to the upkeep and bug fixes of the theme for the future?

Can you also add a SVG favorites icon?

PR #5813, opened 2022-05-15
luci-app-openvpn : Updates "user", "group", "proto" and "mssfix" params conds and defaults

pduveau:
From Philippe Duveau

"proto" parameter is available in client or server mode in the network section.
"mssfix" default value is setted to openvpn.org default value
"user" and "group" are available in service section with secured default values (nobody, nogroup)

Signed-off-by: Philippe Duveau <pduvax@gmail.com>

---

The "proto" parameter is mandatory for other parameters of server or client modes (topology, mssfix...) but the condition was to restrict it to client mode only.

PR #6780, opened 2023-12-22
luci-app-openvpn: Updates user, group, proto and mssfix params and defaults

pduveau:
- proto parameter is available in client or server mode in the network section.
- mssfix default value is setted to openvpn.org default value
- user/group are available in service section with secured default values (nobody, nogroup)

The "proto" parameter is mandatory for other parameters of server or client modes (topology, mssfix...) but the condition was to restrict it to client mode only.


pduveau:
I'm sorry to create a new a new pull request instead of reopening the previous one #5813. I did a forced push before reopen.

systemcrash:
OK. What is the effect of reintroducing the user/group settings? While hidden, they will internally assume some kind of defaults, I think - are they changed by your introduction?

I see `	option user nobody` persisted in `openvpn.config` but not `group`. I also do not see that either of these options are handled in the init scripts. I don't use openvpn so I have to ask.

pduveau:
@systemcrash You are right. I did not see the internally defined option user nobody. I guess the group nobody is inherited from the user. So I will remove this part of the proposed change.

systemcrash:
> @systemcrash You are right. I did not see the internally defined option user nobody (a year ago). I guess the group nobody is inherited from the user. So I will remove this part of the proposed change.

OK - please update your commit message with your real email

systemcrash:
Closed via 9889d3f73524c5ef814ab37e5c9fac81f213046d

Thanks!

PR #6785, opened 2023-12-24
luci-app-usteer: Added tooltip in status page

Ramon00:
Added tooltip while hovering over connected client showing mac, IP(s), hostname in status page

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -91,25 +91,46 @@ function collectWlanAPInfoEntries(connectioninfo_table_entries, wlanAPInfos) {
 	}
 };
 
+function tootltip(mac, IP, hostname) {
+	var body= E([]);
+	body.appendChild(E('div', mac));
+	if (typeof IP !== 'undefined') {
+		for (var IPaddr in IP['ipaddrs']) body.appendChild(E('div', IP['ipaddrs'][IPaddr]));
+		for (var IPaddr in IP['ip6addrs']) body.appendChild(E('div', IP['ip6addrs'][IPaddr]));;
+	}
+	if (hostname !== '') {
+		body.appendChild(E('div', '%h'.format(hostname)));
+	}
+	return body;
+}
 
 function collectWlanAPInfos(compactconnectioninfo_table_entries, wlanAPInfos) {
 	for (var wlan in wlanAPInfos) {
-		var hostl = '';
+		var hostl = E([]);
 		for (var mac in Clients) {
 			if (typeof Clients[mac] !== 'undefined')
 				if (typeof Clients[mac][wlan] !== 'undefined')
-					if (String(Clients[mac][wlan]['connected']).valueOf() == String('true').valueOf()) {
+					if (String(Clients[mac][wlan]['connected']).valueOf() === 'true') {
 						var foundname = mac;
+						var IP = '';
+						var hostname = '';
 						var macUp = mac.toUpperCase();
 						if (typeof Hosts[macUp] !== 'undefined') {
-							if ((String(Hosts[macUp]['ipaddrs'][0]).length > 0) && (typeof Hosts[macUp]['ipaddrs'][0] !== 'undefined')) {
-								foundname = Hosts[macUp]['ipaddrs'][0];
+							if ((typeof Hosts[macUp]['ipaddrs'][0] !== 'undefined') && (String(Hosts[macUp]['ipaddrs'][0]).length > 0)) {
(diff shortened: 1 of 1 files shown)


applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -91,25 +91,55 @@ function collectWlanAPInfoEntries(connectioninfo_table_entries, wlanAPInfos) {
 	}
 };
 
+function tootltip(mac, IP, hostname) {
+	var body= E([]);
+	var i=0;
+	body.appendChild(E('div', mac));
+	if (typeof IP !== 'undefined') {
+		while ((typeof IP['ipaddrs'][i] !== 'undefined') && (String(IP['ipaddrs'][i]).length > 0)) {
```
systemcrash, line 99: Please revise these `while` loops to `forEach` loops. This makes them less ambiguous and avoids variable `i` declaration and assignment errors. You also avoid indices.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -91,25 +91,55 @@ function collectWlanAPInfoEntries(connectioninfo_table_entries, wlanAPInfos) {
...
+			body.appendChild(E('div', IP['ipaddrs'][i]));
+			i++;
+		}
+		
+		i=0;
+		while ((typeof IP['ip6addrs'][i] !== 'undefined') && (String(IP['ip6addrs'][i]).length > 0)) {
```
systemcrash, line 105: using `&& String...` might behave unexpectedly: what is the truth test here? That you coerced something to a string? Or that length is > 0? Perhaps skip String()

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -91,25 +91,55 @@ function collectWlanAPInfoEntries(connectioninfo_table_entries, wlanAPInfos) {
...
 
 function collectWlanAPInfos(compactconnectioninfo_table_entries, wlanAPInfos) {
 	for (var wlan in wlanAPInfos) {
-		var hostl = '';
+		var hostl = E([]);
 		for (var mac in Clients) {
 			if (typeof Clients[mac] !== 'undefined')
 				if (typeof Clients[mac][wlan] !== 'undefined')
-					if (String(Clients[mac][wlan]['connected']).valueOf() == String('true').valueOf()) {
+					if (String(Clients[mac][wlan]['connected']).valueOf() === String('true').valueOf()) {
```
systemcrash, line 122: Must one coerce everything to strings here? Doesn't the following work identically?:

`Clients[mac][wlan]['connected'] === 'true'`

systemcrash (changes requested):

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
Ramon00: I agree it is a bit messy, I made it into `for (var IPaddr in IP['ipaddrs']) body.appendChild(E('div', IP['ipaddrs'][IPaddr]));` now. So much cleaner. Thx

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
Ramon00: I tested for length >0 as I encountered a empty string as only entry in the past. With the change above this is now not needed anymore.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
Ramon00: Somehow that does not work. But `if (String(Clients[mac][wlan]['connected']).valueOf() === 'true') ` that does work. So I will leave it like this.

systemcrash:
Do you think it's safe to do this with the mac and not null etc check it like IPs and hostname ? 
`	body.appendChild(E('div', mac));`

Ramon00:
> Do you think it's safe to do this with the mac and not null etc check it like IPs and hostname ?

@systemcrash yes that is safe because the mac is _the_ client identifier used. So there is always a mac, there may be one or more IP addresses and there may be a host name. On the main router mosty everything is available, on the AP it may only be a mac address.

Ramon00:
Oh that reminds me to 'escape' the hostname. Let me add that

systemcrash:
OK. Shall we leave this to percolate for a diem and merge?

systemcrash (approved):

Ramon00:
@systemcrash sounds like a plan!

PR #6788, opened 2023-12-26
luci-app-alist: add new package

1715173329:
Add LuCI interface for the AList package.

Ref: https://github.com/openwrt/packages/pull/22902

### DIFF
--- a/applications/luci-app-alist/Makefile
+++ b/applications/luci-app-alist/Makefile
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 ImmortalWrt.org
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI app for AList
+LUCI_DEPENDS:=+alist
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
+++ b/applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
@@ -1,0 +1,144 @@
+// SPDX-License-Identifier: Apache-2.0
+
+'use strict';
+'require form';
+'require poll';
+'require rpc';
+'require uci';
+'require validation';
+'require view';
+
+var callServiceList = rpc.declare({
+	object: 'service',
+	method: 'list',
+	params: ['name'],
+	expect: { '': {} }
+});
+
+function getServiceStatus() {
+	return L.resolveDefault(callServiceList('alist'), {}).then(function (res) {
+		var isRunning = false;
+		try {
+			isRunning = res['alist']['instances']['instance1']['running'];
+		} catch (e) { }
+		return isRunning;
+	});
+}
+
+function renderStatus(isRunning, port) {
+	var spanTemp = '<span style="color:%s"><strong>%s %s</strong></span>';
+	var renderHTML;
+	if (isRunning) {
+		var button = String.format('&#160;<a class="btn cbi-button" href="http://%s:%s" target="_blank" rel="noreferrer noopener">%s</a>',
+			window.location.hostname, port, _('Open Web Interface'));
+		renderHTML = spanTemp.format('green', _('AList'), _('RUNNING')) + button;
+	} else {
+		renderHTML = spanTemp.format('red', _('AList'), _('NOT RUNNING'));
+	}
+
+	return renderHTML;
+}
+
+var stubValidator = {
+	factory: validation,
+	apply: function(type, value, args) {
+		if (value != null)
+			this.value = value;
+
+		return validation.types[type].apply(this, args);
+	},
+	assert: function(condition) {
+		return !!condition;
+	}
+};
+
+return view.extend({
+	load: function() {
+		return Promise.all([
+			uci.load('alist')
+		]);
+	},
+
+	render: function(data) {
+		var m, s, o;
+		var webport = uci.get(data[0], 'config', 'listen_http_port') || '5244';
+
+		m = new form.Map('alist', _('AList'),
+			_('A file list/WebDAV program that supports multiple storages, powered by Gin and Solidjs.') + '<br />' +
+			_('Default webUI/WebDAV login username is %s and password is %s.').format('<code>admin</code>', '<code>password</code>'));
+
+		s = m.section(form.TypedSection);
+		s.anonymous = true;
+		s.render = function () {
+			poll.add(function () {
+				return L.resolveDefault(getServiceStatus()).then(function (res) {
+					var view = document.getElementById('service_status');
+					view.innerHTML = renderStatus(res, webport);
+				});
+			});
+
+			return E('div', { class: 'cbi-section', id: 'status_bar' }, [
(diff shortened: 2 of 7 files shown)


applications/luci-app-alist/Makefile
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 ImmortalWrt.org
```
systemcrash, line 3: This might be problematic. I need a second opinion here. @hnyman @jow- @nbd168 @hauke

1715173329: I'm not sure what do you mean by "problematic".
Is it a license issue? This repo is also licensed under Apache-2.0.

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
```diff
@@ -0,0 +1,151 @@
...
+
+		o = s.option(form.Value, 'site_max_connections', _('Max connections'),
+			_('The maximum number of concurrent connections at the same time (0 = unlimited).'));
+		o.datatype = 'uinteger';
+		o.default = '0';
+		o.rmempty = false;
+
+		o = s.option(form.Flag, 'site_tls_insecure', _('Allow insecure connection'),
+			_('Allow connection even if the remote TLS certificate is invalid (<strong>not recommended</strong>).'));
+		o.default = o.disabled;
```
stokito, line 126: `o.default = o.disabled;`

looks like the line should be removed

1715173329: probably.. but looks like it's still common to declear this explicitly

systemcrash: The init script you submitted does not have a default. So it would be best to remove the default here.

stokito: I mean you really take a value of the `o.disabled` which is always probably false or undefined.
The `default` for flag values is false

1715173329: > The init script you submitted does not have a default. So it would be best to remove the default here.

In my function `uci_json_add_boolean` it defaults to `false` ;)

> The `default` for flag values is false

Yes I see it so I said "it's still common to declear this explicitly". Anyway removed ;)

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
```diff
@@ -0,0 +1,150 @@
...
+			uci.load('alist')
+		]);
+	},
+
+	render: function(data) {
+		var m, s, o;
+		var webport = uci.get(data[0], 'config', 'listen_http_port') || '5244';
+
+		m = new form.Map('alist', _('AList'),
+			_('A file list/WebDAV program that supports multiple storages, powered by Gin and Solidjs.<br />' +
```
systemcrash, line 67: please put any and all  html breaks outside of translatable string blocks and just concatenate them as simple strings.  This and the next line can be independent translatable strings.

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
```diff
@@ -0,0 +1,150 @@
...
+				'Default login username is <code>admin</code> and password is <code>password</code>.'));
```
systemcrash, line 68: Put the blocks in code as %s variables via a format() function.

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
```diff
@@ -0,0 +1,150 @@
...
+		}
+
+		s = m.section(form.NamedSection, 'config', 'alist');
+
+		o = s.option(form.Flag, 'enabled', _('Enable'));
+		o.default = o.disabled;
+		o.rmempty = false;
+
+		o = s.option(form.Value, 'listen_addr', _('Listen address'));
+		o.default = '0.0.0.0';
```
systemcrash, line 92: Change from default to placeholder, especially if the underlying init script handles defaults. Do the same for other defaults supplied below.

systemcrash (changes requested):

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
1715173329: done

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
1715173329: done

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
1715173329: done

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
```diff
@@ -0,0 +1,144 @@
...
+		]);
+	},
+
+	render: function(data) {
+		var m, s, o;
+		var webport = uci.get(data[0], 'config', 'listen_http_port') || '5244';
+
+		m = new form.Map('alist', _('AList'),
+			_('A file list/WebDAV program that supports multiple storages, powered by Gin and Solidjs.') + '<br />' +
+				_('Default login username is <code>%s</code> and password is <code>%s</code>.').format('admin', 'password'));
```
systemcrash, line 68: Change to:
```js
				_('Default login username is %s and password is %s.').format('<code>admin</code>', '<code>password</code>'));
```

Making it easier for translators.

systemcrash (changes requested):

systemcrash:
To what or where does the user login? A webdav service aggregator?

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
1715173329: Okay will do it tomorrow.

systemcrash: OK, that's the last thing I can see. Then we can merge.

systemcrash:
> > To what or where does the user login? A webdav service aggregator?
> 
> AList has a webUI to manage users, storage providers etc., WebDAV also uses the same username/password.

I think it beneficial to provide more context to users that this is where they log in to: it's not immediately clear what the login is for, and being transparent here is helpful.

applications/luci-app-alist/htdocs/luci-static/resources/view/alist/config.js
1715173329: done

1715173329:
> I think it beneficial to provide more context to users that this is where they log in to: it's not immediately clear what the login is for, and being transparent here is helpful.

added

systemcrash (approved):

1715173329:
Thank you!

systemcrash:
Merged. Thanks for this interesting contribution. 

You're welcome to open another PR if there's anything that needs changing later.

PR #6791, opened 2023-12-27
luci-app-sshtunnel: servers: ProxyCommand option

stokito:
A leftover from #6772
Since the https://github.com/openwrt/packages/pull/22975 was merged we can now add the ProxyCommand to the Luci app.

cc @systemcrash

### DIFF
--- a/applications/luci-app-sshtunnel/Makefile
+++ b/applications/luci-app-sshtunnel/Makefile
@@ -7,7 +7,7 @@ LUCI_TITLE:=LuCI support for SSH Tunnels (sshtunnel package)
 
 PKG_MAINTAINER:=Sergey Ponomarev <stokito@gmail.com>
 LUCI_DEPENDS:=+luci-base +sshtunnel
-PKG_VERSION:=1.0.0
+PKG_VERSION:=1.1.0
 PKG_RELEASE:=1
 
 include ../../luci.mk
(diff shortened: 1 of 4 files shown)


stokito:
thank you

hnyman:
There is now a weblate conflict.
Hopefully @systemcrash can fix it, as I am travelling without proper access...

stokito:
I checked and my translation is better than from the WebLate. Also one key was changed (typo "serve" to "server"). So the WebLate changes needs to be discarded

systemcrash:
> There is now a weblate conflict. Hopefully @systemcrash can fix it, as I am travelling without proper access...

Should be able to. Can you see where? I don't immediately see a conflict.

stokito:
It looks like it's too complicated to fix the conflict in the WebLate
https://hosted.weblate.org/projects/openwrt/luciapplicationsdockerman/#alerts

Maybe I'll remove my changes? And later will translate from the WebLate.

systemcrash:
I've reverted this commit via 23df9c2fbf2eafdb208b1461ba0d823d989819d7

systemcrash:
OK, looks like we are back in business.

stokito:
ok, and should I make a new PR without the po files?

systemcrash:
That's certainly one way to do it.

PR #6795, opened 2023-12-29
luci-app-irqbalance: add new package

ElaineR02:
Adding LuCI interface for irqbalance.
It displays the IRQ snapshot first, then the config options. It also checks whether the input is vaild in the last two options.

### DIFF
--- a/applications/luci-app-irqbalance/htdocs/luci-static/resources/view/irqbalance.js
+++ b/applications/luci-app-irqbalance/htdocs/luci-static/resources/view/irqbalance.js
@@ -1,0 +1,116 @@
+'use strict';
+'require view';
+'require fs';
+'require form';
+'require ui';
+'require rpc';
+
+const callServiceList = rpc.declare({
+    object: 'service',
+    method: 'list',
+    params: ['name'],
+    expect: { '': {} },
+});
+
+function getServiceStatus() {
+    return L.resolveDefault(callServiceList('irqbalance'), {}).then(function (res) {
+        try {
+            return res['irqbalance']['instances']['irqbalance']['running'];
+        } catch (e) {
+            return false;
+        }
+    });
+}
+
+function renderStatus(isRunning) {
+    const spanTemp = '<span style="color:%s"><strong>%s</strong></span>';
+
+    return isRunning
+        ? String.format(spanTemp, 'green', _('RUNNING'))
+        : String.format(spanTemp, 'red', _('NOT RUNNING'));
+}
+
+return view.extend({
+    load() {
+        return L.resolveDefault(fs.read_direct('/proc/interrupts'));
+    },
+
+    render(data) {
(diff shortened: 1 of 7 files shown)


applications/luci-app-irqbalance/htdocs/luci-static/resources/view/irqbalance.js
```diff
@@ -0,0 +1,116 @@
...
+        o.placeholder = '2';
+        o.datatype = 'uinteger';
+        o.optional = true;
+
+        o = s.option(form.Value, 'interval', _('Interval'), _('Value in seconds.'));
+        o.placeholder = '10';
+        o.datatype = 'uinteger';
+        o.optional = true;
+
+        o = s.option(form.Value, 'banned_cpulist', _('Excluded CPU list'), _('Specify excluded CPU list, can be an integer or integers separated by commas.') + '<br />' + _('Allowed value(s): %s.').format(cpuNum.join(', ')));
```
systemcrash, line 92: What is an 'excluded CPU'? Please clarify for users what that is.

applications/luci-app-irqbalance/htdocs/luci-static/resources/view/irqbalance.js
```diff
@@ -0,0 +1,116 @@
...
+        o.validate = function (section_id, value) {
+            for (const i of value.split(',')) {
+                if (!cpuNum.includes(i) && i != '') {
+                    return _('Invalid');
+                }
+            }
+            return true;
+        }
+
+        o = s.option(form.DynamicList, 'banirq', _('Banned IRQs'), _('Specify banned IRQs.') + '<br />' + _('Allowed values: %s.').format(irqNum.join(', ')));
```
systemcrash, line 104: Here is a setting called "banned" (past tense), and yet in the next row, is "allowed" (past tense). Which is it? Avoid dissonance. 

Does "allowed" here mean "values to choose from"? And "banned" means "to block"?

I note the setting is 'banirq', which is not past tense, but *present* simple tense.

systemcrash (changes requested):

applications/luci-app-irqbalance/htdocs/luci-static/resources/view/irqbalance.js
```diff
@@ -0,0 +1,116 @@
...
+                    view.innerHTML = renderStatus(res);
+                });
+            });
+
+            return E('div', { class: 'cbi-section' }, [
+                E('p', { id: 'status' }, _('Loading...'))
+            ]);
+        }
+
+        s = m.section(form.TypedSection, 'irqbalance', _('IRQs snapshot'));
```
systemcrash, line 60: What does "IRQs snapshot" mean? Plural here might hinder genitive. Is this something like "Snapshot of current IRQs"? Please make this simpler to understand at a glance to first time users understand this quickly and translators know what to do with it.

systemcrash (changes requested):

ElaineR02:
Wording issues have been fixed. Thanks.

systemcrash (approved):

Issue #6139, opened 2022-12-07
RFC: creating an enhanced version of LuCI statistics graphs?


hnyman:
**Creating an alternative version of LuCI statistics with enhanced graphics?**

We are currently using the year 2003 version of rrdtool 1.0.50 in LuCI statistics. 
The benefit is that the RRD library 1.0.50 is really small, compact and has no external dependencies. It is really efficient on flash space.

The drawback is that we miss all the developments for RRDtool in the last 19 years and the graphs are rather basic and show their age.
In practice there is also no support for the current RRDtool for other needs than LuCI statistics.

The graphs from the current RRDtool 1.8.0 are more polished and crisp. 
The 1.8.0 is also aware of the time_t size and can adjust to 64bit time in 32bit systems.

One year ago I briefly tested with new RRDtool, when looking at the musl 64bit time_t issues. See 
* rejected proposal to accept 64bit time_t and modify the ancient RRDtool for that:  https://github.com/openwrt/packages/pull/16703 
* current modification of keeping 32 bit time in 32 bit CPUs: https://github.com/openwrt/packages/pull/16835 

We ended up patching the ancient RRDtool version to use 32bit time with 32bit CPUs despite musl always has 64bit time_t.
That is a short-term kludge, but offers no compatibility with the current RRDtool versions and is not future-proof.
(However, that enables the old databases to be compatible with the current musl. That is useful to those (few) that have preserved stats data in sysupgrades.)

Triggered by the recent forum discussion in  https://forum.openwrt.org/t/update-rrdtool-up-to-1-4-or-higher/139473 I have now implemented a new upgrade to RRDtool 1.8.0, and I am considering the possible roadmap scenarios for LuCI stats. 

The crucial aspect of RRDtool 1.8.0 is that RRDtool has in 2005-2008 moved from the ancient self-contained lib to more complex graphics libraries cairo, pango, freetype etc. (requiring also glib2). 
So, the size of the required libraries is approx. 3 MB larger (as measured from a compressed sysupgrade image for RT3200). 
That is a lot of flash space :-(

Due to the size increase, I see no feasibility in wholly replacing the old rrdtool 1.0.50 with the new version.

Instead, there might be two alternative versions of rrdtool and LuCI statistics.

**For that purpose, I have created the patch set that**

https://github.com/hnyman/packages/tree/rrdtool180
* creates the necessary graphics libs in feeds/packages:  harfbuzz, fribidi, pango, cairo
[...]

PR #6798, opened 2023-12-30
[DRAFT] luci-app-statistics: DSL graphs update

Weissnix4711:
- Split up SNR and attenuation into two graphs
- Changes type of attenuation instances to gauge. Not sure if this is the correct type for attenuation data, but snr didn't quite fit either.
- Reorder things to be in the order: down/near first and positive y, up/far second and flipped to negative y
- Sensible units, formatting and totals
- Add more RETX errors
- CRC errors
- Errored seconds

I'm still on 22.03, due to the issues with the HH5a, hence I cannot properly test this. My current test setup uses a collectd-exec and a shell script I hastily threw together which calls ubus to gather the DSL metrics. Crucially, not the same as the actual lua script now in the collectd package, so there is plenty of room for things to go wonky.

Not sure if the graph titles work correctly. %pi is meant to be plugin instance. For what I threw together using collectd-exec ages ago, this is `dsl0`. I'm not honestly sure how the lua script does this.

I think I have added all the necessary changes to the lua script for gathering the extra data, but if someone could test that would be greatly appreciated. https://github.com/openwrt/packages/pull/23011

Paging @bluebrother because I know I'm rewriting much of their code.

Screenshots:

es
crc

Not that there's much to show for the errors lol. Nor flags, my shell script didn't really deal with bools well


systemcrash:
So does attenuation ever become negative (i.e. amplification)?
Uptime probably should not be a monotonically increasing graph, but I can't think how else it would be better, if collectd is on the system. If the system is down, it cannot record up or down time. 

Does the relevant graph depend on its position in an array? 

Hmm. This is from a hh5a? Not to derail this discussion, but how was it to install that device? Did you require serial console as per the wiki?

Weissnix4711:
> So does attenuation ever become negative (i.e. amplification)?

No, to my knowledge it could never be negative. This is why I'd suggest it's good to plot both up and down on the same graph, with up simply being flipped to the negative side.

> Uptime probably should not be a monotonically increasing graph, but I can't think how else it would be better

Neither could I. I've basically just left uptime as it was before, with addition of some small formatting tweaks.

> Does the relevant graph depend on its position in an array?

How you mean, sorry?

> Did you require serial console as per the wiki?

Yes, precisely. I've reflashed multiple HH5a's (or rebranded clones) by soldering directly to the test points on the board.

bluebrother (commented):
Can you explain the reason for these changes? To me these are just style changes, not fixes, and other files in the same folder use the formatting this file originally had: use `"` for strings, not have a whitespace betwen `function` and `(`, not use a trailing comma on the last item in a list.

Did you editor auto-format and introduce those changes?

Weissnix4711:
Admittedly, yes this PR was thrown together rather hastily. There were a lot of stylistic changes in there which were enforced by my IDE's auto formatting. I have just reverted the single quotes and random whitespace. Hopefully that's a bit better and makes the diff more readable.

systemcrash:
Hmm, after a bit more thought, perhaps uptime can simply be displayed as a clock, years, months, days, (weeks?), hours, minutes, seconds, rather than a graph whose limit is axis height will effectively never change shape, only the axis units crammed in will get further between. I don't think rrdtool will do this, but just a large rendered string at top or bottom would be fine.

Weissnix4711:
Not sure I agree to be honest. It can be useful to see the history of when the line previously went down, which a simple counter wouldn't show.

However, I have given it a thought and yes you're right, we need something better than graphing uptime in seconds. Newer versions of rrdtool have the ability to format a time duration into a number of days, hours, minutes, etc. See [the docs](https://oss.oetiker.ch/rrdtool/doc/rrdgraph_graph.en.html). This would be great, but we're still stuck on v1.0.x (newest is v1.8.0, not sure when the time/duration formatting was introduced). I have opened an issue in the packages repo: https://github.com/openwrt/packages/issues/23087.

systemcrash:
I think things like this were noted earlier issue #6139 - some of those are not simple to fix. 

> which a simple counter wouldn't show.

A counter would reset to zero (including everything else on platforms that don't save the graph data to storage), but it's not obvious without seeing the line start from zero again.

systemcrash:
@Weissnix4711 

> However, I have given it a thought and yes you're right, we need something better than graphing uptime in seconds

How about instead of uptime, just graph line up status, which is basically boolean. That clearly shows when the line is not up.

systemcrash:
ping

Neustradamus:
@Weissnix4711: Have you progressed on this PR?

systemcrash:
Closed by 8c9e377820442ca61821589170ddba7ab91f97a1

PR #6810, opened 2024-01-06
luci-mod-network: Expose the auto scan limit channel option

db260179:
/etc/config/wireless has had the feature of limiting the auto scanned channels:

'option channels'

User can set individual channels or a hypen list i.e. 36 48 or 36-100 as per the hostapd documentation (https://w1.fi/cgit/hostap/plain/hostapd/hostapd.conf)

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -934,6 +934,10 @@ return view.extend({
 				o = ss.taboption('general', CBIWifiFrequencyValue, '_freq', '<br />' + _('Operating frequency'));
 				o.ucisection = s.section;
 
+				o = ss.taboption('general', form.Value, 'channels', _('Limit auto scan channels'),
+					_('Define ranges or individual channels e.g. 100 140 34-100'));
+				o.optional = true;
+
 				if (hwtype == 'mac80211') {
 					o = ss.taboption('general', form.Flag, 'legacy_rates', _('Allow legacy 802.11b rates'), _('Legacy or badly behaving devices may require legacy 802.11b rates to interoperate. Airtime efficiency may be significantly reduced where these are used. It is recommended to not allow 802.11b rates where possible.'));
 					o.depends({'_freq': '2g', '!contains': true});


systemcrash:
What advantages does this bring?

db260179:
> What advantages does this bring?

So basically there are a few devices that i own, and im sure other people have had similiar issues, where the auto scan has unreliably set licensed channels (DFS or licensed only) thus preventing clients from connecting to your wifi router. 

This just allows the user to manage this better by setting channels you can use.

Also this option has been available for years in openwrt but never exposed to the web ui.

hnyman:
> What advantages does this bring?

See related discussion in https://forum.openwrt.org/t/5-ghz-wlan-craps-out-upon-radar-detection-start-dfs-cac-failed/71018/12

If no fallback channels list is provided, and the wifi driver detects a radar during operations, the radio might get disabled. That has happened for me:
```
Sat Mar  5 02:00:18 2022 daemon.notice hostapd: wlan0: DFS-RADAR-DETECTED freq=5500 ht_enabled=0 chan_offset=0 chan_width=5 cf1=5570 cf2=0
Sat Mar  5 02:00:18 2022 daemon.notice hostapd: dfs_downgrade_bandwidth: no DFS channels left, waiting for NOP to finish
```

But I think that the "auto" is not needed. The 'channels' fallback list should work also if a preferred DFS channel is provided for 'channel' option. Like below: (note, HE40 channels)
```
        option channel '100'
        option channels '100,116,132,52,36'
```
So that part of the PR looks dubious.

castillofrancodamian:
@systemcrash Have a second AP with DFS channels. When a radar is detected, it switches to another predefined DFS channel.

db260179:
> > What advantages does this bring?
> 
> See related discussion in https://forum.openwrt.org/t/5-ghz-wlan-craps-out-upon-radar-detection-start-dfs-cac-failed/71018/12
> 
> If no fallback channels list is provided, and the wifi driver detects a radar during operations, the radio might get disabled. That has happened for me:
> 
> ```
> Sat Mar  5 02:00:18 2022 daemon.notice hostapd: wlan0: DFS-RADAR-DETECTED freq=5500 ht_enabled=0 chan_offset=0 chan_width=5 cf1=5570 cf2=0
> Sat Mar  5 02:00:18 2022 daemon.notice hostapd: dfs_downgrade_bandwidth: no DFS channels left, waiting for NOP to finish
> ```
> 
> But I think that the "auto" is not needed. The 'channels' fallback list should work also if a preferred DFS channel is provided for 'channel' option. Like below: (note, HE40 channels)
> 
> ```
>         option channel '100'
>         option channels '100,116,132,52,36'
> ```
> 
> So that part of the PR looks dubious.

So the suggestion is remove the 'auto' check part as the 'chanlist' is a fallback list?

Maybe i should change it to a tick box option instead?

db260179:
Actually looking at the official openwrt documentation - https://openwrt.org/docs/guide-user/network/wifi/basic#Common%20options it does state:

> channels	list	no	(regulatory domain specific)	
Use specific channels, when channel is in “auto” mode. 
This option allows hostapd to select one of the provided channels when a channel should be automatically selected.
Channels can be provided as range using hyphen ('-') or individual channels can be specified by space (' ') separated values.

So i cant see how this chanlist would work if not in auto mode?

hnyman:
> i cant see how this chanlist would work if not in auto mode?

The upstream hostapd commit is a bit unclear on that topic. (The OpenWrt wiki text is directly coped from there)

https://w1.fi/cgit/hostap/commit/?id=70ee1be2486f96f9ce9290ed8827cc2f7d141dee

But the current code selecting a new DFS channel after a radar was detected, does not place any restriction for having "auto" as far as I can see.
https://w1.fi/cgit/hostap/tree/src/ap/dfs.c#n240

The function call path goes as:
hostapd_dfs_radar_detected --> hostapd_dfs_start_channel_switch --> dfs_get_valid_channel --> dfs_find_channel

db260179:
> > i cant see how this chanlist would work if not in auto mode?
> 
> The upstream hostapd commit is a bit unclear on that topic. (The OpenWrt wiki text is directly coped from there)
> 
> https://w1.fi/cgit/hostap/commit/?id=70ee1be2486f96f9ce9290ed8827cc2f7d141dee
> 
> But the current code selecting a new DFS channel after a radar was detected, does not place any restriction for having "auto" as far as I can see. https://w1.fi/cgit/hostap/tree/src/ap/dfs.c#n240
> 
> The function call path goes as: hostapd_dfs_radar_detected --> hostapd_dfs_start_channel_switch --> dfs_get_valid_channel --> dfs_find_channel

Yes it seems the hostapd code doesn't require auto to be set, but there seems to be conflict of information with the actual code and the information openwrt has stated?

So for now i dont see the point of removing the 'auto' check, as it really is only relevant for that mode?

hnyman:
> there seems to be conflict of information with the actual code and the information openwrt has stated?
Like I said earlier the text has just been copied from upstream commit (config file example) to OpenWrt wiki, in its original a bit unclear form. 

> for now i dont see the point of removing the 'auto' check, as it really is only relevant for that mode?

And my point is just that I think that "channels"  should work also when a single DFS channel has been selected and later a radar is detected and a new DFS channel is needed.  "auto" is not required in that scenario.  (I haven't seen a radar lockup

You might remove the "auto" check and just expose the option always. 

PS. Spelling:  "seperated" --> "separated"

systemcrash:
Agreed on these points below.


> And my point is just that I think that "channels" should work also when a
> single DFS channel has been selected and later a radar is detected and a
> new DFS channel is needed. "auto" is not required in that scenario. (I
> haven't seen a radar lockup
>
> You might remove the "auto" check and just expose the option always.
>
> PS. Spelling: "seperated" --> "separated"
>

db260179:
> > there seems to be conflict of information with the actual code and the information openwrt has stated?
> > Like I said earlier the text has just been copied from upstream commit (config file example) to OpenWrt wiki, in its original a bit unclear form.
> 
> > for now i dont see the point of removing the 'auto' check, as it really is only relevant for that mode?
> 
> And my point is just that I think that "channels" should work also when a single DFS channel has been selected and later a radar is detected and a new DFS channel is needed. "auto" is not required in that scenario. (I haven't seen a radar lockup
> 
> You might remove the "auto" check and just expose the option always.
> 
> PS. Spelling: "seperated" --> "separated"

Ok, so i have removed the auto part dependency.

Not sure the wording of this feature is correct so please advise to change to something?

hnyman:
I think that the text is otherwise ok, ammiguous enough, but didn't yet fix the spelling of "seperated" --> "separated".

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -934,6 +934,10 @@ return view.extend({
 				o = ss.taboption('general', CBIWifiFrequencyValue, '_freq', '<br />' + _('Operating frequency'));
 				o.ucisection = s.section;
 
+				o = ss.taboption('general', form.Value, 'channels', _('Limit auto scan channels'),
+					_('Space seperated or list range of channels with a hyphen i.e. 100 140 34-100'));
```
systemcrash, line 938: Change to e.g.:
```js
					_('Define ranges or individual channels e.g. 100 140 34-100'));
```

systemcrash (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -934,6 +934,10 @@ return view.extend({
...
+				o.optional = true;
```
jow-, line 939: Please add a validation function for this option.

db260179: Will adding

`o.datatype = 'uinteger';`

I would think making sure numbers are added only? or what a good other type to use?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -934,6 +934,11 @@ return view.extend({
 				o = ss.taboption('general', CBIWifiFrequencyValue, '_freq', '<br />' + _('Operating frequency'));
 				o.ucisection = s.section;
 
+				o = ss.taboption('general', form.Value, 'channels', _('Limit auto scan channels'),
+					_('Define ranges or individual channels e.g. 100 140 34-100'));
+				o.datatype = 'uinteger';
```
systemcrash, line 939: if the datatype check is `uinteger`, ranges won't validate. Are ranges actually possible?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
systemcrash: I might have used a naïve regex along the lines of `[\d- ]+` to limit the accepted character types in this field, and let hostapd deal with what's usable.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
db260179: Probably, but we are dealing with 2g, 5g ranges so could do 1-165?

db260179: Up for any example or suggestions on this?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
systemcrash: Review other code examples, or test that this works:
```js
		o.validate = function(section_id, value) {
			const regex = /[\d -]+/;
			if (!regex.test(value)) {
				return _('Error: %s; expecting space separated numbers or hyphenated number ranges.').format(value);
			}
			return true;
		}
```

jow-: > and let hostapd deal with what's usable.

Does it "deal" with unusable values or will it just refuse to start? In either case the regex above is incomplete. The validation function should check for proper syntax, e.g. no unfinished ranges (`1-` or `-12`, ensure that start value is smaller than end value etc.)

systemcrash: I just tested it again with broken ranges and it starts fine (broken ranges are ignored, and the currently chosen working channel is added into `chanlist` if no other valid ranges were found in `chanlist`). In this case, there is no need for ad absurdum checks (although they don't hurt to have), although your improved regex is also good.

systemcrash: Extended testing:

While invalid ranges don't prevent hostapd startup, they propagate into the config, and likely won't be useful when it comes time to change channels. hostapd seems to stop processing chanlist after the first invalid range.

Space separated ranges: 
`        option channels '36-40 44-48 56-60'` results in `chanlist=36-40 44-48 56-60`

Comma separated ranges: 

`        option channels '36-40,44-48,56-60'` results in `chanlist=36-40,44-48,56-60`

Space+comma separated ranges:

`        option channels '36-40, 44-48, 56-60'` results in `chanlist=36-40, 44-48, 56-60`

Invalid ranges: processing stops after first range

`        option channels 36-, 44-, 56-60'` results in `chanlist=36-,`


Summary: using a widget here might not be a bad idea.

systemcrash: Please test that this works:
```js

		o.validate = function(section_id, value) {
			/* Regular expression to match valid value pattern
			* 36-40 44-48 56-60
			* 36-40,44-48,56-60
			* 36-40, 44-48, 56-60
			* 36, 44, 56-60
			* Fail: 36-, 44-, 56-60
			*/
			var regex = /^(\d+-\d+|\d+)(\s*,\s*(\d+-\d+|\d+))*$/;

			// Check if value matches the pattern
			if (!regex.test(value)) {
				return _('Error: %s; expecting space or comma separated numbers or hyphenated number ranges.').format(value);
			}

			// Split value by commas or spaces
			var ranges = value.split(/\s*,\s*|\s+/);

			// Validate each range
			for(var range of ranges) {
				// If the range has a '-' character
				if (range.includes("-")) {
					var [start, end] = range.split("-").map(Number);
					// Check that start and end are valid integers and start is less than or equal to end
					if (isNaN(start) || isNaN(end) || start > end) {
						return _('Error: %s; expecting space or comma separated numbers or hyphenated number ranges.').format(value);
					}
				} else { // It's a single integer
					if (isNaN(parseInt(range)))
						return _('Error: %s; expecting space or comma separated numbers or hyphenated number ranges.').format(value);
				}
			}
			return true;
		}

```

systemcrash (changes requested):

PR #6815, opened 2024-01-08
[23.05] luci-app-firewall: add ICMPv6 MLD rules

jonathanunderwood:
This cherry picks 73abc0ee6d64bab49889b1d5fbca09f9a8e64745 from master - it would be great if we can get this fix into the next point release of 23.05.


feckert:
Done with commit https://github.com/openwrt/luci/commit/971459df2604ecdef87ad945314a9887e0dff517
The commit should be carried out with a cherry-pick commit. This is what you were missing.
Next time please use the command `git cherry-pick -x <sha256sum>`

PR #6818, opened 2024-01-08
luci-base: Title logon username fields

AlreadyM:
add title just with empty ,keep the page same before and better for user to repalce it with any string what we want to show。


AlreadyM:
can apply this PR now？

PR #6820, opened 2024-01-09
luci-base: Title logon username fields

AlreadyM:
add the title property with empty  keep the same like before。

the effect is let know where to write what the want to show。

will show like this

屏幕截图 2024-01-08 203628

### DIFF
--- a/modules/luci-base/ucode/template/sysauth.ut
+++ b/modules/luci-base/ucode/template/sysauth.ut
@@ -22,7 +22,7 @@
 			<div class="cbi-value">
 				<label class="cbi-value-title">{{ _('Username') }}</label>
 				<div class="cbi-value-field">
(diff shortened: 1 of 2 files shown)


Chuenhua (approved):
SomeBody Help Me, Merge This PR！

yurtpage:
Are you trying to add a title with some text, right? Then it should be not just empty but with a translation like:

    title="{{ _('Input your username to login') }}"

Still I do believe that the title is not needed. It's obvious that the login input is your username.
Please add more details about your motivation and a problem that you are trying to solve.

Chuenhua:
No，just empty！keep same behavior like before. the title property empty add for the bash command can repleace the point。it`ll be good at someone want to custom will script. what need to do is just repleace it，and will keep the same display with empty.

PR #6829, opened 2024-01-12
luci-app-usteer: added nslookup of AP name

Ramon00:
added nslookup of AP name

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -10,6 +10,22 @@
 
 var Hosts, Remotehosts, Remoteinfo, Localinfo, Clients;
 
+var dns_cache = [];
+
+function SplitWlan(wlan) {
+	var wlansplit = [];
+	if (typeof wlan.split('#')[1] !== 'undefined') {
+		wlansplit=wlan.split('#');
+		if (typeof dns_cache[wlansplit[0]] !== 'undefined') {
+			wlansplit[0]=dns_cache[wlansplit[0]];
+		}
+	} else {
+		wlansplit[0]=_('This AP'); 
+		wlansplit[1]=wlan; 
+	}
+	return wlansplit;
+}
+
 
 function collectHearingClient(client_table_entries, mac) {
 	if (typeof Clients[mac] !== 'undefined') {
@@ -24,8 +40,10 @@ function collectHearingClient(client_table_entries, mac) {
 				SSID = Remoteinfo[wlanc]['ssid'];
 				freq = Remoteinfo[wlanc]['freq'];
 			}
+			var wlansplit=SplitWlan(wlanc);
 			client_table_entries.push([
-				'<nobr>' + wlanc + '</nobr>',
+				'<nobr>' + '%h'.format(wlansplit[0]) + '</nobr>',
+				'<nobr>' + '%h'.format(wlansplit[1]) + '</nobr>',
 				SSID,
 				freq,
 				Clients[mac][wlanc]['connected'] === true ? 'Yes' : 'No',
@@ -57,7 +75,8 @@ var HearingMap = form.DummyValue.extend({
 			);
 			var client_table = E('table', {'class': 'table cbi-section-table','id':'client_table'+macn}, [
 				E('tr', {'class': 'tr table-titles'}, [
-					E('th', {'class': 'th', 'style': 'width:35%'}, _('IP & Interface','Combination of IP and interface name in usteer overview')),
+					E('th', {'class': 'th'}, _('AP','Name or IP address of access point')),
+					E('th', {'class': 'th'}, _('Interface name','interface name in usteer overview')),
 					E('th', {'class': 'th', 'style': 'width:25%'}, _('SSID')),
 					E('th', {'class': 'th', 'style': 'width:15%'}, _('Frequency','BSS operating frequency in usteer overview')),
 					E('th', {'class': 'th', 'style': 'width:15%'}, _('Connected','Connection state in usteer overview')),
@@ -74,10 +93,14 @@ var HearingMap = form.DummyValue.extend({
 });
 
 
+
+
 function collectWlanAPInfoEntries(connectioninfo_table_entries, wlanAPInfos) {
 	for (var wlan in wlanAPInfos) {
+		var wlansplit=SplitWlan(wlan);
 		connectioninfo_table_entries.push([
-			'<nobr>' + wlan + '</nobr>',
+			'<nobr>' + '%h'.format(wlansplit[0]) + '</nobr>',
+			'<nobr>' + '%h'.format(wlansplit[1]) + '</nobr>',
 			wlanAPInfos[wlan]['bssid'],
 			wlanAPInfos[wlan]['ssid'],
 			wlanAPInfos[wlan]['freq'],
@@ -133,8 +156,10 @@ function collectWlanAPInfos(compactconnectioninfo_table_entries, wlanAPInfos) {
 						);
 					}
 		}
+		var wlansplit=SplitWlan(wlan);
 		compactconnectioninfo_table_entries.push([
-			'<nobr>'+wlan+'</nobr>', 
+			'<nobr>' + '%h'.format(wlansplit[0]) + '</nobr>',
+			'<nobr>' + '%h'.format(wlansplit[1]) + '</nobr>',
 			wlanAPInfos[wlan]['ssid'],
 			wlanAPInfos[wlan]['freq'],
 			wlanAPInfos[wlan]['load'],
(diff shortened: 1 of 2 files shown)


Ramon00:
if anybody can test, that would be great

applications/luci-app-usteer/root/usr/share/rpcd/acl.d/luci-app-usteer.json
```diff
@@ -6,6 +6,7 @@
         "uci": [ "get" ],
         "usteer": [ "remote_info","remote_hosts", "local_info", "get_clients"],
         "luci-rpc": [ "getHostHints" ]
+        "network.rrdns": [ "lookup" ]
```
McGiverGim, line 9: I think this needs a comma:
```suggestion
        "luci-rpc": [ "getHostHints" ],
        "network.rrdns": [ "lookup" ]
```

McGiverGim:
Tested but it seems not to work if I enable to use IPv6 in the config.
I have observed that the `Hosts` variable contains both, the IPv4 and the IPv6 of the hosts, for example, this are my two APs:
image

Maybe you can search for the IP at the `Hosts` variable?

applications/luci-app-usteer/root/usr/share/rpcd/acl.d/luci-app-usteer.json
Ramon00: i think so too :)

Ramon00:
I actually did not test IPv6. Seems the issue is that it uses a link local address, which does not appear to resolve with an nslookup. I can manually search the hosthints but well that is kinda inefficients if there is a large list... So hmmm

McGiverGim:
It's not similar to the search for mac, ip, name, etc. of each device?

Ramon00:
All the other lookups are just indexed by mac address. Another issue is that  all the link local addresses of the other APs are only available on the router (at least not on my setup), while the IPv4 can be made availble elsewhere via dnsmasq. So im not sure how to proceed. I could do the one for IPv4 and the other for IPv6 i suppose

Ramon00:
To be honest I dont know if that is a problem, not in my setup, but not sure how it scales... Anyway I added a search in the hosthints. Give it a try

McGiverGim:
> Give it a try

Working perfectly!! In my case only two APs, and about 15 devices in hearing map, so I can't speak about performance. It works fast in my case.

Maybe some people want to have the MAC too, not only the name. To me is enough like it is now, but only a suggestion.

McGiverGim:
I suppose in the worst case, if someone complains for performance, exists the possibility to add a check in the hearing tab. When checked, do the translation, if not let the mac alone.

Ramon00:
> about 15 devices in hearing map
Its about the amount of APs times the amount of IP address in the host hints. Anyway, the way it is implemented now is that it only does it one time the page is loaded, not every refresh. So I think it should be fine.

McGiverGim:
@Ramon00 I think you have closed this by mistake?

McGiverGim (approved):

Ramon00:
@systemcrash Could you do a review of the code and merge if ok?

systemcrash:
I don't use usteer, but the code largely looks acceptable. @McGiverGim seems to verify its stability.

Performance we can iterate on.

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -144,22 +169,55 @@ function collectWlanAPInfos(compactconnectioninfo_table_entries, wlanAPInfos) {
...
 function collectRemoteHosts (remotehosttableentries,Remotehosts) {
+	var ipAddrs = [];
 	for (var IPaddr in Remotehosts) {
-			remotehosttableentries.push([IPaddr, Remotehosts[IPaddr]['id']]);
-		}
+		if (typeof dns_cache[IPaddr] === 'undefined') ipAddrs.push(IPaddr);
+	}
+
+	L.resolveDefault(callNetworkRrdnsLookup(ipAddrs, 1000, 1000), {}).then(function(replies) {
+				for (var index in ipAddrs) {
```
systemcrash, line 187: Is `index` an enumerable property of `ipAddrs`? Or an iterable object? The latter, I think, should use:

```js
				for (var index of ipAddrs) {
```

e.g.
```js
let list = [4, 5, 6];

for (let i in list) {
   console.log(i); // "0", "1", "2",
}

for (let i of list) {
   console.log(i); // "4", "5", "6"
}
```

In this case, simply:
```js
for (var ip of ipAddrs) {
  if (!ip)
    continue;
  if (replies[ip]) {
...
```
```

File: applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js [hunk 3]
```diff
@@ -144,22 +169,55 @@ function collectWlanAPInfos(compactconnectioninfo_table_entries, wlanAPInfos) {
...
+					var address = ipAddrs[index];
+					if (!address)
+						continue;
+					if (replies[address]) {
+						dns_cache[address] = replies[address];
+						continue;
+					} else {
+						for (var IPaddr in Remotehosts)
+							for (var mac in Hosts) {
```
systemcrash, line 196: I think this should be an `of` loop

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -144,22 +169,55 @@ function collectWlanAPInfos(compactconnectioninfo_table_entries, wlanAPInfos) {
...
+	object: 'network.rrdns',
+	method: 'lookup',
+	params: [ 'addrs', 'timeout', 'limit' ],
+	expect: { '': {} }
+});
+
+
 function collectRemoteHosts (remotehosttableentries,Remotehosts) {
+	var ipAddrs = [];
 	for (var IPaddr in Remotehosts) {
```
systemcrash, line 182: Replace this for loop using a filter:

```js
const getUndefinedDnsCacheIPs = (Remotehosts, dns_cache) =>
  Object.keys(Remotehosts).filter(IPaddr => !dns_cache.hasOwnProperty(IPaddr));
ipAddrs = getUndefinedDnsCacheIPs(Remotehosts, dns_cache);
```
(I'm freestyling now so it'll need testing)

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -144,22 +169,55 @@ function collectWlanAPInfos(compactconnectioninfo_table_entries, wlanAPInfos) {
...
+					var address = ipAddrs[index];
+					if (!address)
+						continue;
+					if (replies[address]) {
+						dns_cache[address] = replies[address];
+						continue;
+					} else {
+						for (var IPaddr in Remotehosts)
+							for (var mac in Hosts) {
+								for (var IPaddr2 in Hosts[mac]['ip6addrs'])
```
systemcrash, line 197: Same for these two `for` loops I think

systemcrash (changes requested):

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
Ramon00: Ah yes, using "of" is indeed better

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
Ramon00: actually no, i think "in" is the correct usage here.  Agree?
`root@OpenWrt6:~# ubus call usteer "remote_hosts"
{
        "192.168.64.2": {
                "id": -454639439
        },
        "192.168.64.1": {
                "id": -812651663
        }
}
root@OpenWrt6:~# ubus call luci-rpc "getHostHints"
{
        "00:11:22:33:44:55": {
                "ipaddrs": [
                        "192.168.64.65"
                ],
                "ip6addrs": [

                ],
                "name": "xyz"
        },

'

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
Ramon00: nice freestyle, worked without editing :)

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
Ramon00: Ok have a look. I replaced them with a .find(...)

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
systemcrash: I think `of`. you want the actual IP, and not a loop counting integer, correct?

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
systemcrash: :D

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
systemcrash: Looks great.

systemcrash:
So while I'm not a big fan of global variables like `var dns_cache = [];`, the only other way around it is variable passing. Not a deal breaker. Everything works for you as intended, yes?

systemcrash (approved):

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
systemcrash: Just a concern at this point: is it possible to enter that code block and the mac or Hosts is empty, producing a theoretical e.g.:

dns_cache[address]=Hosts[undefined]['name'];

You know what's happening best. Just checking.

Ramon00:
yes everything works as intended

systemcrash:
OK - thank you :)

Ramon00:
@systemcrash  And thank you :)

PR #6832, opened 2024-01-13
luci-app-ksmbd: update kmod version to vermagic

pesa1234:
ksmbd now is inside on kernel, switch to vermagic to load the correct kmod version


stokito (approved):

stokito:
Why did you closed the PR? Also you made merge instead of rebase so the commit history log contains not needed commit.

stokito:
Thank you for the response, I just wanted to double check that you didn't closed it accidentally.

PR #6840, opened 2024-01-18
luci-mod-network: dhcp tags improvements

jjm2473:
1. host tag add candidates
2. add a tab for tags

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -322,6 +322,7 @@ return view.extend({
 		s.tab('general', _('General Settings'));
 		s.tab('advanced', _('Advanced Settings'));
 		s.tab('leases', _('Static Leases'));
+		s.tab('tags', _('Tag'));
 		s.tab('files', _('Resolv and Hosts Files'));
 		s.tab('hosts', _('Hostnames'));
 		s.tab('ipsets', _('IP Sets'));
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -1024,6 +1029,20 @@ return view.extend({
 		if (has_dhcpv6)
 			o = s.taboption('leases', CBILease6Status, '__status6__');
 
+		o = s.taboption('tags', form.SectionValue, '__tags__', form.GridSection, 'tag', null,
+			_('Use tags to advertise different gateway and/or DNS to different hosts. Hosts can be associated with tags in "%s" tab').format(_('Static Leases')) + '<br />' +
+			_('DHCP Options') + ': ' + _('Define additional DHCP options,  for example "<code>6,192.168.2.1,192.168.2.2</code>" which advertises different DNS servers to clients.') +
+			_('"<code>3,192.168.2.1</code>" which advertises different gateway to clients.') + '<br />' +
+			_('Note: Do not use "odhcpd" or network interface (such as "lan", "wan", "wan6", etc.) as name when adding tags, conflicts will occur. It is recommended to prefix the name with "t_" to avoid such conflicts.'));
```
jow-, line 1036: We should add a section name validator to check for this. Iirc the generic logic will already ensure that a chosen section name is unique, in case it is not, we should add a validator.

jjm2473: I have tested on 22.03, section name conflict occurs, sections with same name will merge to a single section, section type will be 'tag'

systemcrash:
Closing in favour of #7178 (which includes this functionality)

PR #6845, opened 2024-01-19
luci-app-ksmbd: update kmod from version to vermagic

pesa1234:
Now the kmod version on luci show "?", if we switch from version to vermagic it will show correct version of kmod.


pesa1234:
immagine

systemcrash:
Thank you :)

PR #6846, opened 2024-01-19
luci-mod-network: add disabled option for interface

stokito:
The PR adds a UI option to disable a network interface.
E.g. I have two ISPs with different WAN config  (PPPoE with creds and just DHCP) and if one is not working I switching to another and enabling it's interface.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -4539,7 +4539,8 @@ var UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 						for (var j = 0; this.changes && this.changes.network && j < this.changes.network.length; j++) {
 							var chg = this.changes.network[j];
 
-							if (chg[0] == 'set' && chg[1] == iif && (chg[2] == 'proto' || chg[2] == 'ipaddr' || chg[2] == 'netmask'))
+							if (chg[0] == 'set' && chg[1] == iif &&
+								((chg[2] == 'disabled' && chg[3] == '1') || chg[2] == 'proto' || chg[2] == 'ipaddr' || chg[2] == 'netmask'))
 								return iif;
 						}
 					}
--- a/modules/luci-base/po/ar/base.po
+++ b/modules/luci-base/po/ar/base.po
@@ -2435,6 +2435,10 @@ msgstr "تعطيل التشفير"
 msgid "Disable Inactivity Polling"
 msgstr "تعطيل أخذ عينات الخمول"
 
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:598
+msgid "Disable this interface"
+msgstr "تعطيل الواجهة"
+
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:889
 msgid "Disable this network"
 msgstr "تعطيل هذه الشبكة"
--- a/modules/luci-base/po/bg/base.po
+++ b/modules/luci-base/po/bg/base.po
@@ -2399,6 +2399,10 @@ msgstr "Деактивиране на криптирането"
 msgid "Disable Inactivity Polling"
 msgstr ""
 
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:598
+msgid "Disable this interface"
+msgstr "Забраняване на тази интерфейс"
+
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:889
 msgid "Disable this network"
 msgstr "Забраняване на тази мрежа"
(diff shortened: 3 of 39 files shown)


modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -4007,15 +4007,15 @@ var CBIFlagValue = CBIValue.extend(/** @lends LuCI.form.FlagValue.prototype */ {
 	 * Sets the input value to use for the checkbox checked state.
 	 *
 	 * @name LuCI.form.FlagValue.prototype#enabled
-	 * @type number
+	 * @type string
 	 * @default 1
```
systemcrash, line 4011: This change is not correct: since luci accepts 1 as true and 0 as false, whether quoted or not.

systemcrash:
Everything else looks good - see comment.

systemcrash:
Please split the JSDoc fixes to a separate PR and retain only here:

luci-mod-network: disable interface

modules/luci-base/htdocs/luci-static/resources/form.js
stokito: the comment of a commit:


    luci-base: FlagValue fix type of enabled and disabled properties
    The form.Flag checkbox allows to specify a value for enabled/disabled e.g. for a dropbear:
    
        o = s.option(form.Flag, 'PasswordAuth');
        o.enabled  = 'on';
    
    So the property type should be a string but not just a number.


It's even marked by IntelliJ IDE as a mistyping.

stokito:
@systemcrash Please merge both PRs

jow-:
@stokito - what happens when the disable checkbox is checked on the interface the client is connect to and the changes are applied? Will it trigger the connectivity warning? If not, you probably need to extend https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/ui.js#L4542 as well to include a check for the `disabled` option.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -4539,7 +4539,8 @@ var UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 						for (var j = 0; this.changes && this.changes.network && j < this.changes.network.length; j++) {
 							var chg = this.changes.network[j];
 
-							if (chg[0] == 'set' && chg[1] == iif && (chg[2] == 'proto' || chg[2] == 'ipaddr' || chg[2] == 'netmask'))
+							if (chg[0] == 'set' && chg[1] == iif &&
+								(chg[2] == 'disabled' || chg[2] == 'proto' || chg[2] == 'ipaddr' || chg[2] == 'netmask'))
```
jow-, line 4543: The warning for `disabled` likely only makes sense if the option is getting enabled, can you try this?

```javascript
if (chg[0] == 'set' && chg[1] == iif &&
	((chg[2] == 'disabled' && chg[3] == '1') || chg[2] == 'proto' || chg[2] == 'ipaddr' || chg[2] == 'netmask'))
```

jow-:
Looks fine for me. Final remark/question - should this option really appear on the generic settings tab? My hunch is that the majority of users will never need it...

stokito:
I don't know for sure but I do have a usage scenario.

> I have two ISPs with different WAN config (PPPoE with creds and just DHCP) and if one is not working I switching to another and enabling it's interface without entering creds.

It may not used that often but in the same time just a simple checkbox don't over complicate things.

systemcrash:
> Looks fine for me. Final remark/question - should this option really appear on the generic settings tab? My hunch is that the majority of users will never need it...

In light of this, one should clarify that `Disable` actually means `Disable this interface`.

systemcrash:
@stokito update the string and LGTM

replace `Disable` with `Disable this interface`

stokito:
I changed to `Disable this interface`.
Also I added translations where I was able to translate. For most languages it should be correct, for some I just copied from "disabled" but for the `ur` there wasn't a translation even for it.

As far I know the new translations should be marked in Weblate as "unchecked" so translators will re-check them eventually.

systemcrash:
> I changed to `Disable this interface`. Also I added translations where I was able to translate. For most languages it should be correct, for some I just copied from "disabled" but for the `ur` there wasn't a translation even for it.

I reviewed the translations and unfortunately some still do not refer concretely to "this" interface. I still think it best to leave those to competent translators. 

Just include the blank po templates and we are done :)

stokito:
It still should be enough to understand and definitely better than nothing (yet). As I said, translators will improve this if they want.
Just imagine that everything is written in Urdu that you completely don't have a clue and if it will be something like just "disable" you will be happy.
In this cases even a stupid machine translation can make a huge deal. But the problem with it is that if may translate with a different or confusing meaning like homonyms.
In this case this is not happening because the "disable" translation was copied.
So please merge it as is.
"The best is the enemy of good"

systemcrash:
Closed via 8a0ee6689f18e9f14965b45f8c79211644c9b58a

systemcrash:
Thanks :)

PR #6862, opened 2024-01-23
luci-mod-network: wireless.js: fix 802.11r dependencies

jesferman:
Don't allow enabling 802.11r when using WPA, it needs at least WPA2.

This is because 802.11r advertises FT support in-part through the Authentication and Key Management (AKM) suites in the Robust Security Network (RSN) Information Element, which was included in the 802.11i amendment and WPA2 certification program.

Pre-standard WPA did not include the RSN IE, but the WPA IE. This IE can not advertise the AKM suite for FT.

Also restrict ft_psk_generate_local to psk2 and psk-mixed. Using it with non-psk encryptions breaks roaming.

Same changes sent to OpenWRT's hostapd in this [PR](https://github.com/openwrt/openwrt/pull/14456)

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1548,9 +1548,9 @@ return view.extend({
 					var has_80211r = L.hasSystemFeature('hostapd', '11r') || L.hasSystemFeature('hostapd', 'eap');
 
 					o = ss.taboption('roaming', form.Flag, 'ieee80211r', _('802.11r Fast Transition'), _('Enables fast roaming among access points that belong to the same Mobility Domain'));
-					add_dependency_permutations(o, { mode: ['ap', 'ap-wds'], encryption: ['wpa', 'wpa2', 'wpa3', 'wpa3-mixed'] });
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1577,7 +1577,7 @@ return view.extend({
 					o.rmempty = true;
 
 					o = ss.taboption('roaming', form.Flag, 'ft_psk_generate_local', _('Generate PMK locally'), _('When using a PSK, the PMK can be automatically generated. When enabled, the R0/R1 key options below are not applied. Disable this to use the R0 and R1 key options.'));
-					o.depends({ ieee80211r: '1' });
+					add_dependency_permutations(o, { ieee80211r: ['1'], mode: ['ap', 'ap-wds'], encryption: ['psk2', 'psk-mixed'] });
```
jow-, line 1580: Is the omission of SAE intentional?

jesferman: Yes, it must only be used for psk2 and psk/psk2 mixed

jow-:
Merged, thank you!

PR #6863, opened 2024-01-23
luci-app-upnp: Update plugin title/menu to include PCP as supported

Self-Hosting-Group:
and revise wording

- Revise wording, remove redundancies and mention IPv6 limitations
- Mention `UPnP IGD` as `UPnP` is probably not specific enough
- Include current wording in `LUCI_TITLE` for easy finding
- Fix 4 multiple failing strings in translations and remove old comments

Updated title: `UPnP IGD & PCP/NAT-PMP Service`
Current title: `Universal Plug & Play`
Updated menu option: `UPnP IGD & PCP/NAT-PMP`
Current menu option: `UPnP`
Updated wording: `port forwards` (used in Network -> Firewall)
Current wording: `redirects`

Port Control Protocol (PCP) is the successor to NAT-PMP and has similar
protocol concepts and packet formats, but adds IPv6 support.
*PCP standard:*
https://datatracker.ietf.org/doc/html/rfc6887
https://en.wikipedia.org/wiki/Port_Control_Protocol
*NAT-PMP std. (see 9.1 Simplicity - 9.3 for some diff. to UPnP IGD):*
https://datatracker.ietf.org/doc/html/rfc6886
*Port Mapping Protocols Overview:*

### DIFF
--- a/applications/luci-app-upnp/Makefile
+++ b/applications/luci-app-upnp/Makefile
@@ -6,7 +6,7 @@
 
 include $(TOPDIR)/rules.mk
 
-LUCI_TITLE:=Universal Plug & Play configuration module
+LUCI_TITLE:=UPnP IGD & PCP/NAT-PMP (Universal Plug and Play) conf. module
 LUCI_DEPENDS:=+luci-base +miniupnpd +rpcd-mod-ucode
 
 include ../../luci.mk
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
@@ -28,7 +28,7 @@ handleDelRule = function(num, ev) {
 };
 
 return baseclass.extend({
-	title: _('Active UPnP Redirects'),
+	title: _('Active UPnP IGD & PCP/NAT-PMP Port Forwards'),
 
 	load: function() {
 		return Promise.all([
@@ -67,7 +67,7 @@ return baseclass.extend({
 			];
 		});
 
-		cbi_update_table(table, rows, E('em', _('There are no active redirects.')));
+		cbi_update_table(table, rows, E('em', _('There are no active port forwards.')));
 
 		return table;
 	}
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
@@ -63,7 +63,7 @@ return view.extend({
 			];
 		});
 
-		cbi_update_table(nodes.querySelector('#upnp_status_table'), rows, E('em', _('There are no active redirects.')));
+		cbi_update_table(nodes.querySelector('#upnp_status_table'), rows, E('em', _('There are no active port forwards.')));
 
 		return;
 	},
@@ -72,8 +72,8 @@ return view.extend({
 
 		var m, s, o;
 
-		m = new form.Map('upnpd', [_('Universal Plug & Play')],
-			_('UPnP allows clients in the local network to automatically configure the router.'));
+		m = new form.Map('upnpd', [_('UPnP IGD & PCP/NAT-PMP Service')],
+			_('UPnP IGD & PCP/NAT-PMP allows clients on the local network to automatically configure port forwards on the router. Also called Universal Plug and Play.'));
 
 		s = m.section(form.GridSection, '_active_rules');
 
@@ -107,43 +107,43 @@ return view.extend({
 				];
 			});
 
-			cbi_update_table(table, rows, E('em', _('There are no active redirects.')));
+			cbi_update_table(table, rows, E('em', _('There are no active port forwards.')));
 
 			return E('div', { 'class': 'cbi-section cbi-tblsection' }, [
-					E('h3', _('Active UPnP Redirects')), table ]);
+					E('h3', _('Active Service Port Forwards')), table ]);
 		}, o, this);
 
-		s = m.section(form.NamedSection, 'config', 'upnpd', _('MiniUPnP settings'));
+		s = m.section(form.NamedSection, 'config', 'upnpd', _('Service Settings'));
 		s.addremove = false;
 		s.tab('general',  _('General Settings'));
 		s.tab('advanced', _('Advanced Settings'));
 
-		o = s.taboption('general', form.Flag, 'enabled', _('Start UPnP and NAT-PMP service'));
+		o = s.taboption('general', form.Flag, 'enabled', _('Start service'));
 		o.rmempty  = false;
 
-		s.taboption('general', form.Flag, 'enable_upnp', _('Enable UPnP functionality')).default = '1'
-		s.taboption('general', form.Flag, 'enable_natpmp', _('Enable NAT-PMP functionality')).default = '1'
+		s.taboption('general', form.Flag, 'enable_upnp', _('Enable UPnP IGD protocol')).default = '1'
+		s.taboption('general', form.Flag, 'enable_natpmp', _('Enable PCP/NAT-PMP protocol')).default = '1'
 
 		s.taboption('general', form.Flag, 'secure_mode', _('Enable secure mode'),
-			_('Allow adding forwards only to requesting ip addresses')).default = '1'
+			_('Allow adding port forwards only to requesting IP addresses')).default = '1'
 
-		s.taboption('general', form.Flag, 'igdv1', _('Enable IGDv1 mode'),
-			_('Advertise as IGDv1 device instead of IGDv2')).default = '0'
+		s.taboption('general', form.Flag, 'igdv1', _('Enable UPnP IGDv1 mode'),
+			_('Advertise as UPnP IGDv1 device (no IPv6) instead of IGDv2')).default = '0'
 
 		s.taboption('general', form.Flag, 'log_output', _('Enable additional logging'),
 			_('Puts extra debugging information into the system log'))
 
-		s.taboption('general', form.Value, 'download', _('Downlink'),
+		s.taboption('general', form.Value, 'download', _('Download speed'),
 			_('Value in KByte/s, informational only')).rmempty = true
 
-		s.taboption('general', form.Value, 'upload', _('Uplink'),
+		s.taboption('general', form.Value, 'upload', _('Upload speed'),
 			_('Value in KByte/s, informational only')).rmempty = true
 
 		o = s.taboption('general', form.Value, 'port', _('Port'))
 		o.datatype = 'port'
 		o.default  = 5000
 
-		s.taboption('advanced', form.Flag, 'system_uptime', _('Report system instead of daemon uptime')).default = '1'
+		s.taboption('advanced', form.Flag, 'system_uptime', _('Report system instead of service uptime')).default = '1'
 
 		s.taboption('advanced', form.Value, 'uuid', _('Device UUID'))
 		s.taboption('advanced', form.Value, 'serial_number', _('Announced serial number'))
@@ -164,7 +164,7 @@ return view.extend({
 		o = s.taboption('advanced', form.Value, 'presentation_url', _('Presentation URL'))
 		o.placeholder = 'http://192.168.1.1/'
 
-		o = s.taboption('advanced', form.Value, 'upnp_lease_file', _('UPnP lease file'))
+		o = s.taboption('advanced', form.Value, 'upnp_lease_file', _('Service lease file'))
 		o.placeholder = '/var/run/miniupnpd.leases'
 
 		s.taboption('advanced', form.Flag, 'use_stun', _('Use STUN'))
@@ -178,8 +178,8 @@ return view.extend({
 		o.datatype    = 'port'
 		o.placeholder = '0-65535'
 
-		s = m.section(form.GridSection, 'perm_rule', _('MiniUPnP ACLs'),
-			_('ACLs specify which external ports may be redirected to which internal addresses and ports'))
+		s = m.section(form.GridSection, 'perm_rule', _('Service ACLs'),
+			_('ACLs specify which external ports can be forwarded to which client addresses and ports, IPv6 always allowed.'))
 
 		s.sortable  = true
 		s.anonymous = true
@@ -187,15 +187,15 @@ return view.extend({
 
 		s.option(form.Value, 'comment', _('Comment'))
 
-		o = s.option(form.Value, 'ext_ports', _('External ports'))
+		o = s.option(form.Value, 'ext_ports', _('External Port'))
 		o.datatype    = 'portrange'
 		o.placeholder = '0-65535'
 
-		o = s.option(form.Value, 'int_addr', _('Internal addresses'))
+		o = s.option(form.Value, 'int_addr', _('Client Address'))
 		o.datatype    = 'ip4addr'
 		o.placeholder = '0.0.0.0/0'
 
-		o = s.option(form.Value, 'int_ports', _('Internal ports'))
+		o = s.option(form.Value, 'int_ports', _('Client Port'))
 		o.datatype    = 'portrange'
 		o.placeholder = '0-65535'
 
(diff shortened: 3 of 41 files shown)


applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -72,8 +72,8 @@ return view.extend({
 
 		var m, s, o;
 
-		m = new form.Map('upnpd', [_('Universal Plug & Play')],
-			_('UPnP allows clients in the local network to automatically configure the router.'));
+		m = new form.Map('upnpd', [_('UPnP IGD & PCP/NAT-PMP')],
+			_('UPnP IGD & PCP/NAT-PMP allows clients in the local network to automatically configure the router.'));
```
stokito, line 76: it's better to use "and" instead of &

applications/luci-app-upnp/Makefile
```diff
@@ -6,7 +6,7 @@
 
 include $(TOPDIR)/rules.mk
 
-LUCI_TITLE:=Universal Plug & Play configuration module
+LUCI_TITLE:=UPnP IGD & PCP/NAT-PMP configuration module
```
stokito, line 9: What if make it "Universal Plug & Play (UPnP IGD) and NAT-PMP/PCP".
Then it will be easier to find

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
stokito: Search for ["when to use ampersand vs and"](https://www.google.com/search?q=when+to+use+ampersand+vs+and) says:
> In formal writing like academic papers or business correspondence, spell out the word and instead of using the ampersand. The exception to this rule is if the ampersand is part of an official title, such as a company name.


Translators may be afraid to translate the & and leave it as is.

systemcrash: I think in screen and menu real-estate, we should be more efficient. Although I agree with the stylistic finesse. One could technically get away with 'n' but that might take a while to get used to.

applications/luci-app-upnp/Makefile
systemcrash: How about both? 
```Makefile
LUCI_TITLE:=Universal Plug & Play (UPnP) IGD & PCP/NAT-PMP configuration module
```

systemcrash:
The default build of miniupnpd has NAT-PMP/PCP baked in, correct? If so then I think we can say this is ready and merge it.

applications/luci-app-upnp/po/cs/upnp.po
```diff
@@ -229,6 +231,55 @@ msgstr "Použít STUN"
 msgid "Value in KByte/s, informational only"
 msgstr "Pouze informační hodnoty (v KByte/s)"
 
+#~ msgid ""
```
stokito, line 234: You might just rename the i18n keys.
Now the rea lost

stokito (changes requested):

stokito:
> NAT-PMP/PCP baked in, correct

yes 

41 file changed, many questionable changes.
I don't know if it worth it.

systemcrash:
Only in exceptional cases. If wording changes, it's a strong signal that
should also happen for i18n. Retranslation is currently the only way to
ensure this.


> You might just rename the i18n keys.
> Now the rea lost
>

systemcrash:
Ampersand is fine. It's space efficient and is synonymous. Leave it be 😆

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
```diff
@@ -67,7 +67,7 @@ return baseclass.extend({
 			];
 		});
 
-		cbi_update_table(table, rows, E('em', _('There are no active redirects.')));
+		cbi_update_table(table, rows, E('em', _('There are no active port mappings.')));
```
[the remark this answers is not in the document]
systemcrash: For example which languages?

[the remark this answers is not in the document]
systemcrash: In any case, what language does miniupnpd itself use for opened ports?

systemcrash:
> > Ampersand is fine. It's space efficient and is synonymous. Leave it be 😆
> 
> Should we use an ampersand for the title/description or should we use `,` or `and` there?

The sense of an ampersand is generally for coupled things, not to chain unrelated things together to save space. Leave the ampersand. Comma makes it seem like two unrelated things are in the same package.

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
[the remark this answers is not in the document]
stokito: Ok, what about "forwarding"? It's still may be difficult to translate but this is more precise term as for me.

> For example which languages?

Probably all. The mapping translates to "cartography". The most similar word is "accordance" but it's more abstract.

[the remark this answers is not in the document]
systemcrash: I prefer mapping especially if the original specs use this terminology. Pinhole makes sense for ipv6. Port forwarding is acceptable since that's what firewall uses, I think.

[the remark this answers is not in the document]
hnyman: Port mapping sounds strange to me. Forwarding is more explanatory from end-user perspective

hnyman:
The commit will fail formalities...
No committer name,  no signed-off-by line

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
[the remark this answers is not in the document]
systemcrash: The English language has its idiosyncracies. But mapping sounds great to my native ears.  Forward seems to be preferred then. Can we do that @Self-Hosting-Group ?

systemcrash:
> Should we keep `MiniUPnPd ACLs` and `MiniUPnPd settings` or the simpler `Service ACLs` and `Service Settings` or just `Settings`? An update of the documentation is planned.

`Service ACLs` and `Service Settings` sits well with me, unless we must segregate settings based on which RFC we are talking about here. 

Just commit subject and SOB to be fixed then.

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -153,18 +153,18 @@ return view.extend({
 		o.datatype    = 'uinteger'
 		o.placeholder = 30
 
-		o = s.taboption('advanced', form.Value, 'clean_ruleset_threshold', _('Clean rules threshold'))
+		o = s.taboption('advanced', form.Value, 'clean_ruleset_threshold', _('Clean rules threshold of number of rules cleaned once'))
```
stokito, line 156: Keep the name as is but add a description instead:

```
o = s.taboption('advanced', form.Value, 'clean_ruleset_threshold', _('Clean rules threshold'), _('Clean not used port forwardings once their number reached the threshold'))
```

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
```diff
@@ -28,7 +28,7 @@ handleDelRule = function(num, ev) {
 };
 
 return baseclass.extend({
-	title: _('Active UPnP Redirects'),
+	title: _('Active UPnP IGD & PCP/NAT-PMP Port Forwards'),
```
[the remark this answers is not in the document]
stokito: And BTW "forwards" or "forwardings"? @systemcrash please clarify.

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -153,18 +153,18 @@ return view.extend({
...
 		o.datatype    = 'uinteger'
 		o.placeholder = 20
 
-		o = s.taboption('advanced', form.Value, 'clean_ruleset_interval', _('Clean rules interval'))
+		o = s.taboption('advanced', form.Value, 'clean_ruleset_interval', _('Clean rules (delay) interval'))
```
stokito, line 160: It's not a delay but an interval.

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -178,8 +178,8 @@ return view.extend({
 		o.datatype    = 'port'
 		o.placeholder = '0-65535'
 
-		s = m.section(form.GridSection, 'perm_rule', _('MiniUPnP ACLs'),
-			_('ACLs specify which external ports may be redirected to which internal addresses and ports'))
+		s = m.section(form.GridSection, 'perm_rule', _('Service ACLs'),
```
stokito, line 181: Maybe call it "Forwarding ACL"

applications/luci-app-upnp/po/ca/upnp.po
```diff
@@ -234,6 +238,63 @@ msgstr ""
 msgid "Value in KByte/s, informational only"
 msgstr "Valor en KByte/s, només per informació"
 
+#~ msgid ""
```
stokito, line 241: This was an old translation key and now it's commented after you runned an i18n-update.
This happened because the key (e.g. original message) was changed.

The key was:
'ACLs specify which external ports may be **redirected** to which internal addresses and ports'
But the new is:
'ACLs specify which external ports may be **forwarded** to which internal addresses and ports'


The existing translation will be lost and translators from the Weblate won't see the old translation that they should just slightly change.
Instead you can change the translation key and then the existing translation will be kept. Yes, it will be outdated but it would be easier to slightly fix it.

In other places you just changed "UPNP" to "UPNP & NAT-PMP". This again made the strings not same and translation lost. But you can replace them in existing languages. The `&` will need for a translation (e.g. in Russian it would be `UPNP и NAT-PMP`) but you can leave it as `&` and that should be fine for now.

This would be a boring thing to do but you can use bulk replaces across files in your IDE.
We should keep the old translations because it may take entirety while someone will re-translate messages.

stokito (changes requested):

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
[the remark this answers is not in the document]
stokito: Well... many OpenWrt devs aren't native speakers :) The "Port Forwards" means "the port (itself) is forwards somethings".
On the Wiki it's called forwarding. https://en.wikipedia.org/wiki/Port_forwarding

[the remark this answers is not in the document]
systemcrash: > Well... many OpenWrt devs aren't native speakers :) The "Port Forwards" means "the port (itself) is forwards (something)".
> On the Wiki it's called forwarding. https://en.wikipedia.org/wiki/Port_forwarding

"It" - Forwarding - is the concept, a port forward is the correct noun. 

Neither native speakers, nor grammar anywhere ever advocates for "is forwards". We have gerunds for that. The "Port Forwards" is a plural countable noun in the case of firewall. So "configure a port forward under x" or "this port is forwarded to x". It's tenuous yet possible to say "this port forwards to x", but only because it 'is configured' to do so.

@Self-Hosting-Group you're fine here.

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
systemcrash: Yeah, subtext - keep setting title to three words or fewer.

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
systemcrash: Yeah, let's reword to "Intra-cleanse interval". And a subtext like ("%s means immediate").format("<code>0</code>");

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
systemcrash: Fine as is.

applications/luci-app-upnp/po/ca/upnp.po
systemcrash: Unless he's willing to augment every existing translation, this is the correct way to do it.

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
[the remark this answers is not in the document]
systemcrash: 1. yes
2. Ruleset cleanse interval (where 0= disabled)
3. We don't support those older ones, which is why wiki generally are updated to point to the latest or last two releases
   * one page
   * up to you
   * with what package?

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
systemcrash: o = s.taboption('advanced', form.Value, 'clean_ruleset_threshold', _('Ruleset cleanse interval'), _('Clean unused port forwards every X minutes. 0 is disabled.'))

systemcrash:
```js
o = s.taboption('advanced', form.Value, 'clean_ruleset_threshold', _('Ruleset cleanse threshold'), _('Start to remove unused rules when the amount of rules exceeds this value.'))
```

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
systemcrash:
```js
		o = s.taboption('advanced', form.Value, 'clean_ruleset_interval', _('Interval between cleanse runs'), _('Unit: seconds. %s means disable').format("<code>0</code>"));
```

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
systemcrash: Expired is better.

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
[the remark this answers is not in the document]
systemcrash: Good.

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
systemcrash: Fix the PR checks please.

miniupnp: expired means expired.
@stokito is right about the threshold.

stokito:
As far I understood the Sign-off author is like a sign on a legal document. Thus you must sign it with a real name.
A few PRs that weren't merged because author doesn't want to expose it's name and email.
I hope this policy will be changed because some authors may not just value their privacy but it may be dangerous for them.
Anyway, if you can't set the real name then I or someone else can make the same PR but authored with own name. I hope you won't be resentful

systemcrash:
> The Linux community rule talks about a known identity that allows pseudonyms and I think the patches come there more indirectly and it's probably a more critical piece of software. [Linux kernel documentation update of submitting-patches.rst](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/Documentation/process/submitting-patches.rst?id=d4563201f33a022fc0353033d9dfeb1606a88330)

And those are the checks performed in the openwrt repos; identity, not Real Name. The links talk about `because the sign-off needed to be something we could check back with.` The issue in this case is the SOB line. Because not everyone uses github, I imagine this is why the recourse necessary to have a working email - "to check back with", however horrific that wording may be: a means of contacting the developer.

systemcrash:
OK

PR #6869, opened 2024-01-25
 luci-app-upnp: fix name of access-control-list

reidar-cederqvist:
give upnp a unique acl name that does not conflict with the one of ddns.

### DIFF
--- a/applications/luci-app-upnp/root/usr/share/luci/menu.d/luci-app-upnp.json
+++ b/applications/luci-app-upnp/root/usr/share/luci/menu.d/luci-app-upnp.json
@@ -6,7 +6,7 @@
 			"path": "upnp/upnp"
 		},
 		"depends": {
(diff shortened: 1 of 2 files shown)


reidar-cederqvist:
@systemcrash sorry I am unfamiliar with SOB, I tried to match the commit message from other PRs, please advice if there is anything else you would like me to change

reidar-cederqvist:
I have pushed the change to a work-branch in my fork, but unfortunately it is not possible to change the source branch in an open PR, I can open a new PR if it is preferred that the source branch is other than master

systemcrash:
You've done everything almost correctly. The commit has:

luci-app-ddns: fix

But this regards (I assume)

luci-app-upnp: fix

reidar-cederqvist:
Yes, thanks for noticing, it is fixed now

PR #6877, opened 2024-01-31
luci-app-openvpn: add OpenVPN option disable-dco

sch-m:
Perhaps a user might want to disable DCO support: let's add this option.

This PR depends on https://github.com/openwrt/packages/pull/23207

### DIFF
--- a/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-advanced.lua
+++ b/applications/luci-app-openvpn/luasrc/model/cbi/openvpn-advanced.lua
@@ -371,6 +371,10 @@ local knownParams = {
 			{ "net30", "p2p", "subnet" },
 			translate("'net30', 'p2p', or 'subnet'"),
 			{dev_type="tun" } },
+		{ Flag,
+			"disable_dco",
+			0,
+			translate("Disable Data Channel Offloading (DCO) support") },
 	} },
 
 	{ "vpn", translate("VPN"), {


applications/luci-app-openvpn/luasrc/model/cbi/openvpn-advanced.lua
```diff
@@ -371,6 +371,10 @@ local knownParams = {
 			{ "net30", "p2p", "subnet" },
 			translate("'net30', 'p2p', or 'subnet'"),
 			{dev_type="tun" } },
+		{ Flag,
+			"disable_dco",
+			0,
+			translate("Disable data channel offloading support") },
```
systemcrash, line 377: revise to the following to help users searching settings based on acronyms:
```lua
			translate("Disable Data Channel Offloading (DCO) support") },
```

systemcrash (changes requested):

applications/luci-app-openvpn/luasrc/model/cbi/openvpn-advanced.lua
sch-m: OK, I'll change that.

systemcrash (approved):

sch-m:
@systemcrash You've merged this commit also to openwrt-23.05, but there is no DCO support in openwrt-23.05.

PR #6834, opened 2024-01-14
luci-app-cloudflared: add new package

animegasan:
Application for cloudflared settings via LuCI

**Preview Base settings**
image

**Preview Logs**
image

PR #6887, opened 2024-02-03
luci-app-cloudflared: add Tunnels status page

stokito:
The page allows to see if the tunnel is healthy and has connections.
This can be used for a basic troubleshooting without opening the Cloudflare dashboard.

luci-app-cloudflared tunnels

Maintainer: me and @animegasan
CC @1715173329


animegasan (commented):
image

Try adding write permissions for Config file path and Certificate of Origin. Because I tried deleting and replacing files but it wasn't allowed

stokito:
@animegasan thank you! Fixed in the 
c49d4f0bbc2431eb8596fad8da9e09a838a42dac

animegasan (approved):

applications/luci-app-cloudflared/root/usr/share/rpcd/acl.d/luci-app-cloudflared.json
```diff
@@ -6,11 +6,17 @@
...
-			"file": [ "/var/log/cloudflared.log" ]
+			"file": {
+				"/etc/cloudflared/*": [ "read" ],
+				"/var/log/cloudflared.log": [ "read" ]
+			}
 		},
 		"write": {
 			"uci": [ "cloudflared" ],
-			"file": [ "/etc/cloudflared/" ]
+			"file": {
```
systemcrash, line 16: On the principle of giving the least necessary permissions, please don't do this. Use the permissions I outlined earlier in #6834 
```json
		"write": {
			"uci": ["cloudflared"],
			"file": {
				"/etc/cloudflared/*": ["read"],
				"/etc/cloudflared/config.yml": ["write"],
				"/etc/cloudflared/config.yaml": ["write"],
				"/etc/cloudflared/cert.pem": ["write"],
			}
```

@animegasan can you verify that the above permissions context also works for your test case?

systemcrash (changes requested):

applications/luci-app-cloudflared/htdocs/luci-static/resources/view/cloudflared/tunnels.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * Copyright (C) 2024 Sergey Ponomarev <stokito@gmail.com>
+ */
+
+'use strict';
+'require view';
+'require fs';
+
+function listTunnels() {
+	let command = '/usr/bin/cloudflared';
```
systemcrash, line 11: I would prefer that we not directly call binaries here, and that this is wrapped in an rpc call. ( Remember to add it to permissions )

systemcrash (changes requested):

applications/luci-app-cloudflared/root/usr/share/rpcd/acl.d/luci-app-cloudflared.json
animegasan: > On the principle of giving the least necessary permissions, please don't do this. Use the permissions I outlined earlier in #6834
> 
> ```json
> 		"write": {
> 			"uci": ["cloudflared"],
> 			"file": {
> 				"/etc/cloudflared/*": ["read"],
> 				"/etc/cloudflared/config.yml": ["write"],
> 				"/etc/cloudflared/config.yaml": ["write"],
> 				"/etc/cloudflared/cert.pem": ["write"],
> 			}
> ```
> 
> @animegasan can you verify that the above permissions context also works for your test case?

I used it like this and it worked
```json
                "write": {
                        "uci": [ "cloudflared" ],
                        "file": {
                                "/etc/cloudflared/config.yml": ["write"],
                                "/etc/cloudflared/config.yaml": ["write"],
                                "/etc/cloudflared/cert.pem": ["write"],
                                "/usr/bin/cloudflared *": [ "exec" ]
                        }
                }
```

applications/luci-app-cloudflared/root/usr/share/rpcd/acl.d/luci-app-cloudflared.json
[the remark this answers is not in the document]
stokito: this is just a different notation: a simple array. But for the more granular permissions we need to use an object.

applications/luci-app-cloudflared/root/usr/share/rpcd/acl.d/luci-app-cloudflared.json
stokito: I don't see a problem that the entire /etc/cloudlfared is accessible for the app. I may have a `config.yml.bak` file there and specify it. Other parts of system won't be affected.
So let's leave it as is.

applications/luci-app-cloudflared/htdocs/luci-static/resources/view/cloudflared/tunnels.js
stokito: This is not critical. One of the main reason to use the rpcd is to receive JSON but the cloudflared already gives us an output in JSON. Another one important reason is to call it from other services but that's not case here.
We may add rpcd service script later and rewrite the code.
But today this perfectly works and serves the purpose.

applications/luci-app-cloudflared/Makefile
```diff
@@ -9,7 +9,7 @@ LUCI_DEPENDS:=+cloudflared
 LUCI_DESCRIPTION:=LuCI support for Cloudflare Zero Trust Tunnels
 
 PKG_MAINTAINER:=Hilman Maulana <hilman0.0maulana@gmail.com>, Sergey Ponomarev <stokito@gmail.com>
-PKG_VERSION:=1.0
+PKG_VERSION:=1.1
```
aparcar, line 12: Just out of curiosity, why would you chang the versions manually instead of just letting LuCI do it automatically?

stokito: This is probably makes sense for code Luci modules and it will set a version from git.
Here we have an own semantic version.

systemcrash (approved):

stokito:
@systemcrash I fixed the merge conflict in the Spanish translation. Please merge before someone again changed translation on the Weblate.

PR #6893, opened 2024-02-07
luci-base: extend Expires directive to prevent caching

dhewg:
Some browsers apparently act on fixed dates only, so add that too.
Tested on firefox.

### DIFF
--- a/modules/luci-base/root/www/index.html
+++ b/modules/luci-base/root/www/index.html
@@ -4,7 +4,8 @@
 	<head>
 		<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate" />
 		<meta http-equiv="Pragma" content="no-cache" />
-                <meta http-equiv="Expires" content="0" />
+		<meta http-equiv="Expires" content="0" />
+		<meta http-equiv="Expires" content="Thu, 01 Jan 1970 00:00:00 GMT" />
 		<meta http-equiv="refresh" content="0; URL=cgi-bin/luci/" />
 		<style type="text/css">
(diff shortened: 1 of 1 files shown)


systemcrash:
Thank you for this PR. 

Is 1980 a specific choice? Because fresh booting a device after a clean install without network (or NTP) typically puts the date at 1970. Choosing a specific date could interfere here in unexpected ways until device syncs time with NTP or browser. So maybe it makes sense to expire it at some point during the Apollo moon landing :D

dhewg:
Good question, this commit is a bit older. I didn't make it up myself, 
but I don't recall where I took it from. But it appears it's a common 
date to solve the issue at hand:
https://www.google.com/search?hl=en&q=http-equiv%3D%22Expires%22+content%3D%22Tue%2C+01+Jan+1980+1%3A00%3A00+GMT%22

And in case you meant the date of the openwrt device, that doesn't 
matter here. The date of the box where the browser runs on does. And if 
that's too far off it'll break https://, so nowadays I guess we can 
assume it's not? :)

systemcrash:
I could not find any justification for using exactly that date. POSIX systems start their clocks at 1970, so I figure that's a fair point to use. If it breaks https (after the connection was made to load the page), then it's a sign the date works, surely.

dhewg:
It wasn't meant as a justification. Browers are 50% code and 50% hacks&workarounds to me. So I'd stick to known working values instead of getting eye cancer by researching their code base :D

Anyway, I changed to to 1970/1/1, which seems to works too

systemcrash:
OK. I'm going to give others some time to comment on this.

systemcrash:
Good - thanks :)

PR #6895, opened 2024-02-07
luci-mod-network: put each IPv6 address on a separate line

Alphix:
Right now, when there's active DHCPv6 leases and more than one address per host (e.g. because ULA is used together with GUA prefixes), the two IPv6 addresses will be printed on one line (which may or may not get broken up depending on the length of the addresses, which also looks inconsistent).

Putting each address on a separate line makes it much easier to read the addresses (IMHO).


systemcrash:
Good - thank you.

Alphix:
> Good - thank you.

@systemcrash great, thanks. Any chance this could be cherry-picked for the 23.05 branch as well?

PR #6897, opened 2024-02-08
luci-app-opkg: show disk space consistent with overview

stangri:
* Show Disk space graph consistent with the Status->Overview page
* Brought up in https://forum.openwrt.org/t/software-space-is-going-to-full/187112
* Leading whitespaces reformatted by vscode
* Localizable disk space progress bar title
* Kudos to https://forum.openwrt.org/u/psherman for coming up with final design

I'm attaching screenshot of the change, which is consistent how space is displayed in Status->Overview
image

### DIFF
--- a/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
+++ b/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
@@ -1088,11 +1088,11 @@ function updateLists(data)
 
 	return (data ? Promise.resolve(data) : downloadLists()).then(function(data) {
 		var pg = document.querySelector('.cbi-progressbar'),
-		    mount = L.toArray(data[0].filter(function(m) { return m.mount == '/' || m.mount == '/overlay' }))
-		    	.sort(function(a, b) { return a.mount > b.mount })[0] || { size: 0, free: 0 };
+			mount = L.toArray(data[0].filter(function(m) { return m.mount == '/' || m.mount == '/overlay' }))
+				.sort(function(a, b) { return a.mount > b.mount })[0] || { size: 0, free: 0 };
 
-		pg.firstElementChild.style.width = Math.floor(mount.size ? ((100 / mount.size) * mount.free) : 100) + '%';
-		pg.setAttribute('title', '%s (%1024mB)'.format(pg.firstElementChild.style.width, mount.free));
+		pg.firstElementChild.style.width = Math.floor(mount.size ? (100 / mount.size) * (mount.size - mount.free) : 100) + '%';
+		pg.setAttribute('title', _('%s used (%1024mB used of %1024mB, %1024mB free)').format(pg.firstElementChild.style.width, mount.size - mount.free, mount.size, mount.free));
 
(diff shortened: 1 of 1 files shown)


applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -1088,11 +1088,11 @@ function updateLists(data)
...
 		var pg = document.querySelector('.cbi-progressbar'),
-		    mount = L.toArray(data[0].filter(function(m) { return m.mount == '/' || m.mount == '/overlay' }))
-		    	.sort(function(a, b) { return a.mount > b.mount })[0] || { size: 0, free: 0 };
+			mount = L.toArray(data[0].filter(function(m) { return m.mount == '/' || m.mount == '/overlay' }))
+				.sort(function(a, b) { return a.mount > b.mount })[0] || { size: 0, free: 0 };
 
-		pg.firstElementChild.style.width = Math.floor(mount.size ? ((100 / mount.size) * mount.free) : 100) + '%';
-		pg.setAttribute('title', '%s (%1024mB)'.format(pg.firstElementChild.style.width, mount.free));
+		pg.firstElementChild.style.width = Math.floor(mount.size ? (100 / mount.size) * (mount.size - mount.free) : 100) + '%';
+		pg.setAttribute('title', '%s used (%1024mB of %1024mB)'.format(pg.firstElementChild.style.width, mount.size - mount.free, mount.free));
```
McGiverGim, line 1095: The %s used... is a phrase that can be translated?

stangri: Made the whole disk space progress bar title localizable.

McGiverGim: Thanks!

McGiverGim (approved):

hnyman:
I did not run test it, but looks nice.

However, I am thinking if the percentage number is actually useful at all, as the remaining free space in MiB is the most relevant info at the package installation page (to determine how much can still be installed).

I wonder if it might be more useful to drop the percentage number from text (but keep the graph bar), and change the text to be something like "7.2 MiB used of 16 MiB, 8.8 MiB still free"

stangri:
> However, I am thinking if the percentage number is actually useful at all, as the remaining free space in MiB is the most relevant info at the package installation page (to determine how much can still be installed).
> 
> I wonder if it might be more useful to drop the percentage number from text (but keep the graph bar), and change the text to be something like "7.2 MiB used of 16 MiB, 8.8 MiB still free"

This matches the suggestion on the forum as well, to display the available space in MiB better. Let me sleep on it/play with various options and I'll update PR within 72-96 hours.

systemcrash (approved):

stangri:
@jow- @hnyman please re-review.

stangri:
It's a small enough change I feel I can merge it with reviews from @hnyman and @systemcrash -- thank you gentlemen for quick responses.

@jow-, @McGiverGim thank you for the feedback helping improve this PR!

PR #6899, opened 2024-02-09
luci-app-tor: allow to upload custom config

stokito:
Maintainer: me
CC: @ja-pa

The PR contains a fix for luci-app-tor when list of hidden services didn't worked well.
For a user it just didn't show an onion domain.

Additionally I added a separate tab to configure the Tor itself.
But we have only one option: list of configs to include.
Let's configure at least this.

Additionally I added an upload field to upload a custom config and add it to a list of includes.

luci-app-tor upload custom config

This will allow to a user to change the Tor settings without SSH.
For example by just downloading a sample conf from a tutorial, edit and upload. Still this is not convenient but at least something for today.


systemcrash:
Good - thank you :)

PR #6904, opened 2024-02-11
luci-mod-status: Adding scroll buttons on syslog and kernellog status pages

Ramon00:
My log files can grow rather big, typically if i look at the log files then i want to look at the bottom, not the top, so I was missing a scroll down button. 

So I added these:
image

image

Please let me know if you have any comments or suggestions.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
@@ -21,16 +21,37 @@ return view.extend({
 	render: function(logdata) {
 		var loglines = logdata.trim().split(/\n/);
 
+
+		var scrollDownButton = E('button', { 
+				'id': 'scrollDownButton',
+				'class': 'cbi-button cbi-button-neutral'
+			}, _('Scroll to tail', 'scroll to bottom (the tail) of the log file')
+		);
+		scrollDownButton.addEventListener('click', function() {
+			window.scrollTo(0, document.body.scrollHeight);
+		});
+
+		var scrollUpButton = E('button', { 
+				'id' : 'scrollUpButton',
+				'class': 'cbi-button cbi-button-neutral'
+			}, _('Scroll to head', 'scroll to top (the head) of the log file')
+		);
+		scrollUpButton.addEventListener('click', function() {
+			window.scrollTo(0, 0);
+		});
+
 		return E([], [
 			E('h2', {}, [ _('System Log') ]),
 			E('div', { 'id': 'content_syslog' }, [
+				E('div', {'style': 'padding-bottom: 20px'}, [scrollDownButton]),
 				E('textarea', {
 					'id': 'syslog',
 					'style': 'font-size:12px',
 					'readonly': 'readonly',
 					'wrap': 'off',
 					'rows': loglines.length + 1
-				}, [ loglines.join('\n') ])
+				}, [ loglines.join('\n') ]),
+				E('div', {'style': 'padding-bottom: 20px'}, [scrollUpButton])
 			])
 		]);
 	},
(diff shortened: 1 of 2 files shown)


systemcrash:
Might I suggest Tail and Head, especially for those who might already be
familiar with those utilities that do exactly these things.

Ramon00:
@systemcrash yes i use that, but sometimes I want to just check the log from my phone, then i have to scroll all the way down... Anyway if you think this is not a good addition then fine, just reject this PR, i will just use it on my own openwrt installs, just thought it might be useful for somebody else as well. Not sure though why we would not want to make the LuCi logfile page more useful... Or do you mean I can use tail from LuCi? Please tell me how.

hnyman:
> do you mean I can use tail from LuCi? Please tell me how.

Example of reversing log.
https://forum.openwrt.org/t/status-system-log-new-text-should-be-at-the-top-not-bottom/151740/4

Ramon00:
@systemcrash Which package is that in?

jow-:
The button style looks wrong

Ramon00:
@jow- if you can show me an example what button style you would like, then I can adjust it, np. I just put in the most simple button i could find

Ramon00:
@jow- Like this perhaps: 
image

modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
```diff
@@ -21,16 +21,37 @@ return view.extend({
...
+			}, _('Scroll down')
+		);
+		scrollDownButton.addEventListener('click', function() {
+			window.scrollTo(0, document.body.scrollHeight);
+		});
+
+		var scrollUpButton = E('button', { 
+				'id' : 'scrollUpButton',
+				'class': 'cbi-button cbi-button-neutral'
+			}, _('Scroll up')
```
systemcrash, line 37: You could change instances to just `Tail` - a simpler way of saying more concretely `Scroll to bottom` or `Scroll to tail`, which is more definitive than just `Scroll down`.

Likewise for the other button names. You decide, but put something more concrete.

systemcrash (changes requested):

modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
Ramon00: To understand correctly, you just want to name the button "tail" and "head"? If you think that is more direct then im fine with that.

systemcrash: It's short and easier to translate than three words. :)

Maybe `Scroll to tail` is best - more context for the i18n people.

Ramon00: yeah i can do that, i can also add a translation hint separately.

systemcrash (approved):

systemcrash:
I just tested this, and it seems to work fine. Can you put a bit more padding between scroll to head button and the log field?

Ramon00:
@systemcrash I encapsulated the bottom button now in a div and added 20px padding below the button.  The padding between the button and the logfile itself is in the 'textarea' already, so i cannot really remove it. I can also add paddint to the top button either at above and/or below the button.  Just a matter of what the preference is i suppose

systemcrash:
> @systemcrash I encapsulated the bottom button now in a div and added 20px padding below the button. The padding between the button and the logfile itself is in the 'textarea' already, so i cannot really remove it. I can also add paddint to the top button either at above and/or below the button. Just a matter of what the preference is i suppose

Below the botton at the top, so it's not adjoining the log field

Ramon00:
Ah you didnt mean the scroll to head button, but the scroll to tail button. Something like this? 
image
and 
image

systemcrash:
Exactly so

Ramon00:
ok cool.

systemcrash:
Please rebase onto current master.

Ramon00:
> Please rebase onto current master.

done

systemcrash:
Thank you, nicely done :)

Ramon00:
And thank you for the review and the suggestions :)

PR #6908, opened 2024-02-13
luci-app-usteer: Fix no wireless defined setup

McGiverGim:
Some users use the usteer app in a "non wireless" device. With the current code it crashes because there is no SSID defined. Making the ui element a DynamicList fixes the issue because it allows free text input.

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -609,7 +609,7 @@ return view.extend({
 		o.optional = true;
 		o.datatype = 'list(string)';
 
-		o = s.taboption('settings', form.MultiValue, 'ssid_list', _('SSID list'), _('List of SSIDs to enable steering on'));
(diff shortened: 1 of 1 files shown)


Ramon00:
That seems to fix the "crash". Maybe one nice to have: remove the doubles from the list?

systemcrash:
> That seems to fix the "crash". Maybe one nice to have: remove the doubles from the list?

Can you show an example of this? Just to be sure we aren't actually removing dupe SSIDs, where BSSID may differ.

systemcrash:
Thanks @McGiverGim

Ramon00:
@systemcrash the pulldown menu give each SSID twice (probably once per radio). Also the current implementation means I cannot even select the two occurances of each SSID.

Ramon00:
example
image

McGiverGim:
I will push a new PR tomorrow removing the duplicates. I don't have duplicates between my radios so I didn't notice it. Thanks for testing!

McGiverGim:
Added a new PR @Ramon00 if you want to test it https://github.com/openwrt/luci/pull/6912

PR #6910, opened 2024-02-14
luci-app-adblock-fast: bugfix: dnsmasq_instance & smartdns_instance p…

stangri:
…arsing

better code for parsing dnsmasq_instance & smartdns_instance lists to present the proper options in WebUI.


systemcrash:
Small nitpick - and I'm sure you thought of this - but can't the code be reused as a function instead?

PR #6912, opened 2024-02-14
luci-app-usteer: filter repeated SSIDs in config

McGiverGim:
When the SSID name was the same in different radio, they appear repeated in the options. This commit filter them.

Signed-off-by: Miguel Angel Mulero Martinez

This change was after a discussion in of https://github.com/openwrt/luci/pull/6908

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -611,7 +611,7 @@ return view.extend({
 
 		o = s.taboption('settings', form.DynamicList, 'ssid_list', _('SSID list'), _('List of SSIDs to enable steering on'));
 		WifiNetworks.forEach(function (wifiNetwork) {
-			if (wifiNetwork.getSSID()) {
+			if (wifiNetwork.getSSID() && (!o.keylist || o.keylist.indexOf(wifiNetwork.getSSID()) === -1)) {
 				o.value(wifiNetwork.getSSID())
 			}
 		});


systemcrash:
Could you walk me through what the change does please?

McGiverGim:
Some user like @Ramon00 use the app in a non wireless router. The UI elements crashes when the list of options is empty, and the config file have values. I think it searches the element and when it does not found it, crashes.
For this reason I modified the element to admit custom elements, in this way it does not crash.

McGiverGim:
This UI element contains the list of wireless SSIDs available. When the 2.4 and 5 GHz networks have the same SSID, the name appears twice in the list.
This commit filters them to insert only once each SSID name in the list of SSIDs available.

systemcrash:
Sorry, I wasn't clear. I meant the if conditions now:
```js
			if (wifiNetwork.getSSID() && (!o.keylist || o.keylist.indexOf(wifiNetwork.getSSID()) === -1)) {
				o.value(wifiNetwork.getSSID())
			}
```

I just had to stare at it for another minute to parse it :)

systemcrash:
Does the change work for you, @Ramon00 ?

McGiverGim:
I arrived to the condition after testing and looking into the debugger...
- wifiNetwork.getSSID() -> In the list of wifi networks, they appear some that don't have SSID. In my case, I have a 802.11s mesh network, this is in the list, but this method does not return a SSID, for this reason I verify it.
- !o.keylist -> What I want to check, is if the SSID was added before to the UI element or not. This method verifies that the list of the UI is not empty. I've verified that the list, after initializing the UI, is not an empty array, simply is not initialized, so that is what this checks. 
- o.keylist.indexOf(wifiNetwork.getSSID()) === -1)) -> When the list is not empty, this checks if the SSID was added before or not.

systemcrash:
>  `!o.keylist` -> ... This method verifies that the list of the UI is not empty. I've verified that the list, after initializing the UI, is not an empty array, simply is not initialized, so that is what this checks.

I read the code as:  This method verifies that the list of the UI *is* empty (in which case it is safe to add any SSID, the first), and

> `o.keylist.indexOf(wifiNetwork.getSSID()) === -1))` -> When the list is not empty, this checks if the SSID was added before or not.

I read the code as:  When the list is not empty, this returns true when the SSID is *NOT* already in the list

This is your intention, I guess - because that's what seems to me would be appropriate here.

Ramon00:
I tested it and it for me it works fine.

systemcrash:
Ok. Thank you @McGiverGim and @Ramon00

PR #6919, opened 2024-02-16
luci-mod-system: Improve translation

try496:


jow-:
Unfortunately the term `Auto` is too generic without any context, change the translation call to `_('Auto', 'Language selection')` and update the .pot files so that translators get a hint in what context they're translating "Auto" in.

PR #6928, opened 2024-02-18
luci-theme-bootstrap: Always show scrollbar

thomasschroeder:
With this addition to the body tag, on desktop browsers the vertical scrollbar
is always visible and the content will not be shifted by changing tabs (like
at System/Software when the update tab has no to little content).

I couldn't see any negative side effects with Chromium browsers or Firefox
under Windows/Linux and Android, but I don't have access to Apple devices
for tests.


thomasschroeder:
Screenshot before:
grafik

Screenshot with the change:
grafik

Screenshot with a modal dialog:
grafik

Do you need further information for this PR?

thomasschroeder:
@systemcrash Are there any information missing or is my PR not ok from formal side?

jow-:
Merged, thanks!

Issue #6413, opened 2023-06-03
LuCI does not warn about dangers with upgrading packages


kannes:
## Steps to reproduce:
1. go to: System → Software → Updates

## Actual behavior:
1. Mr. McSmartypants sees huge list of available updates
2. Mr. McSmartypants knows to keep network devices up to date
3. Mr. McSmartypants decides to upgrade all those packages
4. Mr. McSmartypants wonders why there is no "upgrade all" button and scours the web
5. Mr. McSmartypants finds issues like https://github.com/openwrt/luci/issues/4602#issuecomment-729783278
6. Mr. McSmartypants finds https://openwrt.org/meta/infobox/upgrade_packages_warning
7. Mr. McSmartypants needs to reinstall/restore to make sure his router did not get screwed up :(

## Expected behavior:
The page could have a informative message saying what the wiki page says ("Upgrading packages may cause serious problems, including soft-bricking your device!"), maybe with a link to the wiki page https://openwrt.org/meta/infobox/upgrade_packages_warning to prevent dutiful users not to do obvious mistakes.

## Title
I am sorry but I do not know which package this concerns, I will update the title when someone tells me :)

systemcrash:
Closed via https://github.com/openwrt/luci/commit/4d23adc8d938214475f9c68b3c4c7c5a55c29743

kannes:
Awesome, thank you so much! This will make LuCI much safer to use for newbies.

PR #6932, opened 2024-02-19
luci-app-opkg: add warning about modifying packages

dannil:
A proposal for a fix for #6413 

It is a well-known fact that opkg doesn't check ABI-compatibility between dependencies and upgrading packages in a stable release is therefore not recommended. In addition, installing and removing packages blindly may result in rendering the device inaccessible as well, so caution should be taken when doing this.

This commit adds a warning which should help notify especially new users about these behaviors with a link to the wiki which advises to not upgrade packages.

See https://openwrt.org/meta/infobox/upgrade_packages_warning for more info, which is the same link that is used in the second paragraph.

Tested on Chrome and Firefox, both desktop and mobile (Android). All CSS used is theme-compatible and should therefore look at least decent across different themes.

Screenshot_44

Screenshot_45

<details>
  <summary>First iteration of design, not relevant anymore</summary>

BootstrapLight:
Screenshot_39

BootstrapDark:
Screenshot_40

Material:
Screenshot_41

OpenWrt2020:
Screenshot_42
</summary>

### DIFF
--- a/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
+++ b/applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
@@ -1129,6 +1129,14 @@ return view.extend({
 
 			E('h2', {}, _('Software')),
 
+			E('div', { 'class': 'cbi-map-descr' }, [
+				E('span', _('Install additional software and upgrade existing packages with opkg.')),
+				E('br'),
+				E('span', _('<strong>Warning!</strong> Package operations can <a %s>break your system</a>.').format(
+					'href="https://openwrt.org/meta/infobox/upgrade_packages_warning" target="_blank" rel="noreferrer"'
+				))
+			]),
+
 			E('div', { 'class': 'controls' }, [
 				E('div', {}, [
 					E('label', {}, _('Disk space') + ':'),
--- a/applications/luci-app-opkg/po/ar/opkg.po
+++ b/applications/luci-app-opkg/po/ar/opkg.po
@@ -15,7 +15,12 @@ msgstr ""
 msgid "%s used (%1024mB used of %1024mB, %1024mB free)"
 msgstr ""
 
-#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1155
+#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1135
+msgid ""
+"<strong>Warning!</strong> Package operations can <a %s>break your system</a>."
+msgstr ""
+
+#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1163
 msgid "Actions"
 msgstr "إجراءات"
 
@@ -27,7 +32,7 @@ msgstr ""
 msgid "Automatically remove unused dependencies"
 msgstr "قم بإزالة التبعيات غير المستخدمة تلقائيًا"
 
-#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1210
+#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1218
 msgid "Available"
 msgstr "متاح"
 
@@ -51,11 +56,11 @@ msgstr ""
 msgid "Cancel"
 msgstr "إلغاء"
 
-#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1142
+#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1150
 msgid "Clear"
 msgstr "إجلاء"
 
-#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1159
+#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1167
 msgid "Configure opkg…"
 msgstr "تكوين opkg …"
 
@@ -65,7 +70,7 @@ msgstr "التبعيات"
 
 #: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:750
 #: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:938
-#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1228
+#: applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js:1236
 msgid "Description"
 msgstr "الوصف"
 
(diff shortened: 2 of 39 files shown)


systemcrash:
Not optimal if it causes a pop-up or a button that must be clicked away.

dannil:
> Luci has UI notification API that can be used like:
> 
> ```
>  ui.addNotification(null, E('p', 'Werning!'), 'error');
> ```
> 
> Maybe you can use it instead?

Having a pop-up show up every time you navigate to the updates tab will most likely be very annoying to the user.

jow-:
Please also add warning banners that *installing* packages and *removing* packages can cause soft bricks or break access to the ui.

systemcrash:
Shall we add more banners and warnings that using openwrt will also
potentially soft brick their devices? 😉


Please also add warning banners that *installing* packages and *removing*
> packages can cause soft bricks or break access to the ui.
>

dannil:
I've updated the PR title, commit message, and screenshots to reflect the suggestion by @jow-

dannil:
@stokito @systemcrash @jow- is there any other changes we want or does it make sense to start thinking of merging it?

applications/luci-app-opkg/htdocs/luci-static/resources/view/opkg.js
```diff
@@ -1206,6 +1206,13 @@ return view.extend({
 				])
 			]),
 
+			E('div', { 'class': 'alert-message warning' }, [
+				E('p', _('Warning! Installing, removing, and upgrading packages may cause serious problems, including soft-bricking your device!')),
+				E('p', _('See <a %s">explanation on wiki</a> why especially upgrading should be avoided.').format(
```
stokito, line 1211: Replace `<a %s">` with `<a %s>`

stokito (changes requested):

systemcrash:
Is it possible to reduce the warning text to a concise nugget? That page is now exploding with lines, colours and text and this does not make it easier to handle.  How about:
```
Warning! Package operations can break your system.')),
				E('p', _('Read <a %s>this</a>.').format(
```

dannil:
> Is it possible to reduce the warning text to a concise nugget? That page is now exploding with lines, colours and text and this does not make it easier to handle. How about:
> 
> ```
> Warning! Package operations can break your system.')),
> 				E('p', _('Read <a %s>this</a>.').format(
> ```

Absolutely, I like that idea.

stokito (approved):

systemcrash:
And now... to acknowledge this big warning dialogue. 

@yurtpage suggested

> Luci has UI notification API that can be used like:
> 
> ```
>  ui.addNotification(null, E('p', 'Werning!'), 'error');
> ```
> 
> Maybe you can use it instead?

But maybe something more resilient. Like an 'acknowledge' button, that when clicked, stores a boolean flag in the config that determines whether the warning is shown at page load. We've done our diligence, the user can hide the warning. 

What do you think @dannil ?

dannil:
> And now... to acknowledge this big warning dialogue.
> 
> @yurtpage suggested
> 
> > Luci has UI notification API that can be used like:
> > ```
> >  ui.addNotification(null, E('p', 'Werning!'), 'error');
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe you can use it instead?
> 
> But maybe something more resilient. Like an 'acknowledge' button, that when clicked, stores a boolean flag in the config that determines whether the warning is shown at page load. We've done our diligence, the user can hide the warning.
> 
> What do you think @dannil ?

Yes, I've actually thought about that too. I think we could do something akin to how the "Confirm" button behaves in warning dialogs (such as when you don't have a password set on your root user), here's an example from nftables:

Screenshot_43

Do we think it would be necessary to have a checkbox somewhere else (like under the "Configure opkg..." dialog) that would enable users to show the warning again? I personally think it shouldn't be needed.

stangri:
1. I believe a nicer way would be: `Warning: package operations can break your system (more information).` and link the wiki article from `more information`?
2. If not, is it possible to move the `Read this` text to the line before and avoid the line break?

stokito:
Honestly I think it would be better to make same introduction paragraph as other pages have:

luci-app-opkg


```html
<div class="cbi-map-descr">
Install additional software and upgrade existing packages with opkg.. 
<br><b>Warning!</b> Package operations <a href="href="https://openwrt.org/meta/infobox/upgrade_packages_warning" target="_blank" rel="noreferrer">can break your system</a>
</div>
```

I.e. the warning can be just a sentence in the introduction.

We may add more important notes e.g. about feeds and disk space usage.
Given that this UI messages are more likely to be translated than a Wiki page it will improve cautious of users. That's why we need to add at least a few words that ABI may broken and that core modules shouldn't be updated at all. Yes, wiki explains this better but who reads it?

BTW the article in Wiki is in inbox e.g. in "draft" state. On the wiki itself not so much of information about package management https://openwrt.org/docs/guide-user/additional-software/start

systemcrash:
> Honestly I think it would be better to make same introduction paragraph as other pages have:


Also good. Less intrusive yet ever present.

dannil:
> Honestly I think it would be better to make same introduction paragraph as other pages have:
> 
> luci-app-opkg
> 
> ```
> <div class="cbi-map-descr">
> Install additional software and upgrade existing packages with opkg.. 
> <br><b>Warning!</b> Package operations <a href="href="https://openwrt.org/meta/infobox/upgrade_packages_warning" target="_blank" rel="noreferrer">can break your system</a>
> </div>
> ```
> 
> I.e. the warning can be just a sentence in the introduction.
> 
> We may add more important notes e.g. about feeds and disk space usage. Given that this UI messages are more likely to be translated than a Wiki page it will improve cautious of users. That's why we need to add at least a few words that ABI may broken and that core modules shouldn't be updated at all. Yes, wiki explains this better but who reads it?
> 
> BTW the article in Wiki is in inbox e.g. in "draft" state. On the wiki itself not so much of information about package management https://openwrt.org/docs/guide-user/additional-software/start

I like this as well, especially the less intrusive part. Also, I've had trouble to implement a UCI config for opkg (since the package itself doesn't define this I need to do some hacky solution with creating the file from the LuCI app, which I don't like) which can entirely be avoided with the above suggestion. 

I'll implement this solution in the following days since it's both much less intrusive and way easier technically to implement since we don't need the while "Hide" logic.

stokito (approved):

systemcrash:
Merged. Thanks @dannil !

PR #6936, opened 2024-02-21
luci-theme-material: improve theming system colors and contrasts

dannil:
This changed spawned from trying to make certain text colors readable (such as white text on white background in many dialogs) introduced by commit 4ee2bc6 (ref: #5841) which made obvious the need to improve the color and contrast situation in the theme.

Alot of colors were duplicated in cascade.css and made for a hard time to align colors across different elements. This commit tries to rectify that by introducing variables for all commonly used colors.

* All base colors (white, blue, red, green, blue, yellow, etc) has been consolidated and moved to common variables
* Introduced more specific selectors for info levels to avoid colors bleeding over to other elements
* Removed duplicated properties which were overriden at the next row

Before:
Screenshot_70
Screenshot_71
Screenshot_74

After:
Screenshot_1
Screenshot_2
Screenshot_3

The slight change in the red color for the danger message is intended to align it with the other red colors which was already in use in buttons styled as dangerous (such as `Perform reset` in `Backup / Flash Firmware`). 

I haven't found any regressions in the modules I've tested these changes work against but any help in validating these changes are appreciated.

### DIFF
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -181,7 +181,7 @@ select {
 	padding: .36rem .8rem;
 	color: #555;
 	border: thin solid #ccc;
-	background-color: #fff;
+	background-color: var(--white-color);
 	background-image: none;
 }
 
@@ -212,8 +212,7 @@ select,
 select:not([multiple="multiple"]):focus,
 input:not(.cbi-button):focus,
 .cbi-dropdown:focus {
-	border-color: #09c;
-	border-color: var(--main-color);
+	border-color: var(--primary-color);
 }
 
 .cbi-dropdown,
@@ -345,12 +344,9 @@ small {
 	top: 4rem;
 	float: left;
 	overflow-x: auto;
-	width: 15%;
 	width: calc(0% + 15rem);
-	height: 100%;
 	height: calc(100% - 4rem);
-	background-color: #fff;
-	background-color: var(--menu-bg-color);
+	background-color: var(--white-color);
 	transition: visibility 400ms, width 400ms;
 }
 
@@ -383,10 +379,8 @@ small {
 }
 
 header {
-	color: #fff;
-	color: var(--header-color);
-	background: #09c;
-	background: var(--header-bg);
+	color: var(--secondary-color);
+	background: var(--primary-color);
 }
 
 header > .fill > .container {
@@ -413,8 +407,7 @@ header > .fill > .container > .brand {
 	cursor: default;
 	vertical-align: text-bottom;
 	text-decoration: none;
-	color: #fff;
-	color: var(--header-color);
+	color: var(--white-color);
 }
 
 header > .fill > .container > .status {
@@ -436,10 +429,8 @@ header > .fill > .container > .status > * {
 	display: flex;
 	flex-wrap: wrap;
 	align-items: center;
-	color: #fff;
-	color: var(--header-color);
-	background: #09c;
-	background: var(--header-bg);
+	color: var(--white-color);
+	background: var(--secondary-color);
 	padding: .5rem;
 	transition: box-shadow .2s;
 	box-shadow: 0 2px 5px rgb(0 0 0 / 26%);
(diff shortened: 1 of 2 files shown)


dannil:
I've noticed some regressions so put it in draft.

stokito:
@dannil  Maybe you can also take a look on the old #3542

dannil:
@stokito sure thing, I can give it a go!

systemcrash:
Can we get an overview of how these changes look in dark mode? :)

dannil:
> Can we get an overview of how these changes look in dark mode? :)

Unfortunately Material doesn't have a dark mode so there's nothing to show :(

systemcrash:
Thanks @dannil

DittelHome:
The new updated theme has some problems ...
Network => Interface => edit:
grafik

System => Custom Commands:
grafik

systemcrash:
@dannil perhaps there are some dynamically created classes which no longer match a CSS tag?

@DittelHome open the browser "inspect" menu on those items, and compare its classes to the classes of enclosing elements. Can we get a screenshot of that?

systemcrash:
Nice, if slightly confusing, effect.

systemcrash:
Exactly. Can you open it side-by-side with the weird display elements?

DittelHome:
OK, sorry, it was a caching problem. I have deleted the cache (firefox), now it looks ok.
The only thing that isn't pretty is: System => Custom Commands

grafik

But that has been the case for a long time...

dannil:
@DittelHome thanks for reporting it anyway. Yes, I've noticed the custom commands one before and that's been there for a long time, I've been thinking of fixing it tho so may be updated in the future.

feckert:
@dannil @systemcrash I have rebased my branches and I have noticed some strange coloring of my customized luci-theme-material. After a git search it was due to this change. **All** names for the color have changed. So my customization of the luci-theme-material does not get applied, because all names of the my colors have changed.

In principle, I think it's good that all the colors that are used could be configured via variables. 

@systemcrash But I am not completely happy with this merge! It breaks the whole concept of this theme!

**The idea behind this theme is that the colors could be customized. So we would could change the color for the menu, hover, submenu or header**

@dannil  Therefore, the colors should be named after the function and not after the color.
As was previously the case. You already did this in this merge with the following colors

- var(--notice-color);
- var(--on-notice-color);
- var(--danger-color);
- var(--on-danger-color);
- var(--warning-color);
- var(--on-warning-color);
- var(--success-color);
- var(--on-success-color);
- var(--on-error-color);
- var(--error-color);

I would therefore ask you to revise it.

systemcrash:
Thanks for the input @feckert. Where are these customisations of yours? Are they in a public repo?
Also, where does one customise? In code as done here in this PR?

feckert:
@systemcrash 

> Where are these customisations of yours?

I custom the `custom.css` by replacing this with my own one.

> Are they in a public repo?

No. But that is the theme made for. Therefor the name of the file `custom.css`.

> Also, where does one customise?

In the `custom.css` file. It looks the same but with different definitions of the used colors.

systemcrash:
> @systemcrash
> 
> > Where are these customisations of yours?
> 
> I custom the `custom.css` by replacing this with my own one.
> 
> > Are they in a public repo?
> 
> No. But that is the theme made for. Therefor the file is named `custom.css`.

Are you saying that `custom.css` is only for local customisation? And all mods should go in the other .css file, and custom is to override those?

> > Also, where does one customise?
> 
> In the `custom.css` file. Till the change the file looks the same but with different definitions of the used colors. By replacing the `custom.css` with my `custom.css` I could define my own color theme.

Thanks. I see. I thought it might have been a bit more straightforward, like some edit box somewhere. ( Future idea, @dannil ? )

feckert:
@systemcrash 
> > No. But that is the theme made for. Therefor the file is named `custom.css`.
> 
> Are you saying that `custom.css` is only for local customisation? And all mods should go in the other .css file, and custom is to override those?

No, it's fine that the colors are stored in `custom.css` file.  I simply replaced the file with scp with mine. But now I can no longer use my customized color theme `custom.css` file because the variable names no longer exist

> 
> > > Also, where does one customise?
> > 
> > 
> > In the `custom.css` file. Till the change the file looks the same but with different definitions of the used colors. By replacing the `custom.css` with my `custom.css` I could define my own color theme.
> 
> Thanks. I see. I thought it might have been a bit more straightforward, like some edit box somewhere. ( Future idea, @dannil ? )

That would be nice to have. But for now, it would be enough to switch back to variable naming that indicate to which feature the color belongs. As you already have done.

- var(--notice-color);
- var(--on-notice-color);
- var(--danger-color); ....

systemcrash:
> That would be nice to have. But for now, it would be enough to switch back to variable naming that indicate to which feature the color belongs. As you already have done.

While we're here. Where might a good place for this be? System -> Admin? Or e.g. a new System -> Themes?

Better to just provide a single text edit box? Or have more 'guard rails' and colour pickers for the individual elements? 

I guess the pickers would mandate that all themes comply with it ( Maybe there's a smart way to parse these things from an active theme. ), rather than a text edit box so any css can comply with any theme. The latter seems the path of least resistance.

dannil:
> **The idea behind this theme is that the colors could be customized. So we would could change the color for the menu, hover, submenu or header**

I was not aware of this, is this documented anywhere that this was a supported use case and if not, should we document this in the CSS itself so it doesn't happen again? I would never have guessed that the CSS variables was a stable API in and of itself, which is why I did the changes. Irregardless, if this was the purpose then I definitely agree we should add it back; it should be quite trivial to do this to the removed functionality color variables (main-color, header-bg, header-color, etc) so your rebase would behave as before, I think I could have this done by today or tomorrow.

systemcrash:
Yeah, jam a cogent comment in there. I would not have guessed either...

dannil:
> > 
> 
> No, it's fine that the colors are stored in `custom.css` file. I simply replaced the file with scp with mine. But now I can no longer use my customized color theme `custom.css` file because the variable names no longer exist

Would this work with the added variables in https://github.com/openwrt/luci/pull/6936/commits/2bdabf14c30320ef0433828b1a080f6b8fe9c430 as well? I imagine if the workflow was just to SCP over the custom.css file with your own derivative then for example `--success-color` would be lost and the element specifying that color would just inherit the parent(s) element background color, or is your `custom.css` a merge between the one provided by the theme and your own colors? 

If the first case, how was the plan to handle the addition of colors in a backwards-compatible way while avoiding to hardcode colors in `cascade.css`? Just trying to figure out the exact flow that was in mind to assure future compatibility.

PR #6943, opened 2024-02-27
luci-app-cloudflared: Overhaul tunnels.js

animegasan:
I overhauled tunnels.js to make it more organized. And upgraded it to version 1.2

Some changes are:
1. Make data from tunnel into a table
2. Add a class for the table structure
3. Wrap each table with a class section

### Before
image

### After
image


systemcrash:
OK. Thanks. 

In future avoid PR from your master branch - it gets messy when you need to rebase on the current master.

hnyman:
This contains obvious duplicates (at least one). Weblate complaints about that.
I will remove them.
The new translation keys were not added into the proper places alphabetically, so let's fix also that...

Example duplicate, added just before the already existing identical msgid:
```
+#: applications/luci-app-cloudflared/htdocs/luci-static/resources/view/cloudflared/tunnels.js:54
+msgid "Connections"
+msgstr ""
+
+#: applications/luci-app-cloudflared/htdocs/luci-static/resources/view/cloudflared/tunnels.js:90
 msgid "Connections"
 msgstr ""
```

PR #6945, opened 2024-02-28
luci-app-ntpc: remove app

dannil:
As ntpclient package is dead, we can delete the app.

See https://github.com/openwrt/luci/pull/6945#issuecomment-1973923917 and https://github.com/openwrt/packages/pull/23569

~~Converted the app to use client-side JavaScript instead of Lua. The only real change to the previous implementation is I changed the `Current system time` field to match the format in `Status -> System -> Local Time` as I could not get the `%c` date format to work with the format function; if this is possible feel free to comment this. Package doesn't have a PKG_VERSION so by definition it hasn't been incremented.~~

Screenshot_9

- [X] Tested on: (x86/generic, OpenWrt 23.05.2, Chrome and Firefox) :white_check_mark:


systemcrash:
Also, would it be interesting to try and calculate the latency between browser and openwrt and attempt to compensate for the offset? Code golf perhaps 😎 

Entirely optional.

dannil:
> Great! I didn't even notice we had this app.
> 
> Could you hint the host name and port fields for when they're empty?
> 
> Maybe also add in the package description that this is a client.

That sounds like good additions, I will add that.

> Also, would it be interesting to try and calculate the latency between browser and openwrt and attempt to compensate for the offset? Code golf perhaps 😎
> 
> Entirely optional.

Haha, I can give it a shot, but I don't promise anything, if anything I can experiment with it and open a PR later on for it.

dannil:
@systemcrash I've updated the code with your suggestions, do you think we need a better description than that or should it suffice?

applications/luci-app-ntpc/Makefile
```diff
@@ -7,7 +7,8 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=NTP time synchronisation configuration module
-LUCI_DEPENDS:=+luci-base +luci-compat +ntpclient
+LUCI_DESCRIPTION:=LuCI support for configuring the ntpclient package
```
dannil, line 10: @systemcrash do you think we need a better description than that or should it suffice?

systemcrash: Golden. Done I think.

systemcrash:
One question: the local Date() clock monotonically increases with time?

dannil:
> One question: the local Date() clock monotonically increases with time?

No, once you enter the app, the clock fetches the current timestamp and displays that, with no auto-refresh or anything fancy like that. That is also how the previous implementation in Lua behaved so I didn't change that, but it wouldn't be hard to implement this if we wanted, like how the overview page auto-refreshes the Local Time every 5 seconds.

systemcrash:
OK - I'm trying to reconcile what System -> Time Sync does not do. With such similar data and function, perhaps the two could be merged if they are not already the same.

dannil:
> OK - I'm trying to reconcile what System -> Time Sync does not do. With such similar data and function, perhaps the two could be merged if they are not already the same.

I think the biggest difference is that System -> Time Sync is built to support [ntpd](https://openwrt.org/docs/guide-user/advanced/ntp_configuration) while this app was built to support [ntpclient](https://github.com/openwrt/packages/blob/master/net/ntpclient/Makefile), so consolidating them would be quite a challenge of the top of my head as we would need to adapt `luci-mod-system` to support configuring multiple NTP daemons. 

We could of course think of adding UI options of features that ntpclient has that ntpd doesn't _but_ that it currently supports but I think that's entirely separate PR:s to do that work.

systemcrash (approved):

dannil:
Nice find, I didn't even know that the project was dead. I'm not opposed to just deleting the whole package, I don't know what the processes are regarding that if there's some users that are currently relying on it?

> The existing system can make calls to other ntpd sources to synchronise, which is basically what ntpclient does, right?

I'm not entirely sure but yes I think so, I created this PR purely to port it to JavaScript and what I did was to check that the UCI config values that the Lua implementation wrote to was still updated correctly after the conversion.

```
$ uci show ntpclient
ntpclient.@ntpserver[0]=ntpserver
ntpclient.@ntpserver[0].hostname='0.openwrt.pool.ntp.org'
ntpclient.@ntpserver[0].port='123'
ntpclient.@ntpserver[1]=ntpserver
ntpclient.@ntpserver[1].hostname='1.openwrt.pool.ntp.org'
ntpclient.@ntpserver[1].port='123'
ntpclient.@ntpserver[2]=ntpserver
ntpclient.@ntpserver[2].hostname='2.openwrt.pool.ntp.org'
ntpclient.@ntpserver[2].port='123'
ntpclient.@ntpserver[3]=ntpserver
ntpclient.@ntpserver[3].hostname='3.openwrt.pool.ntp.org'
ntpclient.@ntpserver[3].port='123'
ntpclient.@ntpdrift[0]=ntpdrift
ntpclient.@ntpdrift[0].freq='0'
ntpclient.@ntpclient[0]=ntpclient
ntpclient.@ntpclient[0].interval='600'
```

systemcrash:
> Nice find, I didn't even know that the project was dead. I'm not opposed to just deleting the whole package, I don't know what the processes are regarding that if there's some users that are currently relying on it?

Not sure how anyone can rely on a package which has no sources any longer. We can check back in a few days and see whether the host comes back. 

I've made a PR otherwise here openwrt/packages#23569

applications/luci-app-ntpc/po/sk/ntpc.po
```diff
@@ -12,66 +12,57 @@ msgstr ""
...
+#: applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js:71
+#: applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js:74
 msgid "Port"
 msgstr "Port"
 
-#: applications/luci-app-ntpc/luasrc/model/cbi/ntpc/ntpc.lua:5
-#: applications/luci-app-ntpc/luasrc/model/cbi/ntpc/ntpcmini.lua:6
-msgid "Synchronizes the system time"
+#: applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js:24
+msgid "Synchronizes the system time."
```
stokito, line 50: Please remove the dot at end and this will keep a translation

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
```diff
@@ -0,0 +1,78 @@
...
+'require view';
+
+var callSystemInfo = rpc.declare({
+	object: 'system',
+	method: 'info'
+});
+
+return view.extend({
+	load: function () {
+		return Promise.all([
```
stokito, line 14: Here you have only one promise so you can remove the Promise.all()

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
```diff
@@ -0,0 +1,78 @@
...
+			L.resolveDefault(callSystemInfo(), {})
+		]);
+	},
+
+	render: function (data) {
+		var system_info = data[0];
+
+		var m, s1, time, interval, count, freq, s2, s3, hostname, port;
```
stokito, line 22: Usually for section and option separate vars aren't created.

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
```diff
@@ -0,0 +1,78 @@
...
+
+		m = new form.Map('ntpclient', _('Time Synchronisation'), _('Synchronizes the system time.'));
+
+		s1 = m.section(form.TypedSection, 'ntpclient', _('General'));
```
stokito, line 26: So instead of s1 use:
`s = m.section(form.TypedSection, 'ntpclient', _('General'));`

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
```diff
@@ -0,0 +1,78 @@
...
+				date.getUTCFullYear(),
+				date.getUTCMonth() + 1,
+				date.getUTCDate(),
+				date.getUTCHours(),
+				date.getUTCMinutes(),
+				date.getUTCSeconds()
+			);
+		}
+
+		time = s1.option(form.DummyValue, '_time', _('Current system time'));
```
stokito, line 45: `o = s.option(form.DummyValue, '_time', _('Current system time'));`

stokito (approved):
Nice work.
Please become a maintainer and add PKG_MAINTAINER

dannil:
> Nice work. Please become a maintainer and add PKG_MAINTAINER

@systemcrash has a PR open in https://github.com/openwrt/packages/pull/23569 to drop support for ntpclient since the sources seems to be dead (https://github.com/openwrt/luci/pull/6945#issuecomment-1973863400), so we may drop the app entirely

applications/luci-app-ntpc/po/sk/ntpc.po
dannil: Fixed

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
dannil: Fixed

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
dannil: Fixed

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
dannil: Fixed

applications/luci-app-ntpc/htdocs/luci-static/resources/view/ntpc.js
dannil: Fixed

stokito:
Interesting, I found that Luci already has NTP config in a System page if the sysntpd from BusyBox was enabled during a compilation.

luci NTP servers configuration
 
It's settings are stored to /etc/config/system:

    config timeserver 'ntp'
	    option enabled '1'
	    option enable_server '0'
	    list server '0.openwrt.pool.ntp.org'
	    list server '1.openwrt.pool.ntp.org'

So yes, probably the feature may be removed.
Still it may be used in some old firmware of some vendors. So maybe the app will be useful for them.
When removing the app please tag the issue so they may find the PR and maybe use it once last time.

hnyman:
Busybox ntp has been the default server and client at least for a decade, and there has been config for that in System settings (luci-mod-system) all the time.

So, I really wonder if somebody is actively using this. This app has not been modified for some 10 years...
@dannil 
you got interested in this. do you actually use it?

dannil:
> Busybox ntp has been the default server and client at least for a decade, and there has been config for that in System settings (luci-mod-system) all the time.
> 
> So, I really wonder if somebody is actively using this. This app has not been modified for some 10 years... @dannil you got interested in this. do you actually use it?

No, I basically picked this as a "test platform" to get into developing in LuCI (the whole "we don't develop apps in Lua anymore" seemed a good issue to start with). I'm fine to drop support for the whole app it if we conclude that the ntp functionality in System settings fulfill the same goals.

systemcrash:
OK, we can axe ntpc. 

@dannil please truncate this PR just to removal.

BKPepe:
It needs to be backported to OpenWrt 23.05.

dannil:
I've opened #6970 to backport it.

PR #6947, opened 2024-03-01
luci-app-speedtest: Add new package

animegasan:
# LuCI for Speedtest

image

I feel this is very helpful for those of us who don't want the hassle of going to the terminal just to check the internet speed of OpenWrt devices. Especially for those who are remote accessing from outside the network.


systemcrash:
I’m going to close this because you have opened a pr from your master branch. Don’t do that. Otherwise the idea is good. Please open a new pr with your feature on a separate branch.

PR #6948, opened 2024-03-01
luci-app-speedtest: add new package

animegasan:
# LuCI for Speedtest

image

I feel this is very helpful for those of us who don't want the hassle of going to the terminal just to check the internet speed of OpenWrt devices. Especially for those who are remote accessing from outside the network.

#6947


systemcrash:
Please address the test formalities

PR #6953, opened 2024-03-02
luci-base: use correct regex for time validation

dannil:
Before this change, values further in time than 23:59:59 was allowed, such as 24:00:00 and 23:60:00. Leap seconds is accounted for so 60 is allowed in the seconds parameter.

First reported in https://github.com/openwrt/luci/issues/6965 and https://forum.openwrt.org/t/openwrt-23-05-2-service-release/177709/303

Screenshot_29
Screenshot_28
Screenshot_21
Screenshot_22

After fix:
Screenshot_30

- [ ] Incremented :up: any `PKG_VERSION` in the Makefile (package doesn't have one, not applicable)
- [X] Tested on: (x86/generic, OpenWrt 23.05.2, Chrome and Firefox) :white_check_mark:
- [X] \( Preferred ) Mention: @ the original code author for feedback @jow- 

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/validation.js
+++ b/modules/luci-base/htdocs/luci-static/resources/validation.js
@@ -563,7 +563,7 @@ var ValidatorFactory = baseclass.extend({
 		},
 
 		timehhmmss: function() {
-			return this.assert(this.value.match(/^[0-6][0-9]:[0-6][0-9]:[0-6][0-9]$/),
+			return this.assert(this.value.match(/^(?:[01]\d|2[0-3]):[0-5]\d:(?:[0-5]\d|60)$/),
 				_('valid time (HH:MM:SS)'));
 		},
 


hnyman:
Your fix is wrong.
It disables entering e.g. 18:00:00, as the second digit in first number can only be 0-3:
`^[0-2][0-3]`

That part of the regex needs further tweaking.

Likely have `(0-1)(0-9) or 2(0-3)` or something similar for that number.

dannil:
> Your fix is wrong. It disables entering e.g. 18:00:00, as the second digit in first number can only be 0-3: `^[0-2][0-3]`
> 
> That part of the regex needs further tweaking.
> 
> Likely have `(0-1)(0-9) or 2(0-3)` or something similar for that number.

Ah of course, thanks, totally missed that case. Will update it.

dannil:
I added a capture group for the hour parameter, so now it allows either 00-19 or 20-23 but no values outside of that range

Also; regex is hard.

dannil:
I just discovered that the capture groups doesn't work in Lua since it's actually pattern matching, not regular expressions, being used in the match function. The only place the `timehhmmss` validator is used today is in `luci-app-firewall` and doesn't rely on luci-compat, so I've dropped my change for `datatypes.lua` since new apps won't be developed in Lua anyway. I don't think it's worth spending time on getting it to work for the Lua code.

systemcrash:
> Before this change, values "further in time" than 23:59:59 was allowed, such as 24:00:00, 23:60:00 and 23:00:60

Not only can regex be difficult but time is as well. So I’m going to mention a few frequently encountered problems as we humans attempt to match civilian time with solar time: 

- Leap seconds make …:60 happen. I’ve seen them. They’re real. 
- Leap minutes are theoretically possible making …:60:.. happen.
- During DST changes, clocks in some time zones can contain 25 hours. So the above mentioned “further in time” values are actually… correct. Poke around in some tzinfo files and their tools and you’ll quickly understand what I mean. 

Yes they’re seldom encountered but they’re necessary.

Would you like some linked references or is the above explanation sufficient? 

Time keeping is not a boring subject. 🤓

systemcrash:
regex nitpick - will JS handle a non-capturing group? `(?:...)` Just code-golf at this point.

dannil:
> > Before this change, values "further in time" than 23:59:59 was allowed, such as 24:00:00, 23:60:00 and 23:00:60
> 
> Not only can regex be difficult but time is as well. So I’m going to mention a few frequently encountered problems as we humans attempt to match civilian time with solar time:
> 
>     * Leap seconds make …:60 happen. I’ve seen them. They’re real.
> 
>     * Leap minutes are theoretically possible making …:60:.. happen.
> 
>     * During DST changes, clocks in some time zones can contain 25 hours. So the above mentioned “further in time” values are actually… correct. Poke around in some tzinfo files and their tools and you’ll quickly understand what I mean.
> 
> 
> Yes they’re seldom encountered but they’re necessary.
> 
> Would you like some linked references or is the above explanation sufficient?
> 
> Time keeping is not a boring subject. 🤓

I'm actually aware of all of this, however since the `timehhmmss` validator doesn't contain any more information other than time (we would need date and time zones as input to the validator to accurately determine if for example the leap seconds are valid) this isn't possible to do right now unless the it's rewritten.

systemcrash:
> we would need date and time zones as input to the validator to accurately determine if for example the leap seconds are valid

Disregard this complexity. We just need to 'generally' account for those possibilities - 'good enough'. Trying to cover every eventuality can make you go insane. 

```
* tzh_typecnt ttinfo entries, each defined as follows:

              struct ttinfo {
                   int32_t        tt_utoff;
                   unsigned char  tt_isdst;
                   unsigned char  tt_desigidx;
              };

         ... Also, in
         realistic applications tt_utoff is in the range [-89999, 93599]
         (i.e., more than -25 hours and less than 26 hours); this allows
         easy support by implementations that already support the POSIX-
         required range [-24:59:59, 25:59:59].
```

and

```
       When a positive leap second occurs, readers should append an
       extra second to the local minute containing the second just
       before the leap second.  If this occurs when the UTC offset is
       not a multiple of 60 seconds, the leap second occurs earlier than
       the last second of the local minute and the minute's remaining
       local seconds are numbered through 60 instead of the usual 59;
       the UTC offset is unaffected.
```

So if an event is scheduled to occur on a (just before) midnight leapsecond, but we cannot enter it, it will never be encountered to be triggered.

hnyman:
Let's not overly complicate things for really rare corner-case occurrences. 
Good enough validation is the goal.

modules/luci-base/htdocs/luci-static/resources/validation.js
```diff
@@ -563,7 +563,7 @@ var ValidatorFactory = baseclass.extend({
 		},
 
 		timehhmmss: function() {
-			return this.assert(this.value.match(/^[0-6][0-9]:[0-6][0-9]:[0-6][0-9]$/),
+			return this.assert(this.value.match(/^(?:[01]\d|2[0-3]):[0-5]\d:[0-5]\d$/),
```
systemcrash, line 566: This should give us what we need
```js
			return this.assert(this.value.match(^(?:[01]\d|2[0-4]):[0-5]\d:(?:[0-5]\d|60)$),
```

systemcrash (changes requested):

brada4:
@systemcrash youre right, seconds is 00-60
ie 0..23:0..59:0..**60**
```
root@suse:~# nft -c -f -
table inet test {
  chain test {
    meta hour "23:59:60"
}
}^D
root@suse:~# nft -c -f -
table inet test {
  chain test {
    meta hour "23:60:59"
}
}^D
/dev/stdin:3:15-24: Error: wrong time format
    meta hour "23:60:59"
              ^^^^^^^^^^
root@suse:~# nft -c -f -
table inet test {
  chain test {
    meta hour "23:58:60"
}
}^D
root@suse:~#

24:00:00 is invalid too
```

systemcrash:
> @systemcrash youre right, seconds is 00-60 ie 0..23:0..59:0..**60**
> 
> ```
> root@suse:~# nft -c -f -
> table inet test {
>   chain test {
>     meta hour "23:59:60"
> }
> }^D
> root@suse:~# nft -c -f -
> table inet test {
>   chain test {
>     meta hour "23:60:59"
> }
> }^D
> /dev/stdin:3:15-24: Error: wrong time format
>     meta hour "23:60:59"
>               ^^^^^^^^^^
> root@suse:~# nft -c -f -
> table inet test {
>   chain test {
>     meta hour "23:58:60"
> }
> }^D
> root@suse:~#
> 
> 24:00:00 is invalid too
> ```

Buggy OSes :D  

OK - we'll go with just leap seconds for now.

modules/luci-base/htdocs/luci-static/resources/validation.js
brada4: Confirm ;-)

brada4:
Now frontend validation 1:1 reflects backend validation, including leap second, no chance planting invalid automatically disabled rule.

modules/luci-base/htdocs/luci-static/resources/validation.js
dannil: I've updated the regex with the suggestion, thanks!

systemcrash (approved):

systemcrash:
Nicely done. Thanks!

PR #6957, opened 2024-03-02
luci-mod-network: wireless.js: implement method to set Operating Channel Validation

huaracheguarache:
Setting Operating Channel Validation has long been possible in the wireless config file with the `ocv` parameter, and this pull request implements a method to set it in luci. It is done by adding a drop-down menu below the 802.11w settings in the wireless security tab as seen in the screenshot below:

Screenshot from 2024-03-02 23-02-13

OCV can only be used in conjunction with 802.11w, so I've added a dependency check that makes sure that 802.11w is either set to optional or mandatory. The OCV setting has two enabled values, where one is a workaround mode that allows buggy clients to connect. This setting is only relevant in ap or ap-wds mode, so I've added a validator function that will warn the user if they've selected this option while in sta or sta-wds mode.

I've done some limited testing on my setup with an R7800 running in ap mode with WPA3-SAE+802.11r encryption. It's probably a good idea if more people could take this for a spin to make sure that it works properly. I've also added a commit that updates the associated translation files, as per @dannil's suggestion. I would appreciate any comments about the positioning of the drop-down menu, the text strings I've used etc.

- [x] Tested on: (ipq806x, 25381-336a531c15, Firefox 123.0) :white_check_mark:

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1842,6 +1842,27 @@ return view.extend({
 						o.placeholder = '201';
 						o.rmempty = true;
 
+						if (L.hasSystemFeature('hostapd', 'ocv') || L.hasSystemFeature('wpasupplicant', 'ocv')) {
+							o = ss.taboption('encryption', form.ListValue, 'ocv', _('Operating Channel Validation'), _("Note: Workaround mode allows a STA that claims OCV capability to connect even if the STA doesn't send OCI or negotiate PMF."));
+							o.value('0', _('Disabled'));
+							o.value('1', _('Enabled'));
+							o.value('2', _('Enabled (workaround mode)'));
+							o.default = '0';
+							o.depends('ieee80211w', '1');
+							o.depends('ieee80211w', '2');
+
+							o.validate = function(section_id, value) {
+								var modeopt = this.section.children.filter(function(o) { return o.option == 'mode' })[0],
+								modeval = modeopt.formvalue(section_id);
+
+								if ((value == '2') && ((modeval == 'sta') || (modeval == 'sta-wds'))) {
+									return _('Workaround mode can only be used when acting as an access point.');
+								}
+
+								return true;
+							}
+						}
+
 						o = ss.taboption('encryption', form.Flag, 'wpa_disable_eapol_key_retries', _('Enable key reinstallation (KRACK) countermeasures'), _('Complicates key reinstallation attacks on the client side by disabling retransmission of EAPOL-Key frames that are used to install keys. This workaround might cause interoperability issues and reduced robustness of key negotiation especially in environments with heavy traffic load.'));
 						add_dependency_permutations(o, { mode: ['ap', 'ap-wds'], encryption: ['psk2', 'psk-mixed', 'sae', 'sae-mixed', 'wpa2', 'wpa3', 'wpa3-mixed'] });
 
--- a/modules/luci-base/po/ar/base.po
+++ b/modules/luci-base/po/ar/base.po
@@ -1006,7 +1006,7 @@ msgstr "اسمح للمستخدم <em>الجذر</em> بتسجيل الدخول 
 msgid "Allowed IPs"
 msgstr "عناوين IP المسموح بها"
 
-#: protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js:86
+#: protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js:88
 msgid "Allowed network technology"
 msgstr "تقنية الشبكة المسموح بها"
 
@@ -1172,7 +1172,7 @@ msgid ""
 "Assign prefix parts using this hexadecimal subprefix ID for this interface."
 msgstr "قم بتعيين أجزاء البادئة باستخدام معرف السداسي العشري هذا لهذه الواجهة."
 
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:2323
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:2344
 #: modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js:366
 msgid "Associated Stations"
 msgstr "المحطات المرتبطة"
@@ -1327,7 +1327,7 @@ msgstr "BSS الانتقال"
 
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:158
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:182
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:1876
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:1897
 #: modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js:427
 #: modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js:149
 msgid "BSSID"
@@ -1607,7 +1607,7 @@ msgstr ""
 #: modules/luci-compat/luasrc/view/cbi/simpleform.htm:52
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:188
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1258
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:2197
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:2218
 #: modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js:128
 #: modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js:297
 #: modules/luci-mod-system/htdocs/luci-static/resources/view/system/sshkeys.js:209
@@ -1752,7 +1752,7 @@ msgstr "يغير كلمة مرور المسؤول للوصول إلى الجها
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:128
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:184
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:504
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:1874
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:1895
 #: modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js:424
 #: modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js:170
 msgid "Channel"
@@ -1771,7 +1771,7 @@ msgstr "عرض القناة"
 msgid "Check filesystems before mount"
 msgstr "افحص أنظمة الملفات قبل التحميل"
 
-#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:2157
+#: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:2178
 msgid "Check this option to delete the existing networks from this radio."
 msgstr "حدد هذا الخيار لحذف الشبكات الموجودة من هذا الراديو."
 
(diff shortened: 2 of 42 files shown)


dannil:
For full completeness you should run `./build/i18n-sync.sh` so the translation files gets updated with your new strings, see https://github.com/openwrt/luci/wiki/i18n#translation-files.

huaracheguarache:
Ok, now I've included a commit that updates the translation files. Thanks for the tip, @dannil!

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1835,6 +1835,34 @@ return view.extend({
...
+							o.value('1', _('Enabled'));
+							o.value('2', _('Enabled (workaround mode)'));
+
+							add_dependency_permutations(o, { mode: ['ap', 'ap-wds', 'sta', 'sta-wds'], encryption: ['psk-mixed', 'psk2', 'sae', 'sae-mixed', 'wpa2', 'wpa3', 'wpa3-mixed', 'owe'] });
+
+							o.validate = function(section_id, value) {
+								var mfpopt = this.section.children.filter(function(o) { return o.option == 'ieee80211w' })[0],
+								mfpval = mfpopt.formvalue(section_id);
+
+								if ((value == '1' || value == '2') && mfpval == '0') {
```
systemcrash, line 1850: Doesn't it work just to add a depends on `mfpval`?

It returns a string which mentions values whose names might change or not be translated.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1835,6 +1835,34 @@ return view.extend({
...
+									return _('Requires 802.11w to be set to optional or required.');
+								}
+
+								var modeopt = this.section.children.filter(function(o) { return o.option == 'mode' })[0],
+								modeval = modeopt.formvalue(section_id),
+								modetitle = modeopt.vallist[modeopt.keylist.indexOf(modeval)];
+
+								if ((value == '2') && ((modeval == 'sta') || (modeval == 'sta-wds'))) {
```
systemcrash, line 1858: Same here for depends

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
huaracheguarache: Sorry, I'm very much a noob so please bear with me. What do you mean by adding a depends on `mfpval`? I looked at the other uses of the validate method in the script and they all return strings in some form.

systemcrash: https://openwrt.github.io/luci/jsapi/LuCI.form.ListValue.html#depends

huaracheguarache: I can remove that part of the validator function and just add the 802.11w dependency to the add_dependency_permutations part of the code:

```javascript
add_dependency_permutations(o, { mode: ['ap', 'ap-wds', 'sta', 'sta-wds'], encryption: ['psk-mixed', 'psk2', 'sae', 'sae-mixed', 'wpa2', 'wpa3', 'wpa3-mixed', 'owe'], ieee80211w: ['1', '2']});
```

The reason I had that validator function in the first place is because I thought it would be a good idea to alert the user about the requirement of 802.11w for ocv, but I think it's ok to drop this to avoid the string translation/renaming issue.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
huaracheguarache: I can't really get around having a validator function here because I need to notify the user that the workaround mode will not work in client mode.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
systemcrash: I don't trust that function which is why I recommended just depends. I'm not certain that the permutations function does what it should, until one has tested every variant. We had problems that meant we had to revert recently because of dep permutations. For a PR which was here for 3 years...

huaracheguarache: Ah, now I get it. I just had a look at the dependencies of 802.11w and realised that it covered what I needed. I've now changed the code to use depends calls. Thanks!

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
systemcrash: So in other words, in what modes *will* it work?

depends also has available: `!contains` and `!reverse`, and the following:
` opt.depends("test.section1.foo", "bar")`
Require the "foo" form option within the "section1" section to be set to "bar". 

so it seems like one can instead do:
```js

							o.depends({ieee80211w: '1', ...});
							o.depends({ieee80211w: '2', ...});
```
where `...` is the necessary magic of "and `mode` is not sta*"

huaracheguarache: Workaround mode will only work in `sta` and `sta-wds` mode. Just to be clear, do you want me to get rid of the validator function altogether? I might very well be missing something here, but I don't think it's possible to use depends to make individual options in the drop-down menu depend on some other value. Depends causes the entire ocv option to appear or be hidden based on whether the dependencies are fulfilled. That means that I would have to create two different taboption instances for ocv, with one for when mode is either `ap` or `ap-wds`, and the other for when mode is `sta` or `sta-wds`. The question then is how to deal with the info text about workaround mode. Do I only show it when mode is set to `sta` or `sta-wds`? Seems a bit messy, but I guess that could work.

systemcrash: > Workaround mode will only work in `sta` and `sta-wds` mode. 

Now we're getting somewhere. 

```js
							//this means: 
							o.depends('ieee80211w', '1'); // 'ocv' depends on 11w enabled... 
							o.depends('ieee80211w', '2'); // OR 'ocv' depends on 11w compat mode
```

We need:

```js
							o.depends('ieee80211w', '1'); // 'ocv' depends on 11w enabled... 
							o.depends({ieee80211w: '2', ...}); // OR 'ocv' depends on 11w compat mode AND *Workaround mode will only work in `sta` and `sta-wds` mode.*
```

huaracheguarache: Sorry, I think I got you confused with the workaround mode stuff. 802.11w has two possible "enabled" values:

```
1 - optional
2 - required
```

As long as ieee80211w is set to one of these values ocv will work in both ap (`ap`, `ap-wds`) and client (`sta`, `sta-wds`) modes. That means we can't account for mode having to be either `sta` or `sta-wds` when using ocv workaround mode with the help of depends. So, we're back to either keeping the validator function (which we don't want to because of the translation issues you pointed out), or to have two different drop-down lists depending on whether ap or client modes have been selected.

systemcrash: > That means we can't account for mode having to be either `sta` or `sta-wds` when using ocv workaround mode with the help of depends.


We *can't*?

https://github.com/openwrt/luci/blob/e2a106db27a0b45983aed3f80b12bead2673fd93/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L1695

https://github.com/openwrt/luci/blob/e2a106db27a0b45983aed3f80b12bead2673fd93/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js#L1007

huaracheguarache: Sure, those can show/hide the entire dropdown list based on whether the dependencies are fulfilled or not, but that doesn't help us in this case. As far as I can see, we can't show/hide individual options in the dropdown list by using depends. If that was possible then we could simply have shown the "workaround mode" option when mode was set to `sta` or `sta-wds`.

huaracheguarache: Just realised that I said that workaround mode will only work in `sta` and `sta-wds` mode, when it's actually in `ap` and `ap-wds` mode. Sorry about that!

systemcrash: Well, if you want to do it that way, gate the `o.value('2')` behind an `if` condition to prevent even showing it. 

We should probably also default the setting (when it is not already set) to 0.

huaracheguarache: I couldn't figure out how to do that, so I decided to keep the validator function and change the string so that it's static and doesn't contain values that can change. Is that ok with regard the the translation issue you pointed out earlier? I've also set the default value to 0.

systemcrash: try this?
```js
...
							var modeopt = this.section.children.filter(function(o) { return o.option == 'mode' })[0],
							modeval = modeopt.formvalue(section_id);
							if (!((modeval == 'sta') || (modeval == 'sta-wds'))) {
								o.value('2', _('Enabled (workaround mode)'));
							}
							o.depends('ieee80211w', '1');
							o.depends('ieee80211w', '2');

```

huaracheguarache: That doesn't work because I just end up with a "this is undefined" error. I assume I need to put this inside some method or something? But right now I'm just wondering what the issue is with the PR at this point. Is it not acceptable as is?

systemcrash: Largely yes. But if we can prevent making users feel dumb or that they did something wrong, and having them try to figure out what's wrong, it's best to prevent those cases altogether (preventing workaround mode when not in sta* modes).

huaracheguarache: Fair enough. I'll try to look at it some more to see if I can make that work.

huaracheguarache: I think my js knowledge is just too limited to be able to figure out how to be able to get the formvalue of mode. I need that for the if-statement that adds the workaround mode value. The code works fine inside the validation method, but when I bring it outside like you suggested below I get the "this is undefined" error.

> try this?
> 
> ```js
> ...
> 							var modeopt = this.section.children.filter(function(o) { return o.option == 'mode' })[0],
> 							modeval = modeopt.formvalue(section_id);
> 							if (!((modeval == 'sta') || (modeval == 'sta-wds'))) {
> 								o.value('2', _('Enabled (workaround mode)'));
> 							}
> 							o.depends('ieee80211w', '1');
> 							o.depends('ieee80211w', '2');
> ```

Very naively, I decided to try to replace `this` with `o`, and I get the `section_id` from `o.section.section` (not sure if this is correct?):

```javascript
var modeopt = o.section.children.filter(function(o) { return o.option == 'mode' })[0],
section_id = o.section.section,
modeval = modeopt.formvalue(section_id);
if (!((modeval == 'sta') || (modeval == 'sta-wds'))) {
    o.value('2', _('Enabled (workaround mode)'));
}
```

But trying to get the formvalue just ends up giving me a "this.root is undefined" error.

systemcrash:
```js
						var mode = uci.get('wireless', section_id, 'mode');
```
? Although `this` should work. Perhaps you just need to clear your cache and reload the page?

huaracheguarache: While `uci.get` is able to return the current value in the config, I don't think it's optimal because it's not reading the value that's currently selected in Luci. A user would be able to change the mode to `sta` or `sta-wds` and without saving open the security tab and then set operating channel validation to `Enabled (workaround mode)`.

Also, I tried refreshing while ignoring cached content (ctrl-shift-r) and `this` still didn't work. I guess it's because the script uses strict mode: _When invoked as a standalone function (not attached to an object: func()), this typically refers to the global object (in non-strict mode) or undefined (in strict mode)._[1]

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

systemcrash: Ah yeah, the in-memory version. Then your current version looks most appropriate.

huaracheguarache: Can this be merged now?

systemcrash:
@huaracheguarache could you rebase your changes please?

PR #6967, opened 2024-03-05
luci-app-statistics: Add UI to configure Mqtt

McGiverGim:
This commit adds UI to configure the Mqtt plugin of the luci-app-statistics.

This PR https://github.com/openwrt/packages/pull/20662 added the possibility of use UCI commands some time ago, now this PR adds the LuCi way. Some screen captures:

image

image

image

The UCI part was based on the cURL plugin, so I did the same in the UI part.

~~I've tested it and one thing is not working: the mandatory fields (not optional) are not being validated, so the user can left this fields empty. I don't know why it not works, maybe someone can point me to the bug.~~ I've fixed it but I'm not too sure if it's the correct way: to be mandatory it needs to have `optional = false` AND `rmempty = false`.

### DIFF
--- a/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/mqtt.js
+++ b/applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/mqtt.js
@@ -1,0 +1,88 @@
+'use strict';
+'require baseclass';
+'require form';
+
+return baseclass.extend({
+    title: _('Mqtt Plugin Configuration'),
+
+    addFormOptions: function(s) {
+        let o, ss;
+
+        o = s.option(form.Flag, 'enable', _('Sends or receives data via mqtt'));
+
+        o = s.option(form.SectionValue, '__blocks', form.GridSection, 'collectd_mqtt_block');
+        o.depends('enable', '1');
+
+        ss = o.subsection;
+        ss.anonymous = true;
+        ss.addremove = true;
+
+        o = ss.option(form.ListValue, 'blocktype', _('Type'));
+        o.value('Publish', _('Publish'));
+        o.value('Subscribe', _('Subscribe'));
+        o.default = 'Publish';
+
+        o = ss.option(form.Value, 'name', _('Name'));
+        o.optional = false;
+        o.rmempty = false;
+
+        o = ss.option(form.Value, 'Host', _('Host'));
+        o.datatype = 'host';
+        o.optional = false;
+        o.rmempty = false;
+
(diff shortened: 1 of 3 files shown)


McGiverGim:
Fixed the problem with the `optional = false` not being respected. I've fixed it but I'm not too sure if it's the correct way: to be mandatory it needs to have `optional = false` AND `rmempty = false`. Is this the expected behaviour?

systemcrash:
> Fixed the problem with the `optional = false` not being respected. I've fixed it but I'm not too sure if it's the correct way: to be mandatory it needs to have `optional = false` AND `rmempty = false`. Is this the expected behaviour?

Verify here: https://openwrt.github.io/luci/jsapi/LuCI.form.TextValue.html

Usually `optional = false` is enough. rmempty just ensures that empty strings are not saved to the config. Sometimes important depending on how the underlying config is handled.

systemcrash:
Looks good. @feckert ?

applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/mqtt.js
```diff
@@ -0,0 +1,88 @@
...
+        o = ss.option(form.Value, 'Topic', _('Topic'));
+        o.depends('blocktype', 'Subscribe');
+        o.optional = true;
+        o.modalonly = true;
+    },
+
+    configSummary: function(section) {
+        return _('Mqtt plugin enabled');
+    }
+});
```
jow-, line 88: Please add a final newline here.

applications/luci-app-statistics/root/usr/share/luci/statistics/plugins/mqtt.json
```diff
@@ -0,0 +1,4 @@
+{
+	"title": "Mqtt",
+	"category": "output"
+}
```
jow-, line 4: Please add a final newline here.

applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/mqtt.js
McGiverGim: Done!

applications/luci-app-statistics/root/usr/share/luci/statistics/plugins/mqtt.json
McGiverGim: Done!

systemcrash:
Thanks @McGiverGim

PR #6972, opened 2024-03-07
dnsmasq: Add options for new EDNS0 options

schuettecarsten:
The pull request https://github.com/openwrt/openwrt/pull/14349 adds four new options to configure EDNS0 upstream support. This patch adds the new settings to DNSMASQ web UI.

image

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -677,6 +677,26 @@ return view.extend({
 		);
 		o.placeholder = '/etc/dnsmasq.servers';
 
+		s.taboption('forward', form.Flag, 'add_mac',
+			_('Add requestor MAC'),
+			_('Add the MAC address of the requestor to DNS queries which are forwarded upstream.'));
+
+		s.taboption('forward', form.Flag, 'strip_mac',
+			_('Remove MAC address before forwarding query'),
+			_('Remove any MAC address information already in downstream queries before forwarding upstream.'));
+
+		o = s.taboption('forward', form.Value, 'add_subnet',
+			_('Add subnet address to forwards'),
+			_('Add a subnet address to the DNS queries which are forwarded upstream.') + ' ' + '<br />' +
+			_('%s adds the /24 and /96 subnets of the requestor for IPv4 and IPv6 requestors, respectively.').format('<code>24,96</code>') + ' ' + '<br />' +
+			_('%s adds 1.2.3.0/24 for IPv4 requestors and ::/0 for IPv6 requestors.').format('<code>1.2.3.4/24</code>') + ' ' + '<br />' +
+			_('%s adds 1.2.3.0/24 for both IPv4 and IPv6 requestors.').format('<code>1.2.3.4/24,1.2.3.4/24</code>'));
+		o.optional = true;
+
+		s.taboption('forward', form.Flag, 'strip_subnet',
+			_('Remove subnet address before forwarding query'),
+			_('Remove any subnet address already present in a downstream query before forwarding it upstream.'));
+
 		o = s.taboption('general', form.Flag, 'allservers',
 			_('All servers'),
 			_('Query all available upstream resolvers.') + ' ' + _('First answer wins.'));


dannil:
Make sure you don't change the indentation of the files, it makes the diff impossible to review.

schuettecarsten:
> Please don't change the indentation of the files, it makes the diff impossible to review.

Oh, sorry, I've set fork to ignore tabs/spaces, so I did not see that. Unfortunately, some js fils use tabs, some use spaces, so it's difficult to choose the correct editor setting and make sure that it does not change indentation on save. Force-pushed....

dannil:
> > Please don't change the indentation of the files, it makes the diff impossible to review.
> 
> Oh, sorry, I've set Fork to ignore tabs/spaces, so I did not see that. Unfortunately, some js fils use tabs, some use spaces, so it's difficult to choose the correct editor setting and make sure that it does not change indentation on save. Force-pushed....

If you have a modern, competent editor it can usually dynamically learn to use tab or spaces (and the amount of spaces) for the tab character depending on the rest of the content of the file, so you need to manually configure it (VSCode for example).

schuettecarsten:
> > > Please don't change the indentation of the files, it makes the diff impossible to review.
> > 
> > 
> > Oh, sorry, I've set Fork to ignore tabs/spaces, so I did not see that. Unfortunately, some js fils use tabs, some use spaces, so it's difficult to choose the correct editor setting and make sure that it does not change indentation on save. Force-pushed....
> 
> If you have a modern editor there's usually functionality to dynamically learn to use tab or spaces (and the amount of spaces) for the tab character depending on the rest of the content of the file, so no need to manually configure it (VSCode for example), but yeah entirely depends on your editor.

It was VSCode detecting mixed indent and deciding to reformat the whole file. :-D

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -647,6 +647,23 @@ return view.extend({
...
+			_('Add requestor MAC address to forwards'),
+			_('Add the MAC address of the requestor to DNS queries which are forwarded upstream.'));
+
+		s.taboption('forward', form.Flag, 'strip_mac',
+			_('Remove MAC address before forwarding query'),
+			_('Remove any MAC address information already in downstream queries before forwarding upstream.'));
+
+		o = s.taboption('forward', form.Value, 'add_subnet',
+			_('Add subnet address to forwards'),
+			_('Add a subnet address to the DNS queries which are forwarded upstream, e.g. {32,128} for full requestor address.'));
```
systemcrash, line 660: This is not a subnet address. This is a network prefix length. 
Hint the field with e.g `1.2.3.4/24` or `24,96`. Some explanation is necessary:

```js
			_('Add a subnet address to the DNS queries which are forwarded upstream.') + ' ' + '<br />' +
			_('%s adds the /24 and /96 subnets of the requestor for IPv4 and IPv6 requestors, respectively.').format('<code>24,96</code>') + ' ' + '<br />' +
			_('%s adds 1.2.3.0/24 for IPv4 requestors and ::/0 for IPv6 requestors.').format('<code>1.2.3.4/24</code>') + ' ' + '<br />' +
			_('%s adds 1.2.3.0/24 for both IPv4 and IPv6 requestors.').format('<code>1.2.3.4/24,1.2.3.4/24</code>'));
```

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -647,6 +647,23 @@ return view.extend({
...
+		o.optional = true;
+
+		s.taboption('forward', form.Flag, 'strip_subnet',
```
systemcrash, line 663: This option is mutex with `add_subnet` - so *every option here* needs either) some error checking is necessary or) `depends` to hide parameters. e.g. checking strip subnet hides add subnet and so on.

systemcrash (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
schuettecarsten: No, all options are independent from each other, see dnsmasq manpage for details.

systemcrash: I did 😎 the point being , what’s the use of checking both to add a Mac and remove one.

schuettecarsten: It is possible to pass --add-mac and --strip-mac at the same time and you are right, it looks like this does not make sense. In my understanding, it will remove any MAC address information that is included in the incoming DNS request and then add new MAC information of the direct requestor to the upstream request. This might be important if the sender of the DNS request is e.g. pi-hole. Then the MAc address in the request and the requestor might be different.

systemcrash: Nope. I've read it again. They're mutually exclusive.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -647,6 +647,26 @@ return view.extend({
 		);
 		o.placeholder = '/etc/dnsmasq.servers';
 
+		s.taboption('forward', form.Flag, 'add_mac',
```
systemcrash, line 650: This is not a flag. Or, it could use a flag but it must then be interpreted (0=base64, etc). Manpage: `--add-mac[=base64|text] `

So this can be a dropdown, with two pre-defined options: `base64` and `text`, or nothing.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -647,6 +647,26 @@ return view.extend({
...
+			_('Add requestor MAC address to forwards'),
```
systemcrash, line 651: Make it more concise, like: 
```js
			_('Add requestor MAC, as'),
```

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Devs @jow- and @rmilecki recently stated a preference for `enabled`. So `enabled` ideally triggers the naked `--add-mac`. But that needs to happen in https://github.com/openwrt/openwrt/pull/14349

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
schuettecarsten: @systemcrash
They are not exclusive. If BOTH the add and strip options are set, incoming EDNS0 options are REPLACED.
https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2022q1/016043.html

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
schuettecarsten: Yes, if https://github.com/openwrt/openwrt/pull/14349 is updated to support values for add-mac, I will also update this patch.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Finally. Clarity. Yeah I dug into the C code and that lines up.

This will need to be made clear in these options also. Either via a dummy option or under each option.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash:
```javascript

o = s.taboption('forward', form.Value, 'add_mac',
	_('Add requestor MAC'),
	_('Add the MAC address of the requestor to DNS queries which are forwarded upstream.'));
o.optional = true;
o.rmempty = true;
o.value('', _('off'));
o.value('enabled', _('Enabled'));
o.value('base64');
o.value('text');
```

schuettecarsten:
Will be replaced by updated version

PR #6975, opened 2024-03-09
luci-app-upnp: revise wording to include PCP

stokito:
Replaces #6863

@Self-Hosting-Group @systemcrash please review and merge before it stalled and got merge conflicts

### DIFF
--- a/applications/luci-app-upnp/Makefile
+++ b/applications/luci-app-upnp/Makefile
@@ -6,7 +6,7 @@
 
 include $(TOPDIR)/rules.mk
 
-LUCI_TITLE:=Universal Plug & Play configuration module
+LUCI_TITLE:=Universal Plug and Play (UPnP IGD) & PCP/NAT-PMP configuration module
 LUCI_DEPENDS:=+luci-base +miniupnpd +rpcd-mod-ucode
 
 include ../../luci.mk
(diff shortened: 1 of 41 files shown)


stokito:
Thank you

stokito:
Screenshot of Active Port Forwards

Do you mean applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js:31

Previously it was `Active UPnP Redirects`.

In your PR it was changed `Active UPnP IGD & PCP/NAT-PMP Port Forwards`. I feel that this is too much and simplified it to `Active Port Forwards`. This allowed to reuse the same key in applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js:113 which previously was `Active UPnP Redirects`.

I don't feel like it's that important to show UPnP there because it's anyway clear for those who knows or confusing for those who don't. Even more, I would like to see also static port forwardings there because I may forget to remove obsolete. But IMHO that's not worth to be implemented.

stokito:
If add the `UPnP IGD & PCP/NAT-PMP` in a middle then we should make the same change on the settings page because this is a same component. But then it will look too clumsy as for me:

Screenshot luci-app-upnp Active forwardings

IMHO everybody knows the UPnP and it should be just enough to mention the NATPMP in a description. Something like this:

> **UPnP IGD**
> UPnP IGD & PCP/NAT-PMP allows clients on the local network to automatically configure port forwards on the router. Also called Universal Plug and Play.
> 
> **Active UPnP IGD Port Forwards**


But anyway, I have nothing to add here and it was too much discussions that it's hard to follow. If you wish to change anything ask someone to send a PR.

PR #6982, opened 2024-03-12
Make Realtime Graphs Respect Bootstrap Dark Mode

JimMatthew:
Currently, the realtime graphs, under Status > Realtime Graphs are always white, even in Bootstrap Dark theme.

This change will make the graphs dark in darkmode.  I wanted to get some guidance on whether this is the right direction.
I used !important to set the color because I needed to override the JS which is setting the background white directly. I looked at fixing this in the javascript but there doesn't seem to be common variables among the different themes, so this way seemed the simplest. 

This is my first PR so please be patient if I have done anything incorrectly.

Example of dark graph.
Currently the graph lines do not show up well, I can work on that if requested.

darkmodeGraph

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -2565,3 +2565,15 @@ div.cbi-value var.cbi-tooltip-container,
 [data-page="admin-system-admin-sshkeys"] .cbi-dynlist {
 	max-width: none;
 }
+
+[data-darkmode="true"] [data-page="admin-status-realtime-load"] #view > div[style] {
+	background-color: var(--background-color-high)!important;
+}
+
+[data-darkmode="true"] [data-page="admin-status-realtime-bandwidth"] #view > div > div > div > div[style] {
+	background-color: var(--background-color-high)!important;
+}
+
+[data-darkmode="true"] [data-page="admin-status-realtime-connections"] #view > div[style] {
+	background-color: var(--background-color-high)!important;
+}


systemcrash:
Looks good! Just pay attention to other PR checks and resolve those. Squash your changes into a single commit and force push. Remember your signed-off-by line in the commit message.

JimMatthew:
> Looks good! Just pay attention to other PR checks and resolve those. Squash your changes into a single commit and force push. Remember your signed-off-by line in the commit message.

Thanks, did as requested.

systemcrash:
Almost! Just add subject line to commit, line should start like "themes: ..."

systemcrash:
Good! Thanks.

thomasschroeder:
Great update, would be highly beneficial for the "Channel Analysis" page as well.

One more suggestion: Changing the colour of the grid to the one of the table grid lines would make it much easier to read.
grafik

JimMatthew:
@systemcrash if I make improvements do I make a new pull request?

JimMatthew:
> Great update, would be highly beneficial for the "Channel Analysis" page as well.
> 
> One more suggestion: Changing the colour of the grid to the one of the table grid lines would make it much easier to read. grafik

what do you think of this? I left the border black because imo it looks better
darkmodeGraph2

thomasschroeder:
> @thomasschroeder what do you think of this? I left the border black because imo it looks better darkmodeGraph2

Looks good to me!

But one additional point - it seems like the wireless tab is missing in your original PR:
grafik

JimMatthew:
thanks, I don't use wireless so my build doesn't have those tabs

PR #6987, opened 2024-03-13
luci-base: fix license

feckert:
Mention: @jow- @Ansuel 

Description:
During the update of luasrcdiet with commit b5d5e5bf13e572af7f1589a01bfd338691d5f063 the license was incorrectly changed from 'Apache-2.0' to 'MIT'.

As the name implies, this is the basic package that must be installed on the target for the LuCI to work at all. So changing the license from 'Apache-2.0' to 'MIT' should not have happend, because this information belongs to the package 'luci-base' and to the 'luasrcdiet'.

Therefore, with commit b0d8fff3e40568af633032aacb0624da16f4266c the 'luasrcdiet' was moved to its own package. However, it was forgotten to adapt the licence to 'Apache-2.0' again.

While we are at it, also set the 'PKG_LICENSE_FILES' to 'LICENSE' and 'NOTICE'.

### DIFF
--- a/modules/luci-base/Makefile
+++ b/modules/luci-base/Makefile
@@ -27,7 +27,7 @@ LUCI_DEPENDS:=\
 	+ucode-mod-html \
 	+liblucihttp-ucode
 
-PKG_LICENSE:=MIT
+PKG_LICENSE:=Apache-2.0
 
 HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/$(PKG_NAME)
 


systemcrash:
Is the license file in the repo root also correct? 

> Mention: @jow- @Ansuel
> 
> Description: During the update of luasrcdiet with commit [b5d5e5b](https://github.com/openwrt/luci/commit/b5d5e5bf13e572af7f1589a01bfd338691d5f063) the license was incorrectly changed from 'Apache-2.0' to 'MIT'.
> 

*their* license seems correct, which was what was inherited. Is the license in [the source repo](https://github.com/jirutka/luasrcdiet) *incorrect*?

jow-:
The PKG_LICENSE in the Makefile refers to the LuCI code license, which is Apache 2.0. The same Makefile also references LuaSrcDiet which is under MIT license. Luckily both licenses are equally permissive so a quick fix might be simply checking PKG_LICENSE to `Apache-2.0 MIT`.

feckert:
@jow- I apologise for asking, but the package 'LuaSrcDiet' has been removed with commit https://github.com/openwrt/luci/commit/b0d8fff3e40568af633032aacb0624da16f4266c from the LuCI feed and added with commit https://github.com/openwrt/packages/commit/46d68b869951854206b0c95423480881d3356c90 to the packages feed.

To be correct the license of `luci-base` should `Apache-2.0` again and not `MIT` *and* `Apache-2.0`?

The `PKG_LICENSE_FILES` handling is not  correct in this pullrequest, because the needed files are two directories above of this Makefile. I have to look at this again.

See https://github.com/openwrt/luci/blob/master/LICENSE
See https://github.com/openwrt/luci/blob/master/NOTICE

systemcrash:
@feckert @jow- do any of these license fixes need picking to branches?

feckert:
@systemcrash If it is simply possible, then I will cherry-pick the next few days

PR #6992, opened 2024-03-14
luci-theme-bootstrap: Change textarea font-family to monospace

skramstad:
Better readability for pages that use textarea, similar to `admin/status/logs`.

The pages below use textarea today:
- Startup -> Local Startup
- Flash Firmware -> Configuration
- Diagnostics

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -259,6 +259,7 @@ input[type="search"]::-webkit-search-decoration {
 textarea {
 	overflow: auto;
 	vertical-align: top;
+	font-family: monospace;
 }
 
 .control-group {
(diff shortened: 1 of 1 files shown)


skramstad:
@stangri Hi, and thanks for the feedback. New PR with a little better description and an example(page) for how it will look after the change to monospace.

Since most of the text is more like a "console", it makes sense to use monospace than Helvetica.

stangri:
@skramstad thank you, any chance you could add the before/after (or at least after) screenshots for the pages you've referenced?

systemcrash:
Before and after (monospace more consistent, especially for kernel logs)

Screenshot 2024-03-27 at 19 25 04

Screenshot 2024-03-27 at 19 25 10

systemcrash:
Thanks @skramstad !

thomasschroeder:
> Before and after (monospace more consistent, especially for kernel logs)

For kernel- and syslogs there was an older commit 04d9820 - I think this old one can be reverted as it's now a generic solution. Also in for syslog there's a special font color: `color: var(--text-color-highest);` - do you think it should be changed to the default textarea value as well?

skramstad:
> Thanks @skramstad !

No worries, and thanks for posting screenshots 👍

systemcrash:
@skramstad ?

skramstad:
> @skramstad ?
> 
> > this old one can be reverted

If I understand this question right, yes that old commit can be reverted.

PR #6994, opened 2024-03-15
luci-app-nlbwmon: add missing PKG_LICENSE

feckert:
@jow- As you may have seen, I am currently updating the license information in some makefiles. I have not found anything for the luci-app-nlbwmon. So that I do not commit something wrong!
Is `luci-app-nlbwmon` licensed  under the Apache 2.0?

### DIFF
--- a/applications/luci-app-nlbwmon/Makefile
+++ b/applications/luci-app-nlbwmon/Makefile
@@ -1,5 +1,7 @@
 include $(TOPDIR)/rules.mk
 
+PKG_LICENSE:=Apache-2.0
+
 LUCI_TITLE:=Netlink based bandwidth accounting
 LUCI_DEPENDS:=+luci-base +nlbwmon
 


jow-:
Yes, you can - unless otherwise stated - assume that all LuCI apps are Apache 2.0 licenses.

feckert:
> Note however that the bundled chart.min.js belongs to the https://github.com/chartjs/Chart.js project and is licensed under MIT

Good to know.
Should I then add MIT or should I move this to a separate package?
It looks like the minified is checked in.
Is there no official source for this?

systemcrash:
Probably this one.

https://cdnjs.com/libraries/Chart.js

jow-:
Iirc it is a custom minified build that only includes the chart core and the doughnut plugin, it is not any official chart.js distribution file. It probably is enough to add a header like the following:

```javascript
/* Custom minified build of Chart.js supporting only doughnut style graphs
 * Released under the MIT license
 * https://github.com/chartjs/Chart.js
 */
 ```

**Comment by feckert** [comment]
@jow- I moved the `chart.js` file as it is  in this pullrequest into its own package.
So this library could gets its own `MIT` license tag.
I want to switch to an official upstream version of `chart.js` in the feature, so I could see the source and not minified subset of the official version.
Also I need this lib to generate nice graphs in other apps.

**Comment by systemcrash** [comment]
When I compile the sources, every time I run make, this line is included/output:
```
WARNING: Makefile 'package/feeds/luci/luci-app-nlbwmon/Makefile' has a dependency on 'luci-lib-chartjs', which does not exist
```

FAO @feckert @jow-

**Comment by jow-** [comment]
```
./scripts/feeds update luci
./scripts/feeds install luci-lib-chartjs
```

systemcrash:
> ./scripts/feeds install luci-lib-chartjs

Hmm - I ran the update part about 10 times already. But that install bit fixed it. Thanks @jow-

PR #7002, opened 2024-03-20
luci-mod-status: Move "Scroll to tail" button to header line on syslog and kernellog status pages

CharlesMengCA:
Layout after the changes.

Button

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
@@ -51,9 +51,11 @@ return view.extend({
 		});
 
 		return E([], [
-			E('h2', {}, [ _('Kernel Log') ]),
+			E('div', { 'style': 'display:flex; align-items:center; gap:5rem; padding-bottom:1rem' }, [
+				E('h2', {}, [ _('Kernel Log') ]),
+				scrollDownButton
+         ]),
 			E('div', { 'id': 'content_syslog' }, [
-				E('div', {'style': 'padding-bottom: 20px'}, [scrollDownButton]),
 				E('textarea', {
 					'id': 'syslog',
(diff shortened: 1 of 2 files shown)


systemcrash:
Hello! Thank you for this PR. 

Would you like to move the button for aesthetic reasons? Balance?

I think I will NACK this for a few reasons:
- for the majority L-to-R readers it's an extra eye-scan to read the button content from the header text. 
- You now have to move the mouse quite far - some extra mileage, especially when you might actually need to swap between the kernel and system logs. More exercise than necessary :)

CharlesMengCA:
Just look at the github headers, you will see it is very common to put the header and buttons on the same row.

Capture

This way will save some space on vertical direction. Normally most of monitors are 4:3, 16:9 or even wider, the vertical space is kind of tight and shouldn't be wasted.

Most the time, when I am on the log page, I will use the mouse wheel to scroll down the log to see if there is an error or warning.

Moving the button away will give me more view area at the beginning.

Ramon00:
> * You now have to move the mouse quite far - some extra mileage, especially when you might actually need to swap between the kernel and system logs. More exercise than necessary :)

I also like the current layout, indeed, if i need to skip between kernel and syslog then i dont need move my mouse too much. 

> Most the time, when I am on the log page, I will use the mouse wheel to scroll down the log to see if there is an error or warning.
> 
> Moving the button away will give me more view area at the beginning.

That kind of kills the argument about the viewing space, you can just use the scroll wheel to move the button off the screen.

CharlesMengCA:
Moved button beside the header.

Just don't like two separated line for a short header and a single button.

Button

systemcrash:
@Ramon00 - updated. Good?

Ramon00:
@systemcrash I think its a good compromise!

CharlesMengCA:
What's next? just waiting for the approval? :)

PR #7006, opened 2024-03-21
luci.mk: make APK version compliant

aparcar:
APK expects hashes to be prefixed with a ~

@jow- please have a look


jow-:
Fine with me!

PR #7008, opened 2024-03-21
luci-theme-material: add back removed color variables

dannil:
Spawned from https://github.com/openwrt/luci/pull/6936#issuecomment-2011632254. Everything should look as before.

Partially reverts 2bdabf14c30320ef0433828b1a080f6b8fe9c430 by adding back the deleted color variables as the custom.css was used by users to override the theme with their own colors. To maintain backwards compatibility with users we should not remove/rename these as it will result in colors being applied incorrectly.

Ping @feckert 

- [X] Tested on: (x86/64, 23.05.2, Firefox) :white_check_mark:

### DIFF
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -212,7 +212,7 @@ select,
 select:not([multiple="multiple"]):focus,
 input:not(.cbi-button):focus,
 .cbi-dropdown:focus {
-	border-color: var(--primary-color);
+	border-color: var(--main-color);
 }
 
 .cbi-dropdown,
@@ -346,7 +346,7 @@ small {
 	overflow-x: auto;
 	width: calc(0% + 15rem);
 	height: calc(100% - 4rem);
-	background-color: var(--white-color);
+	background-color: var(--menu-bg-color);
 	transition: visibility 400ms, width 400ms;
 }
 
@@ -379,8 +379,8 @@ small {
 }
 
 header {
-	color: var(--secondary-color);
-	background: var(--primary-color);
+	color: var(--header-color);
+	background: var(--header-bg);
 }
 
 header > .fill > .container {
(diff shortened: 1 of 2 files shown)


systemcrash:
Ready @dannil ?

dannil:
> Ready @dannil ?

Yes I would think so, I was waiting for feedback from @feckert tho.

feckert:
@dannil Thanks for your work and the fix. Is working as expected :+1:

feckert:
@dannil I've only just seen it.
Next time, please make sure that the length of the commit description does not exceed the 80 character limit per line!.

systemcrash:
> @dannil I've only just seen it. Next time, please make sure that the length of the commit description does not exceed the 80 character limit per line!.

Is this just something we blindly repeat without understanding why? Torvalds removed this restriction a number of years back, since we have high-res desktops capable of more than 80 characters wide since forever, and editors that can horizontally scroll since even more forever.

jow-:
I want to keep that limitation, makes it easier to work with git log in small terminal windows.

systemcrash:
Solved problem apparently: https://stackoverflow.com/questions/2119942/how-to-wrap-git-commit-comments

dannil:
Would it make sense to think about adding the 80 character limit to the [formal.yml](https://github.com/openwrt/actions-shared-workflows/blob/main/.github/workflows/formal.yml) task?

feckert:
@systemcrash I have found an interesting article a bout this topic https://hackaday.com/2020/06/18/ask-hackaday-are-80-characters-per-line-still-reasonable-in-2020/

From my point of view, the following statement says it all. 

`In other words, requires us to move our eyes, or even the entire head, disrupting the reading flow and increasing the risk to overlook something important.`

systemcrash:
Salient points. Consider this: the reason ebooks and HTML don't use 80 character limits anywhere (and why PDFs look and behave like dog-shit on portable readers) is...

RE-FLOW

If text is one contiguous block, you can reformat it at will (which is what github attempts to do) and set whatever width you want. 80 characters be damned. I'd be happy with 30 or 40 (for faster reading, as your point).

dannil:
I've created https://forum.openwrt.org/t/enforcement-of-commit-subject-and-description-in-ci/193391 to follow up the discussion on the formal requirements for commits.

Issue #7022, opened 2024-03-23
luci-app-ttyd: Nginx reverse proxy


alex9434:
## Steps to reproduce:

1. Connect to Luci with cloudflared or behind firewall with only 80 and 443 open ports
2. Go to services -> terminal

## Actual behavior:
Empty page for terminal

## Expected behavior:
Terminal windows

Lucy-app-ttyd uses separate port (default 7681) for ttyd. I would like to use nginx reverse proxy to have ttyd on port 80 (and 443). That requires to change the web interface to use a path for ttyd to use nginx reverse proxy.

ysc3839:
This issue has been there for some years. Because there are many different SSL/reverse proxy configurations. It's difficult to detect it automatically. See also #5347.

alex9434:
Hi @ysc3839 
My suggestion is different than #5347 . I want to use nginx as reverse proxy (could be an option in the config)

```
        location /ttyd {
                proxy_set_header Host $host;
                proxy_set_header X-Forwarded-Proto $scheme;
                proxy_set_header X-Forwarded-Port $server_port;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_pass http://localhost:7681;
                proxy_http_version 1.1;
                proxy_set_header Upgrade $http_upgrade;
                proxy_set_header Connection "upgrade";
        }
```
However, that would require the web-interface to use the path /ttyd for requests.

alex9434:
Great, will test and report back

alex9434:
Hi @ysc3839, I have successfully compiled and installed your commit. Apologies for the delay. I have added the above snipped to `/etc/nginx/conf.d/ttyd.locations` (and restarted nginx).

What exactly do I need to put in `URL_override` in the LuCI config for ttyd? Just putting in `/ttyd` does not seem to work. Other other options that I need to chose to make this work?

ysc3839:
Did you start ttyd? The error is `Connection refused`, which means no program is listening that port.
And you should use UNIX socket for localhost reverse proxy, because it has better performance.

alex9434:
Hi @ysc3839, ttyd is running but without active instances.
```
service ttyd status
active with no instances
```
Which is strange because I have created an instance that is enabled in LuCI. When I create an instance with Unix socket in `/var/run/ttyd.sock`, the file does not appear. What is going on?

alex9434:
[...]

Issue #5347, opened 2021-09-08
Integrate luci-ssl and luci-app-ttyd


alexforencich:
It would be great if the same SSL certificates could be used for both luci-ssl and luci-app-ttyd.  I tried pointing luci-app-ttyd at the /etc/uhttpd.crt and /etc/uhttpd.key files, but unfortunately this does not work as it seems uhttpd.key is in some binary format that ttyd does not understand and hence ttyd does not start.  Not sure what the best solution to this would be - perhaps change the format for the automatically-generated uhttpd.key to something that both uhttpd and ttyd can understand, or alternatively to patch ttyd to understand whatever format is used for uhttpd.key.

alexforencich:
See also #5048

professor-jonny:
> As a side-note, installing luci-app-ttyd alongside luci-ssl without enabling SSL in luci-app-ttyd is rather useless as modern browsers tend to block the non-secure iframe that luci-app-ttyd uses to connect to ttyd when the page is served over HTTPS. So having a nice way to make this work out-of-the-box would be quite convenient.

Did you manage to generate compatable certs for ttyd and get it to work?
If so could you point me to this information ?

ysc3839:
I think the final solution is to use some http reverse proxy server(nginx) to allow apps to share a single SSL port(443). However it may not easy to generate config file automatically. Still requires manual configure.
I think enabling SSL is an advanced option because you have to use a domain name to get a CA signed certificate. Ignoring SSL error is unsafe and the error page will show again in some days after you ignore it. This is not user friendly and I think most users will not enable SSL.
So just leave this issue and let advanced users configure manually.

alexforencich:
Oh yeah, using a reverse proxy is a much better idea - no oddball open ports to deal with on the firewall end, no separate SSL configuration, and no issues with mixes SSL/non-SSL parts of the page.  It darn well better be possible to set that up automatically as part of the package so no manual intervention is required.

okibcn:
In any case, the compatibility between uhttpd and ttyd should be addressed. Creating all the SSL stuff is already complex. luci-uhttpd has managed to automate most of the tasks. @ysc3839, what would be the config files related to that alternative: I assume it involves installing more packages. At least `nginx` and `nginx-ssl`, isn't it?

alexforencich:
[...]

PR #7026, opened 2024-03-25
luci-app-ttyd: add option for UNIX socket and URL override

ysc3839:
- [x] Tested on: (armsr, snapshot, Edge) :white_check_mark:
- [x] Description: Related issues: #5347 and #7022

### DIFF
--- a/applications/luci-app-ttyd/htdocs/luci-static/resources/view/ttyd/config.js
+++ b/applications/luci-app-ttyd/htdocs/luci-static/resources/view/ttyd/config.js
@@ -17,12 +17,21 @@ return view.extend({
 		o = s.option(form.Flag, 'enable', _('Enable'));
 		o.default = true;
 
+		s.option(form.Flag, 'unix_sock', _('UNIX socket'), _('Bind to UNIX domain socket instead of IP port'));
+
 		o = s.option(form.Value, 'port', _('Port'), _('Port to listen (default: 7681, use `0` for random port)'));
+		o.depends('unix_sock', '0');
 		o.datatype    = 'port';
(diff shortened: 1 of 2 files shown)


systemcrash:
Good. Thanks.  @ysc3839

ZeaKyX:
Hi @ysc3839 @systemcrash
I noticed that in this change there’s a note: 

> Note that reverse proxied pages are NOT protected by password like LuCI.


However, in the [Unraid system](https://unraid.net/), the built-in ttyd supports being reverse proxied by Nginx to a custom URL, and it is also integrated with the system’s web management authentication. If the web management page is not logged in, opening the ttyd link will redirect to the login page.
image

May I ask if it’s possible for LuCI to implement a similar feature?

ysc3839:
@ZeaKyX I guess it's using nginx's built-in authentication mechanism. In this case, ttyd is behind nginx, so it's protected.
However, by default LuCI uses uhttpd as reverse proxy server, but uhttpd doesn't have authentication mechanism. Client request goes directly to LuCI, and then LuCI checks token.

ZeaKyX:
> @ZeaKyX I guess it's using nginx's built-in authentication mechanism. In this case, ttyd is behind nginx, so it's protected. However, by default LuCI uses uhttpd as reverse proxy server, but uhttpd doesn't have authentication mechanism. Client request goes directly to LuCI, and then LuCI checks token.

大佬好，本来这个url功能就是为nginx准备的吧，我看原issue里也是装的nginx替代uhttpd。
关于装了nginx-full后openwrt的登录保护，大佬有什么头绪吗？之前研究半天没找到是怎么实现的😂

PR #7029, opened 2024-03-26
luci.mk: more APK version compatibility changes

aparcar:
The initial commit did not take care of corner cases which could happen if building under build conditions other than using Git.

There are three cases for the `findrev` function:

- git -> `24.079.58964~7943616` (remove prefixed `git-`)

APK can't handle `git-` as prefix, it could be `0_git<integer>` but this seems rather confusing.

- date -> `0.240326.34906` ( add leading 0)

Add a leading zero so the version is always lower compared to using Git. This makes it easier to distinguish from a Git based version.

- unknown -> `0` (instead of `unknown`)

APK can't handle `unknown` so set it to a simple zero instead.

### DIFF
--- a/luci.mk
+++ b/luci.mk
@@ -85,18 +85,18 @@ define findrev
       if [ -n "$$1" ]; then
         secs="$$(($$1 % 86400))"; \
         yday="$$(date --utc --date="@$$1" "+%y.%j")"; \
-        printf 'git-%s.%05d~%s' "$$yday" "$$secs" "$$2"; \
+        printf '%s.%05d~%s' "$$yday" "$$secs" "$$2"; \
       else \
-        echo "unknown"; \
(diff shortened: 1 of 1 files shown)


hnyman:
>  date -> 0.240326.34906 ( add trailing 0)

You probably meant "leading 0" or "preceding 0" ?
In commit message, too.

aparcar:
> You probably meant "leading 0" or "preceding 0" or "prepending 0"?

Fixed, thanks.

@jow- can you please have a look?

hnyman:
@jow- 
Any objections?

It would likely make sense to first commit these APK related changes before adjusting .zst hashes in LuCI repo, as these APK changes might affect the calculated hashes.

jow-:
Yeah, okay with me, please go ahead.

Issue #7027, opened 2024-03-25
luci-theme-material: system log "Scroll to tail" and "Scroll to head" buttons don't work


dannil:
## Steps to reproduce:

1. Go to: Status -> System Log
2. Click either "Scroll to tail" or "Scroll to head" buttons

## Actual behavior:

Clicking "Scroll to tail" and "Scroll to head" doesn't do anything

## Expected behavior:
Clicking "Scroll to tail" should jump to the bottom of the page and "Scroll to head" should jump to the top of the page, just as the Bootstrap theme does.

## Additional Information:

I also saw this issue in SNAPSHOT before the `23.05.3` release as well so has been there for a while at least.

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='23.05.3'
DISTRIB_REVISION='r23809-234f1a2efa'
DISTRIB_TARGET='x86/64'
DISTRIB_ARCH='x86_64'
DISTRIB_DESCRIPTION='OpenWrt 23.05.3 r23809-234f1a2efa'
DISTRIB_TAINTS=''
```

systemcrash:
confirmed

systemcrash:
Actually, this is only a problem when the side-panel is visible. If I open the debug window and widen it enough or narrow the window enough such that the side-panel (automatically) disappears, then the buttons work.

McGiverGim:
The problem is that OpenWrt can use themes. The actual code moves the scroll of the "window", that is were the bootstrap theme has the scroll:
https://github.com/openwrt/luci/blob/b69bce077d7d198f9618881b79e1a13b878d466f/modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js#L30-L32

That's true for the bootstrap theme, but not for material or argon theme, this and other themes use DIV for the scroll.

I've a custom code that "searchs" for a scroll in the parents and moves it. It works but I don't know if under some circumstances can move a "wrong" scroll.

I will do more tests and push a PR to let others test it and see if it works or not.

PR #7032, opened 2024-03-27
luci-mod-status: fix syslog scroll for some themes

McGiverGim:
The actual code moves the scroll in the window. This works for the bootstrap theme, because the scroll is at window level. But this does not work for other themes, like material.

This commit changes the move of the scroll by "focusing" the window in the opposite button element in the syslog page. In this way the move is automatically done by the browser.

Another solution is to "search" in the parent until we find the scroll and move it, but seems less solid.

This fixes https://github.com/openwrt/luci/issues/7027

This has been tested on Chrome and Edge, any other test will be appreciated. I've tested too the solution "searching" for the scroll in the parent, and it works, but I'm not too sure where to stop the search if the scroll does not exist or if it can find a different scroll at other place, so I used this other solution.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
@@ -22,7 +22,7 @@ return view.extend({
 			}, _('Scroll to tail', 'scroll to bottom (the tail) of the log file')
 		);
 		scrollDownButton.addEventListener('click', function() {
-			window.scrollTo(0, document.body.scrollHeight);
+			scrollUpButton.focus();
 		});
 
 		var scrollUpButton = E('button', { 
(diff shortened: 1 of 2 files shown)


McGiverGim:
I noticed the dmesg log had the same problem, so I modified the PR to fix both scrolls.

McGiverGim:
Little modification, to use the variables created for the buttons and available at this context, in place of search for them in the document each time.

systemcrash:
Good. Thanks @McGiverGim

dave14305:
The scroll button now does nothing on iPad OS Safari, or iOS Safari. LuCI openwrt-23.05 branch git-24.086.45142-09d5a38. Just updated with a fresh 23.05.3 imagebuilder build with the newer LuCI packages.

I have an older 23.05.3 AP with LuCI openwrt-23.05 branch git-24.073.29889-cd7e519 where the scroll button still works fine on the same Apple devices.

PR #7033, opened 2024-03-27
luci-mod-status: persist sorting of Wi-Fi associated stations table

thomasschroeder:
Similar to commit 6039925 from @jow-

Add an ID attribute to the dynamically generated Wi-Fi associated stations table to persist row ordering choice across reloads.

At the luci-mod-network view, the Wi-Fi associated stations table already has the same ID.


systemcrash:
Merged. Thanks @thomasschroeder

PR #7037, opened 2024-03-31
luci-app-usteer: Escaped more strings to harden

Ramon00:
I hardened up the code a bit. I do not think this is strictly needed as those strings should be clean to start (mostly numbers, mac, IP addresses etc.) but cant hurt either and will make it more resistant to code changes in the future.

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -45,10 +45,10 @@ function collectHearingClient(client_table_entries, mac) {
 			client_table_entries.push([
 				'<nobr>' + '%h'.format(wlansplit[0]) + '</nobr>',
 				'<nobr>' + '%h'.format(wlansplit[1]) + '</nobr>',
-				SSID,
-				freq,
+				'%h'.format(SSID),
+				'%h'.format(freq),
 				Clients[mac][wlanc]['connected'] === true ? 'Yes' : 'No',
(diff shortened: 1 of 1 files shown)


systemcrash:
What is this %h formatting parameter? I cannot seem to find it documented anywhere.

Ramon00:
Somebody taught me that in another PR comment, it cleans out scripting tags to prevent injections. To be honest I never did look it up if that actually works... If you know a better method then im always willing to learn.

systemcrash:
Sounds like he said she said. I cannot find this documented. Anywhere. I'd like to see where %h is documented, described, or at least demonstrated to do what was asserted. If it does so, I'm happy to have it, but I'm turning up nothing.

Ramon00:
i just tested it, and if i have e.g. <script> in a string it prints it literally, without this formatting, it gets interpreted. But indeed I also cannot find a reference. Note that this is used in more luci js code. So well, not sure if this is an "undocumented" feature or not.  But like I said if we have a better method to clean strings before printing them im fine with that too.  In general, for an app like this, it gets information from an external source, so it is a good idea to make sure it does not contain hidden tags which may facilitate XSS. IMHO its this method or a different one, leaving it out is worse.

Ramon00:
@dannil  thx

systemcrash:
> @systemcrash @Ramon00 the `h` placeholder is defined in https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/cbi.js#L614-L616 (along with the rest of the `format` function) which invokes the [esc](https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/cbi.js#L518-L531) function.

Theeeere it iiiis. 🤦

systemcrash:
Merged. Thanks @Ramon00

PR #7009, opened 2024-03-23
luci-app-adblock-fast: prepare migration to APK

stangri:
Maintainer: me
Compile tested: x86_64, Sophos XG-135r3, OpenWrt 23.05.2

PR #7042, opened 2024-04-03
luci.mk: include release in version number

stangri:
Fixes issue mentioned in https://github.com/openwrt/luci/pull/7009#issuecomment-2035504461 where the IPK filename doesn't contain release from the luci-app Makefile.


stangri:
@jow- @neheb @aparcar @hnyman any feedback? This makes the versioning in IPK filenames consistent with the packages repo.

jow-:
WIthout having runtime tested it, it looks good to me so far. It will only change filenames for Makefiles that *do* declare PKG_REVISION, right?

stangri:
> The terminology in the PR is a bit confusing: the commit talks about "revision" but actually changes the handling of PKG_RELEASE. Please adjust.

Fixed in both commit message and PR.

> I guess that @jow- meant that it shouldn't affect the bulk of packages that are not using PKG_VERSION at all and just rely on the automatic versioning. (REVISION was likely misspelling of VERSION, not RELEASE, I think...)

Yes, like the previous code, it should only affect the version if the PKG_VERSION is set, otherwise it uses PKG_SRC_VERSION.

stangri:
@jow- @hnyman please merge if no further feedback.

PR #7056, opened 2024-04-11
luci-base: conversion of ucitrack handling from uci to json

feckert:
Description:

The `ucitrack` configuration under `/etc/config` defines which service must be restarted on a `LuCI` change. This `uci`  file defines how all this works together. Additionally there are `uci-defaults` scripts in the various applications that add additional configurations parameters to this `ucitrack` file if a service must be reloaded/restarted if a `uci` change was made in this application.

There are some problems with the current implementation:
* The `uci` should be used to configure the system and not for this kind of reload/restart dependency handling on `LuCI` changes.
*  On a system update with configuration restore of the `uci` file `ucitrack` the new behavior on `LuCI` reload/restart could not take into account because the new file is not be used.

In these two repositories, the change must then also be made after the merge.
https://github.com/openwrt/packages/pull/23879
https://github.com/openwrt/routing/pull/1062


jow-:
In general this seems fine. My gut feeling is that a lot of these are redundant nowadays, every package with a procd enabled init script that installs a reload handler or does uci to command line conversion should be covered by generic logic already. Only ucitrack entries for servies having legacy init scripts or no init scripts at all should remain (or preferably, the underlying packages should be ported to procd)

systemcrash:
@feckert 

For those out there (posterity, and googlers who find this) who have PRs in the works, they just migrate from ucitrack to json file and everything just 'works'? 

Delete this:
```sh
#!/bin/sh

uci -q batch <<-EOF >/dev/null
	delete ucitrack.@app[-1]
	add ucitrack app
	set ucitrack.@app[-1].init=appd
	commit ucitrack
EOF

exit 0
```

Do this:
```json
{
	"config": "appd",
	"init": "appd"
}
```



(How) Does one handle the following in JSON? e.g.

```sh
#!/bin/sh

rm -f /tmp/luci-indexcache
exit 0
```

systemcrash:
I also did not understand this bit:

> the new behavior on LuCI reload/restart could not take into account because the new file is not be used

feckert:
Paul Donald ***@***.***> schrieb am Do., 11. Apr. 2024, 19:43:

> @feckert <https://github.com/feckert>
>
> For those out there (posterity, and googlers who find this) who have PRs
> in the works, they just migrate from ucitrack to json file and everything
> just 'works'?
>
> Delete this:
>
> #!/bin/sh
>
> uci -q batch <<-EOF >/dev/null	delete ***@***.***[-1]	add ucitrack app	set ***@***.***[-1].init=appd	commit ucitrackEOF
> exit 0
>
> Do this:
>
> {
> 	"config": "appd",
> 	"init": "appd"
> }
>
> (How) Does one handle the following in JSON? e.g.
>
> #!/bin/sh
>
> rm -f /tmp/luci-indexcacheexit 0
>
> This is not needed anymore. This is already done generally in luci.mk
https://github.com/openwrt/luci/blob/291dd24819eecf5c6b878d5b82d917cfea765505/luci.mk#L228

>

feckert:
@systemcrash 
> 
> (How) Does one handle the following in JSON? e.g.
> 
> ```shell
> #!/bin/sh
> 
> rm -f /tmp/luci-indexcache
> exit 0
> ```
This is no longer needed. We already add this to every package in luci.mk in general. See the following lines in `luci.mk`.
https://github.com/openwrt/luci/blob/291dd24819eecf5c6b878d5b82d917cfea765505/luci.mk#L227-L230


> I also did not understand this bit:
> 
> > the new behavior on LuCI reload/restart could not take into account because the new file is not be used

This means that if I keep the `/etc/config/ucitrack` file during a system upgrade and this file has changed, the new file will not be used. And the LuCI will not react as intended to a LuCI change.

Issue #6422, opened 2023-06-08
luci-base: menu doesn't regenerate on UCI change


wryun:
## Steps to reproduce:

1. Add a uci dependency to a JSON menu item (e.g. `"depends": {"uci": {"system": {"foo": 1}}`)
2. Modify that configuration option
3. Reload the page

## Actual behavior:

The menu won't change unless you logout (to clear the session) and remove /tmp/luci-indexcache*

## Expected behavior:

When saving UCI items from the frontend, the frontend should automatically regenerate the menu.

## Additional Information:

Custom build of 22.03.5.

jow-:
Not sure we really want to flush the menu *on every single* uci change. The menu depends mechanism was not really meant for that level of dynamic usage. Primary use case was masking features (such as hiding wifi or switch config if the relevant sections are not present).

wryun:
I looked into this because we were using it to show log file pages only in certain situations (i.e. when particular options are enabled).

I'd argue that this makes it behave in a less surprising way, the cost of rebuilding the menu is not that high and that uci confirm is already a somewhat slow process. But I can understand that it's less than elegant given that most of the time it wouldn't change.

Alternatively, if the idea is that it should really only be for entire sections, perhaps the individual options dependency support could be removed. Do you see a use case for this when from a user perspective the device has to be rebooted for changes to 'take'?

jjm2473:
@jow- I have notice that, too. After wifi module auto config, menu not shown "Wireless". Same issue appears on Docker, if I open luci on browser on boot, Docker maybe not start on that time, menu will not show Docker menu even after docker started.

PR #7058, opened 2024-04-12
luci-base: force menu to regenerate after uci change

wryun:
Because the menu JSON can have 'depends' in them, uci changes should force the menu to regenerate.

Checklist suggested incrementing PKG_VERSION, but AFAIK for luci this is automatic.

Closes openwrt/luci#6422

- [x] Tested on: (arm64, 23.05, Firefox + Chrome) :white_check_mark:
- [x] \( Optional ) Closes: e.g. openwrt/luci#issue-number

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/uci.js
+++ b/modules/luci-base/htdocs/luci-static/resources/uci.js
@@ -919,6 +919,8 @@ return baseclass.extend(/** @lends LuCI.uci.prototype */ {
 							window.setTimeout(try_confirm, 250);
 						else
 							return Promise.reject(rv);
+					} else {
+						document.dispatchEvent(new CustomEvent('uci-applied'));
 					}
 
 					return rv;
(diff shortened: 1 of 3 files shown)


systemcrash:
Sooo, how does this differ from what happened in https://github.com/openwrt/luci/commit/97ebdcbddb9cad76dc551086fcb887e55886a069 ?

jow-:
This time it tests for a `.lua` file extension before trying to execute a path as Lua controller. I still don't like this change though, uci depends in menus are meant for feature detection, not for real time display choices. There's other, non-uci depends such as filesystem or acl, those would still show the old caching behavior, leading to inconsistencies.

systemcrash:
> There's other, non-uci depends such as filesystem or acl, those would still show the old caching behavior, leading to inconsistencies.

Could @wryun fix those also in this PR?

jow-:
It still wouldn't change the fact that the menu would needlessly get regenerated on every single change.

systemcrash:
> It still wouldn't change the fact that the menu would needlessly get regenerated on every single change.

Neddless, perhaps. But regeneration still is useful in some cases. Can't we gate those conditions for to make triggering it more simple? Evidently there's a need.

systemcrash:
Closing this in favour of #7725

PR #7085, opened 2024-04-26
luci-mod-network: manage new packet steering script

pesa1234:
Due to the changing on script from this commit https://github.com/openwrt/openwrt/commit/7ebcf2fb9c5db8b75666761f5e767c91abb1f437 I suggest these changing to select "old style" or "new style" with steering_flow

Tested on my Gl.iNet Mt6000


modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1595,9 +1595,19 @@ return view.extend({
 			_('This prefix is randomly generated at first install.'));
 		o.datatype = 'cidr6';
 
-		o = s.option(form.Flag, 'packet_steering', _('Packet Steering'), _('Enable packet steering across all CPUs. May help or hinder network speed.'));
-		o.optional = true;
-
+		o = s.option(form.ListValue, 'packet_steering', _('Packet Steering'),_('Enable packet steering across all CPUs. May help or hinder network speed.'));
```
feckert, line 1598: please add a space after the last `,`

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1595,9 +1595,19 @@ return view.extend({
...
+				o.value('', _('Disabled'));
+				o.value('1',_('Enabled'));
+				o.value('2',_('Enabled - (old style)'));
+				o.optional = true;
```
feckert, line 1602: The indentation is not correct

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1595,9 +1595,19 @@ return view.extend({
...
+				
+		var steer_flow = uci.get('network', 'globals', 'steering_flows');	
+		
+		o = s.option(form.Value, 'steering_flows', _('Steering flows'),_('Steer up to 128 locally received flows per device to the CPU receiving them.'));
+				o.value('', _('none'));
+				o.value('128', _('Default: 128'));
+				o.depends('packet_steering', '1');
+				o.default = steer_flow;
```
feckert, line 1610: The indentation is not correct

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1595,9 +1595,19 @@ return view.extend({
...
-
+		o = s.option(form.ListValue, 'packet_steering', _('Packet Steering'),_('Enable packet steering across all CPUs. May help or hinder network speed.'));
+				o.value('', _('Disabled'));
+				o.value('1',_('Enabled'));
+				o.value('2',_('Enabled - (old style)'));
+				o.optional = true;
+				
+		var steer_flow = uci.get('network', 'globals', 'steering_flows');	
+		
+		o = s.option(form.Value, 'steering_flows', _('Steering flows'),_('Steer up to 128 locally received flows per device to the CPU receiving them.'));
```
feckert, line 1606: please add a space after the last `,`

feckert (changes requested):
Please fix this codeing style issues first.

pesa1234:
> Please fix this codeing style issues first.

Done, sorry and thanks

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1595,9 +1595,19 @@ return view.extend({
...
+		o = s.option(form.ListValue, 'packet_steering', _('Packet Steering'), _('Enable packet steering across all CPUs. May help or hinder network speed.'));
+		o.value('', _('Disabled'));
+		o.value('1',_('Enabled'));
+		o.value('2',_('Enabled - (old style)'));
 		o.optional = true;
-
+				
+		var steer_flow = uci.get('network', 'globals', 'steering_flows');	
+		
+		o = s.option(form.Value, 'steering_flows', _('Steering flows'), _('Steer up to 128 locally received flows per device to the CPU receiving them.'));
```
systemcrash, line 1606: `to the CPU receiving them` needs re-wording.

pesa1234: `to the core who is receiving`
 Better or what do you suggest? Thank you

systemcrash: Consider: 

I send this to you who is (already) receiving them (anyway).

It's effectively already said. Do you have a source that describes this? I could make a suggestion.

pesa1234: Thanks for suggestion, so, the function is ```
queues = glob(`/sys/class/net/${dev}/queues/rx-*/rps_flow_cnt`);
	for (let queue in queues) {
		if (debug || do_nothing)
			warn(`echo ${local_flows} > ${queue}\n`);
		if (!do_nothing)
			writefile(queue, `${local_flows}`);
	}
```
With the new script we can set the rps.
To understand what Felix do I have read this https://access.redhat.com/documentation/it-it/red_hat_enterprise_linux/7/html/performance_tuning_guide/sect-red_hat_enterprise_linux-performance_tuning_guide-networking-configuration_tools at RPS section 9.3.6.

Thanks again.

Comment (systemcrash) [reply modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js:1606]: First line:

used to direct packets to specific CPUs for processing

So something like:

steers up to 128 packet flows per device to specific CPUs for processing 

Does that capture the essence?
Is there some reason it's not named rps here? Steering is more generalized nowadays with multiple options for it as you see at the link.

Comment (systemcrash) [reply modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js]: If we're going to use the RPS acronym, then it needs explaining (it's not common vernacular). Have a look through the code-base and see how the mouse-overs for acronyms are handled.

e.g. `<abbr title=\"Domain Name System\">DNS</abbr>`

Comment (pesa1234) [reply modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js]: Done, thanks

**Comment by pesa1234** [comment]
Since I understood from @nbd168 that the steering_flow option is valid even if we choose the script with old style, the setting is available also with that setting.

**Review by systemcrash — APPROVED** [review]

**Comment by systemcrash** [comment]
@feckert ?

**Comment by pesa1234** [comment]
Please check now, maybe it is more clear.

File: modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js [hunk 6]
```diff
@@ -1595,9 +1595,21 @@ return view.extend({
...
+		o = s.option(form.ListValue, 'packet_steering', _('Packet Steering'), _('Enable packet steering across CPUs. May help or hinder network speed.'));
+		o.value('', _('Disabled'));
+		o.value('1',_('Enabled'));
+		o.value('2',_('Enabled (all CPUs)'));
 		o.optional = true;
-
+				
+		var steer_flow = uci.get('network', 'globals', 'steering_flows');	
+		
+		o = s.option(form.Value, 'steering_flows', _('Steering flows (<abbr title="Receive Packet Steering">RPS</abbr>)'), _('Used to direct packets to specific CPUs for processing. Select "none" or 128. When you set a value it steers up to value select of packet flows per device to specific CPUs for processing. Note: this setting is for local services on the device only (not for forwarding).'));
```
systemcrash, line 1606: Revise to:
```js
		o = s.option(form.Value, 'steering_flows', _('Steering flows (<abbr title="Receive Packet Steering">RPS</abbr>)'), _('Directs packets to specific CPUs where the local socket owner listens (the local service).') + ' ' + _('Note: this setting is for local services on the device only (not for forwarding).'));
```
Then it is abundantly clear, I hope.

systemcrash (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1595,9 +1595,21 @@ return view.extend({
...
+		o = s.option(form.ListValue, 'packet_steering', _('Packet Steering'), _('Enable packet steering across CPUs. May help or hinder network speed.'));
+		o.value('', _('Disabled'));
+		o.value('1',_('Enabled'));
+		o.value('2',_('Enabled (all CPUs)'));
 		o.optional = true;
-
+				
+		var steer_flow = uci.get('network', 'globals', 'steering_flows');	
+		
+		o = s.option(form.Value, 'steering_flows', _('Steering flows (<abbr title="Receive Packet Steering">RPS</abbr>)'), _('Directs packet flows to specific CPUs where the local socket owner listens (the local service).') + ' ' + _('Note: this setting is for local services on the device only (not for forwarding).'));
```
systemcrash, line 1606: Sorry, @pesa1234 could you split those into different lines e.g.:

```js
		o = s.option(form.Value, 'steering_flows', _('Steering flows (<abbr title="Receive Packet Steering">RPS</abbr>)'),
			_('Directs packet flows to specific CPUs where the local socket owner listens (the local service).') + ' ' +
			_('Note: this setting is for local services on the device only (not for forwarding).'));
```

systemcrash (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
systemcrash: then we're done here.

systemcrash: There was a FP but no change :/

systemcrash:
Closed via c433e9f2c17476996e222ca8de2738cc9ea1aeff

systemcrash:
Thanks @pesa1234!

systemcrash:
> I'm also a native English speaker, but I understood what it meant. However, I can see how some people might be a little confused, although I still don't think @pesa1234's version is correct. And I think it's now the only setting in LuCI that has recommendations/suggestions built into an option? And it also allows you to input text into something that's only supposed to accept numbers.

Locally received. As opposed to what? Remotely received? Locally received is always true. Where else does a local system receive? Anyway. It's in now, and it's clear. What would be an improvement over what we have now?

PR #7086, opened 2024-04-26
luci-mod-network: add support for `netifd` bonding device 

hitech95:
This pull request addresses a gap in Luci's functionality by introducing support for device type bonding configuration.
While Netifd already incorporates this capability, Luci currently lacks the means to manage it effectively through its web interface.

immagine
immagine


systemcrash:
hi @hitech95 thanks for this PR. It looks useful. It will certainly satisfy a number of users who have asked for it.

I know that e.g. @jow- has a deeper understanding of this (and opinions on why it might (not) work).

In the meantime, search through the other PRs for `bonding`.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -519,7 +520,389 @@ return baseclass.extend({
...
+				return 'mii';
+			}
+		};
+		o.depends('type', 'bonding');
+
+		// monitor_interval
+		o = this.replaceOption(s, 'devadvanced', form.Value, 'monitor_interval', _('Monitor Interval'));
+		o.description = _('Specifies the link monitoring frequency in milliseconds');
+		o.placeholder = '100';
+		// FIXME - we have to put a limit/range here somehow?
```
systemcrash, line 829: Example
```js
		o.datatype = 'range(0,500)';
```

hitech95: I know that I can use the `range `method to specify the range as I already did on other controls, but the limit **might** depends on what you have selected as the preferred monitor system.
As the same value is used to configure MII or ARP monitoring.

See for `arp_interval` and `miimon` options in:
https://www.kernel.org/doc/Documentation/networking/bonding.txt

systemcrash: What's sensible and what's absurd here? If they depend on others, then o.datatype is inflexible. Using `o.validate` allows access to other parameters already set via e.g.
```js
var value = this ? this.section.formvalue('devadvanced', 'xyz'): 1000
```
Then.... decide based on those other values, or what's set.

hitech95:
@systemcrash I've modified the limit above, the netifd takes an integer so I've used that limit in there as I wasn't able to find the real upper limit in the kernel documentation.

I've also added dependencies for another parameter. and the "primary" interface selection for the active backup policy.

systemcrash:
@hitech95 Is this ready as-is?

@svanheule is this something you could test in a switch environment to verify bonding functionality?

systemcrash:
@jow- any opinions?

svanheule:
> @svanheule is this something you could test in a switch environment to verify bonding functionality?

I'm afraid I'm currently not set up for this and I won't have time soon either to do so.

hitech95:
> @hitech95 Is this ready as-is?
Yep it should be, this uses netifd built in bonding support instead of the external protocol.
I'm using it on my prod router machine (x86).

Tested mainly on LACP on 2 uplinks.
During initial tests I've also tested backup mode.

systemcrash:
@hitech95 If you want to get this in sooner, I recommend drumming up a bit of support in the forums and see whether anyone else can verify this and throw it into their installation. Some of the switch threads would be a good start!

systemcrash:
@svanheule wonder if your situation has changed any?

svanheule:
@systemcrash It hasn't and is unlikely to change the coming months. Most efficient way is probably to set something up yourself :)

systemcrash:
@hitech95 What device do you test on? Might be faster to find some guinea pigs in the forums who have the same device to be able to test.

hitech95:
@systemcrash 
I've it currently up and tested on x89 VM and I've tested onthe mediatek target.
That said I have it running in production with LACP to a GS1900 switch (now with 3Eth uplink).
But on the other device I have only checked that the correct parameters has been set to the network configuration.
I have not tested the netifd stuff is working as I've assumed it was working as it is included by default.

I'm gonna ask some guys if they can test it out too in the forum.

systemcrash:
I threw it out to [a forum thread](https://forum.openwrt.org/t/support-for-rtl838x-based-managed-switches/57875/2812) for Realtek (GS1900), and the word is that the driver is currently broken (but not necessarily an older version?). But others may be inclined?

hitech95:
I'm trying to get help from some friends to crete a virtial lab to test things out.
Unfortunatly we a re a bit busy with work/univeristy.

I'll update you when some more test are completed.
We can (probably) test on the follwing hardware:
 - mediatek
 - ramips
 - ath79 (not 100% sure if I have a compatible device as all my ath79 are quite old)
 - x86 (can test also with Openwrt and Debian, probably also Opnsense not sure what is supported over there)
 
 The main stopper here is the lack of vendor devices with OEM firmwares to test against. 
 (All my network devices runs openwrt with the exceptio of my GS1900)

systemcrash:
Any progress on this?

hitech95:
not jet :(

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -519,7 +520,395 @@ return baseclass.extend({
...
+				return '';
+			}
+		};
+		o.depends({'type': 'bonding', 'policy': 'balance-xor'});
+		o.depends({'type': 'bonding', 'policy': '802.3ad'});
+		o.depends({'type': 'bonding', 'policy': 'balance-tlb'});
+
+		o = this.replaceOption(s, 'devadvanced', form.Value, 'ad_actor_system', _('MAC address for LACPDUs'));
+		o.description = _('This specifies the mac-address for the actor in protocol packet exchanges (LACPDUs). The value cannot be NULL or multicast.');
+		o.datatype = 'macaddr';
```
systemcrash, line 660: Perhaps a couple of error checks here to prevent NULL and multicast.

Can we add any predefined values here? e.g. `01:80:C2:XX:XX:XX`? Or should this be the underlying MAC of the device(s) - which one could also probe for and provide?

howels:
Need to ascertain if bonding is supported on Realtek yet, then I can test here also.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
hitech95: Added to my TO-DO list

systemcrash:
Hi @hitech95. Can you tell me a bit more about this? I was thinking about merging this for 24 and see how it fares in the wild. More devices are capable of this than just switches. 

You compared it with the proto-bonding - does your change-set also require e.g. kmod-bonding? In which case we might need to check for the kmod. If it's installed, show the extra options, kind of thing.

hitech95:
Sorry for the late reply.
This uses `kmod-bonding`, and use the netifd implementation under the hood.
It uses the "device" configuration instead of the "protocol" one.

The netifd implementation (from nbd) went a bit under the radar:
See: https://github.com/openwrt/netifd/commit/5ba9744

There is also some lack of documentation in the wiki about hose new parameters.

systemcrash:
Merged. Let's see what time does with this option. :)

abceleung:
> Merged. Let's see what time does with this option. :)

Hi, would this package be included in the upcoming 24.10.0 release?

hnyman:
> be included in the upcoming 24.10.0 release?

It is already included...
Just look at the 24.10 commit history and you can find it there.

https://github.com/openwrt/luci/commits/openwrt-24.10/

PR #7087, opened 2024-04-27
luci-mod-status: Lease Time Retrieval for IPv6 Interfaces

hitech95:
This pull request introduces a streamlined method for retrieving lease time information on IPv6 interfaces. While IPv4 interfaces typically have a single lease time (available in the interface `data`), IPv6 interfaces commonly feature multiple lease times due to factors like multiple delegated prefixes or multiple standard /64 subnet.

Users typically receive only one prefix delegation from their ISP, leading this PR to the retrieval of lease time from the first available prefix or IPv6 address.

To enhance usability and provide users with a more comprehensive view of lease time information, however, alternative approaches should be discussed.

Suggestions include returning an array of lease times or aggregating the various possible values using methods like finding the minimum or maximum.

These enhancements aim to improve the clarity the lease time management on IPv6 interfaces.

immagine

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/network.js
+++ b/modules/luci-base/htdocs/luci-static/resources/network.js
@@ -2141,12 +2141,27 @@ Protocol = baseclass.extend(/** @lends LuCI.network.Protocol.prototype */ {
 	 */
 	getExpiry: function() {
 		var u = this._ubus('uptime'),
-		    d = this._ubus('data');
+		    d = this._ubus('data'),
+		    v6_prefixes = this._ubus('ipv6-prefix'),
+		    v6_addresses = this._ubus('ipv6-address');
 
-		if (typeof(u) == 'number' && d != null &&
-		    typeof(d) == 'object' && typeof(d.leasetime) == 'number') {
-			var r = d.leasetime - (u % d.leasetime);
-			return (r > 0 ? r : 0);
+		if (typeof(u) == 'number' && d != null) {
+
+			// DHCPv4 or leasetime in data
+			if(typeof(d) == 'object' && typeof(d.leasetime) == 'number') {
+				var r = d.leasetime - (u % d.leasetime);
+				return (r > 0 ? r : 0);
+			}
+
+			// DHCPv6, we can have multiple IPs and prefixes
+			if (Array.isArray(v6_prefixes) || Array.isArray(v6_addresses)) {
+				var prefixes = [...v6_prefixes, ...v6_addresses];
+
+				if(prefixes.length && typeof(prefixes[0].valid) == 'number') {
+	          		var r = prefixes[0].valid;
+          			return (r > 0 ? r : 0);
+				}
+			}
 		}
 
 		return -1;
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/30_network.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/30_network.js
@@ -21,7 +21,7 @@ function renderbox(ifc, ipv6) {
 	    active = (dev && ifc.getProtocol() != 'none'),
 	    addrs = (ipv6 ? ifc.getIP6Addrs() : ifc.getIPAddrs()) || [],
 	    dnssrv = (ipv6 ? ifc.getDNS6Addrs() : ifc.getDNSAddrs()) || [],
-	    expires = (ipv6 ? null : ifc.getExpiry()),
+	    expires = ifc.getExpiry(),
 	    uptime = ifc.getUptime();
 
 	return E('div', { class: 'ifacebox' }, [


systemcrash:
please fix the subject line of your first commit, which has expity instead of expiry

hitech95:
My bad :D

systemcrash:
In which situations might we notice a difference? Specifically multiple IPv6?

hitech95:
Right now the lease time for DHCPv6 interfaces is not shown:
`expires = (ipv6 ? null : ifc.getExpiry()),`
hard filtered here and getExpiry expect the leaste time to be present inside the `data` object of the intefrace.
For v6 this is not the case.

This add a way to return that information (by reading the correct fields), but has some limitations as a single intarface can have multiple v6 IPs/prefixes with different lases.

So I'm open for discussion on how we can handle such cases.

systemcrash:
I think at its simplest, multiple panels can be displayed - one per IPv6 subnet/IP.

hitech95:
> I think at its simplest, multiple panels can be displayed - one per IPv6 subnet/IP.

I think that we have to discuss with @jow- how to handle this.
Having multiple panels and having an array is a UI and API rework as we have to retun an array on a quite stable API.

systemcrash:
> > I think at its simplest, multiple panels can be displayed - one per IPv6 subnet/IP.
> 
> I think that we have to discuss with @jow- how to handle this. Having multiple panels and having an array is a UI and API rework as we have to retun an array on a quite stable API.

@jow- is usually up to his eyeballs in stuff to do, even though an abstract design discussion would be nice. If you're up for it, since you seem to know your way around, make a change proposal which can be reviewed - which would certainly go faster.

In the meantime we could merge this.

hitech95:
> > > I think at its simplest, multiple panels can be displayed - one per IPv6 subnet/IP.
> > 
> > 
> > I think that we have to discuss with @jow- how to handle this. Having multiple panels and having an array is a UI and API rework as we have to retun an array on a quite stable API.
> 
> @jow- is usually up to his eyeballs in stuff to do, even though an abstract design discussion would be nice. If you're up for it, since you seem to know your way around, make a change proposal which can be reviewed - which would certainly go faster.
> 
> In the meantime we could merge this.

Ok, I'll try to see what I can do without changing too APIs. I'm gonna create/think around the UI first.
But I also have another proposal too that I want to open first.

> Please include the description of the pull request in the commit description, this makes it easier to understand the change in the future if this gets merged. I don't always want to search for the pull request for the commit so that I get the change information.

Yes, my bad.

hitech95:
Pushed with the right commit description.

systemcrash:
Question: is it possible for [getExpiry()](https://github.com/openwrt/luci/pull/7087/commits/d262cda19fe5143f9d7220fd3daf89fc1ff015a8) not to return anything? Let's say we're in some fucked up state, and we got no v4 or v6 addresses. Boom. Can we also handle a default just `return ''`? Presently each return is in an if clause.

hitech95:
> Question: is it possible for [getExpiry()](https://github.com/openwrt/luci/pull/7087/commits/d262cda19fe5143f9d7220fd3daf89fc1ff015a8) not to return anything? Let's say we're in some fucked up state, and we got no v4 or v6 addresses. Boom. Can we also handle a default just `return ''`? Presently each return is in an if clause.

The current behaviour of `getExpity` is to return `-1` when no valid `data.leasetime` has been found.
This is the current way to singnal it: https://openwrt.github.io/luci/jsapi/LuCI.network.Protocol.html#getExpiry

My is logic to return only the first occurrence of leastime with the following logic:
- if there is the data.leasetime field is a number and is present calculate its real-world value and return it. (this logic is unchanged)
- if we have a prefix or address v6 (the ubus call returns an array even for v4 interfaces) i merge the two array and if I have at least one element I get the first element. For v6 the field is called "valid".
- otherwise `return -1`
 
 
 For example a ifstatus on my `LAN_6` interface (DHCPv6 client):
 ```
root@EX5601:~# ifstatus LAN_6
{
        "up": true,
        "pending": false,
        "available": true,
        "autostart": true,
        "dynamic": false,
        "uptime": 1193371,
        "l3_device": "br-lan",
        "proto": "dhcpv6",
        "device": "br-lan",
        "metric": 0,
        "dns_metric": 0,
        "delegation": false,
        "ipv4-address": [

        ],
        "ipv6-address": [
                {
                        "address": "XXXX:XXXX:XXXX:X::XXX",
                        "mask": 128,
                        "preferred": 2766,
                        "valid": 2766
                },
                {
                        "address": "fdf9:826f:b55:1::452",
                        "mask": 128,
                        "preferred": 2766,
                        "valid": 42571
                },
                {
                        "address": "XXXX:XXXX:XXXX:X:XXXX:XXXX:XXXX:XXXX",
                        "mask": 64,
                        "preferred": 2765,
                        "valid": 7057
                },
                {
                        "address": "fdf9:826f:b55:1:1271:b3ff:fe9c:9872",
                        "mask": 64
                }
        ],
        "ipv6-prefix": [

        ],
        "ipv6-prefix-assignment": [

        ],
        ....
        "data": {
                "passthru": "00170010fdf9826f0b5500010000000000000001"
        }
}        
```

**Comment by systemcrash** [comment]
So for this PR, everything is ready? The multi-panel stuff happens later, if at all.

**Comment by hitech95** [comment]
Yes for this PR it is ready. I've been a bit busy.
I hope to have a design to discuss for the end of the week.



Il mar 14 mag 2024, 14:40 Paul Donald ***@***.***> ha scritto:

**Comment by systemcrash** [comment]
OK Good! Thanks for the nice addition. Merged.

**Comment by hitech95** [comment]
The simplest way to improve the functionality would be to include an expire information for each IP that might have an expiration associated fo example:
immagine
I'm not 100% sold on this solution, I've also tried with bold "Expire" text but its even worse.

But I think that the best would be to have it inside the title something like this, maybe not in bold mut somehow marked:
immagine

I know its hard to see but one have the dotted pattern on the bottom:
immagine

I'm working on that too but as of now I dotn have any complte example to who other than this draft I'm woring on:
immagine

Using nested cards require a bigger rework as it changes the classes and the structure for themes.

All of this will assume a proper IPv6 expiry API that return for each IP its expire time in a object/map way so we can use the Ipv6 address as a key.

**Comment by tiagogaspar8** [comment]
What if you have something like:
```
Address A: xxx
Address B: xxx
Prefix Delegated: xxx
Gateway: xxx
Address A Expires: xxx
Address B Expires: xxx
Delegated Prefix Expires: xxx
Gateway Expires: xxx
```
Or:
```
Address A: xxx
Address B: xxx
Prefix Delegated: xxx
Gateway: xxx
Expires (A): xxx
Expires (B): xxx
Expires (PD): xxx
Expires (GW): xxx
```
One could argue what is actually needed for a user that looks at the interface.

(Awesome pr btw)

PR #7090, opened 2024-04-27
luci-app-banip: improve translation

stokito:
Maintainer: @dibdot

The app has a lot of options and translations for them.
This translations are precious: there are a lot of non technical people who use OpenWrt who can't even read English letters.
 
I noticed that when you changing a name of an option we need to translate all the keys again for all languages.

For example the `Countries` was changed to `Countries (RIR)`.
This is very easy to add the RIR to all existing translations with a bulk replace.
But instead I made it simpler: I moved the RIR out of translation and the key become just the  `Countries` again and old translations were saved.

Similarly we had a button `restart` that was renamed to `Restart`. While the difference only in case this made all old translations commented out.
I restored them by simply renaming their key from the `restart` to the new `Restart`.
Yes, their translations remain in lower case but I was just too lazy to upper case them.

But then you again renamed the `Restart` to `Apply & Restart`. So I just renamed the key in translation too. Users will see just the old `Restart` in their lanuage but that's really a minor thing. The Weblate must show such changed strings as "Unchecked" so translators can find them easily and easily improve to the `Apply & Restart`.
The rename wasn't really changed a meaning of the button so in this case it's totally safe to just rename its key in po files.

Another problem is that some options have the ` (default)` suffix. In fact the Luci allows to specify `o.default` so that this is not needed but creates a burden fro translators.
I fixed that too.

So please review the PR. This changes should significantly reduce amount of time for translators (e.g. me too).

Once you reviewed please merge the Weblate changes first #7084
Then I'll fix merge conflicts and we can merge it.

### DIFF
--- a/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
+++ b/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
@@ -574,7 +574,7 @@ return view.extend({
 		o.rmempty = true;
 
 		o = s.taboption('adv_log', form.ListValue, 'ban_loglimit', _('Log Limit'), _('Parse only the last stated number of log entries for suspicious events. To disable the log monitor at all set it to \'0\'.'));
-		o.value('0', _('0 (disable)'));
+		o.value('0', _('Disable'));
 		o.value('50', '50');
 		o.value('100', '100');
 		o.value('250', '250');
--- a/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
+++ b/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
@@ -319,11 +319,12 @@ return view.extend({
 		o.rmempty = true;
 
 		o = s.taboption('general', form.ListValue, 'ban_fetchretry', _('Download Retries'), _('Number of download attempts in case of an error (not supported by uclient-fetch).'));
-		o.value('1', _('1'));
-		o.value('3', _('3'));
-		o.value('5', _('5 (default)'));
-		o.value('10', _('10'));
-		o.value('20', _('20'));
+		o.value('1', '1');
+		o.value('3', '3');
+		o.value('5', '5');
+		o.value('10', '10');
+		o.value('20', '20');
+		o.default = '5';
 		o.optional = true;
 		o.rmempty = true;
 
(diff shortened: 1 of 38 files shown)


dibdot:
@stokito many thanks - LGTM! Feel free to merge whenever it suits you.

systemcrash:
@stokito - I think a more viable result for this PR set is to, as you've done correctly, 'un'translate all the number strings, but *everything* that contains the word default, remove `default` from the string, and just add it as a translatable, used everywhere. Then it's only translated once.
e.g.
```diff
-		o.value('100', _('100 (default)'));
+		o.value('100', '100' + ' ' + _('(default)'));
```

Perhaps it looks a bit messy, but it's ruthlessly efficient.

It's not a bad idea to signal to users what the default is and that it is selected.

systemcrash:
The whole 'default' string thing makes me wonder how we might signal this more efficiently within luci - just flag a string with something special and it tacks on `(default)` at the end. It seems like there is room and use for something like this. 

@jow- ?

stokito:
The default value will be pre-selected so you see it.
But if you changed the value and after some time you wish to back to a  default then you don't know the initial default.
This can be solved globally by adding the ` (Default)` on the UI form.List element. 
I'll create another PR

stokito:
Ok, the that's seems too difficult to implement because the UI controls don't see the default value. We may back to this later.

systemcrash:
While you've marked a number of `o.default = ...` it's not evident to the user what that default is, or whether it is selected, as you've correctly pointed out.

The framework default bit can be left for another rainy day. :)

dibdot:
@stokito I prepare another functional update for banIP - is it OK for you to incorporate most of your changes in my next update?

stokito:
I basically waiting for the weblate pr to get merged and then resolve conflict in the pr. After that you'll be able to merge it and proceed.

29 апр. 2024 г. 22:33:02 Dirk Brenken ***@***.***>:

stokito:
I resolved conflicts @dibdot, @systemcrash @hnyman please squash and merge before new conflicts occur

stokito:
@hnyman then we should add the ` (default)` suffix to all options in all apps. This is not so critical here to see the default.
We may solve this on the UI components level but that may need for some refactoring.

systemcrash:
> @hnyman then we should add the ` (default)` suffix to all options in all apps. This is not so critical here to see the default. We may solve this on the UI components level but that may need for some refactoring.

Welp, for now we could go with my suggestion from a few days ago. Then go for the bigger fish later.

dibdot:
@hnyman 
Maybe a better compromise ...
image

add a placeholder to select boxes with a default, e.g.:
```
o = s.taboption('general', form.ListValue, 'ban_fetchretry', _('Download Retries'), _('Number of download attempts in case of an error (not supported by uclient-fetch).'));
		o.value('1');
		o.value('3');
		o.value('5');
		o.value('10');
		o.value('20');
		o.default = '5';
		o.placeholder = _('-- back to default --');
		o.create = true;
		o.optional = true;
		o.rmempty = true;
```

PR #7092, opened 2024-04-29
luci-app-keepalived: move status page to status menu

feckert:
@jempatel 

The main change her is moving the overview page to the status menu entry.
So the status page of keepalived is new under the status menu pages.
Old:
old

New:
new

While we are at it also add the missing cbi CSS class so it looks better on luci-theme-material and add a missing cbi-descr.

### DIFF
--- a/applications/luci-app-keepalived/htdocs/luci-static/resources/view/keepalived/overview.js
+++ b/applications/luci-app-keepalived/htdocs/luci-static/resources/view/keepalived/overview.js
@@ -48,14 +48,14 @@ return view.extend({
 								}
 							}
 						}
-						return  [ 
+						return  [
 							target.data.iname,
 							target.data.ifp_ifname,
 							state,
 							target.stats.advert_sent,
 							target.stats.advert_rcvd,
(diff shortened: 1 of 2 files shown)


systemcrash:
@feckert was your new image supposed to be of VRRP? :)

feckert:
@systemcrash Yes this is intentional. I think it makes more sense to add into the menu the use case and not the name of the implementation.

systemcrash:
I see. A bit confusing. VRRP is a protocol. Keepalived is... a daemon. Yours might be better. Alas, I don't use either of those actively.

systemcrash:
So users who just installed this keepalive thing naturally look in the menus for keepalive in status, and may be confused when they do not find it.

feckert:
I don't see it that way. If someone installs the keepalived, then they must already know that this has to do with VRRP and will also find this in the GUI?

I have also been working with the [Opensense](https://docs.opnsense.org/manual/gui.html) GUI recently. And it is very structured.I miss this a bit with OpenWrt and would like to improve this step by step.
It can't hurt to do it  the same as other systems do it.

PR #7093, opened 2024-04-29
luci-app-usteer: Allow editing of settings if usteer is not running

Ramon00:
Apparently, an invalid parameter stop usteer from starting.  This PR allows the editing of parameters while usteer is stopped. In that case the status tab is not shown.

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -368,13 +368,15 @@ return view.extend({
 
 		if (!('usteer' in data[0])) {
 			m = new form.Map('usteer', _('Usteer'),
-				_('Usteer is not running. Make sure it is installed and running.') +
+				_('Usteer is not running. Make sure it is installed and running.') +' '+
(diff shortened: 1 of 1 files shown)


Jonsonmon:
Very important PR

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -368,13 +368,15 @@ return view.extend({
 
 		if (!('usteer' in data[0])) {
 			m = new form.Map('usteer', _('Usteer'),
-				_('Usteer is not running. Make sure it is installed and running.') +
+				_('Usteer is not running. Make sure it is installed and running.') +' '+
+				_('An incorrect parameter can cause usteer to fail to startup.') +' '+
```
systemcrash, line 372: It should at least be `to start up`

Ramon00: ok

systemcrash:
Good. Thanks @Ramon00 !

Ramon00:
thx @systemcrash

PR #7096, opened 2024-04-30
luci-mod-network: steering flow - permit only integer value

pesa1234:
An user can insert a strange value right now, with this commit, only integer value can be insert.

Tested working ok


systemcrash:
Change commit title to: `luci-mod-network: steering flow - permit only integer value`

Issue #5568, opened 2021-12-02
Add auto refresh to System Log and Kernel Log pages


shoulders:
This would be a useful feature to have instead of refreshing the page manually.

It could be done in realtime (if possible) and if not you could put 

- checkbox : enable autorefresh
- number input: number of seconds between refreshes

dannil:
This was implemented with #7101

PR #7101, opened 2024-05-01
luci-mod-status: auto-refresh system log and kernel log

dannil:
Implements #5568. See attached video. I have by design not implemented any "follow tail" logic for the log as I was unsure how it should interact with the "Scroll to tail" and "Scroll to head" buttons that are already there (should it follow the tail even if the user hasn't clicked "Scroll to tail" yet?), but any suggestion how and if we should handle that is welcome.

I was a bit unsure if PKG_RELEASE should be incremented for modules; luci-mod-status hasn't received any since 938e54df17d6a48fc8d84fe5893ff67ffe4a3e6e even though the diff since that commit is quite big.

https://github.com/openwrt/luci/assets/7290072/815a0dea-fa35-4146-bba0-b0cfa2fdb87b

- [X] Tested on: (x86/64, 23.05.3, Firefox) :white_check_mark:
- [X] \( Optional ) Closes: e.g. openwrt/luci#5568

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
@@ -1,21 +1,37 @@
 'use strict';
 'require view';
 'require fs';
+'require poll';
 'require ui';
 
 return view.extend({
-	load: function() {
-		return fs.exec_direct('/bin/dmesg', [ '-r' ]).catch(function(err) {
+	retrieveLog: async function() {
+		return fs.exec_direct('/bin/dmesg', [ '-r' ]).then(logdata => {
+			const loglines = logdata.trim().split(/\n/).map(function(line) {
+				return line.replace(/^<\d+>/, '');
+			});
+			return { value: loglines.join('\n'), rows: loglines.length + 1 };
+		}).catch(function(err) {
 			ui.addNotification(null, E('p', {}, _('Unable to load log data: ' + err.message)));
 			return '';
 		});
 	},
 
-	render: function(logdata) {
-		var loglines = logdata.trim().split(/\n/).map(function(line) {
-			return line.replace(/^<\d+>/, '');
-		});
+	pollLog: async function() {
+		const element = document.getElementById('syslog');
+		if (element) {
+			const log = await this.retrieveLog();
+			element.value = log.value;
+			element.rows = log.rows;
+		}
+	},
 
+	load: async function() {
+		poll.add(this.pollLog.bind(this));
+		return await this.retrieveLog();
+	},
+
+	render: function(loglines) {
 		var scrollDownButton = E('button', { 
 				'id': 'scrollDownButton',
 				'class': 'cbi-button cbi-button-neutral',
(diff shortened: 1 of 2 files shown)


systemcrash:
While I like this concept, something like this should be configurable, since everyone is used to today's behaviour: you load the page and you can stare at those logs for hours and nothing changes. Both behaviours are valuable: seeing live action, and freezing a log to understand something better.

I feel a log config tab coming on. Where one can have `tail logs` which chases output, perhaps among other configurables. 

Doable?

feckert:
@systemcrash In my opinion, a configuration option is not necessary. We'll see if it bothers users. There is already a "Refresh" button in the view. If someone does not like this new feature, it can be disabled by pressing this button.

I think this could be merged.

systemcrash:
@feckert there is no refresh button here...? So I would prefer not to introduce a behaviour change without an option to control it.

dannil:
> @feckert there is no refresh button here...? So I would prefer not to introduce a behaviour change without an option to control it.

I think @feckert means the blue "Refreshing" text in the upper rightcorner  of LuCI, if you click that you disable the system-wide polling (technically everything that hasn't overriden the default timeout of the `poll` library), which includes the system log, that's why I didn't take into consideration of adding that as an option.

Screenshot_29
Screenshot_30

feckert:
> > @feckert there is no refresh button here...? So I would prefer not to introduce a behaviour change without an option to control it.
> 
> I think @feckert means the blue "Refreshing" text in the upper rightcorner of LuCI, if you click that you disable the system-wide polling (technically everything that hasn't overriden the default timeout of the `poll` library if I remember correctly), which includes the system log in this diff, that's why I didn't at first think of taking that into consideration.
> 

@dannil  yes this is exactly what i mean. Thanks for the clarification.
@systemcrash I think that fulfills the requirement.
The behavior can be set as before when the logs were not refreshed.
I think this could be merged?

dannil:
> > > 
> > 
> > 
> > I think @feckert means the blue "Refreshing" text in the upper rightcorner of LuCI, if you click that you disable the system-wide polling (technically everything that hasn't overriden the default timeout of the `poll` library if I remember correctly), which includes the system log in this diff, that's why I didn't at first think of taking that into consideration.
> 
> This simply does not exist on the log pages today, at least not for me, OpenWrt 23.05.3 r23809-234f1a2efa. In fact, none of the status pages except for real+time graphs show this bad boy:
> 
> > Screenshot_29 Screenshot_30
> 
> So if at least that can be added for those pages, then this is good to go.

It will be added with this PR. How the flow works is that the [start method of the poll class](https://github.com/openwrt/luci/blob/eee323cb6683e6f1f1da7059e49eb8f4498bedf6/modules/luci-base/htdocs/luci-static/resources/luci.js#L410-L423) dispatches the `poll-start` event. The [setupDOM function](https://github.com/openwrt/luci/blob/68b2ce84ece62e3aec98b9be31e61ec5f6cdba2d/modules/luci-base/htdocs/luci-static/resources/luci.js#L2618-L2626) listens to this event dynamically, and if it discovers there's such an event dispatched, it dynamically adds the "Refreshing" button/indicator (whatever you wanna call it), enabling users to stop the polling themselves. 

This is why some pages (such as Status -> Overview and System -> System) gets a refreshing indicator dynamically and others don't, even though those pages themselves doesn't have code to add the indicator. I verified this by swapping between this branch and `openwrt-23.05`, deploying `luci-mod-status` in between; the indicator disappeared in `openwrt-23.05` but showed up on this branch (due to the usage of the `poll` class).

systemcrash:
Ah good! I learned something new :)

feckert:
@dannil Thanks merged!

systemcrash:
Thanks @dannil

PR #7104, opened 2024-05-01
[RFC] luci-app-lxc: remove bypassing GPG validation

### LINKED ISSUE #7098: luci-app-lxc: Impossible to find any container templates for aarch64 targets
## Steps to reproduce:
On an aarch64 board, go to: Services→ LXC Containers → Template

## Actual behavior:
Even though the default repo at images.linuxcontainers.org has aarch64 templates the following message is displayed:
"There are no templates for your architecture (aarch64) available, please select another containers URL."

## Expected behavior:
Users should be able to select templates for the arm64 (as listed in the repo) target architecture

## Additional Information:
Command lxc-create works normally when the architecture is set to arm64 (instead of aarch64 that luci-app-lxc uses).

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='SNAPSHOT'
DISTRIB_REVISION='r26116-0fe8a215e3'
DISTRIB_TARGET='rockchip/armv8'
DISTRIB_ARCH='aarch64_generic'
DISTRIB_DESCRIPTION='OpenWrt SNAPSHOT r26116-0fe8a215e3'
DISTRIB_TAINTS='no-all busybox'
```

dannil:
Fixes #7098. Ping @xaviergr. 

Since https://github.com/lxc/lxc/commit/58520263041b6864cadad96278848f9b8ce78ee9 which is part of LXC 5.0, not being able to bypass GPG validation anymore has been enforced by upstream in lxc-download. When LXC was bumped to 5.0.1 in https://github.com/openwrt/packages/commit/d957a2293b2a21b1edca1aa92e141bad8292251a this broke passing the --no-validate option to lxc-download, which resulted in getting the output of the --help command line option as the RPC output of lxc-download as it doesn't understand the --no-validate option. This in turn broke the string parsing for compatible distributions and their versions.

To solve this, this commit removes the --no-validate option entirely as it has been removed by upstream.

I've put it as RFC as I don't use LXC myself so don't have a good way of verify that it actually works, but with these changes I can at least download the available templates and create LXC containers properly so I guess it works? The output of the current lxc-download invocation results in the following output:
```
root@labmouse:~# sh /usr/share/lxc/templates/lxc-download --list --no-validate --server images.linuxcontainers.org 2>/dev/null
LXC container image downloader

Special arguments:
[ -h | --help ]: Print this help message and exit
[ -l | --list ]: List all available images and exit

Required arguments:
[ -d | --dist <distribution> ]: The name of the distribution
[ -r | --release <release> ]: Release name/version
[ -a | --arch <architecture> ]: Architecture of the container

Optional arguments:
[ --variant <variant> ]: Variant of the image (default: "default")
[ --server <server> ]: Image server (default: "images.linuxcontainers.org")
[ --flush-cache ]: Flush the local copy (if present)
[ --force-cache ]: Force the use of the local copy even if expired

LXC internal arguments (do not pass manually!):
[ --name <name> ]: The container name
[ --path <path> ]: The path to the container
[ --rootfs <rootfs> ]: The path to the container's rootfs
[ --mapped-uid <map> ]: A uid map (user namespaces)
[ --mapped-gid <map> ]: A gid map (user namespaces)
```

Without the --no-validate option:
```
root@labmouse:~# sh /usr/share/lxc/templates/lxc-download --list --server images.linuxcontainers.org 2>/dev/null
Downloading the image index

---
DIST    RELEASE ARCH    VARIANT BUILD
---
almalinux       8       amd64   default 20240430_23:08
almalinux       8       arm64   default 20240430_23:08
almalinux       9       amd64   default 20240430_23:08
almalinux       9       arm64   default 20240430_23:08
alpine  3.16    amd64   default 20240501_13:00
<... snip here, a huge result of distros and versions... />
```

This bug is present in 23.05 (which I run on my dev machine which had the issue in #7098 but for x86/64) due to https://github.com/openwrt/packages/commit/d957a2293b2a21b1edca1aa92e141bad8292251a being part of master when that was branched so we may need a cherry-pick, unsure tho, you maintainers know best here ;)

Also, a general question: I discovered that the _major bump_ of lxc to v5 was backported to 22.03 in https://github.com/openwrt/packages/commit/29fdb5086dfe24f255238d471b345332dff705c2, is this a common practice to backport major bumps of libraries (which means the API can break completely) or did it just not pass quality checks properly in this case? 

- [X] Tested on: (x86/64, 23.05.3, Firefox) :white_check_mark:
- [X] \( Preferred ) Mention: @ the original code author for feedback @danrl 
- [X] \( Optional ) Closes: e.g. openwrt/luci#7098

### DIFF
--- a/applications/luci-app-lxc/luasrc/controller/lxc.lua
+++ b/applications/luci-app-lxc/luasrc/controller/lxc.lua
@@ -55,11 +55,10 @@ end
 
 function lxc_get_downloadable()
 	local target = lxc_get_arch_target(url)
-	local ssl_status = lxc_get_ssl_status()
 	local templates = {}
 
-	local f = io.popen('sh /usr/share/lxc/templates/lxc-download --list %s --server %s 2>/dev/null'
-		%{ ssl_status, url }, 'r')
+	local f = io.popen('sh /usr/share/lxc/templates/lxc-download --list --server %s 2>/dev/null'
+		%{ url }, 'r')
 	local line
 	for line in f:lines() do
 		local dist, version, dist_target = line:match("^(%S+)%s+(%S+)%s+(%S+)%s+default%s+%S+$")
@@ -80,10 +79,9 @@ function lxc_create(lxc_name, lxc_template)
 		return
 	end
 
-	local ssl_status = lxc_get_ssl_status()
 	local lxc_dist, lxc_release = lxc_template:match("^(.+):(.+)$")
-	luci.sys.call('/usr/bin/lxc-create --quiet --name %s --bdev best --template download -- --dist %s --release %s --arch %s --server %s %s'
-		%{ lxc_name, lxc_dist, lxc_release, lxc_get_arch_target(url), url, ssl_status })
+	luci.sys.call('/usr/bin/lxc-create --quiet --name %s --bdev best --template download -- --dist %s --release %s --arch %s --server %s'
+		%{ lxc_name, lxc_dist, lxc_release, lxc_get_arch_target(url), url })
 
 	while (nx.fs.access(path .. lxc_name .. "/partial")) do
 		nx.nanosleep(1)
@@ -179,14 +177,4 @@ function lxc_get_arch_target(url)
 		end
 	end
 	return target
-end
-
-function lxc_get_ssl_status()
-	local ssl_enabled = uci:get("lxc", "lxc", "ssl_enabled")
-	local ssl_status = "--no-validate"
-
-	if ssl_enabled and ssl_enabled == "1" then
-		ssl_status = ""
-	end
-	return ssl_status
 end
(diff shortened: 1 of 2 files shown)


systemcrash:
Good work @dannil !

Seems like an upstream miss not to turn the `--no-verify` option into a no-op and just eat it.

Anyway, looks like lxc in 23 is at 5.0.3. So this, if it provides relief to lxf users, should go in.

systemcrash:
Fix your SOB :)

systemcrash:
So this should be cherry-picked to 22 and 23, ya?

systemcrash:
Thanks @dannil !

dannil:
> So this should be cherry-picked to 22 and 23, ya?

Yes exactly, every branch that runs LXC 5.0.0 or newer (which is 22 and 23 in addition to master) needs to have it picked.

PR #7109, opened 2024-05-03
luci-app-sqm: remove deprecated ucitrack

qosmio:
Package sqm-scripts already uses procd to manage uci config change triggers, so ucitrack is not necessary.


feckert:
The current `sqm-scripts` which used in the master branch version switched to `procd` .
See: https://github.com/tohojo/sqm-scripts/blob/33a89d8e7f6c82acc0adced84cf5cd95447e6bb5/platform/openwrt/sqm-init

feckert:
@qosmio Thanks merged!

PR #7117, opened 2024-05-12
luci-base: Do not error out if getDevices returns null

lu-zero:
This is potentially either a fix or a workaround for Batman-adv interfaces that are not bound to a device but linked to an wireless device.

Probably it is cleaner to make a `batadv_vlan.js` that returns an empty list or return the empty list if the js file looked up is missing.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/tools/widgets.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/widgets.js
@@ -17,6 +17,15 @@ function getGroups() {
     });
 }
 
+function getDevices(network) {
+    if (network.isBridge()) {
+        var devices = network.getDevices();
+        return devices ? devices : [];
+    } else {
+        return L.toArray(network.getDevice());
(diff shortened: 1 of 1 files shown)


jow-:
What exact problem is this PR fixing?

lu-zero:
Since other protocols return null, I guess this might work better in the end. Please advise if `network.getDevices()` must not return null and the problem is somewhere else.

stachdude:
@lu-zero thanks for working on this, been struggling with that for a while now.

lu-zero:
another way to avoid it is to not declare the batadv interfaces bridges.

systemcrash:
Please fix your SOB line in your commit message.

lu-zero:
Sure, but is this change acceptable or the batadv interface should never be marked as bridge?

systemcrash:
I'm not expert enough in bat stuff to answer that, but this is still a useful helper function. :)

PR #7118, opened 2024-05-12
luci-proto-batman-adv: Add minimal support for batadv_vlan

lu-zero:
Make so getDevice for the interface does not return null.

See https://github.com/onemarcfifty/luci-proto-batman-adv/issues/1

### DIFF
--- a/protocols/luci-proto-batman-adv/htdocs/luci-static/resources/protocol/batadv_vlan.js
+++ b/protocols/luci-proto-batman-adv/htdocs/luci-static/resources/protocol/batadv_vlan.js
@@ -1,0 +1,50 @@
+'use strict';
+'require uci';
+'require form';
+'require network';
+
+network.registerPatternVirtual(/^bat.+$/);
+
+return network.registerProtocol('batadv_hardif', {
+	getI18n: function() {
+		return _('Batman Interface');
+	},
+
+	getIfname: function() {
+		return this._ubus('l3_device') || this.sid;
+	},
+
+	getOpkgPackage: function() {
+		return 'kmod-batman-adv';
+	},
+
+	isFloating: function() {
+		return false;
+	},
+
+	isVirtual: function() {
+		return false;
+	},
+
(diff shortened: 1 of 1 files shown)


protocols/luci-proto-batman-adv/htdocs/luci-static/resources/protocol/batadv_vlan.js
```diff
@@ -0,0 +1,58 @@
...
+	containsDevice: function(ifname) {
+		return (network.getIfnameOf(ifname) == this.getIfname());
+	},
+	
+	renderFormOptions: function(s) {
+		var dev = this.getL3Device() || this.getDevice(), 
+			o;
+
+		o = s.taboption('general', form.ListValue, 'master', _('Batman Device'),
+				_('This is the batman-adv device where you want to link the physical Device from above to. If this list is empty, then you need to create one first. If you want to route mesh traffic over a wired network device, then please select it from the above Device selector. If you want to assign the batman-adv interface to a Wi-fi mesh then do not select a Device in the Device selector but rather go to the Wireless settings and select this Interface as a network from there.'));
```
systemcrash, line 42: Please rethink this mouthful. And never use "where you want to" or triple prepositions "from above to" - just.. what? Split this into separate translatable sentences. e.g. `_('...') + ' ' + _('...')` 

You should never have to say "then please select it from the above Device selector" if everything's function is clear. Make it clear.

systemcrash (changes requested):

protocols/luci-proto-batman-adv/htdocs/luci-static/resources/protocol/batadv_vlan.js
lu-zero: sorry that was copied and it is needless.

systemcrash:
Anyone else use batman?

lu-zero:
Actually looking at the other protocols getDevices returns null as well...

PR #7119, opened 2024-05-12
luci-mod-system: flash.js: add mtdblock name to filename

skramstad:
When you download a partition, the name is `hostname+mtdX.bin` and that can be confusing if you go back and try to remember what `mtdX` was if you have 10+ partitions.

So let us add the partition name to the filename like: 
```
hostname.mtdX.boot.bin
hostname.mtdX.firmware.bin
```
Note: If there is a better, easier and simpler way to do the same thing, please let me know.

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
@@ -170,15 +170,17 @@ return view.extend({
 	},
 
 	handleBlock: function(hostname, ev) {
-		var mtdblock = dom.parent(ev.target, '.cbi-section').querySelector('[data-name="mtdselect"] select').value;
+		var mtdblock = dom.parent(ev.target, '.cbi-section').querySelector('[data-name="mtdselect"] select');
+		var mtdnumber = mtdblock.value;
+		var mtdname = mtdblock.selectedOptions[0].text;
 		var form = E('form', {
 			'method': 'post',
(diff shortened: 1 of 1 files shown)


systemcrash:
I like this. Is there a way we could also get the device type in the file name which is trivially within reach?

skramstad:
> I like this. Is there a way we could also get the device type in the file name which is trivially within reach?

Sorry, but I'm not sure. I'm not so familiar with Luci.

modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
```diff
@@ -170,15 +170,17 @@ return view.extend({
 	},
 
 	handleBlock: function(hostname, ev) {
-		var mtdblock = dom.parent(ev.target, '.cbi-section').querySelector('[data-name="mtdselect"] select').value;
+		var mtdblock = dom.parent(ev.target, '.cbi-section').querySelector('[data-name="mtdselect"] select');
+		var mtdblockValue=mtdblock.value;
```
jow-, line 174: Please adhere to the existing code style and put spaces around `=`. Also consider renaming the variables to reflect their meaning, e.g. `mtdnumber` and `mtdname`.

skramstad: Thanks @jow-. Added spaces and changed to `mtdnumber` and `mtdname` as requested.

PR #7121, opened 2024-05-13
luci-app-adblock-fast: update to 1.1.2-1

stangri:
Maintainer: me
Compile tested: x86_64, Sophos XG-135r3, OpenWrt 23.05.3
Run tested: x86_64, Sophos XG-135r3, OpenWrt 23.05.3

Description:
* update license
* update PKG_VERSION/PKG_RELEASE
* improve visibility in menuconfig
* add warning for compatiblity check fails with principal package
* report compatiblity check fails with principal package from RPCD script

Depends on https://github.com/openwrt/packages/pull/24138


dannil:
@stangri is `1.1.2-1` a valid APK version? I found https://github.com/openwrt/openwrt/issues/15656#issuecomment-2155761936 and assumed it should be `1.1.2-r1` but maybe it's another issue?

PR #7124, opened 2024-05-15
luci-mod-system: flash.js: fix mtdname on download

skramstad:
The previous change did not work as intended for
partitions with `_` in the name. I.e `0:qsee_1`.
It would output an error `Invalid characters in filename`.

So fix this by matching and replacing any
character except for `a-z` and `0-9` with `-`.

Example `0:qsee_1` = `0-qsee-1`.

Tested on WAX620 with partition name `0:qsee_1`.

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
@@ -172,7 +172,7 @@ return view.extend({
 	handleBlock: function(hostname, ev) {
 		var mtdblock = dom.parent(ev.target, '.cbi-section').querySelector('[data-name="mtdselect"] select');
 		var mtdnumber = mtdblock.value;
-		var mtdname = mtdblock.selectedOptions[0].text;
+		var mtdname = mtdblock.selectedOptions[0].text.replace(/([^a-zA-Z0-9]+)/g, '-');
 		var form = E('form', {
 			'method': 'post',
 			'action': L.env.cgi_base + '/cgi-download',


skramstad:
@jow- After some more testing I found out an issue with `_` partition names. 
Sorry about that, so I made a fix. 

But if there's a better way, please do share.

jow-:
You probably should handle uppercase letters as well, iirc there's some vendor mtd partitions out there having capitalized names.

Ansuel:
@jow- yep in qcom device using SMEM AFAIK all the partition are uppercase 0:TEST 1:FOO

skramstad:
Thanks for the feedback. Added `A-Z`.

systemcrash:
> The previous change did not work as intended for partitions with `_` in the name. I.e `0:qsee_1`. It would output an error `Invalid characters in filename`.

At what point, line or so, do you get the error?

skramstad:
> > The previous change did not work as intended for partitions with `_` in the name. I.e `0:qsee_1`. It would output an error `Invalid characters in filename`.
> 
> At what point, line or so, do you get the error?

I get `Invalid characters in filename` in a new window, nothing else. This is when I try to download a partition with name `0:qsee_1`. Nothing else is in the output, so hard for me to know what's really going on.

systemcrash:
What happens in the script console? Right click -> inspect -> ...

skramstad:
> What happens in the script console? Right click -> inspect -> ...

There's nothing I can see in the script. What I can see is that it's doing a POST to `/cgi-bin/cgi-download` and in return we get `Error 400 Invalid characters in filename`.
The post code comes from this:
```
var form = E('form', {
	'method': 'post',
	'action': L.env.cgi_base + '/cgi-download',
	'enctype': 'application/x-www-form-urlencoded'
}, [
	E('input', { 'type': 'hidden', 'name': 'sessionid', 'value': rpc.getSessionID() }),
	E('input', { 'type': 'hidden', 'name': 'path',      'value': '/dev/mtdblock%d'.format(mtdnumber) }),
	E('input', { 'type': 'hidden', 'name': 'filename',  'value': '%s.mtd%d.%s.bin'.format(hostname, mtdnumber, mtdname) })
]);

ev.currentTarget.parentNode.appendChild(form);

form.submit();
form.parentNode.removeChild(form);
```
Is this the problem?
```
E('input', { 'type': 'hidden', 'name': 'filename',  'value': '%s.mtd%d.%s.bin'.format(hostname, mtdnumber, mtdname) })
```

systemcrash:
Your solution will work, although filenames are just local semantics. It would seem that the problem is in URL-encoding. We are after all using `	'enctype': 'application/x-www-form-urlencoded'`

If an underscore is in the filename, the whole filename probably needs URL encoding. IIRC, There are helper functions to do this. [See this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI).

systemcrash:
See also [this bit](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST).

jow-:
The error is omitted by cgi-io, it only allows certain characters for the content-disposition (filename) header.

skramstad:
So is there anything I need to do? Or is the `.replace` OK?

jow-:
One minor thing, with such replacements I prefer to use `[a-zA-Z0-9]+` to replace multiple consecutive invalid characters with just one replacement character.

jow-:
Yes I meant emitted, bad phone auto correction. Using urlencoding will not help, invalid characters are tested after the filename parameter has been fully decoded.

systemcrash:
Merged. Thanks @skramstad!

Issue #7137, opened 2024-05-26
[Feature request] Easily open an access to Luci from the Internet


stokito:
Today it's not so easy for a regular user to get access to OpenWrt from the internet. Usage scenarios:
1. Remote administering: SSH access, LUCI, maybe just call Wake on Lan.
2. Collecting metrics and health checks: maybe just some endpoints.
3. Self-hosting services
   * Files: SFTP to files on a disk, WebDAV, NextCloud, Syncthing
   * Apps: Transmission, IoT hubs like HomeAssistant, email server, Docker-based services, proxy
   * VPNs

They all may have different requirements and policies.
For example, if you have a Transmission UI on a separate port, then it should be generally ok to allow the WAN access.
But the WAN access to Luci is potentially more dangerous.
A user may want to add some restrictions like:
 * IP allowlist or banip
 * Use upfront external WAF and DDoS prevention services Cloudflare Proxy
 * Use a Zero Trust VPNs: Cloudflare Zero Trust, ZeroTier, Tailscale etc.
 * Tunnel services to bypass NAT also adds an encryption: SSH tunnels or a Tor onion service.

Some router manufactures already provide a GUI wizard to enable access from internet:
* [Keenetic](https://help.keenetic.com/hc/en-us/articles/360003145220-Remote-access-to-the-web-interface) has an option also to get the signed cert from LetsEncrypt
* [GL.iNet](https://docs.gl-inet.com/router/en/4/faq/remote_access_web_admin_panel/) recommends to set up a WG VPN and use it.

From the LUCI perspective, the UI should help as much as it possible to avoid misconfiguration and security disasters.
Enabling access should be easy to do to avoid mistakes at the same time a user should fully understand what are risks and how to mitigate them.

For example, opening of SSH port for WAN should be done at once click. But we should ask for a confirmation and try to force security enhancements:
* Change port or even start a separate Dropbear instance with root login disabled
* Use a pubkey auth and show an instruction how to generate a key
* (Maybe) install banip

The firewall rule to allow the SSH access from WAN must be already preinstalled but disabled by default.
This will also mitigate the mistakes of manual adding.

When allowing access to LUCI we should require HTTPS and propose to configure ACME and DDNS.
For the Self-hosting needs we'll need for some reverse proxy or use a separate port.

Many things to unpack here, but for now my proposition is to enhance documentation first, then add UCI configs disabled by default and then try to implement UI.

[...]

PR #7138, opened 2024-05-26
Dropbear: Enable SSH from WAN firewall rule

stokito:
A draft implementation of an easy way to enable the [Allow SSH from WAN](https://github.com/openwrt/openwrt/pull/15562) rule from UI. 

The whole idea is to make a Luci GUI wizard to enable WAN access https://github.com/openwrt/luci/issues/7137

Luci Dropbear enable WAN screenshot

The key idea is to have a firewall rule pre-defined and the GUI should only enable it.
This is a simplest implementation. It will work only for 22 port.
We may extend it later but it already covers the typical usage.

What is missing: we should show precautions for a users that this is not safe at all.

And same for the uhttpd:

luci uhttpd enable WAN screenshot

Here it may be not enough to just open an access. A user may want to switch IP address of uhttpd. Additionally the rfc1918 filter should be disabled. So I copied them from luci-app-uhttpd.

This needs somehow to be managed in graceful way.

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
@@ -1,10 +1,19 @@
 'use strict';
 'require view';
 'require form';
+'require uci';
 'require tools.widgets as widgets';
 
 return view.extend({
+	load: function() {
+		return Promise.all([
+			uci.load('firewall')
+		]);
+	},
+
 	render: function() {
+		var sshFirewallRule = uci.get('firewall', 'wan_ssh_allow');
+
 		var m, s, o;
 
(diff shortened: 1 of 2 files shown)


stangri:
> it already covers the typical usage.

I'm pretty sure the typical usage is to not allow SSH/HTTP/HTTPS on WAN port. Why overload UI (and risk unintentional exposure for newbies) for the very fringe use cases?

systemcrash:
> I'm pretty sure the typical usage is to not allow SSH/HTTP/HTTPS on WAN port. Why overload UI (and risk unintentional exposure for newbies) for the very fringe use cases?

I had the same thought.

Neustradamus:
@stokito: It is needed to have 2 separate rules, one for HTTP and another one for HTTPS.
It must be possible to enable HTTPS and not HTTP.

In more, it is important to choice an external port (to have not default 80/443/22):
- HTTP
- HTTPS
- SSH

PR #7140, opened 2024-05-28
luci-mod-system: Dropbear: remove custom flag values

stokito:
Instead of saving 'on' or 'off' for flag values, use the default 1 and 0.
This makes code simpler.

See related https://github.com/openwrt/openwrt/pull/15579


systemcrash:
merged, thanks!

PR #7147, opened 2024-06-01
luci-app-acme: improve UI for inexperienced users

stokito:
Maintainer: @tohojo

1. A cleanup of translations.
2. Autodetect of domain when adding
3. Import domains from DDNS
4. Logging tab

luci-app-acme screenshot

CC @hgl @stangri @systemcrash

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
@@ -89,11 +89,14 @@ return view.extend({
 		o = s.taboption('general', form.ListValue, 'validation_method', _('Validation method'),
 			_('Standalone mode will use the built-in webserver of acme.sh to issue a certificate. ' +
 				'Webroot mode will use an existing webserver to issue a certificate. ' +
-				'DNS mode will allow you to use the DNS API of your DNS provider to issue a certificate.')
+				'DNS mode will allow you to use the DNS API of your DNS provider to issue a certificate.') + '<br />' +
+			_('Validation via TLS ALPN') + ': ' + _('Validate via TLS port 443.') + '<br />' +
+			'<a href="https://letsencrypt.org/docs/challenge-types/" target="_blank">' + _('See more') + '</a>'
 		);
 		o.value('standalone', 'HTTP-01' + _('Standalone'));
 		o.value('webroot', 'HTTP-01' + _('Webroot Challenge Validation'));
 		o.value('dns', 'DNS-01 ' + _('DNS Challenge Validation'));
+		o.value('alpn', 'TLS-ALPN-01 ' + _('Validation via TLS ALPN'));
 		o.default = 'standalone';
 
 		if (!hasDnsApi) {
@@ -108,6 +111,20 @@ return view.extend({
 				window.open(link, '_blank', 'noopener');
 			};
 		}
+
+		o = s.taboption('general', form.Value, 'listen_port', _('Listen port'),
+			_('Port where to listen for ACME challenge requests. The port will be temporarily open during validation.') + '<br />' +
+			_('It may be needed to change if your web server is behind reverse proxy and uses a different port.') + '<br />' +
+			_('Standalone') + ': ' + _('Default') + ' 80.' + '<br />' +
+			_('Webroot Challenge Validation') + ': ' + _('To temporary open port you can specify your web server port e.g. 80.') + '<br />' +
+			_('Validation via TLS ALPN') + ': ' + _('Default') + ' 443.'
+		);
+		o.optional = true;
+		o.placeholder = '80';
+		o.depends('validation_method', 'standalone');
+		o.depends('validation_method', 'webroot');
+		o.depends('validation_method', 'alpn');
+		o.modalonly = true;
 
 		o = s.taboption('general', form.DynamicList, "domains", _("Domain names"),
 			_("Domain names to include in the certificate. " +


applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -86,10 +86,12 @@ return view.extend({
 
 		o = s.taboption('challenge_dns', form.ListValue, 'dns', _('DNS API'),
 			_("To use DNS mode to issue certificates, set this to the name of a DNS API supported by acme.sh. " +
-				"See https://github.com/acmesh-official/acme.sh/wiki/dnsapi for the list of available APIs. " +
+				'See %s for the list of available APIs. ' +
 				"In DNS mode, the domain name does not have to resolve to the router IP. " +
 				"DNS mode is also the only mode that supports wildcard certificates. " +
-				"Using this mode requires the acme-dnsapi package to be installed."));
+				"Using this mode requires the acme-dnsapi package to be installed.")
+				.format('<a href="https://github.com/acmesh-official/acme.sh/wiki/dnsapi" target="_blank">DNS API</a>')
```
tohojo, line 158: This means the link target won't be translated, right? That's a bit odd...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -30,6 +30,9 @@ return view.extend({
 				"point at the router's public IP address. " +
 				"Once configured, issuing certificates can take a while. " +
 				"Check the logs for progress and any errors.") + '<br/>' +
+				_('You can sign certificates for free with:') + '<br/>' +
+				'<a href="https://ZeroSSL.com/" target="_blank">ZeroSSL</a> ' + _('Default') + '<br/>' +
+				'<a href="https://LetsEncrypt.org/" target="_blank">LetsEncrypt</a> ' + '<br/>' +
```
tohojo, line 45: We override the upstream default in the acme.sh wrapper to use letsencrypt by default. The upstream change to zerossl was for commercial reasons (they are sponsoring the acme.sh project), and I personally consider it pretty dodgy. So please don't mention them here. Instead, we should write something like "letsencrypt will be used by default, but a custom ACME server can be used by specifying the `acme_server` setting".

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -524,7 +524,7 @@ return view.extend({
 		};
 
 		o = s.taboption('advanced', form.Value, "acme_server", _("ACME server URL"),
-			_('Use a custom CA instead of Let\'s Encrypt.') +	' ' + _('Custom ACME server directory URL.'));
+			_('Use a custom CA.') +	' ' + _('Custom ACME server directory URL.'));
```
tohojo: As mentioned in the other comment, letsencrypt is indeed the default, so please keep this.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -524,15 +524,17 @@ return view.extend({
 		};
 
 		o = s.taboption('advanced', form.Value, "acme_server", _("ACME server URL"),
-			_('Use a custom CA.') +	' ' + _('Custom ACME server directory URL.'));
+			_('Use a custom CA.') +	' ' + _('Custom ACME server directory URL.') + '<br />' +
+			'<a href="https://github.com/acmesh-official/acme.sh/wiki/Server" target="_blank">' + _('See more') + '</a>'
```
tohojo, line 590: Notionally, the UI is independent of underlying implementation (i.e., it could be something other than acme.sh), so we should link to a generic documentation of the ACME protocol if we are linking to anything. Or just drop this...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: Also, the wiki page you're linking to below should probably be updated to reflect this as well :)

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -63,21 +63,22 @@ return view.extend({
...
 
+		o = s.taboption('general', form.ListValue, 'validation_method', _('Validation method'),
+			_('Standalone mode will use the built-in webserver of acme.sh to issue a certificate. ' +
+				'Webroot mode will use an existing webserver to issue a certificate. ' +
+				'DNS mode will allow you to use the DNS API of your DNS provider to issue a certificate.')
+		);
+		o.value('standalone', _('Standalone'));
+		o.value('webroot', _('Webroot'));
+		o.value('dns', _('DNS'));
+		o.default = 'webroot';
```
tohojo: Why this change? I personally consider the validation method an "advanced" setting, it should be possible for a user to just specify the domain and have everything work without caring about the validation method, no?

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -78,6 +78,18 @@ return view.extend({
 				"The first name will be the subject name, subsequent names will be alt names. " +
 				"Note that all domain names must point at the router in the global DNS."));
 		o.datatype = "list(string)";
+		o.validate = function (section_id, value) {
+			if (!/^[*a-z0-9.-]*$/.test(value)) {
```
tohojo, line 130: Should probably include upper-case letters as well? They don't really matter, but seems a bit strict to force people to lower-case everything?

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -73,6 +78,19 @@ return view.extend({
...
+			let opkgPackage = 'acme-acmesh-dnsapi';
+			o = s.taboption('general', form.Button, '_install');
+			o.depends('validation_method', 'dns');
+			o.title = _('Package is not installed');
+			o.inputtitle = _('Install package %s').format(opkgPackage);
+			o.inputstyle = 'apply';
+			o.onclick = function () {
+				window.open(L.url('admin/system/opkg') +
+					'?query=' + opkgPackage, '_blank', 'noopener');
+			};
```
tohojo, line 118: While this is neat, I feel like having an actual install button goes a bit against a separation of concerns principle. We are installing certificates here, we shouldn't be installing software; so I think it's better to just have a notice and let users go to the software install page themselves to install the missing package...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -579,6 +585,15 @@ return view.extend({
 	}
 })
 
+function _isFqdn(domain) {
+	// Is not an IP i.e. starts from alphanumeric and has least one dot
+	return /[a-z0-9-]\..*$/.test(domain);
```
tohojo, line 614: I don't think this regex does what you want it to. It will match, for instance, '1.'.

Maybe something like:
`return (! /^[0-9\.]*$/.test(domain)) && /\./.test(domain);`

I.e., something that has at least one dot in it, but is not made exclusively of numbers and dots?

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -17,12 +17,15 @@ return view.extend({
 				return certs;
 			}),
 			L.resolveDefault(fs.stat('/usr/lib/acme/client/dnsapi'), null),
+			fs.lines('/proc/sys/kernel/hostname'),
```
tohojo, line 20: Shouldn't this be taken from the uci system config instead of reading it from the kernel?

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -58,6 +60,28 @@ return view.extend({
...
+				for (let ddnsDomain of ddnsDomains) {
+					let sectionId = uci.add('acme', 'cert', ddnsDomain.name);
+					uci.set('acme', sectionId, 'domains', ddnsDomain.domains);
+					uci.set('acme', sectionId, 'validation_method', 'dns');
+					uci.set('acme', sectionId, 'dns', ddnsDomain.dnsApi);
+					uci.set('acme', sectionId, 'credentials', ddnsDomain.credentials);
+				}
+				uci.save();
+				window.location.reload();
+			};
```
tohojo, line 82: While this is neat, I also worry that it's a bit on the "magic" side. At the very least, we should protect against a user accidentally clicking the import button multiple times and ending up with multiple identical configuration sections. Probably adding some help text to explain what the button does, so a user can see this before clicking it would be nice :)

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: Also, how does this work with the luci acl list (regardless of whether we're reading it from the kernel or from uci)?

tohojo (changes requested):
I generally applaud the effort to make things more user friendly, but have some comments, see below.

Also a general one: You're adding new reads to the file system but not updating the acl config accordingly. So will things actually work correctly? How did you test all the permutations?

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: Well, the idea is that if you set a custom value for the ACME server you know what you're doing and so that takes precedence over the value of `staging`. I think this is fine; but we can certainly add a warning if someone sets a custom value and also enables `staging`; but in terms of UI, I think the "custom server" field should just be hidden as an "advanced option" somewhere...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: That's a bold assumption :)
Also, further down you're doing the same with "DNS alias mode" which definitely sounds like something that should be translatable...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: Hmm, actually, even if the router does have a FQDN assigned, the system hostname is usually just the local part of it. So I'm not sure this hack is worth having at all...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: Right, but then we should explain that in the help text (with a note like "look at the sections added below after clicking this to make sure things look right" or something to that effect). And we can still check for duplicates before just adding a bunch of new sections...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: I understand what you're trying to do, I'm just saying that the regex you wrote does not accomplish this task. `/[a-z0-9-]\..*/` will match any numerical IP, for instance...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: Sure, but the domain is still the primary input, and the validation method is a detail for special cases (I don't think wildcard certificates are terribly common).

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -5,39 +5,53 @@
...
 				"certificates for the web interface. " +
 				"Note that the domain names in the certificate must already be configured to " +
 				"point at the router's public IP address. " +
 				"Once configured, issuing certificates can take a while. " +
-				"Check the logs for progress and any errors.") + '<br/>' +
-				_("Cert files are stored in") + ' <em>/etc/ssl/acme<em>'
+				"Check the logs for progress and any errors.") + '<br />' +
+				_('You can sign certificates for free with:') + '<br />' +
+				'<a href="https://LetsEncrypt.org/" target="_blank">LetsEncrypt</a> ' + _('Default') + '<br />' +
+				'<a href="https://ZeroSSL.com/" target="_blank">ZeroSSL</a> ' + '<br />' +
```
tohojo, line 45: Let's replace this with something like: "By default, the letsencrypt ACME server will be used, but you can supply a custom ACME URL to use any other provider supporting ACME certificate issuance. See %URL for a (partial) list of these"

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
systemcrash: Acronyms and abbreviations such as DNS and API almost never get translated any more because they are so ubiquitous - a quick look [through weblate](https://hosted.weblate.org/search/openwrt/?q=dns&sort_by=-priority%2Cposition&checksum=) should give you a good idea.

dannil: It's true that they don't need to be translated, however the combination of them may differ for different languages; for example in Swedish (native speaker) it would be correct to write `DNS-API` (or `DNS-värdtjänst` if you translate the API part), with a hyphen between them. So making that assumption is at least definitely not correct, however if we need to care about it that's a different question.

systemcrash: I generally agree that most stuff should be i18n able, but ossified abbreviations can be skipped.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
systemcrash: Hmm - I think it's OK. In the p910nd case I built, there was so much butt-hurt trying to include the common print modules, so to avoid clicking and typing, I added buttons, since it's conceivable that many users need this combo which is not available out of the box - all users needed one or the other module, but nobody wanted both by default at the cost of an extra few kb. Anything we can do to make a smooth experience is worthwhile IMO.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
systemcrash: This regex definitely does not do what you think it does. 
`[a-z0-9-]` matches these character ranges, but no quantifier follows it. So: 
a.xyz.com ?

Test on https://regexr.com/

There is already a hostname datatype test.

systemcrash: Might be an idea to test for Ipv4 and 6, and if not those, assume it's an fqdn instead?

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -59,25 +84,57 @@ return view.extend({
...
+			if (!value) {
+				return true;
+			}
+			if (!/^[*a-z0-9][a-z0-9.-]*$/.test(value)) {
+				return _('Invalid domain. Allowed lowercase a-z, numbers and hyphen -');
+			}
+			if (value.startsWith('*')) {
+				let method = this.section.children.filter(function (o) { return o.option == 'validation_method'; })[0].formvalue(section_id);
+				if (method && method !== 'dns') {
+					return _('A domain name with wildcard * available only when the Validation Method: DNS');
```
systemcrash, line 128: revise to:
```js
					return _('* wildcards require DNS validation');
```

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -527,6 +603,98 @@ return view.extend({
 	}
 })
 
+function _isFqdn(domain) {
+	// Is not an IP i.e. starts from alphanumeric and has least one dot
+	return /[a-z0-9-]\..*$/.test(domain) && !/[0-9-]\..*$/.test(domain);
```
systemcrash, line 608: Please provide a few examples of what should and should not match here.

systemcrash (changes requested):

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
systemcrash: Not even 0a.net will match. Test it here: https://regexr.com/

your regex currently allow
`<single character of lowercase or 0-9>.anything including spaces` and do not allow an initial wildcard.

Suggestions: 
- split on `.` and examine each array element. 
- Test for IPv4 only - bail early in that case. 
- Test for IPv6 only - bail early in that case. 

This likely will not handle punycode correctly. So your best bet is to handle cases which are not accepted and reject based on those.

systemcrash: Try:
_isFqdn('flying pig 0a.net')

systemcrash: Try also:
_isFqdn('0a.big no no')

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -46,6 +57,20 @@ return view.extend({
 		o = s.option(form.Flag, "debug", _("Enable debug logging"));
 		o.rmempty = false;
 
+		if (ddnsDomains) {
+			let ddnsDomainsList = [];
+			for (let ddnsDomain of ddnsDomains) {
+				ddnsDomainsList.push(ddnsDomain.domains[0]);
+			}
```
dannil, line 64: If you want to clean it up, I'm fine with either tho.
```js
let ddnsDomainsList = ddnsDomains.map(d => d.domains[0]);
```

tohojo:
> Chrome console: 
> _isFqdn('0a.net')
> true _isFqdn('0a')

Right, but:
_isFqdn("a0.net")
false 

Maybe something like:
```js
function _isIP(str) {
  return /^[0-9\.]+$/.test(str) || /[\[\]:]/.test(str);
}

function _isFqdn(str) {
  return !_isIP(str) && /\./.test(str);
}
```

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -648,7 +648,7 @@ function _collectDdnsDomains() {
 		if (credentials.length > 0) {
 			ddnsDomains.push({
 				sectionId: ddnsService['.name'],
-				domains: [ddnsService['domain']],
+				domains: [ddnsService['domain'], '*.' + ddnsService['domain']],
```
tohojo, line 651: Why? I don't think we should be encouraging people to randomly issue wildcard certs...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -665,6 +657,45 @@ function _collectDdnsDomains() {
 	return ddnsDomains;
 }
 
+function _importDdns(ddnsDomains) {
+	alert(_('After import check the added domain certificate configurations.'));
```
tohojo, line 661: `alert()` is not a good way to send messages to the uses. Pretty sure luci has some way of adding messages, please use that instead.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -665,6 +657,45 @@ function _collectDdnsDomains() {
...
+		let sectionId = ddnsDomain.sectionId;
+		// ensure unique sectionId
+		if (certSectionNames.has(sectionId)) {
+			sectionId += '_' + new Date().getTime();
+		}
+		if (ddnsDomain.domains) {
+			for (let d of ddnsDomain.domains) {
+				let dupDomainSection = certSectionDomains.get(d);
+				if (dupDomainSection) {
+					alert(_('The domain %s in DDNS %s already was configured in %s. Please check it after the import.').format(d, sectionId, dupDomainSection));
```
tohojo, line 683: If it's already configured we should just skip it instead of bugging the user...

stokito:
Normally the message should never occur. It would be better to let a user to know for any duplicate. Or we can just override and a user will check the Changes to apply to see what changed

stokito:
Nothing wrong with wildcards per se. Having them by default may simplify next configurations steps for new subdomains e.g. adding of nextcloud.example.duckdns.org.
An advanced user can remove the wildcard if not needed.

stokito:
Yes, the alert() is annoying, still it's much easier to implement it. Ideally a user should never see it

tohojo:
Sergey Ponomarev ***@***.***> writes:

> Nothing wrong with wildcards per se. Having them by default may simplify next configurations steps for new subdomains e.g. adding of nextcloud.example.duckdns.org.

I disagree. A wildcard certificate assigns a lot more trust to the
device than a single domain, so the principle of least privilege
dictates that these should only be used when absolutely necessary.

> An advanced user can remove the wildcard if not needed.

No, an advanced user can add it if needed :)

tohojo:
Sergey Ponomarev ***@***.***> writes:

> Yes, the alert() is anoying, still it's much easier to implement it.
> Ideally a user should never see it

No, don't offload pain on the user to make things easier to implement.
"Should never see it" is not a good argument, it will definitely happen
more than expected.

systemcrash:
Do not use alert. It's is extremely difficult to manage with keyboard
only.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
systemcrash: Please remove this `alert()`

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
systemcrash: Please remove this `alert()`

systemcrash:
Please remove `alert` and sync i18n.

systemcrash:
ping @stokito

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
stokito: Replaced with ui.showModal()

stokito:
Please re-review the PR. Now it doesn't contains changes in stings to make it easier to review and merge. I'll send another small PR for them.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
stokito: This import functionality is intended for non experienced users.
The DDNS is most likely used for personal and home usage so wildcard is not that risky. People may want to use subdomains for separate home services like HomeAssistant.my.ddns.test or NextCloud.my.ddns.test. The wildcard will significantly simplify things. If someone is paranoid they are free to edit manually the generated config.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -518,13 +509,24 @@ return view.extend({
...
 		o.modalonly = true;
 
+		o = s.taboption('advanced', form.Flag, 'staging', _('Use staging server'),
+			_(
+				'Get certificate from the Letsencrypt staging server ' +
+				'(use for testing; the certificate won\'t be valid).'
+			)
+		);
+		o.depends('acme_server', '');
+		o.depends('acme_server', 'letsencrypt');
```
tohojo: This won't work - the acmesh script has this logic:
```sh
	if [ "$acme_server" ]; then
		set -- "$@" --server "$acme_server"
	# default to letsencrypt because the upstream default may change
	elif [ "$staging" = 1 ]; then
		set -- "$@" --server letsencrypt_test
	else
		set -- "$@" --server letsencrypt
	fi
```

So if `acme_server` is set to `letsencrypt`, the `staging` field won't actually do anything.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
tohojo: I still don't think we should swap these; please drop this commit.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -16,11 +16,13 @@ return view.extend({
 				}
 				return certs;
 			}),
+			L.resolveDefault(fs.stat('/usr/lib/acme/client/dnsapi'), null),
```
tohojo: You're not adding this to the ACL file

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -573,6 +579,15 @@ return view.extend({
 	}
 });
 
+function _isFqdn(domain) {
+	// Is not an IP i.e. starts from alphanumeric and has least one dot
+	return /[a-z0-9-]\..*$/.test(domain) && !/[0-9-]\..*$/.test(domain);
+}
```
tohojo: This test doesn't work:

```js
_isFqdn("a1.com")
false 
```

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -18,6 +19,7 @@ return view.extend({
 			}),
 			L.resolveDefault(fs.stat('/usr/lib/acme/client/dnsapi'), null),
 			L.resolveDefault(fs.lines('/proc/sys/kernel/hostname'), ''),
+			L.resolveDefault(uci.load('ddns')),
```
tohojo: ...nor this

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
tohojo: That's a big assumption; those services may not even be running on the router. I don't think we should be doing the less secure thing by default. Add a note about it, sure, but don't just add the `*.`

tohojo (changes requested):

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
stokito: fixed

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
stokito: ok, removed the `*.`

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
stokito: fixed, but I used `fs.list()` instead of the `fs.stat()`

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
stokito: fixed

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
stokito: you a right, I changed it to a more simple code: just check if the domain has an alphanumeric TLD

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
stokito: Here we guessing a typical usage. In the most ideal case a user will have a DDNS configured and just click on the "Import from DDNS" and that's it.

Otherwise a user need to configure manually.
We can't move the Validation select to Advanced tab because this is the main option that shows other tabs. While the Advanced tab contains really not important options.
If a user will use the Standalone mode (default) or Webroot then nothing needs to be changed. If user will use the DNS mode then it need to select it and switch to DNS Validation tab to configure it.
The domain ideally will be pre-populated so it doesn't need to be changed. That's why it comes second after the Valitation mode selector.
Anyway, this is minor thing, I can't just drop a commit without a full rebase so let's keep it as is.

stokito:
I also added the new ALPN validation method and the listen_port option:
TLS ALPN validation

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -67,20 +86,69 @@ return view.extend({
 		o = s.taboption('general', form.Flag, "enabled", _("Enabled"));
 		o.rmempty = false;
 
+		o = s.taboption('general', form.ListValue, 'validation_method', _('Validation method'),
+			_('Standalone mode will use the built-in webserver of acme.sh to issue a certificate. ' +
+				'Webroot mode will use an existing webserver to issue a certificate. ' +
+				'DNS mode will allow you to use the DNS API of your DNS provider to issue a certificate.') + '<br />' +
+			_('Validation via TLS ALPN') + ': ' + _('Validate via TLS port 443.') + '<br />' +
```
tohojo, line 93: It's a bit weird that the first three uses "will use" and the last one uses ":" to delimit the name and description; shouldn't we be consistent here?

tohojo (changes requested):
One small nit, otherwise LGTM

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
stokito: I reused the key that is used in the option name. You are right, and we should make it consistent and also split the long description to three separate strings to simplify translation. I will do this and some other changes in a separate PR because this PR is already overloaded but also to avoid conflicts with Weblate translation.

tohojo: Alright, that's fair...

tohojo (approved):

PR #7149, opened 2024-06-04
luci-mod-system: Fix startup UI typo, update po/pot files

joshmcorreia:
Fixed a typo in `luci/modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js`

Signed-off-by: Josh Correia <joshmcorreia@gmail.com>


jow-:
Merged, thanks!

PR #7151, opened 2024-06-05
luci-base/ui.js: improve the appearance of the overlay dialogue on connectivity change

feckert:
Before with LuCI-Theme bootstrap:
before

After with LuCI-Theme bootstrap:
after

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -4734,7 +4734,7 @@ var UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 						E('h4', _('Connectivity change')),
 						E('p', _('The network access to this device could be interrupted by changing settings of the "%h" interface.').format(affected)),
 						E('p', _('If the IP address used to access LuCI changes, a <strong>manual reconnect to the new IP</strong> is required within %d seconds to confirm the settings, otherwise modifications will be reverted.').format(L.env.apply_rollback)),
-						E('div', { 'class': 'right' }, [
+						E('div', { 'class': 'center' }, [
 							E('button', {
 								'class': 'btn cbi-button cbi-button-action important',
 								'click': resolveFn.bind(null, true)


jow-:
I don't like that text duplication at all.

feckert:
@jow- 
How about this?
after

feckert:
@jow- I have now solved it like this.
after
I haven't moved the `Cancel` button all the way to the right because I don't want to change much of the CSS now.
I hope that's okay for you.

Ansuel:
I wonder can't we just put everything in a div and add the right class only to cancel button? Or put the first 2 buttons in a dedicated div with class left and cancel button with right div?

systemcrash:
> @jow- I have now solved it like this. after I haven't moved the `Cancel` button all the way to the right because I don't want to change much of the CSS now. I hope that's okay for you.

So... this is red/blue split is [bad for monochromatic colour blind users](https://www.color-blindness.com/coblis-color-blindness-simulator/). But otherwise OK. This is probably more generic to the colours in the whole theme.

I still cannot comprehend what "apply with revert after connectivity loss" entails. Beside "apply and keep settings" ... still confusing. The only thing that seems safe is just "cancel", so it feels like Cancel should get the "Safe" colour treatment, and the other two.... something more dangerous :)

jow-:
- "Apply with revert after connectivity loss" means applied changes are rolled back if there's no connection to the gui within 90 seconds after apply.
- "Apply and keep settings" means always keep and commit settings to the filesystem, even if connection to the gui is lost
- "Cancel" means do nothing and close modal

systemcrash:
Frankly, all of the text in this modal is a deathtrap.

The explanation does not seem relevant to the options presented.

Recommend:

```js
						E('h4', _('Connectivity change')),
						E('p', _('Changing "%h" settings could disable access to this device.').format(affected)),
						E('p', _('Change of IP requires <strong>connecting to the new IP</strong> within %d seconds to retain the changes.').format(L.env.apply_rollback)),
						E('p', _('Choose an option to proceed.').format(L.env.apply_rollback)),
```

> * "Apply with revert after connectivity loss" means applied changes are rolled back if there's no connection to the gui within 90 seconds after apply.

Never in a hundred years would I have inferred that meaning.

☠️

Reword: `Test apply, reverting if GUI remains unreachable`


>     * "Apply and keep settings" means settings are kept and committed to the filesystem in any case, even if connection to the gui is lost and the browser does not check in with LuCI within 90 seconds

Reword: `Apply immediately❗`

>     * "Cancel" means do nothing and close modal

Fine.

jow-:
Sounds good, go ahead and change it

systemcrash:
> Sounds good, go ahead and change it

Done. a6f8361a145d820c0820b44b9ac238b813623043. Sorry @feckert you'll need to rebase now (but that should be trivial).

PR #7152, opened 2024-06-06
luci-base: refactoring overlay for connectivity change

feckert:
Before with LuCI-Theme bootstrap:
before

After with LuCI-Theme bootstrap:
after

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -4732,22 +4732,21 @@ var UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 
 					UI.prototype.changes.displayStatus('warning', [
 						E('h4', _('Connectivity change')),
-						E('p', _('"%h" interface changes could inhibit access to this device.').format(affected)),
-						E('p', _('Any IP change requires <strong>connecting to the new IP</strong> within %d seconds to retain the changes.').format(L.env.apply_rollback)),
+						E('p', _('Changes have been made to the existing connection via "%h". This could inhibit access to this device. Any IP change requires <strong>connecting to the new IP</strong> within %d seconds to retain the changes.').format(affected,L.env.apply_rollback)),
 						E('p', _('Choose how to apply changes:')),
-						E('div', { 'class': 'right' }, [
-							E('button', {
-								'class': 'btn',
-								'click': rejectFn,
-							}, [ _('Cancel') ]), ' ',
+						E('div', { 'class': 'center' }, [
 							E('button', {
 								'class': 'btn cbi-button-action important',
 								'click': resolveFn.bind(null, true)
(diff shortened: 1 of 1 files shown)


systemcrash:
Otherwise centered is good. Just try to retain the gerunds after "Apply" so it's clear what's about to happen. Do you want the actions after apply to read similarly or something?

systemcrash:
"Apply, changes now" is odd. Two verbs, or noun verb? The other button is okay, though unreachable alone is tenuous. That button is the first time the GUI is mentioned, and it's not normally obvious that system and GUI are supposed to be client server (and the distinct and separate).

systemcrash:
Keep gerunds after Apply.

jow-:
"Apply changes now" is not entirely correct. Changes are still applied in a delayed manner, the difference is whether they'll get rolled back or not if the browser does not interact with LuCI within the next 90 seconds. Can't suggest something suitable now but the verb should not refer to the "timing" of the change application but to the "stickiness" (for lack of a better word).

systemcrash:
Yes. So "Apply, committing..." or "Apply, reverting..." pattern encompasses this distinction @jow- mentions.

jow-:
In the generic uci dialog we're using "Apply unchecked" as button label. Would that work?

feckert:
@systemcrash 
The wording `commiting` doesn't seem to fit either, see @castillofrancodamian objection from yesterday.
https://github.com/openwrt/luci/commit/a6f8361a145d820c0820b44b9ac238b813623043#r142789901

> In the generic uci dialog we're using "Apply unchecked" as button label. Would that work?
@jow- I like that.

systemcrash:
Verb + adverb (apply immediately), Verb + adjectival phrase (apply unchecked), verb + gerund (apply, doing x) all work, just a bit differently. Unchecked creates an expectation that all settings actually get checked somehow in another path, in this case the "checked" path. This construction focuses on the first action to the detriment of the successive one. (The gerund construction handles the second one.)

systemcrash:
"Simplify as much as possible, but not further" 😅

jow-:
Centering the buttons would make the layout for this modal different compared to any other modal. I'd prefer consistency here

feckert:
Here are a few suggestions.

jow-:
Actually I had it backwards, "no action, cancel" should be the lower left and "do something" the lower right corner. Pressing `Esc` in an open modal will also perform a click on the first button (which should then be the non-destructive one).

Considering this, the suggested CSS should be:

```css
.modal > .button-row {
  display: flex;
  justify-content: space-between;
}

.modal > .button-row > button:not(:last-of-type) {
  margin-right: .5em;
}

.modal > .button-row > button:first-of-type {
  margin-right: auto;
}
```

... with the result looking like this:
image

systemcrash:
These button titles are the closest to sane and understandable. 
 
The "changes have been made" is not yet concrete (are they or are they not in effect?). Changes are queued up, but are not in effect yet. Thus my changes had "could inhibit" only at the start.

PR #7162, opened 2024-06-16
luci-app-keepalived: Fix option name for script falls. 

kotyara75:
- [ ] Incremented :up: any `PKG_VERSION` in the Makefile: I didn't find where exactly.
- [X] Tested on: (MT7621, 23.05, Safari, Chrome) :white_check_mark:
- [X] Description: Before the fix, changing the 'fail' parameter for a VRRP script had no effect because keepalived expects 'fall', not 'fail'. After changing it to 'fall', it started working as expected, waiting for the given number of fails before triggering the fail state.


systemcrash:
Had to check the manual which would have been good to reference. 

https://www.keepalived.org/manpage.html

Good find - good fix!

PR #7163, opened 2024-06-16
luci-app-attendedsysupgrade: enforce correct imagebuilder version

dannil:
By passing the version_code parameter, we can error out if the imagebuilder that responded to the request isn't the same, mirroring the current behavior of auc.

Screenshot_8

This is spawned from a recent issue with the Attended Sysupgrade server as can be seen in later entries of https://forum.openwrt.org/t/auc-sysupgrade-does-not-work-for-me/196782, https://github.com/openwrt/asu/issues/848 and with the probable root cause in https://github.com/openwrt/asu/issues/853. Due to the LuCI app not having this check, the incorrect version of the imagebuilder may be used for a request, and still be returned to the user as a valid request, which (I guess?) in worst case could downgrade a user's systems instead. 

Adding this check mirrors the behaviour that `auc` currently has, see image below. The relevant code path for the Attended Sysupgrade code is at https://github.com/openwrt/asu/blob/main/asu/build.py#L72-L79, which enforces this check on the API.

Screenshot_7

- [X] Tested on: (x86/64, SNAPSHOT r26637-05aec66d53, Firefox) :white_check_mark:

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -476,6 +476,7 @@ return view.extend({
 					request: {
 						profile,
 						version: candidates[0][0],
+						version_code: revision,
 						packages: Object.keys(packages).sort(),
 					},
 				};
@@ -536,6 +537,7 @@ return view.extend({
 												...firmware,
 												packages: mapdata.request.packages,
 												version: mapdata.request.version,
+												version_code: mapdata.request.version_code,
 												profile: mapdata.request.profile
 											};
 											this.pollFn = L.bind(function () {


systemcrash:
Dan the man with the big fixes! 

Do you ha e any opinion here @aparcar?

aparcar:
Sure let's do it

dannil:
@systemcrash @aparcar I may have implemented this check incorrectly; is the `version_code` supposed to be the expected _remote_ revision for the branch you specify (23.05, SNAPSHOT etc.) and not the currently running revision? I'm looking at the API docs at https://sysupgrade.openwrt.org/ui/#/ but I'm actually not sure what the docs mean by `It is possible to send the expected revision`. 

I almost guess it should be the remote revision, since you want to verify that the imagebuilder you get back is for the version you requested to upgrade to. In that case it should not be `revision` that `version_code` is set to, but the latest remote revision, which would be the `remote_revision` variable.

systemcrash:
Let's wait to hear from @aparcar who can likely clarify.

It sounds like the revision that you are aiming for, and not the one you
are currently on. Because for all intents and purposes, I don't know of any
openwrt version which actually cares about the version you're currently on.
That might become interesting for stats at some point although nothing
more.

dannil:
After testing, I'm right now unable to upgrade SNAPSHOT through the LuCI app due to this. Looking at how firmware selector handles the `version_code` at https://github.com/openwrt/firmware-selector-openwrt-org/blob/master/www/index.js#L106, the `#image-code` selector is the latest revision (it obviously needs to be cause the firmware selector can't know the current revision cause it's not running on the device itself), so I'm 100% sure it needs to be the remote one, I've created #7166 to hopefully solve it.

Screenshot_11

efahl:
> is the `version_code` supposed to be the expected _remote_ revision for the branch you specify

Yes, from the new version to which you are upgrading.  It's in the platform file for the device, for example, open this one and look down at the bottom:

https://sysupgrade.openwrt.org/json/v1/snapshots/targets/x86/64/generic.json

PR #7184, opened 2024-07-05
luci-base: Correct how textarea's wrap works

morytyann:
- [x] Tested on: (aarch64_cortex-a53/x86, OpenWrt23.05.3, Chrome/Edge/Vivaldi) :white_check_mark:

Currently the wrap flag of textarea doesn't works properly, this PR fix it. [Reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#wrap)

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -511,7 +511,7 @@ var UITextarea = UIElement.extend(/** @lends LuCI.ui.Textarea.prototype */ {
 			'style': style,
 			'cols': this.options.cols,
 			'rows': this.options.rows,
-			'wrap': this.options.wrap ? '' : null
+			'wrap': this.options.wrap ? 'soft' : 'off'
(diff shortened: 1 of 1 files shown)


systemcrash:
@dannil @stokito I thought you might have a better understanding about this than I have.

systemcrash:
@morytyann please update your commit and initial PR message once done

dannil:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#wrap), `off` is a non-standard solution so we should at least not use that one if possible if we're fixing the wrap behavior anyway.

dannil:
> @dannil I checked and it works fine in both Chrome and FF:
> 
> https://codepen.io/pcgeek86/pen/ywgWRb

Nice, then it should be fine to merge it.

dannil (approved):

systemcrash:
Ok, thank you @dannil and @stokito!

systemcrash:
Thanks @morytyann.

PR #7187, opened 2024-07-08
luci-mod-status: display log_file in status if defined

Chris1189:
Commit includes wrapper syslog in /usr/libexec. If a log file is configured, the output of this file is displayed. Otherwise the output of logread is displayed.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
@@ -7,12 +7,11 @@
 return view.extend({
 	retrieveLog: async function() {
 		return Promise.all([
-			L.resolveDefault(fs.stat('/sbin/logread'), null),
-			L.resolveDefault(fs.stat('/usr/sbin/logread'), null)
+			L.resolveDefault(fs.stat('/usr/libexec/syslog-wrapper'), null)
 		]).then(function(stat) {
-			var logger = stat[0] ? stat[0].path : stat[1] ? stat[1].path : null;
+			var logger = stat[0].path;
 
-			return fs.exec_direct(logger, [ '-e', '^' ]).then(logdata => {
+			return fs.exec_direct(logger).then(logdata => {
 				const loglines = logdata.trim().split(/\n/);
 				return { value: loglines.join('\n'), rows: loglines.length + 1 };
 			}).catch(function(err) {
--- a/modules/luci-mod-status/root/usr/libexec/syslog-wrapper
+++ b/modules/luci-mod-status/root/usr/libexec/syslog-wrapper
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+. /lib/functions.sh
+
+main() {
+	local log_file
+
+	log_file="$(uci_get system @system[0] log_file "")"
+	if [ -f "$log_file" ]; then
+		cat "$log_file"
+		return
+	fi
+
+	if [ -f /sbin/logread ]; then
+		/sbin/logread -e ^
+	else
+		/usr/sbin/logread -e ^
+	fi
+}
+
+main
--- a/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
+++ b/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
@@ -32,10 +32,7 @@
 			"cgi-io": [ "exec" ],
 			"file": {
 				"/bin/dmesg -r": [ "exec" ],
-				"/sbin/logread": [ "stat" ],
-				"/sbin/logread -e ^": [ "exec" ],
-				"/usr/sbin/logread": [ "stat" ],
-				"/usr/sbin/logread -e ^": [ "exec" ]
+				"/usr/libexec/syslog-wrapper": [ "exec" ]
 			},
 			"ubus": {
 				"file": [ "stat" ]


modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
```diff
@@ -32,10 +32,7 @@
 			"cgi-io": [ "exec" ],
 			"file": {
 				"/bin/dmesg -r": [ "exec" ],
-				"/sbin/logread": [ "stat" ],
-				"/sbin/logread -e ^": [ "exec" ],
-				"/usr/sbin/logread": [ "stat" ],
-				"/usr/sbin/logread -e ^": [ "exec" ]
+                                "/usr/libexec/syslog/syslog": [ "exec" ]
```
systemcrash, line 35: replace all those spaces with tabs.

modules/luci-mod-status/root/usr/libexec/syslog/syslog
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
systemcrash, line 1: rework this file using tabs instead of spaces.

systemcrash:
Is it certain that the wrapper script passes the same exit codes as logread?

In general this is a good idea, although I think other devs would want to see the wrapper go in to a different repo. I don't think there's any real merit in doing so, since the GUI is the only place it gets leveraged here.

modules/luci-mod-status/root/usr/libexec/syslog/syslog
Chris1189: done and sorry, missed those spaces.

Chris1189:
> Is it certain that the wrapper script passes the same exit codes as logread?
> 
> In general this is a good idea, although I think other devs would want to see the wrapper go in to a different repo. I don't think there's any real merit in doing so, since the GUI is the only place it gets leveraged here.

This is a good point. I added a variable error_code to exit the script with the return  code from logread.

If there is the need to move the wrapper somewhere else, I will not obstruct that.

Chris1189:
I made the necessary changes, @systemcrash 
The exit codes are identical to logread.

I am looking forward to your review.
Thank you!

modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
```diff
@@ -7,12 +7,11 @@
...
 		return Promise.all([
-			L.resolveDefault(fs.stat('/sbin/logread'), null),
-			L.resolveDefault(fs.stat('/usr/sbin/logread'), null)
+			L.resolveDefault(fs.stat('/usr/libexec/syslog-wrapper'), null)
 		]).then(function(stat) {
-			var logger = stat[0] ? stat[0].path : stat[1] ? stat[1].path : null;
+			var logger = stat[0].path;
 
-			return fs.exec_direct(logger, [ '-e', '^' ]).then(logdata => {
+			return fs.exec_direct(logger).then(logdata => {
```
systemcrash, line 14: It looks like you excised the ternary operator above - which more than anything looks like a sanity check in case `stat[n]` is null. Maybe a null check and handling isn't bad here after all. We would already be in some weird place because to arrive at this state implies that the wrapper is absent for some reason. I guess it's no worse anyway, because the old 'resolves' had null as a default. What does `exec_direct` do in that case?

modules/luci-mod-status/root/usr/libexec/syslog-wrapper
```diff
@@ -0,0 +1,21 @@
...
+
+. /lib/functions.sh
+
+main() {
+	local log_file
+
+	log_file="$(uci_get system @system[0] log_file "")"
+	if [ -f "$log_file" ]; then
+		cat "$log_file"
+		return
```
systemcrash, line 11: A few points: `cat $file` and then `return`. This effectively means that `cat` return code is bubbled up from the script, right? Why do we prefer `cat` when we can do, e.g. `logread -F`? Can you test with that? Then we remain with the same tools and facilities (no risk of different return codes being passed).

modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
Chris1189: I could also bring back the ternary operator
` var logger = stat[0] ? stat[0].path : null;`,
but as you already said it would be a weird place and the Promise already returns null in that case.

**From the [docs](https://openwrt.github.io/luci/jsapi/LuCI.fs.html)**:
`exec_direct` executes the command while bypassing ubus with cgi-io helper applet at `cgi-bin/cgi-exec'. It is useful to fetch large command outputs, which may exceed the ubus message size limits. It also enforces the same access permission rules as the ubus base exec call.

modules/luci-mod-status/root/usr/libexec/syslog-wrapper
Chris1189: Sorry, but English is not my native language. The translation of your text suggests "bubble up" in the context of filter something. 
But the mere translation of "bubble up" means something like inflate.

I will show you the output of `cat file`:
grafik

Unlike `logread` this enables us to show older log messages than from the time booted.

`logread -F` does not return something but writes the content of it to the file specified after `-F`. I already tried it in the creation process, because I thought like you that the function would be cleaner.

Chris1189: Is something left to do for me?
Or can it be pulled to master?

systemcrash:
Merged. Thanks @Chris1189

systemcrash:
@stokito we determined that -F is a write only parameter, not read. Ie destination file.

is there a good way to reconcile these disparities?

systemcrash:
Perhaps it’s time for an option to read from a specific file.

So the only problem is log filtering…? The -e option only ever filters for
line start. So unless the log pages get some filter option, what’s the win?
 (Js can filter by itself without relying on underlying binaries).

jow-:
I agree with @stokito here, don't implement parallel infrastructure in the gui, don't attempt to read log files directly when cli level commands exist for that purpose.

systemcrash:
Is that what you assess has happened here? The script still leverages `logread`. But until `logread` itself gains a parameter to read from other log files, the wrapper script uses cat.

jow-:
> Is that what you assess has happened here? 

What happens here is that LuCI starts implementing log related infrastructure code, this belongs into base first.

Chris1189:
As @systemcrash mentioned `logread` only puts log messages to a specific file with `-F`.
In `/etc/config/system` the option `log_file` redirects log messages to the set file. We wanted to display contents of that file because this way we can see log messages from before boot.

From the messages you wrote to the discussion I gathered that you wish to add the filter option `-e` from logread to the syslog-wrapper?

My intent was not to implement parallel infrastructure but to add a feature. I want to read the file if it is set.

Issue #3629, opened 2020-02-14
Luci does not support dnat advanced nat configuration


pgwipeout:
Luci supports snat advanced nat configuration, but it does not support dnat.
Dnat is necessary to ensure such things as captive portals and forced dns can be properly redirected without resorting to custom rules.

jow-:
It does. All DNAT options offered by fw3 are in the Redirct Rules tab.

PR #7189, opened 2024-07-09
luci-app-firewall: allow src_dport to be empty in port forwardings (DNAT)

xrgzs:
- [x] Tested on: (x86_64, openwrt-23.05, chrome-124) :white_check_mark:

I want to `redirect` some IPs to another IP using DNAT.

https://github.com/openwrt/luci/issues/3629

The following configurations in `/etc/config/firewall` works well for me:

```
config redirect
  option name 'Hijack-CF-1'
  option src 'lan'
  option src_dip '104.21.0.0/16'
  option dest_ip '104.18.95.126'
  option target 'DNAT'
  
config redirect
  option name 'Hijack-CF-2'
  option src 'lan'
  option src_dip '172.64.0.0/13'
  option dest_ip '104.18.95.126'
  option target 'DNAT'
```

However, the front end returns error messages because of empty `src_port`, which is not necessary in my case.

So I pull this request to allow `src_port` to be set empty so I can click `Save` without any error message form luci directly.

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/forwards.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/forwards.js
@@ -278,7 +278,7 @@ return view.extend({
 		o = s.taboption('general', form.Value, 'src_dport', _('External port'),
 			_('Match incoming traffic directed at the given destination port or port range on this host'));
 		o.modalonly = true;
-		o.rmempty = false;
+		o.rmempty = true;
 		o.datatype = 'neg(portrange)';
 		o.depends({ proto: 'tcp', '!contains': true });
 		o.depends({ proto: 'udp', '!contains': true });


systemcrash:
So there's a risk that this change creates more support interactions. Instead of removing checks, counterbalance it with a checkbox which bypasses the port field and its error check instead.

systemcrash:
Also, regarding your use case: is it possible to forward one IP and all its ports to another IP to a single port? (Any) in your screenshot.

xrgzs:
> Also, regarding your use case: is it possible to forward one IP and all its ports to another IP to a single port? (Any) in your screenshot.

`Any` is not a valid value for IP port.



I tried `0-65535` and it showed up fine.



I think it should be consistent with what is in the config file, which is not "Required" in the official OpenWrt Wiki.

https://openwrt.org/docs/guide-user/firewall/firewall_configuration#options4

xrgzs:
Changing protocol to `Any` solved my problem. 

image

Then get the fixed configurations:

```
config redirect
	option name 'Hijack-CF-1'
	option src 'lan'
	option src_dip '104.21.0.0/16'
	option dest_ip '104.18.95.126'
	option target 'DNAT'
	list proto 'all'

config redirect
	option name 'Hijack-CF-2'
	option src 'lan'
	option src_dip '172.64.0.0/13'
	option dest_ip '104.18.95.126'
	option target 'DNAT'
	list proto 'all'
```

The difference form the perious one is `list proto 'all'`.

I think this pr can be closed.

xrgzs:
Changing protocol to `Any` solved my problem. 

image

Then get the fixed configurations:

```
config redirect
	option name 'Hijack-CF-1'
	option src 'lan'
	option src_dip '104.21.0.0/16'
	option dest_ip '104.18.95.126'
	option target 'DNAT'
	list proto 'all'

config redirect
	option name 'Hijack-CF-2'
	option src 'lan'
	option src_dip '172.64.0.0/13'
	option dest_ip '104.18.95.126'
	option target 'DNAT'
	list proto 'all'
```

The difference form the perious one is `list proto 'all'`.

I think this pr can be closed.

PR #7194, opened 2024-07-14
luci-proto-modemmanager: add peerdns option

memoz:
ModemManager is a layer 3 "protocol", so there should be an option to allow users to ignore operator-supplied DNS(s).

Tested on Banana PI BPI-R4 (MediaTek Filogic 880, MT7988A), current snapshot, Google Chrome 126.0.6478.127.

Question: should I touch the PKG_VERSION of luci-proto-modemmanager since there's no change there?

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -239,6 +239,7 @@ function has_peerdns(proto) {
 	case 'pptp':
 	case 'openvpn':
 	case 'sstp':
+	case 'modemmanager':
 		return true;
 	}
 


feckert:
This is only available, from my point of view, if the modemmanager is using the ppp https://github.com/openwrt/packages/blob/master/net/modemmanager/files/lib/netifd/proto/modemmanager.sh#L101C3-L101C13 to start a connection.

memoz:
> This is only available, from my point of view, if the modemmanager is using the ppp https://github.com/openwrt/packages/blob/master/net/modemmanager/files/lib/netifd/proto/modemmanager.sh#L101C3-L101C13 to start a connection.

Umm...it is a PPP connection in my case, but no authentication is required. I don't see that option in the Luci interface without this change. I'm all ears if there's a better way. Also, peerdns should apply to DHCP as well.

feckert:
I have now had a look at the modemmanager proto.
That's basically correct, but we still have to do something in modemmanger proto to get it right.

At first we have check as in the [umbim](https://github.com/openwrt/openwrt/blob/master/package/network/utils/umbim/files/lib/netifd/proto/mbim.sh#L226) and [uqmi](https://github.com/openwrt/openwrt/blob/master/package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh#L425) if `peerdns` is enabled. Currently we always set for the static part (ipv4/ipv6) [dns1-ipv4](
https://github.com/openwrt/packages/blob/master/net/modemmanager/files/lib/netifd/proto/modemmanager.sh#L184), [dns2-ipv4](
https://github.com/openwrt/packages/blob/master/net/modemmanager/files/lib/netifd/proto/modemmanager.sh#L188) , [dns1-ipv6](https://github.com/openwrt/packages/blob/master/net/modemmanager/files/lib/netifd/proto/modemmanager.sh#L250) and [dns2-ipv6](https://github.com/openwrt/packages/blob/master/net/modemmanager/files/lib/netifd/proto/modemmanager.sh#L254) independently of the `peerdns` configuration value. In order for the correct behavior and independently of ppp, static or dhcp, we would also have to check whether `peerdns` is set. And only if peerdns is set, then and only then we should set the dns1 or dns2.

memoz:
Thank you for the explanation, @feckert . My apologies, but I thought my modem (RM500Q-GL) was using the PPP method, but it turned out to be Static, and for some reason, the `peerdns` option is working.

Syslog excerpt:
> Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): IPv4 connection setup required in interface wwan: static
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding IPv4 address [redacted], netmask 255.255.255.248
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding default IPv4 route via [redacted]
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding primary DNS at [redacted]
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding secondary DNS at [redacted]
Wed Jul 17 03:55:39 2024 daemon.notice netifd: Interface 'wwan' is now up
Wed Jul 17 03:55:39 2024 daemon.notice netifd: Network device 'wwan0' link is up
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): IPv6 connection setup required in interface wwan: static
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding IPv6 address [redacted], prefix 64
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding default IPv6 route via [redacted]
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding primary DNS at [redacted]
Wed Jul 17 03:55:39 2024 daemon.notice netifd: wwan (4912): adding secondary DNS at [redacted]
Wed Jul 17 03:55:39 2024 user.notice firewall: Reloading firewall due to ifup of wwan (wwan0)
Wed Jul 17 03:55:39 2024 daemon.warn netifd: You have delegated IPv6-prefixes but haven't assigned them to any interface. Did you forget to set option ip6assign on your lan-interfaces?

`ifconfig` output:
> wwan0     Link encap:UNSPEC  HWaddr 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00
          inet addr:[redacted]  P-t-P:10.38.141.252  Mask:255.255.255.248
          inet6 addr: [redacted]/128 Scope:Global
          inet6 addr: fe80::128e:bf8d:8744:745/64 Scope:Link
          UP POINTOPOINT RUNNING NOARP MULTICAST  MTU:1400  Metric:1
          RX packets:19097 errors:0 dropped:0 overruns:0 frame:0
          TX packets:36672 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:4958872 (4.7 MiB)  TX bytes:6267750 (5.9 MiB)

By working, I mean the DNSes from the mobile network weren't picked up by Dnsmasq as indicated in `/tmp/resolv.conf.d/resolv.conf.auto`. My custom v4 and v6 servers are there instead under `# Interface wwan`. I also found the IPv4 DNSes from the mobile network in `/tmp/resolv.conf.ppp` which was modified or created around the same time as shown in the log.

I'd like to do things properly, but I'm not sure if I can get my modem to use DHCP or PPP for testing.

memoz:
Unfortunately, I accidentally tore off the lid of my RM500Q-GL when replacing the heatsink. I have no actual need for a 5G modem, so won't be buying another. Closing this pull request since I can't work on it anymore.

PR #7195, opened 2024-07-15
luci-proto-wireguard: disable peers from the tab

cryobry:
It's clumsy to enable/disable peers from the modal, save some clicks

- [:white_check_mark:] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch
- [:white_check_mark:] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [:white_check_mark:] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [:white_check_mark:] Incremented :up: any `PKG_VERSION` in the Makefile
- [:white_check_mark:] Tested on: (architecture, openwrt version, browser)
image
- [:white_check_mark:] Description: (describe the changes proposed in this PR)

When a user wants to switch a wireguard peer, currently they have to enter a modal to disable the active peer, close the modal, enter a modal to enable the inactive peer, and close the modal. This change allows users to disable/enable peers from the peers tab overview.


systemcrash:
Please revise it so that the single checkbox does not consume lots of table width. Find some value between 5-10%, and update your screenshot:

https://openwrt.github.io/luci/jsapi/LuCI.form.Value.html#width

e.g. `		o.width = "5%";`

cryobry:
The narrowest I could make it without wrapping the header was 12%:

image

systemcrash:
Wrap the header (it's OK if the header consumes more vertical space)

cryobry:
What if we switched the name from 'Peer disabled' to just 'Disabled' since it's implied by the tab name and the option description ('Enable / Disable peer.') makes it very explicit in the modal?
The vertically resized header looks bad. I can make the change, just wanted to ask before commiting.

systemcrash:
Changing disabled to enabled is not trivial: Either you have a boolean inversion in the GUI which creates confusion because it's not simple to understand, or you change all the underlying scripts in the openwrt repo (which is fucking impossible to get stuff into, for various good reasons).

Disabled was chosen such that this was possible:
```sh
	if [ "${disabled}" -eq 1 ]; then
		# skip disabled peers
		return 0
	fi
```

systemcrash:
OK, good. ( The only drawback now is that the % width applies to every translation so some header wrapping is inevitable )

systemcrash:
Merged. Thanks!

PR #7200, opened 2024-07-18
luci-base: fix regression add by button-row class

feckert:
This should fix the regression that was introduced by https://github.com/openwrt/luci/pull/7153


feckert:
@systemcrash @dannil @zdz2019  All the problems mentioned should be solved with this PR.

zxlhhyccc:
Problem resolved, hope to merge soon.
By the way, would it be better if these two buttons were swapped?
image

PR #7203, opened 2024-07-20
luci-app-attendedsysupgrade: use versioned endpoints

dannil:
The unversioned endpoint /api/overview has been removed from ASU which breaks the request when searching for an available sysupgrade if the device is on a stable release. In the case where the unversioned endpoint is added back to prevent this regression in 23.05.3 and earlier (so upgrade is once again possible) we should anyway rely on the versioned endpoints as those replace the unversioned ones.

Ref: https://github.com/openwrt/asu/issues/896. See http://sysupgrade.openwrt.org/docs for the current API.

    * x86/64, 23.05.3, Firefox -> upgraded to 23.05.4 successfully
    * x86/64, SNAPSHOT r26870-4a2f712f85, Firefox -> upgraded to SNAPSHOT r26971-66177c081f successfully

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -399,10 +399,9 @@ return view.extend({
 		let { url, revision, advanced_mode, branch } = data;
 		let { version, target, profile, packages } = firmware;
 		let candidates = [];
-		let request_url = `${url}/api/overview`;
-		if (version.endsWith('SNAPSHOT')) {
-			request_url = `${url}/api/v1/revision/${version}/${target}`;
-		}
+
+		const endpoint = version.endsWith('SNAPSHOT') ? `revision/${version}/${target}` : 'overview';
+		const request_url = `${url}/api/v1/${endpoint}`;
 
 		ui.showModal(_('Searching...'), [
 			E(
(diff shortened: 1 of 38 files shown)


Lanchon:
@dannil thanks for the PR.

regardless of whether this PR is merged, are you planing to add the old endpoint back to the server?

if not i guess users could update their luci-app-attendedsysupgrade package, the do the upgrade. but this breaks ASU in various ways. some people maintain routers that do not have internet connection yet used to be easily upgraded via luci-app-attendedsysupgrade (which requires access to the internet on the browser, not in the router). users who are using older snapshots may not be able to upgrade the luci asu package, though i am less clear on that one.

in any case, the endpoint change breaks the ASU premise and workflow. IMHO it would be a good thing to restore it.

dannil:
> regardless of whether this PR is merged, are you planing to add the old endpoint back to the server?

Good question! I'm waiting to hear back from @aparcar in https://github.com/openwrt/asu/issues/896, I've never contributed to ASU itself but adding back the old endpoint (doesn't have to be an endpoint on the API but instead a 301 Redirect to `/api/v1/overview`) seems to me the only proper solution, since we obviously won't be able to backport this to already running systems, so the fix needs to be server side on ASU.

dannil:
A redirect was added for `/api/overview` -> `json/v1/overview.json` in https://github.com/openwrt/asu/issues/896#issuecomment-2241326073 so the regression for already running system is fixed.

systemcrash:
versioning API endpoints is almost a must. The other issue is resolved, does that mean this can go in as-is?

dannil:
> versioning API endpoints is almost a must. The other issue is resolved, does that mean this can go in as-is?

Yes, it should be fine to merge this.

systemcrash:
Merged. Thanks @dannil

PR #7241, opened 2024-08-15
[23.05] luci-app-natmap: change bind port data type from port to portrange

heiher:
This commit changes the data type of bind port from `port` to `portrange`, supporting [sequential allocation of ports](https://github.com/heiher/natmap/blob/master/README.md?plain=1#L54-L57).

cc @ysc3839


heiher:
Re: https://github.com/openwrt/packages/pull/24796

PR #7249, opened 2024-08-20
luci-base: form.js: decode HTML entities in AbstractElement.stripTags()

mikma:
- [x] Tested on: (x86-64, openwrt-23.05.3, Firefox 129.0 on Android) :white_check_mark:
- [x] Mention: @feckert, @jow-
- [x] Description: This PR fixes a problem with HTML entities which were visible in their encoded form in the mobile view. This happened for example when displaying a GridSection with a Value option containing `&nbsp;` in the title. Without this change only HTML entities in titles that also contains tags are decoded before they are stored in data-title attributes.


systemcrash:
ping @jow-

systemcrash:
Merged! Thanks @mikma.

PR #7262, opened 2024-08-30
luci-proto-3g/ppp/pppossh: fix default keepalive values

rany2:
The default PPP keepalive value in OpenWRT is actually `5 1` not `0 5`.


nihilus (approved):

systemcrash:
Should be fine since they're just placeholder values. Thanks @rany2

PR #7267, opened 2024-09-03
luci-app-pbr: sync with principal package

stangri:
Depends on https://github.com/openwrt/packages/pull/24920

One update I'd like some feedback for is changing include of `luci.mk` from relative to an absolute path, that was requested by some users of upstream repo. If it's going to be a problem in OpenWrt repo, please let me know.

### DIFF
--- a/applications/luci-app-pbr/Makefile
+++ b/applications/luci-app-pbr/Makefile
@@ -6,8 +6,8 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=luci-app-pbr
 PKG_LICENSE:=AGPL-3.0-or-later
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>
-PKG_VERSION:=1.1.6
-PKG_RELEASE:=20
+PKG_VERSION:=1.1.7
+PKG_RELEASE:=7
 
 LUCI_TITLE:=Policy Based Routing Service Web UI
 LUCI_URL:=https://github.com/stangri/luci-app-pbr/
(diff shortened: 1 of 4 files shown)


applications/luci-app-pbr/Makefile
```diff
@@ -24,6 +24,6 @@ help
 	Version: $(PKG_VERSION)-$(PKG_RELEASE)
 endef
 
-include ../../luci.mk
+include $(TOPDIR)/feeds/luci/luci.mk
```
krant, line 27: @stangri this change causing that in my local setup (and I'm not even building it):
```
ERROR: please fix package/feeds/applications/luci-app-pbr/Makefile - see logs/package/feeds/applications/luci-app-pbr/dump.txt for details
```
My luci.mk lives in `openwrt/package/feeds/luci.mk` and for all other packages it causes no problem.

stangri: Thanks for the feedback, could you please provide the content of the `logs/package/feeds/applications/luci-app-pbr/dump.txt`?

krant: Here you are:
```
Makefile:27: /home/krant/openwrt/feeds/luci/luci.mk: No such file or directory
make[2]: *** No rule to make target '/home/krant/openwrt/feeds/luci/luci.mk'.  Stop.
```

stangri: @krant thanks for your prompt reply, I'll revert the change

Issue #7229, opened 2024-08-05
luci-app-usteer: SyntaxError An invalid or illegal string was specified


artemklevtsov:
## Steps to reproduce:

1. go to: Network → Interfaces → Usteer

## Actual behavior:

```
SyntaxError
An invalid or illegal string was specified
```

## Expected behavior:
Normal page load.

## Additional Information:
OpenWrt version information from system `/etc/openwrt_release`

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='23.05.4'
DISTRIB_REVISION='r24012-d8dd03c46f'
DISTRIB_TARGET='ipq806x/generic'
DISTRIB_ARCH='arm_cortex-a15_neon-vfpv4'
DISTRIB_DESCRIPTION='OpenWrt 23.05.4 r24012-d8dd03c46f'
DISTRIB_TAINTS=''
```
May be important: I use luci with nginx.

/etc/nginx/uci.conf:

```
# This file is re-created when Nginx starts.
# Consider using UCI or creating files in /etc/nginx/conf.d/ for configuration.
# Parsing UCI configuration is skipped if uci set nginx.global.uci_enable=false
# For details see: https://openwrt.org/docs/guide-user/services/webserver/nginx

worker_processes auto;

user root;

include module.d/*.module;

events {}

http {
        access_log off;
        log_format openwrt
                '$request_method $scheme://$host$request_uri => $status'
                ' (${body_bytes_sent}B in ${request_time}s) <- $http_referer';

        include mime.types;
        default_type application/octet-stream;
        sendfile on;

        client_max_body_size 128M;
        large_client_header_buffers 2 1k;
        server_names_hash_bucket_size 128;

        gzip on;
        gzip_vary on;
        gzip_proxied any;

        root /www;

        server { #see uci show 'nginx._lan'
                listen 443 ssl default_server;
                listen [::]:443 ssl default_server;
                server_name _lan;
                include restrict_locally;
                include conf.d/luci.locations;
                ssl_certificate /etc/nginx/conf.d/_lan.crt;
                ssl_certificate_key /etc/nginx/conf.d/_lan.key;
                ssl_session_cache shared:SSL:32k;
                ssl_session_timeout 64m;
                access_log off; # logd openwrt;
        }

        server { #see uci show 'nginx._redirect2ssl'
                listen 80;
                listen [::]:80;
                server_name _redirect2ssl;
                return 302 https://$host$request_uri;
        }

[...]

PR #7283, opened 2024-09-15
luci-theme-openwrt: change to HTML (like all other themes) instead of XML to fix incompatibility

Ramon00:
luci-theme-openwrt: change to HTML (like all other themes) instead of XML to fix incompatibility

see also https://github.com/openwrt/luci/issues/7229


systemcrash:
Let's roll with HTML. Thanks for taking the time to fix this @Ramon00 !

PR #7292, opened 2024-09-23
luci-app-https-dns-proxy: better description/link for force_dns

stangri:
* better description/link for force_dns in WebUI


Matthaiks:
It seems the msgid "Forces Router DNS use on local devices, also known as DNS Hijacking." needs to be updated in the POT file as well.

PR #7300, opened 2024-10-02
luci-app-filemanager: initial checkin

rdmitry0911:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`) :white_check_mark:
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages :white_check_mark:
- [x] Incremented :up: any `PKG_VERSION` in the Makefile :white_check_mark:
- [x] Description: (describe the changes proposed in this PR) Full featured File Manager for LUCI. Easy navigation thru directory structure and selected files editing. Effectively replaces nano + openssh-sftp-server 
Screenshot 2024-10-02 at 2 31 22 PM
Screenshot 2024-10-02 at 2 31 49 PM

### DIFF
--- a/applications/luci-app-filemanager/Makefile
+++ b/applications/luci-app-filemanager/Makefile
@@ -1,0 +1,13 @@
+# This is free software, licensed under the Apache License, Version 2.0 .
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI File Manager module
+LUCI_DEPENDS:=+luci-base
+
+PKG_LICENSE:=Apache-2.0
+PKG_MAINTAINER:=Dmitry R <rdmitry0911@gmail.com>
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
+++ b/applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
@@ -1,1433 +1,17 @@
-// Enable strict mode for better error checking
 'use strict';
+'require view';
+'require fs';
+'require ui';
+'require dom';
+'require rpc';
+'require view.system.filemanager.md as md';
+'require view.system.filemanager.md_help as md_help';
+'require view.system.filemanager.HexEditor as HE';
 
-// Require necessary modules from the LuCI framework
-'require view'; // For working with views
-'require fs'; // For filesystem operations
-'require ui'; // For user interface components
-'require dom'; // For DOM manipulation
-'require rpc'; // For remote procedure cajls
-
-/**
- * Parses a limited subset of Markdown and converts it to HTML.
- *
- * Supported Markdown elements:
- * - Headings (#, ##, ###)
- * - Bold text (**text** or __text__)
- * - Unordered lists (- or *)
- * - Ordered lists (1., 2., etc.)
- * - Paragraphs
- *
- * @param {string} markdown - The Markdown-formatted string.
- * @returns {string} - The resulting HTML string.
- */
-function parseMarkdown(markdown) {
-	// Split the input into lines
-	const lines = markdown.split('\n');
-	const html = [];
-	let inList = false;
-	let listType = ''; // 'ul' or 'ol'
-
-	lines.forEach((line) => {
-		let trimmedLine = line.trim();
-
-		if (trimmedLine === '') {
-			// Empty line signifies a new paragraph
-			if (inList) {
-				html.push(`</${listType}>`);
-				inList = false;
-				listType = '';
-			}
-			return; // Skip adding empty lines to HTML
-		}
-
-		// Check for headings
-		if (/^###\s+(.*)/.test(trimmedLine)) {
-			const content = trimmedLine.replace(/^###\s+/, '');
-			html.push(`<h3>${escapeHtml(content)}</h3>`);
-			return;
-		} else if (/^##\s+(.*)/.test(trimmedLine)) {
-			const content = trimmedLine.replace(/^##\s+/, '');
-			html.push(`<h2>${escapeHtml(content)}</h2>`);
-			return;
-		} else if (/^#\s+(.*)/.test(trimmedLine)) {
-			const content = trimmedLine.replace(/^#\s+/, '');
-			html.push(`<h1>${escapeHtml(content)}</h1>`);
-			return;
-		}
-
-		// Check for ordered lists
-		let orderedMatch = trimmedLine.match(/^(\d+)\.\s+(.*)/);
-		if (orderedMatch) {
-			const [, number, content] = orderedMatch;
-			if (!inList || listType !== 'ol') {
-				if (inList) {
-					html.push(`</${listType}>`);
-				}
-				html.push('<ol>');
-				inList = true;
-				listType = 'ol';
-			}
-			html.push(`<li>${parseInlineMarkdown(escapeHtml(content))}</li>`);
-			return;
-		}
-
-		// Check for unordered lists
-		let unorderedMatch = trimmedLine.match(/^[-*]\s+(.*)/);
-		if (unorderedMatch) {
-			const content = unorderedMatch[1];
-			if (!inList || listType !== 'ul') {
-				if (inList) {
-					html.push(`</${listType}>`);
-				}
-				html.push('<ul>');
-				inList = true;
-				listType = 'ul';
-			}
-			html.push(`<li>${parseInlineMarkdown(escapeHtml(content))}</li>`);
-			return;
-		}
-
-		// If currently inside a list but the line doesn't match a list item, close the list
-		if (inList) {
-			html.push(`</${listType}>`);
-			inList = false;
-			listType = '';
-		}
-
-		// Regular paragraph
-		html.push(`<p>${parseInlineMarkdown(escapeHtml(trimmedLine))}</p>`);
-	});
-
-	// Close any open list tags at the end
-	if (inList) {
-		html.push(`</${listType}>`);
-	}
-
-	return html.join('\n');
-}
-
-/**
- * Parses inline Markdown elements like bold text.
- *
- * Supported inline elements:
- * - Bold text (**text** or __text__)
- *
- * @param {string} text - The text to parse.
- * @returns {string} - The text with inline Markdown converted to HTML.
- */
-function parseInlineMarkdown(text) {
-	// Convert **text** and __text__ to <strong>text</strong>
-	return text
-		.replace(/\*\*(.+?)\*\*/g, '<strong>$1</strong>')
-		.replace(/__(.+?)__/g, '<strong>$1</strong>');
-}
-
-/**
- * Escapes HTML special characters to prevent XSS attacks.
- *
- * @param {string} text - The text to escape.
- * @returns {string} - The escaped text.
- */
-function escapeHtml(text) {
-	const map = {
-		'&': '&amp;',
-		'<': '&lt;',
-		'>': '&gt;',
-		'"': '&quot;',
-		"'": '&#039;',
-	};
-	return text.replace(/[&<>"']/g, function(m) {
-		return map[m];
-	});
-}
 
 function pop(a, message, severity) {
 	ui.addNotification(a, message, severity)
 }
-
-// hexedit.js based on https://github.com/dnlmlr/hexedit-js
-
-const _NON_PRINTABLE_CHAR = "\u00B7";
-
-/**
- * Converts a byte to its corresponding character.
- * If the byte is not printable, returns a non-printable character.
- *
- * @param {number} b - The byte to convert.
- * @returns {string} - The corresponding character.
- */
-function _byteToChar(b) {
-	// If the byte is not printable, use a dot instead
-	return (b >= 32 && b <= 126) ? String.fromCharCode(b) : _NON_PRINTABLE_CHAR;
-}
-
-/**
- * HexEditor class to handle hex editing functionalities.
- */
-class HexEditor {
-	/**
-	 * Constructs a HexEditor instance.
-	 *
-	 * @param {HTMLElement} hexeditDomObject - The DOM element for the hex editor.
-	 */
-	constructor(hexeditDomObject) {
-		this.hexedit = _fillHexeditDom(hexeditDomObject);
-		this.offsets = this.hexedit.querySelector('.offsets');
-		this.hexview = this.hexedit.querySelector('.hexview');
-		this.textview = this.hexedit.querySelector('.textview');
-		this.hexeditContent = this.hexedit.querySelector('.hexedit-content');
-		this.hexeditHeaders = this.hexedit.querySelector('.hexedit-headers'); // Reference to headers
-
-		this.bytesPerRow = 16;
-		this.startIndex = 0; // Starting index for virtual scrolling
-		this.data = new Uint8Array(0); // Initialize with empty data
-
-		this.selectedIndex = null; // Currently selected byte index
-		this.editHex = true; // Flag to determine edit mode (hex or text)
-		this.currentEdit = ""; // Current edit buffer
-		this.readonly = false; // Read-only mode flag
-		this.ctrlPressed = false; // Control key pressed flag
-
-		this.matches = []; // Array to store all match positions and lengths
-		this.currentMatchIndex = -1; // Index of the current match
-		this.currentSearchType = null; // Current search type ('ascii', 'hex', 'regex')
-		this.activeView = null; // Active view based on focus ('hex' or 'text')
-		this.previousSelectedIndex = null; // To track previous selection for color restoration
-
-		// Storage of the last search pattern for each search type
-		this.lastSearchPatterns = {
-			ascii: '',
-			hex: '',
-			regex: ''
-		};
-
-		this._registerEventHandlers();
-
-		// Initialize ResizeObserver for dynamic row calculation
-		this.resizeObserver = new ResizeObserver(() => {
-			this.calculateVisibleRows();
-		});
-		this.resizeObserver.observe(this.hexeditContent);
-
-		// Initialize Search Functionality
-		this.addSearchUI();
-	}
-
-	/**
-	 * Adds the search interface with input fields, status fields, and navigation buttons.
-	 */
-	addSearchUI() {
-		// Create search container
-		const searchContainer = document.createElement('div');
-		searchContainer.classList.add('hexedit-search-container');
-
-		// Helper function to create search groups
-		const createSearchGroup = (type, placeholder) => {
-			const container = document.createElement('div');
-			container.classList.add('hexedit-search-group');
-
-			const input = document.createElement('input');
-			input.type = 'text';
-			input.placeholder = placeholder;
-			input.classList.add('hexedit-search-input');
-			input.id = `hexedit-search-${type}`;
-
-			const status = document.createElement('span');
-			status.classList.add('hexedit-search-status');
-			status.id = `hexedit-search-status-${type}`;
-			status.textContent = '0/0'; // Initial status
-
-			const prevButton = document.createElement('button');
-			prevButton.innerHTML = '&#8593;'; // Up arrow
-			prevButton.classList.add('hexedit-search-button');
-			prevButton.title = `Previous ${type.toUpperCase()} Match`;
-
-			const nextButton = document.createElement('button');
-			nextButton.innerHTML = '&#8595;'; // Down arrow
-			nextButton.classList.add('hexedit-search-button');
-			nextButton.title = `Next ${type.toUpperCase()} Match`;
-
-			container.appendChild(input);
-			container.appendChild(status);
-			container.appendChild(prevButton);
-			container.appendChild(nextButton);
-
-			// Add event listeners for buttons
-			prevButton.addEventListener('click', () => this.handleFindPrevious(type));
-			nextButton.addEventListener('click', () => this.handleFindNext(type));
-
-			// Add event listener for Enter key
-			input.addEventListener('keydown', (e) => {
-				if (e.key === 'Enter') this.handleFindNext(type);
-			});
-
-			return container;
-		};
-
-		// Create ASCII search group
-		const asciiGroup = createSearchGroup('ascii', _('Search ASCII'));
-
-		// Create HEX search group
-		const hexGroup = createSearchGroup('hex', _('Search HEX (e.g., 4F6B)'));
-
-		// Create RegExp search group
-		const regexGroup = createSearchGroup('regex', _('Search RegExp (e.g., \\d{3})'));
-
-		// Append all search groups to the search container
-		searchContainer.appendChild(asciiGroup);
-		searchContainer.appendChild(hexGroup);
-		searchContainer.appendChild(regexGroup);
-
-		// Insert the search container above the hexedit headers
-		if (this.hexeditHeaders) {
-			this.hexedit.insertBefore(searchContainer, this.hexeditHeaders);
-		} else {
-			// Fallback: append to hexedit if headers are not found
-			this.hexeditContent.insertBefore(searchContainer, this.hexeditContent.firstChild);
-		}
-	}
-
-	/**
-	 * Handles the "Find Next" button click for a specific search type.
-	 *
-	 * @param {string} searchType - The type of search ('ascii', 'hex', 'regex').
-	 */
-	handleFindNext(searchType) {
-		const inputElement = document.getElementById(`hexedit-search-${searchType}`);
-		const currentPattern = inputElement.value.trim();
-
-		// Check if the search pattern has changed
-		if (this.lastSearchPatterns[searchType] !== currentPattern) {
-			// Update the last search pattern
-			this.lastSearchPatterns[searchType] = currentPattern;
-
-			// Set the current search type and active view
-			this.currentSearchType = searchType;
-			this.activeView = (searchType === 'hex') ? 'hex' : 'text';
-
-			// Perform search
-			this.performSearch(searchType);
-		} else {
-			// If the search pattern has not changed, just go to the next match
-			if (this.currentSearchType === searchType && this.matches.length > 0) {
-				// Set activeView based on currentSearchType
-				this.activeView = (this.currentSearchType === 'hex') ? 'hex' : 'text';
-
-				// Navigate to the next match relative to the current cursor position
-				const cursorPosition = this.selectedIndex !== null ? this.selectedIndex : 0;
-				const nextMatchIndex = this.findNextMatch(cursorPosition);
-				if (nextMatchIndex !== -1) {
-					this.navigateToMatch(nextMatchIndex);
-				} else {
-					// If there is no next match, go to the first one
-					this.navigateToMatch(0);
-				}
-			}
-		}
-	}
-
-	/**
-	 * Handles the "Find Previous" button click for a specific search type.
-	 *
-	 * @param {string} searchType - The type of search ('ascii', 'hex', 'regex').
-	 */
-	handleFindPrevious(searchType) {
-		const inputElement = document.getElementById(`hexedit-search-${searchType}`);
-		const currentPattern = inputElement.value.trim();
-
-		// Check if the search pattern has changed
-		if (this.lastSearchPatterns[searchType] !== currentPattern) {
-			// Update the last search pattern
-			this.lastSearchPatterns[searchType] = currentPattern;
-
-			// Set the current search type and active view
-			this.currentSearchType = searchType;
-			this.activeView = (searchType === 'hex') ? 'hex' : 'text';
-
-			// Perform search
-			this.performSearch(searchType);
-		} else {
-			// If the search pattern has not changed, just go to the previous match
-			if (this.currentSearchType === searchType && this.matches.length > 0) {
-				// Set activeView based on currentSearchType
-				this.activeView = (this.currentSearchType === 'hex') ? 'hex' : 'text';
-
-				// Navigate to the previous match relative to the current cursor position
-				const cursorPosition = this.selectedIndex !== null ? this.selectedIndex : this.data.length;
-				const prevMatchIndex = this.findPreviousMatch(cursorPosition);
-				if (prevMatchIndex !== -1) {
-					this.navigateToMatch(prevMatchIndex);
-				} else {
-					// If there is no previous match, go to the last one
-					this.navigateToMatch(this.matches.length - 1);
-				}
-			}
-		}
-	}
-
-	/**
-	 * Finds the index of the next match after the given cursor position.
-	 *
-	 * @param {number} cursorPosition - The current cursor position.
-	 * @returns {number} - The index in the matches array or -1 if not found.
-	 */
-	findNextMatch(cursorPosition) {
-		for (let i = 0; i < this.matches.length; i++) {
-			if (this.matches[i].index > cursorPosition) {
-				return i;
-			}
-		}
-		// If there are no matches after the cursor position, return -1
-		return -1;
-	}
-
-	/**
-	 * Finds the index of the previous match before the given cursor position.
-	 *
-	 * @param {number} cursorPosition - The current cursor position.
-	 * @returns {number} - The index in the matches array or -1 if not found.
-	 */
-	findPreviousMatch(cursorPosition) {
-		for (let i = this.matches.length - 1; i >= 0; i--) {
-			if (this.matches[i].index < cursorPosition) {
-				return i;
-			}
-		}
-		// If there are no matches before the cursor position, return -1
-		return -1;
-	}
-
-	/**
-	 * Performs the search based on the specified search type.
-	 *
-	 * @param {string} searchType - The type of search ('ascii', 'hex', 'regex').
-	 */
-	performSearch(searchType) {
-		let pattern = '';
-		switch (searchType) {
-			case 'ascii':
-				pattern = document.getElementById('hexedit-search-ascii').value.trim();
-				break;
-			case 'hex':
-				pattern = document.getElementById('hexedit-search-hex').value.trim();
-				break;
-			case 'regex':
-				pattern = document.getElementById('hexedit-search-regex').value.trim();
-				break;
-			default:
-				console.warn(`Unknown search type: ${searchType}`);
-				pattern = '';
-				break;
-		}
-
-		// Reset previous search results
-		this.clearSearchHighlights();
-		this.matches = [];
-		this.currentMatchIndex = -1;
-
-		if (!pattern) {
-			// Update status field to 0/0
-			this.updateSearchStatus(searchType, 0, 0);
-			console.log('No search pattern entered.');
-			return;
-		}
-
-		try {
-			// Determine search type and perform search
-			if (searchType === 'ascii') {
-				this.searchASCII(pattern);
-			} else if (searchType === 'hex') {
-				this.searchHEX(pattern);
-			} else if (searchType === 'regex') {
-				this.searchRegex(pattern);
-			}
-		} catch (error) {
-			console.log(`Error during search: ${error.message}`);
-			// Update status field to 0/0 on error
-			this.updateSearchStatus(searchType, 0, 0);
-			return;
-		}
-
-		// After searching, highlight all matches and navigate to the first one
-		if (this.matches.length > 0) {
-			this.highlightAllMatches(searchType);
-			this.currentMatchIndex = 0;
-			this.navigateToMatch(this.currentMatchIndex);
-			// Update status field with actual match count
-			this.updateSearchStatus(searchType, this.currentMatchIndex + 1, this.matches.length);
-			console.log(`Found ${this.matches.length} matches.`);
-		} else {
-			// Update status field to 0/0 if no matches found
-			this.updateSearchStatus(searchType, 0, 0);
-			console.log('No matches found.');
-		}
-	}
-
-	/**
-	 * Highlights all matched patterns in the hex and text views based on search type.
-	 *
-	 * @param {string} searchType - The type of search ('ascii', 'hex', 'regex').
-	 */
-	highlightAllMatches(searchType) {
-		// Rendering will handle highlights based on this.matches
-		this.searchTypeForHighlight = searchType; // Store current search type for rendering
-
-		// Set active view based on search type
-		if (searchType === 'ascii' || searchType === 'regex') {
-			this.activeView = 'text'; // Text view is active
-		} else if (searchType === 'hex') {
-			this.activeView = 'hex'; // Hex view is active
-		}
-
-		// Focus the corresponding view
-		this.focusActiveView();
-
-		this.renderDom(); // Re-render to apply the highlights
-	}
-
-	/**
-	 * Navigates to a specific match by its index.
-	 *
-	 * @param {number} matchIndex - The index in the matches array to navigate to.
-	 */
-	navigateToMatch(matchIndex) {
-		if (this.matches.length === 0) {
-			// Update status field to 0/0 if no matches
-			this.updateSearchStatus(this.currentSearchType, 0, 0);
-			console.log('No matches to navigate.');
-			return;
-		}
-
-		// Ensure matchIndex is within bounds
-		if (matchIndex < 0 || matchIndex >= this.matches.length) {
-			console.log('navigateToMatch: matchIndex out of bounds.');
-			return;
-		}
-
-		this.currentMatchIndex = matchIndex;
-		const match = this.matches[matchIndex];
-
-		// Update activeView based on currentSearchType during navigation
-		this.activeView = (this.currentSearchType === 'hex') ? 'hex' : 'text';
-
-		// Set selected index to the match start
-		this.setSelectedIndex(match.index);
-		console.log(`Navigated to match ${matchIndex + 1} at offset ${match.index.toString(16)}`);
-
-		// Update status field
-		this.updateSearchStatus(this.currentSearchType, this.currentMatchIndex + 1, this.matches.length);
-	}
-
-	/**
-	 * Searches for an ASCII pattern and stores all match positions.
-	 *
-	 * @param {string} pattern - The ASCII pattern to search for.
-	 */
-	searchASCII(pattern) {
-		const dataStr = new TextDecoder('iso-8859-1').decode(this.data);
-		const regex = new RegExp(pattern, 'g');
-		let match;
-		while ((match = regex.exec(dataStr)) !== null) {
-			this.matches.push({
-				index: match.index,
-				length: pattern.length
-			});
-			// Prevent infinite loops with zero-length matches
-			if (match.index === regex.lastIndex) {
-				regex.lastIndex++;
-			}
-		}
-		console.log(`searchASCII: Found ${this.matches.length} matches.`);
-	}
-
-	/**
-	 * Searches for a HEX pattern and stores all match positions.
-	 *
-	 * @param {string} pattern - The HEX pattern to search for (e.g., "4F6B").
-	 */
-	searchHEX(pattern) {
-		// Remove spaces and validate hex string
-		const cleanedPattern = pattern.replace(/\s+/g, '');
-		if (!/^[0-9a-fA-F]+$/.test(cleanedPattern)) {
-			throw new Error('Invalid HEX pattern.');
-		}
-		if (cleanedPattern.length % 2 !== 0) {
-			throw new Error('HEX pattern length must be even.');
-		}
-
-		// Convert hex string to byte array
-		const bytePattern = new Uint8Array(cleanedPattern.match(/.{1,2}/g).map(byte => parseInt(byte, 16)));
-
-		for (let i = 0; i <= this.data.length - bytePattern.length; i++) {
-			let found = true;
-			for (let j = 0; j < bytePattern.length; j++) {
-				if (this.data[i + j] !== bytePattern[j]) {
-					found = false;
-					break;
-				}
-			}
-			if (found) {
-				this.matches.push({
-					index: i,
-					length: bytePattern.length
-				});
-			}
-		}
-		console.log(`searchHEX: Found ${this.matches.length} matches.`);
-	}
-
-	/**
-	 * Searches using a regular expression and stores all match positions.
-	 *
-	 * @param {RegExp} regexPattern - The regular expression pattern to search for.
-	 */
-	searchRegex(regexPattern) {
-		const regex = new RegExp(regexPattern, 'g');
-		const dataStr = new TextDecoder('iso-8859-1').decode(this.data);
-		let match;
-		while ((match = regex.exec(dataStr)) !== null) {
-			const byteIndex = match.index; // With 'iso-8859-1', char index == byte index
-			const length = match[0].length;
-			this.matches.push({
-				index: byteIndex,
-				length: length
-			});
-			// Prevent infinite loops with zero-length matches
-			if (match.index === regex.lastIndex) {
-				regex.lastIndex++;
-			}
-		}
-		console.log(`searchRegex: Found ${this.matches.length} matches.`);
-	}
-
-	/**
-	 * Scrolls the editor to make the match at the specified index visible.
-	 *
-	 * @param {number} index - The byte index of the match.
-	 */
-	scrollToMatch(index) {
-		const lineNumber = Math.floor(index / this.bytesPerRow);
-		const lineHeight = 16; // Height of one row in pixels
-
-		// Calculate new scroll position to ensure the matched line is visible
-		const newScrollTop = Math.max(0, (lineNumber * lineHeight) - ((this.visibleRows / 2) * lineHeight));
-
-		console.log(`scrollToMatch called with index: ${index}`);
-		console.log(`lineNumber: ${lineNumber}`);
-		console.log(`newScrollTop: ${newScrollTop}`);
-
-		// Update the scrollTop property to trigger handleScroll
-		this.hexeditContent.scrollTop = newScrollTop;
-	}
-
-	/**
-	 * Clears previous search highlights.
-	 */
-	clearSearchHighlights() {
-		// Remove previous highlights
-		this.hexview.querySelectorAll('.search-highlight').forEach(span => {
-			span.classList.remove('search-highlight');
-		});
-		this.textview.querySelectorAll('.search-highlight').forEach(span => {
-			span.classList.remove('search-highlight');
-		});
-
-		// Reset active view
-		this.activeView = null;
-
-		// Reset all search status fields to 0/0
-		['ascii', 'hex', 'regex'].forEach(type => {
-			this.updateSearchStatus(type, 0, 0);
-		});
-	}
-
-	/**
-	 * Calculates the number of visible rows based on the container's height.
-	 */
-	calculateVisibleRows() {
-		const lineHeight = 16; // Height of one row in pixels
-		const containerHeight = this.hexeditContent.clientHeight;
-		this.visibleRows = Math.floor(containerHeight / lineHeight);
-		this.visibleByteCount = this.bytesPerRow * this.visibleRows;
-		console.log(`calculateVisibleRows: visibleRows=${this.visibleRows}, visibleByteCount=${this.visibleByteCount}`);
-		this.renderDom(); // Re-render to apply the new rows
-	}
-
-	/**
-	 * Sets the data to be displayed in the hex editor.
-	 *
-	 * @param {Uint8Array} data - The data to set.
-	 */
-	setData(data) {
-		this.data = data;
-		this.totalRows = Math.ceil(this.data.length / this.bytesPerRow);
-		console.log(`setData: data length=${this.data.length}, totalRows=${this.totalRows}`);
-		this.calculateVisibleRows(); // Ensure visibleRows are calculated before rendering
-	}
-
-	/**
-	 * Retrieves the current data from the hex editor.
-	 *
-	 * @returns {Uint8Array} - The current data.
-	 */
-	getData() {
-		return this.data;
-	}
-
-	/**
-	 * Handles the scroll event for virtual scrolling.
-	 *
-	 * @param {Event} event - The scroll event.
-	 */
-	handleScroll(event) {
-		const scrollTop = this.hexeditContent.scrollTop;
-		const lineHeight = 16; // Approximate height of a byte row in pixels
-		const firstVisibleLine = Math.floor(scrollTop / lineHeight);
-		const newStartIndex = firstVisibleLine * this.bytesPerRow;
-
-		console.log(`handleScroll: scrollTop=${scrollTop}, firstVisibleLine=${firstVisibleLine}, newStartIndex=${newStartIndex}`);
-
-		// Update startIndex and re-render the DOM if necessary
-		if (newStartIndex !== this.startIndex) {
-			this.startIndex = newStartIndex;
-			this.renderDom(); // Re-render visible data
-			console.log(`handleScroll: Updated startIndex and rendered DOM.`);
-		}
-	}
-
-	/**
-	 * Renders the visible portion of the hex editor based on the current scroll position.
-	 */
-	renderDom() {
-		// Clear existing content
-		[this.offsets, this.hexview, this.textview].forEach(view => view.innerHTML = '');
-		const lineHeight = 16; // Approximate line height in pixels
-		const totalLines = Math.ceil(this.data.length / this.bytesPerRow);
-
-		// Set the height of the content area to simulate the total height
-		const contentHeight = totalLines * lineHeight;
-		[this.offsets, this.hexview, this.textview].forEach(view => view.style.height = `${contentHeight}px`);
-		// Create fragments to hold the visible content
-		const offsetsFragment = document.createDocumentFragment();
-		const hexviewFragment = document.createDocumentFragment();
-		const textviewFragment = document.createDocumentFragment();
-
-		// Calculate the start and end lines to render
-		const startLine = Math.floor(this.startIndex / this.bytesPerRow);
-		const endIndex = Math.min(this.startIndex + this.visibleByteCount, this.data.length);
-		const endLine = Math.ceil(endIndex / this.bytesPerRow);
-
-		const paddingTop = startLine * lineHeight;
-
-		// Apply padding to offset the content to the correct vertical position
-		this.offsets.style.paddingTop = paddingTop + 'px';
-		this.hexview.style.paddingTop = paddingTop + 'px';
-		this.textview.style.paddingTop = paddingTop + 'px';
-
-		// Render only the visible lines
-		for (let line = startLine; line < endLine; line++) {
-			const i = line * this.bytesPerRow;
-
-			// Offsets
-			const offsetSpan = document.createElement("span");
-			offsetSpan.innerText = i.toString(16).padStart(8, '0');
-			offsetsFragment.appendChild(offsetSpan);
-
-			// Hexview line
-			const hexLine = document.createElement('div');
-			hexLine.classList.add('hex-line');
-
-			// Textview line
-			const textLine = document.createElement('div');
-			textLine.classList.add('text-line');
-
-			for (let j = 0; j < this.bytesPerRow && i + j < this.data.length; j++) {
-				const index = i + j;
-				const byte = this.data[index];
-
-				// Create hex span
-				const hexSpan = document.createElement('span');
-				hexSpan.textContent = byte.toString(16).padStart(2, '0');
-				hexSpan.dataset.byteIndex = index;
-
-				// Apply search highlights based on search type
-				this.matches.forEach(match => {
-					if (index >= match.index && index < match.index + match.length) {
-						hexSpan.classList.add('search-highlight');
-					}
-				});
-
-				hexLine.appendChild(hexSpan);
-
-				// Create text span
-				const charSpan = document.createElement('span');
-				let text = _byteToChar(byte);
-				if (text === " ") text = "\u00A0";
-				else if (text === "-") text = "\u2011";
-				charSpan.textContent = text;
-				charSpan.dataset.byteIndex = index;
-				if (text === _NON_PRINTABLE_CHAR) {
-					charSpan.classList.add("non-printable");
-				}
-
-				// Apply search highlights based on search type
-				this.matches.forEach(match => {
-					if (index >= match.index && index < match.index + match.length) {
-						charSpan.classList.add('search-highlight');
-					}
-				});
-
-				textLine.appendChild(charSpan);
-			}
-
-			hexviewFragment.appendChild(hexLine);
-			textviewFragment.appendChild(textLine);
-		}
-
-		this.offsets.appendChild(offsetsFragment);
-		this.hexview.appendChild(hexviewFragment);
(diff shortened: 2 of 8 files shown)


rdmitry0911:
New version includes: 

- Navigation with resizable window/columns and with list sorting
- Directory and files creation/renaming/removing
- Directory and files attributes changing
- Large files uploading/downloading
- Large files editing
- Storage of Interface settings in config file
Screenshot 2024-10-15 at 8 21 57 AM
Screenshot 2024-10-15 at 8 22 57 AM
Screenshot 2024-10-15 at 8 23 12 AM
Screenshot 2024-10-15 at 8 23 50 AM
Screenshot 2024-10-15 at 8 25 15 AM
Screenshot 2024-10-15 at 8 27 10 AM
Screenshot 2024-10-15 at 8 27 30 AM

systemcrash:
Looking good. Avoid merge commits. Just rebase your work on master.

rdmitry0911:
> Looking good. Avoid merge commits. Just rebase your work on master.

This way is better?

dannil:
Please follow the [contribution guidelines](https://github.com/openwrt/luci/blob/master/CONTRIBUTING.md); squash both your commits together and name it something like `luci-app-filemanager: add app` and put further verbose explanation into the commit body instead of the subject.

dannil (changes requested):
Please follow the [contribution guidelines](https://github.com/openwrt/luci/blob/master/CONTRIBUTING.md); squash both your commits together and name it something like `luci-app-filemanager: add app` and put further verbose explanation into the commit body instead of the subject. And just do a rebase, right now you've included every commit again which (I guess) was newer than your HEAD.

systemcrash:
Amend your commit subject

rdmitry0911:
> Amend your commit subject

What would you propose?

systemcrash:
OK great. I think this is ready. @dannil @stokito @feckert any opinions?

systemcrash:
Do you want to take it for a test-drive @stokito ?

systemcrash:
@rdmitry0911 please address @stokito's comments.

rdmitry0911:
> I tested and it works. My main comments are: 0. Not sure if we need options for styling like width/height, padding etc.
> 
Initially there were no settings for width/height and padding, but finally I added them as different users have different environments and it is convenient to start this app with width/height, padding you like rather then adjust them right next to start. 

>     1. Rename `Current Directory:` to a `Starting Folder:`
> 
No problem, I'll do that. Should I change a version number just for this commit or it is better to collect several changes and commit them together for version changing?

>     2. Use Luci's own uci.js to get options.
> 
>     3. Use Luci's own fs to get list of files instead of calling of `ls`
> 
In fact initially it was done that way, however, using ls, cat and the likes makes the code more compact. And as far as those utils are required for running any system, using them is harmless.

systemcrash:
I would prefer to use the built-in fs lib than suffering monstrous regexp to process ls output. fs has a bunch of stable and usable properties. Then you avoid risks associated with parsing of dates, times and permissions.

rdmitry0911:
> I would prefer to use the built-in fs lib than suffering monstrous regexp to process ls output. fs has a bunch of stable and usable properties. Then you avoid risks associated with parsing of dates, times and permissions.

The problem of using 'fs.list()' is that I can't find a way to get information about symbolic links. They are returned as directories. I have to use 'fs.stat()' for every directory in a list to find links. It produces significant overhead in comparison with 'ls'

rdmitry0911:
> We better not to change the `type` but instead introduce a separate property for symlinks. Maybe we need to support the hardlinks too, not sure.

I would also return a target of a symlink in a separated property to avoid extra fs.stat()

rdmitry0911:
> Makes sense. Also eventually we'll need to add a button to create a symlink. @rdmitry0911 please let us know if you'll add anything to the pr or it can be merged. I hope everybody here are fine to merge the app. Once it merged it may attract more contributors

I think you can merge it.

systemcrash:
Try `lstat`? https://ucode.mein.io/module-fs.html#lstat

systemcrash:
@jow- in da house. Can we get a linkstat `lstat` function in rpcd please?

rdmitry0911:
Is there anything I can do to accelerate approvals for this PR?

systemcrash:
OK, I have a suitable patch for rpcd, which may take a while to get added. But here's how it works:
```sh
root@OpenWrt:/# ubus call file lstat '{"path":"/etc/os-release"}'
{
	"path": "/etc/os-release",
	"type": "symlink",
	"size": 21,
	"mode": 41471,
	"atime": 1727094886,
	"mtime": 1727094886,
	"ctime": 1727094886,
	"inode": 219,
	"uid": 0,
	"gid": 0
}
root@OpenWrt:/# ubus call file stat '{"path":"/etc/os-release"}'
{
	"path": "/etc/os-release",
	"type": "file",
	"size": 528,
	"mode": 33188,
	"atime": 1727094886,
	"mtime": 1727094886,
	"ctime": 1727094886,
	"inode": 1124,
	"uid": 0,
	"gid": 0
}
root@OpenWrt:/# ubus call file stat '{"path":"/var"}'
{
	"path": "/var",
	"type": "directory",
	"size": 520,
	"mode": 17407,
	"atime": 1730221560,
	"mtime": 1730221753,
	"ctime": 1730221753,
	"inode": 1,
	"uid": 0,
	"gid": 0
}
root@OpenWrt:/# ubus call file lstat '{"path":"/var"}'
{
	"path": "/var",
	"type": "symlink",
	"size": 3,
	"mode": 41471,
	"atime": 1727094886,
	"mtime": 1727094886,
	"ctime": 1727094886,
	"inode": 1264,
	"uid": 0,
	"gid": 0
}
root@OpenWrt:/# ubus call file stat '{"path":"/bin/ash"}'
{
	"path": "/bin/ash",
	"type": "file",
	"size": 405522,
	"mode": 33261,
	"atime": 1727094886,
	"mtime": 1727094886,
	"ctime": 1727094886,
	"inode": 15,
	"uid": 0,
	"gid": 0
}
root@OpenWrt:/# ubus call file lstat '{"path":"/bin/ash"}'
{
	"path": "/bin/ash",
	"type": "symlink",
	"size": 7,
	"mode": 41471,
	"atime": 1727094886,
	"mtime": 1727094886,
	"ctime": 1727094886,
	"inode": 13,
	"uid": 0,
	"gid": 0
}
```

rdmitry0911:
> OK, I have a suitable patch for rpcd, which may take a while to get added. But here's how it works:

This patch resolves only lack of individual file information. Another patch for fs.list() is required to get all the information about folder contents at once like using ls -l.

rdmitry0911:
> We may merge it as is and the rpcd chamge can be made later

Agree. The app just works. We can improve it later on

applications/luci-app-filemanager/Makefile
```diff
@@ -0,0 +1,15 @@
+# This is free software, licensed under the Apache License, Version 2.0 .
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI File Manager module
+LUCI_DEPENDS:=+luci-base
+
+PKG_LICENSE:=Apache-2.0
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=1
```
systemcrash, line 10: Remove this. git semantic versioning is fine.

applications/luci-app-filemanager/root/usr/share/rpcd/acl.d/luci-app-filemanager.json
```diff
@@ -0,0 +1,14 @@
+{
+	"luci-app-filemanager": {
+		"description": "Grant access to File Manager",
+		"write": {
```
systemcrash, line 4: There are no permissions for ls if that's what you're going to use.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+                                        E('div', { 'class': 'resizer' })
+                                    ]),
+                                    E('th', { 'data-field': 'size' }, [
+                                        _('Size'),
+                                        E('button', { 'class': 'sort-button', 'data-field': 'size', 'title': 'Sort by Size' }, '↕'),
+                                        E('div', { 'class': 'resizer' })
+                                    ]),
+                                    E('th', { 'data-field': 'mtime' }, [
+                                        _('Last Modified'),
+                                        E('button', { 'class': 'sort-button', 'data-field': 'mtime', 'title': 'Sort by Last Modified' }, '↕'),
```
systemcrash, line 379: All of these title strings should be _('...') i18n able.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+        // Проверяем, существует ли путь и является ли он директорией
+        fs.stat(newPath).then(function(stat) {
+            if (stat.type === 'directory') {
+                currentPath = newPath;
+                pathInput.value = currentPath;
+                self.loadFileList(currentPath).then(function() {
+                    self.initResizableColumns();
+                });
+            } else {
+                ui.addNotification(null, E('p', _('Указанный путь не является директорией.')), 'error');
```
systemcrash, line 572: I'm not sure most users would bother with or be satisfied with russian. Something like `The specified path does not appear to be a directory`.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+            }
+        }).catch(function(err) {
+            ui.addNotification(null, E('p', _('Не удалось получить доступ к указанному пути: %s').format(err.message)), 'error');
```
systemcrash, line 575: english...

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+
+            var statusInfo = document.getElementById('status-info');
+            var statusProgress = document.getElementById('status-progress');
+
+            if (statusInfo) {
+                statusInfo.textContent = _('Uploading: ') + file.name;
+            }
+            if (statusProgress) {
+                statusProgress.innerHTML = '';
+                var progressBarContainer = E('div', { 'class': 'cbi-progressbar', 'title': '0%' }, [
```
systemcrash, line 607: title: 0%?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+
+            var statusInfo = document.getElementById('status-info');
+            var statusProgress = document.getElementById('status-progress');
+            if (statusInfo) statusInfo.textContent = _('No file selected.');
+            if (statusProgress) statusProgress.innerHTML = '';
+        }).catch(function(err) {
+            ui.addNotification(null, E('p', _('Failed to create directory: %s').format(err.message)), 'error');
+        });
+    },
+    handleCreateFileClick: function(ev) { // Новый метод
```
systemcrash, line 703: Comments are good - even better in English.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+        }).catch(function(err) {
+            ui.addNotification(null, E('p', _('Failed to create file: %s').format(err.message)), 'error');
+        });
+    },
+    loadFileList: function(path) {
+        var self = this;
+        return getFileList(path).then(function(files) {
+            var fileList = document.getElementById('file-list');
+            if (!fileList) {
+                ui.addNotification(null, E('p', _('Failed to find the file list container.')), 'error');
```
systemcrash, line 772: What does that mean, to a user who doesn't know what a container is?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+                // Другие ошибки
+                return Promise.reject(new Error(res.stderr.trim()));
+            }
+        } else {
+            var content = res.stdout || '';
+        }
+
+        var editorContainer = document.getElementById('editor-container');
+        if (!editorContainer) {
+            ui.addNotification(null, E('p', _('Editor container not found.')), 'error');
```
systemcrash, line 1084: ??

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+                '.sort-button:focus { outline: none; }' +
+                '#status-bar { margin-top: 10px; padding: 10px; background-color: #f9f9f9; border: 1px solid #ccc; min-height: 40px; display: flex; align-items: center; justify-content: space-between; }' +
+                '#status-info { font-weight: bold; display: flex; align-items: center; }' +
+                '#status-progress { width: 50%; }' +
+                '.cbi-progressbar { width: 100%; background-color: #e0e0e0; border-radius: 5px; overflow: hidden; height: 10px; }' +
+                '.cbi-progressbar div { height: 100%; background-color: #76c7c0; width: 0%; transition: width 0.2s; }'+
+                '.file-manager-header { display: flex; align-items: center; }' +
+                '.file-manager-header h2 { margin: 0; }' +
+                '.file-manager-header input { margin-left: 10px; width: 100%; max-width: 700px; font-size: 18px; }' +
+                '.file-manager-header button { margin-left: 10px; font-size: 18px; }'
```
systemcrash, line 341: All of this feels a bit over-engineered. Why not just set a CSS at the top and use that?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+                self.initResizableColumns();
+            });
+
+            // Очистка уведомлений
+            var statusInfo = document.getElementById('status-info');
+            var statusProgress = document.getElementById('status-progress');
+            if (statusInfo) statusInfo.textContent = _('No file selected.');
+            if (statusProgress) statusProgress.innerHTML = '';
+        }).catch(function(err) {
+            ui.addNotification(null, E('p', _('Failed to create file: %s').format(err.message)), 'error');
```
systemcrash, line 764: This is OK but better to put the format inside the string parentheses (I don't recall the reason why right now). e.g.

_('Failed to create file: %s'.format(err.message)))

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+                                    self.handleFileClick(joinPath(path, file.name));
+                                }
+                            }, file.name)
+                        ]),
+                        E('td', {}, _('File')),
+                        E('td', { 'class': 'size-cell' }, [
+                            E('span', { 'class': 'size-number' }, self.getFormattedSize(file.size).number),
+                            E('span', { 'class': 'size-unit' }, self.getFormattedSize(file.size).unit)
+                        ]),
+                        E('td', {}, new Date(file.mtime * 1000).toLocaleString()),
```
systemcrash, line 856: Does this file time processing account for the local timezone?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+                    }
+                    ui.addNotification(null, E('p', _('File saved successfully.')), 'info');
+                    return self.loadFileList(currentPath).then(function() {
+                        self.initResizableColumns();
+                    });
+                }).then(function() {
+                    // Reset the stored permissions
+                    self.originalFilePermissions = undefined;
+                }).catch(function(err) {
+                    ui.addNotification(null, E('p', _('Failed to restore file permissions: %s').format(err.message)), 'error');
```
systemcrash, line 1207: Same for this `.format`

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+        if (!targetPath.startsWith('/')) {
+            targetPath = joinPath(currentPath, targetPath);
+        }
+        fs.stat(targetPath).then(function(stat) {
+            if (stat.type === 'directory') {
+                self.handleDirectoryClick(targetPath);
+            } else if (stat.type === 'file') {
+                self.handleFileClick(targetPath);
+            } else {
+                ui.addNotification(null, E('p', _('The symlink points to an unsupported type.')), 'error');
```
systemcrash, line 1239: like a dev?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+                var widthParts = widthStr.split(':');
+                if (widthParts.length === 2) {
+                    var field = widthParts[0];
+                    var width = parseInt(widthParts[1], 10);
+                    if (!isNaN(width)) {
+                        config.columnWidths[field] = width;
+                    }
+                }
+            });
+        }
```
systemcrash, line 1449: Can surely improve this by doing something like
```js
	var inputs = {
		columnWidths: document.getElementById('column-widths-input'),
		columnMinWidths: document.getElementById('column-min-widths-input'),
		columnMaxWidths: document.getElementById('column-max-widths-input'),
		padding: document.getElementById('padding-input'),
		paddingMin: document.getElementById('padding-min-input'),
		paddingMax: document.getElementById('padding-max-input'),
		currentDirectory: document.getElementById('current-directory-input'),
		windowWidth: document.getElementById('window-width-input'),
		windowHeight: document.getElementById('window-height-input')
	};

//reusable
	function parseWidthSettings(inputValue, configKey) {
		if (!inputValue) return;
		inputValue.split(',').forEach(function(widthStr) {
			var [field, width] = widthStr.split(':');
			width = parseInt(width, 10);
			if (!isNaN(width)) {
				config[configKey][field] = width;
			}
		});
	}
...
```

And call this function for respective values

applications/luci-app-filemanager/root/usr/share/rpcd/acl.d/luci-app-filemanager.json
systemcrash: (i.e. add some)

applications/luci-app-filemanager/Makefile
rdmitry0911: fixed

applications/luci-app-filemanager/root/usr/share/rpcd/acl.d/luci-app-filemanager.json
rdmitry0911: As far as I understand what I need is 'exec' permission for "/*" With this it works and doesn't work without

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed. All comments and messages are in English now

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Moved to the top

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Yes, just in case may be there will be some more

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1571 @@
...
+    var columnWidthsInput = document.getElementById('column-widths-input');
+    var columnMinWidthsInput = document.getElementById('column-min-widths-input');
+    var columnMaxWidthsInput = document.getElementById('column-max-widths-input');
+    var paddingInput = document.getElementById('padding-input');
+    var paddingMinInput = document.getElementById('padding-min-input');
+    var paddingMaxInput = document.getElementById('padding-max-input');
+    var currentDirectoryInput = document.getElementById('current-directory-input'); // Получаем поле ввода для текущей директории
+    var windowWidthInput = document.getElementById('window-width-input');
+    var windowHeightInput = document.getElementById('window-height-input');
+
```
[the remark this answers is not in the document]
rdmitry0911: done

rdmitry0911:
In Actions section added a checkbox for selecting files and directories and then deleting selected and added a button to make a copy of files and directories

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+// Function to get file list using fs.exec()
+function getFileList(path) {
+    return fs.exec('/bin/ls', ['-lA', '--full-time', path]).then(function(res) {
+        if (res.code !== 0) {
+            var errorMessage = res.stderr ? res.stderr.trim() : 'Unknown error';
+            return Promise.reject(new Error('Failed to list directory: ' + errorMessage));
+        }
+
+        // Check if  res.stdout id defined and set null string as a default
```
systemcrash, line 223: sp

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+
+    return numeric;
+}
+
+// Function to get file list using fs.exec()
+function getFileList(path) {
+    return fs.exec('/bin/ls', ['-lA', '--full-time', path]).then(function(res) {
+        if (res.code !== 0) {
+            var errorMessage = res.stderr ? res.stderr.trim() : 'Unknown error';
+            return Promise.reject(new Error('Failed to list directory: ' + errorMessage));
```
systemcrash, line 220: say which dir

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+        if (lineNumbersDiv.style.display === 'none' || !lineNumbersDiv.style.display) {
+            // Отображаем номера строк
+            lineNumbersDiv.style.display = 'block';
+            this.updateLineNumbers();
+            this.adjustLineNumbersPadding();
+            editorTextarea.addEventListener('input', this.updateLineNumbers.bind(this));
+            editorTextarea.addEventListener('scroll', this.syncScroll.bind(this));
+            this.syncScroll();
+        } else {
+            // Скрываем номера строк
```
systemcrash, line 1812: sp

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+        var editorTextarea = document.getElementById('editor-textarea');
+
+        if (!lineNumbersDiv || !editorTextarea) {
+            return;
+        }
+
+        // Получаем высоту горизонтальной полосы прокрутки
+        var scrollbarHeight = editorTextarea.offsetHeight - editorTextarea.clientHeight;
+
+        // Устанавливаем нижний отступ для line-numbers
```
systemcrash, line 1723: sp

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+            if (statusInfo) {
+                statusInfo.textContent = _('Downloaded: ') + filePath.split('/').pop();
+            }
+        }).catch(function(err) {
+            ui.addNotification(null, E('p', _('Failed to download file: %s'.format(err.message))), 'error');
+        });
+    },
+    handleDeleteFile: function(filePath) {
+        var self = this;
+        if (confirm(_('Are you sure you want to delete this file or directory?'))) {
```
systemcrash, line 1753: do we know which it is before now? Insert a _('file') i18n string depending on a ternary.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+            // File uploaded successfully, now restore original permissions
+            var permissions = self.originalFilePermissions;
+            if (permissions !== undefined) {
+                return fs.exec('chmod', [permissions, filePath]).then(function(res) {
+                    if (res.code !== 0) {
+                        throw new Error(res.stderr.trim());
+                    }
+                }).then(function() {
+                    if (statusInfo) {
+                        statusInfo.textContent = _('File saved successfully: ') + filePath;
```
systemcrash, line 1932: saved or uploaded?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+                    }
+                    ui.addNotification(null, E('p', _('File saved successfully.')), 'info');
+                    return self.loadFileList(currentPath).then(function() {
+                        self.initResizableColumns();
+                    });
+                }).then(function() {
+                    // Reset the stored permissions
+                    self.originalFilePermissions = undefined;
+                }).catch(function(err) {
+                    ui.addNotification(null, E('p', _('Failed to restore file permissions: %s'.format(err.message))), 'error');
```
systemcrash, line 1942: perhaps "apply" is suitable

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+                });
+            } else {
+                if (statusInfo) {
+                    statusInfo.textContent = _('File saved successfully: ') + filePath;
```
systemcrash, line 1946: mutatis mutandis

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+                }
+                ui.addNotification(null, E('p', _('File saved successfully.')), 'info');
```
systemcrash, line 1948: mutatis mutandis

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+    toggleLineNumbers: function() {
+        var lineNumbersDiv = document.getElementById('line-numbers');
+        var editorTextarea = document.getElementById('editor-textarea');
+
+        if (!lineNumbersDiv || !editorTextarea) {
+            return;
+        }
+
+        if (lineNumbersDiv.style.display === 'none' || !lineNumbersDiv.style.display) {
+            // Отображаем номера строк
```
systemcrash, line 1804: sp

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+                    // Delete button
+                    E('span', {
+                        'class': 'delete-button',
+                        'click': function() {
+                            self.handleDeleteFile(joinPath(path, file.name));
+                        }
+                    }, '🗑️')
+                ];
+
+                // Add the download button only if the file type is 'file'
```
systemcrash, line 1321: not for links?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+                            self.handleEditFile(joinPath(path, file.name), file);
+                        }
+                    }, '✏️'),
+                    // Duplicate button
+                    E('span', {
+                        'class': 'duplicate-button',
+                        'click': function() {
+                            self.handleDuplicateFile(joinPath(path, file.name), file);
+                        }
+                    }, '📄'),
```
systemcrash, line 1311: perhaps the two page emoji

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2288 @@
...
+}
+
+function insertCss(cssContent) {
+    var styleElement = document.createElement('style');
+    styleElement.type = 'text/css';
+    styleElement.appendChild(document.createTextNode(cssContent));
+    document.head.appendChild(styleElement);
+}
+
+var cssContent = '' +
```
systemcrash, line 289: maybe rework this as a ` quoted string with {variables}

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: No, for files only

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: I didn't find 2 page emoji and replaced with this one '🔁'

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: I restructured all the comments and put object name in all related messages

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: uploaded. Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+                }).then(function() {
+                    if (statusInfo) {
+                        statusInfo.textContent = _('File "%s"uploaded successfully.').format(fileName);
+                    }
+                    ui.addNotification(null, E('p', _('File "%s"uploaded successfully.').format(fileName)), 'info');
+                    return self.loadFileList(currentPath).then(function() {
+                        self.initResizableColumns();
+                    });
+                }).catch(function(err) {
+                    ui.addNotification(null, E('p', _('Failed to apply permissions for file "%s": %s').format(fileName, err.message)), 'error');
```
systemcrash, line 1651: to file, not for file

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+                    'class': 'edit-button',
+                    'click': function() {
+                        self.handleEditFile(joinPath(path, file.name), file);
+                    }
+                }, '✏️'), E('span', {
+                    'class': 'duplicate-button',
+                    'click': function() {
+                        self.handleDuplicateFile(joinPath(path, file.name), file);
+                    }
+                }, '🔁'), E('span', {
```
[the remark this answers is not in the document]
rdmitry0911: put this one

systemcrash:
OK. I need some fresh eyes on this. @dannil could you give it a once-over?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+var sortDirection = 'asc'; // Sort direction (ascending/descending)
+var configFilePath = '/etc/config/filemanager'; // Path to the configuration file
+
+// Configuration object to store interface settings
+var config = {
+    // Column widths in the file table
+    columnWidths: {
+        'name': 150,
+        'type': 100,
+        'size': 100,
```
systemcrash, line 24: Indent using tabs. Reformat the file using tabs not spaces

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+                    'change': function(ev) {
+                        self.handleCheckboxChange(ev);
+                    }
+                });
+                var actionButtons = [checkbox, E('span', {
+                    'class': 'edit-button',
+                    'click': function() {
+                        self.handleEditFile(joinPath(path, file.name), file);
+                    }
+                }, '📑'), E('span', {
```
systemcrash, line 1121: sorry if this wasn't clear: edit can be the pencil icon

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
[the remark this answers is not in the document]
systemcrash: while copy can be the two-page icon

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+        }).catch(function(err) {
+            ui.addNotification(null, E('p', _('Failed to delete selected files and directories: %s').format(err.message)), 'error');
+        });
+    },
+    // Function to load the file list
+    loadFileList: function(path) {
+        // Get the list of files and display them in the table
+        var self = this;
+        selectedItems.clear();
+        return getFileList(path).then(function(files) {
```
systemcrash, line 1082: Can you format the columns such that the space between the left window edge and filename isn't zero. A small amount of space is needed.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+        }, _('Create File')), E('button', {
+            'id': 'delete-selected-button',
+            'class': 'btn action-button',
+            'style': 'display: none;',
+            'click': this.handleDeleteSelected.bind(this)
+        }, _('Delete Selected'))])]), E('div', {
+            'id': 'content-editor',
+            'class': 'cbi-tab',
+            'style': 'display:none;'
+        }, [E('p', {}, _('Select a file from the list to edit it here.')), E('div', {
```
systemcrash, line 681: Can this be displayed on the condition that we are not editing a file?

rdmitry0911: Yes, it can and it does

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+				var windowHeightValue = parseInt(inputs.windowHeight.value.trim(), 10);
+				if (!isNaN(windowWidthValue)) {
+					config.windowSizes.width = windowWidthValue;
+				}
+				if (!isNaN(windowHeightValue)) {
+					config.windowSizes.height = windowHeightValue;
+				}
+			}
+			saveConfig().then(function() {
+				ui.addNotification(null, E('p', _('Settingsuploaded successfully.')), 'info');
```
systemcrash, line 1920: Settings *space* updated successfully

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+							config[key] = value; // Other settings
+						}
+					}
+				});
+			}
+		});
+	}).catch(function(err) {
+		console.error("Failed to load config: " + err);
+	});
+}
```
systemcrash, line 140: There may be a number of ways to improve key+value handling, but how about the following, then it's a bit less intense, using a switch case to clean up the ifs:
```js

function parseKeyValuePairs(input, delimiter, callback) {
    const pairs = input.split(',');
    pairs.forEach((pair) => {
        const [key, value] = pair.split(delimiter);
        if (key && value) callback(key.trim(), value.trim());
    });
}

async function loadConfig() {
    try {
        const content = await fs.read(configFilePath);
        const lines = content.trim().split('\n');

        lines.forEach((line) => {
            if (!line.includes('option')) return;

            const splitLines = line.split('option').filter(Boolean);

            splitLines.forEach((subline) => {
                const formattedLine = "option " + subline.trim();
                const match = formattedLine.match(/^option\s+(\S+)\s+'([^']+)'$/);

                if (!match) return;

                const [, key, value] = match;

                switch (key) {
                    case 'columnWidths':
                    case 'columnMinWidths':
                    case 'columnMaxWidths':
                        parseKeyValuePairs(value, ':', (k, v) => {
                            config[key] = config[key] || {};
                            config[key][k] = parseInt(v, 10);
                        });
                        break;

                    case 'currentDirectory':
                        config.currentDirectory = value;
                        break;

                    case 'windowSizes':
                        parseKeyValuePairs(value, ':', (k, v) => {
                            config.windowSizes = config.windowSizes || {};
                            const sizeValue = parseInt(v, 10);
                            if (!isNaN(sizeValue)) {
                                config.windowSizes[k] = sizeValue;
                            }
                        });
                        break;

                    default:
                        config[key] = value;
                }
            });
        });
    } catch (err) {
        console.error('Failed to load config: ' + err.message);
    }
}
```

systemcrash (changes requested):
changes

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
systemcrash: Screenshot 2024-11-19 at 17 15 20

systemcrash:
Please test your colour scheme in a few themes. Some file-names are difficult to read in some:
e.g. black on almost black in dark mode.

Screenshot 2024-11-19 at 17 06 57

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+			'id': 'status-progress'
+		})]), E('div', {
+			'class': 'cbi-page-actions'
+		}, [E('button', {
+			'class': 'btn action-button',
+			'click': this.handleUploadClick.bind(this)
+		}, _('Upload File')), E('button', {
+			'class': 'btn action-button',
+			'click': this.handleMakeDirectoryClick.bind(this)
+		}, _('Make Directory')), E('button', {
```
systemcrash, line 669: recommended:
Create Folder

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,1980 @@
...
+			'type': 'text',
+			'id': 'current-directory-input',
+			'value': config.currentDirectory,
+			'style': 'width:100%; margin-bottom:10px;'
+		})]), E('div', {
+			'class': 'cbi-page-actions'
+		}, [E('button', {
+			'class': 'btn cbi-button-save custom-save-button',
+			'click': this.handleSaveSettings.bind(this)
+		}, _('Save'))])])])])])])]);
```
systemcrash, line 749: someone is happy :D

systemcrash:
Once everything has been completed, update the checks in your initial post. 

Also, please update the commit message to include only one subject. ( remove the others at the end )

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Implemented

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: It happens some times :)

systemcrash:
Handle a user mashing ENTER for this field to change paths

Screenshot 2024-11-19 at 17 38 28

rdmitry0911:
> Please test your colour scheme in a few themes. Some file-names are difficult to read in some: e.g. black on almost black in dark mode.
> 
> Screenshot 2024-11-19 at 17 06 57

Fixed. Hope the visibility is enachanced

rdmitry0911:
> There's a nice bug in settings.
> 
> Click settings to go to settings, and tab opens. Click settings again and the first two settings disappear, and reset to 0. ( width + height )

Fixed

rdmitry0911:
> Handle a user mashing ENTER for this field to change paths
> 
> Screenshot 2024-11-19 at 17 38 28

Added

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

rdmitry0911: Done

rdmitry0911: Fixed

rdmitry0911: I can't resolve this conversation

rdmitry0911:
> Perhaps a recommendation for the next version: handling drag and drop when in a target folder.
> 
> You can see how it is handled when adding SSH keys and in the wireguard proto handler for adding configs.

Implemented

systemcrash:
> Also, an "Invert Selection" button for multiple selections would be great.
> 
> Use case: select files to keep, invert, delete.

Last feature?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2313 @@
...
+				var windowHeightValue = parseInt(inputs.windowHeight.value.trim(), 10);
+				if (!isNaN(windowWidthValue)) {
+					config.windowSizes.width = windowWidthValue;
+				}
+				if (!isNaN(windowHeightValue)) {
+					config.windowSizes.height = windowHeightValue;
+				}
+			}
+			saveConfig().then(function() {
+				ui.addNotification(null, E('p', _('Settings uploaded successfully.')), 'info');
```
systemcrash, line 2248: Settings saved

systemcrash: rebase on master. I've extended addNotification with a timeout parameter, so that notifications remove themselves automatically. Rework your 'info' type notifications to include this.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2313 @@
...
+			promise = promise.then(function() {
+				return fs.exec(cmd[0], cmd[1]).then(function(res) {
+					if (res.code !== 0) {
+						return Promise.reject(new Error(res.stderr.trim()));
+					}
+				});
+			});
+		});
+		promise.then(function() {
+			ui.addNotification(null, E('p', _('Changes to %s "%s"uploaded successfully.').format(_('item'), newItemName)), 'info');
```
systemcrash, line 2172: space?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2313 @@
...
+		}).then(function() {
+			var permissions = self.originalFilePermissions;
+			if (permissions !== undefined) {
+				return fs.exec('chmod', [permissions, filePath]).then(function(res) {
+					if (res.code !== 0) {
+						throw new Error(res.stderr.trim());
+					}
+				}).then(function() {
+					if (statusInfo) {
+						statusInfo.textContent = _('File "%s"uploaded successfully.').format(fileName);
```
systemcrash, line 1972: space?

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Fixed

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,2313 @@
...
+					}
+					ui.addNotification(null, E('p', _('File "%s"uploaded successfully.').format(fileName)), 'info');
```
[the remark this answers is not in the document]
rdmitry0911: Fixed

rdmitry0911:
> > Also, an "Invert Selection" button for multiple selections would be great.
> > Use case: select files to keep, invert, delete.
> 
> Last feature?

Don't quite understand the functionality. Just invert selection? Or need extra options?

> > Also, an "Invert Selection" button for multiple selections would be great.
> > Use case: select files to keep, invert, delete.
> 
> Last feature?

Implemented with alt-click

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
systemcrash: I don't see the timeout number in use: this means that all of those info popups about settings saved and file uploaded stick around pushing the GUI further down:

For a five second timeout try:
```js
				ui.addNotification(null, E('p', _('Settings saved.')), 5000, 'info');
```

systemcrash:
So invert selection looks good, but there's no visual indicator of its availability. Users would have to know about it already and be informed about 'click alt'. Adding a button 'Invert Selection' once a selection is made is a good signal - you can keep the alt behaviour around, though. That's good once you know it ;)

rdmitry0911:
> So invert selection looks good, but there's no visual indicator of its availability. Users would have to know about it already and be informed about 'click alt'. Adding a button 'Invert Selection' once a selection is made is a good signal - you can keep the alt behaviour around, though. That's good once you know it ;)

I'm going to implement a help tab with all the explanations. The less items we have in the interface the more easy it will be to maintain.

rdmitry0911:
New update with help tab and HEX editor/viewer

rdmitry0911:
> changes

Don't quite understand what to do

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,3616 @@
...
+					}
+				}, _('Toggle Line Numbers')),
+				E('button', {
+					'class': 'btn',
+					'id': 'toggle-hex-mode',
+					'style': 'margin-left: 10px;',
+					'click': function() {
+						self.toggleHexMode(filePath);
+					}
+				}, _('Toggle to Hex Mode'))
```
systemcrash, line 3487: Reorder these so that toggle of hex/text is consistent

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,3616 @@
...
+// Main exported view module
+return view.extend({
+	editorMode: 'text',
+	hexEditorInstance: null,
+	// Define the Help content in Markdown format
+	helpContentMarkdown: `
+# LuCI OpenWrt File Manager Application Help
+
+## Introduction
+The **LuCI OpenWrt File Manager** is a comprehensive web-based tool designed to facilitate efficient file management on your OpenWrt-powered device. It provides an intuitive interface with multiple functionalities, enabling users to navigate directories, manage files, edit content, and customize the application's settings seamlessly.
```
systemcrash, line 1349: I'm not sure what to do about this untranslatable 'wall of text'. Most things are self-explanatory based on how the UI reacts and changes.

systemcrash:
> Don't quite understand the functionality. Just invert selection? Or need extra options?

Just an invert selection button which arises once a selection has been made - then it's intuitive. But you're not keen on having one...

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Just to make it more compact :)
Done

rdmitry0911:
> > Don't quite understand the functionality. Just invert selection? Or need extra options?
> 
> Just an invert selection button which arises once a selection has been made - then it's intuitive. But you're not keen on having one...

Yea, I left that for a future contributors if there are any :)

rdmitry0911:
Added ASCII, HEX and RegExp search to hex editor

rdmitry0911:
Changed comments to English ones

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,4355 @@
...
+		} else if (searchType === 'hex') {
+			pattern = document.getElementById('hexedit-search-hex').value.trim();
+		} else if (searchType === 'regex') {
+			pattern = document.getElementById('hexedit-search-regex').value.trim();
+		}
+
+		console.log('performSearch called with type and pattern:', {
+			searchType,
+			pattern
+		});
```
systemcrash, line 419: drop

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,4355 @@
...
+		const textviewFragment = document.createDocumentFragment();
+
+		// Calculate the start and end lines to render
+		const startLine = Math.floor(this.startIndex / this.bytesPerRow);
+		const endIndex = Math.min(this.startIndex + this.visibleByteCount, this.data.length);
+		const endLine = Math.ceil(endIndex / this.bytesPerRow);
+
+		const paddingTop = startLine * lineHeight;
+
+		console.log(`renderDom: startLine=${startLine}, endLine=${endLine}, paddingTop=${paddingTop}`);
```
systemcrash, line 726: drop

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,4355 @@
...
+	}
+
+	/**
+	 * Renders the visible portion of the hex editor based on the current scroll position.
+	 */
+	renderDom() {
+		// Clear existing content
+		this.offsets.innerHTML = '';
+		this.hexview.innerHTML = '';
+		this.textview.innerHTML = '';
```
systemcrash, line 703: perhaps something like this to avoid triplicate:
```js
[this.offsets, this.hexview, this.textview].forEach(view => view.innerHTML = '');
```

it's only slightly more compact - but it cleans up the rest of the code a bit. Your call.

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,4355 @@
...
+
+		const lineHeight = 16; // Approximate line height in pixels
+		const totalLines = Math.ceil(this.data.length / this.bytesPerRow);
+
+		// Set the height of the content area to simulate the total height
+		const contentHeight = totalLines * lineHeight;
+		this.offsets.style.height = contentHeight + 'px';
+		this.hexview.style.height = contentHeight + 'px';
+		this.textview.style.height = contentHeight + 'px';
```
systemcrash, line 712: similarly here:
```js
[this.offsets, this.hexview, this.textview].forEach(view => view.style.height = `${contentHeight}px`);
```

systemcrash:
OK - no more new features as my review time is finite, and this thing could grow forever.

rdmitry0911:
> OK - no more new features as my review time is finite, and this thing could grow forever.

Ok. No more new features :)

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
```diff
@@ -0,0 +1,4355 @@
...
+	 */
+	performSearch(searchType) {
+		let pattern = '';
+		if (searchType === 'ascii') {
+			pattern = document.getElementById('hexedit-search-ascii').value.trim();
+		} else if (searchType === 'hex') {
+			pattern = document.getElementById('hexedit-search-hex').value.trim();
+		} else if (searchType === 'regex') {
+			pattern = document.getElementById('hexedit-search-regex').value.trim();
+		}
```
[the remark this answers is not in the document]
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

applications/luci-app-filemanager/htdocs/luci-static/resources/view/system/filemanager.js
rdmitry0911: Done

rdmitry0911:
> changes

What should I do with this request? I don't quite understand

systemcrash:
> > changes
> 
> What should I do with this request? I don't quite understand

Please link to the comment, it's lost in the noise of hundreds of comments....

rdmitry0911:
> changes

Removed 2 notification lines in Settings tab
Indent using tabs.

systemcrash:
Still don't understand. Please paste a *link* to the comment so I can see the context :)

In general I don't think we should notify on success without it auto-disappearing. Otherwise UI gets flooded. Was it that?

rdmitry0911:
Spaces between the contents and view borders in navigation tab have lost somewhere. Here they are

rdmitry0911:
> > Spaces between the contents and view borders in navigation tab have lost somewhere. Here they are
> 
> ? last force-push?

Yes

systemcrash (approved):

systemcrash:
OK - the file-manager beast :) Some abstract comments for this or next version:

- Bring it up to ES6 standard where possible
- Convert `var` into either `let` or `const` (helps scope variables to blocks)
- Use `forEach` loops instead of `for` loops.
- Use `switch` cases (more consistently) with fall-through instead of `if...else...etc...`

I'll take it for another test-drive tomorrow.

rdmitry0911:
> 1\. Markdown render is nice but the such files are rarely used on routers.

Correct, but removing it we save 7k of full text code or about 1.5k on disk. It provides an option to easily show to a user any formatted text. I think it worth it.


> 2\. Hex Editor is definitely not needed.

It is easy to remove it, however, I'd leave it here as I believe there are some users who would appreciate this functionality
 

> 3\. Settings are almost useless

Disagree. Different themes require different window sizes for both navigation and editing. And working in a specific folder will require extra efforts to get there after a session timeout 


> 4\. Help... I really not sure if this needed and even if needed it can be a link.

I'm not sure either. I have 2 use cases in my mind. Working without public internet access and a placeholder for some specific information that developer would like the user to read.

rdmitry0911:
> 1. The md is not really needed at all. BTW the format is by nature developed to be readable even without a viewer. Given that it adds 7kb (that's a lot) and it may have potential XXS problems I would rather remove it.
> 2. When last time you used HEX editor? Do you think a typical user will edit some binaries? BTW you included the library with a different license (MIT). At least split it to a separate file.
> 3. You are mixing theming and settings. The "last working dir" can be just stored to `localStoreage` in a browser itself and no need to make it as an option.
> 4. "Working without public internet" that's a rare case because almost always you at least have a mobile internet. The file manage is not critical thing. You may use tips and descriptions in the FM itself.
> 
> More importantly we need to understand the end goal here. There is a few options:
> 
> 1. Full featured File Manager with bells and whistles that can be used as poor man's Cloud/NAS e.g. edit notes, watch photos and videos, listen music etc. Then here you are reinventing the existing products like NextCloud, [FileBrowser](https://filebrowser.org/), [WebDAV.js](https://github.com/dom111/webdav-js), [File Management browser app](https://chromewebstore.google.com/detail/file-management-webdav/famepaffkmmhdefbapbadnniioekdppm) etc. This is a long way and we can't help you because we focusing on the core OpenWrt features.
> 2. An ability to work with files to configure a system when there is no Luci app for this. For example to edit the Lighttpd config. This means that we need a minimal functionality.
> 3. Have the file manager in the Luci itself out of the box as my luci-app-filebrowser was initially intended. This means that we need to change features of built-in FileUpload component and split it to several smaller components. This is a complicated task that I have in my todo list. In this case the 2 can be a playground where we can test the functionality (like duplicate folder) and then bring it to the core.
> 
> Please clarify your intention.
> 
> I'll switch to other tasks but you may mention me @stokito for a final review.

@stokito My intention was (and is) to have a filemanager app for luci that would effectively replace scp, nano and navigation part of mc. I put in this app all the functionality I'm using and nothing else. I think it is still light-weighted. I agree, that it might be squeezed by taking off hex editor and help and it'll save about 8-9 kB of a disk space. However, modern routers have enough disk space and 10kB is not a big deal. If other maintainers and reviewers share your position about stripping the functionality, I'll do that.

rdmitry0911:
I also strip out a HexEditor to a separate file and move all separate files to a filemanager folder to simplify code maintenance

rdmitry0911:
By the way, I'm thinking about implementing a plug-in system for filemanger so that extra functionality would be available simply by loading a file into a plugin folder. What do you think is the best way to implement that?

systemcrash:
Closed by 3447700b3870e6cd2338a74312288a9a96d9519b.

> By the way, I'm thinking about implementing a plug-in system for filemanger so that extra functionality would be available simply by loading a file into a plugin folder. What do you think is the best way to implement that?

Keep everything internal to this app (within its folder). What functionality did you have in mind?

systemcrash:
Also wanted to say good work on this app @rdmitry0911. Thanks for your patience.

rdmitry0911:
> Closed by [3447700](https://github.com/openwrt/luci/commit/3447700b3870e6cd2338a74312288a9a96d9519b).
> 
> > By the way, I'm thinking about implementing a plug-in system for filemanger so that extra functionality would be available simply by loading a file into a plugin folder. What do you think is the best way to implement that?
> 
> Keep everything internal to this app (within its folder). What functionality did you have in mind?

I think I can split all the current functionality into 5 plugins: navigation, editor, hexeditor, help and settings. The main program will be responsible for tabs creation according to available plugins (each plugin runs in a separate tab) and calling a plugin when a user selects a correspondent tab. In this case the minimal set will be a program with nav and editor plugins but user can install any additional plugins available. At the moment shell plugin comes my mind.  However, I believe, if this app proves to be useful, there will be contributors for plugins.

rdmitry0911:
> Also wanted to say good work on this app @rdmitry0911. Thanks for your patience.

Also thank you, guys for your time :)

systemcrash:
Yeah, please open a separate issue.

PR #7301, opened 2024-10-03
luci-proto-ipv6/luci-mod-status: Add SNAT fix in map and displaying tunnel status in luci

missing233:
1. `luci-proto-ipv6`: Ensure that the multiple port ranges used with specific ISP MAP-E implementation are actually SNAT-ed correctly in `map`

2. `luci-mod-status`/`luci-proto-ipv6`: add support for displaying 4in6 tunnel status (correct gateway ipv6 address, local tunnel endpoint address and available portsets for map)

Dependent PR: https://github.com/openwrt/openwrt/pull/16582

### DIFF
--- a/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js
+++ b/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js
@@ -37,6 +37,52 @@ return network.registerProtocol('map', {
 		return (network.getIfnameOf(ifname) == this.getIfname());
 	},
 
+	getGatewayAddr: function () {
+		return this.get('peeraddr');
+	},
+
+	getIPv6Addrs: function () {
+		var d = this._ubus('data');
+		if (L.isObject(d) && typeof (d.ipv6addr) == 'string')
+			return d.ipv6addr;
+		return null;
+	},
+
+	callShowPortsets: function () {
+		var d = this._ubus('data');
+		if (L.isObject(d) && typeof (d.portsets) == 'string') {
+			var portSets = d.portsets;
+			if (portSets) {
+				var portArray = portSets.split(' ');
+				var groupedPorts = [];
+				for (var i = 0; i < portArray.length; i += 2) {
+					groupedPorts.push(portArray.slice(i, i + 2));
+				}
+				portSets = E('table', { style: 'width: 100%; border-collapse: collapse;' },
+					groupedPorts.map(function (portGroup) {
+						return E('tr', {}, [
+							E('td', { style: 'padding: 10px; border: 1px solid #ddd; text-align: center; font-size: 16px;' }, portGroup[0]),
+							portGroup[1] ? E('td', { style: 'padding: 10px; border: 1px solid #ddd; text-align: center; font-size: 16px;' }, portGroup[1]) : E('td', {})
+						]);
+					})
+				);
+			}
+			function showPortsets() {
+				L.ui.showModal(_('Available portsets'), [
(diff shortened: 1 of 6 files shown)


protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js
```diff
@@ -86,5 +86,12 @@ return network.registerProtocol('map', {
 		o.datatype    = 'max(9200)';
 
 		o = s.taboption('advanced', form.Flag, 'legacymap', _('Use legacy MAP'), _('Use legacy MAP interface identifier format (draft-ietf-softwire-map-00) instead of RFC7597'));
+
+		o = s.taboption('advanced', form.Flag, 'isp_fix', _('Enable SNAT fix'), _('Apply SNAT fixes with certain ISPs'));
```
systemcrash, line 90: Perhaps just 
```js
		o = s.taboption('advanced', form.Flag, 'isp_fix', _('Enable SNAT fix'), _('Only for certain ISPs'));
```

Then you avoid the weird variable name in the openwrt script portion also.

missing233: Replaced with `snat_fix`.

protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js
```diff
@@ -86,5 +86,12 @@ return network.registerProtocol('map', {
...
+
+		o = s.taboption('advanced', form.Value, 'dont_snat_to', _('Exclude SNAT ports'), _('List of ports to exclude from SNAT. Separate ports with spaces'));
```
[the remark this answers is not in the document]
missing233: Added error checks.

missing233:
@systemcrash I have added error checking. Since [the original fix script](https://github.com/fakemanhk/openwrt-jp-ipoe) uses spaces to separate ports,  I don't plan to modify this feature.

protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js
```diff
@@ -86,5 +86,34 @@ return network.registerProtocol('map', {
...
 
 		o = s.taboption('advanced', form.Flag, 'legacymap', _('Use legacy MAP'), _('Use legacy MAP interface identifier format (draft-ietf-softwire-map-00) instead of RFC7597'));
+
+		o = s.taboption('advanced', form.Flag, 'snat_fix', _('Enable SNAT fix'), _('Apply SNAT fixes with certain ISPs'));
+
+		o = s.taboption('advanced', form.Value, 'dont_snat_to', _('Exclude SNAT ports'), _('List of ports to exclude from SNAT. Separate ports with spaces'));
+		o.depends('snat_fix', '1');
+		o.datatype = 'string';
+		o.placeholder = '80 443 8080';
+		o.validate = function (section_id, value) {
```
systemcrash, line 96: Perhaps simplify to the following (please test):
```js
o.validate = function (section_id, value) {
	value = value.trim().replace(/\s+/g, ' ');
	if (!value) return true;

	let seen = new Set();
	return value.split(' ').every(port => {
		let portNum = parseInt(port, 10);
		if (!/^\d+$/.test(port) || portNum < 1 || portNum > 65535) {
			return _('Expecting: %s').format(_('valid port value'));
		}
		if (seen.has(port)) {
			return _('Duplicate port found: ') + port;
		}
		seen.add(port);
		return true;
	});
};
```

missing233: It’s not working, but I tried simplifying the code in another way.

PR #7306, opened 2024-10-05
luci-mod-network: split interface name and protocol validation

### LINKED ISSUE #7146: luci-mod-network: Can't configure second GRE tunnel
## Steps to reproduce:

1. Create a GRE interface in Network → Interfaces named `gre`
2. Attempt to create a second GRE tunnel with a different name

## Actual behavior:
Lucy throws an error in the device selector:
image

## Expected behavior:
The GRE tunnel is created.

## Additional Information:
```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='11' #my versioning based on master
DISTRIB_REVISION='r26386-1188cae092'
DISTRIB_TARGET='ath79/generic'
DISTRIB_ARCH='mips_24kc'
DISTRIB_DESCRIPTION='OpenWrt 11 r26386-1188cae092'
DISTRIB_TAINTS='no-all'
```

If the tunnel is created directly in UCI it works and shows up in LuCI correctly

dannil:
The previous implementation of binding the protocol validation to the name validation caused problems when creating a new interface where an already existing interface had the new one's protocol as its name, as the protocol would be used when validating if an interface of that name already existed.

Consider the following case:
1. Interface 'gre' with protocol PPPoE is created
2. Interface 'foo' with protocol GRE tunnel over IPv4 (which resolves to 'gre') tries to be created

Creating interface 'foo' would error out during protocol validation as 'gre' would be passed to the name validation, which in turn would check if an interface with name 'gre' already exists, which it does.

Instead of reusing the validation logic, simply trigger the name validation manually as the protocol changes to properly pass the name of the interface instead of the protocol. This also gives the benefit of keeping all error states related to the name contained in the appropriate UI element.

Fixes #7146

List of existing interfaces:

Screenshot_45

Before:
Screenshot_43

After:
Screenshot_44

All existing validation logic still works as expected:
Screenshot_46
Screenshot_47

- [X] Tested on: (x86/64, SNAPSHOT r27466-f368e2d5ec, Firefox) :white_check_mark:
- [X] \( Preferred ) Mention: @ the original code author for feedback @jow
- [X] \( Optional ) Closes: e.g. #7146 

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -1237,7 +1237,10 @@ return view.extend({
 			};
 
 			proto = s2.option(form.ListValue, 'proto', _('Protocol'));
-			proto.validate = name.validate;
+			proto.onchange = function(ev, section_id, value) {
+				var elem = name.getUIElement(section_id);
+				elem.triggerValidation();
+			};
 
 			device = s2.option(widgets.DeviceSelect, 'device', _('Device'));
 			device.noaliases = false;


systemcrash:
Really good. 

@tiagogaspar8 what’s your SOB, so @dannil can add a tested by to the commit also.

dannil:
I realized, looking at the docs, that only the [triggerValidation() with zero parameters](https://openwrt.github.io/luci/jsapi/LuCI.ui.AbstractElement.html#triggerValidation) is documented, since the one with one parameter is [supposed to be private](https://github.com/openwrt/luci/blob/470bb2b8540e5b4949c66b04e5f583c5c36a5a7b/modules/luci-base/htdocs/luci-static/resources/form.js#L1980-L1984), but this being JS of course this still worked and looking at the code base there's actually already existing code which invokes this private overload (see below). https://github.com/openwrt/luci/blob/f5ccfd3df4453ac44b271fd4743bf96a1b0a1199/protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js#L184-L185)

To adhere to the API I've pushed a fix which works in the exact same way but respects the API boundaries.

@tiagogaspar8 if you would like to check out the new code it wouldn't hurt, I've run it on my end and it works just as before.

tiagogaspar8:
Here you go:
`Signed-off-by: Tiago Gaspar <tiagogaspar8@gmail.com>`

tiagogaspar8:
@dannil works flawlessly, great work again!

systemcrash:
I can merge this unless you want to add the following @dannil to the commit :
`Tested-by: Tiago Gaspar <tiagogaspar8@gmail.com>`

systemcrash:
Also @dannil please amend change commit message `protocol as it's name` to `protocol as its name`

dannil:
> I can merge this unless you want to add the following @dannil to the commit : `Tested-by: Tiago Gaspar <tiagogaspar8@gmail.com>`

> Also @dannil please amend commit message `protocol as it's name` to `protocol as its name`

Done and done.

systemcrash:
Great find @dannil. Thanks for confirming, @tiagogaspar8!

systemcrash:
Picked also to 23.05 branch

PR #7307, opened 2024-10-06
luci-app-eventcat: add package.

mips171:
Eventcat is like Watchcat but written in Lua and can react to interface up/down events. Depends on https://github.com/openwrt/packages/pull/25088

Screenshot 2024-10-06 at 21 02 53

- [X] \( Optional ) Depends on: e.g. openwrt/packages#pr-number in sister repo

### DIFF
--- a/applications/luci-app-eventcat/Makefile
+++ b/applications/luci-app-eventcat/Makefile
@@ -1,0 +1,13 @@
+# This is free software, licensed under the Apache License, Version 2.0 .
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI Support for Eventcat
+LUCI_DEPENDS:=+luci-base +eventcat
+
+PKG_LICENSE:=Apache-2.0
+PKG_MAINTAINER:=Nicholas Smith <nicholas@nbembedded.com>
+
+include ../../luci.mk
+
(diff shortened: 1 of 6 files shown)


systemcrash:
So how does this materially differ from watchcat? It’s unlikely that any new lua additions will be accepted. Ucode seems to be the way forward.

mips171:
>So how does this materially differ from watchcat?
Less code and easier to maintain. Due to quirks of shell's language, I had a lot of grief extending the original watchcat and would rather have an option done in a "real" programming language.

>Ucode seems to be the way forward.
That's unfortunate. I'm not interested in learning a DSL so might have to close this one. Whats the real problem that Ucode solves that couldn't have been solved with a library or improvements to the ubus interface? Making a language is really cool, but maintaining a language for the long term probably isn't a commitment to make lightly.

mips171:
JS isn't exactly a language implementation that is known for being small, clean and consistent.  If going for a DSL I'd argue for at least mimicking a better language. Good luck have fun, I'm out.

mips171:
Closing because it depends on Lua, which I thought was included in the base installation.

systemcrash:
Oh, lua is included in base installs, but the idea of new apps and functionality is to avoid using any more lua (and move away from it entirely).

PR #7322, opened 2024-10-10
RFC: luci-mod-system: Allow selective file restore from backups

systemcrash:
This is useful when you're importing backups intended for different hardware, for example, where e.g. `/etc/config/[network|wireless]` are not suitable for the target platform, but everything else useful.

Current iteration caveats: subsequent extraction and compression steps happen on the target device, to then feed the selective_backup archive to sysupgrade. ( It's not impossible that a users `/etc/sysupgrade.conf` that created the original backup stashed gigs of stuff in there... but this is an extreme and unlikely edge case, in which case they're better off using just the original restore ).

Next version avoids this once https://github.com/openwrt/openwrt/pull/16656 is merged 

Feedback welcome. 

The backup restore overview
Screenshot 2024-10-11 at 01 04 21

the selective dialogue
Screenshot 2024-10-11 at 01 21 16

In case nothing was chosen
Screenshot 2024-10-11 at 01 10 57

- [x] Tested on: 23.05.5
- [x] \( Preferred ) Mention: @jow- @Ansuel @hnyman 

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
@@ -139,6 +139,85 @@ return view.extend({
 			}, this, ev.target));
 	},
 
+	handleSelectiveRestore: function(ev) {
+		return ui.uploadFile('/tmp/backup.tar.gz', ev.target)
+			.then(L.bind(function(btn, res) {
+				btn.firstChild.data = _('Checking archive…');
+				return fs.exec('/bin/tar', [ '-tzf', '/tmp/backup.tar.gz' ]);
+			}, this, ev.target))
+			.then(L.bind(function(btn, res) {
+				if (res.code != 0) {
+					ui.addNotification(null, E('p', _('The uploaded backup archive is not readable')));
+					return fs.remove('/tmp/backup.tar.gz');
+				}
+
+				// Split the stdout to get individual file names
+				var files = res.stdout.split('\n').filter(function(f) { return f.trim() !== ''; });
+
+				// Generate check-boxes for each file with <br /> after each
+				var fileCheckboxes = files.map(function(file) {
+					return E('label', { 'class': 'checkbox' }, [
+						E('input', { 'type': 'checkbox', 'name': 'file', 'value': file }),
+						' ', file, E('br')  // Add line break after each checkbox
+					]);
+				});
+
+				// Helper buttons to check all, none, or inverse
+				var checkAllButton = E('button', {
+					'class': 'btn',
+					'click': function() {
+						// Select all check-boxes
+						document.querySelectorAll('input[name="file"]').forEach(function(cb) {
+							cb.checked = true;
+						});
+					}
+				}, _('Check All'));
+
+				var uncheckAllButton = E('button', {
+					'class': 'btn',
+					'click': function() {
+						// Deselect all check-boxes
+						document.querySelectorAll('input[name="file"]').forEach(function(cb) {
+							cb.checked = false;
+						});
+					}
+				}, _('Check None'));
+
+				var invertButton = E('button', {
+					'class': 'btn',
+					'click': function() {
+						// Deselect all check-boxes
+						document.querySelectorAll('input[name="file"]').forEach(function(cb) {
+							cb.checked = !cb.checked;
+						});
+					}
+				}, _('Invert'));
+
+				// Display the modal with check-boxes and helper buttons
+				ui.showModal(_('Apply backup?'), [
+					E('p', _('The uploaded backup archive appears to be valid and contains the files listed below. Select the files you wish to restore and press "Continue", or "Cancel" to abort.')),
+					E('div', {}, fileCheckboxes), // Display check-boxes for files
+					E('div', { 'class': 'right' }, [
+						checkAllButton, ' ', uncheckAllButton, ' ', invertButton, ' ',
+						E('button', {
+							'class': 'btn',
+							'click': ui.createHandlerFn(this, function(ev) {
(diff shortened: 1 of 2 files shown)


systemcrash:
Better might be to simply amend the `sysupgrade` script to ingest the `-T /tmp/filelist.txt` and do a selective extraction from there, obviating the need for the two decompress/compress steps.

dannil:
I really like this idea. I assume you're waiting atm if https://github.com/openwrt/openwrt/pull/16656 gets merged and if it does rewrite it to use the `sysupgrade` command instead of creating a new tarball?

systemcrash:
Exactly that. I have a rewrite in my local repo. It avoids all the pitfalls. sysupgrade is basically a big fat front-end script to tar.

systemcrash:
If you like it, just jam a thumbs up on the other PR and make some noise there. It might not be the ideal way to do it, but it's the cleanest for this purpose. The busybox tar doesn't have many options to work with, so providing individual filenames at the command prompt isn't there yet, I think.

stokito (approved):

systemcrash:
[Real case in the wild](https://forum.openwrt.org/t/solved-bricked-gl-mt6000-even-reset-doesnt-work/216331/)

hnyman:
> convenience button for selection to exclude wireless and network

...and system
Almost as dangerous, at least the LED config etc.

All other config files that e.g. affect VPN, DNS can be difficult too, as the lan/wan port names in uci config may be quite different.

The selective restore from another device can be rather challenging.

hnyman:
I feel that more useful than "partial restore into the actual config" could be an easy "restore as reference material to /etc/oldconfig" or something similar. So that some SSH console diff/cp/mv work would be needed, but getting the files to the file system would be easy (and safe).

systemcrash:
I looked at that bit: luci loads every file in the /etc/config folder and differing files with synonymous entries would clash. So such a restore requires placement in a different folder. A bit more complicated.

systemcrash:
@hnyman regardin LED config - what is your experience of things there? Potential of loading wrong kernel modules?

hnyman:
Not that dangerous, but network and usb LEDs might be defined wrongly.

modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
```diff
@@ -169,6 +248,94 @@ return view.extend({
...
+		}
+
+		btn.firstChild.data = _('Creating archive of selected files…');
+
+		// Write the selected file names into a temporary file for tar -T option
+		var fileList = selectedFiles.join('\n');
+		return fs.write('/tmp/filelist.txt', fileList)
+			.then(L.bind(function() {
+				// Make selective restore directory
+			return fs.exec('/bin/mkdir', ['-p', '/tmp/selective_restore/']);
```
openwrt-ai, line 268: This line is under-indented by one tab relative to its enclosing `.then(L.bind(function() {` block and the comment above it. The rest of the file uses tab indentation aligned to block depth.

```suggestion
				return fs.exec('/bin/mkdir', ['-p', '/tmp/selective_restore/']);
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
```diff
@@ -169,6 +248,94 @@ return view.extend({
...
+
+				// Await reconnection
+				ui.awaitReconnect(window.location.host, '192.168.1.1', 'openwrt.lan');
+			}, this))
+			.catch(function(e) { 
+				ui.addNotification(null, E('p', e.message)); 
+			})
+			.finally(function() { 
+				btn.firstChild.data = _('Upload archive...'); 
+			});
```
openwrt-ai, line 336: nit: trailing whitespace on the `.catch`/`.finally` lines and their bodies. The sibling `handleRestoreConfirm` writes these on a single line without trailing spaces.

```suggestion
			.catch(function(e) {
				ui.addNotification(null, E('p', e.message));
			})
			.finally(function() {
				btn.firstChild.data = _('Upload archive...');
			});
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
```diff
@@ -169,6 +248,94 @@ return view.extend({
...
+		if (selectedFiles.length === 0) {
+			ui.addNotification(null, E('p', _('No files selected for restoration')));
+			return fs.remove('/tmp/backup.tar.gz').finally(ui.hideModal);
+		}
+
+		btn.firstChild.data = _('Creating archive of selected files…');
+
+		// Write the selected file names into a temporary file for tar -T option
+		var fileList = selectedFiles.join('\n');
+		return fs.write('/tmp/filelist.txt', fileList)
```
openwrt-ai, line 265: On every error path here (the `L.raise(...)` after a failed `mkdir`/`tar`/`sysupgrade`, or any rejection caught below) the temp artifacts created by this flow — `/tmp/filelist.txt`, `/tmp/selective_restore/` and `/tmp/selective_restore.tar.gz`, plus the uploaded `/tmp/backup.tar.gz` — are left on disk. The plain `handleRestoreConfirm` only ever leaves `/tmp/backup.tar.gz`, but this flow accumulates several more. Consider cleaning them up in the `.catch`/`.finally` (e.g. `fs.remove` the staging dir and intermediate files) so repeated failed attempts don't pile up partial state in `/tmp`.

---

openwrt-ai (commented):
Reviewed the single commit at head `58a69ea`. The commit message matches the diff, so no commit-check items. Inline notes below cover an indentation slip, trailing whitespace, and temp-file cleanup on error paths. Noted that this is an RFC and the follow-up rework tied to openwrt/openwrt#16656 (already called out in the description) supersedes the two-step extract/recompress approach.

---

PR #7320, opened 2024-10-10
luci-app-squid: convert to JavaScript

dannil:
The conversion adds the missing parameters in the UI which is configurable in the principal package.

Assumption has been made that the config file is stored in /etc/squid since the principal package limits the sysconfdir to this directory. If that assumption is changed in the future we need to adjust the ACL.

Screenshot_1
Screenshot_2
Screenshot_3

Values are updated appropriately which the squid logs also indicates.

```
root@labmouse:~# uci show squid
squid.squid=squid
squid.squid.config_file='/etc/squid/squid.conf'
squid.squid.http_port='12345'
squid.squid.coredump_dir='/tmp/customsquid'
squid.squid.visible_hostname='foobar'
squid.squid.mime_table='/etc/squid/mime.conf'
squid.squid.pinger_enable='off'
```

```
Thu Oct 17 18:03:13 2024 daemon.notice squid[4163]: Processing Configuration File: /tmp/squid/squid.conf (depth 0)
Thu Oct 17 18:03:13 2024 daemon.notice squid[4163]: ERROR: 'pinger_enable' requires --enable-icmp
Thu Oct 17 18:03:13 2024 daemon.warn squid[4163]: FATAL: Squid is already running: Found fresh instance PID file (/var/run/squid.pid) with PID 4095     exception location: Instance.cc(122) ThrowIfAlreadyRunningWith
Thu Oct 17 18:03:13 2024 daemon.notice squid[4095]: Preparing for shutdown after 0 requests
Thu Oct 17 18:03:13 2024 daemon.notice squid[4095]: Waiting 30 seconds for active connections to finish
Thu Oct 17 18:03:13 2024 daemon.notice squid[4095]: Closing HTTP(S) port [::]:1337     listening port: 1337
Thu Oct 17 18:03:18 2024 daemon.info procd: Instance squid::instance1 pid 4095 not stopped on SIGTERM, sending SIGKILL instead
Thu Oct 17 18:03:18 2024 daemon.notice squid[4220]: Processing Configuration File: /tmp/squid/squid.conf (depth 0)
Thu Oct 17 18:03:18 2024 daemon.notice squid[4220]: ERROR: 'pinger_enable' requires --enable-icmp
Thu Oct 17 18:03:18 2024 daemon.notice squid[4220]: Created PID file (/var/run/squid.pid)
Thu Oct 17 18:03:18 2024 daemon.notice squid[4220]: Set Current Directory to /tmp/customsquid
[...]

PR #7324, opened 2024-10-12
luci-app-tinyproxy: Translate to JS

Ramon00:
luci-app-tinyproxy: Translate to JS

Signed-off-by: Ramon Van Gorkom <Ramon00c00@gmail.com>


Ramon00:
So I think its fully working, except if you apply the settings then the settings only get saved, not reloaded. I tried with the ucitrack file, but somehow that does not work. Can somebody have a look? 
Easy alternative is adding a button to do a manual reload, but that is not so nice...

systemcrash:
Take a look at what @dannil just did in squid.

>
> Can somebody have a look?
> Easy alternative is adding a button to do a manual reload, but that is not
> so nice...
>
>
>

Ramon00:
> Take a look at what @dannil just did in squid.

I tried something like that already, it sort of works, but somehow not always, maybe a timing thing?

Ramon00:
i have to hit the save and apply button twice then it always works

Ramon00:
Ah seems to work now. If somebody can test the new app that would be great

Ramon00:
That was a bit premature. It still sometimes does not work. Seems the `ui.changes.apply` returns before it is actually saved, then the `/etc/init.d/tinyproxy restart` still reads the old config.

Ramon00:
> Yeah, it's a timing issue, it doesn't work in my PR as well (I didn't see it at first since squid takes a while to startup since it checks for existing PID during startup which takes about 2 seconds sometimes), I had naively assumed that the `apply` function returned it's promise so it could be awaited on but looking at the code it doesn't so we need to think of another solution.

Just doing `uci.apply()` works at least here, but that leaves the "unchange changes" in the header until you refresh the page.
So that is not ideal either

dannil:
> > Yeah, it's a timing issue, it doesn't work in my PR as well (I didn't see it at first since squid takes a while to startup since it checks for existing PID during startup which takes about 2 seconds sometimes), I had naively assumed that the `apply` function returned it's promise so it could be awaited on but looking at the code it doesn't so we need to think of another solution.
> 
> Just doing `uci.apply()` works at least here, but that leaves the "unchange changes" in the header until you refresh the page. So that is not ideal either

Yeah, the reason why I used `ui.changes.apply()` is that's how `handleSaveApply` does it if you don't implement a custom handler for it, then you get functionality like checking that persisting actually succeeded and rolling back if it didn't as a bonus which we really shouldn't implement on the app level.

Ramon00:
> 
> Yeah, the reason why I used `ui.changes.apply()` is because that's how `handleSaveApply` does it if you don't implement a custom handler for it, then you get functionality like checking that persisting actually succeeded and rolling back if it didn't as a bonus which we really shouldn't implement on the app level. IMO not using facilities like that would be a big regression in the conversion.

We can call a `ui.changes.renderChangeIndicator()` afterwards to clear it, but then it is a sort of unchecked apply, which will also apply other changes unchecked. So not a real solution.

Ideally `ui.fs` should be extended with a check function if the apply is finished.

Ramon00:
@dannil I may have found a solution:
`		this.super('handleSaveApply', [ev])
			.then(() => fs.exec_direct('/etc/init.d/tinyproxy', ['restart']))`

That seems to work, not exactly sure why, maybe due to a file lock from ubus? Or maybe it is just "luck" (i.e. slower). Would be great if we can make sure

dannil:
> @dannil I may have found a solution: ` this.super('handleSaveApply', [ev]) .then(() => fs.exec_direct('/etc/init.d/tinyproxy', ['restart']))`
> 
> That seems to work, not exactly sure why, maybe due to a file lock from ubus? Or maybe it is just "luck" (i.e. slower). Would be great if we can make sure

Interesting... I assume tinyproxy doesn't have a procd reload trigger as I wrote about in https://github.com/openwrt/luci/pull/7320#issuecomment-2409093435? That would be a way to solve it but would require changes in the principal package, but it's a nice solution since then changes, irregardless of it's done via the CLI or LuCI, is handled the same.

Ramon00:
I think that was premature again... sigh. 
But new idea, reproduce the "unsaved changes" value. Seems to be in  uci.changes()  [https://github.com/openwrt/luci/blob/84e24c08699ab28e077259756cdd5a5b40cfd2a6/modules/luci-base/htdocs/luci-static/resources/ui.js#L4430](url)
going to experiment with that

Ramon00:
Ok 100% elegant solution, add something like:
`		document.addEventListener('uci-applied',function() {
			fs.exec('/etc/init.d/tinyproxy', ['restart']);
		});`

then you can leave the `handleSaveApply` default.

Ramon00:
@systemcrash now that it is fixed, can you give it a test?

Ramon00:
OK test please!

systemcrash:
Just curious - does `reload` work better than `restart`? Sometimes `restart` is a more substantial operation. And `reload` complets in a much shorter time.

Ramon00:
both seem to work equally. Want me to make into reload?

systemcrash:
Could you try using `reload`? I think that usage should be encouraged unless there's a "yeah, unplug and restart is necessary". Maybe a separate reload convenience button available also. There are a few examples of those in the apps. @stangri tends to put those in :)

Ramon00:
Yes reload works I already tested that.  I can add a reload and a restart button. No problem. Will do that tomorrow though.

Ramon00:
How about this: 
image

Ramon00:
or this maybe: 
image
Not sure how to get rid of the large vertical space in the imported html... I stripped all styles and attributes from the loaded html

systemcrash:
Those look great.

systemcrash:
Merged. Thanks @Ramon00 !

PR #7329, opened 2024-10-16
luci-app-attendedsysupgrade: update status keys and messages

efahl:
Maintainer: @aparcar
Tested on: x86/64 SNAPSHOT build

The reported status values have changed on the ASU server several times since the 'steps' list has been updated, so update it to include all possible values.  Although some of the values in the list are no longer presented by the official OpenWrt ASU server, they may still be produced by other ASU servers, so leave them in for now.

Percentages have been adjusted based upon measurement of the build phases for several build configurations (ath79 vs x86, few packages vs many...).  The numbers are still (very) rough estimates, but are more representative than the old values.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -60,11 +60,19 @@ function get_revision_count(revision) {
 
 return view.extend({
 	steps: {
-		init: [10, _('Received build request')],
-		download_imagebuilder: [20, _('Downloading ImageBuilder archive')],
-		unpack_imagebuilder: [40, _('Setting Up ImageBuilder')],
-		calculate_packages_hash: [60, _('Validate package selection')],
-		building_image: [80, _('Generating firmware image')],
+		init:                    [  0, _('Received build request')],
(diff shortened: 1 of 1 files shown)


dannil:
LGTM. I assume we should wait for https://github.com/openwrt/asu/pull/1037 to be merged and deployed to sysupgrade.openwrt.org? Otherwise we'll break the LuCI app on SNAPSHOT.

efahl:
@dannil , thanks for looking at this.

I dug back through the ASU server revisions a couple months and couldn't find where those old status values were dropped, so we've been running without them for a while.  But, I know ImmortalWrt is running an older version ASU server, so it might still be producing them and I wanted to let their server still work when they upgrade their LuCI app...

aparcar:
Should I merge this?

dannil:
> @dannil , thanks for looking at this.
> 
> I dug back through the ASU server revisions a couple months and couldn't find where those old status values were dropped, so we've been running without them for a while. But, I know ImmortalWrt is running an older version ASU server, so it might still be producing them and I wanted to let their server still work when they upgrade their LuCI app...

If they're dropped it should be fine, the only weirdness I could see was that with the new UI the `unpack_imagebuilder` status has the same progress percentage as the newly modified `calculate_packages_hash`, which meant that the progress bar would stay still moving from `unpack_imagebuilder` to `calculate_packages_hash` if you ran the new LuCI app and the old ASU server, but if they're already removed from ASU anyway before this then it's not a problem.

PR #7340, opened 2024-10-22
luci-app-opkg: fix passing wrong option on opkg update/install

Ansuel:
Fix passing wrong option on opkg update/install. While starting to introduce support for APK in the opkg module, it was notice that --force-removal-of-dependent-packages was always passed even with update and install command.

This was probably a leftover/oversight of old one. To fix this, limit this option only on remove and also update the acl.d to support single call to update or install.

Fixes: 9b25031cb29b ("luci-app-opkg: rework backend operations")


Ansuel:
@jow- didn't want to investigate further but why "--force-removal-of-dependent-packages" is even forced with --remove?

systemcrash (approved):

systemcrash:
This is ready to go in as-is?

systemcrash:
Closed by 3a8e9baf32242f30cf65791b61c460413bd115ce

PR #7341, opened 2024-10-22
luci-app-package-manager: rename from luci-app-opkg and add APK support

Ansuel:
Rename luci-app-opkg to luci-app-ipkg and add APK support for it.

The idea is to adapt APK to mimic OPKG output to require minimal changes
to the luci app.

---

@jow- @aparcar @hauke 

TODO:
- [x] Config handling
- [x] Test install/remove
- [ ] Test update scenario
- [x] Handle --autoremove 
- [x] Handle --force-overwrite 
- [x] Handle --force-removal-of-dependent-packages

What works:
- [x] List packages
- [x] Update package list

Depends on:
https://github.com/openwrt/openwrt/pull/16759


systemcrash:
Does this mean all rpc calls or otherwise to `opkg` must be updated?

stangri:
> Rename luci-app-opkg to luci-app-ipkg and add APK support for it.

Why `luci-app-ipkg`? Why not `luci-app-pkg-manager`/`luci-app-package-manager` for example?

systemcrash:
@stangri 

> Why `luci-app-ipkg`? Why not `luci-app-pkg-manager`/`luci-app-package-manager` for example?

standard naming convention ig.

Ansuel:
> Does this mean all rpc calls or otherwise to `opkg` must be updated?

all rpc calls -> none

OPKG is enough sensible to wrap it in helper script and we already minimize it to just opkg-call script hence it's all handled there, we don't call opkg directly.

Ansuel:
> > Rename luci-app-opkg to luci-app-ipkg and add APK support for it.
> 
> Why `luci-app-ipkg`? Why not `luci-app-pkg-manager`/`luci-app-package-manager` for example?

yes can be an idea but for sure the name needs to be changed to something more generic.

hnyman:
+1 for a clear name that is also understandable by newcomers.

`luci-app-packagemanager` would be my favorite

lorand-horvath:
**`luci-app-opm`** - short for Openwrt Package Manager

**`luci-app-pm`** - even shorter

castillofrancodamian:
> +1 for a clear name that is also understandable by newcomers.
> 
> `luci-app-packagemanager` would be my favorite

Or `luci-app-package-manager`. I like it better.

systemcrash:
Full names tend to serve better than ambiguous acronyms and abbreviations for first-timers. 

I like luci-app-package-manager

Ansuel:
Actually this went pretty fast... Should be finished... Maybe the po file needs some bit of touchup...

systemcrash:
@Ansuel 
```sh
build/i18n-sync.sh modules/luci-app-package-manager
```

systemcrash:
Closed by bcd13b918e2f30b8d19027a06e3d773a1b0ec4c0

Ansuel:
@stokito thanks for the comments I will take care of fixing them

systemcrash:
Yo @Ansuel can this get cherry-picked to 23 or is there stuff it requires from master?

Ansuel:
not at all... 23 doesn't even have working apk. Why should we backport?

systemcrash:
Ah, maybe I can just take the useful fix from #7340 to 23?

Ansuel:
@systemcrash yep that was why it was in a dedicated commit :D

Issue #7339, opened 2024-10-22
luci-proto-wireguard: incorrect peer name detection when more than one peer with the same public key exist


this-username-has-been-taken:
If an OpenWrt device is used as a Wireguard client (i.e. no "listen port" is defined for the interface) it is allowed to define more than one peer with the same public key. In such case the peer name in `Status → WireGuard` menu might not be detected correctly because the code will only compare the public key of a peer and take the first match. To fix the issue it is required to check not only the public key of a peer, but also its `endpoint_host` and `endpoint_port`.
***
## Possible solution (doesn't work with FQDNs - see comments below):
Replace line 79 in `luci/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard` file:
```
if (s.public_key == record[1])
```
with the following:
```
if (s.public_key == record[1] && s.endpoint_host + ":" + s.endpoint_port == record[3])
```
The solution is tested and works if the `endpoint_host` is an IP address. ~I didn't test it if the `endpoint_host` is an FQDN-name.~
UPD: Tested FQDN - doesn't work. See comments below.

## Steps to reproduce:

1. Go to: Network → Interfaces → add a new WireGuard interface. The interface should not contain `listen port`, i.e. act as a client. Fill all other parameters (`private key`, `ip address`) with correct values.
2. Add a fake peer on the `Peers` tab with name `Incorrect peer 1`; set `Disabled` property `true`; fill `Public key` with a real public key value (see further); endpoint host and port with random values.
3. Optional. Add another fake Peer with the same parameters, but different name, e.x. `Incorrect peer 2`.
4. Add a real peer with the same public key (i.e. all peers must have the same public key from the real one) with name `Correct peer`; set `Disabled` property `false`; fill endpoint host and port with correct values.
5. The order of peers is critical, i.e. create fake peers first and the real one last.
6. Save and apply.
7. Set the new Wireguard interface up and running (i.e. connect to the real peer).
8. Go to `Status → WireGuard` page.
9. Observe the issue: the page will show Peer: `Incorrect peer 1` and all other parameters from the real peer.

## Actual behavior:
The `Status → WireGuard` page shows incorrect peer name with all other correct parameters.

## Expected behavior:
The `Status → WireGuard` page shows correct peer name and all other correct parameters.

## Additional Information:
Doesn't matter, but here it is:
```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='SNAPSHOT'
[...]

PR #7342, opened 2024-10-23
luci-proto-wireguard: fixed bug with incorrect peer name detection

this-username-has-been-taken:
- [X] Each commit has a valid :black_nib: `Signed-off-by: this-username-has-been-taken <119663930+this-username-has-been-taken@users.noreply.github.com>` row (via `git commit --signoff`)
- [X] Tested on: OpenWrt SNAPSHOT r27859-f9da81d32f; mediatek/filogic; aarch64_cortex-a53; Firefox
- [X] \( Preferred ) Mention: @systemcrash 
- [X] \( Optional ) Closes: https://github.com/openwrt/luci/issues/7339
- [X] Description:

Fixed bug with incorrect peer name detection on `Status -> WireGuard` page when more than one peer with the same public key exist:
1. Peers are now tested not only by public key, but also by enabled/disabled status, peer host (both IP and FQDN are supported) and port.
2. Added required `resolveip` dependency.

### DIFF
--- a/protocols/luci-proto-wireguard/Makefile
+++ b/protocols/luci-proto-wireguard/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=Support for WireGuard VPN
-LUCI_DEPENDS:=+wireguard-tools +ucode +luci-lib-uqr
+LUCI_DEPENDS:=+wireguard-tools +ucode +luci-lib-uqr +resolveip
 LUCI_PKGARCH:=all
 
 PKG_LICENSE:=Apache-2.0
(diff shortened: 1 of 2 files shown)


systemcrash:
Please fix your email.

this-username-has-been-taken:
I'd like to keep the email as well as my credentials private - that's why I have enabled privacy option in Github settings.
AFAIK this email substitute still works, Github will forward messages to me.

systemcrash:
OK. Done. In future, just invent a name and email address. :)

jow-:
This change introduced a potential root shell code injection. The `configHost` argument is taken verbatim from the uci configuration and passed unquoted/unescaped to `resolveip`.

this-username-has-been-taken:
You are right: there is no point in looking for an IP when we do not have an IP yet.
```
function checkPeerHost(configHost, configPort, wgHost) {
	const ips = popen(`resolveip ${shellquote(configHost)} 2>/dev/null`);
	...
```
Did I get your idea right?

jow-:
Yep, that should be fine. To be extra safe perform a few checks by manually placing peer names like `'option endpoint_host '$(logger injection)'` in /etc/config/network

PR #7375, opened 2024-11-02
luci-app-package-manager: Fix upgrade action for apk

hnyman:
cc @Ansuel @aparcar 

Fix upgrade action for apk by passing an attribute via 'handleInstall' to set correctly the command for 'handlePkg' in order to differentiate between install and upgrade actions.  (Note: 'handleManualInstall' sets always 'install')
Adjust acl accordingly.

Remove extra whitespace.

Tested on live system for 
* both apk & opkg enables systems, 
* both install (single package and packages with dependencies to install) and upgrade.

apk
```
Executing package manager

(1/1) Upgrading luci-app-adblock (24.284.61488~02f2d47 -> 24.304.46082~f9899fd)
OK: 70 MiB in 307 packages
```
opkg
```
Executing package manager

Upgrading ca-bundle on root from 20230311-1 to 20240203-r1...
Downloading https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/base/ca-bundle_20240203-r1_all.ipk
Configuring ca-bundle.
```

### DIFF
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -261,6 +261,7 @@ function display(pattern)
 			btn = E('div', {
 				'class': 'btn cbi-button-positive',
 				'data-package': name,
+				'data-action': 'upgrade',
 				'click': handleInstall
 			}, _('Upgrade…'));
 		}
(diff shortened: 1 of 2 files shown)


applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -661,6 +664,7 @@ function handleReset(ev)
 function handleInstall(ev)
 {
 	var name = ev.target.getAttribute('data-package'),
+	    installstyle = ev.target.getAttribute('data-action'),
```
Ansuel, line 667: would call the variable cmd.

hnyman: I wanted to name it differently than 'cmd' for clarity, as handlePkg function already has 'cmd'

hnyman: I will change it to installcmd

PR #7376, opened 2024-11-03
luci-app-package-manager: Show the executed command

hnyman:
Clarify the output of the LuCI package manager actions: 
show also the exact command passed to apk & opkg
(in addition to output & errors from the command)
That will help debugging.

Examples:

installation:
image

removal:
image

image


systemcrash:
Good additions

Ansuel:
Mhh maybe we can just change the call tool to append the command before the result?

hnyman:
You mean in style of
image

or
image


Clear enough. I will change that.

Ansuel:
no i mean to modify the package-manager-call script instead of the .js to include the command in the res json. That way it's autoincluded as first line in the output.

hnyman:
I actually like it in the way above where the command is in its own box, slightly separated from the results.

Ansuel:
> I actually like it in the way above where the command is in its own box, slightly separated from the results.

yep ok for me

PR #7386, opened 2024-11-12
luci-mod-dashboard: custom.css: Add color variables

griffenedge:
Add color variable references from luci-theme-boostrap to facilitate light and dark color scheme switching by default. Fallback to original colors when variables are not declared by theme.

Screenshot comparison of dark color scheme before and after changes

Screenshot comparison of light color scheme before and after changes

### DIFF
--- a/modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/css/custom.css
+++ b/modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/css/custom.css
@@ -3,11 +3,11 @@
 **/
 
 .Dashboard {
-    color: #212529!important;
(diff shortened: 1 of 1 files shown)


systemcrash:
So.. @stokito @dannil does my memory serve me correctly when I think we actually got rid of 'variables' a while back?

dannil:
> So.. @stokito @dannil does my memory serve me correctly when I think we actually got rid of 'variables' a while back?

No, they're still there. I double-checked the variables as well, they're present for the Bootstrap theme (see @stokito's link above).

griffenedge:
It's unclear to me why this request was closed. Would anyone be able to elaborate? @systemcrash @stokito @dannil

Issue #2112, opened 2018-08-25
LCP echo failure threshold 0  does not behave as described in LuCI


bill888uk:
For BT Home Hub 5A configured to use PPPoE or PPPoA protocol for example.

In **LuCI→Network→Interfaces→WAN→Advanced Settings,** there are two default settings displayed:

LCP echo failure threshold    0
LCP echo interval    5

On the page, the 'default' setting **LCP echo failure threshold** to Zero, implies all LCP echo failures would be ignored. (see attached image) But I found this to be incorrect/misleading. 

There is no **keepalive** option found in the WAN interface section of /etc/config/network on Home Hub 5A when selecting PPPoE or PPPOA protocols.  But there appears to be a fixed timeout (possibly 5 seconds) on the Home Hub 5A when I use PPPoE on VDSL connection.  ie. not infinite as quoted in LuCI.

I don't know if this is the same behaviour for all devices.

If **LCP echo failure threshold** is set to a non-Zero value, then **keepalive** option is CREATED in the WAN interface within /e/c/network file.

The description displayed in LuCI for LEDE and OpenWRT 18: 

    'Presume peer to be dead after a given amount of LCP echo failures, use 0 to ignore failures'

perhaps needs to be changed?

eg. perhaps simplest solution is to remove the phrase **'use 0 to ignore failures'**  ?

This closed ticket may be of interest  (I added my observations affecting HH5A running LEDE 17 to end of the closed ticket)
https://bugs.openwrt.org/index.php?do=details&task_id=1259

Would it be possible to add  'keepalive_adaptive' setting into LuCI mentioned in above ticket in the future?

Another ticket on same subject
https://bugs.openwrt.org/index.php?do=details&task_id=854

![0lcpechofailure](https://user-images.githubusercontent.com/26403991/44621333-7ec4ee80-a89c-11e8-88bd-829fa473eaaf.jpg)

jow-:
From `ppp.sh`:

```
        local lcp_failure="${keepalive%%[, ]*}"
        local lcp_interval="${keepalive##*[, ]}"
        ...
        [ "${lcp_failure:-0}" -lt 1 ] && lcp_failure=""
        ...
        proto_run_command "$config" /usr/sbin/pppd \
                ${lcp_failure:+lcp-echo-interval $lcp_interval lcp-echo-failure $lcp_failure $lcp_adaptive} \
```

[...]

PR #7392, opened 2024-11-14
luci-proto-3g/ppp/pppossh: fix being unable to set keepalive to 0

Preport:
Since on openwrt keepalive option defaults to "5 1" when it's not defined 
(see https://github.com/openwrt/openwrt/blob/6720c4ccba256186bf2f1b1edadb851c447e62a5/package/network/services/ppp/files/ppp.sh#L128).
Users must be able to set it to 0 to ignore connection failures.

- [x] Tested on: (mt7621 mipsel, openwrt 23.05.04 firefox) :white_check_mark:
- [x] Mention: @jow- 
- [x] Closes: openwrt/luci#2112

### DIFF
--- a/protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
+++ b/protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
@@ -26,13 +26,13 @@ function write_keepalive(section_id, value) {
 	    f = (f_opt != null) ? +f_opt[0].formvalue(section_id) : null,
 	    i = (i_opt != null) ? +i_opt[0].formvalue(section_id) : null;
 
-	if (f == null || f == '' || isNaN(f))
-		f = 0;
+	if (f === '' || isNaN(f))
+		f = null;
 
 	if (i == null || i == '' || isNaN(i) || i < 1)
 		i = 1;
 
-	if (f > 0)
+	if (f !== null)
 		uci.set('network', section_id, 'keepalive', '%d %d'.format(f, i));
 	else
 		uci.unset('network', section_id, 'keepalive');
--- a/protocols/luci-proto-ppp/htdocs/luci-static/resources/protocol/ppp.js
+++ b/protocols/luci-proto-ppp/htdocs/luci-static/resources/protocol/ppp.js
@@ -26,13 +26,13 @@ function write_keepalive(section_id, value) {
 	    f = (f_opt != null) ? +f_opt[0].formvalue(section_id) : null,
 	    i = (i_opt != null) ? +i_opt[0].formvalue(section_id) : null;
 
-	if (f == null || f == '' || isNaN(f))
-		f = 0;
+	if (f === '' || isNaN(f))
+		f = null;
 
 	if (i == null || i == '' || isNaN(i) || i < 1)
 		i = 1;
 
-	if (f > 0)
+	if (f !== null)
 		uci.set('network', section_id, 'keepalive', '%d %d'.format(f, i));
 	else
 		uci.unset('network', section_id, 'keepalive');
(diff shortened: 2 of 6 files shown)


protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
```diff
@@ -26,13 +26,13 @@ function write_keepalive(section_id, value) {
 	    f = (f_opt != null) ? +f_opt[0].formvalue(section_id) : null,
 	    i = (i_opt != null) ? +i_opt[0].formvalue(section_id) : null;
 
-	if (f == null || f == '' || isNaN(f))
-		f = 0;
+	if (f === '' || isNaN(f) || f < 0)
```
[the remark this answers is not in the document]
Preport: You are right, I forgot that it was a uint

protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
```diff
@@ -26,13 +26,13 @@ function write_keepalive(section_id, value) {
 	    f = (f_opt != null) ? +f_opt[0].formvalue(section_id) : null,
 	    i = (i_opt != null) ? +i_opt[0].formvalue(section_id) : null;
 
-	if (f == null || f == '' || isNaN(f))
-		f = 0;
+	if (f === '' || isNaN(f))
+		f = null;
 
 	if (i == null || i == '' || isNaN(i) || i < 1)
```
systemcrash, line 32: Can't interval be zero also? I don't know for sure. I get the impression that setting it to zero should turn it off. Does that happen in the init scripts?

https://github.com/ppp-project/ppp/blob/616102e93be1c629821cafd83a89d9e6c0467033/pppd/lcp.c#L2492-L2493

Preport: As long as failure is less than zero it should work.

https://github.com/openwrt/openwrt/blob/6720c4ccba256186bf2f1b1edadb851c447e62a5/package/network/services/ppp/files/ppp.sh#L133

https://github.com/openwrt/openwrt/blob/6720c4ccba256186bf2f1b1edadb851c447e62a5/package/network/services/ppp/files/ppp.sh#L145

systemcrash: > As long as failure is less than zero it should work.
> 
> https://github.com/openwrt/openwrt/blob/6720c4ccba256186bf2f1b1edadb851c447e62a5/package/network/services/ppp/files/ppp.sh#L133

This just means that setting it to zero disables it (in the init scipt) - which is inaccessible with the`i < 1` behaviour. I'm not nit-picking your code- just what's going on here in general. That whole logic could do with a rewrite. Are you game?

systemcrash:
@feckert do you also use some flavour of PPP?

protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
Preport: It can't be less than 1 on luci since there's a check for that
https://github.com/openwrt/luci/blob/b9496f2023cb6b5f51dff88f1cf3a3063390dae4/protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js#L145

But if we want users to be able to disable keepalive through the interface by either setting the failure or/and the interval to 0 then yes we'd need to edit the init script too.

I think the init script should be edited regardless to always set the keepalive instead of omitting it when it's 0 since that way pppd may use whatever it's set to on options file.
https://github.com/openwrt/openwrt/blob/6720c4ccba256186bf2f1b1edadb851c447e62a5/package/network/services/ppp/files/etc/ppp/options

systemcrash:
Closed by f3d26a2a560d8ae526c617eb1c28821ec900fa7a

PR #7393, opened 2024-11-15
[luci-app-statistics] fix prepare values from list vlaues

Softovick:
If UCI store  values as a list instead of an option, conversation for the collectd configuration is not work correctly. This is due to the use of a DynamicList type element in the UI (for example, for the RRATimespans field). For this function argument val receives as array instead of a string, so no additional conversion is required.

- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch ✅
- [x] Tested on: (AARCH64, openwrt 23.05) ✅
uci_list_rratimespan


systemcrash:
can you reformat the commit message please ? (line breaks at 80 chars)

Softovick:
> can you reformat the commit message please ? (line breaks at 80 chars)

Yes. Updated, please take a look, is it right now?

PR #7394, opened 2024-11-15
luci-app-keepalived: add notification tab

Chris1189:
This commit adds a tab to modify keepalived.user file. With this commit it is possible to specify notifications.
image

Maintainer: @jempatel


systemcrash:
merged - thanks @Chris1189

PR #7396, opened 2024-11-15
ci: fix build error and account for APK as default

dannil:
See separate commits for details. The first commit should fix all the recent build errors for the task `Test x86_64` ([1](https://github.com/openwrt/luci/actions/runs/11852952401/job/33032096379?pr=7394), [2](https://github.com/openwrt/luci/actions/runs/11764799075/job/32770355810), [3](https://github.com/openwrt/luci/actions/runs/11795975811/job/32856854484)). For reviewers, make sure to check that the workflow run is green and produces the expected results.

* ci: upgrade openwrt/gh-action-sdk workflow to v7

    With the change of the Docker container not including pre-built
    binaries [1], the previous version of the workflow fails as it still
    expects those to be present in the upstream container while the new
    versions runs a file called setup.sh if present to configure the
    container.

    By upgrading openwrt/gh-action-sdk workflow to v7, we make sure that
    setup.sh is run and configures the container [2].

    [1] https://github.com/openwrt/docker/commit/9b55784b18f8d2c684aac4dd21a224320bb9b9ce
    [2] https://github.com/openwrt/gh-action-sdk/commit/0c00b28c11bed5c7697e807d4cb8ef176ca688ef

* ci: account for APK as default on main branch

    With APK defaulted on openwrt:main, we should include packages with the
    .apk file extension as artifacts. Also default the logging to V=s as
    otherwise the APK version check won't be printed to stdout.

- [ ] Tested on: (architecture, openwrt version, browser) :white_check_mark: N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,15 +54,18 @@ jobs:
         env:
           ARCH: ${{ matrix.arch }}-${{ env.BRANCH }}
           FEEDNAME: packages_ci
+          V: s
 
       - name: Move created packages to project dir
-        run: cp bin/packages/${{ matrix.arch }}/packages_ci/*.ipk . || true
+        run: cp bin/packages/${{ matrix.arch }}/packages_ci/* . || true
 
       - name: Store packages
         uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.arch}}-packages
-          path: "*.ipk"
+          path: |
+            *.ipk
+            *.apk
 
       - name: Store logs
         uses: actions/upload-artifact@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
           echo "PACKAGES=$PACKAGES" >> $GITHUB_ENV
 
       - name: Build
-        uses: openwrt/gh-action-sdk@v5
+        uses: openwrt/gh-action-sdk@v7
         env:
           ARCH: ${{ matrix.arch }}-${{ env.BRANCH }}
           FEEDNAME: packages_ci


dannil:
For reference, here's how the APK version check looks like today, exemplified by manipulating `PKG_VERSION` for `luci-app-adblock-fast` to an invalid value.

* today: https://github.com/dannil/luci/commit/6fd7271edb15403b77c68b2fb7fe7ebe0d1938c6: https://github.com/dannil/luci/actions/runs/11864787802/job/33068858146#step:5:8635
* with `V=s`: https://github.com/dannil/luci/commit/5b6f1d2231a16e1222df5704bff5e3d71c39d5d0: https://github.com/dannil/luci/actions/runs/11864801201/job/33068895111#step:5:16424

.github/workflows/build.yml
```diff
@@ -54,15 +54,18 @@ jobs:
         env:
           ARCH: ${{ matrix.arch }}-${{ env.BRANCH }}
           FEEDNAME: packages_ci
+          V: s
 
       - name: Move created packages to project dir
-        run: cp bin/packages/${{ matrix.arch }}/packages_ci/*.ipk . || true
+        run: cp bin/packages/${{ matrix.arch }}/packages_ci/* . || true
```
hnyman, line 60: I wonder if this should be `*.?pk` instead ?  
To match both `.apk` and `.ipk`, but not removing the selection criteria totally leaving just the `*` wildcard there.

Ansuel: Should we bite the bulled and be more explicit?
```
cp bin/packages/${{ matrix.arch }}/packages_ci/*.ipk . || true
cp bin/packages/${{ matrix.arch }}/packages_ci/*.apk . || true
```

systemcrash: Prefer one-liners :D

dannil: I'm fine with either solution, I don't have any real preference as long as we can capture the .apk files.

dannil: I saw that actions-shared-workflows uses the [same principle](https://github.com/openwrt/actions-shared-workflows/blob/main/.github/workflows/multi-arch-test-build.yml#L119) and then [copies the relevant files from the working directory](https://github.com/openwrt/actions-shared-workflows/blob/main/.github/workflows/multi-arch-test-build.yml#L165-L171) so maybe we should keep it the way it is now?

Ansuel: Well if we want to keep consistency yes, we should use the * directly.

systemcrash:
Awesome

PR #7399, opened 2024-11-16
luci-light: move firewall and ppp to luci collection

olemmela:
Allow luci compilation without firewall and ppp proto. A Firewall or ppp is not needed for the AP basic functions.

This way it is possible to save memory and flash space in low memory enviroments.


hnyman:
I don't like the removal of the firewall. It is so commonly used.

You probably didn't notice, but your change would have removed the firewall also from both LuCI SSL collections, as they depend on luci-light.

If you really need minimal LuCI, just select the individual components but not firewall.

olemmela:
I did not notice that collections are not mandatory. And all needed Luci component can be also individually selected.

I'll close this PR

PR #7400, opened 2024-11-16
luci-app-package-manager: support new APK repositories.d configs

Ansuel:
Support new APK repositories.d configs following OPKG was to better split distfeeds and customfeeds.

Support for ""legacy"" repositories is kept to support migration period and to support custom scenario where user use this file instead.


hnyman:
Seems to work. 
I built a test build with this included, and editing both files works ok.

Ansuel:
Think this should be merged ASAP if everyone is ok with the format for the additional 2 files...

I didn't want to complicate this too much with introducing logic to list files in directory (as that would also list files in the keys directory and is a security risk)

hnyman:
Yeah, like I said yesterday, looks good.
Just merge it.

PR #7415, opened 2024-11-21
luci-base: make items of UIDynamicList drag-sortable

Ramon00:
This PR makes items in UIDynamicList drag-sortable. 

Should work as is, but could use two improvements:
1) does not work on mobile. Maybe this can be fixed in CSS?  --> Update: Basic touch functionality added with last commit
2) Click detection is pixel width based, as the `::after` is not  accessable from js (or at least i do not know a method).  To fix this would require a lot of code refactoring.

Any suggestions, comments, or improvements appreciated,

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2715,7 +2715,6 @@ var CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection.p
 	handleDragStart: function(ev) {
 		if (!scope.dragState || !scope.dragState.node.classList.contains('drag-handle')) {
 			scope.dragState = null;
-			ev.preventDefault();
 			return false;
 		}
 
@@ -2726,6 +2725,7 @@ var CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection.p
 
 	/** @private */
 	handleDragOver: function(ev) {
+		if (scope.dragState === null ) return;
 		var n = scope.dragState.targetNode,
 		    r = scope.dragState.rect,
 		    t = r.top + r.height / 2;
@@ -2746,6 +2746,7 @@ var CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection.p
 
 	/** @private */
 	handleDragEnter: function(ev) {
+		if (scope.dragState === null ) return;
 		scope.dragState.rect = ev.currentTarget.getBoundingClientRect();
 		scope.dragState.targetNode = ev.currentTarget;
 	},
@@ -2771,6 +2772,7 @@ var CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection.p
 
 	/** @private */
 	handleDrop: function(ev) {
+		if (scope.dragState === null ) return;
 		var s = scope.dragState;
 
 		if (s.node && s.targetNode) {
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -2265,7 +2265,96 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
 			this.addItem(dl, this.values[i], label);
 		}
 
+		this.initDragAndDrop(dl);
+
 		return this.bind(dl);
+	},
+
+	/** @private */
+	initDragAndDrop: function(dl) {
+		let draggedItem = null;
+		let placeholder = null;
+
+		dl.addEventListener('dragstart', (e) => {
+			if (e.target.classList.contains('item')) {
+				draggedItem = e.target;
+				e.target.classList.add('dragging');
+			}
+		});
+
+		dl.addEventListener('dragend', (e) => e.target.classList.remove('dragging'));
+
+		dl.addEventListener('dragover', (e) => e.preventDefault());
+
+		dl.addEventListener('dragenter', (e) => e.target.classList.add('drag-over'));
+
+		dl.addEventListener('dragleave', (e) => e.target.classList.remove('drag-over'));
+
+		dl.addEventListener('drop', (e) => {
+			e.preventDefault();
+			e.target.classList.remove('drag-over');
+			const target = e.target.classList.contains('item') ? e.target : dl.querySelector('.add-item');
+			dl.insertBefore(draggedItem, target);
+		});
+
+		dl.addEventListener('click', (e) => {
+			if (e.target.closest('.item')) {
+				const span = e.target.closest('.item').querySelector('SPAN');
+				if (span) {
+					const range = document.createRange();
+					range.selectNodeContents(span);
+					const selection = window.getSelection();
+					if (selection.rangeCount === 0 || selection.toString().length === 0) {
+						selection.removeAllRanges();
+						selection.addRange(range);
+					} else selection.removeAllRanges();
+				}
+			}
+		});
+
+		dl.addEventListener('touchstart', (e) => {
+			const touch = e.touches[0];
+			const target = e.target.closest('.item');
+			if (target) {
+				draggedItem = target;
+
+				placeholder = draggedItem.cloneNode(true);
+				placeholder.className = 'placeholder';
+				placeholder.style.height = `${draggedItem.offsetHeight}px`;
+				draggedItem.parentNode.insertBefore(placeholder, draggedItem.nextSibling);
+				draggedItem.classList.add('dragging')
+			}
+		});
+
+		dl.addEventListener('touchmove', (e) => {
+			if (draggedItem) {
+				const touch = e.touches[0];
+				const currentY = touch.clientY;
+
+				const items = Array.from(dl.querySelectorAll('.item'));
+				const target = items.find(item => {
+					const rect = item.getBoundingClientRect();
+					return currentY > rect.top && currentY < rect.bottom;
+				});
+
+				if (target && target !== draggedItem) {
+					const insertBefore = currentY < target.getBoundingClientRect().top + target.offsetHeight / 2;
+					dl.insertBefore(placeholder, insertBefore ? target : target.nextSibling);
+				}
+
+				e.preventDefault();
+			}
+		});
+
+		dl.addEventListener('touchend', (e) => {
+			if (draggedItem && placeholder) {
+				dl.insertBefore(draggedItem, placeholder);
+				draggedItem.classList.remove('dragging')
+				placeholder.parentNode.removeChild(placeholder);
+				placeholder = null;
+				draggedItem = null;
+			}
+		});
 	},
 
 	/** @private */
@@ -2287,7 +2376,7 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
 	/** @private */
 	addItem: function(dl, value, text, flash) {
 		var exists = false,
-		    new_item = E('div', { 'class': flash ? 'item flash' : 'item', 'tabindex': 0 }, [
+		    new_item = E('div', { 'class': flash ? 'item flash' : 'item', 'tabindex': 0, 'draggable': true }, [
 				E('span', {}, [ text || value ]),
 				E('input', {
 					'type': 'hidden',
@@ -2359,7 +2448,17 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
 			return;
 
 		if (item) {
-			this.removeItem(dl, item);
+			// Get bounding rectangle of the item
+			var rect = item.getBoundingClientRect();
+
+			// Get computed styles for the ::after pseudo-element
+			var afterStyles = window.getComputedStyle(item, '::after');
+			var afterWidth = parseFloat(afterStyles.width) || 0;
+
+			// Check if the click is within the ::after region
+			if (rect.right - ev.clientX <= afterWidth) {
+				this.removeItem(dl, item);
+			}
 		}
 		else if (matchesElem(ev.target, '.cbi-button-add')) {
 			var input = ev.target.previousElementSibling;
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -624,9 +624,11 @@ select,
 	border-radius: 3px;
 	color: var(--text-color-high);
 	position: relative;
-	pointer-events: none;
+	pointer-events: auto; /* needed for drag-and-drop in UIDynamicList */
 	overflow: hidden;
 	word-break: break-all;
+	cursor: move; /* drag-and-drop */
+	user-select: text; /* text selection in drag-and-drop */
 }
 
 .cbi-dynlist > .item::after {
@@ -645,8 +647,33 @@ select,
 	pointer-events: auto;
 }
 
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .item.drag-over { 
+	border-top: 1px solid var(--text-color-highest);
+}
+
+/* Make item being dragged in UIDynamicList partially transparent*/
+.cbi-dynlist > .item.dragging {
+	opacity: 0.5;
+}
+
+/* prevent pointer changing when over the span element in UIDynamicList */
+.cbi-dynlist > .item > span {
+	pointer-events: none;
+}
+
 .cbi-dynlist > .add-item {
 	display: flex;
+}
+
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .add-item > .cbi-input-text.drag-over {
+	border-top: 1px solid var(--text-color-highest);
+}
+
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .add-item > .cbi-button-add.drag-over {
+	border-top: 1px solid var(--text-color-highest);
 }
 
 .cbi-dynlist > .add-item > input,
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -1378,10 +1378,12 @@ body:not(.Interfaces) .cbi-rowstyle-2:first-child {
 	max-width: 25rem;
 	margin-right: 2em;
 	padding: .5em .25em .25em 0;
-	pointer-events: none;
+	pointer-events: auto; /* needed for drag-and-drop in UIDynamicList */
 	color: #666;
 	border-bottom: 2px solid rgba(0, 0, 0, .26);
 	outline: 0;
+	cursor: move; /* drag-and-drop */
+	user-select: text; /* text selection in drag-and-drop */
 }
 
 .cbi-dynlist[name="sshkeys"] > .item {
@@ -1403,9 +1405,21 @@ body:not(.Interfaces) .cbi-rowstyle-2:first-child {
 	background-color: var(--red-color-high);
 }
 
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .item.drag-over {
+	border-top: 1px solid black;
+}
+
+/* Make item being dragged in UIDynamicList partially transparent*/
+.cbi-dynlist > .item.dragging {
+	opacity: 0.5;
+}
+
+/* prevent pointer changing when over the span element in UIDynamicList */
 .cbi-dynlist > .item > span {
 	white-space: normal;
 	word-break: break-word;
+	pointer-events: none;
 }
 
 .cbi-dynlist > .add-item {
@@ -1413,6 +1427,16 @@ body:not(.Interfaces) .cbi-rowstyle-2:first-child {
 	align-items: center;
 	width: 100%;
 	min-width: 16rem;
+}
+
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .add-item > .cbi-input-text.drag-over {
+	border-top: 1px solid black;
+}
+
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .add-item > .cbi-button-add.drag-over {
+	border-top: 1px solid black;
 }
 
 .cbi-dynlist > .add-item:not([ondrop]) > input {
--- a/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css
+++ b/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css
@@ -1173,9 +1173,11 @@ textarea {
 	position: relative;
 	overflow: hidden;
 	transition: box-shadow .25s ease-in-out;
-	pointer-events: none;
+	pointer-events: auto; /* needed for drag-and-drop in UIDynamicList */
 	flex: 1 1 100%;
 	word-break: break-all;
+	cursor: move; /* drag-and-drop */
+	user-select: text; /* text selection in drag-and-drop */
 }
 
 .cbi-dynlist > .item::after {
@@ -1196,6 +1198,20 @@ textarea {
 	pointer-events: all;
 }
 
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .item.drag-over {
+	border-top: 1px solid black;
+}
+
+/* Make item being dragged in UIDynamicList partially transparent*/
+.cbi-dynlist > .item.dragging {
+	opacity: 0.5;
+}
+/* prevent pointer changing when over the span element in UIDynamicList */
+.cbi-dynlist > .item > span {
+	pointer-events: none;
+}
+
 .cbi-dynlist[disabled] > .item::after {
 	pointer-events: none;
 }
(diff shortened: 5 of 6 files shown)


systemcrash:
I like the idea of something like this, but perhaps inheriting the dynamiclist to make a new type with these properties? Unless it's sufficient to set e.g. 'draggable' to true on existing instantiations.

See how it's done in CBITableSection in form.js which is one of the few which implements dragging. That's probably where it should be done.

muink:
@Ramon00 It works!

But drag is not work in `TableSection` & `GridSection with editable = true`.

Ramon00:
@systemcrash We can make it into a new proto, but that then requires each instance that uses this to be updated.  Maybe its better to just add an option "draggable" default it to true and only if its explicitly set to false then disable it? The way I implemented it means the layout is 100% untouched, so there should not be compatibility issues.  Im not sure if there are situations where you dont want to have this?

@muink Let me have a look if it can be "enabled" there as well.

systemcrash:
draggability of elements should have some visual indicator. It's not obvious that those elements whose behavioural expectations have ossified, are movable. It would appear that elements in a table list vs in a dialogue window get confused regarding how drag should behave.

Ramon00:
@systemcrash there is a visual indication, the mouse pointer changes shape. Anyway there are two ways to do this 1) make a new function, add drag icon. Just to make sure it does not break anything else. Then for each location this is used, change to the other location and test 2) Use current function, dont add drag icon, just enable everywhere. 

@muink The form.js CBITableSection did not allow passing of the events. So i fixed that. Just 1 minor thing I need to fix is the opacity=0.5 of the CBITableSection.  But please have a test already.

Ramon00:
@muink  visual is fixed. Please have a test.

Ramon00:
@systemcrash  and other. Please have a look if there are any bugs and if you like this. 
if you do not like it, just close the commit. I am fine either way.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2265,8 +2266,106 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
...
+
+	dl.addEventListener('touchend', (e) => {
+		if (draggedItem && placeholder) {
+			dl.insertBefore(draggedItem, placeholder);
+			draggedItem.style.opacity = "";
+			placeholder.parentNode.removeChild(placeholder);
+			placeholder = null;
+			draggedItem = null;
+		}
+	});
```
systemcrash, line 2366: indent

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2287,7 +2386,7 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
 	/** @private */
 	addItem: function(dl, value, text, flash) {
 		var exists = false,
-		    new_item = E('div', { 'class': flash ? 'item flash' : 'item', 'tabindex': 0 }, [
+			new_item = E('div', { 'class': flash ? 'item flash' : 'item', 'tabindex': 0, 'draggable': true }, [
```
systemcrash, line 2389: leave as spaces for now

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -2759,7 +2760,6 @@ var CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection.p
 	/** @private */
 	handleDragEnd: function(ev) {
 		var n = ev.target;
-
```
systemcrash, line 2762: drop

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2359,7 +2458,8 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
 			return;
 
 		if (item) {
-			this.removeItem(dl, item);
+ 			if (item.getBoundingClientRect().right - ev.clientX<25)
+				this.removeItem(dl, item);
 		}
```
systemcrash, line 2463: This may need firming up. This is to ensure we don't start to drag but delete the item. Maybe we can derive the general dimensions of the X part, since it may vary through different themes. Try this:
```js

		if (item) {
			// Get bounding rectangle of the item
			var rect = item.getBoundingClientRect();

			// Get computed styles for the ::after pseudo-element
			var afterStyles = window.getComputedStyle(item, '::after');
			var afterWidth = parseFloat(afterStyles.width) || 0;

			// Check if the click is within the ::after region
			if (rect.right - ev.clientX <= afterWidth) {
				this.removeItem(dl, item);
			}
		}
```

I get the move cursor anyway juuust at the right edge in bootstrap, but that's a byproduct of how the design lays elements out. So that's OK.

systemcrash:
I gave it a quick test-drive. Works nicely. Great work @Ramon00. Address the comments and we can get this in.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2265,8 +2266,106 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
 			this.addItem(dl, this.values[i], label);
 		}
 
+		this.initDragAndDrop(dl);
+
 		return this.bind(dl);
 	},
+	initDragAndDrop: function(dl) {
+		let draggedItem = null;
+		let touchStartY = 0;
```
systemcrash, line 2275: drop this. it's unused.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2265,8 +2266,106 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
...
+				draggedItem = e.target;
+				e.target.style.opacity = 0.5;
+			}
+		});
+
+		dl.addEventListener('dragend', (e) => {
+			if (e.target.classList.contains('item')) {
+				e.target.style.opacity = "";
+			}
+		});
```
systemcrash, line 2289: maybe 
```js
	dl.addEventListener('dragend', (e) => resetStyles(e.target));
```

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2265,8 +2266,106 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
...
+			if (e.target.classList.contains('item')) {
+				e.target.style.borderTop = "2px solid #aaa";
+			}
+		});
+
+		dl.addEventListener('dragleave', (e) => {
+			if (e.target.classList.contains('item')) {
+				e.target.style.borderTop = "";
+			}
+		});
```
systemcrash, line 2305: maybe
```js
dl.addEventListener('dragleave', (e) => resetStyles(e.target));
```

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2265,8 +2266,106 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
...
+			e.preventDefault();
+			if (e.target.classList.contains('item')) {
+				e.target.style.borderTop = "";
+				dl.insertBefore(draggedItem, e.target);
+			} else {
+				var ai = dl.querySelector('.add-item');
+				ai.parentNode.insertBefore(draggedItem, ai);
+				e.target.style.borderTop = "";
+			}
+		});
```
systemcrash, line 2317: maybe
```js
		dl.addEventListener('drop', (e) => {
			e.preventDefault();
			if (draggedItem) {
				resetStyles(e.target);
				const target = e.target.classList.contains('item') ? e.target : dl.querySelector('.add-item');
				dl.insertBefore(draggedItem, target);
			}
		});
```

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2265,8 +2266,106 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
...
+				return currentY > rect.top && currentY < rect.bottom;
+			});
+
+			if (target && target !== draggedItem) {
+				if (currentY < target.getBoundingClientRect().top + target.offsetHeight / 2) {
+					dl.insertBefore(placeholder, target);
+				} else {
+					dl.insertBefore(placeholder, target.nextSibling);
+				}
+			}
```
systemcrash, line 2352: maybe
```js
				if (target && target !== draggedItem) {
					const insertBefore = currentY < target.getBoundingClientRect().top + target.offsetHeight / 2;
					dl.insertBefore(placeholder, insertBefore ? target : target.nextSibling);
				}
```

systemcrash:
One side effect I've just noticed is that it's nigh on difficult to double or triple click to select text contents which I often do to  duplicate similar entries. Any ideas?

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2265,8 +2266,106 @@ var UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype */
...
 
+		this.initDragAndDrop(dl);
+
 		return this.bind(dl);
 	},
+	initDragAndDrop: function(dl) {
+		let draggedItem = null;
+		let touchStartY = 0;
+		let placeholder = null;
+
```
systemcrash, line 2277: add this for later
```js

		const resetStyles = (target) => {
			if (target?.classList.contains('item')) {
				target.style.opacity = "";
				target.style.borderTop = "";
			}
		};
```

systemcrash:
> One side effect I've just noticed is that it's nigh on difficult to double or triple click to select text contents which I often do to duplicate similar entries. Any ideas?

If I do something like this, we have click to select working:
```js

	initDragAndDrop: function(dl) {
		let draggedItem = null;
		let placeholder = null;
		let selectedItem = null;
...
		const selectText = (element) => {
			const range = document.createRange();
			const textNode = element.querySelector('span');
			if (textNode) {
				range.selectNodeContents(textNode);
				const selection = window.getSelection();
				selection.removeAllRanges();
				selection.addRange(range);
			}
		};

		dl.addEventListener('click', (e) => {
			const item = e.target.closest('.item');
			if (!item) return;

			// Deselect previously selected item
			if (selectedItem && selectedItem !== item) {
				selectedItem.classList.remove('selected');
				window.getSelection().removeAllRanges();
			}

			// Toggle selection state
			if (item.classList.contains('selected')) {
				item.classList.remove('selected');
				selectedItem = null;
				window.getSelection().removeAllRanges();
			} else {
				item.classList.add('selected');
				selectedItem = item;
				selectText(item);
			}
		});

...
```

modules/luci-base/htdocs/luci-static/resources/ui.js
Ramon00: i guess you mean the touchStartY. yes that was a leftover from a test.

modules/luci-base/htdocs/luci-static/resources/ui.js
Ramon00: That may be better, but these events may get triggered by sub-items as well, so actually not sure.

modules/luci-base/htdocs/luci-static/resources/ui.js
Ramon00: See previous comment

modules/luci-base/htdocs/luci-static/resources/ui.js
systemcrash: I tested. Worked fine.

Ramon00: I know it works. Its just that maybe it get applied to something you dont want, e.g. in (future) extensions . Let me just remove it though.

modules/luci-base/htdocs/luci-static/resources/ui.js
Ramon00: I removed the "if", but the resetStyles does not (always) reset the opacity.

modules/luci-base/htdocs/luci-static/resources/ui.js
Ramon00: I did another solution, I added CSS styles, and dynamically added a "dragging" class and "drag-over" class.  Is even more elegant.

Ramon00:
That click code did not work for me. Not sure why.  We could also use the dblclick event, or right click as alternative? Maybe tooltip to clarify "right click to select text"?

One other issue ive noticed is that if you drag with your cursor over the text of the next item that then the "insert line" (border-top line) disappears. Somehow a "dragleave event" gets generated if you move onto the actual text. I tried to circumvent it, but no luck so far. It still works if you drop there though, so its just visual.

Ramon00:
@systemcrash found a bug in the already existing code of likely form.js (so not related to this PR), if you click here:
image
it changes into: 
image

The data filled I filled in is "saved" but not "applied". Not sure if that matters if its applied or not. (Running 23.05.4)

Ramon00:
Ah found the error: 
`				ui.Table.prototype.deriveSortKey((val != null) ? val.trim() : ''),`
if you fill in a number then trim is not available. Maybe something like this:
`				ui.Table.prototype.deriveSortKey((val != null) && ( typeof val.trim === 'function') ? val.trim() : ''),`

or force it to a string?

systemcrash:
> The data filled I filled in is "saved" but not "applied". Not sure if that matters if its applied or not. (Running 23.05.4)

Yeah, that click might be something to do with column sorting. I see it too on 23.05.5. Fun. 


> That click code did not work for me. Not sure why. We could also use the dblclick event, or right click as alternative? Maybe tooltip to clarify "right click to select text"?

Try harder :D

It worked for me on FF. What browser(s) did you try?

systemcrash:
> Ah found the error: ` ui.Table.prototype.deriveSortKey((val != null) ? val.trim() : ''),` if you fill in a number then trim is not available. Maybe something like this: ` ui.Table.prototype.deriveSortKey((val != null) && ( typeof val.trim === 'function') ? val.trim() : ''),`
> 
> or force it to a string?

OK. Good catch with this one. Although your fix works for the sort part, it seems to reveal another problem: deleting an entry from that table then results in a traceback. Do you see the same?

It looks like the delete propagates, if you reload the page, but... not great.

Ramon00:
> It worked for me on FF. What browser(s) did you try?

im using chromium (Edge) at the moment.

> OK. Good catch with this one. Although your fix works for the sort part, it seems to reveal another problem: deleting an entry from that table then results in a traceback, but only after sorting the table on any column. Do you see the same?
> 
> It looks like the delete propagates, if you reload the page, but... not great.
I am not getting the tracback on delete. Maybe thats FF vs chromioum?

systemcrash:
```
uci add dhcp ipset # =cfg096e2a
uci add_list dhcp.@ipset[-1].name='1'
uci add_list dhcp.@ipset[-1].domain='foo.example.com'
uci set dhcp.@ipset[-1].table_family='inet'
uci add dhcp ipset # =cfg0a6e2a
uci add_list dhcp.@ipset[-1].name='2'
uci add_list dhcp.@ipset[-1].domain='foo2.example.com'
uci set dhcp.@ipset[-1].table_family='inet'
```
1 apply your fix exactly as proposed
2 load a page of ipsets where the first entry is called ipset 1 and the second ipset is called ipset 2
3 click on the sort header
4 delete an entry

Traceback:
```
raise http://192.168.1.130/luci-static/resources/luci.js?v=24.311.67894~9ddbc11:2301
    handleCallReply http://192.168.1.130/luci-static/resources/rpc.js?v=24.311.67894~9ddbc11:94
    promise callback*parseCallReply http://192.168.1.130/luci-static/resources/rpc.js?v=24.311.67894~9ddbc11:66
    promise callback*call http://192.168.1.130/luci-static/resources/rpc.js?v=24.311.67894~9ddbc11:41
    declare http://192.168.1.130/luci-static/resources/rpc.js?v=24.311.67894~9ddbc11:342
    declare http://192.168.1.130/luci-static/resources/rpc.js?v=24.311.67894~9ddbc11:302
    reorderSections http://192.168.1.130/luci-static/resources/uci.js?v=24.311.67894~9ddbc11:194
    save http://192.168.1.130/luci-static/resources/uci.js?v=24.311.67894~9ddbc11:945
    promise callback*save http://192.168.1.130/luci-static/resources/uci.js?v=24.311.67894~9ddbc11:937

```

Ramon00:
> 4 delete an entry

image

systemcrash:
Yeah, same crash. It's the rpc uci order call. One too many entries gets passed right after delete.

Ramon00:
ah found the fix for text selection `	user-select: text;` in the css code. Will push an update shortly, that leaves the visual "issue" with the insert line. Could just remove the line to make it simpler?

systemcrash:
I might have a fix for that. Just testing. Arrays and off-by-ones can be weird.

systemcrash:
please re-base and re-test - that crash bug should be gone now. 🤞

Ramon00:
ok will rebase, and one more thing to change. That `const span = e.target.closest('.item').querySelector('SPAN');` can generate an exception if .item is not present. Let me add a check on that.

Ramon00:
ok seems to work. Just that line still disappears, if I move over the text part.

systemcrash:
did you `git pull` the latest changes? Your commit deletes some recently added stuff.

Ramon00:
{67A6BBD9-EDA3-491E-A089-DFF9CA900F7D}
{2A4AD10C-ECA4-4225-9E8B-DF82B0FF8AFD}
This i mean

Ramon00:
> did you `git pull` the latest changes? Your commit deletes some recently added stuff.

git pull --rebase origin master

systemcrash:
> {67A6BBD9-EDA3-491E-A089-DFF9CA900F7D} {2A4AD10C-ECA4-4225-9E8B-DF82B0FF8AFD} This i mean

I'm not sure it's important we fix that. Drag and drop otherwise works. The drag seems to be targetting the 'target' cell, and not the parent div that contains it.

systemcrash:
> > did you `git pull` the latest changes? Your commit deletes some recently added stuff.
> 
> git pull --rebase origin master

Your commit is still based on 53e36e3293bee2ea2ce0fbc1250074c44cfe8335. And not fefb9acf57ad11012c336e4d056b074620b87d2f

Ramon00:
> > > did you `git pull` the latest changes? Your commit deletes some recently added stuff.
> > 
> > 
> > git pull --rebase origin master
> 
> Your commit is still based on [53e36e3](https://github.com/openwrt/luci/commit/53e36e3293bee2ea2ce0fbc1250074c44cfe8335). And not [fefb9ac](https://github.com/openwrt/luci/commit/fefb9acf57ad11012c336e4d056b074620b87d2f)

ok let me fix... not sure what happened there

Ramon00:
ok i think its fixed

systemcrash:
Just clean that double subject line from the commit, and it's done I reckon.

Ramon00:
Alright. Thx for the help!

Ramon00:
> add

Ah nice. CSS to the rescue!

muink:
It's normal now. @systemcrash

systemcrash:
Hi @Ramon00 I want to get this merged. I was wondering whether you can add a peppering of code comments in the CSS additions? I want the style entries to leave no doubt as to why they're there. CSS seems to be a mysterious dumping ground, and future editors unwilling to change anything there because nobody knows why something went in. I'll make some proposals for at least the first file.

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -624,9 +624,11 @@ select,
 	border-radius: 3px;
 	color: var(--text-color-high);
 	position: relative;
-	pointer-events: none;
+	pointer-events: auto;
```
systemcrash, line 627: `	pointer-events: auto; // needed for drag-and-drop in at least UIDynamicList`

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -624,9 +624,11 @@ select,
 	border-radius: 3px;
 	color: var(--text-color-high);
 	position: relative;
-	pointer-events: none;
+	pointer-events: auto; 
 	overflow: hidden;
 	word-break: break-all;
+	cursor: move;
+	user-select: text;
```
systemcrash, line 631: Perhaps group these with the pointer-events: change above.

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -645,10 +647,30 @@ select,
...
 }
 
+.cbi-dynlist > .add-item > .cbi-input-text.drag-over {
+	border-top: 1px solid var(--text-color-highest);
+}
+
+.cbi-dynlist > .add-item > .cbi-button-add.drag-over {
+	border-top: 1px solid var(--text-color-highest);
+}
+
```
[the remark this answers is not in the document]
Ramon00: no,  there is a div with class "add-item". That contains two elements an input element with "cbi-input-text" class and a div item  (button) with cbi-button-add class.  So this triggers when you have that cascade of elements with those classes, and in addition the "drag-over" is present (applied in js code) when you hover over it. So this adds a line on either the button or the input text.

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
Ramon00: This only triggers when the "cbi-dynlist" class is applied. That should only be applied in a UIdynamicList function. So let not add that "at least" part

Ramon00: Also I dont think // comments are allowed in CSS. /* */ only as far as I know

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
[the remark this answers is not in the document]
systemcrash: Sorry, missed the button/input bit 🤦 haha

You're right on the comments, /* */ are acceptable.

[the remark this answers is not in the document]
Ramon00: ah no worries. And changes are almost done. Just copying them to the other css files.

themes/luci-theme-material/htdocs/luci-static/material/cascade.css
```diff
@@ -1403,9 +1405,21 @@ body:not(.Interfaces) .cbi-rowstyle-2:first-child {
 	background-color: var(--red-color-high);
 }
 
+/* indication line for drag-and-drop in UIDynamicList*/
+.cbi-dynlist > .item.drag-over {
+	border-top: 1px solid black;
+}
+
+/* transluce item being dragged in UIDynamicList*/
```
systemcrash, line 1413: translucent?

Ramon00: Category any noun can be verbed :) 
I was trying to save some bytes, ill make it "make item partially transparent while being dragged in UIDynamicList"

systemcrash:
Excellent. I think our work here is done :)

Ramon00:
ok go for it! Thx for the help

systemcrash:
merged. Thanks @Ramon00 !

muink:
Recently I noticed that drag sorting does not trigger the `onchange` and `validate` functions to run.
We need to add event listeners to execute the [onchange](https://github.com/openwrt/luci/blob/f001a35ad14e69d26e2e8fdfb6f0c0f6fb00b17a/modules/luci-base/htdocs/luci-static/resources/form.js#L3787) and [validate](https://github.com/openwrt/luci/blob/f001a35ad14e69d26e2e8fdfb6f0c0f6fb00b17a/modules/luci-base/htdocs/luci-static/resources/ui.js#L2279) functions when the drag is completed.
@Ramon00

systemcrash:
I don't think they should. The fields are read only after entering the data. Can you show us a case when it's necessary?

Ramon00:
All values should be validated already, so triggering a validate would make it automatically pass? I could imagine a change in order needing to trigger something, but i cannot think of a good example.

muink:
This is an inconvenient place I encountered in [my project](https://github.com/muink/openwrt-fchomo).
sort
`Type/Factor ++` will submit changes to `Entry` through the [onchange()](https://github.com/muink/openwrt-fchomo/commit/436971038ce518893ac4b9ef408dd64fea6793de) or [validate()](https://github.com/muink/openwrt-fchomo/commit/ff9159514a3c314b2756cc6220c88ab28843bbc1) function.  
But `Type/Factor ++` itself `o.write = function() {};`, so any changes are not submitted to `Entry` will be discarded when the `Save` button is pressed.
I think it is necessary to at least trigger onchange. The value it receives is the entire array, which should not cause problems.

Ramon00:
I guess if we add 
`		dl.dispatchEvent(new CustomEvent('cbi-dynlist-change', {
			bubbles: true,
			detail: {
				instance: this,
				element: dl,
				value: value,
				add: true
			}
		}));`

in a few locations in the code then this should generate both validate and onchange. Let me make a PR for this.

Ramon00:
https://github.com/openwrt/luci/pull/7533

PR #7416, opened 2024-11-21
collections: select package manager for SSL collections

robimarko:
- [x] Tested on: (qualcommax/ipq807x, main) :white_check_mark:

Currently, the regular and nginx collections include package manager by default but SSL collections dont, so lets include package manager on SSL collections as well.

### DIFF
--- a/collections/luci-ssl-openssl/Makefile
+++ b/collections/luci-ssl-openssl/Makefile
@@ -14,7 +14,10 @@ LUCI_DESCRIPTION:=LuCI with OpenSSL as the SSL backend (libustream-openssl). \
  OpenSSL cmd tools (openssl-util) are used by uhttpd for SSL key generation \
  instead of the default px5g. (If px5g is installed, uhttpd will prefer that.)
 
-LUCI_DEPENDS:=+luci-light +libustream-openssl +openssl-util
+LUCI_DEPENDS:=+luci-light \
+	+libustream-openssl \
+	+openssl-util \
+	+luci-app-package-manager
 
 PKG_LICENSE:=Apache-2.0
 
--- a/collections/luci-ssl/Makefile
+++ b/collections/luci-ssl/Makefile
@@ -10,7 +10,10 @@ LUCI_TYPE:=col
 LUCI_BASENAME:=ssl
 
 LUCI_TITLE:=LuCI with HTTPS support (mbedtls as SSL backend)
-LUCI_DEPENDS:=+luci-light +libustream-mbedtls +px5g-mbedtls
+LUCI_DEPENDS:=+luci-light \
+	+libustream-mbedtls \
+	+px5g-mbedtls \
+	+luci-app-package-manager
 
 PKG_LICENSE:=Apache-2.0
 


systemcrash (approved):

robimarko:
That was a weird reasoning, I dont see any reason why these collections should differ in the basic feature set.

sanderpqr:
When I use the imagebuilder to create a 24.10-SNAPSHOT image, adding `luci-ssl` automatically adds `luci-app-package-manager`. This makes sense in MASTER, but not in 24.10 where APK hasn't replaced OPKG. I'm guessing this is caused by this commit and [Merge branch 'master' into openwrt-24.10](https://git.openwrt.org/?p=project/luci.git;a=commit;h=2ce85291bdd70accbb1863bfb32ddf18bcb7b3fe)?

sanderpqr:
Ah, that clarifies it. Thanks.

dhewg:
On 22/11/2024 10:30 am, Robert Marko wrote:
> That was a weird reasoning, I dont see any reason why these collections 
> should differ in the basic feature set.

Weird how? What I find weird is why a ssl provider should depend on a 
package frontend? :P

The idea was that the top level luci collection depends on the package 
manager, but luci-light doesn't. With that you had the option to combine 
one of those with a ssl provider of your choice (so the ssl collections 
are second tier, an addition to the top level ones).

Due to this luci-app-package-manager was dragged in my config 
(luci-light + luci-ssl-openssl), which I don't need or want, no matter 
if opkg or apk. Now that's gone, and we may as well remove luci-light then?

But I'd prefer a revert, pretty please?

hnyman:
@dhewg 
It is not just a "SSL provider", it is the default basic LuCI service set with SSL.
I think that the majority of the users do want the package manager at least ocacsionally, so having it in the default LuCI SSL collection looks normal to me. And dropping it was strange.

All LuCI collection are virtual, so it is easy for you to drop something in your own build, if you want a minimal set.

Or we might add a luci-mod-light-ssl or luci-mod-mini-ssl collection? or something like that.
(as they are just virtual dependency groups without real content, that would not increase maintenance.)

dhewg:
On 27/11/2024 8:25 am, Hannu Nyman wrote:
> It is not just a "SSL provider", it is the default basic LuCI service 
> set with SSL.

Well, my above linked PR changed things a bit in that regard, so it 
basically was just a ssl provider for almost 2 years now ;)

> I think that the majority of the users do want the package manager at 
> least ocacsionally, so having it in the default LuCI SSL collection 
> looks normal to me. And dropping it was strange.

Right, but I had the impression that quite some users were happy about 
having the option to not drag in the package manager.

> All LuCI collection are virtual, so it is easy for you to drop something 
> in your own build, if you want a minimal set.

I can just revert the patch locally, no problem, but imagebuilder users 
won't have that choice now.
> Or we might add a luci-mod-light-ssl or luci-mod-mini-ssl collection? or 
> something like that.
> (as they are just virtual dependency groups without real content, that 
> would not increase maintenance.)

That was discussed back then, and discarded because of many combinations 
and the extra burden.

PR #7426, opened 2024-11-26
luci-app-ddns: fix toggle state button

Chill-Try:
- [x] Tested on: (x86_64, v23.05.5, chrome) :white_check_mark:
- [x] \( Preferred ) Mention: @systemcrash 
- [x] \( Optional ) Closes: openwrt/luci/issues/7423#issuecomment-2498227219
- [x] Description: Fixed toggle state button. This is my first time using pull request on github. Please forgive me if there is anything wrong.

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -167,9 +167,12 @@ return view.extend({
 	},
 
 	handleToggleDDns: function(m, ev) {
-		return this.callInitAction('ddns', 'enabled')
-			.then(L.bind(function (action) { return this.callInitAction('ddns', action ? 'disable' : 'enable')}, this))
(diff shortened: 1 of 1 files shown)


systemcrash:
See 8d26279bab90ee4cee9eda01c9054e8a6c7b890c. This might close this issue.

Try2079:
I think it won't work. Because `this.callInitAction()` only return `true`, it never return `false`.

```lua
# sys.lua
local function init_action(action, name)
	name = fs.basename(name)
	if fs.access(init.dir..name) then
		return call("env -i %s%s %s >/dev/null" %{ init.dir, name, action })
	end
end
```
```javascript
// luci
setInitAction: {
		args: { name: 'name', action: 'action' },
		call: function(request) {
			switch (request.args.action) {
			case 'enable':
			case 'disable':
			case 'start':
			case 'stop':
			case 'restart':
			case 'reload':
				const rc = init_action(request.args.name, request.args.action);

				if (rc === false)
					return { error: 'No such init script' };

				return { result: rc == 0 };

			default:
				return { error: 'Invalid action' };
			}
		}
	},
```

PR #7407, opened 2024-11-18
luci-app-upnp: Finalise changes for OpenWrt 24.10 release

Self-Hosting-Group:
- Complete plugin wording/UI revision
  Add missing protocol term in status/overview UI and shorten menu option to `UPnP IGD & PCP`, otherwise two lines, rearrange table columns, improve wording and add help texts and note upstream change miniupnp/miniupnp@9339f0e
- Simplify UI by moving rarely used options to advanced settings, rest to service setup and rearrangement
- Hide dependent UI fields if UPnP IGD protocol is disabled
- Remove `clean_ruleset_interval/threshold` UI options as not working
- Require client address and improve defaults in ACL entry UI
- Manual adaption of changed non-translatable protocol acronyms in translations, see #7175

This PR completes committed changes #6863 (merged in #6975) and continues #7217 for the UPnP IGD & PCP/NAT-PMP autonomous port mapping service.

### Service setup UI:

openwrt-settings
_[Screenshot 23.05 release](https://github.com/user-attachments/assets/38cbb287-e85d-4fd8-bed6-47a3a4a2cfbc)_

### Advanced settings UI:

openwrt-advanced-settings
_[Screenshot 23.05 release](https://github.com/user-attachments/assets/d0b9bcfc-d3af-4941-af6a-19e5bfcd3f81)_

### ACL entry UI:

openwrt-acl-entry

### Status/overview UI:

openwrt-status-overview

The **Port Control Protocol (PCP)** is the successor to NAT-PMP, has similar protocol concepts and packet formats, but adds support for **IPv6 port maps** and options/extensions. For more information, see:
**Port Mapping Protocols Overview and Comparison 2024 - About UPnP IGD & PCP/NAT-PMP**
https://github.com/Self-Hosting-Group/wiki/wiki/Port-Mapping-Protocols-Overview

Maintainer: @jow-

PR #7433, opened 2024-11-30
luci-app-upnp: restore some i18n keys

stokito:
Leftovers from #7407
CC @Self-Hosting-Group


systemcrash:
These changes look fine

systemcrash:
There were previous attempts where there was... a lot of discussion. I'm fairly sure we went through those options that were removed.

systemcrash:
Merged, thanks @stokito !

PR #7438, opened 2024-12-01
declare rpc calls with const

stokito:
This is a code style change in order to migrate to the ES6 syntax. 
I'll prose to gradually convert without no rush. Tools like the IntelliJ can reformat almost everything but results needs to be checked. For example we can replace `for (var i = 0;;)` with the `for (let i = 0;;)` but actually we may have a code after the `for` that checks for the `i` var value. On the other side almost all such indexed `for` loops can be replaced with `for of` (foreach).

But here I wish to start from the quick win and replace the most common problem of declaring with `var`. 

The `var rpcCall = rpc.declare({})` can be always unambiguously converted to `const` since they are never changed.
In the same time the `var m, s, o` can be always converted to the `let m, s, o`.

When new devs will start to copy-paste code they'll use the new version.

I made a command to see most popular usages of the `var`:
```sh
grep -h --include=*.js -r  "var " applications/ | tr -d ' \t' | sort | uniq -c | sort
```

### DIFF
--- a/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
+++ b/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
@@ -200,7 +200,7 @@ return view.extend({
 			}
 		}
 
-		var m, s, o;
+		let m, s, o;
 
 		m = new form.Map('rpcd', _('LuCI Logins'));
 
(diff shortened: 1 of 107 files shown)


systemcrash:
Good idea. Can we start with just const for the rpcd calls? That's easier to review. Also check SOB...

systemcrash:
Closed by 22d48305afcb3095468756f087610a091d69a116.

Thanks @stokito

jow-:
Maybe you can also pick up https://github.com/openwrt/luci/pull/6597 and continue on top of it

PR #6830, opened 2024-01-13
Add CODEOWNERS and fix missing PKG_MAINTAINER

stokito:
The CODEOWNERS will request for a reviewer automatically by a path.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

The luci-app-bcp38 had a maintainer from luci-app-acme and that looks like a copy-paste error. So it's maintainer was changed to an author @danrl

Daniel F. Dickinson changed email address to <dfdpublic@wildtechgarden.ca>

luci-all-lxl has a maintainer Petar Koretic <petar.koretic@sartura.hr> but there is no corresponding GitHub account. So Dirk Brenken was added as a second maintainer: he answered on an issue of the app.

When maintainer wasn't set the initial author was used, or most contributor or Jo-Philipp Wich as a default.

If anyone can't be a maintainer he or she should create an issue and the CODEOWNERS file will be updated with a new one.

PR #7439, opened 2024-12-01
CODEOWNERS with list of maintainers

stokito:
The second attempt of the #6830 that was closed after a discussion.
My proposition is to add the CODEOWNERS file to help for contributors to quickly find a maintainer (as I did myself today). Even better, for those devs that already have a write access they'll automatically be assigned as reviewers for a PR for the particular app.

The other steps like creating `luci-dev` team as I described in my comment can be done later. But the CODEOWNERS can start helping us today and save time for contributors. We may add a link to it into the PR template.

It won't change anything in terms of security so the PR is safe to merge.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,131 @@
+applications/luci-app-acl @jow-
+applications/luci-app-acme @tohojo
+applications/luci-app-adblock @hnyman @dibdot
+applications/luci-app-adblock-fast @stangri
+applications/luci-app-advanced-reboot @stangri
+applications/luci-app-alist @1715173329
+applications/luci-app-apinger @jempatel
+applications/luci-app-aria2 @kuoruan
+applications/luci-app-attendedsysupgrade @aparcar
+applications/luci-app-babeld @PolynomialDivision
+applications/luci-app-banip @dibdot
+applications/luci-app-bcp38 @tohojo
+applications/luci-app-bmx7 @rogerpueyo @p4u
+applications/luci-app-clamav @ratkaj @lperkov
+applications/luci-app-cloudflared @animegasan @stokito
+applications/luci-app-commands @jow-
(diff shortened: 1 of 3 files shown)


Ramon00:
just merge this?

hnyman:
Not much has changed since #6830 which did not receive approval. 
(and like jow noted in that PR, the proposed list already contains quite many ancient users who haven't been active for a while)

Ramon00:
I guess this is in the catagory "better than nothing". But anyway, somebody needs to make decision, either merge or not. Waiting for the all singing and dancing solution will likely mean nothing will happen.

systemcrash:
I've read earlier revisions of info on this document and found some things left unexplained. More recent revisions seem much more clear. 

https://github.com/github/docs/blob/main/content/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners.md

But... unless a user mentioned in the CODEOWNERS file has write permissions to this repo, it's nothing more than a glorified reference document (so no behaviour changes will result, at present standards; a good thing). The document needs to be maintained, however. Do you commit to maintaining it?

hnyman:
@stokito 
You keep pushing for this, but so far I have seen no green light from any of the core OpenWrt developers who should have a say, neither in this PR or in the other (where there was hesitancy and opposition).

Well, looking for more closely:
Actually Github makes this easy for us: Apparently currently you can't be made as CODEOWNERS list member without write access. This PR with visible error messages shows it quite clearly:

image

image

And as there should be no commit access given to that amount of people, this is an automatic no-go.

systemcrash:
So... I don't think having this is bad or causes any undesirable behaviour changes. Githubs documentation on the feature is now much more clear on it. It just requires being kept up to date, something, which I think @stokito seems OK with maintaining?

stokito:
FYI: I updated the CODEOWNERS and removed the recently removed apps.

PR #7443, opened 2024-12-02
luci-app-acme: `cfgvalue` does not receive the `set_value` parameter when it is called in `CBIValue.render`.

tsanie:
- [X] Tested on: (x86_64, v24.10-SNAPSHOT, Edge 132.0.2957.11) :white_check_mark:
- [X] Description: The value of "Key type" will be overwritten to `undefined` when the drop-box after rendering. (except the first migration from the old `keylength` config parameter)

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
@@ -484,12 +484,7 @@ return view.extend({
 		o.rmempty = false;
 		o.optional = true;
 		o.modalonly = true;
-		o.cfgvalue = function(section_id, set_value) {
-			if (set_value != null) {
-				// no migration required, need to store data
-				this.data = this.data || {};
-				this.data[section_id] = set_value;
-			}
+		o.cfgvalue = function(section_id) {
 			var keylength = uci.get('acme', section_id, 'keylength');
 			if (keylength) {
 				// migrate the old keylength to a new keytype
@@ -502,7 +497,7 @@ return view.extend({
 					default: return ''; // bad value
 				}
 			}
-			return this.data ? this.data[section_id] : null;
+			return this.super.apply(this, ['cfgvalue'].concat(arguments));
 		};
 		o.write = function(section_id, value) {
 			// remove old keylength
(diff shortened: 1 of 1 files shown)


applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -497,7 +502,7 @@ return view.extend({
 					default: return ''; // bad value
 				}
 			}
-			return set_value;
```
tsanie, line 500: As code https://github.com/openwrt/luci/blob/b1968d9cdd6425821553138e0487df8840785afa/modules/luci-base/htdocs/luci-static/resources/form.js#L3681
the `set_value` will be `undefined`

tohojo:
I'm sorry, but I only have a very limited grasp of how all the callback stuff in luci works, so I don't feel qualified to review this. Please find a luci expert to take a look...

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
systemcrash: your code branch simplifies to:
```
if(set_value){
 something = set_value;
}
...
return something;
```

which is the same as:
```
if(set_value) return set_value;
return null;
```

How does this differ? What is the error you receive?

tsanie: Yes, that’s the problem.

The `cfgvalue` function should always return the selected value which isn’t `undefined`, so we need to cache it and return it later. (like the super method does) https://github.com/openwrt/luci/blob/b1968d9cdd6425821553138e0487df8840785afa/modules/luci-base/htdocs/luci-static/resources/form.js#L1826

The `set_value` will be passed in the `load` function.https://github.com/openwrt/luci/blob/b1968d9cdd6425821553138e0487df8840785afa/modules/luci-base/htdocs/luci-static/resources/form.js#L913
But in the later `render` function, the second parameter is undefined.https://github.com/openwrt/luci/blob/b1968d9cdd6425821553138e0487df8840785afa/modules/luci-base/htdocs/luci-static/resources/form.js#L3681

Issue:
Now the dropdown box always shows the undefined value even I have the right config `ec256`.
image

tsanie: Got it, the `super` method is a good way to resolve it.

PR #7444, opened 2024-12-02
luci-app-upnp: Critical hotfix revert 7a7f9ec and fix old translations

Self-Hosting-Group:
- Hotfix that reverts commit 7a7f9ec, which seemed cosmetic and unimportant, but changes defaults and disables both protocols/(service) on next save/apply
- Manually add the `IPv6 always allowed` note (in english) to the translations, as this is important, and fix the old texts that were not intended to be carried over but were committed in 9c403f5 and 2e46cfc, and re-remove the obsolete translations restored by the additional commits

For the intro/ACL texts (and some others), the consideration was that the original text was already old, had been changed several times by developers rather than translators, and had undergone four changes since the 23.05 release (missing IPv6 hint, ACLs->ACL, redirected->mapped, shortened sentence), so I decided to have them retranslated to maintain the meaning and integrity of the translations. I think there should be a policy that if e.g. security relevant or the meaning or multiple parts of a sentence have changed, a new translation should be enforced.

IMHO, it was not worth the effort/trouble (the 2nd time) of the additional hasty untested code/commits shortly after merging PR #7407, also because it introduced buggy code that caused a regression and added old text that should have been retranslated, also when I see how fast the translation works. The newly added strings were translated into chinese in less than 6 hours after the PR was merged, now 100% in four languages.

I think we can say that the LuCi plugin is now in good shape after this latest hotfix, except for the 24.10 daemon package which has the issue that the nftable rules are not deleted correctly and the port map listing functions always return a port of `0`, see miniupnp/miniupnp#768, miniupnp/miniupnp@d07b0a1 and point 4 in https://github.com/openwrt/luci/pull/7407#issuecomment-2493345487. A fix can be added as a separate commit to openwrt/packages#24988 if desired.

### DIFF
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
@@ -37,6 +37,7 @@ return baseclass.extend({
 	},
 
 	render: function(data) {
+
 		var table = E('table', { 'class': 'table', 'id': 'upnp_status_table' }, [
 			E('tr', { 'class': 'tr table-titles' }, [
 				E('th', { 'class': 'th' }, _('Client Name')),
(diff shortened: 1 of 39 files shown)


systemcrash:
Please drop all changes except for Host, Port, and .default inclusion.

systemcrash:
And that `return` statement

Issue #7436, opened 2024-12-01
<luci-proto-wireguard>: The peer name detection fails when the endpoint_host is an IPv6 address.


Stream-z:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### screenshots or captures

![capture](https://github.com/user-attachments/assets/1b032dfe-ca2e-458d-9def-e22069e50dfb)

### Actual behaviour

When the endpoint_host is set to an IPv6 address, the peer name appears as "Untitled peer" on the ``Status → WireGuard`` page. When set to an IPv4 address, the peer name is displayed correctly.

### Expected behaviour

1. When the endpoint_host is set to an IPv6 address,the peer name should be displayed correctly on the ``Status → WireGuard`` page.

### Steps to reproduce

1.go to:Network → Interfaces → wireguard → peer tab
2.modify the peer's endpoint_host to an IPv6 address or a domain name that resolves to an IPv6 address.

### Additional Information

```shell
NAME="OpenWrt"
VERSION="23.05.5"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt 23.05.5"
VERSION_ID="23.05.5"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r24106-10cc5fcd00"
OPENWRT_BOARD="x86/64"
OPENWRT_ARCH="x86_64"
OPENWRT_TAINTS=""
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt 23.05.5 r24106-10cc5fcd00"
```

### What browsers do you see the problem on?

Chrome

Stream-z:
I am not very familiar with this area of knowledge, so I tried using AI tools to locate the issue. I found that the problem might be in the ``checkPeerHost`` function in the ``luci.wireguard`` file:
```javascript
function checkPeerHost(configHost, configPort, wgHost) { 
    const ips = popen(`resolveip ${configHost} 2>/dev/null`);
    if (ips) {
        for (let line = ips.read('line'); length(line); line = ips.read('line')) {
            const ip = rtrim(line, '\n');
[...]

Issue #7347, opened 2024-10-24
luci-proto-wireguard: incorrect peer name detection when devise acts as a server


this-username-has-been-taken:
Unfortunately a previously fixed bug (see https://github.com/openwrt/luci/issues/7339) caused another issue: if an OpenWrt device is used as a Wireguard server (i.e. "listen port" is defined for the interface) then peer names on the `Status → WireGuard` page are not detected at all: `Untitled peer` is shown instead of the real names.

***
## Suggested solution:
Technically it is possible to save more than one peer with the same public/private key pair on the `Peers` tab, but after the interface is restarted only one peer will be active (the latter one). All other peers with the same public/private key pairs will be ignored and not shown in the `wg show all dump` command output. Thus we do not need to check peer addresses in this case - it is enough to only check if a public key matches.
In the view of the above to fix the issue I suggest to modify line 92 in `luci/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard` file from:
```
if (!s.disabled && s.public_key == record[1] && checkPeerHost(s.endpoint_host, s.endpoint_port, record[3]))
```
to:
```
if (!s.disabled && s.public_key == record[1] && (!s.endpoint_host || checkPeerHost(s.endpoint_host, s.endpoint_port, record[3])))
```
I.e. do not check a peer's host if no endpoint is defined (interface acts as a server).

## Steps to reproduce:

1. go to: Network → Interfaces → create a Wireguard interface. The interface should contain `listen port`, i.e. act as a server. Fill all other parameters (private key, ip address, etc) with correct values.
2. Add one or more peers on the `Peers` tab. None of the peers should have `endpoint host` defined.
3. Save & Apply.
4. Restart the interface.
5. Navigate to `Status → WireGuard` page.
6. Observe the issue:
![Peer_names_issue](https://github.com/user-attachments/assets/e8c645ef-eceb-4d32-a782-4434265afc43)

## Actual behavior:
All the peers on the `Status → WireGuard` page have `Untitled peer` name.

## Expected behavior:
All the peers on the `Status → WireGuard` page have correct names.

## Additional Information:
OpenWrt version information from system `/etc/openwrt_release`

Doesn't matter, but here it is:
```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='SNAPSHOT'
DISTRIB_REVISION='r27902-6835ff8cbc'
DISTRIB_TARGET='mediatek/filogic'
[...]

PR #7445, opened 2024-12-02
luci-proto-wireguard: Fix incorrect peer detection for IP v6

this-username-has-been-taken:
This commit fixes incorrect peer detection when using IP v6 by deleting all the square the square brackets from the wgHost variable.

Signed-off-by: Tom Haley <this_username_has_been_taken2@proton.me>

- [ ] Incremented :up: any `PKG_VERSION` in the Makefile **Not applicable: luci-proto-wireguard does not have `PKG_VERSION` in the Makefile**
- [X] Tested on: x86/64, OpenWrt 23.05.5, Chrome :white_check_mark:
- [X] \( Preferred ) Mention: @systemcrash
- [X] \( Optional ) Closes: openwrt/luci#7436

### DIFF
--- a/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
+++ b/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
@@ -17,10 +17,11 @@ function command(cmd) {
 
 function checkPeerHost(configHost, configPort, wgHost) {
 	const ips = popen(`resolveip ${configHost} 2>/dev/null`);
+	const hostIp = replace(wgHost, /\[|\]/g, "");
 	if (ips) {
 		for (let line = ips.read('line'); length(line); line = ips.read('line')) {
 			const ip =  rtrim(line, '\n');
-			if (ip + ":" + configPort == wgHost) {
+			if (ip + ":" + configPort == hostIp) {
 				return true;
 			}
 		}


systemcrash:
One question @this-username-has-been-taken - this change only affects what is used in the luci layer, and not what is passed to wg? I know wg needs the [] for IPv6.

systemcrash:
Also fix commit formalities (commit SOB?)

this-username-has-been-taken:
@systemcrash , yes - this fix only affects Luci and has nothing to do with the Wireguard connection itself.
The idea of the algorithm is:
1. Luci runs `wg show all dump` command and gets information about connected Peer.
2. Luci iterates through defined Wireguard Peers in the config to determine which one is used (up and running). Here lays the problem. If a Peer is an IP v4 Peer - everything is OK, but if a Peer is an IP v6 then the `wg show all dump` returns its address with square brackets, but config file stores its address without them hence the comparison fails and we get `Undefined peer` in the GUI.

I have also fixed the Sign-off message.

systemcrash:
OK - I just hope this doesn't conflict with changes introduced by e.g. #7347 and #7339. But since it's you who's fixing it, you've likely considered that :)

systemcrash:
Don't worry about it. We all make mistakes. The more you know...

xyyx:
Hi. I have luci-proto-wireguard  git-24.333.73023-d885ca5 version. Compared luci.wireguard and added last change. But all connected peers changes name to "Untitled peer". How to debug? wg show all dump gave correct IPv4 addresses

this-username-has-been-taken:
@xyyx 
Hi!
Let's investigate a bit further. Can you please share:
1. If you are using Wireguard as a server or client?
2. Your current luci.wireguard file.
3. Your `/etc/config/network` configuration containing Wireguard Peers.
4. Output of the `wg show all dump` command.
Please mask all the sensitive information, but in a way that we can see how IPs, keys, and other parameters are defined.

this-username-has-been-taken:
I was able to replicate the issue.
Indeed, a user can define a peer without port: in this case a default port (51820) will be used, but configPort will equal to `null`.

I suggest to fix the code and handle this situation by adding a condition into if block:
```
if (configPort && (ip + ":" + configPort == hostIp) || (ip == substr(hostIp, 0, index(hostIp, ":")))) {
                                return true;
                        }
```
If a port is defined we will use it and IP. If it is not defined we will compare only IP, but we also should trim the port from the `hostIp` variable because `wg show all dump` will return it anyway.
I have tested the fix on the live OpenWrt, but only with IP v4 - it works fine.

systemcrash:
How about the following which would be more readable and parse-able:
```js
if (configPort && (ip + ":" + configPort == hostIp) ) {
  return true;
} else if ( ip == substr(hostIp, 0, index(hostIp, ":")) ) {
  return true;
}
```
(please test, obviously)

this-username-has-been-taken:
I prefer one-liners after Groovy. But if you think that `else if` is more readable - I am OK with that.
I kindly ask to check this PR first: https://github.com/openwrt/luci/pull/7453 - otherwise there will be conflict.

systemcrash:
else if is more maintainable, in case other conditions arise in the future. As much as I love one-liners :)

this-username-has-been-taken:
OK.
I have just checked `else if` variant with IP v4 - everything works fine.
I can't check for IP v6. But it should also work fine.

Should I create a new branch which includes changes from https://github.com/openwrt/luci/pull/7453 and commit there?

systemcrash:
Just add a new commit to the existing PR. That should be fine.

xyyx:
> @xyyx Hi! Let's investigate a bit further. Can you please share:
> 
> 1. If you are using Wireguard as a server or client?
> 2. Your current `luci.wireguard` file.
> 3. Your `/etc/config/network` configuration containing Wireguard Peers.
> 4. Output of the `wg show all dump` command.
>    Please mask all the sensitive information, but in a way that we can see how IPs, keys, and other parameters are defined.

1. I use as server.
2. Used https://github.com/this-username-has-been-taken/luci/blob/4b10f99b9ff808aeffd37ee4976646a3bc33ea1b/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
config wireguard_wireguard
	option description 'tv_box'
	option public_key 'xxx'
	option private_key 'xxx'
	list allowed_ips '172.16.0.5/32'
	option route_allowed_ips '1'
	option endpoint_host 'xxx'
	option endpoint_port '1234'

config wireguard_wireguard
	option description 'msk'
	option public_key 'xxx'
	option private_key 'xxx'
	option route_allowed_ips '1'
	option endpoint_host 'xxx'
	option endpoint_port '1234'
	list allowed_ips '172.16.0.6/32'
	list allowed_ips '192.168.11.0/24' 

after_connect
before_connect

this-username-has-been-taken:
Hm... Why do you put the Endpoint Host (`endpoint_host`) and the Endpoint Port (`endpoint_port`) parameters into a Peer definition (`msk`, `tv_box`, etc.) when configuring the wireguard interface as a server?

They should be omitted in this case: try to leave them blank and check if the peer is detected properly.

The idea behind it is:
1. Either you use a wireguard interface as a client: in this scenario you must put an Endpoint Host (`endpoint_host`) and optionally an Endpoint Port (`endpoint_port`) in a Peer's config.
2. Or you use a wireguard interface as a server: in this scenario you define an IP and a port for the interface (i.e. one for all Peers) under the General Settings tab (which equals to `config interface '{interface name}'` section in the config file). After that you shouldn't put these parameters in a Peer's config, because it is the only way to check if a wireguard interface runs in a "client" or a "server" mode while detecting the Peer's name. In the "client"mode we should also check the Peer's IP and port (if provided); in the "server" mode - we shouldn't.

xyyx:
Oh, it's my  mistake. I removed endpoints from peers. Thanks a lot

this-username-has-been-taken:
No problem and have a nice day! :)

PR #7463, opened 2024-12-10
luci-compat: CBI skip fields that do not satisfy depends

jjm2473:
If depends is not satisfied, the fvalue will be nil.

@jow- 

- [x] Tested on: (aarch64, openwrt 22.03, chrome) :white_check_mark:

A simple test code (`model/cbi/abc.lua`):

```lua
local m, s, o

m = Map("floatip", translate("FloatingGateway"))

s=m:section(NamedSection, "main", translate("Global settings"))
s.anonymous=true

o = s:option(Flag, "enabled", translate("Enable"))
o.rmempty = false

o = s:option(ListValue, "role", translate("Node Role"))
o.rmempty = false
o.widget = "select"
o:value("main", translate("Preempt Node"))
o:value("fallback", translate("Fallback Node"))

o = s:option(Value, "set_ip", translate("Floating Gateway IP"))
o.rmempty = false
o.datatype = "or(ip4addr,cidr4)"

o = s:option(Value, "check_ip", translate("Preempt Node IP"))
o.rmempty = false
o.datatype = "ip4addr"
o:depends("role", "fallback")

return m
```

Without this patch, when "role" changes from "fallback" to "main", save will failed with 'missing' error (`One or more required fields have no value!`).

### DIFF
--- a/modules/luci-compat/luasrc/cbi.lua
+++ b/modules/luci-compat/luasrc/cbi.lua
@@ -1443,7 +1443,7 @@ function AbstractValue.parse(self, section, novld)
 				--luci.util.append(self.map.events, self.events)
 			end
 		end
-	else							-- Unset the UCI or error
+	elseif fvalue ~= nil then			-- Unset the UCI or error
 		if self.rmempty or self.optional then
 			if not self.alias or
 			   not self.section.aliased or


jjm2473:
there is a bug when commit empty list
```
o = s:option(StaticList, "interfaces", translate("Interfaces"))
o:value('lan', 'lan')
o:value('wan', 'wan')

```

jow-:
Please do not introduce new options into the legacy Lua cbi library.

jjm2473:
> Please do not introduce new options into the legacy Lua cbi library.

It's not an option, 'nilasempty' just a flag for internal usage, frontend pass nil when list empty.


If we don't do this, how can we fix the luci bug? 

[luci-app-dockerman](https://github.com/openwrt/luci/blob/2da1efba3859df23fca35096b8e9d7e0821e0d8e/applications/luci-app-dockerman/luasrc/model/cbi/dockerman/configuration.lua#L17) is also affected by this bug. First switch "Remote Endpoint" to true, set "Remote Host" to "10.1.1.2", set "Remote Port" to "2375", save and apply; then switch "Remote Endpoint" to false. At this time, no matter how you set parameters, it will not be saved.

jjm2473:
The key point is that you should not check 'rmempty = false' when depends is not satisfied.

jjm2473:
Dockerman issue: https://github.com/openwrt/luci/issues/7508

systemcrash:
In this case, I think docker's `o.rmempty = true` for those two options could be the correct way forward. Have you tried that?

jjm2473:
@systemcrash 

> In this case, I think docker's `o.rmempty = true` for those two options could be the correct way forward. Have you tried that?

Maybe that can save. But when "Remote Endpoint" enabled,  user must set "Remote Host" and "Remote Port", so `o.rmempty` must be false.

systemcrash:
> so `o.rmempty` must be false.

why?

jjm2473:
> why?

When "Remote Endpoint"  mode, "Remote Host" and "Remote Port" are all required.

Issue #7478, opened 2024-12-17
luci: Reveal password button does not work when password manager enabled


andrewdodd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### screenshots or captures

https://github.com/user-attachments/assets/214f9ad2-8440-4714-ada4-0a89c512e037

### Actual behaviour

When I click the "reveal / hide" button on a password input, nothing happens if I have the LastPass password manager enabled.

### Expected behaviour

The password input should reveal the value.

### Steps to reproduce

1. Install a password manager (like LastPass)
2. Load any page with a password input
3. Enter a value in the password input
4. Confirm the password input is hiding the value from you
5. Click the reveal password button
6. See that the password is not revealed to you
7. Disable the password manager (from your browser's extension setting)
8. Reload the page and repeat the above steps to see that the password can now be revealed

### Additional Information

```shell
NAME="OpenWrt"
VERSION="23.05.5"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt 23.05.5"
VERSION_ID="23.05.5"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r24106-10cc5fcd00"
OPENWRT_BOARD="bcm27xx/bcm2711"
OPENWRT_ARCH="aarch64_cortex-a72"
OPENWRT_TAINTS=""
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt 23.05.5 r24106-10cc5fcd00"
```

### What browsers do you see the problem on?

Firefox

PR #7479, opened 2024-12-17
luci: Make password reveal work with password managers

andrewdodd:
Password managers (like LastPass etc) tend to add additional elements into the DOM for their own context menus. If this happens between the hide/reveal button and the password input, then the logic to reveal the password breaks.

This change updates the onclick handler to look backward in the DOM for the first sibling that is an input, and to then toggle the password/text type on that element.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -368,7 +368,7 @@ const UITextfield = UIElement.extend(/** @lends LuCI.ui.Textfield.prototype */ {
 			'id': this.options.id ? `widget.${this.options.id}` : null,
 			'name': this.options.name,
 			'type': 'text',
-			'class': this.options.password ? 'cbi-input-password' : 'cbi-input-text',
+			'class': `password-input ${this.options.password ? 'cbi-input-password' : 'cbi-input-text'}`,
 			'readonly': this.options.readonly ? '' : null,
 			'disabled': this.options.disabled ? '' : null,
 			'maxlength': this.options.maxlength,
@@ -384,8 +384,15 @@ const UITextfield = UIElement.extend(/** @lends LuCI.ui.Textfield.prototype */ {
 					'title': _('Reveal/hide password'),
 					'aria-label': _('Reveal/hide password'),
 					'click': function(ev) {
-						const e = this.previousElementSibling;
-						e.type = (e.type === 'password') ? 'text' : 'password';
+						// DOM manipulation (e.g. by password managers) may have inserted other
+						// elements between the reveal button and the input. This searches for
+						// the first <input> inside the parent of the <button> to use for toggle.
+						const e = this.parentElement.querySelector('input.password-input')
+						if (e) {
+							e.type = (e.type === 'password') ? 'text' : 'password';
+						} else {
+							console.error('unable to find input corresponding to reveal/hide button');
+						}
 						ev.preventDefault();
 					}
 				}, '∗')


andrewdodd:
This is a possible patch for https://github.com/openwrt/luci/issues/7478

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -384,8 +384,18 @@ const UITextfield = UIElement.extend(/** @lends LuCI.ui.Textfield.prototype */ {
 					'title': _('Reveal/hide password'),
 					'aria-label': _('Reveal/hide password'),
 					'click': function(ev) {
-						const e = this.previousElementSibling;
-						e.type = (e.type === 'password') ? 'text' : 'password';
+						var e = this.previousElementSibling;
```
systemcrash, line 387: Please amend to `let e = ...`

andrewdodd: Done, thanks for the catch!

systemcrash:
Drop the merge commit. Rebase. Squash commits.

systemcrash:
Make sure to rebase as other changes in master are conflicting.

andrewdodd:
I had a look at the test build, but the log is so long I can't really find where to look for clues ("error"; "failure" and "failed" appear too much to use them to find the failure).

Is there something I should try here?

andrewdodd:
@systemcrash , sorry to hassle, but I'm not sure if there is something I should do next, could you please let me know?

andrewdodd:
I removed the change to the `applications/luci-app-dockerman/luasrc/view/dockerman/cbi/inlinevalue.htm` file, as it is not essential and I think might be causing the build to fail.

andrewdodd:
Hi @stokito , perhaps I can explain the change.

The existing code works by having the `onclick` handler for the reveal button look for the previous element in the DOM, and changing its type between password and text.

The existing code assumes that the previous element an `<input>`.

Password managers typically insert another DOM element, which invalidates this assumption.

The change works by looking back through the chain of previous element siblings to find the first `<input>` and then changing the type of that between "password" and "text".

The change also introduces an `else { console.log() }` to indicate if the handler was unable to find a corresponding `<input>`, instead of failing silently as per the current implementation.

andrewdodd:
Hi Stokito,

I had a go with [`.closest()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest), but I couldn't get it to be any simpler.

I think the issue is that `closest()` "traverses the element and its parents", but what we need is the sibling.

E.g. we typically have a DOM like:
```
<element>
  <input type='password'>
  <button>Reveal/Hide>
</element>
```

But sometimes extra items show up in between:

```
<element>
  <input type='password'>
  <other things>
  <button>Reveal/Hide>
</element>
```

And `closest()` searches up the tree from `button -> element -> etc`, not to the siblings.

I'm happy to redo the PR, but I'm not sure with what. Do you have any tips, or is this one sufficient?

andrewdodd:
I tried to keep this change as close to the original code. Instead of _"look to the sibling before me"_, it is _"look to the sibling before that is an `input`"._ This is still just one thing that the code thinks about. 

I worry that adding a class to the element and then using a "go to my parent, then search inside" now requires two concepts to be in the code. Are you sure this will be a less complex approach?

andrewdodd:
I had a go and I think it is similar complexity, but I updated the PR to use the approach you suggested.

systemcrash:
I don't think things like these are bad ideas per sé, but it's for a very exclusive problem regarding a browser plugin that externally mutates the DOM. The luci design paradigm involves putting elements within elements, although the JS should ideally work consistently. Can you show that this change works with and without the plugin?

andrewdodd:
Hi,

I agree this is a minor problem, but it is not just about a specific plugin or DOM manipulation.

The core problem is that the reveal button onclick handler is imprecise about what to toggle the `type` of - it just chooses *"whatever element is left of me"*. Ideally the onclick would precisely identify the `<input>` it corresponds to.

This PR changes the logic to be more precise - to *"choose the first `<input>` with class `password-input` that is in my immediate parent"*.

This guarantees the toggle will only be attempted on an `input`. It is still possible for the onclick to pick the incorrect element, if something has inserted another `<input class='password-input'>` before the expected target input within the reveal button's immediate parent - but this is very unlikely.

Here's a video of this change working both with and without the plugin enabled:

https://github.com/user-attachments/assets/22c98917-ccc2-4cd2-856f-4ddf73d589ee

PR #7480, opened 2024-12-18
luci-app-pbr: update to 1.1.8

stangri:
Maintainer: me
Compile tested: x86_64, Dell EMC Edge620, OpenWrt 24.10.0-rc2
Run tested: x86_64, Dell EMC Edge620, OpenWrt 24.10.0-rc2

Description:
* match version with principal package
* update README version
* use var instead of const for RPCD objects


systemcrash:
> use var instead of const for RPCD objects

Particular reason for this? Curious.

stangri:
> > use var instead of const for RPCD objects
> 
> Particular reason for this? Curious.

I think it came from a linter.

PR #7481, opened 2024-12-18
luci-app-upnp: Add `Expires` to port map listing

Self-Hosting-Group:
### Status/overview UI:

openwrt-status-overview
_[Screenshot 23.05 release](https://github.com/user-attachments/assets/4a9748fb-6a4b-420b-bfee-1ec4cce88d67)_

The **Port Control Protocol (PCP)** is the successor to NAT-PMP, has similar protocol concepts and packet formats, but adds support for **IPv6 port maps** and options/extensions. For more information, see:
**Port Mapping Protocols Overview and Comparison 2024: About UPnP IGD & PCP/NAT-PMP**

Maintainer: @jow-

Previous improvement: #7407

### DIFF
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
@@ -4,8 +4,6 @@
 'require rpc';
 'require uci';
 
-
-
 const callUpnpGetStatus = rpc.declare({
 	object: 'luci.upnp',
 	method: 'get_status',
@@ -37,7 +35,6 @@ return baseclass.extend({
 	},
 
 	render: function(data) {
-
 		var table = E('table', { 'class': 'table', 'id': 'upnp_status_table' }, [
 			E('tr', { 'class': 'tr table-titles' }, [
 				E('th', { 'class': 'th' }, _('Client Name')),
@@ -45,6 +42,7 @@ return baseclass.extend({
 				E('th', { 'class': 'th' }, _('Client Port')),
 				E('th', { 'class': 'th' }, _('External Port')),
 				E('th', { 'class': 'th' }, _('Protocol')),
+				E('th', { 'class': 'th right' }, _('Expires')),
 				E('th', { 'class': 'th' }, _('Description')),
 				E('th', { 'class': 'th cbi-section-actions' }, '')
 			])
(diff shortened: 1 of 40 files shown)


applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -49,12 +53,27 @@ return view.extend({
...
+			second = Math.floor(expires_sec % 60);
+			if (hour >= 1) {
+				expires_str += hour + 'h ';
+				expires_str += padnum(minute, 2) + 'm ';
+				expires_str += padnum(second, 2) + 's';
+			} else if (minute >= 1) {
+				expires_str += minute + 'm ';
+				expires_str += padnum(second, 2) + 's';
+			} else if (expires_sec >= 1) expires_str += second + 's';
+
```
systemcrash, line 69: test and consider:
```js
    const expires_sec = (rule.expires * 1000 - Date.now()) / 1000;
    const hour = Math.floor(expires_sec / 3600);
    const minute = Math.floor((expires_sec % 3600) / 60);
    const second = Math.floor(expires_sec % 60);

    const expires_str = 
        hour > 0 ? `${hour}h ${padnum(minute, 2)}m ${padnum(second, 2)}s` :
        minute > 0 ? `${minute}m ${padnum(second, 2)}s` :
        expires_sec > 0 ? `${second}s` :
        '';
```

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -36,6 +34,12 @@ function handleDelRule(num, ev) {
 	callUpnpDeleteRule(num);
 }
 
+function padnum(num, length) {
+	num = num.toString();
+	while (num.length < length) num = "0" + num;
+	return num;
+}
+
```
systemcrash, line 42: test this and revise to:
```js
const padnum = (num, length) => num.toString().padStart(length, "0");
```

systemcrash: We don't require it, but we generally aim for most new code as ES6 now.

systemcrash: (this function now appears to be missing)

systemcrash: Ah, good. 👀

systemcrash:
OK, did you re-run i18n-sync after the js changes?

systemcrash:
Above?

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
```diff
@@ -53,12 +51,24 @@ return baseclass.extend({
 		var rules = Array.isArray(data[0].rules) ? data[0].rules : [];
 
 		var rows = rules.map(function(rule) {
+			const padnum = (num, length) => num.toString().padStart(length, "0");
+			const expires_sec = (rule.expires * 1000 - Date.now()) / 1000;
```
systemcrash, line 55: In the unlikely event that the expires property is unavailable, how about handling it a bit better (so in worst case we get 0):
```js
			const expires_sec = ((rule?.expires * 1000 || Date.now()) - Date.now()) / 1000;
```

systemcrash:
Sure. Test the CSS for that column. But it might be a bit over-engineered to do so. Maybe `Expiry` would be better?

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
systemcrash: Although this is where we handle the value in the front end.

systemcrash: Feels safer that way. Yes please.

systemcrash:
Add the equivalent `style` to that column.

systemcrash:
Merged. Thanks @Self-Hosting-Group

PR #7484, opened 2024-12-18
luci-mod-system: use DirectInterface for bound dropbear interface

dannil:
Since https://github.com/openwrt/openwrt/commit/3f96246e97215c4c76ca407a8bca8f3f5de32e1c the correct UCI option for specifying the bound interface is DirectInterface instead of Interface.

- [X] Tested on: (x86/64, 24.10.0-rc2, Firefox) :white_check_mark:
- [ ] \( Preferred ) Mention: @ the original code author for feedback @jow- 

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
@@ -17,7 +17,7 @@ return view.extend({
 		o = s.option(form.Flag, 'enable', _('Enable Instance'), _('Enable <abbr title="Secure Shell">SSH</abbr> service instance'));
 		o.default  = o.enabled;
 
-		o = s.option(widgets.NetworkSelect, 'Interface', _('Interface'), _('Listen only on the given interface or, if unspecified, on all'));
+		o = s.option(widgets.NetworkSelect, 'DirectInterface', _('Interface'), _('Listen only on the given interface or, if unspecified, on all'));
 		o.nocreate    = true;
 
 		o = s.option(form.Value, 'Port', _('Port'));


dannil:
One thing I noticed is that if you upgrade from 23.05.5 to 24.10.0-rc2 is that the UI will show the interface as "unspecified", as the option isn't specified. Do we have a way to handle that or do we even need to? It's purely a cosmetic issue as the init script still checks for `Interface` if specified, and if both `DirectInterface` and `Interface` is specified, `DirectInterface` takes precedence.

systemcrash:
It looks like we're forced to choose. I've used Interface for years and never had a single problem. I have several IPs on a single interface and each work fine.

dannil:
> It looks like we're forced to choose. I've used Interface for years and never had a single problem. I have several IPs on a single interface and each work fine.

As long as `Interface` is supported for specifying interfaces everything will continue to work as before, it's the wording of "taking precedence" that made me act as it sounds like `DirectInterface` will replace `Interface` at some point in the future for binding to interfaces, `Interface` is meant to bind to raw IP addresses and is the intended usage going forward, at least after the referenced commit.

systemcrash:
> `Interface` is meant to bind to IP addresses and is the intended usage going forward

Ah, so they changed the behaviour of the existing option?

dannil:
> > `Interface` is meant to bind to IP addresses and is the intended usage going forward
> 
> Ah, so they changed the behaviour of the existing option?

That is at least my understanding, yes. You can see that they added a warning for specifying interfaces as an option for `Interface` at https://github.com/openwrt/openwrt/commit/3f96246e97215c4c76ca407a8bca8f3f5de32e1c#diff-af469bfcaffdaa7e73a3e127c5a312143b09002c82d226f9fae9cc042227f189R210

systemcrash:
The code comments do not give me much confidence:
```sh
	# 'DirectInterface' should specify single interface
	# but end users may misinterpret this setting
	DirectInterface=$(normalize_list "${DirectInterface}")
	# 'Interface' should specify single interface
	# but end users are often misinterpret this setting
	Interface=$(normalize_list "${Interface}")
```

systemcrash:
What we have today in luci works, so we can provide a parallel option. But only when it's clear what should do what. The original issue seems to discuss the problem when someone does the following, which is what the init script perhaps choked on
```
option Interface 'lan'
option Interface 'foo'
...
```

systemcrash:
Luci paradigm is currently correct. It allows one to create multiple instances, each instance with exactly one Interface.

See @stangri comment here https://github.com/openwrt/packages/issues/25555

PR #7492, opened 2024-12-21
luci-app-file-plug-manager: Full featured file manager with plugins

rdmitry0911:
- minimal working set: file-plug-manager + Navigation plugin. All other plugins are optional
 - Directory Navigation and Traversal
 - File and Directory Operations
 - Bulk Selection and Management
 - drug'n'drop to/from local computer
 - drug'n'drop to router local folders
 - copy/move mode for drug'n'drop switching with 'Alt' key
 - Editing text files with find/replace functionality
 - Hex editor
 - Save/load settings
 - Dumb Terminal

- [x ] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x ] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x ] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [x ] Incremented :up: any `PKG_VERSION` in the Makefile
- [ramips/mt7621, x86/64, OpenWrt 23.05.2 Chrome, Firefox ] Tested on: (architecture, openwrt version, browser) :white_check_mark:
- [x ] \( Preferred ) Screenshot or mp4 of changes:
- [luci-app-filemanager with plugins support and enchanced functionality ] Description: (describe the changes proposed in this PR)
Screenshot 2024-12-19 at 2 06 03 PM
Screenshot 2024-12-19 at 2 03 44 PM
Screenshot 2024-12-19 at 2 02 39 PM
Screenshot 2024-12-19 at 2 02 13 PM
Screenshot 2024-12-19 at 2 01 11 PM
Screenshot 2024-12-19 at 2 00 53 PM


systemcrash:
Perhaps revise to `drag and drop`. And at least a monospace font for any terminal. Do we need a terminal in a file manager?

rdmitry0911:
> I don't thing this worth a for separate app. Just add the "plugins" into the existing FM. The main problem is maintaining of this: you added an advanced functionality that is out of scope the Luci. And this is a complicated task by itself, for example showing an image with zoom and rotate.
> 
> I think the more stable solution would be to use an existing web file managers that are many lacks of contributors e.g.
> 
> 1. FileManager.org the most popular web UI with own backend in Go and frontend written in Vue https://github.com/filebrowser/filebrowser/tree/master/frontend
> 2. For the WebDAV
>    
>    1. [Chrome App](https://chromewebstore.google.com/detail/file-management-webdav/famepaffkmmhdefbapbadnniioekdppm) with [open sources](https://github.com/WebDAVDevs/webdav-browser-addon) but not sure if it can be reused.
>    2. https://github.com/dom111/webdav-js the simplest and outdated, written in React
> 3. WYSIWYG editors like FCKEditor and TinyMCE have [own file managers](https://www.tiny.cloud/docs/tinymce/latest/full-featured-premium-demo/)
>    Screenshot TinyMCE image gallery file manager
> 4. NextCloud and others
> 
> So if we might use the same codebase for all file managers that would be just great. But not sure if this possible.

I agree and I'm reconsidering the idea. I'll try to find a way to keep the router footprint as small as possible and extend the functionality by using client's or other external system resources.

applications/luci-app-file-plug-manager/htdocs/luci-static/resources/view/system/file-plug-manager.js
```diff
@@ -0,0 +1,903 @@
...
+	loadPlugins: function(buttonsContainer, contentsContainer) {
+		var self = this;
+
+		var dispatcherInfo = self.info();
+		self.pluginsRegistry[dispatcherInfo.name] = self;
+		self.default_plugins['Dispatcher'] = dispatcherInfo.name;
+
+		var pluginsPath = '/www/luci-static/resources/view/system/file-plug-manager/plugins/';
+
+		fs.exec('/bin/ls', [pluginsPath]).then(function(result) {
```
openwrt-ai, line 501: Discovering plugins by shelling out to `/bin/ls` on the hard-coded server path `/www/luci-static/...` is fragile: it breaks on any deployment whose docroot is not `/www` (configurable via uhttpd `home`), and it spends an rpcd `file.exec` round-trip just to enumerate this app's own static assets. Since the plugin filenames are fixed and shipped with the package, a static list (or `fs.list()` against the resource dir) avoids both the exec dependency and the path assumption.

---

applications/luci-app-file-plug-manager/htdocs/luci-static/resources/view/system/file-plug-manager/plugins/Navigation.js
```diff
@@ -0,0 +1,2848 @@
...
+		return specialBits + parseInt(numericPermissions, 10); // Add "special" bits
+	},
+
+	/**
+	 * Parses a single line of `ls -lA --full-time` output.
+	 * @param {string} line - A single line from the `ls` command output.
+	 * @returns {Object|null} - Returns an object with file information or null if parsing fails.
+	 */
+
+	parseLsLine: function(line) {
```
openwrt-ai, line 1785: Parsing `ls -lA --full-time` text with a regex is fragile and avoidable. LuCI's `fs.list(path)` (ubus `file.list`) already returns structured entries (`name`, `type`, `size`, `mtime`) with `stat` available for permissions/owner, so directory listings don't depend on the exact `ls` column layout or locale. Concrete failure modes of the current regex: filenames containing whitespace/newlines, and `new Date('2024-01-01 12:00:00.000000000 +0000')` (the `--full-time` format with nanoseconds) does not parse reliably across browsers, yielding `Invalid Date` for `mtime`. Recommend switching the listing path to `fs.list()`/`fs.stat()`.

---

openwrt-ai (commented):
## Commit checks
- `d543b4a` "luci-app-file-plug-manager: Full featured file manager with plugins" — this commit squashes two separate commit messages into one: it carries two `subject` lines ("Full featured file manager..." and "Fixed application name and monospace fonts usage across plugins") and two `Signed-off-by` blocks. The effective subject is also the entire bulleted feature list rather than a short summary line. Please reword into a single commit message with one concise subject and one Signed-off-by.

Note for context (not a line-level finding): the ACL in `acl.d/luci-app-file-plug-manager.json` grants `file: { "/*": [list, read, write, exec] }`, `ubus file ["*"]`, and `cgi-io upload/download`. This is identical to the existing `luci-app-filemanager` ACL, so it is consistent with in-tree precedent — flagging only so the broad (effectively full-root) scope is a conscious choice, given this app additionally ships a terminal plugin built on `fs.exec()`.

---

PR #7495, opened 2024-12-22
luci-mod-system: give crontab (aka scheduled tasks) some helpers

systemcrash:
- [x] Tested on: 23.05.5, Firefox :white_check_mark:

Screenshot 2024-12-23 at 15 51 44

Any input from @dannil @stokito @Ramon00 @hnyman @rdmitry0911 ?

Gave crontab a small makeover. It's raw and rudimentary but it is already an improvement over editing the raw crontab.

Good if others can take it for a spin or suggest improvements. ( inter-row and element styling )

I do not recommend putting any safeguards in for the command field, since we were already editing the crontab directly.

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontabhelper.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontabhelper.js
@@ -1,0 +1,334 @@
+'use strict';
+'require view';
+'require fs';
+'require ui';
+
+var isReadonlyView = !L.hasViewPermission() || null;
+
+
+const yearly  = { minute: '@yearly',  command: '', comment: '', };
+const monthly = { minute: '@monthly', command: '', comment: '', };
+const weekly  = { minute: '@weekly',  command: '', comment: '', };
+const daily   = { minute: '@daily',   command: '', comment: '', };
+const hourly  = { minute: '@hourly',  command: '', comment: '', };
+const a_task  = { minute: '*', hour: '*', day: '*', month: '*', weekday: '*', command: '', comment: '', };
+const alias   = { minute: '@', hour: '*', day: '*', month: '*', weekday: '*', command: '', comment: '', };
+
+const width = 'width:100px;';
+const double_width = 'width:300px';
+const padding = 'padding:3px;';
+const centre = 'text-align:center;';
+
+return view.extend({
+	load() {
+		return L.resolveDefault(fs.read('/etc/crontabs/root'), '');
+	},
+
+	handleSave(ev) {
+		const tasks = Array.from(document.querySelectorAll('.crontab-row')).map(row => {
+			const getFieldValue = (fieldName) => {
+				const mode = row.querySelector(`.${fieldName}-mode`)?.value;
+
+				switch (mode) {
+				case 'custom':
+					const custom = row.querySelector(`.${fieldName}-custom`)?.value?.trim();
+					return custom;
+				case 'ignore':
+					return '*';
+				case 'interval':
+					const interval = row.querySelector(`.${fieldName}-interval`)?.value?.trim();
+					return interval ? `*/${interval}` : '*'; // Every Xth unit
+				case 'specific':
+					const specific = row.querySelector(`.${fieldName}-specific`)?.value?.trim();
+					return specific || '*'; // Specific units
+				}
+			};
+
+			const comment = row.querySelector('.comment')?.value?.trim();
+			const minute  = row.querySelector('.minute')?.value?.trim();
+
+			// if it's a # comment row, just stuff the comment and return
+			if (minute == comment)
+				return { iscomment: true, comment: comment };
+			else
+				return {
+					minute: getFieldValue('minute') || '*',
+					hour: getFieldValue('hour') || '*',
+					day: getFieldValue('day') || '*',
+					month: getFieldValue('month') || '*',
+					weekday: getFieldValue('weekday') || '*',
+					command: row.querySelector('.command')?.value?.trim(),
+					comment: comment ? `# ${comment}` : '',
+				};
+		});
+
+		const value = tasks.map(task => {
+			if (task.iscomment)
+				return `${task.comment}`;
+			else if (task.minute[0] !== '@')
+				return `${task.minute} ${task.hour} ${task.day} ${task.month} ${task.weekday} ${task.command} ${task.comment}`;
+			else
+				return `${task.minute} ${task.command} ${task.comment}`;
+		}).join('\n') + '\n';
+
+		return fs.write('/etc/crontabs/root', value).then(() => {
+			ui.addTimeLimitedNotification(null, E('p', _('Contents have been saved.')), 5000, 'info');
+			return fs.exec('/etc/init.d/cron', [ 'reload' ]);
+		}).catch(e => {
+			ui.addNotification(null, E('p', _('Unable to save contents: %s').format(e.message)));
+		});
+	},
+
+	render(crontab) {
+		const tasks = (crontab || '').split('\n').filter(line => line.trim()).map(line => {
+			if (line.startsWith('#'))
+				return {
+					// stash comment lines for saving later
+					minute:  line,
+					comment: line,
+				};
+			const parts = line.split(/\s+/);
+			const commentIndex = parts.findIndex(part => part.startsWith('#'));
+			// exclude the '#' character from comments existing in valid command rows:
+			if (commentIndex !== -1) parts[commentIndex] = parts[commentIndex].substring(1)?.trim();
+			const comment = commentIndex !== -1 ? parts.slice(commentIndex).join(' ') : '';
+
+			if(parts[0][0] == '@') {
+				const updatedTask = { ...alias };
+				updatedTask.minute = parts[0];
+				updatedTask.command = commentIndex !== -1 ? parts.slice(1, commentIndex).join(' ') : parts.slice(1).join(' ');
+				updatedTask.comment = comment;
+				return updatedTask;
+			}
+
+			return {
+				minute:  parts[0] || '',
+				hour:    parts[1] || '',
+				day:     parts[2] || '',
+				month:   parts[3] || '',
+				weekday: parts[4] || '',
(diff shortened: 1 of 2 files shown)


systemcrash:
> Interesting how many people and how use the cron. The UI improvements also increase a size but it looks like that's not a problem anymore. From the screenshot I can tell that:

Not many users? Possibly because it remains cryptic.

>     1. The original editor needs to be preserved so if if someone just copy-paste from instructions.

Fair point. I thought retaining both original window would be possible. Requires two save buttons and routines.

>     2. The command field needs to be wider.

Fixable. But it then affects whichever column it appears in.

>     3. We may use the same visual language of a table/sections

Also thought about this. Probably more maintainable. 

Take it for a test-drive and see how it works for you.

Ramon00:
The layout needs some work, especially if the table is empty. See e.g.: 
image
Looks like you need to fill in an hour value in the empty box instead of the command. Is it supposed to work already because if I hit save then I just get the rotating circle thing (on 23.05.5).

systemcrash:
> The layout needs some work, especially if the table is empty. See e.g.: image Looks like you need to fill in an hour value in the empty box instead of the command. Is it supposed to work already because if I hit save then I just get the rotating circle thing (on 23.05.5).

Yeah, that’s the command value. The minute column could occupy six columns to ensure the command column is placed correctly. But then which column is minute? 

It should work. Check the console.

Ramon00:
image

hnyman:
To me the proposed editor looks confusing compared to the direct crontab file editing. 

It might be better to just provide more help text about the fields with the old crontab editor.

systemcrash:
Probably. I think this can be provided side-by-side instead of replacing the existing crontab editor.

Ramon00:
yes side-by-side would be perfect, i.e. show and allow editing of the resulting cron rules. Maybe also good to add a comment field?

systemcrash:
> yes side-by-side would be perfect, i.e. show and allow editing of the resulting cron rules. Maybe also good to add a comment field?

Side-by-side it is. A comment field only has a place if the crontab itself also permits it. Otherwise it's a code comment at the end of the command. e.g. `command.sh param # comment`

systemcrash:
> image

Fixed.

> Maybe also good to add a comment field?

Added.

> Maybe it would be enough to add a link to a wiki https://openwrt.org/docs/guide-user/base-system/cron.

Added.

Ramon00:
> Fixed

I just tested it and indeed now it is functional.

Cosmetic remark, maybe add some pixels space here between the two boxes? 
image

One more remark, if I click on the pulldown select "every Xth" it then changes to: 
image
If i then hit save, i get a 
`*/5 undefined undefined undefined undefined`
in the file.

systemcrash:
Good find. Fixed now. That'll get converted so that next page load it's no longer an alias but a normal task format.

Anyway, the pretty layout needs some love.

systemcrash:
maybe @dannil or @stokito know some good CSS to center the table as a whole instead of all the rows stretching off to the right of the window?

systemcrash:
Screenshot 2025-04-23 at 18 16 16

openwrt-ai (commented):
Reviewed 1 new commit. The commit message header/body match the changes (new `crontabhelper.js` view + menu entry, valid Signed-off-by). No commit-message issues. Inline findings concern two un-extractable/malformed `_()` i18n calls, a comment-row detection scheme (`minute == comment`) that can silently drop valid task rows, and a missing `|| ''` fallback on the command field that can serialize the literal string `undefined`. CI status on the head commit is green; these are static-review findings.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontabhelper.js
```diff
@@ -0,0 +1,334 @@
...
+				]),
+				E('div', { 'class': `${fieldName}-input interval-input`, 'style': mode === 'interval' ? width : 'display:none;' }, [
+					E('label', {}, _('Every')),
+					E('input', { 'type': 'number', 'min': min, 'max': max, 'class': `${fieldName}-interval`, 'value': intervalValue, 'style': width }),
+					E('span', {}, _(label.toLowerCase()))
+				]),
+				E('div', { 'class': `${fieldName}-input specific-input`, 'style': mode === 'specific' ? width : 'display:none;' }, [
+					E('label', {}, _('At')),
+					E('input', { 'type': 'text', 'class': `${fieldName}-specific`, 'value': specificValue, 'style': width, 'placeholder': '0,5,10,...' }),
+					E('span', {}, _(label.toLowerCase() + _('s (CSV)', 'pluralisation for hours, minutes, etc')))
```
openwrt-ai, line 289: This `_()` call cannot be translated and is malformed. It is called with a runtime-concatenated argument (`label.toLowerCase() + _(...)`), so the i18n scanner (`i18n-scan.pl` / xgettext) can never extract the string, and a `_()` is nested inside the argument of another `_()`, which is not a supported pattern. `label.toLowerCase()` ('minute', 'hour', ...) is also not a catalog key and won't be translated. Build the visible label from a fixed translatable format string instead:

```suggestion
					E('span', {}, _('%s (CSV)').format(label.toLowerCase()))
```

Note `%s` will still embed the untranslated English unit word; if per-unit translation is needed, pass an already-translated label into `createTimeDropdown` rather than lowercasing here.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontabhelper.js
```diff
@@ -0,0 +1,334 @@
...
+					}, _('Custom'))
+				]),
+				E('div', { 'class': `${fieldName}-input ignore-input`, 'style': mode === 'ignore' ? '' : 'display:none;' }, [
+					E('input', { 'type': 'text', 'class': fieldName, 'value': '*', 'style': mode === 'ignore' ? 'display:none;': width,
+					})
+				]),
+				E('div', { 'class': `${fieldName}-input interval-input`, 'style': mode === 'interval' ? width : 'display:none;' }, [
+					E('label', {}, _('Every')),
+					E('input', { 'type': 'number', 'min': min, 'max': max, 'class': `${fieldName}-interval`, 'value': intervalValue, 'style': width }),
+					E('span', {}, _(label.toLowerCase()))
```
openwrt-ai, line 284: Same i18n issue as line 289: `_(label.toLowerCase())` passes a dynamic, lowercased string to `_()`, so it won't be extracted into the message catalog and won't be translated. Pass a pre-translated label down from `createTimeDropdown`'s callers (which already hold the English label) instead of translating a runtime-lowercased value here.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontabhelper.js
```diff
@@ -0,0 +1,334 @@
...
+					return specific || '*'; // Specific units
+				}
+			};
+
+			const comment = row.querySelector('.comment')?.value?.trim();
+			const minute  = row.querySelector('.minute')?.value?.trim();
+
+			// if it's a # comment row, just stuff the comment and return
+			if (minute == comment)
+				return { iscomment: true, comment: comment };
```
openwrt-ai, line 52: Detecting a comment row via `minute == comment` can silently drop valid task rows. For a real task row, `.minute` here is the value of the `.minute`-class input, but a regular task row easily produces `minute == comment` too — most notably when both are empty/whitespace (`undefined == undefined`, or `'' == ''`), or whenever the user happens to type the same text in both fields. Such a row is then classified `iscomment` and serialized at line 67 as just `${task.comment}`, discarding minute/hour/.../command entirely. Recommend marking comment rows explicitly (e.g. a `data-iscomment` attribute or a dedicated class set in `renderCommentRow`) and testing that, instead of value equality.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontabhelper.js
```diff
@@ -0,0 +1,334 @@
...
+			else
+				return {
+					minute: getFieldValue('minute') || '*',
+					hour: getFieldValue('hour') || '*',
+					day: getFieldValue('day') || '*',
+					month: getFieldValue('month') || '*',
+					weekday: getFieldValue('weekday') || '*',
+					command: row.querySelector('.command')?.value?.trim(),
```
openwrt-ai, line 60: `command` has no `|| ''` fallback unlike every adjacent field (`getFieldValue('minute') || '*'`, etc.). If `.command` is absent or `?.value` is undefined, `command` becomes `undefined` and is templated into the crontab line at line 69/71 as the literal text `"undefined"` — the same class of symptom (`*/5 undefined undefined ...`) reported earlier in this thread. Was this path meant to be covered when that was marked fixed?

```suggestion
					command: row.querySelector('.command')?.value?.trim() || '',
```

---

openwrt-ai (commented):
Inline findings for this PR (the four code comments failed to attach to my earlier review on this commit due to an API issue; reposting them here). They concern: two un-extractable/malformed `_()` i18n calls in `createTimeDropdown`, comment-row detection via `minute == comment` that can silently drop valid task rows, and a missing `|| ''` fallback on the command field that can serialize the literal string `undefined`. Commit message and Signed-off-by are fine. CI on the head commit is green; these are static-review findings.

---

feckert:
@systemcrash I think that’s an interesting feature. Could you have another look at what the AI does not like and perhaps fix it?

jow-:
Did you consider introducing an "add crontab entry" button which spawns a modal containing your proposed form which then simply appends a raw line to the existing textbox on save/close?

Would give the best of both worlds. People following instructions can simply copy paste and people building own entries but not remembering the syntax in detail can format it through the modal wizard.

stokito:
@jow~ this should be the GUI first. But for editing we may call the File Manager with edit mode. Editing may be reused in many places. 

It would be great to check similar UIs:
CPanel
CPanel crontab

https://github.com/alseambusher/crontab-ui
crontab-ui

KCron
KCron new task

Zeit
Zeit new task

KTimer
KTimer

Others: Gcron, FcronQ, GNOME Schedule

PR #7496, opened 2024-12-22
treewide: Add a link to Weblate

stokito:
To help users to find a right place where to translate.

Because some users trying to edit the po files directly. This also should be acceptable because we shouldn't force users to register on a third party tool and learn it.

Ideally we should add a link to translations, github and dourm into the Luci itself. This must increase contributions from regular users.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,9 @@
 # Contributing Guidelines
 
+## Translations
+
+Use [Weblate](https://hosted.weblate.org/engage/openwrt/?utm_source=widget) instead of direct editing of the `*.po` files.
+
 ## Patches and Pull requests:
 
 If you want to contribute a change to LuCI, please either send a patch using git send-email
--- a/README.md
+++ b/README.md
@@ -40,4 +40,6 @@ See [CONTRIBUTING.md](CONTRIBUTING.md) file.
 
 ## Translation status
 
+Use [Weblate](https://hosted.weblate.org/engage/openwrt/?utm_source=widget) instead of direct editing of the `*.po` files.
+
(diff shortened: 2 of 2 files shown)


hnyman:
And what is wrong there?
(You  can't translate source strings, which Weblate shows as developer English.)

hnyman:
Looks like weblate applies a global default link from somewhere for all apps:
`https://github.com/openwrt/luci/blob/%7Bbranch%7D/%7Bfilename%7D#L{line}`

The template link has actually a single { instead of double {{, and first two instances are escaped via % sign, while the last one is a hardcoded {

while it should be
`https://github.com/openwrt/luci/blob/master/{{filename}}#L{{line}}`

image

I fixed manually upnp, so the link there should now work.

But the same "Repository browser" field is empty for all other 100 translatable apps, and the wrong global default just pops from somewhere. 
I haven't yet found any global setting for that.

Ps. we haven't set the translation branch setting separately for each app, so weblate uses the git default, in practice master.

hnyman:
Documenting here that I think that I fixed the `Source string location` links by setting the `Repository browser` setting in luci-base app.  
`https://github.com/openwrt/luci/blob/{{branch}}/{{filename}}#L{{line}}`

luci-base is apparently the underlying root app in weblate config, so some of the changes made to it will also affect the other apps unless overridden by the app settings.  Good to remember in future.

hnyman:
@stokito 
Back to the original idea of this PR:
Why your proposed link addition has the "?utm_source=widget" part?
It is https://hosted.weblate.org/engage/openwrt/?utm_source=widget 
instead of the plain link https://hosted.weblate.org/engage/openwrt/ ?

systemcrash:
Merged. Thanks @stokito

PR #7499, opened 2024-12-23
luci-mod-status: add simple filtering for connections

McGiverGim:
This commit adds a simple filtering input text into the connections status page. You can type some filtering like:
- '100.15' -> Will show only connections that include this '100.15' onto the IP.
- 'UDP 100.15' -> Will show only UDP connections that include this '100.15' onto the IP.

You can use a part of the IP or part of the name if the DNS resolution button is enabled.

It's a very simple filter that can grow in the future if it deserves some interest, maybe adding some regular expression support.

image

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
@@ -27,7 +27,8 @@ var callNetworkRrdnsLookup = rpc.declare({
 var graphPolls = [],
     pollInterval = 3,
     dns_cache = {},
-    enableLookups = false;
+    enableLookups = false,
+    filterText = '';
 
 var recheck_lookup_queue = {};
 
@@ -130,12 +131,27 @@ return view.extend({
 			var src = dns_cache[c.src] || (c.layer3 == 'ipv6' ? '[' + c.src + ']' : c.src);
 			var dst = dns_cache[c.dst] || (c.layer3 == 'ipv6' ? '[' + c.dst + ']' : c.dst);
 
+			const network = c.layer3.toUpperCase();
+			const protocol = c.layer4.toUpperCase();
+			const source ='%h'.format(c.hasOwnProperty('sport') ? (src + ':' + c.sport) : src);
+			const destination = '%h'.format(c.hasOwnProperty('dport') ? (dst + ':' + c.dport) : dst);
+			const transfer = [ c.bytes, '%1024.2mB (%d %s)'.format(c.bytes, c.packets, _('Pkts.')) ];
+
+			if (filterText) {
+				let filterTextExpressions = filterText.split(' ');
+				if (filterTextExpressions.some((element) => element.toUpperCase() !== network && element.toUpperCase() !== protocol 
+						&& !(c.src.includes(element) || source.includes(element))
+						&& !(c.dst.includes(element) || destination.includes(element)))) {
+					continue;
+				}
+			}
+
 			rows.push([
-				c.layer3.toUpperCase(),
-				c.layer4.toUpperCase(),
-				'%h'.format(c.hasOwnProperty('sport') ? (src + ':' + c.sport) : src),
-				'%h'.format(c.hasOwnProperty('dport') ? (dst + ':' + c.dport) : dst),
-				[ c.bytes, '%1024.2mB (%d %s)'.format(c.bytes, c.packets, _('Pkts.')) ]
+				network,
+				protocol,
+				source,
+				destination,
+				transfer,
 			]);
 		}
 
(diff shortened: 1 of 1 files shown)


modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
```diff
@@ -359,23 +367,48 @@ return view.extend({
...
+					style: 'display: flex',
+				}, [
+					E('span', { 
+						class: 'filter',
+						style: 'flex: auto',
+					}, [
+						E('input', {
+							id: 'filter-connections',
+							type: 'text',
+							placeholder: _('Type some filtering for connections...'),
```
systemcrash, line 383: Perhaps revise to something more 'useful' - like an example search?

jow-: Also please use `…` (&hellip;) instead of three ordinary dots.

McGiverGim: I was thinking the same, but it seemed too long to me to put examples, specially for small screens. But I can change it, maybe something short like: 
"Filter like '0.15' or 'tcp 0.15'”
I'm not English native, so any help will be appreciated...

McGiverGim:
I replaced the text with the samples (but I'm ready to change it if someone prefers a better wording) and added comparison with ignore case to protocol and network.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
```diff
@@ -130,12 +131,27 @@ return view.extend({
 			var src = dns_cache[c.src] || (c.layer3 == 'ipv6' ? '[' + c.src + ']' : c.src);
 			var dst = dns_cache[c.dst] || (c.layer3 == 'ipv6' ? '[' + c.dst + ']' : c.dst);
 
+			let network = c.layer3.toUpperCase();
+			let protocol = c.layer4.toUpperCase();
+			let source ='%h'.format(c.hasOwnProperty('sport') ? (src + ':' + c.sport) : src);
+			let destination = '%h'.format(c.hasOwnProperty('dport') ? (dst + ':' + c.dport) : dst);
+			let transfer = [ c.bytes, '%1024.2mB (%d %s)'.format(c.bytes, c.packets, _('Pkts.')) ];
```
systemcrash, line 138: consider:
```js
const network = c.layer3?.toUpperCase() || '';
const protocol = c.layer4?.toUpperCase() || '';
const source = c.hasOwnProperty('sport') ? `${src}:${c.sport}` : src;
const destination = c.hasOwnProperty('dport') ? `${dst}:${c.dport}` : dst;
const transfer = [ c.bytes, `${(c.bytes / 1024 / 1024).toFixed(2)} MB (${c.packets} ${_('Pkts.')})` ];
```

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
```diff
@@ -130,12 +131,27 @@ return view.extend({
...
+
+			if  (filterText) {
+				let filterTextExpressions = filterText.split(' ');
+				if (filterTextExpressions.some((element) => element.toUpperCase !== network && element.toUpperCase() !== protocol
```
systemcrash, line 142: Should that be a function?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
```diff
@@ -130,12 +131,27 @@ return view.extend({
...
+			let protocol = c.layer4.toUpperCase();
+			let source ='%h'.format(c.hasOwnProperty('sport') ? (src + ':' + c.sport) : src);
+			let destination = '%h'.format(c.hasOwnProperty('dport') ? (dst + ':' + c.dport) : dst);
+			let transfer = [ c.bytes, '%1024.2mB (%d %s)'.format(c.bytes, c.packets, _('Pkts.')) ];
+
+			if  (filterText) {
+				let filterTextExpressions = filterText.split(' ');
+				if (filterTextExpressions.some((element) => element.toUpperCase !== network && element.toUpperCase() !== protocol 
+						&& !(c.src.includes(element) || source.includes(element))
+						&& !(c.dst.includes(element) || destination.includes(element)))) {
```
systemcrash, line 144: Can we skip the c.* checks? They're implied in `source` and `destination`, yes?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
systemcrash: How about something simple like `UDP 192.168.1.1`? Then i18n is not necessary.

systemcrash:
It’s low effort, and high value when dealing with lots of endpoints, so it’s a welcome addition I think. @McGiverGim also thinks so.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
McGiverGim: True, I will change it.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
McGiverGim: In reallity it must return a 'truthy' value https://developer.mozilla.org/en-US/docs/Glossary/Truthy

You can use the complete function or like here, an arrow function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some#examples

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
McGiverGim: Not exactly... 
- the `c.src` and `c.dst` contains the IP always
- the `source` and `destination` contains the IP or the HOSTNAME (if the DNS resolution button is enabled) and the PORT

So this is a simple way to filter by IP or hostname. If the DNS resolution is disabled it will filter by IP twice, but if I try to represent this in the condition in the filter to use it only when the DNS resolution is disabled it will be a more complicated function, so I decided to let it simple.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
McGiverGim: The field does not have any "label" indicating what is this for, so for this reason I added the "filter" word to the placeholder. But if you think it's enough adding the sample without text, I can change it, no problem.

McGiverGim:
I think it's useful when you want to see the connections from a host, for example. If this is not needed, then the full list of connections it's not needed neither, because now it's a big list but you can't find anything because it's big and changes continuosly.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
systemcrash: You have `element.toUpperCase !== network` (not a function call)
but you have `element.toUpperCase() !== protocol` (function call)

McGiverGim: True, this is a bug :( I will change it. Good catch! I didn't understand your first explanation, sorry...

McGiverGim: Fixed!

modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
McGiverGim: Do I change this finally? If yes, what about different filters to show different samples with a separator?
`192.168.1.15 / UDP 1.15 / :443`

Maybe it's confusing...

systemcrash: That’s great

McGiverGim: Ok, changed as discused.

systemcrash:
Took it for a spin. Works fine.

systemcrash:
Thanks @McGiverGim

PR #7510, opened 2024-12-26
luci-app-acme: DNS API store all params

### LINKED ISSUE #7504: luci-app-acme: variables don't save using dns-api

- [X] I have searched the existing issues

### screenshots or captures

https://github.com/user-attachments/assets/c8b71a71-5198-4520-a31d-7f45f6ea7739

https://github.com/user-attachments/assets/4f7e78a3-e25f-4844-9ef4-ee9ec60b8aaa

### Actual behaviour

I noticed this while setting up cloudflare api, but i tested a few dns api providers and data input also failed to save.

While using the dns-api function, the various new input field don't save the variables.
First Video Example.

But manually adding the variable in the DNS API credentials area such as CF_Token="Example" saves and also displays in the corresponding box.
Second Video Example

I originally experienced this on my MT6000, but I performed a fresh install of the latest snapshot last night on my Netgear R8000 and confirmed this behavior.

### Expected behaviour

I would expect the fields to save the input data, and not just display the manually entered variables, thus making it redundant. 

### Steps to reproduce

1. Go to Services > ACME Certificates
2. Add cert config
3. Change Validation method to DNS
4. Go to DNS Challenge Validation tab
5. Change DNS API to CloudFlare.com
6. Enter Data. Into any of the Fields. 
   i.e.CF Key, CF Email, CF Token, CF Account ID, CF Zone ID
7. Save
8. Edit new entry to view empty fields.

### Additional Information

```shell
AME="OpenWrt"
VERSION="SNAPSHOT"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt SNAPSHOT"
VERSION_ID="snapshot"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r28576-a9895c4585"
OPENWRT_BOARD="mediatek/filogic"
OPENWRT_ARCH="aarch64_cortex-a53"
OPENWRT_TAINTS="busybox"
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt SNAPSHOT r28576-a9895c4585"
OPENWRT_BUILD_DATE="1734722468"
```

Firefox

### Relevant log output

_No response_

stokito:
Fix  #7504

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
@@ -529,7 +529,7 @@ return view.extend({
 
 
 function _addDnsProviderField(s, provider, env, title, desc) {
-	let o = s.taboption('challenge_dns', form.Value, '_' + env, _(title),
+	let o = s.taboption('challenge_dns', form.Value, '_credentials_' + env, _(title),
 		_(desc));
 	o.depends('dns', provider);
 	o.modalonly = true;
@@ -537,10 +537,27 @@ function _addDnsProviderField(s, provider, env, title, desc) {
 		var creds = this.map.data.get(this.map.config, section_id, 'credentials');
 		return _extractParamValue(creds, env);
 	};
-	o.write = function (section_id, value) {
-		this.map.data.set('acme', section_id, 'credentials', [env + '="' + value + '"']);
-	};
+	o.write = function (section_id, value) { };
+	o.onchange = _handleEditChange;
 	return o;
+}
+
+function _handleEditChange(event, section_id, newVal) {
+	// Add the provider field value directly to the credentials DynList
+	let credentialsDynList = this.map.lookupOption('credentials', section_id)[0].getUIElement(section_id);
+	let creds = credentialsDynList.getValue();
+	let credsMap = _parseKeyValueListToMap(creds);
+	let optName = this.option.substring('_credentials_'.length);
+	if (newVal) {
+		credsMap.set(optName, newVal);
+	} else {
+		credsMap.delete(optName);
+	}
+	let newCreds = [];
+	for (let [key, val] of credsMap) {
+		newCreds.push(key + '="' + val + '"');
+	}
+	credentialsDynList.setValue(newCreds);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


systemcrash:
This, in its present form, does not fix the bug. The data race is still present.

systemcrash:
When handling credentials, you’ll need to pull apart the variable and value, to determine whether an incoming variable already exists, and if so, replace it.

systemcrash:
Not quite. It checks to see whether an identical k+v exist in the set at add time, but not whether an identical k already exists with a different v. Such a check is necessary if a user changes their credentials for a provider.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -529,43 +529,67 @@ return view.extend({
...
 	if (!paramsKeyVals) {
-		return '';
+		return map;
 	}
-	for (let i = 0; i < paramsKeyVals.length; i++) {
-		var paramKeyVal = paramsKeyVals[i];
-		var parts = paramKeyVal.split('=');
-		if (parts.lenght < 2) {
+	for (let paramKeyVal of paramsKeyVals) {
+		let parts = paramKeyVal.split('=');
```
systemcrash, line 583:
```js
		const [name, val] = paramKeyVal.split('=');
```

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -529,43 +529,67 @@ return view.extend({
...
-		var name = parts[0];
-		var val = parts[1];
-		if (name == paramName) {
-			// unquote
-			return val.substring(0, val.length-1).substring(1);
-		}
+		let name = parts[0];
+		let val = parts[1];
+		let unquotedVal = val.substring(0, val.length - 1).substring(1);
+		map.set(name, unquotedVal);
```
systemcrash, line 590:
```js
		map.set(name, val.slice(1, -1) );
```

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
systemcrash: You split anyway. But no check for =. So the check is redundant, I think.

systemcrash: IIRC = is just tail padding in base64, and even if you truncate it away, the base64 still decodes fine.

systemcrash:
Merged. Good work!

PR #7511, opened 2024-12-27
luci-app-squid: remove variable redeclaration

dannil:
Since 22d4830 the form variables m, s, and o was changed to be declared with let instead of var, but the o variable was redeclared in the app, resulting in a redeclaration error when it was changed to instead be declared with let.

See https://forum.openwrt.org/t/openwrt-24-10-0-rc4-fourth-release-candidate/219368/81?u=dannil

- [X] Tested on: (x86/64, 24.10-rc4, Firefox) :white_check_mark:
- [X] \( Preferred ) Mention: @ the original code author for feedback @stokito and @dannil (myself)

Before
Screenshot_11

After
Screenshot_12

### DIFF
--- a/applications/luci-app-squid/htdocs/luci-static/resources/view/squid.js
+++ b/applications/luci-app-squid/htdocs/luci-static/resources/view/squid.js
@@ -49,7 +49,7 @@ return view.extend({
 		s.tab('general', _('General Settings'));
 		s.tab('advanced', _('Advanced Settings'));
 
-		var o = s.taboption('general', form.Value, 'config_file', _('Config file'));
+		o = s.taboption('general', form.Value, 'config_file', _('Config file'));
 		o.datatype = 'string';
 		o.default = '/etc/squid/squid.conf';
 		o.validate = function(section_id, value) {


systemcrash:
There's probably quite a few bugs which are seeing light for the first time since lots of additions went into master which is seeing its first release in a while. This is a good thing.

dannil:
> There's probably quite a few bugs which are seeing light for the first time since lots of additions went into master which is seeing its first release in a while. This is a good thing.

As a secondary point, we may want to check other apps if they have the same problem, I haven't seen any other reports of this issue on the forums but it could simply be that the apps aren't used often.

systemcrash:
> As a secondary point, we may want to check other apps if they have the same problem, I haven't seen any other reports of this issue on the forums but it could simply be that the apps aren't used often.

Also true. Probably the case for stuff in master :)

dannil:
I checked every app which had a diff in 22d48305afcb3095468756f087610a091d69a116 and squid was the only instance of this problem, so we should be safe now.

hnyman:
> probably quite a few bugs which are seeing light for the first time since lots of additions went into master which is seeing its first release in a while. This is a good thing

hnyman:
> probably quite a few bugs which are seeing light for the first time since lots of additions went into master which is seeing its first release in a while. This is a good thing

Well, actually the stable 24.10 branch is seeing the release, not master.  The stable 24.10 should not get any more speculative changes in the rc4 phase...  For example, this 22d4830 has been authored well after 24.10 branching. Even this kind of innocent looking change can cause unanticipated problems, like this case proves.

24.10 should now really be let to stabilise, and no new things should be backported to it before having matured for a while in master. Rebasing on master worked well for the first weeks after branching, before RC builds, but right now we should let 24.10 to get stabilised, and hope that no new mishaps from the already backported new stuff materializes.

PR #7515, opened 2024-12-29
luci-app-acme: render DNS API fields from the Structured Info 

stokito:
Maintainer: @tohojo

To show fields specific to a DNS API provider I made a list of all options and included them directly into the JS file.

In the upcoming ACME.sh release all scripts now contains a [Structured Info](https://github.com/acmesh-official/acme.sh/wiki/DNS-API-Structural-Info-description) with list of options:

```
dns_alviy
Alviy.com
Site: Alviy.com
Docs: github.com/acmesh-official/acme.sh/wiki/dnsapi2#dns_alviy
Options:
 Alviy_token API token. Get it from the https://cloud.alviy.com/token
Issues: github.com/acmesh-official/acme.sh/issues/5115
```

So once release we can make an RPC call to gather the info for all providers right from scripts and render it.

The PR adds the parsing and rendering part. Once the acmesh will be updated in the OpenWrt we can include the script and RPC call.
For now I just generated it and included as a file (slightly stripped to keep disk space).

We might use the file converted to a JSON but the info is really big and it's takes much less space and ram to have a JS parser right on UI.

CC @hgl @systemcrash @dannil @IamRPDev

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
@@ -3,6 +3,7 @@
 'require fs';
 'require uci';
 'require view';
+"require view.dnsapi as dnsapi";
 
 return view.extend({
 	load() {
@@ -16,11 +17,15 @@ return view.extend({
 				}
 				return certs;
 			}),
+			L.resolveDefault(fs.exec_direct('/usr/libexec/acmesh-dnsinfo.sh'), ''),
 		]);
 	},
 
 	render(data) {
 		let certs = data[0];
+		let dnsApiInfoText = data[1];
+		let apiInfos = dnsapi.parseFile(dnsApiInfoText);
+
 		let wikiUrl = 'https://github.com/acmesh-official/acme.sh/wiki/';
 		let wikiInstructionUrl = wikiUrl + 'dnsapi';
 		let m, s, o;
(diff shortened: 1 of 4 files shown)


tohojo:
Neat! Feels like the dnsapi.info.txt file should be part of the acme.sh package, not the luci app, though?

Also, AFAICT the parser is not terribly robust against malformed entries (judging from the unconditional shift() calls)? Not sure how the javascript runtime will handle failures, but IMO the parser should be smart enough to just skip malformed entries instead of aborting the whole thing (which is what I'm assuming will happen if the code throws an exception?).

systemcrash:
Use JSON, whose handling is far more well-defined. But no, a more opaque format, so everything can be done in (ba)sh. Doesn't really matter where the file is, repo-wise, but if the GUI is the only thing that depends on it, here is its home.

tohojo:
> The full parser with some tests is located here https://github.com/yurt-page/acmesh-parse-dnsapi-info
> 
> Right now we parsing own file that is 100% correct. Even now it will always parse to at least api Id and Name.
> 
> I'll try to check it later to improve error handling, maybe half of the parser can be replaced with a regexp :)

The main thing I'm after from a robustness perspective is that a malformed entry will just be skipped, but not interfere with the parsing of the rest of the file. Should be pretty straight forward to add a test for that :)

> In the PR I added a stripped info version 16 Kb. The full info version is 39 Kb and gzipped to 10 Kb. Converted to JSON 115 Kb and gzipped to 12 Kb (see parsed.json). The parser (dnsapi.js) is 3 KB but can be simplified to 1 Kb.
> 
> Maybe we can convert to JSON in shell but it's easier to do in JS, especially that we don't need for a full parsing (with Issues, Author fields).

I don't have a strong opinion about whether the file should generated as JSON at build time, or just be a concatenation of the definitions from the plugin files as you are doing now. But the "at build time" thing is important, IMO. I.e., the file should not be manually generated and included in the Luci package, it should be generated at build time of the acme-acmesh-dnsapi package and included in that package. Otherwise we have to manually keep the list of providers in sync between the two packages, and avoiding that is (presumably) the whole point of having an automatic parser in the first place :)

applications/luci-app-acme/htdocs/luci-static/resources/view/dnsapi.js
```diff
@@ -0,0 +1,110 @@
...
+	Default = '';
+}
+
+function parseFile(infoFileText) {
+	let infoFileLines = infoFileText.split('\n');
+	let infos = [];
+	let startIdx = 0;
+	for (let i = 1; i < infoFileLines.length; i++) {
+		if (infoFileLines[i] == '') {
+			if (i - startIdx > 2) {
```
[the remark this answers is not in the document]
tohojo: Ah, right, gotcha :)

tohojo:
> The PR can be merged today.

Hmm, but why do it in this order? Seems to me that getting the extraction bits in place first, and the merging this afterwards makes more sense? :)

tohojo:
Alright, but I still don't think the info.txt should be part of the GUI package. Transitioning it to the dnsapi package will be a pain later. We can add the manually generated one to acme-acmesh-dnsapi until upstream releases a version that we can support auto-generation from instead?

systemcrash:
ping @stokito

stokito:
The acmesh package was updated so now we can parse the DNS options right from scripts. I added a script to parse them and print to stdout. The script is called from the UI and it's output is parsed into array of API Info objects.
The list of API options is rendered as options that are shown only when the DNS API is selected.
The "Alt options" field will be always shown even for those APIs that doesn't have it. I have a version that can hide it but this makes code even more complicated.
I added Options Title that can describe difference between Opts and OptsAlt.

applications/luci-app-acme/root/usr/libexec/acmesh-dnsinfo.sh
```diff
@@ -0,0 +1,15 @@
...
+  dns_api="${filename%.*}"
+  echo "$dns_api"
+  dns_api_info_var="${dns_api}_info"
+  # shellcheck source=./dnsapi/dns_*.sh
+  . "$f"
+  info=""
+  eval info=\$$dns_api_info_var
+  echo "$info"
+  echo
+done
```
systemcrash: newline at end

systemcrash:
@tohojo has the final word, then.

stokito:
@tohojo friendly remind on this

applications/luci-app-acme/root/usr/libexec/acmesh-dnsinfo.sh
```diff
@@ -0,0 +1,15 @@
...
+do
+  filename=$(basename -- "$f")
+  dns_api="${filename%.*}"
+  echo "$dns_api"
+  dns_api_info_var="${dns_api}_info"
+  # shellcheck source=./dnsapi/dns_*.sh
+  . "$f"
+  info=""
+  eval info=\$$dns_api_info_var
+  echo "$info"
```
tohojo: Why not just `eval echo \"\$$dns_api_info_var\"` ?

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
```diff
@@ -16,11 +17,15 @@ return view.extend({
 				}
 				return certs;
 			}),
+			L.resolveDefault(fs.exec_direct('/usr/libexec/acmesh-dnsinfo.sh'), ''),
```
tohojo, line 20: This is not added to the ACL file?

applications/luci-app-acme/root/usr/libexec/acmesh-dnsinfo.sh
tohojo: ^^

applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
stokito: fixed

PR #7517, opened 2024-12-31
luci-app-ddns: Update global cacert path doc to reflect it is a file not a dir

kontaxis:
Description:
1. The UI (`Dynamic DNS > Global Settings`) describes `cacert` as the path for certificates (plural)
2. However, code in `packages/net/ddns-scripts/files/usr/bin/ddns.sh` expects the parameter to be a file path, i.e., a certificate bundle, not a directory
3. Therefore (1) is misleading and if someone supplies a directory, e.g., `/etc/ssl/certs`, DDNS services list updates silently fail

I would expect the following to succeed given how this parameter is documented:
```
# uci get ddns.global.cacert
/etc/ssl/certs/
# ls /etc/ssl/certs/
ca-certificates.crt
# /usr/bin/ddns service update
Certification file not found (/etc/ssl/certs/)
```

When a file path is used instead, DDNS service update succeeds:
```
# uci get ddns.global.cacert
/etc/ssl/certs/ca-certificates.crt
# /usr/bin/ddns service update
Downloading 'https://raw.githubusercontent.com/openwrt/packages/master/net/ddns-scripts/files/usr/share/ddns/list'
[...]
Download completed (825 bytes)
```

This change updates the language in `Dynamic DNS > Global Settings` to indicate that `cacert` is a file and not a directory.

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -440,8 +440,8 @@ return view.extend({
 
 		}
 
-		o = s.taboption('global', form.Value, 'cacert', _('CA Certs path'));
-		o.description = _('CA certificates path that will be used to download services data. Set IGNORE to skip certificate validation.');
+		o = s.taboption('global', form.Value, 'cacert', _('CA cert bundle file'));
+		o.description = _('CA certificate bundle file that will be used to download services data. Set IGNORE to skip certificate validation.');
 		o.placeholder = 'IGNORE';
 		o.write = function(section_id, value) {
 			if(value == 'ignore')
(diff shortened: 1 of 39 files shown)


applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
```diff
@@ -440,8 +440,8 @@ return view.extend({
 
 		}
 
-		o = s.taboption('global', form.Value, 'cacert', _('CA Certs path'));
-		o.description = _('CA certificates path that will be used to download services data. Set IGNORE to skip certificate validation.');
+		o = s.taboption('global', form.Value, 'cacert', _('CA bundle file path'));
+		o.description = _('File path of the CA bundle that will be used to download services data. Set IGNORE to skip certificate validation.');
```
systemcrash, line 444: In this case, let's avoid ambiguity of path, and refer directly to the file itself. Please re-word to
```
		o = s.taboption('global', form.Value, 'cacert', _('CA Certs bundle file'));
		o.description = _('CA bundle file used to download services data. Set IGNORE to skip certificate validation.');
```

Issue #7519, opened 2025-01-01
luci-base: Sorting of `GridSection` does not work


muink:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### screenshots or captures

![1](https://github.com/user-attachments/assets/95a17c61-b930-42c6-bf14-9f374c216d36)

### Actual behaviour

Click on any table title will not trigger sorting.

### Expected behaviour

Click on any table title to trigger sorting

### Steps to reproduce

Click on any table title.

### Additional Information

```shell
NAME="OpenWrt"
VERSION="24.10.0-rc4"
ID="OpenWrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt 24.10.0-rc4"
VERSION_ID="24.10.0-rc4"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r28211-d55754ce0d"
OPENWRT_BOARD="x86/64"
OPENWRT_ARCH="x86_64"
OPENWRT_TAINTS=""
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt 24.10.0-rc4 r28211-d55754ce0d"
OPENWRT_BUILD_DATE="1734915335"
```

### What browsers do you see the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Relevant log output

```shell
form.js?v=24.365.05279~f843014:215 Uncaught TypeError: Assignment to constant variable.
    at ClassConstructor.eval (form.js?v=24.365.05279~f843014:215:314)
    at NodeList.forEach (<anonymous>)
    at ClassConstructor.handleSort (form.js?v=24.365.05279~f843014:215:85)
    at ClassConstructor.eval (ui.js?v=24.365.05279~f843014:369:29)
eval	@	form.js?v=24.365.05279~f843014:215
handleSort	@	form.js?v=24.365.05279~f843014:215
eval	@	ui.js?v=24.365.05279~f843014:369
```

dannil:
See #7520

systemcrash:
[...]

PR #7520, opened 2025-01-01
luci-base: allow sort value to be reassigned

dannil:
Commit 234e1315073595771a7397d74a039d0eb4c1a89d changed the "val" variable to be a constant, but it's value is reassigned later. Change the variable to be declared with let instead so it can be reassigned.

- [X] Tested on: (x86/64, 24.10.0-rc4, Firefox) :white_check_mark:
- [X] \( Preferred ) Mention: @ the original code author for feedback @systemcrash 
- [X] \( Optional ) Closes: e.g. openwrt/luci#7519

Before: sorting on zone forwardings fails (applies to all GridSections)
Screenshot_18

After: sorting on zone forwardings works correctly (applies to all GridSections)
Screenshot_19
Screenshot_20


systemcrash:
Gratefully accepted!

PR #7524, opened 2025-01-03
luci-mod-network: The expand-hosts seems to also affect the AAAA DNS records.

adelton:

I've spent quite some time trying to figure out why
```
dig @192.168.1.1 -t a laptop.lan
```
returns the A record and
```
dig @192.168.1.1 -t aaaa laptop.lan
```
does not return the AAAA record, even if I saw the IPv6 address assigned to the host and listed in the Active DHCPv6 Leases.

I finally found out that I have the `Expand hosts` unchecked (because I prefer to use the full hostnames) and that it does not affect the A records but does affect the AAAA records. It simply did not occur to me to just check
```
dig @192.168.1.1 -t aaaa laptop
```
and I thought something about AAAA records is broken.

If there was a note in the UI mentioning the relationship of the `Expand hosts` to the IPv6 addresses and AAAA records, it would have saved me nontrivial amount of time. Adding a note to help others who might find themselves in the same situation.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -430,7 +430,7 @@ return view.extend({
 
 		s.taboption('general', form.Flag, 'expandhosts',
 			_('Expand hosts'),
-			_('Add local domain suffix to names served from hosts files.'));
+			_('Add local domain suffix to names served from hosts files; also affects names for IPv6 addresses (AAAA)..'));
 
 		o = s.taboption('logging', form.Flag, 'logqueries',
 			_('Log queries'),


systemcrash:
Some nuance like this is worth investigating. A problem more likely lies in how dnsmasq works. The setting you mention is a plausible offender. But your amendment invalidates an existing translation. It's better to add a new follow-on string. 

E.g.

```js
... + ' ' +_('your string);
```

Because a host has IPs via DHCP doesn't mean a dns lookup will automatically work. You likely need to assign it a static IP under host names in DHCP.

adelton:
Thanks for the hint about preserving the existing translations, updated in 6b1c47c4f3d0dabee68c1e96186d954da831968c.

I specifically kept the wording a little bit vague WRT to when the AAAA records get actually created. But on a stock OpenWrt 23.05.5 (r24106-10cc5fcd00) I see
```
$ for i in openwrt openwrt.lan laptop laptop.lan ; do for t in a aaaa ; do echo "$i $t: $( dig +short @192.168.1.1 -t $t $i )" ; done ; done
openwrt a: 192.168.1.1
openwrt aaaa: fddd:7eab:3d2a::1
openwrt.lan a: 192.168.1.1
openwrt.lan aaaa: fddd:7eab:3d2a::1
laptop a: 192.168.1.142
laptop aaaa: fddd:7eab:3d2a::bb3
laptop.lan a: 192.168.1.142
laptop.lan aaaa: fddd:7eab:3d2a::bb3
```
and when I uncheck the `Expand hosts` and `Save & Apply`, I see
```
$ for i in openwrt openwrt.lan laptop laptop.lan ; do for t in a aaaa ; do echo "$i $t: $( dig +short @192.168.1.1 -t $t $i )" ; done ; done
openwrt a: 192.168.1.1
openwrt aaaa: fddd:7eab:3d2a::1
openwrt.lan a: 
openwrt.lan aaaa: 
laptop a: 192.168.1.142
laptop aaaa: fddd:7eab:3d2a::bb3
laptop.lan a: 192.168.1.142
laptop.lan aaaa: 
```

systemcrash:
Can you show me the content of your `/etc/hosts` file, and your `/etc/config/dhcp` file? (redact or mutate anything considered sensitive)

adelton:
The vanilla content after resetting to defaults is

```
root@OpenWrt:~# cat /etc/hosts
127.0.0.1 localhost

::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
root@OpenWrt:~# cat /etc/config/dhcp

config dnsmasq
	option domainneeded '1'
	option boguspriv '1'
	option filterwin2k '0'
	option localise_queries '1'
	option rebind_protection '1'
	option rebind_localhost '1'
	option local '/lan/'
	option domain 'lan'
	option expandhosts '1'
	option nonegcache '0'
	option cachesize '1000'
	option authoritative '1'
	option readethers '1'
	option leasefile '/tmp/dhcp.leases'
	option resolvfile '/tmp/resolv.conf.d/resolv.conf.auto'
	option nonwildcard '1'
	option localservice '1'
	option ednspacket_max '1232'
	option filter_aaaa '0'
	option filter_a '0'

config dhcp 'lan'
	option interface 'lan'
	option start '100'
	option limit '150'
	option leasetime '12h'
	option dhcpv4 'server'
	option dhcpv6 'server'
	option ra 'server'
	option ra_slaac '1'
	list ra_flags 'managed-config'
	list ra_flags 'other-config'

config dhcp 'wan'
	option interface 'wan'
	option ignore '1'

config odhcpd 'odhcpd'
	option maindhcp '0'
	option leasefile '/tmp/hosts/odhcpd'
	option leasetrigger '/usr/sbin/odhcpd-update'
	option loglevel '4'

```

When I uncheck that `Expand hosts` and `Save & Apply`, the `/etc/hosts` is the same and the diff in `/etc/config/dhcp` is
```diff
--- dhcp.orig	2025-01-06 08:23:26.949500182 +0100
+++ dhcp	2025-01-06 08:23:07.720432382 +0100
@@ -1,25 +1,18 @@
 
 config dnsmasq
 	option domainneeded '1'
-	option boguspriv '1'
-	option filterwin2k '0'
 	option localise_queries '1'
 	option rebind_protection '1'
 	option rebind_localhost '1'
 	option local '/lan/'
 	option domain 'lan'
-	option expandhosts '1'
-	option nonegcache '0'
 	option cachesize '1000'
 	option authoritative '1'
 	option readethers '1'
 	option leasefile '/tmp/dhcp.leases'
 	option resolvfile '/tmp/resolv.conf.d/resolv.conf.auto'
-	option nonwildcard '1'
 	option localservice '1'
 	option ednspacket_max '1232'
-	option filter_aaaa '0'
-	option filter_a '0'
 
 config dhcp 'lan'
 	option interface 'lan'
```

systemcrash:
It's possible your tests may have hit cached entries. Also, the dnsmasq manpage has this to say about host-records (`Unlike hosts-files, names are not expanded, even when --expand-hosts is in effect`):


`--host-record=<name>[,<name>....],[<IPv4-address>],[<IPv6-address>][,<TTL>]`
    Add A, AAAA and PTR records to the DNS. This adds one or more names to the DNS with associated IPv4 (A) and IPv6 (AAAA) records. A name may appear in more than one --host-record and therefore be assigned more than one address. Only the first address creates a PTR record linking the address to the name. This is the same rule as is used reading hosts-files. --host-record options are considered to be read before host-files, so a name appearing there inhibits PTR-record creation if it appears in hosts-file also. Unlike hosts-files, names are not expanded, even when --expand-hosts is in effect. Short and long names may appear in the same --host-record, eg. --host-record=laptop,laptop.thekelleys.org,192.168.0.1,1234::100 


These are defined by `hostrecord` types in the dhcp config. They're not in the GUI yet. 

Compare with `host` entries, which correspond to `--dhcp-host` and do receive the suffix.


So the question is: where do your entries come from? e.g. `laptop` above?

adelton:
I see `laptop` mentioned in
```
/var/dhcp.leases:1727138289 90:2e:16:aa:bb:cc 192.168.1.142 laptop 01:90:2e:16:aa:bb:cc
/var/hosts/odhcpd:fd7f:422f:4895::bb3	laptop
/var/hosts/odhcpd:# br-lan 0004804054c45add8a5b5fdf7aac01234567 667b932a laptop 1727138286 bb3 128 fd7f:422f:4895::bb3/128
```

These records do not change when I toggle `Expand hosts` back and forth (and `Save & Apply`), while the `dig` output changes. (The IPv6 addresses changed since the first time I showed the `dig` outputs above because I've reset the router to defaults which reset ULA prefix.)

systemcrash:
Then this seems related not to 'dns' entries per sé, but DHCP entries. Wording here is important, if we are going to put a message that users will read. In which case, let's revise it to:

`Also affects DHCP-derived names` 

as stated in the dnsmasq manpage. 

There's a big but here: there might be undefined behaviour or a bug in dnsmasq regarding this. I don't see why AAAA should be absent when this is changed, and A get treated.

systemcrash:
Actually, what is the content of your `/etc/resolv.conf` in both cases? Since the lines in the `odhcpd` files are controlled not by dnsmasq but `odhcpd` which uses

https://linux.die.net/man/3/dn_expand

adelton:
> Then this seems related not to 'dns' entries per sé, but DHCP entries. Wording here is important, if we are going to put a message that users will read. In which case, let's revise it to:
> 
> `Also affects DHCP(v6)-derived names`
> 
> as stated in the dnsmasq manpage.

Updated, rebased on master.

> There's a big but here: there might be undefined behaviour or a bug in dnsmasq regarding this. I don't see why AAAA should be absent when this is changed, and A get treated.

I don't disagree. :-)

> Actually, what is the content of your /etc/resolv.conf in both cases? Since the lines in the odhcpd files are controlled not by dnsmasq but odhcpd which uses https://linux.die.net/man/3/dn_expand

The content of the `/etc/resolv.conf` is
```
search lan
nameserver 127.0.0.1
nameserver ::1
```
and the `Expand hosts` checkbox does not seem to have any effect on that content.

systemcrash:
> The content of the `/etc/resolv.conf` is
> 
> ```
> search lan
> nameserver 127.0.0.1
> nameserver ::1
> ```
> 
> and the `Expand hosts` checkbox does not seem to have any effect on that content.

Is it possible that your resolv file gets clobbered after changing the setting? That `search lan` bit might be interpreted differently by various daemons.

adelton:
Can you elaborate on that "clobbered"? (Non-native speaker here.)

systemcrash:
Replaced with something completely different?

adelton:
Ah. No, the content of `/etc/resolv.conf` does not seem to get affected at all by changing the `expandhosts` value:

```
root@OpenWrt:~# grep expand /etc/config/dhcp 
	option expandhosts '1'
root@OpenWrt:~# cat /etc/resolv.conf 
search lan
nameserver 127.0.0.1
nameserver ::1
root@OpenWrt:~# sha256sum /etc/resolv.conf
7d6806759821bb1b18ad3ce81edec5f082c2c0dda30f17907016591fb4d6e84c  /etc/resolv.conf

root@OpenWrt:~# uci del dhcp.cfg01411c.expandhosts
root@OpenWrt:~# uci commit
root@OpenWrt:~# grep expand /etc/config/dhcp 
root@OpenWrt:~# sha256sum /etc/resolv.conf
7d6806759821bb1b18ad3ce81edec5f082c2c0dda30f17907016591fb4d6e84c  /etc/resolv.conf

root@OpenWrt:~# uci set dhcp.cfg01411c.expandhosts=0
root@OpenWrt:~# uci commit
root@OpenWrt:~# grep expand /etc/config/dhcp 
	option expandhosts '0'
root@OpenWrt:~# sha256sum /etc/resolv.conf
7d6806759821bb1b18ad3ce81edec5f082c2c0dda30f17907016591fb4d6e84c  /etc/resolv.conf
root@OpenWrt:~# 
```

This was on a freshly installed OpenWrt 24.10.0-rc5 (r28304-6dacba30a7) x86_64 VM.

Issue #1324, opened 2017-08-22
No Associated Stations displayed when using dynamic VLANs


Mattyx:
Hello,
I'm using SSIDs with WPA2 802.1X and dynamic VLANs.
For those SSIDs the "Associated Stations" section does not show any station.

I use 3 VLANs (with ID 18, 19 and 20).

I think this is due to the fact that these stations are associated to different devices than wlan0 or wlan1.
Using iw (or iwinfo) command I can see stations under the wlan0.20, wlan1.20, etc. devices. The parent device (wlan0 or wlan1) on this specific configuration does not have any station directly associated.

Maybe LuCI doesn't aggregate the additional VLAN interfaces under the "parent" interface?

There is a way to correct this?

Thanks in advance!

xsokolikx:
I'm very interested on this. Any solution to show the associated clients on a wireless dynamic vlan?
Thanks

ASoTNetworks:
I am also interested in seeing this fixed so we don't have to go into the command line to see connected clients on each VLAN interface.

gtxaspec:
2023 and still issue still persists.  clients connected to AP/VLAN do not show up on the associated stations list.

timfoley66:
Experiencing same issue.

spamtrap07:
I observe the same issue on openwrt 23.05.0-rc2.

hizukiayaka:
It is a history problem since the old openwrt time,
https://forum.archive.openwrt.org/viewtopic.php?id=71846

systemcrash:
I think the first reasonable steps to fix this including augmenting (read: patch) the iw command to aggregate station info irrespective of interface, if it doesn't already. Armed with this it should be trivial to figure out their vlans and which ap to roughly display them correctly.

What kind of information does iw have available?

tobiaswaldvogel:
Actually the code seems to be VLAN aware already:

[https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/network.js](url)

```
	getAssocList: function() {
		var tasks = [];
		var ifnames = [ this.getIfname() ].concat(this.getVlanIfnames());

		for (var i = 0; i < ifnames.length; i++)
			tasks.push(callIwinfoAssoclist(ifnames[i]));

		return Promise.all(tasks).then(function(values) {
			return Array.prototype.concat.apply([], values);
		});
	},
```

[...]

Issue #1234, opened 2017-07-13
wifi.getiwinfo returns empty result when called from JsonRpc


alexandertmills:
I'm trying to access information about wlan0 though an HTTP POST request to JsonRpc. After successfully authenticating, I made the following request to` [device]/cgi-bin/luci/rpc/sys`
```
{
	"method" : "wifi.getiwinfo",
	"params" : [ "wlan0" ]
}
```
…which resulted in a 200 code and this response:
```
{
	"id": null,
	"result": [],
	"error": null
}
```
I've tried this with user.checkpasswd and a number of other exported libraries which produced expected results and no issues, but wifi.getiwinfo always returns an empty list. What might the issue be?

antscode:
I found this blog post about the issue (in Chinese, use translate):

https://blog.csdn.net/jt6562/article/details/21399897

The solution provided by the author worked for me. I SSH'd into my OpenWRT box and modified the following file with the code provided in the article above:

/usr/lib/lua/luci/sys.lua

After a reboot, you can call the alternative **wifi.getiwinfo_rpc** method via json rpc.

jow-:
The iwinfo RPC binding has been reworked with 055b621cf5cdc15b24e78408d610a00f6869b8e0 - check the commit message of this commit for further details.

antscode:
Thank you!

PR #7529, opened 2025-01-05
luci: show wifi vlan interface instead of base interface

tobiaswaldvogel:
- [x] Tested on: aarch, mips, mipsel, openwrt version SNAPSHOT, r28432+19-4b6eb631e1, browser Chrome :white_check_mark:
- [x] Description: This patch adds a badge to the associated wireless stations with the vlan id and name. It is displayed in the same color as the network, to which it is bridged, so the color corresponds to the color in the network configuration
page.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/network.js
+++ b/modules/luci-base/htdocs/luci-static/resources/network.js
@@ -639,7 +639,7 @@ function enumerateNetworks() {
 }
 
 
-var Hosts, Network, Protocol, Device, WifiDevice, WifiNetwork;
+var Hosts, Network, Protocol, Device, WifiDevice, WifiNetwork, WifiVlan;
 
 /**
  * @class network
@@ -4150,14 +4150,40 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */ 
 	 */
 	getAssocList: function() {
 		var tasks = [];
-		var ifnames = [ this.getIfname() ].concat(this.getVlanIfnames());
+		var station;
 
-		for (var i = 0; i < ifnames.length; i++)
-			tasks.push(callIwinfoAssoclist(ifnames[i]));
+		for (let vlan of this.getVlans())
+			tasks.push(callIwinfoAssoclist(vlan.getIfname()).then(
+				function(stations) {
+					for (station of stations)
+						station.vlan = vlan;
+
+					return stations;
+				})
+			);
+
+		tasks.push(callIwinfoAssoclist(this.getIfname()));
 
 		return Promise.all(tasks).then(function(values) {
 			return Array.prototype.concat.apply([], values);
 		});
+	},
+
+	/**
+	 * Fetch the vlans for this network.
+	 *
+	 * @returns {Array<LuCI.network.WifiVlan>}
+	 * Returns an array of vlans for this network.
+	 */
+	getVlans: function() {
+		var vlans = [];
+		var vlans_ubus = this.ubus('net', 'vlans');
+
+		if (vlans_ubus)
+			for (let vlan of vlans_ubus)
+				vlans.push(new WifiVlan(vlan));
+
+		return vlans;
 	},
 
 	/**
@@ -4434,6 +4460,79 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */ 
 			params: [ 'addr', 'deauth', 'reason', 'ban_time' ]
 		})(mac, deauth, reason, ban_time);
 	}
+});
+
+/**
+ * @class
+ * @memberof LuCI.network
+ * @hideconstructor
+ * @classdesc
+ *
+ * A `Network.WifiVlan` class instance represents a vlan on a WifiNetwork.
+ */
+WifiVlan = baseclass.extend(/** @lends LuCI.network.WifiVlan.prototype */ {
+	__init__: function(vlan) {
+		this.ifname = vlan.ifname;
+		if (L.isObject(vlan.config)) {
+			this.vid = vlan.config.vid;
+			this.name = vlan.config.name;
+
+			if (Array.isArray(vlan.config.network) && vlan.config.network.length)
+				this.network = vlan.config.network[0];
+		}
+	},
+
+	/**
+	 * Get the name of the wifi vlan.
+	 *
+	 * @returns {string}
+	 * Returns the name.
+	 */
+	getName: function() {
+		return this.name;
+	},
+
+	/**
+	 * Get the vlan id of the wifi vlan.
+	 *
+	 * @returns {number}
+	 * Returns the vlan id.
+	 */
+	getVlanId: function() {
+		return this.vid;
+	},
+
+	/**
+	 * Get the network of the wifi vlan.
+	 *
+	 * @returns {string}
+	 * Returns the network.
+	 */
+	getNetwork: function() {
+		return this.network;
+	},
+
+	/**
+	 * Get the Linux network device name of the wifi vlan.
+	 *
+	 * @returns {string}
+	 * Returns the current network device name for this wifi vlan.
+	 */
+	getIfname: function() {
+		return this.ifname;
+	},
+
+	/**
+	 * Get a long description string for the wifi vlan.
+	 *
+	 * @returns {string}
+	 * Returns a string containing the vlan id and the vlan name,
+	 * if it is different than the vlan id
+	 */
+	getI18n: function() {
+		var name =  this.name && this.name != this.vid ? ' (' + this.name + ')' : '';
+		return 'vlan %d%s'.format(this.vid, name);
+	},
 });
 
 return Network;
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -765,6 +765,25 @@ return view.extend({
 				])
 			];
 
+			var zones = data[4];
+			if (bss.vlan) {
+				var desc = bss.vlan.getI18n();
+				var vlan_network = bss.vlan.getNetwork();
+				var vlan_zone;
+
+				if (vlan_network && zones)
+					for (let zone of zones)
+						if (zone.getNetworks().includes(vlan_network))
+							vlan_zone = zone;
+
+				row[0].insertBefore(
+					E('div', {
+						'class' : 'zonebadge',
+						'title' : desc,
+						'style' : firewall.getZoneColorStyle(vlan_zone)
+					}, [ desc ]), row[0].firstChild);
+			}
+
 			if (bss.network.isClientDisconnectSupported()) {
 				if (table.firstElementChild.childNodes.length < 6)
 					table.firstElementChild.appendChild(E('th', { 'class': 'th cbi-section-actions'}));
@@ -803,7 +822,8 @@ return view.extend({
 		return Promise.all([
 			uci.changes(),
 			uci.load('wireless'),
-			uci.load('system')
+			uci.load('system'),
+			firewall.getZones(),
 		]);
 	},
 
@@ -823,11 +843,11 @@ return view.extend({
 		params: [ 'config', 'section', 'name' ]
 	}),
 
-	render: function() {
+	render: function(data) {
 		if (this.checkAnonymousSections())
 			return this.renderMigration();
 		else
-			return this.renderOverview();
+			return this.renderOverview(data[3]);
 	},
 
 	handleMigration: function(ev) {
@@ -862,7 +882,7 @@ return view.extend({
 		]);
 	},
 
-	renderOverview: function() {
+	renderOverview: function(zones) {
 		var m, s, o;
 
 		m = new form.Map('wireless');
@@ -2404,6 +2424,10 @@ return view.extend({
 							return hosts_radios_wifis;
 						});
 					}, network))
+					.then(L.bind(function(zones, data) {
+						data.push(zones);
+						return data;
+					}, network, zones))
 					.then(L.bind(this.poll_status, this, nodes));
 			}, this), 5);
 
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
@@ -5,6 +5,7 @@
 'require uci';
 'require fs';
 'require rpc';
+'require firewall';
 
 return baseclass.extend({
 	title: _('Wireless'),
@@ -184,6 +185,7 @@ return baseclass.extend({
 			network.getHostHints(),
 			this.callSessionAccess('access-group', 'luci-mod-status-index-wifi', 'read'),
 			this.callSessionAccess('access-group', 'luci-mod-status-index-wifi', 'write'),
+			firewall.getZones(),
 			L.hasSystemFeature('wifi') ? L.resolveDefault(uci.load('wireless')) : L.resolveDefault(),
 		]).then(L.bind(function(data) {
 			var tasks = [],
@@ -216,7 +218,8 @@ return baseclass.extend({
 		    networks = data[1],
 		    hosthints = data[2],
 		    hasReadPermission = data[3],
-		    hasWritePermission = data[4];
+		    hasWritePermission = data[4],
+		    zones = data[5];
 
 		var table = E('div', { 'class': 'network-status-table' });
 
@@ -325,6 +328,24 @@ return baseclass.extend({
 						E('span', this.wifirate(bss.tx))
 					])
 				];
+
+				if (bss.vlan) {
+					var desc = bss.vlan.getI18n();
+					var vlan_network = bss.vlan.getNetwork();
+					var vlan_zone;
+
+					if (vlan_network)
+						for (let zone of zones)
+							if (zone.getNetworks().includes(vlan_network))
+								vlan_zone = zone;
+
+					row[0].insertBefore(
+						E('div', {
+							'class' : 'zonebadge',
+							'title' : desc,
+							'style' : firewall.getZoneColorStyle(vlan_zone)
+						}, [ desc ]), row[0].firstChild);
+				}
 
 				if (networks[i].isClientDisconnectSupported() && hasWritePermission) {
 					if (assoclist.firstElementChild.childNodes.length < 6)


systemcrash:
Q: is this 'instead of' or 'as well as'? And can you show me an example of this in use?

I want to know what to look for to verify that existing behaviour is not affected for those not using dynamic VLANS. 

This regards #1324, correct?

Also, please amend your commit message with 79 char line-wrap to include the description, after you've fixed the strange bit: `When using wifi the vlans the base wifi interface in the network object is show`. Did you mean:

`When using wifi with dynamic vlans, the base wifi interface in the network object is shown`?

tobiaswaldvogel:
I adjusted the wording as you suggested. Yes, it is partly related to [#1234](https://github.com/openwrt/luci/issues/1324). The issue was actually that clients connected to wifi interfaces with dynamic vlans don't show up in the status and the wireless page. **This is not addressed with this patch**. 

For detecting the VLANs Luci relies on the output of `ubus call network.wireless` and the array `vlans` inside the `interface` objects. hostapd does not populate them for dynamic vlans hence Luci does not see them. This would require some changes in hostapd. I had a look at this, but as there is a very complex interaction between netifd and hostapd, so this is not easy to address.

An alternative for the dynamic vlans is using `wifi-vlan` sections instead. They populate VLAN array correctly and and the clients show up.in Luci. However you have to configure the tagged network interfaces, bridges and luci interfaces yourself.
The only catch is, that the base interface is shown instead of the VLAN name. This is what this patch is about.

Here is an example (not using dynamic vlans):

`/etc/config/wireless`
```
config wifi-vlan
        option vid '3'
        option network 'lan'
        option name 'lan'

config wifi-vlan
        option vid '12'
        option network 'guest'
        option name 'gst'

config wifi-iface 'wlan2'
        option ifname 'wlan2'
        option device 'radio_2g'
        option mode 'ap'
        option ssid '2'
        option wpa_psk_file '/etc/config/wireless.wpa_psk'
...
```
This will create the base interface `wlan2`, which is not connected to any network, `wlan2-lan` connected to lan and `wlan2-gst` connect to gst..

Without this patch all clients will be reported with **`2 (wlan2)`**. 
After this patch clients connected to vlan 3 will show up as **`2 (wlan-lan)`** and clients connected to vlan 12 will show up as **`2 (wlan-gst)`**

Configuration of lan and gst in `/etc/config/network`:
```
...
config device
        option type '8021q'
        option name 'eth1-lan'
        option ifname 'eth1'
        option vid '3'

config device
        option type '8021q'
        option name 'eth1-gst'
        option ifname 'eth1'
        option vid '12'
...
config device
        option type 'bridge'
        option name 'br-lan'
        list ports 'eth1-lan'

config device
        option type 'bridge'
        option name 'br-gst'
        list ports 'eth1-gst'
...
```

systemcrash:
Can you show me an example of this in use? A screenshot of the changes.

tobiaswaldvogel:
Before
image

After:
image

tobiaswaldvogel:
Improved description wording and added to the commit

systemcrash:
In this case, why not simply add a VLAN column? Since it's not clear from reading `wlan2-iot` what iot is, it's not denoted anywhere. VLAN Interface here is a misnomer since it seems that no VLAN is revealed, but the underlying 'network'?

You can pack a couple or triplet in the VLAN column, similar to the Sig/Noise column. e.g. header has VLAN: VID, Name, <something else>, and each cell has:
vid
name
<something else>

tobiaswaldvogel:
The interface name is actually the name which was chosen in the wifi-vlan section. so if you prefer the vlan id you can change the option `name` to '3'

```
config wifi-vlan
        option vid '3'
        option network 'lan'
        option name '3'
```

That would create a vlan interface with the suffix -3 instead -lan in my previous example and show up e.g. as wlan2-3

Adding another column might be a little bit tricky as that might break the rendering in mobile devices. I can do a test though.
Right this looks like:

image

systemcrash:
The old notation for VLANs in at least ifconfig and some config notations is `.<vlan>`. But this may be confusing in the wireless domain where VLANs are not possible. So if it really *must* go in the wireless info, it should be: `VLAN: <vlan>`. But let's see if we can avoid that.

tobiaswaldvogel:
So adding a column is not an option at all because that require touching the stylesheets of all themes. There are several rules for displays smaller than 800px or mobile devices to automatically hide some columns like the mac address, and they are based on the column number.

```
	.assoclist tr > *:nth-of-type(2) {
		display: none;
	}
```

Trying to change the layout of the assoclist will certainly break the rendering on some devices/themes and would required thousands of regression tests.

The text itself in the badge comes also via the stylesheet:
```
.assoclist td > .ifacebadge[data-ssid][data-ifname] > span {
	display: none;
}

.assoclist td > .ifacebadge[data-ssid][data-ifname]::after {
	content: attr(data-ssid) " (" attr(data-ifname) ")";
}
```

So it is  ```ssd ( ifname )```

Without changing the stylesheets of all themes we can modify only the values for ssid and ifname.
So something like  ```2 ( wlan2, vlan 3 )``` would be possible. But it makes it unnecessarily long and might be easily truncated with longer ssid names.

So from my point of showing the vlan interface is still the best option. As said before you can name them as you like e.g. just the VLAN number, that would result in ```<interface>-<vlan> ```

tobiaswaldvogel:
I did one more experiment using y zone badge element, which seems to render ok in all scenarios.
The name `(iot)` will only be added if `name` and `vid` in the wifi-vlan section is different

image

The color is the same as the color of the network interface of the option `network` in the wifi vlan section:

image

What do you think?

systemcrash:
@hnyman @dannil @stokito any thoughts on this?

Would also be good to have some tests with those who get auto-vlan assignment (via RADIUS?)

tobiaswaldvogel:
> Would also be good to have some tests with those who get auto-vlan assignment (via RADIUS?)  

That would be great. I understand that in this scenario the vlans would also be defined via wifi-vlan sections and just the assignment is performed by the radius server instead of the wpa_psk file. So I would expect this work as well without any issues.

The only limitation for now is dynamic_vlans, as the code does not update the ubus network,wireless objects with the vlans.

modules/luci-base/htdocs/luci-static/resources/network.js
```diff
@@ -4436,4 +4460,78 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */
...
 
+/**
+ * @class
+ * @memberof LuCI.network
+ * @hideconstructor
+ * @classdesc
+ *
+ * A `Network.WifiVlan` class instance represents a vlan on a WifiNetwork.
+ */
+WifiVlan = baseclass.extend(/** @lends LuCI.network.WifiVlan.prototype */ {
```
systemcrash, line 4471: Perhaps it can just be `Vlan`?

tobiaswaldvogel: I think that might be mistaken for an ethernet vlan interface, which is quite different.

systemcrash: Ofc. You're right. It's in `wireless`.

systemcrash:
Please don't include merges: `Merge branch 'master' into luci_wifi_dynamic_vlan` but rebase.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -2343,6 +2363,8 @@ return view.extend({
 				return render_network_status(inst);
 		};
 
+		m.zones = zones;
```
dannil, line 2366: Is this assignment needed? It it is, you should solve it in another way; correct me if I'm wrong but it looks like you're manipulating the prototype for [LuCi.form.Map](https://openwrt.github.io/luci/jsapi/LuCI.form.Map.html) and adding a new field which is a big no-no.

modules/luci-base/htdocs/luci-static/resources/network.js
```diff
@@ -4436,4 +4460,78 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */
...
+	__init__: function(vlan) {
+		this.ifname = vlan.ifname;
+		if (L.isObject(vlan.config)) {
+			this.vid = vlan.config.vid;
+			this.name = vlan.config.name;
+
+			if (Array.isArray(vlan.config.network))
+				if (vlan.config.network.length)
+					this.network = vlan.config.network[0];
```
dannil, line 4480: Feel free to collapse this to one if-statement so we avoid unnecessary nesting.

dannil (commented):
Overall looks good, haven't run-tested it myself tho since I don't use VLAN's in my setup. You've mixed and matched `var` and `let` in the implemented code, for example in `getVlans`, feel free to use ES6-syntax throughout if you'd like.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
tobiaswaldvogel: Good point. I'll rework it

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
```diff
@@ -185,6 +186,7 @@ return baseclass.extend({
 			this.callSessionAccess('access-group', 'luci-mod-status-index-wifi', 'read'),
 			this.callSessionAccess('access-group', 'luci-mod-status-index-wifi', 'write'),
 			L.hasSystemFeature('wifi') ? L.resolveDefault(uci.load('wireless')) : L.resolveDefault(),
+			firewall.getZones(),
```
systemcrash, line 189: I think it's OK to put this above `L.has...` so you can keep data[x] contiguous.

modules/luci-base/htdocs/luci-static/resources/network.js
tobiaswaldvogel: Makes absolutely sense

modules/luci-base/htdocs/luci-static/resources/network.js
```diff
@@ -4436,4 +4462,77 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */
...
+ */
+WifiVlan = baseclass.extend(/** @lends LuCI.network.WifiVlan.prototype */ {
+	__init__: function(vlan) {
+		this.ifname = vlan.ifname;
+		if (L.isObject(vlan.config)) {
+			this.vid = vlan.config.vid;
+			this.name = vlan.config.name;
+
+			if (Array.isArray(vlan.config.network) && vlan.config.network.length)
+					this.network = vlan.config.network[0];
```
dannil, line 4481: One tab indentation too many.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
```diff
@@ -326,6 +329,24 @@ return baseclass.extend({
...
+						for (let zone of zones)
+							if (zone.getNetworks().includes(vlan_network))
+								vlan_zone = zone;
+
+					row[0].insertBefore(
+						E('div', {
+							'class' : 'zonebadge',
+							'title' : desc,
+							'style' : firewall.getZoneColorStyle(vlan_zone)
+						}, [ desc ]), row[0].firstChild);
```
dannil, line 347: What happens if the code ends up at `firewall.getZoneColorStyle(vlan_zone)` and `vlan_zone` is still null? This code allows for that and it looks like `getZoneColorStyle` would crash with a null pointer in that case.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -765,6 +765,25 @@ return view.extend({
...
+					for (let zone of zones)
+						if (zone.getNetworks().includes(vlan_network))
+							vlan_zone = zone;
+
+				row[0].insertBefore(
+					E('div', {
+						'class' : 'zonebadge',
+						'title' : desc,
+						'style' : firewall.getZoneColorStyle(vlan_zone)
+					}, [ desc ]), row[0].firstChild);
```
dannil, line 784: See previous comment about null pointer.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js
tobiaswaldvogel: If getZoneColorStyle cannot find the zone it returns a style with #eeeeee (light gray), which is the intended behavior (same as in the interface page if there is no firewall assigned to an interface)

```
	getZoneColorStyle: function(zone) {
		var hex = (zone instanceof Zone) ? zone.getColor() : getColorForName((zone != null && zone != '*') ? zone : null);
...
```

function getColorForName(forName) {
	if (forName == null)
		return '#eeeeee';
...
```

File: modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js (hunk 7 above)
Comment (tobiaswaldvogel) [reply modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js:784]: See previous reply

File: modules/luci-base/htdocs/luci-static/resources/network.js (hunk 5 above)
Comment (tobiaswaldvogel) [reply modules/luci-base/htdocs/luci-static/resources/network.js:4481]: Agreed

File: modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js (hunk 6 above)
Comment (dannil) [reply modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/60_wifi.js:347]: Nice, I misread the code and placed the `null` outside the nested parenthesis, then it's fine.

**Comment by systemcrash** [comment]
Hi @tobiaswaldvogel are you satisfied with the current state? If you're active in the forum, it would be nice to get this tested by dynamic vlan users. Those kinds of threads appear from time to time.

**Comment by tobiaswaldvogel** [comment]
Yes, I'm happy with the implementation now. Thanks for all the comments.
Lately I didn't use the forum that much but I'll look for a thread with this topic.
But I also don't mind setting up a test radius server. I also considered that before, in order to avoid synching the wpa_psk_file, but at the end I didn't want to introduce another dependency. But if I don't find anybody I'll prepare a test this weekend.

**Comment by tobiaswaldvogel** [comment]
Finally I managed to complete the radius test. The setup was a little bit more complicated than anticipated, although I worked already with with freeradius before.

Short answer:  
Yes, it works when using wifi-vlan definitions, same as without radius.

Long answer:  
Wifi VLANs require a vlan defintion file, which is set with hostapd option ```vlan_file```. In OpenWRT you have two options how to deal with it: 

- wifi-vlan sections in /etc/config/wireless
  In this case hostapd.sh creates a file ```/var/run/hostapd-<interface-name>.vlan``` and points ```vlan_file``` to it. ```vlan_no_bridge`` will be set to 1 and netifd takes care of adding and removing the wifi vlan interfaces to the corresponding bridge. That will also populate the ubus vlan arrays, which are used in this PR

- Manually  created ```vlan_file``` + ```vlan_bridge```
  In this case it is hostapd taking care of adding the interface to the bridge and even of creating the bridge itself if the option ```dynamic_vlan``` is set. That will not populate the ubus array hence the mapping is not visible and stations connected to a wifi vlan will not show up.  
  I evaluated the following options:  
  Change the the hostapd.sh script to scan the file and add it to ubus as well, but in this case netifd would interfere, which is not what we want.
  Parsing the vlan file itself. That would be cumbersome as well, as reading it directly is not option. It would need to be declared in the rpc acl, which is not possible as it can be place anywhere. Writing a backend extension just to read it is not worth the effort.

So the conclusion is: Use wifi-vlan sections, which has also the advantage to have to whole config in the /etc/config/* files

Another comment on ```vlan_tagged_interface```: With this option you can create the bridges and 802.11q interfaces automatically, which might save a lot of manual bridge and interface definitions, but the downside is that wifi down also tears down the lan bridge, which breaks the connection to the device itself. This is why I stopped using it.

A few more words on my radius test setup.:  
As none of my IOT devices support wpa enterprise I opted for using ppsk (private PSK = PSKs provided from radius). For now the hostapd.sh script evaluates the option ppsk. Setting it results in the hostapd options ```wpa_psk_radius=2``` and ```macaddr_acl=2```

Right now there is a bug in hostapd.sh, which leads adding the radius server twice. Furthermore ```wpa_psk_radius=3``` would be the preferred option. I created a fix and introduced ppsk=2 as new option for mode 3, which is subject of another PR.

Excerpt of ```hostapd.conf```
```
# Optionally, WPA passphrase can be received from RADIUS authentication server
# This requires macaddr_acl to be set to 2 (RADIUS) for wpa_psk_radius values
# 1 and 2.
# 0 = disabled (default)
# 1 = optional; use default passphrase/psk if RADIUS server does not include
#       Tunnel-Password
# 2 = required; reject authentication if RADIUS server does not include
#       Tunnel-Password
# 3 = ask RADIUS server during 4-way handshake if there is no locally
#       configured PSK/passphrase for the STA
```

Mode 3 has the advantage you can use wpa_psk_file and radius in parallel. In addition it does not require macaddr_acl if you are not interested in mac filtering,

Surprisingly VLAN assigment via wpa_psk_file works with ```dynamic_vlan``` set to 0. However vlan from radius requires setting it  to 1 or 2 (more secure).

The final setup in ```wireless``` requires these options:
```
	option ppsk '2'
	option nasid 'ap-c7'     (this is optional)
	option auth_server '<address of the radius server'
	option auth_secret '<shared secret as in radius clients.conf>'
	option dynamic_vlan '2'  
```

I'll write a proper wiki page, when I've got the time as the current one lacks a lot of details. There I'll include also the Freeradius sample configuration.

**Comment by systemcrash** [comment]
> I created a fix and introduced ppsk=2 as new option for mode 3

That'll certainly complicate things 😅. There's the hostapd.sh scripts which assume it's a flag, and some new ucode scripts which check whether it's true (boolean). 

Interesting write-up nonetheless. As regards the content of this PR which is mainly cosmetic improvements, they can go in unmodified for your `ppsk=2` and other test cases?

**Comment by tobiaswaldvogel** [comment]
You are right, that there is no dependency between this PR and the ppsk mode, so it is not relevant in any way for a decision on this PR.

The ppsk option was just for being able to assign VLANs according to the passphrase. With wpa_psk_mode=2 you can only tie it to a MAC address as the PSK is not included in the radius access request.

**Comment by systemcrash** [comment]
Merged! Thanks @tobiaswaldvogel.

**Comment by castillofrancodamian** [comment]
So how do I get the colors like the images? I don't understand how to achieve this.

**Comment by tobiaswaldvogel** [comment]
> So how do I get the colors like the images? I don't understand how to achieve this.

Are you missing just the colors or don't you see any VLAN information?

**Comment by tobiaswaldvogel** [comment]
This implementation relies on the VLAN in the ubus data, which is only populated if you configure your vlans with VLAN sections in ```/etc/config/wireless```

Here is an example:

```
config wifi-vlan
        option vid '3'
        option network 'lan'
        option name 'lan'

config wifi-vlan
        option vid '10'
        option network 'iot'
        option name 'iot'

config wifi-vlan
        option vid '11'
        option network 'local'
        option name 'loc'

config wifi-vlan
        option vid '12'
        option network 'guest'
        option name 'gst'
```

This will generate automatically the vlan file ```/var/run/hostapd-<interface name>.vlan```

It does not work if you configure the vlan file yourself using the ```vlan_file``` parameter, as this does not populate the ubus data. Also fetching the file from the router is not an option because that would required an entry in the Luci ACL and a fixed path.

**Comment by castillofrancodamian** [comment]
@tobiaswaldvogel This is my current setup:

<details><summary>/etc/config/wireless (expand)</summary>
<p>

```
root@OpenWrt:~# cat /etc/config/wireless 

config wifi-device '2g'
        option type 'mac80211'
        option path 'platform/ahb/18100000.wmac'
        option band '2g'
        option channel '11'
        option htmode 'HT20'
        option country 'AR'
        option cell_density '1'

config wifi-iface '2g_LAN'
        option device '2g'
        option network 'lan'
        option mode 'ap'
        option ssid 'LAN'
        option encryption 'psk2+ccmp'
        option ifname '2g_LAN'
        option key '12345678'

config wifi-device '5g'
        option type 'mac80211'
        option path 'pci0000:00/0000:00:00.0'
        option band '5g'
        option channel '36'
        option htmode 'VHT80'
        option country 'AR'
        option cell_density '1'

config wifi-iface '5g_LAN'
        option device '5g'
        option network 'lan'
        option mode 'ap'
        option ssid 'LAN'
        option encryption 'psk2+ccmp'
        option ifname '5g_LAN'
        option key '12345678'

config wifi-iface '2g_INVITADOS'
        option device '2g'
        option mode 'ap'
        option ssid 'INVITADOS'
        option encryption 'psk2+ccmp'
        option ifname '2g_INVITADOS'
        option key '12345678'
        option network 'invitados'

config wifi-iface '5g_INVITADOS'
        option device '5g'
        option mode 'ap'
        option ssid 'INVITADOS'
        option encryption 'psk2+ccmp'
        option ifname '5g_INVITADOS'
        option key '12345678'
        option network 'invitados'
```
</p>
</details> 

<details><summary>/etc/config/network (expand)</summary>
<p>

```
root@OpenWrt:~# cat /etc/config/network 

config interface 'loopback'
        option device 'lo'
        option proto 'static'
        option ipaddr '127.0.0.1'
        option netmask '255.0.0.0'

config globals 'globals'
        option packet_steering '1'

config device
        option name 'br-lan'
        option type 'bridge'
        list ports 'eth0.1'

config interface 'lan'
        option device 'br-lan'
        option proto 'static'
        option ipaddr '192.168.1.1'
        option netmask '255.255.255.0'
        option defaultroute '0'
        option delegate '0'

config switch
        option name 'switch0'
        option reset '1'
        option enable_vlan '1'

config switch_vlan
        option device 'switch0'
        option vlan '1'
        option ports '1 2 3 4 0t'

config device
        option type 'bridge'
        option name 'br-invitados'

config interface 'invitados'
        option proto 'static'
        option device 'br-invitados'
        option ipaddr '192.168.2.1'
        option netmask '255.255.255.0'
        option defaultroute '0'
        option delegate '0'

```

</p>
</details> 

<details><summary>ls /var/run/hostapd-* (expand)</summary>
<p>

```
root@OpenWrt:~# ls /var/run/hostapd-*
/var/run/hostapd-2g_INVITADOS.psk   /var/run/hostapd-2g_LAN.vlan        /var/run/hostapd-5g_LAN.psk         /var/run/hostapd-phy0.conf.prev
/var/run/hostapd-2g_INVITADOS.vlan  /var/run/hostapd-5g_INVITADOS.psk   /var/run/hostapd-5g_LAN.vlan        /var/run/hostapd-phy1.conf
/var/run/hostapd-2g_LAN.psk         /var/run/hostapd-5g_INVITADOS.vlan  /var/run/hostapd-phy0.conf          /var/run/hostapd-phy1.conf.prev
```

</p>
</details> 

`SNAPSHOT r28907-2389793edf`

I don't see the client list showing the network interface colors.

Captura de pantalla_20250227_171606

Captura de pantalla_20250227_171725

PR #7531, opened 2025-01-06
luci-app-ispapp: first version

kmoz000:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <abdelkarim@ispapp.co>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `luci-app-ispapp` first line subject for packages
- [x] Tested on: (aarch64_generic, aarch64_generic-23.05-SNAPSHOT, mozilla/adge latest 2025/01/06) :white_check_mark: [release_tag](https://github.com/ispapp/ispapp-linux-client/releases/tag/testing-aarch64_generic)
- [x] Description: LuCI Support of Ispapp.co Agent, a could managment platform that support most openwrt based devices.
- screenshots: 
image
image psd
image


systemcrash:
Thanks for this PR. We don't accept any new LUA additions. There is also probably no value in adding this to the main repo, since users who wish to install this can add a downloaded APK or add your repo containing it, which has a multi architecture build, so documenting that process on the website is more productive.

The website mentions Microtik devices, but the point of openwrt is a homogenised OS with a unified control interface. Maybe that's a separate product line.

kmoz000:
Thanks for the review.
We were hoping that we can spear OpenWrt users those more steps of adding a separate repo and bridge the same app using the main openwrt/luci feed repo.

Overall, what you said was logical and already implemented.

PR #7533, opened 2025-01-06
luci-base: Generate cbi-dynlist-change event after drag/touch end

Ramon00:
This PR will add generation of the cbi-dynlist-change event after drag/touch end. 

As discussed here: https://github.com/openwrt/luci/pull/7415

I regression tested it and no impact. I do not have progression testing code. so @muink Please test it!

I did validate that it does not generate console errors.


muink:
In my tests it worked fine. Thanks @Ramon00 !

Ramon00:
ok nice. Anybody else wants to give this a test?

systemcrash:
Tested. Still works.

PR #7538, opened 2025-01-08
luci-base/luci-mod-network: handle default protocol for UCI interfaces

erikrk:
The protocol defaults to 'none' in netifd if unspecified.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/network.js
+++ b/modules/luci-base/htdocs/luci-static/resources/network.js
@@ -1655,7 +1655,7 @@ Network = baseclass.extend(/** @lends LuCI.network.prototype */ {
 		if (name == null)
 			return null;
 
-		proto = (proto == null ? uci.get('network', name, 'proto') : proto);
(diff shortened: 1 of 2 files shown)


systemcrash:
The reason 'none' is not included is to force the user to choose one of the installed protos, or to go and install one. Settling on 'none' just allows a user to indefinitely ignore an interface until a later time. What is your use to have 'none'?

erikrk:
The issue is that perfectly valid UCI which may for example have been generated by an alternative management method gets reported erroneously as "Unsupported protocol type" while the interface may in fact be up and running just fine as for example an unmanaged bridge. I think if it is "wrong" to not provide the 'proto' option then netifd should not accept such configuration. The configuration IS accepted by netifd and it is treated exactly as if 'proto' were set to 'none', i.e. 'none' acts as default value for the 'proto' option. This default value should of course as per usual be reflected in LuCI so valid configuration is not shown as faulty.

systemcrash:
Do you have an opinion here @jow- ?

PR #7542, opened 2025-01-09
luci-app-nut: fix wrong nut_server option

VergilGao:
There is no driver_path prefix in the original lua code.
https://github.com/openwrt/luci/blob/f6e6b055d514595941db4a4cc11c3d9902eabefb/applications/luci-app-nut/luasrc/model/cbi/nut_server.lua#L132-L135
and currently the uci configuration file does not require this path prefix by default.

- [x] Tested on: (x86/64, 24.10.0-rc4, Edge) :white_check_mark:
- [x] \( Preferred ) Mention: @systemcrash the original code author for feedback
- [x] \( Optional ) Closes: [openwrt/packages#25552](https://github.com/openwrt/packages/issues/25552)

### DIFF
--- a/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
+++ b/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
@@ -160,7 +160,7 @@ return view.extend({
 		o = s.option(form.ListValue, 'driver', _('Driver'),
 			_('If this list is empty you need to %s'.format('<a href="/cgi-bin/luci/admin/system/package-manager?query=nut-driver-">%s</a>'.format(_('install drivers')))));
 		driver_list.forEach(driver => {
-			o.value(driver_path + driver);
+			o.value(driver);
 		});
 		o.optional = false;
 


systemcrash:
Your fix is correct. But please keep the same amount of line numbers so an i18n refresh is not required.

VergilGao:
> Your fix is correct. But please keep the same amount of line numbers so an i18n refresh is not required.

ok.

systemcrash:
Please rebase on current master and force push

systemcrash:
Please also fix your Author line (email in github - set it to the same as your SOB). This won't be accepted:
`author:"羽先生 <8655163+VergilGao@users.noreply.github.com>"`

VergilGao:
> Please also fix your Author line (email in github - set it to the same as your SOB). This won't be accepted:
> `author:"羽先生 <8655163+VergilGao@users.noreply.github.com>"`

OK，i will do it tomorrow since its late night for me.

systemcrash:
Should be OK. Today would have been better since we're approaching 24 release.

PR #7547, opened 2025-01-09
luci-proto-3g/ppp/pppossh: fix setting of keepalive

erikrk:
Fix regressions from previous fix:

1) It is no longer possible to leave the keepalive option empty. Then
   it becomes keepalive='0 1'

2) In case the interval is different from 1 it is necessary to
   explicitly set the option even if failures is left empty.

### DIFF
--- a/protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
+++ b/protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
@@ -23,14 +23,14 @@ network.registerPatternVirtual(/^3g-.+$/);
 function write_keepalive(section_id, value) {
 	var f_opt = this.map.lookupOption('_keepalive_failure', section_id),
 	    i_opt = this.map.lookupOption('_keepalive_interval', section_id),
-	    f = (f_opt != null) ? +f_opt[0].formvalue(section_id) : null,
-	    i = (i_opt != null) ? +i_opt[0].formvalue(section_id) : null;
+	    f = parseInt(f_opt?.[0]?.formvalue(section_id), 10),
+	    i = parseInt(i_opt?.[0]?.formvalue(section_id), 10);
 
-	if (f === '' || isNaN(f))
-		f = null;
-
-	if (i == null || i == '' || isNaN(i) || i < 1)
+	if (isNaN(i))
 		i = 1;
+
+	if (isNaN(f))
+		f = (i == 1) ? null : 5;
 
 	if (f !== null)
 		uci.set('network', section_id, 'keepalive', '%d %d'.format(f, i));
@@ -142,7 +142,7 @@ return network.registerProtocol('3g', {
 
 		o = s.taboption('advanced', form.Value, '_keepalive_interval', _('LCP echo interval'), _('Send LCP echo requests at the given interval in seconds, only effective in conjunction with failure threshold'));
 		o.placeholder = '1';
-		o.datatype    = 'min(1)';
+		o.datatype    = 'and(uinteger,min(1))';
 		o.write       = write_keepalive;
 		o.remove      = write_keepalive;
 		o.cfgvalue = function(section_id) {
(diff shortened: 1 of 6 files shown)


protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
```diff
@@ -23,15 +23,17 @@ network.registerPatternVirtual(/^3g-.+$/);
...
 
 	if (i == null || i == '' || isNaN(i) || i < 1)
 		i = 1;
 
+	if (f === null || f === '' || isNaN(f))
+		f = (i == 1) ? null : 5;
+
 	if (f !== null)
 		uci.set('network', section_id, 'keepalive', '%d %d'.format(f, i));
 	else
```
systemcrash, line 39: Perhaps optional chaining to avoid all the null checks:
```js
function write_keepalive(section_id, value) {
	const f_opt = this.map.lookupOption('_keepalive_failure', section_id)?.[0]?.formvalue(section_id);
	const i_opt = this.map.lookupOption('_keepalive_interval', section_id)?.[0]?.formvalue(section_id);

	let i = parseInt(i_opt, 10) || 1;
	let f = parseInt(f_opt, 10);

	if (isNaN(f)) {
		f = (i === 1) ? null : 5;
	}

	if (f !== null) {
		uci.set('network', section_id, 'keepalive', `${f} ${i}`);
	} else {
		uci.unset('network', section_id, 'keepalive');
	}
}
```

Please test :)

erikrk: Thanks, updated!

protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
```diff
@@ -142,7 +142,7 @@ return network.registerProtocol('3g', {
 
 		o = s.taboption('advanced', form.Value, '_keepalive_interval', _('LCP echo interval'), _('Send LCP echo requests at the given interval in seconds, only effective in conjunction with failure threshold'));
 		o.placeholder = '1';
-		o.datatype    = 'min(1)';
+		o.datatype    = 'and(uinteger,min(1))';
```
systemcrash, line 145: This is redundant: since `min(1)` keeps it above zero (`uinteger`) anyway.

systemcrash: unless you can show me an example where it's necessary? (it executes more code)

erikrk: It is actually not redundant since `min(1)` allows floating point as well. It has to be an integer.

systemcrash: Ah, of course. Floats.

protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js
systemcrash: Doesn't look like it :)

erikrk: OK, I ended up doing a somewhat more conservative change but the "messy" checks for `null` and `''` are nonetheless all gone. Note that I am fully aware of the merits of `let` and `const` over `var` and so on but it is not the only instance thereof and I do not feel like redoing the whole file. I mostly just wanted to fix the regressions and get things working. It is currently in a broken state.

systemcrash: OK good. Looks fine. That's fine as you've done it, though we encourage new code to use ES6 or better.

systemcrash:
Can you reword the commit message so the wording focuses on 'now' and not 'then'. ( I assume you referred to f3d26a2a560d8ae526c617eb1c28821ec900fa7a )

```
Fix regressions from f3d26a2a560d8ae526c617eb1c28821ec900fa7a:

1) It is now possible to leave the keepalive option empty. 
2) In case the interval was different from 1 it was necessary to
   explicitly set the option even if failures is left empty.
```

erikrk:
Commit message updated now.

systemcrash:
Merged. Thanks @erikrk

Issue #7539, opened 2025-01-08
Feature request: WireGuard interfact should allow for IPv6-PD intput


adelton:
### What would you like to see in luci?

When configuring a 6in4 interface with LuCI, it is possible to specify `IPv6 routed prefix` in the `General Settings` tab. The value(s) then lands in `/etc/config/network` in `list ip6prefix`.

For the WireGuard VPN protocol, the same UI-based approach is desired. Configuring the `ip6prefix` manually in `/etc/config/network` works and is then shown in the `IPv6-PD` field on the interface overview page but making it part of the UI would make things easier.

amanjuman:
This is very much needed.
I wrote a guide for manual setup [here](https://medium.com/@amanjuman/setting-up-wireguard-ipv6-tunnel-with-prefix-delegation-pd-on-openwrt-a56d7d6c6ff9).

vevere:
thanks a lot for the tip (set `ip6prefix `manually in `/etc/config/network`) ... you save my day !
for those who want to easily have their entire LAN behind a VPN that is great

Issue #5059, opened 2021-05-21
luci-add-ddns: IPv6 handling - UI bug/discrepency - prefix delegation


ldir-EDB0:
@Ansuel 

Hi.  Hitting what I think is a bug/oversight in the luci ddns configuration page related to IPv6 interface handling.  My ISP provides dual-stack IPv4 & IPv6 over a vdsl interface.  As far as openwrt is concerned I have logical interfaces of 'wan4' and 'wan6' for IPv4/v6 operating over the same physical interface 'eth0'.  My lan interface is 'eth1'.

Extract from /etc/config/network
`
config interface 'wan4'
        option ifname 'eth0'
        option proto 'dhcp'
        option norelease '1'
        option peerdns '0'
        option delegate '0'
        option hostname '*'

config interface 'wan6'
        option proto 'dhcpv6'
        option reqprefix '56'
        option peerdns '0'
        option norelease '1'
        option reqaddress 'none'
        option ifname 'eth0'
`

My ISP delegates a prefix, thus 'wan6' doesn't get an IPv6 address, instead the logical 'lan' interface (physical eth1) is assigned an IPv6 address based on the delegated prefix.  Thus ddns-scripts needs to know about logical 'wan6' events for prefix change events etc but needs to obtain the IPv6 address for use with the ddns service provider from logical 'lan'.  A working config for ddns is shown below:

Extract from working /etc/config/ddns
`
config service 'myddns_ipv6'
	option use_ipv6 '1'
	option ip_source 'network'
	option ip_network 'lan'
	option interface 'wan6'
`

Unfortunately this combination of 'ip_source/network' vs 'interface' cannot be configured via the luci gui.  When using defining the "IP address source" as 'network' and "Network" as 'lan', the "event network" is also automatically changed to 'lan', thus ddns-scripts doesn't a) start or b) pick up any prefix change events.  Changing the "IP address source" to 'wan6' changes the event network also to 'wan6' which is correct from an 'event' perspective BUT wan6 doesn't actually have an IPv6 address, instead this is delegated to 'lan', thus it cannot report an IPv6 to the ddns provider.

Using the gui will destroy an active, working configuration.

PR #7548, opened 2025-01-09
luci-proto-wireguard: Add the option for setting ip6prefix.

adelton:
- [ ] Tested on: local x86_64 VM with OpenWrt 23.05.5 (r24106-10cc5fcd00)
- [ ] \( Preferred ) Screenshot or mp4 of changes:  Screenshot_2025-01-09_20-20-17
- [ ] \( Optional ) Closes: https://github.com/openwrt/luci/issues/7539
- [ ] Description: Adds the option for setting `ip6prefix` for WireGuard interfaces.

### DIFF
--- a/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
+++ b/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
@@ -168,6 +168,9 @@ return network.registerProtocol('wireguard', {
 		o.datatype = 'ipaddr';
 		o.optional = true;
 
+		o = s.taboption('general', form.DynamicList, 'ip6prefix', _('IPv6 routed prefix'), _('This is the prefix routed to you by the tunnel broker for use by clients'));
+		o.datatype = 'cidr6';
+
 		o = s.taboption('general', form.Flag, 'nohostroute', _('No Host Routes'), _('Optional. Do not create host routes to peers.'));
 		o.optional = true;
 


systemcrash:
This won't see much usage, since it looks like clients should each have an IP statically set from that prefix anyway. Why not just use advanced->prefix filter (choose HE.net interface) and advanced->suffix?

adelton:
> This won't see much usage, since it looks like clients should each have an IP statically set from that prefix anyway.

Well this is the point -- our local community tunnel broker prescribes the static IP to be outside of the prefix. The address is `2a03:abcd:200::XXX`, the prefix allocated for the tunnel is `2a03:abcd:XXX::/48` (where XXX != 200).

> Why not just use advanced->prefix filter (choose HE.net interface) and advanced->suffix?

I tried to put the prefix to the IPv6 prefix filter, and `::1` to the IPv6 suffix. Diffing the `/etc/config/network`, this resulted in the prefix value being stored in `list ip6class` (instead of `list ip6prefix`), and `::1` in `option ip6ifaceid`.

The problem is, with this configuration, the `br-lan` interface did not get the `2a03:abcd:XXX::1/64` address as it gets with with `ip6prefix`. So while the tunnel gets established and IPv6 connectivity works from the router itself, it does not work from machines on the local network. Is there something else I should be setting in the prefix filter and/or suffix to get the same behaviour as `ip6prefix` provides?

adelton:
I notice https://github.com/openwrt/luci/issues/5059 actually describes setup very similar to mine -- the IPv6 address (in my case `2a03:abcd:XXX::1`) on `br-lan`, while prefix handled by the wan6 (in my case the WireGuard setup).

systemcrash:
I could have sworn I had this working. But I have pure IPv6 now, and my recommendation doesn't seem to work on my he.net tunnel... maybe it was just a static IPv6 I had assigned. :/

adelton:
@systemcrash Were you able to five this some more thoughts? Note that we are merely exposing in the UI a configuration which is already possible and functional.

systemcrash:
It's an additional option which takes up screen real-estate and doesn't make any client config dynamic. If you change that setting, every single 'client' also needs to have its static IP changed. WG is independent of a number of settings, and I don't know whether it's possible to extend netifd and odhcpd to auto-assign an IP(v6) to WG clients when the interface gets a (new) prefix, which would be a nice to have and justify having this setting much better.

IPv6 is quite flexible. Statically setting an IPv6 for the WireGuard interface and each client in a ULA range, and using an IPv6 Rule for the 'main' table for anything in the ULA prefix, everything is taken care of, and all the WireGuard clients come from an (one and the same) IPv6 for whichever interface they egress from. The only 'drawback' to such an approach is your IPv6 is the 'main' IPv6 for that interface. But I don't even need to think about what to do when a prefix changes via HE.net or my provider.

adelton:
I do agree that adding the option takes screen real-estate.

But it does so because it exposes an option that I had to use with two of my IPv6 tunnel brokers in my `/etc/config/network`. It's not a nice-to-have option. Without it, things don't work. So the value is shown on the Interfaces overview page but when they user clicks Edit, it's nowhere to be found in that form.

As for the change of the IP addresses of the local IPv6-enabled machines -- well I never needed to change the value because it is prescribed by the tunnel broker and I'm glad if they give me one static prefix.

If I understand correctly, you suggest to use some sort of NAT-ing which might be what some people might desire but I really like and need the global native IPv6 addresses in my network.

Please reconsider applying this PR.

systemcrash:
How about under the advanced tab then?

adelton:
Sure, we can put it there.

Does it mean I use `modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js` for that? Or what is the canonical way to put things to the Advanced tab but for the WireGuard interface only?

systemcrash:
```diff
- o = s.taboption('general', ...
+ o = s.taboption('advanced', ... 
```

adelton:
Ah, thanks. Updated, rebased on master.

protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js
```diff
@@ -193,6 +193,8 @@ return network.registerProtocol('wireguard', {
 			return true;
 		};
 
+		o = s.taboption('advanced', form.DynamicList, 'ip6prefix', _('IPv6 routed prefix'), _('This is the prefix routed to you by the tunnel broker for use by clients'));
```
systemcrash: Change to:
`This is the prefix routed to you by your provider for use by clients`

adelton: Updated.

systemcrash:
Merged. Thanks @adelton

adelton:
Thank you!

Issue #3340, opened 2019-11-21
Interface IPv6 configuration: PD and static IPv6 address


madires:
Hi!

I found an issue with the IPv6 interface configuration when applying an assignment of a PD subprefix and a static IPv6 address in OpenWrt 18.06.5. Typically you get a public IPv6 prefix from your ISP and also want to have a static ULA prefix in your LAN. LuCI allows me to set either the PD assignment or a static IPv6 address. But /etc/config/network supports both at the same time:

config interface 'lan'
        option type 'bridge'
        option ifname 'eth0.1'
        option proto 'static'
        option netmask '255.255.255.0'
        option ip6assign '64'
        option ip6ifaceid '::1'
        option ipaddr '192.168.1.1'
        option ip6addr 'fd00:dead:beef:1::1/64'

The example above works fine. Unfortunately, LuCI displays only the PD configuration and will remove the "option ip6addr" when saving the current configuration. It's annoying to have to add the "option ip6addr" manually again each time. The other workaround is to use an alias interface which is also cumbersome. It would be nice to have both settings (PD and static IPv6 address) at the same time.

Cheers,
 Markus

cvmiller:
As a work-around you can set the ULA in the globals section:

```
config globals 'globals'
	option ula_prefix 'fd00:dead:beef::/64'
```

adelton:
I can see the
```
o.depends('ip6assign', '');
```
is there since 029631ec489b2d28a9e4b3a6ef4f2de9e999b956, in the original form of
```
ip6addr:depends("ip6assign", "")
```

@sbyx, would you remember the original reasoning? The commit message says "Make static IPv6-configuration more usable" but @madires suggests that being able to have both `ip6addr` and `ip6assign` would actually be more useful.

What would break if we just removed that `o.depends('ip6assign', '');` in https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/protocol/static.js#L187?

systemcrash:
I recall looking at this, when I was interested to increase the visibility of (static) IPv6. Both should be configurable.

systemcrash:
PD and static should both be possible, unless we can ascertain a reason why that might be undesirable.

PR #7556, opened 2025-01-10
luci-base: expose static IPv6 configuration even if IPv6-PD is enabled.

adelton:
- [ ] Tested on: x86_64 VM with  OpenWrt 24.10.0-rc5 (r28304-6dacba30a7)
- [ ] \( Optional ) Closes https://github.com/openwrt/luci/issues/3340
- [ ] Description: expose static IPv6 configuration even if IPv6-PD is enabled.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/protocol/static.js
+++ b/modules/luci-base/htdocs/luci-static/resources/protocol/static.js
@@ -184,14 +184,11 @@ return network.registerProtocol('static', {
 		o = s.taboption('general', form.DynamicList, 'ip6addr', _('IPv6 address'));
 		o.datatype = 'ip6addr';
 		o.placeholder = _('Add IPv6 address…');
-		o.depends('ip6assign', '');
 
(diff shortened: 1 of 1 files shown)


adelton:
I had to move this PR one commit back in the master branch, to avoid the issue being addressed by https://github.com/openwrt/luci/pull/7557.

Ramon00:
Would this not make it confusing for most people? I.e. nowhere is it now shown that this is optional if you have IPv6-PD?

adelton:
Well the values are optional since they are not mandatory. :-)

What is your suggestion? Do you think we should add some explicit hints about these being optional and "possibly not needed" when IPv6-PD is enabled? Or?

Ramon00:
yeah maybe add to the hint "optional if IPv6-PD enabled". Alternative would be add an extra toggle  "enable manual IPv6" which only shows up if you have IPv6-PD and which unhides the additional 3 items?

PR #7557, opened 2025-01-10
luci-base: addressing this.vstack[0] is undefined.

adelton:
- [ ] Tested on: x86_64 VM with OpenWrt 24.10.0-rc5 (r28304-6dacba30a7)
- [ ] Description: with a67d68353b2f55ff888e4a7062a31d9fccb70357, any attempt to `Edit` a network interface produces a yellow message `TypeError: this.vstack[0] is undefined`:
Screenshot_2025-01-10_23-04-51
I believe the reason is that `.datatype` cannot be an empty string. With this change the `TypeError` is gone.


systemcrash:
Good catch. That does look better. Thanks @adelton

PR #7558, opened 2025-01-11
luci-base: add member `allowduplicates` to `DynamicList`

muink:
- [x] Tested on: (x86_64, 24.10.0-rc4, Chrome) :white_check_mark:
- [x] \( Preferred ) Mention: @systemcrash 

## Description:

Add member `allowduplicates` to `DynamicList` to allow duplicate underlying form values.

`allowduplicates` defaults is `null`, `DynamicList` will keep the same behavior as before.
If `o.allowduplicates = true`, the underlying form value will not be checked for duplication.

## Screenshot:

image

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -3863,6 +3863,17 @@ const CBIValue = CBIAbstractValue.extend(/** @lends LuCI.form.Value.prototype */
 const CBIDynamicList = CBIValue.extend(/** @lends LuCI.form.DynamicList.prototype */ {
 	__name__: 'CBI.DynamicList',
 
+	/**
+	 * Allows the underlying form controls to have multiple identical values.
+	 *
+	 * Default is `null`. If `true`, the underlying form value will
+	 * not be checked for duplication.
+	 *
+	 * @name LuCI.form.DynamicList.prototype#allowduplicates
+	 * @type boolean
+	 * @default null
+	 */
+
 	/** @private */
 	renderWidget(section_id, option_index, cfgvalue) {
 		const value = (cfgvalue != null) ? cfgvalue : this.default;
@@ -3872,6 +3883,7 @@ const CBIDynamicList = CBIValue.extend(/** @lends LuCI.form.DynamicList.prototyp
 		const widget = new ui.DynamicList(items, choices, {
 			id: this.cbid(section_id),
 			sort: this.keylist,
+			allowduplicates: this.allowduplicates,
 			optional: this.optional || this.rmempty,
 			datatype: this.datatype,
 			placeholder: this.placeholder,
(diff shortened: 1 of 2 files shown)


modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -2505,7 +2505,8 @@ const UIDynamicList = UIElement.extend(/** @lends LuCI.ui.DynamicList.prototype
 			return;
 
 		sbIn.setValues(sbEl, null);
-		sbVal.element.setAttribute('unselectable', '');
+		if (!this.options.duplicate)
+			sbVal.element.setAttribute('unselectable', '');
```
Ramon00, line 2509: Why is there a need for a dependency of "unselectable" on "duplicate"?

muink: Because we can add preset chooses to `DynamicList` through `o.value(key, label)`.
And the selected key choose will become `unselectable`.
This will prevent us from adding the same key.

Ramon00: ah right, just from the name I was thinking it had to do with the ability to select the text, e.g. for copy/paste.

systemcrash:
Although it might be a solution in general, it might go unused but for your use-case @muink. Perhaps there is a better way to capture config?

This seems OK in general. Opinions from others? @hnyman @stokito @dannil ?

muink:
I've tried a few other solutions, but they are all too heavy. This is currently the simplest stupidst solution I can come up with.  
And `uci` can store duplicate lists, it would be strange if `luci` can't do that.

dannil:
> This seems OK in general. Opinions from others? @hnyman @stokito @dannil ?

I'm fine with it, the semantics of a list is to allow duplicates anyway, otherwise you're dealing with sets, and since it doesn't break existing default functionality I don't see any problem.

systemcrash:
Hi @muink - please rename the setting to `allowduplicates`. I like @jow-'s suggestion from the linked ticket, and it's what I would settle on also, since the name makes clear what the boolean value hinges on (and that duplicates is a plural noun).

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -3863,6 +3863,17 @@ const CBIValue = CBIAbstractValue.extend(/** @lends LuCI.form.Value.prototype */
 const CBIDynamicList = CBIValue.extend(/** @lends LuCI.form.DynamicList.prototype */ {
 	__name__: 'CBI.DynamicList',
 
+	/**
+	 * Allows the underlying form controls to have repeated values.
```
systemcrash, line 3867:
```js
	 * Allows the underlying form controls to have multiple identical values.
```

muink:
force pushed.

Issue #2436, opened 2019-01-06
luci-app-ddns: Namecheap DDNS setting are misleading


mayl:
Setting up DDNS with namecheap through LUCI was a little confusing for me.  Basically, LUCI populates the "username" field into the "host" field in the [namecheap update url](https://www.namecheap.com/support/knowledgebase/article.aspx/29/11/how-do-i-use-a-browser-to-dynamically-update-the-hosts-ip). Until I figured that out, I was putting my namecheap account username into the username field and it obviously wasn't working.  It would help newbies like me out if this was a little bit more consistent/clear.

castillofrancodamian:
Can you help me how to configure the DDNS client in OpenWrt with this service?

ryandmaclean:
This had me confused for a while until I found this post. The 'username' field is really the 'host' field as mentioned in the OP. Thus, 'Domain: example.net', and 'username: myhost' updates 'myhost.example.net'.

It's additionally confusing that the 'Domain' field shows 'myhost.example.net' as an example of what to enter as shown below:

![image](https://user-images.githubusercontent.com/8517067/95623299-03eecb80-0a43-11eb-941e-92d785f11712.png)

castillofrancodamian:
> This had me confused for a while until I found this post. The 'username' field is really the 'host' field as mentioned in the OP. Thus, 'Domain: example.net', and 'username: myhost' updates 'myhost.example.net'.
> 
> It's additionally confusing that the 'Domain' field shows 'myhost.example.net' as an example of what to enter as shown below:
> 
> ![image](https://user-images.githubusercontent.com/8517067/95623299-03eecb80-0a43-11eb-941e-92d785f11712.png)

What device are you using and what version of OpenWrt? I would also like you to show me your config file.

feckert:
Have you tried the newest luci-app-ddns? There we have improved the description.
https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L736

castillofrancodamian:
> Have you tried the newest luci-app-ddns? There we have improved the description.
> https://github.com/openwrt/luci/blob/master/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L736

What would "USERNAME" be?

feckert:
You got me there now ;-)
That is of course wrong.
USERNAME must be replaced with DOMAIN.

castillofrancodamian:
[...]

Issue #5297, opened 2021-08-24
luci-app-ddns: Helpful changes


LecrisUT:
The current field names and settings for the `ddns` app are very confusing for both users and maintainers. There are issues opened about this like #2436  and various other by confused users. I would propose the following changes:

1. Allow the script maintainers to add, remove, rename, and set properties (hidden, required etc.) to fields via a `.json` file. For example this can look like:
```json
{
   "name": "My new DDNS script",
   "fields": [ {
      "name": "record name",
      "required": true
   }, {
      "name": "domain",
      "required": true
   }]
}
```
2. For `ddns-scripts`, provide a complete example `curl` script including any possible error catchers. There are some issues out there linked to an incomplete `ddns-script`. This could also be used to add better tests for the CI.
3. Add a `view log` option outside the `edit` dialog, and output any or all the contents in the `\var\log`, `DATFILE`, `ERRFILE`.
4. Lowest priority, but would definitely be appreciated: Provide a contact method for when any of the scripts are failing.

Situations where this is helpful:
- Remove the confusion when you have to input the `record name` into the `username` field.
- Allow `ddns` on local servers like PowerDNS
- Simplify and clarify scripts, e.g. remove the need for `param_opt`

I understand that this could break some of the current `ddns-scripts`, however some of them could use a rewrite to better manage error at the very least.

Issue #3582, opened 2020-01-28
luci-app-ddns requires user / password for keyauth ddns provider.


morzexxx:
When select the keyauth provider the form doesn't change from user/password to just one password in current trunk.

feckert:
Please tell us your OpenWrt/LuCI version?

morzexxx:
OpenWrt SNAPSHOT r12001-2fe464a712 / LuCI Master git-20.019.74462-98ba87a
running in rpi 4B

mrjoel:
I'm also impacted by this, whether by specifying a service or an "update_url" entry that doesn't use username, password, or either it is definitely an irritant to have luci complaining about an otherwise valid configuration. A simple workaround is to just set the username and password fields to "unused" or some other dummy value, but it would be nice to not be required.

lorthirk:
+1. At least for my provider (dynu.com), username is optional since you can send the domain and the password and their API will match the current user.

m3thm4th:
luci-app-ddns is utterly broken in OpenWRT 21.02 because of this issue. All the required fields which are actually not to be used make the application unusable. It used to work with no issues in 19.07.

The `Path to CA-Certificate` field which is added when selecting `Use HTTP Secure` is also required even though the description states it can be ignored. This field could also be automatically filled as its hint to `/etc/ssl/certs`.

xorbug:
I agree, this is tedious the way it is now.
I can be wrong as I just gave quick check, but looks like the 21.02 implementation is missing the parsing of the template url and the subsequent filtering of unneeded fields from the settings (and another small thing, the providers dropdown list is not sorted). Afaics, this was being done in 19.07 instead.

But in general, what is the current format used by these json service files now? I mean I can figure it out from the default files in _net/ddns-scripts_, but is there a generic schema with all possible keys for that json object? If it's not possible already, what about explicitly specifying fields and their attributes (e.g. mandatory/optional) in the service json file itself?

This may be needed because, even if parsing was done (and should already improve the current situation by a lot), there may be some optional field in the url, and labeling a field as required just because of its presence in the url may be undesired. E.g. consider an url containing the `[DOMAIN]` field: the simplest approach would then assume that it is a required field after parsing (so would probably be set for non-empty validation at least), but that same provider could work just fine without it, maybe deriving it from the token. So it could as well be optional, and actually should be.

[...]

PR #7559, opened 2025-01-11
luci-app-ddns: show the URL which will be used.

adelton:
- [x] Tested on: x86_64 VM with OpenWrt 24.10.0-rc5 (r28304-6dacba30a7)
Screenshot_2025-01-11_14-33-50
- [x] \( Optional ) Closes https://github.com/openwrt/luci/issues/2436
- [x] Description: When using the DDNS settings, I was always puzzled a bit by the `Replaces [DOMAIN] in Update-URL (URL-encoded)` comments. It's certainly nice to have that information but I always wondered "what is that Update-URL?" for this service? Then I came across https://github.com/openwrt/luci/issues/2436 which basically discusses that this DDNS provider does not use username and expects host (?) instead. We likely cannot finetune the description of the fields in the UI to this case but maybe if we actually showed the admin the Update-URL which will be used, it might be easier for them to figure out which field to use for what.

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -586,6 +586,7 @@ return view.extend({
 			return _this.handleGetServiceData(service).then(L.bind(function (service_data) {
 				s.service_available = true;
 				s.service_supported = true;
+				s.url = null;
 
 				if (service != '-') {
 					if (!service_data)
@@ -594,6 +595,11 @@ return view.extend({
 						service_data = JSON.parse(service_data);
 						if (ipv6 == "1" && !service_data.ipv6)
 							s.service_supported = false;
+						else if (ipv6 == "1") {
+							s.url = service_data.ipv6.url;
+						} else {
+							s.url = service_data.ipv4.url;
+						}
 					}
 				}
 
@@ -672,6 +678,14 @@ return view.extend({
 					o.cfgvalue = function () {
 						return _("Service doesn't support this IP type")
 					};
+				}
+
+				if (Boolean(s.url)) {
+					o = s.taboption('basic', form.DummyValue, '_url', _("Update URL"));
+					o.rawhtml = true;
+					o.default = '<div style="font-family: monospace;">'
+						+ s.url.replace(/&/g,"&amp;").replace(/</g,"&lt;").replace(/>/g,"&gt;")
+						+ '</div>';
 				}
 
 				var service_switch = s.taboption('basic', form.Button, '_switch_proto');


adelton:
This can also be considered as a fix for https://github.com/openwrt/luci/issues/5297.

adelton:
In https://github.com/openwrt/luci/issues/3582 folks suggest that LuCI should parse the URL for the `[FIELD]`s to then determine what UI fields should be mandatory.

If we don't plan to do that (or do that right now) and we start to show the URL in the UI with this PR, could we make all the currently-mandatory fields optional? And basically make it the responsibility of the user to fill in the fields used in the URL, based on the hints in under the fields?

systemcrash:
I suppose this is a start at being helpful, but it does have exceptions. Some providers have old info that mandates a username+password, but providing a token (i.e. password field) only, works fine. The providers template URLs drive which fields are mandated. Improvements in this area are welcome.

adelton:
Would you prefer this PR to also attempt to do those improvements, or would it make sense to accept the current work to expose the URL, and to subsequent work in a separate PR?

systemcrash:
We're close to 24 release, so minimal (atomic) changes are the easiest to test and review. But this can be kept minimal if done right (mandating fields based on those in the URL). The field presence in the provider URL is fairly easy to detect. It might require an exhaustive review of providers and their URLs (which are all in another repo anyway). This PR as-is satisfies the issue already, no?

adelton:
If we are talking about the https://github.com/openwrt/luci/issues/2436, I believe this PR should significantly alleviate hardship people might experience attempting to fill the form to match the expectations, by showing where each of the values will land.

applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
```diff
@@ -674,6 +680,11 @@ return view.extend({
 					};
 				}
 
+				if (Boolean(s.url)) {
+					o = s.taboption('basic', form.DummyValue, '_url', _("Update URL"));
+					o.default = s.url;
```
systemcrash, line 685: I find that monospace makes the URL easier to parse:
```js
					o.rawhtml = true;
					o.default = '<div style="font-family: monospace;">' + s.url; + '</div>'
```

adelton: We'd have to properly escape that `s.url` as well.

If we want to show this in monospace, I'd much rather add support for `.monospace` to `form.Value` and `form.DummyValue` and configure it more cleanly that way, rather than hacking it with `o.rawhtml`. (A much bigger change, granted.)

systemcrash: I had a thought that perhaps adding a whole `.style` element might be the way to go. But that's for another day. 

The whole point of `.rawhtml` is so you don't have to escape. That you provide raw HTML.

https://openwrt.github.io/luci/jsapi/LuCI.form.DummyValue.html

adelton: Right. And if we are to provide raw HTML, it has to be a valid, properly escaped HTML, doesn't it? And we cannot know that for the URL value.

So I added some `.replace`s; rebased on master.

adelton: Those only have to be escaped in attributes.

PR #7579, opened 2025-01-23
luci-app-attendedsysupgrade: sort latest version array

efahl:
The Attended Sysupgrade app requires that the array of latest versions be in order from newest release to oldest.  Ensure that this is true even when upstream presents them in arbitrary order.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -450,6 +450,9 @@ return view.extend({
 			} else {
 				const latest = response.json().latest;
 
+				// ensure order: newest to oldest release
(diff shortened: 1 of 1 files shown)


efahl:
Should be backported to 24.10.

aparcar:
I think this isn't the best solution, it requires clients to update packages before a firmware upgrade. We should always avoid that

systemcrash:
> I think this isn't the best solution, it requires clients to update packages before a firmware upgrade. We should always avoid that

And that's because of a list sort?

hnyman:
> >    isn't the best solution, it requires clients to update packages before a firmware upgrade. We should always avoid that
>
> And that's because of a list sort?

Likely aparcar means that is does not yet benefit those who are trying to upgrade from versions before this patch. But this will help in future when later upgrading from versions with this patch.

systemcrash:
Sorry, I read "update packages" as "update all packages" instead of "update this package".

PR #7597, opened 2025-01-30
Add firmware upgrade notification to status page

aparcar:
This is likely not cleanly implemented, I'm unfamiliar with the LuCI tricks and JavaScript in general.

The core idea is to show a notification whenever a new firmware is available.

image

Right now it's just a notification, this could be extended to include the luci-app-attendedsysupgrade functionality or a slimmed down version.

@jow- please have a look

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
@@ -2,6 +2,8 @@
 'require baseclass';
 'require fs';
 'require rpc';
+'require uci'
+'require ui';
 
 var callLuciVersion = rpc.declare({
 	object: 'luci',
@@ -18,6 +20,43 @@ var callSystemInfo = rpc.declare({
 	method: 'info'
 });
 
+function showUpgradeModal(type, current_version, new_version) {
+	ui.showModal(_('New Firmware Available'), [
+		E('p', _('A new %s version of OpenWrt is available: %s').format(type, new_version)),
+		E('p', _('Your current version is: %s').format(current_version)),
+		E('p', _('Please check the https://firmware-selector.openwrt.org/ for the firmware upgrade image')),
+		E('div', { class: 'btn', click: ui.hideModal }, _('Close')),
+	]);
+};
+
+function checkDeviceAvailable(boardinfo, new_version) {
+	const profile_url = 'https://downloads.openwrt.org/releases/%s/targets/%s/profiles.json'.format(new_version, boardinfo.release.target);
+	return fetch(profile_url)
+		.then(response => response.json())
+		.then(data => {
+			// special case for x86 and armsr
+			if (Object.keys(data.profiles).length == 1 && Object.keys(data.profiles)[0] == "generic") {
+				return true;
+			}
+
+			for (var i = 0; i < data.profiles.length; i++) {
+				if (data.profiles[i].profile == boardinfo.board_name) {
+					return true;
+				}
+				for (var j = 0; j < data.profiles[i].supported_devices.length; j++) {
+					if (data.profiles[i].supported_devices[j] == boardinfo.board_name) {
+						return true;
+					}
+				}
+			}
+		})
+		.catch(error => {
+			console.error('Failed to fetch profile information:', error);
+			return false;
+		});
+};
+
+
 return baseclass.extend({
 	title: _('System'),
 
@@ -25,8 +64,39 @@ return baseclass.extend({
 		return Promise.all([
 			L.resolveDefault(callSystemBoard(), {}),
 			L.resolveDefault(callSystemInfo(), {}),
-			L.resolveDefault(callLuciVersion(), { revision: _('unknown version'), branch: 'LuCI' })
+			L.resolveDefault(callLuciVersion(), { revision: _('unknown version'), branch: 'LuCI' }),
+			uci.load('luci')
 		]);
+	},
+
+	oneshot: function(data) {
+		var boardinfo = data[0];
+		var check_upgrades = uci.get_first('luci', 'core', 'check_firmware_version') || false;
+		console.log(check_upgrades)
+
+		if (check_upgrades == '1' || check_upgrades == 'true') {
+				fetch('https://downloads.openwrt.org/.versions.json')
+				.then(response => response.json())
+				.then(data => {
+						if (data.oldstable_version
+								&& data.oldstable_version > boardinfo.release.version
+								&& checkDeviceAvailable(boardinfo, data.oldstable_version)) {
+							showUpgradeModal("oldstable", boardinfo.release.version, data.oldstable_version)
+						} else if (data.stable_version
+								&& data.stable_version > boardinfo.release.version
+								&& checkDeviceAvailable(boardinfo, data.stable_version)) {
+							showUpgradeModal("stable", boardinfo.release.version, data.stable_version)
+						} else if (data.upcoming_version
+								&& boardinfo.release.version > data.stable_version
+								&& data.upcoming_version > boardinfo.release.version
+								&& checkDeviceAvailable(boardinfo, data.upcoming_version)) {
+							showUpgradeModal("upcomming", boardinfo.release.version, data.upcoming_version)
+						}
+				})
+				.catch(error => {
+						console.error('Failed to fetch version information:', error);
+				});
+		}
 	},
 
 	render: function(data) {
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js
@@ -40,6 +40,11 @@ function startPolling(includes, containers) {
 				else if (includes[i].content != null)
 					content = includes[i].content;
 
+				if (typeof (includes[i].oneshot) == 'function') {
+					includes[i].oneshot(results ? results[i] : null);
+					includes[i].oneshot = null;
+				}
+
 				if (content != null) {
 					containers[i].parentNode.style.display = '';
 					containers[i].parentNode.classList.add('fade-in');


modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
```diff
@@ -18,17 +20,85 @@ var callSystemInfo = rpc.declare({
...
+							showUpgradeModal("oldstable", boardinfo.release.version, data.oldstable_version)
+						} else if (data.stable_version
+								&& data.stable_version > boardinfo.release.version
+								&& checkDeviceAvailable(boardinfo, data.stable_version)) {
+							showUpgradeModal("stable", boardinfo.release.version, data.stable_version)
+						} else if (data.upcoming_version
+								&& boardinfo.release.version > data.stable_version
+								&& data.upcoming_version > boardinfo.release.version
+								&& checkDeviceAvailable(boardinfo, data.upcoming_version)) {
+							showUpgradeModal("upcomming", boardinfo.release.version, data.upcoming_version)
```
systemcrash, line 93: `upcoming` (or just `new`?)

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
```diff
@@ -18,17 +20,85 @@ var callSystemInfo = rpc.declare({
...
-			L.resolveDefault(callLuciVersion(), { revision: _('unknown version'), branch: 'LuCI' })
+			L.resolveDefault(callLuciVersion(), { revision: _('unknown version'), branch: 'LuCI' }),
+			uci.load('luci')
 		]);
 	},
 
+	oneshot: function(data) {
+		var boardinfo = data[0];
+		var check_upgrades = uci.get_first('luci', 'core', 'check_firmware_version') || false;
+		console.log(check_upgrades)
```
systemcrash, line 75: drop

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
```diff
@@ -18,17 +20,85 @@ var callSystemInfo = rpc.declare({
...
+				}
+				for (var j = 0; j < data.profiles[i].supported_devices.length; j++) {
+					if (data.profiles[i].supported_devices[j] == boardinfo.board_name) {
+						return true;
+					}
+				}
+			}
+		})
+		.catch(error => {
+			console.error('Failed to fetch profile information:', error);
```
systemcrash, line 54: consider adding more context to the error. A user who doesn't know about this feature might be confused by this err string, even though it's a console message.

systemcrash:
Consider using `ui.addNotification()` with perhaps a 60 second timeout,  instead of  `ui.showModal()`. Having a modal that a user must interact with to reach the interface causes friction (e.g. when a user logs in and perhaps needs to take care of something else more pressing). 

```js
// ui.addNotification(a, message, timeout, severity)
ui.addNotification(null, E('p', _('A new firmware is available.')), 60000, 'info');
```

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
```diff
@@ -18,17 +20,85 @@ var callSystemInfo = rpc.declare({
...
 			L.resolveDefault(callSystemInfo(), {}),
-			L.resolveDefault(callLuciVersion(), { revision: _('unknown version'), branch: 'LuCI' })
+			L.resolveDefault(callLuciVersion(), { revision: _('unknown version'), branch: 'LuCI' }),
+			uci.load('luci')
 		]);
 	},
 
+	oneshot: function(data) {
+		var boardinfo = data[0];
+		var check_upgrades = uci.get_first('luci', 'core', 'check_firmware_version') || false;
```
efahl, line 74: I'm not sure but I think that should be `'main'` and not `'core'`, assuming the js versions work like the cli tool:
```
$ uci show luci.core
uci: Entry not found

$ uci show luci.main
luci.main=core
luci.main.lang='auto'
luci.main.mediaurlbase='/luci-static/openwrt2020'
luci.main.resourcebase='/luci-static/resources'
luci.main.ubuspath='/ubus/'
```

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
efahl: Or maybe "release candidate"? since I think that's the only thing ever to populate the `upcoming_release` field in .versions.json...

hnyman:
Is this forced to all users? Also for those who regularly build their own builds every few days...

I am wondering if this should either 
* be non-default, a small LuCI package that is included in the release branch buildbot images but would not get be default into private builds, or
* be called via a menu item (so that it is not in the front page).

In many OEM firmwares that I have seen, the upgrade check/notice is in a system admin menu, not on the front page.

(And hopefully the data is not polled by every 5 seconds)

systemcrash:
> (And hopefully the data is not polled by every 5 seconds)

At most once a week is optimistic, once a month preferred (cronjob trigger?). But that's probably why the function is a oneshot. I suppose it could have a counter, and disable itself after a few tries, only to be activated after the next upgrade / reboot.

aparcar:
It runs every time you open the status page and does nothing in the background, why should it anyway. 

It's not enabled by default but could be if users want that. A modal could open once and ask the user if they want to automatically check for upgrades when opening the web interface. 

Additionally we could add a button somewhere to manually switch.

efahl:
> Additionally we could add a button somewhere to manually switch.

Putting a "Don't show me again" button right on the modal dialog would probably be pretty easy, it could just do `uci.set('luci', 'main', 'check_firmware_version', 0)` like is done at https://github.com/openwrt/luci/blob/master/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js#L270

hnyman:
> Putting a "Don't show me again" button right on the modal dialog would probably be pretty easy, 

I feel that any automatic call-to-home should be opt-in, not something to be opted out. By default, the OpenWrt routers are currently not contacting OpenWrt servers before the user selects some action like opkg/APK that triggers the contact.

That "no automatic call-home" was the OpenWrt stance a few years ago, when there was discussion about collecting usage stats, device counts etc.

aparcar:
This is out of the question, it'd always be an opt in feature. 

I'll see how to implement that and update this PR accordingly

efahl:
It's already opt-in.  The value for `luci.main.check_firmware_version` in undefined, so it results in false.  You have to make an effort to explicitly add it to `/etc/config/luci` to turn it on...

richb-hanover:
A couple thoughts:

> It's already opt-in. The value for `luci.main.check_firmware_version` is undefined...

We could use this to our advantage to handle the "startup case". When that value is undefined, it could be a trigger for LuCI to pop up the "Opt in/Opt out" window. We could treat the Yes and No answers in the obvious way, and treat a click on "ask me later" or simply dismissing the window by leaving it undefined. (We would probably implement similar behavior for the CLI as well...)

Once we get permission, the question is "when does the router check?" I am reluctant to queue up any kind of cron job because it creates a permanent load on our servers. 

My inclination would be to (both) request opt-in/opt-out and check for updates when there's interesting activity in LuCI or the command line, perhaps at login. That way we can be certain that there's a human present who can view and act on the "New Version" notice.

efahl:
> perhaps at login.

Yeah, that's the current behavior.  You log in, it pops up the notice.  I've had this PR running on my main router (23.05.5) for a couple weeks and it simply shows the dialog every morning when I log in for status.  It's pretty non-intrusive...

richb-hanover:
More thoughts:

* What about Snapshot versions? I think they should always be ignored. If you're running a snapshot, you are almost certainly aware that there's a new snapshot available now (or after you wake up tomorrow).

* What about Release Candidates? I think they _should always_ be checked. If you're running a RC, you likely would like to know about a new one.

* The "version update check request" might contain the current version. I realize that this gets close to (or goes over) the line about not collecting user data. (Could we stand another flag to say, Don't supply my current version in the request?)

efahl:
Currently snapshots are ignored by this, you need to go to ASU and do a "check for update" to get the check with those.

RCs are included as part of the `latest` list in https://downloads.openwrt.org/.versions.json, so are treated the same as a regular upgrade.

The download URL includes both the current version and target/subtarget of the device, which is the same as when doing ASU's "check for newer", so if you were already doing that, this won't leak any more information than you already did.

aparcar:
To move things forward I'd take the logic from this PR and add it to `luci-app-attendedsysupgrade`. Instead of messing with the system stuff we should use a *status include* as shown [here](https://github.com/openwrt/luci/blob/master/applications/luci-app-mwan3/htdocs/luci-static/resources/view/status/include/90_mwan3.js)

systemcrash:
Given that we don't live in a perfect world, a link to the current release notes would be my recommendation (firmware selector does not have this prominently displayed anywhere). Rare, but sysupgrades can result in a brick.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
```diff
@@ -18,17 +20,85 @@ var callSystemInfo = rpc.declare({
...
+	const profile_url = 'https://downloads.openwrt.org/releases/%s/targets/%s/profiles.json'.format(new_version, boardinfo.release.target);
+	return fetch(profile_url)
+		.then(response => response.json())
+		.then(data => {
+			// special case for x86 and armsr
+			if (Object.keys(data.profiles).length == 1 && Object.keys(data.profiles)[0] == "generic") {
+				return true;
+			}
+
+			for (var i = 0; i < data.profiles.length; i++) {
```
GeorgeSapkin, line 42: You can rewrite this using `for...of` to ensure you go through every element and don't go out of bounds:

```js
for (const profile of data.profiles) {
}
```

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
```diff
@@ -18,17 +20,85 @@ var callSystemInfo = rpc.declare({
...
+				if (data.profiles[i].profile == boardinfo.board_name) {
+					return true;
+				}
+				for (var j = 0; j < data.profiles[i].supported_devices.length; j++) {
```
GeorgeSapkin, line 46: Ditto here:

```js
for (const device of supported_devices) {
}
```

systemcrash:
I can create a PR or go ahead and commit: I have something which more or less completes this PR, as a (asu) one-shot status include, and which includes a preference checkbox whether to check for updates on the main status page.

( just ignore for the moment that the versions are identical here - the notification fills the table with the upgrade properties like kernel version etc )


Screenshot 2025-04-08 at 19 06 45


Screenshot 2025-04-08 at 18 54 42

GeorgeSapkin:
Does it make sense to show the redundant info like model, arch, etc. that's not going to change in an update? The notification looks too much like the status page right now and it took me some time to even notice it.

systemcrash:
> Does it make sense to show the redundant info like model, arch, etc. that's not going to change in an update? The notification looks too much like the status page right now and it took me some time to even notice it.

We can possibly lose some of them - they are just properties of the upgrade. ( We don't yet handle the situation where a device changes 'platform' (as has previously happened), but we might, although that's an edge case that likely won't be upgradeable anyway). The various other dialogue classes have colours which either obliterate the URLs or stab your eyes with a bright colour. I opted for the less jarring 'notice', as opposed to e.g. 'warn' or 'info'.

systemcrash:
info:

info

warn:

warn

systemcrash:
We can trim it down to the most relevant elements:

Screenshot 2025-04-08 at 19 31 05

GeorgeSapkin:
I'd say _info_ is a natural fit here.

Perhaps values could be shown as some kind of a diff in two columns, like:

||Current|Updated|
|-|-|-|
|Firmware Version|24.10.0|24.10.1|

systemcrash:
> I'd say _info_ is a natural fit here.
> 
> Perhaps values could be shown as some kind of a diff in two columns, like:
> 	Current 	Updated
> Firmware Version 	24.10.0 	24.10.1
> 

I like this idea. Like so?

Screenshot 2025-04-08 at 19 48 58

GeorgeSapkin:
Yes, this looks more useful.

What happens when manual changes are necessary, e.g. resizing a partition or swconfig to DSA switch?

systemcrash:
> What happens when manual changes are necessary, e.g. resizing a partition or swconfig to DSA switch?

Out of the scope of this "new firmware notification" - but @efahl knows a bit better: basically you lose your config. The update info is still relevant, however.

systemcrash:
Closed by f746a52979141df8ffbf7c3774002fead7817724

PR #7600, opened 2025-02-02
luci-app-example: add missing dependencies

GeorgeSapkin:
- [x] Tested on: ath79/generic, OpenWrt 24.10.0-rc7 r28417-daef29c75d on TP-Link Archer C7 v2 :white_check_mark:
- [x] \( Preferred ) Mention: @andibraeu @cricalix 
- [x] Description:
    `luci-app-example` that's installed from 24.10.0-rc7 repos has non-working RPC examples due to missing dependencies used in `applications/luci-app-example/root/usr/libexec/rpcd/luci.example`. This PR adds the required dependencies to the app's `Makefile`.

### DIFF
--- a/applications/luci-app-example/Makefile
+++ b/applications/luci-app-example/Makefile
@@ -4,7 +4,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=LuCI example app for js based luci
-LUCI_DEPENDS:=+luci-base
+LUCI_DEPENDS:=+lua +luci-base +luci-lib-jsonc +luci-lib-nixio +luci-lua-runtime
 LUCI_PKGARCH:=all
 
 PKG_LICENSE:=GPL-2.0


hnyman:
The example app is supposed to be JavaScript based, so without lua.
Adding lua related dependencies looks wrong.

GeorgeSapkin:
> The example app is supposed to be JavaScript based, so without lua. Adding lua related dependencies looks wrong.

The RPC examples (RPC Array Example, RPC JSONMap Table Example and RPC JSONMap Typed Example) call the mentioned lua script in `applications/luci-app-example/root/usr/libexec/rpcd/luci.example` and they don't work without the lua dependencies.

systemcrash:
Then this example app is outdated, and if it's to be an example of what openwrt currently promotes, then those calls should be (written in) ucode. lua is slowly being excised from openwrt. So the only thing for it is to remove anything lua based from the examples.

GeorgeSapkin:
@systemcrash I gave it a shot rewriting the lua part in ucode in a new PR: #7602

PR #7602, opened 2025-02-03
luci-app-example: rewrite RPC side using ucode

GeorgeSapkin:
- [x] Tested on: ath79/generic, OpenWrt 24.10.0-rc7 r28417-daef29c75d on TP-Link Archer C7 v2 :white_check_mark:
- [x] \( Preferred ) Mention: @andibraeu @cricalix @systemcrash
image
- [x] Description:
    `luci-app-example` that's installed from 24.10.0-rc7 repos has non-working RPC examples due to missing lua dependencies used in `applications/luci-app-example/root/usr/libexec/rpcd/luci.example`. According to @systemcrash [0], _lua is slowly being excised from openwrt_. This PR rewrites the RPC side of the example app from lua to ucode, without touching the UI too much.

[0]: https://github.com/openwrt/luci/pull/7600#issuecomment-2629595461


applications/luci-app-example/root/usr/share/rpcd/ucode/example.uc
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env ucode
```
systemcrash, line 1: Technically, if we're to use the shebang and be able to call this uc file externally, set the +x bit on it also.

GeorgeSapkin: Do ucode files under rpcd actually need a shebang? I can only find [one example](https://github.com/openwrt/luci/blob/0f743ad0f29a56704be8d13fb6d52c3617a7918a/applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc#L1) in luci-app-ddns which has it, but all ucode files outside of rpcd don't.

systemcrash:
Merged. Thanks @GeorgeSapkin

PR #7610, opened 2025-02-07
luci-app-attendedsysupgrade: apply changes to package list

efahl:
Utilize the 'package_changes' from the ASU server to update the package list for the build request.

Links: https://github.com/openwrt/asu/commit/2bff208387d69a5a1723032a45e672650c17de62

---
@aparcar @systemcrash Please critique these changes thoroughly, my JS experience is grossly outdated and I'm not at all confident about the stylistic choices I've made.  The new function feels solid to me (except for some missing error checking), but the call sequence at the bottom seems kind of ugly.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -78,6 +78,44 @@ return view.extend({
 	request_hash: '',
 	sha256_unsigned: '',
 
+	applyPackageChanges: async function(packages, data, firmware) {
+		const overview_url = `${data.url}/api/v1/overview`;
+		const revision_url = `${data.url}/api/v1/revision/${firmware.version}/${firmware.target}`;
+
+		let changes, target_revision;
+
+		await Promise.all([
+			request.get(overview_url).then(
+				(response) => {
+					let json = response.json();
+					changes = json.branches[get_branch(firmware.version)].package_changes;
+				},
+				(failed) => {
+					ui.addNotification(null, E('p', _(`Get overview failed ${failed}`)));
+				}
+			),
+			request.get(revision_url).then(
+				(response) => {
+					target_revision = get_revision_count(response.json().revision);
+				},
+				(failed) => {
+					ui.addNotification(null, E('p', _(`Get revision failed ${failed}`)));
+				}
+			),
+		]);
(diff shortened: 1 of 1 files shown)


applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -78,6 +78,44 @@ return view.extend({
...
+		let changes, target_revision;
+
+		await Promise.all([
+			request.get(overview_url).then(
+				(response) => {
+					let json = response.json();
+					changes = json.branches[get_branch(firmware.version)].package_changes;
+				},
+				(failed) => {
+					alert(`Get overview failed ${failed}`);
```
aparcar, line 94: Please use ui.notification or however it's called.

aparcar:
Seems to work, amazing

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
efahl: Done, `ui.addNotification` seems to be the thing other packages use.

efahl:
> amazing

I know, right?  😂  Took me like 50 iterations reworking all that Promises.all and contents to get something that worked.

systemcrash:
Looks legit and clean. If it works - even better 😁

aparcar:
It should be backport to 24 and 23

efahl:
> If it works - even better 😁

Alas, it does not. 😞  I just found a bug, I used `firmware.version`, i.e., the old one, rather than the update-to version when grabbing the revcode and `package_changes`.  Working on it...

efahl:
> It should be backport to 24 and 23

24 is an easy cherrypick as we've kept it current with main.

How should I handle 23?  It's way behind and has a ton of diffs, but it all appears to work if I just copy it to my 23.05.5 test box.  Does a cherry pick work on all that or is there some other way to get the changes merged?

hnyman:
cherry-pick PR including the bug fix from #7611 (?) sounds like the natural approach for 24.10.

23.05 looks more problematic, as there are so many commits in master that have not been backported to 23.05 already earlier. Trying to backport only the latest commit might lead into new problems.

aparcar:
I'd backport it all, the state of the 23er luci app is unable to update to 24.10

PR #7611, opened 2025-02-08
luci-app-attendedsysupgrade: supply proper version to package changes

efahl:
Fix bug in applyPackageChanges, it was using the installed version as the target, resulting in no changes being applied.

@aparcar


hnyman:
So, this is the new fix you mention in #7610 ?
Tested enough for merge?

efahl:
Yes, this should be solid.  Tested on a couple devices a dozen times and works as expected for me.

PR #7613, opened 2025-02-09
Fix package-manager-call

kiddin9:


systemcrash:
@kiddin9 what problem does this fix?

hnyman:
APK is set to be called in the not-apk if/else branch... (Meaning opkg)
It is also supposed to be additive in while statement for opkg and change "cmd", but the current code is actually pretty much no-op as it sets the non-used ipkg_bin variable.

That seems wrong also to me, so I flagged this as review  for @ansuel who authored the big change for supporting also apk. 

The statement has no role in master with APK, but the fix might be really needed in 24.10

hnyman:
Applied in fixed form as eb422f57

Issue #7643, opened 2025-02-22
luci-proto-vxlan: wrong section type for peers


alaviss:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### Actual behaviour

The peer is added with `config vxlan_peers`. Potential culprit: https://github.com/openwrt/luci/blob/dbf2c4272a503e9db5ae23b193cb1d62a54b56bc/protocols/luci-proto-vxlan/htdocs/luci-static/resources/protocol/vxlan.js#L182

### Expected behaviour

The peer is added with `config vxlan_peer` (without `s`). This is expected by the vxlan configuration script: https://github.com/openwrt/openwrt/blob/6936deac67e5f0d950bcc26e56bc90b15b12db12/package/network/config/vxlan/files/vxlan.sh#L106

### Steps to reproduce

1. Create a new VXLAN interface
2. Add a peer in the "Peers" tab

### Additional Information

```shell
{
        "kernel": "6.6.77",
        "hostname": "node-01",
        "system": "MediaTek MT7621 ver:1 eco:3",
        "model": "ASUS RT-AX53U",
        "board_name": "asus,rt-ax53u",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "SNAPSHOT",
                "firmware_url": "https://downloads.openwrt.org/",
                "revision": "r28847-e60729c720",
                "target": "ramips/mt7621",
                "description": "OpenWrt SNAPSHOT r28847-e60729c720",
                "builddate": "1739914826"
        }
}
```

### Relevant log output

```shell

```

dxdxdt:
The typo in vxlan6 has not been patched. Ditto in pr #7646

tkkost:
This error has not been patched yet.
Or I understand it wrong. My configuration doesn't work with 'config vxlan_peers'. And it works with 'config vxlan_peer', without 's' . OpenWRT 24.10.2

tkkost:
OpenWRT 24.10.4. This error has not been patched yet. The peer is added with 'config vxlan_peers'.

systemcrash:
1f2929fb49ada94c8a6aa268df048ce3911ccb4f picked in July

tkkost:
> [1f2929f](https://github.com/openwrt/luci/commit/1f2929fb49ada94c8a6aa268df048ce3911ccb4f) picked in July

[...]

PR #7646, opened 2025-02-22
luci-proto-vxlan: fix option name: vxlan_peers to vxlan_peer

stokito:
Fixes: #7643

### DIFF
--- a/protocols/luci-proto-vxlan/htdocs/luci-static/resources/protocol/vxlan.js
+++ b/protocols/luci-proto-vxlan/htdocs/luci-static/resources/protocol/vxlan.js
@@ -179,7 +179,7 @@ return network.registerProtocol('vxlan', {
 		}
 		catch(e) {}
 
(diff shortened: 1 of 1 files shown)


dxdxdt:
There's the same typo copied and pasted in vxlan6 as well. Would you be able to fix that as well? It's just a single character change.

https://github.com/openwrt/luci/blob/master/protocols/luci-proto-vxlan/htdocs/luci-static/resources/protocol/vxlan6.js#L182

systemcrash:
Fixed in 1352d75f1aeb6f9a6745596361261667499f30ac - picked to 24.10 in 1f2929fb49ada94c8a6aa268df048ce3911ccb4f

tkkost:
OpenWRT 24.10.4. This error has not been patched yet. The peer is added with 'config vxlan_peers'.

systemcrash:
OK - there were two really similar commits. Fixed in 28816ce1345c6b435c4ebade6758b55e718fd9f5

tkkost:
It's fixed with a luci-proto-vxlan version 25.328.04704~28816ce

Issue #7326, opened 2024-10-13
Cannot read properties of null (reading 'remove')


jura43:
I fresh installed open-wrt on my Asus AX6000 router. What I configured in DHCP and DNS settings is deleted default instance and created new one with name work. After that I am not able to open DHCP and DNS settings.

## Steps to reproduce:
1. go to: Network → DHCP and DNS 

## Actual behavior:
![image](https://github.com/user-attachments/assets/b4efac27-b2a8-4dec-a9ae-e2330c3c6203)

## Additional Information:
```
root@OpenWrt:~# cat /etc/config/dhcp

config dhcp 'lan'
        option interface 'lan'
        option start '100'
        option limit '100'
        option leasetime '12h'
        option dhcpv4 'server'
        option dhcpv6 'server'
        option ra 'server'
        list ra_flags 'managed-config'
        list ra_flags 'other-config'

config dhcp 'wan'
        option interface 'wan'
        option ignore '1'

config odhcpd 'odhcpd'
        option maindhcp '0'
        option leasefile '/tmp/hosts/odhcpd'
        option leasetrigger '/usr/sbin/odhcpd-update'
        option loglevel '4'

config dhcp 'Work'
        option start '100'
        option limit '200'
        option leasetime '12h'
        list server '1.1.1.1'
        option rebind_protection '0'
        option localservice '0'
        option interface 'Work'
        list dhcp_option '6,192.168.20.1'

config dhcp 'Guest'
        option interface 'Guest'
        option start '100'
        option limit '200'
        option leasetime '12h'
        list dhcp_option '6,1.1.1.1,1.0.0.1'
```

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='SNAPSHOT'
DISTRIB_REVISION='r27751-f8c22c9bff'
DISTRIB_TARGET='mediatek/filogic'
DISTRIB_ARCH='aarch64_cortex-a53'
DISTRIB_DESCRIPTION='OpenWrt SNAPSHOT r27751-f8c22c9bff'
DISTRIB_TAINTS=
```

jura43:
I think I am able to reproduce this.
I create interface, enable DHCP server, save. DHCP and DNS page still working. But if I edit DHCP settings in interface windows for example I change range of IP addresses and save. Then DHCP and DNS stop working.
Here is config before and after editing DHCP in interface menu.
```
config dnsmasq
        option domainneeded '1'
        option boguspriv '1'
        option filterwin2k '0'
[...]

PR #7652, opened 2025-02-25
luci-base: forms.js safe call on this.map.data.remove()

stokito:
Related: #7326

It may not fix, but at least it shouldn't harm


systemcrash:
How does it behave if you install a test instance and as described in the linked issue, delete the main instance.

PR #7653, opened 2025-02-26
luci-theme-material: change dropdown button for better visibility

Chris1189:
My PR is a slight improvement, where the upstream version works fine, but my downstream version benefits heavily from it, so made the request:

On theme material it is barely possible to see the possible options in
the dropdown from my version of luci-theme-material. 
If the dropdown is switched to `--black-color` (EDIT: not `--gray-color-high`) it improves distinguishing the colors in my GUI.

*Before downstream version*
image

*After downstream version*
image

In the upstream version the dropdown is fine and also distinguishable with my modified version:

*Before upstream version*
image
*After upstream version*
image

On both versions (upstream/downstream) the button is no ideal as a display so I changed that also:
*Before*
image
*After*
image

- [x] Tested on: (x86, openwrt 24.10, firefox) :white_check_mark:
- [x] Screenshot or mp4 of changes:

### DIFF
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -1586,6 +1586,10 @@ body:not(.Interfaces) .cbi-rowstyle-2:first-child {
 	color: var(--main-menu-color);
 }
 
+.cbi-dropdown.btn.cbi-button.cbi-button-positive.important > ul > li[display] {
+	color: var(--white-color);
+}
+
 .cbi-dropdown > ul > li[display],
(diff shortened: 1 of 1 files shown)


systemcrash:
These changes look acceptable. Can you apply an identical treatment to the other themes ( so they’re synchronised )?

Chris1189:
I could do that. The only thing is that for example bootstrap looks already good to me:
image

Also openwrt theme looks distinguishable:
image

Openwrt-2020:
image

systemcrash:
In which case, this looks OK. Any opinions from @dannil @stokito @hnyman ?

PR #7670, opened 2025-03-07
treewide: fix notification call function arguments

feckert:
I was wondering why no CSS class was set for a notification. The problem was that the call arguments of the addNotification function have changed. This [commit](https://github.com/openwrt/luci/commit/53e36e3293bee2ea2ce0fbc1250074c44cfe8335) changed the API of the function.
I have now adjusted this. I have not tested everything as I do not use all applications. 
@systemcrash can you have a look at this and cross check my changes.

### DIFF
--- a/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
+++ b/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
@@ -174,9 +174,11 @@ return view.extend({
 	},
 
 	render: function(data) {
-		ui.addNotification(null, E('p', [
-			_('The LuCI ACL management is in an experimental stage! It does not yet work reliably with all applications')
-		]), 'warning');
+		ui.addNotification(null,
+			E('p', [_('The LuCI ACL management is in an experimental stage! It does not yet work reliably with all applications')]),
+			0,
+			'warning'
+		);
 
 		var has_uhttpd = data[0],
 		    known_unix_users = {};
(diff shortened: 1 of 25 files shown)


feckert:
@systemcrash if you have no objections. I would also backorted this directly to openwrt-24.10.

feckert:
That is correct and breaks downstream, but this is already in the stable branch owrt-24.10!
But I  think the feature is not so bad.
Maybe we can add the timeout argument as the last argument in the function?

jow-:
Does not matter if it is in the stable branch or not, it broke the API and apparently nothing was adapted to the new one, so it should be simply backed out.

Adding as last argument is not possible since the function uses a rest argument for CSS classes.

A proper solution would be adding a new function like `flashNotification()` or `addNotificationEx()` which has a different parameter signature.

feckert:
@jow- Sounds good. I will revert API change.
@systemcrash as suggested by @jow if you need this feature you should add a new function `flashNotificition` function.

feckert:
@stokito As jow has already [mentioned](https://github.com/openwrt/luci/pull/7670#issuecomment-2706333125),

1. Add a new function if this feature is required.

or

2. Do a code change to check if classes[0] is an int

I think the second suggestion is not an option from my point of view, as the change would complicate the code.
Also I can see that no one is using this feature upstream.

systemcrash:
I'll take care of 1. There were only a few implementations so far for 2.

PR #7686, opened 2025-03-18
luci-base: ui: filebrowser parent dir fix

realn:
Fix for a bug in file browser parent directory button.

When clicking on a parent directory link in a file browser control (ex. selecting a certificate in OpenVPN app) it will send an incorrect directory request that will end in an empty list.

Main problem is with how current directory is split by splitPath function: 
`/etc/config -> ['', 'etc', 'config']`

and reassembled:
```
for(let i = 0; i < dirs.length; i++) {
cur = cur ? ``${cur}/${dirs[i]}`` : dirs[i];

```
this will result with path strings missing the root directory slash: `etc/config`
sending this into the request will produce an invalid directory -> empty list.

A bonus bug is with how ?? operator treats empty strings: 
```
let foo = '';
return foo ? 'full' : 'empty'; //  -> returns 'empty' 
return foo ?? 'empty';         //  -> returns '' not 'empty'
```

The easiest fix (without completely rewriting the whole algorithm) is to just always add the missing slash before sending the request.

- [ :white_check_mark: ] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch 
- [ :white_check_mark: ] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [ :white_check_mark: ] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [N/A] Incremented :up: any `PKG_VERSION` in the Makefile
- [ :white_check_mark: ] Tested on: ARMv8 Processor rev 1, 25.073.41994~96c3654, LibreWolf(Firefox)/Chromium
- [ :white_check_mark: ] Description: (describe the changes proposed in this PR)

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -3084,7 +3084,7 @@ const UIFileUpload = UIElement.extend(/** @lends LuCI.ui.FileUpload.prototype */
 				i ? ' » ' : '',
 				E('a', {
 					'href': '#',
-					'click': UI.prototype.createHandlerFn(this, 'handleSelect', cur ?? '/', null)
+					'click': UI.prototype.createHandlerFn(this, 'handleSelect', '/' + cur, null)
 				}, dirs[i] != '' ? '%h'.format(dirs[i]) : E('em', '(root)')),
 			]);
 		}


systemcrash:
The problem here is the openvpn app is still lua based. So the problem you've encountered is likely not based in JS - but in lua. See also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

realn:
@systemcrash ok, sorry, i'm a little confused - it's not JS, but lua? How?
It's a ui.js source file and Lua doesn't even have an ?? operator. And the same bug also exists on luci-app-filebrowser that uses ui.js from luci-base. Did i misunderstood the connections?

systemcrash:
Your fix appears to work for JS, for the problem case you mention, but it is then incorrect for those apps that specify a start path e.g.: `/tmp` which would result in path `//tmp` being sent in the rpcd file list call.

realn:
@systemcrash hmm, i don't see how this change could do that. Given path is always split and reassembled, so it always ends up without a root slash - i tested it. Could you give me a hint, where to find a case that you mentioned?

systemcrash:
Inspect the console when using a JS app using the filebrowser component. e.g. Dynamic DNS.

realn:
@systemcrash thanks! Found the case, and tested it.

So, when the `root_directory` property for the control is set, it works as intended, without the bug:
ddns_withroot_nofix_01
ddns_withroot_nofix_02
But the moment we replace value of `root_directory` with '/' or empty - it breaks as in openvpn app.
ddns_noroot_nofix_01
ddns_noroot_nofix_02
ddns_noroot_nofix_03
But when I apply my fix, it works again:
ddns_noroot_fix_01
Also it works with `root_directory` filled - no problems found here.
ddns_withroot_fix_01

systemcrash:
Closed by 5e39cf04c41e5a656e612aa36f75ce47211defee.

systemcrash:
Thanks for trying, in any case, @realn. This has been picked to the 24 branch also.

realn:
Thanks! This bug was very unnerving, so all it matters, that it's now officially fixed :)

systemcrash:
HTH. Your input is welcome. Let us know if you find anything else, in future.

Issue #7687, opened 2025-03-18
luci-app-attendedsysupgrade: xiaomi ax3600 upgrade (24.10.0) error


migueldeblas:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### screenshots or captures

Error in luci-app-attendedsysupgrade
I have a Xiaomi AX3600 IoT
When I try to upgrade (attendedsysupgrade) from 23.05.5. to 24.10.0 Luci does not work. 

The navigation console error:

![Image](https://github.com/user-attachments/assets/35771518-6bd2-454e-b399-435355b01bab)

### Actual behaviour

When I try to upgrade (attendedsysupgrade) from 23.05.5. to 24.10.0
Luci does not work and does not show any message or error.

### Expected behaviour

Generates the firmware and shows a message or error.

### Steps to reproduce

Luci >  System > Attended Sysupgrade > Search for firmware upgrade > 
Select firmware version: 24.10.0
Board Name / Profile: xiaomi,ax3600
Click "Request firmware image"

### Additional Information

```shell
{
        "kernel": "5.15.167",
        "hostname": "AX3600",
        "system": "ARMv8 Processor rev 4",
        "model": "Xiaomi AX3600",
        "board_name": "xiaomi,ax3600",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "23.05.5",
                "revision": "r24106-10cc5fcd00",
                "target": "ipq807x/generic",
                "description": "OpenWrt 23.05.5 r24106-10cc5fcd00"
        }
}
```

### Relevant log output

```shell

```

systemcrash:
Have you updated sysupgrade first?

migueldeblas:
Yes, All packages upgraded (via Luci).

systemcrash:
@efahl does this look like something you recognise?

efahl:
Yes, I'll take a look.

efahl:
Displayed error is due to insufficient error handling right here: https://github.com/openwrt/luci/blob/master/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L101 .  I'll fix both of those functions to be a bit more robust.
```
$ curl -s https://sysupgrade.openwrt.org/api/v1/revision/24.10.0/ipq807x/generic
{"detail":"Failed to fetch revision for 24.10.0/ipq807x/generic","status":404}

[...]

PR #7691, opened 2025-03-19
luci-app-attendedsysupgrade: add more robust handling of downloads

efahl:
Check various aspects of the overview and revision downloads from the sysupgrade site, giving the user better error messages when things go wrong.

Partially addresses issue #7687 with respect to diagnosis, but not root cause.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -87,23 +87,22 @@ return view.extend({
 		let changes, target_revision;
 
 		await Promise.all([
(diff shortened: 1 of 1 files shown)


systemcrash:
I think that looks ok. 

Input from other maintainers?

efahl:
I found an issue (fixed in that last push), if you know Python it's like the old "bare except" problem.

In my first version, the top-level catch was silent, relying on the called function `applyPackageChanges` to generate any messages.   This would only happen for the two handled exceptions, so anything outside of those two would be just ignored and the modal dialog would be closed without any indication as to what happened.

Now the top-level caller shows the error message, regardless of its source or form.  For example, if the for-loop were to fail somehow, then a message is displayed prior to closing the modal dialog and the user is given a message as to why.

migueldeblas:
Sorry. I upgraded it.

PR #7712, opened 2025-04-04
luci-app-antiblock: Gateway selection bug fixed

karen07:
1) Changed from "widgets.NetworkSelect" to "widgets.DeviceSelect"


karen07:
Thanks @systemcrash

PR #7721, opened 2025-04-12
luci-app-statistics: add UseLabels option for sensor

rhjdvsgsgks:
- [x] Tested on: on 24.10 . by editing /usr/share/luci/statistics/plugins/sensors.json directly
- [ ] \( Preferred ) Screenshot or mp4 of changes: no visual changes
- [x] Description: add a new option from upstream. this may useful when sensor have multiple input on a same type (eg. sfp have both tx/rx power)

also https://github.com/openwrt/packages/pull/26308


systemcrash:
See comments on https://github.com/openwrt/packages/pull/26308

Issue #7703, opened 2025-03-27
luci-theme-*: Page `<title>` incomplete since OpenWrt 23.05


Self-Hosting-Group:
Since OpenWrt 23.05 the called page (menu name) is missing in the page title element, `${node.title}` is not set.

When correcting this, it may be better for visibility to show the menu name first, then the hostname, and use `|` as the separator.

https://github.com/openwrt/luci/blob/4207c2c5d51470fd273441f92b0a2c104441454c/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut#L21

Currently missing title:
- `<title>$HOSTNAME - LuCI</title>`

But should be displayed as:
- `<title>$HOSTNAME - $TITLE - LuCI</title>`

Title suggestions (examples):
1. `<title>DHCP and DNS | $HOSTNAME | LuCI</title>`
2. `<title>DHCP and DNS | $HOSTNAME LuCI</title>` (good with OpenWrt hostname)
3. `<title>DHCP and DNS | LuCI | $HOSTNAME</title>`
4. `<title>DHCP and DNS | LuCI/$HOSTNAME</title>` (good with OpenWrt hostname)
5. `<title>DHCP and DNS | $HOSTNAME</title>`
6. `<title>DHCP and DNS | LuCI</title>`

PR #7731, opened 2025-04-17
luci-theme-*: Add missing menu name to page `<title>`

Self-Hosting-Group:
Close: #7703

### DIFF
--- a/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
+++ b/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
@@ -18,7 +18,7 @@
 <html lang="{{ dispatcher.lang }}" {{ darkpref ? `data-darkmode="${darkpref}"` : '' }}>
 	<head>
 		<meta charset="utf-8">
-		<title>{{ striptags(`${boardinfo.hostname ?? '?'}${node ? ` - ${node.title}` : ''}`) }} - LuCI</title>
+		<title>{{ striptags(`${dispatched ? `${dispatched.title} | ` : ''}${boardinfo.hostname ?? '?'}`) }} | LuCI</title>
 		{% if (!darkpref): %}
 			<script>
 				var mediaQuery = window.matchMedia('(prefers-color-scheme: dark)'),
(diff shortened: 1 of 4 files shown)


systemcrash:
I'm not sure whether this was intentional, but this is a question for @jow- who, if he hasn't answered already, this isn't necessary. It's an "opinionated nice to have(?)". The page URL kind of makes this redundant, no?

Headcrabed:
@jow- Ping

systemcrash:
Merged. Thanks @Self-Hosting-Group

Self-Hosting-Group:
Thanks for merging @systemcrash

Nmrawe:
@systemcrash Menu names in <title> tags are in English only. This breaks i18n a bit.

systemcrash:
Can you show an example?

Nmrawe:
Firefox displays tabs with the current OpenWrt like `Overview | OpenWrt | LuCI` or `Wireless | OpenWrt | LuCI`. The first word on the left should be translated on non-English locales but it's in English. I've noticed this issue while testing 25.12-rc2, but I had to revert to 24.10.5 due to some bugs. The older version always displays `OpenWrt - LuCI` on all tabs. According to my password manager's matching window titles function it used to be `OpenWrt - Overview - LuCI` in some old OpenWrt versions with translated middle words.

Self-Hosting-Group:
When testing the luci-app-upnp PR with an old OpenWrt 22.03, I was surprised to see the menu title in the browser tab. I didn't think I could fix this so easily in the current version (missed in lua->ucode migration). I'm happy now that @Nmrawe has continued with this, and @systemcrash made the strings now translated. Thank you both!

However, I've noticed that the translation adjustments haven't been applied everywhere. `<meta name="application-name"...>` and the Apple equivalent is missing, but only exists in the material theme.

https://github.com/openwrt/luci/blob/9972117cde97e376d5d9373177e232038856d593/themes/luci-theme-material/ucode/template/themes/material/header.ut#L33-L40

But why should these tags only apply to this theme? Could these deprecated tags be removed? Also, the current viewport tag prevents zooming. IMHO, all of the meta tags listed above could be removed (except adapted viewport), and the `| LuCI` could possibly be removed from the title, as it is in the footer with details.

Finally, some suggestions that do not need to be implemented, but could possibly improve usability. Currently, the same title for Firewall/Routing in System is displayed for the entries in Network. This could possibly be made clearer by using `${hostname} | ${title} | ${topmenu}`. I also find that the title->hostname swapping (https://github.com/openwrt/luci/issues/8308) is not always useful. A LuCI UCI option with a template string could be a solution.

systemcrash:
It's there for the taking if you make a PR

PR #7741, opened 2025-04-23
luci-app-banip: Trim() error fix.

GithPK:
Unable to open a banip Overview tab.
Update overview.js for the Trim() error fix.

- [x] Tested on: (ARMv8 Processor rev 4, OpenWrt SNAPSHOT r29276-963d320086, EDGE) :white_check_mark:
- [x] Description: (improved trim() function)

### DIFF
--- a/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
+++ b/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/overview.js
@@ -695,9 +695,9 @@ return view.extend({
 		if (feeds && Object.keys(feeds).length) {
 			o = s.taboption('feeds', form.MultiValue, 'ban_feed', _('Blocklist Feed'));
 			for (let i = 0; i < Object.keys(feeds).length; i++) {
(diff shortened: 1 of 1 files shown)


systemcrash:
@dibdot

dibdot:
> Unable to open a banip Overview tab.
That can only happen with a broken custom feed file and should be fixed in the backend by resetting the custom feed file (as mentioned in the readme). Anyway, the proposed trim fix LGTM  - thanks!

GithPK:
> > Unable to open a banip Overview tab.
> 
> That can only happen with a broken custom feed file and should be fixed in the backend by resetting the custom feed file (as mentioned in the readme). Anyway, the proposed trim fix LGTM - thanks!

Thank you for the feedback. If I knew the trick with the feed reset we would not talk ;-) 
I reset the custom feed and it has started working properly.
I must admit, that the Readme is not so clear in this matter as your clear trick suggestion above. Thank you. 
I hope the fix will limit some user's problems,

PR #7744, opened 2025-04-23
luci-theme-bootstrap: unbound darkmode file editor

JimMatthew:
Add css to make the Unbound file editor have a dark background when using Bootstrap dark mode theme.
Currently the unbound file editor has a yellowish background, even in dark mode. This makes the background dark.
I am not sure if this is the best way to do this but the change is fairly straightforward.

Let me know if this is something you are interested in or if there are any changes you would like made.

Current:
Screenshot 2025-04-22 090611

After change:
Screenshot 2025-04-22 090024

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -2621,3 +2621,7 @@ div.cbi-value var.cbi-tooltip-container,
 [data-darkmode="true"] [data-page="admin-status-realtime-connections"] #view > div > div > div > svg > line[style] {
 	stroke: #fff!important;
 }
+
+[data-darkmode="true"] .cbi-input-textarea {
+  background-color: var(--background-color-high);
+}


systemcrash:
This looks OK - but all of the themes should be updated synonymously.

Ramon00:
i thought only bootstrap had a light/dark mode?

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -2621,3 +2621,7 @@ div.cbi-value var.cbi-tooltip-container,
 [data-darkmode="true"] [data-page="admin-status-realtime-connections"] #view > div > div > div > svg > line[style] {
 	stroke: #fff!important;
 }
+
+[data-darkmode="true"] .cbi-input-textarea {
+  background-color: var(--background-color-high);
```
Ramon00, line 2626: The background color is already switched between light and dark (or well it should be). So why do we need to add another conditional define here? I guess maybe it is better to assign the background color higher up in the (parent) CSS structure?

JimMatthew: I think its because in the unbound package, the background is getting set manually
style>
  textarea
  {
    background-color: #fffff0;
    font-family: monospace;
  }
</style>
https://github.com/openwrt/luci/blob/master/applications/luci-app-unbound/luasrc/view/unbound/css-editing.htm

jow-:
Just remove the style overrides and let unbound use the default styles.

systemcrash:
Can you fix that, @JimMatthew ?

PR #7746, opened 2025-04-23
luci-base: remove bad Unicode on clone button

dwfreed:

This Unicode character is intended to be followed by 2 more characters which are combined. It's entirely unnecessary, so just remove it.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2728,11 +2728,11 @@ const CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection
 
 			dom.append(tdEl.lastElementChild,
 				E('button', {
-					'title': btn_title || _('Clone') + '⿻',
+					'title': btn_title || _('Clone'),
 					'class': 'btn cbi-button cbi-button-neutral',
 					'click': ui.createHandlerFn(this, 'handleClone', section_id, true),
 					'disabled': this.map.readonly || null
-				}, [ btn_title || _('Clone') + '⿻' ])
+				}, [ btn_title || _('Clone') ])
 			);
 		}
 


dwfreed:
Note: this should be backported to 24.10 as well

systemcrash:
There are no bad unicode characters - just the choice of fonts displaying them.

https://www.unicode.org/charts/PDF/U2FF0.pdf
⿻ - U+2FFB
"These are visibly displayed graphic characters, not invisible composition controls."

dwfreed:
No useful fonts implement these characters, and they look like garbage to everybody, but sure, show garbage to everybody

dwfreed:
image

systemcrash:
What do you run, windows 3.11? https://fonts.google.com/?preview.text=%E2%BF%BB

dwfreed:
https://www.fontspace.com/unicode/char/2FFB-ideographic-description-character-overlaid lists a total of 13 fonts that implement the character, none of them appear like they'd be default on any standard installation.  https://www.fileformat.info/info/unicode/char/2ffb/fontsupport.htm lists 6, though 2 of them are fallback fonts.  Your own link only shows the actual character for the Noto fonts for China, Japan, and Korea and related locations, as far as I can tell.  All the rest are the same as my screenshot.

dwfreed:
Using your link, I made one change, restricting the language to English, and then scrolled through 1,599 fonts. This excluded the Noto Chinese, Japanese, and Korean fonts I mentioned.  I saw ***4*** fonts that implemented this character:

* LXGW WenKai Mono TC
* LXGW WenKai TC
* Chocolate Classical Sans
* Cactus Classical Serif

ynezz (approved):
LGTM. That icon button makes the UI inconsistent, breaks the rhythm, there is no spacing between icon and label, the dashed outline is hard to see on dark gray and I believe, that the icon should be before the label. I would add `Fixes: #7754` Git trailer.

Ramon00:
> Using your link, I made one change, restricting the language to English, and then scrolled through 1,599 fonts. This excluded the Noto Chinese, Japanese, and Korean fonts I mentioned. I saw _**4**_ fonts that implemented this character:
> 
> * LXGW WenKai Mono TC
> * LXGW WenKai TC
> * Chocolate Classical Sans
> * Cactus Classical Serif

I cannot say in how many it is implemented, but on my screen it just shows as it should: 
image
(both on android and Windows using chromium Edge).

I did not install any strange fonts or anything. 

On some fonts it may look bad, but lets not over exaggerate the problem. Can we not add a check in the code if the character is defined in the font that is used? In that way everybody is happy

Ramon00:
Seems it is implemented in rather default fonts: https://unicodeplus.com/U+2FFB
image

GeorgeSapkin:
Seems to work here on Firefox 138 on Fedora 42, with no extra fonts installed. Ditto for the latest Firefox on Android 15. Not sure what the issue is. Might look better with an extra space, sure.

Ramon00:
Maybe use something like this to detect if the element is defined: 

```
function isUnicodeCharacterSupported(char, element = document.body) {
    const canvas = document.createElement("canvas");
    const ctx = canvas.getContext("2d");

    const font = window.getComputedStyle(element).font;
    ctx.font = font;

    const referenceWidth = ctx.measureText("\uFFFD").width; //fallback character
    const charWidth = ctx.measureText(char).width;

    return charWidth !== referenceWidth;
}
```
then invoke `isUnicodeCharacterSupported("\u2FFB")`

This checks if the width is the same as the refenence width. Not idiot proof but should work most of the time, if accidentally the width is the same the only consequence is that the symbol is missing (not the end of the world).

dwfreed:
Or, you know, just not use a character that requires special font support and serves no practical purpose...  One solution seems massively simpler than the other.

dibdot:
Let's remove this Unicode character as long as it causes problems for many users and we don't have a better, consistent concept for button labels.

castillofrancodamian:
@systemcrash I suggest moving the "Edit" button back to where it was before adding it. The reason for this suggestion is because it's very easy to press the "Clone" button when you mean to press the "Edit" button.

Issue #7692, opened 2025-03-20
improve memory usage ui


pynixwang:
### What would you like to see in luci?

https://github.com/openwrt/luci/blob/70aa0bea68fae83bc8682fdc62aa198fe3e40634/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/20_memory.js#L35

`Total Available` progress bar  makes me confuse. 
`Used` is enough.

all system monitor use a pattern  `used / total`

juronja:
agree, confuses me every time too

PR #7761, opened 2025-05-02
remove Total Available progress bar

pynixwang:
this bar is useless and confuse users.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/20_memory.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/20_memory.js
@@ -32,7 +32,6 @@ return baseclass.extend({
 		    swap = L.isObject(systeminfo.swap) ? systeminfo.swap : {};
 
 		var fields = [
-			_('Total Available'), (mem.available) ? mem.available : (mem.total && mem.free && mem.buffered) ? mem.free + mem.buffered : null, mem.total,
 			_('Used'),            (mem.total && mem.free) ? (mem.total - mem.free) : null, mem.total,
 		];
 


pynixwang:
#7692

hnyman:
I see no reason to remove the available RAM amount indicator.
What there specifically annoys you?

pynixwang:
@hnyman 

duplicated with Used progress bar

—————————————/————————————————————
Used                                             Available


or 

if Available must be displayed, it just need a number not a progress bar

pynixwang:
Available         100M

hnyman:
> duplicated with Used progress bar

???
The "Total available" is not the whole "Total" amount, but the currently available amount (free + recoverable cache/biffers).

image

> all system monitor use a pattern used / total

Except Linux memory info?

```
root@router6000:~# cat /proc/meminfo
MemTotal:        1010112 kB
MemFree:          771100 kB
MemAvailable:     837556 kB
Buffers:           43544 kB
Cached:            67924 kB
...
```

pynixwang:
@hnyman but Linux memory info does not use `progress bar`

the problem is progress bar

Ramon00:
Why is the progress bar the problem? It is just showing graphically what the number says.

pynixwang:
@hnyman @Ramon00 

sorry, if no one feels this is a problem,  I will close this pr.

juronja:
I am late, but I confirm it confuses me every time because both express similar data. The thing is that there is 'duplicate' info on this Summary page about the ram and only ram, hence it CAN be confusing. For consistency and better ux I would argue that one datapoint could be committed. Like the OP says we are more used to look at 'used X of total X' from other places (OSes, Proxmox, esp chips, etc.). And one datapoint is plenty for a Summary page.

PR #7773, opened 2025-05-12
luci-app-banip: additional ipv6 parse option for feeds

fw23git:
- Tested on: aarch64, 24.10.0, Chrome :white_check_mark:
- Description: Additional feed parse option to parse IPV6 in feeds to /64 /48 /32 subnets, respectively


dibdot:
Is this only relevant for IPv6? 
Anyway, please provide a testing URL with a real-world feed.

Thanks!

fw23git:
this is for converting IPV6 into "conservative", larger subnets, using the presence of the double colon in IPV6 as a hint. I added a rule to conserve predefined subnets in the input.

dibdot:
I tend to reject this PR.
- This new rule only works with the dropv6 feed.
- I think rewriting feeds with changed values is generally problematic and will result in much more support requests

Anyway; I'll check options to bring back the option to add custom rules ... but this will take time, as I don't have time for OpenWrt ATM.

PR #7791, opened 2025-05-28
luci-app-attendedsysupgrade: maintain value of owut.rootfs_size in LuCI app

efahl:
If someone has used owut to install an image with a modified rootfs partition size, then using LuCI Attended Sysupgrade would change it back, causing either a build failure or loss of configuration.

We fix this by simply adding the value from the owut config to the build request (or 'null' if it doesn't exist).  The ASU server then builds the image with the correct rootfs size, avoiding these issues.

This value is not and should not be exposed in the LuCI app's options, as it is quite advanced, very fragile and should only be used by those who have researched the issues involved, typically by reading the owut documentation and testing with owut first.

Tested on x86/64, both main snapshot and 24.10.1, using both sysupgrade.o.o and my local ASU server.


systemcrash:
Should this go into 24.10 also?

PR #7794, opened 2025-05-29
luci-app-attendedsysupgrade: avoid redirects, handle redirect errors

efahl:
When users add an extraneous slash to the server url, then the ASU server redirects GET requests.  But, this causes an error on POST requests as the ASU server does not redirect those, so the LuCI client failed inappropriately with an unhandled error.

Fix both of these by first trimming trailing slashes from the server url as it is loaded to avoid the redirects.  Then, also handle all unexpected responses, instead of just a select few.

Links: efahl/owut@724c990d6cc0


systemcrash:
Thanks @efahl. Pick to 24?

efahl:
Yes, definitely, I had forgotten all about that!

systemcrash:
Done. 4319c1efa9320c467cbdf0094106cba3d8265b38

PR #7800, opened 2025-05-30
luci-base: fix table option does not show in non-Bootstrap themes when it has depends

morytyann:
- [x] Tested on: x86_64, OpenWrt SNAPSHOT, Chrome/Edge/Vivaldi :white_check_mark:

The `hidden` class was being set to the wrong element.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -3779,7 +3779,10 @@ const CBIValue = CBIAbstractValue.extend(/** @lends LuCI.form.Value.prototype */
 				E('div', { 'class': 'cbi-value-description' }, this.description.trim()));
 
 		if (depend_list && depend_list.length)
-			optionEl.classList.add('hidden');
+			if (in_table)
+				optionEl.firstChild.classList.add('hidden');
+			else
+				optionEl.classList.add('hidden');
 
 		optionEl.addEventListener('widget-change',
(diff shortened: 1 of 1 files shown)


systemcrash:
Can you illustrate an example of before/after?

feckert:
Have you checked other themes as well? Do they work after and before your change?

morytyann:
openwrt 2020 before
image

openwrt 2020 after
image

openwrt before
image

openwrt after
image

jjm2473:
I think this PR will break `isActive` and `setActive`(no test):

```javascript
	/**
	 * Test whether the option element is currently active.
	 *
	 * An element is active when it is not hidden due to unsatisfied dependency
	 * constraints.
	 *
	 * @param {string} section_id
	 * The configuration section ID
	 *
	 * @returns {boolean}
	 * Returns `true` if the option element currently is active, otherwise it
	 * returns `false`.
	 */
	isActive(section_id) {
		const field = this.map.findElement('data-field', this.cbid(section_id));
		return (field != null && !field.classList.contains('hidden'));
	},

	/** @private */
	setActive(section_id, active) {
		const field = this.map.findElement('data-field', this.cbid(section_id));

		if (field && field.classList.contains('hidden') == active) {
			field.classList[active ? 'remove' : 'add']('hidden');

			if (dom.matches(field.parentNode, '.td.cbi-value-field'))
				field.parentNode.classList[active ? 'remove' : 'add']('inactive');

			return true;
		}

		return false;
	},
```

Theme bug should fix in themes.

PR #7809, opened 2025-06-05
luci-app-ddns: no implicit interface when ip_source=interface

7Ji:
  - architecture: `aarch64`
  - openwrt version: `OpenWrt 24.10.0 r28427-6df0e3d02a / LuCI openwrt-24.10 branch 25.014.55016~7046a1c`
  - browser: Firefox 139.0.1
  - before this, when `ip_source=interface`, `interface` would be derived from the physical interface: 
  before
  - after this, when `ip_source=interface`, `interface` needs to be set explicitly:
  after

  This makes the DDNS configuration `service.xxx.interface` no longer derived from the physical interface name (duplicated from `ip_interface` to be specific) when `ip_source=interface` and must be set explicitly. This setting is needed when ifup happens in hotplug.d and an updater needs to be spawned. And as such it must use the name in uci.network instead of the physical interface name, so user needs to set it explicitly.  
  Without the fix, users might configure DDNS on a WAN conneciton e.g. `pppoe-wan` from which the `interface` would be set to `pppoe-wan` when it should be either `wan` or `wan_6`, which causes the hotplug.d hook failing to find any configured service during ifup events. To "workaround" the issue that "ddns" "would not start on boot" (which is of course intended as `interface` was not configured correctly), users has been trying to either manually start `ddns` service after boot or remove `boot()` hook from ddns init script to force it start.
  references:
  - ddns-scripts [hotplug.d hook](https://github.com/openwrt/packages/blob/08b4fcd5e6b2ec5853c7eedd548bff0d3f541fbe/net/ddns-scripts/files/etc/hotplug.d/iface/ddns#L6)
  - ddns-scripts [updater script](https://github.com/openwrt/packages/blob/08b4fcd5e6b2ec5853c7eedd548bff0d3f541fbe/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_updater.sh#L135) 
  - ddns-scripts [start_daemon_for_all_ddns_sections() function](https://github.com/openwrt/packages/blob/08b4fcd5e6b2ec5853c7eedd548bff0d3f541fbe/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh#L179)
  - https://forum.openwrt.org/t/ddns-no-start-after-boot/94769
  - https://forum.openwrt.org/t/ddns-do-not-start-when-reboot/98374
  - https://forum.openwrt.org/t/ddns-never-starts-after-reboot/141148
  - [my blog post on this](https://7ji.github.io/networking/2025/06/05/why-openwrt-ddns-does-not-start-on-boot.html)


systemcrash:
OK - I'll allow this change: please document the reasoning in the commit message also. e.g.

With this change, the DDNS configuration service.xxx.interface is no longer derived from the physical interface name (duplicated from ip_interface to be specific) when ip_source=interface, and as a result, the interface must be set explicitly.

7Ji:
@systemcrash 

I've updated the commit message, please kindly check if it's OK : )

systemcrash:
Great. Thanks for the fix @7Ji - great write-up also!

systemcrash:
You can flag this in the forum so other users can try it out. I'll pick this to 24.10.

PR #7813, opened 2025-06-09
luci-app-firewall: replace 'Masquerading' with 'IPv4 Masquerading'

n0toose:
Although the description in the UI explains that this setting affects IPv4 only, a more clear label can help some users (hint: me) know that there the existence of this option does not exclude the existence of the separate 'IPv6 Masquerading' option.

---

CC: @mastaal
First time contributor here - couldn't find a `PKG_VERSION` in the corresponding Makefile.

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
@@ -164,7 +164,7 @@ return view.extend({
 		p[1].default = fwDefaults.getOutput();
 		p[2].default = fwDefaults.getForward();
 
-		o = s.taboption('general', form.Flag, 'masq', _('Masquerading'),
+		o = s.taboption('general', form.Flag, 'masq', _('IPv4 Masquerading'),
 			_('Enable network address and port translation IPv4 (NAT4 or NAPT4) for outbound traffic on this zone. This is typically enabled on the <em>wan</em> zone.'));
 		o.editable = true;
 		o.tooltip = function(section_id) {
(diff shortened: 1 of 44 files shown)


n0toose:
As I don't wish to waste anyone's time, I'm disclosing the following:
- I used `sed` to replace all affected `msgid`'s. I have a degree of familiarity with Weblate, although I didn't follow all instructions in `i18n.md` - I tried to use other PRs as a reference, but I'm not sure if I missed anything.
- As the checkboxes indicate, I have not tested the changes yet. (A short message informing me whether that would be appreciated or overkill would be great).
- I did not bump `PKG_VERSION` because I found no such entry in the Makefile.

systemcrash:
The IPv4 bit in the descriptive text after the title isn't enough?

n0toose:
> The IPv4 bit in the descriptive text after the title isn't enough?

I had honestly trouble figuring out where IPv6 was. "Masquerading" felt, and this is vibes-based, "mutually exclusive" to "IPv6 Masquerading", so I *briefly* stopped looking. Explicitness could have helped. "If *IPv4 Masquerading* is here, and IPv6 is not, where could it be...?" (*checks Additional Settings*) 

I considered adding an explanation on how IPv6 Masquerading can be found under "Additional settings", but I didn't manage to quickly spot any instances of such a practice on the UI (although it might have been more helpful). So, I just refrained from doing that.

n0toose:
Like, with the descriptive text, the first thing that popped up in my head was "Oh, they seem to be making it clear that Masquerading is *only* about IPv4 - doesn't OpenWRT also support the same, but for IPv6?"

Sorry for the lengthy answer - it's just that this is purely subjective (thus hard to justify).

systemcrash:
merged. Thanks @n0toose

PR #7825, opened 2025-06-19
luci-mod-network: aligned column names

GrumpyMeow:
luci-mod-network: aligned column names

With this PR i propose to have the DHCPv4- and DHCPv6-leasetables to both use the same column-names.
I noticed that the tables were using different strings during weblate-translations. 

I don't think this is a breaking change. Also non-breaking for translations.


systemcrash:
merged. Thanks @GrumpyMeow

PR #7829, opened 2025-06-22
luci-app-uhttpd: bugfix

Ramon00:
Fix a bug in luci-app-uhttpd: Updating the crt and/or key from the default config causes the https listener section to be deleted: 
image

### DIFF
--- a/applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
+++ b/applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
@@ -57,9 +57,6 @@ return view.extend({
 		lhttps = s.taboption('general', form.DynamicList, 'listen_https', _('HTTPS listener (address:port)'), _('Bind to specific interface:port (by specifying interface address)'));
 		lhttps.datatype = 'list(ipaddrport(1))';
 
-		var cert = uci.get('uhttpd', 'main', 'cert');
-		var key = uci.get('uhttpd', 'main', 'key');
-
 		lhttps.validate = function (section_id, value) {
 			let have_https_listener = false;
 			let have_http_listener = false;
@@ -94,8 +91,6 @@ return view.extend({
 
 			return true;
 		};
-
-		lhttps.depends({ cert, key });
 
 		o = s.taboption('general', form.Flag, 'redirect_https', _('Redirect all HTTP to HTTPS'));
 		o.default = o.enabled;


systemcrash:
Is this just a luci thing or does uhttpd really require both to run https?

Ramon00:
If I apply those settings, i.e. delete the listen https section, uhttpd still runs, but then you cannot connect with https (port 443), only http on port 80.  The bug is here that if you try to change the certificate then you actually disable the whole https part...

Ramon00:
Maybe to add, if i then later on add the 443 ports back in, then it works again as intended. So not the most dramatic bug but well, kinda like a "huh, why did it stop working"-bug.

systemcrash:
It seems like there's a better way to fix this. Granted that if only 'cert' is provided, uhttpd won't start without the corresponding 'key'. How the form values are fetched might be problematic, if those form values were untouched, like the IP and port https listen setting. 

Try adding something like:
```js
lhttps.cfgvalue = function(section_id) {
	return uci.get('uhttpd', section_id, 'listen_https');
};
```

Because only changing the cert doesn't make sense without also changing the key (they must both be updated), so make sure both are changed.

Ramon00:
That does not work. the issue is not the listen_https section, but these

`

		var cert = uci.get('uhttpd', 'main', 'cert');
		var key = uci.get('uhttpd', 'main', 'key');

`
Apparently these get changed into empty or something making the listen_https section disappear (due to the depend).

But why is the depend even there? Why hide it if the crt and key are not in there? maybe the original author wanted to trigger the validate again instead of hiding?

hnyman:
> But why is the depend even there? Why hide it if the crt and key are not in there?

Likely just to be helpful for users, by hiding unapplicable options. Rather typical in LuCI. E.g. some wifi options are hidden if your radio driver or hostapd version does not support those options.
Here the logic probably is that "if there are no local certs to be used by uhttpd, then https is not possible, so do not offer 'HTTPS listen' option at all."

Ramon00:
Could be, but hiding it behind a "enable https" checkmark would have been nicer. I mean why expose the cert and key but hide the listen ports? The other way around would even have made more sense, show the listen hide and then only show the the key and cert after some listen port is made. 

But anyway by default all are visible and I would argue they should be filled in and used on all installations, so the depend does not make that much sense, only on very rare cases will it hide. 

Having said that I am fine with leaving it in but then it needs to be fixed, because the way it is now it is not working. So this needs suggestions for a different fix, if so desired.

hnyman:
You could well remove the dependency, I think.

The dependency seems to originate from the original submission of the app in 2015, in lua:
https://github.com/openwrt/luci/commit/d83642c20519692c5d3883e23d01bb84a48dd85a

Ramon00:
That is the solution in this PR. Come to think of it the other two lines can then be removed as well, let me update the PR tonight

PR #7831, opened 2025-06-22
luci-app-usteer: Extra error checking

Ramon00:
When you directly load the usteer page "/cgi-bin/luci/admin/network/usteer"  then somehow the wifi networks do not get populated, this throws an error. 
image
I added extra checking to prevent this. The result is that the  SSID list is empty: 
image
while normally you get the list of all wifi networks. 

It would be better to additionally fix `network.js`, but to be honest I am not sure why it will not  show, maybe its a timing issue?

Also fixed a console error related to the hostnames by additional error checking.


systemcrash:
Yeah, likely due to timing. Checking for system features (wifi) happens at DOM load time, which should complete first. Then the initNetworkState Promise bundle. I've pushed out 9c0f071bbd1150b324a5c1b2f0e1f642da693128 which you can also try. But these fixes seem otherwise good.

PR #7832, opened 2025-06-23
luci-app-csshnpd: Add new package

cpswan:
Maintainer: @cpswan
Compile tested: x86_64 (snapshot r29619, 24.10.0), aarch64_cortex-a53 (snapshot, 23.05.5)
Run tested: x86_64 (snapshot r29619, 24.10.0), aarch64_cortex-a53 (23.05.5)

Description:

Adding LuCI web interface for [csshnpd package](https://github.com/openwrt/packages/tree/master/net/csshnpd) (thanks @feckert and @BKPepe for shepherding https://github.com/openwrt/packages/pull/26518 through to being merged).

### NoPorts Config screenshot
image

### NoPorts Enrollment tab screenshot prior to enrollment
image

### NoPorts Enrollment tab screenshot if config is incomplete
image

### NoPorts Enrollment tab screenshot once key is in place
image

Video walk through at [NoPorts OpenWrt Installation Guide](https://docs.noports.com/installation/openwrt-installation-guide) (NB predates csshnpd being accepted upstream so starts with LuCI based installation of downloaded .ipks).

### DIFF
--- a/applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
+++ b/applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
@@ -1,0 +1,88 @@
+'use strict';
+'require view';
+'require form';
+
+function validateAtsign(section_id, value) {
+	if (value.length < 1) {
+		return _('Must not be empty and should start with @ (e.g., "@a").');
+	}
+	return true;
+}
+
+function validateDevice(section_id, value) {
+	if (value.length < 1) {
+		return _('Must be at least one character long (e.g., "a").');
+	}
+	if (value.length == 1) {
+		if (!/^[a-z]+$/.test(value)) {
+			return _('First character should be a lowercase letter (e.g., "a").');
+		} else {
+			return true;
+		}
+	}
+	if (!/^[a-z][a-z0-9_-]+$/.test(value)) {
+		return _('Device names may contain a-z 0-9 _ or - (e.g., "my_thing1").');
+	}
+	if (value.length > 36) {
+		return _('Maximum device name length is 36 characters.');
+	}
+	return true;
+}
+
+function validateOTP(section_id, value) {
+	if (value.length != 6) {
+		return _('Must be six characters (e.g., "S3CR3T").');
+	}
+	return true;
+}
+
+function firstAt(section_id, value) {
+	if (value && !value.startsWith('@')) {
+		value = '@' + value; // Ensure @ at start
+	}
+	return this.super('write', [section_id, value]);
+}
+
+return view.extend({
+	render: function() {
+		let m, s, o;
+
+		m = new form.Map('sshnpd', _('NoPorts'),
+			_('Daemon Configuration'));
+
+		s = m.section(form.TypedSection, 'sshnpd', _('sshnpd config'));
+		s.anonymous = true;
+
+		o = s.option(form.Value, 'atsign', _('Device atSign'),
+			_('The device atSign e.g. @device'));
+		o.default = '@device';
+		o.validate = validateAtsign;
+		o.write = firstAt;
+
+		o = s.option(form.Value, 'manager', _('Manager atSign'),
+			_('The manager atSign e.g. @manager'));
+		o.default = '@manager';
+		o.validate = validateAtsign;
+		o.write = firstAt;
+
+		o = s.option(form.Value, 'device', _('Device name'),
+			_('The name for this device e.g. openwrt'));
+		o.default = 'openwrt';
+		o.validate = validateDevice;
+
+		s.option(form.Value, 'args', _('Additional arguments'),
+			_('Further command line arguments for the NoPorts daemon'));
+
+		o = s.option(form.Value, 'otp', _('Enrollment OTP/SPP'),
+			_('One Time Passcode (OTP) for device atSign enrollment'));
+		o.default = '000000';
+		o.validate = validateOTP;
+
+		o = s.option(form.Flag, 'enabled', _('Enabled'),
+			_('Check here to enable the service'));
+		o.default = '1';
+		o.rmempty = false;
+
+		return m.render();
+	},
+});
--- a/applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
+++ b/applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
@@ -1,0 +1,97 @@
+'use strict';
+'require view';
+'require dom';
+'require fs';
+'require ui';
+'require uci';
+'require network';
+
+
+return view.extend({
+	handleCommand: function(exec, args) {
+		let buttons = document.querySelectorAll('.diag-action > .cbi-button');
+
+		for (let i = 0; i < buttons.length; i++)
+			buttons[i].setAttribute('disabled', 'true');
+
+		return fs.exec(exec, args).then(function(res) {
+			let out = document.querySelector('textarea');
+
+			dom.content(out, [ res.stdout || '', res.stderr || '' ]);
+		}).catch(function(err) {
+			ui.addNotification(null, E('p', [ err ]))
+		}).finally(function() {
+			for (let i = 0; i < buttons.length; i++)
+				buttons[i].removeAttribute('disabled');
+		});
+	},
+
+	handleEnroll: function() {
+		return this.handleCommand('at_enroll.sh', "");
+	},
+
+	load: function() {
+		return uci.load('sshnpd').then(function() {
+			let atsign = uci.get_first('sshnpd','','atsign'),
+				keyfile = '/root/.atsign/keys/'+atsign+'_key.atKeys';
+				return L.resolveDefault(fs.stat(keyfile), {});
+		});
+	},
+
+	render: function(res) {
+
+		const has_atkey = res.path;
+		const atsign = uci.get_first('sshnpd','','atsign');
+		const device = uci.get_first('sshnpd','','device');
+		const otp = uci.get_first('sshnpd','','otp');
+		const enrollready = atsign && device && otp && !has_atkey;
+
+		const instructions = E('div', { 'class': 'cbi-map-descr'}, _('Press the Enroll button then run this command on a system where '+atsign+' is activated:'));
+
+		const enrollcmd = E('code','at_activate approve -a '+atsign+' --arx noports --drx '+device);
+
+		let table = E('table', { 'class': 'table' }, [
+				E('tr', { 'class': 'tr' }, [
+					E('td', { 'class': 'td left' }, [
+						E('span', { 'class': 'diag-action' }, [
+							E('button', {
+								'class': 'cbi-button cbi-button-action',
+								'click': ui.createHandlerFn(this, 'handleEnroll')
+							}, [ _('Enroll') ])
+						])
+					]),
+				])
+			]);
+
+		const cmdwindow = E('div', {'class': 'cbi-section'}, [
+			E('div', { 'id' : 'command-output'},
+				E('textarea', {
+					'id': 'widget.command-output',
+					'style': 'width: 100%; font-family:monospace; white-space:pre',
+					'readonly': true,
+					'wrap': 'on',
+					'rows': '20'
+				})
+			)
+			]);
+
+		let view = E('div', { 'class': 'cbi-map'}, [
+			E('h2', {}, [ _('NoPorts atSign Enrollment') ]),
+			atsign ? E([]) : E('div', { 'class': 'cbi-map-descr'}, _('atSign must be configured')),
+			device ? E([]) : E('div', { 'class': 'cbi-map-descr'}, _('Device must be configured')),
+			otp ? E([]) : E('div', { 'class': 'cbi-map-descr'}, _('OTP must be configured. An OTP can be generated using:')),
+			otp ? E([]) : E('code','at_activate otp -a '+atsign),
+			has_atkey ? E('div', { 'class': 'cbi-map-descr'}, _('Existing key found at: '+has_atkey)) : E([]),
+			enrollready ? instructions : E([]),
+			enrollready ? enrollcmd  : E([]),
+			enrollready ? table : E([]),
+			enrollready ? cmdwindow : E([]),
+		]);
+
+		return view;
+	},
+
+	handleSaveApply: null,
+	handleSave: null,
+	handleReset: null
+});
--- a/applications/luci-app-csshnpd/root/usr/share/luci/menu.d/luci-app-csshnpd.json
+++ b/applications/luci-app-csshnpd/root/usr/share/luci/menu.d/luci-app-csshnpd.json
@@ -1,0 +1,30 @@
+{
+	"admin/network/sshnpd": {
+		"title": "NoPorts",
+		"order": 70,
+		"action": {
+			"type": "firstchild"
+		},
+		"depends": {
+			"acl": [ "luci-app-csshnpd" ]
+		}
+	},
+
+	"admin/network/sshnpd/config": {
+		"title": "NoPorts Config",
+		"order": 1,
+		"action": {
+			"type": "view",
+			"path": "sshnpd/config"
+		}
+	},
+
+	"admin/network/sshnpd/enroll": {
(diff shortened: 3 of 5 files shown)


applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+'require view';
+'require form';
+
+// Project code format is tabs, not spaces
+return view.extend({
+	render: function() {
+		var m, s, o;
+
+		/*
```
feckert: I think this comment is no needed this is the default behaviour.

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+'require view';
+'require form';
+
+// Project code format is tabs, not spaces
+return view.extend({
+	render: function() {
+		var m, s, o;
```
feckert: Please use `let` instead of `var` https://stackoverflow.com/questions/762011/what-is-the-difference-between-let-and-var

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+'require view';
+'require form';
+
+// Project code format is tabs, not spaces
```
feckert: Please remove this line

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
```diff
@@ -0,0 +1,53 @@
...
+		The relevant ACL path for writing back the configuration is
+		write > uci > ["example"]
+		*/
+		m = new form.Map('sshnpd', _('NoPorts'),
+			_('Daemon Configuration'));
+
+		s = m.section(form.TypedSection, 'sshnpd', _('sshnpd config'));
+		s.anonymous = true;
+
+		s.option(form.Value, 'atsign', _('Device atSign'),
```
feckert: What about input validation?
It would also be nice if you could add a short description of what this options are for.

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
```diff
@@ -0,0 +1,97 @@
...
+		return uci.load('sshnpd').then(function() {
+			var atsign = uci.get_first('sshnpd','','atsign'),
+				keyfile = '/root/.atsign/keys/'+atsign+'_key.atKeys';
+				return L.resolveDefault(fs.stat(keyfile), {});
+		});
+	},
+
+	render: function(res) {
+
+		var has_atkey = res.path,
```
feckert: This does not correspond to the standard of look and feel and is also difficult to read.
Can you please define more then one variable  with `var` .
Please replace `var` with `let`.
See we have:
```
let has_atkey = res.path;
let atsign = uci.get_first('sshnpd','','atsign');
let device = uci.get_first('sshnpd','','device');
			
```

Cod is always read more often than written and I think it is difficult to make a change because everything is done in one line at the end.

applications/luci-app-csshnpd/root/usr/share/luci/menu.d/luci-app-csshnpd.json
```diff
@@ -0,0 +1,30 @@
...
+		"order": 70,
+		"action": {
+			"type": "firstchild"
+		},
+		"depends": {
+			"acl": [ "luci-app-csshnpd" ]
+		}
+	},
+
+  "admin/network/sshnpd/config": {
```
feckert: Please use `tabs`

applications/luci-app-csshnpd/root/usr/share/luci/menu.d/luci-app-csshnpd.json
```diff
@@ -0,0 +1,30 @@
...
+
+  "admin/network/sshnpd/enroll": {
+    "title": "NoPorts Enrollment",
+    "order": 2,
+    "action": {
+      "type": "view",
+      "path": "sshnpd/enroll"
+    }
+  }
+}
```
feckert: New line

applications/luci-app-csshnpd/root/usr/share/rpcd/acl.d/luci-app-csshnpd.json
```diff
@@ -0,0 +1,15 @@
...
+			"file": {
+				"/usr/bin/at_enroll.sh": ["exec"],
+				"/root/.atsign/keys/*": ["stat"]
+			}
+		},
+		"write": {
+			"uci": [ "sshnpd" ]
+		}
+	}
+}
```
feckert: New line

cpswan:
@feckert thanks for the comprehensive feedback. All the changes you requested should be addressed in https://github.com/openwrt/luci/pull/7832/commits/18169c260449de17fe721642ca52ac870de8f4a3

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
```diff
@@ -0,0 +1,82 @@
...
+			return _('First character should be a lowercase letter (e.g., "a").');
+		} else {
+			return true;
+		}
+	}
+	if (!/^[a-z][a-z0-9_-]+$/.test(value)) {
+		return _('Device names may contain a-z 0-9 _ or - (e.g., "my_thing1").');
+	}
+	if (value.length > 36) {
+		return _('Maximum device length is 36 characters.');
```
systemcrash: device length or device name length?

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
```diff
@@ -0,0 +1,97 @@
...
+'require dom';
+'require fs';
+'require ui';
+'require uci';
+'require network';
+
+
+return view.extend({
+	handleCommand: function(exec, args) {
+		var buttons = document.querySelectorAll('.diag-action > .cbi-button');
```
systemcrash: ES6 i.e. let

applications/luci-app-csshnpd/root/usr/share/rpcd/acl.d/luci-app-csshnpd.json
```diff
@@ -0,0 +1,15 @@
+{
+	"luci-app-csshnpd": {
+		"description": "Grant UCI access for luci-app-csshnpd",
+		"read": {
+			"uci": [ "sshnpd" ],
+			"file": {
+				"/usr/bin/at_enroll.sh": ["exec"],
+				"/root/.atsign/keys/*": ["stat"]
```
systemcrash, line 8: Not sure this is great place to store secrets. They are not included in backups. Not sure about upgrades.

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
```diff
@@ -0,0 +1,97 @@
...
+		let atsign = uci.get_first('sshnpd','','atsign');
+		let device = uci.get_first('sshnpd','','device');
+		let otp = uci.get_first('sshnpd','','otp');
+		let enrollready = atsign && device && otp && !has_atkey;
+
+		let instructions = E('div', { 'class': 'cbi-map-descr'}, _('Press the Enroll button then run this command on a system where '+atsign+' is activated:'));
+
+		let enrollcmd = E('code','at_activate approve -a '+atsign+' --arx noports --drx '+device);
+
+		let table = E('table', { 'class': 'table' }, [
```
systemcrash: can these be `const`?

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
```diff
@@ -0,0 +1,82 @@
...
+
+function validateAtsign(section_id, value) {
+	if (!/^@/.test(value)) {
+		return _('Must start with the "@" character.');
+	}
+	if (value.length < 2) {
+		return _('Must be at least 2 characters long (e.g., "@a").');
+	}
+	return true;
+}
```
systemcrash: Perhaps override the write function to check for the @, and if it is missing, prepend it.

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
cpswan: Fixed in https://github.com/openwrt/luci/pull/7832/commits/9fd92bff9302ca6da3d3b4cba3650a75dc4298a5

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
cpswan: Fixed in https://github.com/openwrt/luci/pull/7832/commits/9fd92bff9302ca6da3d3b4cba3650a75dc4298a5

applications/luci-app-csshnpd/root/usr/share/rpcd/acl.d/luci-app-csshnpd.json
cpswan: If the key is lost it can just be re-enrolled. The master key lives elsewhere.

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
cpswan: Yes, done in https://github.com/openwrt/luci/pull/7832/commits/9fd92bff9302ca6da3d3b4cba3650a75dc4298a5

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
cpswan: Can you think of another package that does (something like) this?
I don't know how to do that from clean cloth, but I should be able to rework from other code.

systemcrash: Search the codebase for something like `.write = `. Read also the luci docu. 

https://openwrt.github.io/luci/jsapi/LuCI.html

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/enroll.js
```diff
@@ -0,0 +1,97 @@
...
+		const atsign = uci.get_first('sshnpd','','atsign');
+		const device = uci.get_first('sshnpd','','device');
+		const otp = uci.get_first('sshnpd','','otp');
+		const enrollready = atsign && device && otp && !has_atkey;
+
+		const instructions = E('div', { 'class': 'cbi-map-descr'}, _('Press the Enroll button then run this command on a system where '+atsign+' is activated:'));
+
+		const enrollcmd = E('code','at_activate approve -a '+atsign+' --arx noports --drx '+device);
+
+		let table = E('table', { 'class': 'table' }, [
```
systemcrash, line 53: You might think it can't be const, but E() creates an object which can mutate. const just prevents it from being replaced. This is fine as is anyway :)

applications/luci-app-csshnpd/root/usr/share/rpcd/acl.d/luci-app-csshnpd.json
systemcrash: Upgrades should prevent this extra work from being a necessity. How about `/etc/atsign` which is included in backups and upgrades.

cpswan: `$HOME/.atsign/keys` is standard across all the supported platforms, so I'm loath to change it just for OpenWrt. NB also this would need to be changed in the csshnpd package too, as that's where the core config lives.

It is possible to specify alternative key locations using the additional args setting and passing in a `-k` flag. But I'd rather not have that flag there by default, because it's then going to get extra fiddly if somebody wants to put their keys somewhere else.

Lastly there is an argument that keys shouldn't be lying around in backups.

applications/luci-app-csshnpd/htdocs/luci-static/resources/view/sshnpd/config.js
cpswan: That's done in https://github.com/openwrt/luci/pull/7832/commits/e6c6c852f6d374065944d4e4ca69e0f9d1f26deb

applications/luci-app-csshnpd/root/usr/share/rpcd/acl.d/luci-app-csshnpd.json
systemcrash: The alternative is to add the keys location to `/etc/sysupgrade.conf` so keys aren't lost. Things for the root user can go in the `/root` folder generally, but does this mean the daemon runs as root, which is itself a compromise vector? Or is it chrooted or sandboxed by procd into ujail?

cpswan: > The alternative is to add the keys location to `/etc/sysupgrade.conf` so keys aren't lost.

Good idea. I'll do that with the next update to the `csshnpd` package.

>Things for the root user can go in the `/root` folder generally, but does this mean the daemon runs as root, which is itself a compromise vector? Or is it chrooted or sandboxed by procd into ujail?

The daemon runs as the user that's going to make the NoPorts connection (usually an SSH connection for OpenWrt), so given that most OpenWrt installs just have a root account, the daemon will generally run as root. [`csshnp.init`](https://github.com/openwrt/packages/blob/master/net/csshnpd/files/csshnpd.init) is designed to support users other than root, but I'm not holding out much hope that many users will actually do that.

cpswan: > The alternative is to add the keys location to `/etc/sysupgrade.conf` so keys aren't lost. 

Or would it make more sense to add `/root/.atsign` to the `define Package/csshnpd/conffiles` section of the Makefile?

systemcrash: It's just not very 'openwrt' but it works. Do what you consider best in this situation - bear in mind that unexpected behaviour shall be documented. A user upgrades and all their keys disappear is cause for concern.

systemcrash:
Also, as a general comment, do you have the possibility to host an openwrt repo for the luci and daemon packages? It may not be not optimal for exposure but it means you have full control of package distro. I added key management for the GUI recently. 

Take a look at https://fantastic-packages.github.io/releases/ for some inspiration.

cpswan:
> Also, as a general comment, do you have the possibility to host an openwrt repo for the luci and daemon packages? It may not be not optimal for exposure but it means you have full control of package distro. I added key management for the GUI recently.

At present we make packages available [in GitHub Releases](https://github.com/atsign-foundation/Atsign_OpenWRT_packages/releases) but I also have an issue open to create a proper package repo - https://github.com/atsign-foundation/Atsign_OpenWRT_packages/issues/47

> Take a look at https://fantastic-packages.github.io/releases/ for some inspiration.

Thanks for the pointer :)

applications/luci-app-csshnpd/root/usr/share/rpcd/acl.d/luci-app-csshnpd.json
cpswan: https://github.com/openwrt/packages/pull/26894 opened so that the .atsign directory is part of conffiles and verified that my atKeys file appears in the Backup file list.

cpswan: and that's now merged

PR #7833, opened 2025-06-23
luci-app-lldpd and luci-app-privoxy: Add missing Maintainer

stokito:
The maintainer field is important to help users to tag `@` a reviewer

### DIFF
--- a/applications/luci-app-privoxy/Makefile
+++ b/applications/luci-app-privoxy/Makefile
@@ -9,6 +9,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=luci-app-privoxy
 
 PKG_LICENSE:=Apache-2.0
+PKG_MAINTAINER:=Paul Donald <newtwen+github@gmail.com>
 
 LUCI_TITLE:=LuCI Support for Privoxy WEB proxy
 LUCI_DEPENDS:=+luci-base +privoxy


hnyman:
@systemcrash seems to be proposed by you. Have you asked him?

stokito:
I hope @systemcrash will be fine with this given that he was the last person who refactored these two apps.
If not, then the "default" maintainer is ~jow. We may create a team and mentions it like @openwrt/luci-devs

In fact I did what a new contributor will make: find the last person who made significant changes to mention it for a review.

systemcrash:
I intentionally didn't put a maintainer in them: (just about ;) everything there is working code (where commit history speaks for itself), so anyone can roll up their sleeves and get busy. I did work on at least LLDPD to bring it to feature parity with the daemon and add stuff that others and I use.

You've done nothing wrong, and you like to see apps have maintainers, but I don't see the need. Some people move on after a while. Just another thing to maintain :)

stokito:
I think it's important to have a maintainer field to help users to find whom to mention. The empty field only means to find someone itself which is difficult. We may use some generic team name instead. This are only the two addons without a maintainer field.

stokito:
@systemcrash it would be great to add the maintainers to save a time for contributors. But ideally it would be add CODEOWNERS #7439
If you don't wish to be noted as a maintainer then close the PR

PR #7838, opened 2025-06-27
luci-mod-status: Cleanup and fix syslog.js bug

Ramon00:
Fixed bug in syslog.js where lines could be hidden even when "any" was selected, and simplified code a bit.
When clicking on "Not" they appear:
image

@systemcrash: Can you have a look? 

Also is there a way to check which logger is used (im using syslog-ng). That does not use a "." between, so the functionallity actually does not fully work? If so I can easily add this to the code. Alternative would be to remove the "."

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
@@ -9,12 +9,12 @@ return view.extend({
 	logFacilityFilter: 'any',
 	invertLogFacilitySearch: false,
 	logSeverityFilter: 'any',
-	invertLogseveritySearch: false,
+	invertLogSeveritySearch: false,
 	logTextFilter: '',
 	invertLogTextSearch: false,
 
 	facilities: [
-		['any', '', _('Any')],
+		['any', 'any', _('Any')],
 		['0',  'kern',   _('Kernel')],
 		['1',  'user',   _('User')],
 		['2',  'mail',   _('Mail')],
@@ -42,7 +42,7 @@ return view.extend({
 	],
 
 	severity: [
-		['any', '', _('Any')],
+		['any','any', _('Any')],
 		['0',  'emerg',   _('Emergency')],
 		['1',  'alert',   _('Alert')],
 		['2',  'crit',   _('Critical')],
@@ -55,7 +55,6 @@ return view.extend({
 
 
 	retrieveLog: async function() {
-		const facility = this.logFacilityFilter;
 
 		return Promise.all([
 			L.resolveDefault(fs.stat('/usr/libexec/syslog-wrapper'), null),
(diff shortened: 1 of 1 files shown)


Ramon00:
Oh one other thing, maybe also add the freetext filtering to the kernel log tab?

systemcrash:
We can make a different wrapper to look at the different logging scenarios. This is intended to handle the built-in logger, logread. Please leave the numbers - they're used to build syslog facilities in general and are helpful to understand ordering and priority - even if they're not in active used here. 


In your picture, only 'info' is present - which lacks the `.`. Perhaps tweak your logging so it matches the facility.severity form.

Ramon00:
> Please leave the numbers

I can do that, just thought to simplify things.

> n your picture, only 'info' is present - which lacks the `.`. Perhaps tweak your logging so it matches the facility.severity form.

That is how I found the bug. If you select any/any/"" then the code should not filter regardless of the content (dots or any other thing being present in the line or not).

> We can make a different wrapper to look at the different logging scenarios. This is intended to handle the built-in logger, logread.

Would be great if it would not rely on a specific layout or maybe if the standardard logger is not used, that then then the dot is not checked, but that is not the most dramatic part, i can add a "." in syslog-ng, its really flexible. 

But let me revert that numbering thing.

Ramon00:
ok have a look

modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
```diff
@@ -185,7 +173,7 @@ return view.extend({
 			self.logFacilityFilter = facilitySelect.value;
 			self.invertLogFacilitySearch = facilityInvert.checked;
 			self.logSeverityFilter = severitySelect.value;
-			self.invertLogseveritySearch = severityInvert.checked;
+			self.invertLogSeveritySearch = severityInvert.checked;
 			self.logTextFilter = filterTextInput.value;
 			self.invertLogTextSearch = filterTextInvert.checked;
 			self.retrieveLog().then(log => {
```
systemcrash: Perhaps also change this to `self.pollLog();` while we're here

systemcrash:
> Oh one other thing, maybe also add the freetext filtering to the kernel log tab?

The browser's own search suffices nicely, no?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/syslog.js
Ramon00: yeah indeed

Ramon00:
> The browser's own search suffices nicely, no?

well, filtering puts lines next to each other while searching just shows one by one.

Ramon00:
one other thing, while we are at it, what do you think about removing the "focus" from the scroll buttons after we clicked on it e.g. add `scrollDownButton.blur();` ?

systemcrash:
Seems to work without it, but for completeness, yeah.

Ramon00:
yes still works, but visually it is not as nice. Give it a spin, if you think its ok as is, then go ahead and merge

systemcrash:
merged. Thanks @Ramon00

Ramon00:
Thx!

PR #7839, opened 2025-06-28
luci-app-https-dns-proxy: DNS4EU support

bjmi:
Support DNS4EU DNS resolvers: https://www.joindns4.eu/for-public#resolver-options


systemcrash:
Looks ok tho should go into @stangri's repo first. 

https://github.com/stangri/source.openwrt.melmac.net

bjmi:
@rikroe has already submitted a DNS4EU config in stangri/luci-app-https-dns-proxy#10.
I'll close this PR.

stangri:
I'm looking to fix the bug with the bootstrap option not loading values from providers' json files before I version bump and merge this and the bootstrap fix in OpenWrt's luci snapshot/release repos. I've experimented with it earlier today and didn't find an acceptable solution, if it takes me much longer I'll sync versions between my and OpenWrt repos without the bootstrap fix.

Issue #7801, opened 2025-05-30
Request to Clarify DHCP Reservation Label in LuCI


nopcarrier:
### What would you like to see in luci?

Hello OpenWRT team,

I appreciate the robust features of OpenWRT and its LuCI interface. I have a suggestion to improve clarity in the DHCP Leases section.

Currently, the button labeled "Set Static" is used to create a DHCP reservation for a device, ensuring it receives the same IP address based on its MAC address. However, the term "Set Static" can be misleading, as it traditionally refers to manually configuring a static IP address on the device itself, bypassing DHCP.

To enhance clarity and align with standard networking terminology, I propose renaming this button to one of the following:

"Add DHCP Reservation"

"Create Static Lease"

"Reserve IP Address"

This change would help users better understand the function and prevent potential confusion.

Thank you for considering this suggestion.

Best regards, 
Dan H

systemcrash:
Hover text is possible.

efahl:
@nopcarrier What version OpenWrt do you use?  Fix might need backport for you to see it...

PR #7841, opened 2025-06-29
modules: dhcp: improve wording for DHCP reservations

efahl:
The use of "static" regarding DHCP reservations can be misleading, as static usually implies a setting on the host device, where what we're actually doing is reserving the address on the server.

Change the button label from "Set Static" to "Reserve IP" to more clearly indicate the action to be taken.  Add some tip text to make it even more clear.

Fixes: https://github.com/openwrt/luci/issues/7801


systemcrash:
merged. Thanks @efahl

PR #7852, opened 2025-07-04
treewide: adjust packages against glibc 2.41

rockdrilla:
Depends on: https://github.com/openwrt/openwrt/pull/19293

### DIFF
--- a/libs/luci-lib-nixio/Makefile
+++ b/libs/luci-lib-nixio/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=NIXIO POSIX library
-LUCI_DEPENDS:=+liblua
+LUCI_DEPENDS:= +USE_GLIBC:libcrypt-compat +liblua
 
 PKG_LICENSE:=Apache-2.0
 


rockdrilla:
I've resigned commit, no other changes are made.

rockdrilla:
Current state of repositories allows to merge this safely in cost of one extra dependency (libcrypt-compat) and sort of UB on which libcrypt.so.1 is loaded - from current glibc 2.38 or from that compatibility package.

rockdrilla:
@systemcrash PTAL

systemcrash:
Giving others a chance to comment. Seems this is reasonable. This module might disappear anyway once all of the lua apps are gone, but that's a bit further down the road.

rockdrilla:
Compile-/run-tested: mediatek/filogic (GL.iNet GL-MT6000), ath79/generic (TP-Link Archer C7 v2).

@robimarko and you? :)

graysky2:
```
Build system: x86/64
Build-tested: x86/64-glibc
Run-tested: x86/64-glibc
```

rockdrilla:
@systemcrash it seems that we should merge this MR [ASAP] as it's done with https://github.com/openwrt/packages/pull/26889.

Issue #7526, opened 2025-01-04
treewide: rpcd scripts needs to be executable in git index?


dannil:
### Is there an existing issue for this?

- [X] I have searched among all existing issues (including closed issues)

### screenshots or captures

Tagging some of the currently most active developers in the repo for discussion.

@systemcrash @stokito @Ramon00 @rdmitry0911 

Spawned from https://forum.openwrt.org/t/openwrt-24-10-0-rc4-fourth-release-candidate/219368/142?u=dannil

It seems that all apps which doesn't have it's respective rpcd scripts set with the executable flag, which at the moment is `luci-app-olsr-viz` and `luci-app-squid`, isn't able to register with ubus and therefore crashes in the UI with the very non-descript error `RPC call to <name of RPC object>
 failed with error -32000: Object not found`

![Screenshot_24](https://github.com/user-attachments/assets/665e6532-4f7b-47da-a3e7-81fe5607f56e)
![Screenshot_25](https://github.com/user-attachments/assets/355c3162-9324-4615-befb-c72edc071960)

The affected apps were found with the following command

```
~/git/openwrt_parent/luci master > find . -type d -name rpcd -exec sh -c 'ls -la "$1"' _ {} \; | grep -v '^d' | grep -v '^total'
-rwxr-xr-x 1 vm vm  713 Jun 25  2024 luci.yggdrasil
-rwxr-xr-x 1 vm vm  791 Jun 25  2024 luci.yggdrasil-jumper
-rwxr-xr-x 1 vm vm  897 Jun 25  2024 luci.battstatus
-rwxr-xr-x 1 vm vm 5016 Jun 25  2024 luci.https-dns-proxy
-rwxr-xr-x 1 vm vm 8501 Jun 25  2024 luci.example
-rwxr-xr-x 1 vm vm 8750 Jan  4 22:45 luci.adblock-fast
-rwxr-xr-x 1 vm vm 9317 Nov 17 11:32 luci.pbr
-rwxr-xr-x 1 vm vm 12577 Jun 25  2024 luci.advanced_reboot
-rwxr-xr-x 1 vm vm 1107 Jun 25  2024 olsr-services
-rwxr-xr-x 1 vm vm 1561 Jun 25  2024 olsrinfo
-rw-r--r-- 1 vm vm  699 Jan  4 23:14 luci.squid
-rw-r--r-- 1 vm vm 1378 Jan  4 22:45 olsrvizinfo
```

[...]

PR #7861, opened 2025-07-13
luci-proto-yggdrasil: fix interface selection

Saiv46:
Package `yggdrasil` doesn't support network names and aliases.

To avoid back-porting and upgrading main package across releases, it's simpler to reflect this behavior in interface instead.

I'm leaving out incrementing package version for @wfleurant as there may be other issues to fix (see #7526).

- [X] Tested on: (OpenWrt 24.10.2, ramips/mt7621, Firefox 140) :white_check_mark:
- [X] ( Preferred ) Mention: @ the original code author for feedback

### DIFF
--- a/protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
+++ b/protocols/luci-proto-yggdrasil/htdocs/luci-static/resources/protocol/yggdrasil.js
@@ -276,7 +276,9 @@ return network.registerProtocol('yggdrasil',
 			o=ss.option(form.Value,"address",_("Peer URI"));
 			o.placeholder="tls://0.0.0.0:0"
 			o.validate=validateYggdrasilPeerUri;
-			ss.option(widgets.NetworkSelect,"interface",_("Peer interface"));
+
+			o=ss.option(widgets.DeviceSelect,"interface",_("Peer interface"));
+			o.noaliases=true;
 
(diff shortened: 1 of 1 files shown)


wfleurant:
ok fine and does this change anything visually? thanks for helping

systemcrash:
We recently changed rp-pppoe from device to network, because pppoe has problems with device names - and works best with networks. Seems the opposite here. They're the same widget, but instead of choosing `wan` you choose `eth1` or so, which doesn't always match expectations. The `@` aliases can be a problem, however.

wfleurant:
@Saiv46 can you test your patch using multiple yggdrasil interfaces? if i recall this should still be supported feature for some time now... i think it is still being used with particular networks if ok, we can merge eh?

Saiv46:
> @Saiv46 can you test your patch using multiple yggdrasil interfaces? if i recall this should still be supported feature for some time now... i think it is still being used with particular networks if ok, we can merge eh?

@wfleurant I didn't quite understand the wording, but this patch has no side effects when using with multiple interfaces - it only prevents launching Yggdrasil instance with invalid interface name (unless network name is equal to interface name, which is rare). Main package script haven't been updated for the last year, so it's safe to assume that feature is working.

Merge when you want to :)

systemcrash:
ping

wfleurant:
sounds like an improvement 👍🙏 thanks

PR #7867, opened 2025-07-16
luci-base: implement support for reversing buttons order on login

stangri:
This adds support for swapping Login/Reset button order by setting luci.main.login_button_order=reverse.

Default order is [Login] [Reset], this has been bothering me as I perfer the order of [Reset] [Login]. The default action of Enter still causes the from to be submitted, even with the order of buttons reversed.

If this is actively unwanted, I can always patch the ut-file in my builds.

### DIFF
--- a/modules/luci-base/ucode/template/sysauth.ut
+++ b/modules/luci-base/ucode/template/sysauth.ut
@@ -35,8 +35,14 @@
 	</div>
 
 	<div class="cbi-page-actions">
-		<input type="submit" value="{{ _('Log in') }}" class="btn cbi-button cbi-button-apply" />
-		<input type="reset" value="{{ _('Reset') }}" class="btn cbi-button cbi-button-reset" />
+		{% let login_button_order = uci.get('luci', 'main', 'login_button_order') ?? "";
+		if (login_button_order == 'reverse'): %}
+			<input type="reset" value="{{ _('Reset') }}" class="btn cbi-button cbi-button-reset" />
+			<input type="submit" value="{{ _('Log in') }}" class="btn cbi-button cbi-button-apply" />
+		{% else %}
+			<input type="submit" value="{{ _('Log in') }}" class="btn cbi-button cbi-button-apply" />
+			<input type="reset" value="{{ _('Reset') }}" class="btn cbi-button cbi-button-reset" />
+		{% endif %}
 	</div>
 </form>
 


systemcrash:
Where might one encounter this order in other sites or projects? Just curious. I’ve never thought about it (for me the order is fine).

otherwise the change looks ok. I’ll wait and see if anyone else has any feedback.

stangri:
> Where might one encounter this order in other sites or projects?

I was trying to recall where would I have seen a similar UI with the action button on the right-side when I was creating PR and I couldn't remember anything. Most sites have [Sign Up] and [Login] buttons aligned vertically to avoid LTR/RTL languages confusion. 

Like I said I don't mind dropping it if I'm the only one who feels the previous order of buttons is odd for LTR languages.

However, this fix might be useful for RTL languages. ;)

stangri:
Just realized that unless this is also implemented for all the other luci pages with [Save & Apply], [Save], [Reset] buttons, as well as modal Configurations/Changes popup which has [Close], [Save], [Revert] buttons, the login page will be inconsistent with the rest of the UI.

Hence a question to a wider audience:
1. Is there a value in working on the rest of the templates so that they would also respect the "Reverse order" setting?
2. Should the setting then be renamed to reflect its significance for RTL languages?

stangri:
When flashing upgrade via WebUI, the action button is on the right.
image
image

stokito:
On the login page the `[Reset]` button itself should be removed. 
The new option `login_button_order` will only complicate things and eat space (while this is not so important). Most people open their router not so often so they may not even notice this.
The action button is usually on right. Maybe on some "opinionated" platforms like MacOS or GNOME this is not true but almost everywhere this is on right. We may check the deign guidelines:

* https://developer.gnome.org/hig/patterns/feedback/dialogs.html
* https://develop.kde.org/hig/layout_and_nav/
* https://developer.apple.com/design/human-interface-guidelines/

The Luci is based on Bootstrap and we can check its docs.

I think it would be good to change the layout to make it more intuitive. This may be problematic for users who follows video tutorials but anyway this tutorials needs to be renewed each few years.

Ramon00:
why is the reset button even there? In bootstrap it is not even there: 
image

I would say remove it altogether.

stangri:
Oh, didn't realize the buttons on the login page are theme-controlled. I guess I should just create PR for the theme to remove the Reset button like @stokito and @Ramon00 suggested.

Thanks!

stangri:
So the `modules/luci-base/ucode/template/sysauth.ut` file needs to be patched still? I don't understand why the Reset button is visible on some themes, but not others.

stokito:
This is clearly just an artifact, let's remove it.

stangri:
Did I understand it correctly that [this](https://github.com/openwrt/luci/pull/7904) should be a desired fix?

systemcrash:
Bootstrap (theme - the only one with the sysauth.ut file) has:

`	<button class="btn cbi-button-positive important">{{ _('Log in') }}</button>`

whereas the default template you want to remove the reset button from has:

```
	<div class="cbi-page-actions">
		<input type="submit" value="{{ _('Log in') }}" class="btn cbi-button cbi-button-apply" />
		<input type="reset" value="{{ _('Reset') }}" class="btn cbi-button cbi-button-reset" />
	</div>
```

Themes override the default ut files. 

I think reset has a place: in my interaction with forms, especially from mobile devices, it's nice to smash reset or clear in case you fudged the input. Users who don't have autofill of a saved password benefit from such functionality. 

Perhaps the themes need the reset button adding. What's the thinking to get rid of reset? Underused button consuming real-estate?

stokito:
@systemcrash I used the reset once or two in my life on some internal corporate sites with old and big forms. Users will survive without it, in the same time we can simplify code and even safe a few bytes on a router. Bikeshedding change but annoying, let's just merge it

stangri:
Closing this in favour of https://github.com/openwrt/luci/pull/7904.

Issue #7819, opened 2025-06-15
Cannot input custom value on android chromium


ledlamp:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### screenshots or captures

![Image](https://github.com/user-attachments/assets/cc208aab-5649-4c39-a0e7-9e54e922fb7c)

### Actual behaviour

when I try to input custom value on my android phone, nothing happens when I press the enter/return key on the touch keyboard

### Expected behaviour

pressing enter/return is the only way to use the custom value but it's not working on mobile

### Steps to reproduce

go to a value input with a dropdown of suggested values like the one in the screenshot and try to input custom value

### Additional Information

```shell
OpenWrt 24.10.1 r28597-0425664679 / LuCI openwrt-24.10 branch 25.161.59576~cb94dcc
```

### What browsers do you see the problem on?

Chrome

### Relevant log output

```shell

```

ledlamp:
It works on Firefox, it doesn't work in Chrome or Vivaldi

Ramon00:
Have a look at the PR I just created [https://github.com/openwrt/luci/pull/7876](https://github.com/openwrt/luci/pull/7876). This should fix the issue. Please test.

PR #7876, opened 2025-07-25
luci-base: allow submitting new value in dropdown lists in android chromium

Ramon00:
Update ui.js to allow submitting from dropdown on android chromium.

Fixes: [https://github.com/openwrt/luci/issues/7819](https://github.com/openwrt/luci/issues/7819)

P.S. maybe we should also fix the `ev.keyCode` as this may also be unreliable (and using e.g. ev.key is more readable)


systemcrash:
Reasonable. Perhaps end the line with a `,` so future change deltas are smaller. Let's see if we get feedback about the fix from the issue.

Ramon00:
yeah let me add a `,`

systemcrash:
Merged. Thanks @Ramon00

1715173329:
This is awesome! I can confirm it fixed the issue.

PR #7879, opened 2025-08-01
luci-app-pbr: sync with principal package

stangri:
Maintainer: me
Compile tested: x86_64, Dell EMC Edge 640vn, OpenWrt 24.10.2
Run tested: x86_64, Dell EMC Edge 640vn, OpenWrt 24.10.2

Description:
Depends on https://github.com/openwrt/packages/pull/27134

### DIFF
--- a/applications/luci-app-pbr/htdocs/luci-static/resources/pbr/status.js
+++ b/applications/luci-app-pbr/htdocs/luci-static/resources/pbr/status.js
@@ -10,12 +10,15 @@ var pkg = {
 	get Name() {
 		return "pbr";
 	},
+	get LuciCompat() {
+		return 14;
+	},
 	get ReadmeCompat() {
 		return "1.1.8";
 	},
 	get URL() {
 		return (
-			"https://docs.openwrt.melmac.net/" +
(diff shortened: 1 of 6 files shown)


applications/luci-app-pbr/htdocs/luci-static/resources/pbr/status.js
```diff
@@ -135,50 +120,70 @@ var status = baseclass.extend({
 	render: function () {
 		return Promise.all([
 			L.resolveDefault(getInitStatus(pkg.Name), {}),
-			//			L.resolveDefault(getGateways(pkg.Name), {}),
+			L.resolveDefault(getUbusInfo(pkg.Name), {}),
 		]).then(function (data) {
```
systemcrash, line 124: since you're returning an array of promises, you can just destructure into an array here like so:

`		]).then(function ([initStatus, ubusInfo]) {`

then you avoid all of the confusing data[x] stuff later on.

stangri: Thank you so much, this is definitely more readable, I'll include it in the next PR!

PR #7881, opened 2025-08-01
luci-mod-battstatus: preparing for other devices

rbqvq:
For standard devices, we can read from `/sys/class/power_supply`

Add board checks to implement different solutions for different boards

### DIFF
--- a/modules/luci-mod-battstatus/Makefile
+++ b/modules/luci-mod-battstatus/Makefile
@@ -7,9 +7,9 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=LuCI Battery Status
-LUCI_DESCRIPTION:=Provides a battery charge indicator in LuCI. Currently only supports the HooToo HT-TM05 travel router.
+LUCI_DESCRIPTION:=Provides a battery charge indicator in LuCI.
 
-LUCI_DEPENDS:=+luci-base +libiwinfo-lua +rpcd-mod-iwinfo +libi2c +i2c-tools
+LUCI_DEPENDS:=+luci-base
 
 PKG_LICENSE:=Apache-2.0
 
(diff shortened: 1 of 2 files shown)


modules/luci-mod-battstatus/Makefile
```diff
@@ -7,9 +7,9 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=LuCI Battery Status
-LUCI_DESCRIPTION:=Provides a battery charge indicator in LuCI. Currently only supports the HooToo HT-TM05 travel router.
+LUCI_DESCRIPTION:=Provides a battery charge indicator in LuCI.
 
-LUCI_DEPENDS:=+luci-base +libiwinfo-lua +rpcd-mod-iwinfo +libi2c +i2c-tools
+LUCI_DEPENDS:=+luci-base
```
systemcrash, line 12: surely the i2c stuff is needed if it's called in the rpc script

rbqvq: Most device use `cat /sys/class/power_supply/BAT0/capacity` are ok
Not ALL device using i2c, let user install it instead of force bundle it

systemcrash: Yet your code does not handle the `cat /sys/class/power_supply/BAT0/capacity` case. Care to add it?

rbqvq: > Yet your code does not handle the `cat /sys/class/power_supply/BAT0/capacity` case. Care to add it?

It just a prepare, I'm write for a out of tree device.
I think most device with battery can use it command.
But my battery device is not in mainline.

rbqvq: And different boards have different names. On x86, it's BAT0, and on msm8916, it's pm8916-*

systemcrash:
Closed by d42dce3ea29762ae85452bf7eebcda41e4bbc4b5

PR #7898, opened 2025-08-11
Luci-App-Unbound Stat Scroll Behavior

JimMatthew:
Currently, when viewing the Unbound Stats under Services -> Unbound DNS -> status, the text in the text area scrolls to the bottom on each poll update. This behavior makes sense when viewing a log file, but not when viewing statistics where you would want your position to remain between poll intervals. 
This commit removes the scroll to bottom behavior on the status text area.


feckert:
@JimMatthew CI/CD is not happy https://github.com/openwrt/luci/actions/runs/16870343079/job/47783790304?pr=7898 Please fix it

JimMatthew:
> @JimMatthew CI/CD is not happy https://github.com/openwrt/luci/actions/runs/16870343079/job/47783790304?pr=7898 Please fix it

Fixed

JimMatthew:
@systemcrash Since you converted / wrote this, what do you think about this change? With the current behavior, it makes it difficult to look at unbound stats because every couple seconds it scrolls to the bottom of the stat text area.

PR #7904, opened 2025-08-12
luci-base: remove reset button from sysauth

stangri:
Derived from this: https://github.com/openwrt/luci/pull/7867

### DIFF
--- a/modules/luci-base/ucode/template/sysauth.ut
+++ b/modules/luci-base/ucode/template/sysauth.ut
@@ -36,7 +36,6 @@
 
 	<div class="cbi-page-actions">
 		<input type="submit" value="{{ _('Log in') }}" class="btn cbi-button cbi-button-apply" />
-		<input type="reset" value="{{ _('Reset') }}" class="btn cbi-button cbi-button-reset" />
 	</div>
 </form>
 


stangri:
requesting more reviewers as I very infrequently open PRs against core components and definitely do not want to merge it myself, as little of a change as it is.

systemcrash (approved):

hnyman:
I use the default Bootstrap theme, so I haven't noticed the whole reset button issue at login there.
And to be honest, the original button order thing/discussion in the other PR seemed funny. Personally I have no clear preference for the button order.  Well, maybe I would prefer first the positive "ok", then "cancel"...

@jow- is maybe the authority here, who also knows the possible fallout possibilities from the ucode change that you are proposing.

stangri:
> I use the default Bootstrap theme, so I haven't noticed the whole reset button issue at login there. And to be honest, the original button order thing/discussion in the other PR seemed funny. Personally I have no clear preference for the button order. Well, maybe I would prefer first the positive "ok", then "cancel"...

So the original discussion had screenshots that in the firmware update WebUI pages the "action" (Upload, Update) button was on the right and the cancel button on the left, contradicting your preference.

> @jow- is maybe the authority here, who also knows the possible fallout possibilities from the ucode change that you are proposing.

Yeah, I can also patch the themes instead but if it is fixed in ucode it might give a clearer direction to theme creators/maintainers. But consensus in a previous PR was that the reset button is unneeded and is disabled in most themes anyways.

stangri:
Unless there are objections, I'll merge this by mid-October.

systemcrash:
Merged. Thanks @stangri !

PR #7913, opened 2025-08-22
luci-base: don't render until luci-loaded is done

wryun:
Because the setupDOM/initDOM methods do strange things related to the old lua server rendered templates
(like remove all the elements currently hidden by dependencies...), we want to be sure that these have finished before the view itself renders.

This also ensures that any caching (e.g. from probeSystemFeatures) is finished before the view render.

This is a hacky fix that we've been using on 23.05, but it's not clear to me what the best one is. I'm raising it as a draft for comment; if someone thinks this is an acceptable solution, I'll add a simple reproduction of a bug and test on master.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/luci.js
+++ b/modules/luci-base/htdocs/luci-static/resources/luci.js
@@ -1893,6 +1893,15 @@
 			DOM.content(vp, E('div', { 'class': 'spinning' }, _('Loading view…')));
 
 			return Promise.resolve(this.load())
+				.then(function (...args) {
+					if (L.loaded) {
+						return Promise.resolve(...args);
+					} else {
+						return new Promise(function (resolve) {
+							document.addEventListener('luci-loaded', resolve.bind(null, ...args), { once: true });
+						});
+					}
+				})
 				.then(LuCI.prototype.bind(this.render, this))
 				.then(LuCI.prototype.bind(function(nodes) {
 					const vp = document.getElementById('view');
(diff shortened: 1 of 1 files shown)


jow-:
The change looks sensible to me

systemcrash:
Seems reasonable

wryun:
Thanks! I'll try to find time to test this on master in the next week and give a repro of a problem.

systemcrash:
ping

wryun:
Thanks for the ping; sorry that I haven't managed to get to this. Next likely time, sadly, is going to be in another week and a half; my next attempt at accountability :)

systemcrash:
This might be a fix for 7698 where they mention a race condition which evidently arises on occasion.

wryun:
I've tested this on master and it seems good. To generate a bug easily, I made a new view:

```
'use strict';

'require view';
'require form';
'require uci';

return view.extend({
        render: function () {
                const m = new form.Map('system', 'System');
                const s = m.section(form.TypedSection, 'system');
                s.option(form.Value, 'hostname', 'Hostname');
                const o = s.option(form.Value, 'disappear', 'Disappearing');
                o.depends('hostname', 'foo');
                return m.render();
        },
});
```

Then tried to access that view directly without logging in, forcing probeSystemFeatures etc. to be called. This meant that the render was called before the load finished, and the failed dependency was entirely removed from the DOM (such that even if I wrote 'foo' into hostname it wouldn't appear). However, on refreshing the page it appeared.

When I replaced luci.js with my modified luci.js, this issue went away.

systemcrash:
Good! Thanks for you diligence @wryun.

I’ll take it for a spin.

PR #7747, opened 2025-04-23
treewide: vectorise iconography

systemcrash:
- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [x] Tested on: 24.10.1, firefox, safari :white_check_mark:
- [x] \( Preferred ) Screenshot or mp4 of changes:
- [x] Description: (describe the changes proposed in this PR)

ping @hnyman @stokito @dannil @castillofrancodamian @GeorgeSapkin @Ramon00 @stangri for some feedback. 

Screenshot 2025-06-10 at 21 32 16

Screenshot 2025-06-10 at 21 33 31

Screenshot 2025-06-10 at 21 30 05

Screenshot 2025-06-10 at 21 36 25

Screenshot 2025-06-10 at 21 29 34

Screenshot 2025-06-10 at 21 38 00

PR #7917, opened 2025-08-23
themes: scale down icons for network interfaces

dbnk0:
Many LuCI icons doubled in size during migration to vector graphics in commit ae5d91da903b6f1d3086d6082ca622231e34f555 (16px to 32px).

Sizing of their instances is mostly controlled by CSS, but there's an exception - network interface boxes of *Interfaces* page in LuCI. Current CSS doesn't specify any particular size requirements to follow, so the icons just scale with the images served, thus effectively doubling in width and height compared to state before the vectorization commit. Such a big icons look odd and take up too much space, especially for bridge interfaces with many ports.

Instead of reverting to the original 16×16px, this commit proposes compromise of 24×24px as most of other icons within LuCI became a bit bigger as well.

Changes were tested on all 4 themes using OpenWrt 24.10.2 (with a custom build of LuCI packages).
Selector `#cbi-network-interface` is necessary otherwise *Port status* on *Overview* page gets smaller icons as well.

<table>
  <tr>
    <th>Before this commit</th>
    <th>After this commit</th>
  </tr>
  <tr>
    <td colspan="2">Desktop</td>
  </tr>
  <tr>
    <td>luci_network_desktop_before</td>
    <td>luci_network_desktop_after</td>
  </tr>
  <tr>
    <td colspan="2">Mobile</td>
  </tr>
  <tr>
    <td>luci_network_mobile_before</td>
    <td>luci_network_mobile_after</td>
  </tr>
</table>

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -2188,6 +2188,11 @@ th[data-sort-direction="desc"]::after { content: "\a0\25bc"; }
 	white-space: normal;
 }
 
+#cbi-network-interface .ifacebox img {
+	width: 24px;
+	height: 24px;
+}
+
 .ifacebadge {
 	display: inline-block;
 	flex-direction: row;
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -2085,6 +2085,11 @@ td > .ifacebadge,
 	padding: .25em;
 }
 
+#cbi-network-interface .ifacebox img {
+	width: 24px;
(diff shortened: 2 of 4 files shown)


dbnk0:
Ping @systemcrash for feedback (as an author of the original *vectorize iconography* PR #7747).

systemcrash:
I'm not against this, but I don't see why we should reduce the size of vector graphics when so much screen real-estate is available. 32x32 is a nice size. The whole point of vector graphics and the size bump is so we don't have to squint to see what the device is any longer. Manual zoom is available but it's an additional step.

dbnk0:
I get your point, but I personally find the new icons too big compared to the near-by text. It just doesn't look right to have a 32px icons surrounded by 13px brackets. My proposed change is a compromise between original tiny icons and new big ones.

dbnk0:
Ping @systemcrash

PR #7919, opened 2025-08-24
luci-app-advanced-reboot: update to 1.1.1

stangri:
Maintainer: me
Compile tested: x86_64, Linksys MX4200v1, OpenWrt 24.10.2
Run tested: x86_64, Linksys MX4200v1, OpenWrt 24.10.2

Description:
This is a massive overwrite of the RPCD code from shell script to ucode which in turn brought:
* update to the device json schema, which is now:
  * more structured
  * no weird names/object in json
  * customizable get, set, save commands per device (still need to be defined in ACL)
  * single partitions array now holding all data for toggling/labels/altMount
  * support for more than 2 bootable partition devices
* a _device_json_transform.jq file to translate schema from old format to new one for any WIP devices
* simplification of RPCD script as object can now be returned
* adjustments to the Javascript to parse new error messages and new RPCD reply structure

I have tested both UBI volumes and dd-based information gathering on alternative partition, I'd still welcome:
* tests on as many Linksys devices for actual partition toggling (rebooting to alternative)
* tests for both information gathering and partition toggling on:
  * "d-link,dgs-1210-28"
  * "mercusys,mr90x-v1"
  * "zyxel,nbg6817"

I also gladly welcome any code reviews for both updated javascript and ucode.


stangri:
I've addressed the issues discovered in testing, unless there's any feedback it should be ready to merge.

If @systemcrash @stokito @champtar @Ramon00 or anyone else have time to review and comment on the new code, I'd be much obliged.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,419 @@
...
+}
+
+/* Read /tmp/sysinfo/board_name */
+function get_board_name() {
+	let s = fs.readfile('/tmp/sysinfo/board_name');
+	if (!s) return null;
+	/* take the first line, drop trailing CR if present */
+	let line = split(s, "\n")[0] ?? "";
+	let parts = split(line, "\r");
+	return parts[0] ?? "";
```
systemcrash: just wrap fs.readfile in a `trim()` https://ucode.mein.io/module-core.html#trim

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,419 @@
...
+function file_exists(p) {
+	try { return fs.stat(p) != null; } catch (e) { return false; }
+}
+
+function str_trim(s) {
+	if (s == null) return s;
+	/* Greedy capture to reliably trim leading/trailing whitespace */
+	let m = match("" + s, /^\s*(.*)\s*$/);
+	return (m && m[1] != null) ? m[1] : ("" + s);
+}
```
systemcrash: replace usage with `trim()`

systemcrash:
Perhaps some more comments for each functions which give examples of sane input values - useful when following each step.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,419 @@
...
+		if (!boards) continue;
+		if (type(boards) != 'array') boards = [ '' + boards ];
+		for (let i = 0; i < length(boards); i++) {
+			let s = '' + boards[i];
+			if (s == rb)
+				return obj;
+		}
+	}
+	return null;
+}
```
systemcrash, line 130: I recommend a single mono JSON file containing all boards - faster read, skip the array checks, and you check the device.board anyway. 

```json
[
	{
		"device": {
			"vendor": "ZyXEL",
			"model": "NBG6817",
			"board": [
				"nbg6817",
				"zyxel,nbg6817"
			]
		},
		...
	},
	{
		"device": {
			...
		},
		...
	},
	...
]
```

stangri: Thank you for review @systemcrash, I believe this was discussed before outside of this repo and my reasons to keep individual per-device json files are:
1. ease of contributions: contributors can create a single new file on their device/in PR instead of editing a large json file.
2. break-proof: even if an individual json file is corrupted/misbalanced somehow, the rest of the devices/jsons will still work.
3. ease of identifying supported devices: by glancing in the devices directory it's easy to confirm if the device is supported or not.

If there's a strong preference from the other contributors/core team for merging individual device jsons into a single main file, I wouldn't be opposed to PR which would allow both main json and individual jsons (for points 1 and 2 above) to work.

stangri:
> Perhaps some more comments for each functions which give examples of sane input values - useful when following each step.

sorry do you mean the RPC functions or all functions in ucode?

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: thank you, I'll try to implement this and other suggestions some time this week.

stokito: The board_name should be generated so it unlikely to have spaces

applications/luci-app-advanced-reboot/root/usr/share/advanced-reboot/devices/xiaomi-ax3600.json
systemcrash, line 1: nitpick: These two xiaomi profiles can change indent to be consistent with the others.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
systemcrash: Fair points. Although:

1. Editing any file still requires consistency.
2. You want to know when things break asap, to identify all errors.
3. Text search? This seems a non-issue. They're also documented in your docs?

This repo has a JSON check - so consistency is fairly assured. Moments after that are: structure which you've detailed, and; accuracy of data which is down to the submitter.

You're not bound to keeping a single file - you can maintain individual files and run jq to output a single file (and install both variants):

`jq -s '.' *.json > devices.json`

Code effectively: 
```js
function find_device_info(romBoardName) {
	let rb = str_trim(romBoardName ?? "");

	// Read one big JSON file instead of many small ones
	let txt = fs.readfile(DEVICES_DIR + "devices.json");
	if (!txt)
		return null;

	// devices.json : [ { device: {...} }, { device: {...} }, ... ]
	let arr = json(txt);
	if (!arr || type(arr) != 'array')
		return null;

	// Loop over each object in the array
	for (let obj in arr) {
		if (!obj || !obj.device) continue;

		let boards = obj.device.board;
		if (!boards) continue;

		// normalise to array
		if (type(boards) != 'array')
			boards = [ '' + boards ];

		for (let i = 0; i < length(boards); i++) {
			let s = '' + boards[i];
			if (s == rb)
				return obj;
		}
	}

	return null;
}
```

systemcrash:
> > Perhaps some more comments for each functions which give examples of sane input values - useful when following each step.
> 
> sorry do you mean the RPC functions or all functions in ucode?

ucode.

applications/luci-app-advanced-reboot/root/usr/share/advanced-reboot/devices/xiaomi-ax3600.json
stangri: thank you for spotting this, we're on the same page about it. ;)

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,483 @@
...
+
+	let out_parts = [];
+	for (let i = 0; i < length(parts); i++) {
+		let p = parts[i] ?? {};
+		let num = p.number != null ? p.number : i;
+		let mtd = p.mtd ?? null;
+		let info = null;
+		let v = null;
+
+		if ("" + curvals[0] == "" + p.param_values[0]) {
```
systemcrash, line 439: This looks like a logic error. Are you checking length or whether you can successfully concatenate a string?

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,483 @@
...
+			}
+		}
+		/* raw-label fallback if needed */
+		if (mtd && p.labelOffsetBytes != null) {
+			if (
+				!info ||
+				info.label == null ||
+				info.label == "" ||
+				info.os == null ||
+				info.os == ""
```
systemcrash: not 100% here (with ucode) but try this?:
```js
if (!info?.label || !info?.os)
```

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,483 @@
...
+		},
+		partitions: out_parts,
+	};
+}
+
+function boot_partition(req) {
+	/* extract target partition number from RPC args */
+	let number = null;
+	if (req && req.args && req.args.number != null) number = int(req.args.number);
+	if (number == null || number != number)
```
systemcrash: `number != number`?

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: failsafe to compare string to either string or number (coming from json).

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,483 @@
...
+		}
+		if (!kver) {
+			let r = command("uname -r 2>/dev/null");
+			if (r) kver = r;
+		}
+	}
+	return { label: label, os: kver };
+}
+
+/* Read current OS name from /etc/os-release, normalized for snapshot */
```
systemcrash: Are you sure? :)

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
systemcrash: The if expression evaluates a boolean result, which seemingly would always be true as a result of `+ p.param_values[0]`...?

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: integrated with revision 7, thank you!

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: accidental vscode autocomplete. it still wants to add it after I removed it.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: do snapshots no longer use /etc/os-release ?

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: seems to work as intended and only evaluates if curvals[0] matches p.param_values[0] tho.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: @stokito trim is for newline/line-break if exists in the file.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
systemcrash: not if it's a comment for
```js
function get_partition_info_current() {
	return get_volume_info("/");
}
```

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
systemcrash: sorry, I see the logic now :D

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: yeah, there used to be two different functions for current and alt info getters and now they use mostly the same code if the alt partition's UBI can be mounted. if you can suggest a more elegant code, please do.

stangri:
Still working on more comments for still "undocumented" functions as per @systemcrash advise.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
systemcrash: My point was: is `/* Read current OS name from /etc/os-release, normalized for snapshot */` a suitable comment for a function that does `return get_volume_info("/");`?

systemcrash:
> Still working on more comments for still "undocumented" functions as per @systemcrash advise.

A few that lack comments is OK - but a cursory hint to what most of the functions do is really helpful, especially to reviewers :)

systemcrash:
LGTM

stangri:
@systemcrash updated the ucode with (mostly AI-generated) function descriptions.

I've also tried to adjust the Makefile to generate the devices.json during build and remove no longer needed directories when building the package.

systemcrash:
ping @stangri

stangri:
@systemcrash I'll be back mid-December, so hoping to push latest fix(es) and merge between then and X-mas.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,579 @@
...
+				};
+			if (savecmd) rc = command(savecmd + " 2>/dev/null; echo $?");
+			if (rc != "0") return { error: "ERR_SAVE_ENV", rom_board_name: board };
+		}
+	} else {
+		let df = read_dual_flag_mtd();
+		if (!df) return { error: "NO_DUAL_FLAG", rom_board_name: board };
+		if (!file_exists(df))
+			return { error: "NO_DUAL_FLAG_BLOCK", rom_board_name: board };
+		if (!write_dual_flag_value(df, target.param_values[0]))
```
systemcrash, line 567: could be undefined. Above you do `let value = target.param_values ? target.param_values[j] : null;` but here you assume it exists.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,579 @@
...
+ * Also removes possible ubiblock nodes for the device.
+ *
+ * @param {Number|String} op_ubi   UBI mtd index previously attached.
+ * @param {Number|String} ubi_vol  UBI volume index (ignored during cleanup).
+ * @returns {void}
+ * @sideeffects  umounts /var/alt_rom, ubidetach, removes ubiblock nodes.
+ */
+function alt_partition_unmount(op_ubi, ubi_vol) {
+	let mtdCount = command(
+		'ubmtdCount | grep "Present UBI devices" | tr "," "\\n" | grep -c "ubi"'
```
systemcrash: is `ubmtdCount` supposed to be available on all mtd devices? I can't see it but there's `ubinfo` which returns the same output.

stangri:
Still in the pipeline. I believe I need to update PR with some fixes and address @systemcrash  questions before it can be merged.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: Thanks for spotting it, surprised it didn't show up in testing, was an erroneous auto-complete, should have been `ubinfo`, I'll push a fix.

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
stangri: there's `if (value == null) continue;` above that. does this address your concern or did I misunderstand something?

lorand-horvath:
@stangri @systemcrash This new `luci-app-advanced-reboot` seems to be broken on my Linksys MR5500. There are at least 3 issues that I've noticed:

1. Trying to reboot to alternative partition - clicking "Reboot to this partition" will return an error: `Invalid request: number is required and must be numeric.`, see screenshot:
image

2. The firmware build version is not detected anymore on the current partition, it previously said something like "OpenWrt SNAPSHOT r32598", now there's no more version specified.

3. The confirmation popup has broken text (missing spaces), i.e. the backslashes in the code to break the long text into separate lines do not work as intended:
image

Can this commit be reverted please?

systemcrash:
Paste the JSON output from the browser console produced when you click 'reboot'; its accompanying 'request' and 'response' from the request respective response tab of the POST request.

applications/luci-app-advanced-reboot/htdocs/luci-static/resources/view/system/advanced-reboot.js
```diff
@@ -190,31 +216,43 @@ return view.extend({
...
-				As your network configuration and WiFi SSID/password on alternative partition might be different,\
-					you might have to adjust your computer settings to be able to access your device once it reboots.<br /><br />\
-				Please also be aware that alternative partition firmware might not provide an easy way to switch active partition\
-					and boot back to the currently active partition.<br /><br />\
-				Click "Proceed" below to reboot device to an alternative partition.'
+As your network configuration and WiFi SSID/password on alternative partition might be different,\
+you might have to adjust your computer settings to be able to access your device once it reboots.<br /><br />\
+Please also be aware that alternative partition firmware might not provide an easy way to switch active partition\
+and boot back to the currently active partition.<br /><br />\
+Click "Proceed" below to reboot device to the selected partition.'
```
systemcrash, line 255: Perhaps just rework this to 
```
'...' + 
'...' + 
```
But I recommend splitting strings to each sentence. e.g.
```
_('sentence.') + '<br />' +
...
```

lorand-horvath:
> Paste the JSON output from the browser console produced when you click 'reboot'; its accompanying 'request' and 'response' from the request respective response tab of the POST request.

@systemcrash There's no error in the console. The request/response looks like this:

request:
`[{"jsonrpc":"2.0","id":5,"method":"call","params":["ff63eb89442e15594fc23d05689c5bbb","luci.advanced-reboot","boot_partition",{"number":"1"}]}]`

response:
```
[
    {
        "jsonrpc": "2.0",
        "id": 5,
        "result": [
            0,
            {
                "error": "INVALID_ARG",
                "detail": "number is required and must be numeric"
            }
        ]
    }
]
```

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,578 @@
...
+ *
+ * @param {{args:{number:String|Number}}} req
+ * @returns {Object} {} on success; {error:..., ...} on failure.
+ * @sideeffects  Calls fw_setenv/fw_saveenv or writes to dual-flag MTD.
+ */
+function boot_partition(req) {
+	/* extract target partition number from RPC args */
+	let number;
+	if (req && req.number != null) number = int(req.number);
+	if (!number)
```
systemcrash, line 507: Perhaps: 
```js
if (!req?.number) {
 return... 
```
and skip the coercion (int(x)).

applications/luci-app-advanced-reboot/root/usr/share/rpcd/ucode/luci.advanced-reboot
```diff
@@ -0,0 +1,578 @@
...
+			return { error: "NO_DUAL_FLAG_BLOCK", rom_board_name: board };
+		if (!write_dual_flag_value(df, target.param_values[0]))
+			return { error: "ERR_SET_DUAL_FLAG", args: [df], rom_board_name: board };
+	}
+	return {};
+}
+
+const methods = {
+	obtain_device_info: { call: obtain_device_info },
+	boot_partition: { args: { number: "String" }, call: boot_partition },
```
systemcrash, line 575: Perhaps define the signature using an int here:
```js
	boot_partition: { args: { number: 0 }, call: boot_partition },
```

applications/luci-app-advanced-reboot/htdocs/luci-static/resources/view/system/advanced-reboot.js
```diff
@@ -231,7 +269,32 @@ return view.extend({
 								"button",
 								{
 									class: "btn cbi-button cbi-button-positive important",
-									click: L.bind(this.handleTogglePartition, this),
+									click: L.bind(function () {
+										this.callBootPartition(String(pn))
```
systemcrash, line 273: And perhaps just pass the int here directly.

systemcrash:
@lorand-horvath what happens if you call from the terminal like so:
`ubus call luci.advanced-reboot boot_partition '{ "number": "1" }'` or `ubus call luci.advanced-reboot boot_partition '{ "number": 1 }'`?

lorand-horvath:
```
# ubus call luci.advanced-reboot boot_partition '{ "number": "1" }'
{
        "error": "INVALID_ARG",
        "detail": "number is required and must be numeric"
}
```

```
# ubus call luci.advanced-reboot boot_partition '{ "number": 1 }'
Command failed: Invalid argument
```

lorand-horvath:
@stangri @systemcrash  This is a mess. Please revert the commit until there's an actual, tested, functional version.

lorand-horvath:
@stangri Care to say something? The package is completely broken. You shouldn't leave it in this state, it's just not fair with respect to the community. @systemcrash Please revert.

ghost:
@systemcrash 
```if (req && req.number != null) number = int(req.number);```
a better option might be:
```if (req && req.args &&  req.args.number != null) number = int(req.args.number);```

I believe that req.**_args_**.number is the key here. After this change it works for two different SOCs.

@stangri Thanks for this version. I can drop my custom patch and just add a json file and the basic functionality work. (for now I'm patching the above, but I'm expecting this to be solved eventually)

**Comment by systemcrash** [comment]
@lorand-horvath if you can confirm that fixes things for you, I can push it out.

**Comment by lorand-horvath** [comment]
@stangri @systemcrash I have installed the latest version 1.1.1-15 from https://github.com/stangri/luci-app-advanced-reboot/releases and it seems to be working fine (I've only tested rebooting to Alternative partition).
```
apk add --allow-untrusted luci-app-advanced-reboot-1.1.1-15_openwrt-25.12_noarch.apk
```
image

hnyman:
@stangri 
Have you noticed this error visible in "make defconfig" etc. configuration actions:

> WARNING: Makefile 'package/feeds/luci/luci-app-advanced-reboot/Makefile' has a build dependency on 'jq/host', which does not exist

stangri:
@hnyman I don't really use full builds, before PR I compile with SDK and rely on CI to figure out any errors. Is `jq` installed on host?

hnyman:
> Is jq installed on host?

No idea. And should have no relevancy.
But you declared a dependency for that host build to be in the OpenWrt build root, although the jq package itself does not declare a host build. So, there is no proper host build.

https://github.com/openwrt/packages/blob/master/utils/jq/Makefile

There is no host build defined in the Makefile.
And for that reason all config evaluations throw a hiccup to your requirement of that `jq/host` .

Error is visible e.g. in all buildbot builds.
https://buildbot.openwrt.org/images/#/builders/23/builds/514/steps/21/logs/stdio

systemcrash:
jq is not available AFAICT on host build. It's available as a package for users tho. That should be removed from host deps. Just gotta jq the JSON before releases :/

stangri:
> jq is not available AFAICT on host build. It's available as a package for users tho. That should be removed from host deps. Just gotta jq the JSON before releases :/

jq on host is very useful tho, I was thinking of merging the luci-app-https-dns-proxy providers files in Makefile same way, so I've created a PR to build it on host and it's linked just above your comment. ;)

systemcrash:
OK - fair! jq is useful - though big :)

stangri:
> OK - fair! jq is useful - though big :)

Hehe, I blame you, it was your idea and your code to merge json files. ;)

systemcrash:
Yeah, a one-time thing. :)

Issue #7210, opened 2024-07-25
luci-mod-network: IPv6 traceroute timeouts in extreme condition


emi-YFcz5XDj:
## Steps to reproduce:

1. Open Network > Diagnostics
2. Traceroute `4321::1234`
3. Wait

## Actual behavior:

Error: XHR request timed out

## Expected behavior:

Traceroute outputs appear in the text box

## Additional Information:

```
NAME="OpenWrt"
VERSION="SNAPSHOT"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt SNAPSHOT"
VERSION_ID="snapshot"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r26992-bbcd5111cb"
OPENWRT_BOARD="mediatek/filogic"
OPENWRT_ARCH="aarch64_cortex-a53"
OPENWRT_TAINTS=""
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt SNAPSHOT r26992-bbcd5111cb"
```

neehhaa06:
Hello! @emi-YFcz5XDj 

It might be helpful for this issue:

fix: #7210

# Quick Fixes

>    Disable uHTTPd Connection Reuse

>    Try a Different Browser

>    Check System Logs

>    Increase RPC Timeout

>    Reboot the Router

>    Update OpenWrt

>    Run Traceroute via SSH

>    Seek Community Help

PR #7920, opened 2025-08-25
luci-base, luci-mod-network: diagnostics: add realtime command output

ysc3839:
- [x] Tested on: (x86-64, snapshot, Chrome 139) :white_check_mark:
Before:

https://github.com/user-attachments/assets/9ad4bb0d-df8b-47da-8718-eb96c657ba89

https://github.com/user-attachments/assets/95e0e895-9e1b-41b9-8ec9-41ecbf5e760d

After:

https://github.com/user-attachments/assets/e306d254-0a19-4a6c-92f1-7f535fc4f3e3
- [x] \( Optional ) Closes: e.g. openwrt/luci#7210
- [x] \( Optional ) Depends on a patch for `cgi-io`: openwrt/cgi-io#2
- [x] Description:

The diagnostics page is using `fs.exec()` to execute command, which blocks until command exit. Users have to wait for a while to see the result.

When doing IPv6 traceroute, this may cause XHR timeout, if target host is unreachable. (See issue #7210)

This commit adds XHR response progress event option in luci.js and fs.js. And use this new event in diagnostics.js to update command output.

This commit also adds a `stderr` option in `fs.exec_direct()`, and requires a patch for `cgi-io`.

Currently `cgi-io` redirects stderr to /dev/null, which makes users unable to see error messages. The patch adds a `stderr` option, and if it's set to `1`, redirects stderr to stdout.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
@@ -8,15 +8,16 @@
 
 return view.extend({
 	handleCommand: function(exec, args) {
-		var buttons = document.querySelectorAll('.diag-action > .cbi-button');
+		var buttons = document.querySelectorAll('.diag-action > .cbi-button'),
+			out = document.querySelector('textarea');
 
 		for (var i = 0; i < buttons.length; i++)
 			buttons[i].setAttribute('disabled', 'true');
 
-		return fs.exec(exec, args).then(function(res) {
-			var out = document.querySelector('textarea');
-
-			dom.content(out, [ res.stdout || '', res.stderr || '' ]);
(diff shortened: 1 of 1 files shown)


feckert:
@ysc3839 I like this feature. Here are a few comments.
* Can you split the one commit into two? One that adds the new API and one for `diagnositcs.js` that uses the new API
* Make CI/CD happy -> Formal issues
* Add a reference to the PR or patch that changes `cgi-io` and get this merged first.

ysc3839:
@feckert Updated.

systemcrash:
Looks Ok

systemcrash:
closed by 122a5ba73e50bfc6c474fcee0d352aa7fab9f05d. Thanks @ysc3839 !

PR #7930, opened 2025-09-01
luci-proto-ipv6: map: allow IPv6 prefix length up to 128

missing233:
IPv6 prefix length is valid in the range 0–128 (RFC 4291). 
LuCI currently restricts it to 0–64.
Extend to 0–128 to support valid cases like /96 and /128.


systemcrash:
The reason it doesn't go any lower is that odhcpd6 seems to lack support for splitting subnets any smaller than a /64. You might want to check up on that. Test with e.g. /96 in your config first, and see whether it even works.

missing233:
> The reason it doesn't go any lower is that odhcpd6 seems to lack support for splitting subnets any smaller than a /64. You might want to check up on that. Test with e.g. /96 in your config first, and see whether it even works.

this change is not related to odhcpd6. the logic here is handled by the `mapcalc` from the `map` package.
it's only meant for the lw4o6 case. in that mode users sometimes need to enter a /128 to set the tunnel’s local IPv6 address. it’s not about splitting subnets smaller than /64, just making the luci accept /128 for this special case.

systemcrash:
If only the special case of /128 is used, how about something like:
`'or(range(0,64),uinteger(128))'`.

I'm unfamiliar with the mapcalc and what it can take - are things other than /96 /128 also possible?

missing233:
> If only the special case of /128 is used, how about something like: `'or(range(0,64),uinteger(128))'`.
> 
> I'm unfamiliar with the mapcalc and what it can take - are things other than /96 /128 also possible?

at least for now the only special case we need to support is /128.
I’ve updated the PR
thanks for the suggestion!

PR #7940, opened 2025-09-10
luci-mod-network: add DNR (RFC9463) support

Alphix:
odhcpd has had support for Discovery of Network-designated Resolvers (DNR, RFC9463), which allows allows devices on the network to discover encrypted DNS resolvers, since December 2024 [1].

This patch adds a basic luci interface to configure DNR records. odhcp actually supports DNR records via DHCPv4, DHCPv6 and RA, but since it's typically only used for the latter two, stashing the option next to IPv6 DNS server options seems reasonable.

[1] https://github.com/openwrt/odhcpd/commit/6d342cc03bf72f27d0838942d009bf1c5df4f954

- [x] Tested on: (architecture: Turris Omnia, openwrt version: 24.10, browser: Chrome 139) :white_check_mark:

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -936,6 +936,20 @@ return view.extend({
 					so.depends('dhcpv6', 'server');
 					so.depends({ dhcpv6: 'hybrid', master: '0' });
 
+					so = ss.taboption('ipv6', form.DynamicList, 'dnr', _('Announce encrypted DNS servers'),
+						_('Specifies a fixed list of encrypted DNS server addresses to announce via DHCPv6/<abbr title="Router Advertisement">RA</abbr> (see %s).')
+						 .format('<a href="%s" target="_blank">RFC9463</a>').format('https://www.rfc-editor.org/rfc/rfc9463') + '<br/>' +
+						_('IPv4 addresses are only supported if <code>odhcpd</code> also handles DHCPv4.') + '<br/>' +
+						_('Syntax: <code>&lt;numeric priority&gt; &lt;domain-name&gt; [IP,...] [SVC parameter ...]</code>') + '<br/>' +
+						_('Example: <code>100 dns.example.com 2001:db8::53,192.168.1.53 alpn=doq port=853</code>') + '<br/>' +
+						_('Note: the <code>_lifetime=&lt;seconds&gt;</code> SVC parameter sets the lifetime of the announced server (use <code>0</code> to indicate a server which should no longer be used).')
+					);
+					so.datatype = 'string';
+					so.depends('ra', 'server');
+					so.depends({ ra: 'hybrid', master: '0' });
+					so.depends('dhcpv6', 'server');
+					so.depends({ dhcpv6: 'hybrid', master: '0' });
+
 					so = ss.taboption('ipv6', form.Flag, 'dns_service', _('Local IPv6 DNS server'),
 						_('Announce this device as IPv6 DNS server.'));
 					so.default = so.enabled;


Alphix:
Screenshot From 2025-09-11 00-57-53

Alphix:
See also https://github.com/openwrt/odhcpd/pull/213#issuecomment-1994588681

systemcrash:
One problem: when was the odhcpd package in openwrt/main last bumped?

Alphix:
> One problem: when was the odhcpd package in openwrt/main last bumped?

Oh, right:
https://github.com/openwrt/openwrt/blob/f26260c7e75112429901fff59ef0a0f5601c5675/package/network/services/odhcpd/Makefile#L16
`PKG_SOURCE_DATE:=2024-05-08`

@systemcrash who would be the right person to ask to bumbp odhcpd in the openwrt/openwrt repo? From the commit logs so far, I'd guess @Ansuel or @dedeckeh ?

systemcrash:
Tag them, but you can make a PR yourself. Although that bump script can be a tricky sometimes. I think it's this one. 

https://git.openwrt.org/?p=maintainer-tools.git;a=blob;f=update_git_source_package.sh;hb=HEAD

Alphix:
> Tag them, but you can make a PR yourself. Although that bump script can be a tricky sometimes. I think it's this one.
> 
> https://git.openwrt.org/?p=maintainer-tools.git;a=blob;f=update_git_source_package.sh;hb=HEAD

Ok, https://github.com/openwrt/openwrt/pull/20022

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -936,6 +936,15 @@ return view.extend({
 					so.depends('dhcpv6', 'server');
 					so.depends({ dhcpv6: 'hybrid', master: '0' });
 
+					so = ss.taboption('ipv6', form.DynamicList, 'dnr', _('Announced Encrypted IPv6 DNS servers'),
+						_('Specifies a fixed list of Encrypted DNS server addresses to announce via DHCPv6/RA (see %s).<br/>Format: <code>&lt;numeric priority&gt; &lt;domain-name&gt; [IP,...] [SVC parameter ...]</code><br/>Example: <code>100 dns.example.com fd01::53,192.168.1.53 alpn=doq port=853</code>',
```
systemcrash: Break this up into multiple text strings so i18n translation is easier.
e.g. 
```
_('...') + '<br/>' + 
_('...') ...
```
Also, prefer syntax to format.

Alphix: Ok, that makes it more readable

Alphix:
@systemcrash Ok, https://github.com/openwrt/openwrt/pull/20022 has been merged and the patch has been updated to address suggested changes.

Quick question: should I add one more line to the help text documenting the `_lifetime=` [fake SVC parameter](https://github.com/openwrt/odhcpd/pull/213#issuecomment-2042279480)?

systemcrash:
Seems like it should be a part of syntax, yes :)

Alphix:
> Seems like it should be a part of syntax, yes :)

@systemcrash Ok, v3 pushed with an additional line explaining the ```_lifetime=<seconds>``` SVC parameter

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -936,6 +936,19 @@ return view.extend({
 					so.depends('dhcpv6', 'server');
 					so.depends({ dhcpv6: 'hybrid', master: '0' });
 
+					so = ss.taboption('ipv6', form.DynamicList, 'dnr', _('Announced Encrypted IPv6 DNS servers'),
```
systemcrash: Can we make this `_('Announce Encrypted IPv6 DNS servers'),` which makes it sound like something we do, rather than something someone else does and we are trying to 'find'? Then it's in line with the description below it.

Alphix: Sure, the current wording was an attempt to align with the language of the surrounding options in the UI (see the screenshot above). Should I change the other options as well?

Also, another possibility would be to just remove the word "Announced" from all three options. The help text for each option already provides that information...

systemcrash: I'm glad you pointed that out - I've taken care of the other strings. Just rebase on master and go with Announce :)

Self-Hosting-Group:
Some terminology differences between IPv4 and IPv6 when configuring DNS resolvers:

**IPv4 DNS resolvers (Advanced Settings)**

1. Use custom DNS servers
2. DNS search domains


**IPv6 DNS resolvers (DHCP Server -> IPv6 Settings)**

1. Announced IPv6 DNS servers
2. Local IPv6 DNS server
3. Announced DNS domains

Self-Hosting-Group:
1. Is there a reason why the example for an encrypted IPv6 DNS resolver lists an IPv4 address (192.168.1.53)?
2. Not manually assigning the priority (using an automatic internal increment when saving) would possibly simplify the configuration
3. Note that port is optional: If no port service parameter is included, this indicates that default port numbers should be used, see https://datatracker.ietf.org/doc/html/rfc9463#section-4.1

Alphix:
> 1. Is there a reason why the example for an encrypted IPv6 DNS resolver lists an IPv4 address (192.168.1.53)?

Yeah, `odhcpd` does support IPv4 DNRs, but only if `odhcpd` is setup to handle DHCPv4 (i.e. if  the `dnsmasq` and `odhcpd-ipv6only` packages have been replaced with the `odhcpd` package, see the commit message). But I could change the title from "Announce Encrypted IPv6 DNS servers" to "Announce encrypted DNS servers" and add a blurb about this to the help text...(which is getting quite long)

> 2. Not manually assigning the priority (using an automatic internal increment when saving) would possibly simplify the configuration

I'm not sure which kind of UI you're imagining, but it would need to support both automatic and manual priority. I guess that's not a blocker to have this merged in its current state though. 

> 3. Note that port is optional: If no port service parameter is included, this indicates that default port numbers should be used, see https://datatracker.ietf.org/doc/html/rfc9463#section-4.1

Yes? The draft help text doesn't imply that a port is mandatory?

Alphix:
> Some terminology differences between IPv4 and IPv6 when configuring DNS resolvers:
> 
> **IPv4 DNS resolvers (Advanced Settings)**
> 
> 1. Use custom DNS servers
> 2. DNS search domains
> 
> **IPv6 DNS resolvers (DHCP Server -> IPv6 Settings)**
> 
> 1. Announced IPv6 DNS servers
> 2. Local IPv6 DNS server
> 3. Announced DNS domains

More consistent language would be great in general in LUCI. But I think that's a bit outside the scope of this PR.

Alphix:
New version pushed, here's what it looks like now:

Screenshot From 2025-09-19 17-56-05

systemcrash:
Merged. Thanks @Alphix

PR #7943, opened 2025-09-16
luci-mod-network: test strength of wpa_key

ckorber:
- [x] Tested on: (x86/64, openwrt-24.10, firefox) :white_check_mark:
image

image

image

This commit uses the function 'checkPassword' from `password.js` in `luci-mod-system` in a modified version to display the strength of the wpa_key.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -13,6 +13,25 @@
 
 const isReadonlyView = !L.hasViewPermission();
 
+function check_password_strength(section_id, value) {
+	var strength = document.querySelector('[data-name="_wpa_key"]'),
+	strongRegex = new RegExp("^(?=.{8,})(?=.*[A-Z])(?=.*[a-z])(?=.*[0-9])(?=.*\\W).*$", "g"),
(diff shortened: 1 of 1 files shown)


hnyman:
Looks nice (in patched live router).

My question is: should it also validate that the entered key is actually allowable?  (not just complex enough, but does it contain non-allowed chars?)

To my knowledge the rule is:
> 8 to 63 printable ASCII characters
> Each character in the pass-phrase must have an encoding in the range of 32 to 126 (decimal), inclusive.

There is already a validation tied to the 'wpakey' datatype, which checks that it is 8-63 chars, but does not take view to printable/not... So, I am wondering how the field-specific validation and the datatype validation play together.

Ps.
Here is also a screenshot with the default theme (instead of non-standard material(?) theme... )

image

systemcrash:
The field should take either a 64 character hex key (i.e. pre-computed 256bit key) or 8-63 ascii characters which will be derived to form the 64 character hex key. The wpakey type checks for these. 

Since the ascii input is put through a PBKDF2 function to produce 256 key bits, 'strength' here is effectively meaningless, especially when the minimum ascii length is anyway 8 (and the included regex have minimums of 6, 7, and 8 respectively).

I like the idea, but this is security theatre which doesn't amount to anything :/

PR #7946, opened 2025-09-16
luci-mod-network,-status: show DHCPv6 lease IAID

Alphix:
The IAID is important information as it allows the user to know which interface on the client device a given DHCPv6 lease corresponds to. odhcpd already exposes this information (e.g. via "ubus call dhcp ipv6leases"), but it would be good to have access to the same information via the web interface, especially since odhcpd does take the IAID into account when allocating addresses.

- [x] Tested on: (mvebu/cortexa9, OpenWRT 24.10, Chrome) :white_check_mark:

Screenshot From 2025-09-17 01-41-13

### DIFF
--- a/libs/rpcd-mod-luci/src/luci.c
+++ b/libs/rpcd-mod-luci/src/luci.c
@@ -355,6 +355,7 @@ struct lease_entry {
 	struct ether_addr mac;
 	char *hostname;
 	char *duid;
+	char *iaid;
 	union {
 		struct in_addr in;
 		struct in6_addr in6;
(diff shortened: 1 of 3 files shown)


feckert:
Please make CI/CD happy https://github.com/openwrt/luci/actions/runs/17782086020/job/50542883144?pr=7946

Alphix:
> Please make CI/CD happy https://github.com/openwrt/luci/actions/runs/17782086020/job/50542883144?pr=7946

@feckert Yeah, it's complaining about the title of the PR/commit. The thing is, the patch touches several modules. I scrolled through `git log` to see what previous patch authors have done in such cases, and the way I formatted it now seems to have been one common approach (see e.g. 9c7e472edb207cb18a46fd191471dcdfc2645da0).

Not sure how to proceed 😄

efahl:
Yeah, since there are no clear guidelines on the how to handle multiple specific packages, I do like the use of braces to list them (and see it with some frequency in the other repos).  But, if you just want to pass [the test](https://github.com/openwrt/luci/blob/667871d68c11c05d952a702b4ddd1478dea64712/.github/workflows/formal.yml#L60), then you could do something like this...

```
$ echo "luci-mod-network,-status: show DHCPv6 lease IAID" | grep -q -e '^[0-9A-Za-z,+/_-]\+: ' -e '^Revert ' ; echo $?
0
```

systemcrash:
change looks otherwise acceptable - perhaps just split the different components to different commits each with relevant subject line.

PR #7952, opened 2025-09-22
luci-app-acme: Fix incorrect "Days until renewal" placeholder

rightaditya:
`acme.sh` uses 60 days as its default if nothing is specified, which is the case if the field is left empty. This renewal time is also in line with Let's Encrypt's recommendation[^1].

[^1]: <https://letsencrypt.org/docs/faq/#what-is-the-lifetime-for-let-s-encrypt-certificates-for-how-long-are-they-valid>

- [x] Tested on: (x86-64, 24.10.2, Firefox 143.0) :white_check_mark:

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme.js
@@ -216,7 +216,7 @@ return view.extend({
 
 		o = s.taboption('advanced', form.Value, 'days', _('Days until renewal'));
 		o.optional    = true;
-		o.placeholder = 90;
+		o.placeholder = 'acme.sh default (60 days)';
 		o.datatype    = 'uinteger';
 		o.modalonly = true;
 


systemcrash:
While this is well intentioned, and currently correct, this will soon become outdated. The trends are downward to shorter lifetimes. We might just as well remove the placeholder.

90 -> 6 https://letsencrypt.org/2025/01/16/6-day-and-ip-certs

See also https://www.digicert.com/blog/tls-certificate-lifetimes-will-officially-reduce-to-47-days

rightaditya:
> See also https://www.digicert.com/blog/tls-certificate-lifetimes-will-officially-reduce-to-47-days

Thanks, I wasn't aware of this.

> We might just as well remove the placeholder.

I don't entirely disagree, but I think it's useful to indicate to the user what the lifetime will be if they don't enter anything. 
The `acme-acmesh` package scripts only give a value to `acme.sh` if this relevant UCI setting has a value, so without anything explicitly set, the `acme.sh` default will be used (which is currently set to 60 days). If and when that changes, the placeholder can be updated to still communicate the default to the user.

Removing the placeholder entirely does have the advantage of not giving the user the impression a placeholder value will be used indefinitely (since it's subject to `acme.sh`'s default) but to really avoid that kind of situation it'd be better to use a default value for the UCI setting rather than just a placeholder. But this would also have to be adjusted as the limits decrease over the next few years.

systemcrash:
@tohojo thoughts?

tohojo:
I don't think there's a constraint that the placeholder has to be numeric, is there? In which case, how about changing the placeholder to "acme.sh default (60 days)" or something to that effect?

hnyman:
tohojo's proposal sounds good. 
I haven't checked the code if there is any validation of placeholder, but in essence it is just text. 
So, explanatory message there sounds good to me.

The digicert link about the 398/200/100/47 days is interesting. 
(Sounds like we may need to change uhttpd related things by latest in 5-6 months. The short lifetime is not really optimal for self-signed certs, so if browsers enforce that, the situation may be uncomfortable for OpenWrt LuCI.)

rightaditya:
Good idea. I've just updated the commit accordingly and force-pushed.

PR #7953, opened 2025-09-23
luci-app-attendedsysupgrade: reorder and reword error instructions

efahl:
Many users have been posting incomplete error information to the forum and in git issues.  They'll not report important information that is above the paragraph indicating where to report it, only the partial information below it.

Remedy this by putting the reporting instructions first, and make explicit that everything below is needed.

---
Here's an "after" shot of the message window with the revised wording and reordering.

image

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -257,12 +257,75 @@ return view.extend({
 			...firmware
 		};
 		let body = [
-			E('p', {}, _('Server response: %s').format(response.detail)),
+			E('p', {}, [
+				_('First check '),
+				E(
+					'a',
+					{ href: 'https://forum.openwrt.org/t/luci-attended-sysupgrade-support-thread/230552' },
+					_('this forum thread')
+				),
+				_('.  If you don\'t find a solution there, report all of the information below.')
+			]),
+
 			E(
-				'a',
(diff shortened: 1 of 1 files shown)


systemcrash:
Pesky users :)

I wonder whether filling the paste buffer with the request data is a help? Or coalescing the request data and error message into a single contiguous paste buffer.

efahl:
Yeah, I had some thoughts about preformatting things like that but being lazy wanted to change as little as possible.  

Maybe that's the best solution though, add a "Copy Errors" button or something that fills the paste buffer with exactly what we want.  If we assume it's going to be pasted to the forum, we could even add markup with hidden sections and backticks...

~~~
[details="Error Message"]
```
Error text...
```
[/details]

[details="STDERR"]
```
blah blah
```
[/details]
~~~

systemcrash:
> fills the paste buffer with exactly what we want.

Why settle for anything less? :D

efahl:
Ok, so is the clipboard.writeText the best way to do that?  I couldn't find anything wrapped up in the LuCI UI stuff, so just bare metal js I guess?

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

systemcrash:
Exactly that one. I've used it before once.

systemcrash:
How's this coming @efahl ?

efahl:
I keep wandering off into other issues, it's still on my list...

systemcrash:
She's a beaut! Thanks @efahl

efahl:
Oh, the code is ugly as hell, lots of hard-coded constants, duplication and copy pasta in there.  I'm planning a major overhaul whenever I Get A Round Tuit™...

PR #7955, opened 2025-09-24
luci-app-nft-qos: drop this application

BKPepe:
Maintainer is no longer active and nft-qos does not work for some time. The last relevant change was in 2022 and since then it does not get any update.

In packages repo, it will be removed in https://github.com/openwrt/packages/pull/27512


BKPepe:
This can be merged.

rezad1393:
I used this package to see individual lan/wifi clients network bandwidth usage via this address http://my-luci-domain.lan/bin/luci/admin/status/realtime/rate

is there anyway to still access that info via other packages?

systemcrash:
The current status pages?

rezad1393:
> The current status pages?

??
there is no real time bandwidth usage based on lan ip in status pages.

systemcrash:
Present bandwidth per client likely requires some integration maths to figure out, but the data is there, I think, if you poll nft, though that only shows data through the firewall.

PR #7956, opened 2025-09-24
luci-app-natmap: sync with natmap package

heiher:
- [x] \( Preferred ) Mention: @ysc3839 
- [x] \( Optional ) Depends on: e.g. openwrt/packages#27515
- [x] Description: Sync with natmap package.

### DIFF
--- a/applications/luci-app-natmap/htdocs/luci-static/resources/view/natmap.js
+++ b/applications/luci-app-natmap/htdocs/luci-static/resources/view/natmap.js
@@ -76,23 +76,37 @@ return view.extend({
 		o.modalonly = true;
 
 		o = s.option(form.Value, 'interval', _('Keep-alive interval'));
-		o.datatype = 'uinteger';
+		o.datatype = 'and(uinteger, min(1))';
 		o.modalonly = true;
 
+		o = s.option(form.Value, 'stun_cycle', _('STUN check cycle'), _('For UDP mode'));
+		o.datatype = 'uinteger';
+		o.modalonly = true;
+		o.depends('udp_mode', '1');
+
 		o = s.option(form.Value, 'stun_server', _('STUN server'));
-		o.datatype = 'host';
+		o.datatype = 'string';
 		o.modalonly = true;
 		o.optional = false;
 		o.rmempty = false;
 
 		o = s.option(form.Value, 'http_server', _('HTTP server'), _('For TCP mode'));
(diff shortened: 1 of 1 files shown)


applications/luci-app-natmap/htdocs/luci-static/resources/view/natmap.js
```diff
@@ -79,19 +79,24 @@ return view.extend({
...
 
 		o = s.option(form.Value, 'http_server', _('HTTP server'), _('For TCP mode'));
-		o.datatype = 'host';
+		o.datatype = 'string';
 		o.modalonly = true;
 		o.rmempty = false;
 
 		o = s.option(form.Value, 'port', _('Bind port'));
-		o.datatype = 'portrange';
+		o.datatype = 'string';
```
systemcrash, line 100: why would this type change?

heiher: This supports the parameter syntax for randomly assigning ports within a specified range: https://github.com/heiher/natmap/blob/f217acdbaede2c9f7399edbf0d25770973b68155/README.md?plain=1#L64

feckert: If we don't have validation (via a new validation check function), then at least in  note there should be an info how the string should be. The LuCI user does not know the exact allowed syntax of this option.

Regardless of this, in my opinion, both together would be the right solution.

systemcrash: Please drop this change until there is proper validation for 0, port, port-range and port~range.

muink: Can add a switch to the page:
https://github.com/muink/luci-app-natmapt/blob/9c28343aa6e1ff0e620d271bc01f465776523489/htdocs/luci-static/resources/view/natmap.js#L395-L400
and automatically replace `-` with `~` during the boot phase in init.
https://github.com/muink/openwrt-natmapt/blob/89ae3a32ff480a39f0aeda4c1b0d5b80ba579f10/files/natmap.init#L183

heiher: Thanks for all the reviews and feedback. I decided to go with @muink's suggestion.

PR #7957, opened 2025-09-24
luci-mod-network: Add flag for address_as_local

markozajc:
See openwrt/openwrt#18610 for more information. These pull requests should be merged together.

- [x] Tested on: ramips/mt7621, OpenWrt 24.10.2, Firefox :white_check_mark:
- [ ] ( Preferred ) Mention: @ the original code author for feedback
- [x] ( Preferred ) Screenshot or mp4 of changes:
   image

- [ ] ( Optional ) Closes: e.g. openwrt/luci#issue-number
- [x] ( Optional ) Depends on: e.g. openwrt/openwrt#18610
- [x] Description: Adds a toggle for the behaviour described in openwrt/openwrt#18610. The toggle defaults to off (old behaviour) to avoid breaking existing workflows.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -676,6 +676,11 @@ return view.extend({
 			_('Allocate IP addresses sequentially, starting from the lowest available address.'));
 		o.optional = true;
 
+		o = s.taboption('general', form.Flag, 'address_as_local',
+			_('Resolve addresses locally'),
+			_('Never send queries for FQDNs in the Address option to an upstream resolver.'));
+		o.optional = true;
+
 		o = s.taboption('filteropts', form.Flag, 'boguspriv',
 			_('Filter private'),
 			customi18n(


efahl:
Tested on 24.10.3, LGTM for main and 24.10 backport...

```
$ cat /etc/config/dhcp
...
config dnsmasq
...
        option address_as_local '1'
```

systemcrash:
Can you add the commit message from the openwrt repo here also? Then it's more understandable why this setting should exist - its rationale if you will.

markozajc:
Sure,
```
`list address` entries in /etc/config/dhcp are sometimes (I'm not sure
about the exact conditions) passed to upstream resolver, bypassing local
resolution. Adding them (minus the IP) to --local prevents this. In the
configuration, this means that

    # /etc/config/dhcp
    list address '/hello.com/world.com/1.2.3.4'
    list address '/foo.com/bar.com/4.3.2.1'

which previously translated into

    # /var/etc/dnsmasq.conf.*
    address=/hello.com/world.com/1.2.3.4
    address=/foo.com/bar.com/4.3.2.1

now becomes

    # /var/etc/dnsmasq.conf.*
    address=/hello.com/world.com/1.2.3.4
    local=/hello.com/world.com/
    address=/foo.com/bar.com/4.3.2.1
    local=/foo.com/bar.com/

This behaviour is controlled by the `address_as_local` boolean option, which
defaults to false (old behaviour). https://github.com/openwrt/luci/pull/7957 adds support for this flag
to LuCI.

A workaround for a small list of domains is to add them to `option local`,
but this is very tedious to do for every `list address` entry and dnsmasq
limits this option to 1024 characters.
```

systemcrash:
… to the commit

markozajc:
Ah, right.

MkQtS:
@systemcrash this can be merged now

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -364,6 +364,11 @@ return view.extend({
 		o.optional = true;
 		o.datatype = 'uinteger';
 		o.placeholder = 150;
+
+		o = s.taboption('general', form.Flag, 'address_as_local',
+			_('Resolve addresses locally'),
+			_('Never send queries for FQDNs in the Address option to an upstream resolver.'));
+		o.optional = true;
```
systemcrash, line 371: I think you can drop optional

markozajc: Are you sure? I noticed that it's set for other `Flag`s, for example:
https://github.com/openwrt/luci/blob/6fdd6e3d14fe60f5509f629cc5ce6793fc9e5b50/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js#L356-L359

It looks like `Flag` options without `optional` are always written to the config, whereas ones with `optional` are only written if the value differs from the default. To me that seems like the better option, but I can change it if that's really the desired behaviour:
https://github.com/openwrt/luci/blob/6fdd6e3d14fe60f5509f629cc5ce6793fc9e5b50/modules/luci-base/htdocs/luci-static/resources/form.js#L4418-L4421

systemcrash: I figure that depends on the default behaviour you set in the dnsmasq init script. I guess optional can stay, then.

systemcrash:
Merged. Thanks for your patience @markozajc !

PR #7971, opened 2025-09-30
luci-app-usteer: fix menu showing on login screen

castiel652:
Currently menu entry is always shown even at the login screen.
Fix this by setting acl dependency.

Aslo fix the indent.

- [x] Tested on: mediatek/filogic, SNAPSHOT, Firefox :white_check_mark:
- [x] \( Preferred ) Mention: @Ramon00  feedback

### DIFF
--- a/applications/luci-app-usteer/root/usr/share/luci/menu.d/luci-app-usteer.json
+++ b/applications/luci-app-usteer/root/usr/share/luci/menu.d/luci-app-usteer.json
@@ -1,10 +1,13 @@
 {
-  "admin/network/usteer": {
-    "title": "Usteer",
-    "order": 80,
-    "action": {
-      "type": "view",
-      "path": "usteer/usteer"
-    }
-  }
+	"admin/network/usteer": {
+		"title": "Usteer",
+		"order": 80,
+		"action": {
+			"type": "view",
+			"path": "usteer/usteer"
+		},
+		"depends": {
+			"acl": [ "luci-app-usteer" ]
+		}
+	}
 }


systemcrash:
Please fix your commit message:)

Aslo fix the indent. -> Also fix the indent.

Ramon00:
What do you mean shown at log in screen? 
I actually don't understand why this depend is needed. Pretty much says it depends on itself?

castiel652:
> What do you mean shown at log in screen? I actually don't understand why this depend is needed. Pretty much says it depends on itself?

image

systemcrash:
Display is typically predicated on the acls being fulfilled. While the app has acls in JSON, it's absent from the menu json.

Ramon00:
To be honest this is a bug in that ui. Default UI shows nothing. Adding this depend feels like a workaround for a bad login page

systemcrash:
Not really, this is how acls are supposed to work. Most of them require a user to be logged in to access more privileged resources. But it's not unthinkable that some functions or displays be available on the login page, e.g. captive portals or some weather app :)

Ramon00:
> Not really, this is how acls are supposed to work. Most of them require a user to be logged in to access more privileged resources. But it's not unthinkable that some functions or displays be available on the login page, e.g. captive portals or some weather app :)

So if you are saying that you can click on the usteer menu item at the login page and have access then I agree, if you click on the link and it goes nowhere then well bad design login page. 

Or is it the intention that there are luci apps that work without a login altogether?

systemcrash:
> So if you are saying that you can click on the usteer menu item at the login page and have access then I agree, if you click on the link and it goes nowhere then well bad design login page.

The app has an ACL JSON - so evidently privileged resources.

> Or is it the intention that there are luci apps that work without a login altogether?

That's the idea.

Ramon00:
> That's the idea

But then we need to patch main UI login page.

castiel652:
@Ramon00 @systemcrash what's the conclusion of this?

Ramon00:
I'm fine to merge this, but this menu at login needs work if somebody really wants apps to work without login. If this functionality is not needed it would be simpler to just get rid of the menu and acl for the menu.

PR #7975, opened 2025-10-04
luci-base: form.js depends(): make value optional

stokito:
The `depends()` has two overloaded versions:
* `depends(field, value)` i.e. with two args
* `depends({'field1': 'val1', 'field2': 'val2'})` i.e. with one `Object` arg
The JsDoc doesn't allow to describe overloading but we can show that the second arg is optional. Then this calls of `depends({})` won't be reported by IDE as wrong.

Additionally I added some fixes typos and missing articles because IDE reports a lot about them and this distracts from code.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -189,7 +189,7 @@ const CBIJSONConfig = baseclass.extend({
  * @hideconstructor
  * @classdesc
  *
- * The `AbstractElement` class serves as abstract base for the different form
+ * The `AbstractElement` class serves as an abstract base for the different form
  * elements implemented by `LuCI.form`. It provides the common logic for
  * loading and rendering values, for nesting elements and for defining common
  * properties.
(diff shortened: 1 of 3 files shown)


systemcrash:
Please separate these two commits. Deploy just the depends one first. Those other 'typos' require more attention, since some are not typos at all.

stokito:
I used IntelliJ IDE and this typo, and this fix was suggested by its grammar check plugin:
1. Missing articles a/the
2. "amount of arguments" should be "number of arguments" because it's countable
3. Some missing comas
4. "satis**i**fied" changed to "satisfied" in variables - the only code change.
5. "The configuration form this section is added to." was changed to "The configuration form **to which** this section is added to." because the grammar checked asked, "Did you mean 'from'?", so I added more clarification.

So the changes aren't really that big, and except for two lasts, they all were changed automatically, so they should be safe. My main goal was to just reduce the number of warnings in the IDE.
These are all cosmetic changes

systemcrash:
Drop the 'grammar' changes.

systemcrash:
Closed by 6c5f78ffd2969b3a81807a126ab66be379a96aee.

systemcrash:
> So the changes aren't really that big and except of two last this all was changed automatically, so they should be safe. My main goal was to just reduce amount ow warning in the IDE. This is all cosmetic changes

Point 2. A word: there can be ambiguity when using 'number' since it can infer ordinal or cardinal properties, unless it's explicit, like "number of seconds since x". Number of arguments might imply "the positional number" so amount is used to mean 'quantity'. A total. 

Point 5. remove the redundant final "to" preposition which is already baked into the "to which".

systemcrash:
Anyway, thanks for the fixes :)

PR #7976, opened 2025-10-04
[WIP,RFC] get odhcpd leases via ubus

Alphix:
This is just a WIP/RFC so far, but right now `rpcd-mod-luci` digs around in the lease files created by `dnsmasq` and `odhcpd` (and it does so twice for each time the leases are updated in the web view).

So, I've wanted for a long time to get `rpcd-mod-luci` to communicate with `odhcpd` via `ubus` (no plans to do something similar for `dnsmasq`) to get current leases.

This PR is my first rough stab at it, and I'd appreciate some feedback before I spend any more time on it. Does it make sense?


Alphix:
@systemcrash and @jow- : I guess the two of you might be interested/most qualified to comment on this?

Alphix:
Hm, I must've been quite tired yesterday, I think it makes more sense to let luci talk straight to `odhcpd` 😆

Issue #7985, opened 2025-10-07
Feature request: Make app-package-manager optional


9Lukas5:
### What would you like to see in luci?

I found this in the menuconfig today:

```
 Symbol: PACKAGE_luci-app-package-manager [=y]                                                                                                                                                              │  
  │ Type  : tristate                                                                                                                                                                                           │  
  │ Defined at tmp/.config-package.in:63424                                                                                                                                                                    │  
  │   Prompt: luci-app-package-manager.................. Package management application                                                                                                                        │  
  │   Location:                                                                                                                                                                                                │  
  │     -> LuCI                                                                                                                                                                                                │  
  │       -> 3. Applications                                                                                                                                                                                   │  
  │         -> luci-app-package-manager.................. Package management application (PACKAGE_luci-app-package-manager [=y])                                                                               │  
  │ Selects: PACKAGE_libc [=y] && PACKAGE_luci-base [=y]                                                                                                                                                       │  
  │ Selected by [y]:                                                                                                                                                                                           │  
  │   - PACKAGE_luci-ssl-openssl [=y] && y                                                                                                                                                                     │  
  │ Selected by [n]:                                                                                                                                                                                           │  
  │   - PACKAGE_luci [=n] && y                                                                                                                                                                                 │  
  │   - PACKAGE_luci-nginx [=n] && y                                                                                                                                                                           │  
  │   - PACKAGE_luci-ssl [=n] && y  
```

And when I diffed against my old config, I found that I was able to disable the package manager in the past (to reduce my image size), but now it's a dependency on the base package.

Can the package manager be made optional again?

systemcrash:
Nothing has changed there for a year. All luci apps depend on luci-base. What's the problem?

jow-:
The empty meta package `luci-ssl-openssl` selects it. Deselect this meta package and you can pick LuCI components individually.

9Lukas5:
Well, it's actually selected by all the meta packages. (` luci`, `luci-nginx`, `luci-ssl` and `luci-ssl-openssl`)
I'm not sure when this change happened, I have a config from spring when I last built and updated everything now in autumn again.

@systemcrash 
Sometime in-between, the change happened, that when I select the the LuCI collection with encryption (no matter if choosing mbedTLS or OpenSSL) the package manager is a hard dependency and cannot be deselected.  
_That_ is the problem I'm referring to.

@jow- I'm not sure what you are suggesting I should do 🧐   
[...]

PR #7995, opened 2025-10-08
collections: Remove package-manager dependency for luci-ssl* meta packages

9Lukas5:
- [x] Tested on: local build machine menuconfig selection works as intended by change :white_check_mark:
- [x] \( Optional ) Closes: openwrt/luci#7985
- [x] Description:
  The package manager app has nothing to do with the traffic being encrypted or not. Removes the dependency, as it's already brought in if normal `luci` is selected instead of the `luci-light` anyways.

### DIFF
--- a/collections/luci-ssl/Makefile
+++ b/collections/luci-ssl/Makefile
@@ -13,7 +13,6 @@ LUCI_TITLE:=LuCI with HTTPS support (mbedtls as SSL backend)
 LUCI_DEPENDS:=+luci-light \
 	+libustream-mbedtls \
 	+px5g-mbedtls \
-	+luci-app-package-manager
 
 PKG_LICENSE:=Apache-2.0
 


hnyman:
Well, package manager was first removed in Q1/2023 by #6211
and then added back in Q4/2024 by  #7416
Hard to imagine that it would be removed again.

Package manager can be easily disabled in private builds like you were told in #7985

W.g. when using OpenSSL, you avoid selecting (the empty virtual package) luci-ssl-openssl, and you select the three individual components: luci-light, libustream-openssl and openssl-util 

Like seen in the structure:
```
LUCI_DEPENDS:=+luci-light \
	+libustream-openssl \
	+openssl-util \
	+luci-app-package-manager
```

hnyman:
I close this as @jow- explained the reasoning in #7985 where the discussion about your proposal has been

9Lukas5:
Uhm, I actually thought the last state of #7985 was to see if there are any reasons for not doing this PR and as none were raised I would have expected this to get merged :thinking: 

As I just realized I left the trailing `\` just as in #7996, I removed them here as well, in case you re-open this PR.

PR #7996, opened 2025-10-08
collections/luci-nginx: reorganize dependencies into sorted single lines

9Lukas5:
- [x] Tested on: local build machine menuconfig selection works as intended by change :white_check_mark:
- [x] \( Optional ) Refers to discussion in: openwrt/luci#7985
- [x] Description:
  Just splits the dependencies in single lines and sorts them alphabetically as all the other meta packages have it already.

### DIFF
--- a/collections/luci-nginx/Makefile
+++ b/collections/luci-nginx/Makefile
@@ -12,9 +12,15 @@ LUCI_BASENAME:=nginx
 LUCI_TITLE:=LuCI interface with Nginx as Webserver
 LUCI_DESCRIPTION:=Standard OpenWrt set including full admin with ppp support and the default Bootstrap theme
 LUCI_DEPENDS:= \
-	+nginx +nginx-mod-luci +luci-mod-admin-full +luci-theme-bootstrap \
-	+luci-app-firewall +luci-app-package-manager +luci-proto-ppp +IPV6:luci-proto-ipv6 \
-	+rpcd-mod-rrdns
+	+IPV6:luci-proto-ipv6 \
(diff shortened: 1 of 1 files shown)


collections/luci-nginx/Makefile
```diff
@@ -12,9 +12,15 @@ LUCI_BASENAME:=nginx
...
-	+rpcd-mod-rrdns
+	+IPV6:luci-proto-ipv6 \
+	+luci-app-firewall \
+	+luci-app-package-manager \
+	+luci-mod-admin-full \
+	+luci-proto-ppp \
+	+luci-theme-bootstrap \
+	+nginx \
+	+nginx-mod-luci \
+	+rpcd-mod-rrdns \
```
systemcrash: remove trailing `\`

9Lukas5: I actually added it on purpose, compared to trailing commas in some data formats in case something changes in the future on the end if the list and it seemed to not cause any issues.

Shall I still remove it?

hnyman: Leaving it there is a style that enables adding new lines without modifying the existing lines. Some people like it. But it can lead to mistakes, as the next line needs to be empty. If somebody edits the file without recognizing that, and uses the empty line for somthing, there will be mistakes.

So, I prefer the style without the trailing backslash at the last line, just like systemcrash

9Lukas5: Ok, will change it on the weekend and ping you again 👍

PR #7998, opened 2025-10-10
luci-mod-network: add global DUID cfg option

Alphix:
- [x] Tested on: (NanoPi R6S, OpenWrt master, Chrome) :white_check_mark:
- [x] \( Optional ) Depends on: https://github.com/openwrt/openwrt/pull/20359

This exposes the default DHCP DUID config option in the LuCI interface.

For reference, see:
https://github.com/openwrt/openwrt/pull/20359
https://github.com/openwrt/odhcpd/pull/274

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -1659,6 +1659,11 @@ return view.extend({
 			_('This prefix is randomly generated at first install.'));
 		o.datatype = 'cidr6';
 
+		o = s.option(form.Value, 'dhcp_default_duid', _('Default DUID'),
+			_('The default <abbr title="DHCP Unique Identifier">DUID</abbr> for this device, used when acting as a DHCP server or client. The client identifier can be overridden on a per-interface basis.') + '<br />' +
+			_('This identifier is randomly generated the first time the device is booted.'));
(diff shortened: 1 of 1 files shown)


systemcrash:
Question: is it really a good idea to expose this setting for users? The real question being is changing this likely to cause interruption?

Alphix:
> Question: is it really a good idea to expose this setting for users? The real question being is changing this likely to cause interruption?

I think that's a tough choice. If we implement a static per-device DUID (which we should), then the DUID that is used by the DHCP clients will change when users upgrade to the next OpenWrt release (since per-interface DUID-LLs are in use right now). My thinking was that users might want to be able to change the DUID in such a case.

At a minimum, I think the DUID should be provided somewhere in the UI (and right next to where the IPv6 ULA is exposed seemed like the logical place to do so), but it could be read-only (to allow users to update the cfg in whatever device is handing out leases).

Alphix:
> > Question: is it really a good idea to expose this setting for users? The real question being is changing this likely to cause interruption?

And I should add that the DUID *is* exposed right now in LuCI, but as a per-device setting (it's under "Advanced Settings", the option is titled "Client ID to send when requesting DHCP" if you go to interfaces and edit an interface which is a DHCPv4/DHCPv6 client)

Alphix:
Now that [PR#20359](https://github.com/openwrt/openwrt/pull/20359) is in, I think this is ready to go as well...

systemcrash:
Thanks @Alphix !

PR #8005, opened 2025-10-10
luci-base: add odhcpd feature detection

Alphix:
For dnsmasq, feature detection is currently supported like this:

	L.hasSystemFeature('dnsmasq', 'dhcpv6')

while for odhcpd, only a basic check is supported:

	L.hasSystemFeature('odhcpd')

With this patch, a similar feature check is also possible for odhcpd, e.g.:

	L.hasSystemFeature('odhcpd', 'dhcpv6')

- [x] Tested on: (NanoPi R6S, OpenWrt master, Chrome) :white_check_mark:

### DIFF
--- a/modules/luci-base/root/usr/share/rpcd/ucode/luci
+++ b/modules/luci-base/root/usr/share/rpcd/ucode/luci
@@ -229,7 +229,6 @@ const methods = {
 				offloading: access('/sys/module/xt_FLOWOFFLOAD/refcnt') == true || access('/sys/module/nft_flow_offload/refcnt') == true,
 				br2684ctl:  access('/usr/sbin/br2684ctl') == true,
 				swconfig:   access('/sbin/swconfig') == true,
-				odhcpd:     access('/usr/sbin/odhcpd') == true,
 				zram:       access('/sys/class/zram-control') == true,
 				sysntpd:    readlink('/usr/sbin/ntpd') != null,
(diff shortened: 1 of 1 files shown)


Alphix:
@systemcrash ping :)

systemcrash:
Does `odhcpd` presence still evaluate as truthy even if the object `{}` is empty (some older version of odhcpd)?

Alphix:
> Does `odhcpd` presence still evaluate as truthy even if the object `{}` is empty (some older version of odhcpd)?

No, it doesn't. I assume that you want me to make sure it does :) I'll update the patch...

Alphix:
Done

Alphix:
> Does `odhcpd` presence still evaluate as truthy even if the object `{}` is empty (some older version of odhcpd)?

I mean, the presence would be truthy if `{}` was empty, but with the previous version of the patch, there would have been no `{}` at all...so odhcpd presence would have been false for older versions of odhcpd. Fixed now

systemcrash:
This feature is in main already, right?

Alphix:
> This feature is in main already, right?

Yes, since https://github.com/openwrt/openwrt/commit/5b1cb76a015449de77094185f03c7de02dae20a0

systemcrash:
Good. Thanks @Alphix

Alphix:
> Good. Thanks @Alphix

Thank you :)

PR #8006, opened 2025-10-11
build: fix Traditional Chinese name

AndyChiang888:
fix Traditional Chinese name

- [X] Tested on: (x86/64, SNAPSHOT, Chrome 141.0.7390.77) :white_check_mark:

### DIFF
--- a/luci.mk
+++ b/luci.mk
@@ -56,8 +56,8 @@ LUCI_LANG.tr=Türkçe (Turkish)
 LUCI_LANG.uk=Українська (Ukrainian)
 LUCI_LANG.vi=Tiếng Việt (Vietnamese)
 LUCI_LANG.yua=Yucateco (Yucatec Maya)
-LUCI_LANG.zh_Hans=简体中文 (Chinese Simplified)
-LUCI_LANG.zh_Hant=繁體中文 (Chinese Traditional)
+LUCI_LANG.zh_Hans=简体中文 (Simplified Chinese)
+LUCI_LANG.zh_Hant=正體中文 (Traditional Chinese)
 #LUCI_LANG_END
 
 # Submenu titles


hnyman:
To my understanding, the aliases are a compatibility setting due to language code change a few years ago.
What is your problem with the aliases?   (as the main language code is already the new one)

* https://github.com/openwrt/luci/issues/3477
* https://github.com/openwrt/luci/commit/48749ca873a5ba385dc372a6a57738024cc2acfb

AndyChiang888:
Chinese Simplified is not only used in mainland China, also in Singapore and Malaysia. Chinese Traditional is not only used in Taiwan, also in Hong Kong and Macau, it should be use with a script subtag instead of a region subtag.
zh-Hant and zh-Hans are valid and supported by browser, not invalid zh_Hant and zh_Hans.
See:
https://schneegans.de/lv/?tags=zh-Hans&Format=text
https://schneegans.de/lv/?tags=zh-Hant&Format=text

hnyman:
> invalid zh_Hant and zh_Hans.

Well, the "valid" and "invalid" changes according to the context and tool used. ICU defines them with underscore, and weblate, the web translation platform that we use, uses that notation. So, we are using the underscore in any case.

https://www.localeplanet.com/icu/
<html><body>
<h1>ICU Locale “Simplified Chinese” (zh_Hans)</h1>
<h2>General Information</h2>

Property | Value
-- | --
Base locale ID | zh_Hans


</body>
</html>


Also zh-cn and zh-tw are valid, so this is more about the implied context about the alias, not about the technical correctness.

Where does that alias actually show up? Elsewhere than in the name of the downloaded language pack?

AndyChiang888:
In addition to the language pack, only /etc/config/luci display. I will update the PR to only fix the Chinese Traditional name.

PR #8010, opened 2025-10-13
luci-mod-status: fix duplicate routing in menu

AndyChiang888:
fixes: 89ae891462ea9928fc2b01e25c1611b1b71352d5
fixes: ad6591fb4b33a09a9fc16982f1fad941527c1c55

Two Routing now.....

- [X] Tested on: (x86/64, openwrt r31419-297057f0f2, Chrome) :white_check_mark:


systemcrash:
Thanks for spotting this @AndyChiang888. The problem with jumping over exclusively to JSON here is that it depends on the IP commands, either tiny or full, and they have to be mandated as a PKG_BUILD_DEPENDS. I'm trying not to do that, and instead giving the option to use JSON if either of those is installed, as it sometimes is. 

I fixed it via 40232e89ff4121bdfff71d97866a8f6563274cf0 instead.

PR #8012, opened 2025-10-15
luci-app-watchcat: allow multiple IP addresses in ping hosts field

ivanddiaz:
The watchcat service backend already supports monitoring multiple IP addresses, but the LuCI web interface was blocking this functionality through its UI validation.

This PR fixes the input validation to accept space-separated IP addresses/hostnames instead of restricting users to a single entry.

**Changes**
- Changed datatype from 'host' to 'list(host)' in the ping hosts field
- Updated default value to show multiple IPs example: '8.8.8.8 1.1.1.1'

**Testing**
Users can now enter multiple monitoring targets like "8.8.8.8 1.1.1.1 google.com" in the Host To Check field. Each host in the list is validated individually.
This aligns the web interface capabilities with what the watchcat service actually supports.

imagen

Depends on: https://github.com/openwrt/packages/pull/27756

### DIFF
--- a/applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js
+++ b/applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js
@@ -56,16 +56,6 @@ return view.extend({
 		o.depends({ mode: "restart_iface" });
 		o.depends({ mode: "run_script" });
 
-		o.load = function(section_id) {
-			return (String(this.map.data.get('watchcat', section_id, 'pinghosts') || '')
-			.trim().split(/\s+/).filter(Boolean))
-		}
-
-		o.write = function(section_id, formvalue) {
-			this.map.data.set('watchcat', section_id, 'pinghosts',
-			(formvalue || []).map(v => String(v).trim()).filter(Boolean).join(' '))
-		}
-
 		o = s.taboption('general', form.ListValue, 'addressfamily',
 				_('Address family for pinging the host'));
 		o.default = 'any';


systemcrash:
This value might serve better as a dynamiclist. Multiple values are anyway concatenated into a single parameter lower down in the shell scripts. 

Then you don't need to change the display string at all.

ivanddiaz:
Hello @systemcrash,

Thank you for the feedback.
I updated the [code](https://github.com/openwrt/luci/pull/8012/commits/c20245c2796f22a5908739540446c8e96d7bb0be) to use [DynamicList](https://openwrt.github.io/luci/jsapi/LuCI.ui.DynamicList.html) as suggested.
The label and backend logic remain the same; only the user interface is improved.

applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js
```diff
@@ -49,13 +49,23 @@ return view.extend({
...
+		o.load = function(section_id) {
+			return (String(this.map.data.get('watchcat', section_id, 'pinghosts') || '')
+			.trim().split(/\s+/).filter(Boolean))
+		}
+
+		o.write = function(section_id, formvalue) {
+			this.map.data.set('watchcat', section_id, 'pinghosts',
+			(formvalue || []).map(v => String(v).trim()).filter(Boolean).join(' '))
+		}
+
```
systemcrash: test your change without this and see how the watchcat init parses multiple values.

ivanddiaz: Hi @systemcrash,

I’ve updated the frontend following your suggestion to use a **dynamiclist**, and it works great — the UI now accepts multiple hosts without any problem.  

However, while testing this, I found a new issue in the backend script (`watchcat.sh`).  

Right now, OpenWrt runs this command without quotes:  

`/bin/sh /usr/bin/watchcat.sh run_script 60 1.1.1.1 8.8.8.8 3 standard eth0 ipv4 /etc/watchcat.user.sh`


This breaks with the following error:  

    time_now='7094.18 7015.09'
    time_now=7094
    time_diff=0
    '[' 0 -lt 8.8.8.8 ]
    sh: 8.8.8.8: bad number
The problem is that both IPs are passed as separate positional arguments, so the script later treats `8.8.8.8` as a number.  

It should be passed as **one quoted argument**, like this:  

`/bin/sh /usr/bin/watchcat.sh run_script 60 "1.1.1.1 8.8.8.8" 3 standard eth0 ipv4 /etc/watchcat.user.sh`


Internally, `watchcat.sh` expects something like:  
```sh
pinghosts="1.1.1.1 8.8.8.8"
```
So, even if the frontend is now correct, the backend still needs a fix to handle this case properly.
I’ll take a couple of days to review and test a safe solution before continuing.

Thanks!

ivanddiaz: @systemcrash hello again,

I’ve found the root cause in /etc/init.d/watchcat.
I’ll open a PR tomorrow in packages/utils/watchcat with the fix and will let you know once it’s ready.

Thank you!

systemcrash: ping

ivanddiaz: I’ve found that the init script read the unlock mmifacename as NULL when “Name of ModemManager Interface" is not set. This causes watchcat.sh to break, because all arguments shift to the left. I’m working on fixing it.

ivanddiaz: Hello @systemcrash 

Backend PR is here:
https://github.com/openwrt/packages/pull/27756

I'll wait for feedback con continue with this.

Thanks for your guidance and patience :).

systemcrash:
@ivanddiaz OK - https://github.com/openwrt/packages/pull/27756 is merged. If you're happy, set this to ready.

ivanddiaz:
> @ivanddiaz OK - [openwrt/packages#27756](https://github.com/openwrt/packages/pull/27756) is merged. If you're happy, set this to ready.

Thank you @systemcrash, marked as ready

systemcrash:
Merged. Thanks @ivanddiaz !

PR #7931, opened 2025-09-02
luci-mod-status,luci-mod-network: support oui to vendor resolving

ckorber:
For easier definition of connected devices, this commit adds support to identify them by vendor name.

Package `upf-neigh` is needed to lookup the vendor name in a hash table. It implements a ubus-call `ubus call fingerprint fingerprint`:

```
root@<redacted> ~ # ubus call fingerprint fingerprint
{
        "7c:c2:55:XX:XX:XX": {
                "vendor": "Super Micro Computer, Inc."
        }
}
```
image

image

- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [x] Tested on: (x86/64, openwrt 24.10, firefox-esr) :white_check_mark:
- [ ] \( Preferred ) Mention: @ the original code author for feedback
- [x] \( Preferred ) Screenshot or mp4 of changes:
- [x] \( Optional ) Closes: e.g. openwrt/luci#2065
- [x] Description: (describe the changes proposed in this PR)

PR #8013, opened 2025-10-15
luci-mod-status,-network: remove checkUfpInstalled

Alphix:
Right now, the checkUfpInstalled RPC call is performed first, and only after that can the other Promises be called. Simplify this by adding a single check for /usr/sbin/ufpd in the rpcd ucode, and replace calls to checkUfpInstalled() with simple L.hasSystemFeature('ufpd') check (which is only done once, and cached).

This also allows the 'stat' permission to be removed from a couple of acls.

- [x] Tested on: (NanoPi R6S, master, Chrome) :white_check_mark:

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/routes.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/routes.js
@@ -114,7 +114,7 @@ return view.extend({
 		return res;
 	},
 
-	parseRoutes(routes, macs, networks, v6) {
+	parseRoutes(routes, networks, v6) {
 		const res = [];
 
 		for (const line of routes.trim().split(/\n/)) {
@@ -226,7 +226,7 @@ return view.extend({
 		cbi_update_table(neigh4tbl, this.parseNeighbs(ip4neigh, macdata, networks, false),
 			E('em', _('No entries available'))
 		);
-		cbi_update_table(route4tbl, this.parseRoutes(ip4route, macdata, networks, false),
+		cbi_update_table(route4tbl, this.parseRoutes(ip4route, networks, false),
 			E('em', _('No entries available'))
 		);
 		cbi_update_table(rule4tbl, this.parseRules(ip4rule),
@@ -235,7 +235,7 @@ return view.extend({
 		cbi_update_table(neigh6tbl, this.parseNeighbs(ip6neigh, macdata, networks, true),
 			E('em', _('No entries available'))
 		);
-		cbi_update_table(route6tbl, this.parseRoutes(ip6route, macdata, networks, true),
+		cbi_update_table(route6tbl, this.parseRoutes(ip6route, networks, true),
 			E('em', _('No entries available'))
 		);
 		cbi_update_table(rule6tbl, this.parseRules(ip6rule),
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/routesj.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/routesj.js
@@ -125,7 +125,7 @@ return view.extend({
 		return res;
 	},
 
-	parseRoutes(routes, macs, networks, v6) {
+	parseRoutes(routes, networks, v6) {
 		const res = [];
 
 		for (const rt of this.parseJSON(routes)) {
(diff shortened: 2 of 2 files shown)


Alphix:
@systemcrash ping

modules/luci-mod-status/htdocs/luci-static/resources/view/status/routes.js
```diff
@@ -124,7 +114,7 @@ return view.extend({
 		return res;
 	},
 
-	parseRoutes(routes, macs, networks, v6) {
+	parseRoutes(routes, networks, v6) {
```
feckert, line 117: Does this change belongs to this commit?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/routesj.js
```diff
@@ -135,7 +125,7 @@ return view.extend({
 		return res;
 	},
 
-	parseRoutes(routes, macs, networks, v6) {
+	parseRoutes(routes, networks, v6) {
```
feckert, line 128: Does this change belongs to this commit?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/routes.js
Alphix: Yeah, maybe I should have mentioned it in the commit msg, the commits introducing the `checkUfpInstalled()` RPC calls introduced an unused variable to `parseRoutes()`...I could split that to a separate commit to make it clearer

modules/luci-mod-status/htdocs/luci-static/resources/view/status/routesj.js
Alphix: Yeah, maybe I should have mentioned it in the commit msg, the commits introducing the `checkUfpInstalled()` RPC calls introduced an unused variable to `parseRoutes()`...I could split that to a separate commit to make it clearer

systemcrash:
This seems reasonable - and your change was actually what I had in mind when @ckorber was doing #7931. Although what we have now is more simple for backporting, whereby a user doesn't also need to update luci-base to get the benefits in luci-mod-network. 

Alas, this way is probably better performance-wise in the long run.

Alphix:
> This seems reasonable - and your change was actually what I had in mind when @ckorber was doing https://github.com/openwrt/luci/pull/7931.

Yeah, I saw the comment in https://github.com/openwrt/luci/pull/7931#issuecomment-3249984709

> Although what we have now is more simple for backporting...

Hm, backporting...is that something we support? I'm kind of working on a separate PR to split `/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js` into two separate views for DNS and DHCP (dhcp.js has grown kind of big, and is already quite difficult to work with). Is that something I should forget about?

systemcrash:
> Hm, backporting...is that something we support? I'm kind of working on a separate PR to split `/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js` into two separate views for DNS and DHCP (dhcp.js has grown kind of big, and is already quite difficult to work with). Is that something I should forget about?

If it's a working useful feature in main and it's just a drop-in cherry-pick, why not?

I agree that the dhcp thing has grown. But they all dig into the dhcp config file, so it's reasonable to have them there, but not unwarranted if a new entry clears things up.

Alphix:
> > Hm, backporting...is that something we support? I'm kind of working on a separate PR to split `/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js` into two separate views for DNS and DHCP (dhcp.js has grown kind of big, and is already quite difficult to work with). Is that something I should forget about?
> 
> If it's a working useful feature in main and it's just a drop-in cherry-pick, why not?

It was more a question...if backport-ability is important, it would rather limit how much can be changed/refactored :) This particular feature (MAC-vendor support) won't be a simple cherry-pick though since there's been several little fixups after it was committed. Anyway...the commit is there, so someone who wants to cherry-pick it can still do it since this would be a subsequent commit?

> I agree that the dhcp thing has grown. But they all dig into the dhcp config file, so it's reasonable to have them there, but not unwarranted if a new entry clears things up.

I'll submit a separate PR for that, but it's quite....big :)

Alphix:
Pushed an updated PR where the changes to `parseRoutes()` is a separate commit to make it clearer that the seemingly unrelated change is also related to this (addresses @feckert's comments).

modules/luci-mod-status/htdocs/luci-static/resources/view/status/routesj.js
feckert, line 1: This file was added 3 days ago with commit https://github.com/openwrt/luci/commit/89ae891462ea9928fc2b01e25c1611b1b71352d5

If we want to backport the whole ufp feature and this PR to the openwrt-24.10 stable branch. We should split this into different commits to make cherry pick easy.

systemcrash: Agreed. We can skip the backport for this particular feature.

Alphix: Does that mean the PR is good to go?

systemcrash:
Merged. Thanks @Alphix !

PR #8016, opened 2025-10-16
luci-mod-network: separate dhcp and dns views

Alphix:
This started when I build a custom image including `odhcpd` with DHCPv4 support and without `dnsmasq` and I was quite surprised to see the whole `DHCP and DNS` tab gone from LuCI, even though it's necessary to configure static leases, also in the absence of `dnsmasq`.

First I tried adding a bunch of `if (dnsmasq || odhcpd)` and `if (dnsmasq)` conditionals to each individual option/tab, but I quickly realized that it was very cumbersome, leading to a lot of extra code and made `dhcp.js` even larger and hairier than before.

In addition, the various per-tab options in `dhcp.js` are currently defined in a rather random order (i.e. not in the order which they actually appear in the UI).

Thus I settled on this approach. Just because there's a whole lot of cfg options in `/etc/config/dhcp` related to both DNS and DHCP, there's no real reason that they have to be presented in that manner in the LuCI UI.

At the same time, I also sorted all the options so that they are listed in the actual order that they appear in the UI.

The end result is two files:
- `modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js`
- `modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js`

Each weighing in at around 850 lines, and which are much more readable (and amenable to further devel) than the previous `dhcp.js` heavyweight at 1500+ lines.

Note that some options (like the interfaces `dnsmasq` should/shouldn't listen on, or the log facility used by `dnsmasq`) are currently duplicated across the views. We might want to consider adding a short help text explaining that, but I intentionally refrained from changing the actual code as much as possible to make this PR easier to review.

- [x] Tested on: (NanoPi R6S, master, Chrome)

### DIFF
--- a/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
+++ b/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
@@ -38,8 +38,7 @@
 		"description": "Grant access to DNS configuration",
 		"read": {
 			"ubus": {
-				"luci-rpc": [ "getDHCPLeases", "getDUIDHints", "getHostHints" ],
-				"fingerprint": [ "fingerprint" ]
+				"luci-rpc": [ "getHostHints" ]
 			},
 			"uci": [ "dhcp" ]
 		},


Alphix:
Some screenshots illustrating the end result:
Screenshot From 2025-10-15 20-10-15
Screenshot From 2025-10-15 20-10-43
Screenshot From 2025-10-15 20-11-09

Alphix:
Pinging @systemcrash

systemcrash:
This is fine. 

Any thoughts from @hnyman ?

systemcrash:
Merged. Thanks @Alphix

PR #8017, opened 2025-10-17
luci-base, luci-app-firewall: enable range of MAC addresses

ckorber:
In order to handle a range of MAC addresses it seems to be useful being able to set it in luci.
Therefore these commits use the `src_mac` field to specify ranges like can be seen in the pictures.
Also negation can be used to exclude a certain range for being handled.

This PR depends on openwrt/firewall4#74

- [x] Tested on: (x86/64, openwrt-24.10, firefox) :white_check_mark:
- [x] Screenshot or mp4 of changes:

If a single MAC address is specified, it is validated by `macaddr` if a range is given it is validated by `macrange`
Screenshot From 2025-10-17 11-14-10
invalid MAC address
image
valid MAC address
Screenshot From 2025-10-17 11-17-00
invalid range
image
valid range
image
also negation is possible
image
upper address is lower than lower macaddr -> error

- [x] Depends on: e.g. openwrt/firewall4#74 in sister repo

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/validation.js
+++ b/modules/luci-base/htdocs/luci-static/resources/validation.js
@@ -770,6 +770,35 @@ const ValidatorFactory = baseclass.extend(/** @lends LuCI.validation.ValidatorFa
 				multicast ? _('valid multicast MAC address') : _('valid MAC address'));
 		},
 
+		macrange(multicast) {
+			const m = this.value.match(/^((?:[a-fA-F0-9]{2}:){5}[a-fA-F0-9]{2})-((?:[a-fA-F0-9]{2}:){5}[a-fA-F0-9]{2})$/);
+			let valid, mac1, mac2;
+
+			let tonumber = (mac) => {
+				const n = mac.replace(/[^a-fA-F0-9]/g, '').toUpperCase();
+				return BigInt('0x' + n);
+			};
+
+			if (m == null)
+				return this.assert(false, multicast ? _('valid multicast MAC address range') : _('valid MAC address range'));
+
+			if (m[1]) {
+				mac1 = tonumber(m[1]);
+				valid = this.apply('macaddr', m[1], [multicast]);
+				if (!valid)
+					return this.assert(false, multicast ? _('valid multicast MAC address') : _('valid MAC address'));
+			}
+
+			if (m[2]) {
+				mac2 = tonumber(m[2]);
+				valid = this.apply('macaddr', m[2], [multicast]);
+				if (!valid)
+					return this.assert(false, multicast ? _('valid multicast MAC address') : _('valid MAC address'));
+			}
+
+			return this.assert(mac1 < mac2, multicast ? _('valid multicast MAC address range') : _('valid MAC address range'));
+		},
+
 		/**
 		 * Assert a valid hostname or IP address.
 		 * @function LuCI.validation.ValidatorFactory.types#host
@@ -1003,7 +1032,7 @@ const ValidatorFactory = baseclass.extend(/** @lends LuCI.validation.ValidatorFa
 			if (v == '.' || v == '..')
 				return this.assert(false, _('valid network device name, not "." or ".."'));
 
-			return this.assert(v.match(/^[^:\/%\s]{1,15}$/), _('valid network device name between 1 and 15 characters not containing ":", "/", "%" or spaces'));
+			return this.assert(v.match(/^[^:/%\s]{1,15}$/), _('valid network device name between 1 and 15 characters not containing ":", "/", "%" or spaces'));
 		},
 
 		/**
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/tools/firewall.js
@@ -516,8 +516,8 @@ return baseclass.extend({
 		const o = s.taboption(tab, this.CBIDynamicMultiValueList, name, label, description);
 
 		o.modalonly = true;
-		o.datatype = 'list(macaddr)';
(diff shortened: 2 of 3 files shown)


openwrt-ai (commented):
Reviewed 2 new commits.

## Commit checks
- `d7d71590bc1` (luci-base: parse for mac range): message body has a typo — "range of MAC addresses int they syntax of" should read "in the syntax of" — and uses ``'''`` as a code fence (Markdown) where plain indentation is the git-log convention. Content otherwise matches the diff.
- `4a5a74ae217` (luci-app-firewall: enable to handle mac ranges): message matches the diff.

One inline question on the `macrange` validator (equal-endpoint handling, plus two minor cleanups). The cross-repo dependency on openwrt/firewall4#74 is already disclosed in the PR description and the "depends on PR in other repo" label, so the missing backend support for the `-` range syntax is expected and not flagged here.

Note: the failing "Lua Lint" check does not appear related to this PR, which touches only JavaScript files.

---

modules/luci-base/htdocs/luci-static/resources/validation.js
```diff
@@ -386,6 +386,35 @@ const ValidatorFactory = baseclass.extend({
...
+			}
+
+			if (m[2]) {
+				mac2 = tonumber(m[2]);
+				valid = this.apply('macaddr', m[2], [multicast]);
+				if (!valid)
+					return this.assert(false, multicast ? _('valid multicast MAC address') : _('valid MAC address'));
+			}
+
+			return this.assert(mac1 < mac2, multicast ? _('valid multicast MAC address range') : _('valid MAC address range'));
```
openwrt-ai, line 799: The final check uses strict `<`, so a degenerate range with identical endpoints (e.g. `00:11:22:33:44:55-00:11:22:33:44:55`) is rejected as invalid. Is that intended? nftables accepts an equal-endpoint range, and a user pasting the same MAC into both halves would get a confusing "valid MAC address range" error. If single-address ranges should be allowed, use `mac1 <= mac2`.

Two smaller notes on this validator:
- The regex requires both capture groups (no `?` quantifier), so once `m != null` both `m[1]` and `m[2]` are always present. The `if (m[1])` / `if (m[2])` guards can never be false and can be dropped.
- `tonumber` parses the addresses only to compare ordering, while `apply('macaddr', ...)` is what actually validates them. Consider computing `mac1`/`mac2` after both `macaddr` checks pass, so ordering is compared on values already known to be well-formed.

---

openwrt-ai (commented):
Follow-up to my previous review: the inline comment referenced there failed to attach, so it is added here. Also correcting that previous note — all CI checks on this head commit are passing; please disregard the "Lua Lint" remark.

---

modules/luci-base/htdocs/luci-static/resources/validation.js
```diff
@@ -770,12 +770,42 @@ const ValidatorFactory = baseclass.extend(/** @lends LuCI.validation.ValidatorFa
...
+		},
+
 		/**
 		 * Assert a valid hostname or IP address.
 		 * @function LuCI.validation.ValidatorFactory.types#host
 		 * @param {boolean} [ipv4only] enforce IPv4 IPs only.
 		 * @returns {@link LuCI.validation.Validator#assert assert()} {boolean}
 		 */
+
```
openwrt-ai: nit: this rebase introduced a stray blank line between the `host` JSDoc block and its function declaration, which doesn't match the rest of this file (every other validator's `*/` is immediately followed by its function). Remove it.

```suggestion
```

---

openwrt-ai (commented):
Reviewed 2 re-pushed commits.

## Commit checks
- `054d5be` "luci-base: parse for mac range" — the typo from the previous version ("int they syntax of") is now fixed. The message body still wraps the example in `'''` (Markdown fence) rather than plain indentation as is the git-log convention; minor, content matches the diff.
- `a38abd2` "luci-app-firewall: enable to handle mac ranges" — message matches the diff.

The `macrange` validator logic is unchanged from the previously reviewed version, so my earlier inline note on the strict `<` (rejects equal-endpoint ranges), the always-true `if (m[1])` / `if (m[2])` guards, and the ordering of `tonumber` vs the `macaddr` checks still stands (see the existing thread on line 799) and is not repeated here.

---

openwrt-ai (commented):
Reviewed 2 re-pushed commits; no new issues found.

The only change versus the previously reviewed head (`a38abd2`) is the removal of a stray blank line above `host()` in `validation.js`; the `macrange` validator logic and the `luci-app-firewall` changes are byte-identical to the prior version. My earlier inline notes on the `macrange` validator (strict `<` rejecting equal-endpoint ranges, the always-true `if (m[1])` / `if (m[2])` guards) therefore still stand and are not repeated here. The `'''` Markdown fence in `c60230e`'s commit body (vs. plain indentation per git-log convention) is also unchanged and was already noted previously.

---

openwrt-ai (commented):
Reviewed 2 re-pushed commits; no new issues found.

The only change to the PR's three files versus the previously reviewed head (`9ba0e35`) is a cosmetic regex cleanup in the `macrange` validator — the unnecessary backslash-escape of `-` outside a character class was dropped (`\-` → `-`), which is correct and a no-op for matching. The `macrange` logic and the `luci-app-firewall` changes are otherwise unchanged. My earlier inline notes on the `macrange` validator (strict `<` rejecting equal-endpoint ranges, the always-true `if (m[1])` / `if (m[2])` guards, and the `tonumber`-vs-`macaddr` ordering) therefore still stand on the existing unresolved thread at `validation.js:799` and are not repeated here. The `'''` Markdown fence in `0f856cf`'s commit body (vs. plain indentation per git-log convention) is also unchanged and was noted previously.

---

Issue #7310, opened 2024-10-08
convert outstanding LUA apps to JS


systemcrash:
Who is up for the task?

Key:

- [ ] **~~do not convert~~**
- [ ] 🗑️  _~~deprecate or remove~~_
- [ ] unassigned 
- [ ] not yet assigned with last known maintainer(s) tagged @octocat
- [x] assigned or accepted by @octocat 
- [x] ~~done~~ @octocat
- [x] ~~done~~ 

@stokito @dannil @Ramon00 ?

- [ ] luci-app-ahcp/luasrc @jow- 
- [ ] luci-app-clamav/luasrc  @ratkaj @lperkov 
- [ ] luci-app-coovachilli/luasrc @sbyx @jow- 
- [ ] luci-app-cshark/luasrc @lperkov 
- [ ] luci-app-dcwapd/luasrc @csonsino 
- [ ] luci-app-dnscrypt-proxy/luasrc @dibdot 
- [ ] luci-app-dockerman/luasrc @lisaac @feckert
- [ ] luci-app-dump1090/luasrc @Noltari 
- [ ] luci-app-dynapoint/luasrc @ascob
- [ ] luci-app-ltqtapi/luasrc @blogic 
- [ ] luci-app-lxc/luasrc @danrl 
- [ ] luci-app-mosquitto/luasrc @karlp
- [ ] luci-app-nft-qos/luasrc @rosysong
- [ ] luci-app-nut/luasrc @danielfdickinson
- [ ] luci-app-ocserv/luasrc @nmav
- [ ] luci-app-openvpn/luasrc @sbyx @jow- 
- [ ] luci-app-privoxy/luasrc @chris5560
- [ ] luci-app-radicale/luasrc @chris5560
- [ ] luci-app-radicale2/luasrc @danielfdickinson
- [ ] luci-app-siitwizard/luasrc @sbyx @jow- 
- [ ] luci-app-splash/luasrc @mmunz 
- [ ] luci-app-squid/luasrc  @ratkaj
- [ ] luci-app-tinyproxy/luasrc @EricLuehrsen
- [ ] luci-app-unbound/luasrc @EricLuehrsen 
- [ ] luci-app-vnstat/luasrc @jow- 
- [ ] luci-app-wifischedule/luasrc @newkit

Some of these might not make sense to convert. Some may have no users whatsoever - only download statistics will reveal info there, but those statistics are currently unavailable. At least a few of the maintainers likely have no time or desire to undertake such a conversion.

[...]

PR #8018, opened 2025-10-19
luci-app-tailscale-community: Init checkin

Tokisaki-Galaxy:
add new package. use to gui manage tailscale configuare https://github.com/Tokisaki-Galaxy/luci-app-tailscale-community/

- [x] Tested on: (aarch64, openwrt 24.10.03, Edge) :white_check_mark:
setting
status

LUCI_DEPENDS:=+tailscale +coreutils-base64

### luci-app-tailscale: Add Tailscale WebUI application

This Pull Request introduces a comprehensive LuCI Web Interface for the Tailscale VPN service. This application provides users with a user-friendly way to configure their Tailscale node settings and monitor the status of their virtual network (tailnet) directly from the OpenWrt administration panel.

I confirm that I have tested this application thoroughly and it adheres to the LuCI coding guidelines, including full i18n support and use of the modern `luasrc` structure.

#### Key Features

*   **Status Overview**: Displays core node information (running status, IPv4/IPv6 address, Tailnet name).
*   **Peer Status Table**: Lists all connected network devices (peers) in the tailnet, showing their hostname, IPs, OS, connection status (Direct/Relay), and last seen time.
*   **Node Settings Configuration**: Allows users to manage instant node configurations via `tailscale set` commands, including:
    *   Accept/Advertise Routes
    *   Advertise as Exit Node
    *   Use Exit Node / Allow LAN Access
    *   SSH, Shields Up, Auto-Updates, Custom Hostname
*   **Daemon Environment Settings**: Provides an interface to set service-level environment variables (e.g., Custom MTU, Memory Reduction optimization `GOGC=10`), which trigger a service restart upon commit.

#### Technical Implementation Details

1.  **CBI & Status Polling**: The application uses the CBI framework for configuration and relies on polling data from the `tailscale status` and `tailscale status --json` commands via a custom data model (`luci.model.tailscale_data`).
2.  **Configuration Separation**:
    *   **Instant Settings**: Node settings are applied immediately on `on_after_commit` using `tailscale set ...`.
    *   **Daemon Settings**: Environment variables are written to `/etc/profile.d/tailscale-env.sh` and require a service restart. The PR code correctly handles the service restart only when these daemon settings are changed.
3.  **Dependency**: This application requires the `tailscale` core package to be installed to function. The `Makefile` correctly sets `+tailscale +coreutils-base64`.
4.  **Internationalization**: Full i18n support is implemented with `po/` files included.

#### Test Notes

Tested on Openwrt 24.10.03 on a aarch64 device using the latest `tailscale` package.

### DIFF
--- a/applications/luci-app-tailscale-community/Makefile
+++ b/applications/luci-app-tailscale-community/Makefile
@@ -1,0 +1,12 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for Tailscale
+LUCI_URL:=https://github.com/tokisaki-galaxy/luci-app-tailscale-community
+PKG_DESCRIPTION:=Provides a LuCI Web management interface for Tailscale, allowing viewing status, configuring nodes and daemons.
+PKG_MAINTAINER:=Tokisaki-Galaxy <moebest@outlook.jp>
+LUCI_DEPENDS:=+tailscale +ip +luci-base
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
+++ b/applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
@@ -1,0 +1,574 @@
+'use strict';
+'require view';
+'require form';
+'require rpc';
+'require ui';
+'require uci';
+'require tools.widgets as widgets';
+
+const callGetStatus = rpc.declare({ object: 'tailscale', method: 'get_status' });
+const callGetSettings = rpc.declare({ object: 'tailscale', method: 'get_settings' });
+const callSetSettings = rpc.declare({ object: 'tailscale', method: 'set_settings', params: ['form_data'] });
+const callDoLogin = rpc.declare({ object: 'tailscale', method: 'do_login', params: ['form_data'] });
+const callDoLogout = rpc.declare({ object: 'tailscale', method: 'do_logout' });
+const callGetSubroutes = rpc.declare({ object: 'tailscale', method: 'get_subroutes' });
+const callSetupFirewall = rpc.declare({ object: 'tailscale', method: 'setup_firewall' });
+let map;
+
+const tailscaleSettingsConf = [
+	[form.ListValue, 'fw_mode', _('Firewall Mode'), _('Select the firewall backend for Tailscale to use. Requires service restart to take effect.'), {values: ['nftables','iptables'],rmempty: false}],
+	[form.Flag, 'accept_routes', _('Accept Routes'), _('Allow accepting routes announced by other nodes.'), { rmempty: false }],
+	[form.Flag, 'advertise_exit_node', _('Advertise Exit Node'), _('Declare this device as an Exit Node.'), { rmempty: false }],
+	[form.Flag, 'exit_node_allow_lan_access', _('Allow LAN Access'), _('When using the exit node, access to the local LAN is allowed.'), { rmempty: false }],
+	[form.Flag, 'runwebclient', _('Enable Web Interface'), _('Expose a web interface on port 5252 for managing this node over Tailscale.'), { rmempty: false }],
+	[form.Flag, 'nosnat', _('Disable SNAT'), _('Disable Source NAT (SNAT) for traffic to advertised routes. Most users should leave this unchecked.'), { rmempty: false }],
+	[form.Flag, 'shields_up', _('Shields Up'), _('When enabled, blocks all inbound connections from the Tailscale network.'), { rmempty: false }],
+	[form.Flag, 'ssh', _('Enable Tailscale SSH'), _('Allow connecting to this device through the SSH function of Tailscale.'), { rmempty: false }],
+	[form.Flag, 'disable_magic_dns', _('Disable MagicDNS'), _('Use system DNS instead of MagicDNS.'), { rmempty: false }]
+];
+
+const accountConf = [];	// dynamic created in render function
+
+const daemonConf = [
+	//[form.Value, 'daemon_mtu', _('Daemon MTU'), _('Set a custom MTU for the Tailscale daemon. Leave blank to use the default value.'), { datatype: 'uinteger', placeholder: '1280' }, { rmempty: false }],
+	[form.Flag, 'daemon_reduce_memory', _('(Experimental) Reduce Memory Usage'), _('Enabling this option can reduce memory usage, but it may sacrifice some performance (set GOGC=10).'), { rmempty: false }]
+];
+
+const derpMapUrl = 'https://controlplane.tailscale.com/derpmap/default';
+let regionCodeMap = {};
+
+// this function copy from luci-app-frpc. thx
+function setParams(o, params) {
+	if (!params) return;
+
+	for (const [key, val] of Object.entries(params)) {
+		if (key === 'values') {
+			[].concat(val).forEach(v =>
+				o.value.apply(o, Array.isArray(v) ? v : [v])
+			);
+		} else if (key === 'depends') {
+			const arr = Array.isArray(val) ? val : [val];
+			o.deps = arr.map(dep => Object.assign({}, ...o.deps, dep));
+		} else {
+			o[key] = val;
+		}
+	}
+
+	if (params.datatype === 'bool')
+		Object.assign(o, { enabled: 'true', disabled: 'false' });
+}
+
+// this function copy from luci-app-frpc. thx
+function defTabOpts(s, t, opts, params) {
+	for (let i = 0; i < opts.length; i++) {
+		const opt = opts[i];
+		const o = s.taboption(t, opt[0], opt[1], opt[2], opt[3]);
+		setParams(o, opt[4]);
+		setParams(o, params);
+	}
+}
+
+function getRunningStatus() {
+	return L.resolveDefault(callGetStatus(), { running: false }).then(function (res) {
+		return res;
+	});
+}
+
+function formatBytes(bytes) {
+	const bytes_num = parseInt(bytes, 10);
+	if (isNaN(bytes_num) || bytes_num === 0) return '-';
+	const k = 1000;
+	const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
+	const i = Math.floor(Math.log(bytes_num) / Math.log(k));
+	return parseFloat((bytes_num / Math.pow(k, i)).toFixed(2)) + ' ' + sizes[i];
+}
+
+function formatLastSeen(d) {
+	if (!d) return _('N/A');
+	if (d === '0001-01-01T00:00:00Z') return _('Now');
+	const t = new Date(d);
+	if (isNaN(t)) return _('Invalid Date');
+	const diff = (Date.now() - t) / 1000;
+	if (diff < 0) return t.toLocaleString();
+	if (diff < 60) return _('Just now');
+
+	const mins = diff / 60, hrs = mins / 60, days = hrs / 24;
+	const fmt = (n, s, p) => `${Math.floor(n)} ${Math.floor(n) === 1 ? _(s) : _(p)} ${_('ago')}`;
+
+	if (mins < 60) return fmt(mins, 'minute', 'minutes');
+	if (hrs < 24) return fmt(hrs, 'hour', 'hours');
+	if (days < 30) return fmt(days, 'day', 'days');
+
+	return t.toISOString().slice(0, 10);
+}
+
+async function initializeRegionMap() {
+	const cacheKey = 'tailscale_derp_map_cache';
+	const ttl = 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds
+
+	try {
+		const cachedItem = localStorage.getItem(cacheKey);
+		if (cachedItem) {
+			const cached = JSON.parse(cachedItem);
+			// Check if the cached data is still valid (not expired)
+			if (Date.now() - cached.timestamp < ttl) {
+				regionCodeMap = cached.data;
+				return;
+			}
+		}
+	} catch (e) {
+		ui.addTimeLimitedNotification(null, [ E('p', _('Error reading cached DERP region map: %s').format(e.message || _('Unknown error'))) ], 7000, 'error');
+	}
+
+	// If no valid cache, fetch from the network
+	try {
+		const response = await fetch(derpMapUrl);
+		if (!response.ok) {
+			return;
+		}
+		const data = await response.json();
+		const newRegionMap = {};
+		for (const regionId in data.Regions) {
+			const region = data.Regions[regionId];
+			const code = (region.RegionCode || '').toLowerCase();
+			const name = region.RegionName || region.RegionCode || `Region ${regionId}`;
+			newRegionMap[code] = name;
+		}
+		regionCodeMap = newRegionMap;
+
+		// Save the newly fetched data to the cache
+		try {
+			const itemToCache = {
+				timestamp: Date.now(),
+				data: regionCodeMap
+			};
+			localStorage.setItem(cacheKey, JSON.stringify(itemToCache));
+		} catch (e) {
+			ui.addTimeLimitedNotification(null, [ E('p', _('Error caching DERP region map: %s').format(e.message || _('Unknown error'))) ], 7000, 'error');
+		}
+	} catch (error) {
+		ui.addTimeLimitedNotification(null, [ E('p', _('Error fetching DERP region map: %s').format(error.message || _('Unknown error'))) ], 7000, 'error');
+	}
+}
+
+function formatConnectionInfo(info) {
+	if (!info) { return '-'; }
+	if (typeof info === 'string' && info.length === 3) {
+		const lowerCaseInfo = info.toLowerCase();
+		return regionCodeMap[lowerCaseInfo] || info;
+	}
+	return info;
+}
+
+function renderStatus(status) {
+	// If status object is not yet available, show a loading message.
+	if (!status || !status.hasOwnProperty('status')) {
+		return E('em', {}, _('Collecting data ...'));
+	}
+
+	const notificationId = 'tailscale_health_notification';
+	let notificationElement = document.getElementById(notificationId);
+	if (status.health != '') {
+		const message = _('Tailscale Health Check: %s').format(status.health);
+		if (notificationElement) {
+			notificationElement.textContent = message;
+		}
+		else {
+			let newNotificationContent = E('p', { 'id': notificationId }, message);
+			ui.addNotification(null, newNotificationContent, 'info');
+		}
+	}else{
+		try{
+			notificationElement.parentNode.parentNode.remove();
+		}catch(e){}
+	}
+
+	if (Object.keys(regionCodeMap).length === 0) {
+		initializeRegionMap();
+	}
+
+	// --- Part 1: Handle non-running states ---
+
+	// State: Tailscale binary not found.
+	if (status.status == 'not_installed') {
+		return E('dl', { 'class': 'cbi-value' }, [
+			E('dt', {}, _('Service Status')),
+			E('dd', {}, E('span', { 'style': 'color:red;' }, E('strong', {}, _('TAILSCALE NOT FOUND'))))
+		]);
+	}
+
+	// State: Logged out, requires user action.
+	if (status.status == 'logout') {
+		return E('dl', { 'class': 'cbi-value' }, [
+			E('dt', {}, _('Service Status')),
+			E('dd', {}, [
+				E('span', { 'style': 'color:orange;' }, E('strong', {}, _('LOGGED OUT'))),
+				E('br'),
+				E('span', {}, _('Please use the login button in the settings below to authenticate.'))
+			])
+		]);
+	}
+
+	// State: Service is installed but not running.
+	if (status.status != 'running') {
+		return E('dl', { 'class': 'cbi-value' }, [
+			E('dt', {}, _('Service Status')),
+			E('dd', {}, E('span', { 'style': 'color:red;' }, E('strong', {}, _('NOT RUNNING'))))
+		]);
+	}
+
+	// --- Part 2: Render the full status display for a running service ---
+
+	// A helper array to define the data for the main status table.
+	const statusData = [
+		{ label: _('Service Status'), value: E('span', { 'style': 'color:green;' }, E('strong', {}, _('RUNNING'))) },
+		{ label: _('Version'), value: status.version || 'N/A' },
+		{ label: _('TUN Mode'), value: status.TUNMode ? _('Enabled') : _('Disabled') },
+		{ label: _('Tailscale IPv4'), value: status.ipv4 || 'N/A' },
+		{ label: _('Tailscale IPv6'), value: status.ipv6 || 'N/A' },
+		{ label: _('Tailnet Name'), value: status.domain_name || 'N/A' }
+	];
+
+	// Build the horizontal status table using the data array.
+	const statusTable = E('table', { 'style': 'width: 100%; border-spacing: 0 5px;' }, [
+		E('tr', {}, statusData.map(item => E('td', { 'style': 'padding-right: 20px;' }, E('strong', {}, item.label)))),
+		E('tr', {}, statusData.map(item => E('td', { 'style': 'padding-right: 20px;' }, item.value)))
+	]);
+
+	// --- Part 3: Render the Peers/Network Devices table ---
+
+	const peers = status.peers;
+	let peersContent;
+
+	if (!peers || Object.keys(peers).length === 0) {
+		// Display a message if no peers are found.
+		peersContent = E('p', {}, _('No peer devices found.'));
+	} else {
+		// Define headers for the peers table.
+		const peerTableHeaders = [
+			{ text: _('Status'), style: 'width: 80px;' },
+			{ text: _('Hostname') },
+			{ text: _('Tailscale IP') },
+			{ text: _('OS') },
+			{ text: _('Connection Info') },
+			{ text: _('RX') },
+			{ text: _('TX') },
+			{ text: _('Last Seen') }
+		];
+
+		// Build the peers table.
+		peersContent = E('table', { 'class': 'cbi-table' }, [
+			// Table Header Row
+			E('tr', { 'class': 'cbi-table-header' }, peerTableHeaders.map(header => {
+				let th_style = 'padding-right: 20px; text-align: left;';
+				if (header.style) {
+					th_style += header.style;
+				}
+				return E('th', { 'class': 'cbi-table-cell', 'style': th_style }, header.text);
+			})),
+
+			// Table Body Rows (one for each peer)
+			...Object.entries(peers).map(([peerid, peer]) => {
+				const td_style = 'padding-right: 20px;';
+
+				return E('tr', { 'class': 'cbi-rowstyle-1' }, [
+					E('td', { 'class': 'cbi-value-field', 'style': td_style },
+						E('span', {
+							'style': `color:${peer.exit_node ? 'blue' : (peer.online ? 'green' : 'gray')};`,
+							'title': (peer.exit_node ? _('Exit Node') + ' ' : '') + (peer.online ? _('Online') : _('Offline'))
+						}, peer.online ? '●' : '○')
+					),
+					E('td', { 'class': 'cbi-value-field', 'style': td_style }, E('strong', {}, peer.hostname + (peer.exit_node_option ? ' (ExNode)' : ''))),
+					E('td', { 'class': 'cbi-value-field', 'style': td_style }, peer.ip || 'N/A'),
+					E('td', { 'class': 'cbi-value-field', 'style': td_style }, peer.ostype || 'N/A'),
+					E('td', { 'class': 'cbi-value-field', 'style': td_style }, formatConnectionInfo(peer.linkadress || '-')),
+					E('td', { 'class': 'cbi-value-field', 'style': td_style }, formatBytes(peer.rx)),
+					E('td', { 'class': 'cbi-value-field', 'style': td_style }, formatBytes(peer.tx)),
+					E('td', { 'class': 'cbi-value-field', 'style': td_style }, formatLastSeen(peer.lastseen))
+				]);
+			})
+		]);
+	}
+
+	// Combine all parts into a single DocumentFragment.
+	// Using E() without a tag name creates a fragment, which is perfect for grouping elements.
+	return E([
+		statusTable,
+		E('div', { 'style': 'margin-top: 25px;' }, [
+			E('h4', {}, _('Network Devices')),
+			peersContent
+		])
+	]);
+}
+
+return view.extend({
+	load() {
+		return Promise.all([
+			L.resolveDefault(callGetStatus(), { running: '', peers: [] }),
+			L.resolveDefault(callGetSettings(), { accept_routes: false }),
+			L.resolveDefault(callGetSubroutes(), { routes: [] })
+		])
+		.then(function([status, settings_from_rpc, subroutes]) {
+			return uci.load('tailscale').then(function() {
+				if (uci.get('tailscale', 'settings') === null) {
+					// No existing settings found; initialize UCI with RPC settings
+					uci.add('tailscale', 'settings', 'settings');
+					uci.set('tailscale', 'settings', 'fw_mode', 'nftables');
+					uci.set('tailscale', 'settings', 'accept_routes', (settings_from_rpc.accept_routes ? '1' : '0'));
+					uci.set('tailscale', 'settings', 'advertise_exit_node', ((settings_from_rpc.advertise_exit_node || false) ? '1' : '0'));
+					uci.set('tailscale', 'settings', 'advertise_routes', (settings_from_rpc.advertise_routes || []).join(', '));
+					uci.set('tailscale', 'settings', 'exit_node', settings_from_rpc.exit_node || '');
+					uci.set('tailscale', 'settings', 'exit_node_allow_lan_access', ((settings_from_rpc.exit_node_allow_lan_access || false) ? '1' : '0'));
+					uci.set('tailscale', 'settings', 'ssh', ((settings_from_rpc.ssh || false) ? '1' : '0'));
+					uci.set('tailscale', 'settings', 'shields_up', ((settings_from_rpc.shields_up || false) ? '1' : '0'));
+					uci.set('tailscale', 'settings', 'runwebclient', ((settings_from_rpc.runwebclient || false) ? '1' : '0'));
+					uci.set('tailscale', 'settings', 'nosnat', ((settings_from_rpc.nosnat || false) ? '1' : '0'));
+					uci.set('tailscale', 'settings', 'disable_magic_dns', ((settings_from_rpc.disable_magic_dns || false) ? '1' : '0'));
+
+					uci.set('tailscale', 'settings', 'daemon_reduce_memory', '0');
+					uci.set('tailscale', 'settings', 'daemon_mtu', '');
+					return uci.save();
+				}
+			}).then(function() {
+				return [status, settings_from_rpc, subroutes];
+			});
+		});
+	},
+
+	render ([status = {}, settings = {}, subroutes_obj]) {
+		const subroutes = (subroutes_obj && subroutes_obj.routes) ? subroutes_obj.routes : [];
+
+		let s;
+		map = new form.Map('tailscale', _('Tailscale'), _('Tailscale is a mesh VPN solution that makes it easy to connect your devices securely. This configuration page allows you to manage Tailscale settings on your OpenWrt device.'));
+
+		s = map.section(form.NamedSection, '_status');
+		s.anonymous = true;
+		s.render = function (section_id) {
+			L.Poll.add(
+				function () {
+					return getRunningStatus().then(function (res) {
+						const view = document.getElementById("service_status_display");
+						if (view) {
+							const content = renderStatus(res);
+							view.replaceChildren(content);
+						}
+
+						// login button only available when logged out
+						const login_btn=document.getElementsByClassName('cbi-button cbi-button-apply')[0];
+						if(login_btn) { login_btn.disabled=(res.status != 'logout'); }
+					});
+				}, 10);
+
+			return E('div', { 'id': 'service_status_display', 'class': 'cbi-value' },
+				_('Collecting data ...')
+			);
+		}
+
+		// Bind settings to the 'settings' section of uci
+		s = map.section(form.NamedSection, 'settings', 'settings', _('Settings'));
+		s.dynamic = true;
+
+		// Create the "General Settings" tab and apply tailscaleSettingsConf
+		s.tab('general', _('General Settings'));
+
+		defTabOpts(s, 'general', tailscaleSettingsConf, { optional: false });
+
+		const en = s.taboption('general', form.ListValue, 'exit_node', _('Exit Node'), _('Select an exit node from the list. If enabled, Allow LAN Access is enabled implicitly.'));
+		en.value('', _('None'));
+		if (status.peers) {
+			Object.values(status.peers).forEach(function(peer) {
+				if (peer.exit_node_option) {
+					const primaryIp = peer.ip.split('<br>')[0];
+					const label = peer.hostname ? `${peer.hostname} (${primaryIp})` : primaryIp;
+					en.value(primaryIp, label);
+				}
+			});
+		}
+		en.rmempty = true;
+		en.cfgvalue = function(section_id) {
+			if (status && status.status === 'running' && status.peers) {
+				for (const id in status.peers) {
+					if (status.peers[id].exit_node) {
+						return status.peers[id].ip.split('<br>')[0];
+					}
+				}
+				return '';
+			}
+			return uci.get('tailscale', 'settings', 'exit_node') || '';
+		};
+
+		const o = s.taboption('general', form.DynamicList, 'advertise_routes', _('Advertise Routes'),_('Advertise subnet routes behind this device. Select from the detected subnets below or enter custom routes (comma-separated).'));
+		if (subroutes.length > 0) {
+			subroutes.forEach(function(subnet) {
+				o.value(subnet, subnet);
+			});
+		}
+		o.rmempty = true;
+
+		const fwBtn = s.taboption('general', form.Button, '_setup_firewall', _('Auto Configure Firewall'));
+		fwBtn.description = _('Experimental: applies minimal firewall and interface setup for Tailscale. It will create/patch network.tailscale (proto none, device tailscale0), add a firewall zone "tailscale" with ACCEPT/ACCEPT/ACCEPT, masq, mtu_fix, and ensure forwarding tailscale<->lan. It reloads network/firewall only if changes are made.');
+		fwBtn.inputstyle = 'action';
+		fwBtn.onclick = function() {
+			const btn = this;
+			btn.disabled = true;
+			return callSetupFirewall().then(function(res) {
+				const msg = res?.message || _('Firewall configuration applied.');
+				ui.addNotification(null, E('p', {}, msg), 'info');
+			}).catch(function(err) {
+				ui.addNotification(null, E('p', {}, _('Failed to configure firewall: %s').format(err?.message || err || 'Unknown error')), 'error');
+			}).finally(function() {
+				btn.disabled = false;
+			});
+		};
+
+		// Create the account settings
+		s.tab('account', _('Account Settings'));
+		defTabOpts(s, 'account', accountConf, { optional: false });
+
+		const loginBtn = s.taboption('account', form.Button, '_login', _('Login'),
+		_('Click to get a login URL for this device.')
+		+'<br>'+_('If the timeout is displayed, you can refresh the page and click Login again.'));
+		loginBtn.inputstyle = 'apply';
+
+		const customLoginUrl = s.taboption('account', form.Value, 'custom_login_url',
+			_('Custom Login Server'),
+			_('Optional: Specify a custom control server URL (e.g., a Headscale instance, https://example.com).')
+			+'<br>'+_('Leave blank for default Tailscale control plane.')
+		);
+		customLoginUrl.placeholder = '';
+		customLoginUrl.rmempty = true;
+
+		const customLoginAuthKey = s.taboption('account', form.Value, 'custom_login_AuthKey',
+			_('Custom Login Server Auth Key'),
+			_('Optional: Specify an authentication key for the custom control server. Leave blank if not required.')
+			+'<br>'+_('If you are using custom login server but not providing an Auth Key, will redirect to the login page without pre-filling the key.')
+		);
+		customLoginAuthKey.placeholder = '';
+		customLoginAuthKey.rmempty = true;
+
+		const logoutBtn = s.taboption('account', form.Button, '_logout', _('Logout'),
+		_('Click to Log out account on this device.')
+		+'<br>'+_('Disconnect from Tailscale and expire current node key.'));
+		logoutBtn.inputstyle = 'apply';
+		logoutBtn.id = 'tailscale_logout_btn';
+
+		loginBtn.onclick = function() {
+			const customServerInput = document.getElementById('widget.cbid.tailscale.settings.custom_login_url');
+			const customServer = customServerInput ? customServerInput.value : '';
+			const customserverAuthInput = document.getElementById('widget.cbid.tailscale.settings.custom_login_AuthKey');
+			const customServerAuth = customserverAuthInput ? customserverAuthInput.value : '';
+			const loginWindow = window.open('', '_blank');
+			if (!loginWindow) {
+				ui.addTimeLimitedNotification(null, [ E('p', _('Could not open a new tab. Please check if your browser or an extension blocked the pop-up.')) ], 10000, 'error');
+				return;
+			}
+			// Display a prompt message in the new window
+			const doc = loginWindow.document;
+			doc.body.innerHTML = 
+				'<h2>' + _('Tailscale Login') + '</h2>' +
+				'<p>' + _('Requesting Tailscale login URL... Please wait.') + '</p>' +
+				'<p>' + _('This can take up to 30 seconds.') + '</p>';
+
+			ui.showModal(_('Requesting Login URL...'), E('em', {}, _('Please wait.')));
+			const payload = {
+				loginserver: customServer || '',
+				loginserver_authkey: customServerAuth || ''
+			};
+			// Show a "loading" modal and execute the asynchronous RPC call
+			ui.showModal(_('Requesting Login URL...'), E('em', {}, _('Please wait.')));
+			return callDoLogin(payload).then(function(res) {
+				ui.hideModal();
+				if (res && res.url) {
+					// After successfully obtaining the URL, redirect the previously opened tab
+					loginWindow.location.href = res.url;
+				} else {
+					// If it fails, inform the user and they can close the new tab
+					doc.body.innerHTML = 
+						'<h2>' + _('Error') + '</h2>' +
+						'<p>' + _('Failed to get login URL. You may close this tab.') + '</p>';
+					ui.addTimeLimitedNotification(null, [ E('p', _('Failed to get login URL: Invalid response from server.')) ], 7000, 'error');
+				}
+			}).catch(function(err) {
+				ui.hideModal();
+				ui.addTimeLimitedNotification(null, [ E('p', _('Failed to get login URL: %s').format(err.message || _('Unknown error'))) ], 7000, 'error');
+			});
+		};
+
+		logoutBtn.onclick = function() {
+			const confirmationContent = E([
+				E('p', {}, _('Are you sure you want to log out?')
+					+'<br>'+_('This will disconnect this device from your Tailnet and require you to re-authenticate.')),
+				
+				E('div', { 'style': 'text-align: right; margin-top: 1em;' }, [
+					E('button', {
+						'class': 'cbi-button',
+						'click': ui.hideModal
+					}, _('Cancel')),
+					' ', 
+					E('button', {
+						'class': 'cbi-button cbi-button-negative',
+						'click': function() {
+							ui.hideModal();
+							ui.showModal(_('Logging out...'), E('em', {}, _('Please wait.')));
+
+							return callDoLogout().then(function(res) {
+								ui.hideModal();
+								ui.addTimeLimitedNotification(null, [ E('p', _('Successfully logged out.')) ], 5000, 'info');
+							}).catch(function(err) {
+								ui.hideModal();
+								ui.addTimeLimitedNotification(null, [ E('p', _('Logout failed: %s').format(err.message || _('Unknown error'))) ], 7000, 'error');
+							});
+						}
+					}, _('Logout'))
+				])
+			]);
+			ui.showModal(_('Confirm Logout'), confirmationContent);
+		};
+
+		// Create the "Daemon Settings" tab and apply daemonConf
+		//s.tab('daemon', _('Daemon Settings'));
+		//defTabOpts(s, 'daemon', daemonConf, { optional: false });
+
+		return map.render();
+	},
+
+	// The handleSaveApply function is executed after clicking "Save & Apply"
+	handleSaveApply(ev) {
+		return map.save().then(function () {
+			const data = map.data.get('tailscale', 'settings');
+
+			// fix empty value issue
+			if(!data.advertise_exit_node) data.advertise_exit_node = '';
+			if(!data.advertise_routes) data.advertise_routes = '';
+			if(!data.exit_node) data.exit_node = '';
+			if(!data.custom_login_url) data.custom_login_url = '';
+			if(!data.custom_login_AuthKey) data.custom_login_AuthKey = '';
+
+			ui.showModal(_('Applying changes...'), E('em', {}, _('Please wait.')));
+
+			return callSetSettings(data).then(function (response) {
+				if (response.success) {
+					ui.hideModal();
+					setTimeout(function() {
+							ui.addTimeLimitedNotification(null, [ E('p', _('Tailscale settings applied successfully.')) ], 5000, 'info');
+					}, 1000);
+					try {
+						L.ui.changes.revert();
+					} catch (error) {
+						ui.addTimeLimitedNotification(null, [ E('p', _('Error saving settings: %s').format(error || _('Unknown error'))) ], 7000, 'error');
+					}
+				} else {
+					ui.hideModal();
+					ui.addTimeLimitedNotification(null, [ E('p', _('Error applying settings: %s').format(response.error || _('Unknown error'))) ], 7000, 'error');
+				}
+			});
+		}).catch(function(err) {
+			ui.hideModal();
+			//console.error('Save failed:', err);
+			ui.addTimeLimitedNotification(null, [ E('p', _('Failed to save settings: %s').format(err.message)) ], 7000, 'error');
+		});
+	},
+
+	handleSave: null,
+	handleReset: null
+});
--- a/applications/luci-app-tailscale-community/po/templates/community.pot
+++ b/applications/luci-app-tailscale-community/po/templates/community.pot
@@ -1,0 +1,470 @@
+msgid ""
+msgstr "Content-Type: text/plain; charset=UTF-8"
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:34
+msgid "(Experimental) Reduce Memory Usage"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:20
+msgid "Accept Routes"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:425
+msgid "Account Settings"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:21
+msgid "Advertise Exit Node"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:400
+msgid "Advertise Routes"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:400
+msgid ""
+"Advertise subnet routes behind this device. Select from the detected subnets "
+"below or enter custom routes (comma-separated)."
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:22
+msgid "Allow LAN Access"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:20
+msgid "Allow accepting routes announced by other nodes."
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:26
+msgid "Allow connecting to this device through the SSH function of Tailscale."
+msgstr ""
+
+#: applications/luci-app-tailscale-community/root/usr/share/rpcd/acl.d/luci-app-tailscale-community.json:3
+msgid "Allow user access to tailscale"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:547
+msgid "Applying changes..."
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:499
+msgid "Are you sure you want to log out?"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:408
+msgid "Auto Configure Firewall"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:506
+msgid "Cancel"
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:450
+msgid "Click to Log out account on this device."
+msgstr ""
+
+#: applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js:429
+msgid "Click to get a login URL for this device."
+msgstr ""
+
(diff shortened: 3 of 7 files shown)


systemcrash:
Problem: LUA

No new lua additions are accepted into luci. See #3378 and #7310  Migrate to ucode, or javascript. Both of which provide base64 en/decode facilities inbuilt.

Tokisaki-Galaxy:
First, let me express my sincere gratitude for your incredible work on the LuCI framework. It's a powerful and flexible platform.

I am currently in the process of refactoring lua into the ucode+js framework. I've successfully implemented the core functionality, but I've hit a stubborn roadblock that I'm hoping you can provide some guidance on.

My application's goal is not to simply manage a UCI config file in the traditional sense. Instead, its primary purpose is to execute a specific command (`tailscale set --option=value ...`) via a ucode RPC call. The UCI file is essentially used as a temporary vehicle to pass the user's settings from the frontend form to the backend script.

The process works flawlessly: the user changes settings, clicks "Save & Apply," my custom `handleSaveApply` function is triggered, it calls the ucode RPC, the `tailscale` command is executed successfully, and the service is configured.

**The Problem:** However, after the `handleSaveApply` function completes and the page reloads, the "Unsaved Changes: N" warning persists at the top of the page. This is confusing for the end-user, as their changes have already been successfully applied.

I understand this is happening because the `form.Map` object still considers itself to be in a "dirty" state. I have tried several methods to resolve this:

1.  **`map.reset()`**: My first thought was to call `this.map.reset()` within the `handleSaveApply` function after the RPC call succeeds. Unfortunately, this does not seem to have the desired effect, and the warning remains after the page reloads.

2.  **Command-Line UCI**: I also considered attempting to clear the changes via a command-line approach, but I quickly realized that the shell's UCI context (and its buffer) is separate from the one used by the LuCI web interface, making this path unviable.

My question is: **What is the correct, canonical method within a custom `handleSaveApply` function to signal to the LuCI framework that the custom actions have completed successfully and that the form's state should be considered 'clean'?**

Here is a simplified version of my `handleSaveApply` function for context:

```javascript
handleSaveApply: function (ev) {
    var map = this.map;
    return map.save().then(function () {
        var data = uci.get('tailscale', 'settings');
        ui.showModal(_('Applying changes...'), E('em', {}, _('Please wait.')));

        return callSetSettings(data).then(function (response) {
            if (response.success) {
                // I have tried map.reset() and this.map.reset() here,
                // but the "Unsaved Changes" warning persists after reload.
                map.reset(); 

                ui.hideModal();
                ui.addNotification(null, E('p', _('Settings applied successfully.')), 'info');
                setTimeout(function () { window.location.reload(); }, 2000);
            } else {
                // Error handling...
                ui.hideModal();
                ui.addNotification(null, E('p', _('Error: %s').format(response.error)), 'error');
            }
        });
    }).catch(function(err) {
        // General error handling...
        ui.hideModal();
        ui.addNotification(null, E('p', _('Failed to save: %s').format(err.message)), 'error');
    });
},
```

As a side note, I'm learning the hard way how critical getting these details right is. During my development process, a misconfigured backend script (triggered by my LuCI view) unfortunately caused a system lock-up, which prevented all remote access (including SSH) and required physical intervention to recover my router. sad:(

Any guidance or example you could offer on the proper way to handle this "stateless" save-and-apply scenario would be immensely helpful and deeply appreciated.

Thank you again for your time and for maintaining such a fantastic platform.

systemcrash:
You're overriding handleSaveApply, so any of the original functionality which performs the save and commit actions is now replaced by your version (meaning nothing was committed that you changed in the form). I'm not sure why you'd `get` the settings when you're saving. The save action reloads the page. You should be loading the settings at page load in the load promise => (`uci.load('tailscale'),`)actions.

systemcrash:
The 'get the settings and send them to the daemon' should generally happen in an init (procd) script, i.e. in its package in the packages repo.

Tokisaki-Galaxy:
Thank you so much for your insightful reply. It was extremely helpful and clarified the "standard" OpenWrt development pattern perfectly. I now understand that the ideal place for the configuration logic is within the service's init script.

I realize I didn't provide the full context in my first message, which is my oversight. The reason I am using a custom handleSaveApply is that my project is not a new package, but rather a LuCI view for the official, pre-existing Tailscale package.

This official package provides its own service file (/etc/init.d/tailscale), which I should not modify. This service file does not read from the UCI configuration file I am managing.

Therefore, my custom handleSaveApply and RPC backend are intentionally designed to act as a "bridge" between LuCI's UCI-based world and the command-based reality of the Tailscale daemon (tailscale set ...). This seems to be the only way to make them work together without altering the official package's files.

Given this "adapter" architecture, I am still facing the original challenge: the "Unsaved Changes" warning persists after my custom logic has successfully applied the settings.

I have re-attempted to use map.reset() inside the success handler of handleSaveApply, but it doesn't seem to clear the dirty state after the page reloads.

So, my question is: Within this specific architecture where a custom handleSaveApply is necessary, what is the correct method to programmatically clear the form's dirty state and prevent the "Unsaved Changes" warning from appearing after a successful apply action?

Any insight you could provide would be the final key to solving this puzzle. Thank you again for your time and invaluable help

systemcrash:
You're better off writing a ucode script to handle this, so it takes your uci settings and writes out the JSON. ucode is basically javascript.

Tokisaki-Galaxy:
image

Thank you for the guidance. Following your feedback, I have completed the refactoring from Lua to ucode and JavaScript, addressing the policy against new Lua additions. The pull request has been updated accordingly and is now ready for your review. Please let me know if further changes are required.

Tokisaki-Galaxy:
Apologies for the extra merge commit. I used the 'Update branch' button to sync with the latest master, and I realize now that this created a merge commit.
If it's easier for you, please feel free to squash all commits into one when you merge. I will use a rebase workflow for future updates. Thank you for your time and guidance!

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,337 @@
...
+        }).catch(function(err) {
+            ui.hideModal(); 
+            console.error('Save failed:', err); 
+            ui.addNotification(null, E('p', _('Failed to save settings: %s').format(err.message)), 'error');
+        });
+    },
+
+    handleSave: null,
+    handleReset: null
+});
```
systemcrash: \n

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,337 @@
...
+                console.log('Found the "Discard" button in the modal. Clicking it...');
+                discardButton.click();
+            } else {
+                console.error('Could not find the "Discard" button in the modal!');
+            }
+        }, 100);
+            
+        } catch (error) {   
+        }
+                    // 重新加载页面以显示最新状态
```
systemcrash: all comments to english plz

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,337 @@
...
+'require ui';
+'require uci';
+'require tools.widgets as widgets';
+
+var callGetStatus = rpc.declare({ object: 'tailscale', method: 'get_status' });
+var callGetSettings = rpc.declare({ object: 'tailscale', method: 'get_settings' });
+var callSetSettings = rpc.declare({ object: 'tailscale', method: 'set_settings', params: ['form_data'] });
+var callDoLogin = rpc.declare({ object: 'tailscale', method: 'do_login' });
+var callGetSubroutes = rpc.declare({ object: 'tailscale', method: 'get_subroutes' });
+var map;
```
systemcrash: Use ES6. const, let, etc.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,337 @@
...
+];
+
+var daemonConf = [
+    [form.Value, 'daemon_mtu', _('Daemon MTU'), _('Set a custom MTU for the Tailscale daemon. Leave blank to use the default value.'), { datatype: 'uinteger', placeholder: '1280' }, { rmempty: false }],
+    [form.Flag, 'daemon_reduce_memory', _('Reduce Memory Usage'), _('Enabling this option can reduce memory usage, but it may sacrifice some performance (set GOGC=10).'), { rmempty: false }]
+];
+function setParams(o, params) {
+    if (!params) return; for (var key in params) {
+        var val = params[key]; if (key === 'values') {
+            for (var j = 0; j < val.length; j++) {
```
systemcrash: ES6 let j = 0; ...

systemcrash: or better

`for (let value in val) {`

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,337 @@
...
+        finalHtml.push('<dl class="cbi-value"><dt>' + _('Service Status') + '</dt>');
+        finalHtml.push('<dd><span style="color:orange;"><strong>' + _('LOGGED OUT') + '</strong></span></br><span>' + _('Please use the login button in the settings below to authenticate.') + '</span></dd></dl>');
+        return finalHtml.join('');
+    }
+    // ** MODIFICATION END **
+    if (status.status != 'running') {
+        finalHtml.push('<dl class="cbi-value"><dt>' + _('Service Status') + '</dt>');
+        finalHtml.push('<dd><span style="color:red;"><strong>' + _('NOT RUNNING') + '</strong></span></dd></dl>');
+        return finalHtml.join('');
+    }
```
systemcrash: see any of the luci-mod-status js files for how we build HTML tables. e.g.
```js
		const route4tbl = E('table', { 'class': 'table' }, [
			E('tr', { 'class': 'tr table-titles' }, [
				E('th', { 'class': 'th', 'title': device_title }, [ _('Device') ]),
				E('th', { 'class': 'th', 'title': target_title }, [ _('Target') ]),
				E('th', { 'class': 'th', 'title': gateway_title }, [ _('Gateway') ]),
				E('th', { 'class': 'th', 'title': source_title }, [ _('Source') ]),
				E('th', { 'class': 'th', 'title': metric_title }, [ _('Metric') ]),
				E('th', { 'class': 'th', 'title': table_title }, [ _('Table') ]),
				E('th', { 'class': 'th', 'title': proto_title }, [ _('Protocol') ]),
			])
		]);
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,337 @@
...
+            }).then(function() {
+                return rpc_data;
+            });
+        });
+    },
+
+    render: function (data) {
+        var status = data[0] || {};
+        var settings = data[1] || {};
+        var subroutes = (data[2] && data[2].routes) ? data[2].routes : [];
```
systemcrash: array destructure:
`render: function ({status, settings, subroutes}) {`

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,337 @@
...
+            if (discardButton) {
+                console.log('Found the "Discard" button in the modal. Clicking it...');
+                discardButton.click();
+            } else {
+                console.error('Could not find the "Discard" button in the modal!');
+            }
+        }, 100);
+            
+        } catch (error) {   
+        }
```
systemcrash: a note flash here in case of error

systemcrash: `                    ui.addNotification(null, E('p', _('Error saving settings: %s').format(error || 'Unknown error')), 'error');`

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,263 @@
...
+TS_MTU=$(uci_get_state daemon_mtu)
+if [ -n "$TS_MTU" ]; then export TS_DEBUG_MTU="$TS_MTU"; fi
+`;
+            const clean_env_script_content = replace(env_script_content, /\r/g, '');
+            if (new_mtu !== "" || new_reduce_mem === "1") {
+                writefile(env_script_path, clean_env_script_content);
+                exec('chmod 755 '+env_script_path);
+            } else {
+                unlink(env_script_path);
+            }
```
systemcrash: don't do this on every config save/apply.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,263 @@
...
+            popen('tailscale login&','r');
+            sleep(interval);
+        }
+        return { error: 'Could not retrieve login URL from tailscale command.' };
+    }
+};
+
+methods.get_subroutes = {
+    call: function() {
+        let cmd = `ip -4 addr show | awk '!/ lo$|tailscale/ && /inet/ { split($2,c,"/");split(c[1],o,"."); p=c[2]; i=0; for(k=1;k<=4;k++) i=or(lshift(i,8),o[k]); m=lshift(4294967295,32-p); n=and(i,m); print rshift(n,24)"."and(rshift(n,16),255)"."and(rshift(n,8),255)"."and(n,255)"/"p }'`;
```
systemcrash: yuck. 

Use `ip -j route`.  But your Makefile then needs a dependency on `ip`, as busybox version doesn't support JSON yet. Many have wireguard, or other things which typically require ip anyway. If you're installing tailscale, you have the space for it.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,263 @@
...
+            try {
+                let status_data = json(join('',status_json_output.stdout));
+                data.version = status_data.Version || 'Unknown';
+                data.TUNMode = status_data.TUN;
+                if (status_data.BackendState == 'Running') {
+                    data.status =  'running';
+                }
+            } catch (e) { /* ignore */ }
+        }
+        let status_plain_output = exec('tailscale status');
```
systemcrash: why two runs of the same command?

systemcrash:
> Apologies for the extra merge commit. I used the 'Update branch' button to sync with the latest master, and I realize now that this created a merge commit. If it's easier for you, please feel free to squash all commits into one when you merge. I will use a rebase workflow for future updates. Thank you for your time and guidance!

Yes - we don't accept merges. Just rebase in future. Thanks!

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
Tokisaki-Galaxy: Because they actually output different content:

```
root@ImmortalWrt:/usr/share/rpcd/ucode# tailscale status --json |head -n5
{
  "Version": "1.80.3-1 (OpenWrt)",
  "TUN": true,
  "BackendState": "Running",
  "HaveNodeKey": true,
```
```
root@ImmortalWrt:/usr/share/rpcd/ucode# tailscale status | head -n3
100.68.180.114  immortalwrt-1        Tokisaki-Galaxy@ linux   -
100.101.92.37   amy-honor            Tokisaki-Galaxy@ android offline
100.101.92.20   amy-pad              Tokisaki-Galaxy@ windows offline
```

Tokisaki-Galaxy:
Thank you again for your detailed review and guidance. I have addressed the feedback you provided and pushed the updated code.
Please let me know if any further changes are needed.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
systemcrash: That looks like a bug according to the docs. https://tailscale.com/kb/1080/cli#status

===

Get the status of your connections to other Tailscale devices.
...
Running tailscale status with the --json flag returns a machine-readable JSON response.
...
Unlike tailscale status, using this flag gives a detailed list of peers and users in your tailnet that makes it well-suited for automation tasks. It also includes additional metadata associated with your device.

`tailscale status --json | jq -r '.Peer[].Relay | select(.!="")' | sort | uniq -c | sort -nr`

===

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+'require tools.widgets as widgets';
+
+let callGetStatus = rpc.declare({ object: 'tailscale', method: 'get_status' });
+let callGetSettings = rpc.declare({ object: 'tailscale', method: 'get_settings' });
+let callSetSettings = rpc.declare({ object: 'tailscale', method: 'set_settings', params: ['form_data'] });
+let callDoLogin = rpc.declare({ object: 'tailscale', method: 'do_login' });
+let callGetSubroutes = rpc.declare({ object: 'tailscale', method: 'get_subroutes' });
+let map;
+
+let tailscaleSettingsConf = [
```
systemcrash: Define all common functions and layout objects as `const` so they're not overwritten.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+        statusTable,
+        E('div', { 'style': 'margin-top: 25px;' }, [
+            E('h4', {}, _('Network Devices')),
+            peersContent
+        ])
+    ]);
+}
+
+return view.extend({
+    load: function() {
```
systemcrash: load() {

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+        return Promise.all([
+            L.resolveDefault(callGetStatus(), { running: '', peers: [] }),
+            L.resolveDefault(callGetSettings(), { accept_routes: false }),
+            L.resolveDefault(callGetSubroutes(), { routes: [] })
+        ])
+        .then(function(rpc_data) {
```
systemcrash: `        .then(function({status, settings, routes}) {`
and unpack as necessary further down as objects or arrays.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+                    uci.set('tailscale', 'settings', 'daemon_mtu', '');
+                    return uci.save();
+                }
+            }).then(function() {
+                return rpc_data;
+            });
+        });
+    },
+
+    render: function (data) {
```
systemcrash: render() {

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+
+        // Create the "Daemon Settings" tab and apply daemonConf
+        s.tab('daemon', _('Daemon Settings'));
+        defTabOpts(s, 'daemon', daemonConf, { optional: false });
+
+        return map.render();
+    },
+
+    // The handleSaveApply function is executed after clicking "Save & Apply"
+    handleSaveApply: function (ev) {
```
systemcrash: handleSaveApply(ev) {

handle all functions like this

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+function getRunningStatus() {
+    return L.resolveDefault(callGetStatus(), { running: false }).then(function (res) {
+        return res;
+    });
+}
+
+function formatBytes(bytes) {
+    let bytes_num = parseInt(bytes, 10);
+    if (isNaN(bytes_num) || bytes_num === 0) return '-';
+    let k = 1024;
```
systemcrash: This means one MB == 1048576 bytes, which is not correct. MB == Million Bytes. Or Mega. SI mandates that one million or one MB == 1000000 bytes. So `k = 1000`.

So either k = 1024 and the units are MiB (Mebibytes, base 2) or k = 1000 (SI, base 10) and the units are MB.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+                val = [val]; let deps = []; for (let j = 0; j < val.length; j++) {
+                    let d = {}; for (let vkey in val[j])
+                        d[vkey] = val[j][vkey]; for (let k = 0; k < o.deps.length; k++) { for (let dkey in o.deps[k]) { d[dkey] = o.deps[k][dkey]; } }
+                    deps.push(d);
+                }
+            o.deps = deps;
+        } else { o[key] = params[key]; }
+    }
+    if (params['datatype'] === 'bool') { o.enabled = 'true'; o.disabled = 'false'; }
+}
```
systemcrash: Please rework this beast into something a bit more readable. Test this:
```js

function setParams(o, params) {
    if (!params) return;

    for (const [key, val] of Object.entries(params)) {
        if (key === 'values') {
            [].concat(val).forEach(v =>
                o.value.apply(o, Array.isArray(v) ? v : [v])
            );
        } else if (key === 'depends') {
            const arr = Array.isArray(val) ? val : [val];
            o.deps = arr.map(dep => Object.assign({}, ...o.deps, dep));
        } else {
            o[key] = val;
        }
    }

    if (params.datatype === 'bool')
        Object.assign(o, { enabled: 'true', disabled: 'false' });
}

```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,356 @@
...
+function defTabOpts(s, t, opts, params) { for (let i = 0; i < opts.length; i++) { let opt = opts[i]; let o = s.taboption(t, opt[0], opt[1], opt[2], opt[3]); setParams(o, opt[4]); setParams(o, params); } }
```
systemcrash: put onto more lines for readability.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
Tokisaki-Galaxy: yes, tailscale bug or you can call feature
when u use `tailscale status`
```
root@ImmortalWrt:~# tailscale status
100.68.180.114  immortalwrt-1        Tokisaki-Galaxy@ linux   -
100.101.92.16   rrhine               tagged-devices windows active; offers exit node; direct 8.129.26.246:41641
...
```
notice, rrhine node use direct mode.
bug when u use `tailscale status --json`
```
...
"nodekey:xxxxxxxxxxx": {
      "ID": "xxxxxxxxxxx",
      "PublicKey": "nodekey:xxxxxxxxxxx",
      "HostName": "RRhine",
      "DNSName": "rrhine.folk-themis.ts.net.",
      "OS": "windows",
      "UserID": 8851821021232338,
      "TailscaleIPs": [
        "100.101.92.16",
        "fd7a:115c:a1e0::5c01:c010"
      ],
      "AllowedIPs": [
        "100.101.92.16/32",
        "fd7a:115c:a1e0::5c01:c010/128",
        "0.0.0.0/0",
        "::/0"
      ],
      "Tags": [
        "tag:server"
      ],
      "Addrs": null,
      "CurAddr": "8.129.26.246:41641",
      "Relay": "blr",
      "RxBytes": 0,
      "TxBytes": 0,
      "Created": "2025-04-16T03:59:07.95271146Z",
      "LastWrite": "2025-10-22T12:31:15.674907628Z",
      "LastSeen": "0001-01-01T00:00:00Z",
      "LastHandshake": "0001-01-01T00:00:00Z",
      "Online": true,
      "ExitNode": false,
      "ExitNodeOption": true,
      "Active": true,
      "PeerAPIURL": [
        "http://100.101.92.16:53989",
        "http://[fd7a:115c:a1e0::5c01:c010]:62479"
      ],
      "CapMap": {
        "suggest-exit-node": null
      },
      "InNetworkMap": true,
      "InMagicSock": true,
      "InEngine": false
    },
... 
```
you can see `"Relay": "blr",`.
In tailscale network, every node link use replay first, when they establish NAT link, they will use DIRECT mode. But tailscale status --json only record relay info

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: I'm very sorry, I didn't notice it at that time. now I use k=1000

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
systemcrash: Isn't that the same info though?:
```
root@ImmortalWrt:~# tailscale status
100.101.92.16   rrhine               tagged-devices windows active; offers exit node; direct 8.129.26.246:41641
```

and

```
...
"nodekey:xxxxxxxxxxx": {
...
      "HostName": "RRhine",
...
      "TailscaleIPs": [
        "100.101.92.16",
        "fd7a:115c:a1e0::5c01:c010"
      ],
      "AllowedIPs": [
        "100.101.92.16/32",
        "fd7a:115c:a1e0::5c01:c010/128",
        "0.0.0.0/0",
        "::/0"
      ],
...
      "CurAddr": "8.129.26.246:41641",
...
    },
...

Comment (systemcrash) [reply applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc]: Where you have JSON, use it and avoid processing individual lines in a text format.

Comment (Tokisaki-Galaxy) [reply applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc]: you are right. I am very sorry.

File: applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
Comment (systemcrash) [inline applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc:1]: use tab spacing

File: applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc [hunk 19]
```diff
@@ -0,0 +1,285 @@
...
+                try{mkdir('/etc/profile.d');} catch (e) { }
+                if (new_mtu !== "" || new_reduce_mem === "1") {
+                    writefile(env_script_path, env_script_content);
+                    exec('chmod 755 '+env_script_path);
+                } else {
+                    unlink(env_script_path);
+                }
+                popen('/bin/sh -c /etc/init.d/tailscale restart &');
+            }
+        }
```
systemcrash: Do you need to double check new_mtu and new_reduce_mem?

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,285 @@
...
+        return { success: true };
+    }
+};
+
+methods.do_login = {
+    call: function() {
+        let status=methods.get_status.call();
+        if (status.status != 'logout') {
```
systemcrash: If you handle JSOn above, you're working with objects and all this becomes simpler here.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
Tokisaki-Galaxy: No, it was a mistake. Has been repaired.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -54,21 +69,102 @@ function getRunningStatus() {
...
+    }
+    const diffDays = Math.floor(diffHours / 24);
+    if (diffDays < 30) {
+        return `${diffDays} ${diffDays === 1 ? 'day' : 'days'} ago`;
+    }
+    const year = lastSeenDate.getFullYear();
+    const month = String(lastSeenDate.getMonth() + 1).padStart(2, '0');
+    const day = String(lastSeenDate.getDate()).padStart(2, '0');
+    return `${year}-${month}-${day}`;
+}
```
systemcrash: Not 100% sure about the last line `toISOString()` which might need adjusting, but try this function: 
```js
function formatLastSeen(d, _) {
	if (!d) return _('N/A');
	if (d === '0001-01-01T00:00:00Z') return _('Now');
	const t = new Date(d);
	if (isNaN(t)) return _('Invalid Date');
	const diff = (Date.now() - t) / 1000;
	if (diff < 0) return t.toLocaleString();
	if (diff < 60) return _('Just now');

	const mins = diff / 60, hrs = mins / 60, days = hrs / 24;
	const fmt = (n, s, p) => `${Math.floor(n)} ${Math.floor(n) === 1 ? _(s) : _(p)} ${_('ago')}`;

	if (mins < 60) return fmt(mins, 'minute', 'minutes');
	if (hrs < 24) return fmt(hrs, 'hour', 'hours');
	if (days < 30) return fmt(days, 'day', 'days');

	return t.toISOString().slice(0, 10);
}


```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -211,18 +306,17 @@ return view.extend({
...
             });
         });
     },
 
-    render: function (data) {
-        let [status = {}, settings = {}, subroutes_obj] = data;
-        let subroutes = (subroutes_obj && subroutes_obj.routes) ? subroutes_obj.routes : [];
-        
-        let s, o, loginBtn, loginUrl;
+    render ([status = {}, settings = {}, subroutes_obj]) {
```
systemcrash: Here you default to falling back to an empty object. So any functions which take these objects as parameters need to check object properties and functions via `?.` and not just `.`.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,244 @@
...
+
+				// peers list
+				for (let p in status_data.Peer) {
+					p = status_data.Peer[p];
+					peer_map[p.ID] = {
+						ip: join('<br>', p.TailscaleIPs) || '',
+						hostname: split(p.DNSName,'.')[0] || '',
+						ostype: p.OS,
+						online: p.Online,
+						linkadress: (p.CurAddr=="") ? p.Relay : p.CurAddr,
```
systemcrash: `						linkadress: (!p.CurAddr) ? p.Relay : p.CurAddr,`

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,244 @@
...
+		/*if (access(env_script_path)==false) {
+			if (form_data.daemon_mtu != "") {
+				try{ mkdir('/etc/profile.d'); } catch (e) { }
+				writefile(env_script_path, env_script_content);
+				exec('chmod 755 '+env_script_path);
+				popen('/bin/sh -c /etc/init.d/tailscale restart &');
+			}
+		}else{
+			if (form_data.daemon_mtu == "") { unlink(env_script_path); }
+		}*/
```
systemcrash: ?

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,244 @@
...
+			sleep(interval);
+		}
+		return { error: 'Could not retrieve login URL from tailscale command.' };
+	}
+};
+
+methods.get_subroutes = {
+	call: function() {
+		try {
+			let cmd = 'ip -j route';
```
systemcrash: much cleaner, right?

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,464 @@
...
+            return;
+        }
+        const data = await response.json();
+        for (const regionId in data.Regions) {
+            const region = data.Regions[regionId];
+            const code = region.RegionCode.toLowerCase();
+            const name = region.RegionName;
+            regionCodeMap[code] = name;
+        }
+        console.log('Region map initialized successfully.');
```
systemcrash: drop

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,464 @@
...
+    } catch (error) {
+        console.error('Error initializing region map:', error);
```
systemcrash: drop

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,464 @@
...
+        return renderedMap.then(function(node) {
+            const correctFwMode = (settings.fw_mode || 'nftables').split(' ')[0];
+            setTimeout(function() {
+                const fwModeSelect = node.querySelector('#widget\\.cbid\\.tailscale\\.settings\\.fw_mode');
+                if (fwModeSelect) {
+                    console.log(`[Workaround] Forcing dropdown value to: '${correctFwMode}' after view has rendered.`);
+                    fwModeSelect.value = correctFwMode;
+                    fwModeSelect.dispatchEvent(new Event('change'));
+                } else {
+                    console.error('[Workaround] Could not find the fw_mode dropdown to apply fix.');
```
systemcrash: drop

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,464 @@
...
+                        ui.addNotification(null, E('p', _('Tailscale settings applied successfully.')), 'info');
+                }, 1000);
+        try {
+            const indicator = document.querySelector('span[data-indicator="uci-changes"][data-clickable="true"]');
+            if (indicator) {
+                indicator.click();
+                setTimeout(function() {
+                    const discardButton = document.querySelector('.cbi-button.cbi-button-reset');
+                    if (discardButton) {
+                        console.log('Found the "Discard" button in the modal. Clicking it...');
```
systemcrash: drop

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,464 @@
...
+                        discardButton.click();
+                    } else {
+                        console.error('Could not find the "Discard" button in the modal!');
```
systemcrash: flash a notification maybe

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,464 @@
...
+                // Reload the page to display the latest status
+                setTimeout(function () { window.location.reload(); }, 2000);
+                } else {
+                    ui.hideModal();
+                    ui.addNotification(null, E('p', _('Error applying settings: %s').format(response.error || 'Unknown error')), 'error');
+                }
+            });
+        }).catch(function(err) {
+            ui.hideModal(); 
+            console.error('Save failed:', err);
```
systemcrash: drop

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
Tokisaki-Galaxy: According to the description of [issue](https://github.com/tailscale/tailscale/issues/8219) , this method is effective to adjust MTU, but I found that I can't reproduce it effectively on the second test equipment.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
Tokisaki-Galaxy: Yes, thank you for pointing it out

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: Yes, in fact, a notice will definitely flash. But if you want to repair it, you must use a troublesome repair method: using new configuration files and services.

By changing the new configuration file, calling the standard restart service process and restarting the new service, the tailscale setting can be changed, but will it be too much trouble? Now the code can also run well.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: can I call console output when I find an error? or unitfy use ui.addnotifition?

systemcrash: *you* can, but not in the main repo

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
Tokisaki-Galaxy: found answer, some new versions of Tailscale may not work well with this method. but old version can use. so add again

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
systemcrash, line 1: tab spacing

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,464 @@
...
+            const data = map.data.get('tailscale', 'settings');
+            ui.showModal(_('Applying changes...'), E('em', {}, _('Please wait.')));
+
+            return callSetSettings(data).then(function (response) {
+                if (response.success) {
+                    ui.hideModal();
+                setTimeout(function() {
+                        ui.addNotification(null, E('p', _('Tailscale settings applied successfully.')), 'info');
+                }, 1000);
+        try {
```
systemcrash: indent try so it doesn't look like unreachable code

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
systemcrash: not sure clicking UI elements for users is the right approach. Call a function directly instead?

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
systemcrash: Also use unix newlines

Tokisaki-Galaxy: ok

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: I'm not sure whether users always use browsers that support ES2020+. I've modified ucode to return '' values in case of failure. Is it necessary to use ?. ?

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: I'll check it out. It should be ok.

Tokisaki-Galaxy: I found it. I use `L.ui.changes.revert (); ` instead of simulating dom clicks.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,266 @@
...
+// Keep in doubt about its usefulness
+const env_script_path = "/etc/profile.d/tailscale-env.sh";
+const ori_env_script_content = `#!/bin/sh
+# This script is managed by luci-app-tailscale-community.
+uci_get_state() { uci get tailscale.settings."$1" 2>/dev/null; }
+if [ "$(uci_get_state daemon_reduce_memory)" = "1" ]; then export GOGC=10; fi
+TS_MTU=$(uci_get_state daemon_mtu)
+if [ -n "$TS_MTU" ]; then export TS_DEBUG_MTU="$TS_MTU"; fi
+`;
+const env_script_content = replace(ori_env_script_content, /\r/g, '');
```
systemcrash: It seems like this should go into a separate file, either in this package, or in the packages repo. Look at e.g. luci-app-statistics

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: Sorry, it has been revised.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,473 @@
...
+	if (status.health != '') {
+		const message = _('Tailscale Health Check: %s').format(status.health);
+		if (notificationElement) {
+			notificationElement.textContent = message;
+		}
+		else {
+			let newNotificationContent = E('p', { 'id': notificationId }, message);
+			ui.addNotification(null, newNotificationContent, 'info');
+		}
+	}else{try{ notificationElement.parentNode.parentNode.remove(); }catch(e){}}
```
systemcrash:
```js
	} else {
		try {
			notificationElement.parentNode.parentNode.remove();
		} catch(e){}
	}
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,473 @@
...
+		customLoginAuthKey.rmempty = true;
+
+		loginBtn.onclick = function() {
+			const customServerInput = document.getElementById('widget.cbid.tailscale.settings.custom_login_url');
+			const customServer = customServerInput ? customServerInput.value : '';
+			const customserverAuthInput = document.getElementById('widget.cbid.tailscale.settings.custom_login_AuthKey');
+			const customServerAuth = customserverAuthInput ? customserverAuthInput.value : '';
+			const loginWindow = window.open('', '_blank');
+			if (!loginWindow) {
+				ui.addNotification(null, E('p', _('Could not open a new tab. Please disable your pop-up blocker for this site and try again.')), 'error');
```
systemcrash: Where you use `ui.addNotification`, consider using `ui.addTimeLimitedNotification` which keeps the notification around for a limited period, especially for success notifications, so we don't overload the user with mandatory clicks. You can use it for errors, although I suggest that they stick around for slightly longer than warnings, alternatively keep the `ui.addNotification` but only for errors (which we hope are a rare event).

e.g. here
```js
				ui.addTimeLimitedNotification(_('header blah'), [
					E('p', _('msg %s').format(x)),
				], 7000, 'error');
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,473 @@
...
+				return;
+			}
+			// Display a prompt message in the new window
+			loginWindow.document.write(_('Requesting Tailscale login URL... Please wait...<br>The looggest time to get the URL is about 30 seconds.'));
```
systemcrash: Split this and fix looggest -> longest

			loginWindow.document.write(_('Requesting Tailscale login URL... Please wait...') + '<br/> _('The longest time to get the URL is about 30 seconds.'));

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,473 @@
...
+			ui.showModal(_('Requesting Login URL...'), E('em', {}, _('Please wait.')));
+			return callDoLogin(payload).then(function(res) {
+				ui.hideModal();
+				if (res && res.url) {
+					// After successfully obtaining the URL, redirect the previously opened tab
+					loginWindow.location.href = res.url;
+				} else {
+					// If it fails, inform the user and they can close the new tab
+					loginWindow.document.write(_('<br>Failed to get login URL. You may close this tab.'));
+					ui.addNotification(null, E('p', _('>Failed to get login URL: Invalid response from server.')), 'error');
```
systemcrash:
```js
					loginWindow.document.write('<br/>' + _('Failed to get login URL. You may close this tab.'));
					ui.addNotification(null, E('p', _('Failed to get login URL: Invalid response from server.')), 'error');
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,473 @@
...
+				}
+			}).catch(function(err) {
+				ui.hideModal();
+				ui.addNotification(null, E('p', _('Failed to get login URL: %s').format(err.message || 'Unknown error')), 'error');
```
systemcrash: `_('Unknown error')`

systemcrash: Same also further down

miyatsu:
@Tokisaki-Galaxy Any plan to add support for netifd to this? See [ZeroTier MR](https://github.com/openwrt/packages/pull/18912)

Tokisaki-Galaxy:
> @Tokisaki-Galaxy Any plan to add support for netifd to this? See [ZeroTier MR](https://github.com/openwrt/packages/pull/18912)

good idea. but i think this is the tailscale task of package. I just provide luci GUI

Tokisaki-Galaxy:
I modified the code as required. Thank you very much for your careful check.

applications/luci-app-tailscale-community/root/usr/share/rpcd/acl.d/luci-app-tailscale-community.json
```diff
@@ -0,0 +1,30 @@
...
+		},
+		"write": {
+			"ubus": {
+				"tailscale": [
+					"set_settings"
+				]
+			}
+		}
+	}
+}
```
systemcrash: add \n at end

Tokisaki-Galaxy:
Thanks for pointing out that it has been added \n

applications/luci-app-tailscale-community/Makefile
systemcrash, line 1: same line format for all files

Tokisaki-Galaxy:
I missed it when I checked this morning. I'm very sorry.

systemcrash:
OK - squash all your commits and rebase on master please.

Tokisaki-Galaxy:
done. I should have finished these, please check them for me.

applications/luci-app-tailscale-community/Makefile
```diff
@@ -0,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for Tailscale
+PKG_DESCRIPTION:=Provides a LuCI Web management interface for Tailscale, allowing viewing status, configuring nodes and daemons.
+PKG_VERSION:=2.2.2
+PKG_RELEASE:=1
```
systemcrash: drop these - we will use git semantic versioning.

applications/luci-app-tailscale-community/po/templates/community.pot
systemcrash, line 1: And now to fix i18n. The i18n scripts work based on the folder name, so the files should be `community.pot?`. In reality it doesn't matter what the files are named, but this is so later sync works.

1 - delete this file and the zh template (save it for later)
2 - in luci root, run `./build/i18n-init.sh applications/luci-app-tailscale-community`
3 - put zh po from templates in the correct zh_han? folder (looks like simplified)
4 - in luci root, run `./build/i18n-sync.sh applications/luci-app-tailscale-community`

applications/luci-app-tailscale-community/Makefile
systemcrash: drop these two lines - we will use git semantic versioning.

Tokisaki-Galaxy: Can you be more specific? If these two lines are not added to the Makefile, it seems that there is no version number in the compiled ipk file.

Tokisaki-Galaxy: I see, thank you.

applications/luci-app-tailscale-community/po/templates/community.pot
Tokisaki-Galaxy: Thank you very much, it has been revised as required.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,523 @@
...
+			L.resolveDefault(callGetSettings(), { accept_routes: false }),
+			L.resolveDefault(callGetSubroutes(), { routes: [] })
+		])
+		.then(function([status, settings_from_rpc, subroutes]) {
+			return uci.load('tailscale').then(function() {
+				if (uci.get('tailscale', 'settings') === null) {
+					// No existing settings found; initialize UCI with RPC settings
+					uci.add('tailscale', 'settings', 'settings');
+					uci.set('tailscale', 'settings', 'fw_mode', settings_from_rpc.fw_mode);
+					uci.set('tailscale', 'settings', 'accept_routes', (settings_from_rpc.accept_routes ? '1' : '0'));
```
systemcrash, line 317: You do uci reads and writes, but you haven't defined them in the ACL. These will fail.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,265 @@
...
+		let status_json_output = exec('tailscale status --json');
+		let peer_map = {};
+		if (status_json_output.code == 0 && length(status_json_output.stdout) > 0) {
+			try {
+				let status_data = json(join('',status_json_output.stdout));
+				data.version = status_data.Version || 'Unknown';
+				data.health = status_data.Health || '';
+				data.TUNMode = status_data.TUN || 'true';
+				if (status_data.BackendState == 'Running') { data.status =  'running'; }
+				if (status_data.BackendState == 'NeedsLogin') { data.status =  'logout'; }
```
systemcrash: recommend you chain all properties which might not exist:
`status_data?.BackendState`

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,265 @@
...
+
+				data.ipv4 = status_data.Self.TailscaleIPs[0] || 'No IP assigned';
```
systemcrash: `status_data?.Self?.TailscaleIPs`

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,265 @@
...
+					settings.accept_routes = status_data.RouteAll || false;
+					settings.advertise_exit_node = status_data.AdvertiseExitNode || false;
+					settings.advertise_routes = status_data.AdvertiseRoutes || [];
+					settings.exit_node = status_data.ExitNodeID || "";
+					settings.exit_node_allow_lan_access = status_data.ExitNodeAllowLANAccess || false;
+					settings.shields_up = status_data.ShieldsUp || false;
+					settings.ssh = status_data.RunSSH || false;
+					settings.runwebclient = status_data.RunWebClient || false;
+					settings.nosnat = status_data.NoSNAT || false;
+					settings.fw_mode = split(uci.get('tailscale', 'settings', 'fw_mode'),' ')[0] || 'nftables';
```
systemcrash: here also

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: You are right, it did appear in some cases, but it never appeared in my own test.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,272 @@
...
+			uci.set('tailscale', 'settings', key, form_data[key]);
+		}
+		uci.save('tailscale');
+		uci.commit('tailscale');
+
+		// process reduce memory https://github.com/GuNanOvO/openwrt-tailscale
+		// some new versions of Tailscale may not work well with this method
+		if (form_data.daemon_mtu != "" || form_data.daemon_reduce_memory != "") {
+			exec('/bin/sh -c '+env_script_path+' &');
+			popen('/bin/sh -c /etc/init.d/tailscale restart &');
```
systemcrash: Have you verified that this actually works as intended? exec exports the variables subsequently available only to that subshell. I haven't tried, but you should probably source the script in the same popen for the restart command so the exported variables are available.

Tokisaki-Galaxy: I see, this is a bug. However, there are problems with this deamon module.
It seems that there is no option to set this function in the tailscale software package provided by opkg source.

Tokisaki-Galaxy:
Deamon settings need further debugging, and it seems that this function is not available in tailscale in opkg source. Temporarily disable

Tokisaki-Galaxy:
https://github.com/openwrt/packages/issues/27740
Due to the problem of tailscale package, this function cannot be provided for the time being. If it is forcibly used, tailscale will be completely unusable.

SuperSandro2000:
I could not identify any issue with the packaging. There is likely some logic bug or missunderstanding here.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+	[form.Flag, 'nosnat', _('Disable SNAT'), _('Disable Source NAT (SNAT) for traffic to advertised routes. Most users should leave this unchecked.'), { rmempty: false }],
+	[form.Flag, 'shields_up', _('Shields Up'), _('When enabled, blocks all inbound connections from the Tailscale network.'), { rmempty: false }],
+	[form.Flag, 'ssh', _('Enable Tailscale SSH'), _('Allow connecting to this device through the SSH function of Tailscale.'), { rmempty: false }]
+];
+
+const accountConf = [];	// dynamic created in render function
+
+const daemonConf = [
+	//[form.Value, 'daemon_mtu', _('Daemon MTU'), _('Set a custom MTU for the Tailscale daemon. Leave blank to use the default value.'), { datatype: 'uinteger', placeholder: '1280' }, { rmempty: false }],
+	[form.Flag, 'daemon_reduce_memory', _('Reduce Memory Usage'), _('Enabling this option can reduce memory usage, but it may sacrifice some performance (set GOGC=10).'), { rmempty: false }]
```
SuperSandro2000:
```suggestion
	[form.Flag, 'daemon_reduce_memory', _('(Experimental) Reduce Memory Usage'), _('Enabling this option can reduce memory usage, but it may sacrifice some performance (set GOGC=10).'), { rmempty: false }]
```
We should make it clear, that this might backfire and that issues could be created by setting this.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+		initializeRegionMap();
+	}
+
+	// --- Part 1: Handle non-running states ---
+
+	// State: Tailscale binary not found.
+	if (status.status == 'not_installed') {
+		return E('dl', { 'class': 'cbi-value' }, [
+			E('dt', {}, _('Service Status')),
+			E('dd', {}, E('span', { 'style': 'color:red;' }, E('strong', {}, _('NO FOUND TAILSCALE'))))
```
SuperSandro2000:
```suggestion
			E('dd', {}, E('span', { 'style': 'color:red;' }, E('strong', {}, _('TAILSCALE NOT FOUND'))))
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+			L.resolveDefault(callGetStatus(), { running: '', peers: [] }),
+			L.resolveDefault(callGetSettings(), { accept_routes: false }),
+			L.resolveDefault(callGetSubroutes(), { routes: [] })
+		])
+		.then(function([status, settings_from_rpc, subroutes]) {
+			return uci.load('tailscale').then(function() {
+				if (uci.get('tailscale', 'settings') === null) {
+					// No existing settings found; initialize UCI with RPC settings
+					uci.add('tailscale', 'settings', 'settings');
+					uci.set('tailscale', 'settings', 'fw_mode', settings_from_rpc.fw_mode);
```
SuperSandro2000: fw_mode needs to default to nftables on current openwrt releases.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+
+		const loginBtn = s.taboption('account', form.Button, '_login', _('Login'),
+		_('Click to get a login URL for this device.')
+		+'<br>'+_('If the timeout is displayed, you can refresh the page and click Login again.'));
+		loginBtn.inputstyle = 'apply';
+		loginBtn.id = 'tailscale_login_btn';
+
+		const customLoginUrl = s.taboption('account', form.Value, 'custom_login_url',
+			_('Custom Login Server'),
+			_('Optional: Specify a custom control server URL (e.g., a Headscale instance, http(s)://ex.com).')
```
SuperSandro2000:
```suggestion
			_('Optional: Specify a custom control server URL (e.g., a Headscale instance, https://example.com).')
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+		logoutBtn.id = 'tailscale_logout_btn';
+
+		loginBtn.onclick = function() {
+			const customServerInput = document.getElementById('widget.cbid.tailscale.settings.custom_login_url');
+			const customServer = customServerInput ? customServerInput.value : '';
+			const customserverAuthInput = document.getElementById('widget.cbid.tailscale.settings.custom_login_AuthKey');
+			const customServerAuth = customserverAuthInput ? customserverAuthInput.value : '';
+			const loginWindow = window.open('', '_blank');
+			if (!loginWindow) {
+				ui.addTimeLimitedNotification(null, [ E('p', _('Could not open a new tab. Please disable your pop-up blocker for this site and try again.')) ], 10000, 'error');
```
SuperSandro2000: Most browsers now days disable certain pop-ups per default. This will not be the default message covering this but only if people install 3rd party tools?

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+				return;
+			}
+			// Display a prompt message in the new window
+			loginWindow.document.write(_('Requesting Tailscale login URL... Please wait...')
+			+ '<br>'+ _('The longest time to get the URL is about 30 seconds.'));
```
SuperSandro2000:
```suggestion
			+ '<br>'+ _('This can take up to 30 seconds.'));
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+			}).catch(function(err) {
+				ui.hideModal();
+				ui.addTimeLimitedNotification(null, [ E('p', _('Failed to get login URL: %s').format(err.message || _('Unknown error'))) ], 7000, 'error');
+			});
+		};
+
+		logoutBtn.onclick = function() {
+			const confirmationContent = E([
+				E('p', {}, _('Are you sure you want to log out?')
+					+'<br>'+_('This will disconnect the device from your Tailnet and require you to re-authenticate later.')),
```
SuperSandro2000:
```suggestion
					+'<br>'+_('This will disconnect this device from your Tailnet and require you to re-authenticate.')),
```

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -0,0 +1,520 @@
...
+						}
+					}, _('Logout'))
+				])
+			]);
+			ui.showModal(_('Confirm Logout'), confirmationContent);
+		};
+
+		// Create the "Daemon Settings" tab and apply daemonConf
+		//s.tab('daemon', _('Daemon Settings'));
+		//defTabOpts(s, 'daemon', daemonConf, { optional: false });
```
SuperSandro2000, line 530: Whats up with this?

Tokisaki-Galaxy: Because luci-app-tailscale-community relies on tailscale on opkg source by default, the test failed on arm devices and x86 devices, and it will not run if this setting is enabled.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: Thank you very much for your help.

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -0,0 +1,271 @@
+#!/usr/bin/env ucode
+
+'use strict';
+
+import { access, popen, readfile, writefile, unlink } from 'fs';
+import { cursor } from 'uci';
+
+const uci = cursor();
+const env_script_path = "/etc/profile.d/tailscale-env.sh";
```
SuperSandro2000: I don't like this extra env file. If the init script needs some adjustements, I am very open to change it instead of creating this file.

Tokisaki-Galaxy: Yes, this method is not elegant. I will try to use a method similar to `sed -i '/  procd_close_instance/i\procd_set_param env TS_DEBUG_MTU=1410' /etc/init.d/tailscale` later. But wait for the opkg source to update the tailscale package. Most users get it from opkg source.

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
Tokisaki-Galaxy: have been modified prompt

Tokisaki-Galaxy:
Hi @systemcrash,

Thanks again for all your detailed feedback.

Just wanted to post a quick update to clarify the current scope of this PR. Regarding the daemon functionality we discussed earlier, I've decided to leave that for a future PR to keep this initial version focused and easier to review.
I believe I have addressed all the other points from your last review. Please let me know if there's anything else I can do to get this ready for merging.

Thanks!

Tokisaki-Galaxy:
Hi @systemcrash 
I see the automated checks have skips. This is my first contribution to the LuCI project, and I'm very excited about it!

Could you please take a look when you get a chance, or let me know if there are any specific guidelines or improvements I should address? Thanks for your time and guidance

Tokisaki-Galaxy:
I saw the sob format problem in ci and solved https://github.com/openwrt/luci/actions/runs/18994048129
I'm very sorry, because of the git setting problem on the computer, SOB was always set to an uncommon email by mistake. I have fixed this problem again, and theoretically there will be no more SOB errors. I'm really sorry.
```
#git log -1                                                                                                                                   
commit 7473c9200f8321c77d2c672d13805d59e2a6970a (HEAD -> add-luci-app-tailscale-community)
Author: Tokisaki-Galaxy <moebest@outlook.jp>
Date:   Sun Nov 16 00:05:13 2025 +0800

    luci-app-tailscale-community: Add new application

    This commit adds a new LuCI application for managing Tailscale on OpenWrt.

    The application provides a web interface to view service status,
    list network peers, and configure various Tailscale settings,
    such as exit nodes, advertised routes, and daemon options.

    Co-authored-by: Sandro <sandro.jaeckel@gmail.com>
    Signed-off-by: Tokisaki-Galaxy <moebest@outlook.jp>
```

applications/luci-app-tailscale-community/Makefile
```diff
@@ -0,0 +1,12 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for Tailscale
+LUCI_URL:=https://github.com/tokisaki-galaxy/luci-app-tailscale-community
+PKG_DESCRIPTION:=Provides a LuCI Web management interface for Tailscale, allowing viewing status, configuring nodes and daemons.
+PKG_MAINTAINER:=Tokisaki-Galaxy <moebest@outlook.jp>
+LUCI_DEPENDS:=+tailscale +ip +luci-base
```
systemcrash, line 7: we can drop the ip dep, since I don't see it used anywhere.

Tokisaki-Galaxy: ip-tiny/BusyBox doesnot support -j (json output). so I use ip-full

Tokisaki-Galaxy:
update pr to lastest version.
 - add auto config tailscale firewall
 - fix config exit node bug

I configured copilot automatic review incorrectly before, and I'm very sorry.

applications/luci-app-tailscale-community/Makefile
systemcrash: That's correct OK.

systemcrash:
Merged. Thanks @Tokisaki-Galaxy !

PR #8023, opened 2025-10-20
modules: fix default values

AndyChiang888:
fix default values for luci-base and luci-mod-{dashboard,status,system}

This reverts commit 523b3ea372d2d3790655374ee83fa4d1dd82c264.

commit 523b3ea372d2d3790655374ee83fa4d1dd82c264 breaks switching to full timezone name, ts default value of 0 causes the full timezone name to never be used. And hc should not be initialize , leave it undefined and let the browser automatically select it based on the locale unless specified by the user.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/tools/views.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/views.js
@@ -73,12 +73,12 @@ var CBILogreadBox = function(logtag, name) {
 			try {
 				const tz = uci.get('system', '@system[0]', 'zonename')?.replaceAll(' ', '_');
 				const ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0;
-				const hc = uci.get('system', '@system[0]', 'clock_hourcycle') || 'h23';
+				const hc = uci.get('system', '@system[0]', 'clock_hourcycle') || 0;
 				const logEntries = await callLogRead(this.fetchMaxRows, false, true);
 				const dateObj = new Intl.DateTimeFormat(undefined, {
 						dateStyle: 'medium',
 						timeStyle: (ts == 0) ? 'long' : 'full',
-						hourCycle: hc,
+						hourCycle: (hc == 0) ? undefined : hc,
 						timeZone: tz
 				});
 
--- a/modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
+++ b/modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/include/10_router.js
@@ -315,12 +315,12 @@ return baseclass.extend({
 			const date = new Date(unixtime * 1000);
 			const zn = uci.get('system', '@system[0]', 'zonename')?.replaceAll(' ', '_') || 'UTC';
 			const ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0;
-			const hc = uci.get('system', '@system[0]', 'clock_hourcycle') || 'h23';
+			const hc = uci.get('system', '@system[0]', 'clock_hourcycle') || 0;
 
 			datestr = new Intl.DateTimeFormat(undefined, {
 				dateStyle: 'medium',
 				timeStyle: (ts == 0) ? 'long' : 'full',
-				hourCycle: hc,
+				hourCycle: (hc == 0) ? undefined : hc,
 				timeZone: zn
 			}).format(date);
 		}
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/10_system.js
@@ -52,12 +52,12 @@ return baseclass.extend({
 			var date = new Date(unixtime * 1000),
 				zn = uci.get('system', '@system[0]', 'zonename')?.replaceAll(' ', '_') || 'UTC',
 				ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0,
-				hc = uci.get('system', '@system[0]', 'clock_hourcycle') || 'h23';
+				hc = uci.get('system', '@system[0]', 'clock_hourcycle') || 0;
 
 			datestr = new Intl.DateTimeFormat(undefined, {
 				dateStyle: 'medium',
 				timeStyle: (ts == 0) ? 'long' : 'full',
-				hourCycle: hc,
+				hourCycle: (hc == 0) ? undefined : hc,
 				timeZone: zn
 			}).format(date);
 		}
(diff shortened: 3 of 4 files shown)


AndyChiang888:
@systemcrash 
Pleask take a look.
I have tested it locally. After initialize default values, the full timezone name can no longer be switched.
In addition, some locales that use the 12-hour system by default will also be forced to use the 24-hour system.

AndyChiang888:
````
const date = new Date(Date.UTC(2020, 11, 20, 3, 23, 16, 738));
const ts = undefined;
const hc = undefined;
console.log('ts: ', ts, 'hc: ', hc)
console.log(
  new Intl.DateTimeFormat("en-GB", {
    dateStyle: "medium",
    timeZone: "Australia/Sydney",
    timeStyle: (ts == 0) ? 'long' : 'full',
    hourCycle: hc
  }).format(date),
);

print:

"ts: " undefined "hc: " undefined
"20 Dec 2020, 14:23:16 Australian Eastern Daylight Time"
````

````
const date = new Date(Date.UTC(2020, 11, 20, 3, 23, 16, 738));
const ts = undefined;
const hc = undefined;
console.log('ts: ', ts, 'hc: ', hc)
console.log(
  new Intl.DateTimeFormat("en-US", {
    dateStyle: "medium",
    timeZone: "Australia/Sydney",
    timeStyle: (ts == 0) ? 'long' : 'full',
    hourCycle: hc
  }).format(date),
);

print:

"ts: " undefined "hc: " undefined
"Dec 20, 2020, 2:23:16 PM Australian Eastern Daylight Time"
````
hourCycle can be undefined, in which case locales will decide whether to use a 12-hour or 24-hour clock.

ts has only two states, either undefined or 0 (that is, when the full timezone name is unchecked).
The default is undefined, in which case the full timezone name will be used. 
If someone want to use the timezone offset, just uncheck it.

**Comment by AndyChiang888** [comment]
Or change it to this:
````
const ts = uci.get('system', '@system[0]', 'clock_timestyle') || 1;
const hc = uci.get('system', '@system[0]', 'clock_hourcycle') || undefined;
````
Clarify that clock_hourcycle should be  undefined when uci.get fails

**Comment by AndyChiang888** [comment]
We just need to set 'long' for timeStyle and the default value of hourCycle is 'h23' in view.js.

**Comment by AndyChiang888** [comment]
@systemcrash 
please take a look.

**Comment by systemcrash** [comment]
The current defaults match what it was before this change. YYYY/MM/DD 24:MM:SS. Removing the defaults means the time value display can differ noticeably until a user figures out the two new time settings.

**Comment by AndyChiang888** [comment]
````
ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0;
timeStyle: (ts == 0) ? 'long' : 'full',
````
This means that there will never be a switch to the full time zone name.
If users do not compare by the full timezone name flag, they have no way of knowing what the full timezone name is.

**Comment by systemcrash** [comment]
> ```
> ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0;
> timeStyle: (ts == 0) ? 'long' : 'full',
> ```
> 
> This means that there will never be a switch to the full time zone name. 

Only if they haven't changed the new setting you provided.

**Comment by AndyChiang888** [comment]
I tested it according to your settings:
````
ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0;
````
It never switches to the full time zone.
Because:
````
o = s.taboption('general', form.Flag, 'clock_timestyle', _('Full TimeZone Name'), _('Unchecked means the timezone offset (E.g. GMT+1) is displayed'));
o.default = o.enabled;
````
clock_timestyle default is checked (undefined),
````
'ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0;
````
This code causes ts to always be 0.

**Comment by AndyChiang888** [comment]
With
````
ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0; 
````

image

Whether Full TimeZone Name is checked or not, only EDT is displayed.

With
````
ts = uci.get('system', '@system[0]', 'clock_timestyle'); 
````

image

image

**Comment by AndyChiang888** [comment]
Also, the time format should not be defaulted to h23 except in view.js.
It should follow the locale's default format unless manually set by the user.

**Comment by systemcrash** [comment]
I supplied the `||` fallback which prevents `undefined` getting passed anywhere, which means exactly that, undefined. Browsers might change behaviour one day whether or not specs change.

I had this default value thing the other way around :/

> o.default = o.enabled;

This means that the default value is enabled. i.e. '1'. Which means also that when it is the default value, i.e. '1', it is removed from the config. So it becomes '0' here by virtue of uci.get returning null or undefined:

> `'ts = uci.get('system', '@system[0]', 'clock_timestyle') || 0;`

If we remove: `> o.default = o.enabled;` then the default is '0' which gets removed, then it becomes '0' after fallback to ` || 0`. 

We should as much as possible prevent undefined unless it's something under 'our' control in the repo. Recommended (defaults to 'long' TZ):
```js
		o = s.taboption('general', form.Flag, 'clock_timestyle', _('Full TimeZone Name'), _('Unchecked means the timezone offset (E.g. GMT+1) is displayed'));

		o = s.taboption('general', form.ListValue, 'clock_hourcycle', _('Time Format'));
		o.value('h12', _('12-Hour Clock'));
		o.value('', _('24-Hour Clock (Default)'));
		o.optional = true;
```

The behaviour here is that `uci.set` internally treats .set( '' ) as 'remove this value' so `clock_hourcycle` disappears and acquires the fallback value at the ` || `.

AndyChiang888:
I designed the default to use the full timezone name, you mean the default to use the time zone offset right?
As for the time format, I need to think about how to deal with it.
After all, browsers can now use hourCycle: undefined

systemcrash:
> I designed the default to use the full timezone name, you mean the default to use the time zone offset right? As for the time format, I need to think about how to deal with it. After all, browsers can now use hourCycle: undefined

`Unchecked means the timezone offset (E.g. GMT+1) is displayed`

So I'm aiming for that. I just had the default thing the wrong way around :/

Browsers may not intentionally change behaviours any more, but how we handle time certainly will change. 

But hour cycle default: 24 is basically how it always was.

AndyChiang888:
In environments such as en-US, hourCycle is not provided or is set to hourCycle: undefined, which uses a 12-hour clock.
If want to avoid the use of undefined, then I will modify it so that hourCycle is not passed to Intl.DateTimeFormat when clock_hourcycle is the default value.

AndyChiang888:
@systemcrash 
How about this one now? If you are worried about the browser changing its behavior, I can also change it to not pass it to Intl.DateTimeFormat when hc is 0

AndyChiang888:
Now default to using timezone offset and the default time format follows the locales standard.

systemcrash:
OK. This is good.

PR #8028, opened 2025-10-22
luci-app: fix translation strings

AndyChiang888:
The current handling method introduces extraneous spaces into the translations.
For instance, 'System Log', which is correctly translated to Japanese as 'システムログ' (without extra spaces), is being displayed as 'システム ログ' under the current approach.
Other CJK/Southeast Asian languages are experiencing similar issues, including Chinese, Thai, and Lao.
Therefore, these elements should be translated together as a single unit, rather than separately.


systemcrash:
Merged. Thanks @AndyChiang888

PR #8030, opened 2025-10-22
luci-base: add ability to hide section titles

Alphix:
The rationale here is that tabbed CBIMaps were introduced in commit 082fd9ff10b.

With tabbed maps, code could typically look like this:
```js
	m = new form.Map('foobar', _('FooBar'));
	m.tabbed = true;
	s = m.section(form.TypedSection, 'foo', _('foo Settings'));
```
The problem is that the title of `s` will be used as the name of the tab rendered in `m`, but also rendered as an `<h3>` right below the tab. IOW, the same information will be presented twice, which looks weird.

Doing this instead...
```js
	m = new form.Map('foobar', _('FooBar'));
	m.tabbed = true;
	s = m.section(form.TypedSection, 'foo');
```
...means that the superfluous `<h3>` won't be rendered (since `s` has no title), but the tab will then simply have the name of the section ("foo"), which can't be translated (bad).

After this change, the tabbed map can be written like this:
```js
	m = new form.Map('foobar', _('FooBar'));
	m.tabbed = true;
	s = m.section(form.TypedSection, 'foo', _('foo Settings'));
	s.hidetitle = true;
```
Which will give the Map tab the name "foo Settings", but won't add a title for the TypedSection right under the tab.

- [x] Tested on: (NanoPi R6S, master, Chrome)

Example screenshot showing the issue (see "Leases" tab and "Leases" header):
Screenshot From 2025-10-22 02-30-12

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2156,6 +2156,17 @@ const CBITypedSection = CBIAbstractSection.extend(/** @lends LuCI.form.TypedSect
 	 */
 
 	/**
+	 * If set to true, the title caption of the form section element which
+	 * is normally rendered before the start of the section content will
+	 * not be rendered in the UI. The default is false, meaning that the
+	 * title is rendered.
+	 *
+	 * @name LuCI.form.TypedSection.prototype#hidetitle
+	 * @type boolean
+	 * @default false
+	 */
+
+	/**
 	 * If set to `true`, mapped section instances are treated as anonymous
 	 * UCI sections, which means that section instance elements will be
 	 * rendered without a title element and that no name is required when adding
@@ -2300,7 +2311,7 @@ const CBITypedSection = CBIAbstractSection.extend(/** @lends LuCI.form.TypedSect
 			'data-tab-title': (this.map.tabbed && !this.parentoption) ? this.title || this.sectiontype : null
 		});
 
-		if (this.title != null && this.title != '')
+		if (this.title != null && this.title != '' && !this.hidetitle)
 			sectionEl.appendChild(E('h3', {}, this.title));
 
 		if (this.description != null && this.description != '')
@@ -2530,7 +2541,7 @@ const CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection
 			'class': 'table cbi-section-table'
 		});
 
-		if (this.title != null && this.title != '')
+		if (this.title != null && this.title != '' && !this.hidetitle)
 			sectionEl.appendChild(E('h3', {}, this.title));
 
 		if (this.description != null && this.description != '')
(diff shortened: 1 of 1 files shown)


Alphix:
@systemcrash ping :)

systemcrash:
Looks OK. Control over this aspect is desirable. Any word from @hnyman ?

systemcrash:
Would it be better named as `hidesectiontitle`?

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -2155,6 +2155,15 @@ const CBITypedSection = CBIAbstractSection.extend(/** @lends LuCI.form.TypedSect
 	 * @default false
 	 */
 
+	/**
+	 * If set to `true`, the title caption of the form section element will not
+	 * be rendered in the UI. The default is `false`.
```
systemcrash: Perhaps something more like:

If set to `true`, the title caption of the form section element which is normally rendered between the tab row and the start of the content will not be rendered in the UI. The default is `false`, meaning that the title is rendered.

Alphix: But a tab row isn't a given, that depends on other things?

systemcrash: It's not a given, but we should give an indication of the effect it has so toggling it is either easy to see/find or why one might want to toggle this flag. I was trying not to refer to `<h3>` which might change one day, but I guess that's another way of referring to it.

Alphix:
> Would it be better named as `hidesectiontitle`?

I was thinking that the current wording would be more logical if one uses a variable like `section` or (more commonly `s`)...then it already reads like `section.hidetitle` (i.e. `s.hidetitle`)...and your suggestion would be `section.hidesectiontitle`.

But I'm not that committed to the current wording 🤷🏻‍♂️

systemcrash:
I like that it's short. But the hidetitle doesn't elude to which title until we read about how the flag works. I guess hidetitle is fine then.

modules/luci-base/htdocs/luci-static/resources/form.js
Alphix: I should add that "the title caption of the form section element" is taken from the description of `.title`...I was hoping that the current comment would convey the message that it's `.title` that won't appear in the UI...

How about:
```
If set to true, the title caption of the form section element which is
normally rendered before the start of the section content will not be
rendered in the UI. The default is false, meaning that the title is
rendered.
```

systemcrash: when you're ready

Alphix:
@systemcrash : PR updated with suggested doc text

modules/luci-base/htdocs/luci-static/resources/form.js
Alphix: Ready when you are, Sergeant Pembry.

systemcrash:
Merged. Thanks @Alphix !

PR #8034, opened 2025-10-25
luci-app-attendedsysupgrade: use the same logic as the main build when requesting rebuilds

Guerra24:
This PR makes rebuild request use the same logic as the main build, fixes an issue with rebuilds requests being counted as unique on the ASU server.

- [x] Tested on: (armv8, r31547-01e1c86f68, Firefox) :white_check_mark:
- [x] \( Preferred ) Mention: @aparcar 
- [x] \( Optional ) Closes: openwrt/asu/issues/1526

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -75,7 +75,7 @@ return view.extend({
 		unpack_imagebuilder:     [ 40, _('Setting Up ImageBuilder')],
 	},
 
-	request_hash: '',
+	request_hash: new Map(),
 	sha256_unsigned: '',
 
 	applyPackageChanges: async function(package_info) {
@@ -224,9 +224,6 @@ return view.extend({
 	},
 
 	handle202: function (response) {
-		response = response.json();
-		this.request_hash = response.request_hash;
-
 		if ('queue_position' in response) {
 			ui.showModal(_('Queued...'), [
 				E(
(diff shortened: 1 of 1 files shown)


systemcrash:
@efahl

efahl:
I'll wait until morning when my brain restart to do the code review, but.

The commit message should probably have something in the body about what we're fixing "The rebuilder loop was doing repeated build requests via POST etc..." and then a brief synopsis of how it's being fixed.

Also probably should add a "Fixes" tag for context to the commit body:

Fixes: https://github.com/openwrt/asu/issues/1526

efahl (changes requested):
Take a look in `handleError` line 255, I think that needs to be server-specific, 

`... this.request_hash,`

Looks like you'll need to pass `server` to that method, too, which should be safe as I don't see `handleError` being called anywhere else, there's just that one call from `handleRequest`.

Or maybe just pass the `request_hash` since we already know it at the point of the call?  That's probably better...
```
-	handleError: function (response, data, firmware) {
+	handleError: function (response, data, firmware, request_hash) {
		response = response.json();
		const request_data = {
			...data,
-			request_hash: this.request_hash,
-			request_hash: request_hash,
```
with corresponding change up at the call site.

Guerra24:
Done, went for the second option.

efahl:
Tested on 24.10.3 x86/64.

Looks good to me, let's ship it!

@systemcrash This needs backport to 24.10, do you want to do it, or should @Guerra24 submit a cherrypick PR on the openwrt-24.10 branch?

systemcrash:
@Guerra24 please fix your github config so that your SoB matches :)

Guerra24:
@systemcrash Done, I think...

systemcrash:
Merged. Thanks @Guerra24 ! @efahl picked to 24.10

PR #8043, opened 2025-10-28
luci-mod-network: move max_{preferred,valid}_lifetime to IPv6-RA tab

1715173329:
- [x] Tested on: (aarch64, snapshot, chrome) :white_check_mark:
- [x] Description:
   These options are RA configuration so they should be put under the IPv6-RA tab instead of the "generic" IPv6 tab.
   Also make them depend on RA server/hybrid mode as these values are pointless without RA server enabled.


systemcrash:
Thanks @1715173329 !

PR #8045, opened 2025-10-28
luci-app-attendedsysupgrade: improve ux during install and reboot

efahl:
During the sysupgrade image installation and reboot process, it is possible for LuCI to reconnect before the device is rebooted. This leads to confusion and possible danger of changes being made that will damage the device.

We set expectations as to what will happen during the installation process with better wording in the spin dialog, and also "lock up" the interface for 10s, allowing sysupgrade to do its job and avoid reconnecting too early

Link: https://forum.openwrt.org/t/luci-attended-sysupgrade-support-thread/230552/268

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -476,27 +476,26 @@ return view.extend({
 					.then((response) => {
 						if (response.sha256sum != sha256) {
 							ui.showModal(_('Wrong checksum'), [
-								E(
-									'p',
-									_('Error during download of firmware. Please try again')
-								),
+								E('p', _('Error during download of firmware. Please try again')),
 								E('div', { class: 'btn', click: ui.hideModal }, _('Close')),
 							]);
 						} else {
 							ui.showModal(_('Installing...'), [
-								E(
-									'p',
-									{ class: 'spinning' },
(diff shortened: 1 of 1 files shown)


applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -476,27 +476,25 @@ return view.extend({
...
-								E(
-									'p',
-									{ class: 'spinning' },
-									_('Installing the sysupgrade. Do not unpower device!')
-								),
+								E('div', { class: 'spinning' }, [
+									E('p', _('Installing the sysupgrade image...')),
+									E('p', _(
+									'Once the image is written, the system will reboot. ' +
+									'This should take at least a minute, so please wait for the login screen.'
```
systemcrash: Could you make these two i18n strings with a `+ ' ' +` in between?

efahl:
Just for completeness:

0111a1c785b3b4491cd34aa51290dbb99f027da7_2_1035x225

That spinning wheel on the left looks a little tight against the text (didn't look in the bootstrap theme, just the 2020)...

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
efahl: Should be good now?

systemcrash:
Thanks @efahl !

PR #8047, opened 2025-10-29
luci-app-attendedsysupgrade: do major cleanup of error handler

efahl:
Extract and isolate the forum-specific constants and formats to a location that is easily modified.  Factor out a lot of common code and simply make everything much cleaner and more readable.

Output to the "Request Data" block was augmented to contain the system board info, and was reordered to put more important data earlier in the output.

Also fix a potential bug in construction of the error html, where the final step was coercing everything to a string prematurely.

---
Fix my embarrassing code from a couple days ago.  I couldn't sleep with it in such a state.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -9,6 +9,32 @@
 'require dom';
 'require fs';
 
+
+/* --------------------------------------------------------------------------
+ * This section should be sufficient to isolate the changes that any forked
+ * versions need to change with a custom support url.
+ */
+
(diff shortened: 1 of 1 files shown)


applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -256,95 +282,69 @@ return view.extend({
...
+		const request_str = JSON.stringify(request_data, null, 4);
 		let body = [
 			E('p', {}, [
-				_('First check '),
-				E(
-					'a',
-					{ href: 'https://forum.openwrt.org/t/luci-attended-sysupgrade-support-thread/230552' },
-					_('this forum thread')
-				),
+				_('First check'), ' ',
```
systemcrash: In English one generally says "First, check..." otherwise "First check" sounds like a noun, as in "first aid".

PR #8049, opened 2025-10-29
luci-proto-wireguard: add IPv6 prefix field

nagual2:
Add support for configuring IPv6 prefix delegation in WireGuard protocol interface.

This patch adds an `ip6prefix` input field to the WireGuard protocol configuration UI, positioned directly below the existing `addresses` field.

### Changes:
- Add `ip6prefix` field using `form.DynamicList` to support


systemcrash:
Neither commit will be accepted in its present form which changes all tabs to spaces. Also fix commit messages and sign-off

PR #8050, opened 2025-10-29
luci-proto-wireguard: add ip6prefix field to protocol configuration UI

nagual2:
Add support for configuring IPv6 prefix delegation in WireGuard protocol interface.

This patch adds an `ip6prefix` input field to the WireGuard protocol configuration UI, positioned directly below the existing `addresses` field.

### Changes:
- Add `ip6prefix` field using `form.DynamicList` to support


systemcrash:
Same problems with the commit in this PR. Please don't open new PRs - fix the problems via the existing PRs.

PR #8052, opened 2025-10-29
luci-app-acme: Use fullchain.crt file to read issueDate

sunnyqeen:
The private key is used for applying the cert
And the cert should be used for reading issueDate

The private key could be reused, so the issueDate showing on the UI is wrong, when next renew hapening

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
@@ -12,7 +12,7 @@ return view.extend({
 			L.resolveDefault(fs.list('/etc/ssl/acme/'), []).then(files => {
 				let certs = [];
 				for (let f of files) {
-					if (f.type == 'file' && f.name.match(/\.key$/)) {
+					if (f.type == 'file' && f.name.match(/\.fullchain\.crt$/)) {
 						certs.push(f);
 					}
 				}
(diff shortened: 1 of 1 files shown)


systemcrash:
@dibdot ? Seems right.

dibdot:
I think @stokito @tohojo should review such changes, on first sight I would replace ...

```
let domain = cert.name.substring(0, cert.name.length - 14);
```
... with ...
```
let domain = cert.name.replace(/\.fullchain\.crt$/, '');
```
... but I've never used that module.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -521,7 +521,7 @@ function _renderCerts(certs) {
 	]);
 
 	let rows = certs.map(function (cert) {
-		let domain = cert.name.substring(0, cert.name.length - 4);
+		let domain = cert.name.replace(/\.fullchain\.crt$/, '');
```
systemcrash, line 524: Is it a fact it will always be called fullchain? How about
```js
		let domain = cert.name.split('.')[0];
```

sunnyqeen: the file name will be like www.yyy.xxxx.com.fullchain.crt
only taking the first part will not work correctly.

PR #8060, opened 2025-11-01
luci-theme-material: fixes long-named items in the main menu

astra1:
This change truncates long item names in the navigation menu.
@systemcrash

lucibefore

luciafter

- [x] Tested on: (x86, openwrt 24.10, firefox) :white_check_mark:

### DIFF
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -343,7 +343,6 @@ small {
 	float: left;
 	overflow-x: auto;
 	width: calc(0% + 15rem);
-	min-width: max-content;
 	height: calc(100% - 4rem);
 	background-color: var(--menu-bg-color);
 	transition: visibility 400ms, width 400ms;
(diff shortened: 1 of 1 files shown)


hnyman:
This fix might be sensible, but it also shows the underlying problem: the overly long menu item itself. 
The menu item should be shortened.  Where is that actually coming from? Not from a normal OpenWrt package, I think.

"Information about 3G/4G/5G connection"
-->
"3G/4G/5G connection status"
"3G/4G/5G connection info"

astra1:
> This fix might be sensible, but it also shows the underlying problem: the overly long menu item itself. The menu item should be shortened. Where is that actually coming from? Not from a normal OpenWrt package, I think.
> 
> "Information about 3G/4G/5G connection" --> "3G/4G/5G connection status" "3G/4G/5G connection info"

Yes, from the 
[luci-app-modemdata](https://github.com/4IceG/luci-app-modemdata) package, and it is possible to request the maintainer (@4IceG) to shorten the menu item name, but this will also be helpful for internationalization, as the translation may be even longer than the English name.

4IceG:
No problem, in new version of the package there will be "Modemdata info" in menu, I think it will be the shortest version.

systemcrash:
I dislike imposing hard limits on layout space - the layout is meant to be flexible to accommodate what goes there. The side bar in that theme is hideable in any case. But addressing the actual title is an improvement.

PR #8062, opened 2025-11-02
luci-mod-network: Add warning for required package in mesh mode in wireless.js 

GithPK:
Some devices (e.g. Arcadyan AW1000 / Qualcomm IPQ807x platforms) expose the “802.11s (Mesh)” mode in LuCI even when only wpad-basic-mbedtls is installed.

However, hostapd from wpad-basic-mbedtls does *not* provide mesh (802.11s) support. Selecting Mesh mode in the UI silently fails, resulting in:
 - mesh interface showing channel “36 (0.000 GHz)”
 - no SSID beaconing
 - hostapd / wpa_supplicant reload loops
 - netifd reporting "command failed: Not supported (-95)"

Installing full `wpad-mbedtls` resolves the issue immediately.

This commit adds a clear warning message in LuCI that 802.11s Mesh requires the full `wpad-mbedtls` package, helping users avoid confusing failures, especially on devices where mesh previously worked or where the UI hides the complexity. The warning appears only when the user selects "802.11s" mode.

No functional changes for any other modes.

More description in:
https://github.com/openwrt/openwrt/issues/20391

- :white_check_mark: This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- :white_check_mark: Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- :white_check_mark: Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- :white_check_mark: Tested on: (AW1000, ARMv8 Processor rev 4, OpenWrt SNAPSHOT r31645-3b21f97641, EDGE) :white_check_mark:
- :white_check_mark: Closes: openwrt/luci issue #20391
- :white_check_mark: Description: (describe the changes proposed in this PR)

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1037,10 +1037,10 @@ return view.extend({
 				o.value('sta', _('Client'));
 				o.value('adhoc', _('Ad-Hoc'));
 
-				let warn = ss.taboption('general', form.Value, '_mesh_warning', _('Warning:'));
+				let warn = ss.taboption('general', form.Value, '_mesh_warning', _(' Warning:'));
 				warn.depends('mode', 'mesh');
 				warn.readonly = true;
-				warn.default = _('Required package: wpad-mbedtls ');
+				warn.default = _(' Required package: wpad-mbedtls');
 				warn.rmempty = false;
 				
 				o = ss.taboption('general', form.Value, 'mesh_id', _('Mesh Id'));


castillofrancodamian:
This applies to all devices. There's also the `wpad-mesh-mbedtls` package, which is for 802.11s.

GithPK:
> castillofrancodamian

Thank you.
So, I assume, that some people could have the same problem, struggling with setting the 802.11s, and wondering why it does not work in spite of setting everything correctly in the LuCI.
That is why I decided to start the change in that matter.

wryun:
It would be nice be to add mesh as one of the tested wpa_supplicant features in rpcd/ucode/luci. You could then gate this option on whether the feature existed. Or (IMO more cleanly, but more confusingly) just not even provide the Mesh Point mode if it's not available.

systemcrash:
> It would be nice be to add mesh as one of the tested wpa_supplicant features in rpcd/ucode/luci. You could then gate this option on whether the feature existed. Or (IMO more cleanly, but more confusingly) just not even provide the Mesh Point mode if it's not available.

This is the current approach everywhere else - detecting features and showing them based on availability. 

In a few other cases, I opted to display that a certain kmod should be installed for some functionality. ( See e.g. bonding )

https://github.com/openwrt/luci/blob/12995312420bbe32d7327beac30e8b52243732b4/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js#L453-L462

feature detection is done in the ucode `luci` in luci-base.

GithPK:
I totally agree with you that checking an ability of the system before letting the user switching to any feature should be a standard behaviour of the GUI.
It is so obvious that it should work this way. So, the question is why it is not implemented in LuCI yet?
Anyway, it is better to make a small first step and motivate somebody who is experienced in .js and could implement it in LuCI.
Do you suggest to cancel this PR and wait untill anybody touch it?

systemcrash:
Closed by 13f9f6f010decfd0016beb659978b4b8d4895a1c

GithPK:
Thank you @systemcrash

GithPK:
@systemcrash 
Hello Paul, I have just upgraded LuCI packages only (apk only, not a whole image) and the mode 802.11s dissapeared from the menu, even though I have wpad-mbedtls installed and the mode has worked recently. 
Could you check it, please?

systemcrash:
Please run: 
```sh
wpa_supplicant -vmesh
echo $?
```
and
```sh
/usr/sbin/hostapd -vmesh
echo $?
```

GithPK:
@systemcrash  

Both return "0"

```
root@owrt:~# /usr/sbin/hostapd -vmesh
root@owrt:~# echo $?
0
root@owrt:~# wpa_supplicant -vmesh
root@owrt:~# echo $?
0
```

systemcrash:
Run in your browser console when the page is loaded:
```
L.hasSystemFeature('hostapd', 'mesh');
L.hasSystemFeature('wpasupplicant', 'mesh');
```

If either fails, I think you need a newer luci-base (which does those checks).

GithPK:
Thank you @systemcrash. 
Some updates have helped. It works fine now.

GithPK:
@systemcrash 
Paul,
Could we utilize your expertise and go another step ahead with this menu issue? I have an idea how to make it even more user friendly.

Issue #7958, opened 2025-09-25
luci-app-tinyproxy:valid domain name reported as invalid


andvalb:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### screenshots or captures

<img width="577" height="181" alt="Image" src="https://github.com/user-attachments/assets/e08fafb6-878b-4889-abe3-a1ff5688f61b" />

### Actual behaviour

Valid and working domain name reported as invalid.
Also help text under the option "Target host" says that "Can be either an IP address or range, a domain name or '.' for any host without domain", but domain name '.' also reported as invalid.

### Expected behaviour

Domain name like '.onion' or .'i2p' are valid and must be reported as valid for this filter option.
Do not use address format validation for them. This is not a url(adddress) but filter.

### Steps to reproduce

go to : Services->Tinyproxy and try to add '.' or '.i2p' or '.onion' to the field  "Target host" of "Upstream proxy" option.

### Additional Information

```shell
{
        "kernel": "6.6.104",
        "hostname": "router",
        "system": "ARMv8 Processor rev 0",
        "model": "Bananapi BPI-R4",
        "board_name": "bananapi,bpi-r4",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "24.10.3",
                "revision": "r28872-daca7c049b",
                "target": "mediatek/filogic",
                "description": "OpenWrt 24.10.3 r28872-daca7c049b",
                "builddate": "1758316778"
        }
}
```

### What browsers do you see the problem on?

Microsoft Edge

### Relevant log output

```shell

```

PR #8071, opened 2025-11-07
luci-app-tinyproxy: Loosen constraint on upstream target

cukiernick:
- [x] Tested on: (ARMv8, OpenWrt 24.10.4, Firefox 144) :white_check_mark:
- [x] \( Optional ) Closes: openwrt/luci#7958

---

Upstream target datatype is too strict, rejecting values starting with a dot, which selects all subdomains of provided domain.

From tinyproxy [docs](https://tinyproxy.github.io/):

> The site can be specified in various forms as a hostname, domain name or as an IP range:
> 
>     name matches host exactly
>     .name matches any host in domain "name"
>     . matches any host with no domain (in 'empty' domain)
>     IP/bits matches network/mask
>     IP/mask matches network/mask

I looked at available datatypes in `modules/luci-compat/luasrc/cbi/datatypes.lua` and `modules/luci-base/htdocs/luci-static/resources/validation.js` and there is no suitable predefined type, so I changed it to `string`

### DIFF
--- a/applications/luci-app-tinyproxy/htdocs/luci-static/resources/view/tinyproxy/tinyproxy.js
+++ b/applications/luci-app-tinyproxy/htdocs/luci-static/resources/view/tinyproxy/tinyproxy.js
@@ -299,7 +299,7 @@ return view.extend({
 
 		ta.rmempty = true;
 		ta.placeholder = '0.0.0.0/0';
-		ta.datatype = 'host(1)';
+		ta.datatype = 'string';
 
 
 		v = s.option(form.Value, 'via', _('Via proxy'),


applications/luci-app-tinyproxy/htdocs/luci-static/resources/view/tinyproxy/tinyproxy.js
```diff
@@ -299,7 +299,7 @@ return view.extend({
 
 		ta.rmempty = true;
 		ta.placeholder = '0.0.0.0/0';
-		ta.datatype = 'host(1)';
+		ta.datatype = 'string';
```
systemcrash, line 302: Use "or". e.g. `o.datatype = 'or(ip4addr("nomask"),ip6addr("nomask"))';` and `or(ipmask("true"),iprange)`. You'll need to customise.

systemcrash: ping @cukiernick

systemcrash: ping

PR #8077, opened 2025-11-08
luci-mod-network: Adds warning for required package in mesh mode in wireless.js

GithPK:
<luci-mod-network>: Adds warning for required package in mesh mode in wireless.js 
Devices (e.g. Arcadyan AW1000 / Qualcomm IPQ807x platforms) expose the “802.11s (Mesh)” mode in LuCI even when only wpad-basic-mbedtls is installed.

However, hostapd from wpad-basic-mbedtls does *not* provide mesh (802.11s) support. Selecting Mesh mode in the UI silently fails, resulting in:
 - mesh interface showing channel “36 (0.000 GHz)”
 - no SSID beaconing
 - hostapd / wpa_supplicant reload loops
 - netifd reporting "command failed: Not supported (-95)"

Installing e.g. full `wpad-mbedtls` resolves the issue immediately.

This commit adds a clear warning message in LuCI that 802.11s Mesh requires the full `wpad-mbedtls` package, helping users avoid confusing failures, especially on devices where mesh previously worked or where the UI hides the complexity. The warning appears only when the user selects "802.11s" mode.

No functional changes for any other modes.

More description in:
https://github.com/openwrt/openwrt/issues/20391

Signed-off-by: Piotr Kołtun <pkoltun@op.pl>

- [X] Each commit has a valid :black_nib: `Signed-off-by: Piotr Kołtun <pkoltun@op.pl>` row (via `git commit --signoff`)
- [X] Each commit and PR title has a valid :memo: `<luci-mod-network>: Add warning for required package in mesh mode in wireless.js` first line subject for packages
- [X] Tested on: ARMv8 Processor rev 4, ARMv8 Processor rev 4, Edge :white_check_mark:
- [X] Description: Adds a clear warning message in LuCI that 802.11s mesh requires
the full wpad-mbedtls package or similar including mesh.

### DIFF
--- a/collections/luci-nginx/Makefile
+++ b/collections/luci-nginx/Makefile
@@ -15,6 +15,7 @@ LUCI_DEPENDS:= \
 	+IPV6:luci-proto-ipv6 \
 	+luci-app-firewall \
 	+luci-app-package-manager \
+	+luci-app-attendedsysupgrade \
 	+luci-mod-admin-full \
 	+luci-proto-ppp \
 	+luci-theme-bootstrap \
--- a/collections/luci-ssl-openssl/Makefile
+++ b/collections/luci-ssl-openssl/Makefile
@@ -17,7 +17,8 @@ LUCI_DESCRIPTION:=LuCI with OpenSSL as the SSL backend (libustream-openssl). \
 LUCI_DEPENDS:=+luci-light \
 	+libustream-openssl \
 	+openssl-util \
-	+luci-app-package-manager
+	+luci-app-package-manager \
+	+luci-app-attendedsysupgrade
 
 PKG_LICENSE:=Apache-2.0
 
--- a/collections/luci-ssl/Makefile
+++ b/collections/luci-ssl/Makefile
@@ -13,7 +13,8 @@ LUCI_TITLE:=LuCI with HTTPS support (mbedtls as SSL backend)
 LUCI_DEPENDS:=+luci-light \
 	+libustream-mbedtls \
 	+px5g-mbedtls \
-	+luci-app-package-manager
+	+luci-app-package-manager \
+	+luci-app-attendedsysupgrade
 
 PKG_LICENSE:=Apache-2.0
 
--- a/collections/luci/Makefile
+++ b/collections/luci/Makefile
@@ -13,7 +13,8 @@ LUCI_TITLE:=LuCI interface with Uhttpd as Webserver (default)
 LUCI_DESCRIPTION:=Standard OpenWrt set including package management and attended sysupgrades support
 LUCI_DEPENDS:= \
 	+luci-light \
-	+luci-app-package-manager
+	+luci-app-package-manager \
+	+luci-app-attendedsysupgrade
 
 PKG_LICENSE:=Apache-2.0
 


systemcrash:
I don't know what you're doing but please stop

GithPK:
> I don't know what you're doing but please stop

I know how it may look like but I am struggling with the test formalities as you can see. 
Trying to find the right way. 
Sorry, if it makes you nervous.

systemcrash:
Spend some time to figure out how to amend commits and force push. Ensure your github email config agrees with your SoB line.

hnyman:
Note that mbedtls is not the only SSL library supported. Also OpenSSL and wolfssl have their own wpad variants... Your warning message ignores that.

GithPK:
> Spend some time to figure out how to amend commits and force push. Ensure your github email config agrees with your SoB line. 

I am just spending time to find out out how to use the web GitHub interface and put proper form of the subject... e.g. no brackets
Email address is not a problem.

GithPK:
> Note that mbedtls is not the only SSL library supported. Also OpenSSL and wolfssl have their own wpad variants... Your warning message ignores that.
> 
Thank you for the comment. You are right. Let us find a more general message, and I will be happy to change it. Do you have anything on your mind?

hnyman:
I am not quite sure, why this would be needed, or if your change would be right. 

The original logic in the file has the intention that the mesh mode is not available in the drop-down list in GUI, if 'have_mesh' is not true.  There is clear logic to evaluate that from the compiled-in features in hostapd and/or wpa-supplicant

`const have_mesh = L.hasSystemFeature('hostapd', 'mesh') || L.hasSystemFeature('wpasupplicant', 'mesh');`

To me it looks like your change would actually expose the unapplicable mode item with explanation, while originally the mode is hidden if there is no support for it. 

One scenario where users might run into the problem is that they already have mesh in their /etc/config/wifi, and then the sysupgrade to a build without mesh support. In that scenario the help text might be useful.
Is that your target usage?  

Ps.
The applicable text might be something like 
`"The mesh / 802.11s functionality requires the full wpad or wpa-supplicant variant instead of the default 'basic' variant."`

GithPK:
> I am not quite sure, why this would be needed, or if your change would be right.
> 
> The original logic in the file has the intention that the mesh mode is not available in the drop-down list in GUI, if 'have_mesh' is not true. There is clear logic to evaluate that from the compiled-in features in hostapd and/or wpa-supplicant
> 
> `const have_mesh = L.hasSystemFeature('hostapd', 'mesh') || L.hasSystemFeature('wpasupplicant', 'mesh');`
> 
> To me it looks like your change would actually expose the unapplicable mode item with explanation, while originally the mode is hidden if there is no support for it.
> 
> One scenario where users might run into the problem is that they already have mesh in their /etc/config/wifi, and then the sysupgrade to a build without mesh support. In that scenario the help text might be useful. Is that your target usage?
> 
> Ps. The applicable text might be something like `"The mesh / 802.11s functionality requires the full wpad or wpa-supplicant variant instead of the default 'basic' variant."`

The reason I have started working on it is mentioned in the issue: https://github.com/openwrt/openwrt/issues/20391
I had been using the 802.11s until I upgraded my system. I did not know why after selecting the 802.11s from a menu the system did  not work. I did not care if it was basic or not. It used to work and then stopped without any given reason.
Why the pull request make sense:
Case No. 1 - in my case I had no info what happened and the warning would have given me any hint if I had seen it.
Case No.2 - If we just  hide the option (which is already done by Paul), somebody (less experinced like e.g. me) will be looking for it but will not find it quickly without asking around
Case No.3 - The 802.11s is in the menu in accordance to the qualcomm driver availble features (sometimes nice to know what can be done looking arround the LuCI menu) if we give info in the menu (not available) and the warning message it will be clear where to start or what to do, at the first sight.

In my opinion, LuCI is for less experienced users who would like to easily set up some functionalities and they should be informed what is possible and needed to do it, and not be cheated by the menu. Therefore, in my opinion, the more hints and warnings the better.

GithPK:
> Ps. The applicable text might be something like `"The mesh / 802.11s functionality requires the full wpad or wpa-supplicant variant instead of the default 'basic' variant."`

Thank you for the proposal.
What do you think about the below message?
"Mesh mode (802.11s) requires a wpad package with SAE support (e.g. wpad-openssl, wpad-wolfssl, or wpad-mbedtls). Please ensure that a compatible package is installed."

GithPK:
I have a problem with the test formalities check.
```
: Commit subject line MUST start with '<package name>: ' (<luci-mod-network>: Add warning for required package in mesh mode in wireless.js)
```
Why the system does not accept neither this:

```<luci-mod-network>: Add warning for required package in mesh mode in wireless.js```
nor that:
```luci-mod-network: Add warning for required package in mesh mode in wireless.js```

**Comment by systemcrash** [comment]
I don't like how big a change this PR makes. Try something like this around line 1040-1050, just after `o = ss.taboption('general', form.ListValue, 'mode', _('Mode'));
`:

```js

				if (!have_mesh) {
					o = ss.taboption('general', form.DummyValue, '_dummy_mesh', _('802.11s unavailable'),
						_('802.11s Mesh mode requires wpad with mesh support (e.g. wpad-mesh or wpad-full).'));
					o.rawhtml = true;
					o.default = '<a id="installmesh" href="%s" target="_blank" rel="noreferrer">%s</a>'
						.format(L.url('admin/system/package-manager') + '?query=wpad-mesh', _('Install mesh wpad'));
				}
```

Or to make it even more discrete so as not to annoy users who don't care about it - ignore the above and change only 'mode' to:
```js

				o = ss.taboption('general', form.ListValue, 'mode', _('Mode') , !have_mesh ? '<a id="installmesh" href="%s" target="_blank" rel="noreferrer">%s</a>'
						.format(L.url('admin/system/package-manager') + '?query=wpad-mesh', _('Install mesh wpad') ) : '');
```

systemcrash:
> `luci-mod-network: Add warning for required package in mesh mode in wireless.js`

This is the only accepted format. `<>` are not literal.

systemcrash:
Make `Signed-off-by: Piotr Kołtun <pkoltun@op.pl>` your Sob. Update github config. Squash all your commits, and amend the last one to fix the subject line. Don't try to do this in any github app, I have no idea whether that works.

GithPK:
> I don't like how big a change this PR makes. Try something like this around line 1040-1050, just after `o = ss.taboption('general', form.ListValue, 'mode', _('Mode')); `:
> 
> Or to make it even more discrete so as not to annoy users who don't care about it - ignore the above and change only 'mode' to:
> 
> ```js
> 				o = ss.taboption('general', form.ListValue, 'mode', _('Mode') , !have_mesh ? '<a id="installmesh" href="%s" target="_blank" rel="noreferrer">%s</a>'
> 						.format(L.url('admin/system/package-manager') + '?query=wpad-mesh', _('Install mesh wpad') ) : '');
> ```

It looks good and leads to the final solution. Thank you.
I will try to fight a little bit with the GitHub yet...

GithPK:
> Make `Signed-off-by: Piotr Kołtun <pkoltun@op.pl>` your Sob. Update github config. Squash all your commits, and amend the last one to fix the subject line. Don't try to do this in any github app, I have no idea whether that works.

I could not squash the commits unfortunately.
I have tried to run another PR and the test formalities went well. But...
I have stopped on the missing tag or image:
Error: buildx failed with: ERROR: failed to build: failed to solve: ghcr.io/openwrt/sdk:x86_64-pk: failed to resolve source metadata for ghcr.io/openwrt/sdk:x86_64-pk: ghcr.io/openwrt/sdk:x86_64-pk: not found

GithPK:
@systemcrash why does it look for the nonexistent image: x86_64-pk ?

systemcrash:
> > Make `Signed-off-by: Piotr Kołtun <pkoltun@op.pl>` your Sob. Update github config. Squash all your commits, and amend the last one to fix the subject line. Don't try to do this in any github app, I have no idea whether that works.
> 
> I could not squash the commits unfortunately. 

That's a you problem. It's up to you to learn how to use git :)

Every time you fail and open a new PR, we all get spammed, which becomes more and more likely to get ignored, or worse, blocked.

fork this repo, and go play with git and simulate making PRs there. 

What tool are you using? Github?

> I have tried to run another PR and the test formalities went well. But... I have stopped on the missing tag or image: Error: buildx failed with: ERROR: failed to build: failed to solve: ghcr.io/openwrt/sdk:x86_64-pk: failed to resolve source metadata for ghcr.io/openwrt/sdk:x86_64-pk: ghcr.io/openwrt/sdk:x86_64-pk: not found

Github weather.

GithPK:
> > > Make `Signed-off-by: Piotr Kołtun <pkoltun@op.pl>` your Sob. Update github config. Squash all your commits, and amend the last one to fix the subject line. Don't try to do this in any github app, I have no idea whether that works.
> > 
> > 
> > I could not squash the commits unfortunately.

I could not squash because the system indicated that it is not possible. I do not investigate further.
 
> That's a you problem. It's up to you to learn how to use git :)

I try to do my best...

> 
> Every time you fail and open a new PR, we all get spammed, which becomes more and more likely to get ignored, or worse, blocked.
> 
I am sorry, I cannot try and get it fixed before you get the email. It is not easy also for me. I cannot find out about any failure before get to know about it together. As you say  "Github weather."

I use the GitHub on the web (https://github.com).

jow-:
Apart from the technicalities, I am not sure if warnings about optional packages are actually wanted. Assume some firmware vendor or community firmware builder custom tailors an image deliberately leaving out functionality (e.g. to reduce space usage or because it is outside of the scope of the firmware), then the warning about missing packages might be undesired.

It is not sure whether the referenced package actually is installable (firmware might not contain a package manager) or available (vendor might not use OpenWrt repos and might not have built other wpad flavors). Even if the package would be available, the version might not be compatible or the remaining flash space too low.

So far the approach in LuCI usually has been using feature checks, not package install hints.

GithPK:
> Apart from the technicalities, I am not sure if warnings about optional packages are actually wanted. Assume some firmware vendor or community firmware builder custom tailors an image deliberately leaving out functionality (e.g. to reduce space usage or because it is outside of the scope of the firmware), then the warning about missing packages might be undesired.
> 
> It is not sure whether the referenced package actually is installable (firmware might not contain a package manager) or available (vendor might not use OpenWrt repos and might not have built other wpad flavors). Even if the package would be available, the version might not be compatible or the remaining flash space too low.
> 
> So far the approach in LuCI usually has been using feature checks, not package install hints.

Thank you for the feedback @jow-.
I hope you have seen the story from the beginning, I have not imagined that scenario myself and I think the solution @systemcrash has implemented is useful for many LuCI users.
The intention here is not to prescribe package installation, but to provide contextual information when a user selects a feature that cannot function without a specific dependency. Experienced developers or vendors may deliberately omit packages, but less experienced users often have no way of understanding why a visible option in LuCI fails silently. A neutral warning (“this feature requires package X, which may not be present in your firmware”) preserves the existing feature‑check approach while reducing confusion. It does not enforce installation, but clarifies the dependency chain for those who otherwise lack the implicit knowledge developers may take for granted.
The PR is closed. Thank you all for the patience and commitment.

PR #8079, opened 2025-11-08
luci-mod-network: wireless.js updates for cipher selection

### LINKED ISSUE #8070: luci-mod-network: cipher selection in WPA3-SAE?

### screenshots or captures

_No response_

### Actual behaviour

@systemcrash: I see in [https://github.com/openwrt/luci/commit/46e8c0b8f7f460901a34d6300abb0b31271c3029](url) GCMP was added in cipher options but your comment says it was not needed to be available for WPA3-SAE. Can you explain why that is not needed? As far as I know CCMP is still allowed in WPA3 next to the stronger GCMP. It would make sense that we should be able to force GCMP for WPA3? Or does hostapd not support forcing it to GCMP?

### Expected behaviour

-

### Steps to reproduce

-

### Additional Information

```shell
-
```

_No response_

### Relevant log output

```shell

```

Ramon00:
luci-mod-network: wireless.js updates for cipher selection. Allows selection to ciphers in WPA3. Can be committed by itself but preferably committed after netifd update (to be sure that default ciphers are set ok).

Depends on: https://github.com/openwrt/netifd/pull/60
Closes #8070

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1276,40 +1276,35 @@ return view.extend({
 					if (value == 'wpa' || value == 'wpa2' || value == 'wpa3' || value == 'wpa3-mixed' || value == 'wpa3-192')
 						uci.unset('wireless', section_id, 'key');
 
-					if (co.isActive(section_id) && e && (c == 'tkip' || c == 'ccmp' || c == 'ccmp256' || c == 'gcmp' || c == 'gcmp256' || c == 'tkip+ccmp'))
-						e += '+' + c;
-
+					if (co.isActive(section_id) && e) {
+						for (const  r of c) {
+							e += '+' + r;
+						}
+					}				
 					uci.set('wireless', section_id, 'encryption', e);
 				};
 
-				o = ss.taboption('encryption', form.ListValue, 'cipher', _('Cipher'));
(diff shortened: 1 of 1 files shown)


Ramon00:
see also [https://github.com/openwrt/luci/issues/8070](url)

Ramon00:
See also [https://github.com/openwrt/netifd/pull/60](https://github.com/openwrt/netifd/pull/60)

Ramon00:
So how do we push this along?

systemcrash:
Get some reviews from owrt members on the linked PR, I figure. Or get some tested-by, and someone who understands the downstream effects of that change.

Ramon00:
Ok let me make a forum post to see if some people want to test and review.

Neustradamus:
Forum link:
- https://forum.openwrt.org/t/security-enhancement-for-wpa3-looking-for-testers-of-prs/242883

menakite:
I'm not registered on the forum, but I've been testing your PRs for a few days now without any issues.
The hostapd configuration in `/var/run/hostapd-phy*.conf` is updated accordingly; in fact, clients that don't support GCMP won't connect when CCMP(-256) is not selected, and vice versa.

Ramon00:
Thank you for testing

> in fact, clients that don't support GCMP won't connect when CCMP(-256) is not selected, and vice versa

That is indeed exactly the intended behavior. 

The only real question I have is what do we want the default selected ciphers to be?

Ramon00:
Apparantly people do not care about wireless secutiry and peformance, hence closing this

PR #8083, opened 2025-11-09
luci-mod-status: Improve DUID/IAID handling for "Reserve IP" button

Alphix:
Note that `duid` list/option from UCI can be *either* `<DUID>` or `<DUID>%<IAID>`, so update the code to handle both cases.
    
Also, make sure the "Reserve IP" button is disabled if the `duid` is unknown (can only happen if dnsmasq is responsible for DHCPv6) and that DUIDs/IAIDs are always rendered in lowercase in the UI.

Finally, make sure the DUID gets stored as a list in UCI.

- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [ ] Incremented :up: any `PKG_VERSION` in the Makefile
- [x] Tested on: (NanoPi R6S, Master, Chrome) :white_check_mark:
- [ ] \( Preferred ) Mention: @ the original code author for feedback
- [ ] \( Preferred ) Screenshot or mp4 of changes:
- [x] \( Optional ) Closes: openwrt/luci#8056
- [ ] \( Optional ) Depends on: e.g. openwrt/packages#pr-number in sister repo
- [x] Description: (describe the changes proposed in this PR)

PR #6713, opened 2023-11-21
luci-mod-status: split dhcp status in dhcp and dhcpv6 includes

Ansuel:
Split dhcp status in separate dhcp and dhcpv6 includes to tidy code up and better organize them instead of defining multiple table and define an empty title for the include file.

PR #8084, opened 2025-11-09
luci-mod-status: use ubus for odhcpd leases

Alphix:
This is just a RFC for now (and it's based on top of https://github.com/openwrt/luci/pull/8083 rather than `master`).

Basically, it changes `luci-mod-status` so that it calls `odhcpd` directly over `ubus` to get the status of DHCPv4/DHCPv6 leases (yes, I know that `odhcpd` isn't responsible for DHCPv4 in a default install).

Before this change, all lease status is provided via the `rpcd-mod-luci` plugin in `rpcd`, which will dig through `leasefiles` to get the status of leases (and it'll still do that for `dnsmasq`-based leases). Communicating directly with `odhcpd` is quicker, simpler and provides access to information which wasn't exposed in the interface before (like per-lease interface information, force reconf status, etc).

This isn't complete right now (for example, it'll render 4 tables, showing both what was returned from `odhcpd` and what was returned from `rpcd`, but that's for comparison, also needs more testing).

However, I'm posting this here now so that I can get some early feedback. Ultimately I'd like to implement a similar change to `luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js`, after which the `odhcpd`-specific code in `rpcd-mod-luci` won't really be necessary anymore.

It might look like a lot of duplication, but there's enough little differences between the data returned from `odhcpd` and `rpcd` (and also the resulting HTML tables carry different information), that there's not much code that can be shared between the different table generation functions.

Down the line, I guess this would also make it easier to split `40_dhcp.js` into three separate files, one for `dnsmasq`, one for `odhcpd-DHCPv4` and one for `odhcpd-DHCPv6` (if that's desirable, cf. https://github.com/openwrt/luci/pull/6713).

- [x] Tested on: (NanoPi R6S, Master, Chrome) :white_check_mark:

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
@@ -39,47 +39,6 @@ return baseclass.extend({
 		return E([]);
 	},
 
-	handleCreateStaticLease(lease, ev) {
-		ev.currentTarget.classList.add('spinning');
-		ev.currentTarget.disabled = true;
-		ev.currentTarget.blur();
-
-		const cfg = uci.add('dhcp', 'host');
-		uci.set('dhcp', cfg, 'name', lease.hostname);
-		uci.set('dhcp', cfg, 'ip', lease.ipaddr);
-		uci.set('dhcp', cfg, 'mac', [lease.macaddr.toUpperCase()]);
-
-		return uci.save()
-			.then(L.bind(L.ui.changes.init, L.ui.changes))
-			.then(L.bind(L.ui.changes.displayChanges, L.ui.changes));
-	},
-
-	handleCreateStaticLease6(lease, ev) {
-		ev.currentTarget.classList.add('spinning');
-		ev.currentTarget.disabled = true;
-		ev.currentTarget.blur();
-
-		const cfg = uci.add('dhcp', 'host');
-		const ip6addr = lease.ip6addrs?.[0]?.replace(/\/128$/, '');
-		const ip6arr = ip6addr ? validation.parseIPv6(ip6addr) : null;
-
-		// Combine DUID and IAID if both available
-		// (note that we know that lease.duid is set here)
-		let duid_iaid = lease.duid.toUpperCase();
-		if (lease.iaid)
-			duid_iaid += `%${lease.iaid}`.toUpperCase();
-
-		uci.set('dhcp', cfg, 'name', lease.hostname);
-		uci.set('dhcp', cfg, 'duid', [duid_iaid]);
-		uci.set('dhcp', cfg, 'mac', [lease.macaddr]);
-		if (ip6arr)
-			uci.set('dhcp', cfg, 'hostid', (ip6arr[6] * 0xFFFF + ip6arr[7]).toString(16));
-
-		return uci.save()
-			.then(L.bind(L.ui.changes.init, L.ui.changes))
-			.then(L.bind(L.ui.changes.displayChanges, L.ui.changes));
-	},
-
 	renderLeases(dhcp_leases, host_hints, macaddr) {
 		const leases = Array.isArray(dhcp_leases.dhcp_leases) ? dhcp_leases.dhcp_leases : [];
 		const leases6 = Array.isArray(dhcp_leases.dhcp6_leases) ? dhcp_leases.dhcp6_leases : [];
@@ -229,4 +188,44 @@ return baseclass.extend({
 		]);
 	},
 
+	handleCreateStaticLease(lease, ev) {
+		ev.currentTarget.classList.add('spinning');
+		ev.currentTarget.disabled = true;
+		ev.currentTarget.blur();
+
+		const cfg = uci.add('dhcp', 'host');
+		uci.set('dhcp', cfg, 'name', lease.hostname);
+		uci.set('dhcp', cfg, 'ip', lease.ipaddr);
+		uci.set('dhcp', cfg, 'mac', [lease.macaddr.toUpperCase()]);
+
+		return uci.save()
+			.then(L.bind(L.ui.changes.init, L.ui.changes))
+			.then(L.bind(L.ui.changes.displayChanges, L.ui.changes));
+	},
+
+	handleCreateStaticLease6(lease, ev) {
+		ev.currentTarget.classList.add('spinning');
+		ev.currentTarget.disabled = true;
+		ev.currentTarget.blur();
+
+		const cfg = uci.add('dhcp', 'host');
+		const ip6addr = lease.ip6addrs?.[0]?.replace(/\/128$/, '');
+		const ip6arr = ip6addr ? validation.parseIPv6(ip6addr) : null;
+
+		// Combine DUID and IAID if both available
+		// (note that we know that lease.duid is set here)
+		let duid_iaid = lease.duid.toUpperCase();
+		if (lease.iaid)
+			duid_iaid += `%${lease.iaid}`.toUpperCase();
+
+		uci.set('dhcp', cfg, 'name', lease.hostname);
+		uci.set('dhcp', cfg, 'duid', [duid_iaid]);
+		uci.set('dhcp', cfg, 'mac', [lease.macaddr]);
+		if (ip6arr)
+			uci.set('dhcp', cfg, 'hostid', (ip6arr[6] * 0xFFFF + ip6arr[7]).toString(16));
+
+		return uci.save()
+			.then(L.bind(L.ui.changes.init, L.ui.changes))
+			.then(L.bind(L.ui.changes.displayChanges, L.ui.changes));
+	},
 });
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
@@ -11,6 +11,12 @@ const callLuciDHCPLeases = rpc.declare({
 	expect: { '': {} }
 });
  
+const callODHCPD4Leases = rpc.declare({
+	object: 'dhcp',
+	method: 'ipv4leases',
+	expect: { 'device': {} }
+});
+
 const callODHCPD6Leases = rpc.declare({
 	object: 'dhcp',
 	method: 'ipv6leases',
@@ -49,6 +55,7 @@ return baseclass.extend({
 	load() {
 		return Promise.all([
 			callLuciDHCPLeases(),
+			L.hasSystemFeature('odhcpd', 'dhcpv4') ? L.resolveDefault(callODHCPD4Leases()) : null,
 			L.hasSystemFeature('odhcpd', 'dhcpv6') ? L.resolveDefault(callODHCPD6Leases()) : null,
 			network.getHostHints(),
 			L.hasSystemFeature('ufpd') ? callUfpList() : null,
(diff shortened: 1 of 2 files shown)


Alphix:
Screenshot From 2025-11-09 18-04-54

"Old" tables on top, "new" tables below

Alphix:
@systemcrash I have the feeling that your javascript-fu is much better than mine, so feel free to go through this and see if it can be made more elegant...

systemcrash:
test-beef haha XD

systemcrash:
Could you add `abbr` hover info for the three abbreviations in the table headers? Ideally stuffed with i18n `_('blah')` strings. 
e.g. `<abbr title=\"Domain Name System\">DNS</abbr>`. So there's DUID, FR, and IAID. 

Perhaps we should distinguish the table info source - somehow hint why there are two different tables (dnsmsaq, odhcpd).

Or merge the entries and use a mono table.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -127,29 +133,98 @@ return baseclass.extend({
...
+			}
+
+			return columns;
+		}, this)), E('em', _('There are no active leases')));
+
+		return table4;
+	},
+
+	createOdhcpdTable4(iface_leases4) {
+		console.log("DHCPv4 leases " + JSON.stringify(iface_leases4));
```
systemcrash, line 171: not in the final product. but ok for now.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -127,29 +133,98 @@ return baseclass.extend({
...
+		);
+
+		cbi_update_table(table4, leases4.map(L.bind(function(lease) {
+			let exp;
+			if (lease.valid == 0xffffffff)
+				exp = E('em', _('unlimited'));
+			else if (lease.valid <= 0)
+				exp = E('em', _('expired'));
+			else
+				exp = '%t'.format(lease.valid);
```
systemcrash, line 202: just use interpolated string.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -189,32 +267,162 @@ return baseclass.extend({
...
+		);
+
+		cbi_update_table(table6, leases6.map(L.bind(function(lease) {
+			let exp;
+			if (lease.valid == 0xffffffff)
+				exp = E('em', _('unlimited'));
+			else if (lease.valid <= 0)
+				exp = E('em', _('expired'));
+			else
+				exp = '%t'.format(lease.valid);
```
systemcrash, line 342: same

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -189,32 +267,162 @@ return baseclass.extend({
...
+
+			const hint = lease.macaddr ? this.mac_hints.filter(function(h) { return h[0] == lease.macaddr })[0] : null;
+
+			let host = null;
+			if (hint && lease.hostname && lease.hostname != hint[1] && lease.ip6addrs[0] != hint[1])
+				host = '%s (%s)'.format(lease.hostname, hint[1]);
```
systemcrash, line 348: same

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -127,29 +133,98 @@ return baseclass.extend({
 			else
 				exp = '%t'.format(lease.expires);
 
-			const hint = lease.macaddr ? machints.filter(function(h) { return h[0] == lease.macaddr })[0] : null;
+			const hint = lease.macaddr ? this.mac_hints.filter(function(h) { return h[0] == lease.macaddr })[0] : null;
+			let host = null;
+
+			if (hint && lease.hostname && lease.hostname != hint[1])
+				host = '%s (%s)'.format(lease.hostname, hint[1]);
```
systemcrash, line 140: interpolated

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
```diff
@@ -11,114 +11,120 @@ const callLuciDHCPLeases = rpc.declare({
...
+	expect: { 'device': {} }
+});
+
 const callUfpList = rpc.declare({
 	object: 'fingerprint',
 	method: 'fingerprint',
 });
 
+function DUIDtoMAC(duid) {
+	if (duid == null || duid == '')
```
systemcrash, line 32: `	if (!duid)`?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
Alphix: Oversight

Alphix:
> Could you add `abbr` hover info for the three abbreviations in the table headers? Ideally stuffed with i18n `_('blah')` strings. e.g. `<abbr title=\"Domain Name System\">DNS</abbr>`. So there's DUID, FR, and IAID.

Yeah, that makes sense...

> Perhaps we should distinguish the table info source - somehow hint why there are two different tables (dnsmsaq, odhcpd).
> 
> Or merge the entries and use a mono table.

I don't think the tables can really be merged (and they might diverge even further in the future wrt the data they expose....for example, I'm planning to add a "remove lease" button to the odhcpd table)...but we could change the table titles? As in "Active DHCPv4 Leases" -> "Active DHCPv4 Leases (dnsmasq)" or "Active DHCPv4 leases (odhcpd)"?

Alphix:
> test-beef haha XD

With a DUID including a lot of `0x...BEEF`

Ansuel:
@systemcrash at least it wasn't DEAD (this time)

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
Alphix: How do I do that for `%t`? That's a `cbi.js` addition to the String class, isn't it?

Alphix:
@jow- (sorry, bit of braindump here)...since you seem to have done a lot of the heavy lifting on `rpcd-mod-luci`, I'd like to get your feedback on this as well. The question is if it's a good idea to talk straight to `odhdpd` over `ubus` or whether that is bonkers and I should focus on getting `rpcd-mod-luci` to talk to `odhcpd` over `ubus` instead?

For context, I want to get `rpcd-mod-luci` to stop digging through the `leasefile` that `odhcpd` writes (so that we're free to change the format of the file, which is needed for some new functionality, like persisting leases across `odhcpd` restarts or even reboots, if the user is ok with writing leases to flash). Also, it seems a lot more "correct"/straightforward to talk directly to `odhcpd` over `ubus` to get the desired information rather than dig through its statefiles or going the `LuCI -- (ubus) --> rpcd -- (ubus) --> odhcpd` path.

Now, converting `luci-mod-status` is pretty straightforward (as can be seen in this PR), but `luci-mod-network` is a different beast...it'll call `getHostHints`, `getDUIDHints` and `getDHCPLeases` (meaning `luci-mod-status` digs through the leasefile four times, twice for `getDHCPLeases`).

`getDUIDHints` and `getDHCPLeases` are easy enough to synthesise in JavaScript from the data that `odhcpd` returns, but `getHostHints` can't really be skipped since it consults so many other sources (`netlink`, `uci`, `leasefiles`, `/etc/ethers`, `getifaddrs()` and `RDNS`)...it's also used by a lot of other modules...the more I look at it the more questionable it seems...

So...should I just throw in the towel and focus on getting `rpcd-mod-luci` to get more data via `ubus`, even if it'll add an extra intermediary?

systemcrash:
@Alphix did you want to bring this up to current or is this superseded?

Alphix:
> @Alphix did you want to bring this up to current or is this superseded?

It needs further work, I think it might be material for after the next OpenWrt branch... Basically I think I need to get rpcd-mod-luci to talk to odhcpd first... Then I can think about more direct approaches after that...

Self-Hosting-Group:
@Alphix What do you think about using `Hostname` for IPv6 DHCP, as for IPv4?
`E('th', { 'class': 'th' }, _('Hostname')),`
(And the use of `Host` for `Associated Stations` could possibly be changed to `Device`.)

Alphix:
> @Alphix What do you think about using `Hostname` for IPv6 DHCP, as for IPv4? `E('th', { 'class': 'th' }, _('Hostname')),`

Sure, see https://github.com/openwrt/luci/pull/8140

> (And the use of `Host` for `Associated Stations` could possibly be changed to `Device`.)

I don't have any opinion on that, but feel free to create a PR and ping @systemcrash

GoetzGoerisch:
While you are at it, a comment from my end.

Currently using `dnsmasq` (v4) and `odhcpd-ipv6only` in coexistense if there is a static lease with a DNS entry from `dnsmasq` the FQDN is shown in brackets next to the hostname in the v4 and v6 lease table.

If we could get consinstent this, that would be great.

Alphix:
> While you are at it, a comment from my end.
> 
> Currently using `dnsmasq` (v4) and `odhcpd-ipv6only` in coexistense if there is a static lease with a DNS entry from `dnsmasq` the FQDN is shown in brackets next to the hostname in the v4 and v6 lease table.
> 
> If we could get consinstent this, that would be great.

That's more complicated. The FQDN comes from `luci-rpc`s `getHostHints()` method, where the MAC address is used as the lookup key. The MAC address isn't always available (especially for DHCPv6 leases).

And just to be clear, that's pre-existing functionality, not something I've messed around with :)

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
Cyprianhenry: https://www.blockchain.com/blog/posts/introducing-multichain-deposits?utm_source=Iterable&utm_medium=email&utm_campaign=campaign_15979404

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
openwrt-ai, line 171: `createOdhcpdTable4()` is missing the null guard that its sibling `createOdhcpdTable6()` has (`if (!iface_leases6) return null;`). In `load()`, `odhcpd_leases4` is set to literal `null` when `L.hasSystemFeature('odhcpd', 'dhcpv4')` is false, and `createOdhcpdTable4(odhcpd_leases4)` is then called unconditionally from `renderLeases()`. With a null argument, `Object.entries(iface_leases4)` (and the `JSON.stringify` on the debug line above it) will throw, so on any system where odhcpd does not advertise dhcpv4 the leases page would fail to render. Add the same early `if (!iface_leases4) return null;` guard used in the v6 path.

---

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js
openwrt-ai, line 140: In `createDnsmasqTable4()`, `const vendor = this.ufp_list?.[lease.macaddr.toLowerCase()]?.vendor ?? null;` dereferences `lease.macaddr` unconditionally. The old code guarded this (`if (macaddr) vendor = macaddr[lease.macaddr.toLowerCase()]`), and the button block a few lines down still treats `lease.macaddr` as possibly falsy (`if (!this.isReadonlyView && lease.macaddr)`). If a dnsmasq DHCPv4 lease ever lacks `macaddr`, `lease.macaddr.toLowerCase()` throws. Consider guarding, e.g. `const vendor = lease.macaddr ? (this.ufp_list?.[lease.macaddr.toLowerCase()]?.vendor ?? null) : null;`.

---

openwrt-ai (commented):
Reviewed the full PR diff (9 commits, first automated review). This is an RFC/draft, so most of the design-level points (comparison tables, debug logging, abbr tooltips) are already covered by the author and @systemcrash and are not repeated here. Two concrete robustness issues and one nit below.

nit: the split into separate dnsmasq/odhcpd tables introduces duplicate DOM `id` attributes — both `createDnsmasqTable4()` and `createOdhcpdTable4()` emit `id: 'status_leases4'`, and both v6 tables emit `id: 'status_leases6'`. Duplicate element IDs are invalid HTML and will break any `getElementById`/CSS-id selectors; consider distinct ids (e.g. `status_leases4_odhcpd`).

---

PR #8090, opened 2025-11-12
luci-base: improve docs for uci.get_bool

efahl:
The get_bool function documentation was missing an option value, so add it.  Provide a link into the developer's guide on how it's used in scripts, for better background.

---
Stumbled upon this while perusing

https://openwrt.org/docs/guide-developer/config-scripting#reading_booleans
https://github.com/openwrt/openwrt/blob/5ab6e6011feb5d3fd782a8fe19c5ce3acfecf712/package/base-files/files/lib/functions.sh#L153
https://github.com/openwrt/packages/blame/6423781254b9f3e52c6102fb2cbcd9f99f2445a3/net/mwan3/files/usr/share/rpcd/ucode/mwan3#L32

I need to fix ucode uci to have a `get_bool`, too...


systemcrash:
Merged. Thanks @efahl !

PR #8092, opened 2025-11-14
luci-theme-*: use ES6 syntax

eamonxg:
Updated to arrow functions for clearer `this` binding, replaced slice.call with Array.from for more readable array conversion, and adopted ES6 block-scoped declarations.

### DIFF
--- a/themes/luci-theme-openwrt-2020/htdocs/luci-static/resources/menu-openwrt2020.js
+++ b/themes/luci-theme-openwrt-2020/htdocs/luci-static/resources/menu-openwrt2020.js
@@ -4,7 +4,7 @@
 
 return baseclass.extend({
 	__init__() {
-		ui.menu.load().then(L.bind(this.render, this));
+		ui.menu.load().then((tree) => this.render(tree));
 	},
 
 	render(tree) {


eamonxg:
Hi @systemcrash , do you think this update is necessary?
I believe the ES6 syntax makes the code clearer and easier to read.

feckert:
Please split this PR in several commits for each theme.

eamonxg:
> Please split this PR in several commits for each theme.

Done.

themes/luci-theme-bootstrap/htdocs/luci-static/resources/view/bootstrap/sysauth.js
```diff
@@ -3,32 +3,35 @@
...
 
 		document.querySelector('input[type="password"]').focus();
 
 		return '';
 	},
 
-	addFooter: function() {}
-});
+	addFooter() {},
+});
```
feckert, line 39: Missing new line

themes/luci-theme-bootstrap/htdocs/luci-static/resources/view/bootstrap/sysauth.js
```diff
@@ -3,32 +3,35 @@
...
+		const form = document.querySelector('form');
+		const btn = document.querySelector('button');
 
-		var dlg = ui.showModal(
+		const dlg = ui.showModal(
 			_('Authorization Required'),
-			[].slice.call(document.querySelectorAll('section > *')),
-			'login'
+			Array.from(document.querySelectorAll('section > *')),
+			'login',
```
feckert: Why is there a `,`

themes/luci-theme-bootstrap/htdocs/luci-static/resources/view/bootstrap/sysauth.js
```diff
@@ -3,32 +3,35 @@
...
 		);
 
-		form.addEventListener('keypress', function(ev) {
-			if (ev.key == 'Enter')
-				btn.click();
+		form.addEventListener('keypress', (ev) => {
+			if (ev.key === 'Enter') btn.click();
```
feckert: please move btn.click into a new line.

eamonxg:
> I am not so familiar with the newest javascript. But what does this mean new `arrow` function? Can you give me more context about that?

Sure. In ES6, arrow functions don’t create their own `this`.
 Instead, they **inherit the `this` value from the surrounding scope**, which makes their behavior easier to understand in callback situations.

In this code:

```js
return baseclass.extend({
    __init__() {
        ui.menu.load().then((tree) => this.render(tree));
    },
});
```

the arrow function inherits `this` from `__init__()`, so `this.render(tree)` works exactly the same as:

```js
ui.menu.load().then(L.bind(this.render, this));
```

The change simply removes the need for the custom `L.bind()` helper.
 For developers familiar with modern JavaScript, the arrow function form is more intuitive and readable than the Luci-specific binding syntax.

themes/luci-theme-bootstrap/htdocs/luci-static/resources/view/bootstrap/sysauth.js
eamonxg: Sorry about that — the trailing comma was added automatically by my formatter ([Prettier](https://prettier.io/docs/)), which is my default JS formatting tool.

I noticed that LuCI uses its own formatting script here:
https://github.com/openwrt/luci/blob/fa319cfe95508d2b3662fcb73f0eb19a6c752087/applications/luci-app-example/README.md#L40

I ran it before submitting, but it didn’t remove the comma.
I’ll fix it manually now.

themes/luci-theme-bootstrap/htdocs/luci-static/resources/view/bootstrap/sysauth.js
eamonxg: done~

themes/luci-theme-bootstrap/htdocs/luci-static/resources/view/bootstrap/sysauth.js
eamonxg: fixed~

feckert:
@eamonxg Thanks for your explanation. `L.bind(this ...)` has always annoyed me and makes the code really very confusing. The question now is, do we only want to support ES6 browsers? What about the other browsers that don't have a new JavaScript engine yet?
@systemcrash Is this even relevant, or do all current browsers already support ES6 from their JavaScript engine?

jow-:
@feckert - you can safely rely on ES6 features now, other parts of LuCI, mainly luci-base, already require them anyway.

systemcrash:
Yes @feckert ES6 is fine. Some places even use newer standards…

feckert:
Thanks for the clarification.

feckert:
@eamonxg Thanks merged

PR #8093, opened 2025-11-15
luci-app-firewall: move NAT6 to general like NAT4

AndyChiang888:
move NAT6 to general like NAT4.
This makes the NAT6 flag more visually apparent.

image

image

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/zones.js
@@ -173,9 +173,22 @@ return view.extend({
 			var masq_dest = uci.get('firewall', section_id, 'masq_dest')
 			if ((!family || family.indexOf('6') == -1) && (masq_src || masq_dest))
 				return _('Limited masquerading enabled');
-
 			return null;
 		};
+
+		if (fw4) {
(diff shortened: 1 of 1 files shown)


AndyChiang888:
@systemcrash @hnyman How does this look?

hnyman:
I understand the idea about similarity with IPv4, but now IPv6 masquerading looks too prominent to me, as it is rarely needed.

With IPv4, the masquerading/NAT is typical in LAN, while with IPv6 it should be needed only rarely. Somehow advice about that should be visible. Otherwise users will unnecessarily enable IPv6 NAT.

AndyChiang888:
NAT6 can add a hint that is generally unnecessary, and I think placing it alongside NAT4 would be more fitting.
It feels a bit odd to put it in the advanced right now.

hnyman:
> This is typically not enabled on the <em>wan</em> zone.

I would write the second sentence more explanatorily for novices, something like:

"Typically not needed, as native IPv6 routing is usually used instead of NAT6."

AndyChiang888:
👌

1715173329:
Please see more discussion here #6317
I agreed to not expose this option in general tab, generally speaking you have more choices rather than using nat66.
At least this should be invisible on main page, i.e. add `o.modalonly = true;`.

AndyChiang888:
I changed `o.editable = true;` to `o.modalonly = true;`, on main page is invisible.

PR #8096, opened 2025-11-15
luci-app-attendedsysupgrade: rework logic in check-for-new-firmware

efahl:
Fix a bug in the check for new firmware feature, where it would indicate that an older version should be installed.

While we're here, simplify the top-level logic to make it more readable and maintainable, and clean up the table generator.

---
Should be clean for both snapshot and 24.10, tested on 24.10.4.


systemcrash:
Merged. Thanks @efahl !

PR #8098, opened 2025-11-16
[24.10] luci-app-attendedsysupgrade: rework logic in check-for-new-firmware

efahl:
Fix a bug in the check for new firmware feature, where it would indicate that an older version should be installed.

While we're here, simplify the top-level logic to make it more readable and maintainable, and clean up the table generator.

(cherry picked from commit 3034f05d6503406d714b1b50dba5d4c157720dd3)


systemcrash:
Merged. Thanks @efahl !

PR #8099, opened 2025-11-18
luci-mod-network: Implement WiFi QR Codes

systemcrash:
This implements T:, R:, S: and H: parameters standardized within WiFi QR codes.

This leaves the SAE-PK related options unimplemented. A future UI will likely handle SAE-PK options differently than in hostapd_bss_options.

Current parameters are based on:
https://www.wi-fi.org/system/files/WPA3%20Specification%20v3.5.pdf#page=33

Wifi security tab:

Screenshot 2025-11-18 at 03 41 53

Clicking through to generate QR presents a UI
Screenshot 2025-11-18 at 03 43 52

ping @hnyman @efahl @Alphix @Ansuel @feckert 

- [X] Tested on: 24.10 :white_check_mark:

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -10,6 +10,7 @@
 'require network';
 'require firewall';
 'require tools.widgets as widgets';
+'require uqr';
 
 const isReadonlyView = !L.hasViewPermission();
 
@@ -26,6 +27,23 @@ function render_radio_badge(radioDev) {
 		' ',
 		radioDev.getName()
 	]);
+}
+
+function buildSVGQRCode(data, code, options, dummy=false) {
+	const opts = {
+		pixelSize: 4,
+		whiteColor: 'white',
+		blackColor: 'black',
+		ecc: 'M',
+		...options
+	};
+	const svg = uqr.renderSVG(data, opts);
+	if (dummy)
+		return svg;
+	else {
+		code.style.opacity = '';
+		dom.content(code, Object.assign(E(svg), { style: 'width:100%;height:auto' }));
+	}
 }
 
 function render_signal_badge(signalPercent, signalValue, noiseValue, wrap, mode) {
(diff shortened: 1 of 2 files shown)


castillofrancodamian:
I find it more useful if the button is in the wireless interfaces overview or in the LuCI overview.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -28,6 +29,21 @@ function render_radio_badge(radioDev) {
...
...7h1v1h-1zM16 17h1v1h-1zM17 17h1v1h-1zM18 17h1v1h-1zM4 18h1v1H4zM5 18h1v1H5zM6 18h1v1H6zM7 18h1v1H7zM8 18h1v1H8zM9 18h1v1H9zM10 18h1v1h-1zM14 18h1v1h-1zM16 18h1v1h-1zM17 18h1v1h-1zM21 18h1v1h-1zM22 18h1v1h-1zM23 18h1v1h-1zM4 19h1v1H4zM10 19h1v1h-1zM12 19h1v1h-1zM13 19h1v1h-1zM15 19h1v1h-1zM16 19h1v1h-1zM19 19h1v1h-1zM21 19h1v1h-1zM23 19h1v1h-1zM24 19h1v1h-1zM4 20h1v1H4zM6 20h1v1H6zM7 20h1v1H7zM8 20h1v1H8zM10 20h1v1h-1zM12 20h1v1h-1zM13 20h1v1h-1zM15 20h1v1h-1zM18 20h1v1h-1zM19 20h1v1h-1zM20 20h1v1h-1zM22 20h1v1h-1zM23 20h1v1h-1zM24 20h1v1h-1zM4 21h1v1H4zM6 21h1v1H6zM7 21h1v1H7zM8 21h1v1H8zM10 21h1v1h-1zM13 21h1v1h-1zM15 21h1v1h-1zM16 21h1v1h-1zM19 21h1v1h-1zM21 21h1v1h-1zM23 21h1v1h-1zM24 21h1v1h-1zM4 22h1v1H4zM6 22h1v1H6zM7 22h1v1H7zM8 22h1v1H8zM10 22h1v1h-1zM13 22h1v1h-1zM15 22h1v1h-1zM18 22h1v1h-1zM19 22h1v1h-1zM20 22h1v1h-1zM21 22h1v1h-1zM22 22h1v1h-1zM4 23h1v1H4zM10 23h1v1h-1zM12 23h1v1h-1zM13 23h1v1h-1zM14 23h1v1h-1zM17 23h1v1h-1zM18 23h1v1h-1zM20 23h1v1h-1zM22 23h1v1h-1zM4 24h1v1H4zM5 24h1v1H5zM6 24h1v1H6zM7 24h1v1H7zM8 24h1v1H8zM9 24h1v1H9zM10 24h1v1h-1zM12 24h1v1h-1zM13 24h1v1h-1zM14 24h1v1h-1zM16 24h1v1h-1zM17 24h1v1h-1zM18 24h1v1h-1zM22 24h1v1h-1zM24 24h1v1h-1z"/></svg>';
```
efahl: Would it be a good idea to move `qrIcon` to a static resource and share it with other packages?  I see similar code in `luci-proto-wireguard`...

efahl:
> I find it more useful if the button is in the wireless interfaces overview or in the LuCI overview.

But isn't the point to set the password, generate a new QR and print it off so you can post it (in a hotel room or by your front door or wherever guests need it).  Seems like you want it right next to that "set your password" part for discovery...

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
systemcrash: That's gone now. We generate a dummy qr-code on the fly instead.

systemcrash:
> I find it more useful if the button is in the wireless interfaces overview or in the LuCI overview.

Yeah, I figure you could put this in other more prominent places in the GUI, with the drawback that those other places will invoke the uqr lib dependency to generate the QR code. By the password seems the most evident placing.

Edit: any placement would ideally put it behind a button so casual shoulder surfers wouldn't capture the sensitive content.

efahl:
Just tested on a setup with WPA3, so as expected the password options weren't up to snuff, but what I wanted to see was if the SSID would work, because I've got goofy unicode emojis in the ids on that device:

image

And yes, it did, the phone read the whole thing just fine.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1464,6 +1482,177 @@ return view.extend({
...
+					const wifiHide = this.section.formvalue(section_id, 'hidden') === '1'; // H
+
+					/* trdisable:
+					0 WPA3-Personal
+					1 SAE-PK
+					2 WPA3-Enterprise
+					3 WiFi-Enhanced Open */
+					let trdisable = ''; // R
+					switch (true) {
+					case (wifiEncr === 'sae'): trdisable = 0; break; // 'sae' i.e. WPA3-Personal
```
efahl, line 1545: Should this one be a `startswith`, too, to catch the `sae-mixed` case?

systemcrash: My reading of Table 6 https://www.wi-fi.org/system/files/WPA3%20Specification%20v3.5.pdf#page=36 is no, that if it's pure SAE, then transition disable is zero. Other mixed encryptions elide trdisable.

efahl: Got it.  I was looking at the conditional up at 1371 and thought that implied the same password constraints, the table clarifies since -mixed includes PSK, it would be zero.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1464,6 +1482,177 @@ return view.extend({
...
+
+							if (this.isValid(section_id)) {
+								conf.firstChild.data = wifiQRGenerator(section_id);
+								code.style.opacity = '.5';
+
+								buildSVGQRCode(conf.firstChild.data, code, {ecc: ecc.getValue()});
+							}
+						};
+
+						qro = qrs.option(form.ListValue, 'ecc', _('QR ECC'));
```
stokito, line 1592: (minor, just opinion) The ECC can be hard coded to M or even L. Higher layers are needed for street ad or some paper printed materials that may have a lot of errors. If we can simplify something then it's probably better to do.

systemcrash: Users may (or may not) take this QR and print it out, in which case having the choice of ECC is valuable.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
systemcrash: In that case ('*-mixed') it seems it would just not be included.

Neustradamus:
@systemcrash: Nice PR, good job!

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
stokito: Ok, then "QR ECC" may be "Error correction level (ECC)" for those who don't know

systemcrash:
Closed by 805b2db67012ee76ea8baf57cf83c4dc31b72832

PR #8104, opened 2025-11-20
luci-mod-system: add translation string for language

AndyChiang888:
add translation string for language

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
@@ -234,7 +234,7 @@ return view.extend({
 		o.uciconfig = 'luci';
 		o.ucisection = 'main';
 		o.ucioption = 'lang';
-		o.value('auto');
+		o.value('auto', _('auto'));
 
 		var l = Object.assign({ en: 'English' }, uci.get('luci', 'languages')),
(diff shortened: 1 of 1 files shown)


systemcrash:
Merged. Thanks @AndyChiang888 !

modules/luci-mod-system/htdocs/luci-static/resources/view/system/system.js
```diff
@@ -234,7 +234,7 @@ return view.extend({
 		o.uciconfig = 'luci';
 		o.ucisection = 'main';
 		o.ucioption = 'lang';
-		o.value('auto');
+		o.value('auto', _('auto'));
```
jow-, line 237: A translation context hint might make sense here:
`_('auto', 'Automatic language choice')`

The term `auto` is very generic and might not translate the same in all contexts and languages.

AndyChiang888: For now, it seems that most of the auto translations are correct. If someone brings up an issue, we can adjust them then.
See:
https://hosted.weblate.org/search/openwrt/luci/?offset=&q=auto&sort_by=num_words&checksum=
and
https://github.com/openwrt/luci/blob/fdac3c0100ec4a02c6fa1f0149678889c17f53b6/modules/luci-base/po/templates/base.pot#L12452

PR #8112, opened 2025-11-24
luci-app-wol: replace fs.stat/exec with safe RPC backend, fix ACLs

mdevolde:
- [x] Tested on: aarch64, OpenWrt 24.10.4, Opera One(version : 124.0.5705.42) :white_check_mark: 
- [x] \( Preferred ) Mention: @mdevolde

518189561-3d2b61e8-5241-49e5-9058-54c46eb06236

518201649-0ff4c5e8-3eb5-43c6-90fc-0a18e9200d47

Screenshots of functional access to `luci-app-wol` for a restricted user (restricted to these permissions : read `unauthenticated`, read `luci-base`, read `luci-app-wol`, write `luci-app-wol`).

- [x] Description:

In its current state, if we create a restricted user as mentioned above, access to `luci-app-wol` does not work because it cannot verify that the two executables `/usr/bin/etherwake` and `/usr/bin/wol` exist via `fs.stat`.
Furthermore, it cannot be executed via `fs.exec` either, as ACLs are missing in `luci-app-wol.json` (`ubus file`).
Finally, an ACL concerning the use of `DeviceSelect` is also missing.

Instead of adding broad `ubus file.*` ACLs, a dedicated `luci.wol` backend has been added, providing two RPC methods:
- `stat` to verify that the two executables exist and are executable
- `exec` to execute one of these two executables, with parameters.

The missing ACL for using `DeviceSelect` has also been added (`getNetworkDevices`).

Adding this backend and the appropriate ACLs makes the `luci-app-wol` app functional, even for a restricted user, by assigning only the permissions necessary for it to work properly.

Signed-off-by: Martin Devolder <martin.devolder2@gmail.com>

### DIFF
--- a/applications/luci-app-wol/htdocs/luci-static/resources/view/wol.js
+++ b/applications/luci-app-wol/htdocs/luci-static/resources/view/wol.js
@@ -8,8 +8,25 @@
 'require form';
 'require tools.widgets as widgets';
 
+const ETHERWAKE_BIN = '/usr/bin/etherwake';
+const WAKEONLAN_BIN = '/usr/bin/wakeonlan';
+
 return view.extend({
 	formdata: { wol: {} },
+
+	callStat: rpc.declare({
+		object: 'luci.wol',
+		method: 'stat',
+		params: [ ],
+		expect: { }
+	}),
+
+	callExec: rpc.declare({
+		object: 'luci.wol',
+		method: 'exec',
+		params: [ 'name', 'args' ],
+		expect: { }
+	}),
 
 	callHostHints: rpc.declare({
 		object: 'luci-rpc',
@@ -19,17 +36,15 @@ return view.extend({
 
 	load: function() {
 		return Promise.all([
-			L.resolveDefault(fs.stat('/usr/bin/etherwake')),
-			L.resolveDefault(fs.stat('/usr/bin/wol')),
+			L.resolveDefault(this.callStat()),
 			this.callHostHints(),
 			uci.load('etherwake')
 		]);
 	},
 
-	render: function(data) {
-		var has_ewk = data[0],
-		    has_wol = data[1],
-		    hosts = data[2],
+	render([stat, hosts]) {
+			var has_ewk = stat && stat.etherwake,
+				has_wol = stat && stat.wakeonlan,
 		    m, s, o;
 
 		this.formdata.has_ewk = has_ewk;
@@ -44,8 +59,8 @@ return view.extend({
 			o = s.option(form.ListValue, 'executable', _('WoL program'),
 				_('Sometimes only one of the two tools works. If one fails, try the other one'));
 
-			o.value('/usr/bin/etherwake', 'Etherwake');
-			o.value('/usr/bin/wol', 'WoL');
+			o.value(ETHERWAKE_BIN, 'Etherwake');
+			o.value(WAKEONLAN_BIN, 'Wakeonlan');
 		}
 
 		if (has_ewk) {
@@ -67,7 +82,7 @@ return view.extend({
 			});
 
 			if (has_wol)
-				o.depends('executable', '/usr/bin/etherwake');
+				o.depends('executable', ETHERWAKE_BIN);
 		}
 
 		o = s.option(form.Value, 'mac', _('Host to wake up'),
@@ -88,7 +103,7 @@ return view.extend({
 			o = s.option(form.Flag, 'broadcast', _('Send to broadcast address'));
 
 			if (has_wol)
-				o.depends('executable', '/usr/bin/etherwake');
+				o.depends('executable', ETHERWAKE_BIN);
 		}
 
 		return m.render();
@@ -96,16 +111,17 @@ return view.extend({
 
 	handleWakeup: function(ev) {
 		var map = document.querySelector('#maincontent .cbi-map'),
-		    data = this.formdata;
+		    data = this.formdata,
+		    self = this;
 
 		return dom.callClassMethod(map, 'save').then(function() {
 			if (!data.wol.mac)
 				return alert(_('No target host specified!'));
 
-			var bin = data.executable || (data.has_ewk ? '/usr/bin/etherwake' : '/usr/bin/wol'),
+			var bin = data.wol.executable || (data.has_ewk ? ETHERWAKE_BIN : WAKEONLAN_BIN),
 			    args = [];
 
-			if (bin == '/usr/bin/etherwake') {
+			if (bin == ETHERWAKE_BIN) {
 				args.push('-D', '-i', data.wol.iface);
 
 				if (data.wol.broadcast == '1')
@@ -114,16 +130,16 @@ return view.extend({
 				args.push(data.wol.mac);
 			}
 			else {
-				args.push('-v', data.wol.mac);
+				args.push(data.wol.mac);
 			}
 
 			ui.showModal(_('Waking host'), [
 				E('p', { 'class': 'spinning' }, [ _('Starting WoL utility…') ])
 			]);
-
-			return fs.exec(bin, args).then(function(res) {
+			
+			return self.callExec(bin, args).then(function(res) {
 				ui.showModal(_('Waking host'), [
-					res.stderr ? E('p', [ res.stdout ]) : '',
+					res.stdout ? E('p', [ res.stdout ]) : '',
 					res.stderr ? E('pre', [ res.stderr ]) : '',
 					E('div', { 'class': 'right' }, [
 						E('button', {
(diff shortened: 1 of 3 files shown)


applications/luci-app-wol/htdocs/luci-static/resources/view/wol.js
```diff
@@ -19,16 +33,16 @@ return view.extend({
...
-			L.resolveDefault(fs.stat('/usr/bin/etherwake')),
-			L.resolveDefault(fs.stat('/usr/bin/wol')),
+			L.resolveDefault(this.callStat({'path': '/usr/bin/etherwake'})),
+			L.resolveDefault(this.callStat({'path': '/usr/bin/wol'})),
 			this.callHostHints(),
 			uci.load('etherwake')
 		]);
 	},
 
 	render: function(data) {
```
systemcrash: might as well just do this: 
```js
render([stat_ewk, stat_wol, hosts]) {
```

applications/luci-app-wol/root/usr/libexec/rpcd/luci.wol
```diff
@@ -0,0 +1,90 @@
+#!/bin/sh
```
systemcrash: recommend this be written as ucode, then you avoid manual json handling pitfalls, and just build objects directly.

jow-:
Nice work! Since you convert to a dedicated ubus backend anyway, you can probably hide away the stat/exec logic and simply provide a `wol.wake` method which takes a MAC address argument and internally invokes the correct tool.

applications/luci-app-wol/htdocs/luci-static/resources/view/wol.js
mdevolde: fixed

applications/luci-app-wol/root/usr/libexec/rpcd/luci.wol
mdevolde: I prefer to keep the sh version, as I do not have enough experience with ucode to develop a similar, clean solution. Since the sh utility is quite small, I think it remains readable despite the verbose JSON manipulation.

mdevolde:
@jow- I think that proceeding in this way would cause the module to lose functionality, because currently, if the user has both binaries on their OpenWRT, they can decide which one to use.

mdevolde:
Rebased and force-pushed with additional updates.
Adapted `luci-app-wol` to recent versions of OpenWrt where the `wol` binary is no longer provided. The code now switches to `wakeonlan`, which is the current replacement for the `wol` package. Updated the paths, existence tests, executable selection, and corrected the call by removing the irrelevant -v option.
Also corrected an incorrect reference to `data.executable` by using `data.wol.executable`.

systemcrash:
What (or which check) is data.wol predicated on?

Also, you might as well replace all instances of `/bin/blah` with a const string at the start of the file, I think.

mdevolde:
`data.wol` always exists because the view initializes `formdata` with:
```js
formdata: { wol: {} },
```

All form fields belong to that `wol` section, and Luci’s `form.JSONMap` populates `formdata.wol.*` during the `save()` call.

Therefore `data.wol.executable` and `data.wol.mac` are always defined after `save()`, `data.wol.iface` and `data.wol.broadcast` can be defined if `data.wol.executable` is `/usr/bin/etherwake`, after `save()`. So, the previous reference to `data.executable` was incorrect.

Here is the typical structure of the object:
```json
{
    "wol": {
        ".anonymous": false,
        ".name": "wol",
        ".type": "wol",
        ".index": 0,
        "executable": "/usr/bin/etherwake",
        "mac": "4E:38:20:0E:C2:F2",
        "iface": "br-guest",
        "broadcast": "1"
    },
    "has_ewk": true,
    "has_wol": true
}
```

Good idea for the constants, I'll do that now.

mdevolde:
~~Hi @daveth3t3chg33k, as I specified in @systemcrash's review comments, I preferred to write it in `sh` because I have less experience writing a similar utility in `ucode`.~~

If this is really a blocking factor for the maintainers, I will look into it.

*Finally done*

applications/luci-app-wol/root/usr/libexec/rpcd/luci.wol
systemcrash: Something like this to get you started (not run-tested):
```js
#!/usr/bin/ucode

'use strict';

import { access, popen, error } from 'fs';

const etherwake = '/usr/bin/etherwake';
const wakeonlan = '/usr/bin/wakeonlan';


function shellquote(s) {
	return `'${replace(s, "'", "'\\''")}'`;
}


const methods = {

	stat: {
		call: function(request) {
			const result = {};

			if (access(etherwake)) {
				result.etherwake = true;
			}

			if (access(wakeonlan)) {
				result.wakeonlan = true;
			}

			return result;
		}
	},

	exec: {
		args: { name: 'name', action: 'action' },
		call: function(request) {
			const result = {};
			if (request.args.name == etherwake || request.args.name == wakeonlan) {

				const fd = popen(shellquote(request.args.name));

				result.stdout = fd.read('all');
				result.stderr = fd.error();
				result.code = 0;

				return result;
			} else {
				result.stderr = 'disallowed';
				result.code = 1;
			}
		}
	}
};

return { wakeonlan: methods };

```

mdevolde: Thank you for your help. I will look into it, improve it and run some tests.

systemcrash: Check https://ucode.mein.io/ for references

mdevolde:
@systemcrash  I converted the utility to `ucode`.
It works correctly.
The only drawback compared to the `sh` version is the difficulty of retrieving `stderr` in `popen`.
However, displaying `stderr` in the app interface only provided optional information for the user, so this is not a limitation.

applications/luci-app-wol/root/usr/libexec/rpcd/luci.wol
mdevolde: done

mdevolde:
I forgot the shell quote, I'll add it tomorrow.

mdevolde:
I added shellquote.

systemcrash:
Can you wrap your commit message to <= 80 chars please?

mdevolde:
@systemcrash Done, each line of the commit title and description is now less than 80 characters long.

applications/luci-app-wol/root/usr/share/rpcd/acl.d/luci-app-wol.json
```diff
@@ -3,14 +3,14 @@
 		"description": "Grant access to wake-on-lan executables",
 		"read": {
 			"ubus": {
-				"luci-rpc": [ "getHostHints" ]
+				"luci.wol": [ "stat" ],
+				"luci-rpc": [ "getHostHints", "getNetworkDevices" ]
```
systemcrash, line 7: Are you sure we need `getNetworkDevices`? I don't see it used.

mdevolde: Yep, it's used internally by `widget.DeviceSelect`

systemcrash: So the widget didn't work without it?

mdevolde: In `modules\luci-base\htdocs\luci-static\resources\tools\widgets.js`, at line 564, `DeviceSelect` use `network.getDevices()`.
In `modules\luci-base\htdocs\luci-static\resources\network.js`, at line 1159, `getDevices` use `initNetworkState()`.
In `modules\luci-base\htdocs\luci-static\resources\network.js`, at line 359, `initNetworkState` use `callLuciNetworkDevices()` :
```js
var callLuciNetworkDevices = rpc.declare({
	object: 'luci-rpc',
	method: 'getNetworkDevices',
	expect: { '': {} }
});
```

Without this permission, `DeviceSelect` cannot retrieve existing devices, and users who choose to use `etherwake` cannot select one of the existing devices in the web interface of `luci-app-wol`.

This is for the first field here:
image

systemcrash: Good catch!

systemcrash:
Merged. Thanks @mdevolde !

mdevolde:
@systemcrash Thank you for the follow-up and the regular reviews. 

Are you planning to backport this fix to
the current stable branch or will it be available in the next release?

systemcrash:
I usually give it a week and see how it fares in master - with any follow-up fixes coming shortly thereafter. It will land in the next release, and weeklies.

mdevolde:
Okay, thanks again.

PR #8115, opened 2025-11-26
luci-theme-bootstrap: disable Dark Reader

mikma:
I propose to disable [Dark Deader](https://darkreader.org/) browser extension in the bootstap theme since the theme has built-in dark and light modes, as well as a mode which auto-detects the user's preference. Dark Reader also doesn't work well with the bootstrap theme, as seen below when using Dark Reader:

Before the change in this PR:
before

After the change in this PR:
after


systemcrash:
Is darkreader the browser's own built in reader mode?

mikma:
It's a browser extension. I added a link in the PR.

systemcrash:
On the one hand we carry a 'stop' for a proprietary browser extension, on the other, it's trivial.

systemcrash:
Merged. Thanks @mikma

PR #8122, opened 2025-11-28
luci-mod-network: bump reassociation_deadline default to 20000

rany2:
With OpenWRT commit [1], the default for reassociation_deadline is now 20000. Update LuCI to reflect that change.

[1]: https://github.com/openwrt/openwrt/commit/a7790ce41099549cf6c97765561ac716d102ae5e


systemcrash:
Thanks @rany2! Merged.

PR #8129, opened 2025-11-30
luci-mod-network,-status: add IPv6-only preferred option, fix some table titles

Alphix:
Some simple patches to add support for IPv6-only preferred, and to make the table titles for DHCP leases a bit more consistent.

- [x] Tested on: (NanoPi R6S, main, Chrome) :white_check_mark:

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
@@ -36,13 +36,13 @@ const callUfpList = rpc.declare({
 const CBILeaseStatus = form.DummyValue.extend({
 	renderWidget(section_id, option_id, cfgvalue) {
 		return E([
-			E('h4', _('Active DHCP Leases')),
+			E('h4', _('Active DHCPv4 Leases')),
 			E('table', { 'id': 'lease_status_table', 'class': 'table' }, [
 				E('tr', { 'class': 'tr table-titles' }, [
 					E('th', { 'class': 'th' }, _('Hostname')),
 					E('th', { 'class': 'th' }, _('IPv4 address')),
 					E('th', { 'class': 'th' }, _('MAC address')),
-					E('th', { 'class': 'th' }, _('Lease time remaining'))
+					E('th', { 'class': 'th' }, _('Remaining time'))
 				]),
 				E('tr', { 'class': 'tr placeholder' }, [
 					E('td', { 'class': 'td' }, E('em', _('Collecting data...')))
@@ -62,7 +62,7 @@ const CBILease6Status = form.DummyValue.extend({
 					E('th', { 'class': 'th' }, _('IPv6 addresses')),
 					E('th', { 'class': 'th' }, _('DUID')),
 					E('th', { 'class': 'th' }, _('IAID')),
-					E('th', { 'class': 'th' }, _('Lease time remaining'))
+					E('th', { 'class': 'th' }, _('Remaining time'))
 				]),
 				E('tr', { 'class': 'tr placeholder' }, [
 					E('td', { 'class': 'td' }, E('em', _('Collecting data...')))


Alphix:
@systemcrash ping :)

hnyman:
Please avoid those no-op character case changes in translated strings.
Your changes will invalidate all translations.

(And I do not understand why address should be Address in a multi-word item.

Alphix:
> Please avoid those no-op character case changes in translated strings. Your changes will invalidate all translations.

I'm well-aware that the change will invalidate translations, but right now the UI is inconsistent, there's titles like "Static Lease" and titles like "Lease time remaining" (i.e. different Title Capitalization rules are being applied in the very same tables). I can't really fix that without causing work for the translators, which is unfortunate, but also unavoidable?

> (And I do not understand why address should be Address in a multi-word item.

I don't have a strong opinion on this, but whatever title capitalization scheme we use, it should be consistent. Using this style (consistent with e.g. the Chicago Manual of Style) created the smallest diff...so that's what I went for.

systemcrash:
@hnyman the alternative is to grep through the po files and modify the anchor string there to match the commit. But this frequently upsets weblate until one of us goes and fixes it :O 

To be perfectly honest, capital letters stick out more - and cause increased cognitive load. Capitalising the first item is preferred since it reads more smoothly, and capitalising each word is often done when you're introducing an acronym. There are mixed styles in the code base and both are OK.

Alphix:
> To be perfectly honest, capital letters stick out more - and cause increased cognitive load. Capitalising the first item is preferred since it reads more smoothly, and capitalising each word is often done when you're introducing an acronym. There are mixed styles in the code base and both are OK.

I am fine with either approach, but I think the reverse (i.e. a non-"Title Capitalization" style patch) would be a bigger diff (and we recently changed some other places to use "Title Captialization"). Also, right now we have different styles in the very same tables, which looks weird...so I still think this should be changed (one way or another)...

systemcrash:
I'll take this for a spin when I'm back. But if you feel up to including the po files for the a->A change, then we retain some i18n.

Alphix:
> I'll take this for a spin when I'm back. But if you feel up to including the po files for the a->A change, then we retain some i18n.

Yeah, there's no hurry. I'll have a look at whether I can fiddle with the `.po` files.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -36,13 +36,13 @@ const callUfpList = rpc.declare({
...
-			E('h4', _('Active DHCP Leases')),
+			E('h4', _('Active DHCPv4 Leases')),
 			E('table', { 'id': 'lease_status_table', 'class': 'table' }, [
 				E('tr', { 'class': 'tr table-titles' }, [
 					E('th', { 'class': 'th' }, _('Hostname')),
-					E('th', { 'class': 'th' }, _('IPv4 address')),
-					E('th', { 'class': 'th' }, _('MAC address')),
-					E('th', { 'class': 'th' }, _('Lease time remaining'))
+					E('th', { 'class': 'th' }, _('IPv4 Address')),
+					E('th', { 'class': 'th' }, _('MAC Address')),
```
systemcrash: let's skip these a->A mods, and just stick to the meat of this change.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -36,13 +36,13 @@ const callUfpList = rpc.declare({
 const CBILeaseStatus = form.DummyValue.extend({
 	renderWidget(section_id, option_id, cfgvalue) {
 		return E([
-			E('h4', _('Active DHCP Leases')),
+			E('h4', _('Active DHCPv4 Leases')),
```
systemcrash, line 39: We should also update 40_dhcp.js in luci-mod-status so we're consistent (which you've done), and this one we should sed the .po files so the anchor strings don't orphan the i18n efforts.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -59,10 +59,10 @@ const CBILease6Status = form.DummyValue.extend({
 			E('table', { 'id': 'lease6_status_table', 'class': 'table' }, [
 				E('tr', { 'class': 'tr table-titles' }, [
 					E('th', { 'class': 'th' }, _('Hostname')),
-					E('th', { 'class': 'th' }, _('IPv6 addresses')),
+					E('th', { 'class': 'th' }, _('IPv6 Addresses')),
```
systemcrash: skip this one...

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -781,7 +781,7 @@ return view.extend({
 			so.value(mac, hint ? '%s (%s)'.format(mac, hint) : mac);
 		});
 
-		so = ss.option(form.Value, 'ip', _('IPv4 address'), _('The IPv4 address for this host, or <em>ignore</em> to ignore DHCP requests from this host.'));
+		so = ss.option(form.Value, 'ip', _('IPv4 Address'), _('The IPv4 address for this host, or <em>ignore</em> to ignore DHCP requests from this host.'));
```
systemcrash: skip this one...

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
```diff
@@ -819,7 +819,7 @@ return view.extend({
 		});
 
 		so = ss.option(form.Value, 'leasetime',
-			_('Lease time'),
+			_('Lease Time'),
```
systemcrash: skip this one...

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
Alphix: Not sure what you want me to do? Here the original string actually _does_ change (not just capitalisation), so translations _should_ be invalidated? What should I do with the `.po` files?

Alphix:
@systemcrash I think I've addressed all your comments but one :)

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js
systemcrash: Normally we run `i18n-sync.sh` when strings change. But the gnu tools will discard existing i18ns for strings dropped (`Active DHCP Leases` => `blah blah`) and add new ones (`Active DHCPv4 Leases` => ''), when we can retain the i18ns and translators just add the v4 bit on the next pass in their own languages. So we retain the older i18n which is a closer and more useful match in the short-mid term than ''.

systemcrash: See 875c2a958f4e72137860f594398f7b520f9947ce

Alphix: But then the translators won't ever know that the title changed, will they?

PR #8134, opened 2025-12-03
[24.10] luci-mod-network: Solve the problem of no address assignment with multiple interfaces enabled for DHCP

laipeng668:
Due to missing configuration options, when the DHCP service is enabled on multiple interfaces, it cannot assign addresses to clients under specific conditions. These options are now being added to resolve the issue.


Alphix:
This is only relevant if you're using `odhcpd` as your DHCPv4 server, right?

And I should add that this has been fixed in `master`: https://github.com/openwrt/luci/blob/7df4840b1c077acef60a6a00e8477f686c22b882/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L694

laipeng668:
> This is only relevant if you're using `odhcpd` as your DHCPv4 server, right?
> 
> And I should add that this has been fixed in `master`:

Yes, the issue has already been fixed in the `master` branch, and all non-master branches have this problem.
Unfortunately, except for `24.10` branch, the other branches are no longer maintained.

systemcrash:
Please cherry pick from master.

PR #8146, opened 2025-12-12
luci-mod-status: show DHCPv6 stats as tooltip

eamonxg:
In the previous view, the DHCPv6 statistics displayed in the IPv6 Upstream interface boxes contained many fields, which made the boxes very tall. This stretched the adjacent IPv4 Upstream boxes and left empty space in the middle.

*Previous view: all DHCPv6 stats visible, boxes height increased.*
image

---

I suggest keeping the DHCPv6 statistics in tooltips. This way, the box content remains compact, while users can still view the details on hover.

*Current view: DHCPv6 stats shown as tooltips, box body stays concise.*
image

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/30_network.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/30_network.js
@@ -3,6 +3,7 @@
 'require fs';
 'require network';
 'require rpc';
+'require ui';
 
 
 /* returns per odhcp6c active interface JSON like:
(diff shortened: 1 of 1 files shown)


eamonxg:
@systemcrash Do you think we could improve how this part is presented?

systemcrash:
This idea is OK - but it defeats the purpose of 'overview'. Information is hidden and requires actuation from the user to reveal it. But there's no easy way to make it configurable unless we use the recently added 'hide' functionality. The ports overview uses tooltips, so I guess this is fine.

systemcrash:
All set. Thanks @eamonxg

eamonxg:
> This idea is OK - but it defeats the purpose of 'overview'. Information is hidden and requires actuation from the user to reveal it. But there's no easy way to make it configurable unless we use the recently added 'hide' functionality. The ports overview uses tooltips, so I guess this is fine.

That makes sense. Perhaps we could compute a single aggregated DHCPv6 statistics packet total for the overview, and then reveal the detailed breakdown on hover — similar to how it’s handled in the Port Status ifacebox. This way the overview stays concise while still keeping the details accessible.

systemcrash:
That could be an improvement.

PR #8155, opened 2025-12-14
luci-mod-network: update dhcpv6_pd_min_len values

Alphix:
odhcpd will allow a greater range of PD prefix lengths, so update the UI to reflect this.

https://github.com/openwrt/odhcpd/pull/359


Alphix:
The question is if we should go for:

```
so.placeholder='62';
```

or:
```
so.default='62';
so.optional=true;
```

The disadvantage of the latter is that it'll remove the option if the user sets "62" (and then it'll change if we change the default in `odhcpd`)....

systemcrash:
the latter with placeholder. we try and synch changes anyway.

Alphix:
> the latter with placeholder. we try and synch changes anyway.

Instructions unclear. Segfault. Should I go for `default`, `optional` and `placeholder` or...what?

systemcrash:
as it is now :)

Alphix:
@systemcrash ok, and the PD change is in odhcpd master and 25.12 now, so should be ok to go in?

systemcrash:
Is that commit in openwrt/openwrt yet?

GoetzGoerisch:
Yes https://github.com/openwrt/openwrt/commit/3b58539817a99dc940cd2bb337ad61896a291aa9

GoetzGoerisch:
This needs to be backported to 25.12 if I'm not mistaken?

systemcrash:
picked. Thanks :)

Issue #8160, opened 2025-12-16
Feature request: Content Security Policy


josteink:
### What would you like to see in luci?

The router is one of the most security-sensitive devices I have on my network. If that get's hijacked, all else will fall.

Therefore it is incredibly important to do defense in depth, and one incredibly powerful security measure is yet not implemented by OpenWRT & Luci: [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP).

Without a Content Security Policy, a browser will allow any script from any source to do anything it want inside the active browser window. This means the smallest kind of JS-error may result in complete takeover of the browser-session.

A Content Security Policy header serves to tell the browser "only these things are allowed, and only from these sources", effectively preventing 99% of browser-based attacks. And it's really easy to implement.

Doing some simple experimentation in the browser, I've added this HTML meta-tag after the fact, and it seems to work on the pages I've tested:

```html
<meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' 'trusted-types-eval'; img-src 'self' data:; style-src 'self' 'unsafe-inline';">
```

This CSP effectively limits the Luci web-app to only allow:

* scripts from its own origin/IP, and nowhere else.
* css from its own origin/IPv
* images from its own origin/IP or data: URLs, and nowhere else.
* fetch/etc from own origin/IP, and nowhere else.
* javascript to use inline scripts, eval() and trusted types.

I think this represents a serious security hardening, especially given how little effort it takes to implement.

This is obviously best inserted as a HTTP response header, rather than a HTML meta-tag, but this was what I was able to experiment with, without the need for fully building Luci from source :) 

Could this perhaps make it for 25.12 ? 😁

Ramon00:
Sounds like a really good idea, why dont you make a PR for this?

josteink:
I haven' really done much lua before, apart from nginx-config files, but just from grepping the source, I can see the entry point is `http.uc`.

I'd guess something like the following patch would be sufficient, but I wouldn't know how to test it:

```diff
$ git diff
diff --git a/modules/luci-base/ucode/http.uc b/modules/luci-base/ucode/http.uc
index e7f64ae6e9..0beffbc8e6 100644
--- a/modules/luci-base/ucode/http.uc
+++ b/modules/luci-base/ucode/http.uc
@@ -504,6 +504,9 @@ const Class = {
[...]

PR #8161, opened 2025-12-16
luci-base: http.uc: add CSP headers to http response

josteink:
- [X] Tested on: (Atheros AR9132 rev 2, 25.12.0-RC1, Chromium) :white_check_mark:
- [X] \( Preferred ) Mention: @josteink
image
image

- [X] \( Optional ) Closes: e.g. openwrt/luci#8160

This PR adds support for Content Security Policy to Luci Web.

Content Security Policy is a very effective security hardening for browser-based applications:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP

CSP for OpenWRT/Luci is implemented in 4 different modes:

- None (Disabled - least secure) - default
- Permissive (More secure. Should permit extensive customization)
- Strict (Most secure. Only allow what is required by OpenWRT/Luci by default) - block everything else
- Custom - allow user to define own CSP policy

This option is stored in `/etc/config/uhttpd` and also manageable via the Luci Web UI under "System" -> "Administration" -> "uHTTPd" -> "HTTP(s) Acces" -> "General settings".

### DIFF
--- a/applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
+++ b/applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
@@ -96,6 +96,28 @@ return view.extend({
 		o.default = o.enabled;
 		o.rmempty = false;
 
+		const csp_mode_option = s.taboption('general', form.RichListValue, "csp_mode", _('Content-Security-Policy'), _('Configure CSP headers to improve security.'));
+		csp_mode_option.value('none', _('None (default)'), _('Least secure. CSP disabled.'));
+		csp_mode_option.value('strict', _('Strict'), _('Most secure setting compatible with OpenWRT default installs.'));
+		csp_mode_option.value('permissive', _('Permissive'), _('Less secure than Strict, but better than None.<br>Use with integrations incompatible with Strict.'));
+		csp_mode_option.value('custom', _('Custom'), _('For experts only.'));
+		csp_mode_option.default = 'none';
+		csp_mode_option.rmempty = false;
+
+		const csp_policy_option = s.taboption('general', form.Value, 'csp_policy', _('Custom CSP Policy String'), _('The Content-Security-Policy header-value used in custom-mode.') + "<br />" + _(' WARNING: Wrong values for this setting can render the web-UI inaccessible and require recovery by SSH.'));
+		csp_policy_option.default = "default-src 'none'; script-src 'self' 'unsafe-inline' 'unsafe-eval' 'trusted-types-eval'; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'; connect-src 'self' https://sysupgrade.openwrt.org;";
+
+		csp_mode_option.onchange = function(ev, section_id, value) {
+			const policy_element = csp_policy_option.getUIElement(section_id);
+			const node = policy_element.node.querySelector('input');
+			const isCustom = value === 'custom';
+			if (isCustom) {
+				node.removeAttribute('readonly', 'readonly');
+			} else {
+				node.setAttribute('readonly', 'readonly');
+			}
+		};
+
 		o = s.taboption('general', form.Flag, 'rfc1918_filter', _('Ignore private IPs on public interface'), _('Prevent access from private (RFC1918) IPs on an interface if it has an public IP address'));
 		o.default = o.enabled;
 		o.rmempty = false;
--- a/modules/luci-base/ucode/http.uc
+++ b/modules/luci-base/ucode/http.uc
@@ -13,6 +13,10 @@ import {
 	stdin, stdout, mkstemp
 } from 'fs';
 
+import { cursor } from 'uci';
+
+let uci = cursor();
+
 // luci.http module scope
 export let HTTP_MAX_CONTENT = 1024*100;		// 100 kB maximum content size
 
@@ -503,6 +507,33 @@ const Class = {
 
 		if (!this.headers?.['x-content-type-options'])
 			this.header('X-Content-Type-Options', 'nosniff');
+
+		if (!this.headers?.['content-security-policy']) {
+			// Read UCI settings
+			let csp_mode = uci.get('uhttpd', 'main', 'csp_mode') || 'none';
+			
+			// Define preset policies
+			let presets = {
+                // TODO: strict: default-src: none
+				'strict': 'default-src \'none\'; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob:; style-src \'self\' \'unsafe-inline\'; connect-src \'self\' https://sysupgrade.openwrt.org;',
+				'permissive': 'default-src \'self\' https://*; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob: https://*; style-src \'self\' \'unsafe-inline\' https://*;',
+			};
+			
+			let csp_policy = null;
+			
+			if (csp_mode == 'custom') {
+				// For custom mode, read the user's policy
+				csp_policy = uci.get('uhttpd', 'main', 'csp_policy');
+			}
+			else if (csp_mode in presets) {
+				// For permissive/strict, use preset
+				csp_policy = presets[csp_mode];
+			}
+			// else mode is 'none' - don't set CSP header
+			
+			if (csp_policy)
+				this.header('Content-Security-Policy', csp_policy);
+		}
 
 		this.output('Status: ');
 		this.output(this.status_code);
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
@@ -14,6 +14,28 @@ return view.extend({
 		o = s.option(form.Flag, 'redirect_https', _('Redirect to HTTPS'), _('Enable automatic redirection of <abbr title="Hypertext Transfer Protocol">HTTP</abbr> requests to <abbr title="Hypertext Transfer Protocol Secure">HTTPS</abbr> port.'));
 		o.rmempty = false;
 
+		const csp_mode_option = s.option(form.RichListValue, "csp_mode", _('Content-Security-Policy'), _('Configure CSP headers to improve security.'));
+		csp_mode_option.value('none', _('None (default)'), _('Least secure. CSP disabled.'));
+		csp_mode_option.value('strict', _('Strict'), _('Most secure setting compatible with OpenWRT default installs.'));
+		csp_mode_option.value('permissive', _('Permissive'), _('Less secure than Strict, but better than None.<br>Use with integrations incompatible with Strict.'));
+		csp_mode_option.value('custom', _('Custom'), _('For experts only.'));
+		csp_mode_option.default = 'none';
+		csp_mode_option.rmempty = false;
+
+		const csp_policy_option = s.option(form.Value, 'csp_policy', _('Custom CSP Policy String'), _('The Content-Security-Policy header-value used in custom-mode.') + "<br />" + _(' WARNING: Wrong values for this setting can render the web-UI inaccessible and require recovery by SSH.'));
+		csp_policy_option.default = "default-src 'none'; script-src 'self' 'unsafe-inline' 'unsafe-eval' 'trusted-types-eval'; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'; connect-src 'self' https://sysupgrade.openwrt.org;";
+
+		csp_mode_option.onchange = function(ev, section_id, value) {
+			const policy_element = csp_policy_option.getUIElement(section_id);
+			const node = policy_element.node.querySelector('input');
+			const isCustom = value === 'custom';
+			if (isCustom) {
+				node.removeAttribute('readonly', 'readonly');
+			} else {
+				node.setAttribute('readonly', 'readonly');
+			}
+		};
+
 		return m.render();
 	}
 });


efahl:
> fetch/etc from own origin/IP, and nowhere else.

I didn't try it, but does this prevent "System -> Backup/Flash" from downloading .img files directly?  How about Attended Sysupgrade, which does a bunch of POST and GET fetches to do its upgrades?

You can see some of this at https://github.com/openwrt/luci/blob/a0531c7822bc928626b801c7f3bbdc49163a25e8/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L116

josteink:
Good catch. Attended sysupgrade fails, and will need adjustment to the CSP headers. I'll add rules as required to ensure it passes.

> Connecting to 'https://sysupgrade.openwrt.org/api/v1/revision/25.12-SNAPSHOT/ath79/generic?1765915536852' violates the following Content Security Policy directive: "default-src 'self'".
> Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback. The action has been blocked.

I'm not sure what to test in System -> Backup/Flash. I don't see anything there causing (or which should cause) HTTP requests anywhere?

efahl:
> I'm not sure what to test in System -> Backup/Flash. I don't see anything there causing (or which should cause) HTTP requests anywhere?

Oh, ok, I didn't dig into it, I thought it did a `.get` when you uploaded the `Flash image...`

josteink:
Adding https://sysupgrade.openwrt.org fixes the default//OOB config, but I see that it will break with custom configurations:

image

I guess probably the simplest, most pragmatic option is just to set connect-src to allow anything. The CSP will still be a major improvement over what is already there (which is no hardening at all).

josteink:
With the testing I've been able to to, it seems like attended sysupgrade now works as well!

efahl:
Yeah, changing the upstream url, or adding one to "Rebuilders" complicates it.  Could the ASU code add exceptions somehow for the duration of its use???

You can use https://sysupgrade.guerra24.net as primary server, or added as a rebuilder, if you felt like testing that...

josteink:
I went for the pragmatic and simple solution: allow fetch from any HTTPS-host. 

It’s not perfect, but it’s still clearly a major improvement over what is there today due to the restricted script and default permissions. 

In absence of any other areas to validate/test, I don’t plan any other updates to this PR.

efahl:
Hmm, maybe another one.  I have luci-app-statistics installed on one of my APs, and am getting this when I try to view the generated images:

```
Content-Security-Policy: The page’s settings blocked the loading of a resource (img-src) at blob:https://wap01/0bd0c7d6-fc6c-4917-a773-8847d822c46e because it violates the following directive: “img-src 'self' data:”
```
Hacked it up to contain `blob:` and it works:
```
507                 if (!this.headers?.['Content-Security-Policy'])
508                         this.header('Content-Security-Policy',
509                         "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' 'trusted-types-eval';"
510                         + "img-src 'self' data: blob:;"
511                         + "style-src 'self' 'unsafe-inline';"
512                         + "connect-src 'self' https://*;");
```

efahl:
And firefox says
```
Content-Security-Policy: Couldn’t parse invalid host 'trusted-types-eval'
```
But how do browsers deal with these "bad" ones?  Just ignore them?  In which case, who cares if it "errors"?

That one specifically seems to have very narrow support, but if it is supported, seems like it should be there.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src#browser_compatibility

systemcrash:
For reasons you're now discovering, this is why we've not bothered with extended security measures like this, since most of the customisation people do or use depends on the freedom to do so. There are perhaps only a few apps which call to external resources like this, and the sources are entirely verifiable. Introducing this change will likely break some functionality somewhere... :/

josteink:
> Hacked it up to contain blob: and it works:

Nice. That sounds like a reasonable addition as well. 

> But how do browsers deal with these "bad" ones? Just ignore them? In which case, who cares if it "errors"?

They ignore them. They can’t apply rules/restrictions they don’t know about. 

This one in particular is required by Safari and does no harm elsewhere as far as I’ve tested.

josteink:
> For reasons you're now discovering, this is why we've not bothered with extended security measures like this, since most of the customisation people do or use depends on the freedom to do so.

These are IMO simple issues to work around or loosen up. 

If people are qualified to customize these things, they’re probably qualified to update or remove this header too? 

> There are perhaps only a few apps which call to external resources like this

People might get malicious browser extensions which wants to inject scripts into your router interface.

This would help protect against that to a certain extent. 

> and the sources are entirely verifiable. Introducing this change will likely break some functionality somewhere... :/

IMO script-src is the most important one to get in and the rest can be loosened where there’s any plausible reason for doing so.

josteink:
From a brainstorming POV, would it be possible to implement this as a luci-app? luci-app-csp?

Then those who want it can install it and it could contain a config-pane (with these defaults) which injects the headers into `http.uc`?

Ramon00:
It could be made configurable indeed. Add a option somewhere maybe in the system menu or something, different degrees of allowing things. Add a help text to explain.

mdevolde:
> could contain a config-pane

Or just a text area to write your CSP from start to finish, which is injected into `http.uc` (this way, anyone who wants to apply a CSP can completely customise their CSP according to their config).

So, an interface to do it properly is available, without having the end user tinkering with ucode files, but the CSP must be written by the end user.

So, nothing broken by installing the app.

josteink:
> Or just a text area to write your CSP from start to finish, which is injected into http.uc (this way, anyone who wants to apply a CSP can completely customise their CSP according to their config).

This sounds more viable from a practical point of view, and for my first Luci submission 😄 

Provide a reasonable default, and allow whatever the user wants. If non-empty, use as a CSP headers. If empty, don't provide a CSP header.

I think that's OK for a version 1 implementation, isn't it?

Ramon00:
> I think that's OK for a version 1 implementation, isn't it?

Depends on the default, if the default is empty then this is NOK (people would not know what to fill in).

I do like the idea of user defined. 

I would still go for drop down, 1) none 2) loose 3) tight 4) user defined. {previewing the new header) This is most user friendly and will serve everybody. I'm guessing 99% of the people will use 1), 2) or 3).

Do not choose for the easy solution just because you do not have the experience to implement, there are plenty of people willing to help out.

I do wonder what the default should be, 1) or 2) or 3)...

josteink:
I've added and pushed a first iteration of using UCI for settings and a minimal UI for it.

1. checkbox to enable/disable
2. default policy as derived from the discussion in this PR.

Do we have any good examples in the luci interface of a dropdown which automatically populates the value of an accompagnying text-field?

modules/luci-base/ucode/http.uc
```diff
@@ -13,6 +13,8 @@ import {
 	stdin, stdout, mkstemp
 } from 'fs';
 
+import { cursor } from 'uci';
```
mdevolde, line 16: It seems that the convention is to define the cursor here, and not in the function. ([ref](https://github.com/openwrt/luci/blob/a0531c7822bc928626b801c7f3bbdc49163a25e8/modules/luci-base/ucode/dispatcher.uc#L17))
```suggestion
import { cursor } from 'uci';

let uci = cursor();
```

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +506,24 @@ const Class = {
 		if (!this.headers?.['x-content-type-options'])
 			this.header('X-Content-Type-Options', 'nosniff');
 
+		if (!this.headers?.['Content-Security-Policy']) {
+			// Read UCI settings - default to enabled with default policy
+			let uci_cursor = cursor();
```
mdevolde: If the cursor is set above, there is no need to set it here.
This suggestion should be applied only if the previous one is applied.
```suggestion
```

Ramon00:
> Do we have any good examples in the luci interface of a dropdown which automatically populates the value of an accompagnying text-field?

Yes there should be plenty around, for example look at luci-app-usteer which I made.

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +506,24 @@ const Class = {
...
+			let enable_csp = uci_cursor.get('uhttpd', 'main', 'enable_csp');
```
systemcrash: Fall back to `0` here, since this emulates behaviour as it is today. 
```js
let enable_csp = uci_cursor.get('uhttpd', 'main', 'enable_csp') || 0;
```

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +506,24 @@ const Class = {
...
+			
+			// Default to enabled ('1') if setting doesn't exist (backward compatibility)
+			if (enable_csp == null || enable_csp == '1') {
```
systemcrash:
```js
if (enable_csp) {
```

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
```diff
@@ -96,6 +96,15 @@ return view.extend({
 		o.default = o.enabled;
 		o.rmempty = false;
 
+		o = s.taboption('general', form.Flag, 'enable_csp', _('Enable Content-Security-Policy'), _('Add CSP headers to HTTP responses to improve security. Disabling may be needed for custom integrations.'));
+		o.default = o.enabled;
+		o.rmempty = false;
+
+		o = s.taboption('general', form.Value, 'csp_policy', _('CSP Policy'), _('Custom Content-Security-Policy header value. Leave empty to use the default policy.'));
```
systemcrash: Use a textbox.

modules/luci-base/ucode/http.uc
efahl: Better yet, use `get_bool`, which handles all the weird cases for you:

https://openwrt.github.io/luci/jsapi/LuCI.uci.html#get_bool

systemcrash: Don’t. This is ucode, not uci 🙃

efahl: BLAH  Again my pre-coffee brane fails me by translating uc to js.  I have a patch to ucode-mod-uci that adds `get_bool`, just need to submit the PR...

Ok, here ya go https://github.com/jow-/ucode/pull/356

systemcrash: Pretty good!

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
josteink: This is the first time I'm trying to build something in the Luci UI. From what I can tell, this is how most textboxes in the `uhttpd.js` file is defined.

Could you be more concrete wrt to what you think should be "textboxed" here?

modules/luci-base/ucode/http.uc
josteink: I tried this but it doesn't work because `enable_csp` when disabled (via UI) is `"0"` (a string) which evaluates to a truthy value. Checking against `"1"` does not have these problems.

modules/luci-base/ucode/http.uc
josteink: Fixed as originally suggested by @systemcrash .

josteink:
The formality check fails because the second commit starts with package name `luci-app-uhttpd:` which seemingly is wrong.

However, this is the package name used in all previous git commits on this file, and also matches the root folder it's in.

Help anyone? 😄 

**Edit:** I see the formality checks are a fairly new thing. Is this an error with my PR or with the formality check?

hnyman:
> The formality check fails because the second commit starts with package name luci-app-uhttpd:

It actually starts with a space before "luci-app-uhttpd", if you closely to the check results.

```
Checking subject:
 luci-app-uhttpd: add configuration options for CSP
[fail] Commit subject line MUST start with `<package name>: `
```
Visible difference to the succeeding commit:
```
Checking subject:
luci-base: http.uc: add CSP HTTP response headers
[pass] Commit subject line format seems OK
```

GeorgeSapkin:
> It actually starts with a space before "luci-app-uhttpd", if you closely to the check results.

Haven't seen this one before. I guess this is something to check for.

josteink:
> It actually starts with a space before "luci-app-uhttpd", if you closely to the check results.

Got it. Fixed now!

> @Ramon00 wrote:
>
> I would still go for drop down, 1) none 2) loose 3) tight 4) user defined. {previewing the new header) This is most user friendly and will serve everybody. I'm guessing 99% of the people will use 1), 2) or 3).

I instead went with "None", "Permissive", "Strict" & "Custom", but now there's dropdowns there 😄 

I'll test more thoroughly tomorrow to ensure there's no stupid bugs, but any comments on the "form" of the UI solution would be appreciated.

Ramon00:
> I instead went with "None", "Permissive", "Strict" & "Custom", but now there's dropdowns there 😄

Close enough 😄

But question: strict, does that block everything? I mean I probably would not mind having a default that blocks everyting, if i need to upload a new firmware image or somthing, i can always lower it temporarily, you know.

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
```diff
@@ -96,6 +96,52 @@ return view.extend({
...
+
+			// For custom mode, return stored value or empty
+			if (mode === 'custom') {
+				return stored_policy || '';
+			} else {
+				return '';
+			}
+		};
+
+		// Only save to UCI when in custom mode
```
Ramon00: This is a choice here. If you often switch between some custom and e.g. "permissive", then probably you would want to have the custom stored, even if it is currently not selected, such that with 1 click you can go back to the custom one. Otherwise you need to store it and copy/past it in. Not sure how likely that scenario is though

josteink:
> But question: strict, does that block everything? I mean I probably would not mind having a default that blocks everyting, if i need to upload a new firmware image or somthing, i can always lower it temporarily, you know.

Everything works. The intent is for it to be as strict as it can be without breaking any **default** OpenWRT/Luci functionality.

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
josteink: I agree that increases chances of user error. Probably not a good design. 

I’ll change it.

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +506,33 @@ const Class = {
 		if (!this.headers?.['x-content-type-options'])
 			this.header('X-Content-Type-Options', 'nosniff');
 
+		if (!this.headers?.['Content-Security-Policy']) {
+			// Read UCI settings
+			let uci_cursor = cursor();
+			let csp_mode = uci_cursor.get('uhttpd', 'main', 'csp_mode') || 'none';
```
mdevolde: Use the new defined cursor.
This suggestion should be applied only if the previous ones are applied.
```suggestion
			let csp_mode = uci.get('uhttpd', 'main', 'csp_mode') || 'none';
```

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +506,33 @@ const Class = {
...
+			let presets = {
+				'strict': 'default-src \'self\'; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob:; style-src \'self\' \'unsafe-inline\'; connect-src \'self\' https://sysupgrade.openwrt.org;',
+				'permissive': 'default-src \'self\' https://*; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob: https://*; style-src \'self\' \'unsafe-inline\' https://*;',
+			};
+			
+			let csp_policy = null;
+			
+			if (csp_mode == 'custom') {
+				// For custom mode, read the user's policy
+				csp_policy = uci_cursor.get('uhttpd', 'main', 'csp_policy');
```
mdevolde: Use the new defined cursor.
This suggestion should be applied only if the previous ones are applied.
```suggestion
				csp_policy = uci.get('uhttpd', 'main', 'csp_policy');
```

Ramon00:
> Everything works. The intent is for it to be as strict as it can be without breaking any **default** OpenWRT/Luci functionality.

I get that, but there are uses for a "fort knox" setting that may break some things, like new fw upload, but seeing as I do not upload new fw very often, it is not that much work to switch it to a lower setting to upload a new fw. However, with the custom option, i can always fill it in, so im fine either way.

josteink:
> I get that, but there are uses for a "fort knox" setting that may break some things, like new fw upload, but seeing as I do not upload new fw very often, it is not that much work to switch it to a lower setting to upload a new fw. However, with the custom option, i can always fill it in, so im fine either way.

CSP is a browser-only technology. Forbidding firmware-uploads from the main luci web interface might still permit it from other origins unless it is explicitly forbidden on the server-side. 

If the idea is to implement a «fort Knox» kind of security regime it would be better to start with a server-side lockdown if conditionally enabled functionality. 

That would be a way bigger change CSP and way outside the scope of this PR 😄

Ramon00:
> CSP is a browser-only technology. Forbidding firmware-uploads from the main luci web interface might still permit it from other origins unless it is explicitly forbidden on the server-side.

nevermind, you do not get what i mean

modules/luci-base/ucode/http.uc
josteink: Agreed. Looks better!

modules/luci-base/ucode/http.uc
josteink: Same here.

modules/luci-base/ucode/http.uc
josteink: Same here.

modules/luci-base/ucode/http.uc
josteink: Same here.

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
josteink: I've changed it now to not discard the value for the custom-header.

However I was not able to hide the text-field when not in custom-mode and at the same time retain the setting-value. I therefore resorted to keeping it always visisble, but made it read-only when not in custom-mode.

 Does that look like a better solution for you?

josteink:
Latest changes used RichListValue and enables some more descriptive content for those not 100% up to speed on CSP.

image

Looking pretty good, if I have to say so myself!

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
```diff
@@ -96,6 +96,28 @@ return view.extend({
...
+	    var csp_mode_option = s.taboption('general', form.RichListValue, "csp_mode", _('Content-Security-Policy'), _('Configure CSP headers to improve security.'));
+	    csp_mode_option.value('none', _('None (disabled)'), _('Least secure.'));
+		csp_mode_option.value('permissive', _('Permissive'), _('More secure. Might be required for some integrations.'));
+		csp_mode_option.value('strict', _('Strict'), _('Most secure setting compatible with OpenWRT default settings.'));
+		csp_mode_option.value('custom', _('Custom'), _('For experts only.'));
+		csp_mode_option.default = 'none';
+		csp_mode_option.rmempty = false;
+
+		var csp_policy_option = s.taboption('general', form.Value, 'csp_policy', _('Custom CSP Policy String'), _('The Content-Security-Policy header-value used in custom-mode.') + "<br />" + _(' WARNING: Wrong values for this setting can render the web-UI inaccessible and require recovery by SSH.'));
+		csp_policy_option.default = 'default-src \'none\'; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob:; style-src \'self\' \'unsafe-inline\'; connect-src \'self\' https://sysupgrade.openwrt.org;'
```
mdevolde: @systemcrash Wouldn't it be more appropriate, for here and for `http.uc`, to create a dedicated section in `/etc/config/uhttpd` where we put the different CSP default values, rather than hardcoding them in the js and ucode?

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +508,33 @@ const Class = {
...
 			this.header('X-Content-Type-Options', 'nosniff');
 
+		if (!this.headers?.['Content-Security-Policy']) {
+			// Read UCI settings
+			let csp_mode = uci.get('uhttpd', 'main', 'csp_mode') || 'none';
+			
+			// Define preset policies
+			let presets = {
+                // TODO: strict: default-src: none
+				'strict': 'default-src \'none\'; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob:; style-src \'self\' \'unsafe-inline\'; connect-src \'self\' https://sysupgrade.openwrt.org;',
```
mdevolde, line 518: Same question as above.

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +508,33 @@ const Class = {
...
+				'permissive': 'default-src \'self\' https://*; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob: https://*; style-src \'self\' \'unsafe-inline\' https://*;',
```
mdevolde, line 519: Same question as above.

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
```diff
@@ -96,6 +96,28 @@ return view.extend({
 		o.default = o.enabled;
 		o.rmempty = false;
 
+	    var csp_mode_option = s.taboption('general', form.RichListValue, "csp_mode", _('Content-Security-Policy'), _('Configure CSP headers to improve security.'));
+	    csp_mode_option.value('none', _('None (disabled)'), _('Least secure.'));
```
GeorgeSapkin: Mixing tabs and spaces here and few other places.

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
```diff
@@ -96,6 +96,28 @@ return view.extend({
...
+		csp_mode_option.value('custom', _('Custom'), _('For experts only.'));
+		csp_mode_option.default = 'none';
+		csp_mode_option.rmempty = false;
+
+		var csp_policy_option = s.taboption('general', form.Value, 'csp_policy', _('Custom CSP Policy String'), _('The Content-Security-Policy header-value used in custom-mode.') + "<br />" + _(' WARNING: Wrong values for this setting can render the web-UI inaccessible and require recovery by SSH.'));
+		csp_policy_option.default = 'default-src \'none\'; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob:; style-src \'self\' \'unsafe-inline\'; connect-src \'self\' https://sysupgrade.openwrt.org;'
+
+		csp_mode_option.onchange = function(ev, section_id, value) {
+			var policy_element = csp_policy_option.getUIElement(section_id);
+            var node = policy_element.node.querySelector('input');
```
GeorgeSapkin: What's the reason for sticking with `var` instead of `const`/`let`?

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
josteink: Yes and no?

On the one hand I'd like to avoid code-repetition, embrace modularity and extensibility. That talks in favour of your suggestions.

On the other hand I'd like to make it simple to use. Not just now, but over time.

I think "strict" and "permissive" are tightly connected to the OpenWRT codebase and release itself. What is possible and/or required for "strict" may change over time. If we store this in `/etc/config/uhttpd` somewhere, user's will typically choose to preserve these settings when doing upgrades.

If the default values for those presets have changed between releases, that will be a upgrade which can fail in unepexted/undesirable ways. Keeping these presets fixed in code avoids that.

And I think that is a bigger problem/risk than these presets being hardcoded in code. That's an argument for keeping things as is.

I'm leaning more towards that direction, but I'm not 100% set on this if someone can explain a way we can make it "upgrade-safe".

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
josteink: It seemed to be a lot of `vars` there already so I just stayed with the pattern. If it's OK to go `let` and `const` I have no issues with that :)

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
josteink: My editor is giving me a hard time editing tab-based files here. I thought I had those cleaned up, but clearly not.

applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js
mdevolde: @josteink I don't understand your concerns. It doesn't change anything from the user's point of view, the strings will just be stored in the `httpd` config file, and not in the code any more.
It will be easier for developers to maintain, and it doesn't change anything from the user's point of view.
The section containing them will not be editable from JS.
I am waiting for the opinion of a maintainer, in case there are any contraindications.

josteink: Just chiming in with a concrete example while awaiting the @systemcrash 's opinion:

1. The user installs OpenWRT release with CSP support (lets say 25.12.1)
2. CSP presets/profiles are stored in `/etc/config/uhttpd` - based on that release
3. A new Luci/OpenWRT release comes out later (lets says 26.10.0), with some different JS in the browser which requires permissions not granted by the existing CSP preset/profiles. We update the profiles as well to ensure it works.
4. User upgrades to 26.12.0 and checks the option "keep existing config".
5. He will get the config file from 25.12.1 with the old presets, but the Luci code which requires the 26.10.0 presets.
6. Luci may fail to load as a result. User may be unhappy. Depending on his ability to SSH in and edit the relevant config files, he may just say the upgrade "bricked" his router.

That's the scenario I am worried about.

If I'm worrying about something which is trivial to solve, by all means, I'm happy to go that route. Just let me know how!

Just in case it's not obvious: This is my first OpenWRT contribution and I'm fairly new to the source-code 😄

mdevolde: @josteink I understand your concern better now, I don't know if there's a way to remedy this.
What's unfortunate at the moment, from a maintainability standpoint, is the repetition of the "strict" CSP in JS and ucode, as well as their decentralisation (not all written CSP at the same place).

mdevolde: Maybe a script in `/etc/uci-defaults/` could be a solution, I'm curious to hear a maintainer's opinion on this matter.

josteink: Custom is going to be custom, no matter what, so that's obviously something which should go in `/etc/config/uhttpd` as it does now.

>  is the repetition of the "strict" CSP in JS and ucode

Technically speaking the CSP in the JS is not "strict" but a default-value for "custom", as a convenience, based on "strict". I thought it would be a "nice-to-have" sort of thing, but if it makes things complicated/decentralized, we could just remove the default as well, leaving the only 2 profiles in the ucode part of the codebase.

Either way I'm curious to hear about other, possibly better solutions.

mdevolde: @systemcrash ping

systemcrash:
The problems I see at the moment are as follows:
* settings in an app that might not be installed

This potentially introduces new behaviour, yet does not surface a control element. The behaviour is in base, but the config element is in a separate app. luci-app-uhttpd seems like a good place for it. But luci-app-uhttpd configures uhttpd behaviour. This change is about controlling the front-end behaviour. It seems to me the better placement would be under system->admin->http, which is about the only decent place in base.

This addition correctly 'fails open' - if there is no config present, it remains in a disabled state. You might want to flag that in the GUI - that disabled is the default state.

It also might be a bit inconsistent, but use " to wrap the string, and ' to quote the elements in it.

What settings can render the UI inaccessible?

josteink:
> The problems I see at the moment are as follows:
> 
> * settings in an app that might not be installed

I may be misunderstanding something, but the file I modified (uhttpd.js) for the settings is where these kind of HTTP controls used to be already.

> This potentially introduces new behaviour, yet does not surface a control element. 

This only introduces a new behaviour, if configured. What control elements are you looking for, beside the UI provided?

> The behaviour is in base, but the config element is in a separate app. luci-app-uhttpd seems like a good place for it. But luci-app-uhttpd configures uhttpd behaviour. This change is about controlling the front-end behaviour. It seems to me the better placement would be under system->admin->http, which is about the only decent place in base.

The setting you are referring to seems (to my untrained eye) to be uhttpd.js. Check the URL for the admin-page you are referring to:

image

When grepping the source for "Redirect to HTTPS", this setting is only found in `uhttpd.js`.

```
% git grep "Redirect all HTTP to HTTPS" | grep .js
applications/luci-app-uhttpd/htdocs/luci-static/resources/view/uhttpd/uhttpd.js:		o = s.taboption('general', form.Flag, 'redirect_https', _('Redirect all HTTP to HTTPS'));
````

> This addition correctly 'fails open' - if there is no config present, it remains in a disabled state. You might want to flag that in the GUI - that disabled is the default state.

Good idea. I'll make it so.

> It also might be a bit inconsistent, but use " to wrap the string, and ' to quote the elements in it.

Fine by me :smile:

> What settings can render the UI inaccessible?

Only setting which can render the UI inaccessible is "custom" with a bad policy which excludes required JS scripts, in the area clearly marked as "experts only", and with a safe default provided.

**Edit:** I see my grep was incorrect and there is a different module I can (should?) insert these changes into.

I'll make it so too!

@systemcrash : I assume `modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js` is the right file then, to avoid dependency on non-default app? :) 

Or should I keep it in both?

**Comment by josteink** [comment]
All comments should in theory be resolved? Anything else I should change? 😄

**Comment by systemcrash** [comment]
If you're aware of any settings already which might be problematic, we might surface a time limited warning when those are entered. I propose a `.validate` that returns true, but raises a `ui.addTimeLimitedNotification(...)`, if you know a combo which can do this. 

I suggest that such a long string warrants a textbox, so everything is visible at once. 

If the setting is available in both places (app and mod system), that's OK.

**Comment by josteink** [comment]
> If you're aware of any settings already which might be problematic, we might surface a time limited warning when those are entered.

No settings are by themselves "problematic" more so than the others.

* None. "Problematic" because it offers no browser-based security against external abuse, and you don't want your router to be insecure.
* Permissive. Works with any default OpenWRT install. Could be seen as "problematic" because it permits more attack-vectors than strictly needed (like custom attended sysupgrade feeds), unless the user has a specific need to loosen up security.
* Strict. Works with any default OpenWRT install. Could be "problematic" in the sense that it works now, but you install some other luci-apps and for whatever reason they don't seem to be working (only showing CSP warnings in the browser console) somewhat obscuring the cause of the failure.  (This is what I will be running 100% of the time on all of my devices)
* Custom. "Problematic" because there's no way for us to know up front if a policy is going to work or not. This will be 100% dependent on the users deployment and customizations.

**Right now, if you do a default OpenWRT install on any device, apply any of the suggested modes (including custom with default value policy) and click "Save & Apply", everything default in OpenWRT will keep on working 100%.** You will not be prevented from entering the same admin-section to change the setting later.

Does that clear up the expected behaviour sufficiently?

>  I propose a `.validate` that returns `true`, but raises a `ui.addTimeLimitedNotification(...)`, if you know a combo which can do this.

The only place where there is a real risk is when using "custom" mode. We cannot validate up front if that will be successful or not.

So I guess that means `.validate()`  should return `true` if current value is `"custom"` ?

**Comment by josteink** [comment]
@systemcrash : I tried the following code to just to test. It works, kinda, but I always get the validate-handler invoked twice on changes, leading to the ui popup to appear twice as well.

```js
		csp_mode_option.validate = function(section_id, value) {
			// triggers UI warning when using custom-mode with UI timer based fallback
			if (value === "custom")
			{
			  ui.addTimeLimitedNotification(_('Danger zone'), E('p', _('Custom CSP header will be used as is. If deemed invalid by the browser, you will need to use SSH to reset the value in /etc/config/uhttpd.')), 60000, 'warning');
			}
			return true;
		};
```

image

Is this a general luci problem, or is it anything else I'm doing wrong? 😄

systemcrash:
No, validate is triggered on the cfgvalue and load and I think another one of the function calls, IIRC. :/

That custom string? I can't see anything. That's why it should be a text box. 

What I mean are policies like the following: 
```
Content-Security-Policy:
  default-src 'none';
  script-src 'none';
  style-src 'none';
  img-src 'none';
  font-src 'none';
  connect-src 'none';
  frame-src 'none';
```

So scanning for the key-word `none` is probably sufficient to trigger a warning.

josteink:
> No, validate is triggered on the cfgvalue and load and I think another one of the function calls, IIRC. :/

Is there a convention somewhere on how to avoid double warnings then, or should I just leave it?

> That custom string? I can't see anything. That's why it should be a text box.

What's the best/right way to do that? I don't seen any `Form.TextBox` type to use here.

> So scanning for the key-word none is probably sufficient to trigger a warning.

To be absoultely clear, the default-value for custom (based on strict) intentionally defaults to `none` unless another policy is set (`default-src 'none')`.

That is, you use the other policies to whitelist only the things you know you need. So this warning will also trigger for the default value.

Also, for instance, for `script-src` it's not enough to just say "from this origin", you also have to allow what type of unsafe behaviours (like `eval`) you want to allow.

Creating a CSP-validator which is meaningful is going to be a much bigger job than just getting the basic settings in there, and it will be 10x more complicated than the actual feature already proposed.

If someone later feels CSP is great, and they want a better interactive editor for that... I would delegate that responsibility to a separate, installable `luci-app`, like we have `luci-app-uhttpd` for exactly the same reasons for more of the "advanced" uhttpd settings.

It is my opinion that this PR, as is, will provide a substantial security improvement to LUCI compared to how it was before.

Let's resolve the above issues, and leave the CSP-editor/validator aspects to eventually be solved later, if there turns out to be a need?

systemcrash:
> Is there a convention somewhere on how to avoid double warnings then, or should I just leave it?

There is, IIRC. I just can't remember it now... let me have a think. 


Ah, sorry - take a look at: https://openwrt.github.io/luci/jsapi/LuCI.form.TextValue.html

There are some in use in the code-base. Take a peek at how others are used. 

> That is, you use the other policies to whitelist only the things you know you need.

I took a look [at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#fetch_directive_syntax) which says:

the single value `'none'`, indicating that the specific resource type should be completely blocked

So we can at least add a note about that in the GUI: e.g. `'none' completely blocks the specific resource type`, and if the user saws off the branch they're standing on, it's on them :)

josteink:
No new changes in latest pull. Have been busy during Christmas as everyone else 😄

Just did a rebase on upstream git master to ensure I'm not walking into any merge-conflicts.

systemcrash:
Finally took this for a spin. I get `Content-Security-Policy: Couldn't parse invalid host 'trusted-types-eval'`. It's not a valid source expression. Let's trim that out.

josteink:
> Finally took this for a spin. I get `Content-Security-Policy: Couldn't parse invalid host 'trusted-types-eval'`. It's not a valid source expression. Let's trim that out.

Thats required in Safari or else it doesn’t work.

In Chrome it just gets ignored because it’s not a supported yet. It doesn’t cause any issues apart from that one warning.

modules/luci-base/ucode/http.uc
```diff
@@ -504,6 +508,33 @@ const Class = {
 		if (!this.headers?.['x-content-type-options'])
 			this.header('X-Content-Type-Options', 'nosniff');
 
+		if (!this.headers?.['Content-Security-Policy']) {
+			// Read UCI settings
```
systemcrash: This will assign it every time - because each of `this.headers` gets squashed to lowercase in the addition function:
```js

	header: function(key, value) {
		this.headers ??= {};
		this.headers[lc(key)] = value;
	},
```

josteink: Good call. I didn't know that. Fixed!

systemcrash:
@josteink how do you feel about making this part of a broader plugin system? Take a look at #8281 and my latest comments there. Basically, various header types shall be handled by separate ucode plugins, whereby CSP is its own plugin. The goal is to have a minimal change delta in http.uc/dispatcher.uc, and that the user can completely disable extra header functionality (by not installing plugins) and move the new uc functionality to e.g. headerplugins.csp.uc, which lives in ucode/headerplugins/csp.uc.

We can make a new plugins/ folder in the repo, and this can live in plugins/http-headers-csp/ucode/headerplugins/csp.uc

Each plugin
- announces which headers it handles in standard camel-case, 
- http.uc loads plugins in the headerplugins path 
- returns its header(s) based on its configuration (done already in uhttpd.js, although a more dynamic approach is better)

So this:

```js
		if (!this.headers?.['content-security-policy']) {
			// Read UCI settings
			let csp_mode = uci.get('uhttpd', 'main', 'csp_mode') || 'none';
			
			// Define preset policies
			let presets = {
                // TODO: strict: default-src: none
				'strict': 'default-src \'none\'; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob:; style-src \'self\' \'unsafe-inline\'; connect-src \'self\' https://sysupgrade.openwrt.org;',
				'permissive': 'default-src \'self\' https://*; script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' \'trusted-types-eval\'; img-src \'self\' data: blob: https://*; style-src \'self\' \'unsafe-inline\' https://*;',
			};
			
			let csp_policy = null;
			
			if (csp_mode == 'custom') {
				// For custom mode, read the user's policy
				csp_policy = uci.get('uhttpd', 'main', 'csp_policy');
			}
			else if (csp_mode in presets) {
				// For permissive/strict, use preset
				csp_policy = presets[csp_mode];
			}
			// else mode is 'none' - don't set CSP header
			
			if (csp_policy)
				this.header('Content-Security-Policy', csp_policy);
		}
```

becomes (pseudocode):
```js
		for (let hp of headerplugins) {
			if(!this.headers.includes(hp.my_headers)) // static method.
				hp.insert_headers(this.headers); // static method.
			// perhaps anything else we have not thought of yet
		}
```

I don't mind putting together a structure and example plugin to give the idea.

josteink:
That example looks simple enough. 

But how do you plan to handle configuration for the various plugins?

As always: the base code is just a few lines of code. Most complexity actually lies in the configuration part. 

Do you have any concrete plans to make sure that not every developer is on his own?

Also: an example plugin could be the current HTTP->HTTPS redirection code also currently in http.uc next to the additions from my PR.

systemcrash:
I'm thinking a structure like: plugins -> class -> type, so here plugins -> http -> headers

And we progressively build the minimal changes where they need to go. Like for headers as discussed above. 

I'm working on a dynamic loading of plugin config pages. Install adds config page, plugin page displays it. Click - done. System (http) checks preference for class (on/off) and type (enabled/disabled) and loads. 

Something like that.

Config I think would go in luci -> plugin (section type).

josteink:
I must admit I've been somewhat fatigued by this whole PR process. This is by no means my first patch and I know different projects have different standards and processes.

But what for me started as a simple 1-line HTTP header patch which improved security 100x has in lack of better words been subject to more bike-shedding than I'm used to, and as a result I've lost most of my motivation for landing it as things are now.

If @systemcrash or someone else lands a plugin system which my use-case fits nicely into I can consider creating a PR compatible with that system.

Or this patch can be considered in the form it is now, and I will be interested to complete it without other major changes.

The choice is yours, I guess.

systemcrash:
@josteink sorry for the wait - but the rudimentary plugin handling is now in master. Might I suggest that in a revised plugin UI, that presets are available as you have it, but that each 'chunk' of the value has its own configurable field?
i.e. `default-src`, and `script-src` etc.

josteink:
Sorry for the long wait. I've been busy across the line.

I've now gotten a spare device, provisioned with OpenWRT SNAPSHOT, so that I can test the plugin system.

Due to how everything is changed (and this PR is with a certain degree of bike shedding)... I'll leave this PR be, and start a new one:

https://github.com/openwrt/luci/pull/8645

systemcrash:
Closed via #8645

PR #8176, opened 2025-12-25
luci-app-statistics: major upgrade

icegood:
1) add ability to log each input plugin in
own interval

2) make configurable which data will be
shown on ui for xpu plugin (system/user/etc)

3) Add smart plugin support (based on what rrdtool could collect) Therefore try to run rrdtool firs and after that configure what attributes to show

4) small fixes in rrdtool converter. Say, DEF/CDEF key size not more 30 symbols

5) Some fixes in common UI: make possible to get list without predefined values. Useful in our case when disk has not attributes collected by rrdtool yet

6) A bit towards fixing of
https://github.com/openwrt/packages/issues/26736
by not making all of these softlinks to work. Appropriate commit to colectd will be
 added. However this one is believed to work independenltly (therefore, no draft)

@jow- 

- [x ] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [ x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [ x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [ x] Incremented :up: any `PKG_VERSION` in the Makefile
- [ x] Tested on: (x86/64 opnwrt series 24) :white_check_mark: 
- [x ] \( Preferred ) Mention: @ the original code author for feedback
- [ x] \( Preferred ) Screenshot or mp4 of changes:
- [ x] \( Optional ) Closes: e.g. openwrt/luci#issue-number
- [ x] \( Optional ) Depends on: e.g. openwrt/packages#pr-number in sister repo
- [ x
stat1
stat2
stat3
stat4
stat5
stat6
stat7
stat8
] Description: (describe the changes proposed in this PR)


applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/cpu.js
```diff
@@ -1,33 +1,40 @@
...
 			_('By setting this, CPU is not aggregate of all processors on the system'));
+		
 		o.default = '1';
+		
 		o.rmempty = false;
 		o.depends('enable', '1');
 
 		o = s.option(form.Flag, 'ReportByState', _('Report by state'),
-			_('When set to true, reports per-state metric (system, user, idle)'));
+			_('When set to true, reports per-state metric e.g. (system, user, idle)'));
```
systemcrash, line 24: drop this change

applications/luci-app-statistics/root/usr/libexec/stat-genconfig
```diff
@@ -32,36 +32,50 @@ const plugins = {
...
+	[ 'Globals' ],
+	[]
+];
 
+function parse_units(ustr) {
 	// unit map
 	const map = {
 		y  : 60 * 60 * 24 * 366,
-		m  : 60 * 60 * 24 * 31,
+		mon  : 60 * 60 * 24 * 31,
```
systemcrash, line 46: This will not always hold true but I guess it's fine since it's already there. You've also redefined the behaviour for `m`, so I think m must be retained as the month calculation.

systemcrash:
@hnyman can you take this change-set for a spin?

hnyman:
I will look into it, but it would much more clear with several commits instead of one, as there are quite unrelated changes.

(And there are also unnecessary whitespace changes & a few wording changes)

One thing that am not sure if I like at the first glance is the redefinition of m to mon. I believe that y m d h are used as time definitions also in other places in uci (DHCP leasetime and some others?) , so having here a different definition would be strange.

modules/luci-base/htdocs/luci-static/resources/form.js
systemcrash, line 1: drop all changes to this file

applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/collectd.js
```diff
@@ -71,21 +79,31 @@ return view.extend({
...
+		var m, s, o, enabled;
+
+		for (var i = 0; i < plugins.length; i++)
+			plugins[plugins[i].name] = plugins[i];
+
+		m = new form.Map('luci_statistics', _('Luci statistics settings'));
+		m.tabbed = true;
+		// omit for a while. don't know why 'config_file' not shown in this tab
+		//this.addGeneralTab(m);
+		this.addCollectdTab(m);
```
systemcrash, line 99: ?

applications/luci-app-statistics/htdocs/luci-static/resources/view/statistics/plugins/smart.js
```diff
@@ -0,0 +1,173 @@
...
+				this.vallist = getLabels([]);
+				return this.super('render', [option_index, section_id, in_table]);
+			}
+			var disk = diskChild.cfgvalue(section_id);
+			// Clear existing values
+			const attrs = await getDiskAttributes(disk, s);
+			this.keylist = attrs;
+			this.vallist = getLabels(attrs);
+			return this.super('render', [option_index, section_id, in_table]);
+		};
```
systemcrash, line 157: what are you trying to do here?

feckert:
@icegood  No response from the PR creator. 
This needs to be split into smaller commits. Such a large commit change cannot be reviewed.

I am closing this. Feel free to reopen it once you make the requested changes.

PR #8187, opened 2026-01-03
luci-app-acme: support --cert-profile option

ngehrsitz:
This makes the new `--cert-profile` option from https://github.com/openwrt/packages/pull/28237 available using Luci.

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
@@ -299,6 +299,10 @@ return view.extend({
 		o.datatype    = 'uinteger';
 		o.modalonly = true;
 
+		o = s.taboption('advanced', form.Value, 'cert_profile', _('Certificate Profile'));
+		o.optional = true;
+		o.modalonly = true;
+
 
 		s = m.section(form.GridSection, '_certificates');
 
--- a/applications/luci-app-acme/po/ar/acme.po
+++ b/applications/luci-app-acme/po/ar/acme.po
@@ -40,7 +40,7 @@ msgstr "إعدادات متقدمة"
 msgid "Alternative DNS API options"
 msgstr ""
 
-#: applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js:419
+#: applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js:423
 msgid "Cancel"
 msgstr ""
 
@@ -48,11 +48,15 @@ msgstr ""
 msgid "Cert files are stored in"
 msgstr "يتم تخزين ملفات الشهادات في"
 
+#: applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js:302
+msgid "Certificate Profile"
+msgstr ""
+
 #: applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js:76
 msgid "Certificate config"
 msgstr "تكوين الشهادة"
 
-#: applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js:537
+#: applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js:541
 msgid "Certificates"
 msgstr "الشهادات"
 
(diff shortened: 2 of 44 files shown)


applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -299,6 +299,10 @@ return view.extend({
 		o.datatype    = 'uinteger';
 		o.modalonly = true;
 
+		o = s.taboption('advanced', form.Value, 'cert_profile', _('Certificate Profile'));
+		o.optional = true;
+		o.modalonly = true;
```
ngehrsitz, line 304: Is there any way to control the order?
grafik
Ideally this would be below the URL, but it doesn´t appear to be the order these are defined in?

systemcrash: order? It doesn't appear.

ngehrsitz: This screenshot was captured without the change. My question was in general if you can control the order since the order here does not match the one in the code.

systemcrash: Order of what? The order the elements appear in? Code order = appearance order

ngehrsitz: In the code it is:
1. Keytype
https://github.com/openwrt/luci/blob/6efab6839b9c1319d10444e8327a7d36eb17904a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js#L245-L275
2. URL
https://github.com/openwrt/luci/blob/6efab6839b9c1319d10444e8327a7d36eb17904a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js#L277-L284
3. Staging
https://github.com/openwrt/luci/blob/6efab6839b9c1319d10444e8327a7d36eb17904a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js#L286-L294
4. Days
https://github.com/openwrt/luci/blob/6efab6839b9c1319d10444e8327a7d36eb17904a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js#L296-L300

But in the Screenshot it´s 
1. Staging
2. Keytype
3. URL
4. Days

Thus not the order it is defined in...

systemcrash:
@stokito any thoughts?

stokito (approved):

stokito:
Thank you for the PR, it looks good. Maybe only it would be better to add a description with a known profile names and with a link to LE profiles page. Also we may call the `acme.sh --list-profiles` and show them in the drop down, but that would be over-complication.

I think we should merge it but...
For now, as far I understood the only reason to use them is to issue a cert for an IP address with the `shortlived` profile. And that's it: other profiles simply makes no sense at all. So they just created a new abstraction layer for nothing, if only they have a plan to use this latter to reduce compatibility, increase fragmentation, complicate understanding, make it prone for errors etc. I personally don't have any trust in the LE as an organization and I don't think that we, as users should follow them. Even in the docs they mentioned that the profiles was an idea of the "CA/Browser Forum" which is literally a syndicate of monopolists (DigiCert, Google) against whom initially the LE was formed. Instead we should move in a direction of decentralized web and stop to trust to the Certificate Authorities because they are not Authorities if they obey other Authorities IRL.

The IP address cert is always was a questionable thing. In fact, even if you don't have a domain... well okay, get one for free from DDNS. But this is another tool to use and if we can avoid it, then yes the IP cert would be better. And still, a user must see some domain for a safety reason. Otherwise it will be easier to make a fishing attack. Not sure if browsers do allow the IP certs and maybe they are only for tools like wget/curl but even in this case why would you use them with a router.

Ok, back to the PR. Can we get an IP cert without a profile (maybe just specify shorter days)? If yes, then lets just forget about profiles and make a single checkbox "IP addr cert". We also have a strict check for a domain name on the UI so a user anyway won't be able to issue such a cert.

ngehrsitz:
My use case for shortlived certs is that I want to notice any issues with my certificate automation as fast as possible. With 90 day certificates it can take a long time until I actually notice that it´s broken. With shortlived certs I can just monitor if the Certificate Transparency notifications come in every week.
Linking to LE does not seem appropriate to me since we shouldn´t assume that LE is the users CA.
We can´t simplify this to a checkbox or dropdown approach unless we implement a dynamic resolution with `acme.sh --list-profiles` since other CAs will have different profile names.

systemcrash:
What does `acme.sh --list-profiles` output?

stokito:
https://github.com/acmesh-official/acme.sh/wiki/Profile-selection

@ngehrsitz please turn the PR from Draft so it can be merged. From code perspective all looks good, the feature looks like needed. You may ignore my ranting

systemcrash:
Merged. Thanks @ngehrsitz

PR #8191, opened 2026-01-04
luci-base: Make some tables better searchable

Ramon00:
This changes the "name" in specific tables from being painted to being rendered. This allows the use of "ctrl-f" in browsers to e.g. search for firewall rule names.

Can anybody test this out? If no issues, I will then also change the other themes. 

P.S. Any reason why there is 2nd section in the css file:
```
.cbi-section-table-titles.named::before,
.cbi-section-table-descr.named::before,
.cbi-section-table-row[data-title]::before {

```
Maybe just remove this as well?

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -2576,6 +2576,9 @@ const CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection
 			if (this.extedit || this.rowcolors)
 				trEl.classList.add(!(tableEl.childNodes.length % 2)
 					? 'cbi-rowstyle-1' : 'cbi-rowstyle-2');
+			if  (sectionname && (!this.anonymous || this.sectiontitle)) {
+				trEl.appendChild(E('td', {'class': 'td cbi-value-field cbi-value-first-field'}, [ (sectionname && (!this.anonymous || this.sectiontitle)) ? sectionname : null ]));
+			}
 
 			for (let j = 0; j < max_cols && nodes[i].firstChild; j++)
 				trEl.appendChild(nodes[i].firstChild);
@@ -2616,10 +2619,17 @@ const CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection
 
 		if (has_titles) {
 			const trEl = E('tr', {
-				'class': `tr cbi-section-table-titles ${anon_class}`,
+				'class': `tr cbi-value-first-field cbi-section-table-titles ${anon_class}`,
 				'data-title': (!this.anonymous || this.sectiontitle) ? _('Name') : null,
 				'click': this.sortable ? ui.createHandlerFn(this, 'handleSort') : null
 			});
+			if (!this.anonymous || this.sectiontitle) {
+				trEl.appendChild(E('th', {
+						'class': 'th cbi-section-table-cell',
+						'data-sortable-row': this.sortable ? '' : null
+						},	(!this.anonymous || this.sectiontitle) ? _('Name') : null
+					));
+			}
 
 			for (let i = 0, opt; i < max_cols && (opt = this.children[i]) != null; i++) {
 				if (opt.modalonly)
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -2036,12 +2036,7 @@ form.inline { display: inline; margin-bottom: 0; }
 .cbi-section-table-titles.named::before,
 .cbi-section-table-descr.named::before,
 .cbi-section-table-row[data-title]::before {
-	content: attr(data-title) " ";
-	display: table-cell;
-	padding: 10px 10px 9px;
-	line-height: 18px;
-	font-weight: bold;
-	vertical-align: middle;
+	content: none;
 }
 
 .cbi-section-table-titles.named::before,
@@ -2321,6 +2316,10 @@ div.cbi-value var,
 .td.cbi-value-field var {
 	font-style: italic;
 	color: #0069d6;
+}
+
+.cbi-value-first-field {
+	font-weight: bold;	
 }
 
 div.cbi-value var[data-tooltip],
--- a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
+++ b/themes/luci-theme-material/htdocs/luci-static/material/cascade.css
@@ -1304,12 +1304,7 @@ td > table > tbody > tr > td,
 }
 
 .td[data-title]::before {
-	font-weight: bold;
-	display: none;
-	padding: .25rem 0;
-	content: attr(data-title) ":\20";
-	text-align: left;
-	white-space: nowrap;
+	content: none;
 }
 
 .tr.placeholder .td[data-title]::before {
@@ -1318,16 +1313,7 @@ td > table > tbody > tr > td,
 
 .tr[data-title]::before,
 .tr.cbi-section-table-titles.named::before {
-	font-weight: bold;
-	display: table-cell;
-	align-self: center;
-	flex: 1 1 5%;
-	padding: .25rem;
-	content: attr(data-title) "\20";
-	text-align: center;
-	vertical-align: middle;
-	white-space: normal;
-	word-wrap: break-word;
+	content: none;
 }
 
 .cbi-rowstyle-1 {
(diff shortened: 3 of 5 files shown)


systemcrash:
Is there a quick litmus test that I can use to verify the changes? I use FF and search works fine there.

vincejv:
@systemcrash the issue is reproducible on Chromium based browsers

### Steps to reproduce

1. Go to Luci > Network Firewall > Traffic Rules
2. Press `CTRL+F` to bring up browser search functionality
3. Enter search term that can be found in the page
4. Does not return anything as reported by browser as seen in the screenshot

image

systemcrash:
Thanks @vincejv - I'm aware. I don't run Chrome anything, which is why I ask for another way to verify...

jow-:
Please test on a mobile screen after this change, that content approach was used to be able to dynamically hide/re-position row titles to headlines when decomposing the table into cards.

Ramon00:
> Thanks @vincejv - I'm aware. I don't run Chrome anything, which is why I ask for another way to verify...

I think the progession part is sufficiently tested. I think testing regression impact is more important, i.e. is the visual layout affected? So Firefox test would be great to do as well.

> Please test on a mobile screen after this change, that content approach was used to be able to dynamically hide/re-position row titles to headlines when decomposing the table into cards.

If i view it portrait on mobile its pretty identical (other than a grey background bar missing in e.g. the rule name). 

Oh also the text is not in bold anymore (mobile or desktop), its same style as the other text, if needed I can make it bold again, if people prefer that. (The bold was not that functional I thought, but im fine either way). Just let me know what people prefer.

vincejv:
@systemcrash 
> I use FF and search works fine there.
> which is why I ask for another way to verify...

it doesn't work on Safari either, tested on iPhone with iOS 26.2, now if someone can do the same test on Mac.

Ramon00:
> @systemcrash
> 
> > I use FF and search works fine there.
> > which is why I ask for another way to verify...
> 
> it doesn't work on Safari either, tested on iPhone with iOS 26.2, now if someone can do the same test on Mac.

Chromium based browsers suffer from this "feature" (=not searching painted items). This is a design choice as those ::before elements should only contain visual things, not relevant information. We can question if this is the right choice in the browser or not, but regardless,  chromium based browsers suffer from this. I think the market share is enough of chromium based browsers to warrant fixing this. 

What I do think needs maybe more testing is if on all browsers the patch does not mess up the layout anywhere (we dont want to substitute one issue for another).

vincejv:
safari is not chromium based though. However chromium based browsers uses blink engine which was forked from WebKit.

this design decision could stem back from WebKit. So if we fix this upstream, we have to file 2 separate issues for both WebKit/Safari and Chromium. However, since search only “works” correctly in Firefox, I agree with you that it makes sense to follow the behavior of the two engines with much larger user bases.

Ramon00:
I agree 
(the point i was trying to make, that i dont think it makes sense to do a full inventory of which browser has the issue and which browser does not, before fixing it)

Ramon00:
Updated all themes now.

Ramon00:
This is what it will look now:

Bootstrap
image


Material
image


Openwrt
image


Openwrt-2020
image

systemcrash:
What happens on mobile? :)

Ramon00:
Looks fine to me:

image

jow-:
The label text ("printer port") should be slightly bigger and bold afair

Ramon00:
> The label text ("printer port") should be slightly bigger and bold afair

Currently the font is not bigger, the only delta is it being bold.
But i can make it bigger and bold by reusing a css item or making a new one. To be honest I do not think the bold or bigger is very functional, it does not make the pages read easier or be better to understand, but im fine with adding that. 

How big do you want it to be?

systemcrash:
`font-weight: bold;` should probably be retained, then.

Ramon00:
image

How about that?

Ramon00:
I can add something like `font-size: 110%;` if that makes it look better

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -2641,10 +2619,17 @@ const CBITableSection = CBITypedSection.extend(/** @lends LuCI.form.TableSection
...
-				'class': `tr cbi-section-table-titles ${anon_class}`,
+				'class': `tr cbi-value-first-field cbi-section-table-titles ${anon_class}`,
 				'data-title': (!this.anonymous || this.sectiontitle) ? _('Name') : null,
 				'click': this.sortable ? ui.createHandlerFn(this, 'handleSort') : null
 			});
+			if (!this.anonymous || this.sectiontitle) {
+				trEl.appendChild(E('th', {
+						'class': 'th cbi-section-table-cell',
+						'data-sortable-row': this.sortable ? '' : null
+						},	(!this.anonymous || this.sectiontitle) ? _('Name') : null
```
systemcrash, line 2630:
```js
						},	[ (!this.anonymous || this.sectiontitle) ? _('Name') : null ]
```

Ramon00: To be honest, i dont think  the square brackets are needed here, we only add a single element not an array. Looks like appendChild also only supports a single element. Maybe it is actually better to remove them in line 2580 as well? 

I did do a quick search in the luci code base and both are used. So both seem to work. Let me know if you want me to add it here or remove on 2580

systemcrash:
Merged via 720c96ce5b85583c8c6db6ef31857dc8a3eeb600. Thanks @Ramon00 !

eamonxg:
Hi there ,

First of all, thank you for your work on updating LuCI!  
I really appreciate the improvements in the new version. I wanted to bring up a small backward compatibility issue I noticed regarding the table first column, which might help improve theme compatibility.

In previous versions of LuCI, the first column of tables was displayed using  
`.tr[data-title]::before` and `.td[data-title]::before` (with `data-title`).  
In a recent update, this implementation was changed to use actual DOM elements instead.

Currently, themes handle this change by simply disabling the `::before` pseudo-element in CSS (for example, using `content: none`).  
While this works fine with the new version, there is a potential compatibility issue:

If a user is still on an **older version of LuCI** that relies on `::before` for the first column, but installs the **latest theme**, the first column disappears because `::before` is directly disabled.  
Simply overriding the pseudo-element in CSS can unintentionally break the display for older LuCI versions.

## Illustration of the Issue

| Scenario | Outcome | Screenshot / Illustration |
|----------|---------|---------------------------|
| Old LuCI (uses `::before`) + **old theme** | ✅ First column displays correctly |has-before |
| Old LuCI (uses `::before`) + **new theme** | ❌ First column disappears | no-before

eamonxg:
Hi @Ramon00 @systemcrash ,

I have a suggestion that might help maintain backward compatibility:

One idea is to add a simple flag to indicate that `::before` is no longer used in the new implementation, for example:

```html
<td data-title="xxx" data-no-before></td>
<tr data-title="xxx" data-no-before></tr>
```


With this flag, themes could use a selector to conditionally apply the old ::before styles, enabling backward compatibility, for example:
```css
.td[data-title]:not([data-no-before])::before {
  /* original styles */
}

.tr[data-title]:not([data-no-before])::before {
  /* original styles */
}
```
This would allow older versions of LuCI that still rely on ::before to continue displaying the first column correctly, while the new implementation remains unaffected.

systemcrash:
Hmm - thanks for the information @eamonxg. I won't pick this to previous versions - only to 25.12. Those are some good suggestions.

eamonxg:
> Hmm - thanks for the information @eamonxg. I won't pick this to previous versions - only to 25.12. Those are some good suggestions.

Thanks for the reply, I hope this could be considered, as third-party themes need to support both versions before 25.12 and the latest version. A simple `data-no-before` flag would make handling this much easier without affecting the new implementation.

Ramon00:
Yeah indeed you can't just update or install  one of the two halfs. Adding style data into the js code makes updating in the future more complicated. I guess just using this on 25.12 is probably a good compromise, as it only affects the test releases where you can expect some "bugs" and only have limited audience, while still keeping the code not too complicated.

eamonxg:
> Adding style data into the js code makes updating in the future more complicated

Since the goal is to avoid making the code more complicated, what about completely removing the `named `and` data-title` logic that was used for displaying the ::before pseudo-element?
image

Ramon00:
> what about completely removing the `named `and` data-title` logic that was used

I am all for removing obsolete code. The thing is, you have to be careful not to break something if that data-title tag was still used.  Leaving it in may also make debugging easier.  Removing those tags  will of course not fix the issue of installing a later version of a theme with an old version of the luci code. Regardless the issue will be gone once rc4 gets build (i guess rc3 got tagged just two days too early, maybe for some of the builds its already ok...)

systemcrash:
You can still image build in a few days a get the new theme.

Issue #8195, opened 2026-01-04
luci-base: build products created in source tree, never cleaned


9ary:
The luci-base host-compile step runs make directly in the source tree and leaves build products there.
https://github.com/openwrt/luci/blob/0d55a89cb2c7b781553d41a519d4875d9d15c1f3/modules/luci-base/Makefile#L67-L69

To reproduce, build OpenWrt normally, then inspect feeds/luci:
```console
$ git clean -fdxn
Would remove modules/luci-base/src/contrib/lemon
Would remove modules/luci-base/src/jsmin
Would remove modules/luci-base/src/jsmin.o
Would remove modules/luci-base/src/lib/lmo.o
Would remove modules/luci-base/src/lib/plural_formula.c
Would remove modules/luci-base/src/lib/plural_formula.h
Would remove modules/luci-base/src/lib/plural_formula.o
Would remove modules/luci-base/src/po2lmo
Would remove modules/luci-base/src/po2lmo.o
```

Additionally, some of these files aren't picked up by the clean step:
https://github.com/openwrt/luci/blob/0d55a89cb2c7b781553d41a519d4875d9d15c1f3/modules/luci-base/src/Makefile#L18-L19
The above doesn't cover `jsmin{,.o}` and `po2lmo{,.o}`.

I noticed because I ran into a build failure due to changes in how my distro builds GCC, which made old objects (that had lingered since April!) no longer linkable.

<details><summary>Config seed</summary>

```
# Target
CONFIG_TARGET_rockchip=y
CONFIG_TARGET_rockchip_armv8=y
CONFIG_TARGET_rockchip_armv8_DEVICE_friendlyarm_nanopi-r5c=y
# CONFIG_TARGET_ROOTFS_EXT4FS is not set
CONFIG_TARGET_ROOTFS_PARTSIZE=512

# Packages
CONFIG_PACKAGE_luci-ssl=y
CONFIG_PACKAGE_htop=y
CONFIG_PACKAGE_iperf3=y
CONFIG_PACKAGE_kmod-usb-net-rndis=y
CONFIG_PACKAGE_openssh-sftp-server=y
CONFIG_PACKAGE_tailscale=y
CONFIG_PACKAGE_vim-fuller=y
```
</details>

PR #8196, opened 2026-01-05
luci-base: clean overlooked files

9ary:
Closes: https://github.com/openwrt/luci/issues/8195

This fixes the primary symptom (subsequent builds reusing stale products), but not the root cause (the build running in-tree).


systemcrash:
please change updates to closes in your commit message

9ary:
The commit message is deliberate: this patch does not fix the problem as I reported it which is that the build litters the source tree, it only deals with the consequences. If the original issue is a wontfix then I'll change it.

systemcrash:
Merged. Thanks @9ary

PR #8210, opened 2026-01-09
luci-app-radicale2, rpcd-mod-rad2-enc: drop packages

BKPepe:
Radicale2 has been dropped from the packages feed as it fails to compile due to PEP 625 compliance issues.

While the package compiles locally with lowercase filenames, the build system expects uppercase names. However, PEP 625 enforces lowercase for distribution filenames (sdists/wheels).

Since radicale3 is available as a replacement (https://github.com/openwrt/packages/pull/28211), radicale2 is being removed. LuCI support for radicale3 will be added shortly.

**This needs to be backported to the stable releases e.g. OpenWrt 25.12, OpenWrt 24.10.**
Radicale3 will be added there within 48 hours once radicale3 is successfully compiled on buildbots.


danielfdickinson:
See https://github.com/openwrt/luci/pull/8216 which adds an app for radicale3

systemcrash:
picked to 25 and 24

muink:
Removing `luci-app-radicale2` caused [weblate](https://hosted.weblate.org/projects/openwrt/luciapplicationsattendedsysupgrade/#alerts) to fail to synchronize the repository.

systemcrash:
Thanks for the heads up @muink. It's fixed now.

PR #8204, opened 2026-01-08
luci-base: rpc: fix luci/getMountPoints crash on path containing '\xxx'

jjm2473:
`char` should be `chr`

Fixes https://github.com/openwrt/openwrt/issues/21459

- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [x] Incremented :up: any `PKG_VERSION` in the Makefile
- [x] Tested on: (architecture, openwrt version, browser) :white_check_mark: armsr/arm64, openwrt 25.12.0-rc2, Chrome
- [x] \( Preferred ) Mention: @jow- 
- [ ] \( Preferred ) Screenshot or mp4 of changes:
- [x] \( Optional ) Closes: e.g. openwrt/openwrt#21459
- [x] Description: (describe the changes proposed in this PR)

Fix `char` function undefined.

PR #8214, opened 2026-01-10
[openwrt-24.10] luci-base: rpc: fix luci/getMountPoints crash on path containing '\xxx'

jjm2473:
getMountPoints crash on path containing '\xxx'.

`char` should be `chr`

Fixes https://github.com/openwrt/openwrt/issues/21459

(cherry picked from commit 5ad9c3609e1616836d6bc6125a6c983231a2de9c)

This is a backport of https://github.com/openwrt/luci/pull/8204


jjm2473:
ping @jow- @systemcrash

PR #8216, opened 2026-01-11
luci-app-radicale3: add app for radicale3

danielfdickinson:
      bcm72xx/bcm2712 (rpi-5), SNAPSHOT (r32598-93e98b4082), Firefox ESR
- N/A \( Preferred ) Mention: @ the original code author for feedback

image

- N/A \( Optional ) Closes: e.g. openwrt/luci#issue-number
  https://github.com/openwrt/packages/pull/28322

radicale2 has been removed from packages as it can no longer be
built due to PyPi service changes. It has been replaced with
radicale3: https://github.com/openwrt/packages/pull/28211

Here we add a JavaScript-based LuCI app for radicale3, in place of
the old Lua-based LuCI app for radicale2.

In order to support the user management (encrypting passwords) we add a helper rpcd
program. It is a Python program because radicale3 is a Python program, so we know Python is
already on the system, and it allows using radicale itself for part of the helper.

rpcd-mod-rad3-enc: add radicale3 password hasher

This replaces the rpcd-mod-rad2-enc helper for radicale2.

@jow- @aparcar @hnyman @BKPepe @stangri I would welcome some review of this app before finalizing it.

### DIFF
--- a/applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
+++ b/applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
@@ -1,0 +1,478 @@
+'use strict';
+'require dom';
+'require form';
+'require view';
+'require rpc';
+'require uci';
+'require ui';
+
+// cspell:words Radicale addremove authtype cfgsections cfgvalue formvalue
+// cspell:words hostport htpasswd ipaddrport noopener noreferrer nsections
+// cspell:words origcfgsections plainpass ppconfirm rmempty sslon taboption
+// cspell:words uinteger multifilesystem hidetitle packagelist libpass
+// cspell:words passlib cffi
+
+const callServiceList = rpc.declare({
+	object: 'service',
+	method: 'list',
+	params: ['name'],
+	expect: {
+		radicale3: {},
+	},
+});
+
+const callEncrypt = rpc.declare({
+	object: 'rad3-enc',
+	method: 'encrypt',
+	params: ['type', 'plainpass'],
+	expect: {},
+});
+
+const callPackageList = rpc.declare({
+	object: 'rpc-sys',
+	method: 'packagelist',
+	params: [],
+	expect: {},
+});
+
+return view.extend({
+	load() {
+		return Promise.all([
+			L.resolveDefault(callServiceList('radicale3')),
+			L.resolveDefault(callPackageList()),
+			uci.load('radicale3'),
+		]);
+	},
+	render([service_list_r3, package_list_r3]) {
+		let radicale_address, radicale_host, radicale_port, running, button;
+		let has_ssl, package_list, python_ssl_version;
+
+		radicale_host = uci.get('radicale3', 'server', 'host') || ['127.0.0.1:5232'];
+		radicale_host = String(radicale_host?.[0]);
+		radicale_address = radicale_host.substring(0, radicale_host.lastIndexOf(':')) || '127.0.0.1';
+
+		button = '';
+
+		if ((radicale_address == '127.0.0.1') || (radicale_address == '::1')) {
+			ui.addNotification(_('Need a listen address'),
+				_('Radicale needs a non-loopback IP address for your browser to access the web interface'));
+		} else {
+			if (radicale_address == '0.0.0.0' || radicale_address == '::') {
+				radicale_address = window.location.hostname;
+			}
+			radicale_port = radicale_host.substring(radicale_host.lastIndexOf(':') + 1) || '5232';
+
+			running = Object.keys(service_list_r3.instances || {}).length > 0;
+
+			if (running)
+				button = '&#160;&#160;<a class="btn" href="' + window.location.protocol + '//' + radicale_address + ':' + radicale_port + '" target="_blank" rel="noreferrer noopener">' + _('Open Web Interface') + '</a>';
+		}
+
+		package_list = package_list_r3.packages;
+		python_ssl_version = package_list['python3-openssl'];
+		has_ssl = (python_ssl_version != null) && (python_ssl_version != '');
+
+		let m, s, o, at, hte, plainpass, ppconfirm, pass;
+
+		m = new form.Map('radicale3', 'Radicale3', _('Radicale v3 CalDav/CardDAV Server Configuration') + button);
+
+		s = m.section(form.NamedSection, 'server', 'section', _('Server Settings'));
+		s.addremove = true;
+		s.anonymous = false;
+
+		s.tab('main', _('Main'));
+
+		o = s.taboption('main', form.DynamicList, 'host', _('Host:port'));
+		o.optional = true;
+		o.datatype = 'or(hostport(0),ipaddrport(0))';
+		o.default = ['127.0.0.1:5232', '::1:5232'];
+
+		s.tab('advanced', _('Advanced'));
+
+		if (has_ssl) {
+			let sslon, cert_file, key_file, ca_file;
+
+			sslon = s.taboption('main', form.Flag, 'ssl', _('SSL'), _('Enable SSL connections'));
+			sslon.rmempty = true;
+			sslon.default = sslon.disabled;
+
+			cert_file = s.taboption('main', form.FileUpload, 'certificate', _('Certificate'));
+			cert_file.rmempty = true;
+			cert_file.optional = false;
+			cert_file.depends('ssl', sslon.enabled);
+
+			key_file = s.taboption('main', form.FileUpload, 'key', _('Private Key'));
+			key_file.rmempty = true;
+			key_file.optional = false;
+			key_file.depends('ssl', sslon.enabled);
+
+			sslon.cfgvalue = function (section_id) {
+				return ((cert_file.cfgvalue(section_id) != null) && (key_file.cfgvalue(section_id) != null));
+			};
+
+			ca_file = s.taboption('advanced', form.FileUpload, 'certificate_authority', _('Client Certificate Authority'), _('For verifying client certificates'));
+			ca_file.rmempty = true;
+			ca_file.depends('ssl', sslon.enabled);
+
+			o = s.taboption('advanced', form.Value, 'ciphers', _('Allowed Ciphers'), _('See python3-openssl documentation for available ciphers'));
+			o.rmempty = true;
+			o.depends('ssl', sslon.enabled);
+
+			o = s.taboption('advanced', form.Value, 'protocol', _('Use Protocol'), _('See python3-openssl documentation for available protocols'));
+			o.rmempty = true;
+			o.depends('ssl', sslon.enabled);
+			o.placeholder = 'PROTOCOL_TLSv1_2';
+		} else {
+			o = s.taboption('main', form.DummyValue, '_no_ssl', _('No SSL'), _('No SSL support available. Please install python3-openssl.'));
+			o.write = function () {};
+			o.cfgvalue = function () {
+				return '';
+			};
+		}
+
+		o = s.taboption('advanced', form.Value, 'max_connection', _('Max Connections'), _('Maximum number of simultaneous connections'));
+		o.rmempty = true;
+		o.placeholder = 20;
+		o.datatype = 'uinteger';
+
+		o = s.taboption('advanced', form.Value, 'max_content_length', _('Max Content Length'), _('Maximum size of request body (bytes)'));
+		o.rmempty = true;
+		o.datatype = 'uinteger';
+		o.placeholder = 100000000;
+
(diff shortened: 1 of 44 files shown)


applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,405 @@
...
+
+const callEncrypt = rpc.declare({
+	object: 'rad3-enc',
+	method: 'encrypt',
+	params: ['type', 'plainpass'],
+	expect: {}
+});
+
+return view.extend({
+	load: function () {
```
systemcrash: `load() {` (do same with all other functions)

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,405 @@
...
+		return Promise.all([
+			L.resolveDefault(callServiceList('radicale3')),
+			uci.load('radicale3')
+		]);
+	},
+	render: function (res) {
```
systemcrash: `render([service_list_r3]) {` and use that instead of generic res.

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,405 @@
...
+		var radicale_host = uci.get('radicale3', 'server', 'host') || ['127.0.0.1:5232'];
```
systemcrash: Use ES6 const or let everywhere instead of global var.

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,405 @@
...
+		if (radicale_address == '0.0.0.0' || radicale_address == '::') {
+			radicale_address = window.location.hostname;
+		}
+		var radicale_port = radicale_host.substring(radicale_host.lastIndexOf(':') + 1) || '5232';
+
+		var running = Object.keys(res[0].instances || {}).length > 0;
+
+		var button = '';
+		if (running)
+			button = '&#160;&#160;<a class="btn" href="' + window.location.protocol + '//' + radicale_address + ':' + radicale_port + '" target="_blank" rel="noreferrer noopener">' + _('Open Web Interface') + '</a>';
```
systemcrash: If you form a link with the loopback address, what is the actual target on the client?

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,405 @@
...
+		pass.password = true;
+		pass.rmempty = false;
+		pass.depends({ 'radicale3.auth.type': 'htpasswd' });
+		pass.depends({ 'radicale3.auth.type': '' });
+
+		pass.parse = async function (section_id) {
+			var active = this.isActive(section_id);
+
+			var plainValue = '';
+			var encValue = '';
```
systemcrash: ES6...

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,405 @@
...
+					const authtype = hte.formvalue('auth') || 'plain';
+					var variant = at.formvalue(section_id) || 'plain';
+
+					if (authtype != 'autodetect') {
+						variant = authtype;
+					}
+
+					const encryptPromise = callEncrypt(variant, plainValue).then(function (res) {
+						if (!res.encrypted_password) {
+							throw new Error(_('Unable to encrypt plaintext password: %s').format(res.error));
```
systemcrash: ui.timelimitednotification

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: The default for the package is to only bind to use loopback address. The user must set a listen address for it to listen other than on loopback. I think I'll change this to an 'if loopback inform user to add a listen address'.

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: Done

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: Done

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: Implemented

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,422 @@
...
+
+					if (authtype != 'autodetect') {
+						variant = authtype;
+					}
+
+					const encryptPromise = callEncrypt(variant, plainValue).then(function (res) {
+						if (!res.encrypted_password) {
+							ui.addTimeLimitedNotification(
+								_('Unable to encrypt plaintext password'),
+								dom.create('<p>' + res.error + '</p>'),
```
systemcrash: You can pass a string directly. No need for tags, unless it looks better that way.

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,422 @@
+'use strict';
+'require dom';
+'require form';
+'require view';
+'require fs';
```
systemcrash: unused...?

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: Thanks. Will update.

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: Of those, I can drop 'require fs'. The others are used. That was inherited from the Lua for checking if python3-openssl is present - I haven't added that back yet as I haven't had a chance to figure how to achieve that in the JavaScript LuCI.

systemcrash: Either a Makefile depends or fs.exec_direct (or rpc call to a ucode script?) that runs python that tries to load said module: 
```sh
python3 -c 'import openssl' # whatever the module is called in python
```
and you check the Python return code.

systemcrash:
Perhaps the python script can be re-done in ucode for speed if everything is available. 

https://ucode.mein.io/module-digest.html

Not sure about md5 crypt or the 5000 rounds figure.

https://github.com/jow-/ucode/pull/347

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: I realized I could use the ubus call rpc-sys packageslist and check that python3-ssl is installed and am implementing that way.

danielfdickinson:
Currently it doesn't look like everything is available (no rounds configuration that I see, and definitely not bcrypt, or -- the to be added -- argon2). I would like to be able to switch it to ucode at some point, however I don't want to block on that.

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
systemcrash: That works!

danielfdickinson:
One question before I take this off draft: do you want the eslint configuration and related (needed for eslint) package.json removed from the luci-app-radicale3 directory, or is it okay to leave it?

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,479 @@
...
+			L.resolveDefault(callPackageList()),
+			uci.load('radicale3'),
+		]);
+	},
+	render([service_list_r3, package_list_r3]) {
+		let radicale_address, radicale_host, radicale_port, running, button;
+		let has_ssl, package_list, python_ssl_version;
+
+		radicale_host = uci.get('radicale3', 'server', 'host') || ['127.0.0.1:5232'];
+		radicale_host = String(radicale_host[0]);
```
systemcrash: `		radicale_host = String(radicale_host?.[0]);` in case radicale_host is empty for some reason.

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
```diff
@@ -0,0 +1,479 @@
...
+
+			if (!/^[a-zA-Z0-9._-]+$/.test(value)) {
+				return _('Username can only contain letters, numbers, dots, underscores and hyphens');
+			}
+
+			// Check for duplicate usernames
+			const sections = this.section.cfgsections();
+			for (let i = 0; i < sections.length; i++) {
+				if (sections[i] !== section_id) {
+					const existingName = this.cfgvalue(sections[i]);
```
systemcrash: I find `for (const section of sections) { ...` more readable.

danielfdickinson:
@systemcrash Thank you for all your comments and reviewing!

applications/luci-app-radicale3/htdocs/luci-static/resources/view/radicale3.js
danielfdickinson: I knew I should have looked closer at the that bit that Duo inserted - I hadn't thought of checking duplicate usernames, so it's a nice addition, but doing it well is important too.

danielfdickinson:
@systemcrash Anything else I have missed/that could be improved?

Also, I have taken off draft with the eslint configuration still present; do you want me to remove it before merging, or is it good as is?

applications/luci-app-radicale3/root/usr/share/luci/menu.d/luci-app-radicale3.json
```diff
@@ -0,0 +1,14 @@
...
+		"title": "Radicale3",
+		"order": 120,
+		"action": {
+			"type": "view",
+			"path": "radicale3"
+		},
+		"depends": {
+			"acl": [ "luci-app-radicale3" ],
+			"uci": { "radicale3": true }
+		}
```
systemcrash, line 15: This is an optional, but you can require presence of binaries to show the menu items here.

"Either/Or" e.g. from mod-network:
```json
			"fs": [
				{ "/usr/sbin/odhcpd": "executable" },
				{ "/usr/sbin/dnsmasq": "executable" }
			],
```

"And" e.g. 
```json
			"fs": {
				"/etc/init.d/foo": "executable",
				"/usr/bin/foo": "executable"
			},
```

systemcrash:
> One question before I take this off draft: do you want the eslint configuration and related (needed for eslint) package.json removed from the luci-app-radicale3 directory, or is it okay to leave it?

Linting should be separate, since that will apply to the whole repo. But it's interesting to try, nonetheless. (At best it should be a recommendation, and not a hard failure. )

danielfdickinson:
@systemcrash As currently included in the PR it is only for the luci-app-radicale3. Do you mean that if added, it should be added for the whole repo (as a recommendation, not hard failure) and not only for a particular app?

Also, thank you for the tip on the "fs" bit - that will change a PR I have on the packages repo for the `radicale3` package. It means instead depending on the uci config being non-empty that I can depend on the existence of the radicale3 initscript, for the LuCI app.

systemcrash:
> @systemcrash As currently included in the PR it is only for the luci-app-radicale3. Do you mean that if added, it should be added for the whole repo (as a recommendation, not hard failure) and not only for a particular app?

Let's exclude it. It requires some different magic to run in this repo.

danielfdickinson:
Once I'm done testing this (to be pushed once tested) last set of changes (using the "fs") this will be good to go, although working SSL with this app will be pending https://github.com/openwrt/packages/pull/28325 (~~to be updated, which is why I have temporarily made it draft~~).

danielfdickinson:
Should be GTG.

systemcrash:
pick to ... 25?

danielfdickinson:
I'd say yes. Some people will complain because 24 has lost luci-app-radicale2 and will not have a replacement, but I think this is a lot of new work, so doesn't really belong in the existing stable series (25 which is still rc is another matter).

PR #8236, opened 2026-01-20
luci-mod-network & luci-mod-status: Add PSE/PoE configure functionality for modern linux PSE-PD support.

carlosz1986:
We're working currently on backporting the modern kernel PSE driver to 6.12. This will allow us to have a standardized interface to control POE devices. By that we can also port realtek-poe and other drivers to this unified approach.
There is a still a lot to do, but we have a working prototype. More you can read [here](https://forum.openwrt.org/t/support-poe-on-hasivo-devices-s1100wp-8gt-se/244817/44?u=openwrtawesome): 

I made this PR to get first feedback and it includes: 
- add hasPSE() [all status information] and getPSE() [has the device PSE hardware?]
- changes ACL permissions to query PSE data from netifd
- add new POE icon for status page
- changes port status to show poe information, next to link information and data transfer.
- add new tab to device config for POE.

Some pictures & questions:
1) The device config page:
image

- Is this in line what you would expect?
- Would you suggest to only load the tab, if the device supports POE or does that not matter much?

2) Changes of Port status page:
image

- I changed the icon, if POE connection is detected. Is this a solid approach?
- Does it make sense to show the status for POE here?
- Do you think it's okay to show here power values? I wonder, because all other stats are sums. But showing energy (instead of power) is mostly not supported by the driver and requires bigger changes. Also in general I expect that power is more meaningful here.

3) Hover to see more details:
image
- The same questions here: Are the values helpful, what is about power vs energy....

I would appreciate general feedback on code, acls etc. as well, so I can adjust the direction...
Would you merge such a change, or is this rather a plugin module...


modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -358,12 +451,34 @@ return baseclass.extend({
...
-			    pzones = (pmap && pmap.zones.length) ? pmap.zones.sort(function(a, b) { return L.naturalCompare(a.getName(), b.getName()) }) : [ null ];
+			    pzones = (pmap && pmap.zones.length) ? pmap.zones.sort(function(a, b) { return L.naturalCompare(a.getName(), b.getName()) }) : [ null ],
+			    pse = pseMap[port.device],
+			    pseInfo = getPSEStatus(pse),
+			    psePower = formatPSEPower(pse);
+
+			/* Select port icon: poe (delivering), up (carrier), down (no link) */
+			var portIcon = 'down';
+			if (carrier) {
+				portIcon = (pseInfo && pseInfo.isDelivering) ? 'poe' : 'up';
```
bevanweiss: I don't think `up`/`down` and `pse delivering` are necessarily in the same data set.
I'm pretty sure that a port can be delivering power, but have no active link on it (i.e. perhaps auto negotiation failed, or MDIX settings were wrong).
So I think all existing up/down statuses need to remain visible, whilst also adding the pse delivering / enabled status indications.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -377,12 +492,7 @@ return baseclass.extend({
 					E('span', { 'class': 'cbi-tooltip left' }, [ renderNetworksTooltip(pmap) ])
 				]),
 				E('div', { 'class': 'ifacebox-body' }, [
-					E('div', { 'class': 'cbi-tooltip-container', 'style': 'text-align:left;font-size:80%' }, [
```
bevanweiss, line 380: Is this necessary for the pse introduction?  If not, it's likely nicer to have it as a separate commit + PR.

It's ideal to keep commits (and PRs) as singular and specific as possible, in particular around bisecting possible causes of bugs later on.  So we wouldn't want this commit or PR to do multiple things at once.

andrewjlamarche:
@carlosz1986 my opinion is yes/looks good to all all your questions, except that I would only expect the PoE tab to load if PoE equipment is available, so new users are not confused.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -13,6 +13,13 @@ var callGetBuiltinEthernetPorts = rpc.declare({
 	expect: { result: [] }
 });
 
+var callNetworkDeviceStatus = rpc.declare({
```
systemcrash: const

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -236,10 +243,51 @@ function formatSpeed(carrier, speed, duplex) {
...
+		return null;
+
+	var status = pse['c33-power-status'] || pse['podl-power-status'],
+	    power = pse['c33-actual-power'];
+
+	return {
+		status: status,
+		power: power,
+		isDelivering: status === 'delivering' && power > 0
+	};
```
systemcrash: `return { status: pse['...'] || pse['...'], power: pse['..'], ... }`

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -236,10 +243,51 @@ function formatSpeed(carrier, speed, duplex) {
...
+}
+
+function formatPSEPower(pse) {
+	if (!pse)
+		return null;
+
+	var status = pse['c33-power-status'] || pse['podl-power-status'],
+	    power = pse['c33-actual-power'];
```
systemcrash: const

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -236,10 +243,51 @@ function formatSpeed(carrier, speed, duplex) {
...
+
+	if (status === 'delivering' && power) {
+		var watts = (power / 1000).toFixed(1);
+		return E('span', { 'style': 'color:#000' },
+			[ '\u26a1\ufe0e\u202f%s\u202fW'.format(watts) ]);
```
systemcrash, line 271: add some code comments for how these should look when formatted

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -236,10 +243,51 @@ function formatSpeed(carrier, speed, duplex) {
...
+
+function formatPSEPower(pse) {
+	if (!pse)
+		return null;
+
+	var status = pse['c33-power-status'] || pse['podl-power-status'],
+	    power = pse['c33-actual-power'];
+
+	if (status === 'delivering' && power) {
+		var watts = (power / 1000).toFixed(1);
```
systemcrash: const everywhere applicable

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -252,7 +300,27 @@ function formatStats(portdev) {
...
+	if (pse) {
+		var status = pse['c33-power-status'] || pse['podl-power-status'],
+		    power = pse['c33-actual-power'],
+		    powerClass = pse['c33-power-class'],
+		    powerLimit = pse['c33-available-power-limit'];
+
+		items.push(_('PoE status'), status || _('unknown'));
+
+		if (power)
+			items.push(_('PoE power'), '%.1f W'.format(power / 1000));
```
systemcrash, line 315: the '%1000m mW'.format() would produce SI units of mW (e.g. 1Ki mW), but that would probably be weird and less readable. Are you happy with 1dp of precision?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -309,7 +377,31 @@ return baseclass.extend({
 			firewall.getZones(),
 			network.getNetworks(),
 			uci.load('network')
-		]);
+		]).then(function(data) {
+			/* Find devices with PSE configured */
+			var pseDevices = [];
```
systemcrash: const

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -309,7 +377,31 @@ return baseclass.extend({
 			firewall.getZones(),
 			network.getNetworks(),
 			uci.load('network')
-		]);
+		]).then(function(data) {
```
systemcrash: (data) => {

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -358,12 +451,34 @@ return baseclass.extend({
...
+			var portIcon = 'down';
+			if (carrier) {
+				portIcon = (pseInfo && pseInfo.isDelivering) ? 'poe' : 'up';
+			}
+
+			var statsContent = [
+				'\u25b2\u202f%1024.1mB'.format(port.netdev.getTXBytes()),
+				E('br'),
+				'\u25bc\u202f%1024.1mB'.format(port.netdev.getRXBytes())
+			];
```
systemcrash: is this to prime the content with something?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: changed everywhere else ...as well

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: makes sense!

carlosz1986: After thinking about it: Best would be 2 new icons (one for poe active with link, one for poe active without link)... Then it's clear. I will do that...

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: If it's okay for you, I think that is fine ... :-)

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: I guess, it exists to format the transferred traffic amount (B, KB, GB) and a icon up ... That existed already before.
What are your thoughts?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: done:)

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: the "status" & "power" variable is used twice, so I thought it's nicer like this?

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: Could you explain what your concern is? I mean it exists to show the power "Watt" value next to the traffic sums (RX,TX), but otherwise it does not change much... :-)

bevanweiss: If it adds the Power value into the traffic values, then it (or something similar) is required.

Typically when adding a new feature, it should be done with minimal changes. With other 'tidy up' changes done separately, just so potential breakages can be tracked atomically to either the introduction of the feature, or to the tidy up.

So could the Power be added with 'less' changes to how the existing stuff works?
(and have those other changes in a subsequent 'tidyup' PR).
For such a small change as this, it likely won't matter, just something to keep in mind for later.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
bevanweiss: I would think Watts (W) is likely better...
46.0Ki mW definitely won't read as well as 46.0 W
Unsure if two dp adds valuable info though, 46.0 W vs 46.02 W I don't think adds actionable info (although maybe someone with a 48 port PoE switch does care about the 0.01W)

carlosz1986:
> @carlosz1986 my opinion is yes/looks good to all all your questions, except that I would only expect the PoE tab to load if PoE equipment is available, so new users are not confused.

This is now changed... As this requires a async call (similar to what happens in wifi configs), there were more changes required. I hope you can give feedback...

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: I changed it, so we have the typical link Up&Down icons from before and now two others for PSE with and without link. On top it shows PSE status... 


Image

Image

Image

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1571,9 +1579,19 @@ return view.extend({
...
+			/* Query PSE status from netifd to determine if device has PSE capability */
+			if (devName) {
+				return L.resolveDefault(callNetworkDeviceStatus(devName), {}).then((status) => {
+					var hasPSE = (status.pse != null);
+					nettools.addDeviceOptions(s, dev, isNew, rtTables, hasPSE);
+				});
+			} else {
+				nettools.addDeviceOptions(s, dev, isNew, rtTables, false);
+				return Promise.resolve();
+			}
```
systemcrash: If we're adding new variables, let's go with ES6 let and const. Here I think const is fine.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -358,12 +468,36 @@ return baseclass.extend({
 			    duplex = port.netdev.getDuplex(),
 			    carrier = port.netdev.getCarrier(),
 			    pmap = port_map[port.netdev.getName()],
-			    pzones = (pmap && pmap.zones.length) ? pmap.zones.sort(function(a, b) { return L.naturalCompare(a.getName(), b.getName()) }) : [ null ];
+			    pzones = (pmap && pmap.zones.length) ? pmap.zones.sort(function(a, b) { return L.naturalCompare(a.getName(), b.getName()) }) : [ null ],
+			    pse = pseMap[port.device],
+			    pseInfo = getPSEStatus(pse),
+			    psePower = formatPSEPower(pse);
```
systemcrash: Just make these let/const x; one per line.

systemcrash:
When you're done, squash your commits, leave the descriptions intact, 
title e.g.: `luci-base: PoE and PSE details` and add SoB.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
carlosz1986: ok, done

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
carlosz1986: sry, for causing you extra work ;-) that could have been avoided! changed

carlosz1986:
> When you're done, squash your commits, leave the descriptions intact, title e.g.: `luci-base: PoE and PSE details` and add SoB.

I tested again all scenarios (with poe, without ... & with patched openwrt netifd,ubus, without ...) So it does not cause any conflicts.
I added the commit signoff message, squashed and cleaned up.

I assume we are ready to go :-)

systemcrash:
Merged. Thanks @carlosz1986 !

carlosz1986:
> Merged. Thanks @carlosz1986 !

thx to you for helping out :-)

PR #8239, opened 2026-01-21
luci-app-advanced-reboot: bugfix: actual partition switch

stangri:
Maintainer: me
Compile tested: x86_64, Dell EMC Edge 620, OpenWrt 24.10.4
Run tested: x86_64, Dell EMC Edge 620, OpenWrt 24.10.4

Description:
At some point after r7, when I no longer had the dual boot device readily available for testing, a regression was introduced that the RPCD script would not actually perform the partition switch.
Thanks to @bd0426 for report and @Jackie264 for fixing the issue.

advanced-reboot.js
* simplify partition number validation in handleAlternativeReboot
* reformat text copy

RPCD script
* obtain OpenWrt version from "$OPENWRT_RELEASE" not "$PRETTY_NAME" for the snapshot compatibility
* refactor parameter processing for boot_partition function
* add logging/better output on error in boot_partition

test.sh
* introduce a test file for heartbeat check of the RPCD script

### DIFF
--- a/applications/luci-app-advanced-reboot/test.sh
+++ b/applications/luci-app-advanced-reboot/test.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+ubus -S call luci.advanced-reboot obtain_device_info | grep 'NO_BOARD_NAME_MATCH' && \
+ubus -S call luci.advanced-reboot boot_partition '{ "number": "1" }' | grep 'NO_BOARD_NAME_MATCH' && \
+ubus -S call luci.advanced-reboot boot_partition '{ "number": "2" }' | grep 'NO_BOARD_NAME_MATCH'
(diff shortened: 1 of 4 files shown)


lorand-horvath:
@systemcrash Please merge this PR, thanks!

applications/luci-app-advanced-reboot/test.sh
systemcrash, line 1: can this be named something more descriptive than 'test'?

stangri: @systemcrash I'm under impression that's the OpenWrt's CI convention that if a script test.sh exists in top level directory of the package, it is ran as part of the CI post-install. Reference: https://github.com/openwrt/packages/blob/4fd9971c5910d8c8b843a2401786821a573c429b/CONTRIBUTING.md?plain=1#L202

PR #8243, opened 2026-01-21
luci-mod-status: improve hide/show button placement and size

efahl:
Align the hide/show buttons with the header of the section that they control.  Make the buttons as tall as the header line, very much improving accessibility on mobile.

---
In 2020 theme on desktop:
image

Bootstrap theme on mobile:
image

2020 theme on mobile:
image

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js
@@ -129,10 +129,11 @@ return view.extend({
 
 			rv.appendChild(E('div', { 'class': 'cbi-section', 'style': 'display: none' }, [
 				E('div', { 'class': 'cbi-title' },[
-					title != '' ? E('h3', title) : '',
-					E('div', [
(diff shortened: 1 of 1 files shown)


efahl:
Is there a  good trick for making the buttons consistently size in width to `max(len("HIDE"), len("SHOW"))`?

systemcrash:
Hmm, CSS isn't my forte - there probably is. Although this doesn't bother me. The different widths actually help signal there is a differing state.

systemcrash:
IOW I'm fine with this now. Merge?

efahl:
I just added a min-width of 4em on them, makes the hide buttons look much better.  Also tried the Spanish translations, looks fine there, so if this runs through ok, I'm ready to merge...

efahl:
The longer buttons in spanish don't seem to be as ugly as the english ones, probably because the length diff is smaller in percentage terms.  In any case, here's with the 4em min:

image

image

systemcrash:
Great! Thank you @efahl

PR #8245, opened 2026-01-21
luci-theme-openwrt-2020: add missing selector

efahl:
The `.label.notice` selector is defined in the bootstrap theme, and used in luci-base, but is not present in the openwrt2020 theme.

---
Before:
image

After:
image


systemcrash:
Not necessarily a bad thing in this case(!) :)

PR #8246, opened 2026-01-21
luci-mod-status: clean up presentation of dhcp section

efahl:
When no dhcp services are available, return null rather than an empty container, thus suppressing display of the section altogether.

Add a title to the outer container so that the hide/show button has something to control.

When both v4 and v6 tables are empty, display a message indicating this rather than showing nothing below the title.


systemcrash:
I think this is OK, altho rather than straying into misleading territory, perhaps we can write something like 'No active leases found/returned' in case rpc or something is playing silly buggers.

efahl:
Yeah, good point.  I just copied that line from a couple places above, I'll change them all to better represent what's going on.  (On my test box, it's literally true, as it only has a wan connection and no clients unless I pull some wires, so it never occurred to me.)

systemcrash:
Great! Thanks @efahl

Issue #8226, opened 2026-01-16
Status page


efahl:
Two issues, only visually related:

1) In modules/luci-mod-status/htdocs/luci-static/resources/view/status/index.js, there's an extra Hide/Show button displayed on empty sections.  See "Extra button" on below screenshot.

2) In the luci-theme-openwrt-2020 (at least, didnt' look at other themes), the button spacing is very odd, being mashed down on top of the content and far away from the section header with which it is associated.  Could some CSS wizard figure out how to move those up and make them appear to be associated with the proper section header (esp. when the section is hidden)?

<img width="1180" height="1265" alt="Image" src="https://github.com/user-attachments/assets/cadb7088-d13b-406e-ae3e-2f4c7d3c5aac" />

systemcrash:
The hide bit is likely a table that didn't get any data. 

The 'proper' styling which was included in @Ansuel's original commit caused general problems elsewhere in the GUI so I had to reduce it to a bare minimum...

Ansuel:
What did I break :(

systemcrash:
Haha. I fixed it. It was the CSS in the table hide commit from ages ago. See 788e47034becc7806485b40dbef9774a14f8ea9d and 6e263428bdd39a7af4b22df661600553df20bdf2

efahl:
So I'm digging around trying to see what is up with the phantom buttons and in the dhcp include file I see a line with `status_leases ... lases` (bottom line in grep output, below), which makes me think there's a typo, should be `leases`.  But to make sure, I grep around and find it in a bunch of places.  Nothing in css files, just in table defs.

Is this legit, or just something that's been copy pastad erroneously for years?

```
$ git grep '\<lases\>'
applications/luci-app-apinger/htdocs/luci-static/resources/view/apinger/overview.js:                    E('table', { 'class': 'table lases' }, [
applications/luci-app-keepalived/htdocs/luci-static/resources/view/keepalived/overview.js:                      E('table', { 'class': 'table lases' }, [
applications/luci-app-keepalived/htdocs/luci-static/resources/view/status/include/35_keepalived.js:                     E('table', { 'class': 'table lases' }, [
applications/luci-app-libreswan/htdocs/luci-static/resources/view/libreswan/overview.js:                        E('table', { 'class': 'table lases' }, [
modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/40_dhcp.js:            const table = E('table', { 'id': 'status_leases', 'class': 'table lases' }, [
[...]

Issue #7925, opened 2025-08-29
luci-app-attendedsysupgrade: After the update, a settings option appeared on the homepage.


icetea0822:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### screenshots or captures

After recently updating this plugin, the option shown in the attached screenshot appeared on the homepage. This setting belongs to luci-app-attendedsysupgrade, but its placement on the homepage looks odd. 

![Image](https://github.com/user-attachments/assets/5b32f3c5-1d34-423a-b115-71ab502f9f29)

### Actual behaviour

An extra option has appeared on the homepage.

### Expected behaviour

I would appreciate it if this option could either be removed or controlled via a toggle. Thank you.

### Steps to reproduce

Upgrade luci-app-attendedsysupgrade to the latest version

### Additional Information

```shell
`
```

### What browsers do you see the problem on?

Chrome

### Relevant log output

```shell

```

edrikk:
Agree.  
I see you've already commented in the relevant commit as well, but to keep it in this issue:  https://github.com/openwrt/luci/commit/f746a52979141df8ffbf7c3774002fead7817724

PR #8250, opened 2026-01-23
luci-app-attendedsysupgrade: rework check-for-new-firmware

efahl:
Do a complete overhaul of the firmware check:

When the Status -> Overview page is loaded, we check if the configuration is set.  If not, we ask the user to choose between enabled or disabled.  Once this is done, it never appears again (much like the "set password" logic in the shell).

As a result, there is no longer a persistent section on the Overview page with a simple toggle eating real estate and playing havoc with the Hide/Show button scheme.

When the setting is enabled, then every time Status -> Overview is loaded, we do the firmware check and display an alert notice as before.  But, the alert notice now contains a button to disable the alerts, so navigation is still simple, you don't have to dig around to figure out how to turn it off.

The logic for version comparisons was cleaned up and simplified.

Fixes: #7925
Fixes: #8226

---
Initial screen, only shown once if you do enable/disable:
image

Firmware check alert:
image


systemcrash:
Great! Merged. Thanks @efahl

PR #8254, opened 2026-01-23
luci-app-attendedsysupgrade: fix syntax error

Seromantic:
Add the missing semicolon at the end of line 5.
Fix the syntax error popup on the LuCI homepage related to 11_upgrades.js.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/status/include/11_upgrades.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/status/include/11_upgrades.js
@@ -2,7 +2,7 @@
 'require baseclass';
 'require fs';
 'require rpc';
-'require uci'
+'require uci';
 'require ui';
 
 


GoetzGoerisch:
Commit message should include:

```
Fixes: f746a52979141df8ffbf7c3774002fead7817724 ("luci-app-attendedsysupgrade: add firmware upgrade check status include")

```

systemcrash:
Merged. Thanks @Seromantic !

efahl:
How did that slip past me?  I tested this on two different devices, but only one browser?  Does firefox's js engine let that stuff through?

In any case, thanks for the fix @Seromantic!

systemcrash:
It wasn't yours @efahl . You didn't touch it. It was in the original f746a52979141df8ffbf7c3774002fead7817724.

:/

efahl:
Yeah, I just chased it back.  It has been without the semicolon for almost a year, and we're just now noticing it?  Wat da eff.

KA2107:
Please backport this fix to 25.12 branch as well. Thanks in advance.

systemcrash:
Apparently it was never included or executed, then.

systemcrash:
picked to 25

hnyman:
@efahl 

> How did that slip past me?  

> It has been without the semicolon for almost a year, and we're just now noticing it? 

I debugged a bit, and the bug manifests only if you do NOT have the file `/etc/config/attendedsysupgrade`. If you have the uci file already there, the missing semicolon does not cause a visible problem (although I didn't check, what action then happen or do not happen).

So, as you have been authoring the LuCI tool, you have likely had that file there in your test devices

efahl:
@hnyman 

I was on a different track: it only manifests in the minified js, so was thinking either the file had not been run through jsmin previously (makefile change???), or jsmin itself had changed (it has not).  Hmm, or maybe the layout/content of the include file has caused jsmin to produce different output???

In any case, the `include/11_upgrades.js` file should always be installed at the same time with `/etc/config/attendedsysupgrade` as they are parts of the same package (through dependency on `attendedsysupgrade-common` -> https://github.com/openwrt/packages/blob/master/utils/attendedsysupgrade-common/files/attendedsysupgrade.defaults).

hnyman:
Ok, I was looking for that file in /rom and didn't find it, and thought that it was the reason. But you seem to be creating it via uci-defaults.   I use un-minified JS, and the missing file then triggers an error in browser debugger.

Out of curiosity, is there a reason why you force an extra flash write by using uci-defaults to create the config file, instead of including the default config file in /etc/config into the image?

efahl:
> Out of curiosity, is there a reason why you force an extra flash write by using uci-defaults to create the config file, instead of including the default config file in /etc/config into the image?

That's way before my time, looks like it has been that way for at least 8 years...  Seems like a good change to move it to a static solution.

efahl:
@hnyman 
I found it, jsmin was indeed failing.  From a 24.10 box:

```
$ ll /www/luci-static/resources/view/status/include/11*
-rw-r--r--  1 root     root  6153 Dec 17 13:08 /www/luci-static/resources/view/status/include/11_upgrades.js
-rw-r--r--  1 root     root  1967 Dec 17 13:08 /www/luci-static/resources/view/status/include/11_upgrades.js.o
```

efahl:
Should we run eslint (or whatever) on the minified output in the CI? Also should we get a better minifier, jsmin is ancient and unmaintained, and apparently can't handle some constructs like this "unterminated regex" thing?

Reproducible by getting an older 11_upgrades.js and minifying it.  My recent changes "fixed" the syntax that was causing grief for jsmin and created/exposed the bug:

```
$ ./staging_dir/hostpkg/bin/jsmin < 11_upgrades.js
'use strict';'require baseclass';'require fs';'require rpc';'require uci''require ui';const callSystemBoard=rpc.declare({object:'system',method:'board'});function showUpgradeNotification(type,boardinfo,new_version,upgrade_info)
{const table_rows=[[_('Firmware Version'),boardinfo.release.version,upgrade_info.version_number],[_('Revision'),boardinfo.release.revision,upgrade_info.version_code],[_('Kernel Version'),boardinfo?.kernel,upgrade_info.linux_kernel?.version],];const table=E('table',{'class':'table'});table.appendChild(E('tr',{'class':'tr'},[E('th',{'class':'th'},[]),E('th',{'class':'th'},[_('Current')]),E('th',{'class':'th'},[_('Available')])]));table_rows.forEach((cols)=>{table.appendChild(E('tr',{'class':'tr'},[E('td',{'class':'td left','width':'33%'},[cols[0]]),E('td',{'class':'td left'},[cols[1]??'?']),E('td',{'class':'td left'},[cols[2]??'?']),]));});ui.addTimeLimitedNotification(_('New Firmware Available'),[E('p',_('A new %s version of OpenWrt is available:').format(type)),table,E('p',[_('Check')+' ',E('a',{href:`/cgi-bin/luci/admin/system/attendedsysupgrade`},_('Attended Sysupgrade')),' '+_('and')+' ',E('a',{href:`https://openwrt.org/releases/${new_version?.split('.').slice(0, 2).join('.')}/notes-${new_version}`},_('release notes')),]),],60000,'notice');};function shouldUpgrade(installed,available)
{if(!available)return false;if(!installed||installed.includes('SNAPSHOT'))return false
const parse=(v)=>v.split(/[-+]/)[0]?.split('.').map(Number);const parseRC=(v)=>v.split(/[-+]/)[1]?.split('').map(Number);const isPrerelease=(v)=>/-/.test(v);

 const [aParts, bParts] = [parse(available), parse(installed)];

 for (let i = 0; i < Math.max(aParts.length, bParts.length); i++) {
  const numA = aParts[i] || 0;
  const numB = bParts[i] || 0;
  if (numA > numB) return true;
  if (numA < numB) return false;
 }

 const [aRC, bRC] = [parseRC(available), parseRC(installed)];

 if (aRC > bRC) return true;
 if (aRC < bRC) return false;

JSMIN Error: Unterminated set in Regular Expression literal.
```

systemcrash:
It looks like the regex was a bit dicey. Given that - in square brackets typically indicates a range, altho it's 'assumed' fine if it comes first or last, like in yours. So safest would have been to escape it, I guess. `[\-+]`.

PR #8241, opened 2026-01-21
luci-app-nut: nut-server: show/hide genericups config

danielfdickinson:
**Description:** Hide configuration applicable only to genericups when not configuring a genericups. This avoids user confusion in thinking settings such as mfr (shown as 'Manufacturer (Display)') are LuCI display only rather than passed to the driver as config (the settings are used as driver configuration items and end up on the ups.conf file; this can cause a driver other then genericups to fail to start).

Closes #8200
luci-app-nut: mfr, model fields not valid on usbhid-ups driver

- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [x] Tested on: (architecture, openwrt version, browser) :white_check_mark:
  bcm27xx/bcm27xx, SNAPSHOT, Firefox 140.7.0 ESR
- [x] \( Preferred ) Mention: @ the original code author for feedback
  @danielfdickinson @systemcrash 
- [ ] \( Preferred ) Screenshot or mp4 of changes:
- [x] \( Optional ) Closes: e.g. openwrt/luci#issue-number
  #8200 
- [ ] \( Optional ) Depends on: e.g. openwrt/packages#pr-number in sister repo
- [x] Description: (describe the changes proposed in this PR)

Issue #8200, opened 2026-01-07
luci-app-nut: mfr, model fields not valid on usbhid-ups driver


rpavlik:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### screenshots or captures

<img width="1363" height="851" alt="Image" src="https://github.com/user-attachments/assets/0f0c81a4-5e16-4e1c-8028-a0d1aa1b73a5" />

### Actual behaviour

1. Set "Manufacturer (Display)" or "Model (Display)" to something non-empty
2. Observe that they show up in /var/etc/nut/ups.conf
3. Observe that the nut server will not start because it says `mfr` or similar is not a valid option

### Expected behaviour

If this is used only in LuCI for display, then it shouldn't be templated into the nut config file.

Otherwise, if it's only for some drivers, would be great if it were conditional on the driver type.

(Also it doesn't appear to restart nutserver on change but that's a different problem. Further, restart nutserver is not enough to get it to work again, I have to replug the UPS or reboot)

### Steps to reproduce

See "Actual Behavior"

### Additional Information

```shell
{
	"kernel": "6.6.119",
	"hostname": "bluebell",
	"system": "ARMv8 Processor rev 4",
	"model": "OpenWrt One",
	"board_name": "openwrt,one",
	"rootfs_type": "squashfs",
	"release": {
		"distribution": "OpenWrt",
		"version": "24.10.5",
		"revision": "r29087-d9c5716d1d",
		"target": "mediatek/filogic",
		"description": "OpenWrt 24.10.5 r29087-d9c5716d1d",
		"builddate": "1766005702"
	}
}
```

### What browsers do you see the problem on?

Firefox

### Relevant log output

```shell
Wed Jan  7 12:09:01 2026 daemon.info usbhid-ups[11516]: Network UPS Tools - Generic HID driver 0.52 (2.8.1)
Wed Jan  7 12:09:01 2026 daemon.info usbhid-ups[11516]: USB communication driver (libusb 1.0) 0.46
Wed Jan  7 12:09:01 2026 daemon.info usbhid-ups[11516]:
Wed Jan  7 12:09:01 2026 daemon.info usbhid-ups[11516]: Fatal error: 'mfr' is not a valid variable name for this driver.
Wed Jan  7 12:09:01 2026 daemon.info usbhid-ups[11516]:
Wed Jan  7 12:09:01 2026 daemon.info usbhid-ups[11516]: Look in the man page or call this driver with -h for a list of
[...]

PR #8264, opened 2026-01-25
luci-app-nut: improve confusing driver option text

danielfdickinson:
**Description:** Some driver options were noted as '(Display)' even though they are added to the `ups.conf` configuration file (this is because the drivers use them in the messages they emit). Since not all drivers support these options and it is quite a large job to manage the which options apply to which of the many drivers for nut, for now we add informational text to inform the user of the situation.

Discussed in #8200 mfr, model fields not valid on usbhid-ups driver and #8241.

As noted in the latter's comments, more than one driver uses the problematic options, so the solution is not as simple as hiding the options for any driver except that one.

**Screenshot:**

image

  bcm27xx/bcm712 (rpi-5), SNAPSHOT r32783-cf84e8ee86, Firefox 140.7.0 ESR
  Partial: #8200 

### DIFF
--- a/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
+++ b/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
@@ -154,7 +154,7 @@ return view.extend({
 		o.optional = true;
 		o.placeholder = 'private';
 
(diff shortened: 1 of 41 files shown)


systemcrash:
Merged. Thanks @danielfdickinson

PR #8270, opened 2026-01-27
luci-mod-network: add autoneg and duplex options

sion-111:
- [x] This PR is not from my *main* or *master* branch, but a *separate* branch
- [x] Each commit has a valid `Signed-off-by: sion-111 <s.piotrowski91@gmail.com>`
- [x] Each commit and PR title has a valid `<package name>: title` first line subject
- [ ] Incremented any `PKG_VERSION` in the Makefile (not applicable)
- [x] Tested on: Banana Pi BPI-R4, OpenWrt 23.05.x, LuCI (browser: Chromium)
- [ ] Mention the original code author for feedback
- [ ] Screenshot or mp4 of changes
- [ ] Closes:
- [ ] Depends on:

### Description

This change adds Auto negotiation and Duplex options to device settings
in LuCI network configuration.

Leaving the value empty keeps the default kernel behavior.

Tested on Banana Pi BPI-R4.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
@@ -1141,6 +1141,22 @@ return baseclass.extend({
 		o.placeholder = dev ? dev._devstate('qlen') : '';
 		o.datatype = 'uinteger';
 
+
+		o = this.replaceOption(s, 'devgeneral', form.ListValue, 'autoneg', _('Auto negotiation'));
+    		o.value('',_(''));
+		o.value('1', _('Enable'));
+    		o.value('0', _('Disable'));
+    		o.default = '';
+    		o.rmempty = true;
+
+
+    		o = this.replaceOption(s, 'devgeneral', form.ListValue, 'duplex', _('Duplex'));
+    		o.value('',_(''));
+    		o.value('1', _('Full duplex'));
+    		o.value('0', _('Half duplex'));
+    		o.default = '';
+    		o.rmempty = true;
+
 		/* PSE / PoE options */
 		if (hasPSE) {
 			o = this.replaceOption(s, 'devpse', form.ListValue, 'pse', _('PoE (C33)'),


modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1141,6 +1141,22 @@ return baseclass.extend({
 		o.placeholder = dev ? dev._devstate('qlen') : '';
 		o.datatype = 'uinteger';
 
+
+		o = this.replaceOption(s, 'devgeneral', form.ListValue, 'autoneg', _('Auto negotiation'));
```
feckert, line 1145: The documentation [sayes](https://openwrt.org/docs/guide-user/network/network_configuration?s[]=0&s[]=1#section_device) that this is a string. However, the option is actually a Boolean and therefore a flag in Luci.
If the option is not set, 1 is assumed, and should not be saved in the configuration. As this is the standard for `autneg` and `duplex` nowadays.

systemcrash: For devs not familiar with the luci formats, this is a viable way of handling a boolean, but it's much more 'clicking' than just a Flag. Flag is preferred 👍

openwrt-ai, line 1145: `_('')` is wrong here: gettext treats the empty msgid specially (it resolves to the PO catalog's header/metadata entry, not an empty string), so this does not yield a blank label and it pollutes the translation catalog. The in-tree convention for the empty `ListValue` entry is `o.value('', _('default'))` (the dominant form across luci-base and luci-mod-network). Apply the same here and to the `duplex` entry on line 1152.

```suggestion
		o.value('', _('default'));
```

---

openwrt-ai (commented):
Reviewed 1 new commit; 1 new issue found.

Backend coupling verified: netifd does consume both `autoneg` and `duplex` (declared as `BLOBMSG_TYPE_BOOL` device options in `device.c` and applied via ethtool in `system-linux.c`), so these UCI options are wired to a real consumer. Note this confirms the existing reviewer feedback that both are booleans — netifd reads them with `blobmsg_get_bool`, so a `Flag` widget (as suggested by @feckert and @systemcrash) matches the backend type better than a three-state `ListValue`. Not repeating that thread, nor the mixed-tab/space indentation already raised inline.

---

feckert:
@sion-111 Still interested in this change? If so, please implement the comments.

PR #8271, opened 2026-01-27
luci-theme-bootstrap: restyle wireless overview

GeorgeSapkin:
Restyle wireless overview to make it clearer which SSIDs belong to which radios.

More of an RFC, as this is surely not the best way to do it. I wanted to do this purely through theming without breaking any other screens.

Before:

Screenshot From 2026-01-27 21-50-33

After:

Screenshot From 2026-01-27 21-50-53 Screenshot From 2026-01-27 21-50-58

Mobile:

image

Mobile (tiny screen):

image

Tested on 25.12-rc3 with desktop Firefox and Chromium, and mobile Firefox and Chrome on Android. Mobile rendering of `↳` seems to be device-dependent, so this bit might need some rethinking.

cc: @systemcrash

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -317,7 +317,7 @@ a:hover {
 }
 
 .nowrap:not(span) {
-/*	Sometimes you need to wrap: badges on narrow screens look weird 
+/*	Sometimes you need to wrap: badges on narrow screens look weird
 	and nowrap defeats other width constraints, so exclude span */
 	white-space: nowrap;
 }
@@ -653,7 +653,7 @@ select,
 }
 
 /* indication line for drag-and-drop in UIDynamicList*/
-.cbi-dynlist > .item.drag-over { 
+.cbi-dynlist > .item.drag-over {
 	border-top: 1px solid var(--text-color-highest);
 }
 
@@ -2212,6 +2212,39 @@ th[data-sort-direction="desc"]::after { content: "\a0\25bc"; }
 	line-height: 1.2em;
 }
 
+@media screen and (min-device-width: 800px) {
+	body[data-page="admin-network-wireless"] .td[data-name="_badge"] {
+		width: 1%;
+	}
+
+	body[data-page="admin-network-wireless"] .ifacebadge {
+		align-items: center;
+		padding: 5px;
+		text-align: left;
+	}
+
+	body[data-page="admin-network-wireless"] .tr[data-sid^="radio"] .td[data-name="_badge"] .ifacebadge {
+		min-width: 170px;
+	}
+
+	body[data-page="admin-network-wireless"] .tr:not([data-sid^="radio"]) .td[data-name="_badge"] div.center {
+		align-items: center;
+		display: flex;
+		justify-content: flex-end;
+	}
+
+	body[data-page="admin-network-wireless"] .tr:not([data-sid^="radio"]) .td[data-name="_badge"] div.center:before {
+		color: #999;
+		content: '\21B3';
+		font-size: 1.2em;
+		margin-right: 10px;
+	}
+
+	body[data-page="admin-network-wireless"] .tr:not([data-sid^="radio"]) .td[data-name="_badge"] .ifacebadge {
+		min-width: 130px;
+	}
+}
+
 .ifacebadge img {
 	width: 32px;
 	height: 32px;
@@ -2320,7 +2353,7 @@ div.cbi-value var,
 }
 
 .cbi-value-first-field {
-	font-weight: bold;	
+	font-weight: bold;
 }
 
 div.cbi-value var[data-tooltip],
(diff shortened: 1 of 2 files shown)


themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -2212,6 +2212,39 @@ th[data-sort-direction="desc"]::after { content: "\a0\25bc"; }
...
+
+	body[data-page="admin-network-wireless"] .tr:not([data-sid^="radio"]) .td[data-name="_badge"] div.center {
+		align-items: center;
+		display: flex;
+		justify-content: flex-end;
+	}
+
+	body[data-page="admin-network-wireless"] .tr:not([data-sid^="radio"]) .td[data-name="_badge"] div.center:before {
+		color: #999;
+		content: '↳';
```
systemcrash: Might I suggest one of these 
https://www.alt-codes.net/arrow_alt_codes.php

e.g. `&#x21B3;` -> &#x21B3;

GeorgeSapkin: A more robust solution would be using an SVG. Maybe there's an open-source icon pack somewhere.

systemcrash: emoji are where it's at. SVG is an extra fetch.

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -2212,6 +2212,39 @@ th[data-sort-direction="desc"]::after { content: "\a0\25bc"; }
...
+
+	body[data-page="admin-network-wireless"] .tr:not([data-sid^="radio"]) .td[data-name="_badge"] div.center {
+		align-items: center;
+		display: flex;
+		justify-content: flex-end;
+	}
+
+	body[data-page="admin-network-wireless"] .tr:not([data-sid^="radio"]) .td[data-name="_badge"] div.center:before {
+		color: #999;
+		content: '\21B3';
```
systemcrash, line 2238: should this also be `\x`?

GeorgeSapkin: Looks like it's the other way around: `\abcd` is the correct syntax.

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
GeorgeSapkin: That's the issue as these are rendered all over the place depending on the browser, OS and device. But I don't have any feelings about the implementation, so I'm up for any changes.

GeorgeSapkin:
@systemcrash any other feedback? :eyes: I think there must be a much better way to structure this.

systemcrash:
Directly in the rendered HTML?

GeorgeSapkin:
I was trying to avoid changing the HTML output so as not to break any other themes. What I was trying to ask was if there's a better way to structure the styles. As these look somewhat off to me.

jow-:
While outside of the scope of your proposed change, I am wondering if it makes sense to keep the hierarchical structure (group by radio) of the wireless overview at all, especially considering modern wireless characteristics such as MLO etc.

Maybe it should be changed into a flat list of wireless network tiles/cards. The hardware details (radio name, channel, band) could be duplicated in each wireless card (e.g. a band indicator in the upper left corner).

Common operations such as "add radio" and "scan" could prompt for the phy (radio) to use if multiple are present.

GeorgeSapkin:
I find the wireless overview page and sub-views confusing so I agree it could be improved. My biggest gripe is that device settings are in the interface dialog, even though they clearly are per device. Furthermore, some settings should be global (e.g. country for reg domain), yet there are at the interface level. This comes up on the forum from time to time where users try to set different device setting for interfaces on the same radio and it doesn't work. It would make sense to me that device settings would be separate.

I'm curios to find out why the interface is that way now, as I feel there's a hidden fence somewhere.

I haven't thought about this too much, so I can't really imagine a simpler UI, apart from having two separate lists: devices and interfaces, and splitting settings accordingly.

systemcrash:
Yeah, it's not a trivial change. There's quite a bit of lifting going on in network and wifi settings. But the layout is not too difficult. I'm slaying dragons at the moment so I won't get to anything like this in a while.

themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
```diff
@@ -2212,6 +2212,39 @@ th[data-sort-direction="desc"]::after { content: "\a0\25bc"; }
 	line-height: 1.2em;
 }
 
+@media screen and (min-device-width: 800px) {
```
openwrt-ai, line 2215: The desktop rules here are gated on `min-device-width: 800px`, but the matching mobile rules in `mobile.css` live in `@media screen and (max-device-width: 600px)`. That leaves viewports in the 601–799px range (tablets, narrow desktop windows) with neither block applied, so the badge layout and the `↳` indentation cue silently don't render there. Was 800 vs 600 intentional, or should one breakpoint be adjusted so the ranges meet (e.g. both at 600px)?

---

openwrt-ai (commented):
Reviewed 1 commit; see the inline comment on the media-query breakpoint gap.

---

Issue #8273, opened 2026-01-28
Feature request: Support for Passkey (WebAuthn) authentication in LuCI


Tokisaki-Galaxy:
### What would you like to see in luci?

Description
I would like to propose adding support for Passkey (WebAuthn) authentication to the LuCI web interface.

Motivation
Currently, LuCI primarily relies on traditional username/password authentication. As the central gateway of a network, an OpenWrt router is a high-value target. Relying solely on passwords poses several risks:
 
 - Brute-force attacks: Even with strong passwords, HTTP interfaces are often targets for automated cracking attempts.
 - Credential theft: Passwords can be phished or leaked.
 - Complexity vs. Security: Users often choose weak passwords to make it easier to log in from mobile devices.

Proposed Solution
Implementing WebAuthn support would allow users to log in using Passkeys (TouchID, FaceID, Windows Hello, or hardware keys like YubiKey). This would provide:
 
 - Phishing resistance: WebAuthn is inherently secure against phishing.
 - Improved Security: Eliminates the risk of brute-force attacks on the web interface.
 - Better UX: Faster and more secure login experience across devices without typing long passwords.

Additional Context
Since LuCI is a critical component, modernizing its authentication mechanism aligns with the industry trend toward passwordless security. I'd like to start a discussion on the technical feasibility and whether this should be implemented as a core feature or an optional luci-app.

What are your thoughts on this?

systemcrash:
I think it makes sense - I've started on this myself. But every security "feature" we build *must* have a recourse. Users being users inevitably get locked out and need a reset. Fortunately we have ssh or it's possible to reboot the router into a safe mode.

Any amendments to luci shall remain usable at all levels of security, and as much as possible, implementation agnostic, and standards compliant: no yubikey only no matter how attractive that 5kb library is, and it will continue to work without Internet.

> * to make it easier to log in from mobile devices.

I don't think I've ever typed it in on mobile. I just mash my browser autofill buttons. It's a device at home, it's not very interesting. :)

jow-:
Proper HTTPS setup (custom CA trusted by client, not just a selfsigned cert) and a proper, *stable* and *unique* hostname are preconditions for this. These are significant hurdles and very easy to break.

- There is no device agnostic way to extract vendor serials on OpenWrt, but a unique name could be formed based on the primary MAC address of the device
- When a non-unique hostname such as "openwrt.lan" is chosen, it becomes impossible to manage multiple devices
[...]

PR #8280, opened 2026-01-29
luci-plugin-2fa: init checkin

Tokisaki-Galaxy:
- [x] **(N/A)** Incremented :up: any `PKG_VERSION` in the Makefile
- [x] \( Optional ) Closes: 
https://github.com/openwrt/luci/pull/7069
https://github.com/openwrt/luci/issues/8273
- [x] \( Optional ) Depends on: openwrt/luci #8281 

https://github.com/user-attachments/assets/d5a91dd9-b915-440e-932b-7044709e5125

image

the app must changed LuCI core file because:
 - No hook point exists between password verification and session creation
 - External packages cannot inject authentication logic
 - No plugin discovery mechanism in the original code

---

## Security Measures
Constant-time string comparison to prevent timing attacks
Username sanitization to prevent command injection
Array-based popen to prevent shell injection
OTP format validation (exactly 6 digits)
Session destroyed if 2FA verification fails
Uses authenticated session username to prevent bypass attacks

origin repo https://github.com/Tokisaki-Galaxy/luci-app-2fa

### DIFF
--- a/plugins/luci-plugin-2fa/Makefile
+++ b/plugins/luci-plugin-2fa/Makefile
@@ -1,0 +1,23 @@
+#
+# Copyright (C) 2026 tokisaki galaxy <moebest@outlook.jp>
+# Copyright (C) 2024 Christian Marangi <ansuelsmth@gmail.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-plugin-2fa
+
+PKG_LICENSE:=Apache-2.0
+PKG_MAINTAINER:=tokisaki galaxy <moebest@outlook.jp>
+PKG_DESCRIPTION:=LuCI 2-Factor Authentication Plugin
+
+LUCI_TITLE:=LuCI 2-Factor Authentication
+LUCI_DEPENDS:=+luci-base +luci-lib-uqr +ucode-mod-struct +ucode-mod-digest +ucode-mod-log
+LUCI_PKGARCH:=all
+LUCI_URL:=https://github.com/tokisaki-galaxy/luci-plugin-2fa
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot
--- a/plugins/luci-plugin-2fa/root/usr/libexec/generate_otp.uc
+++ b/plugins/luci-plugin-2fa/root/usr/libexec/generate_otp.uc
@@ -1,0 +1,125 @@
+#!/usr/bin/ucode
+
+// Copyright (c) 2024 Christian Marangi <ansuelsmth@gmail.com>
+// Copyright (c) 2026 tokiskai galaxy <moebest@outlook.jp>
+import { cursor } from 'uci';
+import { sha1 } from 'digest';
+import { pack } from 'struct';
+
+const base32_decode_table = (function() {
+	let t = {};
+	for (let i = 0; i < 26; i++) { t[ord('A') + i] = i; t[ord('a') + i] = i; }
+	for (let i = 0; i < 6; i++) { t[ord('2') + i] = 26 + i; }
+	return t;
+})();
+
+function decode_base32_to_bin(string) {
+	let clean = replace(string, /[\s=]/g, "");
+	if (length(clean) == 0) return null;
+
+	let bin = "";
+	let buffer = 0;
+	let bits = 0;
+
+	for (let i = 0; i < length(clean); i++) {
+		let val = base32_decode_table[ord(clean, i)];
+		if (val === null || val === undefined) continue;
+
+		buffer = (buffer << 5) | val;
+		bits += 5;
+
+		if (bits >= 8) {
+			bits -= 8;
+			bin += chr((buffer >> bits) & 0xff);
+		}
+	}
+	return bin;
+}
+
+function calculate_hmac_sha1(key, message) {
+	const blocksize = 64;
+	if (length(key) > blocksize) key = hexdec(sha1(key));
+	while (length(key) < blocksize) key += chr(0);
+
+	let o_key_pad = "", i_key_pad = "";
+	for (let i = 0; i < blocksize; i++) {
+		let k = ord(key, i);
+		o_key_pad += chr(k ^ 0x5c);
+		i_key_pad += chr(k ^ 0x36);
+	}
+	let inner_hash = hexdec(sha1(i_key_pad + message));
+	return sha1(o_key_pad + inner_hash);
+}
+
+function calculate_otp(secret_base32, counter_int) {
+	let secret_bin = decode_base32_to_bin(secret_base32);
+	if (!secret_bin) return null;
+
+	let counter_bin = pack(">Q", counter_int);
+
+	let hmac_hex = calculate_hmac_sha1(secret_bin, counter_bin);
+
+	let offset = int(substr(hmac_hex, 38, 2), 16) & 0xf;
+	let binary_code = int(substr(hmac_hex, offset * 2, 8), 16) & 0x7fffffff;
+
+	return sprintf("%06d", binary_code % 1000000);
+}
+
+let username = ARGV[0];
+let no_increment = false;
+let custom_time = null;
+let plugin_uuid = null;
+
+for (let i = 1; i < length(ARGV); i++) {
+	let arg = ARGV[i];
+	if (arg == '--no-increment') {
+		no_increment = true;
+	} else if (substr(arg, 0, 7) == '--time=') {
+		let time_str = substr(arg, 7);
+		if (match(time_str, /^[0-9]+$/)) {
+			custom_time = int(time_str);
+			if (custom_time < 946684800 || custom_time > 4102444800) custom_time = null;
+		}
+	} else if (substr(arg, 0, 9) == '--plugin=') {
+		let uuid_str = substr(arg, 9);
+		if (match(uuid_str, /^[0-9a-fA-F]{32}$/)) plugin_uuid = uuid_str;
+	}
+}
+
+if (!username || username == '') exit(1);
+
+let ctx = cursor();
+let otp_type, secret, counter, step;
+
+if (plugin_uuid) {
+	otp_type = ctx.get('luci_plugins', plugin_uuid, 'type_' + username) || 'totp';
+	secret = ctx.get('luci_plugins', plugin_uuid, 'key_' + username);
+	counter = int(ctx.get('luci_plugins', plugin_uuid, 'counter_' + username) || '0');
+	step = int(ctx.get('luci_plugins', plugin_uuid, 'step_' + username) || '30');
+} else {
+	otp_type = ctx.get('2fa', username, 'type') || 'totp';
+	secret = ctx.get('2fa', username, 'key');
+	counter = int(ctx.get('2fa', username, 'counter') || '0');
+	step = int(ctx.get('2fa', username, 'step') || '30');
+}
+
+if (!secret) exit(1);
+
+let otp;
+if (otp_type == 'hotp') {
+	otp = calculate_otp(secret, counter);
+	if (!no_increment && otp) {
+		if (plugin_uuid) {
+			ctx.set('luci_plugins', plugin_uuid, 'counter_' + username, '' + (counter + 1));
+			ctx.commit('luci_plugins');
+		} else {
+			ctx.set('2fa', username, 'counter', '' + (counter + 1));
+			ctx.commit('2fa');
+		}
+	}
+} else {
+	let timestamp = (custom_time != null) ? custom_time : time();
+	otp = calculate_otp(secret, int(timestamp / step));
+}
+
+if (otp) print(otp); else exit(1);
--- a/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
+++ b/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
@@ -1,0 +1,720 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2026 LuCI 2FA Plugin Contributors
+//
+// LuCI Authentication Plugin: Two-Factor Authentication (2FA/OTP)
+//
+// This plugin implements TOTP/HOTP verification as an additional
+// authentication factor for LuCI login.
+//
+// Adapted for master's plugin architecture (luci_plugins UCI config)
+
+'use strict';
+
+import { popen, readfile, writefile, open } from 'fs';
+import { connect } from 'ubus';
+import { cursor } from 'uci';
+import { syslog, LOG_INFO, LOG_WARNING, LOG_AUTHPRIV } from 'log';
+
+const PLUGIN_UUID = 'bb4ea47fcffb44ec9bb3d3673c9b4ed2';
+
+// Default minimum valid time (2026-01-01 00:00:00 UTC)
+// TOTP depends on accurate system time. If system clock is not calibrated
+// (e.g., after power loss on devices without RTC battery), TOTP codes will
+// be incorrect and users will be locked out. This threshold disables TOTP
+// when system time appears uncalibrated.
+const DEFAULT_MIN_VALID_TIME = 1767225600;
+
+// Rate limit state file
+const RATE_LIMIT_FILE = '/tmp/2fa_rate_limit.json';
+const RATE_LIMIT_LOCK_FILE = '/tmp/2fa_rate_limit.lock';
+const DEFAULT_PRIORITY = 15;
+const RATE_LIMIT_STALE_SECONDS = 86400;
+let RATE_LIMIT_LOCK_HANDLE = null;
+let ubus = connect();
+
+function get_priority() {
+	let ctx = cursor();
+	let value = ctx.get('luci_plugins', PLUGIN_UUID, 'priority');
+
+	if (!value || !match(value, /^-?[0-9]+$/))
+		return DEFAULT_PRIORITY;
+
+	return int(value);
+}
+
+function get_system_min_valid_time_fallback() {
+	let newest = 0;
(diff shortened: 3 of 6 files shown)


Neustradamus:

Do not forget to solve:
```
🔶 Author name (Tokisaki-Galaxy) seems to be a nickname or an alias
🔶 Committer name (Tokisaki-Galaxy) seems to be a nickname or an alias
```

stangri:
Looks very polished @Tokisaki-Galaxy!

Does this use TOTP? If the OpenWrt device doesn't have RTC and is offline or generally doesn't have correct time, does SSH become the only option to login?

Is there a README/instructions (ideally a hint on failed attempt) on how to disable 2FA from SSH/CLI for people who may be locked out of WebUI and can't read code ahead of time?

Tokisaki-Galaxy:
@stangri 

Please refer to the newly added video at the top of the description section for details.

The plugin can choose either TOTP or HOTP, but TOTP is recommended.
For RTC clock that is not synchronized (for example, the year is 1970), users can choose the 2fa behavior (strict mode) through options. By default, LAN areas are allowed to bypass 2FA for login, while non-LAN areas are prohibited. It can be configured to allow anyone to bypass 2FA for login.
 
Regarding the documentation for SSH/CLI, I'm not quite sure where it should be placed. Should it be directly included in the web UI? But if users don't read it carefully, they might not be able to log in and it would be impossible to see the result. Do you have any suggestions? 

Previously, it was planned to add backup code, but this was abandoned because it would cause the bitward auto-fill function to become unusable and the complexity would be too high.

Tokisaki-Galaxy (commented):
TOTP standard requires that the key is a string encoded by Base32, which will be used as the key of HMAC after decoding.
But now the code is to encode_base32 plaintext key again _ base 32, which will lead to the inconsistency between the generated OTP and standard TOTP tools (such as Google Authenticator).

The correct way:
decode the Base32 string into binary, and then use it directly in HMAC.
current code is encode_base32, which causes the key to be misplaced.

systemcrash:
@Tokisaki-Galaxy let's see whether the plugin structure needs some modifications to handle 2FA stuff.

Tokisaki-Galaxy:
@systemcrash 
branch have update

applications/luci-app-2fa/root/usr/libexec/generate_otp.uc
```diff
@@ -0,0 +1,396 @@
...
+		if (bits_in_buffer >= 8) {
+			bits_in_buffer -= 8;
+			push(out, (buffer >> bits_in_buffer) & 0xff);
+		}
+	}
+	
+	return out;
+}
+
+function calculate_sha1(binary_string) {
```
systemcrash: ucode has a built-in for this. https://ucode.mein.io/module-digest.html#sha1

Tokisaki-Galaxy: ok, most of this document was written by the author before. I didn't pay much attention to it. I'll revise it. Thank you very much.

Tokisaki-Galaxy:
Refactored `generate_otp.uc` to leverage ucode native modules and built-in features. replaces the manual implementations, improving performance

plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,633 @@
...
+
+// Sanitize username to prevent command injection
+function sanitize_username(username) {
+	if (!match(username, /^[a-zA-Z0-9_.+-]+$/))
+		return null;
+	return username;
+}
+
+// Validate IP address (IPv4 or IPv6)
+function is_valid_ip(ip) {
```
systemcrash, line 100: For this function, something I've done in the past is use the socket library. 
https://ucode.mein.io/module-socket.html#sockaddr

There is also https://ucode.mein.io/module-core.html#iptoarr 

Then simply check error which should inform you whether the IP is valid or not.

applications/luci-app-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,633 @@
...
+	
+	return false;
+}
+
+// Get all LAN interface subnets from OpenWrt network configuration
+function get_lan_subnets() {
+	let subnets = [];
+	
+	// Try ubus call to get LAN interface status
+	let fd = popen('ubus call network.interface.lan status 2>/dev/null', 'r');
```
systemcrash: See `import { connect } from 'ubus';` - take a peek in dispatcher.uc.

applications/luci-app-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,633 @@
...
+	if (fd) {
+		let output = fd.read('all');
+		fd.close();
+		
+		if (output) {
+			let status = json(output);
+			if (status && status['ipv4-address']) {
+				for (let addr in status['ipv4-address']) {
+					if (addr.address && addr.mask) {
+						let ip_parts = split(addr.address, '.');
```
systemcrash: See https://ucode.mein.io/module-core.html#iptoarr

applications/luci-app-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,633 @@
...
+	return state;
+}
+
+// Save rate limit state
+function save_rate_limit_state(state) {
+	writefile(RATE_LIMIT_FILE, sprintf('%J', state));
+}
+
+function lock_rate_limit_state() {
+	let fd = popen('lock -w 5 ' + RATE_LIMIT_LOCK_FILE + ' >/dev/null 2>&1; echo $?', 'r');
```
systemcrash: https://ucode.mein.io/module-fs.file.html#lock

applications/luci-app-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,633 @@
...
+		return false;
+
+	let status = trim(fd.read('all') || '');
+	fd.close();
+
+	return status == '0';
+}
+
+function unlock_rate_limit_state() {
+	let fd = popen('lock -u ' + RATE_LIMIT_LOCK_FILE + ' >/dev/null 2>&1', 'r');
```
systemcrash: https://ucode.mein.io/module-fs.file.html#lock u

applications/luci-app-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,633 @@
...
+	if (fd)
+		fd.close();
+}
+
+// Check rate limit
+function check_rate_limit(ip) {
```
systemcrash: This and consume_rate_limit_attempt are very similar - consider a branch in a single function.

applications/luci-app-2fa/htdocs/luci-static/resources/uqr.js
systemcrash, line 1: I wonder whether we can use luci-lib-uqr instead?

Tokisaki-Galaxy: yes, this plugin inherit old PR, that time donot have this lib uqr

plugins/luci-plugin-2fa/Makefile
```diff
@@ -0,0 +1,23 @@
+#
+# Copyright (C) 2026 tokisaki galaxy <moebest@outlook.jp>
+# Copyright (C) 2024 Christian Marangi <ansuelsmth@gmail.com>
+#
+# This is free software, licensed under the Apache License, Version 2.0.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-app-2fa
```
systemcrash: OK - since it's an auth plugin, let's house it in the plugin folder.

plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,699 @@
...
+import { syslog, LOG_INFO, LOG_WARNING, LOG_AUTHPRIV } from 'log';
+
+const PLUGIN_UUID = 'bb4ea47fcffb44ec9bb3d3673c9b4ed2';
+
+// Default minimum valid time (2026-01-01 00:00:00 UTC)
+// TOTP depends on accurate system time. If system clock is not calibrated
+// (e.g., after power loss on devices without RTC battery), TOTP codes will
+// be incorrect and users will be locked out. This threshold disables TOTP
+// when system time appears uncalibrated.
+const DEFAULT_MIN_VALID_TIME = 1767225600;
```
systemcrash, line 25: There is also the trick that the system uses - it sets the minimum clock value to the last file-change time found in the `/etc` FS. Check `openwrt/package/base-files/files/etc/init.d/sysfixtime` for how it works. If this is available as a 'fallback', that's also not bad.

Tokisaki-Galaxy: I didn't know there was sysfixtime before, so I changed to the algorithm that gives priority to this. Sysfixtime in the system does not provide an interface, so it seems that it cannot be called directly.

plugins/luci-plugin-2fa/Makefile
Tokisaki-Galaxy: Ok, but can it be packaged and compiled correctly by the package management system after it is put into the plugin folder? I haven't tried it.

plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
systemcrash: I wonder whether I'm wrong here - because the system performs this check at startup by default so this is already baked into the system clock. It's just that we don't know that. Since TOTP effectively mandates that client and server are more or less time synched, it probably doesn't make sense to enable TOTP with a 'wrong' clock, or make a 'best-effort' with this fallbacks. Maybe it's better just to avoid this and put a note about enabling NTP in the config UI, or better, flash a warning about no NTP servers being found (and TOTP therefore won't work without them).

plugins/luci-app-2fa/root/usr/libexec/generate_otp.uc
```diff
@@ -0,0 +1,133 @@
...
+import { sha1 } from 'digest';
+import { pack } from 'struct';
+
+function hex_to_bin(hex) {
+	let bin = "";
+	for (let i = 0; i < length(hex); i += 2) {
+		bin += chr(int(substr(hex, i, 2), 16));
+	}
+	return bin;
+}
```
systemcrash: There are https://ucode.mein.io/module-core.html#hexenc and https://ucode.mein.io/module-core.html#hexdec

plugins/luci-app-2fa/root/usr/libexec/generate_otp.uc
```diff
@@ -0,0 +1,133 @@
...
+	}
+	let inner_hash = hex_to_bin(sha1(i_key_pad + message));
+	return sha1(o_key_pad + inner_hash);
+}
+
+function calculate_otp(secret_base32, counter_int) {
+	let secret_bin = decode_base32_to_bin(secret_base32);
+	if (!secret_bin) return null;
+
+	let counter_bin = pack(">I", 0) + pack(">I", counter_int);
```
systemcrash: I wonder whether it's simpler to do:
```js
pack(">Q", counter_int)
```

It should work the same I think. ( The RFC 4226 and 6238 use a single 8 byte integer )

plugins/luci-plugin-2fa/root/usr/libexec/generate_otp.uc
```diff
@@ -0,0 +1,133 @@
...
+function calculate_otp(secret_base32, counter_int) {
+	let secret_bin = decode_base32_to_bin(secret_base32);
+	if (!secret_bin) return null;
+
+	let counter_bin = pack(">I", 0) + pack(">I", counter_int); 
+
+	let hmac_hex = calculate_hmac_sha1(secret_bin, counter_bin);
+
+	let offset = int(substr(hmac_hex, 38, 2), 16) & 0xf;
+	let binary_code = int(substr(hmac_hex, offset * 2, 8), 16) & 0x7fffffff;
```
systemcrash, line 63: Do we get the same result if we do like the RFC?:
```js
offset = hmac[19] & 0xf;
binary_code =
  ((hmac[offset] & 0x7f) << 24) |
  (hmac[offset+1] << 16) |
  (hmac[offset+2] << 8) |
  (hmac[offset+3]);
```
Need to hexdec sha1 output.

plugins/luci-app-2fa/root/usr/libexec/generate_otp.uc
Tokisaki-Galaxy: I try >Q past, but apparent some bug :( later I try again

plugins/luci-plugin-2fa/root/usr/libexec/generate_otp.uc
Tokisaki-Galaxy: yes, in separate repo https://github.com/tokisaki-galaxy/luci-app-2fa/commit/6ca83edc665380b3ac11017bfca529d2933d66ec I past use this.
I think now more simple? If you need, I recovery this

Tokisaki-Galaxy: result same

plugins/luci-app-2fa/root/usr/libexec/generate_otp.uc
Tokisaki-Galaxy: ok, work normal

plugins/luci-plugin-2fa/root/usr/libexec/generate_otp.uc
systemcrash: OK :)

plugins/luci-app-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -0,0 +1,720 @@
...
+		verify: function(http, user) {
+			let client_ip = get_client_ip(http);
+
+			// Check if IP is whitelisted
+		if (client_ip && is_ip_whitelisted(client_ip)) {
+			syslog(LOG_INFO|LOG_AUTHPRIV,
+				sprintf("luci: 2FA bypassed for %s from %s due to IP whitelist",
+					user || '?', client_ip || '?'));
+			return { success: true, whitelisted: true };
+		}
```
systemcrash: indent

systemcrash:
OK - thanks for your patience. This is looking good now. LGTM.

Tokisaki-Galaxy:
Thank you so much for your patience and thorough review!
To be honest, I’m truly grateful—this is the first time I’ve had someone review my PRs with such care and provide such constructive feedback.

As a non-CS university student with no professional work experience yet, I’ve learned a lot through this process. This project has given me my first real taste of large-scale team collaboration, and your guidance has been invaluable in helping me improve.
Thank you again for your time and for helping me grow!

systemcrash:
To highlight that this is a plugin, how about naming it `luci-plugin-2fa`?

Tokisaki-Galaxy:
great! I have rename and rebase commit

plugins/luci-plugin-2fa/Makefile
systemcrash: Here also :)

Tokisaki-Galaxy:
sorry, I forget edit Makefile, have rename and update URL





***@***.***



&nbsp;




------------------&nbsp;Original&nbsp;------------------
From:                                                                                                                        "openwrt/luci"                                                                                    ***@***.***&gt;;
Date:&nbsp;Tue, Apr 7, 2026 09:18 PM
***@***.***&gt;;
***@***.******@***.***&gt;;
Subject:&nbsp;Re: [openwrt/luci] luci-app-2fa: init checkin (PR #8280)





 
@systemcrash commented on this pull request.
 
 
In plugins/luci-plugin-2fa/Makefile:
 &gt; @@ -0,0 +1,23 @@ +# +# Copyright (C) 2026 tokisaki galaxy ***@***.***&gt; +# Copyright (C) 2024 Christian Marangi ***@***.***&gt; +# +# This is free software, licensed under the Apache License, Version 2.0. +# + +include $(TOPDIR)/rules.mk + +PKG_NAME:=luci-app-2fa  
Here also :)

systemcrash:
OK - let's see how it works :)

Tokisaki-Galaxy:
Thank you very much!

Neustradamus:
@Tokisaki-Galaxy: Thanks for your good work and thanks to @systemcrash for merging!

BKPepe:
@Ansuel what do you think about this merged PR? Can you review this?

PR #8282, opened 2026-01-29
luci-mod-network: add speed choices via rpc device status

sion-111:
This change adds a speed selection field to device settings in LuCI.

Available speeds are populated dynamically using `network.device status`
via RPC, while still allowing manual custom values to be entered.

ACL permissions were extended to allow read-only access to
`network.device status`.

Tested on Banana Pi BPI-R4.

- [x] This PR is not from my *main* or *master* branch, but a *separate* branch
- [x] Each commit has a valid `Signed-off-by: sion-111 <s.piotrowski91@gmail.com>`
- [x] Each commit and PR title has a valid `<package name>: title` first line subject
- [ ] Incremented any `PKG_VERSION` in the Makefile (not applicable)
- [x] Tested on: Banana Pi BPI-R4
- [ ] Screenshot or mp4 of changes

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
@@ -8,6 +8,7 @@
 'require baseclass';
 'require validation';
 'require tools.widgets as widgets';
+'require rpc';
 
 function validateAddr(section_id, value) {
 	if (value == '')
@@ -362,6 +363,25 @@ var cbiTagValue = form.Value.extend({
 
 	remove: function() {}
 });
+
+var callDeviceStatus = rpc.declare({
+        object: 'network.device',
+        method: 'status',
+        params: [ 'name' ]
+});
+
+function extractSupportedSpeeds(status) {
+        var sup = (status && (status['link-supported'] || status['link_supported'])) || [];
+        var uniq = Object.create(null);
+
+        L.toArray(sup).forEach(function (v) {
+                var m = String(v).match(/^(\d+)/);
+                if (m) uniq[m[1]] = true;
+        });
+
+        return Object.keys(uniq).sort(function (a, b) { return (+a) - (+b); });
+}
+
 
 return baseclass.extend({
 
@@ -1140,6 +1160,35 @@ return baseclass.extend({
 		o = this.replaceOption(s, 'devgeneral', form.Value, 'txqueuelen', _('TX queue length'));
 		o.placeholder = dev ? dev._devstate('qlen') : '';
 		o.datatype = 'uinteger';
+
+		o = this.replaceOption(s, 'devgeneral', form.Value, 'speed', _('Speed'));
+
+o.editable = true;
+o.placeholder = '0';
+o.datatype = 'uinteger';
+o.rmempty = true;
+o.default=''
+
+
+o.value('', '');
+o.load = function (section_id) {
+
+        var ifn = this.section.formvalue(section_id, 'ifname_single')
+                || this.section.formvalue(section_id, 'name_simple')
+                || uci.get('network', section_id, 'name')
+                || (dev ? dev.getName() : '');
+
+        ifn = String(ifn || '').replace(/\.\d+$/, '');
+
+        if (!ifn) ifn = 'eth1';
+
+        var self = this;
+        return callDeviceStatus(ifn).then(function (st) {
+                extractSupportedSpeeds(st).forEach(function (sp) {
+                        self.value(sp, sp + ' Mbit/s');
+                });
+        });
+};
 
 		/* PSE / PoE options */
 		if (hasPSE) {
--- a/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
+++ b/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
@@ -16,8 +16,8 @@
 				"iwinfo": [ "assoclist", "countrylist", "freqlist", "txpowerlist" ],
 				"luci": [ "getSwconfigFeatures", "getSwconfigPortState" ]
 			},
-			"uci": [ "dhcp", "firewall", "network", "wireless", "system" ]
-		},
+			"uci": [ "dhcp", "firewall", "network", "wireless", "system" ],
+			"network.device": [ "status" ]		},
 		"write": {
 			"cgi-io": [ "exec" ],
 			"file": {


systemcrash:
Can you fix the formalities, please?

systemcrash:
ping

modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
```diff
@@ -16,8 +16,8 @@
 				"iwinfo": [ "assoclist", "countrylist", "freqlist", "txpowerlist" ],
 				"luci": [ "getSwconfigFeatures", "getSwconfigPortState" ]
 			},
-			"uci": [ "dhcp", "firewall", "network", "wireless", "system" ]
-		},
+			"uci": [ "dhcp", "firewall", "network", "wireless", "system" ],
+			"network.device": [ "status" ]		},
```
openwrt-ai, line 20: This ubus grant is placed as a sibling of `"uci"`, outside the `"ubus"` object. rpcd only honors ubus object/method grants nested under the `"ubus"` key, so as written this entry grants nothing and `network.device status` will still be denied for this ACL. It also leaves a stray `]\t\t},` (closing array followed by tabs then the brace). The grant belongs inside the existing `"ubus": { ... }` block above (line 14-18).

Note also that `network.device status` is already granted to the network config page via the `luci-base-network-status` ACL group (`modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json` line 42: https://github.com/openwrt/luci/blob/0a4e7b338a55938aab875644a6a842487f2abcb4/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json#L42), so this addition may be redundant — please verify whether any new grant is actually needed.

```suggestion
				"iwinfo": [ "assoclist", "countrylist", "freqlist", "txpowerlist" ],
				"luci": [ "getSwconfigFeatures", "getSwconfigPortState" ],
				"network.device": [ "status" ]
			},
			"uci": [ "dhcp", "firewall", "network", "wireless", "system" ]
		},
```

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1141,6 +1161,35 @@ return baseclass.extend({
...
+
+        if (!ifn) ifn = 'eth1';
+
+        var self = this;
+        return callDeviceStatus(ifn).then(function (st) {
+                extractSupportedSpeeds(st).forEach(function (sp) {
+                        self.value(sp, sp + ' Mbit/s');
+                });
+        });
+};
```
openwrt-ai, line 1191: This whole block is written at column 0 (and uses spaces for the `o.load` body), whereas the surrounding code in this function is tab-indented one level deeper. Re-indent the option setup and the `load` handler to match the file's existing tab indentation so it reads consistently with the adjacent options (e.g. the `devpse`/`devadvanced` blocks).

Also `o.default=''` is missing the trailing semicolon (every other statement here is terminated), and `o.value('', '')` adds an empty-label/empty-value entry on top of the `placeholder = '0'` — that produces a redundant blank choice in the dropdown; drop it unless an explicit empty choice is intended.

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1141,6 +1161,35 @@ return baseclass.extend({
...
+o.load = function (section_id) {
+
+        var ifn = this.section.formvalue(section_id, 'ifname_single')
+                || this.section.formvalue(section_id, 'name_simple')
+                || uci.get('network', section_id, 'name')
+                || (dev ? dev.getName() : '');
+
+        ifn = String(ifn || '').replace(/\.\d+$/, '');
+
+        if (!ifn) ifn = 'eth1';
```
openwrt-ai, line 1183: Falling back to a hardcoded `'eth1'` when the device name can't be resolved means the dropdown gets populated with the supported speeds of an unrelated interface (which may not even exist) and presents them as valid choices for the device being edited. On a device with no `eth1` the RPC call also just fails. If no name can be determined, skip the RPC and leave the field as a free-form input rather than querying an arbitrary device:

```suggestion
        if (!ifn)
                return;
```

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -363,6 +364,25 @@ var cbiTagValue = form.Value.extend({
...
 });
 
+var callDeviceStatus = rpc.declare({
+        object: 'network.device',
+        method: 'status',
+        params: [ 'name' ]
+});
+
+function extractSupportedSpeeds(status) {
+        var sup = (status && (status['link-supported'] || status['link_supported'])) || [];
```
openwrt-ai, line 374: nit: `status['link_supported']` is dead — netifd emits this array only under the hyphenated key `link-supported` (see `system_if_dump_info` in netifd `system-linux.c`: https://github.com/openwrt/netifd/blob/e45b890fa532bb89380b78008c2cc70bfad2c4c7/system-linux.c#L3463). The underscore variant never appears, so the `|| status['link_supported']` fallback can be dropped.

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1141,6 +1161,35 @@ return baseclass.extend({
 		o.placeholder = dev ? dev._devstate('qlen') : '';
 		o.datatype = 'uinteger';
 
+		o = this.replaceOption(s, 'devgeneral', form.Value, 'speed', _('Speed'));
```
openwrt-ai, line 1164: Question on whether the `speed` option alone is actionable: in netifd, a configured `speed` is only pushed to the NIC inside the `if (s->flags & DEV_OPT_AUTONEG)` / `if (!s->autoneg)` path — see `system_apply_ethtool_settings` in `system-linux.c` (https://github.com/openwrt/netifd/blob/e45b890fa532bb89380b78008c2cc70bfad2c4c7/system-linux.c#L2498-L2505). Without also setting `autoneg '0'` (and typically `duplex`), `speed` does not change the forced link speed. The head branch name/description mentions "duplex and Auto negotiation fields", but this diff only adds `speed`. Is the intent to add those companion options too, or to document that `speed` requires `autoneg` disabled to take effect?

---

openwrt-ai (commented):
Reviewed 1 new commit; see inline comments. Main item: the new `network.device` ACL grant is placed outside the `"ubus"` object so rpcd will not honor it (and the same grant already exists in `luci-base`). Plus indentation/style issues in the speed block, a hardcoded `eth1` fallback, and a question about whether `speed` alone is actionable without `autoneg`/`duplex`.

---

feckert:
@sion-111 till interested in this change? If so, please implement the comments and resolve the conflict.

BKPepe:
I think this can be closed. @feckert Shall I enable stale bot as we do have in the packages repository? Would it help for you?

Issue #8109, opened 2025-11-23
Feature request: Associated Stations history


1PSafe:
### What would you like to see in luci?

Hello,
i d like to see what wifi devices have been connected in the past with history.
AVM fritzbox has such a featury called "not connected stations"

mfg

DeborahOlaboye:
Hi, I'd like to work on this issue.

1PSafe:
> Hi, I'd like to work on this issue.

do you need additional information ?

systemcrash:
Nothing in the system keeps this around - all of the data is live. But have a go and see what turns up.

DeborahOlaboye:
> Nothing in the system keeps this around - all of the data is live. But have a go and see what turns up.

Thanks for this @systemcrash. Yes, I noticed that `iwinfo assoclist` only returns currently associated stations, nothing persists once a client disconnects.

My approach is to add a new `luci-app-wifihistory` package with a lightweight background service that periodically polls `iwinfo assoclist` via ubus, resolves hostnames through `luci-rpc getHostHints`, and merges the results into a persistent JSON file at `/var/lib/wifihistory/history.json`. Stations that disappear from the live assoclist are marked as disconnected rather than removed, preserving their last known signal, hostname, and first/last seen timestamps.

ptpt52:
Hi @1PSafe and @DeborahOlaboye,


To clarify, do you consider it necessary to record the history of devices that have connected and disconnected multiple times? Would adding this functionality be beneficial for the overall feature?

PR #8293, opened 2026-02-03
luci-app-wifihistory: add WiFi station history tracking

DeborahOlaboye:
>

- [ ] \( Preferred ) Mention: @1PSafe for feedback
- [x] \( Optional ) Closes: openwrt/luci#8109
- [ ] \( Optional ) Depends on:
- [x] Description:

## Description

Add a new `luci-app-wifihistory` package that tracks WiFi associated stations over time, addressing the lack of persistent station history in OpenWrt.

Currently `iwinfo assoclist` only returns live data once a client disconnects, it disappears. This package solves that by:

- Running a lightweight background service that polls `iwinfo assoclist` via ubus every 30 seconds
- Resolving hostnames/IPs through `luci-rpc getHostHints`
- Persisting all seen stations to `/var/lib/wifihistory/history.json`
- Marking stations as disconnected (rather than removing) when they leave the live assoclist

The LuCI view at **Status > Station History** displays a table with:
- Connected/Disconnected status
- MAC address, Hostname, IP
- Network (SSID)
- Last known Signal/Noise
- First seen / Last seen timestamps

Similar to AVM FritzBox's "not connected stations" feature.

### DIFF
--- a/applications/luci-app-wifihistory/Makefile
+++ b/applications/luci-app-wifihistory/Makefile
@@ -1,0 +1,11 @@
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for WiFi Station History
+LUCI_DEPENDS:=+luci-base +rpcd-mod-iwinfo +ucode +ucode-mod-fs +ucode-mod-ubus
+LUCI_DESCRIPTION:=Track and display history of WiFi associated stations
+
+PKG_LICENSE:=Apache-2.0
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
+++ b/applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
@@ -1,0 +1,181 @@
+'use strict';
+'require view';
+'require poll';
+'require rpc';
+'require ui';
+
+const callGetHistory = rpc.declare({
+	object: 'luci.wifihistory',
+	method: 'getHistory',
+	expect: { history: {} }
+});
+
+const callClearHistory = rpc.declare({
+	object: 'luci.wifihistory',
+	method: 'clearHistory',
+	expect: { result: true }
+});
+
+function formatDate(epoch) {
+	if (!epoch || epoch <= 0)
+		return '-';
+
+	return new Date(epoch * 1000).toLocaleString();
+}
+
+return view.extend({
+	load() {
+		return callGetHistory();
+	},
+
+	updateTable(table, history) {
+		const stations = Object.values(history);
+
+		/* Sort connected stations first, then by most recently seen */
+		stations.sort((a, b) => {
+			if (a.connected !== b.connected)
+				return a.connected ? -1 : 1;
+
+			return (b.last_seen || 0) - (a.last_seen || 0);
+		});
+
+		const rows = [];
+
+		for (const s of stations) {
+			let hint;
+			if (s.hostname && s.ipv4 && s.ipv6)
+				hint = '%s (%s, %s)'.format(s.hostname, s.ipv4, s.ipv6);
+			else if (s.hostname && (s.ipv4 || s.ipv6))
+				hint = '%s (%s)'.format(s.hostname, s.ipv4 || s.ipv6);
+			else if (s.ipv4 || s.ipv6)
+				hint = s.ipv4 || s.ipv6;
+			else
+				hint = '-';
+
+			let sig_value = '-';
+			let sig_title = '';
+
+			if (s.signal && s.signal !== 0) {
+				if (s.noise && s.noise !== 0) {
+					sig_value = '%d/%d\xa0%s'.format(s.signal, s.noise, _('dBm'));
+					sig_title = '%s: %d %s / %s: %d %s / %s %d'.format(
+						_('Signal'), s.signal, _('dBm'),
+						_('Noise'), s.noise, _('dBm'),
+						_('SNR'), s.signal - s.noise);
+				}
+				else {
+					sig_value = '%d\xa0%s'.format(s.signal, _('dBm'));
+					sig_title = '%s: %d %s'.format(_('Signal'), s.signal, _('dBm'));
+				}
+			}
+
+			let icon;
+			if (!s.connected) {
+				icon = L.resource('icons/signal-none.svg');
+			}
+			else {
+				/* Estimate signal quality as percentage:
+				 * Map dBm range [-110, -40] to [0%, 100%] */
+				const q = Math.min((s.signal + 110) / 70 * 100, 100);
+				if (q == 0)
+					icon = L.resource('icons/signal-000-000.svg');
+				else if (q < 25)
+					icon = L.resource('icons/signal-000-025.svg');
+				else if (q < 50)
+					icon = L.resource('icons/signal-025-050.svg');
+				else if (q < 75)
+					icon = L.resource('icons/signal-050-075.svg');
+				else
+					icon = L.resource('icons/signal-075-100.svg');
+			}
+
+			rows.push([
+				E('span', {
+					'class': 'ifacebadge',
+					'style': s.connected ? '' : 'opacity:0.5',
+					'title': s.connected ? _('Connected') : _('Disconnected')
+				}, [
+					E('img', { 'src': icon, 'style': 'width:16px;height:16px' }),
+					E('span', {}, [ ' ', s.connected ? _('Yes') : _('No') ])
+				]),
+				s.mac,
+				hint,
+				s.network || '-',
+				E('span', { 'title': sig_title }, sig_value),
+				formatDate(s.first_seen),
+				formatDate(s.last_seen)
+			]);
+		}
+
+		cbi_update_table(table, rows, E('em', _('No station history available')));
+	},
+
+	handleClearHistory(ev) {
+		return ui.showModal(_('Clear Station History'), [
+			E('p', _('This will permanently delete all recorded station history. Are you sure?')),
+			E('div', { 'class': 'right' }, [
+				E('button', {
+					'class': 'btn',
+					'click': ui.hideModal
+				}, _('Cancel')), ' ',
+				E('button', {
+					'class': 'btn cbi-button-negative',
+					'click': ui.createHandlerFn(this, function() {
+						return callClearHistory().then(L.bind(function() {
+							ui.hideModal();
+							return callGetHistory().then(L.bind(function(history) {
+								this.updateTable('#wifi_history_table', history);
+							}, this));
+						}, this));
+					})
+				}, _('Clear'))
+			])
+		]);
+	},
+
+	render(history) {
+		const isReadonlyView = !L.hasViewPermission();
+
+		const v = E([], [
+			E('h2', _('Station History')),
+			E('div', { 'class': 'cbi-map-descr' },
+				_('This page displays a history of all WiFi stations that have connected to this device, including currently connected and previously seen devices.')),
+
+			E('div', { 'class': 'cbi-section' }, [
+				E('div', { 'class': 'right', 'style': 'margin-bottom:1em' }, [
+					E('button', {
+						'class': 'btn cbi-button-negative',
+						'disabled': isReadonlyView,
+						'click': ui.createHandlerFn(this, 'handleClearHistory')
+					}, _('Clear History'))
+				]),
+
+				E('table', { 'class': 'table', 'id': 'wifi_history_table' }, [
+					E('tr', { 'class': 'tr table-titles' }, [
+						E('th', { 'class': 'th' }, _('Connected')),
+						E('th', { 'class': 'th' }, _('MAC address')),
+						E('th', { 'class': 'th' }, _('Host')),
+						E('th', { 'class': 'th' }, _('Network')),
+						E('th', { 'class': 'th' }, '%s / %s'.format(_('Signal'), _('Noise'))),
+						E('th', { 'class': 'th' }, _('First seen')),
+						E('th', { 'class': 'th' }, _('Last seen'))
+					])
+				])
+			])
+		]);
+
+		this.updateTable(v.querySelector('#wifi_history_table'), history);
+
+		poll.add(L.bind(function() {
+			return callGetHistory().then(L.bind(function(history) {
+				this.updateTable('#wifi_history_table', history);
+			}, this));
+		}, this), 5);
+
+		return v;
+	},
+
+	handleSaveApply: null,
+	handleSave: null,
+	handleReset: null
+});
--- a/applications/luci-app-wifihistory/root/usr/sbin/wifihistory
+++ b/applications/luci-app-wifihistory/root/usr/sbin/wifihistory
@@ -1,0 +1,124 @@
+#!/usr/bin/env ucode
+
+'use strict';
+
+import { readfile, writefile, open, mkdir, rename } from 'fs';
+import { connect } from 'ubus';
+
+const HISTORY_DIR = '/var/lib/wifihistory';
+const HISTORY_FILE = HISTORY_DIR + '/history.json';
+const LOCK_FILE = '/var/lock/wifihistory.lock';
+
+function load_history() {
+	let content = readfile(HISTORY_FILE);
+	if (content == null)
+		return {};
+
+	try {
+		return json(content) || {};
+	}
+	catch (e) {
+		return {};
+	}
+}
+
+function save_history(data) {
+	let tmp = HISTORY_FILE + '.tmp';
+
+	writefile(tmp, sprintf('%J', data));
+	rename(tmp, HISTORY_FILE);
+}
+
+function poll_stations() {
+	let ubus = connect();
+	if (!ubus) {
+		warn('Failed to connect to ubus\n');
+		return;
+	}
+
+	let now = time();
+	let history = load_history();
+	let seen_macs = {};
+
+	let wifi_status = ubus.call('network.wireless', 'status');
+	if (!wifi_status)
+		return;
+
+	let hints = ubus.call('luci-rpc', 'getHostHints') || {};
+
+	for (let radio in wifi_status) {
+		let ifaces = wifi_status[radio]?.interfaces;
+		if (!ifaces)
+			continue;
+
+		for (let iface in ifaces) {
+			let ifname = iface?.ifname;
+			if (!ifname)
+				continue;
+
+			let info = ubus.call('iwinfo', 'info', { device: ifname });
+			let ssid = info?.ssid || '';
+
+			let assoc = ubus.call('iwinfo', 'assoclist', { device: ifname });
+			if (!assoc?.results)
+				continue;
+
+			for (let bss in assoc.results) {
+				let mac = bss?.mac;
+				if (!mac)
+					continue;
+
+				mac = uc(mac);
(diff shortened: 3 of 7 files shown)


applications/luci-app-wifihistory/root/etc/uci-defaults/40_luci-app-wifihistory
feckert, line 1: I think this file is not necessary. It will be called only once during first boot and then the file is deleted.
The command for creating the directory `/var/lib/wifihistory` is already done in the start sequence of `/etc/init.d/wifihistory`.
And the other stuff is already done via a generic posinst script in [luci.mk](https://github.com/openwrt/luci/blob/master/luci.mk#L231)

applications/luci-app-wifihistory/root/usr/libexec/rpcd/luci.wifihistory
```diff
@@ -0,0 +1,47 @@
...
+
+HISTORY_FILE="/var/lib/wifihistory/history.json"
+
+get_history() {
+	if [ -f "$HISTORY_FILE" ]; then
+		local data
+		data="$(cat "$HISTORY_FILE")"
+		json_init
+		json_add_object "history"
+		json_load_file "$HISTORY_FILE" 2>/dev/null
```
feckert: Are this lines needed? I do not see a json_dump?

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
```diff
@@ -0,0 +1,221 @@
+#!/bin/sh
+
+# Copyright (C) 2025 The LuCI Team <luci@lists.subsignal.org>
+# Licensed under the Apache License, Version 2.0
+
+. /usr/share/libubox/jshn.sh
+
+HISTORY_DIR="/var/lib/wifihistory"
+HISTORY_FILE="${HISTORY_DIR}/history.json"
+LOCK_FILE="/var/lock/wifihistory.lock"
```
feckert: Mybe this variables can be moved to a common file because `/usr/libexec/rpcd/luci.wifihistory` and `/usr/sbin/wifihistory` uses theme?

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
```diff
@@ -0,0 +1,221 @@
...
+get_wifi_interfaces() {
+	local ifaces=""
+	local ubus_wifi
+
+	ubus_wifi="$(ubus call network.wireless status 2>/dev/null)"
+	[ -z "$ubus_wifi" ] && return
+
+	json_load "$ubus_wifi"
+	json_get_keys radios
+	for radio in $radios; do
```
feckert: Variable declaration `local radio`, `local radios` is missing

applications/luci-app-wifihistory/root/usr/libexec/rpcd/luci.wifihistory
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+. /usr/share/libubox/jshn.sh
+
+HISTORY_FILE="/var/lib/wifihistory/history.json"
+
+get_history() {
+	if [ -f "$HISTORY_FILE" ]; then
+		local data
```
jow-: I suggest something like this:

```shell
printf '{"history":'
cat "$HISTORY_FILE"
printf '}'
```

Much simpler and takes virtually no memory, compared to the json_load_file approach, especially when the history file becomes very large.

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
```diff
@@ -0,0 +1,221 @@
...
+		for seen_mac in $seen_macs; do
+			if [ "$seen_mac" = "$station_mac" ]; then
+				is_connected=1
+				break
+			fi
+		done
+
+		[ "$first" = "1" ] && first=0 || new_history="${new_history},"
+
+		new_history="${new_history}\"${station_mac}\":{\"mac\":\"${s_mac}\",\"hostname\":\"${s_hostname}\",\"ipv4\":\"${s_ipv4}\",\"ipv6\":\"${s_ipv6}\",\"network\":\"${s_network}\",\"ifname\":\"${s_ifname}\",\"connected\":$([ "$is_connected" = "1" ] && echo "true" || echo "false"),\"signal\":${s_signal:-0},\"noise\":${s_noise:-0},\"first_seen\":${s_first_seen:-0},\"last_seen\":${s_last_seen:-0}}"
```
feckert: Please use json_* API via [json_set_namespace](https://github.com/search?q=repo%3Aopenwrt%2Fopenwrt%20json_set_namespace&type=code)

applications/luci-app-wifihistory/root/usr/libexec/rpcd/luci.wifihistory
systemcrash: If it does, the max ubus can handle is like 256Kb or something?

DeborahOlaboye: Okay noted. I will implement the change.

DeborahOlaboye: > I suggest something like this:
> 
> ```shell
> printf '{"history":'
> cat "$HISTORY_FILE"
> printf '}'
> ```
> 
> Much simpler and takes virtually no memory, compared to the json_load_file approach, especially when the history file becomes very large.

Done. Rewrote the `rpcd` plugin in `ucode`

DeborahOlaboye: > If it does, the max ubus can handle is like 256Kb or something?

Good point, noted.

applications/luci-app-wifihistory/root/etc/uci-defaults/40_luci-app-wifihistory
DeborahOlaboye: > I think this file is not necessary. It will be called only once during first boot and then the file is deleted. The command for creating the directory `/var/lib/wifihistory` is already done in the start sequence of `/etc/init.d/wifihistory`. And the other stuff is already done via a generic posinst script in [luci.mk](https://github.com/openwrt/luci/blob/master/luci.mk#L231)

Removed, thanks.

applications/luci-app-wifihistory/root/usr/libexec/rpcd/luci.wifihistory
DeborahOlaboye: Removed. The `rpcd` plugin is now a `ucode` service with no dead code.

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
DeborahOlaboye: Noted. Both files now use the same constant path directly.

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
DeborahOlaboye: Fixed. The service is now in `ucode` which uses `let` scoping throughout.

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
DeborahOlaboye: Replaced all manual JSON string building with ucode's native `sprintf('%J', data)` for serialization.

feckert:
Thanks for rewriting this in ucode :+1:
* Can you please squash the commits so there is only one commit.
* Please also fix CI/CD formal check https://github.com/openwrt/luci/pull/8293#issuecomment-3840371643

DeborahOlaboye:
> Thanks for rewriting this in ucode 👍
> 
> * Can you please squash the commits so there is only one commit.
> * Please also fix CI/CD formal check [luci-app-wifihistory: add WiFi station history tracking #8293 (comment)](https://github.com/openwrt/luci/pull/8293#issuecomment-3840371643)

Alright, done.

feckert:
CI/CD for `Test Formatlities` is still not happy

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
```diff
@@ -0,0 +1,186 @@
...
+});
+
+function formatDate(epoch) {
+	if (!epoch || epoch <= 0)
+		return '-';
+
+	var d = new Date(epoch * 1000);
+	return '%04d-%02d-%02d %02d:%02d:%02d'.format(
+		d.getFullYear(), d.getMonth() + 1, d.getDate(),
+		d.getHours(), d.getMinutes(), d.getSeconds());
```
systemcrash: Look at how this is done in luci-base.

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
```diff
@@ -0,0 +1,186 @@
...
+
+return view.extend({
+	load: function() {
+		return callGetHistory();
+	},
+
+	updateTable: function(table, history) {
+		var rows = [];
+		var stations = [];
+
```
systemcrash: Use ES6. 
```js
load() {...
updateTable(table, history) {...

```

Use const or let everywhere.

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
```diff
@@ -0,0 +1,186 @@
...
+		stations.sort(function(a, b) {
+			if (a.connected !== b.connected)
+				return a.connected ? -1 : 1;
+
+			return (b.last_seen || 0) - (a.last_seen || 0);
+		});
+
+		for (var i = 0; i < stations.length; i++) {
+			var s = stations[i];
```
systemcrash: let station of stations

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
DeborahOlaboye: Okay, noted and implemented.

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
DeborahOlaboye: Okay, noted and implemented.

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
DeborahOlaboye: Okay, noted and implemented. Thank you for the review.

DeborahOlaboye:
> CI/CD for `Test Formatlities` is still not happy

Hopefully, CI/CD for `Test Formatlities` should be passing now after the latest checks.

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
```diff
@@ -0,0 +1,175 @@
...
+					sig_title = '%s: %d %s'.format(_('Signal'), s.signal, _('dBm'));
+				}
+			}
+
+			let icon;
+			if (!s.connected) {
+				icon = L.resource('icons/signal-none.svg');
+			}
+			else {
+				const q = Math.min((s.signal + 110) / 70 * 100, 100);
```
feckert, line 79: Could you please add a comment explaining what this calculation is for and what you are doing there? So that others who are not familiar with this app but need to change something know what it is for?

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
```diff
@@ -0,0 +1,175 @@
...
+
+return view.extend({
+	load() {
+		return callGetHistory();
+	},
+
+	updateTable(table, history) {
+		const stations = Object.values(history);
+
+		stations.sort((a, b) => {
```
feckert, line 35: Can you also write a comment here explaining why this is being done? Or move it to a sub-function with a function name that is self-explanatory and describes what is being done here.

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
```diff
@@ -0,0 +1,122 @@
...
+	for (let mac in history)
+		if (!seen_macs[mac])
+			history[mac].connected = false;
+
+	ubus.disconnect();
+
+	save_history(history);
+}
+
+let lock_fd = open(LOCK_FILE, 'w');
```
feckert, line 109: I think this is not a correct lock in ucode See https://ucode.mein.io/module-fs.file.html#lock

applications/luci-app-wifihistory/root/usr/share/rpcd/acl.d/luci-app-wifihistory.json
```diff
@@ -0,0 +1,22 @@
...
+		"read": {
+			"ubus": {
+				"iwinfo": [ "assoclist" ],
+				"luci.wifihistory": [ "getHistory" ]
+			},
+			"file": {
+				"/var/lib/wifihistory/history.json": [ "read" ]
+			}
+		},
+		"write": {
```
feckert, line 10: If we don't have a write lock in the ACL, then we shouldn't be able to delete the history either. Search for `isReadonlyView` in the repository. I would therefore set the delete button to `disabled` if you don't have a write ACL for this app.

applications/luci-app-wifihistory/Makefile
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (C) 2025 The LuCI Team <luci@lists.subsignal.org>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=LuCI support for WiFi Station History
+LUCI_DEPENDS:=+luci-base +rpcd-mod-iwinfo
```
feckert: Please exptend the `LUCI_DEPENDS` with the needed ucode package?

applications/luci-app-wifihistory/Makefile
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (C) 2025 The LuCI Team <luci@lists.subsignal.org>
+#
```
feckert: Please remove this lines. This are not needed

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
```diff
@@ -0,0 +1,122 @@
...
+if (!lock_fd) {
+	warn('Failed to open lock file\n');
+	exit(1);
+}
+
+let pp = popen('mkdir -p ' + HISTORY_DIR, 'r');
```
feckert: Can this not be done with ucode? Do we need there a shell execution fork?

DeborahOlaboye:
> LGTM We're almost done. Please take another look at my review comments.
> 
> Thanks for your work 👍

Alright, thank you.

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
DeborahOlaboye: Alright, added.

applications/luci-app-wifihistory/htdocs/luci-static/resources/view/status/wifihistory.js
DeborahOlaboye: Alright, comment added.

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
DeborahOlaboye: Oh, figured that out. Fixed.

applications/luci-app-wifihistory/root/usr/share/rpcd/acl.d/luci-app-wifihistory.json
DeborahOlaboye: I have now added `L.hasViewPermission()` check, the Clear History button is now disabled when the user lacks write ACL. Thank you.

applications/luci-app-wifihistory/Makefile
DeborahOlaboye: Added `+ucode +ucode-mod-fs +ucode-mod-ubus` to `LUCI_DEPENDS`.

applications/luci-app-wifihistory/Makefile
DeborahOlaboye: Removed.

applications/luci-app-wifihistory/root/usr/sbin/wifihistory
DeborahOlaboye: Replaced with `mkdir()` from ucode's `fs` module.

DeborahOlaboye:
> LGTM We're almost done. Please take another look at my review comments.
> Thanks for your work 👍

All review comments addressed. Thanks for the thorough review.

applications/luci-app-wifihistory/root/usr/share/rpcd/acl.d/luci-app-wifihistory.json
```diff
@@ -0,0 +1,22 @@
...
+			"ubus": {
+				"luci.wifihistory": [ "clearHistory" ]
+			},
+			"file": {
```
feckert: I don't think that's necessary, as writing from this file is wrapped in the [rpcd](applications/luci-app-wifihistory/root/usr/share/rpcd/ucode/wifihistory.uc) call, isn't it?

applications/luci-app-wifihistory/root/usr/share/rpcd/acl.d/luci-app-wifihistory.json
```diff
@@ -0,0 +1,22 @@
+{
+	"luci-app-wifihistory": {
+		"description": "Grant access to WiFi station history",
+		"read": {
+			"ubus": {
+				"iwinfo": [ "assoclist" ],
+				"luci.wifihistory": [ "getHistory" ]
+			},
+			"file": {
+				"/var/lib/wifihistory/history.json": [ "read" ]
```
feckert: I don't think that's necessary, as reading from this file is wrapped in the [rpcd](applications/luci-app-wifihistory/root/usr/share/rpcd/ucode/wifihistory.uc) call, isn't it?

applications/luci-app-wifihistory/root/usr/share/rpcd/acl.d/luci-app-wifihistory.json
DeborahOlaboye: You're right, the file access is handled through the rpcd ucode plugin, so direct file ACL entries aren't needed. I'd remove it.

applications/luci-app-wifihistory/root/usr/share/rpcd/acl.d/luci-app-wifihistory.json
DeborahOlaboye: You're right, the file access is handled through the rpcd ucode plugin, so direct file ACL entries aren't needed. I'd remove it.

DeborahOlaboye:
@feckert , you're right, the file access is handled through the rpcd ucode plugin, so direct file ACL entries aren't needed. I have removed both.

systemcrash:
Merged. Thanks @DeborahOlaboye !

PR #8296, opened 2026-02-04
luci-app-tailscale-community: fix default state file path

kuanyili:
Maintainer: @Tokisaki-Galaxy
Tested on: x86/64, OpenWrt 25.12.0-rc4

Align default state file path with `tailscale` package.

(See https://github.com/openwrt/packages/blob/master/net/tailscale/files/tailscale.init)

When `tailscale.settings` doesn't exist in UCI, this LuCI App tries to create one with settings parsed from the state file. The parser first looks for `tailscale.settings.state_file` but this value sure wouldn't exist under the circumstances and will fall back to default path.

Fix the default path for the parser to locate the state file properly.

### DIFF
--- a/applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
+++ b/applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
@@ -95,7 +95,7 @@ methods.get_settings = {
 	call: function() {
 		let settings = {};
 		uci.load('tailscale');
-		let state_file_path = uci.get('tailscale', 'settings', 'state_file') || "/var/lib/tailscale/tailscaled.state";
+		let state_file_path = uci.get('tailscale', 'settings', 'state_file') || "/etc/tailscale/tailscaled.state";
 		if (access(state_file_path)) {
 			try {
 				let state_content = readfile(state_file_path);


Tokisaki-Galaxy:
My original intention was to follow the FHS (Filesystem Hierarchy Standard) used by generic Linux distributions as fallback. However, considering that /var in OpenWrt is a tmpfs (RAM-based), keeping the state file there would cause loss of identity after a reboot.
Thank you very much for pointing out

kuanyili:
Got you.

I think it is safe to assume that the user has standard `tailscale` package installed since this LuCI App listed it as a dependency. On the other hand, if standard `tailscale` is not installed and `/etc/config/tailscale` does not exist, `uci.load('tailscale')` will also fail and the web UI won't load. So we should almost never have the need to put parsing result in use, maybe just for the user who run `uci delete tailscale.settings`? (would be rare)

This PR is just some nitpicking, the package actually works really well! Thanks for the great work! 👍

systemcrash:
How often does that statefile update? Flash systems don't fare well if the state file updates often. In odhcpd, we made it an option to save the statefile to flash. Otherwise everything is stashed in /tmp. Things the user should be aware of.

kuanyili:
The state file of Tailscale stores keys and config, does not change across reboots from my observation.

PR #8303, opened 2026-02-05
Activate luci-app-attendedsysupgrade and luci based on buildbot config

hauke:
This will remove the dependency to `luci-app-attendedsysupgrade` from the luci collections to allow disabling it in custom builds. `luci-ssl` and `luci-app-attendedsysupgrade` will be activated in buildbot builds.

### DIFF
--- a/applications/luci-app-attendedsysupgrade/Makefile
+++ b/applications/luci-app-attendedsysupgrade/Makefile
@@ -9,6 +9,8 @@ LUCI_DEPENDS:=+luci-base +attendedsysupgrade-common +cgi-io
 PKG_MAINTAINER:=Eric Fahlgren <ericfahlgren@gmail.com>, Paul Spooren <paul@spooren.de>
 PKG_LICENSE:=GPL-2.0
 
+LUCI_DEFAULT:=y if BUILDBOT
+
 include ../../luci.mk
 
 # call BuildPackage - OpenWrt buildroot signature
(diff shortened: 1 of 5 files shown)


efahl:
@sinancetinkaya @peterwillcn
Since you had concerns/comments regarding inclusion of luci-app-attendedsysupgrade...

sinancetinkaya:
> @sinancetinkaya @peterwillcn Since you had concerns/comments regarding inclusion of luci-app-attendedsysupgrade...

Can you push it to 25.12 as well?

sanderpqr:
Why remove `luci-app-attendedsysupgrade` but keep `luci-app-package-manager`? For making a lightweight custom build, both ASU and APK are superfluous imho.

(Fwiw, my personal approach for custom builds is to simply not add `luci-ssl`, but add `luci-light` and `px5g-mbedtls` instead. This also avoids the dependencies on `luci-app-attendedsysupgrade` an `luci-app-package-manager`.)

hauke:
I also plan to cherry pick this to 25.12.

hauke:
Just fixed the complained about the git commit title being too long.

aparcar:
LGTM

hauke:
I removed activating luci-ssl by default to prevent adding LuCI to the main branch builds automatically (00bcca921a0fe97bd40112a40263659bffc685dc). The main branch builds activate `CONFIG_KERNEL_KALLSYMS` which increases the kernel size. We save some space again by leaving out LuCI, contine that for now.

PR #8102, opened 2025-11-19
luci-mod-network: add support for fast-dns-retry, use-stale-cache

saldry:
- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [ ] Incremented :up: any `PKG_VERSION` in the Makefile
- [x] Tested on: (x86/64, openwrt snapshot, Chrome) :white_check_mark:
- [ ] \( Preferred ) Mention: @ the original code author for feedback
- [x] \( Preferred ) Screenshot or mp4 of changes:
- [ ] \( Optional ) Closes: e.g. openwrt/luci#issue-number
- [x] \( Optional ) Depends on: https://github.com/openwrt/openwrt/pull/20823
- [x] Description: (describe the changes proposed in this PR)

Note: I don't know if Weblate likes the help text of `Fast retry time setting`/`fast_dns_retry_param`

pic1

pic2

PR #8314, opened 2026-02-09
luci-mod-network: dns.js: move some related options to cache tab

saldry:
- [x] Tested on: (x86, openwrt SNAPSHOT, Chrome 144) :white_check_mark:

---

Originally part of #8102

- nonegcache, cachesize, min_cache_ttl and max_cache_ttl are moved to cache tab
- improved the help text of nonegcache and min_cache_ttl
- added datatype limit to min_cache_ttl and max_cache_ttl

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js
@@ -266,6 +266,34 @@ return view.extend({
 		recordtypes.forEach(r => {
 			o.value(r);
 		});
+
+		o = s.taboption('cache', form.Flag, 'nonegcache',
+			_('No negative cache'),
+			_('Do not cache negative replies from upstream nameservers, e.g. for non-existent domains.') + '<br/>' +
+			_('Caching negative replies allows dnsmasq to answer identical queries without forwarding them again.'));
+		o.rmempty = false;
+
+		o = s.taboption('cache', form.Value, 'cachesize',
+			_('Size of DNS query cache'),
+			_('Number of cached DNS entries, 10000 is maximum, 0 is no caching.'));
+		o.optional = true;
+		o.datatype = 'range(0,10000)';
+		o.placeholder = 150;
+
+		o = s.taboption('cache', form.Value, 'min_cache_ttl',
+			_('Min cache TTL'),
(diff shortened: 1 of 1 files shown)


systemcrash:
Perhaps it's an opinion, but I chose to put nonegcache not in the cache tab, since dnsmasq 'filters out' those replies. But I suppose all 'cache' entries could live under the cache tab. The 'limits' cache entries do just as well under both tabs. But it's one or the other without a bit more magic.

saldry:
"Cache" is a more clear and specific subject/topic, while "limit" not. So I prefer cache. Actually, there are only dnsforwardmax and ednspacket_max left in "limit" after removing these cache options. Maybe they can also move to other tabs(dnsforwardmax to "general" or "forward", renaming "DNSSEC" to "EDNS" and adding ednspacket_max).

For nonegcache, the *filtering* is performed internally and does not affect/change a dns response. So I think "cache" is better.

Personal opinion, for reference only.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js
```diff
@@ -266,6 +266,34 @@ return view.extend({
 		recordtypes.forEach(r => {
 			o.value(r);
 		});
+
+		o = s.taboption('cache', form.Flag, 'nonegcache',
+			_('No negative cache'),
+			_('Do not cache negative replies from upstream nameservers, e.g. for non-existent domains.') + '<br/>' +
+			_('Caching negative replies allows dnsmasq to answer identical queries without forwarding them again.'));
+		o.rmempty = false;
```
openwrt-ai, line 274: This adds `rmempty = false`, which the original `nonegcache` option did not have. The previous definition used the default Flag behavior (`rmempty = true`), so unchecking removed the option from UCI; with `rmempty = false` an unchecked box now persists `option nonegcache '0'` instead. The commit message only mentions moving the option and improving help text, so was this behavior change intentional? If not, drop the line to preserve the prior behavior:

```suggestion
```

(i.e. remove `o.rmempty = false;`). If it is intentional, it would be worth noting in the commit body.

---

openwrt-ai (commented):
Reviewed 1 new commit; one question on an undocumented behavior change inline.

---

PR #8330, opened 2026-02-17
luci-theme-*: Remove LuCI from title, material changes

Self-Hosting-Group:
- Remove `| LuCI` suffix from title tag
- Make material theme zoomable and remove legacy meta tags
- Unify viewport

### DIFF
--- a/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
+++ b/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
@@ -18,7 +18,7 @@
 <html lang="{{ dispatcher.lang }}" {{ darkpref ? `data-darkmode="${darkpref}"` : '' }}>
 	<head>
 		<meta charset="utf-8">
-		<title>{{ striptags(`${boardinfo.hostname + ' | ' ?? '?'}${dispatched?.title ? `${_(dispatched.title)} | ` : ''}`) }} LuCI</title>
+		<title>{{ striptags(`${boardinfo.hostname ?? '?'}${dispatched?.title ? ` | ${_(dispatched.title)}` : ''}`) }}</title>
 		{% if (!darkpref): %}
 			<script>
 				var mediaQuery = window.matchMedia('(prefers-color-scheme: dark)'),
@@ -29,7 +29,7 @@
 				setDarkMode(mediaQuery);
 			</script>
 		{% endif %}
-		<meta name="viewport" content="initial-scale=1.0">
+		<meta name="viewport" content="width=device-width, initial-scale=1">
 		<meta name="darkreader-lock">
 		<link rel="stylesheet" href="{{ media }}/cascade.css">
 		<link rel="stylesheet" media="only screen and (max-device-width: 854px)" href="{{ media }}/mobile.css" />
--- a/themes/luci-theme-material/ucode/template/themes/material/header.ut
+++ b/themes/luci-theme-material/ucode/template/themes/material/header.ut
@@ -30,14 +30,7 @@
 <html lang="{{ dispatcher.lang }}">
 <head>
 <meta charset="utf-8">
-<meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0" name="viewport"/>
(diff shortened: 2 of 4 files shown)


Self-Hosting-Group:
Hello @systemcrash. Here is the PR.

systemcrash:
@Ramon00 ?

Ramon00:
two remarks:
1) Why remove all the "web-app" meta tags? I guess somebody had a use case there? BTW I do think it is inconsistent, i.e. either do it for all themes or for none. We can make it a configurable as well I suppose. Im fine either way, but i dont really see the use case to be honest. So question is why was it added in the past?
2) I personally just want to have the hostname first, the rest I do not really mind, either with page name or whole menu. I typically do not have multiple tabs for the same router open. So why was it there before, and why was it removed?

systemcrash:
> two remarks:
> 
> 1. Why remove all the "web-app" meta tags? I guess somebody had a use case there? BTW I do think it is inconsistent, i.e. either do it for all themes or for none. We can make it a configurable as well I suppose. Im fine either way, but i dont really see the use case to be honest. So question is why was it added in the past?
> 2. I personally just want to have the hostname first, the rest I do not really mind, either with page name or whole menu. I typically do not have multiple tabs for the same router open. So why was it there before, and why was it removed?

Fair points. They were added a long time ago when browsers were… older standards.

2. probably fanatical use of tags in the templates. Similar to recent comments: for users managing multiple devices and so on. On a normal dev day I have several open.

Self-Hosting-Group:
Thanks for the comments.

> 1. Why remove all the "web-app" meta tags? I guess somebody had a use case there? BTW I do think it is inconsistent, i.e. either do it for all themes or for none. We can make it a configurable as well I suppose. Im fine either way, but i dont really see the use case to be honest. So question is why was it added in the past?

Removed for consistency, as these meta tags were only present in the material/openwrt2020 themes, introduced in 2015 in b825a14. Most of them are legacy tags that would be replaced by a web application manifest (PWA), but this only works (installable) with HTTPS and trusted certificates, which is probably not common with OpenWrt's UI. However, even without installation, some manifest properties should be applied (has to be checked). See: https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable

So, if you prefer, we could add a minimal web app manifest instead with a PR later, or we could add `mobile-web-app-capable`/`apple-mobile-web-app-capable` for consistency to all themes as an alternative with this PR. Note: I see some differences between the legacy meta tags and the manifest, as the latter seems to be for the whole site and not just the URL.

> 2. I personally just want to have the hostname first, the rest I do not really mind, either with page name or whole menu. I typically do not have multiple tabs for the same router open. So why was it there before, and why was it removed?

> 2. Probably fanatical use of tags in the templates. Similar to recent comments: for users managing multiple devices and so on. On a normal dev day I have several open.

I see these user cases:
1. Users having single/multiple tabs open for the same router (home users/devs)
2. Users managing multiple routers (router managers)

By showing the menu in the title, it's already good for user case 1, but IMHO, preferred would be the menu first that I proposed first, inspired by how it is with the OPNsense UI. The software/site can easily be identified by the icon and hostname first, perfect for case 2. I removed the LuCI suffix, as it seems to me unnecessary, as more details are available in the footer.

What should we do now?

It would actually be good to have the top menu (already translated) in the title, as mentioned here: https://github.com/openwrt/luci/pull/7731#issuecomment-3913648304. What would be necessary to have a `dispatched.topmenu`?

systemcrash:
Fix commit formalities plz

Self-Hosting-Group:
I try to fix the failing formality check, `Commit subject line MUST start with <package name>`, I've changed it to `luci-theme:`, which was last time accepted by the CI check.

Self-Hosting-Group:
Even without `dispatched.topmenu`, I think the PR would be a useful addition to the menu name missing from the title at the time, including the newly found title expression fixes, for the 25.12 release ;-)

systemcrash:
Closed via e81c0637e275b5d65a81ded69b43118563803def. Thanks @Self-Hosting-Group

Self-Hosting-Group:
Thanks for merging, @systemcrash!

PR #8334, opened 2026-02-18
openfortivpn: add password2 parameter to proto

andoniElBicho:
Introduce password2 new parameter in openfortivpn proto to enable introducing OTP or other DFA necessary for the VPN connection established.Password2 parameter name and coding has been copied from openconnect package.


systemcrash:
Does this align with the proto handler?

andoniElBicho:
> Does this align with the proto handler?

I hope so, I have created another PR (https://github.com/openwrt/packages/pull/28598). Do I need to include the new parameter in any other place? Thanks

PR #8338, opened 2026-02-21
Improve expiration handling in luci.c

### LINKED ISSUE #8337: rpcd-mod-luci: Fix wronlgy displayed "expire" for active leases with HAVE_BROKEN_RTC in dnsmasq enabled

### screenshots or captures

_No response_

### Actual behaviour

On my glinet router, all DHCPv4 (wrongly) show up as expired in LUCI.

### Expected behaviour

Active DHCPv4 leases should show up as active, like in the leases file.

### Steps to reproduce

The issue as i found it, lies in the expire value in my lease file, wichh has this format:
<expiration_duration> <MAC_address> <IP_address> <hostname> <client_id>
isntead of the (presumably expected)
<expiration_timestamp> <MAC_address> <IP_address> <hostname> <client_id>

For example:
not 1708365600 aa:bb:cc:dd:ee:ff 192.168.1.100 mydevice *
but 3600 aa:bb:cc:dd:ee:ff 192.168.1.100 mydevice *

I found this to be caused by a feature of dnsmasq, the HAVE_BROKEN_RTC flag. (It can be checked using `dnsmasq --version | grep "no-RTC"`)
If set, it stores realtive times, to fix issues on bad hardware, and seems to be recommended in the git repo of dnsmasq.

I tried to patch `luci-static/resources/view/status/include/40_dhcp.js` to detect heuristically, if the flag is set and i need to parse it differently.
```js
		if (hasBrokenRTC) {
			return {
				isUnlimited: false,
				isExpired: lease.expires <= 0,
				expiresAt: now + lease.expires // lease_length is stored in dhcp.leases
			};
		}

		return {
			isUnlimited: false,
			isExpired: lease.expires <= now,
			expiresAt: lease.expires // lease_expires is stored in dhcp.leases
		};
...
```
If most lease expiry values are unreasonably small compared to current time, hey're likely relative durations (seconds), not absolute timestamps (since 1970). Absolute timestamps should be > 100,000,000 (~1973). Relative durations are typically 1-86400 seconds (1 second to 1 day), so < 100,000,000 (~3 years).

But it seems this data is actually being passed wrongly from the rpc backend! There is always 0 coming back: 
```json
{
	"10": {
		"jsonrpc": "2.0",
		"id": 788,
		"result": [
			0,
			{
				"dhcp_leases": [
					{
						"expires": 0,
						"hostname": "Devicename",
						"macaddr": "11:11:11:11:11:11",
						"duid": "1111",
						"ipaddr": "192.168.111.111"
					},
					{
...
```

### Additional Information

```shell
{
        "kernel": "5.4.164",
        "hostname": "xxx",
        "system": "ARMv8 Processor rev 4",
        "model": "GL Technologies, Inc. AX1800",
        "board_name": "glinet,ax1800",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "23.05-SNAPSHOT",
                "revision": "",
                "target": "ipq60xx/generic",
                "description": "OpenWrt 23.05-SNAPSHOT ",
                "tip-revision": "OpenWrt 23.05-SNAPSHOT  / TIP-devel-aa312cd8",
                "tip-version": "devel"
        }
}
```

_No response_

### Relevant log output

```shell

```

nettnikl:
Refactor expiration logic to handle relative durations (as in dnsmasq with HAVE_BROKEN_RTC).

Fixes https://github.com/openwrt/luci/issues/8337

- [x] \( Optional ) Closes: e.g. openwrt/luci#issue-number

### DIFF
--- a/libs/rpcd-mod-luci/src/luci.c
+++ b/libs/rpcd-mod-luci/src/luci.c
@@ -345,6 +345,7 @@ static struct {
 	struct {
 		FILE *fh;
 		bool odhcpd;
+		bool has_relative_leasetime;
 	} *files;
 } lease_state = { };
 
@@ -386,6 +387,9 @@ add_leasefile(const char *path, bool is_odhcpd)
 	lease_state.files = ptr;
 	lease_state.files[lease_state.num].fh = fh;
 	lease_state.files[lease_state.num].odhcpd = is_odhcpd;
+	if (!is_odhcpd) {
+		lease_state.files[lease_state.num].has_relative_leasetime = dnsmasq_has_no_rtc();
+	}
 	lease_state.num++;
 
 	return true;
@@ -472,6 +476,24 @@ lease_open(void)
 	uci_free_context(uci);
 }
 
+static int
+dnsmasq_has_no_rtc(void)
+{
+    FILE *fd = popen("dnsmasq --version 2>/dev/null", "r");
+    if (!fd)
+        return 0;
+
+    char buf[8192] = {0};
+    char line[1024];
+
+    while (fgets(line, sizeof(line), fd))
+        strncat(buf, line, sizeof(buf) - strlen(buf) - 1);
+
+    pclose(fd);
+
+    return strstr(buf, "no-RTC") != NULL;
+}
+
 static struct lease_entry *
 lease_next(void)
 {
@@ -525,17 +547,10 @@ lease_next(void)
 				n = strtol(p, NULL, 10);
 				if (n < 0) {
 					e.expire = -1;  // unlimited
+				} else if (n > lease_state.now)
+					e.expire = n - lease_state.now;  // convert absolute to relative timestamp
 				} else {
-					if (L.hasSystemFeature('dnsmasq', 'rtc')) {
-						// If RTC is disabled, n is a relative duration in seconds
-						e.expire = n;   // already relative, pass through as-is
-					} else {
-						// If hardware times are trusted, n is an absolute timestamp
-						if (n > lease_state.now)
-							e.expire = n - lease_state.now;
-						else if (n >= 0)
-							e.expire = 0;  // expired, cap at 0
-					}
+					e.expire = 0;  // expired
 				}
 
 				strtok(NULL, " \t\n"); /* id */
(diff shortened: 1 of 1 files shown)


nettnikl:
Before i forget, before merge, same change needs to be applied to cgi-bin/luci/admin/network/dhcp i suppose - atleast in case we need to change frontend code

libs/rpcd-mod-luci/src/luci.c
```diff
@@ -523,12 +545,13 @@ lease_next(void)
...
-					e.expire = 0;
-				else
-					e.expire = -1;
+				if (n < 0) {
+					e.expire = -1;  // unlimited
+				} else if (n > lease_state.now)
+					e.expire = n - lease_state.now;  // convert absolute to relative timestamp
+				} else {
+					e.expire = 0;  // expired
+				}
```
openwrt-ai, line 554: Unbalanced braces here will fail to compile. The `else if` branch has no opening brace, but the line below it (`} else {`) tries to close one, so the `}` closes the surrounding block prematurely and the trailing `else` is left dangling. Add braces to the `else if` branch:

```suggestion
				if (n < 0) {
					e.expire = -1;  // unlimited
				} else if (n > lease_state.now) {
					e.expire = n - lease_state.now;  // convert absolute to relative timestamp
				} else {
					e.expire = 0;  // expired
				}
```

---

libs/rpcd-mod-luci/src/luci.c
```diff
@@ -386,6 +387,9 @@ add_leasefile(const char *path, bool is_odhcpd)
 	lease_state.files = ptr;
 	lease_state.files[lease_state.num].fh = fh;
 	lease_state.files[lease_state.num].odhcpd = is_odhcpd;
+	if (!is_odhcpd) {
+		lease_state.files[lease_state.num].has_relative_leasetime = dnsmasq_has_no_rtc();
```
openwrt-ai, line 391: `dnsmasq_has_no_rtc()` is called here but only defined later in the file (around line 480), with no prior prototype. This build uses `-Werror -Wmissing-declarations` (see [src/CMakeLists.txt](https://github.com/openwrt/luci/blob/2e528e0fdc40280b23fe290e170246c35114be19/libs/rpcd-mod-luci/src/CMakeLists.txt#L5)), so the implicit declaration / missing-declaration will turn into a hard compile error. Move the function definition above `add_leasefile`, or add a forward declaration before this call.

---

openwrt-ai (commented):
## Commit checks
- c21cc32 "luci-mod-status: improve lease format detection" — the message says "Use hasSystemFeature to distinguish dnsmasqs lease state format", but this commit introduces `L.hasSystemFeature('dnsmasq', 'rtc')` — JavaScript syntax (single-quoted strings, a JS API) inside a C source file, which cannot compile. It is then replaced in the next commit (2e528e0) by the `popen("dnsmasq --version")` / `no-RTC` grep approach. Consider squashing the series so no intermediate commit is broken and the final detection method matches its commit message.

See the two inline comments for compile-breaking issues that survive into the final commit (brace mismatch in the odhcpd expire block; call to `dnsmasq_has_no_rtc()` before its definition under `-Werror`).

---

PR #8348, opened 2026-02-24
luci-base: implement the move method for CBIJSONConfig

emzet:
As the title says, the original implementation was missing. Here is the actual implementation. It works correctly with `GridSection` (including nested ones) and handles moving items internally, which is useful when you need to save the data as `JSON` file, not just visualize it via the UI. 

The logic is as follows:

- convert the `data` object into an array
- find the indices for `section_id1` and `section_id2` 
- move the `section_id1` item to the correct position
- recalculate the indices for the entire data object
- convert the array back into an object

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -186,7 +186,19 @@ const CBIJSONConfig = baseclass.extend({
 	},
 
 	move(config, section_id1, section_id2, after) {
-		return uci.move.apply(this, [config, section_id1, section_id2, after]);
+		const dataArray = Object.values(this.data).sort((a, b) => a['.index'] - b['.index']);
+		
+		const fromIndex = dataArray.findIndex(section => section['.name'] === section_id1);
+		const toIndex = dataArray.findIndex(section => section['.name'] === section_id2);
+		
+		const [itemToMove] = dataArray.splice(fromIndex, 1);
+		dataArray.splice(toIndex, 0, itemToMove);
+		
+		dataArray.forEach((item, index) => {
+			item['.index'] = index;
+		});
+
+		this.data = Object.fromEntries(dataArray.map(item => [item['.name'], item]));
 	}
 });
 


modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -186,7 +186,19 @@ const CBIJSONConfig = baseclass.extend({
...
 
 	move(config, section_id1, section_id2, after) {
-		return uci.move.apply(this, [config, section_id1, section_id2, after]);
+		const dataArray = Object.values(this.data).sort((a, b) => a['.index'] - b['.index']);
+		
+		const fromIndex = dataArray.findIndex(section => section['.name'] === section_id1);
+		const toIndex = dataArray.findIndex(section => section['.name'] === section_id2);
+		
+		const [itemToMove] = dataArray.splice(fromIndex, 1);
+		dataArray.splice(toIndex, 0, itemToMove);
```
openwrt-ai, line 195: The `after` argument is accepted but never used, so this diverges from the `uci.move` contract this method is meant to replicate. The canonical implementation inserts at `index + !!after` to place `section_id1` either before (`after` falsy) or after (`after` truthy) `section_id2` ([uci.js#L875](https://github.com/openwrt/luci/blob/831250868521747435caadb5bef9988bfbfed638/modules/luci-base/htdocs/luci-static/resources/uci.js#L875)). GridSection drag-and-drop relies on this: `handleDrop` passes `after = true` when dropping below a row ([form.js#L3395](https://github.com/openwrt/luci/blob/831250868521747435caadb5bef9988bfbfed638/modules/luci-base/htdocs/luci-static/resources/form.js#L3395), and an explicit `after=true` call at [form.js#L3699](https://github.com/openwrt/luci/blob/831250868521747435caadb5bef9988bfbfed638/modules/luci-base/htdocs/luci-static/resources/form.js#L3699)). As written, a "drop below" always lands above the target.

Also, the documented contract says that when `section_id2` is null the section moves to the *end* of the config ([uci.js#L832-L833](https://github.com/openwrt/luci/blob/831250868521747435caadb5bef9988bfbfed638/modules/luci-base/htdocs/luci-static/resources/uci.js#L832)). Here a null/absent `section_id2` yields `toIndex === -1`, and `dataArray.splice(-1, 0, item)` inserts *before the last* element rather than at the end.

Suggest honoring both: when `section_id2` is null push to the end, otherwise insert at `toIndex + (after ? 1 : 0)`.

---

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -186,7 +186,19 @@ const CBIJSONConfig = baseclass.extend({
 	},
 
 	move(config, section_id1, section_id2, after) {
-		return uci.move.apply(this, [config, section_id1, section_id2, after]);
+		const dataArray = Object.values(this.data).sort((a, b) => a['.index'] - b['.index']);
+		
+		const fromIndex = dataArray.findIndex(section => section['.name'] === section_id1);
+		const toIndex = dataArray.findIndex(section => section['.name'] === section_id2);
+		
+		const [itemToMove] = dataArray.splice(fromIndex, 1);
```
openwrt-ai, line 194: `findIndex` returns `-1` when `section_id1` is absent, and `dataArray.splice(-1, 1)` then removes the *last* section and reinserts it elsewhere — silently corrupting the config instead of failing. `uci.move` guards this and returns `false` when either section is missing ([uci.js#L863](https://github.com/openwrt/luci/blob/831250868521747435caadb5bef9988bfbfed638/modules/luci-base/htdocs/luci-static/resources/uci.js#L863)). Recommend bailing out when `fromIndex === -1` (and, for a non-null `section_id2`, when `toIndex === -1`) before splicing. Relatedly, the public API documents a boolean return ([uci.js#L842-L844](https://github.com/openwrt/luci/blob/831250868521747435caadb5bef9988bfbfed638/modules/luci-base/htdocs/luci-static/resources/uci.js#L842)); this override returns `undefined`, so callers can no longer detect failure.

---

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -186,7 +186,19 @@ const CBIJSONConfig = baseclass.extend({
 	},
 
 	move(config, section_id1, section_id2, after) {
-		return uci.move.apply(this, [config, section_id1, section_id2, after]);
+		const dataArray = Object.values(this.data).sort((a, b) => a['.index'] - b['.index']);
+
```
openwrt-ai, line 190: nit: this and the other intra-method blank lines (after `toIndex` and after the `splice`) contain trailing tab whitespace; surrounding code keeps blank lines empty.

---

openwrt-ai (commented):
Reviewed the `move` override for `CBIJSONConfig`; see inline comments. Key concerns: the `after` argument is ignored (breaking GridSection "drop below"), missing sections produce `findIndex` of `-1` and silently corrupt the config via `splice(-1, ...)`, a null `section_id2` no longer moves to the end, and the override drops the documented boolean return value.

---

PR #8367, opened 2026-03-02
luci-app-usteer: add sta cipher and AKM info in tooltip

Ramon00:
Add sta cipher and AKM in tooltip to luci-app-usteer if hostapd_cli is installed

example of the full tooltip
image

It hostapd_cli is not installed it will show `Install hostapd_cli for AKM and cipher info` instead of the cipher and AKM.

For the sta's not on the current AP it does not show anything (as hostapd_cli can only show information on the current AP)

P.S. If so desired, I can add more information, e.g. the flags=[AUTH][ASSOC][AUTHORIZED][WMM][MFP][HT][VHT][HE]
or maybe he_capab, vht_caps_info, vht_capab, ht_caps_info, ext_capab? See `/usr/sbin/hostapd_cli -i phy1-ap0 all_sta` for a full list

P.P.S Alternative or maybe in addition, this can also be done in luci-mod-network/wireless.js?

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -2,6 +2,7 @@
 'require view';
 'require rpc';
 'require poll';
+'require fs';
 'require dom';
 'require ui';
 'require form';
@@ -12,6 +13,7 @@
 let Hosts, Remotehosts, Remoteinfo, Localinfo, Clients, WifiNetworks;
 
 const dns_cache = [];
+const hostapdClientData = [];
 
 function SplitWlan(wlan) {
 	let wlansplit = [];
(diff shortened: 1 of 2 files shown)


systemcrash:
Isn't this info all available via network functions already (a la network/wifi overview)?

Ramon00:
nope currently no way to check the actual cipher and AKM that the sta is using in LuCi. As far as I can tell it is also not exposed via ubus.

Ramon00:
ugh... where did that merge commit come from, let me kill that

applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
```diff
@@ -337,7 +423,13 @@ return view.extend({
...
 		Localinfo = data[4];
 		Clients = data[5];
+<<<<<<< HEAD
 		
+=======
+
+		getCipherAKM();	 
+
+>>>>>>> ea7fa9b017 (luci-app-usteer: add sta info in tooltip)
 		const remotehosttableentries = [];
```
systemcrash: ...

Ramon00: indeed ...

systemcrash:
I think you need to rebase on master to fix the conflict.

Ramon00:
I did that already, seems like something is out of sync on my side, but let me try again tonight

systemcrash:
Are you sure those maps will be `00-0f-ac-1` and not something like `00-0f-ac-01`?

Ramon00:
Yes im sure, try running `/usr/sbin/hostapd_cli -i phy1-ap0 all_sta` it will show a key like `dot11RSNAStatsSelectedPairwiseCipher=00-0f-ac-4` and `AKMSuiteSelector=00-0f-ac-4`. It is really strangely encoded. The last digit is actually decimal.

systemcrash:
OK - if you're happy, in she goes (just gonna let the checks finish)

Ramon00:
ok thx. I could make the same tooltip for luci-mod-network wireless.js, if you think that makes sense

systemcrash:
Closed via 60c348801dd4f9dc147ddf9109d238c97e8a2d1f

PR #8368, opened 2026-03-02
luci-app-https-dns-proxy: providers.json support; new provider

stangri:
Maintainer: me
Compile tested: x86_64, Dell EMC Edge620, OpenWrt 25.12.0-rc4
Run tested: x86_64, Dell EMC Edge620, OpenWrt 25.12.0-rc4

Description:
This commit adds support for:
* new unified provider.json support to save space compared to individual provider json files
* new v.recipes provider (thanks @morytyann!)
* functional testing
* minor js/shell code improvements


systemcrash:
@stangri I wonder whether it makes sense to merge the different force_http* settings into a force_http_ver - I don't think it's possible to use 1.1 and 3 together...

PR #8381, opened 2026-03-03
[DRAFT] luci-layer2: create a minimal collection for layer-2 devices

efahl:
The current luci-light collection includes many router (layer-3) specific package dependencies that are inappropriate for inclusion on switches and APs.  Let's create an even lighter top-level package to accommodate those devices as they rapidly are becoming more popular.

---
Testers strongly encouraged to grab this and test it out.  Some discussion at https://forum.openwrt.org/t/support-for-rtl838x-based-managed-switches/57875/4178

### DIFF
--- a/collections/luci-layer2/Makefile
+++ b/collections/luci-layer2/Makefile
@@ -1,0 +1,25 @@
+#
+# Copyright (C) 2008-2014 The LuCI Team <luci@lists.subsignal.org>
+#
+# This is free software, licensed under the Apache License, Version 2.0 .
+#
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TYPE:=col
+LUCI_BASENAME:=layer2
+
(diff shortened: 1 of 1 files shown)


systemcrash:
Good idea. Smaller footprint for managed switches. Even better now that Lua is gone :)

efahl:
Anyone willing to discuss the color of the bike shed?  What do you think of the name `luci-layer2`?  The `layer2` seems to me to be oddly specific and simultaneously non-descriptive, but I couldn't come up with something better that was reasonably short...

systemcrash:
I thought about that yesterday. I also thought that 'switch' was too general. Perhaps luci-managed-switch?

efahl:
I actually started doing this for APs, and then later thought it would be good for switches, too.  I originally named it `luci-ap`, only subsequently migrating to the `layer2` name to encompass both use cases.

systemcrash:
I guess layer 2 is fine then.

systemcrash:
Let's see how that works out in 25.12.

PR #8382, opened 2026-03-04
luci-base: fix label association warning for Dropdown widget

stangri:
## Summary

- Add `data-widget-id` attribute to the Dropdown widget's container `<div>`, matching the pattern already used by Checkbox (`ui.js` line 647)
- This allows `renderFrame()`'s existing fallback selector (`form.js` line 4454) to find the widget, resolving browser console accessibility warnings

## Problem

The Dropdown widget renders as a `<div>`, which is not a [labelable HTML element](https://html.spec.whatwg.org/multipage/forms.html#category-label). When used with `MultiValue` or other dropdown-based form options, `renderFrame()` creates a `<label for="widget.cbid...">` but no element with that `id` or `data-widget-id` exists in the Dropdown output.

This causes browser DevTools to report:
- "A `<label>` element with a `for` attribute doesn't match any element id"

## Fix

Single-line addition — add `data-widget-id` to `UIDropdown.render()`:

```javascript
'data-widget-id': this.options.id ? `widget.${this.options.id}` : null,
```

This leverages the existing fallback mechanism in `renderFrame()`:

```javascript
const elem = node.nextElementSibling.querySelector(`#${node.getAttribute('for')}`)
    ?? node.nextElementSibling.querySelector(`[data-widget-id="${node.getAttribute('for')}"]`);
```

## Test plan

- Open any LuCI page using `MultiValue` widget (e.g., any app with a multi-select dropdown)
- Check browser DevTools console — label association warning should no longer appear
- Verify clicking the label text still focuses/opens the dropdown

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -1090,8 +1090,17 @@ const UIDropdown = UIElement.extend(/** @lends LuCI.ui.Dropdown.prototype */ {
 			'multiple': this.options.multiple ? '' : null,
 			'optional': this.options.optional ? '' : null,
 			'disabled': this.options.disabled ? '' : null,
+			'data-widget-id': this.options.id ? `widget.${this.options.id}` : null,
 			'tabindex': -1
 		}, E('ul'));
+
+		if (this.options.id)
+			sb.appendChild(E('select', {
+				'id': `widget.${this.options.id}`,
+				'style': 'display:none',
+				'aria-hidden': 'true',
+				'tabindex': -1
+			}));
 
 		let keys = Object.keys(this.choices);
 
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -4451,7 +4451,7 @@ const CBIValue = CBIAbstractValue.extend(/** @lends LuCI.form.Value.prototype */
 					'for': 'widget.cbid.%s.%s.%s'.format(config_name, section_id, this.option),
 					'click': (ev) => {
 						const node = ev.currentTarget;
-						const elem = node.nextElementSibling.querySelector(`#${node.getAttribute('for')}`) ?? node.nextElementSibling.querySelector(`[data-widget-id="${node.getAttribute('for')}"]`);
+						const elem = node.nextElementSibling.querySelector(`[data-widget-id="${node.getAttribute('for')}"]`) ?? node.nextElementSibling.querySelector(`#${node.getAttribute('for')}`);
 
 						if (elem) {
 							elem.click();


systemcrash:
A bit of a can of worms here. I can only reproduce this on Chrome which generates those notices, and I cannot seem to silence the notice with this change alone. The checkboxes which use 'label' for the text label (also) generate the notice.

stangri:
@systemcrash sorry, you're right, the previous commit alone, wasn't sufficient. Just double-checked that this fixes the console errors on Chrome when using MultiValue widgets in luci apps.

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -1090,9 +1090,18 @@ const UIDropdown = UIElement.extend(/** @lends LuCI.ui.Dropdown.prototype */ {
...
 			'tabindex': -1
 		}, E('ul'));
 
+		if (this.options.id)
+			sb.appendChild(E('select', {
+				'id': `widget.${this.options.id}`,
+				'style': 'display:none',
+				'aria-hidden': 'true',
+				'tabindex': -1
+			}));
```
openwrt-ai, line 1103: Appending this hidden `<select>` makes it the last child of `sb`, which breaks the choice-list population below. `render()` builds the list with `sb.lastElementChild.appendChild(E('li', …))` (lines [1123](https://github.com/openwrt/luci/blob/a070c2c88ad841ceaf2cfcea8c38703bae078bd3/modules/luci-base/htdocs/luci-static/resources/ui.js#L1123) and [1144](https://github.com/openwrt/luci/blob/a070c2c88ad841ceaf2cfcea8c38703bae078bd3/modules/luci-base/htdocs/luci-static/resources/ui.js#L1144)), which previously targeted the `E('ul')` created as the div's only child. With the `<select>` appended after it, `sb.lastElementChild` is now the `<select>`, so every `<li>` option ends up inside the hidden, `display:none` `<select>` instead of the `<ul>`. `bind()` then does `sb.querySelector('ul')` and `ul.querySelectorAll('li')` and finds an empty list — so any Dropdown/MultiValue with an `id` (the normal case, since form widgets always pass `id: this.cbid(...)`) renders with no selectable items.

Since the `data-widget-id` attribute you added on the `<div>` plus the reordered fallback in `form.js` already let the label handler resolve the widget, the simplest fix is to drop this hidden `<select>` entirely. If it is genuinely needed for the labelable-element association, then the list population must target the `<ul>` explicitly (e.g. `sb.firstElementChild` / `sb.querySelector('ul')`) instead of `sb.lastElementChild` at both append sites.

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline issue found.

---

PR #8392, opened 2026-03-06
libs/luci-lib-ipkg: fix insecure temporary file usage in ipkg.lua

quetta-1030:
Fix potential security vulnerability in luci-lib-ipkg where fixed temporary file names could lead to:
- Race conditions when multiple processes run simultaneously
- Symlink attacks in /tmp directory
- Information disclosure through predictable file names

Changes:
- Use os.tmpname() to generate secure random temporary file names
- Update _action() function to use secure temp files
- Add security comments explaining the fix

This prevents attackers from exploiting predictable temporary file names to potentially escalate privileges or disclose sensitive information through symlink attacks in the /tmp directory.

### DIFF
--- a/libs/luci-lib-ipkg/luasrc/model/ipkg.lua
+++ b/libs/luci-lib-ipkg/luasrc/model/ipkg.lua
@@ -27,13 +27,17 @@ local function _action(cmd, ...)
 		cmdline[#cmdline+1] = util.shellquote(v)
 	end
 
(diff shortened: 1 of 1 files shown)


systemcrash:
A nice principled fix. What you might not be aware of is that nothing in this repo uses Lua any longer, which has been deprecated for several years and is scheduled for removal altogether. Package mangement is completely handled via JS, now, so ipkg.lua is... unused.

PR #8398, opened 2026-03-08
luci-mod-network: wireless.js: fix field names for private key file and password in EAP configuration

fungus-dev:
- [x] Tested on: (x86/64, OpenWrt 25.12.0, Chrome) :white_check_mark:

wireless.js
 fixed field names for private key file and password in EAP configuration which were causing issues with wpa_supplicant configuration file. 

Signed-off-by: Vishnu Verma <vermavishnuverma@gmail.com>

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -2014,10 +2014,10 @@ return view.extend({
 					o = ss.taboption('encryption', form.FileUpload, 'client_cert', _('Path to Client-Certificate'));
 					add_dependency_permutations(o, { mode: ['sta', 'sta-wds'], encryption: ['wpa', 'wpa2', 'wpa3', 'wpa3-mixed', 'wpa3-192'], eap_type: ['tls'] });
 
-					o = ss.taboption('encryption', form.FileUpload, 'priv_key', _('Path to Private Key'));
+					o = ss.taboption('encryption', form.FileUpload, 'private_key', _('Path to Private Key'));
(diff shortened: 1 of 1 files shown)


fungus-dev:
[supplicant.uc](
https://github.com/openwrt/openwrt/blob/557b092273e7ecb4f59305e2be63789522182145/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L181-L192) expects keys `'private_key,private_key_passwd'` but luci sets them as `'priv_key,'priv_key_pwd'` , causing them to not be added in the wpa_supplicant configuration file and generate the following error
```
daemon.notice: wpa_supplicant[2217]: phy1-sta0: CTRL-EVENT-EAP-PROPOSED-METHOD vendor=0 method=13
daemon.notice: wpa_supplicant[2217]: EAP-TLS: Private key not configured
```

systemcrash:
While this is an acceptable fix - it should be fixed in the openwrt repo - which would prevent breakage for users of this configuration when they upgrade to 25.12. Or is this still only master?

fungus-dev:
This issue popped up in OpenWrt 25. Versions 24 and earlier used [hostapd.sh](https://github.com/openwrt/openwrt/blob/9c04f527c62b38faf25e39e958679d2c717c618d/package/network/config/wifi-scripts/files/lib/netifd/hostapd.sh#L1472) to append values with correct field names.  OpenWrt 25.12.0 now uses [ucode](https://github.com/openwrt/openwrt/blob/37a1070b37b5da97a622e2d8a2d69fef73b9d641/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L184) script for handling config generation which does not handle this naming convention and instead tries to add fields that do not exist.

systemcrash:
Merged. Thanks @fungus-dev

PR #8399, opened 2026-03-08
luci-mod-status: fix syslog page with syslog-ng

Ramon00:
If you have syslog-ng installed instead of logd then syslog page displays errors.
This commit fixes that.

P.S. Every 5 seconds an error is added to the page via ui.addNotification, not sure if we want to fix that in addition? It does not look very elegant... We could just replace the error box with an error message in the log file box?

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/tools/views.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/views.js
@@ -2,6 +2,7 @@
 'require poll';
 'require rpc';
 'require uci';
+'require fs';
 'require ui';
 'require view';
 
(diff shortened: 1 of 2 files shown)


systemcrash:
syslog-ng provides a wrapper (it has but not the one called here - yet).

Also, how do you install syslog-ng? Does it remove logd?

Ramon00:
This is the wrapper  `/usr/libexec/syslog-wrapper`  that I know works (and also that is the one that was used in the past).

Not sure if it removes logd automatically, but I always remove logd then install syslog-ng.

Ramon00:
two additional things 1) people (including myself) may think that the logfile actually gets created on the suggested location (while nothing is there!):
image
2) why limit to 1000 lines?

systemcrash:
It's a 'suggestion' - not a default. I suppose inserting a keyword might fix that?

1000 lines is roughly 32kb. It's a good performance balance. 

> This is the wrapper `/usr/libexec/syslog-wrapper` that I know works (and also that is the one that was used in the past).

Check the syslog-ng package for its wrapper script. See whether both align.

Ramon00:
> Check the syslog-ng package for its wrapper script. See whether both align.

syslog-ng is just writing to `/var/log/messages`  The script [syslog-wrapper](https://github.com/openwrt/luci/blob/master/modules/luci-mod-status/root/usr/libexec/syslog-wrapper) reads it via `logread` script. This is a specific syslog-ng version of logread. Its actually explained [here ](https://openwrt.org/docs/guide-user/perf_and_log/log.syslog-ng3)

systemcrash:
https://github.com/openwrt/packages/blob/master/admin/syslog-ng/files/logread.sh

Ramon00:
i dont get your point, you want me to call logread directly instead of syslog-wrapper?

systemcrash:
So it works fine to invoke syslog-ng via the wrapper script?

Ramon00:
Yes I tested it like the PR is now and that works fine

systemcrash:
Merged. Thanks @Ramon00

PR #8405, opened 2026-03-08
luci-app-firewall: fix custom rules page for fw4

### LINKED ISSUE #7428: luci: custom firewall rules don't show up with fw4

- [X] I have searched the existing issues

### screenshots or captures

_No response_

### Actual behaviour

In the network>firewall tab of the GUI, the custom rules tab does not show up
I was able to fix this by removing the dependency of /usr/share/fw3/helpers.conf from the
/usr/share/luci/menu.d/luci-app-firewall.json file

### Expected behaviour

The custom tab in the firewall will show up and allow the user to put nftables commands for rules

### Steps to reproduce

go to: Network -> Firewall and look at the tabs on a device using firewall4

### Additional Information

```shell
NAME="OpenWrt"
VERSION="23.05.2"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt 23.05.2"
VERSION_ID="23.05.2"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r23630-842932a63d"
OPENWRT_BOARD="ipq40xx/generic"
OPENWRT_ARCH="arm_cortex-a7_neon-vfpv4"
OPENWRT_TAINTS="no-all busybox"
OPENWRT_DEVICE_MANUFACTURER="OpenWrt"
OPENWRT_DEVICE_MANUFACTURER_URL="https://openwrt.org/"
OPENWRT_DEVICE_PRODUCT="Generic"
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE="OpenWrt 23.05.2 r23630-842932a63d"
```

_No response_

### Relevant log output

_No response_

etb-source:
This fixes the Custom Rules page on firewall4 systems.

The page is currently hidden by a dependency on
`/usr/share/fw3/helpers.conf`, which is fw3-specific and not needed for firewall4.

This change removes that dependency so the page is shown again on fw4,
and updates the page text to refer to nftables syntax instead of iptables.

- [X] Tested on: ipq806x/generic, OpenWrt 24.10.5, Firefox
- [X] Description: This change removes the dependency on `/usr/share/fw3/helpers.conf`, which is fw3-specific and not needed for firewall4, so the page is shown again on fw4, and updates the page text to refer to nftables syntax instead of iptables.

Closes #7428

### DIFF
--- a/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/custom.js
+++ b/applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/custom.js
@@ -23,7 +23,7 @@ return view.extend({
 	render(fwuser) {
 		return E([
 			E('h2', _('Firewall - Custom Rules')),
-			E('p', {}, _('Custom rules allow you to execute arbitrary iptables commands which are not otherwise covered by the firewall framework. The commands are executed after each firewall restart, right after the default ruleset has been loaded.')),
+			E('p', {}, _('Custom rules allow you to define custom firewall rules which are not otherwise covered by the firewall framework. On firewall4 systems, these rules must use nftables syntax.'))
 			E('p', {}, E('textarea', { 'style': 'width:100%', 'rows': 25 }, [ fwuser != null ? fwuser : '' ]))
 		]);
 	},
--- a/applications/luci-app-firewall/root/usr/share/luci/menu.d/luci-app-firewall.json
+++ b/applications/luci-app-firewall/root/usr/share/luci/menu.d/luci-app-firewall.json
@@ -65,8 +65,5 @@
 			"type": "view",
 			"path": "firewall/custom"
 		},
-		"depends": {
-			"fs": { "/usr/share/fw3/helpers.conf": "file" }
-		}
 	}
 }


systemcrash:
This change co-opts the fw3 rules page and changes it to a fw4 rules page. What do fw3 users do?

etb-source:
This patch is not intended to remove fw3 support.

It only removes the fw3-specific visibility dependency for the Custom Rules page:
`/usr/share/fw3/helpers.conf`.

The concrete problem in #7428 is that this extra dependency hides the page on fw4 systems, even though fw4 is the default firewall backend on current OpenWrt releases.

So the intent here is narrower:
- keep the existing page
- make it visible again on fw4
- update the page text so it no longer refers to iptables when used on fw4

For fw3 users, this patch does not remove the page or replace fw3-specific functionality; it only stops gating the page on a fw3-specific helper file.

applications/luci-app-firewall/htdocs/luci-static/resources/view/firewall/custom.js
```diff
@@ -23,7 +23,7 @@ return view.extend({
 	render(fwuser) {
 		return E([
 			E('h2', _('Firewall - Custom Rules')),
-			E('p', {}, _('Custom rules allow you to execute arbitrary iptables commands which are not otherwise covered by the firewall framework. The commands are executed after each firewall restart, right after the default ruleset has been loaded.')),
+			E('p', {}, _('Custom rules allow you to define custom firewall rules which are not otherwise covered by the firewall framework. On firewall4 systems, these rules must use nftables syntax.'))
```
openwrt-ai, line 26: The trailing comma was dropped, so this `E('p', ...)` and the following `E('p', ...)` element are now juxtaposed with no separator inside the array passed to `E([...])`. This is a JavaScript syntax error and the entire Custom Rules view will fail to parse/render.

```suggestion
			E('p', {}, _('Custom rules allow you to define custom firewall rules which are not otherwise covered by the firewall framework. On firewall4 systems, these rules must use nftables syntax.')),
```

---

applications/luci-app-firewall/root/usr/share/luci/menu.d/luci-app-firewall.json
```diff
@@ -65,8 +65,5 @@
 			"type": "view",
 			"path": "firewall/custom"
 		},
```
openwrt-ai, line 67: Removing the `"depends"` block left a trailing comma after the `"action"` object (`},` immediately followed by `}`), which is invalid JSON. LuCI's dispatcher parses these files with ucode's strict `read_jsonfile` (`/usr/share/luci/menu.d/*.json`, see `dispatcher.uc`), so the parse will fail and the *entire* `luci-app-firewall.json` menu file is dropped — taking down all firewall menu entries (Zones, Port Forwards, Traffic Rules, NAT, IP Sets, Custom Rules), not just this page. Drop the trailing comma:

```suggestion
		}
```

---

openwrt-ai (commented):
Reviewed 1 new commit; 2 inline issues found (both will break functionality on merge).

---

PR #8413, opened 2026-03-12
luci-app-adguardhome: add new app

GeorgeSapkin:
Add LuCI UI for AdGuard Home configuration.

If AdGuard Home service is running, restart it automatically when configuration is applied.

Depends on:
- https://github.com/openwrt/packages/pull/28757
- https://github.com/openwrt/packages/pull/28781

## Run Testing Details

- **OpenWrt Version:** 25.12-rc3
- **OpenWrt Target/Subtarget:** x86/64
- **OpenWrt Device:** QEMU

image

image

image

### DIFF
--- a/applications/luci-app-adguardhome/htdocs/luci-static/resources/view/adguardhome/config.js
+++ b/applications/luci-app-adguardhome/htdocs/luci-static/resources/view/adguardhome/config.js
@@ -1,0 +1,261 @@
+'use strict';
+
+'require dom';
+'require form';
+'require fs';
+'require poll';
+'require rpc';
+'require view';
+
+const DEFAULT_CONFIG_FILE = '/etc/adguardhome/adguardhome.yaml';
+const DEFAULT_WORK_DIR = '/var/lib/adguardhome';
+const DEFAULT_USER = 'adguardhome';
+const DEFAULT_GROUP = DEFAULT_USER;
+
+const DEFAULT_GOGC = '0';
+const DEFAULT_GOMAXPROCS = '0';
+const DEFAULT_GOMEMLIMIT = '0';
+
+const PATH_REGEX = new RegExp('^/etc(/[^/]+)?/?$');
+
+const POLL_INTERVAL = 5;
+
+const RUNNING_SPAN = `<span style="color: var(--success-color-high); font-weight: bold">${_('Running')}</span>`;
+const NOT_RUNNING_SPAN = `<span style="color: var(--error-color-high); font-weight: bold">${_('Not running')}</span>`;
+
+const STORAGE_KEY = 'luci-app-adguardhome';
+
+function getServiceInfo(name) {
+	const fn = rpc.declare({
+		object: 'service',
+		method: 'list',
+		params: ['name'],
+		expect: { [name]: { instances: { [name]: {} }}},
+	});
+	return () => fn(name);
+}
+
+const getAGHServiceInfo = getServiceInfo('adguardhome');
+
+async function getStatus() {
+	try {
+		const res = await getAGHServiceInfo();
+		const isRunning = res?.instances?.adguardhome?.running;
+		return isRunning ?? false;
+	} catch (e) {
+		console.error(e);
+		return false;
+	}
+}
+
+function getStatusValue(isRunning) {
+	return isRunning ? RUNNING_SPAN : NOT_RUNNING_SPAN;
+}
+
+async function getVersion() {
+	try {
+		const res = await fs.exec('/usr/bin/AdGuardHome', ['--version']);
+		const version = res.stdout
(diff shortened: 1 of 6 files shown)


systemcrash:
Another yaml config monster :) 

Those go variables are globals, yes?

GeorgeSapkin:
> Those go variables are globals, yes?

Not sure I understand what you mean by globals here. These are per-process.

systemcrash:
Per-process, implying they control the behaviour of all Go binaries on the system?

GeorgeSapkin:
> Per-process, implying they control the behaviour of all Go binaries on the system?

They can be set for each Go process individually. They could be set on a system-level, but that doesn't make sense.

Here's how it's used for AGH:
- https://github.com/openwrt/packages/blob/e192c5983e50169ed984d48b094798869a0a780b/net/adguardhome/files/adguardhome.init#L66-L68

```bash
	[ "$gc" -le 0 ] || procd_append_param env GOGC="$gc"
	[ "$maxprocs" -le 0 ] || procd_append_param env GOMAXPROCS="$maxprocs"
	[ "$memlimit" -le 0 ] || procd_append_param env GOMEMLIMIT="$memlimit"
```

systemcrash:
Ah, OK. That looks like it's set only for the agh sub-process. Is this ready, then?

GeorgeSapkin:
If nothing jumps out as wrong. Particularly around polling and updating a UI element. I couldn't figure out something that looked idiomatic to update the _Service Status_ field.

applications/luci-app-adguardhome/htdocs/luci-static/resources/view/adguardhome/config.js
```diff
@@ -0,0 +1,261 @@
...
+			'general',
+			form.Flag,
+			'advanced_settings',
+			_('Advanced Settings'),
+		);
+		advSettingsOpt.default = '0';
+		advSettingsOpt.rmempty = false;
+		advSettingsOpt.load = () => sessionStorage.getItem(STORAGE_KEY) || '0';
+		advSettingsOpt.remove = () => {};
+		advSettingsOpt.write = (_, value) => sessionStorage.setItem(STORAGE_KEY, value);
```
systemcrash, line 205: This is a bit odd. Why store a key for agh in a different uci config file?

applications/luci-app-adguardhome/htdocs/luci-static/resources/view/adguardhome/config.js
```diff
@@ -0,0 +1,261 @@
...
+		memLimitOpt.depends('advanced_settings', '1');
+		memLimitOpt.placeholder = DEFAULT_GOMEMLIMIT;
+		memLimitOpt.retain = true;
+
+		const rendered = await map.render();
+
+		const statusNode = map.findElement('data-field', statusOpt.cbid('status_section'));
+		poll.add(updateStatus(statusNode), POLL_INTERVAL);
+
+		return rendered;
```
systemcrash, line 259: This is OK, although look through the codebase and see how it's done elsewhere. 

Typically return map.render().then(...);

applications/luci-app-adguardhome/htdocs/luci-static/resources/view/adguardhome/config.js
GeorgeSapkin: This is a session storage entry that marks that advanced settings tab is visible. Nothing is stored in UCI.

systemcrash: So user logs out and in and the settings are hidden anew?

systemcrash: The retain = 1 should help, otherwise the dependent settings get erased.

GeorgeSapkin: I'm not familiar with all the knobs yet. Can you make a small snippet of what needs to be modified in this case?

systemcrash: Do you want to store something permanently or only while you're logged in?

GeorgeSapkin: Something for the duration of the session, hence why I used the session storage. But while logged in also works if that's more idiomatic.

systemcrash: Then what you have is fine. It's not idiomatic with conventions in this repo so others might find this and think it's a bug.

systemcrash:
Let's see how she fares. master only?

GeorgeSapkin:
> Let's see how she fares. master only?

Yeah, it depends on config changes on the service side that are not in stable yet.

PR #8415, opened 2026-03-12
luci-app-upnp: Revision, new access control and UCI options…

### LINKED ISSUE #2985: luci-app-upnp: Cannot see added IPv6 UPnP/PCP rules in LuCI interface
When dynamically adding a IPv6 Firewall rule (AddPinhole) making a request to miniupnpd 2.1 in OpenWrt 18.06 it adds the rule correctly to ip6tables chain and allow incoming traffic towards the target device in the LAN.

However on luci-app-upnp it doesn't show anything in the UPnP/PCP LuCI interface as it does for IPv4 added rules. This may be a problem for users that may not have familiarity to command line if a added rule must be removed before the lease time expires or when there is no lease time to avoid security issues.

luci-app-upnp version used is git-19.218.48077-4468345-1

### LINKED ISSUE #5197: luci-app-upnp: IPv6 pinholes aren't displayed
Powered by LuCI openwrt-19.07 branch (git-21.189.23240-7b931da) / OpenWrt 19.07.7 r11306-c4a6851c72
luci-app-upnp	git-21.189.23240-7b931da-1
miniupnpd	2.2.0-4
miniupnpc	2.1.20190408-2 (on router)

I'm not sure if this is a problem with the luci app, a limitation of miniupnpd or that Transmission isn't setting pinholes, or a limitation of miniupnpc for which it has no method of getting a pinhole listing.

I guess what I need is help trying to debug whether pinholing is working at all

[https://miniupnp.tuxfamily.org/forum/viewtopic.php?p=4799](url)

```
davygrvy@puukukui:~$ upnpc -L
upnpc : miniupnpc library test client, version 2.1.
 (c) 2005-2019 Thomas Bernard.
Go to http://miniupnp.free.fr/ or https://miniupnp.tuxfamily.org/
for more information.
List of UPNP devices found on the network :
 desc: http://192.168.0.1:5000/rootDesc.xml
 st: urn:schemas-upnp-org:device:InternetGatewayDevice:1

Found valid IGD : http://192.168.0.1:5000/ctl/IPConn
Local LAN ip address : 192.168.0.241
 i protocol exPort->inAddr:inPort description remoteHost leaseTime
 0 TCP 51413->192.168.0.241:51413 'NAT-PMP 51413 tcp' '' 986
 1 TCP  8333->192.168.0.241:8333  'Bitcoin Core v0.21.1' '' 604681
 2 UDP 51413->192.168.0.241:51413 'Transmission at 51413' '' 482188
 3 UDP  9308->192.168.0.217:9308  '192.168.0.217:9308 to 9308 (UDP)' '' 460923
davygrvy@puukukui:~?
```

### LINKED ISSUE #8313: Feature request: luci-app-upnp - provide or comment known-good STUN host/port options.
### What would you like to see in luci?

GL-iNet GL-BE6500 has a "Network Setting" -> "Drop-in Gateway" feature to allow chaining it after an old ISP router to take over DHCP, DNS, NAT, Firewall, for its own subnet, but uPNP won't work, because it can only see the old ISP routers address, not the WAN address. STUN resolves this gotcha, but I had to find https://forum.openwrt.org/t/solution-to-upnp-not-working/135748 to figure out what STUN was, and what value to use. There are probably safer other value too, because Google can't really be trusted.

### LINKED ISSUE #8677: luci-app-upnp : Missing field for "internal_iface"

### screenshots or captures

Image
Image
Image

### Actual behaviour

It is not possible to set the internal_iface property of /etc/config/upnpd config file from the GUI

### Expected behaviour

Add a field for selecting interfaces to be used for internal_iface.
Write the selected interfaces in the config file /etc/config/upnpd in the field internal_iface

### Steps to reproduce

1. Install luci-app-upnp
2. In the GUI, in the top banner, select Services and click on UPnP IGD & PCP
3. Go in the Service Setup tab, there is no field for setting internal_iface
4. Go in the Advanced Settings tab, there is no field for setting internal_iface

### Additional Information

```shell
{
        "kernel": "6.12.87",
        "hostname": "hostname",
        "system": "MediaTek MT7621 ver:1 eco:3",
        "model": "TP-Link ER605 v2",
        "board_name": "tplink,er605-v2",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "25.12.4",
                "revision": "r32933-4ccb782af7",
                "target": "ramips/mt7621",
                "description": "OpenWrt 25.12.4 r32933-4ccb782af7",
                "builddate": "1778712129"
        }
}
```

Firefox

### Relevant log output

```shell

```

### LINKED ISSUE #12843: [openwrt-23.05] iptables fails to display DNAT rules (was: luci-app-upnp:"There are no active redirects",but miniupnpd works fine.)
I built the latest OpenWrt-23.05, which includes dual WAN and Passwall support. I disabled nftables and compiled fw3 because mwan3 is incompatible with fw4 and Passwall. The debug logs of miniupnpd and the firewall both indicated that they were working fine. However, UPnP did not display any entries in the Luci interface. The same configuration worked perfectly in OpenWrt-22.03.

DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='23.05-SNAPSHOT'
DISTRIB_REVISION='r23058-e2e7dd4450'
DISTRIB_TARGET='x86/64'
DISTRIB_ARCH='x86_64'
DISTRIB_DESCRIPTION='OpenWrt 23.05-SNAPSHOT r23058-e2e7dd4450'
DISTRIB_TAINTS='no-all mklibs'
image

Self-Hosting-Group:
As this PR is extensive, the descriptions of the individual commits are collapsed here:

<details><summary><b>1. Improve existing UI</b></summary>

- Workaround daemon bug to fix listing of active port maps with iptables >= 1.8.8, and relax nftables parsing for the new release
- Change `Active Service Port Maps` to `Active Port Maps`, use the same wording for the table headings and ACL as on the overview status page
- Change description field header to `Added via / description`, always include the protocol and clearer/less redundant protocol labels
- Disable IPv6 mapping (`ipv6_disable`): UI option added, UCI exists
- Don't clear dependent options if UPnP IGD/STUN temporarily disabled
- Set `notify_interval` minimum to 900 s (default), as recommended by [UDA 1.1] (2x=1800 in the standard), because daemon/OpenWrt wrongly suggested 30x less in the past, and to reduce multicast traffic and power consumption in wireless networks, clearer help
- Report system instead of service uptime, UUID and service lease file: UI options removed (disabled, to keep translations) as rarely used
- Slightly improve some text strings
- Show hint if no suitable configuration found (missing related update), link to package manager, and show form as read-only to prevent changes
- Configure service lease file by default in rpcd ucode for full UI functionality without option set
- Refactoring by moving ubus connection as jow-/ucode@a58fe47 was merged
- Adapt to JavaScript ES6 and remove some line breaks and format files
- A commit has been prepared that manually adjusts the translations, without losing them

Fixes: https://github.com/openwrt/openwrt/issues/12843

[UDA 1.1]: https://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf#page=30
</details>

<details><summary><b><i>2. Add UPnP IGD Adjustments tab</i></b></summary>

And rearrange as many options

(to merge with prior)
</details>

<details><summary><b>3. Revision and adapt to new UCI options</b></summary>

The following settings UCI options been added or changed, and the previous options are migrated on updating:

- Only display `Active Port Maps` if the service is enabled and `Access Control List` if it is used
- Enable protocols (`enable_protocols`): Combined UI option added
- Allow CGNAT/STUN (`allow_cgnat`): Accept new values for IPv4 CGNAT use and update help with newer wording of RFC 5780
- STUN server (`stun_host`): Allow port inclusion
- STUN port: Removed, as now accepted in STUN server
- Override external IPv4 (`external_ip`): UI option added for CGNAT use
- Allow third-party mapping (`allow_third_party_mapping`): Inverted from secure mode and optionally extended to PCP
- Log level (`log_output`): Allow info log level
- UPnP IGD compatibility (`upnp_igd_compat`): Reworded/extensible
- Download/upload speed (`download_kbps`/`upload_kbps`): In kbit/s and datatype set, now, interface link speed by default
- Router/friendly name (`friendly_name`): UI option added to set name displayed in Windows Explorer, model/serial number removed
- Enable Networks / Access Control (`internal_network`): Section added to select the enabled networks and their access control. By:
  - Internal network (`interface`): UI option added to select the local/internal (LAN) network interface to enable the service for
  - Access defaults (`access_defaults`): UI option added to select access defaults for ports that all devices on the network can map
  - Accept extra ports (`accept_ports`): UI option added to accept these ports or port ranges on the network as well
  - Reject ports (`reject_ports`): UI option added to reject ports on the network; overrides other settings
  - Check ACL (`check_acl`): UI option added to check the ACL entries first; these extend/override the defaults
- Slightly improve introduction text

More details on changed options can be found in the dependent package PR

Fixes: https://github.com/openwrt/luci/issues/8313
Fixes: https://github.com/openwrt/luci/issues/8677
Depends on: https://github.com/openwrt/packages/pull/28765
</details>

<details><summary><b><i>4. Rename UCI section to settings v2.0</i></b></summary>

Rename UCI section `config` (v1.0) -> `settings` (v2.0), helps on migration and to distinguish the updated config from the previous one

(to merge with prior)
</details>

<details><summary><b><i>5. Update ACL options, migrate section</i></b></summary>

- Disableable and cloneable ACL entries, always translated `Action`
- Improve UI with direct editability, clearer help wording, and rename to `Access Control List`
- The ACL is now rejected last if not accepted by access defaults. Add (disabled) ACL template entries on migration
- Migrate ACL entries to the new section name `acl_entry`
- The following ACL UCI options been added or changed, and the previous options are migrated on updating:

acl_entry UCI options       | Change                     | Previous name
----------------------------|----------------------------|--------------
action                      | New/updated values     (1) |
int_port                    | Remove colon separator (2) | int_ports
ext_port                    | Remove colon separator (2) | ext_ports
descr_filter                | New option             (3) |

1. Allow disabled, and update action option to use the nftables terms (allow/deny -> accept/reject). To avoid adding inverted actions when changing via LuCI, ensure any missing are set, as LuCI and UCI had not matching action defaults. Missing actions are now ignored/logged
2. Ensure that the hyphen (-) is only used as a port range separator by migration, as the colon (:) is not valid in LuCI
3. Add missing UCI option to set a regular expression to check for a UPnP IGD IPv4 port map description, and fix the current collision with the comment field which was not noticed due to a daemon bug https://github.com/openwrt/packages/pull/24495 https://github.com/miniupnp/miniupnp/pull/853

(to merge with prior)
</details>

<details><summary><b><i>6. List active IPv6 port maps</i></b></summary>

Supports nftables and ip6tables; IPv6 port maps not yet deletable, as not parsing extra IPv6 lease file

(Note: Daemon's IPv6 lease file differs from the IPv4 format in order and separator. With IPv4, the remote IP field is missing and not retained during daemon restarts. IMHO, it would be better to have a complete/combined IPv4/IPv6 lease file upstream, with unique mapping IDs for reliable deletion in the 1st field, with a standardised order and more fields, and then use it without rewriting the parsing.)

(to merge with prior)

Fixes: https://github.com/openwrt/luci/issues/2985
Fixes: https://github.com/openwrt/luci/issues/5197
Fixes: https://github.com/openwrt/luci/pull/8567
</details>

<details><summary><b><i>7. Service-wide access control settings</i></b></summary>

This implements variant B

upnpd.settings UCI options   | Change                    | Previous name
-----------------------------|---------------------------|--------------
access_defaults              | New option            (1) |
accept_ports                 | New option            (1) |
reject_ports                 | New option            (2) |
check_acl                    | New option            (1) |

Notes:
1. New options added for default ports that all devices can map, client-specific permissions using the access control list (ACL) can extend/override the defaults. Access defaults: accept-high-ports accept-web+high-ports/accept-web-ports/accept-all-ports
2. Reject ports; overrides other settings. By default reject unsafe: FTP (21), Telnet (23), DCE/NetBIOS/SMB (135/137-139/445), RDP (3389)

(to merge with prior)
</details>

(The italic commits are intended to be merged with the prior ones after review)

#### Screenshots

The new service-wide access control functionality… can best be described using the LuCI screenshots:

<details><summary><b>New Access Control tab</b></summary>

luci-access-control
</details>

<details><summary><b>Advanced Settings with new CGNAT functionality</b></summary>

luci-advanced
</details>

<details><summary><b>UPnP IGD Adjustments tab (new)</b></summary>

luci-igd
</details>

<details><summary><b>LuCI notification if the related package is not updated</b></summary>

luci-notification
</details>

<details><summary><b>Full LuCI screenshot</b></summary>

luci-screenshot
</details>

Depends on packages PR: https://github.com/openwrt/packages/pull/28765
Tested on: OpenWrt 24.10.8/25.12.5 with iptables/nftables
Maintainer: @jow-

**miniupnpd: Core functionality issues**
https://github.com/Self-Hosting-Group/miniupnpd-issues

The Port Control Protocol (PCP) is the successor to NAT-PMP, shares similar protocol concepts and packet formats, but supports IPv6 port mapping and options/extensions. For more information, see:
**Port Mapping Protocols Overview and Comparison 2026+: About UPnP IGD & PCP/NAT-PMP**

### DIFF
--- a/applications/luci-app-upnp/Makefile
+++ b/applications/luci-app-upnp/Makefile
@@ -6,7 +6,7 @@
 
 include $(TOPDIR)/rules.mk
 
-LUCI_TITLE:=UPnP IGD & PCP/NAT-PMP configuration module | Universal Plug and Play
+LUCI_TITLE:=LuCI support for UPnP IGD & PCP/NAT-PMP service | originally: Universal Plug and Play
 LUCI_DEPENDS:=+luci-base +miniupnpd +rpcd-mod-ucode
 
 PKG_LICENSE:=Apache-2.0
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
@@ -70,15 +70,19 @@ return baseclass.extend({
 				rule.proto,
 				expires_str,
 				'%h'.format(rule.descr),
-				E('button', {
+				rule.num != '0' ? E('button', {
 					'class': 'btn cbi-button-remove',
 					'click': L.bind(handleDelRule, this, rule.num),
 					'title': _('Delete')
+				}, [_('Delete')]) : E('button', {
+					'class': 'btn cbi-button-remove',
+					'title': _('Not yet deletable; UPnP IGDv2 IPv6 port maps expire sooner, have no description'),
+					'style': 'cursor: not-allowed'
 				}, [_('Delete')])
 			];
 		});
 
-		cbi_update_table(table, rows, E('em', _('There are no active port maps.')));
+		cbi_update_table(table, rows, E('em', _('There are no active IPv4/IPv6 port maps')));
 
 		return table;
 	}
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
@@ -7,14 +7,14 @@
 const callUpnpGetStatus = rpc.declare({
 	object: 'luci.upnp',
 	method: 'get_status',
-	expect: {  }
+	expect: {}
 });
 
 const callUpnpDeleteRule = rpc.declare({
 	object: 'luci.upnp',
 	method: 'delete_rule',
-	params: [ 'token' ],
-	expect: { result : "OK" },
+	params: ['token'],
+	expect: { result: 'OK' },
 });
 
 function handleDelRule(num, ev) {
@@ -35,23 +35,23 @@ return baseclass.extend({
 	},
 
 	render: function(data) {
-		var table = E('table', { 'class': 'table', 'id': 'upnp_status_table' }, [
+		const table = E('table', { 'class': 'table', 'id': 'upnp_status_table' }, [
 			E('tr', { 'class': 'tr table-titles' }, [
-				E('th', { 'class': 'th' }, _('Client Name')),
-				E('th', { 'class': 'th' }, _('Client Address')),
-				E('th', { 'class': 'th' }, _('Client Port')),
-				E('th', { 'class': 'th' }, _('External Port')),
+				E('th', { 'class': 'th' }, _('Hostname')),
+				E('th', { 'class': 'th' }, _('IP address')),
+				E('th', { 'class': 'th' }, _('Port')),
+				E('th', { 'class': 'th' }, _('External port')),
 				E('th', { 'class': 'th' }, _('Protocol')),
 				E('th', { 'class': 'th right' }, _('Expires')),
-				E('th', { 'class': 'th' }, _('Description')),
+				E('th', { 'class': 'th' }, _('Added via / description')),
 				E('th', { 'class': 'th cbi-section-actions' }, '')
 			])
 		]);
 
-		var rules = Array.isArray(data[0].rules) ? data[0].rules : [];
+		const rules = Array.isArray(data[0].rules) ? data[0].rules : [];
 
-		var rows = rules.map(function(rule) {
-			const padnum = (num, length) => num.toString().padStart(length, "0");
+		const rows = rules.map(function(rule) {
+			const padnum = (num, length) => num.toString().padStart(length, '0');
 			const expires_sec = rule?.expires || 0;
 			const hour = Math.floor(expires_sec / 3600);
 			const minute = Math.floor((expires_sec % 3600) / 60);
@@ -72,8 +72,9 @@ return baseclass.extend({
 				'%h'.format(rule.descr),
 				E('button', {
 					'class': 'btn cbi-button-remove',
-					'click': L.bind(handleDelRule, this, rule.num)
-				}, [ _('Delete') ])
+					'click': L.bind(handleDelRule, this, rule.num),
+					'title': _('Delete')
+				}, [_('Delete')])
 			];
 		});
 
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
@@ -147,6 +147,7 @@ return view.extend({
 		s = m.section(form.NamedSection, 'settings', 'upnpd', _('Service Settings'));
 		s.addremove = false;
 		s.tab('setup', _('Service Setup'));
+		s.tab('access_control', _('Access Control'));
 		s.tab('advanced', _('Advanced Settings'));
 		s.tab('igd', _('UPnP IGD Adjustments'));
 
@@ -161,6 +162,18 @@ return view.extend({
 		o.default = 'all';
 		o.widget = 'radio';
 
+		o = s.taboption('setup', widgets.NetworkSelect, 'internal_iface', _('Enable networks'),
+			_('Select local/internal (LAN) network interfaces to enable the service for'));
+		o.nocreate = true;
+		o.multiple = true;
+		o.rmempty = false;
+		o.filter = function(section_id, value) {
+			return (value == 'wan' || value == 'wan6') ? '' : value;
+		};
+		o.write = function(section_id, formvalue) {
+			uci.set('upnpd', section_id, 'internal_iface', Array.isArray(formvalue) ? formvalue.join(' ') : formvalue);
+		};
+
 		o = s.taboption('setup', form.ListValue, 'upnp_igd_compat', _('UPnP IGD compatibility'),
 			_('Set compatibility mode (act as device) to workaround IGDv2-incompatible clients; %s are known to only work with %s (or) <br />Emulate/report a specific/different device to workaround/support/handle/bypass/assist/mitigate... (Alternative text welcome)').format('Sony PS, Activision CoD…', 'IGDv1'));
 		o.value('igdv1', _('IGDv1 (IPv4 only)'));
@@ -168,6 +181,35 @@ return view.extend({
 		o.depends('enable_protocols', 'upnp-igd');
 		o.depends('enable_protocols', 'all');
 		o.retain = true;
+
+		o = s.taboption('access_control', form.ListValue, 'access_defaults', _('Access defaults'),
+			_('Set access control defaults for ports that all devices can map'));
+		o.value('', _('None / accept extra ports only'));
+		o.value('accept-high-ports', _('Accept ports >= 1024'));
+		o.value('accept-web+high-ports', _('Accept HTTP/HTTPS + ports >= 1024'));
+		o.value('accept-web-ports', _('Accept HTTP/HTTPS ports'));
+		o.value('accept-all-ports', _('Accept all ports'));
+
+		o = s.taboption('access_control', form.Value, 'accept_ports', _('Accept extra ports'));
+		o.datatype = 'list(portrange)';
+
+		o = s.taboption('access_control', form.Value, 'reject_ports', _('Reject ports'),
+			_('Reject unsafe/insecure/risky FTP/Telnet/DCE/NetBIOS/SMB/RDP ports by default; overrides other settings; use %s for none').format('<code>0</code>'));
+		o.datatype = 'list(portrange)';
+		o.placeholder = '21 23 135 137-139 445 3389';
+		o.modalonly = true;
+
+		o = s.taboption('access_control', form.Flag, 'check_acl', _('Check ACL'),
+			_('Extend or override access defaults by device-specific permissions using the access control list (ACL)') + '<br />' +
+			_('Sequence:') + ' 1. ' + _('Reject ports') + ', 2. ' + _('ACL entries (if checked)') + ', 3. ' + _('Access defaults') + ', 4. ' + _('Accept extra ports'));
+		o.default = '1';
+		o.onchange = function(ev, section_id, value) {
+			let acl = document.getElementById('cbi-upnpd-acl_entry');
+			value == 0 ? acl.style.display = 'none' : acl.style.display = 'block';
+		};
+
+		s.taboption('access_control', form.Flag, 'ipv6_disable', _('Disable IPv6 mapping'),
+			_('IPv6 is currently always accepted unless disabled'));
 
 		o = s.taboption('advanced', form.RichListValue, 'allow_cgnat', _('Allow %s/%s', 'Allow %s/%s (%s = CGNAT, %s = STUN)')
 			.format('<a href="https://en.wikipedia.org/wiki/Carrier-grade_NAT" target="_blank" rel="noreferrer"><abbr title="Carrier-grade NAT">CGNAT</abbr></a>',
@@ -198,8 +240,6 @@ return view.extend({
 		o.value('1', _('Enabled'));
 		o.value('upnp-igd', _('Enabled') + ' (' + _('UPnP IGD only') + ')');
 		o.value('pcp', _('Enabled') + ' (' + _('PCP only') + ')');
-
-		s.taboption('advanced', form.Flag, 'ipv6_disable', _('Disable IPv6 mapping'));
 
 		o = s.taboption('advanced', form.Flag, 'system_uptime', _('Report system instead of service uptime'));
 		o.default = '1';
@@ -281,72 +321,16 @@ return view.extend({
 		o.depends('enable_protocols', 'all');
 		o.retain = true;
 
-		s = m.section(form.GridSection, 'internal_network', '<h5>' + _('Enable Networks / Access Control') + '</h5>',
-			_('Select local/internal (LAN) network interfaces to enable the service for.') + ' ' +
-			_('Set access control defaults for ports that all devices on a network can map.') + ' ' +
-			_('IPv6 is currently always accepted unless disabled. (Alternative text welcome)'));
-		s.anonymous = true;
-		s.addremove = true;
-		s.cloneable = true;
-		s.sortable = true;
-		s.nodescriptions = true;
-		s.modaltitle = _('UPnP IGD & PCP/NAT-PMP') + ' - ' + _('Edit Network Access Control Settings');
-
-		o = s.option(widgets.NetworkSelect, 'interface', _('Internal network'),
-			_('Select the local/internal (LAN) network interface to enable the service for'));
-		o.nocreate = true;
-		o.editable = true;
-		o.rmempty = false;
-		o.retain = true;
-		o.filter = function(section_id, value) {
-			return (value == 'wan' || value == 'wan6') ? '' : value;
-		};
-
-		o = s.option(form.ListValue, 'access_defaults', _('Access defaults'),
-			_('Set access control defaults for ports that all devices on the network can map'));
-		o.value('', _('None / accept extra ports only'));
-		o.value('accept-high-ports', _('Accept ports >= 1024'));
-		o.value('accept-web+high-ports', _('Accept HTTP/HTTPS + ports >= 1024'));
-		o.value('accept-web-ports', _('Accept HTTP/HTTPS ports'));
-		o.value('accept-all-ports', _('Accept all ports'));
-		o.editable = true;
-		o.retain = true;
-
-		o = s.option(form.Value, 'accept_ports', _('Accept extra ports'));
-		o.datatype = 'list(portrange)';
-		o.retain = true;
-
-		o = s.option(form.Value, 'reject_ports', _('Reject ports'),
-			_('Reject unsafe/insecure/risky FTP/Telnet/DCE/NetBIOS/SMB/RDP ports on the network by default; overrides other settings; use %s for none').format('<code>0</code>'));
-		o.datatype = 'list(portrange)';
-		o.placeholder = '21 23 135 137-139 445 3389';
-		o.modalonly = true;
-		o.retain = true;
-
-		o = s.option(form.Flag, 'check_acl', _('Check ACL'),
-			_('Extend or override access defaults by device-specific permissions using the access control list (ACL)') + '<br />' +
-			_('Sequence:') + ' 1. ' + _('Reject ports') + ', 2. ' + _('ACL entries (if checked)') + ', 3. ' + _('Access defaults') + ', 4. ' + _('Accept extra ports'));
-		o.default = '1';
-		o.editable = true;
-		o.retain = true;
-
 		s = m.section(form.GridSection, 'acl_entry', '<h5>' + _('Access Control List') + '</h5>',
 			_('The access control list (ACL) specifies which IP addresses and ports can be mapped.') + ' ' +
-			_('ACL entries are checked in order, then rejected if not matched and not accepted by access defaults. (should be part of extra tab)'));
+			_('ACL entries are checked in order, then rejected if not matched and not accepted by access defaults. (To do: should be part of access control tab)'));
 		s.anonymous = true;
 		s.addremove = true;
 		s.cloneable = true;
 		s.sortable = true;
 		s.modaltitle = _('UPnP IGD & PCP/NAT-PMP') + ' - ' + _('Edit ACL Entry');
-		// To do: ACL part of extra tab with dependency on option as immediately, and network section part of service setup tab
-		let acl_used = false;
-		for (let ifnr = 0; uci.get('upnpd', `@internal_network[${ifnr}]`, 'interface'); ifnr++) {
-			if (uci.get('upnpd', `@internal_network[${ifnr}]`, 'check_acl') != '0') {
-				acl_used = true;
-				break;
-			}
-		}
-		s.disable = !acl_used;
+		// To do: ACL part of access control tab and hide (or dependency on option) instead of disable as immediately, and to not break onchange function
+		s.disable = uci.get('upnpd', 'settings', 'check_acl') == '0';
 
 		o = s.option(form.Value, 'comment', _('Comment'));
 		o.default = _('unspecified');
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
@@ -322,11 +322,14 @@ return view.extend({
 		o.editable = true;
 		o.retain = true;
 
-		s = m.section(form.GridSection, 'perm_rule', _('Service Access Control List'),
-			_('ACL specify which client addresses and ports can be mapped, IPv6 always allowed.'));
+		s = m.section(form.GridSection, 'acl_entry', '<h5>' + _('Access Control List') + '</h5>',
+			_('The access control list (ACL) specifies which IP addresses and ports can be mapped.') + ' ' +
+			_('ACL entries are checked in order, then rejected if not matched and not accepted by access defaults. (should be part of extra tab)'));
 		s.anonymous = true;
 		s.addremove = true;
+		s.cloneable = true;
 		s.sortable = true;
+		s.modaltitle = _('UPnP IGD & PCP/NAT-PMP') + ' - ' + _('Edit ACL Entry');
 		// To do: ACL part of extra tab with dependency on option as immediately, and network section part of service setup tab
 		let acl_used = false;
 		for (let ifnr = 0; uci.get('upnpd', `@internal_network[${ifnr}]`, 'interface'); ifnr++) {
@@ -337,23 +340,46 @@ return view.extend({
 		}
 		s.disable = !acl_used;
 
-		s.option(form.Value, 'comment', _('Comment'));
+		o = s.option(form.Value, 'comment', _('Comment'));
+		o.default = _('unspecified');
 
 		o = s.option(form.Value, 'int_addr', _('IP address'));
 		o.datatype = 'ip4addr';
-		o.placeholder = '0.0.0.0/0';
+		o.rmempty = false;
+		o.editable = true;
+		o.retain = true;
+		o.modalonly = false;
 
-		o = s.option(form.Value, 'int_ports', _('Port'));
-		o.datatype = 'portrange';
-		o.placeholder = '1-65535';
+		o = s.option(form.Value, 'int_addr', _('IP address'),
+			_('Enter device\'s IPv4 address, or an address and netmask'));
+		o.datatype = 'ip4addr';
+		o.rmempty = false;
+		o.retain = true;
+		o.modalonly = true;
 
-		o = s.option(form.Value, 'ext_ports', _('External port'));
+		o = s.option(form.Value, 'int_port', _('Port'));
 		o.datatype = 'portrange';
-		o.placeholder = '1-65535';
+		o.placeholder = '1-65535 (' + _('any port') + ')';
+		o.editable = true;
+		o.retain = true;
+
+		o = s.option(form.Value, 'ext_port', _('External port'));
+		o.datatype = 'portrange';
+		o.placeholder = '1-65535 (' + _('any port') + ')';
+		o.editable = true;
+		o.retain = true;
+
+		o = s.option(form.Value, 'descr_filter', _('Description filter'),
+			_('Regular expression to check description of UPnP IGD IPv4 port maps'));
+		o.placeholder = '.* (' + _('any description') + ')';
+		o.modalonly = true;
 
 		o = s.option(form.ListValue, 'action', _('Action'));
-		o.value('allow', _('Allow'));
-		o.value('deny', _('Deny'));
+		o.value('accept', _('Accept'));
+		o.value('reject', _('Reject'));
+		o.value('disabled', _('Disabled'));
+		o.editable = true;
+		o.retain = true;
 
 		return m.render().then(L.bind(function(m, nodes) {
 			if (uci.get('upnpd', 'settings', 'enabled') != '0') {
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
@@ -85,16 +85,16 @@ return view.extend({
 				'The %s (%s = UPnP IGD & PCP/NAT-PMP) protocols/service enable permitted devices on local networks to autonomously set up IPv4/IPv6 port maps/forwards on this router.')
 			.format(protocols)
 		);
-		if (!uci.get('upnpd', 'config')) {
+		if (!uci.get('upnpd', 'settings')) {
 			ui.addNotification(null, E('div', '<h4>' + _('No suitable configuration was found!') + '</h4><p>' +
-				_('No suitable config (LuCI app %s) found in %s. A related package update (daemon or LuCI app) may be missing.').format('v1.0', '<code>/etc/config/upnpd</code>') + '<br />' +
+				_('No suitable config (LuCI app %s) found in %s. A related package update (daemon or LuCI app) may be missing.').format('v2.0', '<code>/etc/config/upnpd</code>') + '<br />' +
 				_('Use the software package manager, update lists, and install the related update. Config is migrated with the daemon package update.') + '</p>' +
 				'<a class="btn" href="/cgi-bin/luci/admin/system/package-manager?query=UPnP%20IGD%20&amp;%20PCP/NAT-PMP">' + _('Go to package manager…') + '</a>'), 'warning');
 			m.readonly = true;
 		}
 
 		s = m.section(form.GridSection, '_active_rules');
-		s.disable = uci.get('upnpd', 'config', 'enabled') == '0';
+		s.disable = uci.get('upnpd', 'settings', 'enabled') == '0';
 
 		s.render = L.bind(function(view, section_id) {
 			const table = E('table', { 'class': 'table cbi-section-table', 'id': 'upnp_status_table' }, [
@@ -136,7 +136,7 @@ return view.extend({
 			]);
 		}, o, this);
 
-		s = m.section(form.NamedSection, 'config', 'upnpd', _('Service Settings'));
+		s = m.section(form.NamedSection, 'settings', 'upnpd', _('Service Settings'));
 		s.addremove = false;
 		s.tab('setup', _('Service Setup'));
 		s.tab('advanced', _('Advanced Settings'));
@@ -356,7 +356,7 @@ return view.extend({
 		o.value('deny', _('Deny'));
 
 		return m.render().then(L.bind(function(m, nodes) {
-			if (uci.get('upnpd', 'config', 'enabled') != '0') {
+			if (uci.get('upnpd', 'settings', 'enabled') != '0') {
 				poll.add(L.bind(function() {
 					return Promise.all([
 						callUpnpGetStatus()
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
@@ -138,6 +138,7 @@ return view.extend({
 		s.addremove = false;
 		s.tab('setup', _('Service Setup'));
 		s.tab('advanced', _('Advanced Settings'));
+		s.tab('igd', _('UPnP IGD Adjustments'));
 
 		o = s.taboption('setup', form.Flag, 'enabled', _('Enable service'),
 			_('Enable the autonomous port mapping service'));
@@ -149,23 +150,13 @@ return view.extend({
 		o = s.taboption('setup', form.Flag, 'enable_natpmp', _('Enable PCP/NAT-PMP protocols'));
 		o.default = '1';
 
-		s.taboption('setup', form.Flag, 'ext_allow_private_ipv4', _('Allow private IPv4'),
+		s.taboption('advanced', form.Flag, 'ext_allow_private_ipv4', _('Allow private IPv4'),
 			_('Enable forwarding for private/reserved IPv4 address'));
 
 		o = s.taboption('setup', form.Flag, 'igdv1', _('UPnP IGDv1 compatibility mode'),
 			_('Advertise as IGDv1 (IPv4 only) device instead of IGDv2'));
 		o.default = '1';
 		o.rmempty = false;
-		o.depends('enable_upnp', '1');
-		o.retain = true;
-
-		o = s.taboption('setup', form.Value, 'download', _('Download speed'),
-			_('Report maximum download speed in kByte/s'));
-		o.depends('enable_upnp', '1');
-		o.retain = true;
-
-		o = s.taboption('setup', form.Value, 'upload', _('Upload speed'),
-			_('Report maximum upload speed in kByte/s'));
 		o.depends('enable_upnp', '1');
 		o.retain = true;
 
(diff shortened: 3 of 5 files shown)


applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+			_('Check the ACL entries first; these extend/override the defaults') + '<br />' +
+			_('Sequence: 1. Reject ports, 2. ACL entries (if checked), 3. Preset ports, 4. Accept ports'));
 		o.default = '1';
+		o.editable = true;
 		o.rmempty = false;
-		o.depends('enable_upnp', '1');
+		o.onchange = function(ev, section_id, value) {
+			let acl = document.getElementById("cbi-upnpd-acl_entry");
+			value == 0 ? acl.style.display = 'none' : acl.style.display = 'block';
+		};
```
openwrt-ai, line 216: If the page is loaded with `check_acl == '0'`, the ACL grid section at line 360 is skipped via `s.disable = true`, so the element `#cbi-upnpd-acl_entry` does not exist in the DOM. When the user then toggles `check_acl` to enabled, `document.getElementById("cbi-upnpd-acl_entry")` returns `null` and the next statement throws `TypeError: Cannot set properties of null (setting 'style')`, leaving the form in a broken state. Either render the section unconditionally and only hide it via `style.display` (so the element is always present), or guard the assignment with `if (acl)` and re-render on toggle.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+		s.sortable = true;
+		s.modaltitle = _('UPnP IGD & PCP') + ' - ' + _('Edit ACL Entry');
+		// Preferably: ACL part of access control tab with depends for section as immediately. Nice to have: Add button (+input) calls function and opens modal pre-filled
+		let acl_used = false;
+		for (let ifnr = 0; uci.get('upnpd', `@internal_network[${ifnr}]`, 'interface'); ifnr++) {
+			if (uci.get('upnpd', `@internal_network[${ifnr}]`, 'check_acl') != '0') {
+				acl_used = true;
+				break;
+			}
+		}
```
openwrt-ai: `acl_used` is computed here but never read anywhere. Either wire it into the `s.disable` decision below (the loop seems intended to drive that, since per-network `check_acl` overrides exist in the dependent packages PR) or drop the dead block. Same applies to the `// Preferably:` TODO line 352 — these notes belong in the commit message / an issue, not in committed source.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+		o.depends('enable_protocols', 'upnp-igd');
+		o.depends('enable_protocols', 'all');
+		o.retain = true;
 
-		s.taboption('advanced', form.Flag, 'log_output', _('Enable additional logging'),
-			_('Puts extra debugging information into the system log'));
+		o = s.taboption('igd', form.Value, 'uuid', _('Device UUID'));
+		// o.depends('enable_protocols', 'upnp-igd');
+		// o.depends('enable_protocols', 'all');
+		o.depends('to-disable-as-rarely-used', '1');
```
openwrt-ai: Using a non-existent UCI option name (`to-disable-as-rarely-used`) as a sentinel dependency to permanently hide options is a hack and silently turns the `taboption()` calls into dead UI: the options can never be edited, but they still get loaded/validated by the form machinery. Five sites do this (lines 266, 277, 305, 311, 324). If the goal is just to preserve translation strings, consider keeping the `_('…')` literals in a `// translators: keep…` block (or a dummy unused array) and removing the `taboption()` calls entirely; otherwise gate them behind a real config flag. As-is, a future reader will not understand what `'to-disable-as-rarely-used'` means.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+		o = s.taboption('setup', form.ListValue, 'enable_protocols', _('Enable protocols'));
+		o.value('all', _('All protocols'));
+		o.value('upnp-igd', _('UPnP IGD'));
+		o.value('pcp+nat-pmp', _('PCP and NAT-PMP'));
+		o.default = 'all';
+		o.widget = 'radio';
+
+		o = s.taboption('setup', widgets.NetworkSelect, 'internal_iface', _('Enable networks'),
+			_('Select local/internal (LAN) network interfaces to enable the service for'));
+		o.exclude = 'wan'; // wan6 should also be excluded
```
openwrt-ai: `o.exclude` on `widgets.NetworkSelect` accepts only a single interface name, so `wan6` is not actually excluded here despite the (acknowledged) TODO. The `validate` function below catches the `'wan6'` case for a single selection, but with `multiple = true` a value like `'lan wan6'` joins to a space-separated string and the `== 'wan6'` check no longer matches. Either iterate `value.split(' ')` in `validate`, or use a `filter` callback to exclude both `wan` and `wan6` from the picker. The placeholder TODO should be resolved before merge.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+		o.nocreate = true;
+		o.editable = true;
+		o.multiple = true;
+		o.rmempty = false;
+		o.retain = true;
+		o.validate = function(section_id, value) {
+			return (value == '' || value == 'wan' || value == 'wan6') ? _('Expecting: %s').format(_('internal network interfaces')) : true;
+		};
+		o.write = function(section_id, formvalue) {
+			uci.set('upnpd', section_id, 'internal_iface', formvalue.join(' '));
```
openwrt-ai: `formvalue.join(' ')` will throw if `formvalue` is `null`/`undefined` — `NetworkSelect.formvalue()` can return `null` when nothing is selected (and `rmempty = false` plus `validate` may not catch every path that reaches `write`). Guard with `Array.isArray(formvalue) ? formvalue.join(' ') : ''` (or call `uci.unset(...)` for empty), otherwise saving an empty selection can blow up.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+		m = new form.Map('upnpd', _('UPnP IGD & PCP/NAT-PMP Service'),
+			_('The %s protocols/service enable permitted devices on local networks to autonomously set up IPv4/IPv6 port maps (forwards) on this router.',
+				'The %s (%s = UPnP IGD & PCP/NAT-PMP) protocols/service enable permitted devices on local networks to autonomously set up IPv4/IPv6 port maps (forwards) on this router.')
+			.format(protocols)
 		);
+		if (!uci.get('upnpd', 'settings')) {
+			ui.addNotification(null, E('div', '<h4>' + _('No suitable configuration was found!') + '</h4><p>' +
+				_('No suitable config (LuCI app %s) found in %s. A related package update (daemon or LuCI app) may be missing.').format('v2.0', '<code>/etc/config/upnpd</code>') + '<br />' +
+				_('Use the software package manager, update lists, and install the related update. Config is migrated with the daemon package update.') + '</p>' +
+				'<a class="btn" href="/cgi-bin/luci/admin/system/package-manager?query=UPnP%20IGD%20&%20PCP/NAT-PMP">' + _('Go to package manager…') + '</a>'), 'warning');
```
openwrt-ai: The package-manager link is built by string concatenation with `&` left unencoded inside the query string (`query=UPnP%20IGD%20&%20PCP/NAT-PMP`), which the browser parses as two query params and the second one (` PCP/NAT-PMP`) is lost / interpreted as a new key. Encode the whole value:

```suggestion
				_('Use the software package manager, update lists, and install the related update. Config is migrated with the daemon package update.') + '</p>' +
				'<a class="btn" href="/cgi-bin/luci/admin/system/package-manager?query=' + encodeURIComponent('UPnP IGD & PCP/NAT-PMP') + '">' + _('Go to package manager…') + '</a>'), 'warning');
```

Also, the surrounding `E('div', '<h4>…</h4>')` passes raw HTML as a text child — the existing pattern in luci-base is `E('div', {}, [ E('h4', …), E('p', …), … ])` so the markup is built as DOM and translated strings can't accidentally inject HTML. Worth reworking while you're touching this.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+			_('ACL entries are checked in order and, by default, rejected with no preset. (should be part of access control tab)'));
 		s.anonymous = true;
 		s.addremove = true;
-
-		s.option(form.Value, 'comment', _('Comment'));
-
-		o = s.option(form.Value, 'int_addr', _('Client Address'));
+		s.cloneable = true;
+		s.sortable = true;
+		s.modaltitle = _('UPnP IGD & PCP') + ' - ' + _('Edit ACL Entry');
```
openwrt-ai: nit: modal title says `UPnP IGD & PCP` while the rest of the UI / package title is `UPnP IGD & PCP/NAT-PMP`. Either keep NAT-PMP here too, or drop it from the Map title at line 94 for consistency.

---

applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp
```diff
@@ -7,107 +9,133 @@ import { access, open, popen } from 'fs';
...
+								num: '0',
+								intaddr: arrtoip(iptoarr(m[2])),
+								intport: +m[4],
+								extport: +m[4],
+								proto: uc(m[3])
+							});
+						}
+					}
+					iptipv6.close();
+				}
```
openwrt-ai, line 78: The new ip6tables block is nested inside `if (ipt)` (line 48), so on a pure-nftables system where the `iptables` binary is missing — and `popen()` therefore returns `null` — the ip6tables chain will not be parsed either, even when `ip6tables` itself exists. The same applies to the IPv6 nftables block at line 99 being nested inside `if (nft)`, although that one is less likely to surface in practice. The two protocol families are independent and should be at the same scope:

```suggestion
				ipt.close();
			}

			const iptipv6 = popen('ip6tables --line-numbers -xnvL MINIUPNPD 2>/dev/null');
			if (iptipv6) {
				for (let line = iptipv6.read('line'); length(line); line = iptipv6.read('line')) {
					let m = match(line, /^([0-9]+) .+ ([0-9a-f:]+)  ([ut].p) dpt:([0-9]+)/);
					if (m) {
						push(rules, {
							num: '0',
							intaddr: arrtoip(iptoarr(m[2])),
							intport: +m[4],
							extport: +m[4],
							proto: uc(m[3])
						});
					}
				}
				iptipv6.close();
			}
```

(and apply the same de-nesting to the IPv6 nftables block below).

---

applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp
```diff
@@ -7,107 +9,133 @@ import { access, open, popen } from 'fs';
 import { connect } from 'ubus';
 import { cursor } from 'uci';
 
-// Establish ubus connection persistently outside of the call handler scope to
-// prevent premature GC'ing. Can be moved into `get_status` callback once
-// https://github.com/jow-/ucode/commit/a58fe4709f661b5f28e26701ea8638efccf5aeb6
-// is merged.
-const ubus = connect();
+const uci = cursor();
+const leasefilepath = uci.get('upnpd', 'settings', 'lease_file') || '/var/run/miniupnpd.leases';
```
openwrt-ai: `leasefilepath` is captured at module load time from `uci.get('upnpd', 'settings', 'lease_file')`. If the admin changes that option in `/etc/config/upnpd` and reloads the daemon but not rpcd, `luci.upnp` keeps using the stale path until rpcd restarts — a behaviour change from the pre-patch code, which re-read the cursor on every `get_status`/`delete_rule` call. Worth either reading it inside each `call:` (as before) or documenting the rpcd-restart requirement in the commit message.

---

applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env ucode
```
openwrt-ai, line 1: This file is loaded as a ucode module by rpcd via the `exec` mode (see `/usr/share/rpcd/ucode.json`), not executed as a script — the shebang has no effect and would only matter if the file gained `+x`. Other in-tree ucode rpcd modules (e.g. [`luci-app-firewall/.../firewall.ucode.json` consumers](https://github.com/openwrt/luci/blob/30933b757cc1c1d5affa36d9e1f29f2c635264ee/applications/luci-app-upnp/root/usr/share/rpcd/acl.d/luci-app-upnp.json)) don't carry one. Worth dropping to match convention:

```suggestion
// Copyright 2022 Jo-Philipp Wich <jo@mein.io>
// Licensed to the public under the Apache License 2.0.
```

---

applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp
```diff
@@ -7,107 +9,133 @@ import { access, open, popen } from 'fs';
...
-const ubus = connect();
+const uci = cursor();
+const leasefilepath = uci.get('upnpd', 'settings', 'lease_file') || '/var/run/miniupnpd.leases';
 
 const methods = {
 	get_status: {
 		call: function(req) {
-			const uci = cursor();
-
+			const ubus = connect();
```
openwrt-ai, line 16: The original placement of `const ubus = connect();` at module scope had a [comment explaining](https://github.com/openwrt/luci/blob/f699752316ac4651ae4f4f966510d352670ca87a/applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp#L10-L13) that it was a workaround for premature GC of `ubus` handles, only safe to move into the callback after [jow-/ucode@a58fe47](https://github.com/jow-/ucode/commit/a58fe4709f661b5f28e26701ea8638efccf5aeb6) ships. Confirm the rpcd/ucode in the target OpenWrt branch already contains that commit before moving the call inside `call:`, otherwise `get_status` will hit the same intermittent crash the original comment described.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
+		o.value('accept-web+high-ports', _('Accept HTTP/HTTPS + ports >= 1024'));
+		o.value('accept-web-ports', _('Accept HTTP/HTTPS ports only'));
+		o.value('accept-all-ports', _('Accept all ports'));
+		o.editable = true;
+		o.retain = true;
+
+		o = s.taboption('access_control', form.Value, 'accept_ports', _('Accept ports'));
+		o.retain = true;
+		o.validate = function(section_id, value) {
+			return value.replace(/\d+-\d+ ?|\d+ ?/g, '') == '' ? true : _('Expecting: %s').format(_('valid port or port range (port1-port2)'));
```
openwrt-ai: The validator only checks that everything in the string matches `digit{,-}digit{ ?}` shapes — it does not reject port numbers > 65535, ranges where `port1 > port2`, or port 0. e.g. `'99999'`, `'8080-21'`, `'0-65535'` all pass. Consider switching to luci's existing `datatype = 'list(portrange)'` (with space as the separator) which already enforces both constraints, instead of a custom regex. Same applies to `reject_ports` at line 218 below.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,338 @@ return view.extend({
...
 		}, o, this);
 
-		s = m.section(form.NamedSection, 'config', 'upnpd', _('Service Settings'));
+		s = m.section(form.NamedSection, 'settings', 'upnpd', _('Service Settings'));
 		s.addremove = false;
 		s.tab('setup', _('Service Setup'));
+		s.tab('access_control', _('Access Control'), '<h5>' + _('Access Control Defaults') + '</h5>' +
+			_('Set defaults for ports that all devices can map.') + ' ' +
+			_('Client-specific permissions using the access control list (ACL) can extend/override the defaults.') + ' ' +
+			_('IPv6 is currently always accepted unless disabled. (alternative text welcome)'));
```
openwrt-ai: nit: the tab description ends with `(alternative text welcome)` — and line 190 / line 346 carry similar placeholder copy (`(should be part of access control tab)`, `(alternative text welcome)`). These read like author-to-reviewer notes, not user-facing UI text. They should be resolved (or moved to the PR description) before merge.

---

openwrt-ai (commented):
## Commit checks

- `c29a7b9ea6` "luci-app-upnp: update ACL options, migrate section" — the body's last two bullets describe shell-init refactors (`is_port_or_range`/`upnpd_get_port_range`, renaming `conf_rule_add` to `upnpd_add_acl_entry`, removing a shellcheck warning), but the only file touched is `htdocs/.../upnp/upnp.js`. Those changes belong to the dependent `openwrt/packages` PR; either drop them from this commit's body or split the message to clarify what landed here.
- Several commits (`de0657afcb`, `75a1884052`, `c29a7b9ea6`, `338e2fdc73`, `30933b757c`) carry `(to merge with prior)` markers — please squash before merging so the history doesn't ship those notes.

(General note, not blocking inline: this PR's UCI schema is co-dependent on openwrt/packages#28765 — without that, the migration to `settings`/`acl_entry`/`internal_network` does not happen and the LuCI page renders the "No suitable configuration was found" banner on existing deployments. Worth surfacing this dependency in the PR title/body too, not just the commit message.)

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,326 @@ return view.extend({
...
+		o = s.taboption('access_control', form.Flag, 'check_acl', _('Check ACL'),
+			_('Check the ACL entries first; these extend/override the defaults') + '<br />' +
+			_('Sequence: 1. Reject ports, 2. ACL entries (if checked), 3. Preset ports, 4. Accept ports'));
 		o.default = '1';
+		o.editable = true;
 		o.rmempty = false;
-		o.depends('enable_upnp', '1');
+		o.onchange = function(ev, section_id, value) {
+			let acl = document.getElementById("cbi-upnpd-acl_entry");
+			value == 0 ? acl.style.display = 'none' : acl.style.display = 'block';
```
openwrt-ai, line 215: Null dereference when toggling `check_acl` from off. When the page loads with `check_acl='0'`, the `acl_entry` section is hidden via `s.disable = uci.get('upnpd', 'settings', 'check_acl') == '0'` (line ~348), which means it is never rendered and `cbi-upnpd-acl_entry` does not exist in the DOM. If the user then enables this flag, `document.getElementById("cbi-upnpd-acl_entry")` returns `null` and `acl.style.display` throws `TypeError: acl is null` — i.e. exactly the show-the-ACL path is broken.

Guard against the missing element (and re-render so the section actually appears), e.g.:

```suggestion
		o.onchange = function(ev, section_id, value) {
			let acl = document.getElementById("cbi-upnpd-acl_entry");
			if (acl)
				acl.style.display = (value == 0) ? 'none' : 'block';
		};
```

A cleaner alternative is to drop the `s.disable` + manual DOM toggle entirely and let the framework hide the section (the option's value is already persisted), avoiding the render/DOM divergence.

---

openwrt-ai (commented):
Re-reviewed after the force-push that reworked the series (the previous review's commits are no longer in history, so this pass treats the current diff as new). One new issue found inline; the earlier notes about the `(to merge with prior)` squash markers and the openwrt/packages#28765 co-dependency still apply and are not repeated here.

---

openwrt-ai (commented):
The branch was rebased since the last review (the previously reviewed commit `98ea0ba` is no longer in history), so this pass treats the current 7-commit series as new.

## Commit checks

- `cfe6b2e` "luci-app-upnp: revision and adapt to new UCI options" — the message describes an *"Enable Networks / Access Control (`internal_network`)"* section with a sub-option *"Internal network (`interface`)"*, but the code adds no `internal_network` section and no `interface` option. The network selector is implemented as option `internal_iface` in the `settings` section, and the `access_defaults`/`accept_ports`/`reject_ports`/`check_acl` options the bullet lists under that section are placed in `settings` (the Access Control tab) instead. The later commit `10aaf18` "service-wide access control settings" notes it "implements variant B", which looks like it superseded the per-network design this body still describes. Is the `cfe6b2e` body just stale relative to the variant-B implementation? If so, please reword it to match what actually lands (`internal_iface` in `settings`, service-wide access-control options) so the history isn't misleading.

Not re-flagged (already disclosed in the PR description): the `(to merge with prior)` markers on 5 of 7 commits (the body states the italic commits are to be squashed after review), and the co-dependency on openwrt/packages#28765 for the `config`→`settings`/`acl_entry` UCI migration. I confirmed against the current `openwrt/packages` master that `internal_iface` and `lease_file` are real backend options, so the renamed schema is backend-grounded once that PR merges.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -66,185 +68,326 @@ return view.extend({
...
+		o.retain = true;
+		o.modalonly = false;
 
-		o = s.option(form.Value, 'int_ports', _('Client Port'));
+		o = s.option(form.Value, 'int_addr', _('IP address'),
+			_('Enter device\'s IPv4 address, or an address with netmask'));
+		o.datatype = 'ip4addr';
+		o.rmempty = false;
+		o.retain = true;
+		o.modalonly = true;
```
openwrt-ai, line 357: `int_addr` is now declared twice in this `acl_entry` GridSection — as an editable grid column (line 345) and again as a `modalonly` field here — both bound to the same UCI key. Two `form.Value` options with the same name in one section produce two widgets sharing the DOM id `cbi-upnpd-<sid>-int_addr`: the edit modal renders the field twice, `getElementById`/validation resolve only the first, and both write back the same value on save. The grid-column option already covers the field, so add the help text to it and drop the duplicate:

```suggestion
		o = s.option(form.Value, 'int_addr', _('IP address'),
			_('Enter device\'s IPv4 address, or an address with netmask'));
		o.datatype = 'ip4addr';
		o.rmempty = false;
		o.editable = true;
		o.retain = true;
```

---

openwrt-ai (commented):
The 7-commit series was rebased since my last review (`10aaf18` is no longer in history), so this pass re-examines the current head `5550b4a`. One new inline issue below.

Still applicable from earlier passes, not re-flagged here as they are already disclosed in the PR description: the `(to merge with prior)` squash markers on the italic commits, and the co-dependency on openwrt/packages#28765 for the `settings` / `acl_entry` / `int_addr` UCI schema (I re-confirmed against `openwrt/packages` master that `int_addr` and `internal_iface` are real miniupnpd backend options). The unresolved null-dereference when toggling `check_acl` from off (existing inline thread) also still stands in the current code.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
```diff
@@ -70,14 +70,19 @@ return baseclass.extend({
...
 					'class': 'btn cbi-button-remove',
-					'click': L.bind(handleDelRule, this, rule.num)
-				}, [ _('Delete') ])
+					'click': L.bind(handleDelRule, this, rule.num),
+					'title': _('Delete')
+				}, [_('Delete')]) : E('button', {
+					'class': 'btn cbi-button-remove',
+					'title': _('Not yet deletable; UPnP IGDv2 IPv6 port maps expire sooner, have no description'),
+					'style': 'cursor: not-allowed'
+				}, [_('Delete')])
```
openwrt-ai, line 81: The IPv6 variant of the button is not actually disabled — it only carries `cursor: not-allowed`. It still renders in the normal (non-greyed) `cbi-button-remove` style, is keyboard-focusable and tab-reachable, and pressing it silently does nothing, so it looks like a working Delete that fails without feedback. In-tree buttons use the `disabled` attribute for this, e.g. [`images.js:490`](``https://github.com/openwrt/luci/blob/c264a4aabea9b6bed2b10b9ba433353afb3c350c/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/images.js#L490).``

```suggestion
				}, [_('Delete')]) : E('button', {
					'class': 'btn cbi-button-remove',
					'title': _('Not yet deletable; UPnP IGDv2 IPv6 port maps expire sooner, have no description'),
					'disabled': ''
				}, [_('Delete')])
```

The same applies to the duplicated block in `htdocs/luci-static/resources/view/upnp/upnp.js` at lines 132-136.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -59,185 +61,326 @@ return view.extend({
...
+		o.value('accept-all-ports', _('Accept all ports'));
+
+		o = s.taboption('access_control', form.Value, 'accept_ports', _('Accept extra ports'));
+		o.datatype = 'list(portrange)';
+
+		o = s.taboption('access_control', form.Value, 'reject_ports', _('Reject ports'),
+			_('Reject unsafe/insecure/risky FTP/Telnet/DCE/NetBIOS/SMB/RDP ports by default; overrides other settings; use %s for none').format('<code>0</code>'));
+		o.datatype = 'list(portrange)';
+		o.placeholder = '21 23 135 137-139 445 3389';
+		o.modalonly = true;
```
openwrt-ai: `modalonly` is a no-op here: `s` is a `form.NamedSection`, and the property is only honoured by grid sections — see [form.js:1569-1570](https://github.com/openwrt/luci/blob/c264a4aabea9b6bed2b10b9ba433353afb3c350c/modules/luci-base/htdocs/luci-static/resources/form.js#L1569-L1570) ("This property has no effect on options that are not children of grid section elements") and the only consumer at [form.js:4060](https://github.com/openwrt/luci/blob/c264a4aabea9b6bed2b10b9ba433353afb3c350c/modules/luci-base/htdocs/luci-static/resources/form.js#L4060) inside `CBIGridSection.renderChildren`.

It is a leftover from the `internal_network` GridSection that this same commit removes, where it did have an effect. `Reject ports` therefore renders in the Access Control tab like every other option, which reads like the intended result — so the line is just dead:

```suggestion
```

---

applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp
```diff
@@ -7,107 +9,133 @@ import { access, open, popen } from 'fs';
...
+			// Workaround daemon bug with iptables >= 1.8.8 by parsing MINIUPNPD-POSTROUTING
+			// See https://github.com/miniupnp/miniupnp/issues/837
+			//const ipt = popen('iptables --line-numbers -t nat -xnvL MINIUPNPD 2>/dev/null');
+			const ipt = popen('iptables --line-numbers -t nat -xnvL MINIUPNPD-POSTROUTING 2>/dev/null');
 			if (ipt) {
 				for (let line = ipt.read('line'); length(line); line = ipt.read('line')) {
-					let m = match(line, /^([0-9]+)\s+([a-z]+).+dpt:([0-9]+) to:(\S+):([0-9]+)/);
-
+					//let m = match(line, /^([0-9]+) .+ ([ut].p) dpt:([0-9]+) to:(\S+):([0-9]+)/);
+					let m = match(line, /^([0-9]+) .+ \* +([0-9.]+) .+ ([ut].p) spt:([0-9]+) .+ ports: ([0-9]+)/);
```
openwrt-ai, line 53: The superseded `popen()` call and `match()` pattern are left in as commented-out code. The two explanatory comments above already record why the chain changed, so the dead lines add nothing but will rot as soon as the daemon bug is fixed upstream and someone has to decide whether they are still the correct fallback.

```suggestion
			// Workaround daemon bug with iptables >= 1.8.8 by parsing MINIUPNPD-POSTROUTING
			// See https://github.com/miniupnp/miniupnp/issues/837
			const ipt = popen('iptables --line-numbers -t nat -xnvL MINIUPNPD-POSTROUTING 2>/dev/null');
			if (ipt) {
				for (let line = ipt.read('line'); length(line); line = ipt.read('line')) {
					let m = match(line, /^([0-9]+) .+ \* +([0-9.]+) .+ ([ut].p) spt:([0-9]+) .+ ports: ([0-9]+)/);
```

Same for the three commented-out `o.depends('enable_protocols', …)` pairs in `view/upnp/upnp.js` at lines 283-284, 289-290 and 302-303.

---

applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp
```diff
@@ -7,107 +9,133 @@ import { access, open, popen } from 'fs';
...
+				if (nftipv6) {
+					for (let line = nftipv6.read('line'), num = 1; length(line); line = nftipv6.read('line')) {
+						let m = match(line, /^\t\tiif ".+" th dport ([0-9]+) @nh,192,128 0x([0-9a-f]+) @nh,48,8 (0x6|0x11) accept/);
+						if (m) {
+							push(rules, {
+								num: '0',
+								intaddr: arrtoip(iptoarr(join(':', [substr(m[2], 0, 4),
+									substr(m[2], 4, 4), substr(m[2], 8, 4), substr(m[2], 12, 4), substr(m[2], 16, 4),
+									substr(m[2], 20, 4), substr(m[2], 24, 4), substr(m[2], 28, 4)
+								]))),
```
openwrt-ai, line 111: The fixed offsets `substr(m[2], 0, 4)` … `substr(m[2], 28, 4)` assume `m[2]` is always exactly 32 hex characters, but the regex captures `0x([0-9a-f]+)` with no length constraint. `nft` prints raw-payload values as a hex *number*, and mpz-based hex formatting strips leading zeros — so an address whose first nibble is zero (e.g. `0064:ff9b::/96` NAT64, or any `0…`-prefixed prefix) yields a 31-or-fewer character capture and every slice shifts, producing a wrong `intaddr` (or `null` out of `iptoarr()`, i.e. a blank column) rather than an error.

Have you checked what `nft --handle list chain inet fw4 upnp_forward` prints for such an address? If leading zeros can indeed be dropped, left-padding `m[2]` to 32 chars before slicing (or matching `0x([0-9a-f]{1,32}) `) would make this robust.

Separately, `num++` on line 113 is dead — the loop always pushes the literal `num: '0'`, so the counter is never read. Dropping it (and the `num = 1` initialiser on line 100) would make it obvious that IPv6 entries deliberately carry no delete token.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
```diff
@@ -59,185 +61,326 @@ return view.extend({
...
-			_('Allow adding port maps for requesting IP addresses only'));
+		o.filter = function(section_id, value) {
+			return (value == 'wan' || value == 'wan6') ? '' : value;
+		};
+		o.write = function(section_id, formvalue) {
+			uci.set('upnpd', section_id, 'internal_iface', Array.isArray(formvalue) ? formvalue.join(' ') : formvalue);
+		};
+
+		o = s.taboption('setup', form.ListValue, 'upnp_igd_compat', _('UPnP IGD compatibility'),
+			_('Set compatibility mode (act as device) to workaround IGDv2-incompatible clients; %s are known to only work with %s (or) <br />Emulate/report a specific/different device to workaround/support/handle/bypass/assist/mitigate... (Alternative text welcome)').format('Sony PS, Activision CoD…', 'IGDv1'));
```
openwrt-ai, line 178: Author-to-reviewer notes are still inside translatable, user-facing strings: this description ends with `to workaround/support/handle/bypass/assist/mitigate... (Alternative text welcome)`, and line 326 ships `(To do: should be part of access control tab)` in the ACL section description. Both go through `_()`, so they land in the `.po` catalogues and get handed to translators and end users. These need resolving (or moving into the PR description / a `// To do:` source comment) before merge.

---

applications/luci-app-upnp/Makefile
```diff
@@ -6,7 +6,7 @@
 
 include $(TOPDIR)/rules.mk
 
-LUCI_TITLE:=UPnP IGD & PCP/NAT-PMP configuration module | Universal Plug and Play
+LUCI_TITLE:=LuCI support for UPnP IGD & PCP/NAT-PMP service | originally: Universal Plug and Play
```
openwrt-ai, line 9: nit: `LUCI_TITLE` is the package title shown by the package manager, and `| originally: Universal Plug and Play` reads like a changelog note rather than a title. No other `applications/*/Makefile` in the tree carries a `|` separator with historical wording. Suggest keeping just the current name:

```suggestion
LUCI_TITLE:=LuCI support for UPnP IGD & PCP/NAT-PMP service
```

If the old wording is kept purely so users searching for "Universal Plug and Play" still find the package, `LUCI_DESCRIPTION` is the better place for it.

---

openwrt-ai (commented):
The branch was rebased again since my last review (`5550b4a` is no longer in history), so this pass treats the current 7-commit series at `c264a4a` as new.

## Commit checks

- `74e90a0` "luci-app-upnp: revision and adapt to new UCI options" — the body tabulates the renamed/added UCI options fairly exhaustively, but two changes in the same commit are not mentioned:
  - the UCI rename `port` → `http_port` (SOAP/HTTP port). Every other rename in this commit is listed (`stun_host`, `external_ip`, `download_kbps`/`upload_kbps`, `friendly_name`, "STUN port: Removed"), so this one reads like an oversight. Worth adding, especially since the current backend still uses `port` ([`miniupnpd.init:75`](https://github.com/openwrt/packages/blob/7b12d68dd352028f8d3c5f09ba159322de74aaeb/net/miniupnpd/files/miniupnpd.init#L75)) and the rename only exists in the dependent packages PR.
  - the `LUCI_TITLE` change in `applications/luci-app-upnp/Makefile` — unrelated to "adapt to new UCI options"; see the inline note.

Backend-coupling check (per the repo's review rules): against `openwrt/packages` master, `access_defaults`, `accept_ports`, `reject_ports`, `check_acl`, `enable_protocols`, `allow_cgnat`, `allow_third_party_mapping`, `upnp_igd_compat`, `download_kbps`, `upload_kbps`, `friendly_name`, `descr_filter`, `http_port` and the `settings` / `acl_entry` section names have no consumer in `net/miniupnpd`, which still reads the `config` section. That is expected given openwrt/packages#28765 and is already disclosed in the PR body, so not flagged as a finding — noting it only to confirm the dependency really is hard-blocking. The `lease_file` fallback `/var/run/miniupnpd.leases` in `luci.upnp` does match [`upnpd.config:13`](https://github.com/openwrt/packages/blob/7b12d68dd352028f8d3c5f09ba159322de74aaeb/net/miniupnpd/files/upnpd.config#L13).

Not re-flagged, since the existing (still-anchored) threads on those lines remain accurate: the `check_acl` `onchange` null-dereference at `upnp.js:207` (also acknowledged by the `// To do:` at line 332), the ip6tables/nft-IPv6 blocks nested inside `if (ipt)` / `if (nft)`, the shebang, the ubus-connect move, and the unencoded `&` in the package-manager query string at `upnp.js:96`. The duplicate `int_addr` declaration from the last pass is now correctly split via `modalonly = false` / `true` — that one looks resolved.

---

applications/luci-app-upnp/htdocs/luci-static/resources/view/status/include/80_upnp.js
```diff
@@ -70,14 +70,19 @@ return baseclass.extend({
...
 			];
 		});
 
-		cbi_update_table(table, rows, E('em', _('There are no active port maps.')));
+		cbi_update_table(table, rows, E('em', _('There are no active IPv4/IPv6 port maps')));
```
timur-davletshin, line 85: Tiniest suggestion: `No active IPv4/IPv6 port maps` will be enough since both `DHCP leases` and `Associated Stations` use similar style messages.

applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp
```diff
@@ -7,107 +9,136 @@ import { access, open, popen } from 'fs';
...
 			if (leasefile) {
 				for (let line = leasefile.read('line'); length(line); line = leasefile.read('line')) {
-					const record = split(line, ':', 6);
-
-					if (length(record) == 6) {
+					const record = split(line, ':', 7);
+					if (length(record) == 6 || length(record) == 7) {
+						// Daemon version >2.3.11 changes lease file format
+						let recshift = 0;
+						if (length(record) == 7) recshift = 1;
```
openwrt-ai, line 26: Field count can't discriminate the two lease formats: ucode's `split` with `limit` splits at most `limit - 1` times and puts the remainder in the last element ([lib.c:1840-1842](https://github.com/jow-/ucode/blob/64cf18aa55c67e73b9acd0262a44d50b41774ed8/lib.c#L1840-L1842)), so raising 6 → 7 makes an *old*-format line whose description contains a colon (a client-supplied string, e.g. the classic `Skype UDP at 192.168.1.5:23423 (2537)`) also yield 7 elements and take the `recshift = 1` path — `intaddr`, `intport` and `expires` then all read one field too far. The previous `limit` of 6 was what kept colons inside the description. Discriminate on content instead: `record[3]` is the internal port (all digits) in the old format and the internal address in the new one.

```suggestion
					let record = split(line, ':', 6);
					// Daemon version >2.3.11 inserts a remote host field before the internal address
					const recshift = (length(record) == 6 && !match(record[3], /^[0-9]+$/)) ? 1 : 0;
					if (recshift) record = split(line, ':', 7);
					if (length(record) == 6 + recshift) {
```


---

openwrt-ai (commented):
## Commit checks

- `569489cc` "revision and adapt to new UCI options" — the body describes an `internal_network` section with an `interface` sub-option; the code adds `internal_iface` in `settings`, and `access_defaults`/`accept_ports`/`reject_ports`/`check_acl` land in `settings` via `ec863ba` instead. Reword to match what the commit actually adds.
- `569489cc` — two changes in the diff are absent from the message: the `port` → `http_port` UCI rename (every other rename is listed) and the `LUCI_TITLE` edit in `applications/luci-app-upnp/Makefile`.

---

PR #8418, opened 2026-03-13
luci-app-rustdesk-server: add keep.d file

luminoso:

Add a keep.d so that user doesn't loose the rustdesk private key generated and being used by the service when upgrading OpenWRT.

I've included the binaries that the user has to put in root, until I have some time to create the packages and then make this app depend on those packages.

### DIFF
--- a/applications/luci-app-rustdesk-server/Makefile
+++ b/applications/luci-app-rustdesk-server/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
-PKG_VERSION:=20250610
-PKG_RELEASE:=4
+PKG_VERSION:=20260313
+PKG_RELEASE:=1
 PKG_NAME:=luci-app-rustdesk-server
 PKG_MAINTAINER:=Guilherme Cardoso <luminoso@gmail.com>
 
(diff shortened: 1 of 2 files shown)


systemcrash:
Are you sure the binaries need to be kept? They're automatically installed if owut or asu is used.

luminoso:
@systemcrash I would say so, since the user must (for now) upload the binaries. 

I'm planning to create the binaries packages, but for now the user has to manually build and upload them.

Do you suggest to not keep the binaries, just the config? I'm also fine with that. What's critical is the private key

systemcrash:
Ah yes. That. It'd be awkward to mandate a new binary install, altho the front-end checks for their presence. Let's start with the metadata first, then get the package in.

luminoso:
@systemcrash sorry, just to be celar. So do we keep it like this or not?

My plan is:

1. Keep like it is in this PR
2. Create the packages

Then:
3.1 Make this luci-app depend on the packages
3.2 Remove the binaries from keep.d

systemcrash:
OK. We shall proceed :)

meiser79:
Hi, may I kindly ask to backport this PR to openwrt-25.12?

luminoso:
Yes. PR to openwrt-25.12: #8530

PR #8419, opened 2026-03-13
luci-app-usteer: Add parameters specifically for usteer-ng

Ramon00:
usteer-ng is the forked version of usteer with more options and improvements. 
This commit allows setting the additional options in luci-app-usteer. 

Next to that a small bug is fixed.

### DIFF
--- a/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
+++ b/applications/luci-app-usteer/htdocs/luci-static/resources/view/usteer/usteer.js
@@ -10,7 +10,7 @@
 'require network';
 'require tools.widgets as widgets';
 
-let Hosts, Remotehosts, Remoteinfo, Localinfo, Clients, WifiNetworks;
+let Hosts, Remotehosts, Remoteinfo, Localinfo, Clients, WifiNetworks, Initscript;
 
 const dns_cache = [];
 const hostapdClientData = [];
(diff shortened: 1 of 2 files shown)


systemcrash:
The angle you're going for here is that the new settings work or are benign. How about detecting which agent is installed via a fs.stat or fs.exec command -v and revealing settings based on that result? It might prevent users lodging bug reports 'this setting doesn't work'....

Ramon00:
I can try to see if fs.stat works

domingues:
usteer-ng has new default values, probably they should also be reflected here. https://github.com/NilsRo/usteer-ng/pull/3/changes/b69ac48b18863983b0ff0ed9f021c635a930dbe3

Even better is to push for the original project to wake up to life... https://github.com/openwrt/usteer/pull/10

Ramon00:
> I can try to see if fs.stat works

Ok I added an elegant detection



> usteer-ng has new default values

Its only 1 parameter, also i cannot really detect what the default is. The only thing I could do is check if one of the other parameters are included and then change the default based on that. But to be honest, I dont think it is that important, especially because usteer-ng is not officially supported. @systemcrash what do you think, also switch on that, or just leave it slightly less complex?

systemcrash:
Well, you can elide a default for the parameter, and retain whatever value is provided, especially if it's a flag. 

We can put this in and see how it fares. master only?

Ramon00:
Ok i added a switch for that one parameter. Im fine either way, this or the previous version.

Ramon00:
optimized a bit

systemcrash:
Merged. Thanks @Ramon00

PR #8420, opened 2026-03-15
luci-app-nut: fix sync between LuCI and init.d parameters

usamanassir:

### DIFF
--- a/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js
+++ b/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js
@@ -87,10 +87,10 @@ return view.extend({
 		o = s.option(form.Value, 'fsdmsg', _('Forced shutdown message'));
 		o.optional = true;
 
-		o = s.option(form.Value, 'comokmsg', _('Communications restored message'));
+		o = s.option(form.Value, 'commokmsg', _('Communications restored message'));
 		o.optional = true;
 
-		o = s.option(form.Value, 'combadmsg', _('Communications lost message'));
+		o = s.option(form.Value, 'commbadmsg', _('Communications lost message'));
 		o.optional = true;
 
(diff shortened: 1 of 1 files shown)


systemcrash:
@danielfdickinson does this agree?

danielfdickinson:
I think part of the problem may be incorrect values in the <https://github.com/openwrt/packages> version of nut. I would prefer to see a PR ensuring packages.git is correct before updating LuCI to match it. I've been rather busy, but if I am able to find some time, I will such a PR myself.

usamanassir:
@danielfdickinson, I simply matched the names to https://github.com/openwrt/packages/blob/master/net/nut/files/nut-monitor.init.
If you think it is better, I can change nut-monitor.init instead. important thing is to get both of them using that same names.

danielfdickinson:
@usamanassir I have created https://github.com/openwrt/packages/pull/28875 which makes sure the names are correct from NUTs point of view. This PR should match that PR.

I have also created https://github.com/openwrt/luci/pull/8434 to keep that part in sync.

usamanassir:
@danielfdickinson, wont (https://github.com/openwrt/packages/pull/28875)https://github.com/openwrt/packages/pull/28875 break old nut instlattions. Meaning, once a working system gets the new version, it will stop working.

danielfdickinson:
> @danielfdickinson, wont ([openwrt/packages#28875](https://github.com/openwrt/packages/pull/28875) break old nut instlattions. Meaning, once a working system gets the new version, it will stop working.

Care was taken to prevent breaking old installations. Have you actually tried? If you have an installation that breaks, please provide configuration and logs of the broken configuration. Also please provide those comments on https://github.com/openwrt/packages/pull/28875

danielfdickinson:
@usamanassir I believe https://github.com/openwrt/luci/pull/8533 and the related https://github.com/openwrt/packages/pull/28875 resolve the parameters names issue.

usamanassir:
Agree. Closing the PR.

PR #8428, opened 2026-03-16
luci-base: fix getBuiltinEthernetPorts crashing due to missing realpath import

MI2-2:
The getBuiltinEthernetPorts x86_64 workaround block uses realpath() to resolve PCI subsystem symlinks, but realpath was not included in the 'fs' module import statement. This caused the ubus method to crash with "Unknown error" on hardware where NICs have PCI subsystem symlinks at different sysfs depths (e.g. onboard NIC vs PCIe card NIC).

Additionally, the original readlink() call only returns the raw symlink target string without resolving it to an absolute path. On hardware where different NICs have subsystem symlinks of different depths:

  eth0: /sys/class/net/eth0/device/subsystem -> ../../../bus/pci
  eth1: /sys/class/net/eth1/device/subsystem -> ../../../../bus/pci
  eth2: /sys/class/net/eth2/device/subsystem -> ../../../../bus/pci

readlink() returns different strings ("../../../bus/pci" vs "../../../../bus/pci") even though both resolve to /sys/bus/pci, causing eth1 and eth2 to be incorrectly excluded from the result.

Fix both issues by:
1. Adding realpath to the 'fs' module import statement
2. Replacing readlink() with realpath() in the workaround block so that symlinks of varying depths are correctly normalized before comparison

スクリーンショット 2026-03-16 200219

- [x] N/A (luci-base uses auto-generated version from git hash)

* Mention: @jow for feedback

* Tested on: x86_64, OpenWrt 24.10 (kernel 6.6.71), Firefox

* Description:
  Fix two bugs in getBuiltinEthernetPorts that prevent PCI Ethernet
  ports from being detected on x86_64 hardware where onboard NIC and
  PCIe card NICs have sysfs subsystem symlinks at different depths.

  Bug 1: realpath() was called in the x86_64 workaround block but was
  not included in the 'fs' module import, causing ubus call luci
  getBuiltinEthernetPorts to crash with "Unknown error".

  Bug 2: The original readlink() returns raw symlink strings that
  differ by depth ("../../../bus/pci" vs "../../../../bus/pci"),
  causing ports to be incorrectly filtered out even though all paths
  resolve to /sys/bus/pci. Replacing with realpath() normalizes all
  paths before comparison.

Replaces #8425

### DIFF
--- a/modules/luci-base/root/usr/share/rpcd/ucode/luci
+++ b/modules/luci-base/root/usr/share/rpcd/ucode/luci
@@ -3,7 +3,7 @@
 
 'use strict';
 
-import { stdin, access, dirname, basename, open, popen, glob, lsdir, readfile, readlink, error } from 'fs';
+import { stdin, access, dirname, basename, open, popen, glob, lsdir, readfile, readlink, realpath, error } from 'fs';
 import { cursor } from 'uci';
 
 import { init_list, init_index, init_enabled, init_action, conntrack_list, process_list } from 'luci.sys';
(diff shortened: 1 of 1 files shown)


systemcrash:
Closed by ab72c4457a18334a99a175da6eb6a1a53c63af7e - please don't open more

It's already picked to 25.12 via 6efc20a91376ad2ef7c12f784d51761431771ffa

PR #8430, opened 2026-03-16
plugin architecture

systemcrash:

Global enable for all plugins and enable for specific plugin classes
Screenshot 2026-03-16 at 21 19 42

Screenshot 2026-03-16 at 21 20 17

Screenshot 2026-03-16 at 21 20 06

This introduces a plugin system for the UI - primarily to configure the behaviour of ucode plugins - and some example plugins. 

This should cover the user-case for @josteink in #8161. 

Some additional hook points are necessary to take care of 2FA type plugins for @Tokisaki-Galaxy and #8280 and #8281, I think.

### DIFF
--- a/modules/luci-base/ucode/http.uc
+++ b/modules/luci-base/ucode/http.uc
@@ -13,6 +13,12 @@ import {
 	stdin, stdout, mkstemp
 } from 'fs';
 
+import {
+	openlog, syslog, closelog, LOG_NOTICE, LOG_LOCAL0
+} from 'log';
+
+import { run_plugins } from 'luciplugins';
+
 // luci.http module scope
 export let HTTP_MAX_CONTENT = 1024*100;		// 100 kB maximum content size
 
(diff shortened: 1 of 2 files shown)


modules/luci-mod-system/htdocs/luci-static/resources/view/system/plugins.js
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		for (const group of new Set([...groups].sort())) {
+
+			s = m.section(form.GridSection, group, groups[group]);
+
+			s.sectiontitle = function(section_id) {
+				const plugin = plugins[section_id];
+
+				return plugin.title;
```
Tokisaki-Galaxy: If a plugin is uninstalled, its UCI configuration remains. The `GridSection` will attempt to render it, and s.sectiontitle will throw erro when accessing plugin.title of an undefined plugin, causing the UI to crash. It should be changed to `return plugin ? plugin.title : section_id;`?

Tokisaki-Galaxy:
Hi systemcrash,
Amazing design! Iam happy to help add the auth mechanism hooks once the UI/UX part is settled.
One question: since UCI configs aren't removed when a plugin file is deleted, how should we handle "ghost" configurations? Currently, if a plugin file is missing but the config remains, it might lead to orphaned entries or even UI rendering errors (like in s.sectiontitle).

systemcrash:
If the js plugin is gone - the UI portion doesn't show. You can take this PR for a test-drive to see how it behaves. So remaining config isn't a problem to worry about. Users might notice it when they look in the config files tho.

stangri:
Thanks for this — the architecture itself looks interesting and the HTTP header example is a solid use case.

One naming concern worth raising: LuCI already has a well-established term for optional, independently installable UI components — "applications" (`luci-app-*`). In many other ecosystems (browsers, VS Code, WordPress), "plugin" is the common word for exactly that same concept, which already causes occasional confusion when people come to OpenWrt from those environments.

Introducing "plugins" as a new term here — even for a technically distinct thing — risks compounding that confusion rather than resolving it. Would it make sense to stay within the existing "applications" vocabulary and perhaps qualify it (e.g. "system applications", "core applications") rather than adding a second term that overlaps with what users already call `luci-app-*` packages?

systemcrash:
Thanks for raising these concerns @stangri. I see what you mean. 

The luci-app-* paradigm is a UI layer for an app dependency found in another repo, as per standard structure. It's possible to write an app which contains plugins, or is exclusively a plugin, since path structure determines the plugin, as long as it's ucode and fulfils plugin properties.

Having a distinct grouping differentiates them: they're (can be) independent of apps from other repos and are intended to augment UI behaviour. Their initial use cases being HTTP headers and 2FA stuff. Other hook points are possible, e.g. in the luci script which determines features (a la dnsmasq).

systemcrash:
Now in master.

PR #8452, opened 2026-03-21
luci-app-attendedsysupgrade: format error details

efahl:
On certain rare errors, the ASU server returns a JSON formatted block of details.  The app was just displaying this object directly, resulting in displaying a useless 'Object [object]'.  If the error details are not a string, then stringify it so we can make sense of the error.


systemcrash:
Merged. Thanks @efahl

PR #8454, opened 2026-03-21
luci-mod-network: show option also for sae

Ramon00:
For sae the cipher overrule is not exposed.
There can be cases where this is needed.
See e.g. https://github.com/openwrt/openwrt/issues/22200


Ramon00:
To strengthen my argumentation: Makes no sense that e.g. gcmp-128 cannot be chosen for SAE (where it is part of the standard), but can be chosen for wpa2 (where it is not part of the standard).

castiel652:
Now I have a question.
Will this cause issue with 
https://github.com/openwrt/openwrt/commit/254dd7d7d9c88e1b05dad44615213de9a0a4e462
https://github.com/openwrt/openwrt/commit/b48925fd9cf6a4891ce0c2085c18ec0af58b540c

Ramon00:
If you force a setting which is not supported then that will give issues, but that is not unique to sae. E. G. you force WPA to gcmp256 then that will very likely not work(for everybody), but there may be cases where this is needed...

Ramon00:
@systemcrash wanne merge this, or shall I close this?

castiel652:
Upon further research I think this PR is safe to merge. 👍

systemcrash:
Closed by a4f0c52fbf5db5a9a51c81866d7f3ac688ec95f4. 

Thanks @Ramon00

PR #8463, opened 2026-03-25
package-manager: migrate to apk-tools JSON API

Konstantin-Glukhov:
Replace legacy opkg status parsing with native apk-tools JSON queries. This modernizes the backend calls and improves dependency resolution.

- Implement 'apk query --format json' for package data retrieval.
- Add robust regex for versioned dependencies (e.g., name>=version).
- Update version comparison to handle APK date-based revisions.
- Fix label-input associations to resolve accessibility warnings.
- Retain deprecated opkg fallback logic for LuCI master.

NOTE: The change rational is discussed in https://github.com/openwrt/openwrt/pull/22512

- [x] Tested on: (ARMv7, OpenWrt 25.12.1, chrome/firefox) :white_check_mark:

### DIFF
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -152,7 +152,7 @@ function parseList(s, dest)
 
 				switch (installed) {
 				case 'installed':
-					pkg.installed = true;
+					pkg.status = ['installed'];
 					break;
 				}
 				break;
@@ -199,6 +199,59 @@ function parseList(s, dest)
 			});
 		}
 	}
+}
+
+function parseApkQueryJson(s, dest) {
+	// Parse the raw JSON text string into an array
+	let rawData;
+	try {
+		rawData = JSON.parse(s);
+	} catch (e) {
+		console.error("Failed to parse APK JSON:", e);
+		return;
+	}
+	// Ensure rawData is actually an array before iterating
+	if (!Array.isArray(rawData))
+		return;
+	// Ensure our storage objects exist on the destination
+	dest.pkgs = dest.pkgs || {};
+	dest.providers = dest.providers || {};
+
+	// Single pass through each item in the new JSON array
+	for (const item of rawData) {
+
+		// Rename 'file-size' to 'size' while pulling out 'name'
+		const { name, 'file-size': size, ...attributes } = item;
+
+		// Construct the package object with the expected 'size' key
+		const pkg = { name, size, ...attributes };
+
+		// Add/Update the package in the main map
+		dest.pkgs[name] = pkg;
+
+		// Determine all provided names (a package always provides itself)
+		const provides = [name, ...(Array.isArray(pkg.provides) ? pkg.provides : [])];
+		for (const p of provides) {
+			dest.providers[p] = dest.providers[p] || [];
+			if (!dest.providers[p].includes(pkg))
+				dest.providers[p].push(pkg);
+		}
+	}
+}
+
+function parsePackageData(s, dest) {
+	if (!s) return;
+
+	// Check if the input is JSON (starts with [)
+	if (s.trim().charAt(0) === '[') {
+		parseApkQueryJson(s, dest);
+	} else {
+		parseList(s, dest);
+	}
+}
+
+function isPkgInstalled(pkg) {
+	return pkg && Array.isArray(pkg.status) && pkg.status.includes('installed');
 }
 
 function display(pattern)
@@ -249,7 +302,7 @@ function display(pattern)
 			const avail = packages.available.pkgs[name];
 			const inst  = packages.installed.pkgs[name];
 
-			if (!inst || !inst.installed)
+			if (!isPkgInstalled(inst))
 				continue;
 
 			if (!avail || compareVersion(avail.version, pkg.version) <= 0)
@@ -267,7 +320,7 @@ function display(pattern)
 			}, _('Upgrade…'));
 		}
 		else if (currentDisplayMode === 'installed') {
-			if (!pkg.installed)
+			if (!isPkgInstalled(pkg))
 				continue;
 
 			ver = truncateVersion(pkg.version || '-');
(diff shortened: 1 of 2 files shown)


systemcrash:
Two comments: we need to maintain support for opkg until 24.10 goes out of support, in case anything gets backported to 24.10. It looks like opkg gets removed here.

I'd like an OK from at least @efahl also.

efahl:
I'm still digesting it before review.  

And although my conceptual opinion is that this is a good cleanup, getting rid of a bunch of old junk, I think we need some input from @Ansuel as he's the one who migrated from opkg-only to dual package manager support, and surely has some opinions on this.

Konstantin-Glukhov:
> I'm still digesting it before review.
> 
> And although my conceptual opinion is that this is a good cleanup, getting rid of a bunch of old junk, I think we need some input from @Ansuel as he's the one who migrated from opkg-only to dual package manager support, and surely has some opinions on this.

I was not aware that 24.10 needs to be maintained in the master branch. I will merge the prior stuff back.

efahl:
> I was not aware that 24.10 needs to be maintained in the master branch. I will merge the prior stuff back.

I'd wait a bit on that, it may be the case that this is fine for master and 25.12, and that we can just maintain 24.10 package manager independently.  

(I personally would like that, but it's not my call. 🙂)

Konstantin-Glukhov:
> > I was not aware that 24.10 needs to be maintained in the master branch. I will merge the prior stuff back.
> 
> I'd wait a bit on that, it may be the case that this is fine for master and 25.12, and that we can just maintain 24.10 package manager independently.
> 
> (I personally would like that, but it's not my call. 🙂)

OK, if it is decided to have the dual functionality (opk+apk), then I have the change is tested and staged for a commit.

systemcrash:
Retain opkg for now, please.

XCas13:
@Konstantin-Glukhov Please also consider changing `install) action="add" ;;` to `install) action="add --allow-untrusted" ;;` for APK in package-manager-call. This will make the "Download Package" button useful again 🙂 as I described it in #8482. Unless the package developers/maintainers have any objections to this. Don't want to create another PR with changes that will conflict to yours 🙂

systemcrash:
> to `install) action="add --allow-untrusted" ;;` for APK in package-manager-call

If anything, this should probably have a checkbox for this flag, so user's don't blindly install harmful code.

Ansuel:
I'm against allow untrusted... also the APK are now also signed so official .apk should also work from the upload page.

systemcrash:
Fair. I figure the problem is apk from external repos (I've been contemplating a 'repo' file format that one can drag+drop onto the repokeys page, so adding repos would be easy) or from their own build environment.

XCas13:
There is no point to use Upload Package for trusted packages, because it's simply possible to install them let's call it "official way". The "Upload package" button is really used for uploading untrusted packages. Otherwise there is no point to keep it in Luci.  It's fine to have a check box to allow untrusted packages and/or some warning, but users should have a choice.

Ansuel:
@XCas13 what about situation when the device doesn't have connection ?

XCas13:
> @XCas13 what about situation when the device doesn't have connection ?

Agree, this case is also real, but imho most often users will be trying to install untrusted packages. I'm not asking about creating a security breach, but about better user experience. You can connect by ssh and run `apk add --allow-untrusted` anyway. However, it's annoying that you have to do this, instead of having nice and quick way to install packages through Luci.

Konstantin-Glukhov:
I pushed the change that leave opkg functionality alone. If you decide to add '--allow-untrusted' option, I would happily do this. Just let me know.

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -201,6 +201,48 @@ function parseList(s, dest)
 	}
 }
 
+function parseApkQueryJson(s, dest) {
+	// Parse the raw JSON text string into an array
+	const rawData = JSON.parse(s);
```
systemcrash: This needs wrapping in a try/catch.

Konstantin-Glukhov: > This needs wrapping in a try/catch.

Addressed.

systemcrash:
@efahl this looks ready - does it line up with your understanding of APK JSON output?

efahl:
The `apk` query output calls the size of the package `file-size`, where `opkg` lists it as `Size:`, which gets mapped to `pkg.size` in the js here.  I don't see anything that is mapping `file-size` -> `size` anywhere.

Konstantin-Glukhov:
> The `apk` query output calls the size of the package `file-size`, where `opkg` lists it as `Size:`, which gets mapped to `pkg.size` in the js here. I don't see anything that is mapping `file-size` -> `size` anywhere.

Addressed.

systemcrash:
Any final comments @efahl ? If not LGTM and drop the patch in main.

efahl:
Looks good, commit away!

map-b:
Before (`apk list --full`)

```
Package: firewall4
Version: 2025.03.17~b6e51575-r1
Depends: kmod-nft-core, kmod-nft-fib, kmod-nft-nat, kmod-nft-offload, libc, nftables-json, ucode, ucode-mod-fs, ucode-mo
d-ubus, ucode-mod-uci
Provides: firewall4-any, uci-firewall
Description: This package provides an nftables-based implementation of the UCI firewall sharing the same configuration f
ormat.
License: ISC
Installed-Size: 130657
Size: 30944

Package: firewall4
Version: 2025.03.17~b6e51575-r2
Depends: kmod-nft-core, kmod-nft-fib, kmod-nft-nat, kmod-nft-offload, libc, nftables-json, ucode, ucode-mod-fs, ucode-mod-ubus, ucode-mod-uci
Provides: firewall4-any, uci-firewall
Description: This package provides an nftables-based implementation of the UCI firewall sharing the same configuration format.
License: ISC
Installed-Size: 130657
Size: 30950
```
After (`apk query`)
```
"package": "firewall4-2025.03.17~b6e51575-r2",
    "name": "firewall4",
    "version": "2025.03.17~b6e51575-r2",
    "description": "This package provides an nftables-based implementation of the UCI firewall sharing the same configuration format.",
    "arch": "mipsel_24kc",
    "license": "ISC",
    "origin": "feeds/base/network/config/firewall4",
    "maintainer": "Jo-Philipp Wich <jo@mein.io>",
    "installed-size": 130657,
    "file-size": 30950,
    "provider-priority": 100,
    "depends": [
      "kmod-nft-core",
      "kmod-nft-fib",
      "kmod-nft-nat",
      "kmod-nft-offload",
      "libc",
      "nftables-json",
      "ucode>=2022.03.22",
      "ucode-mod-fs",
      "ucode-mod-ubus",
      "ucode-mod-uci"
    ],
    "provides": [
      "firewall4-any",
      "uci-firewall=2025.03.17~b6e51575-r2"
    ],
```

The versioned provides appear with the version in the `query` case.
- https://github.com/openwrt/luci/pull/8593

PR #8472, opened 2026-03-27
luci-app-crowdsec-firewall-bouncer: rename UCI config to crowdsec-firewall-bouncer

leow149:
- [x] Tested on: mediatek/filogic (Banana Pi BPI-R4), OpenWrt 25.12.0-rc5 :white_check_mark:
- [x] ( Preferred ) Mention: @ne20002
- [x] ( Optional ) Depends on: openwrt/packages#28991
- [x] Description: Update UCI config name from `crowdsec` to `crowdsec-firewall-bouncer` in
  `form.js` and the rpcd ACL file, to match the config file rename in the companion packages PR
  openwrt/packages#28991. Without this change, the LuCI app would silently read and write from
  the wrong UCI config file.

### DIFF
--- a/applications/luci-app-crowdsec-firewall-bouncer/htdocs/luci-static/resources/view/crowdsec-firewall-bouncer/form.js
+++ b/applications/luci-app-crowdsec-firewall-bouncer/htdocs/luci-static/resources/view/crowdsec-firewall-bouncer/form.js
@@ -7,7 +7,7 @@ return view.extend({
 	render: function() {
 		let m, s, o;
 
-		m = new form.Map('crowdsec', _('CrowdSec'),
+		m = new form.Map('crowdsec-firewall-bouncer', _('CrowdSec'),
 			_('Gain %s protection against malicious IPs.'.format('<a href="http://www.crowdsec.net">crowd-sourced</a>')) + '<br/>' +
 			_('Benefit from the most accurate CTI in the world.'));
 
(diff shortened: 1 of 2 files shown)


dwardor:
@leow149 if you want this to be merged, you need to adress the check failures

applications/luci-app-crowdsec-firewall-bouncer/htdocs/luci-static/resources/view/crowdsec-firewall-bouncer/form.js
```diff
@@ -7,7 +7,7 @@ return view.extend({
 	render: function() {
 		let m, s, o;
 
-		m = new form.Map('crowdsec', _('CrowdSec'),
+		m = new form.Map('crowdsec-firewall-bouncer', _('CrowdSec'),
```
openwrt-ai, line 10: This rename only works once the backend rename lands. In `openwrt/packages` master the bouncer still reads the old config name: the init script does `config_load "${CONFIGURATION}"` with `CONFIGURATION=crowdsec`, and the Makefile installs `./files/crowdsec.config` to `/etc/config/crowdsec` (with `procd_add_config_trigger "config.change" "crowdsec"`). So if this LuCI PR is merged before openwrt/packages#28991, the frontend will read/write `/etc/config/crowdsec-firewall-bouncer` while the daemon still loads `/etc/config/crowdsec`, silently breaking the app.

The dependency is noted in the PR description, so this is a merge-ordering question rather than a defect: can you confirm this should not be merged until packages#28991 is in, and ideally that the packages PR also handles migrating any existing `/etc/config/crowdsec` on upgrade so current users don't lose their settings?

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline question about backend/merge-ordering coupling.

---

PR #8490, opened 2026-04-01
luci-mod-status: move getConntrackList from ubus to fs.exec

### LINKED ISSUE #6360: luci: Missleading error message when rpcd fails
## Steps to reproduce:

1. Have ~5k active connections
2. go to: Status → Realtime Graphs → Connections
3. wait a few seconds
4. reload page because it will not respond anymore
5. see error message `/usr/lib/lua/luci/dispatcher.lua:431: /etc/config/luci seems to be corrupt, unable to find section 'main'`

## Actual behavior:
* Error message `/usr/lib/lua/luci/dispatcher.lua:431: /etc/config/luci seems to be corrupt, unable to find section 'main'`

## Expected behavior:
* Maybe `rpcd does not respond`? 

## Additional Information:
At least 22.03.[0,1,2,3] is also effected, 21.03 probably too. See https://github.com/openwrt/luci/issues/4758 for reference.

```
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='22.03.4'
DISTRIB_REVISION='r20123-38ccc47687'
DISTRIB_TARGET='mediatek/mt7622'
DISTRIB_ARCH='aarch64_cortex-a53'
DISTRIB_DESCRIPTION='OpenWrt 22.03.4 r20123-38ccc47687'
```

jjm2473:
to prevent procd from being kicked off by ubusd
  due to large amounts of data.
See https://github.com/openwrt/openwrt/issues/9747

Fixes https://github.com/openwrt/luci/issues/6360

- [x] Tested on: :white_check_mark: qemu arm64, 25.12.1, Chrome
- [x] \( Preferred ) Mention: @jow- @systemcrash 
- [x] \( Optional ) Closes: openwrt/luci#6360
- [x] Description:
Move getConntrackList from ubus to web to prevent procd from being kicked off by ubusd due to large amounts of data.

Has been tested with a fake 5MB /proc/net/nf_conntrack, procd not kicked off (Such a large amount of data can actually cause browser lag.).
image

### DIFF
--- a/modules/luci-mod-status/root/usr/libexec/luci-status-call
+++ b/modules/luci-mod-status/root/usr/libexec/luci-status-call
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+action=$1
+shift
+
+case "$action" in
+	conntrack)
+		if [[ -f /proc/net/nf_conntrack ]]; then
+			/bin/grep -vF TIME_WAIT /proc/net/nf_conntrack
+		elif [[ -x /usr/sbin/conntrack ]]; then
+			/usr/sbin/conntrack -L -o extended | /bin/grep -vF TIME_WAIT
+		fi
+	;;
+esac
--- a/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
+++ b/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
@@ -3,7 +3,9 @@
 		"description": "Grant access to realtime statistics",
 		"read": {
 			"file": {
-				"/etc/services": [ "read" ]
+				"/etc/protocols": [ "read" ],
+				"/etc/services": [ "read" ],
+				"/usr/libexec/luci-status-call": [ "exec" ]
 			},
 			"ubus": {
 				"luci": [ "getConntrackList", "getRealtimeStats" ],
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/connections.js
@@ -12,12 +12,6 @@ const callLuciRealtimeStats = rpc.declare({
 	expect: { result: [] }
 });
 
-const callLuciConntrackList = rpc.declare({
-	object: 'luci',
-	method: 'getConntrackList',
-	expect: { result: [] }
-});
-
 const callNetworkRrdnsLookup = rpc.declare({
 	object: 'network.rrdns',
 	method: 'lookup',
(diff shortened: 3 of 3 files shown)


jjm2473:
image

This is test on 24.10.5.
Create thousands of connections using https://qps.itzmx.com/.

jjm2473:
[v2]:
1. Use `grep` to filter out connections in the TIME_WAIT state on backend to reduce front-end load
2. Ignore unused fields
3. Reduce commit message

Link to v1: https://github.com/jjm2473/luci/commit/bc970eaa19c45c5c5f37cf0bfa5bab98c2d9bc5a

jjm2473:
image

Tested with over 10,000 connections, no issues were found, and the frontend remained stable.

jow-:
That's a good idea, I like it. I just wonder if it would be better to have a dedicated backend route which takes care of the parsing and spits out a JSON for the front end to consume instead of parsing the raw system files in the frontend code directly.

ptpt52:
LGTM

jjm2473:
In fact, modern browsers generally outperform routers, so it might be more appropriate to perform the main computation in the browser?
The backend could use grep or sed to filter out some data to reduce network traffic, since luci (uhttpd) currently doesn't compress data.

jjm2473:
Update: use `grep -vF` instead of `grep -v`, saving ~60% CPU time on backend.

image

jjm2473:
[v3]:
1. Add a wrapper for easy align to the `/usr/sbin/conntrack -L -o extended` fallback in `sys.uc` (the fallback was added by https://github.com/openwrt/luci/commit/a30c8d8872e333dad05878a7b4da49a2e13ad98b ).

Link to v2: https://github.com/jjm2473/luci/commit/f80986d03b25dd6dd00b87a0dc800f83e84800fb
Link to v1: https://github.com/jjm2473/luci/commit/bc970eaa19c45c5c5f37cf0bfa5bab98c2d9bc5a

modules/luci-mod-status/root/usr/libexec/luci-status-call
systemcrash, line 1: Can we rename this to e.g. `luci-status-conntrack-call`?

jjm2473: @systemcrash 
Perhaps in the future, Luci will need to bypass ubusd when reading other data. In that case, this can be written to this file, and the first parameter, `action`, can define new actions. If this extensibility is not needed, I can also rename the file to `luci-status-conntrack-call` as you mentioned and remove the parameter.

modules/luci-mod-status/root/usr/libexec/luci-status-call
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+action=$1
+shift
+
+case "$action" in
+	conntrack)
+		if [[ -f /proc/net/nf_conntrack ]]; then
```
openwrt-ai, line 8: `[[ ... ]]` is a bash keyword and is not implemented by the BusyBox `ash` that provides `/bin/sh` on OpenWrt. With this `#!/bin/sh` shebang the test will fail at runtime with `[[: not found`, so neither branch runs and the conntrack list comes back empty on-device. Use the POSIX single-bracket `[ ... ]` instead.

```suggestion
		if [ -f /proc/net/nf_conntrack ]; then
```

The same applies to the `elif` on line 10 below: `elif [ -x /usr/sbin/conntrack ]; then`.

---

modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
```diff
@@ -3,7 +3,9 @@
 		"description": "Grant access to realtime statistics",
 		"read": {
 			"file": {
-				"/etc/services": [ "read" ]
+				"/etc/protocols": [ "read" ],
+				"/etc/services": [ "read" ],
+				"/usr/libexec/luci-status-call": [ "exec" ]
 			},
 			"ubus": {
 				"luci": [ "getConntrackList", "getRealtimeStats" ],
```
openwrt-ai, line 11: nit: with `callLuciConntrackList` removed from `connections.js`, nothing in luci-mod-status calls the `getConntrackList` ubus method anymore, so this ACL grant is now dead. The backend method itself still lives in luci-base (`luci.sys.conntrack_list`), so leaving it is harmless, but you may want to drop `getConntrackList` from this list to keep the granted permissions in sync with what the frontend actually uses.

---

openwrt-ai (commented):
Reviewed the conntrack-from-fs migration. The `compressIpv6` helper reproduces all of its documented test cases correctly and the parser matches the `sys.uc:conntrack_list` logic it ports. One runtime correctness issue (bashism under `/bin/sh`) and one ACL nit inline.

---

modules/luci-mod-status/root/usr/libexec/luci-status-call
jjm2473: openwrt always support this feature.

jjm2473:
ping @jow- @systemcrash

openwrt-ai (commented):

---

PR #8532, opened 2026-04-06
luci-theme-bootstrap: fix header .pull-right in mobile.css

ghost:
### Description
Noticed the indicators behaving a little odd on mobile, and realised one flex element in the mobile.css was the cause. Removing the flex entry did not cause any further issue, and fixed the indicator misbehaviour.

### Maintainer _(Preferred)_

@jow- 

---

## Tested on
**OpenWrt version:** OpenWrt 25.12.2 (r32802-f505120278)
**LuCI version:** LuCI openwrt-25.12 branch (26.094.61984~16167c7)
**Web browser:** Safari Version 26.5 (21624.2.1.11.1)

---

## Checklist
- [ ] Incremented :up: any `PKG_VERSION` in the Makefile. _(Not applicable for theme)_

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/mobile.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/mobile.css
@@ -279,7 +279,6 @@ header h3 a, header .brand {
 	}
 
 	header .pull-right {
-		flex: 0 1 20%;
 		display: flex;
 		flex-direction: column;
 		padding: 0;


efahl:
Just pondering here...  Was the button sized to the "Refreshing" label and then stayed that size all the time?  If so, is it possible that the menu wrapping to the left would change on button size change as you toggle between "Paused" and "Refreshing"?  (Assuming you'd have to land on some very specific screen size/font size combo, but looks possible, maybe?)

ghost:
> is it possible that the menu wrapping to the left would change on button size change as you toggle between "Paused" and "Refreshing"?

Good idea. Just had a play again.

It doesn't seem to be specific to any existing text, but just the percentage value of flex causing it to misbehave. Removing the value allows the indicator to flex to the text itself, which I think is a more straightforward approach.

Adobe Express - Screen Recording 2026-04-10 at 8 57 33 am

PR #8540, opened 2026-04-10
luci: add .zed to .gitignore

ghost:
Adding `.zed` to `.gitignore` so the [Zed editor](https://zed.dev) does not upload settings and app specific files to the repository.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ docs/**/*.*
 **eslint.config**
 !eslint.config.mjs
 .DS_Store
+.zed


GeorgeSapkin:
Why single out one editor? Just don't stage unnecessary files?

ghost:
You're right, I wasn't aware that there are [several options](https://docs.github.com/en/get-started/git-basics/ignoring-files#excluding-local-files-without-creating-a-gitignore-file) for me to exclude them without having to commit to the repo. I'll put the blame on GitHub Desktop, as it did not provide me any other way except automatically committing the .gitignore file.

Will just add the exclusion locally as per "[Excluding local files without creating a .gitignore file](https://docs.github.com/en/get-started/git-basics/ignoring-files#excluding-local-files-without-creating-a-gitignore-file)".

PR #8543, opened 2026-04-12
luci-app-uci-git-backup: add LuCI interface

mrangger:
## Summary
- add a new `luci-app-uci-git-backup` application with settings and restore views for `uci-git-backup`
- include the LuCI menu entry, rpcd ACL, and German translations for the JS interface
- depend on the backend package proposed in https://github.com/openwrt/packages/pull/29131

### DIFF
--- a/applications/luci-app-uci-git-backup/htdocs/luci-static/resources/view/uci-git-backup/restore.js
+++ b/applications/luci-app-uci-git-backup/htdocs/luci-static/resources/view/uci-git-backup/restore.js
@@ -1,0 +1,120 @@
+'use strict';
+'require view';
+'require fs';
+'require ui';
+
+function parseCommits(logText) {
+	return (logText || '').split(/\n/).filter(Boolean).map(function(line) {
+		const parts = line.split('\t');
+
+		if (parts.length < 3)
+			return null;
+
+		return {
+			sha: parts[0],
+			date: parts[1],
+			msg: parts.slice(2).join('\t')
+		};
+	}).filter(Boolean);
+}
+
+function renderOutput(title, res, reloadAfterClose) {
+	const output = [res.stdout, res.stderr].filter(Boolean).join('\n').trim() || _('No output.');
+	const success = (res.code === 0);
+
+	ui.showModal(title, [
+		E('p', { 'class': success ? 'spinning' : null }, [
+			success ? _('Command completed successfully.') : _('Command failed with exit code %d.').format(res.code)
+		]),
+		E('pre', { 'style': 'white-space: pre-wrap' }, [ output ]),
+		E('div', { 'class': 'right' }, [
+			E('button', {
+				'class': 'btn',
+				'click': function() {
+					ui.hideModal();
+
+					if (reloadAfterClose)
+						window.location.reload();
+				}
+			}, [ _('Dismiss') ])
+		])
+	]);
+}
+
+return view.extend({
+	handleRestoreConfirm: function(sha) {
+		ui.showModal(_('Restoring backup...'), [
+			E('p', { 'class': 'spinning' }, [ _('The selected commit is being restored now.') ])
+		]);
+
+		return fs.exec('/usr/bin/uci-git-restore', [ sha ]).then(function(res) {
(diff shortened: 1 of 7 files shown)


applications/luci-app-uci-git-backup/root/usr/share/rpcd/acl.d/luci-app-uci-git-backup.json
```diff
@@ -0,0 +1,23 @@
+{
+	"luci-app-uci-git-backup": {
+		"description": "Grant access to luci-app-uci-git-backup",
+		"read": {
+			"uci": ["uci_git_backup", "system", "dhcp"],
```
openwrt-ai, line 5: The frontend only ever touches the `uci_git_backup` config — `settings.js` does `uci.load('uci_git_backup')` and `uci.get(CONFIG, …)` and nothing else (no read of `system` or `dhcp`). Granting read access to `system` and `dhcp` here is broader than the app needs and widens the rpcd attack surface for no functional gain. Drop them:

```suggestion
			"uci": ["uci_git_backup"],
```

---

applications/luci-app-uci-git-backup/htdocs/luci-static/resources/view/uci-git-backup/restore.js
```diff
@@ -0,0 +1,120 @@
...
+		};
+	}).filter(Boolean);
+}
+
+function renderOutput(title, res, reloadAfterClose) {
+	const output = [res.stdout, res.stderr].filter(Boolean).join('\n').trim() || _('No output.');
+	const success = (res.code === 0);
+
+	ui.showModal(title, [
+		E('p', { 'class': success ? 'spinning' : null }, [
```
openwrt-ai, line 26: nit: applying the `spinning` class on success leaves the busy-spinner animation running on a modal that is actually finished — the command has already completed and the user is meant to read the output and dismiss. The spinner should only indicate in-progress work. The `else`/failure branch correctly uses no spinner. Consider dropping it for the success case too:

```suggestion
		E('p', {}, [
```

The same pattern is in [`settings.js` `renderOutput` line 26](`https://github.com/openwrt/luci/blob/b83a4be584bbc1c70af0aa9a18e2405d3fac8a18/applications/luci-app-uci-git-backup/htdocs/luci-static/resources/view/uci-git-backup/settings.js#L26).`

---

applications/luci-app-uci-git-backup/htdocs/luci-static/resources/view/uci-git-backup/settings.js
```diff
@@ -0,0 +1,245 @@
...
+	},
+
+	handleTestConnection: function() {
+		const values = this.getCurrentSettingsValues();
+		const args = [
+			'--remote-url', values.remote_url || '',
+			'--branch', values.branch || 'main',
+			'--auth-type', values.auth_type || 'password',
+			'--username', values.username || '',
+			'--password', values.password || ''
```
openwrt-ai, line 94: Passing the password/token as a command-line argument to `/usr/bin/uci-git-test` exposes the secret in the process list (`/proc/<pid>/cmdline`, `ps`) for the lifetime of the test run, where any local user/process can read it. `fs.exec` uses an argv array so there's no shell-injection risk, but argv visibility is the concern. Since the value is already persisted to UCI by `saveCurrentSettings()` immediately before this call, could `uci-git-test` read it from the config (or stdin) instead of receiving it on the command line? This depends on the backend interface in openwrt/packages#29131, so flagging as a question for you to confirm how the backend consumes the credential.

---

openwrt-ai (commented):
Fresh review of this new `luci-app-uci-git-backup` frontend. The backend dependency (`uci-git-backup`, openwrt/packages#29131) is correctly declared in the PR body and the executed scripts (`/usr/bin/uci-git-{backup,test,list,restore}`) are gated behind it via `LUCI_DEPENDS`, so the frontend-only-no-consumer concern is addressed. Inline comments below: one over-broad ACL grant, one secret-handling question on the backend interface, and a spinner nit. None block merge.

---

PR #8548, opened 2026-04-12
luci-mod-system: add current host to reset reconnect

aslafy-z:
### Description

When performing a sysupgrade without keeping settings, `ui.awaitReconnect()`
only tries the hardcoded `192.168.1.1` and `openwrt.lan`. This fails silently
when the device uses a different static IP, forcing the user to manually
navigate to the correct address.

This change prepends `window.location.host` to the reconnect target list in
the factory reset code path, keeping `192.168.1.1` and `openwrt.lan` as
fallbacks. This matches the behavior already used when settings are kept.

Common use case: custom images built with the OpenWrt Image Builder with
pre-configured static IPs. After flashing, the device comes back on a known
address that differs from the defaults.

### Screenshot or video of changes _(If applicable)_

N/A - behavior change only visible during sysupgrade reconnect.

### Maintainer _(Preferred)_

@jow-

---

## Tested on
**OpenWrt version:** 24.10.6
**LuCI version:** git-24.xxx (luci-mod-system)
**Web browser:** Firefox

---

## Checklist

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js
@@ -337,7 +337,7 @@ return view.extend({
 		if (opts['keep'][0].checked)
 			ui.awaitReconnect(window.location.host);
 		else
-			ui.awaitReconnect('192.168.1.1', 'openwrt.lan');
+			ui.awaitReconnect(window.location.host, '192.168.1.1', 'openwrt.lan');
 	},
 
 	handleBackupList(ev) {


systemcrash:
> This fails silently when the device uses a different static IP

In which cases does this happen? The defaults are what is set during a factory reset.

aslafy-z:
@systemcrash flashing does not mean factory reset.  I wrongly mentioned factory reset in the PR desc, and just edited it. Here's the updated section: 

> Common use case: custom images built with the OpenWrt Image Builder with
> pre-configured static IPs. After flashing, the device comes back on a known
> address that differs from the defaults.

systemcrash:
OK - but when flashing without retaining settings ("aka factory reset"), the default is 192.168.1.1. So the question still stands - what do I need to do to find myself in your situation (to reproduce this issue)?

aslafy-z:
The situation is: "do a sysupgrade to a firmware image whose built-in default LAN IP is not 192.168.1.1".

"Without retaining settings" resets to the defaults of the flashed image, not necessarily to the stock OpenWrt defaults. So this is reproducible with custom images that change the LAN address at build time, for example through Image Builder files a baked /etc/config/network, or a uci-defaults script.

Example reproduction:

1. Build an image with network.lan.ipaddr=192.168.42.1
2. Trigger sysupgrade with "do not keep settings"
3. After reboot, the device is reachable at 192.168.42.1, but awaitReconnect() only probes 192.168.1.1 and openwrt.lan

systemcrash:
Merged. Thanks @aslafy-z

PR #8555, opened 2026-04-16
luci-mod-network: fix wrong name

AndyChiang888:
### Description

This should be nftables instead of netfilter.

---

## Tested on
**OpenWrt version:** SNAPSHOT
**LuCI version:** master
**Web browser:** Chrome

---

## Checklist

### DIFF
--- a/modules/luci-base/po/ar/base.po
+++ b/modules/luci-base/po/ar/base.po
@@ -7984,8 +7984,8 @@ msgid "Neighbour cache validity"
 msgstr "صلاحية ذاكرة التخزين المؤقت للمجاوريين"
 
 #: modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js:848
-msgid "Netfilter table name"
-msgstr "اسم جدول Netfilter"
+msgid "Nftables table name"
+msgstr "اسم جدول Nftables"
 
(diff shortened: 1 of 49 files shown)


systemcrash:
nf in nftables == netfilter.

AndyChiang888:
iptables also uses netfilter, so changing it to nftables is more appropriate. The netfilter table actually refers to the nftables table.

systemcrash:
opinion @efahl @hnyman ?

hnyman:
I already had earlier a private email from @AndyChiang888 about this.
Netfilter is the general name for the whole infrastructure, and iptables etc. were the traditional components of it, but they have been replaced by nftables tools a few years ago. 

To me the current 'netfilter' word there is correct enough, so I see no immediate need for changing it. 
But I am not strongly opposing the change, either.

AndyChiang888:
I think this will still confuse some users because dnsmasq's nftset specifies the nftables table name, but luci displays it as the netfilter table name, which could be misleading.

efahl:
I agree with @hnyman.

Use of either the Linux component "netfilter" or the specific project "nftables" to denote them is sort of a wash in this context, as all of that detail is completely hidden from the user in the gui.  Anything will work here; those who are going to dive into the guts and hand write rules/sets already knows what's being used.

But, if you want to be pedantic, you'd change all reference to "ipset" to "nftset", too.

PR #8569, opened 2026-04-20
luci-app-snmpd: add strong SNMPv3 crypto options

micpf:
Add SHA-256, SHA-384, SHA-512, SHA-224 authentication and AES-192,
AES-256 encryption options to the SNMPv3 user configuration.

net-snmp has supported these algorithms for a while, but the LuCI
dropdown only offered SHA(-1), MD5, AES(-128) and DES. Modern security
standards (e.g. EN 18031 CCK-1) require at least 112-bit security
strength, which rules out SHA-1 (80-bit) and DES (56-bit).

The default authentication type is changed from SHA to SHA-256.

The init script already passes `auth_type` and `privacy_type` directly to
net-snmp's `createUser`, so no backend changes are needed.

Tested with net-snmp 5.9.4 on OpenWrt 25.12 (aarch64).


micpf:
Closing to retarget to main branch (stable branch requires cherry-pick marking).

PR #8575, opened 2026-04-22
luci-app-mwan3: fix missing translation call for Weight label

micpf:
The 'Weight' option label in the member configuration page uses
`('Weight')` instead of `_('Weight')`, so the string is never passed
through the translation function. This causes the label to always display
in English regardless of the selected UI language.

The translation for 'Weight' already exists in luci-base, so this
one-character fix is sufficient.


systemcrash:
Merged. Thanks @micpf

PR #8576, opened 2026-04-22
luci-mod-network: fill ieee-oui.txt from ufp-neigh

ckorber:
With ufp-neigh it is possible to get the vendor of a MAC address. Because arp-scan allows us to specify a txt file for lookup, it is suitable to use ufp-neigh to create a txt file on the spot and populate it with vendors of connected devices. This is only used if ufp is running as service.

Depends on changes to ufp.
https://lists.openwrt.org/pipermail/openwrt-devel/2026-April/044755.html

@systemcrash 

---

**OpenWrt version:** owrt-25.12
**Web browser(s):** Firefox
---

## Checklist

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
@@ -4,7 +4,24 @@
 'require fs';
 'require ui';
 'require uci';
+'require rpc';
 'require network';
+
+const callFingerprint = rpc.declare({
+	object: 'fingerprint',
+	method: 'fingerprint',
+	params: [ 'export' ],
+	expect: { '': {} }
+});
+
+const callServiceList = rpc.declare({
+	object: 'service',
+	method: 'list',
+	params: [ 'name' ],
+	expect: { ufp: {} }
+});
+
+let has_ufp;
 
 return view.extend({
 	handleCommand(exec, args) {
@@ -50,8 +67,11 @@ return view.extend({
 
 	handleArpScan(ev, cmd) {
 		const addr = ev.currentTarget.parentNode.previousSibling.value;
+		const ieee_file = '/tmp/arp-scan/ieee-oui.txt';
 
-		return this.handleCommand('arp-scan', [ '-l', '-I', addr ]);
+		if (!has_ufp)
+			return this.handleCommand('arp-scan', [ '-l', '-I', addr ]);
+		return this.handleCommand('arp-scan', [ '-l', '-I', addr, '-O', ieee_file ]);
 	},
 
 	load() {
@@ -60,14 +80,25 @@ return view.extend({
 			L.resolveDefault(fs.stat('/bin/traceroute6') || fs.stat('/usr/bin/traceroute6'), false),
 			L.resolveDefault(fs.stat('/usr/bin/arp-scan'), false),
 			network.getDevices(),
+			L.resolveDefault(callServiceList('ufp').then((res) => {
+				let instances = res.instances;
+
+				for (let i in instances) {
+					if (instances[i].running)
+						return callFingerprint(true);
+				}
+
+				return null;
+			}), null),
 			uci.load('luci')
 		]);
 	},
 
-	render([has_ping6, has_traceroute6, has_arpscan, devices]) {
+	render([has_ping6, has_traceroute6, has_arpscan, devices, ufp]) {
 		const dns_host = uci.get('luci', 'diag', 'dns') || 'openwrt.org';
 		const ping_host = uci.get('luci', 'diag', 'ping') || 'openwrt.org';
 		const route_host = uci.get('luci', 'diag', 'route') || 'openwrt.org';
+		has_ufp = ufp != null ? true : false;
 
 		const table = E('table', { 'class': 'table' }, [
 				E('tr', { 'class': 'tr' }, [
--- a/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
+++ b/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
@@ -78,7 +78,9 @@
 				"/usr/bin/arp-scan": [ "exec", "list" ]
 			},
 			"ubus": {
-				"file": [ "exec", "stat" ]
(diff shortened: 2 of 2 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
```diff
@@ -4,8 +4,17 @@
 'require fs';
 'require ui';
 'require uci';
+'require rpc';
 'require network';
 
+var callFingerprint = rpc.declare({
```
systemcrash: const

modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
```diff
@@ -4,8 +4,17 @@
...
+	object: 'fingerprint',
+	method: 'fingerprint',
+	expect: { '': {} }
+});
+
+var has_ufp;
```
systemcrash: let

modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
```diff
@@ -50,8 +59,11 @@ return view.extend({
 
 	handleArpScan(ev, cmd) {
 		const addr = ev.currentTarget.parentNode.previousSibling.value;
+		const ieee_file = '/tmp/arp-scan/ieee-oui.txt';
```
systemcrash, line 70: Is the file always in this location?

ckorber: Yes, it is planned to reside always there.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
```diff
@@ -60,14 +73,20 @@ return view.extend({
 			L.resolveDefault(fs.stat('/bin/traceroute6') || fs.stat('/usr/bin/traceroute6'), false),
 			L.resolveDefault(fs.stat('/usr/bin/arp-scan'), false),
 			network.getDevices(),
+			fs.exec('service', [ 'ufp', 'status' ]).then((res) => {
+				if (res.stdout.trim() == 'running')
+					return callFingerprint(true);
+				return null;
+			}),
```
systemcrash: There is the ubus object `service` which means we can avoid this exec call, and call via an rpc.declare. The intent is more specific so we need a more specific check (`ufp` -> instances -> instancex -> running). This requires then only the a `ubus` k+v in the acl.d JSON.

I recommend this route since the service call can have long execution times, and if ufp is unavailable, the service call dumps everything else that's running.

ckorber: I rely on only one instance but implemented a for-loop just in case.
So if an instance is running the fingerprint call will be performed, otherwise not.

modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
```diff
@@ -75,10 +75,13 @@
 				"/usr/bin/ping6": [ "exec", "list" ],
 				"/usr/bin/traceroute": [ "exec" ],
 				"/usr/bin/traceroute6": [ "exec", "list" ],
-				"/usr/bin/arp-scan": [ "exec", "list" ]
+				"/usr/bin/arp-scan": [ "exec", "list" ],
+				"/sbin/service": [ "exec" ]
```
systemcrash: this can be dropped now I figure

ckorber:
ping @systemcrash

systemcrash:
> Depends on changes to ufp. https://lists.openwrt.org/pipermail/openwrt-devel/2026-April/044755.html

Are these changes in yet?

ckorber:
> > Depends on changes to ufp. https://lists.openwrt.org/pipermail/openwrt-devel/2026-April/044755.html
> 
> Are these changes in yet?

Sorry, my bad. Forgot about it.

systemcrash:
I can't find a repo for it, so it's a bit of a shame we can't make a PR for it.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js
```diff
@@ -60,14 +80,25 @@ return view.extend({
...
+			callServiceList('ufp').then((res) => {
+				let instances = res.instances;
+
+				for (let i in instances) {
+					if (instances[i].running)
+						return callFingerprint(true);
+				}
+
+				return null;
+			}),
```
openwrt-ai: Unlike the four sibling entries in this `Promise.all`, this chain is not wrapped in `L.resolveDefault`, so a rejection propagates and fails the whole `load()` — blanking ping/traceroute/nslookup/arp-scan, not just ufp detection. The `fingerprint` ubus call is the most likely to reject here: it is a not-yet-merged ufp addition (see PR description), so on any deployed system that has ufp running but lacks the new `export` method, or hits an ACL denial, the entire diagnostics page would break. Wrap it like the [neighbouring `L.resolveDefault(...)` entries](`https://github.com/openwrt/luci/blob/c7397db0cc8f0d8c69f09bdb85a6676401306a29/modules/luci-mod-network/htdocs/luci-static/resources/view/network/diagnostics.js#L80-L82)` so failure degrades to "no ufp" rather than a dead page:

```suggestion
			L.resolveDefault(callServiceList('ufp').then((res) => {
				let instances = res.instances;

				for (let i in instances) {
					if (instances[i].running)
						return callFingerprint(true);
				}

				return null;
			}), null),
```

---

openwrt-ai (commented):
Reviewed the diff at c7397db. One inline comment on `load()` error handling. The ufp-side `fingerprint`/`export` ubus interface this depends on is not yet merged and has no public repo (already discussed in-thread), so the backend coupling can't be verified here; that blocker is correctly noted in the PR description.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The `load()` chain is now wrapped in `L.resolveDefault(..., null)`, which resolves the prior comment about a rejected `fingerprint` call breaking the whole diagnostics page.

---

PR #8579, opened 2026-04-22
luci: add user-defined label for physical switch ports

mg-dev-01:
Add support for a UCI label option on network device sections (config device), allowing users to assign a human-friendly display name to each physical DSA switch port (e.g. lan1 → 'NAS', wan → 'ISP ONT').

Labels are shown in:

Network > Interfaces > Devices grid (Label column)
Network > Interfaces > Bridge VLAN filtering port headers
Network > Interfaces > device Configure modal (switch ports only)
Status > Overview port status boxes
Only DSA switch ports are affected, identified via the public API:
dev.getType() === 'switch' && dev.getParent() != null

The DSA parent switch device (e.g. eth0) and all other device types (bridges, VLANs, WiFi) are unaffected. swconfig devices see no UI changes and no breakage.

Screenshot :
Screenshot_20260422_215833

Maintainer (preferred)
@jow-

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
@@ -58,12 +58,21 @@ function updateDevBadge(node, dev) {
 	const type = dev.getType();
 	const up = dev.getCarrier();
 
+	let label = null;
+	uci.sections('network', 'device', function(s) {
+		if (s.name === dev.getName() && s.label)
+			label = s.label;
+	});
+
 	dom.content(node, [
-		E('img', {
-			'class': 'middle',
-			'src': L.resource('icons/%s%s.svg').format(type, up ? '' : '_disabled')
-		}),
-		'\x0a', dev.getName()
+		E('span', { 'style': 'text-align:center;display:block' }, [
+			E('img', {
+				'class': 'middle',
+				'src': L.resource('icons/%s%s.svg').format(type, up ? '' : '_disabled')
+			})
+		]),
+		E('span', { 'style': 'text-align:center;display:block' }, [dev.getName()]),
+		E('small', { 'style': 'font-weight:normal;text-align:center;display:block' }, [label || '\u00a0'])
 	]);
 
 	return node;
@@ -1128,6 +1137,12 @@ return baseclass.extend({
 		o = this.replaceOption(s, 'devgeneral', form.Value, 'txqueuelen', _('TX queue length'));
 		o.placeholder = dev ? dev._devstate('qlen') : '';
 		o.datatype = 'uinteger';
+
+		if (dev && dev.getType() === 'switch' && dev.getParent() != null) {
+			o = this.replaceOption(s, 'devgeneral', form.Value, 'label', _('Label'),
+				_('Optional human-friendly name for this port, shown in the UI alongside the device name.'));
+			o.rmempty = true;
+		}
 
 		/* PSE / PoE options */
 		if (hasPSE) {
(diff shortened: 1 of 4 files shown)


systemcrash:
This looks OK. Thoughts @hnyman @feckert ?

hnyman:
Sounds interesting.

(Why the unicode char definition for empty case? Is it really needed. Preferably ASCII only.)

systemcrash:
> (Why the unicode char definition for empty case? Is it really needed. Preferably ASCII only.)

Thought the same at review.

efahl:
Hmm, maybe the nbsp is to force consistent vertical spacing?  Didn't test so this is just speculation: if the div or whatever is empty, it might just disappear, causing the contents to wander up and down across the rows???

hnyman:
> maybe the nbsp is to force consistent vertical spacing?

I don't criticize nbsp by itself, but just the Unicode char. As OpenWrt is supposed to work also with rather simple/old browsers, it would be prerably to avoid complex symbols and stick to ASCII, if possible. 

Use just a normal space instead?

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -58,12 +58,21 @@ function updateDevBadge(node, dev) {
...
-		}),
-		'\x0a', dev.getName()
+		E('span', { 'style': 'text-align:center;display:block' }, [
+			E('img', {
+				'class': 'middle',
+				'src': L.resource('icons/%s%s.svg').format(type, up ? '' : '_disabled')
+			})
+		]),
+		E('span', { 'style': 'text-align:center;display:block' }, [dev.getName()]),
+		E('small', { 'style': 'font-weight:normal;text-align:center;display:block' }, [label || '\u00a0'])
```
openwrt-ai, line 75: nit: maintainer @hnyman asked in the PR thread to avoid the Unicode NBSP (` `) and stick to ASCII for compatibility with simple/old browsers. A plain space works as the empty-case placeholder here.

```suggestion
		E('small', { 'style': 'font-weight:normal;text-align:center;display:block' }, [label || ' '])
```

Same applies to the `portLabel || ' '` fallback in `29_ports.js`.

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -58,12 +58,21 @@ function updateDevBadge(node, dev) {
 	const type = dev.getType();
 	const up = dev.getCarrier();
 
+	let label = null;
+	uci.sections('network', 'device', function(s) {
+		if (s.name === dev.getName() && s.label)
+			label = s.label;
+	});
```
openwrt-ai, line 65: Unlike the other two display sites — the Devices grid (`interfaces.js`) and the Status overview (`29_ports.js` builds its `labelMap` only when `s.name && s.label`, but the modal and grid both additionally gate on `dev.getType() === 'switch' && dev.getParent() != null`) — this lookup picks up a `label` from any `config device` section, not just DSA switch ports. In practice the form only writes `label` on switch ports, so this is harmless today, but a `label` set on another device section (e.g. a bridge or manually) would render here while being suppressed everywhere else. Is the broader scope here intentional, or should it match the `getType() === 'switch' && getParent() != null` gate used elsewhere?

---

openwrt-ai (commented):
Reviewed the single commit at head 24e4f5a (fresh review). The commit message matches the diff. Two inline notes below — one nit about the Unicode NBSP placeholder (already raised by a maintainer in the thread) and one question about the label-lookup gating in `updateDevBadge` differing from the other two display sites.

---

PR #8586, opened 2026-04-25
luci-mod-network: mptcp option uses '' instead of "off"

AndyChiang888:
In the mptcp options, "off" is the default option, which has the same effect as not defining it. Using '' will prevent unnecessary parameters from being written to the flash.

Link https://openwrt.org/docs/guide-user/network/mptcp

### Maintainer _(preferred)_

@<github-user>

---

## Tested on

**OpenWrt version:** SNAPSHOT
**LuCI version:** master

---

## Checklist

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -1220,11 +1220,12 @@ return view.extend({
 					o = nettools.replaceOption(s, 'advanced', form.RichListValue, 'multipath', _('Multi-Path TCP'),
 						_('Multi-Path TCP') + ' %s'.format('<a href="%s" target="_blank">RFC8684</a>').format('https://www.rfc-editor.org/rfc/rfc8684.html') + '<br/>' +
 						_('For packets originating from this device, e.g. VPN.') );
-					o.value('off', _('Off'), _('Disables this interface for MPTCP'));
+					o.value('', _('Off'), _('Disables this interface for MPTCP'));
 					o.value('on', _('On'), _('No special configuration'));
 					o.value('master', _('Master'), _('Sets default route for all traffic'));
 					o.value('backup', _('Backup'), _('Hot standby; use this interface; do not forward traffic until no other interface is available (faster)'));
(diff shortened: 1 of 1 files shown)


systemcrash:
@dangowrt can you confirm what the values should be? The wiki says otherwise. 

https://openwrt.org/docs/guide-user/network/mptcp

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1217,10 +1217,10 @@ return view.extend({
 				o.placeholder = '0';
 
 				if (L.hasSystemFeature('mptcp')) {
-					o = nettools.replaceOption(s, 'advanced', form.RichListValue, 'multipath', _('Multi-Path TCP'),
+					o = nettools.replaceOption(s, 'advanced', form.ListValue, 'multipath', _('Multi-Path TCP'),
```
systemcrash: drop this - the extra descriptions for the other options are very helpful

AndyChiang888: RichListValue cannot use null values.
So I don't know how to handle this properly, whether RichListValue should be allowed to use null values?

AndyChiang888: RichListValue cannot use empty value.
So I don't know how to handle this properly, whether RichListValue should be allowed to use empty value?

AndyChiang888: ping @systemcrash

systemcrash: how about (something like) `o.default = ''` ?

AndyChiang888: image


Adding `o.default = '';` doesn't work either; it still gives this error.
Might need to modify `RichListValue` to use an empty value.

systemcrash: Ah, it probably needs an `o.optional = true;` to allow empty values.

AndyChiang888: Adding `o.optional = true;` resolved this issue.

PR #8592, opened 2026-04-27
luci-mod-system, luci-app-acl: add password policy

ckorber:
### Description
In order to enforce a password policy, this PR does the following:
1. Change `ubus call luci setPassword` that it can be used for rpcd also
2. Use it in `luci-app-acl`
3. Add a password policy tab (length, digits, uppercase/lowercase, special characters)
4. Enforce it on `luci-mod-system` and `luci-app-acl`

**Administration:**
assets/bd648ad7-41fd-4b8d-b1ba-8b952d822192" />
image

**Policy:**
image

**ACL:**

image

### Maintainer _(preferred)_

@systemcrash 

---

## Tested on

**Web browser(s):** Firefox

---

## Checklist

### DIFF
--- a/modules/luci-base/root/usr/share/rpcd/ucode/luci
+++ b/modules/luci-base/root/usr/share/rpcd/ucode/luci
@@ -46,6 +46,69 @@ function callPackageVersionCheck(pkg) {
 	return version;
 }
 
+function set_new_pwd(user, pwd) {
+	const ctx = cursor();
+	let cmd, fd, value;
+
+	cmd = sprintf('/usr/sbin/uhttpd -m %s', shellquote(pwd));
+	fd = popen(cmd);
+	value = trim(fd.read('line'));
+
+	fd.close();
+
+	ctx.foreach('rpcd', 'login', s => {
+		if (s.username == user) {
+			ctx.set('rpcd', s['.name'], 'password', value);
+			ctx.commit();
+		}
+	});
+}
+
+function check_oldpwd(user, pwd) {
+	let cmd, fd, value;
+	const ctx = cursor();
+	let valid = false;
+
+	cmd = sprintf('/usr/sbin/uhttpd -m %s', shellquote(pwd));
+	fd = popen(cmd);
+	value = trim(fd.read('line'));
+	fd.close();
+
+	ctx.foreach('rpcd', 'login', s => {
+		if (s.username == user && s.password == value)
+			valid = true;
+	});
+
+	return valid;
+}
+
+function check_user(source, entry) {
+	let found = false;
+
+	switch (source) {
+		case 'unix': {
+			const fd = open('/etc/passwd');
+
+			for (let line = fd.read('line'); length(line); line = fd.read('line')) {
+				let user = split(line, /:/)[0];
+				if (user == entry)
+					found = true;
+			}
+			break;
+		}
+
+		case 'rpcd': {
+			const ctx = cursor();
+			ctx.foreach('rpcd', 'login', s => {
+				if (s.username == entry)
+					found = true;
+				});
+		}
+		break;
+	}
+	return found;
+}
+
 const methods = {
 	getVersion: {
 		call: function(request) {
@@ -469,14 +532,58 @@ const methods = {
 	},
 
 	setPassword: {
-		args: { username: 'root', password: 'password' },
+		args: {
+			username: 'root',
+			password: 'password',
+			oldpassword: '',
+			rpcd: false,
+		},
 		call: function(request) {
-			const u = shellquote(request.args.username);
-			const p = shellquote(request.args.password);
+			const user = request.args.username;
+			const pwd = request.args.password;
+			const oldpwd = request.args.oldpassword;
+			const type = request.args.rpcd == true ? 'rpcd' : 'unix';
+			const known_user = check_user(type, user);
+			const uci = cursor();
 
-			return {
-				result: system(`(echo ${p}; sleep 1; echo ${p}) | /bin/busybox passwd ${u} >/dev/null 2>&1`) == 0
-			};
+			if (type == 'unix') {
+				const u = shellquote(user);
+				const p = shellquote(pwd);
+
+				return {
+					result: system(`(echo ${p}; sleep 1; echo ${p}) | /bin/busybox passwd ${u} >/dev/null 2>&1`) == 0
+				};
+			}
+
+			if (!access('/usr/sbin/uhttpd', 'x'))
+				return { 
+					result: 0 ,
+					msg: "uhttpd not installed"
+				};
+
+			if (known_user) {
+				if (!oldpwd)
+					return { result: 0 };
+
+				/*
+				* check if login is valid
+				*/
+				if (!check_oldpwd(user, oldpwd))
+					return { result: 0 };
+			} else {
+				/*
+				 * create user to keep logic of luci-app-acl
+				 */
+				const sid = uci.add('rpcd', 'login');
+				uci.set('rpcd', sid, 'username', user);
+				uci.commit('rpcd');
+			}
+
+			/*
+			 * encrypt password and assign it to rpcd login
+			 */
+			 set_new_pwd(user, pwd);
+			 return { result: 1 };
 		}
 	},
 
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
@@ -2,11 +2,17 @@
 'require view';
 'require dom';
 'require ui';
+'require fs';
+'require uci';
 'require form';
 'require rpc';
 
 var formData = {
-	password: {
+	data: {
+		rpc_user: null,
+		user: null,
+		rpcd: null,
+		oldpw: null,
 		pw1: null,
 		pw2: null
 	}
@@ -15,8 +21,8 @@ var formData = {
 var callSetPassword = rpc.declare({
 	object: 'luci',
 	method: 'setPassword',
-	params: [ 'username', 'password' ],
-	expect: { result: false }
+	params: [ 'username', 'password', 'oldpassword', 'rpcd' ],
+	expect: { result: 1 }
 });
 
 return view.extend({
@@ -40,13 +46,66 @@ return view.extend({
 		return true;
 	},
 
-	render: function() {
-		var m, s, o;
+	load: function() {
+		return Promise.all([
+			L.resolveDefault(fs.stat('/usr/sbin/uhttpd'), null),
+			fs.lines('/etc/passwd'),
+			uci.load('rpcd')
+		]);
+	},
 
-		m = new form.JSONMap(formData, _('Router Password'), _('Changes the administrator password for accessing the device'));
+	render: function([has_uhttpd, passwd]) {
+		var m, s, o, rpcd;
+
+		const known_unix_users = {};
+
+		for (let p of passwd) {
+			const parts = p.split(/:/);
+
+			if (parts.length >= 7)
+				known_unix_users[parts[0]] = true;
+		}
+
+		m = new form.JSONMap(formData, _('Password'), _('Changes the password for accessing the device as (rpcd) user'));
 		m.readonly = !L.hasViewPermission();
 
-		s = m.section(form.NamedSection, 'password', 'password');
+		s = m.section(form.NamedSection, 'data', 'data');
+
+		if (has_uhttpd) {
+			let logins = [];
+
+			uci.sections('rpcd', 'login', s => logins.push(s.username));
+
+			rpcd = s.option(form.Flag, 'rpcd', _('Change password for rpcd user'));
+			rpcd.default = false;
+
+			o = s.option(form.Value, 'rpc_user', _('rpcd username'));
+			for (let user of logins) {
+				if (user == 'root')
+					continue;
+
+				o.value(user, _('%s').format(user));
+			}
+			o.rmempty = false;
+			o.depends({ 'rpcd': '1' });
+
+			o = s.option(form.Value, 'user', _('Router username'));
+			for (let user in known_unix_users)
+				o.value(user, _('%s').format(user));
+			o.rmempty = false;
+			o.depends({ 'rpcd': '0' });
+
+			o = s.option(form.Value, 'oldpw', _('Old Password'));
+			o.password = true;
+			o.depends({ 'rpcd': '1' });
+		}
+
+		if (!has_uhttpd) {
+			o = s.option(form.Value, 'user', _('Router username'));
+			for (let user in known_unix_users)
+				o.value(user);
+			o.rmempty = false;
+		}
 
 		o = s.option(form.Value, 'pw1', _('Password'));
 		o.password = true;
@@ -72,22 +131,39 @@ return view.extend({
 		var map = document.querySelector('.cbi-map');
 
 		return dom.callClassMethod(map, 'save').then(function() {
-			if (formData.password.pw1 == null || formData.password.pw1.length == 0)
+			let rpc_user = formData.data.rpc_user;
+			let user = formData.data.user;
+			let rpcd = formData.data.rpcd;
+			let oldpw = formData.data.oldpw;
+
+			if (rpc_user && (oldpw == null || oldpw.length == 0))
 				return;
 
-			if (formData.password.pw1 != formData.password.pw2) {
+			if (formData.data.pw1 == null || formData.data.pw1.length == 0)
+				return;
+
+			if (formData.data.pw1 != formData.data.pw2) {
 				ui.addNotification(null, E('p', _('Given password confirmation did not match, password not changed!')), 'danger');
 				return;
 			}
 
-			return callSetPassword('root', formData.password.pw1).then(function(success) {
+			return callSetPassword(
+				rpc_user ? rpc_user : user,
+				formData.data.pw1,
+				oldpw ? oldpw : '',
+				rpcd ? true : false,
+			).then(function(success) {
 				if (success)
 					ui.addNotification(null, E('p', _('The system password has been successfully changed.')), 'info');
 				else
 					ui.addNotification(null, E('p', _('Failed to change the system password.')), 'danger');
 
-				formData.password.pw1 = null;
-				formData.password.pw2 = null;
+				formData.data.rpc_user = null;
+				formData.data.user = null;
+				formData.data.rpcd = null;
+				formData.data.pw1 = null;
+				formData.data.pw2 = null;
+				formData.data.oldpw = null;
 
 				dom.callClassMethod(map, 'render');
 			});
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
@@ -6,6 +6,7 @@
 'require uci';
 'require form';
 'require rpc';
+'require tools.password as pwtool';
 
 var formData = {
 	data: {
@@ -32,6 +33,12 @@ return view.extend({
 		    mediumRegex = new RegExp("^(?=.{7,})(((?=.*[A-Z])(?=.*[a-z]))|((?=.*[A-Z])(?=.*[0-9]))|((?=.*[a-z])(?=.*[0-9]))).*$", "g"),
 		    enoughRegex = new RegExp("(?=.{6,}).*", "g");
 
+		const pw_enabled = uci.get('rpcd', 'policy', 'enabled');
+		const pw_length = uci.get('rpcd', 'policy', 'pw_length');
+		const pw_digits = uci.get('rpcd', 'policy', 'digits');
+		const pw_ul = uci.get('rpcd', 'policy', 'uc_lc');
+		const special = uci.get('rpcd', 'policy', 'special_characters');
+
 		if (strength && value.length) {
 			if (false == enoughRegex.test(value))
 				strength.innerHTML = '%s: <span style="color:red">%s</span>'.format(_('Password strength'), _('More Characters'));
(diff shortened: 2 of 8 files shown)


systemcrash:
The only parameter which realistically improves security is length (bits).

Related: https://xkcd.com/936/

Seems OK. 

@hnyman @feckert @efahl ?

efahl:
Seems like an awful lot of code for something people should already know and be doing as a matter of course.  Who is the target audience?

ckorber:
@efahl I agree with you that people should be aware of this, but that probably doesn’t apply to everyone.

The target audience is system administrators who want to enforce a password policy company-wide. By default, these requirements would be disabled.

As for `setPassword`, it’s simply an extension to apply it in `luci-app-acl` and to include `rpcd user` in the administration as well.

hnyman:
Typically OpenWrt is administrator-only device  without other users than root.
Setting policy to yourself is nonsensical.
And password can also be changed from the ssh console, while this only tackles the LuCI way.

So, I do not really see the use case.

feckert:
Those who aren’t administrators actually always use LuCI to configure the system. I understand all the arguments that you really need to know what you’re doing. But by default, it would be **disabled** anyway. If we look at OpenWRT as a system for experts, I would argue the same way.

But in my/our case, we often get requests from customers asking why we don’t have this killer feature ;-). Every other system that uses OpenWRT patches LuCI in somehow to have this kind of requirement.

This is also a preparatory commit. After all, LuCI is, in principle, capable of handling multiple accounts. And I do think it’s useful to be able to specify how strong a password should be – whether that’s actually secure or not. It’s simply meant to prevent people from using ‘1234’ as a password. 
By the way that would then also be the root SSH password if you’re not using ssh-keys

efahl:
Could this be an optional plugin, so that it could be added to only those devices that need it, rather than burdening all of them?

systemcrash:
Good idea with the plugin bit. I’d have to think where to hook in, since I intended to augment dispatcher behaviour. And this seems to go a bit further, interface wise, so app seems the way to go. 

It’s probably better as what it is now.

@feckert do you see this sufficiently meeting those needs of users you mentioned?

ckorber:
I added it now as a plugin. And it seems to do the job.

systemcrash:
That was fast. In its current state, there are concerns about separation of responsibilities. Like you can uninstall the plugin, but the changes to the password page still refer to the plugin ID, and thus would still act after config UI removal. 

So abstractly, perhaps password policy can become a plugin hook point - and the plugin page attempts to process all plugins in the password-policy class. This would better happen at the ucode level, otherwise the UI JS becomes complex to parse all plugins of that class to implement a policy. As you might see, this seems a bit over-engineered, but it would nonetheless work.

ckorber:
I tried to implement it in the recent push. Please tell me if I did what you had in mind, @systemcrash .
Now it works like a plugin. But I am uncertain whether I overengineered the overengineered idea :wink:

ckorber:
ping @systemcrash

systemcrash:
The structure is not incorrect, but I question the usefulness of having (a whole class of) plugins to manage password policies - one win being as @efahl noted: so admins can choose. Is that not possible with a master on/off policy knob for the original implementation?

How many different properties of passwords do we want to mandate? entropy? Upper/lower case? Golden ratios and relationships to pi? With so few properties adding security to a policy aside from entropy, it's likely better to have it the way it was originally in the PR a la https://github.com/openwrt/luci/commit/2ca30de4c383d1a875fb7de384704a31ba1819e7

ckorber:
I thought about adding a feature how long a password is valid, but I also think that the first approach was smaller. I changed it to how it used to be and added an `enabled` flag.

Do you think this can be merged, @systemcrash ?

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -46,6 +46,67 @@ function callPackageVersionCheck(pkg) {
 	return version;
 }
 
+function set_new_pwd(user, pwd) {
```
systemcrash, line 49: These two functions differ only slightly - can they be combined?

Also, it's a good idea to check for uhttpd and bail early if absent - not everyone uses uhttpd, IIUC. Not sure what to do in that case....

ckorber: They differ only slightly, but the first sets a password, while the other simply prompts for one.
There is little point in combining them, as this would eliminate the separation of tasks.

Your point with uhttpd is valid. But in that case the user relies on cli not luci, right? I mean it is necessary for `luci-mod-system`.

I could exit early if uhttpd is not available.

ckorber:
@systemcrash : I return with failure on cli if uhttpd is not installed and rpcd should be changed. That is a valid return and should not break the existing workflow.

ckorber:
ping @systemcrash

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -469,14 +530,57 @@ const methods = {
...
-				result: system(`(echo ${p}; sleep 1; echo ${p}) | /bin/busybox passwd ${u} >/dev/null 2>&1`) == 0
-			};
+			if (type == 'unix') {
+				const u = shellquote(user);
+				const p = shellquote(pwd);
+
+				return {
+					result: system(`(echo ${p}; sleep 1; echo ${p}) | /bin/busybox passwd ${u} >/dev/null 2>&1`) == 0
+				};
+			}
```
systemcrash, line 556: Are we not better served by making a ubus call to this rpcd facility to set the password?

https://github.com/openwrt/rpcd/blob/4476687403d8c6e988eaa1ecc892e998838e6da9/sys.c#L434

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -46,6 +46,67 @@ function callPackageVersionCheck(pkg) {
...
+
+		case 'rpcd': {
+			const ctx = cursor();
+			ctx.foreach('rpcd', 'login', s => {
+				if (s.username == entry)
+					found = true;
+				});
+		}
+	}
+	return found;
```
openwrt-ai, line 109: The `switch` has no `break` between cases, so both `'unix'` and `'rpcd'` always execute regardless of `source`. Wrapping each case body in `{}` does not stop fallthrough in ucode.

Effect: when called with `source='unix'`, `check_user` also iterates `rpcd` logins; so a unix user that happens to share a name with an rpcd login (or vice versa) flips `found` incorrectly. In `setPassword`'s `if (known_user) { ... } else { uci.add('rpcd','login') ... }` branch this can silently route to the wrong path. Add a `break;` (or `return found;`) at the end of each case.

```suggestion
	switch (source) {
		case 'unix': {
			const fd = open('/etc/passwd');

			for (let line = fd.read('line'); length(line); line = fd.read('line')) {
				let user = split(line, /:/)[0];
				if (user == entry)
					found = true;
			}
			break;
		}

		case 'rpcd': {
			const ctx = cursor();
			ctx.foreach('rpcd', 'login', s => {
				if (s.username == entry)
					found = true;
				});
			break;
		}
	}
	return found;
}
```

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -46,6 +46,67 @@ function callPackageVersionCheck(pkg) {
...
+	fd = popen(cmd);
+	value = trim(fd.read('line'));
+	fd.close();
+
+	ctx.foreach('rpcd', 'login', s => {
+		if (s.username == user && s.password == value)
+			valid = true;
+	});
+
+	return valid;
```
openwrt-ai, line 82: This old-password check can never succeed for any user whose password has already been set.

`uhttpd -m <pwd>` calls [`crypt(optarg, "$1$")` in uhttpd's main.c](https://github.com/openwrt/uhttpd/blob/master/main.c#L472), which under glibc/musl generates a fresh random salt when the salt is just `"$1$"`. So `value` here is a new hash with a new salt each invocation, while `s.password` is the hash stored when the password was originally set (different salt). The string comparison `s.password == value` will essentially always be false, and `setPassword` then returns `{ result: 0 }` for known users with a non-empty stored password.

The correct verification is to re-hash the supplied old password using the *stored* hash as the salt and compare to the stored hash (the same scheme [rpcd uses in `rpc_login_test_password`](https://github.com/openwrt/rpcd/blob/master/session.c#L816)). Since rpcd already implements this end-to-end against a username/password pair on its `session.login` ubus call, calling it from here would avoid re-implementing crypt(3) handling in ucode entirely — and aligns with the suggestion already raised in this PR about reusing `system.password_set` from rpcd.

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -46,6 +46,67 @@ function callPackageVersionCheck(pkg) {
 	return version;
 }
 
+function set_new_pwd(user, pwd) {
+	const ctx = cursor();
+	let cmd, fd, value;
+
+	cmd = sprintf('/usr/sbin/uhttpd -m %s', pwd);
```
openwrt-ai: `pwd` is interpolated into a shell command via `popen()` without quoting. An authenticated caller with rpcd `setPassword` write permission can pass a password like `'; reboot; #` and execute arbitrary shell on the device. The `'unix'` branch of `setPassword` already routes through `shellquote()` for both user and password — the `'rpcd'` path should do the same. Same issue at the identical `popen('/usr/sbin/uhttpd -m %s', pwd)` call in `check_oldpwd` below.

```suggestion
	cmd = sprintf('/usr/sbin/uhttpd -m %s', shellquote(pwd));
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
```diff
@@ -37,16 +56,84 @@ return view.extend({
...
 
-	render: function() {
-		var m, s, o;
+	load: function() {
+		return Promise.all([
+			L.resolveDefault(fs.stat('/usr/sbin/uhttpd'), null),
+			fs.lines('/etc/passwd'),
+			uci.load('rpcd'),
+		]);
+	},
```
openwrt-ai: There are two `load:` keys on the same object literal (here and at line 30). JavaScript silently keeps only the last definition, so the first one (`uci.load('luci')`) never runs. This is the ESLint `no-dupe-keys` failure on this PR's [eslint job](https://github.com/openwrt/luci/actions/runs/26213292567/job/77129224444).

Because `uci.load('luci')` is dropped, `uci.get('luci', 'main', 'password_length')` in `handleSave` (line 170) always returns `undefined`, so the legacy `password_length` check is dead code. Merge both `load:` calls into one and either drop the legacy length check entirely (the policy already enforces length via `checkPassword`) or move `password_length` under the new `rpcd.policy` namespace used by the rest of the PR.

```suggestion
	load: function() {
		return Promise.all([
			L.resolveDefault(fs.stat('/usr/sbin/uhttpd'), null),
			fs.lines('/etc/passwd'),
			uci.load('rpcd'),
			uci.load('luci'),
		]);
	},
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
```diff
@@ -72,22 +159,46 @@ return view.extend({
...
+
+			if (rpc_user && (oldpw == null || oldpw.length == 0))
+				return;
+
+			var passwd_len = uci.get('luci', 'main', 'password_length') || "0";
+
+			if (formData.data.pw1.length < passwd_len || formData.data.pw2.length < passwd_len) {
+				ui.addNotification(null, E('p', _('The specified password does not have the permitted length!')), 'danger');
+				return;
+			}
```
openwrt-ai: `formData.data.pw1.length` and `formData.data.pw2.length` are dereferenced here, but the null-guard for `pw1`/`pw2` lives below at line 177. If the user clicks "Save" with an empty form, `pw1` is `null` and this throws `TypeError: Cannot read properties of null`. Move the length policy check after the existing null check, or drop it entirely — `checkPassword` is already wired up as `pw1.validate` and uses the new policy via `pwtool.checkLength`, so this hard-coded gate (which reads the orphaned `luci.main.password_length` option) duplicates that logic.

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -46,6 +46,67 @@ function callPackageVersionCheck(pkg) {
...
+	fd = popen(cmd);
+	value = trim(fd.read('line'));
+
+	fd.close();
```
openwrt-ai, line 57: `popen()` can return `null` (e.g. fork failure); `fd.read('line')` then throws and aborts the rpc handler with no useful result. The caller in `setPassword` does check `access('/usr/sbin/uhttpd', 'x')` before invoking this, but `set_new_pwd` is also called unconditionally on the "create user" branch, and `check_oldpwd` has no `access()` guard at all — invoking the rpcd setPassword with `rpcd=true` on a system without uhttpd reaches `check_oldpwd` and crashes the handler. Either move the `access('/usr/sbin/uhttpd', 'x')` check above the `check_user`/`known_user` logic, or null-check `fd` here and in `check_oldpwd`.

---

applications/luci-app-acl/Makefile
```diff
@@ -5,7 +5,7 @@ include $(TOPDIR)/rules.mk
 PKG_LICENSE:=Apache-2.0
 
 LUCI_TITLE:=LuCI account management module
-LUCI_DEPENDS:=+luci-base
+LUCI_DEPENDS:=+luci-base +luci-mod-system
```
openwrt-ai: nit: this makes the standalone `luci-app-acl` application drag in `luci-mod-system` purely because `acl.js` does `require tools.password as pwtool`, and that helper happens to live under `modules/luci-mod-system/htdocs/luci-static/resources/tools/`. If the password policy helpers are intended to be reused by other apps (acl is the first consumer), they belong in `luci-base` (`modules/luci-base/htdocs/luci-static/resources/tools/password.js`) — that's where shared LuCI helpers go and `luci-app-acl` already depends on `luci-base`. Otherwise consider inlining the four small checks in `acl.js` so this dependency reversal isn't needed.

---

openwrt-ai (commented):
## Commit checks
- b4ddef8734 "luci-mod-system: enforce password policy" — touches `applications/luci-app-acl/Makefile`, which is out of scope for a `luci-mod-system` commit. The Makefile dependency bump belongs with the next commit (8cb78b783f "luci-app-acl: enforce password policy") that actually adds the `tools.password` import in acl.js.

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
ckorber: It's been that way since ded8ccf93ec5163be35c41501869110e5dab30d1.
I just wanted to keep the functionality. It's doable, but is it worth the effort?

modules/luci-base/root/usr/share/rpcd/ucode/luci
ckorber: `uhttpd -m <pwd>` creates in my setup reproducible hashes to the set `oldpwd`, so the first part is not true.

`known_user` is in this case from `rpcd`, not `/etc/passwd` and `system.password_set` applies to unix users as far as I saw that.

modules/luci-base/root/usr/share/rpcd/ucode/luci
ckorber: That is a valid point!

modules/luci-base/root/usr/share/rpcd/ucode/luci
ckorber: This is not correct! If there is no `uhttpd` the function returns and `check_oldpwd` is never called. `access('/usr/sbin/uhttpd', 'x')` already guards against that.

applications/luci-app-acl/Makefile
ckorber: Moved the helper functions.

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -46,6 +46,69 @@ function callPackageVersionCheck(pkg) {
...
+	const ctx = cursor();
+	let valid = false;
+
+	cmd = sprintf('/usr/sbin/uhttpd -m %s', shellquote(pwd));
+	fd = popen(cmd);
+	value = trim(fd.read('line'));
+	fd.close();
+
+	ctx.foreach('rpcd', 'login', s => {
+		if (s.username == user && s.password == value)
```
openwrt-ai, line 78: `uhttpd -m` runs `crypt(optarg, "$1$")`, which generates a fresh random salt on every invocation (see [`auth.c` crypt usage](https://github.com/openwrt/uhttpd/blob/1b624f8f814ed568608d756512892416e0431d77/auth.c#L191)). So `value` here is a `$1$<randomsalt>$...` hash that will essentially never be byte-for-byte equal to the stored `s.password`, and `check_oldpwd` returns `false` for the correct old password.

To verify a password you must crypt the candidate using the *stored hash* as the salt setting, the same way uhttpd does (`crypt(pass, realm->pass)`), then compare. Re-hashing with `$1$` and string-comparing cannot work.

Consequence: the existing-rpcd-user change path is broken — and `acl.js` calls `setPassword(user, value, '', true)` with an empty old password, so editing any existing ACL user's password now hits `if (!oldpwd) return { result: 0 }` and fails too.

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -469,14 +532,57 @@ const methods = {
...
+				* check if login is valid
+				*/
+				if (!check_oldpwd(user, oldpwd))
+					return { result: 0 };
+			} else {
+				/*
+				 * create user to keep logic of luci-app-acl
+				 */
+				const sid = uci.add('rpcd', 'login');
+				uci.set('rpcd', sid, 'username', user);
```
openwrt-ai, line 578: The new `login` section is added on this `uci` cursor but never committed, and `set_new_pwd()` below creates its own fresh `cursor()` whose `foreach('rpcd', 'login', ...)` cannot see the uncommitted section. As a result, when `setPassword` is asked to create a new rpcd user (the unknown-user branch used by luci-app-acl), the username row is written but the password is never set. Commit this cursor before calling `set_new_pwd()` (or pass the section id/cursor into it), otherwise the new account ends up with no password.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
```diff
@@ -72,22 +154,46 @@ return view.extend({
...
+			let oldpw = formData.data.oldpw;
+
+			if (rpc_user && (oldpw == null || oldpw.length == 0))
+				return;
+
+			var passwd_len = uci.get('luci', 'main', 'password_length') || "0";
+
+			if (formData.data.pw1.length < passwd_len || formData.data.pw2.length < passwd_len) {
+				ui.addNotification(null, E('p', _('The specified password does not have the permitted length!')), 'danger');
+				return;
```
openwrt-ai: `luci.main.password_length` is never written anywhere in this PR (the policy lives under `rpcd.policy.*`), so `passwd_len` is always the `"0"` default and this branch is dead code. It also duplicates the policy length check already enforced via `pwtool.checkLength` in the field validator. Recommend dropping this block and relying on the policy validation instead.

---

openwrt-ai (commented):
The branch was rebased and the commits reworked since my last review; the earlier scope concern (acl Makefile in the luci-mod-system commit) is resolved. Reviewed the full reworked series.

## Commit checks
- 18f3af786b "luci-mod-system: enforce password policy" — adds `modules/luci-base/htdocs/luci-static/resources/tools/password.js`, a luci-base file, under a `luci-mod-system:` subject. The shared password tool belongs to a `luci-base:` commit; consider splitting it out so the subject prefix matches the touched package.

The inline comments flag two functional bugs in the `setPassword` rework (old-password verification compares non-deterministic crypt outputs; a newly-created rpcd login is never committed so its password is silently dropped) plus a dead `luci.main.password_length` validation block.

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
ckorber: Not my observation

openwrt-ai (commented):
Reviewed the changes since the last review (`ec769ce`). The two new edits resolve the prior findings: the dead `luci.main.password_length` block was dropped from `password.js`, and the freshly-created rpcd `login` section is now committed (`uci.commit('rpcd')`) before `set_new_pwd()` runs. Commit messages match their changes.

On the remaining `check_oldpwd` thread: re-checking [`crypt(optarg, "$1$")` in uhttpd main.c:472](https://github.com/openwrt/uhttpd/blob/1b624f8f814ed568608d756512892416e0431d77/main.c#L472), passing the fixed setting `"$1$"` (no salt characters) is deterministic under musl, so `s.password == value` does match when the stored hash was itself written via `uhttpd -m` — my earlier "fresh random salt" concern does not hold on the OpenWrt target. Withdrawing that finding.

No new issues found.

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
systemcrash: Might we sidestep the uhttpd things (which avoids problems for users who run a different proxy)? If so it seems like a positive. It seems like a miss not to use a facility built for exactly this purpose, although I find your current approach acceptable, since I'm not sure how your machinery would change when using the rpcd method. If you're happy, let's get this merged.

ckorber:
> Might we sidestep the uhttpd things (which avoids problems for users who run a different proxy)? If so it seems like a positive. It seems like a miss not to use a facility built for exactly this purpose, although I find your current approach acceptable, since I'm not sure how your machinery would change when using the rpcd method. If you're happy, let's get this merged.

@systemcrash : Yes, I also don't know if that changes behaviour. Thank you for your review!

systemcrash:
Merged. Thanks @ckorber !

Issue #8256, opened 2026-01-24
luci-app-attendedsysupgrade: unoptimized js lurks around


brada4:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### screenshots or captures

```
opkg search  /www/luci-static/resources/view/system/repokeys.js.
o
luci-mod-system - 26.021.66732~4b823e3
```

### Actual behaviour

unoptimized file is served to users while optimized is dropped next to it without any purpose

### Expected behaviour

no surplus files thrown around

### Steps to reproduce

```
find /www/ -type f -name \*.js.o
```

### Additional Information

```shell
{
        "kernel": "6.6.119",
        "hostname": "OpenWrt",
        "system": "Common KVM processor",
        "model": "QEMU Standard PC (Q35 + ICH9, 2009)",
        "board_name": "qemu-standard-pc-q35-ich9-2009",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "24.10.5",
                "revision": "r29087-d9c5716d1d",
                "target": "x86/64",
                "description": "OpenWrt 24.10.5 r29087-d9c5716d1d",
                "builddate": "1766005702"
        }
}
```

### What browsers do you see the problem on?

Microsoft Edge

### Relevant log output

```shell

```

systemcrash:
Huh. Fun.

brada4:
jobserver race?

brada4:
> JSMIN Error: Unterminated set in Regular Expression literal.

brada4:
https://github.com/openwrt/luci/blob/a52d779164ad1180aaadbc4e0737a65d4e3cf449/luci.mk#L256

```
++ || rm file.o
```

efahl:
Ha, I was just looking at that makefile yesterday and thought the same thing.  

(I'm still trying to figure out the root cause of the missing semicolon working for so long in the ASU include file, which led me there.)

efahl:
AHA.  On 24.10 device, jsmin failed:
```
$ ll /www/luci-static/resources/view/status/include/11*
-rw-r--r--    1 root     root          6153 Dec 17 13:08 /www/luci-static/resources/view/status/include/11_upgrades.js
-rw-r--r--    1 root     root          1967 Dec 17 13:08 /www/luci-static/resources/view/status/include/11_upgrades.js.o
```

brada4:
i am exploring attaching input line number to jsmin error ;-)

[...]

PR #8593, opened 2026-04-27
luci-app-package-manager: handle versioned provides from apk query output and some rework for performance and correctness

### LINKED ISSUE #8563: luci-app-package-manager: incorrectly reports missing dependency "uci-firewall" under apk-based OpenWrt 25.12

### screenshots or captures

Image

### Actual behaviour

LuCI Package Manager reports that "luci-app-firewall" depends on
"uci-firewall", which is shown as missing.

However, this dependency is NOT actually missing. It is a virtual
dependency provided by firewall4-any via APK PROVIDES mechanism.

- LuCI UI shows:
  "uci-firewall required but not available"

- This blocks or warns upgrade/install operations in web UI.

Relevant logs:

```
root@OpenWrt:~# cat /lib/apk/db/installed | grep -A20 luci-app-firewall
P:luci-app-firewall
V:26.106.72997~485787b
A:noarch
S:18703
I:78019
T:Firewall and port forwarding application
U:https://github.com/openwrt/luci
L:Apache-2.0
o:feeds/luci/feeds/luci/applications/luci-app-firewall
m:OpenWrt LuCI community
D:libc luci-base uci-firewall
p:luci-app-firewall-any
f:S
F:lib
F:lib/apk
F:lib/apk/packages
R:luci-app-firewall.list
Z:Q1pWdCtFoQLOepwRI4cjipeJiy/0Y=
F:usr
F:usr/share
F:usr/share/luci
F:usr/share/luci/menu.d
R:luci-app-firewall.json
Z:Q1avXsw/kkOQClXMpbhiVlbj+NYxg=
F:usr/share/rpcd
F:usr/share/rpcd/acl.d
R:luci-app-firewall.json
Z:Q1bMyj48fE59/Dulg2/hG/jw94bo4=
F:usr/share/ucitrack
R:luci-app-firewall.json
Z:Q1jij9uohlSnXBgHdLYW1576BNmPQ=
F:www
F:www/luci-static
F:www/luci-static/resources
F:www/luci-static/resources/tools
R:firewall.js
Z:Q1T0kcjrKQsjHxzqE24yUQOzb+ijU=
F:www/luci-static/resources/view
F:www/luci-static/resources/view/firewall
R:custom.js
Z:Q16YpJFQ/l0Laa2rf+U7JOdxz1Uec=
R:forwards.js
Z:Q199xZMPdeeOqF15mRyLMoZmGwthk=
R:ipsets.js
Z:Q1DvG3BTzrnAdnusQSS3NeZeqjMF0=
R:rules.js
Z:Q12OgAtioF6P7de4g0C8XihdBxajc=
R:snats.js
Z:Q1J1w9ISjOuSDh4BVHKYK4hOIVxg4=
R:zones.js
--
T:Translation for luci-app-firewall - Русский (Russian)
U:
L:Apache-2.0
o:feeds/luci/feeds/luci/applications/luci-app-firewall
m:Jo-Philipp Wich <jo@mein.io>
D:libc luci-app-firewall
p:luci-i18n-firewall-ru-any
f:S
F:etc
F:etc/uci-defaults
R:luci-i18n-firewall-ru
Z:Q1dmyo6HiZIjP2W7TgvmZcL3Qay1I=
F:lib
F:lib/apk
F:lib/apk/packages
R:luci-i18n-firewall-ru.list
Z:Q1uiiqXKtXHQF0TiFAoJEiZeusbKY=
F:usr
F:usr/lib
F:usr/lib/lua
F:usr/lib/lua/luci
F:usr/lib/lua/luci/i18n
R:firewall.ru.lmo
Z:Q10OTGhBmzPQ1wVdHee/lte3tb74U=

C:Q1K+FzqXaKyGScl4rzB7i+yS4ZWBI=
--
D:libc luci-app-firewall luci-mod-admin-full luci-proto-ipv6 luci-proto-ppp luci-theme-bootstrap rpcd-mod-rrdns uhttpd uhttpd-mod-ubus
p:luci-light-any
f:S
F:lib
F:lib/apk
F:lib/apk/packages
R:luci-light.list
Z:Q147DEQpj8HBSa+/TImW+5JCeuQeQ=

C:Q1YN9CinfJlUuM40yfCXa9L7NLltk=
P:luci-lua-runtime
V:26.101.22673~0c81d2d
A:aarch64_cortex-a53
S:31506
I:132246
T:LuCI Lua runtime libraries
U:https://github.com/openwrt/luci
L:Apache-2.0
o:feeds/luci/feeds/luci/modules/luci-lua-runtime
m:OpenWrt LuCI community
D:libc liblucihttp-lua libubus-lua lua luci-base luci-lib-base luci-lib-ip luci-lib-jsonc luci-lib-nixio ucode-mod-lua
root@OpenWrt:~#
```

```
root@OpenWrt:~# grep -R "uci-firewall" /lib/apk/db/
/lib/apk/db/installed:p:firewall4-any uci-firewall=2025.03.17~b6e51575-r2
/lib/apk/db/installed:D:libc luci-base uci-firewall
root@OpenWrt:~# apk search -d uci-firewall
firewall-2025.10.03~3a65fde5-r2
firewall4-2025.03.17~b6e51575-r2
root@OpenWrt:~#
```

### Expected behaviour

LuCI should resolve APK PROVIDES dependencies correctly, i.e.:

- "uci-firewall" is provided by "firewall4-any"
- dependency should be considered satisfied
- no error/warning should be shown

### Steps to reproduce

1. If you already has `luci-app-firewall` installed, just go to http://router_ip/cgi-bin/luci/admin/system/package-manager and try to update, for example, `luci` package.

### Additional Information

```shell
{
        "kernel": "6.12.74",
        "hostname": "OpenWrt",
        "system": "ARMv8 Processor rev 4",
        "model": "GL.iNet GL-MT6000",
        "board_name": "glinet,gl-mt6000",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "25.12.2",
                "revision": "r32802-f505120278",
                "target": "mediatek/filogic",
                "description": "OpenWrt 25.12.2 r32802-f505120278",
                "builddate": "1774469393"
        }
}
```

Firefox

### Relevant log output

```shell

```

### LINKED ISSUE #8581: luci-app-vnstat2: library not available

### screenshots or captures

Image

### Actual behaviour

on OpenWrt 25.12.2 r32802-f505120278 / LuCI openwrt-25.12 branch 26.112.70337~8405d22, the missing package doesn't allow to install

### Expected behaviour

library missing

### Steps to reproduce

try to install

### Additional Information

```shell
OpenWrt 25.12.2 r32802-f505120278 / LuCI openwrt-25.12 branch 26.112.70337~8405d2
```

_No response_

### Relevant log output

```shell

```

### LINKED ISSUE #23004: can not find uci-firewall !
Image

The right Name is luci-firewall.

### LINKED ISSUE #23136: missing package ip when installing wireguard-tools on openwrt 25.12.2
### Describe the bug

When trying to install wireguard I get the error that the dependent package ip is not available.
I think this is because the package has been split into it-full and ip-tiny and there should be a meta-package for ip that solves this dependency. I have installed ip-full but wireguard does not start, but the ip command now works.

### OpenWrt version

r32802-f505120278

### OpenWrt release

25.12.2

### OpenWrt target/subtarget

ramips/mt7621

### Device

Zyxel WSM20

### Image kind

Official downloaded image

### Steps to reproduce

Image

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### Additional info

_No response_

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

### LINKED ISSUE #23286: luci-app-ddns not available in 25.12.3
After upgrading to OpenWrt version `25.12.3`, I noticed that the `luci-app-ddns` package was not working in the router settings. When trying to install it manually, I see that all dependencies for ddns-scripts are already installed, but when trying to install `ddns-scripts`, it complains that the `ip` dependency cannot be installed because it was not found.

Image

Has this package been deprecated in favor of another one, or will it be updated in the future? In version `25.12.2` it was installed and worked normally.

### LINKED ISSUE #23308: Gawk 5.3.2-r1 dependancy error for libncursesw6
### Describe the bug

Try to install gawk (or anything that depends on gawk) via luci and an error is indicated that libncursesw6 is not available:

### OpenWrt version

r32912-6639b15f62

### OpenWrt release

25.12.3

### OpenWrt target/subtarget

qualcommax/ipq60xx

### Device

Linksys MR7350

### Image kind

Official downloaded image

### Steps to reproduce

From luci menu 

System->Software 

and enter gawk into filter input. Then select install for:

gawk	5.3.2-r1	288.93 KiB	GNU awk

### Actual behaviour

I am presented with:

Details for package gawk

    Version: 5.3.2-r1
    Size: ~288.93 KiB compressed
    Dependencies:
        libncursesw6 Not available
        libreadline8 (~144.71 KiB) Not installed

Description

GNU awk

    Required dependency package libncursesw6 is not available in any repository.

### Expected behaviour

No dependency error should exist

### Additional info

No error generated when installing gawk from cli:

root@OpenWrt:~# apk add gawk
(1/4) Installing terminfo (6.4-r3)
  Executing terminfo-6.4-r3.post-install
(2/4) Installing libncurses6 (6.4-r3)
  Executing libncurses6-6.4-r3.post-install
(3/4) Installing libreadline8 (8.3-r1)
  Executing libreadline8-8.3-r1.post-install
(4/4) Installing gawk (5.3.2-r1)
  Executing gawk-5.3.2-r1.post-install
  * add alternative: /usr/bin/awk -> /usr/bin/gawk
OK: 28.4 MiB in 182 packages

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

### LINKED ISSUE #29416: ddns-scripts: required ip packages dependency is not available
### Package Name

ddns-scripts

### Maintainer

(@feckert)

### OpenWrt Version

25.12.3

### OpenWrt Target/Subtarget

ramips/mt7621

### Steps to Reproduce

1. install  `luci-app-ddns` from gui
2. package's dependency tree shows `ip` package is unavailable for `ddns-scripts` package
3. `apk info ip` found `ip-full` and `ip-tiny` packages

### Actual Behaviour

Required dependency package `ip` is not available in any repository.

### Confirmation Checklist

- [x] The package is maintained in this repository.
- [x] I understand that issues related to [the base OpenWrt repository](https://github.com/openwrt/openwrt/issues) or [LuCI repository](https://github.com/openwrt/luci/issues) will be closed.
- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

map-b:
### Description

'apk query --format json...' outputs the complete provides fields including the version when corresponding.

```
"provides": [
      "firewall4-any",
      "uci-firewall=2025.03.17~b6e51575-r2"
    ],
```
    
Versus the previous 'apk info --full'

```
Package: firewall4
    Version: 2025.03.17~b6e51575-r2
    Depends: kmod-nft-core, kmod-nft-fib, kmod-nft-nat, kmod-nft-offload, libc, nftables-json, ucode, ucode-mod-fs, ucode-mod-ubus, ucode-mod-uci
    Provides: firewall4-any, uci-firewall
```

Strip the `'=<version>'` suffix before gathering the providers chain. This was causing some incorrent dependecy issues (ie. uci-firewall not detected as satisfied) in the luci package-manager app.

Fixes: https://github.com/openwrt/luci/issues/8563
Fixes: https://github.com/openwrt/luci/issues/8581
Fixes: https://github.com/openwrt/openwrt/issues/23004
Fixes: https://github.com/openwrt/openwrt/issues/23136
Fixes: https://github.com/openwrt/openwrt/issues/23286
Fixes: https://github.com/openwrt/openwrt/issues/23308
Fixes: https://github.com/openwrt/packages/issues/29416
Fixes: 1624418f64db1bef1ec8e5fafd42bec63cd0adc0 ("package-manager: migrate to apk-tools JSON API")

---------------------
Changes from @Konstantin-Glukhov:

Rework renderDependencies: improve provider resolution performance and complete the RegExp for accounting for all possible APK dependency operators.

---

## Tested on

**OpenWrt version:** OpenWrt 25.12.3
**LuCI version:** 
**Web browser(s):** Firefox 150.0

---

## Checklist

### DIFF
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -525,8 +525,8 @@ function versionSatisfied(ver, ref, vop)
 
 function pkgStatus(pkg, vop, ver, info)
 {
-	info.errors = info.errors || [];
-	info.install = info.install || [];
+	info.errors ??= [];
+	info.install ??= new Set();
 
 	if (isPkgInstalled(pkg)) {
 		if (vop && !versionSatisfied(pkg.version, ver, vop)) {
@@ -538,7 +538,7 @@ function pkgStatus(pkg, vop, ver, info)
 			});
 
 			if (repl) {
-				info.install.push(repl);
+				info.install.add(repl);
 				return E('span', {
 					'class': 'label',
 					'data-tooltip': _('Requires update to %h %h')
@@ -559,7 +559,7 @@ function pkgStatus(pkg, vop, ver, info)
 	}
 	else if (!pkg.missing) {
 		if (!vop || versionSatisfied(pkg.version, ver, vop)) {
-			info.install.push(pkg);
+			info.install.add(pkg);
 			return E('span', { 'class': 'label' }, _('Not installed'));
 		}
 
@@ -596,12 +596,15 @@ function renderDependencyItem(dep, info, flat)
 
 		let text = pkg.name;
 
+		if (vop && ver)
+			text += ' (%s %s)'.format(vop, ver);
+
 		if (pkg.installsize)
 			text += ' (%1024mB)'.format(pkg.installsize);
 		else if (pkg.size)
 			text += ' (~%1024mB)'.format(pkg.size);
 
-		li.appendChild(E('span', { 'data-tooltip': pkg.description },
+		li.appendChild(E('span', { 'data-tooltip': (pkg.description || '') + (vop && ver && pkg.version ? '\nAvailable: %s'.format(pkg.version) : '') || null },
 			[ text, ' ', pkgStatus(pkg, vop, ver, info) ]));
 
 		(pkg.depends || []).forEach(function(d) {
@@ -647,7 +650,7 @@ function renderDependencies(depends, info, flat)
 		//   ([^)]+)?         [3] optional version string
 		//   \)?$              optional closing paren + end
 		const m = deps[i].match(/^([^><=~\s]+)\s?\(?([><=~]+)?\s?([^)]+)?\)?$/);
-		if (!m || info.seen[m[1]])
+		if (!m || info.seen[m[0]])
 			continue;
 
 		// Provider cache (pC): dep name -> Set of package names (incrementally populated)
@@ -661,13 +664,17 @@ function renderDependencies(depends, info, flat)
 					pC[m[1]].add(p.name);
 		}
 
-		info.seen[m[1]] = {
+		// Index by full string so entries with different constraints on
+		// the same package each get their own line:
+		//   "ucode (>= 2022.03.22)" vs "ucode"
+		//   "dovecot2.3.21>=2.3.0" vs "dovecot2.3.21<2.4.0"
+		info.seen[m[0]] = {
 			name:    m[1],
 			pkgs:    [...pC[m[1]]],
 			version: [m[2] || null, m[3] || null]
 		};
 
-		items.push(renderDependencyItem(info.seen[m[1]], info, flat));
+		items.push(renderDependencyItem(info.seen[m[0]], info, flat));
 	}
 
 	if (items.length)
@@ -713,7 +720,7 @@ function handleInstall(ev)
 	const deps = renderDependencies(pkg.depends, depcache);
 	let tree = null, errs = null, inst = null, desc = null;
 
-	if (depcache.errors && depcache.errors.length) {
+	if (depcache.errors?.length) {
 		errs = E('ul', { 'class': 'errors' });
 		depcache.errors.forEach(function(err) {
 			errs.appendChild(E('li', {}, err));
@@ -724,7 +731,7 @@ function handleInstall(ev)
 	let totalpkgs = 1;
 	let suggestsize = 0;
 
-	if (depcache.install && depcache.install.length)
+	if (depcache.install?.size)
 		depcache.install.forEach(function(ipkg) {
 			totalsize += ipkg.installsize || ipkg.size || 0;
 			totalpkgs++;
@@ -763,7 +770,7 @@ function handleInstall(ev)
 
 			i18n_tree = renderDependencies(i18n_packages, i18ncache, true);
 
-			if (i18ncache.install && i18ncache.install.length) {
+			if (i18ncache.install?.size) {
 				i18ncache.install.forEach(function(ipkg) {
 					suggestsize += ipkg.installsize || ipkg.size || 0;
 				});
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -624,54 +624,50 @@ function renderDependencyItem(dep, info, flat)
 	return li;
 }
 
+let pC = null;
+
 function renderDependencies(depends, info, flat)
 {
 	const deps = depends || [];
 	const items = [];
+	info.seen ??= Object.create(null);
 
-	info.seen = info.seen || [];
+	pC ??= Object.create(null);
 
 	for (let i = 0; i < deps.length; i++) {
-		let dep, vop, ver;
-
 		if (deps[i] === 'libc')
 			continue;
 
-		// This regex handles "name", "name>=ver", and "name (>=ver)"
-		const match = deps[i].match(/^([^><=~\s]+)\s*\(?([><=~]+)?\s*([^)]+)?\)?$/);
-
-		if (match) {
-			// Destructure the match array: [full_match, name, operator, version]
-			const [, matchedDep, matchedVop, matchedVer] = match;
-			dep = (matchedDep || '').trim();
-			vop = (matchedVop || '').trim() || null;
-			ver = (matchedVer || '').trim() || null;
-		} else {
-			// Fallback if the string is just a plain name with no operators
-			dep = deps[i].trim();
-			vop = ver = null;
-		}
-
-		if (info.seen[dep])
+		// This regex handles "name", "name>=ver", "name (>= ver)" with operators: > < = ~ >= <=
+		// Regex (capture groups numbered):
+		//   ^([^><=~\s]+)    [1] package name
+		//   \s?\(?            optional space + optional paren
+		//   ([><=~]+)?       [2] optional operators: > < = ~ >= <=
+		//   \s?               optional space before version
+		//   ([^)]+)?         [3] optional version string
+		//   \)?$              optional closing paren + end
+		const m = deps[i].match(/^([^><=~\s]+)\s?\(?([><=~]+)?\s?([^)]+)?\)?$/);
+		if (!m || info.seen[m[1]])
 			continue;
 
-		const pkgs = [];
+		// Provider cache (pC): dep name -> Set of package names (incrementally populated)
+		//   .installed.providers[ip] = [ ip, ip-tiny ]
+		//   .available.providers[ip]  = [ ip, ip-full ]
+		//   pC[ip]                    = Set( ip, ip-tiny, ip-full )
+		if (!pC[m[1]]) {
+			pC[m[1]] = new Set();
+			for (const src of [packages.installed.providers, packages.available.providers])
+				for (const p of (src[m[1]] || []))
+					pC[m[1]].add(p.name);
+		}
 
-		(packages.installed.providers[dep] || []).forEach(function(p) {
-			if (pkgs.indexOf(p.name) === -1) pkgs.push(p.name);
-		});
-
-		(packages.available.providers[dep] || []).forEach(function(p) {
-			if (pkgs.indexOf(p.name) === -1) pkgs.push(p.name);
-		});
-
-		info.seen[dep] = {
-			name:    dep,
-			pkgs:    pkgs,
-			version: [vop, ver]
+		info.seen[m[1]] = {
+			name:    m[1],
+			pkgs:    [...pC[m[1]]],
+			version: [m[2] || null, m[3] || null]
 		};
 
-		items.push(renderDependencyItem(info.seen[dep], info, flat));
+		items.push(renderDependencyItem(info.seen[m[1]], info, flat));
 	}
 
 	if (items.length)
@@ -1152,6 +1148,7 @@ function updateLists(data)
 
 	packages.available = { providers: {}, pkgs: {} };
 	packages.installed = { providers: {}, pkgs: {} };
+	pC = null;
 
 	return (data ? Promise.resolve(data) : downloadLists()).then(function(data) {
 		const pg = document.querySelector('.cbi-progressbar');


Konstantin-Glukhov:
I think it is better to fix the logic in `renderDependencies(depends, info, flat)`
Here is the proposed change:

```
let providerCache = null;

function renderDependencies(depends, info, flat)
{
	const deps = depends || [];
	const items = [];

	const PKG_PATTERN = new RegExp(
		/^([^><=~\s]+)/.source + // Name stops at space or operator
		/\s*\(?\s*/.source +     // Optional space and opening paren
		/([><=~]+)?/.source +    // Optional comparison symbols
		/\s*/.source +           // Optional space
		/([^)]+)?/.source +      // Optional Version
		/\s*\)?$/.source         // Optional closing paren and end of string
	);

	function getProviderMap() {
		// Only build the map once (Lazy Loading)
		if (providerCache) return providerCache;

		providerCache = {};

		function index(source) {
			if (!source) return;
			for (const key in source) {
				const m = key.match(PKG_PATTERN);
				const baseName = m ? m[1] : key;
				if (!providerCache[baseName]) providerCache[baseName] = new Set();
				for (const p of source[key]) providerCache[baseName].add(p.name);
			}
		}

		index(packages.installed.providers);
		index(packages.available.providers);

		return providerCache;
	}

	const activeMap = getProviderMap();

	// Use an object {} for key-based lookups (faster than an array [])
	info.seen = info.seen || {};

	for (let i = 0; i < deps.length; i++) {
		// Initialize: vop/ver are optional (null), dep is mandatory (undefined)
		let dep, vop = null, ver = null;

		if (deps[i] === 'libc')
			continue;

		const match = deps[i].match(PKG_PATTERN);
		if (match) {
			dep = match[1];
			vop = match[2] || null;
			ver = match[3] ? match[3].trim() : null;
		} else {
			// Fallback if the string is just a plain name with no operators
			dep = deps[i].trim();
		}

		// Skip if name is invalid or already processed
		if (!dep || info.seen[dep])
			continue;

		info.seen[dep] = {
			name:    dep,
			pkgs:    Array.from(activeMap[dep] || []),
			version: [vop, ver]
		};

		items.push(renderDependencyItem(info.seen[dep], info, flat));
	}

	if (items.length)
		return E('ul', { 'class': 'deps' }, items);

	return null;
}
```

Konstantin-Glukhov:
I have submitted a PR to this branch to address critical logic and performance gaps in the current draft.
The Logic Fix:
The current implementation fails to link dependencies when APK provider strings include versioning or specific operators (e.g., uci-firewall=2025.03.17~...). While a simple string split at the population stage was suggested, my approach preserves the original metadata while using a robust regex-based lookup. This correctly handles the variety of APK operators (=, >=, ~) and optional parentheses without being destructive to the source data.
The Performance Fix:
I have implemented a lazy-loaded global provider cache. Instead of iterating through the entire provider database for every dependency check, it pre-indexes the data once. This ensures that dependency lookups are near-instant, preventing UI lag as the package list grows.
I believe these changes are necessary to make the APK-based dependency rendering stable and performant for production.

map-b:
> The current implementation fails to link dependencies when APK provider strings include versioning or specific operators (e.g., uci-firewall=2025.03.17~...). While a simple string split at the population stage was suggested, my approach preserves the original metadata while using a robust regex-based lookup. This correctly handles the variety of APK operators (=, >=, ~) and optional parentheses without being destructive to the source data.

Just note that in the provides cases(virtual or versioned) are different to dependencies, only the '=<version>' casuistic is expected, no other APK operators. And maybe it's acceptable the "destructive" approach since is not expected that that prefix to be useful at all in this context?

Totally agree in the performance refactor although the logic ends up a bit ofuscated ;) but that's just a unskilled human limitation :). Maybe merge the fix and go ahead with your proposal? ~~I haven't fully tested this yet...🤭~~

Konstantin-Glukhov:
I think this is ready for review

efahl:
Is there some way to eliminate the cache and reduce the code size?  Code size if far more important that "performance" on OpenWrt devices.

map-b:
Since there is no Claude review in this repo, just though in spending the little free vscode quota before mid-month expiration. The little specimen: Raptor mini ( was auto selected)

imagen

imagen

@efahl No opinion but some questions, the cache itself would be only the first block and more or less is the same as the two foreach, at the end, that it replaces? What is the measure, LoC, total code length since the code is minimified? just the final package size? would be that versus the performace gained the two metrics to account for?

Should we give some numbers through making some benchmark?

Konstantin-Glukhov:
> Is there some way to eliminate the cache and reduce the code size? Code size if far more important that "performance" on OpenWrt devices.

There are 47 additions & 22 deletions, so the net increase is only 25 lines.
That said, two options worth considering if size is still a concern:

- Removing the opkg section would shrink it further
- The whole file could be refactored employing modern JS, which might help as well

efahl:
> Removing the opkg section would shrink it further

That was my original hope, but we need to wait a release for that.

I'm for anything that slows the tide of growth (we'll never stop it, but we can pay attention to how fast it's moving).  Things like https://github.com/openwrt/openwrt/commit/bf9e4ed9451745781d1027a33e62da8a57461dc6 should be stalled as long as possible (that's an 8/64 device).

systemcrash:
SOP is to maintain master so picks to branches are clean - opkg will be dropped when 24.10 goes out of support (currently 24ish months).

I'm not fully up to speed here but IIUC this basically makes apk handle JSON cleanly, and perhaps some other perks?

efahl:
> I'm not fully up to speed here but IIUC this basically makes apk handle JSON cleanly, and perhaps some other perks?

Sort of the other way around: `apk` has very good native support for output in JSON, which makes handling it directly in LuCI much simpler and robust.  The old "parse some semi-structured text" scheme used with `opkg`, and adapted into an OpenWrt-only patch on `apk list`, was brittle as are all such solutions.  (For an example, run `apk query --format json --fields all 'apk*'` sometime.)

So, that previous PR that reworked the package manager app should ultimately result in a much smaller footprint in LuCI js code, simultaneously improving its reliability.  Being able to gut out the `opkg` support will clean up the code considerably, hopefully in the 26.x release (or maybe just put current `luci-app-package-manager` on a 24.10 branch sooner and do the drop in 25.12).

map-b:
Should I push these changes? Only the cache reset (edge case but real)?
- Rename `activeMap` -> `activeProviderMap`
- Reset cache if `updateLists` is called.

How about the `info.seen` initialization concern: `Object.create(null)` instead of `{}`?

```diff
diff --git a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
index 402cbba9a9..36d41ccf8f 100644
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -654,7 +654,7 @@ function renderDependencies(depends, info, flat)
 		return providerCache;
 	}
 
-	const activeMap = getProviderMap();
+	const activeProviderMap = getProviderMap();
 
 	// Use an object {} for key-based lookups (faster than an array [])
 	info.seen = info.seen || {};
@@ -675,7 +675,7 @@ function renderDependencies(depends, info, flat)
 		if (deps[i] === 'libc')
 			continue;
 
-		// This regex handles "name", "name>=ver", and "name (>=ver)"
+		// This regex handles "name", "name>=ver", "name (>=ver)", ...
 		const match = deps[i].match(PKG_PATTERN);
 		if (match) {
 			dep = match[1];
@@ -692,7 +692,7 @@ function renderDependencies(depends, info, flat)
 
 		info.seen[dep] = {
 			name:    dep,
-			pkgs:    Array.from(activeMap[dep] || []),
+			pkgs:    Array.from(activeProviderMap[dep] || []),
 			version: [vop, ver]
 		};
 
@@ -1177,6 +1177,7 @@ function updateLists(data)
 
 	packages.available = { providers: {}, pkgs: {} };
 	packages.installed = { providers: {}, pkgs: {} };
+	providerCache = null;
 
 	return (data ? Promise.resolve(data) : downloadLists()).then(function(data) {
 		const pg = document.querySelector('.cbi-progressbar');

```

Konstantin-Glukhov:
> How about the info.seen initialization concern: Object.create(null) instead of {}?

Did I miss something? I didn't see the concern raised. Was it a part of a different thread?

map-b:
Cheap AI review, doesn't matter. I'm not sure what are we waiting for...

General view, I think we are close to a point where core maintainers start to use agents with expensive frontier AI models and that's it. No idea how is going to be the process of engaging new contributors/maintainers. How it was in the past? Good and bad feelings (like every revolution) 🙃

Konstantin-Glukhov:
I am fine with you change. Maintainers should merge this PR if they want that bug addressed.

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -623,50 +624,75 @@ function renderDependencyItem(dep, info, flat)
...
+		return providerCache;
+	}
+
+	const activeMap = getProviderMap();
+
+	// Use an object {} for key-based lookups (faster than an array [])
+	info.seen = info.seen || {};
+
+	const PKG_PATTERN = new RegExp(
+		/^([^><=~\s]+)/.source + // [1] Name stops at space or operator
```
efahl: Creating a regex, then extracting the text used to create it is probably not the best way to represent a string.  You could simply use the strings to make compilation and execution faster, and reduce the code size:

```diff
- 	/^([^><=~\s]+)/.source + // [1] Name stops at space or operator
+	'^([^><=~\s]+)' + // [1] Name stops at space or operator
```

efahl:
Can you remove the second commit from this PR, so that it only contains the actual bug fix, please?  (Or @systemcrash, can you just grab the one-liner from https://github.com/openwrt/luci/pull/8593/changes/e997ab852efd60128e1bed7486a4b599981bb1cf and merge it without all the unnecessary stuff?)

systemcrash:
> Can you remove the second commit from this PR, so that it only contains the actual bug fix, please? (Or @systemcrash, can you just grab the one-liner from [e997ab8](https://github.com/openwrt/luci/commit/e997ab852efd60128e1bed7486a4b599981bb1cf) and merge it without all the unnecessary stuff?)

OK - I've pushed that out to 25.12 for now. Let's see see how it weathers for a while. The follow-up stuff can remain in this PR.

map-b:
~Just a personal preference maybe. For a case like this with multiple linked issues. Next time, please, modify if required and merge the PR so the bot closes all the related issues and everything  ends up linked together.~  EDIT: Ok, only one "fixes:" which was not in the commit itself was missed. *I was surprised that the commit was not enough as I thought, ~confirmed it's enough ;)~ EDIT: No, it only closed the issues in Luci repo. Just to confirm merging the PR would close in other repos too, right? At least I saw that behavior in the openwrt repo.

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
Konstantin-Glukhov: > Creating a regex, then extracting the text used to create it is probably not the best way to represent a string. You could simply use the strings to make compilation and execution faster, and reduce the code size:
> 
> ```diff
> - 	/^([^><=~\s]+)/.source + // [1] Name stops at space or operator
> +	'^([^><=~\s]+)' + // [1] Name stops at space or operator
> ```

I agree that avoiding unnecessary `RegExp` construction can slightly reduce allocations/code size here, although the regex is not actually being executed — only instantiated so `.source` can be read.

That said, the suggested replacement is not equivalent. In a JavaScript string literal, `\s` is not preserved as a regex escape sequence, so this:

```js id="k4zqxa"
'^([^><=~\s]+)'
```

does not produce the same pattern as:

```js id="e5gb4l"
/^([^><=~\s]+)/
```

The equivalent string form would need:

```js id="n1z4r1"
'^([^><=~\\s]+)'
```

Personally, I think the regex literal + `.source` form is clearer and less error-prone because it avoids the double-escaping rules required by string literals. I’d prefer the small overhead over making the pattern harder to read and maintain.

map-b:
@efahl Why did you raise the code size concern? Did you checked the resulting .apk size or just looking at the code? I took(lightly) this rabbit-hole about code size, at least, in the luci app cases, where it is minified, compressed and ...?

The PR raw code changes delta is about 800 bytes and the resulted package .apk increased from 11 to 17KB (6000 bytes of increment)-> strange, it seemed disproportionate (tm). So after some blind unfruitfully searches, a last dumb discussion with google (logout) IA search mode, pointed out some possible cause about the regexp, specifically that the `.source +` syntax could cause problems in older minifiers.

So tried again with the RAW regexp and "voila", the size return to the expected value. 17KB -> 11KB

Doubts/notes:
- No obvious error from compile logs. The package is compiled but without minified sources.
- Raise an error if this happens?
- Upgrade the minifier to allow modern syntax?
- There could more cases with non-minified sources somewhere else in the repos?

@Konstantin-Glukhov Some idea to support the best syntax for regex. (I have just learnt it when you proposed it recently but I agree with you that it seems the cleanest/best option)

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -624,50 +623,76 @@ function renderDependencyItem(dep, info, flat)
...
 {
 	const deps = depends || [];
 	const items = [];
 
-	info.seen = info.seen || [];
+	function getProviderMap() {
+		// Only build the map once (Lazy Loading)
+		if (providerCache) return providerCache;
+
+		providerCache = {};
```
GeorgeSapkin: Sorry for piling on this lively discussion, but please consider using the `Map` type instead of an empty object. Objects are not empty and you risk introducing unintended behavior.

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -624,50 +623,76 @@ function renderDependencyItem(dep, info, flat)
...
+		index(packages.available.providers);
+
+		// providerCache[ip] = Set( ip, ip-tiny, ip-full )
+		return providerCache;
+	}
+
+	const activeProviderMap = getProviderMap();
+
+	// Use an object {} for key-based lookups (faster than an array [])
+	info.seen = info.seen || {};
```
GeorgeSapkin:
```suggestion
	info.seen ??= {};
```

But looks like this could be a `Map` as well:

```suggestion
	info.seen ??= new Map();
```

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
Konstantin-Glukhov: > Sorry for piling on this lively discussion, but please consider using the `Map` type instead of an empty object. Objects are not empty and you risk introducing unintended behavior.

Thanks for the feedback! You are completely right about the prototype collision risks.

Given LuCI's strict code size and footprint limitations, switching to a Map introduces slightly heavier syntax throughout the file (.set(), .get(), .has()).

To keep the code size minimal and highly minifiable, I suggest using Object.create(null) instead of {}. This creates a completely blank dictionary with zero inherited properties or prototype risks, while preserving the lightweight, low-overhead object syntax.

Konstantin-Glukhov:
> @Konstantin-Glukhov Some idea to support the best syntax for regex. (I have just learnt it when you proposed it recently but I agree with you that it seems the cleanest/best option)

If the minimizer cannot be upgraded, I am afraid using raw strings is the only option.

map-b:
- https://github.com/openwrt/luci/issues/8256 this is an old known friend.

jsmin is the key word, good news is that there is room for improvement in minifing (potential overall code size optimization @efahl) if not making a more complex one, at least, for allowing the valid js syntax consistently.

If not, I'm willing for leave the clean regex syntax commented out together with a jsmin explanation comment.

@georgesapkin should we include some CI test checking for correct minification?

GeorgeSapkin:
@map-b I'm not sure what these tests would look like. If you can come up with a PoC I can have a look.

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
GeorgeSapkin: What are the size implications of using Map?

Konstantin-Glukhov: Probably 30 bytes :)

map-b:
Just another instance of the recurrent issue/topic, when can be worthy to switch to a more complex, slower minifier?
2026 May candidates: `esbuild` (cli go static linked binary 10MB), `minify` (cli go static linked binary 5MB)

- `minify` failed with the "global return" syntax in LuCI (https://github.com/openwrt/luci/blob/98fefa8b5b5e9e95e381813e883ce074428f8e2d/eslint.config.mjs#L162-L164). (Worked with a dirty workaroud)
- `esbuild` worked at first.

Just to note here a quick tiny test results with package-manager.js source code example (also tested successfully with the problematic RegExp syntax with esbuild).

| | raw | jsmin | esbuild (10ms) |
| :---: | --- | --- | --- |
| package-manager.js | 37333 B (37 KB) | 29455 B (28,7 KB) | 26208 B (25,5 KB) | 

There are previous "upgrade the minifier" discussions that ended up with a NAK because of the need to include new dependencies, specially node.js. The best we have now is "go" but that is also too much to add to the default openwrt building system.

If it can be accepted just to download the static binary as a host tool... . It is available in major distros, if that opens some path.

*Seeing parallel discussion with csstidy, I have not tested it but this should cover it too.
EDIT: Just found this benchmark repo linked from esbuild docs https://github.com/privatenumber/minification-benchmarks#readme, and "tested" that cminify candidate but it fails in the test example with inline CSS backslashes ending lines (ES5 style?) and the gain seems too similar to jsmin...

---------------
Coming back to this PR, `jsmin` force us to not use certain JS syntax so follow that and define the regex in the most compact way too. And I'm tempted to try a "extreme" refactor of the proposed changes to "manually" minimize the delta and abuse of comments since they are free (in code size terms) to make the minified delta = +0.

George, there were actually errors in the logs... so probably the best would be to stop the compilation there? There are more clues like the .o semi-proccesed file plus the unminified source. (The final .apk size unminified would be 13KB but with that extra file ended up in 17KB in this case)
```
find /builder/build_dir/target-x86_64_musl/luci-app-package-manager/ipkg-all/luci-app-package-manager/www/ -type f -name '*.js' | while read src; do if jsmin < "$src" > "$src.o"; then mv "$src.o" "$src"; fi; done
JSMIN Error: Unterminated set in Regular Expression literal.
```

systemcrash:
I use the eslint config primarily for local linting and js doc compiling. I do not want to invoke it as a requirement in the build environment given the recent attacks through package management vectors. So any minify should (ideally) be a self-contained binary. 

Fixing the global return problem means switching to a modular architecture which means a re-write of luci :)

map-b:
I missed adding the links, eslint != esbuild, not even though in adapt/modify any code base. I referenced the eslint config file because it describes, a bit,the LuCI JS specialness. I show the result because it worked out of the box, if not I would simply keep it for myself and not wasting your attention. Download the 10MB binary in an alpine x64 builder container, decompress and use it like jsmin. Excuse me if I misunderstood you. Only intention is sharing the findings and see the actual viability.

openwrt-ai (commented):

---

map-b:
I had some fun with magic agents, if anyone is curious about minifiers https://map-b.github.io/luci-build-stats/?a=jsmin&b=without_PR_package_manager (*Assisted-by: OpenCode and free models quota.) and that is the detailed size delta from this PR current state (including the already merged little change) +241 Bytes minified, that, when compressed, would be less (or more) (42  , 414, 982 or 1400 Bytes? depending of compression strategy). For example, 10,7 KB to 10,8 KB (+122 Bytes) in the compressed .apk and a non-lineal increment in the squashfs image of +1,4KB.

Anyway I have almost ready a manual minified version with a negative(minified) delta.

<details><summary>Minimal (*old v2, I forgot to use the ??= operator)</summary>
<p>

```diff
@@ -623,51 +623,51 @@
 
 	return li;
 }
+
+let pC = null;
 
 function renderDependencies(depends, info, flat)
 {
 	const deps = depends || [];
 	const items = [];
+	info.seen = info.seen || Object.create(null);
 
-	info.seen = info.seen || [];
+	// Provider cache: dep name -> Set of package names (built once, reused)
+	//   .installed.providers[ip] = [ ip, ip-tiny ]
+	//   .available.providers[ip]  = [ ip, ip-full ]
+	//   pC[ip]                    = Set( ip, ip-tiny, ip-full )
+	if (!pC) {
+		pC = Object.create(null);
+		for (const src of [packages.installed.providers, packages.available.providers])
+			for (const name in src)
+				for (const pkg of src[name])
+					(pC[name] = pC[name] || new Set()).add(pkg.name);
+	}
 
 	for (let i = 0; i < deps.length; i++) {
-		let dep, vop, ver;
+		let dep, vop = null, ver = null;
 
 		if (deps[i] === 'libc')
 			continue;
 
-		// This regex handles "name", "name>=ver", and "name (>=ver)"
-		const match = deps[i].match(/^([^><=~\s]+)\s*\(?([><=~]+)?\s*([^)]+)?\)?$/);
+		// This regex handles "name", "name>=ver", "name (>=ver)" with operators: > < = ~ >= <=
+		// Regex (capture groups numbered):
+		//   ^([^><=~\s]+)    [1] package name
+		//   \s?\(?            optional space + optional paren
+		//   ([><=~]+)?       [2] optional operators: > < = ~ >= <=
+		//   \s?               optional space before version
+		//   ([^)]+)?         [3] optional version string
+		//   \)?$              optional closing paren + end
+		const m = deps[i].match(/^([^><=~\s]+)\s?\(?([><=~]+)?\s?([^)]+)?\)?$/);
+		dep = m && m[1];
+		vop = m && m[2] || null;
+		ver = m && m[3] || null;
 
-		if (match) {
-			// Destructure the match array: [full_match, name, operator, version]
-			const [, matchedDep, matchedVop, matchedVer] = match;
-			dep = (matchedDep || '').trim();
-			vop = (matchedVop || '').trim() || null;
-			ver = (matchedVer || '').trim() || null;
-		} else {
-			// Fallback if the string is just a plain name with no operators
-			dep = deps[i].trim();
-			vop = ver = null;
-		}
-
-		if (info.seen[dep])
+		if (!dep || info.seen[dep])
 			continue;
 
-		const pkgs = [];
-
-		(packages.installed.providers[dep] || []).forEach(function(p) {
-			if (pkgs.indexOf(p.name) === -1) pkgs.push(p.name);
-		});
-
-		(packages.available.providers[dep] || []).forEach(function(p) {
-			if (pkgs.indexOf(p.name) === -1) pkgs.push(p.name);
-		});
-
 		info.seen[dep] = {
 			name:    dep,
-			pkgs:    pkgs,
+			pkgs:    [...pC[dep] || []],
 			version: [vop, ver]
 		};
 
@@ -1152,6 +1152,7 @@
 
 	packages.available = { providers: {}, pkgs: {} };
 	packages.installed = { providers: {}, pkgs: {} };
+	pC = null;
```
</p>
</details> 

I decided to make the regex more stricter (*not because of the code size, which also). Only three classes of variations in dependencies format: 
- `name`,
- `name<operator>version` (APK query json output) or
- `name<space>(<operator><space>version)` (OPKG)

*`APK list --full `strips the operator and version

<details><summary>APK </summary>
<p>

```
root@OpenWrt:/www/luci-static/resources# apk query --format json --fields depend
s dovecot-pigeonhole
[
  {
    "depends": [
      "dovecot2.3.21>=2.3.0",
      "dovecot2.3.21<2.4.0",
      "libc"
    ]
  }
]

root@OpenWrt:/www/luci-static/resources# apk list --full | grep -A 5 dovecot-pig
eonhole
Package: dovecot-pigeonhole
Version: 0.5.21.1-r1
Depends: dovecot2.3.21, dovecot2.3.21, libc
```

</p>
</details> 

<details><summary>OPKG</summary>
<p>

```
root@OpenWrt:~# opkg depends firewall4
firewall4 depends on:
	ucode (>= 2022.03.22)
	libc
        ...
root@OpenWrt:~# opkg depends -A dovecot-pigeonhole
dovecot-pigeonhole depends on:
	dovecot (>= 2.3.21)
```

</p>
</details>

map-b:
I got a -264 bytes delta but then it appeared an old "bug" or a not fully implemented feature and ended up with -83 bytes with the "fixes". Now the EXTRA_DEPENDS are showed correctly in LuCI.

Screenshot 2026-06-10 at 21 06 35

Konstantin-Glukhov:
The following changes will not make the code much bigger but will add clarity.

Can we move the regex outside of the loop as a constant?
```
const depRegex = /^([^><=~\s]+)\s?\(?([><=~]+)?\s?([^)]+)?\)?$/;

for (let i = 0; i < deps.length; i++) {
    const m = deps[i].match(depRegex);
    ...
}
```
Can we destructure the regex result?
```
        const [full, name, operator, version] = m;

        // Populate provider cache if not already done
        if (!pC[name]) {
            pC[name] = new Set();
            for (const src of [packages.installed.providers, packages.available.providers]) {
                for (const pkg of src[name] || []) {
                    pC[name].add(pkg.name);
                }
            }
        }

        // Record the dependency entry
        info.seen[full] = {
            name,
            pkgs: [...pC[name]],
            version: [operator || null, version || null]
        };
```

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -596,12 +597,15 @@ function renderDependencyItem(dep, info, flat)
...
+		if (vop && ver)
+			text += ' (%s %s)'.format(vop, ver);
+
 		if (pkg.installsize)
 			text += ' (%1024mB)'.format(pkg.installsize);
 		else if (pkg.size)
 			text += ' (~%1024mB)'.format(pkg.size);
 
-		li.appendChild(E('span', { 'data-tooltip': pkg.description },
+		li.appendChild(E('span', { 'data-tooltip': pkg.description + (vop && ver ? '\nAvailable: %s'.format(pkg.version) : '') },
```
openwrt-ai: When `pkg.description` is `undefined`, the string concatenation now produces a literal `"undefined"` tooltip. Previously `'data-tooltip': pkg.description` passed `undefined`, which `E()` skips because [`attr[key] == null` at luci.js:1503](https://github.com/openwrt/luci/blob/34ed6cbc9f8cfde6a70dd861e1c3dc8a93c79e5e/modules/luci-base/htdocs/luci-static/resources/luci.js#L1503) drops the attribute entirely. Now `pkg.description + ''` evaluates to the string `"undefined"`, so `setAttribute('data-tooltip', 'undefined')` runs. This is reachable: the fallback `pkg` is `{ name: dep.name }` (no `description`) when no provider is found, and any provider package lacking a description hits it too — the user sees a tooltip reading "undefined".

Guard the description and fall back to `null` so an empty result still drops the attribute:

```suggestion
		li.appendChild(E('span', { 'data-tooltip': (pkg.description || '') + (vop && ver ? '\nAvailable: %s'.format(pkg.version) : '') || null },
```

---

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -559,7 +559,8 @@ function pkgStatus(pkg, vop, ver, info)
 	}
 	else if (!pkg.missing) {
 		if (!vop || versionSatisfied(pkg.version, ver, vop)) {
-			info.install.push(pkg);
+			if (!info.install.some(p => p.name === pkg.name))
```
openwrt-ai: This dedup fixes the install-size double-count on the "Not installed" path, but the "Needs upgrade" path a few lines up — [`info.install.push(repl)` at line 541](``https://github.com/openwrt/luci/blob/34ed6cbc9f8cfde6a70dd861e1c3dc8a93c79e5e/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js#L541)`` — has no equivalent guard. Since this PR now intentionally indexes `info.seen` by the full dependency string (`m[0]`), the same package can be rendered on multiple dependency lines, so `repl` can be pushed more than once and re-introduce the double-count this commit is fixing, just on the upgrade path. Should the same `!info.install.some(p => p.name === ...)` guard be applied there?

---

openwrt-ai (commented):
Reviewed 2 new commits since the last review (`4a28e94`). Both commit messages match their diffs. Two inline findings on the new `renderDependencies`/`renderDependencyItem` changes — one confirmed tooltip regression, one question about dedup coverage on the upgrade path.

---

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
map-b: I left this path out, thinking that, at least, one version constraint would be valid... A safer approach, I changed info.install to Set and "reverted" this change to info.install.add(pkg), the same with repl so no duplicated is possible.

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
map-b: Done

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
map-b: so `info.seen ??= Object.create(null)` , right?

openwrt-ai (commented):
Reviewed 1 new commit since the last review (`34ed6cb`). The amended `5ffe7a9` resolves both prior inline findings: the `data-tooltip` now falls back to `null` (no more `"undefined"` tooltip), and `info.install` is converted to a `Set` with `.add()` on both the `repl` upgrade path and the not-installed path, so the double-count concern is closed. `.size`/`?.` guards and the `??=`/`Object.create(null)` changes are all consistent with Set/object semantics; no remaining array-style usages of `info.install`. No new issues found.

---

PR #8610, opened 2026-05-06
luci-app-acme: call right  dnsapi package when uacme is used as backend

orangepizza:
### Description

it handles case with case when uacme is used for backend of luci--app-acme (it need to install uacme-dnsapi-adaptor to install sciripts handle acme.sh DNS API) and fixed typos nearby

### Maintainer _(preferred)_

@tohojo 

---

## Tested on

**OpenWrt version:** 
**LuCI version:** 
**Web browser(s):** 
snapshot, x86 vm, firefox 150

---

## Checklist

- [x] Incremented :up: any `PKG_VERSION` in the Makefile.  Don't have version number in makefile for luci

### DIFF
--- a/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
+++ b/applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
@@ -22,6 +22,7 @@ return view.extend({
 			L.resolveDefault(fs.list('/usr/lib/acme/client/dnsapi/'), null),
 			L.resolveDefault(fs.lines('/proc/sys/kernel/hostname'), ''),
 			L.resolveDefault(uci.load('ddns')),
+			L.resolveDefault(fs.stat('/usr/sbin/uacme'), null),
 		]);
 	},
 
@@ -35,6 +36,7 @@ return view.extend({
 		let ddnsDomains = _collectDdnsDomains();
 		let wikiUrl = 'https://github.com/acmesh-official/acme.sh/wiki/';
 		let wikiInstructionUrl = wikiUrl + 'dnsapi';
+		let hasUacme = data[5] != null;
 		let m, s, o;
 
 		m = new form.Map("acme", _("ACME certificates"),
@@ -54,7 +56,8 @@ return view.extend({
 
 		o = s.option(form.Value, "account_email", _("Account email"),
 			_('Email address to associate with account key.') + '<br/>' +
-			_('If a certificate wasn\'t renewed in time then you\'ll receive a notice at 20 days before expiry.')
+			_('CA may or may not send email to notify a cert may expire. ' +
+			'Let\'s Encrypt doesn\'t send email; ZeroSSL sends a notification email 30 days before expiry.')
 		);
 		o.rmempty = false;
 		o.datatype = "minlength(1)";
@@ -89,18 +92,19 @@ return view.extend({
 		o = s.taboption('general', form.ListValue, 'validation_method', _('Validation method'),
 			_('Standalone mode will use the built-in webserver of acme.sh to issue a certificate. ' +
 				'Webroot mode will use an existing webserver to issue a certificate. ' +
-				'DNS mode will allow you to use the DNS API of your DNS provider to issue a certificate.') + '<br />' +
+				'DNS mode will allow you to use the DNS API of your DNS provider to issue a certificate.' ) + '<br />' +
 			_('Validation via TLS ALPN') + ': ' + _('Validate via TLS port 443.') + '<br />' +
 			'<a href="https://letsencrypt.org/docs/challenge-types/" target="_blank">' + _('See more') + '</a>'
 		);
-		o.value('standalone', 'HTTP-01' + _('Standalone'));
-		o.value('webroot', 'HTTP-01' + _('Webroot Challenge Validation'));
+		o.value('standalone', 'HTTP-01 ' + _('Standalone'));
+		o.value('webroot', 'HTTP-01 ' + _('Webroot Challenge Validation'));
 		o.value('dns', 'DNS-01 ' + _('DNS Challenge Validation'));
 		o.value('alpn', 'TLS-ALPN-01 ' + _('Validation via TLS ALPN'));
 		o.default = 'standalone';
 
 		if (!hasDnsApi) {
-			let dnsApiPkg = 'acme-acmesh-dnsapi';
+			// Switch package based on which ACME client backend
+			let dnsApiPkg = hasUacme ? 'uacme-dnsapi-adapter' : 'acme-acmesh-dnsapi';
 			o = s.taboption('general', form.Button, '_install');
 			o.depends('validation_method', 'dns');
 			o.title = _('Package is not installed');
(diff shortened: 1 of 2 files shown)


orangepizza:
have another for support dns-persist-01 challenge, but it's not in CA's production so it may be other PR after it releases

systemcrash:
@stokito @tohojo

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -54,7 +56,8 @@ return view.extend({
 
 		o = s.option(form.Value, "account_email", _("Account email"),
 			_('Email address to associate with account key.') + '<br/>' +
-			_('If a certificate wasn\'t renewed in time then you\'ll receive a notice at 20 days before expiry.')
+			_('CA may or may not send email to notify cert may expire.' + 
+			'Let\' encrypt doesn\'t send email, ZeroSSL sends notificateion email 30 days before expirey')
```
openwrt-ai: This replacement introduces several errors. The two string literals concatenate with no space, so it renders `...may expire.Let' encrypt...`; `Let\' encrypt` produces `Let'` (misplaced apostrophe, lowercase) where it should be `Let's Encrypt`; and there are two typos: `notificateion` and `expirey`.

```suggestion
			_('CA may or may not send email to notify a cert may expire. ' +
			'Let\'s Encrypt doesn\'t send email; ZeroSSL sends a notification email 30 days before expiry.')
```

---

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -35,6 +36,7 @@ return view.extend({
 		let ddnsDomains = _collectDdnsDomains();
 		let wikiUrl = 'https://github.com/acmesh-official/acme.sh/wiki/';
 		let wikiInstructionUrl = wikiUrl + 'dnsapi';
+		let hasUacme = data[5] != null;
```
openwrt-ai, line 39: Detection keys solely on the presence of the `/usr/sbin/uacme` binary. If a user has `uacme` installed but acme.sh is the active backend, the install button will point at `uacme-dnsapi-adapter` rather than `acme-acmesh-dnsapi`. Is keying on the binary alone sufficient, or should this also reflect the configured/active backend? (In practice `uacme-dnsapi-adapter` [depends on `acme-acmesh-dnsapi`](https://github.com/openwrt/packages/blob/a5137522a7eb35e7de481068ca953430adbfab88/net/uacme/Makefile#L64), so the dnsapi scripts get installed either way — but the offered package may still not match the user's intent.)

---

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -125,7 +129,7 @@ return view.extend({
 		o.depends('validation_method', 'webroot');
 		o.depends('validation_method', 'alpn');
 		o.modalonly = true;
-
+
```
openwrt-ai: nit: this change replaces a clean blank line with one containing trailing whitespace.

```suggestion

```

---

openwrt-ai (commented):
Reviewed the uacme backend-detection change. The `uacme-dnsapi-adapter` package name in the code matches the packages feed, and `/usr/sbin/uacme` is the correct detection path (installed by the `uacme` package). A few inline notes below, mostly around the rewritten `account_email` help text which introduces typos and a missing-space rendering issue.

---

stokito (approved):
Looks OK, but I didn't tested.

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
orangepizza: it adapter brings original acmesh-dnsapi as it's own depend, so it shouldn't matter

orangepizza:
fixed typos ai suggested

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
```diff
@@ -54,7 +56,8 @@ return view.extend({
 
 		o = s.option(form.Value, "account_email", _("Account email"),
 			_('Email address to associate with account key.') + '<br/>' +
-			_('If a certificate wasn\'t renewed in time then you\'ll receive a notice at 20 days before expiry.')
+			_('CA may or may not send email to notify cert may expire.' + 
+			'Let\' Encrypt doesn\'t send email; ZeroSSL sends notificateion email 30 days before expiry')
```
openwrt-ai: A few issues from the earlier round are still present here. The two string fragments concatenate with no separating space, so this renders `...may expire.Let' Encrypt...`. `Let\' Encrypt` produces `Let'` + a space (the apostrophe is misplaced) where it should be `Let's Encrypt`. And `notificateion` is a typo. There is also trailing whitespace after the `+` on line 59.

```suggestion
			_('CA may or may not send email to notify a cert may expire. ' +
			'Let\'s Encrypt doesn\'t send email; ZeroSSL sends a notification email 30 days before expiry.')
```

---

openwrt-ai (commented):
Reviewed the force-pushed commit. The trailing-whitespace nit and the commit message are now fixed, but the `account_email` help-text typos from the previous round are only partially addressed — one inline note below.

---

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
openwrt-ai, line 39: Off-by-one on the `data` index: this reads the wrong array element, so uacme is never actually detected.

The `load()` array is (0-indexed):

```
[0] fs.stat('/etc/init.d/acme')
[1] uci.load('acme')
[2] fs.list('/etc/acme')
[3] fs.list('/usr/lib/acme/client/dnsapi/')
[4] fs.lines('/proc/sys/kernel/hostname')
[5] uci.load('ddns')
[6] fs.stat('/usr/sbin/uacme')   <-- the entry this PR added
```

The uacme stat is at index **6**, but `hasUacme` keys on `data[5]`, which is `uci.load('ddns')`. `uci.load()` resolves to a defined value, so `data[5] != null` is essentially always true. The net effect is that `hasUacme` is true even on acme.sh-only systems, and the install button always offers `uacme-dnsapi-adapter` — the opposite of this PR's intent (and the inverse of the previously discussed concern). On a system where `uacme` is genuinely absent the detection should be false, but here it never is.

```suggestion
		let hasUacme = data[6] != null;
```

---

openwrt-ai (commented):
Reviewed the latest commit. The `account_email` help-text typos and trailing-whitespace items from the previous round are now resolved. One new issue: the `hasUacme` detection reads the wrong `data` index (off-by-one), so uacme is effectively never correctly detected — inline note below.

---

applications/luci-app-acme/htdocs/luci-static/resources/view/acme/acme.js
openwrt-ai: Please disregard this comment — it is incorrect, my apologies.

I miscounted the `load()` array. The actual order is:

```
[0] fs.list('/etc/ssl/acme/')  (certs)
[1] fs.exec_direct('/usr/libexec/acmesh-dnsinfo.sh')
[2] fs.list('/usr/lib/acme/client/dnsapi/')
[3] fs.lines('/proc/sys/kernel/hostname')
[4] uci.load('ddns')
[5] fs.stat('/usr/sbin/uacme')   <- added by this PR
```

`uacme` is at index **5**, so `let hasUacme = data[5] != null;` is correct as written. My suggested `data[6]` change would have introduced a bug rather than fixed one. No change is needed here.

---

PR #8616, opened 2026-05-10
luci-base: fix "null" text appearing in modal

AndyChiang888:
fix "null" text appearing in modal

### Maintainer _(preferred)_

@<github-user>

---

## Tested on

**OpenWrt version:** SNAPSHOT
**LuCI version:** master

---

## Checklist

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/luci.js
+++ b/modules/luci-base/htdocs/luci-static/resources/luci.js
@@ -1379,7 +1379,7 @@
 				for (let i = 0; i < children.length; i++)
 					if (this.elem(children[i]))
 						node.appendChild(children[i]);
-					else if (children !== null && children !== undefined)
+					else if (children[i] !== null && children[i] !== undefined)
 						node.appendChild(document.createTextNode(`${children[i]}`));
 
 				return node.lastChild;
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -4993,7 +4993,7 @@ const UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 								if (info)
 									infoNode = E('p', _('%s').format(info));
 
-								UI.prototype.showModal(_('Uploading file…'), [ progress, infoNode ? infoNode : null ]);
+								UI.prototype.showModal(_('Uploading file…'), [ progress, infoNode ]);
 
 								const data = new FormData();
 


systemcrash:
Looks like an acceptable fix @feckert

ckorber:
Ah, my bad!
I forgot about that. I tested @AndyChiang888 's change and it works now when input is `null`.

feckert:
@systemcrash Works as expected. Thanks.

@AndyChiang888 Please make CI/CD happy :-)

AndyChiang888:
My email address changed, so CI/CD failed. I'll fix it.

jow-:
Maybe just `[ progress, infonode ?? '' ]`

AndyChiang888:
Will using '' cause DOM rendering issue?

ckorber:
I checked the suggestion from @jow- . It works on my machine.

AndyChiang888:
The display effect is the same, but will there be one more text node if use '' ?

jow-:
I wonder why this bug appears in the first place. The nodes array passed as 2nd argument to [`showModal()`](https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/ui.js#L4224-L4237) is passed to [`DOM.append()`](https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/luci.js#L1374-L1399) which is supposed to filter `null` and `undefined` elements automatically.

I do not want to do too much bikeshedding on your trivial proposed fix, but `null` appearing in the first place might hint at a deeper bug which could potentially affect other things as well.

AndyChiang888:
https://github.com/openwrt/luci/blob/c346174ed14008e98f8ba103ae9b289880319d38/modules/luci-base/htdocs/luci-static/resources/luci.js#L1382
~~Should this line be changed to (children[i] !== null && children[i] !== undefined)?~~

I've tested it, and this line is causing this issue.

AndyChiang888:
ping @systemcrash @feckert @jow- @ckorber

ckorber:
Works as expected, as I looked it up :+1:

modules/luci-base/htdocs/luci-static/resources/luci.js
```diff
@@ -1379,7 +1379,7 @@
 				for (let i = 0; i < children.length; i++)
 					if (this.elem(children[i]))
 						node.appendChild(children[i]);
-					else if (children !== null && children !== undefined)
+					else if (children[i] !== null && children[i] !== undefined)
```
systemcrash, line 1382: Weird - a null check on the array but `createTextNode` on the indexed item? Perhaps this whole for loop should go in `{}` so that the else is explicitly enclosed. Reading this block, it seems to me that the else case should be `children[i]`.

jow-: This fix is actually correct and the original code had a long standing bug, it did not check the looped items but always the array itself.

feckert:
replaced by https://github.com/openwrt/luci/pull/8627 as comment in https://github.com/openwrt/luci/pull/8616#discussion_r3232799707

PR #8617, opened 2026-05-10
[25.12] luci-app-openvpn: security fix

dibdot:
- fixed a critical path traversal vulnerability in `luci-app-openvpn`, 
  that allows authenticated users to upload arbitrary files
  and achieve root code execution.


systemcrash (approved):
Not a lua expert, but looks acceptable.

feckert:
We should also backport this to openwrt-24.10

aparcar:
Maybe it's just my taste but please use more descriptive commit subjects. Based on the increase of "security reports", soon every other commit would be called "<package>: security fix"... 

That note apart, I really appreciate you fixing this, thanks!

BKPepe:
> We should also backport this to openwrt-24.10

For reference, it was backported, well, cherry-picked: https://github.com/openwrt/luci/commit/739d307be35a7ae5a48d064efabce5bc2978b80f

PR #8619, opened 2026-05-11
luci-proto-qmi: validate PDP index and context input

Dnominated:
This update adds input validation for the PDP context index within the LuCI QMI protocol configuration page. 

It ensures that users provide a valid index, preventing potential connection failures caused by incorrect or out-of-range PDP context values being passed to the underlying qmi.sh script.

### DIFF
--- a/protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
+++ b/protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
@@ -143,13 +143,13 @@ return network.registerProtocol('qmi', {
 		o = s.taboption('advanced', form.Value, 'profile',
 			_('APN profile index'));
 		o.placeholder = '1';
-		o.datatype = 'uinteger';
+		o.datatype = 'range(1,9)';
 
 		if (L.hasSystemFeature('ipv6')) {
 			o = s.taboption('advanced', form.Value, 'v6profile',
 				_('IPv6 APN profile index'));
 			o.placeholder = '1';
-			o.datatype = 'uinteger';
+			o.datatype = 'range(1,9)';
 			o.depends('pdptype', 'ipv4v6');
 		};
 	}


systemcrash:
Is this strictly necessary? You limit the range where the QMI values are only about 4-5. uint seems fine for this today.

Dnominated:
_> Is this strictly necessary? You limit the range where the QMI values are only about 4-5. uint seems fine for this today._

I'm not sure what you are saying with 'you limit the range where the QMI values are only about 4-5. '
PDP Contexts (index of stored APN's and other values) "can be up to 11": https://www.3g4g.co.uk/Tutorial/ZG/zg_pdp.html

**Reading a little further about implementation of PDP Contexts on modern 5G cellular modems:**
"Modern 5G cellular modems do not use "PDP Contexts" as the primary session management mechanism; instead, they utilize PDU Sessions (Protocol Data Unit Sessions). "

You got me there. $250+ for a 5G cellular modem is outside my budget. I've been using 4GLTE modems ($20-50), that use up to 11, apparently, although I've not seen a number that high in the models I use.

The intent is to limit the PDP Context to a valid number, not just any number. And then use the underlying QMI.sh to correctly program the selected Context (with the entered APN) to the correct context, which, as of now **it does not**. The existing faulty QMI.sh ALWAYS programs the APN etc values to PDP Context ID #1, not the one put into Luci or into /etc/config/network.

systemcrash:
> _> Is this strictly necessary? You limit the range where the QMI values are only about 4-5. uint seems fine for this today._
> 
> I'm not sure what you are saying with 'you limit the range where the QMI values are only about 4-5. ' PDP Contexts (index of stored APN's and other values) "can be up to 11": https://www.3g4g.co.uk/Tutorial/ZG/zg_pdp.html
> 

Can be up to 11... *concurrently*. 

What I'm saying is what valid PDP are:

https://github.com/openwrt/uqmi/blob/7914da43cddaaf6cfba116260c81e6e9adffd5ab/common/qmi-enums-wds.h#L1264-L1269


A uint is an unsigned integer, any positive number. But you limit it to ... 1-9. If the above definitions are correct, then this isn't wrong...


> The intent is to limit the PDP Context to a valid number, not just any number. And then use the underlying QMI.sh to correctly program the selected Context (with the entered APN) to the correct context, which, as of now **it does not**. The existing faulty QMI.sh ALWAYS programs the APN etc values to PDP Context ID #1, not the one put into Luci or into /etc/config/network.

That would be a qmi script bug which needs fixing then.

@feckert do you have any idea about this?

protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js
```diff
@@ -143,13 +143,13 @@ return network.registerProtocol('qmi', {
 		o = s.taboption('advanced', form.Value, 'profile',
 			_('APN profile index'));
 		o.placeholder = '1';
-		o.datatype = 'uinteger';
+		o.datatype = 'range(1,9)';
```
openwrt-ai, line 146: `range(1,9)` accepts fractional values. The `range` validator parses with [`parseDecimal` (validation.js:1021-1023)](https://github.com/openwrt/luci/blob/adb015747f44f9694310d8db26e7404fa75477d5/modules/luci-base/htdocs/luci-static/resources/validation.js#L1021-L1023), whose regex `/^-?\d+(?:\.\d+)?$/` allows e.g. `1.5` or `2.7`. So this change regresses the integer guarantee the previous `uinteger` provided while adding the range bound — yet the backend treats the value as an integer ([`profile_id="${profile:-1}"` -> `--modify-profile "3gpp,$profile_id"` in qmi.sh](https://github.com/openwrt/openwrt/blob/246b216df57de45af20a1e66b337b41f63d11dda/package/network/utils/uqmi/files/lib/netifd/proto/qmi.sh#L337-L340)), and the option is declared `proto_config_add_int`. A non-integer index would be invalid. Combine both constraints so only positive integers in range are accepted:

```suggestion
		o.datatype = 'and(uinteger,range(1,9))';
```

Same applies to the `v6profile` line 152 below.

Separately, the upper bound of `9` is being debated in the PR comments. 3GPP profile indices on many modems extend beyond 9 (the discussion references "up to 11"), so this would reject indices that some hardware accepts. Worth confirming the intended ceiling before fixing it at 9.

---

openwrt-ai (commented):
Reviewed the new commit. One inline note on the `range(1,9)` datatype change.

---

PR #8620, opened 2026-05-11
luci-app-strongswan-swanctl: merge SAs and connections

lvoegl:
Merges strongSwan conns and SAs into a detailed table of connections and a table of children for each of them.

### Description

The status used to display only Security Associations, ignoring connections. This change merges connections and related Security Associations into one table / detailed modal view. Also adds Start/Stop buttons to connections and children.

### Screenshots

Connection Table:

Connection Details:

Connection Children Table:

### Maintainer

@lvoegl, Nicholas Smith (no @?)

---

## Tested on
**OpenWrt version:**  OpenWrt 25.12.3 (r32912-6639b15f62)
**LuCI version:** LuCI openwrt-25.12 branch (26.124.37917~fa51906)
**Web browser(s):** Firefox (150.0.2)

Issue #7435, opened 2024-11-30
[CI] Avoid losing existing translations when the i18n key was changed 


stokito:
### How can we otherwise help?

Translations are good even when they bad. But we have periodically a situation when some i18n key i.e. the English text was changed and the old translation doesn't work anymore.
For example if the `_('ACLs')` was changed to `_('ACL')`.
We may keep the old translations which now will be not exactly accurate but it's better than nothing. The Weblate will show such new but translated strings as "Needs for a review" and translators can easily fix working.

Usually after running of the `i18n-sync.sh` such orphaned keys are commented out and moved to the bottom of po file. I trying to restore such translations but if the key was re-translated in the Weblate we end up with a conflict. So it was spent a lot of time of translators and devs.

We need some solution to avoid such i18n keys breaking. This may be a simple CI job that will run the `i18n-sync.sh` and check for the orphaned keys. We may require that all the orphaned keys were renamed or removed before merging.

dannil:
I think this is a bigger issue on the "integrity" of the source texts that was brought up in #7175.

systemcrash:
This doesn't happen often enough, nor is it a problem that is not resolvable. I don't really like CI adding its own commits to repair something. But running the i18n scripts and flagging a failure if a PR touches po files and its results don't match is possible. It's a bit of work... can you make it work @stokito ?

- a new PR touching .po(t) files triggers CI process
- pull the luci repo
- run the i18n sync scripts
- if its results differ, and there are outstanding 'changes' - warn

This will require turning off line numbers and filenames in the po files to reduce churn or false-positives due to code only changes.

systemcrash:
I've come up with some changes for the build scripts to omit line number comments like:
`#: applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js:432` by default. This avoids churn for code-only changes.

One advantage is that we see an average reduction in po file size by about 25%.

Just means a big commit with lots of reduction. All ~3500 files.

Next problem: CI. In my tests, running e.g. `msgmerge` on a github workflow, it randomly decides to amend juuuust a few messages which only change the line length, wrap, but git will mark this as `modified`. The CI runner runs *identical* commands to here. So the only thing I can think of is disparity in gettext tools versions. And... they differ 🤦 ( mine: 0.22.5, CI: 0.21 ). 

I don't want a CI runner making commits and starting merge wars with weblate. 🫤

hnyman:
[...]

Issue #7089, opened 2024-04-27
i18n updates


systemcrash:
Proposal
FAO: @hnyman @aparcar @stokito @stangri 

To prevent/remove location `#: applications/luci-app-blah/...:374` lines, via `msgmerge --no-location`.

In our automated systems, these lines are responsible for change and churn. Does the JS GUI need them?
This can be applied to templates and po files.

A second optimisation is altogether dropping the pruned lines `#~ ...`.

hnyman:
I asked about removing those location lines from .po files a few years ago, but if I remember right, @jow- thinks that they are necessary.

hnyman:
> second optimisation is altogether dropping the pruned lines #~ ...

I have actually done that manually a few times for the .po files where has been lots of pruned lines.

hnyman:
Reference to earlier issue about this.

https://github.com/openwrt/luci/issues/5655

systemcrash:
I mean, does anyone manually translate using poedit? Or does everything go via weblate now?

systemcrash:
I think a better solution for any outliers using poedit is ... another script :D 
A script which pre-processes what they want, adds the locations, job done, remove locations. It's a workflow admittedly, but, I dunno. What do others think?

systemcrash:
> > second optimisation is altogether dropping the pruned lines #~ ...
> 
> I have actually done that manually a few times for the .po files where has been lots of pruned lines.

We can at least fix this in the perl scripts to drop those lines. The commit history shows the changes anyway.

stokito:
> dropping the pruned lines #~ ....

no, this are useful. In the https://lokalise.com/ it shows such old strings as a suggestion when you editing a translation. This helps a lot because you can simply click on them and change only that parts that was changed and needs to be re-translated.

Also sometimes you can simply recover the old translations like I did here:
https://github.com/openwrt/luci/pull/7090/commits/60461c07fffb66e7c5aa02804d7c6c204b06f0cc


The line numbers are needed too.

But if something that I would like to improve is:
1. Move the po files to a separate repo. So I can again subscribe to the repo changes without being annoyed by translation commits. This also will significantly clear history.
2. Get rid of all the translation scripts (e.g. i18n-sync.sh) : this must be a part of the gettext package or something like that. Maybe contribute them or find a replacement (GNOME or Qt must have something like that).
[...]

Issue #5655, opened 2022-01-23
Drop the lines numbers from .po(t) with i18n sync/scan ?


hnyman:
@jow- 
Could we drop the line numbers from the .pot and .po files?
They cause constant trouble with PRs, and also cause a bit too much commit spam compared to the benefits, I think.
We could leave the file annotations, even multiple per same source file, so that the amount of string instances could be seen.

Typical example is two-month old PR  #5563 that I just merged, but had to edit the patch due to the changed annotation line numbers (+1 to every one, in this case).

And right now #5654 has lots of (proper by themselves) line number annotation corrections, which may get invalidated before the PR gets committed.

(To be honest, since 901a0821f I can can't even figure out the exact place where the line numbers are added in i18n-scan.pl, so no PR from me :-(

jow-:
No. These source references allow graphical tools like poedit to display source context while translating. They are useful.

PR #8621, opened 2026-05-12
I18N: sync with luci.mk

feckert:
@hnyman 
[Based on the latest changes](https://github.com/openwrt/luci/commit/6918410c7a0ba0d63f094d5a6cbe05471abc6fda) sync I18N with luci.mk with the command `./build/i18n-add-language.sh`


hnyman:
I do not like creating empty translation files before there is translator interest for them. Not sure if that is wise due to the (unnecessary) repo growth.

Due to the still included context line numbers, the regular syncing of language files creates quite megabytes of fluffy churn to the git repo.  @jow- has so far wanted to keep the line numbers for poedit usage, but as pretty much all translation is done via weblate, the context numbers are pretty unused in reality.

(I enabled the compilation of those few languages, as their translation of luci-base had progressed well.)

feckert:
@hnyman I completely agree with you. That’s a lot of changes.

I’ve just come across a problem: I wanted to install ‘luci-app-strongswan-swanctl’ to check PR #8620, and there was no German translation. So I ran ‘./build/i18n-sync.sh’ in luci toplevel directory, but that didn’t add the German PO file. Then I noticed that many translations were missing, so I ran the command `./i18n-add-language.sh`, but that didn’t add the German translation either. So I ran `./i18n-add-language.sh` in the `luci-app-strongswan-swanctl` directory, and after that the German translation was available.

I think this needs to be improved somehow.
Perhaps we could define standard languages that are always generated, and others that are less common only on request?

It’s all a bit inconsistent! How is one supposed to know when a translation is needed and when it isn’t?

hnyman:
Well, nowadays the translation files for each application are typically generated from the weblate GUI interface by the translators. 
(I would guess that 99% of translations are done via weblate).

There are currently about 120 activated languages in weblate, of which maybe 10 are really active and another 10 have semi-regular activity. About 40 language are enabled in luci.mk for compilation.
It makes no sense to create empty languages for rare lanaguages and rare apps.
 
The codebase is so huge, that full translation exists only for a really few languages.

systemcrash:
I come to the conclusion that if the user translating really needs accurate line numbers, they can run a refresh locally, rather than depending on the repo being up to date, or we trim out line numbers altogether and skip *that* churn. 

Adding and removing line numbers is a simple flag which can be integrated into the various ./build scripts ( I think I have that change queued up somewhere ).

Agree that we don't need to actively add po(t) files for available languages. Just do case by case.

hnyman:
Discussion about removal of the line numbers in :

https://github.com/openwrt/luci/issues/7435#issuecomment-2510197762
https://github.com/openwrt/luci/issues/5655
https://github.com/openwrt/luci/issues/7089

feckert:
If everything runs through Weblate, the numbers won’t matter anymore only if you want exactly know where this is used.

If we remove this line, then this would also prevent major changes from occurring if the source code shifts by just a single line. I’d be in favor of removing the line numbers.
But I’m not authorised to decide that on my own. We would need a consensus on this.

Then, hopefully, the problem that not all packages in LuCI have the same translation packages would also be solved. We could then standardize the workflow for creating PO files and wouldn’t always have to consider whether a language should be included or not. I can’t keep track of which languages is needed or not.

Whenever I create a new packages or update the luci.mk with a new languages, I just imagine running the `command: ./build/i18n-add-language.sh`
and then everything should be generated.

applications/luci-app-https-dns-proxy/po/fa/https-dns-proxy.po
```diff
@@ -0,0 +1,851 @@
+msgid ""
```
openwrt-ai, line 1: nit: `luci-app-https-dns-proxy` is maintained out-of-tree at https://github.com/mossdef-org/luci-app-https-dns-proxy, not in `openwrt/luci`. New `.po` files added here (this commit adds `fa`, `fil`, `ka`, ... ) would belong upstream rather than in this repo. Consider dropping the `luci-app-https-dns-proxy` commit from this series and adding the languages in the upstream repo instead.

---

openwrt-ai (commented):
Fresh review of all 104 commits (new `.po` template files added treewide via `i18n-add-language.sh`). The 968 added files are valid empty translation templates and the diff is additions-only (no existing translations are touched). The design question (whether to pre-create empty `.po` files and the line-number churn) is already under discussion by maintainers, so I'm not re-raising it.

## Commit checks
- `bc56a24` "luci-app-anitblock: sync supported languages with luci.mk" — subject misspells the package; the commit touches `applications/luci-app-antiblock/` (`anitblock` -> `antiblock`).
- `863a6d9` "luci-app-filbrowser: sync supported languages with luci.mk" — subject misspells the package; the commit touches `applications/luci-app-filebrowser/` (`filbrowser` -> `filebrowser`).
- `2cef040` "luci-app-pagckage-manager: sync supported languages with luci.mk" — subject misspells the package; the commit touches `applications/luci-app-package-manager/` (`pagckage` -> `package`).

---

PR #8623, opened 2026-05-12
luci-app-snmpd: remove version on shared library reference

AuthorReflex:
### Description
In version 5.9.5.2 of the main package, the version of the shared library was updated. To prevent subsequent package changes, the main package was updated to install a symlink without specifying the version, and the call from the interface was modified to reference to symlink without version.

image

### Maintainer _(preferred)_

@ckorber

---

## Tested on

**OpenWrt version:**  OpenWrt trunk (r34414-348303a9e2), snmpd version  5.9.5.2 

---

## Checklist

Depends on: https://github.com/openwrt/packages/pull/28867

### DIFF
--- a/applications/luci-app-snmpd/htdocs/luci-static/resources/view/snmpd/snmpd.js
+++ b/applications/luci-app-snmpd/htdocs/luci-static/resources/view/snmpd/snmpd.js
@@ -9,7 +9,7 @@
 'require view';
 
 var testlibsslv3 = function() {
-	return fs.exec('/usr/bin/ldd', [ '/usr/lib/libnetsnmp.so.40' ]).then(function (res) {
+	return fs.exec('/usr/bin/ldd', [ '/usr/lib/libnetsnmp.so' ]).then(function (res) {
 		return res.stdout.includes('libssl');
 	});
 };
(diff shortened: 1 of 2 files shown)


ckorber:
Looks good to me.
Any objection @systemcrash ?

systemcrash:
@ckorber do you consider the symlink reliable? It's a standard part of the install routine now?

Otherwise if you're satisfied with this, I am.

feckert:
@systemcrash In my opinion, that works.
We just have to wait until the PR for the net-snmp update is merged.
We're still waiting for feedback from @stintel

openwrt-ai (commented):

---

feckert:
@AuthorReflex thanks merged! 
Depending PR has been merged.

jow-:
That main package change works contrary to the intent of ABI tracking though:

Quoting https://openwrt.org/docs/guide-developer/package-policies#contents
>  The binary libbar1 package should only contain libbar.so.1 and libbar.so.1.2.3 as the common libbar.so symlink would clash with a libbar2 package providing version 2.0.0 of libbar. 

As it stands, the change broke forward compatibility and the next time libnetsnmp is updated to a new ABI version, file level conflicts will ensue.

feckert:
Then I guess we'll have to iterate through the lib directory, find the first libsnmp, and check whether it supports that `libsnmp` feature? But that's not the best solution for the future either if more then one libsnmp version are installed :-(

jow-:
Why not modify the ldd check to simply probe `/usr/sbin/snmpd` ? This should implicitly check whatever lib version the daemon was linked against without having to deal with version suffixes at all.

ckorber:
> Why not modify the ldd check to simply probe `/usr/sbin/snmpd` ? This should implicitly check whatever lib version the daemon was linked against without having to deal with version suffixes at all.

I opened a PR for that: [Link](https://github.com/openwrt/luci/pull/8753)

PR #8624, opened 2026-05-12
luci-app-lxc: fix authenticated path traversal and ACL bypass

dibdot:
* ucode fixes:
  - tighten `is_valid_lxc_name` regex to `^[A-Za-z0-9_][A-Za-z0-9_-]{0,63}$`
  - apply the validator in `lxc_configuration_get` and `lxc_configuration_set` before any filesystem access
  - reject the `'lxc error: …'` sentinel string returned by `lxc_get_config_path()` on failure, rather than concatenating it into a path.
  - shellquote `LXC_URL` in `lxc_get_downloadable` and `lxc_create`
* ACL fix: add `depends.acl = ["luci-app-lxc"]` to each of the five backend entries, 
  so the routes share the same authorization gate as the view

### DIFF
--- a/applications/luci-app-lxc/root/usr/share/luci/menu.d/luci-app-lxc.json
+++ b/applications/luci-app-lxc/root/usr/share/luci/menu.d/luci-app-lxc.json
@@ -11,7 +11,6 @@
 			}
 		}
 	},
-
 	"admin/services/lxccm/overview": {
 		"title": "Overview",
 		"order": 10,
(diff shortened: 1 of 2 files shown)


applications/luci-app-lxc/root/usr/share/luci/menu.d/luci-app-lxc.json
```diff
@@ -20,67 +19,104 @@
...
-			"methods": [ "cookie:sysauth_https", "cookie:sysauth_http" ],
+			"methods": [
+				"cookie:sysauth_https",
+				"cookie:sysauth_http"
+			],
 			"login": true
 		}
 	}
-}
+}
```
systemcrash: newline

systemcrash (approved):

applications/luci-app-lxc/root/usr/share/luci/menu.d/luci-app-lxc.json
dibdot: Thanks, fixed.

PR #8625, opened 2026-05-13
luci-app-https-dns-proxy: update dns.sb bootstrap IPs

DH-41:
Update the bootstrap DNS addresses for DNS.SB to match their current official recommendations. This ensures reliable resolution of the DoH endpoint.

Reference: https://dns.sb/doh/


stangri:
1. bootstrap DNS are not the IP addresses of DoH servers
2. any updates belong in upstream repo

PR #8628, opened 2026-05-13
luci-base: replace ufpd access check with service run check

feckert:
It is not enough simply to check whether the tool is installed. It is essential to check whether the service is actually running. After all, we need the functionality provided by the service.

@systemcrash I'm not sure whether I should do this via a system call, or is there another way?

### DIFF
--- a/modules/luci-base/root/usr/share/rpcd/ucode/luci
+++ b/modules/luci-base/root/usr/share/rpcd/ucode/luci
@@ -239,7 +239,6 @@ const methods = {
 				relayd:     access('/usr/sbin/relayd') == true,
 				apk:        access('/usr/bin/apk') == true,
 				wifi:       access('/sbin/wifi') == true,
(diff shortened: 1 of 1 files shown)


modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -332,6 +331,11 @@ const methods = {
 				fd.close();
 			}
 
+			result.ufpd = false;
+			if (access('/usr/sbin/ufpd')) {
+				result.ufpd = system(`/etc/init.d/ufp running >/dev/null 2>/dev/null`) == 0
```
systemcrash, line 336: Reuse `getInitList`?

feckert: I rather doubt that’s any better. After all, I’m only interested in one service (ufp), not all of them. Besides, I’d then have to intercept that in LuCI and add an extra RPC call. But I am still unsure

PR #8634, opened 2026-05-16
luci-mod-network: ra_preference option uses '' instead of "medium"

AndyChiang888:
"medium" is the default ra_preference option in odhcpd.
Which has the same effect as not defining it.
Using '' will prevent unnecessary parameters from being written to the flash.

Link https://github.com/openwrt/odhcpd/blob/master/README.md

### Maintainer _(preferred)_

@<github-user>

---

## Tested on

**OpenWrt version:** SNAPSHOT
**LuCI version:** master
**Web browser(s):** Chrome

---

## Checklist

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -897,9 +897,8 @@ return view.extend({
 
 					so = ss.taboption('ipv6-ra', form.ListValue, 'ra_preference', _('Router Priority'),
 						_('A tie-breaker for clients and their routes when multiple routers exist on the same network.'));
-					so.default = 'medium';
+					so.value('', _('Medium'));
 					so.value('low', _('Low'));
-					so.value('medium', _('Medium'));
 					so.value('high', _('High'));
 					so.depends('ra', 'server');
 


systemcrash:
Merged. Thanks @AndyChiang888

jow-:
The problem is that when `medium` is set in the config, the dropdown will now display a wrong preselection. I think an `o.default = 'medium'` would handle this in a cleaner way.

Edit: I see it had `o.default = 'medium'` before. The real fix is investigating why this did not have the desired effect.

AndyChiang888:
I previously tried sorting it in the following order:
```
so.default = '';
so.value('low', _('Low'));
so.value('', _('Medium'));
so.value('high', _('High'));
```
However, the empty value kept appearing first, so I adjusted it to:
```
so.value('', _('Medium'));
so.value('low', _('Low'));
so.value('high', _('High'));
```
I'm not sure why the empty value always takes precedence in the sorting order instead of sitting in the middle.

jow-:
Well, I think your fix is wrong. It should be investigated why o.default = 'medium' does not prevent 'medium' from getting written instead of randomly mangling choice values

AndyChiang888:
We can revert it back after figure out why, but ⁠o.default⁠ has always been writing non-empty values. I thought it was designed that way.

jow-:
Hm, yes - I believe it behaves different for flags (bools) compared to string options.

systemcrash:
In my test - if the value is absent, medium is written if `o.default = 'medium'`.

PR #8635, opened 2026-05-16
luci-proto-batman-adv: add status page

JFroch:
### Description

This PR adds a status page for Batman-adv that shows mesh infos such as hardware interfaces and connected neighbors.
The status page doesn´t work with batctl-tiny as that doesn´t expose the debug tables. In case batctl-tiny is installed the page displays a simple text that informs the user of the requirement.

Screenshot From 2026-05-16 02-40-28

image

### Maintainer

@onemarcfifty 

---

## Tested on

**OpenWrt version:** 25.12.4 r32933-4ccb782af7
**LuCI version:** LuCI openwrt-25.12 branch 26.134.75701~cd18b8e
**Web browser(s):** Firefox 150.0.3

---

## Checklist

### DIFF
--- a/protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js
+++ b/protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js
@@ -1,0 +1,155 @@
+'use strict';
+'require view';
+'require rpc';
+'require poll';
+'require dom';
+'require ui';
+
+
+var callGetBatadvInstances = rpc.declare({
+	object: 'luci.batadv',
+	method: 'getBatadvInstances'
+});
+
+function msToStr(milliseconds){
+	let ago;
+
+	if (milliseconds < 1000)
+		ago = _('%dms ago').format(milliseconds);
+	else if (milliseconds < 60000)
+		ago = _('%ds ago').format(milliseconds / 1000);
+	else
+		ago = _('%dm ago').format(milliseconds / 60000);
+
+	return ago;
+}
+
+function renderNeighborTable(instanceName, neighbors, hostnames) {
+	var t = new L.ui.Table(
+		[
+			_('Neighbor'),
+			_('Last seen'),
+			_('Reported Throughput'),
+			_('Connected on Hardware Interface')
+		],
+		{
+			id: 'neighs-' + instanceName,
+			placeholder: _('No neighbors connected')
+		},
+		E('em', [
+			_('No neighbors connected')
+		])
+	);
+
+	t.update(neighbors.map(function(peer) {
+		return [
+			hostnames[peer.neigh_address] ? `${peer.neigh_address} (${hostnames[peer.neigh_address]})` : peer.neigh_address,
+			[ +peer.last_seen_msecs, msToStr(peer.last_seen_msecs)],
+			[ +peer.throughput, '%.1mb/s'.format(+peer.throughput*1000) ], //peer throughput is reported in kbit/s
+			peer.hard_ifname,
+		];
+	}));
+
+	return t.render();
+}
+
+function renderHardifTable(instanceName, hardifs) {
+	var t = new L.ui.Table(
+		[
+			_('Interface'),
+			_('Address'),
+			_('Active'),
+			_('Hop Penalty'),
+			_('ELP Interval'),
+			_('Throughput Override')
+		],
+		{
+			id: 'hardifs-' + instanceName,
+			placeholder: _('No hardifs associated')
+		},
+		E('em', [
+			_('No hardifs associated')
+		])
+	);
+
+	t.update(hardifs.map(function(iface) {
+		return [
+			iface.hard_ifname,
+			iface.hard_address,
+			iface.active,
+			+iface.hop_penalty,
+			+iface.elp_interval,
+			+iface.throughput_override != 0 ? [+iface.throughput_override, '%.1mb/s'.format(iface.throughput_override*100000)] : [+iface.throughput_override, _('Not overridden')], //hardif throughput override is reported in 100kbit/s
+		];
+	}));
+
+	return t.render();
+}
+
+return view.extend({
+	renderIfaces: function(ifaces) {
+		var res = [
+			E('h2', [ _('Batman-adv Status') ])
+		];
+
+		if(ifaces.batctl_debug_tables){
+			for (var instanceName in ifaces.devices) {
+				res.push(
+					E('h3', [ _('Device "%h"', 'Batman-adv device heading').format(instanceName) ]),
+					E('p', [
+						E('span', { 'class': 'ifacebadge' }, [
+							E('img', { 'src': L.resource('icons', 'tunnel.svg'), 'style': 'width:32px;height:32px' }),
+							'\xa0',
+							instanceName
+						]),
+						E('span', { 'style': 'opacity:.8' }, [
+							' · ',
+							_('Algorithm %h', 'Batman-adv algorithm').format(ifaces.devices[instanceName].algo_name),
+							' · ',
+							E('code', { 'click': '' }, [ ifaces.devices[instanceName].mesh_address ])
+						])
+					]),
+					E('h5', [ _('Hardware Interfaces') ]),
+					renderHardifTable(instanceName, ifaces.devices[instanceName].hardifs),
+					E('h5', [ _('Neighbors') ]),
+					renderNeighborTable(instanceName, ifaces.devices[instanceName].neighbors, ifaces.hostnames)
+				);
+			}
+
+			if (res.length == 1)
+				res.push(E('p', { 'class': 'center', 'style': 'margin-top:5em' }, [
+					E('em', [ _('No Batman-adv interfaces configured or active.') ])
+				]));
+
+		}
+		else {
+			res.push(E('h5', { 'class': 'center', 'style': 'margin-top:5em' }, [
+				E('em', [ _('This page requires either batctl-default or batctl-full to work but none of them are installed.') ])
+			]));
+		}
+
+		return E([], res);
+	},
+
+	render: function() {
+		poll.add(L.bind(function () {
+			return callGetBatadvInstances().then(L.bind(function(ifaces) {
+				dom.content(
+					document.querySelector('#view'),
+					this.renderIfaces(ifaces)
+				);
+			}, this));
+		}, this), 5);
+
+		return E([], [
+			E('h2', [ _('Batman-adv Status') ]),
+			E('p', { 'class': 'center', 'style': 'margin-top:5em' }, [
+				E('em', [ _('Loading data…') ])
+			])
+		]);
+	},
+
+	handleReset: null,
+	handleSaveApply: null,
+	handleSave: null
+});
--- a/protocols/luci-proto-batman-adv/root/usr/share/rpcd/acl.d/luci-proto-batman-adv.json
+++ b/protocols/luci-proto-batman-adv/root/usr/share/rpcd/acl.d/luci-proto-batman-adv.json
@@ -1,0 +1,12 @@
+{
+	"luci-proto-batman-adv": {
+		"description": "Grant access to LuCI Batman-adv procedures",
+		"read": {
+			"ubus": {
+				"luci.batadv": [
+					"getBatadvInstances"
+				]
+			}
+		}
+	}
+}
--- a/protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
+++ b/protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
@@ -1,0 +1,65 @@
+// Copyright 2026 Jan Froch <jan@froch.eu>
+// Licensed to the public under the Apache License 2.0.
+
+'use strict';
+
+import { cursor } from 'uci';
+import { popen, open } from 'fs';
+
+function command(cmd) {
+	return trim(popen(cmd)?.read?.('all'));
+}
+
+const apkproc = popen(`apk info 2>/dev/null | grep '^batctl-\\(default\\|full\\)$'`);
+apkproc.read('all');
+const batctl_debug_tables = !apkproc.close();
+
+const methods = {
+	getBatadvInstances: {
+		call: function() {
+			const data = {};
+			let device;
+
+			data.batctl_debug_tables = batctl_debug_tables;
+
+			if(batctl_debug_tables){
+				data.devices = {};
+				const fh = open('/etc/bat-hosts', 'r');
+				let bat_hosts = {};
+
+				if (fh){
+					for (let line = fh.read("line"); length(line); line = fh.read("line")){
+						let entry = split(rtrim(line, '\n'), /\s+/);
+						if(length(entry) == 2){
+							bat_hosts[entry[0]] = entry[1];
+						}
+					}
+					fh.close();
+				}
+
+				data.hostnames = bat_hosts;
+				const uci = cursor();
+				uci.load("network");
+				uci.foreach('network', 'interface', (s) => {
+					if (s.proto == 'batadv'){
+						device = s['.name'];
+						let out = command(`batctl ${device} mj 2>/dev/null`);
+						if (out) {
+							data.devices[device] = json(out);
+							out = command(`batctl ${device} nj 2>/dev/null`);
+							data.devices[device].neighbors = out ? json(out) : [];
+							out = command(`batctl ${device} hj 2>/dev/null`);
+							data.devices[device].hardifs = out ? json(out) : [];
+						}
+					}
+				});
+
+				uci.unload();
+			}
+
+			return data;
+		}
+	}
+};
+
+return { 'luci.batadv': methods };
--- a/protocols/luci-proto-batman-adv/root/usr/share/luci/menu.d/luci-proto-batman-adv.json
+++ b/protocols/luci-proto-batman-adv/root/usr/share/luci/menu.d/luci-proto-batman-adv.json
@@ -1,0 +1,14 @@
+{
+	"admin/status/batman-adv": {
+		"title": "Batman-adv",
+		"order": 92,
+		"action": {
+			"type": "view",
+			"path": "batman-adv/status"
+		},
+		"depends": {
+			"acl": [ "luci-proto-batman-adv" ],
+			"uci": { "network": true }
+		}
+	}
+}


protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js
```diff
@@ -0,0 +1,155 @@
...
+
+function msToStr(milliseconds){
+	let ago;
+
+	if (milliseconds < 1000)
+		ago = _('%dms ago').format(milliseconds);
+	else if (milliseconds < 60000)
+		ago = _('%ds ago').format(milliseconds / 1000);
+	else if (milliseconds > 60000)
+		ago = _('%dm ago').format(milliseconds / 60000);
```
openwrt-ai: A `last_seen_msecs` value of exactly `60000` matches neither `< 60000` nor `> 60000`, so `ago` stays unset and `msToStr` returns `undefined` (which then becomes the cell display value). Use `else` so the final bucket is inclusive:

```suggestion
		else
			ago = _('%dm ago').format(milliseconds / 60000);
```

---

protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
```diff
@@ -0,0 +1,55 @@
+// Copyright 2026 Jan Froch <jan@froch.eu>
+// Licensed to the public under the Apache License 2.0.
+
+'use strict';
+
+import { cursor } from 'uci';
+import { popen, open } from 'fs';
+
+const batctl_tiny = rtrim(popen(`apk info | grep batctl 2>/dev/null`).read('all'), '\n') == 'batctl-tiny';
```
openwrt-ai: This package-detection relies on `apk`, which only exists on apk-based images; on opkg-based OpenWrt installs `apk info` fails, the comparison is false, and the page would attempt the full `batctl ... mj/nj/hj` path even when batctl-tiny is the installed variant — the opposite of the intended fallback. Additionally, the exact-string match `== 'batctl-tiny'` is fragile: `apk info | grep batctl` can return multiple lines (e.g. both `batctl` and a `batctl-tiny`-named dep), in which case the equality breaks. Is detecting batctl-tiny robustly across both package managers in scope here, or is apk-only intentional? No other rpcd ucode backend in luci does package detection this way, so a more robust probe (e.g. checking whether `batctl ... mj` actually produces JSON) may be preferable.

---

protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
```diff
@@ -0,0 +1,55 @@
...
+
+				data.hostnames = bat_hosts;
+				const uci = cursor();
+				uci.load("network");
+				uci.foreach('network', 'interface', (s) => {
+					if (s.proto == 'batadv'){
+						device = s['.name'];
+						data.devices[device] = json(popen(`batctl ${device} mj 2>/dev/null`));
+						data.devices[device].neighbors = json(popen(`batctl ${device} nj 2>/dev/null`));
+						data.devices[device].hardifs = json(popen(`batctl ${device} hj 2>/dev/null`));
```
openwrt-ai: `json()` throws on premature EOF / parse errors, and `popen()` returns `null` if the spawn fails. So if any `batctl <dev> mj|nj|hj` produces empty or non-JSON output (command missing, table unavailable, error path silenced by `2>/dev/null`), `json(...)` raises an exception and aborts the entire `getBatadvInstances` call — the status page then shows no data for any instance, not just the failing one. Consider reading the output first and guarding before parsing, e.g. `let out = popen(...)?.read?.('all'); data.devices[device] = length(out) ? json(out) : {};` so one bad table doesn't take down the whole response.

---

protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js
```diff
@@ -0,0 +1,155 @@
...
+}
+
+return view.extend({
+	renderIfaces: function(ifaces) {
+		var res = [
+			E('h2', [ _('Batman-adv Status') ])
+		];
+
+		if(!ifaces.batctl_tiny){
+		for (var instanceName in ifaces.devices) {
```
openwrt-ai: nit: the body of this `if (!ifaces.batctl_tiny) {` block (the `for` loop and the trailing `if (res.length == 1)` check through line 116) is not indented one level deeper than the surrounding block, unlike the rest of the file. Re-indenting the block contents makes the branch structure readable.

---

openwrt-ai (commented):
Reviewed 1 new commit; posted inline findings on the status view and the rpcd ucode backend. Main items: a `msToStr` gap that returns `undefined` at exactly 60000 ms, apk-only/fragile batctl-tiny detection, and unguarded `json(popen(...))` calls that can abort the whole RPC when a `batctl` table is empty or fails.

---

protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
```diff
@@ -0,0 +1,65 @@
...
+import { cursor } from 'uci';
+import { popen, open } from 'fs';
+
+function command(cmd) {
+	return trim(popen(cmd)?.read?.('all'));
+}
+
+const apkproc = popen(`apk info 2>/dev/null | grep '^batctl-\\(default\\|full\\)$'`);
+apkproc.read('all');
+const batctl_debug_tables = !apkproc.close();
```
openwrt-ai, line 15: The rework addresses the earlier fragility (no more exact-string match, and `?.read?.` guards a failed spawn), but detection is still apk-only. On opkg-based images (OpenWrt <= 24.10, still supported) `apk info` does not exist, so the pipeline exits non-zero, `batctl_debug_tables` becomes `false`, and the page reports "none installed" even when `batctl-default`/`batctl-full` is in fact installed via opkg — the inverse of the intended behavior.

The in-tree convention for cross-package-manager detection is to branch on `access()`, see [`callPackageVersionCheck` in luci-base ...rpcd/ucode/luci:24-45](https://github.com/openwrt/luci/blob/5bbf45397756830f30205bd6242ea69e93ee8802/modules/luci-base/root/usr/share/rpcd/ucode/luci#L24-L45) (`access('/bin/opkg')` else `access('/usr/bin/apk')`).

A package-manager-agnostic alternative that sidesteps this entirely: probe the actual capability rather than the package — e.g. set `batctl_debug_tables` based on whether `batctl <dev> mj` produces JSON, or check for the presence of the debug binary/tables directly. Is apk-only intentional here, or should opkg be handled too?

---

openwrt-ai (commented):
Re-reviewed the changes since 10d39ce. Three of the four earlier findings are resolved (`msToStr` final bucket is now inclusive, `json(...)` calls are guarded via the new `command()` helper, indentation fixed). One concern from the prior round persists in the reworked code — see the inline comment on the package detection.

---

protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
JFroch: When merged would like to get this backported to 25.12, but 24.10 is not a goal, so I´d say using apk to check is not an issue.

Checking for JSON output is not a good alternative Imo as the command outputs JSON only if the batman device is actually online, if it hasn´t started yet or is disabled completely this check would fail, so the only alternative I see would be to parse the helptext for the availability of JSON commands.

@systemcrash what is your opinion on using apk to check for capability requirements in this case?
If there are no objections I would say this PR is ready for review.

systemcrash: I take that implicitly, while the AI might not have picked up on that. 

Until now, we've added individual binary detection in the luci ucode script which is used by feature detection, rather than asking apk for specific information.

What is the litmus test to know that the batman binary would have JSON support? If binaries after a certain date, then just check for the binary, and add a dep to the Makefile (mandating a specific version - check various Makefile for this syntax).

JFroch: The state right now is that this package depends on `kmod-batman-adv` which in turn depends on `batctl` which is provided by `batctl-default`, `batctl-full` or `batctl-tiny` of which the last one doesn´t have the statistics tables and JSON support. I would like to do a runtime check and just "disable" this new status page/inform the user in case `batctl-tiny` is installed as the protocol configuration part works fine with it.

I will have a look at the feature detection ucode script.

systemcrash: As an alternative, perhaps just try to call the 'ideal' variant which has JSON plumbing, and check the return code of the call. Non-zero indicates error conditions (binary absent or the binary itself returned error) - whereby popen returns null on error - so it should be possible toggle features based on that return value.

JFroch: The three packages are mutually exclusive as they provide the same `batctl` binary, just with different compile options, so something like `popen('batctl-default 2>/dev/null')` unfortunately isn´t possible.

What could be done is grep for the JSON queries in the helptext and use the return code of that:

<details>
  <summary>Relevant part of the helptext. This is empty on the '-tiny' binary.</summary>

```
JSON queries: display results of netlink query as JSON
 	meshif <netdev> bla_backbone_json|bbj      
 	meshif <netdev> bla_claim_json|clj         
 	meshif <netdev> dat_cache_json|dcj         
 	meshif <netdev> gateways_json|gwj          
 	hardif <netdev> hardif_json|hj             
 	meshif <netdev> hardifs_json|hj            
 	meshif <netdev> mcast_flags_json|mfj       
 	meshif <netdev> mesh_json|mj               
 	meshif <netdev> neighbors_json|nj          
 	meshif <netdev> originators_json|oj        
 	meshif <netdev> transtable_global_json|tgj 
 	meshif <netdev> transtable_local_json|tlj  
 	vlan <vdev> vlan_json|vj                   
 	meshif <netdev> vid <vid> vlan_json|vj 
```
</details>

```
const jsonproc = popen(`batctl 2>&1 >/dev/null | grep 'mesh_json'`);
jsonproc.read('all');
const batctl_debug_tables = !jsonproc.close();
```

systemcrash: Ah, I see. Is there no `--flag` to test for whose absence provokes a non-zero return? Or what is the return code if you try to query using one of the json params? Perhaps that requires knowing which netdev to query.

The makefile (i.e. install) can still mandate a particular flavour of the binary, however. (which simplifies code)

JFroch: Not that I know of, the only options are `-h` and `-v `. The return code of for example `batctl bat1 mj` can be used, but only if the interface is up, if it is down it also returns '1', the same as with the command not being available.

Yes, but that could lead to problems for users with `batctl-tiny` installed when trying to use ASU, no? (impossible package selection)
And also users that deliberately chose to install the tiny version for space reasons are forced to either install the heavier version, or are possibly not able to use luci to configure the interfaces anymore.
16k (tiny) vs 45k (default) is not that much of a difference, do you think that could have a real world impact?

At last it would be a possibility to just don´t check for JSON support and let it silently fail when -tiny is installed and change the text from the current `No Batman-adv interfaces configured or active.` to something like `No Batman-adv interfaces configured or active or the batctl version you have installed doesn´t expose statistics`.

protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
```diff
@@ -0,0 +1,65 @@
...
+				uci.foreach('network', 'interface', (s) => {
+					if (s.proto == 'batadv'){
+						device = s['.name'];
+						let out = command(`batctl ${device} mj 2>/dev/null`);
+						if (out) {
+							data.devices[device] = json(out);
+							out = command(`batctl ${device} nj 2>/dev/null`);
+							data.devices[device].neighbors = out ? json(out) : {};
+							out = command(`batctl ${device} hj 2>/dev/null`);
+							data.devices[device].hardifs = out ? json(out) : {};
```
openwrt-ai: `batctl <dev> nj` and `hj` emit JSON *arrays*, and the view calls `.map()` on both values ([`neighbors.map(...)` at status.js:44](``https://github.com/openwrt/luci/blob/f1c2196993c813fe667da264574a66acdb51955e/protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js#L44)`` and [`hardifs.map(...)` at status.js:75](``https://github.com/openwrt/luci/blob/f1c2196993c813fe667da264574a66acdb51955e/protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js#L75)).`` The `{}` fallback is an object, so on the path where `mj` succeeds but `nj`/`hj` returns nothing the view throws `neighbors.map is not a function` and the entire page renders blank instead of showing an empty table. An empty array degrades to the table placeholder instead.

```suggestion
							out = command(`batctl ${device} nj 2>/dev/null`);
							data.devices[device].neighbors = out ? json(out) : [];
							out = command(`batctl ${device} hj 2>/dev/null`);
							data.devices[device].hardifs = out ? json(out) : [];
```


---

protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js
```diff
@@ -0,0 +1,155 @@
...
+		E('em', [
+			_('No neighbors connected')
+		])
+	);
+
+	t.update(neighbors.map(function(peer) {
+		return [
+			hostnames[peer.neigh_address] ? `${peer.neigh_address} (${hostnames[peer.neigh_address]})` : peer.neigh_address,
+			[ +peer.last_seen_msecs, msToStr(peer.last_seen_msecs)],
+			[ +peer.throughput, '%.1mb/s'.format(+peer.throughput*1000) ],
```
openwrt-ai: The scale factor disagrees with the one used for the very same unit on line 82 of this file, which multiplies by `100000` and documents why (`//Value is reported in 100kbit/s`). batman-adv reports `throughput` (neighbours) and `throughput_override` (hardif) in the same 100 kbit/s units, so `*1000` renders values 100x too low — a 10 Mbit/s neighbour shows up as `100.0 Kb/s`. (`%m` in LuCI's formatter divides by 1000 per prefix step, see [`case 'm'` in cbi.js:819](https://github.com/openwrt/luci/blob/f1c2196993c813fe667da264574a66acdb51955e/modules/luci-base/htdocs/luci-static/resources/cbi.js#L819).)

```suggestion
			[ +peer.throughput, '%.1mb/s'.format(+peer.throughput*100000) ],
```


---

protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js
```diff
@@ -0,0 +1,155 @@
...
+	var t = new L.ui.Table(
+		[
+			_('Neighbor'),
+			_('Last seen'),
+			_('Reported Throughput'),
+			_('Connected on Hardware Interface')
+		],
+		{
+			id: 'neighs-' + instanceName,
+			placeholder: 'No neighbors connected'
```
openwrt-ai: nit: this user-visible string is not wrapped in `_()`, and it is the one that actually reaches the screen: `ui.Table.update()` resolves the placeholder as `placeholderText ?? this.options.placeholder ?? …` ([ui.js:3859](https://github.com/openwrt/luci/blob/f1c2196993c813fe667da264574a66acdb51955e/modules/luci-base/htdocs/luci-static/resources/ui.js#L3859)) and re-renders it on every poll ([ui.js:3937-3941](https://github.com/openwrt/luci/blob/f1c2196993c813fe667da264574a66acdb51955e/modules/luci-base/htdocs/luci-static/resources/ui.js#L3937-L3941)), so `options.placeholder` wins over the translated `E('em', [ _('No neighbors connected') ])` node passed as the third constructor argument. Same applies to `'No hardifs associated'` on line 68 and `'Not overridden'` on line 82. (The in-tree sibling this view is modelled on, [luci-proto-wireguard status.js:89-94](``https://github.com/openwrt/luci/blob/f1c2196993c813fe667da264574a66acdb51955e/protocols/luci-proto-wireguard/htdocs/luci-static/resources/view/wireguard/status.js#L89-L94),`` passes only `id` in the options object and relies on the third argument.)

```suggestion
			placeholder: _('No neighbors connected')
```


---

protocols/luci-proto-batman-adv/root/usr/share/rpcd/acl.d/luci-proto-batman-adv.json
```diff
@@ -0,0 +1,14 @@
...
+		"description": "Grant access to LuCI Batman-adv procedures",
+		"read": {
+			"ubus": {
+				"luci.batadv": [
+					"getBatadvInstances"
+				]
+			}
+		},
+		"write": {
+		}
```
openwrt-ai: nit: the empty `"write"` object grants nothing and has no precedent in tree — read-only ACL files simply omit the key, e.g. [`luci-mod-dsl.json`](``https://github.com/openwrt/luci/blob/f1c2196993c813fe667da264574a66acdb51955e/modules/luci-mod-dsl/root/usr/share/rpcd/acl.d/luci-mod-dsl.json#L1-L10).``

```suggestion
		}
```


---

openwrt-ai (commented):
Re-reviewed at f1c2196. The four package files are byte-identical to what I reviewed at 5bbf453 — the new head is a rebase onto current master (the only delta under `protocols/luci-proto-batman-adv/` is master's own `getOpkgPackage` → `getPackageName` rename in `batadv_vlan.js`, untouched by this PR). The open `batctl` capability-detection thread is unchanged and still awaiting a decision; I have not re-posted on it.

Four findings on unchanged code that I had not raised before are inline: an object-vs-array fallback in the rpcd backend that can throw in the view, a throughput scale factor that disagrees with the one used for the same unit elsewhere in the same file, and two nits (untranslated table placeholders, empty ACL `write` block).

No commit-message issues; the single commit's subject and body match its diff.

---

protocols/luci-proto-batman-adv/htdocs/luci-static/resources/view/batman-adv/status.js
JFroch: While the hardif throughput override is reported in 100kbit/s the neighbor throughput is reported in kbit/s.

openwrt-ai (commented):

The delta since f1c2196 is limited to the three fixups for my previous round — `[]` instead of `{}` for the `nj`/`hj` fallbacks, `_()` around the `placeholder` / `Not overridden` strings, and the removal of the empty ACL `write` block — plus the two unit comments on the throughput columns. All of those check out, and I accept the clarification that neighbour `throughput` and hardif `throughput_override` are reported in different units; I could not reach the batman-adv sources from this environment to verify independently, so I am not re-opening that thread.

The `batctl` capability-detection thread (apk-only probe, evaluated once at rpcd script load) is still the only open question from earlier rounds and is unchanged in this head; I have not re-posted on it.

The failing `eslint` check on this head is not caused by this PR — the reported errors are in `luci-app-adblock/.../overview.js:676` and warnings in `luci-app-lldpd/.../config.js`, which are picked up because the branch contains master commits ahead of the PR base SHA. `protocols/luci-proto-batman-adv/` lints clean.

---

protocols/luci-proto-batman-adv/root/usr/share/rpcd/ucode/luci.batadv
JFroch: @hnyman could you have a look and give advice on how I should implement this feature detection? Thank you!

PR #8638, opened 2026-05-18
Alternative view for configuring switch ports and VLANs

mirko:
Seeing OpenWrt supporting more and more dedicated switches with more than just a handful of ports, the current interface for configuring and maintaining ports and VLANs becomes cumbersome to use, especially with frequent changes.

Current (personal) pain points are:
 - To reach the current switch/VLAN config page, one has to click through: `Network` -> `Interfaces` -> `Devices` -> `Configure... (the respective switch/bridge)` -> `Bridge VLAN filtering` which - at least for dedicated switches - is not sth. what I feel comfortable having to do every time I want to make a change 
 - While ~5 ports might still fit next to each other, starting with 2 figure numbers, we end up scrolling horizontally. Finding the port to configure on a 24 or 48 port switch will result in plenty of horizontal scrolling, while at the same time you have to memorise which row maps to which VLAN-ID, as those scroll out the moment you scroll right 
 - Every port needs to be configured by its own, no way to bulk-conf them, apply the same VLAN config onto multiple ports (e.g. applying the same tag onto a large number or all ports)
 - Ports and VLAN-IDs are defined by sole numbers. There's no way to annotate them, e.g. a VLAN ID as "iot" or "mgmt", or a port as e.g. "uplink" or "server-mirko".

### Disclaimer
**Large parts of the code involved in this PR were created/generated with the help of LLMs. It was not properly reviewed and I don't expect anyone to do so with potential AI-slop thrown at.

This PR is merely a starting point for a discussion on how to address some of the pain points and provide a mockup -- or prototype at best -- showing how I can imagine seeing those issues being tackled.
**

### Description
The disclaimer above being stated: This PR provides a functional prototype for how I personally think the/my in the beginning stated pain points could be addressed by the following changes / additions:

 - Adding a dedicated "Switch/VLAN config" entry within the toplevel "Network"-dropdown (you might remember we already had such for swconfig back then, before it got obsoleted by DSA and configuration via vlan aware bridges)
 - Avoiding horizontal scrolling by showing switch ports in multiple rows if not fitting screen width
 - Allowing selection of multiple ports to configure them all the same way at once (includes easily removing all memberships from all selected ports with just one click)
 - Introducing labels for ports and VLANs -- they're optional annotations and don't serve any functional purpose except hinting the user, stored in respective sections within /etc/config/luci 

owrt-luci-swcfg_2

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/switchconfig.css
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/switchconfig.css
@@ -1,0 +1,391 @@
+/*
+ * Switch config view — port-centric VLAN assignment
+ * Injected as a stylesheet <link> in switchconfig.js load() (LuCI has no L.loadCSS).
+ * Scoped under .switchconfig-pc to avoid polluting other views.
+ */
+
+/* ------------------------------------------------------------------ layout */
+
+/*
+ * Reserve bottom padding for the fixed .assign-dock.
+ * LuCI nests .switchconfig-pc inside #maincontent, so the fixed dock
+ * sits relative to the viewport, not this element.
+ */
+.switchconfig-pc {
+	padding-bottom: clamp(14rem, 38vh, 26rem);
+	box-sizing: border-box;
+}
+
+/* ------------------------------------------------------------------ assign dock (fixed bottom bar) */
+
+.switchconfig-pc .assign-dock {
+	position: fixed;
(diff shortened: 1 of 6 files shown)


mirko:
This should have been explicitly a draft -- and conversion didn't appear to work. Hence closing in favour of https://github.com/openwrt/luci/pull/8639

PR #8639, opened 2026-05-18
Alternative view for configuring switch ports and VLANs

mirko:
Seeing OpenWrt supporting more and more dedicated switches with more than just a handful of physical ports, the current interface for configuring and maintaining ports and VLANs becomes cumbersome to use, especially with frequent changes.

### Description
Register a new "Switch/VLAN config" entry in the Network top-level
menu and add the view at view/network/switch-vlan.js.

The menu entry is only visible when the network configuration
contains either a bridge device with vlan_filtering explicitly
enabled or at least one bridge-vlan section (mirroring the implicit
enable behaviour of the existing Bridge VLAN filtering dialog).

A warning about this yet being an experimental view which might
eat all your traffic is shown upon accessing this view.

### Disclaimer
**Large parts of the code involved in this PR were created/generated with the help of LLMs. Given my (lack of) JavaScript skills, me having reviewed the code to the best of my abilities might not be enough.**

### Current (personal) pain points are:
 - To reach the current switch/VLAN config page, one has to click
   through: Network -> Interfaces -> Devices -> Configure... (the
   respective switch/bridge) -> Bridge VLAN filtering which - at
   least for dedicated switches - is not sth. I feel comfortable
   having to do every time I want to make a basic change.

 - While ~5 ports might still fit next to each other, starting with
   2 figure numbers (at the latest), we end up scrolling
   horizontally. Finding a higher port to configure on a 24 or 48
   port switch inevitably results in plenty of horizontal scrolling,
   while at the same time you have to memorise which row maps to
   which VLAN-ID, as those scroll out the moment you start
   scrolling.

 - Every port needs to be configured by its own, no way to bulk-conf
   them, to apply the same VLAN config onto multiple ports (e.g.
   applying the same tag onto a large number or all ports).

 - Ports and VLAN-IDs are defined by sole numbers. There's no way to
   annotate them, e.g. a VLAN ID as "iot" or "mgmt", or a port as
   e.g. "uplink" or "server-mirko".

### How this PR tries to address above stated issues
**The disclaimer above being stated:** This PR provides a functional and isolated component (no changes to other parts of code) for how I personally think the/my in the beginning stated pain points could be addressed by the following changes / additions:
 - Add a dedicated "Switch/VLAN config" entry within the top-level
   "Network" dropdown (you might remember we already had such for
   swconfig back then, before it got obsoleted by DSA and conf via
   vlan aware bridges).

 - Avoid horizontal scrolling by showing switch ports in multiple
   rows if not fitting screen width.

 - Allow selection of multiple ports to configure them all the same
   way at once.

 - Introduce labels for ports and VLANs; they're optional annotations
   and don't serve any functional purpose except hinting the user,
   hence stored in respective sections within /etc/config/luci.

 - Visualising which VLANs are assigned to which ports by colour-
   coding the ports when hovering over configured VLANs.

### Discussion points
 - There's another pull request regarding port labels:
   https://github.com/openwrt/luci/pull/8579 -- however we want
   port and VLAN labels. Not sure how to combine those efforts.

 - Shall we only allow this view for dedicated switch devices?
   We don't have a device class "switch" or alike yet, but could
   easily depend on the "realtek" target, as it's the only supported
   platform currently.

 - Identifying the physical switch: apart from the possibility of
   there being more than one switch, how to identify an appropriate
   one:
    - Go for the bridge named "switch" by sole naming convention.
    - Filter for vlan-aware bridges (which itself is not well
      defined unfortunately), and filter for ony eth/sfp ports, so
      we don't include e.g. wifi ifaces), -- as does the "Bridge
      VLAN filtering" view -- and choose the first one from the list.
    - Filter for physical ports as the "Port status" view does
      (Overview -> Status) and choose the bridge all of them are
      members of; would call luci.getBuiltinEthernetPorts() and
      hence need further ACL permissions.

 - Stashing / saving changes: when and how to stash changes.
   Narrowed down, there are 2 obvious options:
    - uci.save() on every click and change - logging all the
      changes for review when clicking on "Unsaved changes" in the
      top-right corner.
      Due to the uci structure fo configuring VLANs being a bit un-
      fortunate, back-and-forth changes can't be easily tracked but
      stack up.
    - Stash changes inside luci only until being "Saved", which
      then saves them all at once. That works well for models
      where the underlying view has the "Save & Apply" / "Save" /
      "Reset" button-triple, but not really for a full-page view.

 - Should the "local"-flag be en- or disabled by default?

 - We currently do not support configurations of ingress != egress
   VLAN (or in other terms: PVID / primary VLAN != untagged VLAN)

Short screencast showing base features:
https://github.com/user-attachments/assets/94e09cff-3565-4707-82a9-de8971a8f126

Screenshot:
openwrt-luci-switch-vlan-config

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
@@ -1,0 +1,494 @@
+/* Helper module for the bridge-vlan UCI model.
+ *
+ * Provides the read side (find/parse/enumerate bridge-vlan sections and their
+ * port specs) and the write side (anti-stacking port-list writer plus an
+ * out-of-band store for user-defined port/VLAN labels in /etc/config/luci).
+ *
+ * Used by the Switch/VLAN view (view/network/switch-vlan.js); kept separate
+ * from the view so the UCI/port-spec primitives are reusable. */
+
+'use strict';
+'require uci';
+'require network';
+'require baseclass';
+
+const MIN_VLAN_ID = 1;
+const MAX_VLAN_ID = 4094;
+
+const LUCI_LABEL_SECTION_TYPE = 'switchvlan';
+const LUCI_PORT_LABELS_NAME = 'port_labels';
+const LUCI_VLAN_LABELS_NAME = 'vlan_labels';
+
+/* netifd defaults a bridge-vlan's "local" flag to true when the option is
+ * absent (see config_init_vlan_entry() in netifd's config.c), so an unset
+ * value must be read as local, not the other way around. Parse the UCI value
+ * as a proper boolean and treat absence as true. */
+function parseLocal(value) {
+	if (value == null || value === '')
+		return true;
+
+	switch (String(value).toLowerCase()) {
+	case '0':
+	case 'false':
+	case 'no':
+	case 'off':
+	case 'disabled':
+		return false;
+	default:
+		return true;
+	}
+}
+
+/* =============================================================================
+ * Port spec parsing & formatting
+ * ============================================================================= */
+
+function parsePortSpec(spec) {
+	const m = String(spec).match(/^([^:]+)(?::([ut*]*))?$/);
+
+	if (!m)
+		return null;
+
+	const flags = m[2] || '';
+
+	return {
+		port: m[1],
+		tagged: /t/.test(flags),
+		untagged: /u/.test(flags) || flags === '',
+		pvid: /\x2a/.test(flags) || flags === ''
+	};
+}
+
+function formatPortSpec(port, role) {
+	if (role === 'untagged' || role === 'native')
+		return port;
+
+	if (role === 'tagged')
+		return '%s:t'.format(port);
+
+	return null;
+}
+
+/* =============================================================================
+ * Bridge selection & enumeration
+ * ============================================================================= */
+
+function getBridgeDeviceName(deviceSection) {
+	if (!deviceSection)
+		return null;
+
+	return deviceSection.name || deviceSection['.name'] || null;
+}
+
+/* uci.get is mandatory here: the .ports field on a uci.sections() callback
+ * argument is a snapshot and goes stale after uci.set/unset on this option. */
+function getVlanSectionPorts(sectionId) {
+	return L.toArray(uci.get('network', sectionId, 'ports'));
+}
+
+function isVlanFilteringEnabled(deviceSection) {
+	if (!deviceSection)
+		return false;
+
+	if (deviceSection.type !== 'bridge')
+		return false;
+
+	if (deviceSection.vlan_filtering === '1')
+		return true;
+
+	const devname = getBridgeDeviceName(deviceSection);
+	let has_vlans = false;
+
+	uci.sections('network', 'bridge-vlan', function(bvs) {
+		if (bvs.device == devname)
+			has_vlans = true;
+	});
+
+	return has_vlans;
+}
+
+/* A bridge qualifies only when every member resolves to a physical
+ * ethernet/DSA port. Bridges spanning wifi APs, tunnels, wireguard, VLAN
+ * sub-interfaces or nested bridges are excluded on purpose — they're legal
+ * vlan-aware bridges but the T/U-on-physical-ports UI doesn't apply. */
+function findActiveBridge() {
+	let match = null;
+
+	uci.sections('network', 'device', s => {
+		if (match || !isVlanFilteringEnabled(s))
+			return;
+
+		const ports = collectBridgePorts(s);
+
+		if (!ports.length)
+			return;
+
+		const allSwitchworthy = ports.every(p => {
+			const t = network.instantiateDevice(p).getType();
+			return t === 'switch' || t === 'ethernet';
+		});
+
+		if (allSwitchworthy)
+			match = s;
+	});
+
+	return match;
+}
+
+function collectBridgePorts(deviceSection) {
+	const seen = {};
+	const devname = getBridgeDeviceName(deviceSection);
+	const section_id = deviceSection ? deviceSection['.name'] : null;
+
+	if (!deviceSection)
+		return [];
+
+	L.toArray(uci.get('network', section_id, 'ports')).forEach(function(port) {
+		seen[port] = true;
+	});
+
+	if (devname) {
+		const br = network.instantiateDevice(devname);
+		const brports = (br && typeof br.getPorts === 'function') ? br.getPorts() : null;
+
+		if (brports) {
+			brports.forEach(function(portdev) {
+				const name = portdev ? portdev.getName() : null;
+
+				if (name)
+					seen[name] = true;
+			});
+		}
+
+		uci.sections('network', 'bridge-vlan', function(bvs) {
+			if (bvs.device != devname)
+				return;
+
+			getVlanSectionPorts(bvs['.name']).forEach(function(spec) {
+				const m = String(spec).match(/^([^:]+)(?::[ut*]+)?$/);
+
+				if (m)
+					seen[m[1]] = true;
+			});
+		});
+	}
+
+	return Object.keys(seen).sort(L.naturalCompare);
+}
+
+function collectBridgeVlans(deviceSection) {
+	const out = [];
+
+	if (!deviceSection)
+		return out;
+
+	uci.sections('network', 'bridge-vlan', function(bvs) {
+		if (bvs.device != getBridgeDeviceName(deviceSection))
+			return;
+
+		const vid = +bvs.vlan;
+
+		if (!(vid >= MIN_VLAN_ID && vid <= MAX_VLAN_ID))
+			return;
+
+		out.push({
+			section_id: bvs['.name'],
+			vlan: vid,
+			local: parseLocal(bvs.local),
+			ports: getVlanSectionPorts(bvs['.name']).slice()
+		});
+	});
+
+	out.sort(function(a, b) { return a.vlan - b.vlan; });
+
+	return out;
+}
+
+/* =============================================================================
+ * VLAN configuration validation
+ * ============================================================================= */
+
+function checkUnsupportedConfig(deviceSection) {
+	const issues = [];
+
+	if (!deviceSection)
+		return issues;
+
+	const portUntagged = {};
+
+	uci.sections('network', 'bridge-vlan', function(bvs) {
+		if (bvs.device != getBridgeDeviceName(deviceSection))
+			return;
+
+		const vid = +bvs.vlan;
+
+		if (!(vid >= MIN_VLAN_ID && vid <= MAX_VLAN_ID)) {
+			issues.push(_('VLAN ID %s is outside the valid range %d–%d.').format(
+				bvs.vlan, MIN_VLAN_ID, MAX_VLAN_ID));
+			return;
+		}
+
+		getVlanSectionPorts(bvs['.name']).forEach(function(spec) {
+			const parsed = parsePortSpec(spec);
+
+			if (!parsed) {
+				issues.push(_('Port specification "%s" on VLAN %d cannot be parsed.').format(spec, vid));
+				return;
+			}
+
+			/* Order matters: a spec carrying both the tagged and PVID flags
+			 * (e.g. "lan1:t*") is reported as the PVID-with-tagged-egress
+			 * case, not the tagged-and-untagged one. */
+			if (parsed.tagged && parsed.pvid)
+				issues.push(_('Port "%s" has the primary VLAN flag together with tagged egress on VLAN %d (PVID ≠ untagged VLAN).').format(parsed.port, vid));
+			else if (parsed.tagged && parsed.untagged)
+				issues.push(_('Port "%s" is both tagged and untagged on VLAN %d.').format(parsed.port, vid));
+			else if (parsed.pvid && !parsed.tagged && !parsed.untagged)
+				issues.push(_('Port "%s" is the primary VLAN on VLAN %d but has no untagged egress (PVID ≠ untagged VLAN).').format(parsed.port, vid));
+
+			if (parsed.untagged) {
+				if (portUntagged[parsed.port] != null && portUntagged[parsed.port] !== vid) {
+					issues.push(_('Port "%s" is configured as untagged on more than one VLAN (%d and %d).').format(
+						parsed.port, portUntagged[parsed.port], vid));
+				}
+
+				portUntagged[parsed.port] = vid;
+			}
+		});
+	});
+
+	return issues;
+}
+
+/* =============================================================================
+ * Per-port state derivation (native + tagged VLANs for each member)
+ * ============================================================================= */
+
+function buildPortState(deviceSection, ports) {
+	const state = {};
+
+	ports.forEach(function(name) {
+		state[name] = { name: name, native: null, tagged: [] };
+	});
+
+	if (!deviceSection)
+		return state;
+
+	uci.sections('network', 'bridge-vlan', function(bvs) {
+		if (bvs.device != getBridgeDeviceName(deviceSection))
+			return;
+
+		const vid = +bvs.vlan;
+
+		getVlanSectionPorts(bvs['.name']).forEach(function(spec) {
+			const parsed = parsePortSpec(spec);
+
+			if (!parsed || !state[parsed.port])
+				return;
+
+			if (parsed.untagged)
+				state[parsed.port].native = vid;
+			else if (parsed.tagged)
+				state[parsed.port].tagged.push(vid);
+		});
+	});
+
+	for (let name in state)
+		state[name].tagged.sort(function(a, b) { return a - b; });
+
+	return state;
+}
+
+/* =============================================================================
+ * Port/VLAN label store (out-of-band, in /etc/config/luci)
+ * ============================================================================= */
+
+function labelOptionKey(name) {
+	return String(name).replace(/[^A-Za-z0-9_]/g, '_');
+}
+
+function ensureLabelSections() {
+	let port_sid = null, vlan_sid = null;
+
+	uci.sections('luci', LUCI_LABEL_SECTION_TYPE, function(s) {
+		if (s['.name'] === LUCI_PORT_LABELS_NAME)
+			port_sid = s['.name'];
+		else if (s['.name'] === LUCI_VLAN_LABELS_NAME)
+			vlan_sid = s['.name'];
+	});
+
+	if (!port_sid)
+		port_sid = uci.add('luci', LUCI_LABEL_SECTION_TYPE, LUCI_PORT_LABELS_NAME);
+
+	if (!vlan_sid)
+		vlan_sid = uci.add('luci', LUCI_LABEL_SECTION_TYPE, LUCI_VLAN_LABELS_NAME);
+
+	return { port: port_sid, vlan: vlan_sid };
+}
+
+function readPortLabel(portName) {
+	const value = uci.get('luci', LUCI_PORT_LABELS_NAME, labelOptionKey(portName));
+	return value != null ? String(value) : '';
+}
+
+function readVlanLabel(vlanId) {
+	const value = uci.get('luci', LUCI_VLAN_LABELS_NAME, labelOptionKey(vlanId));
+	return value != null ? String(value) : '';
+}
+
+function writePortLabel(portName, label) {
+	ensureLabelSections();
+
+	const key = labelOptionKey(portName);
+
+	if (label == null || label === '')
+		uci.unset('luci', LUCI_PORT_LABELS_NAME, key);
+	else
+		uci.set('luci', LUCI_PORT_LABELS_NAME, key, label);
+}
+
+function writeVlanLabel(vlanId, label) {
+	ensureLabelSections();
+
+	const key = labelOptionKey(vlanId);
+
+	if (label == null || label === '')
+		uci.unset('luci', LUCI_VLAN_LABELS_NAME, key);
+	else
+		uci.set('luci', LUCI_VLAN_LABELS_NAME, key, label);
+}
+
+/* =============================================================================
+ * Bridge-vlan ports writer (anti-stacking; bypasses LuCI's missing
+ * "undo pending change" API by poking uci.state directly)
+ * ============================================================================= */
+
+function vlanPortsEqual(a, b) {
+	const norm = list => L.toArray(list).slice().sort(L.naturalCompare);
+	const sa = norm(a);
+	const sb = norm(b);
+	return sa.length === sb.length && sa.every((v, i) => v === sb[i]);
+}
+
+function getBaseVlanPorts(section_id) {
+	if (uci.state.creates.network?.[section_id])
+		return L.toArray(uci.state.creates.network[section_id].ports);
+	return L.toArray(uci.state.values.network?.[section_id]?.ports);
+}
+
+function clearPendingVlanPorts(section_id) {
+	const c = uci.state.changes.network;
+	const d = uci.state.deletes.network;
+
+	if (c?.[section_id]) {
+		delete c[section_id].ports;
+		if (!Object.keys(c[section_id]).length)
+			delete c[section_id];
+	}
+
+	if (c && !Object.keys(c).length)
+		delete uci.state.changes.network;
+
+	if (d?.[section_id] && d[section_id] !== true) {
+		delete d[section_id].ports;
+		if (!Object.keys(d[section_id]).length)
+			delete d[section_id];
+	}
+
+	if (d && !Object.keys(d).length)
+		delete uci.state.deletes.network;
+}
+
+/* Anti-stacking writer for bridge-vlan ports. Two non-obvious behaviours:
+ *  1. If the new list equals the loaded baseline, the local pending change
+ *     is *cleared* (poking uci.state directly via clearPendingVlanPorts —
+ *     LuCI's uci module has no public "undo a pending change" API). Without
+ *     this, toggling T then T leaves a phantom entry in "Unsaved Changes".
+ *  2. An empty list calls uci.unset, not uci.set([]); OpenWrt UCI rejects
+ *     empty list values on set. */
+function setBridgeVlanPorts(section_id, ports) {
+	const normalized = L.toArray(ports).slice().sort(L.naturalCompare);
+	const base = getBaseVlanPorts(section_id);
+
+	if (vlanPortsEqual(normalized, base)) {
+		if (uci.state.creates.network?.[section_id]) {
+			if (base.length)
+				uci.set('network', section_id, 'ports', base.slice().sort(L.naturalCompare));
+			else
+				uci.unset('network', section_id, 'ports');
+		}
+		else {
+			clearPendingVlanPorts(section_id);
+		}
+		return;
+	}
+
+	if (normalized.length)
+		uci.set('network', section_id, 'ports', normalized);
+	else
+		uci.unset('network', section_id, 'ports');
+}
+
+/* =============================================================================
+ * Label maintenance
+ * ============================================================================= */
+
+function pruneOrphanLabels(validPorts, validVlans) {
+	const portKeys = {}, vlanKeys = {};
+
+	validPorts.forEach(function(p) { portKeys[labelOptionKey(p)] = true; });
+	validVlans.forEach(function(v) { vlanKeys[labelOptionKey(v)] = true; });
+
+	const toRemove = [];
+
+	uci.sections('luci', LUCI_LABEL_SECTION_TYPE, function(s) {
+		const isPortSection = (s['.name'] === LUCI_PORT_LABELS_NAME);
+		const isVlanSection = (s['.name'] === LUCI_VLAN_LABELS_NAME);
+
+		if (!isPortSection && !isVlanSection)
+			return;
+
+		const keys = isPortSection ? portKeys : vlanKeys;
+
+		for (let opt in s) {
+			if (opt.charAt(0) === '.')
+				continue;
+
+			if (!keys[opt])
+				toRemove.push({ section: s['.name'], option: opt });
+		}
+	});
+
+	toRemove.forEach(function(entry) {
+		uci.unset('luci', entry.section, entry.option);
+	});
+
+	return toRemove.length;
+}
+
+return baseclass.extend({
+	MIN_VLAN_ID: MIN_VLAN_ID,
+	MAX_VLAN_ID: MAX_VLAN_ID,
+	LUCI_LABEL_SECTION_TYPE: LUCI_LABEL_SECTION_TYPE,
+	LUCI_PORT_LABELS_NAME: LUCI_PORT_LABELS_NAME,
+	LUCI_VLAN_LABELS_NAME: LUCI_VLAN_LABELS_NAME,
+
+	parsePortSpec: parsePortSpec,
+	formatPortSpec: formatPortSpec,
+	parseLocal: parseLocal,
+	isVlanFilteringEnabled: isVlanFilteringEnabled,
+	findActiveBridge: findActiveBridge,
+	collectBridgePorts: collectBridgePorts,
+	collectBridgeVlans: collectBridgeVlans,
+	checkUnsupportedConfig: checkUnsupportedConfig,
+	buildPortState: buildPortState,
+	labelOptionKey: labelOptionKey,
+	ensureLabelSections: ensureLabelSections,
+	readPortLabel: readPortLabel,
+	readVlanLabel: readVlanLabel,
+	writePortLabel: writePortLabel,
+	writeVlanLabel: writeVlanLabel,
+	pruneOrphanLabels: pruneOrphanLabels,
+	vlanPortsEqual: vlanPortsEqual,
+	setBridgeVlanPorts: setBridgeVlanPorts
+});
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
@@ -1,0 +1,583 @@
+/* Switch/VLAN configuration view */
+
+/* Conflict notifications via ui.addNotification (flex layout; theme zeros ul padding) */
+#maincontent > .alert-message .svc-vlan-conflict {
+	min-width: 0;
+	width: 100%;
+}
+
+#maincontent > .alert-message .svc-vlan-conflict > p {
+	margin: 0 0 0.35em;
+}
+
+#maincontent > .alert-message .svc-vlan-conflict-ports {
+	margin: 0.35em 0 0;
+	padding: 0;
+}
+
+#maincontent > .alert-message .svc-vlan-conflict-port {
+	margin: 0.15em 0 0;
+	padding: 0;
+}
+
+#maincontent > .alert-message .svc-vlan-conflict-port::before {
+	content: '•\00a0';
+}
+
+#maincontent > .alert-message .svc-vlan-conflict ul {
+	margin: 0.35em 0 0;
+	padding-left: 1.35em;
+	list-style-position: inside;
+}
+
+#switch-vlan-view {
+	--svc-port-selected-bg: #e8f4fc;
+	--svc-port-selected-border: #3c7aa8;
+	--svc-port-selected-ring: rgba(60, 141, 188, 0.28);
+	/* Teal + amber-orange: distinguishable from grey port text */
+	--svc-vlan-untagged: #0d9488;
+	--svc-vlan-tagged: #d97706;
+	--svc-vlan-untagged-bg: rgba(13, 148, 136, 0.12);
+	--svc-vlan-tagged-bg: rgba(217, 119, 6, 0.14);
+	--svc-vlan-untagged-stripe: rgba(13, 148, 136, 0.2);
+	--svc-vlan-tagged-stripe: rgba(217, 119, 6, 0.2);
+}
+
+:root[data-darkmode="true"] #switch-vlan-view {
+	--svc-port-selected-bg: rgba(126, 184, 232, 0.22);
+	--svc-port-selected-border: #b3d4f0;
+	--svc-port-selected-ring: rgba(126, 184, 232, 0.45);
+	--svc-vlan-untagged: #2dd4bf;
+	--svc-vlan-tagged: #fbbf24;
+	--svc-vlan-untagged-bg: rgba(45, 212, 191, 0.18);
+	--svc-vlan-tagged-bg: rgba(251, 191, 36, 0.2);
+	--svc-vlan-untagged-stripe: rgba(45, 212, 191, 0.32);
+	--svc-vlan-tagged-stripe: rgba(251, 191, 36, 0.36);
+}
+
+#switch-vlan-view .svc-block {
+	margin: 1em 0;
+}
+
+.svc-block-header {
+	display: flex;
+	flex-wrap: wrap;
+	align-items: center;
+	justify-content: space-between;
+	gap: 0.5em;
+	margin-bottom: 0.5em;
+}
+
+.svc-block-header h3 {
+	margin: 0;
+	font-size: 1.1em;
+}
+
+.svc-port-header-actions {
+	display: inline-flex;
+	flex-wrap: wrap;
+	align-items: center;
+	gap: 0.4em;
+}
+
+#switch-vlan-view .cbi-section-descr {
+	margin-bottom: 1em;
+}
+
+.svc-empty {
+	margin: 0.5em 0;
+	color: #888;
+	font-style: italic;
+}
+
+/* Port tiles */
+.svc-ports {
+	display: grid;
+	grid-template-columns: repeat(auto-fill, minmax(11em, 1fr));
+	gap: 0.5em;
+}
+
+.svc-port-tile {
+	border: 1px solid #ccc;
+	border-radius: 4px;
+	padding: 0.5em 0.6em;
+	background: var(--background-color-medium, #fafafa);
+	color: var(--text-color-high, #000);
+	cursor: pointer;
+	user-select: none;
+	display: flex;
+	flex-direction: column;
+	gap: 0.35em;
+	transition: background 0.05s linear, border-color 0.05s linear, box-shadow 0.05s linear;
+	min-width: 0;
+}
+
+.svc-port-tile:hover {
+	border-color: #888;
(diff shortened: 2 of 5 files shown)


systemcrash:
@jow- @hnyman @feckert @efahl any thoughts?

I think this makes handling (tagged vs untagged) VLANs more intuitive. Labels are a nice addition which another PR recently proposed.

efahl:
Yes, I like it.  Should get @psherman to look at it, as he's the primary VLAN support giver on the forums and should be able to provide a very nuanced review.  Does he have a github account?

mirko:
@systemcrash could you point me to the other concept of labels you're referring to?

JFroch:
I think this would be a really nice addition and want to propose another feature:
In case that lldpd is installed the name of devices detected on a port could also be shown like `lldp: AP12`, similar to the new label field. That way you can instantly see for which connected device the vlans are edited.

lldpctl already has an option to output as json (`lldpctl -f json`) so adding the info should be fairly straighforward.

jow-:
The UX concept is nice, the design of the port membership cards could need some improvements. Right now every single line of text has a different font and color there, makes it look cluttered and somewhat unfinished.

What about something resembling the existing port status tiles with different color variations:

image

mirko:
> The UX concept is nice, the design of the port membership cards could need some improvements. Right now every single line of text has a different font and color there, makes it look cluttered and somewhat unfinished.
> 
> What about something resembling the existing port status tiles with different color variations:
> image

I'm all in for improvements! I'm really the wrong person to do UI/UX and being partially color-blind doesn't help either.
Main incentive was driven by not having seen any intuitive switch configuration panel so far and the above mentioned pain points with the current "Bridge VLAN filtering"-view.
Your approach definitely looks better - will try to combine things. Also I'm currently identifying and addressing the parts the LLM just went rogue and generated convoluted code.

@jow- do you have above posted example as html / css version? if so, please pass it to me!

efahl:
> being partially color-blind

I think that makes your feedback even more important.  I had some status that was indicated _only_ by color in `owut` and felt pretty stupid when I realized what I had done (fixed it by adding labels in addition to the color).

systemcrash:
> @systemcrash could you point me to the other concept of labels you're referring to?

See #8579 

> lldpctl already has an option to output as json (`lldpctl -f json`) so adding the info should be fairly straighforward.

Not a bad idea... but not in this iteration or PR

> 
> I think that makes your feedback even more important. I had some status that was indicated _only_ by color in `owut` and felt pretty stupid when I realized what I had done (fixed it by adding labels in addition to the color).

There are a few different types (or blindness) that sooner or later, we're all guilty of it.

systemcrash:
Maybe (don't know whether P is the right letter, but) P(+) for PoE(+) enabled ports, or an option to enable PoE if there is budgeting. Just an idea for later since PoE stuff has not fully landed yet. @bevanweiss @carlosz1986

mirko:
PoE support is on my list, but it's undergoing changes internally right now (getting rid of userspace helpers, generic support should go via kernel and netifd - which got support merged in just today)

mirko:
Ok, here's version 2. I took your input - and mangled and blended it so the outcome is surely not anymore what you intended it to be.
Admittedly, the screenshot does not look more intuitive / well-arranged - but hear me out! :)

Changes:
 - embraced my partial colour-blindness and went for well-distinguishable color-schemes as well as differently shaped shades/patterns (@efahl, this is on you)
 - got rid of the sticky dock with the dropdowns at the bottom for several reasons: summed up: it used up vertical space for functionalitity affecting VLANs which are already listed above and don't need to be maintained in 2 further places (the dropdowns, which are questionable UI elements by itself)
 - hence, moved all this into the VLAN rows, incl. buttons working like checkboxes with an additional tri-state
 - hovering over VLAN rows highlights ports having this VLAN assigned (tagged / untagged)
 - buttons for selecting ports having this VLAN assigned (tagged / untagged)
 - buttons for assigning VLAN membership on selected ports (and indicating current membership)
 - bailing out and showing error(s) when attempting invalid configurations (choosing same VLAN as tagged and untagged); in a "all or nothing"-fashion (relevant if attempting such change on multiple selected ports)


But there's also a couple of technical questions:

(1) How to identify the (correct) switch / bridge? First: there might be more than one switch / vlan aware bridge.. I'm fine with this view - by definition - only selecting the first one (however the first one is determined), as this view already only allows common configurations (and throws an error if it sees e.g. PVID (primary VLAN) != egress VLAN (untagged) in /etc/config/network).
But still, how to actually filter for a vlan-aware bridge, especially those consisting of actual physical switchports? Currently there's multiple conditions of which either being met is already enough. E.g. is a bridge currently seen as vlan-aware by just having vlan-bridge sections.
Or shall we just very simply go for a bridge named "switch" and ignore / bail out on everything else?

(2) When and how to stash changes. Narrowed down, there're 2 obvious options:
 - uci.save() on every click and change - nicely logging all the changes for review when clicking on "Unsaved changes" in the top-right
 - stash changes inside luci only until being "Saved", which then saves them all it once. That works well for models, where the underlying view as the "Save & Apply"-, "Save"- & "Reset" button-triple, but not really for a full-page view.
Adding routes or firewall rules also immediately uci-stashes them.
So for now I went with the former (debounced uci-saving everything on every change). However there's no clean way for actually only adding/removing a tag to an interface.
This view as well as the original view first prune the whole vlan-section and then (re-)add the interfaces (minus removed ones). E.g. adding a tag on one interface which is present on X interfaces results in X+2 changes. Removing a tag doesn't show its removal in the unsaved changes, but its lack of being re-added.


https://github.com/user-attachments/assets/936253d2-2569-4e4f-b354-a0587d3062d1

owrt-luci-swcfg_5
(keep in mind, this screenshot is an extreme situation, trying to show all features - and is mainly like that (convoluted and unreadable), because I was hovering over a configuerd VLAN being assigned to many ports)

mirko:
I think the code is in a much more better shape now and streamlined where possible.
Also I tried to annotate the source with as much comments as possible to explain somehow cumbersome appearing mitigations.
It appears to be working well so far for me, but we can't overcome certain more underlying limitations, mostly involved by the UCI format/structure and how "unsaved changes" are being tracked in luci.

Now the big elephant in the room: it's partly LLM generated - and I'd feel shabby to actually ask people to review this. On the other side, I believe this view provides a much better user-experience and is totally self-contained (no changes to existing code, except ACLs for labels and top-level menu for easier access).

Suggestions? Worth converting to a PR?

systemcrash:
Squash your commits to 1, add your own SoB, and a co-authored SoB for the LLM. I'll need to take this for a spin on a VM...

mirko:
> Squash your commits to 1, add your own SoB, and a co-authored SoB for the LLM. I'll need to take this for a spin on a VM...

Of course, before making this an official PR that's what will be done. I'll take that as there's no strong objection going for a proper PR wrapping up the current state (with obvious commit history cleanups beforehand)

modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
```diff
@@ -0,0 +1,501 @@
...
+
+		const vid = +bvs.vlan;
+
+		if (!(vid >= MIN_VLAN_ID && vid <= MAX_VLAN_ID))
+			return;
+
+		out.push({
+			section_id: bvs['.name'],
+			vlan: vid,
+			local: bvs.local === '1',
```
openwrt-ai: The `local` flag is read as `bvs.local === '1'`, treating an absent option as not-local. netifd defaults the opposite way: in [config.c the bridge-vlan `local` attribute defaults to `true` when unset](https://github.com/openwrt/netifd/blob/e45b890fa532bb89380b78008c2cc70bfad2c4c7/config.c#L449). Two consequences:

- A `bridge-vlan` section without an explicit `local` option renders the "Local" checkbox unchecked, even though netifd treats it as local.
- `handleVlanLocalChange` unsets the option when the box is cleared (switch-vlan.js L735), which netifd interprets as local=true again — so unchecking "Local" and applying does not actually disable it.

To round-trip faithfully, treat absent as local (`bvs.local !== '0'`) and write `local '0'` explicitly when unchecked rather than unsetting.

---

openwrt-ai (commented):
## Commit checks

- The 43-commit history is a long iterate-in-place series with many introduce-then-fix and pure-restyle steps (e.g. `lighten selected port tiles` followed by `emphasise selected port tiles`; auto-stash added in `auto-stash Switch/VLAN UCI edits` then again in `auto-stash Switch/VLAN T/U toggles`; styling adjusted across ~12 commits). For 5 files / +2197 net, this should be squashed into a small set of logical commits before merge.

(The author identity / formality issues flagged by the failing "Test Formalities" check are left to the formality bot and not repeated here.)

---

openwrt-ai (commented):

The series was squashed from 43 commits down to two logical commits, which addresses the squash concern from the previous review. Both new commit messages match their changes. The new `parseLocal()` helper and the explicit `local='0'` writes correctly match netifd's default (`config_init_vlan_entry()` sets `vlan->local = true` when the option is absent), so the previous `local === '1'` reading was indeed wrong; the fix looks correct. The only failing check is Test Formalities (author identity), which is left to the formality bot.

---

mirko:
Kindly allowing myself to ping @systemcrash and @efahl again, as now converted into an actual PR and cleaned up "Points for discussion" (underneath "Pull request details")

systemcrash:
Just rename the second commit subject to something like:

luci-mod-network: experimental Switch/VLAN config notice

How do the fonts look? Are they more consistent now?

mirko:
> How do the fonts look? Are they more consistent now?

Now sure what you mean or rather what's missing from attached video / screenshot in order to answer that?

jow-:
Sorry for not following up earlier @mirko 

I like your latest draft, it's indeed different from what I suggested but I think it is fine - a lot more consistent and calmer look.

systemcrash:
@mirko get your second commit fixed, and LGTM

mirko:
@systemcrash @jow- @efahl thanks for the positive feedback, but I think before actually approving we should talk about the discussion points beforehand. Like the other label PR and when / how to stash/save changes via UCI.
I re-wrote the whole PR-details section when converting the draft to PR, so I'd at least like to know you're aware of the latest state and have scrolled through the points in question.

systemcrash:
There's another pull request regarding port labels:
    https://github.com/openwrt/luci/pull/8579 -- however we want
    port and VLAN labels. Not sure how to combine those efforts.

Handle them here, if it's not much change. 

    Shall we only allow this view for dedicated switch devices?
    We don't have a device class "switch" or alike yet, but could
    easily depend on the "realtek" target, as it's the only supported
    platform currently.

The dedicated switch devices are for the older switch platforms. There are still a few of those around. Best not to mix the types.

    Identifying the physical switch: apart from the possibility of
    there being more than one switch, how to identify an appropriate
    one:
        Go for the bridge named "switch" by sole naming convention.
        Filter for vlan-aware bridges (which itself is not well
        defined unfortunately), and filter for ony eth/sfp ports, so
        we don't include e.g. wifi ifaces), -- as does the "Bridge
        VLAN filtering" view -- and choose the first one from the list.
        Filter for physical ports as the "Port status" view does
        (Overview -> Status) and choose the bridge all of them are
        members of; would call luci.getBuiltinEthernetPorts() and
        hence need further ACL permissions.

Is it complex to enumerate through the available distributed switches?


    Stashing / saving changes: when and how to stash changes.
    Narrowed down, there are 2 obvious options:
        uci.save() on every click and change - logging all the
        changes for review when clicking on "Unsaved changes" in the
        top-right corner.
        Due to the uci structure fo configuring VLANs being a bit un-
        fortunate, back-and-forth changes can't be easily tracked but
        stack up.
        Stash changes inside luci only until being "Saved", which
        then saves them all at once. That works well for models
        where the underlying view has the "Save & Apply" / "Save" /
        "Reset" button-triple, but not really for a full-page view.


The current behaviour is all or nothing save, which is what roll-back is designed to protect. Most advanced users configuring VLANs are aware of the danger of sawing off the branch they're standing on. Roll-back is helpful here.

    Should the "local"-flag be en- or disabled by default?

Does it have a defined default today? Otherwise consider what fail open or fail closed mean in context. If the expectation is connectivity, fail to that, if the expectation is security, fail towards discontinued access.

    We currently do not support configurations of ingress != egress
    VLAN (or in other terms: PVID / primary VLAN != untagged VLAN)

If these situations can create loops (where STP is designed to help prevent), a best-effort should be to prevent such configs.

mirko:
> Handle them here, if it's not much change.

No change, as labels for both VLANs and ports are currently part of this PR

> The dedicated switch devices are for the older switch platforms. There are still a few of those around. Best not to mix the types.

I'm not sure what you're referring to. From what I know the "realtek"-target was introduced for dedicated switches and not used otherwise before. Which are the "older" ones?

> Is it complex to enumerate through the available distributed switches?

This is not how it works. We work on (vlan-aware-)bridges - in any way.

I proposed 3 ways, all with their pros and cons (copy&paste):

1) Go for the bridge named "switch" by sole naming convention.
2) Filter for vlan-aware bridges (which itself is not well
    defined unfortunately), and filter for ony eth/sfp ports, so
    we don't include e.g. wifi ifaces), -- as does the "Bridge
    VLAN filtering" view -- and choose the first one from the list.
3) Filter for physical ports as the "Port status" view does
    (Overview -> Status) and choose the bridge all of them are
    members of; would call luci.getBuiltinEthernetPorts() and
    hence need further ACL permissions.

> The current behaviour is all or nothing save, which is what roll-back is designed to protect. Most advanced users configuring VLANs are aware of the danger of sawing off the branch they're standing on. Roll-back is helpful here.

Hm, if I understand you correctly, that's how it's currently *not* implemented. Every change get's (uci-)saved currently, not just stashed. Whereas the "Reset"-button prunes those (and all other) changes.

> Does it have a defined default today? Otherwise consider what fail open or fail closed mean in context. If the expectation is connectivity, fail to that, if the expectation is security, fail towards discontinued access.

In the current view ("Network" -> "Interfaces" -> "Devices" -> "Configure..." -> "VLAN-aware-bridge" it's off by default. netifd itself, though, treats absence of the local-flag as "on", so the netifd-default is different than the old/current luci view.

openwrt-ai (commented):

The branch was rebased onto the current base and the two commits were reworded; the tree content is byte-identical to the previously reviewed tip (`git diff f7e0e2f..9bb5135` is empty), so the code already reviewed is unchanged. Both reworded commit messages match their changes.

---

systemcrash:
> I'm not sure what you're referring to. From what I know the "realtek"-target was introduced for dedicated switches and not used otherwise before. Which are the "older" ones?

There are newer DSA type platforms, and there are switch (swconfig) platforms (where a dedicated Switch menu appears in luci). A switch platform e.g. ath79, e.g. TP-Link Archer C7 is a switch platform not yet converted (large performance drop to convert to DSA). I'm not sure which category Realtek falls into, but I think it's DSA.

There might be more reliable litmus tests but perhaps `find /sys -path '*dsa*'` tells you of DSA presence. Absence of swconfig might be another.

So enumerating through all available hardware doesn't seem a stretch since they're all devices available to netifd.

We can start with 'switch' initially.

> Hm, if I understand you correctly, that's how it's currently not implemented. Every change get's (uci-)saved currently, not just stashed. Whereas the "Reset"-button prunes those (and all other) changes.

uci behaves like the UI - you stash individual changes which are committed when you apply. Same at the command line. See `uci commit`.

mirko:
> There are newer DSA type platforms, and there are switch (swconfig) platforms

I figured we got rid of swconfig by now, didn't know there's negative implications converting to DSA for some targets.

I think the current mechanism is conservative and good for testing the waters, checking for a configured vlan-aware-bridge with filtering enabled (the check is not straight forward, but that's how it is). That implies filtering will have to be enabled beforehand (by whatever means).

> uci behaves like the UI - you stash individual changes which are committed when you apply. Same at the command line. See uci commit.

It's not like I don't know how UCI works :) It's more a design decision, as it's also not consistent across other luci pages.
Where the current switch/VLAN-configuration view is a modal with just one "Save"-button and just stashes all modifications within luci before saving, editing routes e.g. uci.set()'s such changes immediately, which then can be Saved / Applied.

modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
```diff
@@ -0,0 +1,522 @@
...
+	case 'false':
+	case 'no':
+	case 'off':
+	case 'disabled':
+		return false;
+	default:
+		return true;
+	}
+}
+
```
systemcrash, line 41: @efahl added a nice uci.get_bool for exactly this case.

modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
```diff
@@ -0,0 +1,522 @@
...
+				break;
+			}
+
+			if (parsed.untagged) {
+				if (portUntagged[parsed.port] != null && portUntagged[parsed.port] !== vid) {
+					issues.push(_('Port "%s" is configured as untagged on more than one VLAN (%d and %d).').format(
+						parsed.port, portUntagged[parsed.port], vid));
+				}
+
+				portUntagged[parsed.port] = vid;
```
systemcrash: Does it not make sense to fold describePortRole into this switch-case?

modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
```diff
@@ -0,0 +1,522 @@
...
+ * ============================================================================= */
+
+function labelOptionKey(name) {
+	return String(name).replace(/[^A-Za-z0-9_]/g, '_');
+}
+
+function ensureLabelSections() {
+	let port_sid = null, vlan_sid = null;
+
+	uci.sections('luci', LUCI_LABEL_SECTION_TYPE, function(s) {
```
systemcrash, line 313: Why not store the label together with the vlan/port entries? Just invent a new descriptive property name. Unrecognised entries are ignored by netifd et al and don't do anything.

mirko: Because they're only cosmetic and don't have any functional influence. I already think /etc/config/network is too convoluted anyway, especially with PoE/PSE added to that recently. Also VLAN and port labels are "global" now, without any mapping to "which" switch, in case there's more. Taking this into account for future network config migrations will just add complexity to one of the most crucial config files.
Hence rather having own config-blocks within luci, which - if purged / getting lost - will only result in missing labels.

modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
mirko: The problem with the "local"-flag is: netifd's default is inverted. Absence means "true".
Also, we already have the string as it was fetched before, get_bool() would do another (unnecessary) lookup. not happy about it either, but also not having a better idea atm.

tell me if you'd prefer:
```
function parseLocal(section_id) {
	const v = uci.get('network', section_id, 'local');
	if (v == null || v === '')
		return true;                                  // netifd default
	return uci.get_bool('network', section_id, 'local');
}
```
with the unnecessary extra-`get` which `get_bool` would introduce here.

modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
mirko: Probably :) Flattens and reduces code quite a bit:
```
diff --git a/modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js b/modu
les/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
index 29c904a8e5..011e26f412 100644
--- a/modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
@@ -208,28 +208,6 @@ function collectBridgeVlans(deviceSection) {
  * VLAN configuration validation
  * ============================================================================= */
 
-function describePortRole(parsed) {
-       if (!parsed)
-               return null;
-
-       if (parsed.tagged && parsed.pvid)
-               return 'pvid-tagged';
-
-       if (parsed.tagged && parsed.untagged)
-               return 'tagged-and-untagged';
-
-       if (parsed.tagged)
-               return 'tagged';
-
-       if (parsed.untagged)
-               return 'untagged';
-
-       if (parsed.pvid)
-               return 'pvid-only';
-
-       return null;
-}
-
 function checkUnsupportedConfig(deviceSection) {
        const issues = [];
 
@@ -258,21 +236,15 @@ function checkUnsupportedConfig(deviceSection) {
                                return;
                        }
 
-                       const role = describePortRole(parsed);
-
-                       switch (role) {
-                       case 'tagged-and-untagged':
-                               issues.push(_('Port "%s" is both tagged and untagged on VLAN %d.').format(parsed.port, vid));
-                               break;
-
-                       case 'pvid-tagged':
+                       /* Order matters: a spec carrying both the tagged and PVID flags
+                        * (e.g. "lan1:t*") is reported as the PVID-with-tagged-egress
+                        * case, not the tagged-and-untagged one. */
+                       if (parsed.tagged && parsed.pvid)
                                issues.push(_('Port "%s" has the primary VLAN flag together with tagged egress on VLAN %d (PVID ≠ untagged VLAN).').format(parsed.port, vid));
-                               break;
-
-                       case 'pvid-only':
+                       else if (parsed.tagged && parsed.untagged)
+                               issues.push(_('Port "%s" is both tagged and untagged on VLAN %d.').format(parsed.port, vid));
+                       else if (parsed.pvid && !parsed.tagged && !parsed.untagged)
                                issues.push(_('Port "%s" is the primary VLAN on VLAN %d but has no untagged egress (PVID ≠ untagged VLAN).').format(parsed.port, vid));
-                               break;
-                       }
 
                        if (parsed.untagged) {
                                if (portUntagged[parsed.port] != null && portUntagged[parsed.port] !== vid) {
```

modules/luci-mod-network/htdocs/luci-static/resources/tools/bridgevlan.js
efahl: I've run into this exact issue before, https://github.com/jow-/ucode/pull/356#issuecomment-3676181271, where there's already a `get` value (in that case you can't use `get_bool` because of the way the values are enumerated).

How about we add `uci.parse_bool` like I'm proposing in that ucode.uci module PR?  And/or maybe add a `default_if_missing` parameter...

mirko: Sounds good to me, I'm happy to use whatever abstraction / convenience functions are being provided.
Question is: is that a blocker? IMHO we can easily adjust, once this (or even more) helpers approach. But if you feel like `parse_bool()` should come first, then I'll of course accept that, too :)

mirko:
@systemcrash @efahl @jow- sorry for (mass-)pinging you, but I'd appreciate some input on how to continue here.
I feel like I addressed all of your points to the degree where either I'm explaining my reasoning or asking for your input.
If something (else) is missing, please tell me (again) what to do. Thanks!

efahl:
Seems like it's in pretty good shape, so I'd say commit it and let people start playing with it...

mirko:
force-pushed this discussed change: https://github.com/openwrt/luci/pull/8639#discussion_r3355376189 - if you think the other points are addressed / explained sufficiently, I'm happy if it gets merged.

Just to avoid misunderstandings / deadlocks: while I technically have commit access, I would like someone else to actually trigger the merge, as I've never really been involved in luci development and my commit access just comes from "back then"-times.

openwrt-ai (commented):

The branch was rebased again. The only tree change since the previously reviewed tip (`9bb5135`) is in `tools/bridgevlan.js`: `describePortRole()` was removed and its logic inlined into `checkUnsupportedConfig()`. The inlined conditionals (`tagged && pvid`, then `tagged && untagged`, then `pvid && !tagged && !untagged`) are behaviorally equivalent to the previous helper-plus-switch, including the comment about ordering precedence. Both commit messages match their changes; the other four files are byte-identical to the last reviewed tip.

The failing `eslint` check is not caused by this PR — it aborts on a missing unrelated file (`luci-app-advanced-reboot/.../devices.json`) before evaluating any rules; the PR's own files are listed and parsed without complaint.

---

hnyman:
Seems like the comments are mainly positive, so let's merge this, and then wait for feedback...

efahl:
A little nit... Just popped it up on a VM using the OpenWrt2020 theme.  Got some weird CSS going on with the port label field:
image

mirko:
> A little nit... Just popped it up on a VM using the OpenWrt2020 theme. Got some weird CSS going on with the port label field: image

It looks entirely different to how I see it. No colour-coding at all, other dimensions of elements, etc. Margings wrong / Elements not fitting their boundaries.
Can you provide details about rendering engine, etc.?

EDIT: ah, OpenWrt2020-theme. Honestly, it didn't even cross my mind trying with a different than the bootstrap-theme. Is this usually a precondition for new features / GUI-elements, testing with all or a list of available themes?

efahl:
> Is this usually a precondition for new features / GUI-elements, testing with all or a list of available themes?

Yes, no, maybe?  I have no good feel for who uses what themes (other than that I get enough ASU error reports from people using argon that I know it has a following, possibly small?).  Maybe we should do a poll on the forums, to see if multi-theme testing would be a good thing to codify...


I tried it on a real device with more ports, to make sure colors worked and other stuff, looks good there but for the same field-too-wide.  I'm using whatever Firefox is on this Windows box, 152.0.1 now that I check.

image

efahl:
Oh, and since I have more 25.12 devices than snapshots, I've just been downloading the apk and installing it "off version".  Working fine, above image is from a 25.12.4 minipc with 4x NICs and no bridge hardware.

mirko:
I'm travelling atm and don't have (remote-)access to such a device right now.

The issue with the label fields are `min-width`'s set by these themes globally.

If you want, feel free to test the following patch, but be aware, it's **llm-assisted** and **untested**:
```
diff --git a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
index a6ea7ad4bb..676988eb33 100644
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
@@ -41,6 +41,42 @@
        --svc-vlan-tagged-bg: rgba(217, 119, 6, 0.14);
        --svc-vlan-untagged-stripe: rgba(13, 148, 136, 0.2);
        --svc-vlan-tagged-stripe: rgba(217, 119, 6, 0.2);
+       --svc-row-bg: var(--background-color-medium, #fafafa);
+       --svc-row-hover-bg: var(--background-color-high, #f3f3f3);
+       --svc-row-add-bg: #f7faff;
+}
+
+/* Theme overrides: several LuCI themes set large min-widths on text inputs
+ * (e.g. openwrt-2020: 20rem, material: 25rem in .cbi-value) and global
+ * button/input resets (openwrt-2020 zeros borders on *). Scope fixes here. */
+#switch-vlan-view input[type="text"] {
+       min-width: 0;
+       max-width: 100%;
+       width: 100%;
+       box-sizing: border-box;
+       min-height: unset;
+       line-height: normal;
+       font-size: inherit;
+       color: var(--text-color-high, inherit);
+       background-color: var(--background-color-high, #fff);
+       background-image: none;
+       border: 1px solid var(--border-color-low, #ddd);
+       border-radius: 2px;
+       box-shadow: none;
+}
+
+#switch-vlan-view input[type="text"]:focus {
+       border-color: var(--primary-color-high, #3c8dbc);
+       outline: none;
+}
+
+#switch-vlan-view .svc-vlan-btn-tagged,
+#switch-vlan-view .svc-vlan-btn-untagged,
+#switch-vlan-view .svc-vlan-btn-remove,
+#switch-vlan-view .svc-vlan-btn-add,
+#switch-vlan-view .svc-vlan-assign-btn {
+       box-shadow: none;
+       background-image: none;
 }
 
 :root[data-darkmode="true"] #switch-vlan-view {
@@ -53,6 +89,9 @@
        --svc-vlan-tagged-bg: rgba(251, 191, 36, 0.2);
        --svc-vlan-untagged-stripe: rgba(45, 212, 191, 0.32);
        --svc-vlan-tagged-stripe: rgba(251, 191, 36, 0.36);
+       --svc-row-bg: var(--background-color-medium, #2a2a2a);
+       --svc-row-hover-bg: var(--background-color-high, #333);
+       --svc-row-add-bg: var(--background-color-high, #2d333b);
 }
 
 #switch-vlan-view .svc-block {
@@ -86,7 +125,7 @@
 
 .svc-empty {
        margin: 0.5em 0;
-       color: #888;
+       color: var(--text-color-low, #888);
        font-style: italic;
 }
 
@@ -98,7 +137,7 @@
 }
 
 .svc-port-tile {
-       border: 1px solid #ccc;
+       border: 1px solid var(--border-color-medium, #ccc);
        border-radius: 4px;
        padding: 0.5em 0.6em;
        background: var(--background-color-medium, #fafafa);
@@ -110,10 +149,11 @@
        gap: 0.35em;
        transition: background 0.05s linear, border-color 0.05s linear, box-shadow 0.05s linear;
        min-width: 0;
+       overflow: hidden;
 }
 
 .svc-port-tile:hover {
-       border-color: #888;
+       border-color: var(--border-color-high, #888);
        background: var(--background-color-high, #f3f3f3);
 }
 
@@ -168,18 +208,12 @@
 .svc-port-link[data-state="down"] .svc-port-dot { background: #b0b0b0; }
 
 .svc-port-label {
-       width: 100%;
-       box-sizing: border-box;
        font-size: 0.92em;
-       color: inherit;
        padding: 2px 4px;
-       border: 1px solid #ddd;
-       border-radius: 2px;
-       background: var(--background-color-high, #fff);
 }
 
 .svc-port-label:focus {
-       border-color: #3c8dbc;
+       border-color: var(--primary-color-high, #3c8dbc);
        outline: none;
 }
 
@@ -218,29 +252,30 @@
 
 /* Highlight ports when hovering a VLAN row (teal = untagged, amber = tagged) */
 .svc-port-tile.highlight-untagged {
-        border-color: var(--svc-vlan-untagged);
-        background:
-                repeating-linear-gradient(
-                        90deg,
-                        var(--svc-vlan-untagged-stripe) 0 24px,
-                        var(--svc-vlan-untagged-bg) 24px 48px
-                );
+       border: 2px solid var(--svc-vlan-untagged);
+       background:
+               repeating-linear-gradient(
+                       90deg,
+                       var(--svc-vlan-untagged-stripe) 0 24px,
+                       var(--svc-vlan-untagged-bg) 24px 48px
+               );
 }
 
 .svc-port-tile.highlight-tagged {
-        border-color: var(--svc-vlan-tagged);
-        background:
-                repeating-linear-gradient(
-                        180deg,
-                        var(--svc-vlan-tagged-stripe) 0 24px,
-                        var(--svc-vlan-tagged-bg) 24px 48px
-                );
+       border: 2px solid var(--svc-vlan-tagged);
+       background:
+               repeating-linear-gradient(
+                       180deg,
+                       var(--svc-vlan-tagged-stripe) 0 24px,
+                       var(--svc-vlan-tagged-bg) 24px 48px
+               );
 }
 
 /* VLAN list — table layout keeps column widths and dividers aligned */
 .svc-vlans {
        display: table;
        width: 100%;
+       max-width: 100%;
        border-collapse: separate;
        border-spacing: 0 0.3em;
        table-layout: auto;
@@ -281,9 +316,10 @@
        text-align: right;
 }
 
-/* No width — extra space goes here (auto layout + narrow 1% siblings) */
+/* Label column absorbs remaining width; keep inputs inside the cell */
 .svc-vlan-cell-label {
-       min-width: 10em;
+       min-width: 0;
+       overflow: hidden;
 }
 
 .svc-vlan-cell-label,
@@ -312,35 +348,35 @@
 }
 
 .svc-vlan-row > .svc-vlan-cell-id {
-       border: 1px solid #ddd;
+       border: 1px solid var(--border-color-low, #ddd);
        border-right: none;
        border-radius: 4px 0 0 4px;
-       background: #fafafa;
+       background: var(--svc-row-bg);
 }
 
 .svc-vlan-row > .svc-vlan-cell-label,
 .svc-vlan-row > .svc-vlan-cell-local,
 .svc-vlan-row > .svc-vlan-cell-select,
 .svc-vlan-row > .svc-vlan-cell-assign {
-       border-top: 1px solid #ddd;
-       border-bottom: 1px solid #ddd;
-       background: #fafafa;
+       border-top: 1px solid var(--border-color-low, #ddd);
+       border-bottom: 1px solid var(--border-color-low, #ddd);
+       background: var(--svc-row-bg);
 }
 
 .svc-vlan-row > .svc-vlan-cell-action {
-       border: 1px solid #ddd;
+       border: 1px solid var(--border-color-low, #ddd);
        border-left: 1px solid var(--border-color-low, #ccc);
        border-radius: 0 4px 4px 0;
-       background: #fafafa;
+       background: var(--svc-row-bg);
 }
 
 .svc-vlan-row:hover > .svc-vlan-cell {
-       background: #f3f3f3;
-       border-color: #bbb;
+       background: var(--svc-row-hover-bg);
+       border-color: var(--border-color-medium, #bbb);
 }
 
 .svc-vlan-row.svc-vlan-add-row > .svc-vlan-cell {
-       background: #f7faff;
+       background: var(--svc-row-add-bg);
 }
 
 /* Dashed outline on the outside of the add row only — not between columns */
@@ -367,7 +403,7 @@
 
 @keyframes svc-flash {
        0%   { background: #fff5b1; }
-       100% { background: #fafafa; }
+       100% { background: var(--svc-row-bg); }
 }
 
 .svc-vlan-id {
@@ -376,26 +412,17 @@
 }
 
 .svc-vlan-id-input {
-       width: 100%;
-       max-width: 100%;
-       min-width: 0;
-       box-sizing: border-box;
        padding: 2px 4px;
        font-family: monospace;
 }
 
 .svc-vlan-label {
-       width: 100%;
-       box-sizing: border-box;
        padding: 2px 4px;
-       border: 1px solid #ddd;
-       border-radius: 2px;
-       background: var(--background-color-high, #fff);
 }
 
 .svc-vlan-label:focus,
 .svc-vlan-id-input:focus {
-       border-color: #3c8dbc;
+       border-color: var(--primary-color-high, #3c8dbc);
        outline: none;
 }
 
@@ -446,24 +473,26 @@
 /* T = tagged (amber), U = untagged (teal) — select + assign buttons */
 #switch-vlan-view .svc-vlan-btn-tagged:not(:disabled),
 #switch-vlan-view .svc-vlan-btn-assign-tagged:not(.svc-vlan-assign-disabled) {
-       border-color: var(--svc-vlan-tagged);
+       border: 1px solid var(--svc-vlan-tagged);
        background-color: var(--svc-vlan-tagged-bg);
+       color: var(--svc-vlan-tagged);
 }
 
 #switch-vlan-view .svc-vlan-btn-untagged:not(:disabled),
 #switch-vlan-view .svc-vlan-btn-assign-untagged:not(.svc-vlan-assign-disabled) {
-       border-color: var(--svc-vlan-untagged);
+       border: 1px solid var(--svc-vlan-untagged);
        background-color: var(--svc-vlan-untagged-bg);
+       color: var(--svc-vlan-untagged);
 }
 
 #switch-vlan-view .svc-vlan-btn-tagged .svc-vlan-btn-prefix,
 #switch-vlan-view .svc-vlan-btn-assign-tagged .svc-vlan-btn-prefix {
-       color: var(--svc-vlan-tagged);
+       color: inherit;
 }
 
 #switch-vlan-view .svc-vlan-btn-untagged .svc-vlan-btn-prefix,
 #switch-vlan-view .svc-vlan-btn-assign-untagged .svc-vlan-btn-prefix {
-       color: var(--svc-vlan-untagged);
+       color: inherit;
 }
 
 #switch-vlan-view .svc-vlan-btn-assign-tagged.svc-vlan-assign-checked,
@@ -478,12 +507,12 @@
 
 #switch-vlan-view .svc-vlan-btn-assign-tagged.svc-vlan-assign-checked {
        background: linear-gradient(#b45309, var(--svc-vlan-tagged));
-       border-color: #b45309;
+       border: 1px solid #b45309;
 }
 
 #switch-vlan-view .svc-vlan-btn-assign-untagged.svc-vlan-assign-checked {
        background: linear-gradient(#0f766e, var(--svc-vlan-untagged));
-       border-color: #0f766e;
+       border: 1px solid #0f766e;
 }
 
 #switch-vlan-view .svc-vlan-btn-assign-tagged.svc-vlan-assign-mixed {
@@ -494,7 +523,8 @@
                transparent 55%,
                var(--svc-vlan-tagged-bg) 55%
        );
-       border-color: var(--svc-vlan-tagged);
+       border: 1px solid var(--svc-vlan-tagged);
+       color: var(--svc-vlan-tagged);
 }
 
 #switch-vlan-view .svc-vlan-btn-assign-untagged.svc-vlan-assign-mixed {
@@ -505,7 +535,8 @@
                transparent 55%,
                var(--svc-vlan-untagged-bg) 55%
        );
-       border-color: var(--svc-vlan-untagged);
+       border: 1px solid var(--svc-vlan-untagged);
+       color: var(--svc-vlan-untagged);
 }
 
 :root[data-darkmode="true"] #switch-vlan-view .svc-vlan-btn-assign-tagged.svc-vlan-assign-checked,
@@ -520,12 +551,12 @@
 
 :root[data-darkmode="true"] #switch-vlan-view .svc-vlan-btn-assign-tagged.svc-vlan-assign-checked {
        background: linear-gradient(#f59e0b, var(--svc-vlan-tagged));
-       border-color: var(--svc-vlan-tagged);
+       border: 1px solid var(--svc-vlan-tagged);
 }
 
 :root[data-darkmode="true"] #switch-vlan-view .svc-vlan-btn-assign-untagged.svc-vlan-assign-checked {
        background: linear-gradient(#14b8a6, var(--svc-vlan-untagged));
-       border-color: var(--svc-vlan-untagged);
+       border: 1px solid var(--svc-vlan-untagged);
 }
 
 .svc-vlan-btn-prefix {
@@ -556,16 +587,16 @@
        }
 
        .svc-vlan-row {
-               border: 1px solid #ddd;
+               border: 1px solid var(--border-color-low, #ddd);
                border-radius: 4px;
-               background: #fafafa;
+               background: var(--svc-row-bg);
                padding: 0.4em 0.6em;
                margin-bottom: 0.3em;
        }
 
        .svc-vlan-row.svc-vlan-add-row {
                border-style: dashed;
-               background: #f7faff;
+               background: var(--svc-row-add-bg);
        }
 
        .svc-vlan-row > .svc-vlan-cell-label,
```

efahl:
> * Introduce labels for ports and VLANs; they're optional annotations
>       and don't serve any functional purpose except hinting the user,
>       hence stored in respective sections within /etc/config/luci.

This was discussed before and I sort of glossed over it, but now that I'm actually looking deeply at the configs generated, I too think the labels belong in `config/network`.

My rationale comes from being a pretty hard-core CLI user, so I edit the config files a _lot_ more than I use LuCI to change things.  So, when I label my port `ethX` as the device to which it's connected, and then edit the `network` config, I want to see that info locally without jumping over to the `luci` config (as a CLI user, how would I even know to look in that mostly-unrelated config?).

```ini
$ cat /etc/config/network
...
config device
        option name 'br-lan'
        option type 'bridge'
        list ports 'eth1'
        list ports 'eth2'
        list ports 'eth3'

config switchvlan 'port_labels'   <<< Moved to network from luci config
        option eth1 'swi03'
        option eth2 'swi04'
        option eth3 'swi05'
...
config bridge-vlan
        option device 'br-lan'
        option label 'home'  <<< Stick it right on the vlan
        option vlan '10'
        option local '1'
        list ports 'eth3:t'

config bridge-vlan
        option device 'br-lan'
        option name 'guest'  <<< Or should it be "name" instead of "label"
                             <<< for consistency with other sections?
        option vlan '20'
        option local '1'
        list ports 'eth3:t'
```


Having the `vlan_labels` values right on the `bridge-vlan` entry simply as `label` seems very much a natural home, in addition to adding some context when manually editing.

mirko:
I understand the label-arguments, but I still stick to what I wrote here: https://github.com/openwrt/luci/pull/8639#discussion_r3355293431

tl;dr: it's non-functional, just cosmetics, we will sooner or later be in need of another migration (version `n` to `n+1`) and I don't see how cosmetics/labels only used for luci should be a potential hazard for core network issues.

Also, /etc/config/network does not have a definite understanding which port belongs to which switch (as far as i get it at least)

efahl:
Regarding the non-functional `bridge-vlan` label, I was using the precedent of the `firewall` config, which does a similar thing with a completely non-functional `name`.  Leaving it out has zero impact on behavior, but boy is it nice to have when you're looking through the config file...
```ini
config rule
        option name 'Subnet-Access'
        option src 'wan'
        list src_ip 'fd0a:bad:dad::0/48'
        option dest 'lan'
        option target 'ACCEPT'
```

hnyman:
>  I have no good feel for who uses what themes (other than that I get enough ASU error reports from people using argon that I know it has a following, possibly small?). 

Official themes that exists in the LuCI repo: https://github.com/openwrt/luci/tree/master/themes

* bootstrap
* openwrt-2020
* material  (somewhat unmaintained for a while, almost dropped)
* openwrt (ancient, original 2010 look)

Bootstrap and openwrt-2020 are the two more widely used official themes.  Good practice is to ensure compatibility at least with them.

Argon etc. are third party themes supported by someone somewhere.  (There have earlier been some other themes accepted to LuCI repo, but they are usually abandoned by the author after a while, and themes start to rot, so currently the view about new themes is rather negative.)

efahl:
I tested the patch on both OpenWrt2020 and Material themes, both are fixed.  Here's material, just for the record.

image

mirko:
> Regarding the non-functional `bridge-vlan` label, I was using the precedent of the `firewall` config, which does a similar thing with a completely non-functional `name`. Leaving it out has zero impact on behavior, but boy is it nice to have when you're looking through the config file...
> 
> ```ini
> config rule
>         option name 'Subnet-Access'
>         option src 'wan'
>         list src_ip 'fd0a:bad:dad::0/48'
>         option dest 'lan'
>         option target 'ACCEPT'
> ```

I don't mind at all having this discussion. I see the benefits, and follow your arguments.
But the decision to move such labels into /etc/config/network shouldn't be made lightly and should have a proper thread beforehand.

mirko:
> I tested the patch on both OpenWrt2020 and Material themes, both are fixed. Here's material, just for the record.
> image

PR including these fixes: https://github.com/openwrt/luci/pull/8738

PR #8642, opened 2026-05-19
luci-base: fix uci write bug when input value equals default

AndyChiang888:
Avoid executing unnecessary write operations to flash/UCI when the user's form selection or input exactly matches the component's `default` configuration.

Previously, if a field had a valid value identical to its default, it would bypass the empty check and trigger a `write()` call, leading to configuration bloating.

This change ensures that such values are correctly intercepted and handled via `this.remove()` to keep the underlying configuration file clean and precise.

### Maintainer _(preferred)_

@jow- @systemcrash 

---

## Tested on

**OpenWrt version:** SNAPSHOT
**LuCI version:** master
**Web browser(s):** Chrome

---

## Checklist

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -879,6 +879,7 @@ return view.extend({
 					o.modalonly = true;
 					o.multiple = false;
 					o.default = 'wan';
+					o.rmempty = false;
 					o.depends("ip_source", "web");
 					o.depends("ip_source", "script");
 					o.depends("ip_source", "interface");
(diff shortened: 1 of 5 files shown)


AndyChiang888:
@jow- @systemcrash 
Fixing this issue is too time-consuming (not the issue itself, but a series of chain reactions), and I don't know if there are other places that need to be addressed.

I felt like I'd dug a bomb that was about to explode.

AndyChiang888:
I tested an older lua based app, and the same issue occurred with `o.default` set and no `o.rmempty = false`.
This issue has existed for many years, leading people to believe that `o.default` is designed this way.

AndyChiang888:
ping @systemcrash @jow- @hnyman

systemcrash:
Merged. Thanks @AndyChiang888

PR #8648, opened 2026-05-21
luci-app-https-dns-proxy: fix get_providers() invalid JSON in directory fallback

### LINKED ISSUE #8647: luci-app-https-dns-proxy: get_providers() emits invalid JSON (trailing comma) in directory fallback
## Bug

`get_providers()` in `root/usr/libexec/rpcd/luci.https-dns-proxy` produces invalid JSON when it falls back to reading provider files from the directory instead of the pre-built `providers.json`.

The loop unconditionally appends a comma after every file, including the last:

```sh
for f in "$providersDir"/*; do
    cat "$f"
    echo ','   # comma after EVERY file, including the last
done
```

This produces:

```json
{"https-dns-proxy":[
{...provider 1...},
{...provider 2...},
{...last provider...},
]}
```

Trailing commas are illegal JSON. Any strict parser (browsers, `jq`, `python3 -m json.tool`) rejects the output, causing the UI to silently receive no providers.

A second defect in the same function: if `$providersDir` is empty or absent, the glob `"$providersDir"/*` expands to a literal path that does not exist. `cat` fails silently on stderr while the loop still emits `,` on stdout, producing `[,]` — also invalid JSON.

## When does this trigger?

The fallback path is taken when `providers.json` has not been pre-built (e.g. the package was installed but the post-install script did not run, or the file was manually removed). Normal installs that have the pre-built file are unaffected.

## Fix

A pull request is attached. Key changes:

- Emit the separator *before* each item after the first (no trailing comma)
- Add `[ -f "$f" ] || continue` and a `*.json` glob so an empty/absent directory yields a valid empty array
- Make `providersDir`/`providersJson` overridable via `_LUCI_PROVIDERS_DIR`/`_LUCI_PROVIDERS_JSON` environment variables (production default unchanged) so the fallback path is testable without touching live system paths
- Add three new tests in `tests/run_tests.sh` section 04 covering: valid JSON output across all 45 providers, correct provider count, and empty-directory edge case

unglazed7010:
Fixes #8647

## Summary

- `get_providers()` emitted a trailing comma after every provider file in the directory-scan fallback, producing JSON rejected by strict parsers and silently breaking the provider list in the UI
- An empty `$providersDir` caused `cat` to fail on a literal glob string while still emitting `,`, producing `[,]`

## Patch

**`root/usr/libexec/rpcd/luci.https-dns-proxy`**

- Flip comma placement: emit separator *before* each item after the first (not after every item)
- Add `[ -f "$f" ] || continue` guard and narrow glob to `*.json` so an empty/absent directory yields a valid empty array instead of broken output
- Allow `_LUCI_PROVIDERS_DIR` / `_LUCI_PROVIDERS_JSON` env-var overrides (default unchanged) to make the fallback path exercisable in the test harness without touching live system paths

**`tests/run_tests.sh`**

New section 04 — three tests:
1. Directory fallback output is valid JSON (python3 json parser accepts it)
2. All 45 providers are present in the output (count check)
3. Empty providers directory yields a valid empty array

## Test run results

```
## 04: getProviders JSON output

  PASS: getProviders dir-fallback returns valid JSON
  PASS: getProviders dir-fallback: all 45 providers present
  PASS: getProviders empty dir-fallback: valid empty array

Ran 335 tests, 335 passed, 0 failed
```

(was 332 tests before this patch; all pre-existing tests still pass)

Signed-off-by: unglazed7010 <thatblueghoul@gmail.com>

### DIFF
--- a/applications/luci-app-https-dns-proxy/root/usr/libexec/rpcd/luci.https-dns-proxy
+++ b/applications/luci-app-https-dns-proxy/root/usr/libexec/rpcd/luci.https-dns-proxy
@@ -11,8 +11,8 @@
 # ubus -S call luci.https-dns-proxy getProviders '{"name": "https-dns-proxy" }'
 
 readonly packageName="https-dns-proxy"
-readonly providersDir="/usr/share/${packageName}/providers"
-readonly providersJson="/usr/share/${packageName}/providers.json"
+readonly providersDir="${_LUCI_PROVIDERS_DIR:-/usr/share/${packageName}/providers}"
+readonly providersJson="${_LUCI_PROVIDERS_JSON:-/usr/share/${packageName}/providers.json}"
 
 . "${IPKG_INSTROOT}/lib/functions.sh"
 . "${IPKG_INSTROOT}/usr/share/libubox/jshn.sh"
@@ -103,15 +103,17 @@ get_platform_support() {
 }
 
 get_providers() {
-	local f
+	local f first=1
 	echo '{"https-dns-proxy":'
 	if [ -s "$providersJson" ]; then
 		cat "$providersJson"
 	else
 		echo '['
-		for f in "$providersDir"/*; do
+		for f in "$providersDir"/*.json; do
+			[ -f "$f" ] || continue
+			[ "$first" = "1" ] || printf ','
+			first=0
 			cat "$f"
-			echo ','
 		done
 		echo ']'
 	fi
--- a/applications/luci-app-https-dns-proxy/tests/run_tests.sh
+++ b/applications/luci-app-https-dns-proxy/tests/run_tests.sh
@@ -459,6 +459,88 @@ else
 fi
 
 ###############################################################################
+#                     04: getProviders JSON output                            #
+###############################################################################
+
+printf "\n##\n## 04: getProviders JSON output\n##\n\n"
+
+if [ -f "$RPC_SCRIPT" ] && [ -d "$PROVIDERS_DIR" ]; then
+	_tmp_stubs="$(mktemp -d)"
+	_tmp_empty="$(mktemp -d)"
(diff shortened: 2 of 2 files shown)


egc112:
Please direct your PR upstream to:
https://github.com/mossdef-org/luci-app-https-dns-proxy

Thanks for your contribution

applications/luci-app-https-dns-proxy/root/usr/libexec/rpcd/luci.https-dns-proxy
```diff
@@ -103,15 +103,17 @@ get_platform_support() {
...
 	if [ -s "$providersJson" ]; then
 		cat "$providersJson"
 	else
 		echo '['
-		for f in "$providersDir"/*; do
+		for f in "$providersDir"/*.json; do
+			[ -f "$f" ] || continue
+			[ "$first" = "1" ] || printf ','
+			first=0
 			cat "$f"
```
openwrt-ai, line 116: `luci-app-https-dns-proxy` is maintained out-of-tree, not in `openwrt/luci`. Changes to this RPC backend (and the accompanying tests) should be submitted to the upstream repository so they don't get overwritten on the next sync: https://github.com/mossdef-org/luci-app-https-dns-proxy

The fix itself looks reasonable (the trailing-comma / `[,]` bug is real), but it needs to land upstream rather than here. Please redirect this PR there.

---

openwrt-ai (commented):
Reviewed 1 new commit. This PR modifies `luci-app-https-dns-proxy`, which is maintained out-of-tree at https://github.com/mossdef-org/luci-app-https-dns-proxy and should be submitted there rather than reviewed/merged in `openwrt/luci`. See the inline comment for details; the underlying fix appears sound but belongs upstream.

---

Issue #8613, opened 2026-05-08
luci-app-firewall: changing cloneable rule option 'enabled' on cli and cloning appends to end of table


ckorber:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### screenshots or captures

Before:
<img width="1601" height="145" alt="Image" src="https://github.com/user-attachments/assets/725558f2-f22c-4ef1-9a39-19008e7e4d8b" />

After:

<img width="1598" height="223" alt="Image" src="https://github.com/user-attachments/assets/7306f097-0953-4e2a-8591-845fdc1e88cc" />

### Actual behaviour

Feature `cloneable` works on the web interface, but if I edit a rule via cli

and clone it afterwards it gets appended to the end and not beneath the original.

I looked into the algorithm in luci, but somehow I think it is a bug because it doesn't work well with package uci. But on the otherhand it felt weird to open an issue in uci for this case.

### Expected behaviour

It should get appended beneath the original.

### Steps to reproduce

1. Change rule on cli
```
uci set firewall.@rule[-2].enabled='1' <-- was '0' before.
uci commit
reload_config
```
2. Clone the rule

### Additional Information

```shell
luci master but reproducable on all branches
```

### What browsers do you see the problem on?

Firefox

### Relevant log output

```shell

```

systemcrash:
Well, there's a flag passed to the clone operation which determines whether its move position immediately follows the original or goes to the end. The general order of rules reflects how it is laid out in the config file, and it gets put at the end anyway if it's 'new'. It might help you to observe its actual position in the file pre- and post- edit (and clone).

ckorber:
You are right, I will append that information:

**Before manipulation**:
```
...
firewall.cfg1692bd=rule
firewall.cfg1692bd.name='Allow-IPsec-NAT-T-Input' <-- This rule will be changed and cloned
firewall.cfg1692bd.src='wan'
firewall.cfg1692bd.dest_port='4500'
firewall.cfg1692bd.proto='udp'
firewall.cfg1692bd.target='ACCEPT'
firewall.cfg1692bd.enabled='0'
firewall.cfg1792bd=rule
firewall.cfg1792bd.name='Allow-OpenVPN-Input'
firewall.cfg1792bd.src='wan'
firewall.cfg1792bd.dest_port='1194'
firewall.cfg1792bd.proto='tcp udp'
firewall.cfg1792bd.target='ACCEPT'
[...]

PR #8649, opened 2026-05-21
luci-base: fix order of clone bug

ckorber:
### Description

If a rule was manipulated via cli, the rule was appended to the end if cloned.
Somehow the reason appears to be that `name` was incorrectly assigned, because name was the event Object this way.

### Maintainer _(preferred)_

@systemcrash 

---

## Tested on

**Web browser(s):** Firefox

---

## Checklist

- [x] _(Optional)_ Includes what Issue it closes (e.g. openwrt/luci#8613).


ckorber:
There was also this hidden bug in `form.js` that removed `enabled` on cloning. This prevented `reorderSections` to fire, because deletes got skipped.

systemcrash:
Thanks for your persistent effort to figure this out. The `ev` omission seems obvious now...

PR #8657, opened 2026-05-25
luci-mod-network: add interface/device description

YusDyr:
## Summary

The UCI `option description` field on both `config interface` and `config device` sections in `/etc/config/network` is already a valid, storable option, but had no representation in the LuCI UI.

This change adds:

**Interfaces (Network → Interfaces):**
- Description shown as a subtitle inside the interface badge when set via `option description`
- For WireGuard/AmneziaWG interfaces: automatically falls back to peer `description` fields — no manual setup needed
- Editable Description field at the top of the interface configure modal (General Settings tab)

**Devices (Network → Interfaces → Devices tab):**
- Read-only Description column in the Devices grid
- For `dev:xxx` entries without a `config device` section (e.g. VPN tunnel devices): falls back to the matching interface description or WireGuard/AmneziaWG peer descriptions
- Editable Description field in the device configure modal (General device options tab)

## Use case

- Label physical LAN ports on DSA-based routers (e.g. `lan2` → `NVR/Cameras`, `lan3` → `TV`)
- VPN tunnel interfaces auto-label from peer config (e.g. `awg0` shows `RedShield Georgia` from its peer's `description`)
- Any interface/device where the UCI name alone is not self-explanatory

No new UCI options or schema changes — purely exposes the existing `description` field in the UI.

## Screenshots

**Interfaces tab** — description subtitle in interface badge (awg tunnels auto-labeled from peers):

Interfaces

**Devices tab** — Description column (LAN ports labeled manually, AWG devices auto-labeled from peers):

Devices

**Device configure modal** — editable Description field in General device options:

DeviceEdit

See also: #8659 (adds description to the Port Status widget on the Overview page)

## Testing

- Interface badge shows description subtitle when `option description` is set or peer descriptions exist; hidden when unset
- WireGuard/AmneziaWG interface badges auto-show peer descriptions without any manual config
- Devices tab shows Description column with correct fallback logic: device config → interface config → peer descriptions → `-`
- Configure modals include editable Description field that correctly writes/removes `option description`
- No regressions on existing interface/device options

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -577,6 +577,12 @@ return view.extend({
 					return L.naturalCompare(a.getProtocol(), b.getProtocol());
 				});
 
+				o = s.taboption('general', form.Value, 'description', _('Description'));
+				o.modalonly = true;
+				o.optional = true;
+				o.rmempty = true;
+				o.placeholder = _('optional interface label');
+
 				o = s.taboption('general', form.DummyValue, '_ifacestat_modal', _('Status'));
 				o.modalonly = true;
 				o.cfgvalue = L.bind(function(section_id) {
@@ -1464,12 +1470,24 @@ return view.extend({
 			if (!net)
 				return;
 
+			const desc = uci.get('network', section_id, 'description') ||
+				['amneziawg', 'wireguard'].reduce(function(found, proto) {
+					if (found) return found;
+					const peers = uci.sections('network', proto + '_' + section_id);
+					const labels = peers.map(function(p) { return p.description }).filter(Boolean);
+					return labels.length ? labels.join(', ') : null;
+				}, null);
 			const node = E('div', { 'class': 'ifacebox' }, [
 				E('div', {
 					'class': 'ifacebox-head',
 					'style': firewall.getZoneColorStyle(zone),
 					'title': zone ? _('Part of zone %q').format(zone.getName()) : _('No zone assigned')
-				}, E('strong', net.getName())),
+				}, [
+					E('strong', net.getName()),
+					desc ? E('span', {
+						'style': 'display:block; font-size:0.75em; font-weight:normal; opacity:0.85; white-space:nowrap; overflow:hidden; text-overflow:ellipsis'
+					}, [ desc ]) : null
+				].filter(Boolean)),
 				E('div', {
 					'class': 'ifacebox-body',
 					'id': '%s-ifc-devices'.format(section_id),
@@ -1587,14 +1605,22 @@ return view.extend({
 			const dev = getDevice(s.section);
 			const devName = dev ? dev.getName() : null;
 
+			function addDescription() {
+				const o = s.taboption('devgeneral', form.Value, 'description', _('Description'));
+				o.optional = true;
+				o.rmempty = true;
+			}
+
 			/* Query PSE status from netifd to determine if device has PSE capability */
 			if (devName) {
-				return L.resolveDefault(callNetworkDeviceStatus(devName), {}).then((status) => {
+				return L.resolveDefault(callNetworkDeviceStatus(devName), {}).then(function(status) {
 					const hasPSE = (status.pse != null);
 					nettools.addDeviceOptions(s, dev, isNew, rtTables, hasPSE);
+					addDescription();
 				});
 			} else {
 				nettools.addDeviceOptions(s, dev, isNew, rtTables, false);
+				addDescription();
 				return Promise.resolve();
 			}
 		};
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1464,12 +1470,24 @@ return view.extend({
 			if (!net)
 				return;
 
+			const desc = uci.get('network', section_id, 'description') ||
+				['amneziawg', 'wireguard'].reduce(function(found, proto) {
```
openwrt-ai, line 1474: This WireGuard/AmneziaWG peer-description fallback is duplicated verbatim here and in the device column textvalue (line 1806 in this file: `https://github.com/openwrt/luci/blob/b42c213e3aa7e3a818e0f51890d89b63ded8744c/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L1806).` Consider extracting a small helper (e.g. `peerDescriptions(ifname)`) and calling it from both sites so the peer-section naming logic lives in one place.

---

openwrt-ai (commented):
Reviewed the single new commit; the change exposes the existing `description` UCI field in the interfaces/devices UI as described, and the commit message accurately matches the diff. WireGuard peer-section naming (`wireguard_<iface>`) and the peer `description` option were verified against luci-proto-wireguard. CI on the head commit passes (Test Formalities success). One minor maintainability note inline.

---

YusDyr:
Rebased onto current master — the branch had gone stale and was conflicting, which is presumably why it sat.

The conflict was in `addModalOptions`: master has since reworked it to destructure the netifd status and pass `parseLinks(...)` into `nettools.addDeviceOptions()` (f4c5c76, 93d2f79). I kept master's version verbatim and only re-applied the `addDescription()` call in both the PSE and non-PSE branches.

Also dropped the inline style on the badge subtitle in favour of a shared `.ifacebox-descr` class, added to every theme that styles `.ifacebox-head` in a separate `themes:` commit. That came out of @efahl's review on #8659, and the same objection applied here. Rendered in all three themes there: https://github.com/openwrt/luci/pull/8659#discussion_r4019197272

The `themes:` commit is on both branches since either may merge first; the second one will drop it as already applied.

No functional change beyond the rebase.

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -986,6 +986,16 @@ body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) .iface
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:nth-child(3){order:1;flex:0 0 100%;width:100%;margin:var(--fs-space-0-5) 0 var(--fs-space-1)}
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:first-child{flex:0 0 auto;width:100%;min-width:0;overflow:hidden;text-overflow:ellipsis;line-height:var(--fs-leading);font-size:var(--fs-type);font-weight:var(--fs-weight);color:var(--fs-text)}
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]):has(img[src*="_down.svg"]) > .ifacebox-head:first-child{color:var(--fs-dim)}
+
+.ifacebox-descr {
```
openwrt-ai: nit: this file writes every rule on a single line with no indentation, so a tab-indented multi-line block does not match it. It also lands mid-way through the `admin-status-overview` run in `@layer page` instead of beside the other component rules in [`@layer base` (cascade.css:216-220)](``https://github.com/openwrt/luci/blob/84f850af80d0b7b18a6d82f2afe6d4fca043880a/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L216-L220``) — move it there and collapse it to the file's one-line form.

---

themes/luci-theme-material/htdocs/luci-static/material/cascade.css
```diff
@@ -3279,4 +3279,14 @@ body[data-page='admin-status-overview'] .ifacebox-head {
   color: var(--header-color);
 }
 
+.ifacebox-descr {
```
openwrt-ai: nit: this lands inside the `/* === STATUS OVERVIEW: MATCH IFACEBOX COLOUR WITH HEADER === */` block, just before its END marker, filing a generic class under a status-overview-specific section. Move it beside [`.ifacebox-head` at cascade.css:2078](``https://github.com/openwrt/luci/blob/84f850af80d0b7b18a6d82f2afe6d4fca043880a/themes/luci-theme-material/htdocs/luci-static/material/cascade.css#L2078-L2081);`` same applies to `themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css` line 2064, which sits inside the identical marker block ([`.ifacebox-head` is at line 695](``https://github.com/openwrt/luci/blob/84f850af80d0b7b18a6d82f2afe6d4fca043880a/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css#L695-L698)).``

---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
YusDyr: Fixed — the rule now sits next to the base `.ifacebox-head` rule in every theme, and in footstrap it is collapsed to the file's one-line, unindented form.

I had inserted after the *last* `.ifacebox-head` match in each file, which is what dropped it into the `admin-status-overview` runs. New positions:

| theme | before | now |
|---|---|---|
| bootstrap | 2194, after the darkmode `[style]` override | 2180, after `.ifacebox .ifacebox-head` |
| footstrap | 990, inside the port-status run in `@layer page` | 219, in `@layer base`, one line |
| material | 3282, inside the STATUS OVERVIEW marker block | 2083, after `.ifacebox-head` |
| openwrt-2020 | 2064, same marker block | 700, after `.ifacebox-head` |
| openwrt | 1544, after `.ifacebox-head.active` | 1540, after `.ifacebox-head` |

Rendering is unchanged — the rule was never nested inside those blocks, only filed under them, so this is purely about where it lives.

themes/luci-theme-material/htdocs/luci-static/material/cascade.css
YusDyr: Moved — same fix as the footstrap one above: material now has it at 2083 right after `.ifacebox-head`, and openwrt-2020 at 700 after its own `.ifacebox-head`, both out of the STATUS OVERVIEW marker blocks.

Root cause was the same for all of them: I inserted after the last `.ifacebox-head` match in each file rather than the base rule.

openwrt-ai (commented):
## Commit checks

- `7ce030b0` "themes: add .ifacebox-descr class for interface descriptions" — the body's premise doesn't hold against the tree: no view hard-codes a description inline style and this series drops none, and "only three inline font-size styles in all of the view code" is wrong (37 `'style': …font-size` occurrences under `modules/`, `applications/`, `themes/`, `protocols/`). State it as what it is: a new shared class so the description label is theme-tunable.

---

PR #8659, opened 2026-05-25
luci-mod-status: add description to port status

YusDyr:
## Summary

Companion change to #8657 (which adds `option description` support to
the Interfaces and Devices views in `luci-mod-network`).

If a `config device` section in `/etc/config/network` has an
`option description` set for a physical port (e.g. `lan2`, `lan3`),
the description is shown as a subtitle below the port name in the
**Port Status** widget on the Overview page.

- Ports without a description are unaffected
- Long descriptions are truncated with `…`; the full text is shown as a tooltip on hover

## Use case

Label physical LAN ports on DSA-based routers so the port function is
visible at a glance on the Overview page (e.g. `lan2` → `NVR/Cameras`,
`lan3` → `TV`).

## Screenshots

**Port Status widget** — description subtitle under port name:

Port Status

## Testing

- Port with `option description` set: subtitle appears below port name
- Port without `option description`: no subtitle, layout unchanged
- Long description: truncated with `…`, full text on hover

## Tested on

**OpenWrt version:** OpenWrt 24.10.3
**LuCI version:** openwrt-24.10
**Web browser:** Safari

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
@@ -497,8 +497,17 @@ return baseclass.extend({
 
 			statsContent.push(E('span', { 'class': 'cbi-tooltip' }, formatStats(port.netdev, pse)));
 
-			return E('div', { 'class': 'ifacebox', 'style': 'margin:.25em;min-width:70px;max-width:100px' }, [
-				E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:bold' }, [ port.netdev.getName() ]),
+			const devSect = uci.sections('network', 'device').find(s => s.name === port.device);
+			const portDesc = devSect ? devSect.description : null;
+
+			const ifaceboxChildren = [
+				E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:bold' }, [ port.netdev.getName() ])
+			];
+
+			if (portDesc)
+				ifaceboxChildren.push(E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:normal;font-size:80%;overflow:hidden;white-space:nowrap;text-overflow:ellipsis', 'title': portDesc }, portDesc));
+
+			ifaceboxChildren.push(
 				E('div', { 'class': 'ifacebox-body' }, [
 					E('img', { 'src': L.resource('icons/port_%s.svg').format(portIcon) }),
 					E('br'),
@@ -516,7 +525,9 @@ return baseclass.extend({
 				E('div', { 'class': 'ifacebox-body' }, [
 					E('div', { 'class': 'cbi-tooltip-container', 'style': 'text-align:left;font-size:80%' }, statsContent)
 				])
-			]);
+			);
+
+			return E('div', { 'class': 'ifacebox', 'style': 'margin:.25em;min-width:70px;max-width:100px' }, ifaceboxChildren);
 		}));
 	}
 });


modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -497,8 +497,17 @@ return baseclass.extend({
...
-				E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:bold' }, [ port.netdev.getName() ]),
+			const devSect = uci.sections('network', 'device').find(s => s.name === port.device);
+			const portDesc = devSect ? devSect.description : null;
+
+			const ifaceboxChildren = [
+				E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:bold' }, [ port.netdev.getName() ])
+			];
+
+			if (portDesc)
+				ifaceboxChildren.push(E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:normal;font-size:80%;overflow:hidden;white-space:nowrap;text-overflow:ellipsis', 'title': portDesc }, portDesc));
```
efahl, line 508: How do these hard-coded style elements look in the other supported themes?  Seems like maybe a new class is needed, or pick from an existing one that's "close enough"?

(You only show one graphic in the PR, so it's hard to know  how it will look with all of default, 2020 and material.)

openwrt-ai (commented):

---

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
YusDyr: Good catch — the inline styles are gone. Both this PR and #8657 now use a shared `.ifacebox-descr` class, added to every theme that styles `.ifacebox-head` (bootstrap, footstrap, material, openwrt, openwrt-2020) in a separate `themes:` commit.

I looked for an existing class first. `.cbi-value-description` is the closest by name, but it is form-field help text: bootstrap renders a help icon through `::before` and indents by `1.25em`, and material sets `display: table-cell`. Neither works as a badge subtitle, so a new class seemed right.

Here is the label in the three themes you asked about. Third badge in each shot has no description set, to show the empty case:

**default (bootstrap)**
default

**openwrt-2020**
openwrt-2020

**material**
material

Worth noting these are rendered from each theme's `cascade.css` against the markup the view emits, not from a live router — I only have bootstrap installed on my device. Happy to redo them on a real install if you would rather see that.

The class deliberately sets no `text-align`, so each theme keeps its own: bootstrap and 2020 centre the badge text, material left-aligns it, and the description follows suit. That is the part an inline style could not do.

The `themes:` commit is on both this branch and #8657, since either can merge first; whichever goes second will drop it as already applied.

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
```diff
@@ -497,8 +497,17 @@ return baseclass.extend({
...
-				E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:bold' }, [ port.netdev.getName() ]),
+			const devSect = uci.sections('network', 'device').find(s => s.name === port.device);
+			const portDesc = devSect ? devSect.description : null;
+
+			const ifaceboxChildren = [
+				E('div', { 'class': 'ifacebox-head', 'style': 'font-weight:bold' }, [ port.netdev.getName() ])
+			];
+
+			if (portDesc)
+				ifaceboxChildren.push(E('div', { 'class': 'ifacebox-head ifacebox-descr', 'title': portDesc }, portDesc));
```
openwrt-ai, line 508: footstrap lays out the port card by child index — [`.ifacebox-head:nth-child(3)`, `.ifacebox-body:nth-child(2)` and `.ifacebox-body:nth-child(4)` at footstrap/cascade.css:986-1003](``https://github.com/openwrt/luci/blob/f843d4673d9467c1b0897fa7a945d1b3039b44dc/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L986-L1003).`` Inserting this div as child 2 shifts the zone row, speed row and stats row by one, so every port that has a description loses footstrap's flex ordering. Those selectors need to follow the new DOM (or be made position-independent) in the same commit that adds the class.

---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -986,6 +986,16 @@ body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) .iface
...
+
+.ifacebox-descr {
+	display: block;
+	font-weight: normal;
+	font-size: .75em;
+	opacity: .75;
+	overflow: hidden;
+	white-space: nowrap;
+	text-overflow: ellipsis;
+}
```
openwrt-ai: nit: this file is entirely minified one-rule-per-line; this is the only expanded, tab-indented rule in it. It also splits the contiguous `admin-status-overview` port block — better placed after the `.ifacebox-body:nth-child(4)` rules at line 1003.

```suggestion
.ifacebox-descr{display:block;font-weight:normal;font-size:.75em;opacity:.75;overflow:hidden;white-space:nowrap;text-overflow:ellipsis}
```

---

themes/luci-theme-material/htdocs/luci-static/material/cascade.css
```diff
@@ -3279,4 +3279,14 @@ body[data-page='admin-status-overview'] .ifacebox-head {
   color: var(--header-color);
 }
 
+.ifacebox-descr {
```
openwrt-ai: nit: this lands inside the `STATUS OVERVIEW: MATCH IFACEBOX COLOUR WITH HEADER` marker pair, which scopes one unrelated rule; put it after the `END` comment instead. Same in `themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css:2064`.

---

openwrt-ai (commented):
Reviewed 2 new commits.

---

openwrt-ai (commented):

---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
openwrt-ai: fixed, thanks — solved by moving it up next to the other `.ifacebox` rules instead of after line 1003, which works just as well.

---

themes/luci-theme-material/htdocs/luci-static/material/cascade.css
openwrt-ai: fixed, thanks — both themes now place it beside their `.ifacebox-head` rule rather than after the `END` marker.

---

modules/luci-mod-status/htdocs/luci-static/resources/view/status/include/29_ports.js
YusDyr: Redone on a live install, as offered — these are the real Interfaces page, not markup rendered against each theme's `cascade.css`. I installed `luci-theme-material` and `luci-theme-openwrt-2020` alongside bootstrap and switched between them.

The descriptions here come from actual UCI config: four AmneziaWG interfaces whose labels are pulled from their peer sections, and `wan` with `option description` set directly. `lan`, `wan6` and `wan6_alias0` have none, so they show the empty case in the same shot.

**default (bootstrap)**
default

**openwrt-2020**
openwrt-2020

**material**
material

IP addresses are masked in the DOM before capture; MACs are left alone. Nothing else is touched.

Each theme keeps its own typography and alignment, which is what the class buys over the inline style it replaced — bootstrap and 2020 centre the badge text, material left-aligns it, and the description follows in each case. The earlier `cascade.css` renders are still in the branch as `theme-*.png` if the isolated view is useful for comparison.

Issue #8603, opened 2026-05-02
luci-base: i18n does not add strings containing "${}"


AndyChiang888:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### Actual behaviour

i18n does not add strings containing "${}"
https://github.com/openwrt/luci/blob/e033fafcc94f61da7bdd9c20bb715d0fda3136b4/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dhcp.js#L650-L656

### Expected behaviour

base.pot should contain these strings.

### Steps to reproduce

None

### Additional Information

```shell
None
```

### Relevant log output

```shell

```

AndyChiang888:
@hnyman PTAL

hnyman:
I would say that we are using too complex notations in the help texts. This part about DNS tags is originating from @systemcrash. Maybe the text could be simplified?

I see that section in gui like
```
A tag is an alphanumeric label. They are attached to a DHCP client or transaction.
dnsmasq conditionally applies chosen DHCP options when a specific tag is encountered.
In other words: "This tag gets these option(6):<opt-name>,[<value>[,<value>]]".
tags do not do anything by themselves. They are labels that other directives test against.
Note: invalid option(6):<opt-name>,[<value>[,<value>]] combinations may cause dnsmasq to crash silently.

Prepend a tag with ! to invert their domain of application, e.g. to send options to a host lacking a tag.

Use the Add tag button to add a new tag.
```
At least all the 'tagcodestring' instances are shown just as highlighted 'tag' in the text. Could they be just text?

What has been the design principle here, @systemcrash ? Using complex syntax to prevent the translation of actual codewords?

AndyChiang888:
Are there any other places where "${}" is used?
If so, need to fix i18n*.pl?

systemcrash:
I opted to use this format which seemed easier for i18n, especially since the block uses the code tag to format various words. (dnsmasq) Tags are not well understood and require a deal of explanation or experimentation to understand.

AndyChiang888:
How to deal with this issue?
Fix .pl or modify those "${}?"

AndyChiang888:
ping @hnyman @systemcrash

jow-:
Using JavaScript `${placeholder}` expressions in translation strings is a bug. Translation strings are not supposed to get evaluated, they *must* be treated as plain string literals.

The issue is actually twofold:
[...]

PR #8664, opened 2026-05-27
luci-base: extend String.format() with named placeholder support

jow-:
Add support for `%{key}` and `%{key:fmt}` named placeholders in `String.format()`.

The key is looked up across all plain-object arguments passed to `format()`, scanning left-to-right and using the first object that contains the key. This allows freely mixing positional and named substitutions:

```js
'%s test %{foo:s}'.format('blah', {bar: true}, {foo: 123})
// => 'blah test 123'
```

- `%{key}` implies `%{key:s}` (string)
- `%{key:fmt}` supports the full existing format syntax: pad character, justification, minimum width, precision, and all type specifiers (`b`, `c`, `d`, `u`, `f`, `o`, `s`, `x`, `X`, `h`, `q`, `t`, `m`)
- e.g. `%{count:03d}` formats key `count` as a zero-padded three-digit integer

Ref: https://github.com/openwrt/luci/issues/8603#issuecomment-4554956879

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/cbi.js
+++ b/modules/luci-base/htdocs/luci-static/resources/cbi.js
@@ -675,133 +675,165 @@ String.prototype.format = function()
 		return s;
 	}
 
+	const dicts = [];
+	for (let i = 0; i < arguments.length; i++)
+		if (arguments[i] !== null && typeof(arguments[i]) === 'object' && !Array.isArray(arguments[i]))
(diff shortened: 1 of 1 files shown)


systemcrash (approved):

systemcrash:
?

jow-:
Merged via a31eccfa1494e6cb8afc90963ae7081c800d25e0

Issue #8650, opened 2026-05-21
Claude Code Review Report


qarmin:
Recently, I ran Claude Code across several projects(like image-rs, lofty of symphonia) to look for potential issues.

In this projects, a significant portion of reported findings turned out to be incorrect or minor false positives, but a non-trivial subset was valid and included real issues ranging from incorrect comments to actual logic bugs. As a result, the findings generally require manual validation to separate noise from actionable problems.

Full Reports (`findings-interactive.html` - interactive report view for manual inspection, `findings-short.md` - compact list ready to copy into GitHub, `findings-table.html` - compact tabular report version):

[findings-interactive.html](https://github.com/user-attachments/files/28113514/findings-interactive.html)
[findings-short.md](https://github.com/user-attachments/files/28113513/findings-short.md)
[findings-table.html](https://github.com/user-attachments/files/28113512/findings-table.html)

Example findings (this is only a subset of the findings, specifically those most likely to be actual bugs; for the full list, see the reports above):

=================================

### API_2 `HIGH`

**Description:** px5g.util.pem2der calls `data:gsub({...})` with only a table argument. Lua's string.gsub requires `(pattern, repl)`, where a table can be a replacement but not the pattern. This call will error at runtime as soon as it is invoked.

**Locations:**

```
libs/luci-lib-px5g/lua/px5g/util.lua:41-44
   41 | function pem2der(data)
   42 | 	local b64 = data:gsub({["\n"] = "", ["%-%-%-%-%-.-%-%-%-%-%-"] = ""})
   43 | 	return nixio.bin.b64decode(b64)
   44 | end
```

**Fix:** Run two gsub passes: e.g. `data = data:gsub('%-%-%-%-%-.-%-%-%-%-%-', ''):gsub('\n', '')`.

### CPY_1 `HIGH`

**Description:** When saving container updates, dockerman reads BlkioWeight from the form using the wrong key: `toInt(gethc('blkio_weight'))`. The form option was registered as 'BlkioWeight', and the JSON map data uses that exact PascalCase key. As a result, blkio_weight changes are never written and the Docker API receives `undefined`.

**Locations:**

```
applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js:1332-1347
 1332 | 				const createBody = {
 1333 | 
 1334 | 					CpuShares: toInt(gethc('CpuShares')),
 1335 | 					Memory: toInt(gethc('Memory')),
[...]

PR #8665, opened 2026-05-27
luci-app-attendedsysupgrade: filemode should be octal

efahl:
Fix decimal "600" file mode that should have octal prefix.

Link: #8650

Tested on x86/64 main snapshot.


systemcrash:
Merged. Thanks @efahl

PR #8669, opened 2026-05-27
github: add LLM review workflow and project rules

hauke:
Add a nightly LLM-driven PR review for the LuCI repo, calling the reusable workflow in actions-shared-workflows. The extra_repos list pre-clones LuCI's typical backends (rpcd, netifd, firewall4, ucode, uhttpd, packages, openwrt) so the routine can verify cross-repo references without re-cloning per PR.

Also add .github/llm-review-rules.md with two project-specific rules the system prompt does not cover: redirecting PRs that touch out-of-tree luci-app-* packages, and checking that new UCI options have a backend consumer.

### DIFF
--- a/.github/llm-review-rules.md
+++ b/.github/llm-review-rules.md
@@ -1,0 +1,41 @@
+# LLM review rules — openwrt/luci
+
+LuCI-specific patterns to flag. The review routine reads this at
+session start. General contribution rules (commit subject prefix,
+Signed-off-by, real-name author, line length) live in
+[`CONTRIBUTING.md`](../CONTRIBUTING.md) and are already enforced by
+the formality CI bot — don't duplicate those checks.
+
+## Out-of-tree luci apps
+
+A few `luci-app-*` packages are maintained in their own repositories,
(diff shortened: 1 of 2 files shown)


.github/llm-review-rules.md
hauke, line 1: I auto generated this file.
This is the system prompt: https://github.com/openwrt/actions-shared-workflows/blob/main/llm-review-prompts/llm-nightly-digest.md

It will read this file in addition. It should contain topics the LLM should check specific to LuCI.

BKPepe:
Man, this looks promising. I saw it in the main repository, and now it is available for LuCI as well. Good job, Hauke.

​Nevertheless, this would deserve to be in the packages feed too. Currently, I trigger Copilot manually, and sometimes it is nice to see how various AIs complement each other or, in the better case, argue.

​Ideally, this should be moved to shared workflows, wouldn't you agree? I caught a glimpse somewhere, I think on the mailing list, that there is a restriction on the number of runs per day.

systemcrash:
Wait until you see your unread notifications :D

PR #8689, opened 2026-06-07
luci-app-privoxy: add support for https inspection

ricsc:
Enables the user to create a custom ca and enable https inspection

### Description
Add support for https inspection

### Screenshot
https-inspection

### Maintainer 
@hnyman 

---

## Tested on
**OpenWrt version:** TRUNK and 25.12
**LuCI version:** openwrt-25.12 branch 26.155.67407
**Web browser(s):** Chrome,Firefox

---

## Checklist
- [x] Includes what it depends on: openwrt/packages#29664

### DIFF
--- a/applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
+++ b/applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
@@ -260,7 +260,10 @@ return view.extend({
 		let caCertPath = s.taboption("https", form.DummyValue, '_ca_cert_path', _('CA Certificate Path'),
 			_('Path to the CA certificate file. Install this in your browser/trusted CA store on each client.'));
 		caCertPath.rawhtml = true;
-		caCertPath.default = '/etc/privoxy/ssl/ca-cert.pem';
+		caCertPath.cfgvalue = function(section_id) {
+			var dir = uci.get(this.map.config, section_id, 'certdir') || '/etc/privoxy/ssl';
+			return dir + '/ca-cert.pem';
+		};
 
 		// Download button
 		let downloadBtn = s.taboption("https", form.Button, '_download_ca_cert', _('Download CA Certificate'),
@@ -268,8 +271,7 @@ return view.extend({
 		downloadBtn.inputstyle = 'primary';
 		downloadBtn.inputtitle = _('Download CA Certificate');
 		downloadBtn.onclick = L.bind(function() {
-			var certDir = document.querySelector('input[name="w.-privoxy.-privoxy.certdir"]');
-			var dir = certDir ? certDir.value : '/etc/privoxy/ssl';
+			var dir = certdir.formvalue('privoxy') || '/etc/privoxy/ssl';
 			var certPath = dir + '/ca-cert.pem';
 
 			fs.read_direct(certPath, 'blob').then(function(blob) {
@@ -298,12 +300,10 @@ return view.extend({
 			if (confirm(_('Are you sure you want to regenerate the CA certificate? This will cause SSL warnings on all clients until the new certificate is installed.'))) {
 				// Create marker file to trigger certificate regeneration
 				return fs.write('/etc/privoxy/regenerate_ca', '1').then(function() {
-					console.log('UCI: marker file created successfully');
 					return callRcInit('privoxy', 'reload');
 				}).then(function() {
 					L.ui.addNotification(null, E('p', {}, _('CA certificate has been regenerated.')), 'info');
 				}).catch(function(err) {
-					console.error('UCI error:', err);
 					L.ui.addNotification(null, E('p', {}, _('Failed to regenerate CA certificate: ') + err.message), 'error');
 				});
 			}
@@ -459,27 +459,6 @@ return view.extend({
 		o.description = _("Log the applying actions");
 		o.orientation = "horizontal";
 
-
-		// Post-render: Setup dynamic behavior
-		return m.render().then(function(node) {
-			// Update certificate path display
-			function updateCertPath() {
-				var certDir = document.querySelector('input[name="w.-privoxy.-privoxy.certdir"]');
-				var certPathDisplay = document.querySelector('input[name="w.-privoxy.-privoxy._ca_cert_path"]');
-				if (certDir && certPathDisplay) {
-					var dir = certDir.value || '/etc/privoxy/ssl';
-					certPathDisplay.value = dir + '/ca-cert.pem';
-				}
-			}
-
-			// Update path when certdir changes
-			var certDirInput = document.querySelector('input[name="w.-privoxy.-privoxy.certdir"]');
-			if (certDirInput) {
-				certDirInput.addEventListener('input', updateCertPath);
-				updateCertPath();
-			}
-
-			return node;
-		});
+		return m.render();
 	}
 });
--- a/applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
+++ b/applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
@@ -6,7 +6,23 @@
 'require fs';
 'require view';
 
+const callRcInit = rpc.declare({
+	object: 'rc',
+	method: 'init',
+	params: [ 'name', 'action' ],
+	expect: { result: false }
+});
+
 return view.extend({
+
+	handleSaveApply: function(ev, mode) {
+		var Fn = L.bind(function() {
+			callRcInit('privoxy', 'reload');
+			document.removeEventListener('uci-applied', Fn);
+		});
+		document.addEventListener('uci-applied', Fn);
+		this.super('handleSaveApply', [ev, mode]);
+	},
 
 	render: function(data) {
 		const m = new form.Map('privoxy', _('Privoxy'),
@@ -76,7 +92,7 @@ return view.extend({
 		// CONFDIR
 		let confdir = s.taboption('filtering', form.Value, 'confdir', _('Configuration Directory'),
 			_('The directory where the other configuration files are located.'));
-		confdir.default = '/var/etc/privoxy';
+		confdir.default = '/etc/privoxy';
 		confdir.rmempty = false;
 
 		// TEMPLDIR
@@ -143,7 +159,7 @@ return view.extend({
 			_('Value range 1 to 4096, no entry defaults to 4096'));
 		buffer.default = 4096;
 		buffer.rmempty = true;
-		buffer.datatype = 'and(uinteger,min(1),max(4096))'
+		buffer.datatype = 'and(uinteger,min(1),max(4096))';
 
 		// TOGGLE
 		let toggle = s.taboption('access', form.Flag, 'toggle', _('Toggle Status'),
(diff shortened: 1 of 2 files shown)


applications/luci-app-privoxy/root/usr/share/rpcd/acl.d/luci-app-privoxy.json
```diff
@@ -2,10 +2,19 @@
...
+				"/etc/privoxy/ssl/ca-cert.pem": ["read"]
+			}
 		},
 		"write": {
-			"uci": [ "privoxy" ]
+			"uci": [ "privoxy" ],
+			"file": {
+				"/etc/config/privoxy": ["write"],
+				"/etc/privoxy/regenerate_ca": ["write"]
+			}
```
openwrt-ai, line 20: The view calls `callRcInit('privoxy', 'reload')` (the `rc` ubus object's `init` method) in both `handleSaveApply` and the regenerate handler, but this ACL never grants access to it. Without a `"ubus": { "rc": ["init"] }` entry, non-root ACL groups will get an access-denied error on the reload call. Peer apps that use `rc init` grant it in both scopes (see [`luci-app-unbound`](https://github.com/openwrt/luci/blob/f673f16523b7a28fc525027cb1c9c540fa6c4804/applications/luci-app-unbound/root/usr/share/rpcd/acl.d/luci-app-unbound.json) and [`luci-app-ocserv`](https://github.com/openwrt/luci/blob/f673f16523b7a28fc525027cb1c9c540fa6c4804/applications/luci-app-ocserv/root/usr/share/rpcd/acl.d/luci-app-ocserv.json)).

```suggestion
		"file": {
				"/etc/config/privoxy": ["write"],
				"/etc/privoxy/regenerate_ca": ["write"]
			},
			"ubus": {
				"rc": [ "init" ]
			}
```

---

applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
```diff
@@ -203,6 +219,108 @@ return view.extend({
...
+		caCertPath.rawhtml = true;
+		caCertPath.default = '/etc/privoxy/ssl/ca-cert.pem';
+
+		// Download button
+		let downloadBtn = s.taboption("https", form.Button, '_download_ca_cert', _('Download CA Certificate'),
+			_('Click to download the Privoxy CA certificate for installation on clients.'));
+		downloadBtn.inputstyle = 'primary';
+		downloadBtn.inputtitle = _('Download CA Certificate');
+		downloadBtn.onclick = L.bind(function() {
+			var certDir = document.querySelector('input[name="w.-privoxy.-privoxy.certdir"]');
```
openwrt-ai: This selector will never match any element, so the download always silently falls back to the default `/etc/privoxy/ssl`. LuCI form widgets do not emit a `name="w.-..."` attribute. For a `form.Value`, the `<input>` carries `id="widget.cbid.privoxy.privoxy.certdir"` and no `name` attribute at all (see [`form.js` `renderWidget`](https://github.com/openwrt/luci/blob/f673f16523b7a28fc525027cb1c9c540fa6c4804/modules/luci-base/htdocs/luci-static/resources/form.js#L4519) passing only `id`, and [`ui.js` `Textfield.render`](https://github.com/openwrt/luci/blob/f673f16523b7a28fc525027cb1c9c540fa6c4804/modules/luci-base/htdocs/luci-static/resources/ui.js#L375) setting `name` from `options.name`, which is undefined here).

Use the form API instead of DOM scraping, e.g. read the live value via the option's `formvalue`:

```js
var dir = certdir.formvalue('privoxy') || '/etc/privoxy/ssl';
```

The same broken selector is used in the `updateCertPath()` post-render handler (lines 467, 468, 476), so the dynamic path display and the `input` listener never attach either; please convert those to the form API as well.

---

applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
```diff
@@ -203,6 +219,108 @@ return view.extend({
...
+		// Regenerate button
+		let regenBtn = s.taboption("https", form.Button, '_regenerate_ca', _('Regenerate CA Certificate'),
+			_('Click to delete and regenerate the CA certificate (will cause SSL warnings on clients until new cert is installed).'));
+		regenBtn.inputstyle = 'negative';
+		regenBtn.inputtitle = _('Regenerate CA Certificate');
+		regenBtn.onclick = L.bind(function() {
+			if (confirm(_('Are you sure you want to regenerate the CA certificate? This will cause SSL warnings on all clients until the new certificate is installed.'))) {
+				// Create marker file to trigger certificate regeneration
+				return fs.write('/etc/privoxy/regenerate_ca', '1').then(function() {
+					console.log('UCI: marker file created successfully');
```
openwrt-ai: nit: leftover debug logging. `console.log('UCI: marker file created successfully')` here and `console.error('UCI error:', err)` on line 306 should be removed before merge; the user-facing notifications already cover both the success and error paths.

---

openwrt-ai (commented):
Reviewed the HTTPS-inspection changes (1 commit, 2 files). A few issues to address; see inline comments. The backend dependency on openwrt/packages#29660 is disclosed in the checklist, so the new UCI options not yet existing in a consumer is expected and not flagged.

---

applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
```diff
@@ -337,6 +457,26 @@ return view.extend({
...
+					var dir = certDir.value || '/etc/privoxy/ssl';
+					certPathDisplay.value = dir + '/ca-cert.pem';
+				}
+			}
+
+			// Update path when certdir changes
+			var certDirInput = certdir.formvalue('privoxy') || '/etc/privoxy/ssl';
+			if (certDirInput) {
+				certDirInput.addEventListener('input', updateCertPath);
+				updateCertPath();
```
openwrt-ai: This post-render block is broken after the `querySelector` → `formvalue` refactor. `certDir`, `certPathDisplay` and `certDirInput` are now plain strings, but the code still treats them as DOM elements:

- L468 assigns `certPathDisplay.value = ...` on a string — a no-op, the displayed CA path is never updated.
- L475 calls `certDirInput.addEventListener(...)` on a string, which throws `TypeError: ...addEventListener is not a function` during render, aborting the rest of the `.then()`.
- L467 reads `certDir.value` (always `undefined` on a string).

Since `certdir.formvalue('privoxy')` is evaluated once at render time and there's no live DOM element to subscribe to here, the whole `updateCertPath`/listener mechanism can't work as written. Either drop this block entirely (the `_ca_cert_path` DummyValue already shows a static default), or attach an `onchange`/`change` handler on the `certdir` form widget and update the DummyValue via its own `cfgvalue`/widget API rather than string manipulation.

---

applications/luci-app-privoxy/root/usr/share/rpcd/acl.d/luci-app-privoxy.json
```diff
@@ -2,10 +2,22 @@
...
 		"write": {
-			"uci": [ "privoxy" ]
+			"uci": [ "privoxy" ],
+			"file": {
+				"/etc/config/privoxy": ["write"],
+				"/etc/privoxy/regenerate_ca": ["write"]
+			},
+			"ubus": {
+				"rc": [ "init" ]
+			}
```
openwrt-ai: nit: trailing tab after the closing brace.

```suggestion
			}
```

---

openwrt-ai (commented):
Re-reviewed the changes since the last review (1 amended commit). The prior feedback was addressed — `document.querySelector` replaced with `certdir.formvalue('privoxy')` and the debug `console.log`/`console.error` calls removed — but the `formvalue` refactor of the post-render `updateCertPath` block left it operating on strings as if they were DOM elements, which now throws at render time. See inline. One nit on trailing whitespace.

---

applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
```diff
@@ -203,6 +219,108 @@ return view.extend({
...
+		certKeySize.default = '2048';
+		certKeySize.datatype = 'uinteger';
+
+		// CA Certificate status display
+		let caCertPath = s.taboption("https", form.DummyValue, '_ca_cert_path', _('CA Certificate Path'),
+			_('Path to the CA certificate file. Install this in your browser/trusted CA store on each client.'));
+		caCertPath.rawhtml = true;
+		caCertPath.cfgvalue = function(section_id) {
+    		var dir = uci.get(this.map.config, section_id, 'certdir') || '/etc/privoxy/ssl';
+    		return  dir + '/ca-cert.pem</code>';
```
openwrt-ai: The returned string has a stray closing `</code>` with no matching opening `<code>`. Since `caCertPath.rawhtml = true`, this emits a dangling, unbalanced HTML tag into the page (likely a leftover from a previous markup-wrapped version). The path should be plain text.

The two lines are also indented with leading spaces followed by tabs, unlike the rest of the file (pure tabs). Suggestion fixes both:

```suggestion
			var dir = uci.get(this.map.config, section_id, 'certdir') || '/etc/privoxy/ssl';
			return dir + '/ca-cert.pem';
```

---

openwrt-ai (commented):
Re-reviewed the changes since the last review (1 amended commit). The two prior issues are resolved — the broken `updateCertPath` post-render block was removed and the trailing tab in the ACL file was fixed. The `cfgvalue` replacement for the CA path display introduced one new issue; see inline.

---

applications/luci-app-privoxy/htdocs/luci-static/resources/view/privoxy/privoxy.js
```diff
@@ -203,6 +219,108 @@ return view.extend({
...
+		certKeySize.default = '2048';
+		certKeySize.datatype = 'uinteger';
+
+		// CA Certificate status display
+		let caCertPath = s.taboption("https", form.DummyValue, '_ca_cert_path', _('CA Certificate Path'),
+			_('Path to the CA certificate file. Install this in your browser/trusted CA store on each client.'));
+		caCertPath.rawhtml = true;
+		caCertPath.cfgvalue = function(section_id) {
+    		var dir = uci.get(this.map.config, section_id, 'certdir') || '/etc/privoxy/ssl';
+    		return  dir + '/ca-cert.pem';
```
openwrt-ai: nit: this commit removed the dangling `</code>` but left these two lines indented with four leading spaces followed by tabs, unlike the rest of the file (pure tabs). There's also a double space in `return  dir`. The earlier suggestion fixed both; applying it here:

```suggestion
			var dir = uci.get(this.map.config, section_id, 'certdir') || '/etc/privoxy/ssl';
			return dir + '/ca-cert.pem';
```

---

openwrt-ai (commented):
Re-reviewed the one new commit since the last review. It resolves the prior `</code>` finding (the dangling tag in the CA cert path is gone). One leftover whitespace nit remains on the line it touched; see inline.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The only change since the last review converts the leading spaces to tabs and removes the double space in the `caCertPath.cfgvalue` block, resolving the leftover whitespace nit.

---

hnyman:
cc @commodo who is the privoxy maintainer

commodo (approved):
LGTM

The LuCI stuff is a bit foreign to me.
But the config stuff looks good

PR #8696, opened 2026-06-08
luci-base: fix setPassword handling

ckorber:
### Description

The earlier commits didn't work for creating new users in rpcd.
Also, creating new users resulted in duplicated rpcd logins, because save and apply also created the uci section. So therefore, `setPassword` returns the generated hash and let `luci-mod-system` and `luci-app-acl` set the new hash.

### Maintainer _(preferred)_

@systemcrash @AndyChiang888 

---

## Tested on

**LuCI version:** luci-openwrt-25.12
**Web browser(s):** firefox

---

## Checklist

### DIFF
--- a/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
+++ b/applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
@@ -15,7 +15,6 @@ const callSetPassword = rpc.declare({
 	object: 'luci',
 	method: 'setPassword',
 	params: [ 'username', 'password', 'oldpassword', 'rpcd' ],
-	expect: { result: 1 }
 });
 
 function checkPassword(value) {
@@ -309,9 +308,12 @@ return view.extend({
 			const user = this.map.lookupOption('username', section_id)[0].formvalue(section_id);
 
 			if (variant.formvalue(section_id) == 'crypted' && value.substring(0, 3) != '$1$')
-				return callSetPassword(user, value, '', true).then(function(success) {
-					if (!success)
+				return callSetPassword(user, value, '', true).then(function(result) {
+					if (!result || !result.result)
 						throw new Error('Failed to create password');
+
+					if (result.hash)
+						uci.set('rpcd', section_id, 'password', result.hash);
 				});
 		};
 		o.remove = function() {};
--- a/modules/luci-base/root/usr/share/rpcd/ucode/luci
+++ b/modules/luci-base/root/usr/share/rpcd/ucode/luci
@@ -46,8 +46,7 @@ function callPackageVersionCheck(pkg) {
 	return version;
 }
 
-function set_new_pwd(user, pwd) {
-	const ctx = cursor();
+function set_new_pwd(ctx, user, pwd, change) {
 	let cmd, fd, value;
 
 	cmd = sprintf('/usr/sbin/uhttpd -m %s', shellquote(pwd));
@@ -56,17 +55,20 @@ function set_new_pwd(user, pwd) {
 
 	fd.close();
 
-	ctx.foreach('rpcd', 'login', s => {
-		if (s.username == user) {
-			ctx.set('rpcd', s['.name'], 'password', value);
-			ctx.commit();
-		}
-	});
+	if (change) {
+		ctx.foreach('rpcd', 'login', s => {
+			if (s.username == user) {
+				ctx.set('rpcd', s['.name'], 'password', value);
+				ctx.commit('rpcd');
+			}
+		});
+	}
+
+	return value;
 }
 
-function check_oldpwd(user, pwd) {
+function check_oldpwd(ctx, user, pwd) {
 	let cmd, fd, value;
-	const ctx = cursor();
 	let valid = false;
 
 	cmd = sprintf('/usr/sbin/uhttpd -m %s', shellquote(pwd));
@@ -82,7 +84,7 @@ function check_oldpwd(user, pwd) {
 	return valid;
 }
 
-function check_user(source, entry) {
+function check_user(ctx, source, entry) {
 	let found = false;
 
 	switch (source) {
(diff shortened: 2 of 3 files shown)


AndyChiang888:
Thanks, I'll test it later.

modules/luci-base/root/usr/share/rpcd/ucode/luci
```diff
@@ -561,29 +561,21 @@ const methods = {
...
+					set_new_pwd(uci, user, pwd, true);
 					return { result: 0 };
-			} else {
-				/*
-				 * create user to keep logic of luci-app-acl
-				 */
-				const sid = uci.add('rpcd', 'login');
-				uci.set('rpcd', sid, 'username', user);
-				uci.commit('rpcd');
+				}
```
openwrt-ai, line 570: This branch is reached when the old password is missing or wrong, yet it calls `set_new_pwd(..., true)` which commits the new password to `rpcd` before returning `{ result: 0 }`. That defeats the `check_oldpwd` guard: an existing user's password is changed even when authentication fails, and the call still reports failure. Drop the write so the failure path leaves the password untouched.

```suggestion
				if (!oldpwd || !check_oldpwd(uci, user, oldpwd))
					return { result: 0 };
			}
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
```diff
@@ -175,11 +171,23 @@ return view.extend({
 				formData.data.pw1,
 				oldpw ? oldpw : '',
 				rpcd ? true : false,
-			).then(function(success) {
-				if (success)
+			).then(function(result) {
+				if (result) {
```
openwrt-ai: Since `expect: { result: 1 }` was removed from the `callSetPassword` declaration, the promise now resolves to the full ubus object (e.g. `{ result: 0 }` or `{ result: 1, hash }`) instead of a scalar. `if (result)` is therefore truthy even on the backend failure return `{ result: 0 }`, so the "successfully changed" notification fires on failure and the `else` branch is dead code. Test the actual field instead:

```suggestion
				if (result && result.result) {
```

---

applications/luci-app-acl/htdocs/luci-static/resources/view/system/acl.js
```diff
@@ -309,9 +308,12 @@ return view.extend({
 			const user = this.map.lookupOption('username', section_id)[0].formvalue(section_id);
 
 			if (variant.formvalue(section_id) == 'crypted' && value.substring(0, 3) != '$1$')
-				return callSetPassword(user, value, '', true).then(function(success) {
-					if (!success)
+				return callSetPassword(user, value, '', true).then(function(result) {
+					if (!result)
```
openwrt-ai: With `expect: { result: 1 }` removed, this promise resolves to the full ubus object rather than a scalar, so `result` is always truthy (even for the failure return `{ result: 0 }`) and this `throw` can never fire — a failed password set is silently ignored. Check the actual result field:

```suggestion
					if (!result || !result.result)
```

nit: line 316 `uci.set('rpcd', section_id, 'password', result.hash)` is missing the trailing semicolon used elsewhere in this file.

---

openwrt-ai (commented):
Reviewed 4 new commits. Two related issues stand out: the backend's `known_user` failure branch now writes the new password before returning `result: 0` (bypassing the old-password check), and both frontends still treat the resolved promise as a scalar after `expect: { result: 1 }` was dropped, so the failure return `{ result: 0 }` is read as success. Details inline.

---

modules/luci-base/root/usr/share/rpcd/ucode/luci
ckorber: I fixed this part

modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
```diff
@@ -175,11 +171,23 @@ return view.extend({
...
+			).then(function(result) {
+				if (result && result.result) {
+					if (rpcd && result.hash) {
+						let sections = uci.sections('rpcd', 'login');
+
+						for (let s of sections) {
+							if (s.username == rpc_user) {
+								uci.set('rpcd', s['.name'], 'password', result.hash);
+								uci.save('rpcd');
+							}
```
openwrt-ai, line 183: This loop only updates a `rpcd.login` section that already exists, but the backend no longer creates one. The base backend used to `uci.add('rpcd', 'login')` for an unknown user ([previous setPassword](https://github.com/openwrt/luci/blob/46cd82328dfb7fbb15961c63f9e674a41cef2500/modules/luci-base/root/usr/share/rpcd/ucode/luci#L583-L588)); that `add` was removed and for a new user the backend now returns the hash with `change=false` and writes nothing ([`luci:576`](https://github.com/openwrt/luci/blob/104ed5df89f31574d3b3fd9c3d26e69274bf8a8f/modules/luci-base/root/usr/share/rpcd/ucode/luci#L576-L577)).

So if `rpc_user` is a name not already present in `rpcd.login`, no `s.username == rpc_user` matches here and the section is never added — the new user silently isn't created (and the guard requiring an old password for `rpc_user` was also dropped in this PR, so the new-user path reaches here). Is `rpc_user` in this view intended only to select existing logins, with new-user creation left to luci-app-acl? If new users should be creatable here, this branch needs to `uci.add('rpcd', 'login')` + set `username` when no match is found.

---

openwrt-ai (commented):
Reviewed 3 new commits since my last review. All three earlier findings (the failure-path write in the backend, and the `result`/`result.result` truthiness checks in both frontends) are addressed correctly. One remaining question about new-user creation in the system password view is inline. Commit messages match their changes.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/password.js
ckorber: This form is intended to change an existing password and not create one. This is handled by `luci-app-acl`, so this works exactly as expected.

ckorber:
Hi @systemcrash , do you approve?

ckorber:
@systemcrash : Could you please review?

systemcrash:
I'm not sure I like the idea of the front-end code handling sensitive password hashes. It seems to me like rpcd should handle this internally...?

PR #8701, opened 2026-06-13
luci-mod-status: resolve named counters in nftables view

dibdot:
### Description

The nftables status view rendered the byte/packet badge by reading .bytes/.packets directly off a rule's counter statement. That only holds for anonymous (inline) counters, whose JSON statement is an object. Named counter objects (counter name "<n>") are serialized as a bare name string instead, so .bytes was undefined and the badge showed "undefinedB". Named counters are used in banIP for example.

Resolve named counter references against the standalone counter objects of the same family and table before rendering, and skip the badge when a reference cannot be resolved instead of emitting garbage.

While at it, replace some whitespaces with tabs as well.

### Screenshot changes
Old nftables view:
nftables_old

New nftables view:
nftables_new

## Tested on
Bananapi BPI-R3, mediatek/filogic, OpenWrt SNAPSHOT (r34876-a1a3659c29)

---

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js
@@ -143,19 +143,34 @@ const action_translations = {
 };
 
 return view.extend({
+
+	resolveCounter(data, spec, counter) {
+		if (typeof(counter) == 'object')
+			return counter;
+
+		for (let d of data)
+			if (typeof(d.counter) == 'object' &&
+				d.counter.family == spec.family &&
+				d.counter.table == spec.table &&
+				d.counter.name == counter)
+				return d.counter;
+
+		return null;
+	},
+
 	load() {
 		return Promise.all([
-			L.resolveDefault(fs.exec_direct('/usr/sbin/nft', [ '--terse', '--json', 'list', 'ruleset' ], 'json'), {}),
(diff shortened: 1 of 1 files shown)


modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js
```diff
@@ -143,19 +143,34 @@ const action_translations = {
 };
 
 return view.extend({
+
+	resolveCounter(data, spec, counter) {
```
openwrt-ai, line 147: nit: This introduces `typeof (...)` with a space here and on the two touched rule branches (the changed `typeof (se)` lines), but the rest of this file consistently uses `typeof(...)` with no space (lines 193, 222, 326, 355-356, 362, 573, 708, 770). The mixed spacing is a deviation from the file's existing style.

```suggestion
		if (typeof(counter) == 'object')
```

---

openwrt-ai (commented):
Reviewed 1 commit; the named-counter resolution logic is correct (standalone counters are matched against rule references by family/table/name, and the badge is skipped when unresolved). Commit message matches the diff. One style nit inline.

---

openwrt-ai (commented):

The previous style nit (`typeof (...)` → `typeof(...)`) has been addressed across all touched branches, and the named-counter resolution logic is unchanged and correct.

---

PR #8703, opened 2026-06-14
luci-app-attendedsysupgrade: add uci-defaults script support

firlin123:
### Description
Added the ability to specify uci-defaults script (`defaults` field in request to ASU) to be sent to ASU server during firmware generation.

*Note: This is my first PR ever and I'm an ESL speaker. Let me know if changes are needed!*

Upgrade modal screenshot:
Screenshot_20260614_052231

Warning in upgrade modal screenshot:
Screenshot_20260615_211324

---

## Tested on
**Board:** Raspberry Pi Compute Module 4 (bcm27xx/bcm2711)
**OpenWrt version:** OpenWrt SNAPSHOT (r35567-83369ca112)
**LuCI version:** LuCI Master (26.207.70497~fb6b224)
**Web browser(s):** Brave 1.90.121 (Official Build) (64-bit), Chromium: 148.0.7778.96

---

## Checklist

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -35,6 +35,14 @@ function detailsBlock(title, content, pre) {
 
 /* -------------------------------------------------------------------------- */
 
+const ERR_OK = 0;
+const ERR_NO_INIT = 1;
+const ERR_INVALID_FILE = 2;
+const ERR_FILE_TOO_BIG = 3;
+const ERR_SERVER_FAILED = 4;
+const ERR_INIT_DISALLOWED = 5;
+const ERR_DISALLOWED_PATH = 6;
+
 const callPackagelist = rpc.declare({
 	object: 'rpc-sys',
 	method: 'packagelist',
@@ -82,6 +90,23 @@ function get_branch(version) {
  */
 function get_revision_count(revision) {
 	return parseInt(revision.substring(1).split('-')[0]);
+}
+
+/**
+ * Resolves relative path segments (`.` and `..`) and normalizes repeated
+ * slashes in a POSIX filesystem path.
+ *
+ * @param {string} path
+ * The input path to normalize
+ * @returns {string}
+ * The resolved, normalized path
+ */
+function resolve_path(path) {
+	return '/' + path.split('/').reduce((acc, part) => {
+		if (part === '..') acc.pop();
+		else if (part && part !== '.') acc.push(part);
+		return acc;
+	}, []).join('/');
 }
 
 return view.extend({
@@ -538,9 +563,6 @@ return view.extend({
 		let { version, target, profile, packages } = firmware;
 		let candidates = [];
 
-		const endpoint = version.endsWith('SNAPSHOT') ? `revision/${version}/${target}` : 'overview';
-		const request_url = `${url}/api/v1/${endpoint}`;
-
 		ui.showModal(_('Searching...'), [
 			E(
 				'p',
@@ -552,7 +574,17 @@ return view.extend({
 			),
 		]);
 
-		L.resolveDefault(request.get(request_url)).then((response) => {
+		const is_snapshot = version.endsWith('SNAPSHOT');
+		const overview_url = `${url}/api/v1/overview`;
+		let promises = [L.resolveDefault(request.get(overview_url))];
+
+		if (is_snapshot) {
+			const snapshot_url = `${url}/api/v1/revision/${version}/${target}`;
+			promises.push(L.resolveDefault(request.get(snapshot_url)));
+		}
+
+		Promise.all(promises).then(async (results) => {
+			let response = results[is_snapshot ? 1 : 0];
 			if (!response.ok) {
 				ui.showModal(_('Error connecting to upgrade server'), [
 					E(
@@ -569,7 +601,7 @@ return view.extend({
 				]);
 				return;
 			}
-			if (version.endsWith('SNAPSHOT')) {
+			if (is_snapshot) {
 				const remote_revision = response.json().revision;
 				if (
 					get_revision_count(revision) < get_revision_count(remote_revision)
@@ -605,6 +637,42 @@ return view.extend({
 			}
 
 			if (candidates.length) {
+				const server_info = results[0]?.ok ? results[0].json().server : null;
+				let init_script_error = ERR_NO_INIT;
+				let init_script_path = uci.get('attendedsysupgrade', 'owut', 'init_script') || '';
+				let init_script_contents = '';
+
+				if (init_script_path) {
+					init_script_path = resolve_path(init_script_path);
+					if (!init_script_path.startsWith('/rom/etc/uci-defaults/') &&
+						!init_script_path.startsWith('/etc/uci-defaults/') &&
+						!init_script_path.startsWith('/root/')) {
+						init_script_error = ERR_DISALLOWED_PATH;
+					} else if (!server_info) {
+						init_script_error = ERR_SERVER_FAILED;
+					} else if (!server_info.allow_defaults) {
+						init_script_error = ERR_INIT_DISALLOWED;
+					} else if (typeof server_info.max_defaults_length !== 'number') {
+						init_script_error = ERR_SERVER_FAILED;
+					} else {
+						let st = await fs.stat(init_script_path).catch(() => null);
+						if (!st || st.type !== 'file') {
+							init_script_error = ERR_INVALID_FILE;
+						} else if (st.size > server_info.max_defaults_length) {
+							init_script_error = ERR_FILE_TOO_BIG;
+						} else if (st.size == 0) {
+							init_script_error = ERR_OK;
+						} else {
+							init_script_contents = await fs.read(init_script_path).catch(() => null);
+							if (init_script_contents != null) {
+								init_script_error = ERR_OK;
+							} else {
+								init_script_error = ERR_INVALID_FILE;
+							}
+						}
+					}
+				}
+
 				let s, o;
 
 				let mapdata = {
@@ -612,6 +680,7 @@ return view.extend({
 						profile,
 						version: candidates[0][0],
 						packages: Object.keys(packages).sort(),
+						init_script_checked: '1'
 					},
 				};
 
@@ -648,6 +717,58 @@ return view.extend({
 					o = s.option(form.DynamicList, 'packages', _('Packages'));
 				}
 
+				switch (init_script_error) {
+					case ERR_OK:
+						o = s.option(
+							form.Flag,
+							'init_script_checked',
+							_('Use init script'),
+							_('Include %s as init script (uci-defaults).').format(E('code', {}, init_script_path).outerHTML)
+						);
+						o.default = '1';
+						o.rmempty = false;
+						break;
+					case ERR_NO_INIT:
+						break;
+					default: {
+						o = s.option(
+							form.DummyValue,
+							'_warning',
+							_('Init script error')
+						);
+						let msg = _('Unknown error. File %s will not be included as init script (uci-defaults) in the built firmware.');
+						switch (init_script_error) {
+							case ERR_INVALID_FILE:
+								msg = _('File %s is invalid and won\'t be included as init script (uci-defaults) in the built firmware.');
+								break;
+							case ERR_FILE_TOO_BIG:
+								msg = _('File %s exceeds %s and won\'t be included as init script (uci-defaults) in the built firmware.');
+								break;
+							case ERR_SERVER_FAILED:
+								msg = _('Unable to fetch build server limits. File %s will not be included as init script (uci-defaults) in the built firmware.');
+								break;
+							case ERR_INIT_DISALLOWED:
+								msg = _('Chosen build server does not allow init scripts (uci-defaults). File %s will not be included as such in the built firmware.');
+								break;
+							case ERR_DISALLOWED_PATH:
+								msg = _('The path %s is not permitted. Only files in /rom/etc/uci-defaults/, /etc/uci-defaults/, or /root/ can be included.');
+								break;
+						}
+						let alertBox = E('div', { class: 'alert-message' });
+						if (init_script_error === ERR_FILE_TOO_BIG) {
+							const kib = (server_info.max_defaults_length / 1024).toFixed(2);
+							alertBox.innerHTML = msg.format(
+								E('code', {}, init_script_path).outerHTML,
+								E('b', {}, [ kib, ' KiB' ]).outerHTML,
+							);
+						} else {
+							alertBox.innerHTML = msg.format(E('code', {}, init_script_path).outerHTML);
+						}
+						o.default = alertBox;
+						break;
+					}
+				}
+
 				L.resolveDefault(map.render()).then((form_rendered) => {
 					ui.showModal(_('New firmware upgrade available'), [
 						E(
@@ -679,6 +800,9 @@ return view.extend({
 													version: mapdata.request.version,
 													profile: mapdata.request.profile
 												};
+												if (init_script_error === ERR_OK && mapdata.request.init_script_checked === '1') {
+													content.defaults = init_script_contents;
+												}
 												this.pollFn = L.bind(function () {
 													this.handleRequest(url, true, content, data, firmware);
 												}, this);
--- a/applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json
+++ b/applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json
@@ -13,10 +13,17 @@
 				],
 				"uci": [
 					"get"
+				],
+				"file": [
+					"read",
+					"stat"
 				]
 			},
 			"file": {
-				"/sys/firmware/efi": [ "list" ]
+				"/sys/firmware/efi": [ "list" ],
+				"/rom/etc/uci-defaults/*": [ "read" ],
+				"/etc/uci-defaults/*": [ "read" ],
+				"/root/*": [ "read" ]
 			},
 			"uci": [
 				"attendedsysupgrade"


applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -650,7 +668,10 @@ return view.extend({
 													...firmware,
 													packages: packages,
 													version: mapdata.request.version,
-													profile: mapdata.request.profile
+													profile: mapdata.request.profile,
+													defaults: (data.defaults_contents && mapdata.request.defaults_checked === '1') 
+														? data.defaults_contents
```
openwrt-ai: nit: trailing whitespace at the end of both lines (after `=== '1')` and after `data.defaults_contents`).

```suggestion
												defaults: (data.defaults_contents && mapdata.request.defaults_checked === '1')
													? data.defaults_contents
```

---

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
```diff
@@ -0,0 +1,35 @@
...
+
+function get_defaults() {
+	const path = uci.cursor().get("attendedsysupgrade", "client", "defaults_path");
+	if (!path) {
+		return { contents: "", path: "" };
+	}
+
+	const st = fs.stat(path);
+	// Enforce the 40 KB official ASU server limit and check for empty files
+	if (!st || st.type !== "file" || st.size <= 0 || st.size > 40960) {
```
openwrt-ai: When the configured `defaults_path` points to a file that is empty, larger than 40 KB, or not a regular file, `get_defaults` returns `contents: ""` and the frontend then nulls out `path` too ([`if(!defaults_path || !defaults_contents.trim())` in overview.js:719](``https://github.com/openwrt/luci/blob/1888178ed0b2d20d5f81afa8c4082a1b1847b10a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L719)).`` The result is that the "Use uci-defaults" option silently disappears with no indication to the user that their configured file was rejected (e.g. for exceeding the size limit). Is a silent drop intended here, or would surfacing the reason (oversized / unreadable) be preferable so the user can correct the path? Not blocking — just flagging the UX gap.

---

openwrt-ai (commented):
Reviewed the single new commit. The change is well-scoped: the new `40_luci-attendedsysupgrade` uci-defaults script matches the established cache-clear/rpcd-reload convention used by other luci apps, the rpcd ucode `get_defaults` method (`fs.stat`/`fs.readfile`, size cap) is correct API usage, the ACL entry is wired to the new method, and the `defaults` field is correctly omitted from the build request when unset. Two minor inline notes below (trailing whitespace; silent rejection of oversized/invalid files). The commit message line-length warning is already reported by the formality bot.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: Fixed in 2nd commit.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
firlin123: Fixed in 2nd commit.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
```diff
@@ -0,0 +1,60 @@
...
+	const st = fs.stat(path);
+	if (!st || st.type !== "file" || st.size <= 0) {
+		return {
+			contents: "",
+			path: path,
+			error: ERR_INVALID_FILE
+		};
+	}
+	// Enforce the 40 KiB official ASU server limit
+	if (st.size > 40960) {
```
efahl: You can't hard code this number as it varies between servers and can change arbitrarily even on the "official" upstream.  You need to fetch it from the server limits:   `curl https://sysupgrade.openwrt.org/json/v1/overview.json | jsonfilter -e '$.server.max_defaults_length'`

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
```diff
@@ -0,0 +1,60 @@
...
+import * as fs from "fs";
+import * as uci from "uci";
+
+const ERR_OK = 0;
+const ERR_NO_DEFAULTS = 1;
+const ERR_INVALID_FILE = 2;
+const ERR_FILE_TOO_BIG = 3;
+
+function get_defaults() {
+	const path = uci.cursor().get("attendedsysupgrade", "client", "defaults_path");
```
efahl: This value is already in the config in `attendedsysupgrade.owut.init_script`.  See https://openwrt.org/docs/guide-user/installation/sysupgrade.owut#using_a_uci-defaults_script

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -585,6 +596,7 @@ return view.extend({
 						profile,
 						version: candidates[0][0],
 						packages: Object.keys(packages).sort(),
+						defaults_checked: '1'
```
efahl: Shouldn't this be a bool?

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/configuration.js
```diff
@@ -61,6 +61,14 @@ return view.extend({
 		o.default = '0';
 		o.rmempty = false;
 
+		o = s.option(
+			form.Value,
+			'defaults_path',
+			_('Defaults file path'),
+			_('Specify the path to your first-boot script (e.g., <code>/rom/etc/uci-defaults/99-asu-defaults</code> inserted by the Firmware Selector.)')
```
efahl: If no config value exists, but the `/rom/.../99-asu-defaults` file exists, this field should auto-populate with the existing file path.

openwrt-ai (commented):
Reviewed 1 new commit (`d4e5e16` "improve uci-defaults validation"); no new issues found.

The new commit cleanly addresses both notes from my prior review: oversized/invalid/empty files now surface an explicit `alert-message` warning instead of silently dropping the option, and the trailing whitespace is gone. The ucode error-code handling (`fs.stat` type/size checks, `length(trim(contents))` emptiness check) is correct, the `expect`/resolveDefault defaults were updated consistently, and the commit message matches the diff. Wrapping `data.defaults_path` in `E('code', {}, ...).outerHTML` also correctly escapes the path before it reaches `innerHTML`.

The substantive open points (hardcoded 40 KiB limit vs. the server's `max_defaults_length`, reuse of `attendedsysupgrade.owut.init_script`, the bool question, and path auto-population) are already covered in @efahl's inline comments on this head, so I won't duplicate them.

---

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
firlin123: Removed the hard-coded number and now using server's max_defaults_length in the latest commit.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
firlin123: I was not aware that a config for this already existed. Switched to using `attendedsysupgrade.owut.init_script` in the latest commit.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: I tried making it bool but the `form.Flag` only accepts `"1"` and `1` (which it type-coerces to `"1"` anyway).

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/configuration.js
firlin123: Because I switched to `attendedsysupgrade.owut.init_script` as you suggested, I don't control the defaults of this config anymore, `owut` does.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -525,166 +537,259 @@ return view.extend({
...
+					E('div', { class: 'btn', click: ui.hideModal }, _('Close')),
+				]),
+			]);
+			return;
+		}
 
-			// allow to re-install running firmware in advanced mode
-			if (advanced_mode == 1) {
-				candidates.unshift([version, revision]);
+		const server_info = results[0].ok ? results[0].json().server : null;
```
openwrt-ai: `results[0]` can be `undefined` here, throwing `TypeError: Cannot read properties of undefined (reading 'ok')`. The `overview` request is wrapped in [`L.resolveDefault(request.get(overview_url))`](``https://github.com/openwrt/luci/blob/1327ff71ca3d10068a3897a88b284a6282144176/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L542)`` with no default value, and [`resolveDefault(value)` resolves to `undefined` on a rejecting promise](https://github.com/openwrt/luci/blob/1327ff71ca3d10068a3897a88b284a6282144176/modules/luci-base/htdocs/luci-static/resources/luci.js#L3117). For a SNAPSHOT device the `!response.ok` guard above only inspects `results[1]` (the snapshot request), so when the snapshot request succeeds but the `overview` request rejects (timeout/network error), execution reaches this line with `results[0] === undefined` and crashes.

Give the `overview` (and snapshot) requests a default object so the `.ok`/`.json()` accesses stay safe:

```suggestion
		const server_info = results[0]?.ok ? results[0].json().server : null;
```

Better still, also pass `{ ok: false }` as the default to the two `request.get` wrappers on lines 542/546 so the existing `response.ok` check above can't throw either.

Separately: when `server_info` is present and `allow_defaults` is true but `max_defaults_length` is missing (older server), `Math.floor(undefined)` yields `NaN`, which serializes to `null` and the backend returns `ERR_INVALID_ARGS` (7) — a code the frontend doesn't define, so it falls through to the generic "Unknown error" message. Worth guarding for if such servers exist.

---

openwrt-ai (commented):
Reviewed 1 new commit (`1327ff7` "switch to owut config and server-driven limits"). The switch to `attendedsysupgrade.owut.init_script` and the server-supplied `max_defaults_length` correctly addresses @efahl's earlier notes, and the new `ERR_EMPTY_FILE`/`ERR_SERVER_FAILED`/`ERR_INIT_DISALLOWED` error paths are wired consistently between the rpcd ucode and the frontend. One inline note on a `TypeError` reachable when the `overview` request rejects for a SNAPSHOT device.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: Addressed both in the latest force push (didn't think such minor change warrants its own commit).

openwrt-ai (commented):
Reviewed 1 new commit (`c5482d5`, replacing the previously-reviewed `1327ff7`); no new issues found.

The only delta since my last review is the `overview.js` refactor that hoists `init_script_path` out of the branches and guards `results[0]` with optional chaining (`results[0]?.ok`). This correctly addresses the `TypeError` I flagged previously: when the `overview` request rejects (e.g. for a SNAPSHOT device), `results[0]` is the `resolveDefault` fallback rather than a response object, and `?.ok` now yields `null` → `server_info = null` → `ERR_SERVER_FAILED`. The added `typeof server_info.max_defaults_length !== 'number'` check also avoids passing `NaN` into `callGetInitScript(Math.floor(...))` when the server advertises `allow_defaults` without a numeric limit. The commit subject (75 chars) and message match the diff, and Test Formalities now passes on this head.

---

firlin123:
Fixed eslint `no-case-declarations` error in the latest force push

openwrt-ai (commented):
Reviewed the force-pushed head (`4311997e`, replacing the previously-reviewed `c5482d5`); no new issues found.

The only delta since my last review is wrapping the `default:` case body in `overview.js` (`handleCheck`'s `switch (init_script_error)`) in a block `{ }`. This correctly resolves the eslint `no-case-declarations` error caused by the lexical declarations (`let msg`, `let alertBox`, `const kib`) in that case, with no behavioral change. The reworded third commit subject is now 75 chars and matches its diff.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/configuration.js
```diff
@@ -61,6 +66,22 @@ return view.extend({
 		o.default = '0';
 		o.rmempty = false;
 
+		// For now, only display if the owut section exists in uci (owut installed or section is manually added).
+		// We should probably move init script path from attendedsysupgrade.owut.init_script to some common place
```
efahl: Yeah, too late now.  That would require a significant rewrite of how owut gather's uci default values, and corresponding upgrade scripting, so not going to happen.  LuCI already grabs another owut value (rootfs_size), and will always be a second class citizen compared to the far-more-capable owut.

applications/luci-app-attendedsysupgrade/root/etc/uci-defaults/40_luci-attendedsysupgrade
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+rm -rf /var/luci-modulecache/; rm -f /var/luci-indexcache;
+[ -x /etc/init.d/rpcd ] && /etc/init.d/rpcd reload
+exit 0
```
efahl: Why is this whole file here?  The tmp directories are deleted on reboot, and the rpcd reloading isn't anything that attended sysupgrade should care about.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
```diff
@@ -36,13 +53,20 @@ function get_defaults() {
...
 			contents: "",
 			path: path,
 			error: ERR_INVALID_FILE
 		};
 	}
+	if (length(trim(contents)) <= 0) {
+		return {
+			contents: "",
+			path: path,
+			error: ERR_EMPTY_FILE
```
efahl: If people want to include an empty placeholder script, I don't see any reason to disallow that.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/configuration.js
firlin123: So should we not display that config at all and make it configurable only via cli (like the `owut.rootfs_size`)? Or just remove the "for now" phrasing from the comment?

applications/luci-app-attendedsysupgrade/root/etc/uci-defaults/40_luci-attendedsysupgrade
firlin123: I honestly don't know. I just looked at other luci apps with ucode rpcd service and they all seem to have this file so I copied it here too. Is it not needed in this case?

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
firlin123: Allowed empty files in latest revision.

openwrt-ai (commented):
Reviewed 1 new commit (`ae849bc`, replacing the previously-reviewed `4311997e`); no new issues found.

The delta since my last review drops the `ERR_EMPTY_FILE` error path: an empty file (`st.size <= 0`) now returns `ERR_OK` with empty `contents` instead of a warning, and the `length(trim(contents)) <= 0` whitespace-only check is removed. The error-code constants are renumbered consistently between the rpcd ucode (`ERR_FILE_TOO_BIG`=3, `ERR_INVALID_ARGS`=6) and `overview.js` (`ERR_FILE_TOO_BIG`=3, `ERR_SERVER_FAILED`=4, `ERR_INIT_DISALLOWED`=5), and the shared codes 0–3 still agree across both sides. The net effect is that an empty/whitespace-only init script is treated as a benign no-op (sent as an empty `defaults` string), which is reasonable. The `configuration.js` re-indentation in this commit only restores the base-master formatting (it's not part of the cumulative PR diff), so there's no scope concern. The commit subject is 75 chars and matches its diff, and Test Formalities passes on this head.

---

firlin123:
ping?

applications/luci-app-attendedsysupgrade/root/etc/uci-defaults/40_luci-attendedsysupgrade
firlin123: UPD: I looked into how it works (sorry for not doing this earlier) and you're right. Removed it in the latest forced push.

firlin123:
Rebased on the latest master, removed `uci-defaults/40_luci-attendedsysupgrade` as suggested, and squished separate commits into 1.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -552,166 +566,255 @@ return view.extend({
...
-							response.url
-						)
-					),
-					E('pre', {}, response.responseText),
-					E('div', { class: 'right' }, [
-						E('div', { class: 'btn', click: ui.hideModal }, _('Close')),
-					]),
-				]);
-				return;
+		const is_snapshot = version.endsWith('SNAPSHOT');
```
openwrt-ai, line 560: `endpoint` and `request_url` (lines 555-556, just above the `ui.showModal(_('Searching...'))` block) are dead after this change — they come from master, but their only consumer, `L.resolveDefault(request.get(request_url)).then(...)`, is exactly what this hunk replaces with the `overview_url` / `snapshot_url` pair. They survived the rebase because they sit outside the rewritten region. `version.endsWith('SNAPSHOT')` also ends up evaluated twice (line 555 and here).

Please drop lines 555-556 and keep only `is_snapshot`. Note CI won't catch this: [`'no-unused-vars': ['off', ...]` in eslint.config.mjs:173](https://github.com/openwrt/luci/blob/8a8dfe0e85ef973befd90410d8779429b0437c71/eslint.config.mjs#L173).

---

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
```diff
@@ -0,0 +1,76 @@
...
+	const max_size = req?.args?.max_size;
+	if (type(max_size) !== "int") {
+		return {
+			contents: "",
+			path: "",
+			error: ERR_INVALID_ARGS
+		};
+	}
+
+	const path = uci.cursor().get("attendedsysupgrade", "owut", "init_script");
```
openwrt-ai: This method returns the verbatim contents of any absolute path the UCI option names, and rpcd ucode plugins run as root — there is no restriction on where `init_script` may point.

The problem is that the same ACL group grants both halves: `read.ubus."luci.attendedsysupgrade" = ["get_init_script"]` and [`write` → `ubus.uci: ["set","commit"]` + `uci: ["attendedsysupgrade"]`](``https://github.com/openwrt/luci/blob/8a8dfe0e85ef973befd90410d8779429b0437c71/applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json#L28-L39).`` So a LuCI session that holds only `luci-app-attendedsysupgrade` (no shell, no root file access) can `uci set attendedsysupgrade.owut.init_script=/etc/shadow` (or a WireGuard private key, `/etc/config/*`, …), commit, then call `get_init_script` and read the file back. That turns a per-app config permission into arbitrary root file read.

Worth noting the app already scopes direct file access through the ACL `file` section ([`"/sys/firmware/efi": [ "list" ]`](``https://github.com/openwrt/luci/blob/8a8dfe0e85ef973befd90410d8779429b0437c71/applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json#L21-L23)),`` which this method bypasses.

How would you like to handle it? A path allowlist in this function is the obvious fix, but owut's own example config uses `/root/data/my-init-script.sh`, so restricting to `/etc/uci-defaults` + `/rom/etc/uci-defaults` would be narrower than what owut itself accepts. An alternative is to have the frontend read the file via `fs.read` (rpcd-mod-file) and add a path-scoped `file` entry to the ACL, so the restriction lives in the ACL rather than in this method.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/configuration.js
```diff
@@ -61,6 +66,20 @@ return view.extend({
...
+		// Only display if the owut section exists in uci (owut installed or section is manually added).
+		if (uci.get_first('attendedsysupgrade', 'owut') !== null) {
+			s = m.section(form.TypedSection, 'owut', _('OpenWrt Upgrade Tool'));
+			s.anonymous = true;
+
+			o = s.option(
+				form.Value,
+				'init_script',
+				_('Init script'),
+				_('Specify the path to the init script (e.g., <code>/rom/etc/uci-defaults/99-asu-defaults</code> inserted by the Firmware Selector.)')
```
openwrt-ai: nit: the sentence-ending period is inside the parenthetical (`… Firmware Selector.)`), so the sentence has no terminator. Since this is a translatable string it's worth fixing before it reaches Weblate.

```suggestion
				_('Specify the path to the init script (e.g., <code>/rom/etc/uci-defaults/99-asu-defaults</code> inserted by the Firmware Selector).')
```

---

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
```diff
@@ -0,0 +1,76 @@
+"use strict";
```
openwrt-ai: nit: this file uses double quotes throughout, while 11 of the 12 existing `root/usr/share/rpcd/ucode/luci.*` plugins in the tree use single quotes (`'use strict';`, `import { … } from 'fs';`) — see [luci.upnp](https://github.com/openwrt/luci/blob/8a8dfe0e85ef973befd90410d8779429b0437c71/applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp#L4-L8) and [luci.lldpd](``https://github.com/openwrt/luci/blob/8a8dfe0e85ef973befd90410d8779429b0437c71/applications/luci-app-lldpd/root/usr/share/rpcd/ucode/luci.lldpd#L1-L3).`` Only `luci.advanced-reboot` uses double quotes. Matching the majority style would make this consistent with its siblings.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -552,166 +566,255 @@ return view.extend({
...
-						_('The device runs the latest firmware version %s - %s').format(
-							version,
-							revision
-						)
-					),
-					E('div', { class: 'right' }, [
-						E('div', { class: 'btn', click: ui.hideModal }, _('Close')),
-					]),
-				]);
+		switch(init_script_error) {
```
openwrt-ai: nit: missing space after `switch` — the nested `switch (init_script_error)` a few lines below uses the spaced form, as does the rest of the file.

```suggestion
		switch (init_script_error) {
```

Same class of thing: lines 650 and 657 (`let init_script_path = uci.get_first(…)` and `if (!server_info.allow_defaults || typeof …)`) run well past the ~80-column wrapping the rest of this file sticks to.

---

openwrt-ai (commented):
Re-reviewed after the rebase onto master + squash into a single commit (`8a8dfe0`, replacing the previously-reviewed `ae849bc`). The commit message matches the diff, the `40_luci-attendedsysupgrade` uci-defaults script is gone as @efahl asked, and the removed `ERR_EMPTY_FILE` path stays removed. `luci-base` already pulls in `rpcd-mod-ucode`, `ucode-mod-fs` and `ucode-mod-uci`, so no Makefile dependency is needed for the new plugin, and `config owut 'owut'` in owut's own defaults means `uci.get_first('attendedsysupgrade', 'owut', …)` resolves correctly.

Two things worth a look before merge: one dead-code leftover from the rebase, and the arbitrary-file-read reachable through the new RPC method given the ACL group's own write permissions. Three nits inline as well.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: Ooops, didn't notice it during the rebase. Fixed in latest force push.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
firlin123: Switched to reading the file using LuCI `fs` API and limited it to `/rom/etc/uci-defaults/*`, `/etc/uci-defaults/*`, and `/root/*` with rpcd ACLs. Dropped the `get_init_script` rpcd method.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/configuration.js
firlin123: Fixed in the latest forced push.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/ucode/luci.attendedsysupgrade
firlin123: This file is no longer needed. Removed in the latest forced push.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: Fixed in the latest forced push.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json
```diff
@@ -16,7 +16,10 @@
 				]
```
openwrt-ai, line 20: The new `file` path scopes below say *where* `read` may happen, but nothing in this group grants the ubus `file` object's `read` **method** — rpcd checks both: `session_access(sid, "ubus", "file", "read")` for the call itself, then `session_access(sid, "file", path, "read")` for the path.

`luci-base` grants only [`"file": [ "list" ]` under `ubus`, plus `"/*": [ "list" ]`](https://github.com/openwrt/luci/blob/adeb73a839ebfdd46bf1c25138f193bd37f09a5b/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json#L14-L20) — which is why the pre-existing `fs.stat('/sys/firmware/efi')` works without a `file` entry here, but `read` is not covered by it. A session holding only `luci-base` + `luci-app-attendedsysupgrade` therefore gets `Access denied` from `fs.read(init_script_path)`, and the `.catch(() => null)` in `overview.js` turns that into the misleading `ERR_INVALID_FILE` → *"File %s is invalid"* warning. It works in testing today only because `luci-mod-status-index` happens to grant [`"file": [ "list", "read" ]`](``https://github.com/openwrt/luci/blob/adeb73a839ebfdd46bf1c25138f193bd37f09a5b/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status-index.json#L11-L15)`` — an unrelated package this app shouldn't have to depend on.

Adding the method grant to this group's `read.ubus` block fixes it:

```suggestion
				],
				"file": [
					"read"
				]
```

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -552,166 +557,263 @@ return view.extend({
...
+				init_script_error = ERR_SERVER_FAILED;
+			} else if (!server_info.allow_defaults || typeof server_info.max_defaults_length !== 'number') {
+				init_script_error = !server_info.allow_defaults ? ERR_INIT_DISALLOWED : ERR_SERVER_FAILED;
+			} else {
+				let st = await fs.stat(init_script_path).catch(() => null);
+				if (!st || st.type !== 'file') {
+					init_script_error = ERR_INVALID_FILE;
+				} else if (st.size > server_info.max_defaults_length) {
+					init_script_error = ERR_FILE_TOO_BIG;
+				} else {
```
openwrt-ai, line 661: Switching from the ucode plugin to `fs.read()` silently re-breaks empty init scripts, which @efahl asked to keep working ([r3468723253](https://github.com/openwrt/luci/pull/8703#discussion_r3468723253)) and which the dropped `get_init_script` handled explicitly via its `st.size <= 0 → ERR_OK` branch.

rpcd's `file read` bails out on a zero-byte file: after `if (s.st_size == 0) s.st_size = RPC_FILE_MIN_SIZE;` it does [`if ((len = read(fd, wbuf, s.st_size)) <= 0) { rv = UBUS_STATUS_NO_DATA; goto out; }`](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L465-L469). `UBUS_STATUS_NO_DATA` (5) makes [`fs.read()`'s `handleRpcReply`](https://github.com/openwrt/luci/blob/adeb73a839ebfdd46bf1c25138f193bd37f09a5b/modules/luci-base/htdocs/luci-static/resources/fs.js#L91-L95) throw `NoDataError`, the `.catch(() => null)` on the next line swallows it, and the user gets *"File … is invalid and won't be included …"* for a perfectly valid empty placeholder script.

Short-circuit the read when the file is empty:

```suggestion
				} else if (st.size > server_info.max_defaults_length) {
					init_script_error = ERR_FILE_TOO_BIG;
				} else if (st.size === 0) {
					init_script_error = ERR_OK;
				} else {
```

---

openwrt-ai (commented):
Reviewed the force-pushed head (`adeb73a`, replacing the previously-reviewed `8a8dfe0`).

The delta drops the `luci.attendedsysupgrade` rpcd ucode plugin entirely and reads the init script from the frontend with `fs.stat()` / `fs.read()`, scoping access through new path entries in the ACL's `read.file` section plus a client-side `startsWith()` allowlist. That's a good answer to the arbitrary-root-read concern from my last review — the path ACL is enforced server-side by rpcd (`rpc_canonicalize_path()` folds `..` before the check, and `rpc_check_symlink_access()` re-checks the `realpath()`-resolved target), so `/etc/uci-defaults/../../etc/shadow` and symlink-out tricks are both rejected regardless of what the client sends. rpcd's ACL matching uses `fnmatch()` without `FNM_PATHNAME`, so `/root/*` does cover owut's documented `/root/data/my-init-script.sh`. The other three items from my last review (dead `endpoint`/`request_url`, the parenthetical period, `switch (`) are all fixed, `luci-base` already pulls in `rpcd-mod-file` so no Makefile change is needed, and the commit message matches the diff.

Two things to look at before merge, both inline: the ACL grants the file paths but not the ubus `file.read` method, and empty init scripts now regress to an "invalid file" warning.

---

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json
firlin123: Fixed in the latest forced push.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: Fixed in the latest forced push.

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json
```diff
@@ -13,10 +13,17 @@
 				],
 				"uci": [
 					"get"
+				],
+				"file": [
+					"list",
+					"read"
 				]
```
openwrt-ai, line 20: This grants `read` but not `stat`, and `stat` is what the code calls first: [`fs.stat(init_script_path)` at overview.js:654](``https://github.com/openwrt/luci/blob/e6ef0faa751b925460b901f4a9616446b247e8a1/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L654)`` maps to the ubus `file` object's [`stat` method](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L1231), which is distinct from `list`/`read` and is checked by uhttpd — [`uh_ubus_allowed(sid, "file", "stat")`](https://github.com/openwrt/uhttpd/blob/f682ccabee9ed5f39dd3b0aaf96228975eeb478d/ubus.c#L279-L296) — before the request ever reaches rpcd.

Nothing in this app's dependency chain grants it: `luci-base` grants only [`"file": [ "list" ]`](https://github.com/openwrt/luci/blob/e6ef0faa751b925460b901f4a9616446b247e8a1/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json#L18-L21). The sole in-tree grant is `luci-mod-network-diagnostics`' [`"file": [ "exec", "stat" ]`](``https://github.com/openwrt/luci/blob/e6ef0faa751b925460b901f4a9616446b247e8a1/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json#L81-L83)`` — an unrelated package this app shouldn't have to rely on (the same accident that makes the pre-existing `fs.stat('/sys/firmware/efi')` on line 825 look like it works). Without it `fs.stat()` rejects with `Access denied`, the `.catch(() => null)` on the next line turns that into `st === null`, and the user gets *"File %s is invalid and won't be included…"* for a perfectly good script.

`list` on the other hand is dead here — nothing calls `fs.list()`, and `luci-base` already grants both the method and the `"/*": [ "list" ]` path scope.

```suggestion
				"file": [
					"read",
					"stat"
				]
```

No change is needed in the `file` path block below: [`rpc_file_stat()` checks the path with the `"list"` permission](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L733), not `"read"`, and `luci-base`'s `"/*": [ "list" ]` already covers that.

---

openwrt-ai (commented):
Re-reviewed the force-pushed head (`e6ef0fa`, replacing the previously-reviewed `adeb73a`).

The delta is exactly the two items from my last review. The empty-file short-circuit (`st.size == 0` → `ERR_OK`) now sits between the size-limit check and the `fs.read()`, so an empty placeholder script no longer hits rpcd's `UBUS_STATUS_NO_DATA` path and is sent as an empty `defaults` string instead of being reported as invalid — that restores the behaviour @efahl asked for in [r3468723253](https://github.com/openwrt/luci/pull/8703#discussion_r3468723253). `==` rather than `===` is fine here: there's no `eqeqeq` rule in `eslint.config.mjs` and the surrounding file already mixes both (`candidate[0] == version`). All three FormalityCheck runs pass on this head, and the commit message still matches the diff.

The ACL half is not quite complete though — one inline note: the new `read.ubus.file` grant covers `read` but not `stat`, which is the method the code calls first.

---

applications/luci-app-attendedsysupgrade/root/usr/share/rpcd/acl.d/luci-app-attendedsysupgrade.json
firlin123: Fixed in the latest forced push.

openwrt-ai (commented):
Re-reviewed the force-pushed head (`7053add`, replacing the previously-reviewed `e6ef0fa`); no new issues found.

The delta is exactly the one item from my last review: `read.ubus.file` is now `["read", "stat"]` instead of `["list", "read"]`. That's the right shape:

- `stat` is the method `fs.stat(init_script_path)` invokes, and uhttpd gates it by name via [`uh_ubus_allowed(sid, "file", "stat")`](https://github.com/openwrt/uhttpd/blob/f682ccabee9ed5f39dd3b0aaf96228975eeb478d/ubus.c#L279-L296) before the call reaches rpcd, so the grant was genuinely required.
- Dropping `list` is correct — nothing under `luci-app-attendedsysupgrade` calls `fs.list()` (only [`fs.stat` at overview.js:654](``https://github.com/openwrt/luci/blob/7053add6e413563eb0e995cfde227cb819898651/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L654),`` [`fs.read` at :662](``https://github.com/openwrt/luci/blob/7053add6e413563eb0e995cfde227cb819898651/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L662)`` and the pre-existing [`fs.stat('/sys/firmware/efi')` at :825](``https://github.com/openwrt/luci/blob/7053add6e413563eb0e995cfde227cb819898651/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L825)),`` and `luci-base` grants the `list` method anyway.
- The `read.file` path block needs no matching change, as noted last time: [`rpc_file_stat()` checks the path with `"list"`](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L734) (covered by `luci-base`'s `"/*": [ "list" ]`), while [`rpc_file_read()` checks `"read"`](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L431) — which the three `uci-defaults`/`/root` prefixes grant.

The squashed commit message matches the diff, and all three FormalityCheck runs pass on this head.

---

efahl:
My concern at this point is that this is just way too much code churn for such a trivial feature.  The code shouldn't need hundreds of lines to just either read the file if possible, or produce some error if it exists and can't be read.

I'd rather we did no config editing, use the same "advanced feature" rational that was applied in https://github.com/openwrt/luci/commit/42d32dff7f718d27ae7ccc65ebc83d001ebf14e3

firlin123:
> My concern at this point is that this is just way too much code churn for such a trivial feature. The code shouldn't need hundreds of lines to just either read the file if possible, or produce some error if it exists and can't be read.
> 
> I'd rather we did no config editing, use the same "advanced feature" rational that was applied in [42d32df](https://github.com/openwrt/luci/commit/42d32dff7f718d27ae7ccc65ebc83d001ebf14e3)

So just remove all changes to configuration.js? Most of the other changes is just me un-deep-nesting the handleCheck(). And there only tabs change. The actual code stays the same.

efahl:
> un-deep-nesting the handleCheck

That all should be another PR; anything unrelated to the specific feature in the title is incidental and needs to be addressed with its own changes.

firlin123:
Reverted the config page changes and undid the handleCheck refactor. The PR is now strictly focused on adding the uci-defaults script support.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -667,6 +788,11 @@ return view.extend({
 									class: 'btn cbi-button cbi-button-positive important',
 									click: ui.createHandlerFn(this, function () {
 										map.save().then(() => {
+											if (init_script_error === ERR_OK && mapdata.request.init_script_checked === '1') {
+												firmware.defaults = init_script_contents;
+											} else {
+												delete firmware.defaults;
+											}
```
openwrt-ai: Assigning to `firmware` rather than to the local `content` object leaks the init script into the public bug-report blob.

`firmware` is built once in [`load()` at overview.js:757-770](``https://github.com/openwrt/luci/blob/a71c94f385ed9add50544876dd1722f8af80e246/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L757-L770)`` and is then threaded through every later call, including [`handleError()`, which spreads it into `request_data` at overview.js:329-335](``https://github.com/openwrt/luci/blob/a71c94f385ed9add50544876dd1722f8af80e246/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L329-L335).`` That object is `JSON.stringify`'d into `request_str`, rendered in the "Error building the firmware image" modal under [_"report all of the information below"_ (line 343)](``https://github.com/openwrt/luci/blob/a71c94f385ed9add50544876dd1722f8af80e246/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L343),`` and copied to the clipboard as [`detailsBlock('Request Data', request_str, true)`](``https://github.com/openwrt/luci/blob/a71c94f385ed9add50544876dd1722f8af80e246/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L355)`` for pasting into the support forum thread.

So on any ASU build failure the user is now invited to paste the verbatim contents of their uci-defaults script into a public thread — and that is precisely the file where root password hashes, WiFi PSKs and WireGuard private keys live. The same applies to the rebuilder polls, which get `firmware` via [`handleRebuilder()`](``https://github.com/openwrt/luci/blob/a71c94f385ed9add50544876dd1722f8af80e246/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L473-L491).``

The previously-reviewed head (`7053add`) did not have this problem: `defaults` was set only on the local `content` literal, which `handleError()` never sees. Restoring that shape keeps the script out of the report — drop these five lines and put the field back on the `content` object a few lines below (802-807):

```js
const content = {
    ...firmware,
    packages: packages,
    version: mapdata.request.version,
    profile: mapdata.request.profile,
    defaults: (init_script_error === ERR_OK && mapdata.request.init_script_checked === '1')
        ? init_script_contents
        : undefined
};
```

(`JSON.stringify` drops `undefined` values, so the `defaults` key is still omitted from the request when unused, exactly as `delete` achieves here.)

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -605,13 +637,50 @@ return view.extend({
 			}
 
 			if (candidates.length) {
+				const server_info = results[0]?.ok ? results[0].json().server : null;
+				let init_script_error = ERR_NO_INIT;
+				let init_script_path = uci.get_first('attendedsysupgrade', 'owut', 'init_script') || '';
```
openwrt-ai: nit: two small things on this line, both fallout from reverting `configuration.js`.

**PR body is now stale.** With the `owut` section gone from `configuration.js`, this UCI read is the only reference to `init_script` in the tree — the option is CLI-only now (which is the `42d32df` "advanced feature" rationale @efahl asked for). The PR description still leads with a *"Configuration tab screenshot"* showing an **Init script** input field that this revision no longer adds, which will mislead anyone reading the PR before the diff. Worth dropping that screenshot and saying explicitly that the path must be set via `uci set attendedsysupgrade.owut.init_script=…` (or is documented at owut's [uci-defaults page](https://openwrt.org/docs/guide-user/installation/sysupgrade.owut#using_a_uci-defaults_script)).

**Accessor inconsistency.** This uses `uci.get_first(…, 'owut', …)` (matches by section *type*), while the pre-existing read of the same section a hundred lines below uses `uci.get(…, 'owut', …)` (matches by section *name*) — [`rootfs_size_mb` at overview.js:769](``https://github.com/openwrt/luci/blob/a71c94f385ed9add50544876dd1722f8af80e246/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L769).`` Both resolve against owut's shipped `config owut 'owut'`, so this isn't a bug, but two different accessors for the same section in one file is confusing. Any preference which one wins?

---

openwrt-ai (commented):
Re-reviewed the force-pushed head (`a71c94f`, replacing the previously-reviewed `7053add`).

Two deltas since my last review. First, `configuration.js` is reverted to base master byte-for-byte and the `handleCheck` de-nesting is undone (back to `Promise.all(...).then(...)` and a non-`async` `handleCheck`, matching master), so the diff is now down to `overview.js` + the ACL file — that's what @efahl asked for in [#8703 (comment)](https://github.com/openwrt/luci/pull/8703#issuecomment-5182006379). Second, the new `resolve_path()` helper is applied *before* the `startsWith()` allowlist, which is the right ordering; it's belt-and-braces anyway since rpcd folds `..` in `rpc_canonicalize_path()` and re-checks `realpath()`-resolved symlink targets before the ACL decision. The ACL file is unchanged from the head I signed off on last time, the single squashed commit message matches the diff, and all three FormalityCheck runs pass on this head.

One thing to fix before merge, inline: the `defaults` payload moved from the local `content` object onto the shared `firmware` object, which puts the init script contents into the clipboard blob the error modal asks users to paste into the public support thread. Plus one nit.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: I mean, [the docs](https://openwrt.org/docs/guide-user/installation/sysupgrade.owut#using_a_uci-defaults_script) say you shouldn't put any sensitive info in there to begin with, but I get that people do it anyway. Fixed in latest force push.

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
firlin123: Removed config page screenshot. Switched to `uci.get(...)`.

openwrt-ai (commented):
Re-reviewed the force-pushed head (`171a671`, replacing the previously-reviewed `a71c94f`); no new issues found.

The delta is exactly the two fixes from my last review, both correct:

- **`defaults` no longer leaks into the bug report.** The assignment moved off `firmware` and onto the local `content` literal, and the `delete firmware.defaults` counterpart is gone with it (it was only needed because of the mutation). [`handleError()` builds `request_data` from `{...data, ...firmware}`](``https://github.com/openwrt/luci/blob/171a67168a24f0773e76df9f448078b7f52b212c/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L327-L336),`` which `content` never feeds into, so the script contents can no longer reach the "report all of the information below" blob or the clipboard text. Omitting the key entirely when the flag is off keeps `defaults` out of the request body as before. The rebuilder path still receives it via [`handleRebuilder(content, …)`](``https://github.com/openwrt/luci/blob/171a67168a24f0773e76df9f448078b7f52b212c/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L474-L491),`` which is what you want — the rebuild has to use the same `defaults` for the hashes to match.
- **Accessor consistency.** `uci.get_first(…, 'owut', …)` → `uci.get(…, 'owut', …)`, now matching the pre-existing [`rootfs_size_mb` read at overview.js:769](``https://github.com/openwrt/luci/blob/171a67168a24f0773e76df9f448078b7f52b212c/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L769).`` owut ships a named `config owut 'owut'` section, so this still resolves.

The squashed commit message matches the diff, and all three FormalityCheck runs pass on this head. The config-page screenshot has been dropped from the PR body, so the description now matches the CLI-only `attendedsysupgrade.owut.init_script` shape of this revision.

---

Issue #7972, opened 2025-09-30
<luci-app-xfrpc>: <generate wrong xfrpc.ini file >


maxmib:
### Is there an existing issue for this?

- [x] I have searched among all existing issues (including closed issues)

### Actual behaviour

xfrpc failed to start 

/etc/config/xfrpc  is 
config xfrpc 'common'
	option enabled '1'
	option loglevel '7'
	option server_addr 'xx.xx.xx.x'
	option server_port '123'
	option token 'xxxxxxxxxxx'

config xfrpc 'web'
	option type 'tcp'
	option local_ip '127.0.0.1'
	option local_port '80'
	option remote_port 'cccc'

config xfrpc 'AMT'
	option type 'tcp'
	option local_ip '192.168.1.1'
	option local_port 'ccccc'

xfrpc config file above generate wrong xfrpc.ini file in /var/etc/

 /var/etc/xfrpc.ini
[common]
server_addr = xx.xx.xx.x
server_port = xxx
token = xxxxxxxxxxx
[web]
server_addr = xx.xx.xx.x
server_port = xxx
token = xxxxxxxxxxx
[AMT]
server_addr = xx.xx.xx.x
server_port = xxx
token = xxxxxxxxxxx

### Expected behaviour

generate right xfrpc.ini to make xfrpc work 

### Steps to reproduce

go to luci-app-xfrpc (reverse proxy under services) ,change  settings and apply .

### Additional Information

```shell
Hostname	    HC5661
Model	            HiWiFi HC5661
Architecture	    MediaTek MT7620A ver:2 eco:6
Target Platform	    ramips/mt7620
Firmware Version    OpenWrt SNAPSHOT r31209-0147d213ff / LuCI Master 25.271.67054~84ad0f3
Kernel Version	    6.12.49
```

### What browsers do you see the problem on?

Chrome

### Relevant log output

```shell

```

systemcrash:
And what generates the ini file?

maxmib:
> And what generates the ini file?

/etc/init.d/xfrpc generate the ini file,it reads /etc/config/xfrpc then generate /var/etc/xfrpc.ini ,and start xfrpc

wang20150419:
[...]

PR #8704, opened 2026-06-14
luci-app-xfrpc: add comment explaining GridSection type behavior

ximalu:
## Description

This is a documentation-only change to clarify that the GridSection type `'xfrpc'` is intentional.

## Background

Issue #7972 reported that `luci-app-xfrpc` generates configs that prevent xfrpc from starting.
Investigation showed the root cause is a section type mismatch:

- LuCI creates proxy sections with type `xfrpc` (via GridSection)
- The init script (`xfrpc.init`) iterates sections by type (`tcp`, `http`, `https`, `socks5`)
- Since no handler iterates type `xfrpc`, proxy configs were lost in the generated `xfrpc.ini`

## Fix Approach

The functional fix is in `openwrt/packages` (`net/xfrpc/files/xfrpc.init`), which now has a fallback handler that reads the `type` option from `xfrpc`-typed sections and dispatches correctly. This LuCI-side change adds a comment documenting the contract between LuCI and the init script.

## Testing

Verified on ImmortalWrt 24.10.4 (Xiaomi WR30U). TCP proxy sections created via LuCI now generate correct `xfrpc.ini` files and the xfrpc service starts successfully.

### DIFF
--- a/applications/luci-app-xfrpc/htdocs/luci-static/resources/view/xfrpc.js
+++ b/applications/luci-app-xfrpc/htdocs/luci-static/resources/view/xfrpc.js
@@ -104,6 +104,17 @@ return view.extend({
 		o.value(2, _('Alert'))
 		o.value(1, _('Emergency'))
 
+		/*
+		 * Proxy sections grid.
+		 *
+		 * GridSection type is "xfrpc". The init script (xfrpc.init) generates the
+		 * xfrpc.ini file from all UCI sections. It first processes type-specific
+		 * sections (tcp, http, https, socks5) via config_foreach, then falls back
+		 * to handling "xfrpc"-typed sections by reading the "type" option.
+		 *
+		 * This ensures proxies added here via LuCI work correctly regardless of
+		 * whether the user's init script has the fallback fix applied.
+		 */
 		s = m.section(form.GridSection, 'xfrpc', _('Proxy Settings'));
 		s.addremove = true;
 		s.filter = function(s) { return s !== 'common'; };
--- a/applications/luci-app-xfrpc/htdocs/luci-static/resources/view/xfrpc.js
+++ b/applications/luci-app-xfrpc/htdocs/luci-static/resources/view/xfrpc.js
@@ -81,17 +81,12 @@ return view.extend({
 			sent to the server. The server must have a matching token for authorization to succeed.'), 
 			_('By default, this value is "".')));
 
-		o = s.taboption('init', form.SectionValue, 'init', form.TypedSection, 
-			'xfrp', _('Startup Settings'));
-		s = o.subsection;
-		s.anonymous = true;
-		s.dynamic = true;
-
-		o = s.option(form.Flag, 'disabled', _('Disabled xfrpc service'));
+		o = s.taboption('init', form.Flag, 'enabled', _('Enable xfrpc service'));
 		o.datatype = 'bool';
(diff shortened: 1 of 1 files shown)


applications/luci-app-xfrpc/htdocs/luci-static/resources/view/xfrpc.js
```diff
@@ -81,17 +81,12 @@ return view.extend({
...
-			'xfrp', _('Startup Settings'));
-		s = o.subsection;
-		s.anonymous = true;
-		s.dynamic = true;
-
-		o = s.option(form.Flag, 'disabled', _('Disabled xfrpc service'));
+		o = s.taboption('init', form.Flag, 'enabled', _('Enable xfrpc service'));
 		o.datatype = 'bool';
-		o.optional = true;
+		o.default = '1';
```
openwrt-ai, line 86: The frontend defaults `enabled` to `1`, but the backend defaults it to `0`. `start_service()` validates [`'enabled:bool:0'`](https://github.com/openwrt/packages/blob/44da4494fb0595be92f0f73794d515c32b2bf707/net/xfrpc/files/xfrpc.init#L191) and the shipped [`xfrpc.conf`](https://github.com/openwrt/packages/blob/44da4494fb0595be92f0f73794d515c32b2bf707/net/xfrpc/files/xfrpc.conf#L2) sets `option enabled 0`. With `rmempty = false` the value is always written on save, so post-save behavior is fine, but the pre-save UI default disagrees with the daemon default (service off). Note the commit message claims "the init script reads 'enabled' (default 1)" — the init script default is actually `0`. Is defaulting the LuCI checkbox to on intended?

---

applications/luci-app-xfrpc/htdocs/luci-static/resources/view/xfrpc.js
```diff
@@ -81,17 +81,12 @@ return view.extend({
...
 			_('By default, this value is "".')));
 
-		o = s.taboption('init', form.SectionValue, 'init', form.TypedSection, 
-			'xfrp', _('Startup Settings'));
-		s = o.subsection;
-		s.anonymous = true;
-		s.dynamic = true;
-
-		o = s.option(form.Flag, 'disabled', _('Disabled xfrpc service'));
+		o = s.taboption('init', form.Flag, 'enabled', _('Enable xfrpc service'));
```
openwrt-ai, line 84: nit: The PR title ("add comment explaining GridSection type behavior") and body ("This is a documentation-only change") describe only the first commit (efd9ee5). This commit (3816f22) is a functional change: it removes the `SectionValue`+`TypedSection` wrapper, rebinds `loglevel` onto the `common` section, and renames the `disabled` flag to `enabled`. Please update the PR description so it reflects the section restructuring and the option rename, not just the documentation comment.

---

openwrt-ai (commented):
## Commit checks
- `3816f22` "luci-app-xfrpc: fix startup settings section and field naming" — The body states "The init script reads 'enabled' (default 1)", but the backend default is `0`, not `1`: `start_service()` validates `'enabled:bool:0'` and the shipped `xfrpc.conf` ships `option enabled 0`. The rename `disabled`→`enabled` is correct (the daemon does read `enabled`), but the stated default is wrong and the LuCI `o.default = '1'` consequently diverges from the daemon default. See the inline note.

---

PR #8707, opened 2026-06-15
luci-app-singbox: add new package

DmitriyStroganov:
## What this PR adds

A new LuCI application — `luci-app-singbox` — providing a web interface for managing [sing-box](https://github.com/SagerNet/sing-box) on OpenWrt. sing-box is already in the OpenWrt packages feed since 23.05 but has no LuCI front-end in this repo; users currently either edit JSON by hand or use third-party feeds.

This package focuses on **VLESS + REALITY** — the most DPI-resistant proxy transport available today — combined with selective traffic splitting, GeoSite presets, multi-server failover, and real-time monitoring via sing-box's built-in Clash API.

## Why this is useful

- REALITY disguises TLS as a legitimate connection to a popular website (SNI masquerade) and is indistinguishable from normal HTTPS even under active probing. There is no LuCI app in this repo that exposes it.
- Existing LuCI proxy apps (luci-app-shadowsocks, luci-app-v2ray, luci-app-vmess) target older protocols that don't survive modern DPI.
- The dashboard (traffic, active connections, per-server latency) brings feature parity with consumer-friendly proxies on platforms that already have mature GUIs.

## Architecture

UCI is the single source of truth. `generate-config.sh` reads `/etc/config/singbox`, renders sing-box JSON, validates it with `sing-box check`, and only then atomically replaces `/etc/sing-box/config.json`. A typo or missing field never reaches the running daemon.

```
LuCI JS views  ──fs.exec──►  shell scripts  ──UCI→JSON──►  /etc/sing-box/config.json
                                  │                              │
                                  └─ status.sh (Clash API)       ▼
                                  └─ update-subscription.sh   sing-box ──► Clash API :9090
                                  └─ tail-log.sh (logread)    (procd-managed)
```

## Notable implementation choices

- **Validation gate.** Every config runs through `sing-box check` before being deployed — typo in a REALITY key never takes the router offline.
- **Modern `action: reject`** instead of the deprecated `{"type":"block"}` outbound — eliminates the "operation not permitted" warnings from sing-box 1.10+ when blocking already-established sockets.
- **Streaming-endpoint-safe Clash API client.** `/traffic` and `/connections` are SSE-style streams; BusyBox `wget` would hang forever reading them. The status helper forks wget to a temp file and kills it the moment any data is captured.
- **procd init with USB-binary wait.** Optional `wait_for_usb` UCI flag for routers that keep `sing-box` on external storage to save overlay; auto-detects dangling `/usr/bin/sing-box` symlinks.
- **Syslog-based logging.** sing-box writes to stderr; procd relays to syslog; the Logs tab reads via `logread | grep sing-box`. logd's ring buffer handles rotation itself — no SIGHUP / no risk of crash-looping sing-box by signalling a "reload" while TUN is held by the dying process (this was a real production failure mode in earlier iterations of the package).
- **rpcd ACL scoped to the minimum** the front-end actually needs (no blanket `/bin/sh` exec).

## Tested

OpenWrt 24.10.2, mipsel_24kc, sing-box 1.13.13. End-to-end on a real router with real REALITY upstreams:
- 11 tagged releases over one day of iterative dogfooding (see https://github.com/DmitriyStroganov/LAS/releases for the development history).
- All LuCI views (Overview / Servers / Routing / Logs / Settings) verified functional.
- Generated config byte-identical to a hand-tuned raw sing-box config with the same rules.

## Out of scope for this PR (intentionally)

The development repository at https://github.com/DmitriyStroganov/LAS also ships a standalone `install.sh` wrapper and a tar.gz `.ipk` builder for users who want to test the package without adding the feed — those are user-facing conveniences that don't belong in the upstream package and are not included in this PR.

## Screenshots

Five screenshots (Overview / Servers / Routing / Logs / Settings) are in the development repo: https://github.com/DmitriyStroganov/LAS#screenshots

## License

GPL-2.0-or-later, matching the rest of luci-app-*.

---

Signed-off-by: Dmitriy Stroganov <strodi@yandex.ru>

### DIFF
--- a/applications/luci-app-singbox/Makefile
+++ b/applications/luci-app-singbox/Makefile
@@ -1,0 +1,73 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-app-singbox
+PKG_VERSION:=1.1.1
+PKG_RELEASE:=1
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_MAINTAINER:=Dmitriy Stroganov <strodi@yandex.ru>
+
+LUCI_TITLE:=sing-box VPN — VLESS+REALITY management interface
+LUCI_DESCRIPTION:=Web interface for sing-box proxy with VLESS+REALITY support, \
+	traffic splitting, geosite rules, subscription import, and real-time \
+	monitoring via Clash API.
+
+define Package/$(PKG_NAME)
+  SECTION:=luci
+  CATEGORY:=LuCI
+  SUBMENU:=3. Applications
+  TITLE:=$(LUCI_TITLE)
+  DEPENDS:=+sing-box +luci-base +jq
+  PKGARCH:=all
+endef
+
+define Build/Compile
+endef
+
+define Package/$(PKG_NAME)/install
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_DIR) $(1)/usr/share/singbox
+	$(INSTALL_DIR) $(1)/usr/share/rpcd/acl.d
+	$(INSTALL_DIR) $(1)/usr/share/luci/menu.d
+	$(INSTALL_DIR) $(1)/usr/share/ucode/rpc
+	$(INSTALL_DIR) $(1)/www/luci-static/resources/view/singbox
+
+	$(INSTALL_DATA) ./files/etc/config/singbox                   $(1)/etc/config/singbox
+	$(INSTALL_BIN)  ./files/etc/init.d/sing-box                  $(1)/etc/init.d/sing-box
+	$(INSTALL_BIN)  ./files/etc/uci-defaults/singbox-migrate     $(1)/etc/uci-defaults/singbox-migrate
+	$(INSTALL_BIN)  ./files/usr/share/singbox/generate-config.sh $(1)/usr/share/singbox/generate-config.sh
+	$(INSTALL_BIN)  ./files/usr/share/singbox/status.sh          $(1)/usr/share/singbox/status.sh
+	$(INSTALL_BIN)  ./files/usr/share/singbox/update-subscription.sh $(1)/usr/share/singbox/update-subscription.sh
+	$(INSTALL_BIN)  ./files/usr/share/singbox/tail-log.sh        $(1)/usr/share/singbox/tail-log.sh
+	$(INSTALL_DATA) ./files/usr/share/rpcd/acl.d/luci-app-singbox.json   $(1)/usr/share/rpcd/acl.d/luci-app-singbox.json
+	$(INSTALL_DATA) ./files/usr/share/luci/menu.d/luci-app-singbox.json  $(1)/usr/share/luci/menu.d/luci-app-singbox.json
+	$(INSTALL_BIN)  ./files/usr/share/ucode/rpc/singbox.uc       $(1)/usr/share/ucode/rpc/singbox.uc
+
+	$(INSTALL_DATA) ./files/www/luci-static/resources/view/singbox/overview.js $(1)/www/luci-static/resources/view/singbox/overview.js
+	$(INSTALL_DATA) ./files/www/luci-static/resources/view/singbox/servers.js  $(1)/www/luci-static/resources/view/singbox/servers.js
+	$(INSTALL_DATA) ./files/www/luci-static/resources/view/singbox/routing.js  $(1)/www/luci-static/resources/view/singbox/routing.js
+	$(INSTALL_DATA) ./files/www/luci-static/resources/view/singbox/logs.js     $(1)/www/luci-static/resources/view/singbox/logs.js
+	$(INSTALL_DATA) ./files/www/luci-static/resources/view/singbox/settings.js $(1)/www/luci-static/resources/view/singbox/settings.js
+endef
+
+# Files preserved across upgrade — opkg asks before overwriting if local copy differs.
+define Package/$(PKG_NAME)/conffiles
+/etc/config/singbox
+endef
+
+define Package/$(PKG_NAME)/postinst
+#!/bin/sh
+[ -n "$$IPKG_INSTROOT" ] && exit 0
+/etc/init.d/rpcd restart >/dev/null 2>&1 || true
+exit 0
+endef
+
+define Package/$(PKG_NAME)/postrm
+#!/bin/sh
+[ -n "$$IPKG_INSTROOT" ] && exit 0
+/etc/init.d/rpcd restart >/dev/null 2>&1 || true
+exit 0
+endef
+
+$(eval $(call BuildPackage,$(PKG_NAME)))
--- a/applications/luci-app-singbox/files/etc/uci-defaults/singbox-migrate
+++ b/applications/luci-app-singbox/files/etc/uci-defaults/singbox-migrate
@@ -1,0 +1,83 @@
+#!/bin/sh
+# singbox-migrate — one-shot UCI defaults script run on first boot after
+# install. If /etc/sing-box/config.json exists from a pre-luci-app-singbox
+# setup, parse the first REALITY outbound out of it and create a matching
+# `config server` section in /etc/config/singbox, then enable the service.
+#
+# Uses a marker file rather than testing for /etc/config/singbox (which the
+# package always ships as a conffile, so its presence is not a reliable
+# signal of "already migrated").
+
+MARKER="/etc/.singbox.migrated"
+
+[ -f /etc/sing-box/config.json ] || exit 0
+[ -f "$MARKER" ] && exit 0
+
+CONFIG="/etc/sing-box/config.json"
+
+exists() { [ -n "$1" ]; }
+
+get_json_val() {
+	local key="$1"
+	grep -o "\"$key\"[^,]*" "$CONFIG" 2>/dev/null | head -1 | sed 's/.*: *"//' | sed 's/".*//'
+}
+
+get_json_num() {
+	local key="$1"
+	grep -o "\"$key\"[^,]*" "$CONFIG" 2>/dev/null | head -1 | sed 's/.*: *//' | sed 's/,.*//'
+}
+
+uci -q batch << EOF
+set singbox.general.enabled=1
+set singbox.general.stack=gvisor
+set singbox.general.tun_address=172.19.0.1/30
+set singbox.general.tun_mtu=1400
+set singbox.general.auto_route=1
+set singbox.general.strict_route=1
+set singbox.general.clash_api=1
+set singbox.general.clash_api_port=9090
+set singbox.general.final_outbound=direct
+set singbox.urltest.enabled=1
+set singbox.urltest.interval=3m
+set singbox.urltest.tolerance=50
+set singbox.dns.direct_server=auto
+set singbox.dns.tunnel_server=8.8.8.8
+set singbox.dns.strategy=prefer_ipv4
+EOF
+
+SERVERS=$(grep -o '"tag":"[^"]*xray"' "$CONFIG" 2>/dev/null | sed 's/"tag":"//;s/"//')
+for TAG in $SERVERS; do
+	SERVER=$(grep -A20 "\"tag\":\"$TAG\"" "$CONFIG" | grep '"server"' | head -1 | sed 's/.*: *"//;s/".*')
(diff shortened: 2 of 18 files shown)


applications/luci-app-singbox/files/usr/share/rpcd/acl.d/luci-app-singbox.json
```diff
@@ -0,0 +1,30 @@
...
+		"description": "sing-box VPN management",
+		"read": {
+			"ubus": {
+				"file": ["list", "read", "exec"],
+				"uci": ["get"]
+			},
+			"uci": ["singbox"],
+			"file": {
+				"/usr/share/singbox/*": ["read", "exec"],
+				"/etc/init.d/sing-box": ["exec"],
```
openwrt-ai: The `read` ACL grants `exec` on every script under `/usr/share/singbox/*` and on `/etc/init.d/sing-box`. That is not read-only: a session holding only the read scope can run `update-subscription.sh <url>` (fetches an arbitrary URL and writes UCI server sections, per [`update-subscription.sh`](``https://github.com/openwrt/luci/blob/93a7e27bd6ad88f95c58d908d50dfe57f3c9a3f4/applications/luci-app-singbox/files/usr/share/singbox/update-subscription.sh#L48-L51))`` and `/etc/init.d/sing-box start|stop|restart`. The PR body states the ACL is "scoped to the minimum the front-end actually needs (no blanket `/bin/sh` exec)", but state-changing actions exposed under `read` undercuts that. Consider moving the mutating execs (init.d control, `update-subscription.sh`, `generate-config.sh`) into the `write` block and keeping only the read-only `status.sh`/`tail-log.sh` execs under `read`.

---

applications/luci-app-singbox/files/usr/share/ucode/rpc/singbox.uc
```diff
@@ -0,0 +1,121 @@
...
+			api = req[0] || api;
+			srv = req[1];
+		} else if (is_string(req)) {
+			srv = name || req;
+		}
+
+		if (!srv)
+			return { code: 1, stderr: 'missing server name' };
+
+		return run(STATUS_BIN, [ shell_safe(api), 'test', srv ]);
```
openwrt-ai: The `api` argument is passed through `shell_safe()` everywhere, but the server name `srv` is passed raw here. It comes straight from the browser-supplied request (`req.name`/`req.tag`/`req[1]`), so the sanitation is inconsistent with the other methods. It happens to be URL-encoded downstream in `status.sh` before it reaches a command line, but routing it through `shell_safe()` like `api` keeps the trust boundary uniform and avoids relying on that downstream detail.

```suggestion
		return run(STATUS_BIN, [ shell_safe(api), 'test', shell_safe(srv) ]);
```

---

applications/luci-app-singbox/files/usr/share/ucode/rpc/singbox.uc
```diff
@@ -0,0 +1,121 @@
...
+	logs: function(req) {
+		let opts = req;
+		if (!is_object(opts))
+			opts = { lines: req };
+
+		const lines = clamp_int(opts.lines, 1, 2000, 200);
+		const level = opts.level ? shell_safe(opts.level) : '';
+		const search = opts.search ? shell_safe(opts.search) : '';
+
+		let cmd = 'tail -n ' + lines + ' ' + LOG_FILE;
```
openwrt-ai: This `logs` method tails `/tmp/sing-box.log`, but the rest of the package deliberately moved logging to syslog: the init script no longer writes a log file (the `log_output` option is accepted "for backward compatibility but no longer used", see [`init.d/sing-box`](https://github.com/openwrt/luci/blob/93a7e27bd6ad88f95c58d908d50dfe57f3c9a3f4/applications/luci-app-singbox/files/etc/init.d/sing-box#L77-L80)), and the Logs view uses [`tail-log.sh`](https://github.com/openwrt/luci/blob/93a7e27bd6ad88f95c58d908d50dfe57f3c9a3f4/applications/luci-app-singbox/files/usr/share/singbox/tail-log.sh) which reads `logread`. So if any view migrates to this RPC surface as the file comment intends, `logs`/`clearLogs` would operate on a file that is never written. Point this method at `tail-log.sh` (logread) instead of `/tmp/sing-box.log`.

---

applications/luci-app-singbox/files/usr/share/singbox/status.sh
```diff
@@ -0,0 +1,248 @@
+#!/bin/sh
+
+#!/bin/sh
```
openwrt-ai, line 3: nit: duplicated shebang — the file has `#!/bin/sh` on line 1 and again on line 3 with a blank line between.

```suggestion
#!/bin/sh
# status.sh — read-only bridge between sing-box's Clash API and LuCI.
```

---

applications/luci-app-singbox/files/etc/uci-defaults/singbox-migrate
```diff
@@ -0,0 +1,76 @@
...
+	local key="$1"
+	grep -o "\"$key\"[^,]*" "$CONFIG" 2>/dev/null | head -1 | sed 's/.*: *"//' | sed 's/".*//'
+}
+
+get_json_num() {
+	local key="$1"
+	grep -o "\"$key\"[^,]*" "$CONFIG" 2>/dev/null | head -1 | sed 's/.*: *//' | sed 's/,.*//'
+}
+
+cp /etc/config/singbox /etc/config/singbox.bak 2>/dev/null
```
openwrt-ai: This migration helper is effectively unreachable. The package ships `/etc/config/singbox` (declared as a conffile in the Makefile), so by the time uci-defaults runs, `/etc/config/singbox` always exists and line 9 (`[ -f /etc/config/singbox ] && exit 0`) always exits before any import happens. As a result the "import existing pre-LuCI config" path never runs on a normal install, and the `cp /etc/config/singbox /etc/config/singbox.bak` on line 25 always operates on the just-shipped default (and would never run anyway after the early exit). Either drop this script, or gate the early-exit on a marker other than the conffile's presence if the intent is to merge an existing `/etc/sing-box/config.json` into the default UCI.

---

applications/luci-app-singbox/files/usr/share/singbox/generate-config.sh
```diff
@@ -0,0 +1,420 @@
...
+	config_get short_id "$name" reality_short_id
+	config_get tag "$name" name
+
+	[ "$enabled" = "1" ] || return
+	[ "$srv_type" = "vless" ] || return
+	[ -z "$server" ] && return
+
+	[ "$first_outbound" = "1" ] && first_outbound=0 || printf ','
+
+	printf '{"type":"vless","tag":"%s","server":"%s","server_port":%s,"uuid":"%s"' \
```
openwrt-ai, line 195: UCI string values (`server`, `uuid`, `sni`, `tag`, and the `match`/domain entries in [`collect_route_rules`](``https://github.com/openwrt/luci/blob/93a7e27bd6ad88f95c58d908d50dfe57f3c9a3f4/applications/luci-app-singbox/files/usr/share/singbox/generate-config.sh#L259-L320))`` are interpolated raw into JSON via `printf '"%s"'` with no escaping. A value containing a `"` or backslash (e.g. an imported server tag/fragment) produces invalid JSON. The `sing-box check` gate prevents a broken config from being deployed, so this is a robustness/usability issue rather than a security hole — but it means such a server silently fails to generate any config at all. Consider JSON-escaping these fields (at minimum `"` and `\`) before interpolation.

---

applications/luci-app-singbox/Makefile
```diff
@@ -0,0 +1,73 @@
...
+LUCI_DESCRIPTION:=Web interface for sing-box proxy with VLESS+REALITY support, \
+	traffic splitting, geosite rules, subscription import, and real-time \
+	monitoring via Clash API.
+
+define Package/$(PKG_NAME)
+  SECTION:=luci
+  CATEGORY:=LuCI
+  SUBMENU:=3. Applications
+  TITLE:=$(LUCI_TITLE)
+  DEPENDS:=+sing-box +luci-base +luci-compat +jq
```
openwrt-ai: nit: `+luci-compat` looks unneeded — all five views use only the modern client APIs (`require view/form/fs/ui/uci`) and none of the legacy `cbi`/`luci.compat` map JS that `luci-compat` provides. Is it actually required by something here, or can it be dropped to slim the dependency?

---

openwrt-ai (commented):
Fresh review of the new `luci-app-singbox` package (1 commit, 18 files). The commit message accurately describes the diff, so no commit-check issues. Inline comments cover one ACL/privilege concern (mutating execs exposed under the `read` scope), a couple of cross-component consistency bugs (the `logs` RPC method targets a log file the daemon no longer writes; unescaped JSON interpolation in `generate-config.sh`; an unsanitized exec arg in the ucode RPC), and a few nits (effectively-dead `singbox-migrate` due to the shipped conffile, duplicate shebang, possibly-unneeded `luci-compat` dep). CI is green.

---

DmitriyStroganov:
Thanks for the review — all 7 points addressed in the updated commit:

1. **ACL read/write split** — read scope now only permits `status.sh` and `tail-log.sh`. `generate-config.sh`, `update-subscription.sh`, and `/etc/init.d/sing-box` live under write scope only.

2. **ucode `test()` sanitation** — `srv` now goes through `shell_safe()` like `api`.

3. **ucode `logs()`/`clearLogs()`** — rewrote `logs()` to call `tail-log.sh` (logread-backed, matching the syslog migration in the rest of the package). Removed `clearLogs` entirely — logd's ring buffer is shared with system logs and cannot be partially cleared.

4. **status.sh duplicate shebang** — removed.

5. **singbox-migrate unreachable** — switched to a marker file (`/etc/.singbox.migrated`) instead of testing for the shipped conffile.

6. **JSON escaping in generate-config.sh** — added `json_escape()` that escapes `\` and `"`. Applied to every interpolated string value. Verified with a deliberately malicious `match` value containing an embedded double-quote — config now generates and validates clean.

7. **`+luci-compat` dependency** — dropped. All views use only modern APIs.

applications/luci-app-singbox/files/usr/share/ucode/rpc/singbox.uc
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Exposes sing-box management operations over ubus as
+ * `ubus call rpc.singbox.<method> '{...}'`. The LuCI views currently call
+ * the shell scripts directly via fs.exec(); this module provides a stable
+ * RPC surface that mirrors those scripts so views can migrate to it without
+ * exposing arbitrary shell execution to the browser session.
+ *
+ * All user-supplied arguments (server names, log filters, subscription URLs)
+ * are sanitised through shell_safe() before being forwarded to the backend
+ * scripts — defence in depth even when the backend re-escapes.
```
openwrt-ai: nit: this comment claims subscription URLs are sanitised through `shell_safe()`, but `importSubscription` forwards `url` to `SUB_BIN` unescaped — and that is the right call, since `shell_safe()` strips `?`, `&`, `=`, and `%`, which would corrupt almost any real subscription URL. The args are passed as `exec()` argv (no shell), so there is no injection to defend against here. Suggest narrowing the comment to the args that actually go through `shell_safe()` (server name, log level/search) so it doesn't overstate the guarantee.

---

openwrt-ai (commented):
Re-reviewed the changes since my last pass (single amended commit, `93a7e27`→`b3a3f1b`, 6 files). All seven previously-raised points are addressed cleanly:

- ACL is now split read (`status.sh`/`tail-log.sh` exec only) vs write (mutating execs + config writes).
- `json_escape()` is consistently applied to every leaf string interpolation in `generate-config.sh`; pre-built list fragments carry already-escaped values, so the escaping is complete.
- `logs()` is rewritten onto `tail-log.sh` (logread-backed) and `clearLogs` is removed with no dangling references; `logs.js` passes lowercase levels that `grep -i` matches against the uppercase syslog tokens.
- `test()` now routes `srv` through `shell_safe()`.
- `singbox-migrate` switched to a marker file.

One minor doc/code mismatch noted inline. No commit-check issues — the commit message accurately describes the package. CI shows no failures.

---

openwrt-ai (commented):
Reviewed 1 new change since my last pass (`b3a3f1b`→`c03f7f1`, 1 file): the `singbox.uc` header comment is corrected to state that subscription URLs are passed as `exec()` argv (no shell, no injection surface) rather than claiming they go through `shell_safe()`. This matches the actual code (`run(SUB_BIN, [ url ])`) and resolves the doc/code mismatch noted in my previous review. No new issues found.

---

PR #8710, opened 2026-06-15
luci-app-advanced-reboot: update to 1.1.2-6

stangri:
Maintainer: me
Compile tested: x86_64, Dell EMC Edge620, OpenWrt 25.12.4

Description:
Update to 1.1.2-6. The headline change is a security fix
for GHSA-vj96-f37g-37f6.

- Security (GHSA-vj96-f37g-37f6, high): the previous read ACL granted
  file.exec on /bin/sh (and dd, mount, sed, fw_setenv, etc.), so a user
  delegated read-only access to this app could execute arbitrary
  commands as root. Device-info gathering is now handled server-side in
  the ucode rpcd backend, the broad file-exec grants are removed, and
  the privileged actions(boot_partition, system reboot, poweroff) are
  moved into a dedicated write block.
- New device support: Linksys EA7500v3, Linksys MX6200 (with a device
  helper script), and Zyxel WSM20.
- Tests: added a ucode test suite (error handling, obtain_device_info,
  and boot_partition) with a mock fs layer and a runner.
- Build: dropped the jq/host build dependency. Build/Prepare now merges
  the per-device JSON files into devices.json with printf/cat, so the
  merged file is no longer committed and the separate install hook is
  gone.
- Maintenance: project moved to the mossdef-org org; melmac.ca URLs
  updated to mossdef.org across the Makefile, README, and LuCI view.


openwrt-ai (commented):

---

castillofrancodamian:
I'm not sure if this is the right place, but it should be "Particiones de Linksys EA7500v2".

Captura de pantalla_20260616_133509

stangri:
> I'm not sure if this is the right place, but it should be "Particiones de Linksys EA7500v2".
> 
> Captura de pantalla_20260616_133509

@castillofrancodamian Franco, thanks for your feedback -- I don't have control over translations, they are managed at https://hosted.weblate.org/projects/openwrt/luci/en_devel/ where you can contribute your translations for any luci application and I'd be grateful if you contribute a fix for above.

castillofrancodamian:
@stangri I'm not talking about the translation. I'm talking about how the string is ordered, if that's what it's called. The word "Particiones" should be at the beginning, not at the end like in English. It's similar to the "h ago" case in WireGuard's status.

stangri:
@castillofrancodamian generally speaking, if the text label is properly coded, the order of localizable and non-localizable resources can be changed by the translator. In this specific case, I've updated the resource and after next PR/merge, the order of those words can be changed for localizations.

PR #8715, opened 2026-06-18
luci-app-olsr: fix blank olsr pages

gratux:
### Description
On OpenWrt 25.12.4 some OLSR Status pages (Neighbours, Routes, Topology), as well as the settings page when adding a new interface, are completely blank.
This PR introduces three fixes to restore functionality:
- correctly reference `etx_color` and `snr_colors` from `olsr/common_js` (these functions were previously moved, but their call sites had not been updated to match)
  additionally, some code still referenced `common/common_js`. this is now updated to `olsr/common_js`.
- avoid a shell "too long argument list" error when processing huge olsr-jsoninfo data with jshn.sh (in my case the Topology status page tried to process ~370KB of JSON data). The workaround is to not use jshn.sh and instead escape the string manually.
- When adding a new interface on the settings page it resulted in a blank page with no obvious error. c567bfc removed the call to actually create the config section for the interface. This has now been re-added.

| Page | before | after |
|:----:|:------:|:-----:|
| OLSR/Neighbours | before | after |
| Services/OLSR IPv(4\|6) -> Interfaces -> Add | before2 | after2 |

### Maintainer _(preferred)_

unsure who I should ping here.
last person to touch the Makefile was @systemcrash.

---

## Tested on

**OpenWrt version:** OpenWrt 25.12.4 (r32933-4ccb782af7) 
**LuCI version:** LuCI openwrt-25.12 branch (26.167.08537~9ccd99b) 
**Web browser(s):** LibreWolf 152.0-1 

---

## Checklist

- [x] Incremented :up: any `PKG_VERSION` in the Makefile.

### DIFF
--- a/applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
+++ b/applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
@@ -56,13 +56,13 @@ call)
 		json_get_var v4_port v4_port
 		json_get_var v6_port v6_port
 
-		jsonreq4=$(echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' 2>/dev/null)
-		jsonreq6=$(echo "/${otable}" | nc ::1 "${v6_port}" | sed -n '/^[}{ ]/p' 2>/dev/null)
-
-		json_init
-		json_add_string "jsonreq4" "$jsonreq4"
-		json_add_string "jsonreq6" "$jsonreq6"
-		json_dump
+		{
+			printf '{"jsonreq4":"'
+			echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' | sed 's/\\/\\\\/g; s/"/\\"/g'
+			printf '","jsonreq6":"'
+			echo "/${otable}" | nc ::1 "${v6_port}" | sed -n '/^[}{ ]/p' | sed 's/\\/\\\\/g; s/"/\\"/g'
+			printf '"}\n'
+		} 2>/dev/null
 		;;
 	hasipip)
 		result=$(ls /etc/modules.d/ | grep -E "[0-9]*-ipip")
--- a/applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
+++ b/applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
@@ -33,6 +33,19 @@ load_hosts() {
 	json_dump
 }
 
+json_stringescape() {
+	awk '
+		NR>1 { r = r "\\n" }
+		{
+			gsub(/["\\]/, "\\\\&");
+			gsub(/\t/, "\\t");
+			gsub(/\r/, "\\r");
+			r = r $0
+		}
+		END  { print r }
+	' | tr -d '\n'
+}
+
 case "$1" in
 list)
 	json_init
@@ -58,9 +71,9 @@ call)
 
 		{
 			printf '{"jsonreq4":"'
-			echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' | sed 's/\\/\\\\/g; s/"/\\"/g'
+			echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' | json_stringescape
 			printf '","jsonreq6":"'
-			echo "/${otable}" | nc ::1 "${v6_port}" | sed -n '/^[}{ ]/p' | sed 's/\\/\\\\/g; s/"/\\"/g'
+			echo "/${otable}" | nc ::1 "${v6_port}" | sed -n '/^[}{ ]/p' | json_stringescape
 			printf '"}\n'
 		} 2>/dev/null
 		;;
--- a/applications/luci-app-olsr/htdocs/luci-static/resources/view/olsr/frontend/olsrd.js
+++ b/applications/luci-app-olsr/htdocs/luci-static/resources/view/olsr/frontend/olsrd.js
@@ -483,6 +483,7 @@ return view.extend({
 		};
 
 		ifs.handleAdd = function (ev) {
+			uci.add('olsrd', 'Interface');
 			uci
 				.save()
 				.then(function () {
--- a/applications/luci-app-olsr/htdocs/luci-static/resources/view/olsr/frontend/olsrd6.js
+++ b/applications/luci-app-olsr/htdocs/luci-static/resources/view/olsr/frontend/olsrd6.js
@@ -462,6 +462,7 @@ return view.extend({
 		ifs.template = 'cbi/tblsection';
 
 		ifs.handleAdd = function (ev) {
+			uci.add('olsrd6', 'Interface');
 			uci
 				.save()
 				.then(function () {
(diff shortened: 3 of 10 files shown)


applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
```diff
@@ -56,13 +56,13 @@ call)
...
-		jsonreq4=$(echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' 2>/dev/null)
-		jsonreq6=$(echo "/${otable}" | nc ::1 "${v6_port}" | sed -n '/^[}{ ]/p' 2>/dev/null)
-
-		json_init
-		json_add_string "jsonreq4" "$jsonreq4"
-		json_add_string "jsonreq6" "$jsonreq6"
-		json_dump
+		{
+			printf '{"jsonreq4":"'
+			echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' | sed 's/\\/\\\\/g; s/"/\\"/g'
```
openwrt-ai: This sed escapes only `\` and `"`, but the piped olsrd output is multi-line: `sed -n '/^[}{ ]/p'` prints multiple lines, so literal newline (0x0A) characters get embedded into the `jsonreq4` string value. The old `json_add_string`/`json_dump` path escaped those (newline → `\n`); this manual replacement drops that.

The value is ultimately consumed in the browser by [`JSON.parse(json.jsonreq4)`](``https://github.com/openwrt/luci/blob/daaa103c38b6bc0fd0948ec23ad3c63816305b99/applications/luci-app-olsr/htdocs/luci-static/resources/olsr/common_js.js#L102),`` and a literal newline inside a JSON string is invalid per ECMA-262 — `JSON.parse` rejects it (reproduced: Node's `JSON.parse` throws `Bad control character in string literal`). The outer rpcd parse uses json-c (`json_tokener_new()`, [non-strict](https://github.com/openwrt/rpcd/blob/master/plugin.c#L217)) which may tolerate and re-normalize the newline on output, which could be why it works in your screenshots — but that masking is version-dependent and doesn't cover other control chars (tab, etc.).

Is this relying on json-c normalizing the embedded newlines? If so it's fragile; the escaping should also cover newlines and other control characters (`< 0x20`). The same applies to the v6 line below (line 63).

---

openwrt-ai (commented):
Reviewed 3 new commits. The `common/common_js.js` → `olsr/common_js.js` path fix and the `olsr.etx_color`/`olsr.snr_colors` qualifier fix both look correct (the module is exported via `'require olsr.common_js as olsr'` and lives at `resources/olsr/common_js.js`). One inline question on the manual JSON escaping in the rpcd backend.

---

applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
gratux: personal investigation suggests olsrd always responds with a single-line json, but your suggestion is more robust. please recheck with the latest commit.

applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
```diff
@@ -56,13 +56,13 @@ call)
...
-		jsonreq4=$(echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' 2>/dev/null)
-		jsonreq6=$(echo "/${otable}" | nc ::1 "${v6_port}" | sed -n '/^[}{ ]/p' 2>/dev/null)
-
-		json_init
-		json_add_string "jsonreq4" "$jsonreq4"
-		json_add_string "jsonreq6" "$jsonreq6"
-		json_dump
+		{
+			printf '{"jsonreq4":"'
+			echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' | sed 's/\\/\\\\/g; s/"/\\"/g; s/\n/\\n/g'
```
openwrt-ai: The new `s/\n/\\n/g` does not escape the embedded newlines it's meant to handle: `sed` processes input one line at a time and strips the trailing newline before the substitution runs, so the pattern space never contains a `\n` for the multi-line case. The inter-line newlines `sed -n '/^[}{ ]/p'` emits still land verbatim between the `printf` quotes, producing a literal control character inside the JSON string value — which is invalid per ECMA-262 and rejected by [`JSON.parse(json.jsonreq4)`](``https://github.com/openwrt/luci/blob/9effc8a8408da236ca40c720089914745e1de1d9/applications/luci-app-olsr/htdocs/luci-static/resources/olsr/common_js.js#L102``) in the browser.

Reproduced (multi-line input still yields a raw 0x0A in the value):

```
$ { printf '{"jsonreq4":"'; printf '{"a":1}\n{"b":2}\n' | sed -n '/^[}{ ]/p' | sed 's/\\/\\\\/g; s/"/\\"/g; s/\n/\\n/g'; printf '"}'; } | cat -A
{"jsonreq4":"{\"a\":1}$
{\"b\":2}$
"}
```

To actually escape across the whole multi-line stream (and avoid a stray trailing newline), the lines need to be joined explicitly. A single-pass `awk` does this cleanly:

```suggestion
			echo "/${otable}" | nc 127.0.0.1 "${v4_port}" | sed -n '/^[}{ ]/p' | awk '{ gsub(/\\/,"\\\\"); gsub(/"/,"\\\""); printf "%s%s", (NR>1?"\\n":""), $0 }'
```

Same fix applies to the v6 line at line 63 below.

---

openwrt-ai (commented):
Reviewed 1 new commit (`9effc8a8` "handle multi-line jsoninfo response"), which addresses the prior question about embedded newlines. The added `s/\n/\\n/g` does not actually escape them — `sed` works line-by-line, so the multi-line case still emits invalid JSON. One inline comment with a tested replacement.

---

applications/luci-app-olsr/root/usr/libexec/rpcd/olsrinfo
gratux: you're right, the newest commit should fix it

openwrt-ai (commented):
Reviewed 1 new commit (`a3d69b9` "handle multi-line jsoninfo response"); no new issues found. The new `json_stringescape()` awk helper correctly addresses the prior multi-line escaping concern: it escapes `\` and `"` (matching `["\\]` first, so order is safe), `\t` and `\r`, and joins multiple lines with a literal `\n`, producing valid JSON for the embedded `jsonreq4`/`jsonreq6` string values (verified with a JSON parser against multi-line input). The same helper is applied to both the v4 and v6 lines.

---

openwrt-ai (commented):
Reviewed 1 new commit (`be4f0039` "re-add missing interface creation"); no new issues found. Re-adding `uci.add('olsrd'/'olsrd6', 'Interface')` before `uci.save()` is correct: the following `.then` reads the pending `add` entry back out of `uci.changes()` to derive the new section id for navigation, so without the `uci.add()` the section id stayed null and the edit page loaded blank. The fix is applied symmetrically to both the v4 (`olsrd.js`) and v6 (`olsrd6.js`) views, and the commit message matches the diff.

---

gratux:
rebased onto master because ESLint complained about missing files

openwrt-ai (commented):
The branch was rebased onto master (head `10b08da`); the resulting diff is unchanged in content from the state reviewed previously — the common_js path fix, the `olsr.etx_color`/`olsr.snr_colors` qualifiers, the `json_stringescape` awk helper, and the symmetric `uci.add()` interface-creation re-add all match what was already reviewed. Re-verified the awk escaper and the `printf`-based JSON assembly independently; no new issues found. All three FormalityCheck runs are green on the current head.

---

jow-:
Merged, thanks!

pmelange:
Thank you for the changes. When will we see the updated package in 25.12?

gratux:
@pmelange Thanks for the reminder, I created the backport PR for 25.12 #8934.

PR #8718, opened 2026-06-19
luci-app-strongswan-swanctl: filter available algorithms

lvoegl:
strongSwan algorithm dropdowns used to list all known algorithms unaware of which are installed on the system. This pr adds a ucode script to filter installed and available algorithms and only displays those in the algorithm dropdowns.

```sh
$ ubus call luci.swanctl list-algs
{
        "error": null,
        "data": {
                "encryption": [
                        {
                                "name": "aes128",
                                "plugin": "aes",
                                "insecure": false
                        }
                 ],
                 ...
        }
}
```

### Maintainer
@lvoegl , Nicholas Smith

---

## Tested on
OpenWrt version: OpenWrt 25.12.3 (r32912-6639b15f62)
LuCI version: LuCI openwrt-25.12 branch (26.124.37917~fa51906)
Web browser(s): Firefox (151.0.4)


applications/luci-app-strongswan-swanctl/root/usr/share/rpcd/ucode/luci.swanctl
```diff
@@ -0,0 +1,369 @@
...
+				error: null,
+				data: null
+			};
+			const cmd = '/usr/sbin/swanmon get-algorithms';
+			const categories = keys(category_maps);
+			let selected_categories = categories;
+			if (req && req.args && req.args.type)
+				selected_categories = [req.args.type];
+
+			const output = json(popen(cmd)?.read?.('all'));
```
openwrt-ai: `json()` raises an uncaught exception on a non-string/non-object argument or on malformed input, which would abort the whole `list-algs` call rather than returning the intended `{ error, data: null }`. Two failure modes hit this:

- if `/usr/sbin/swanmon` is missing or not executable, `popen(cmd)` returns `null`, so `popen(cmd)?.read?.('all')` is `null` and `json(null)` throws `Passed value is neither a string nor an object`;
- if swanmon emits empty or non-JSON output, `json()` throws a syntax exception.

In both cases the frontend never sees `result.error` (the rpc call itself errors), so the `ui.addNotification(...)` path added in `swanctl.js` is unreachable and all dropdowns silently come up empty. Consider guarding the read result before parsing, e.g. capture `popen(cmd)?.read?.('all')`, bail to `result.error = '...'` when it is null/empty, and only then `json()` it (or wrap the parse so a decode failure populates `result.error`).

Note this same `json(popen(...)?.read?.('all'))` idiom already exists in [`luci.lldpd:6`](https://github.com/openwrt/luci/blob/2f2bc32473e7c232d0811cdeff4d55b24f5d6b67/applications/luci-app-lldpd/root/usr/share/rpcd/ucode/luci.lldpd#L6), so it is not unique to this PR — but here the surrounding code goes out of its way to surface a clean error to the user, and that effort is defeated by the unguarded parse.

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline robustness note. The commit message matches the diff (removes the static `strongswan_algorithms.js` baseclass, adds a `luci.swanctl` rpcd ucode backend that filters algorithms via `swanmon get-algorithms`, wires the view to the new `list-algs` ubus method, and grants the matching ACL). The ucode `for...in` semantics, two-variable object iteration, and category mapping all look correct, and the new ACL ubus grant matches the declared method.

---

openwrt-ai (commented):

The only change since the previous review wraps the `popen`/`json` call in `luci.swanctl` in a try/catch and returns the error to the caller. This addresses the earlier robustness note: ucode's `json()` raises an exception on parse errors, premature EOF, or a non-string/non-object argument (including the `popen` failure path where `popen(cmd)?.read?.('all')` evaluates to `null`), and all of those are now caught. The commit message still matches the overall diff.

---

feckert:
LGTM
Thanks merged!

Issue #7553, opened 2025-01-10
WiFi: mandate WPA3 in >= 6GHz bands


systemcrash:
I want to get some feedback before adding a fix for this. It's my understanding that 6GHz bands (WiFi 6E and beyond) mandate WPA3 and OWE only. See [here](https://www.extremenetworks.com/resources/blogs/wireless-security-in-a-6-ghz-wi-fi-6e-world) and [here](https://www.intel.com/content/www/us/en/products/docs/wireless/wi-fi-7.html#editorialTableBlade-1). Even there may be some devices which don't care about this. 

[Forum Example](https://forum.openwrt.org/t/adding-support-for-linksys-mr7500/170721/56).

How does the One (with WiFi 7) behave here? @Ansuel @aparcar 

WiFi 7 in general @rmandrad ?

A 'force' option as the new default is possible, so those wanting older WPA2 in 6GHz can have it.

rmandrad:
@systemcrash thank you for asking for my feedback. From a user perspective similar to 6e WPA3 (OWE haven't tried it) WIFI7 requires these. The  banapi4 wifi forum has plenty of examples of users falling on this "trap"...

My observation, there are some dependencies - 1. WPA3/OWE if possible, should be forced through the mac80211.sh / hostpad.sh scripts (if not already) or the new ucode implementation. Also, 2. firstboot setup should specify wpa3 (with a default password?)  somehow through the json spec of the device ... probably very hard ...

regarding luci - yes if it is wifi7 other than wpa3/owe the other options should not be available.

worth perhaps also include on this @blogic as he probably has an updated view of the ucode changes and if there is anything on this backlog regarding your question

ps. I haven't heard about a force option to use WPA2 or mixed for wifi7 ... why one would allow this ? as it is against the standard so imho i would keep away of implementing this.

systemcrash:
@rmandrad Could you give [this change](https://github.com/systemcrash/luci/blob/36e14ec0ed04bee26cbf8cf677bbdcf7e93188fa/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js) a spin and verify it works as intended on a 6G band?

rmandrad:
so tested

Test approach

replaced wireless.js, removed the luci-indexcache file and restarted the web server.

Test - Using the  banana pi4 device - with a mt7996 card, radio2 is 6ghz ,operating frequency mode BE  only WPA3* and OWE are shown.
[...]

PR #8723, opened 2026-06-22
luci-mod-network: hide incompatible encryption modes for 6 GHz

micpf:
The WiFi 6E (6 GHz) standard mandates WPA3 and management frame
protection.  While wifi-scripts already silently upgrades mixed modes
(psk-sae → sae) on 6 GHz radios, the LuCI encryption dropdown still
offers all WPA2/WPA/WEP options, leading to confusing hostapd failures
when users pick an unsupported mode.

When the radio band is 6 GHz, hide encryption options that are
incompatible with the standard:
- WPA2-PSK, WPA-PSK, WPA-PSK/WPA2-PSK Mixed Mode
- WPA2-PSK/WPA3-SAE Mixed Mode (6 GHz does not allow WPA2 fallback)
- WPA2-EAP, WPA-EAP, WPA2-EAP/WPA3-EAP Mixed Mode
- WEP

This leaves WPA3-SAE, WPA3-EAP, WPA3-EAP 192-bit, OWE, and No
Encryption as the available choices for 6 GHz interfaces.

There is a companion fix for openwrt/openwrt wifi-scripts that extends
the existing 6 GHz auth_type override to also cover pure WPA2 modes
(psk, eap), upgrading them to their WPA3 equivalents as a safety net.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1360,6 +1360,7 @@ return view.extend({
 
 
 				const crypto_modes = [];
+				const is_6ghz = uci.get('wireless', radioNet.getWifiDeviceName(), 'band') == '6g';
 
 				if (hwtype == 'mac80211') {
 					const has_supplicant = L.hasSystemFeature('wpasupplicant');
@@ -1386,9 +1387,11 @@ return view.extend({
 					const has_sta_wep = L.hasSystemFeature('wpasupplicant', 'wep');
 
 					if (has_hostapd || has_supplicant) {
-						crypto_modes.push(['psk2',      'WPA2-PSK',                    35]);
-						crypto_modes.push(['psk-mixed', 'WPA-PSK/WPA2-PSK Mixed Mode', 22]);
-						crypto_modes.push(['psk',       'WPA-PSK',                     12]);
+						if (!is_6ghz) {
+							crypto_modes.push(['psk2',      'WPA2-PSK',                    35]);
+							crypto_modes.push(['psk-mixed', 'WPA-PSK/WPA2-PSK Mixed Mode', 22]);
+							crypto_modes.push(['psk',       'WPA-PSK',                     12]);
+						}
 					}
 					else {
 						encr.description = _('WPA-Encryption requires wpa_supplicant (for client mode) or hostapd (for AP and ad-hoc mode) to be installed.');
@@ -1396,14 +1399,15 @@ return view.extend({
 
 					if (has_ap_sae || has_sta_sae) {
 						crypto_modes.push(['sae',       'WPA3-SAE',                     31]);
-						crypto_modes.push(['sae-mixed', 'WPA2-PSK/WPA3-SAE Mixed Mode', 30]);
+						if (!is_6ghz)
+							crypto_modes.push(['sae-mixed', 'WPA2-PSK/WPA3-SAE Mixed Mode', 30]);
 					}
 
 					// WPA3-Personal Compatibility Mode uses RSN overriding, which is an AP-only feature
-					if (has_ap_sae)
+					if (has_ap_sae && !is_6ghz)
 						crypto_modes.push(['sae-compat', 'WPA2-PSK/WPA3-SAE Compatibility Mode', 30]);
 
-					if (has_ap_wep || has_sta_wep) {
+					if (!is_6ghz && (has_ap_wep || has_sta_wep)) {
 						crypto_modes.push(['wep-open',   _('WEP Open System'), 11]);
 						crypto_modes.push(['wep-shared', _('WEP Shared Key'),  10]);
 					}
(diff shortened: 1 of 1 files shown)


OpenWRT-fanboy:
It's been a long time coming.

https://github.com/openwrt/luci/issues/7553

micpf:
> It's been a long time coming.
> 
> #7553

I see, so others also fell for this trap. We got this as customer feedback, and I was about to patch our build. But I thought this might be of interest upstream.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1336,6 +1336,7 @@ return view.extend({
 
 
 				const crypto_modes = [];
+				const is_6ghz = uci.get('wireless', radioNet.getWifiDeviceName(), 'band') == '6g';
```
openwrt-ai, line 1363: This filters the dropdown by band, but it does not account for a section that *already* stores a now-hidden mode. If a 6 GHz interface was previously configured with, say, `psk2` (imported config, set before this change, or via another band assignment), that value is no longer pushed into `crypto_modes`, so `encr.value()` never registers it. The encryption ListValue would then render with no matching option selected, and saving the form risks silently dropping or rewriting the stored `encryption` value rather than surfacing the incompatibility to the user.

Worth confirming the migration path: does the dropdown still show (and preserve) an existing incompatible value on a 6 GHz radio, or does it blank out? If it blanks, consider pushing the configured value as a disabled/flagged option so the user sees what is set before changing it.

---

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -1472,7 +1479,8 @@ return view.extend({
 					crypto_modes.push(['wep-shared', _('WEP Shared Key'),         10]);
 				}
 
-				crypto_modes.push(['none',       _('No Encryption'),   0]);
+				if (!is_6ghz)
+					crypto_modes.push(['none',       _('No Encryption'),   0]);
```
openwrt-ai, line 1512: nit: the PR description and this commit disagree about `none`. The PR body lists "No Encryption" among the choices that *remain* available for 6 GHz ("This leaves WPA3-SAE, WPA3-EAP, WPA3-EAP 192-bit, OWE, and No Encryption..."), but the commit message lists "No Encryption" under the *hidden* options, which is what this hunk actually does. Please reconcile the PR body with the commit message/code so the intended behavior is unambiguous.

---

openwrt-ai (commented):
Reviewed the 6 GHz encryption-mode filtering. The band detection (`uci.get('wireless', radioNet.getWifiDeviceName(), 'band') == '6g'`) is consistent with existing usage in this file, and narrowing the dropdown to WPA3/OWE on 6 GHz matches the wifi-scripts behavior in `package/network/config/wifi-scripts/.../ap.uc` (it forces PMF and SAE on 6 GHz). Two points left inline: one on preserving an already-configured incompatible mode on a 6 GHz radio, and one nit on the PR body vs commit message disagreeing about whether "No Encryption" stays available.

---

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
micpf: Regarding the migration path: if a 6 GHz interface already has an incompatible mode stored (e.g. `psk2`), that configuration never actually worked — hostapd rejects it with an "Invalid AKM suite" error and the radio doesn't come up. The ListValue defaults to the first available option when the stored value isn't in the list, so the user sees an obviously different selection and must actively save to apply it. This surfaces the broken config rather than silently preserving it, which seems like the right UX here.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
micpf: Good catch — the PR body was stale from before the "No Encryption" fix. Updated it to match the commit message.

OpenWRT-fanboy:
I don't believe no encryption is allowed on 6 GHz. It's WPA3 or OWE only.

maxskokov:
Code is correct — none is already gated behind !is_6ghz. The PR description was just stale. @OpenWRT-fanboy is right that open networks aren't permitted on 6 GHz per 802.11ax.

OpenWRT-fanboy:
Ah, I see it now. It was mentioned that PR description was updated, but I still see this:

> This leaves WPA3-SAE, WPA3-EAP, WPA3-EAP 192-bit, OWE, and No
> Encryption as the available choices for 6 GHz interfaces.

OpenWRT-fanboy (approved):

openwrt-ai (commented):
Re-reviewed the single commit rebased since my last review (`2045d96` → `203022459942afa47159fc1c69391176e574c979`). The only substantive code change is that `sae-compat` (WPA2-PSK/WPA3-SAE Compatibility Mode) is now also gated behind `!is_6ghz`, which is correct — it relies on a WPA2 fallback that 6 GHz does not allow. My two earlier inline threads (configured-value preservation on a 6 GHz radio, and the PR-body/commit-message wording for `none`) still apply and are left as-is rather than duplicated.

## Commit checks
- `2030224` "luci-mod-network: hide incompatible encryption modes for 6 GHz" — the message carries a `(cherry picked from commit 2045d96ceb2f6107db7de89fb4397a1a78a4f246)` trailer, but this PR targets `master` and `2045d96` is this PR's own earlier (force-pushed) head — it is not in mainline history nor on any openwrt branch. The `git cherry-pick -x` trailer is spurious here and will dangle once merged; please drop it before merge.

---

openwrt-ai (commented):

The force-push (`2030224` → `0a9ab27`) leaves the tree identical and only amends the commit message: the spurious `(cherry picked from commit 2045d96…)` trailer flagged in my previous review has been dropped. Message now matches the change.

---

PR #8727, opened 2026-06-23
luci-base: don't crash in String.format() on an undefined argument

Cris7015:
### Description

The status Overview "Ports" widget (`view/status/include/29_ports.js`) crashes with `TypeError: Cannot read properties of undefined (reading 'toString')` (thrown from `String.format()` in `cbi.js`) on any device where a port's netdev reports no statistics: `stats.rx_bytes` etc. are `undefined`, and `'%1024mB'.format(stats.rx_bytes)` aborts. This takes down the **entire Overview render** — every other page works fine.

**Root cause:** `String.prototype.format()` runs its conversion `switch` only when the argument is defined (`if (param !== undefined)`). For a magnitude conversion (`'%<n>m...'`) the switch is also what clears `pMinLength` (which there holds the divisor, not a field width). With an `undefined` argument the switch is skipped, so `pMinLength` stays set while `subst` is left `undefined`, and the following `if (pMinLength) subst = subst.toString()` throws.

**Fix:** emit an empty substitution and clear the field width when the argument is `undefined`.

### Maintainer
@jow-

---

## Tested on
**OpenWrt version:** OpenWrt SNAPSHOT r34203 (target `econet/en751221`, EcoNet EN751221) — TP-Link Archer XR500v
**LuCI version:** master `26.151.60149~1bffbf4`
**Web browser(s):** Chromium (used to capture the console stack trace)

Before: the Overview throws `Cannot read properties of undefined (reading 'toString')` and renders nothing.
After: the Ports panel renders normally (lan1–4 + wan with their stats). Unit check on the running device: `'%1024mB'.format(undefined)` no longer throws, and defined values still format correctly (`'%1024mB'.format(123456789)` → `"117.74 MiB"`).

---

## Checklist
- [x] This PR is not from my *main*/*master* branch, but a *separate* branch (`fix-string-format-undefined-crash`).
- [x] Each commit has a valid `Signed-off-by: Cristian Papa <pcristian292@gmail.com>` row (reachable email, not a GitHub noreply).
- [x] Each commit and the PR title have a valid `<package name>: title` first-line subject (`luci-base: ...`).
- [x] `PKG_VERSION`: N/A — `luci-base` has no `PKG_VERSION` (it is versioned via `luci.mk`/git), so there is nothing to increment.


modules/luci-base/htdocs/luci-static/resources/cbi.js
```diff
@@ -836,6 +836,10 @@ String.prototype.format = function()
 					break;
 			}
 		}
+		else {
+			subst = '';
+			pMinLength = null;
```
openwrt-ai: nit: this changes the observable result of a plain `%s`/`%d` conversion on an undefined argument from the literal string `"undefined"` (the old code fell through to `out += leftpart + subst` with `subst === undefined`) to `""`. That's a sensible improvement and matches the PR's intent, but it is a wider behavior change than just the `%m` crash — worth a one-line mention in the commit body so it's not mistaken for a pure `%m` fix. Clearing `pMinLength` here is correct (otherwise an empty `subst` would still be space-padded to the field width).

---

openwrt-ai (commented):
Reviewed 1 new commit. The fix is correct: with an undefined argument the `if (param !== undefined)` block was skipped, so for a `%<n>m` conversion `pMinLength` (holding the divisor) stayed set while `subst` stayed undefined, and `subst.toString()` threw. Emitting an empty `subst` and nulling `pMinLength` resolves it. The commit message accurately describes the change. One non-blocking nit inline.

---

Cris7015:
Thanks for the review! Addressed the nit in the commit body: added a line noting that, besides fixing the `%m` crash, this also changes a plain `%s`/`%d` on an undefined argument from the literal `"undefined"` to an empty string (and that clearing `pMinLength` there avoids padding that empty value out to the field width). No code change — force-pushed as b16edfa.

openwrt-ai (commented):

The only change since the previous review is an expanded commit body paragraph documenting the `%s`/`%d`-on-undefined side effect (empty string instead of literal `"undefined"`); the code diff is unchanged and the message accurately describes it.

---

jow-:
It's not entirely correct. Imho the most intuitive behavior of `'%10s'.format(undefined)` would be printing ten spaces (padding plus empty string) instead of nothing.

Cris7015:
Agreed — preserving the field width is the right call. Updated so an undefined argument uses an empty substitution but keeps the parsed width, making `'%10s'.format(undefined)` render ten spaces.

`%m` stays special: it reuses the min-length slot as the metric divisor (e.g. `%1024m`), not a field width, so that path still clears `pMinLength` to avoid padding the empty value out to ~1024 chars.

While at it I moved the `pad` initialization ahead of the `param !== undefined` check, since the padding block now runs for undefined arguments too and would otherwise use an uninitialized/stale pad character. Pushed as 9ac6b20.

Cris7015:
Replaced by #8745.

GitHub no longer allows reopening this PR after its source branch was recreated, so I opened a clean replacement PR from the corrected branch state. The new PR keeps the same fix scope and incorporates the review feedback here: undefined arguments use an empty substitution while preserving normal field-width padding, `%m` remains special because its min-length slot is the metric divisor, and pad initialization is moved before the undefined-argument check.

PR #8729, opened 2026-06-24
luci-base: sync translations

AndyChiang888:
add missing translations
fixes: 4556c5ecf2bc

## Pull request details
Some translations for openfortivpn and openconnect got lost again...
### Maintainer _(preferred)_

@hnyman 

---

## Tested on

**OpenWrt version:** SNAPSHOT
**LuCI version:** master
**Web browser(s):** Chrome

---

## Checklist

### DIFF
--- a/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
+++ b/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
@@ -128,7 +128,7 @@ return network.registerProtocol('openfortivpn', {
 			return callSetCertificateFiles(section_id, null, sanitizeCert(value), null);
 		};
 
-		o = s.taboption('general', form.TextValue, 'ca_file', _('CA certificate (PEM encoded; Use instead of system-wide store to verify the gateway certificate.'));
+		o = s.taboption('general', form.TextValue, 'ca_file', _('CA certificate (PEM encoded. Use instead of system-wide store to verify the gateway certificate.)'));
 		o.optional = true;
 		o.rows = 10;
 		o.monospace = true;


openwrt-ai (commented):

---

hnyman:
Let's try to fix first the underlying reasons first. Wrong quoting / punctuation / missing escapes.

AndyChiang888:
I've fixed the typos in openfortivpn. For the rest, running `build/mkbasepot.sh` on my end didn't trigger any errors, and I didn't find any issues with `"Optional, in seconds. If set to '0', no reconnect is attempted."` either.

hnyman:
continuing the discussion from [4556c5e comments](https://github.com/openwrt/luci/commit/4556c5ecf2bc40f25e40ea20039d7f9e99fbc784#commitcomment-190059327)

Sorry, not applying your PR.
I am going there piecemeal. I already synced the switch/vlan things to get that clear part cleaned away.

Trying the mkbasepot as part of i18n-sync has produced some warnings for some time, like this:
```
perus@ub2604:/OpenWrt/github/luci$ ./build/mkbasepot.sh 
Updating modules/luci-base/po/templates/base.pot ... 
standard input:3366: warning: unterminated string
standard input:3734: warning: unterminated string
standard input:5106: warning: unterminated string
standard input:1007: warning: unterminated string
standard input:126: warning: unterminated string
standard input:181: warning: RegExp literal terminated too early
standard input:105: warning: unterminated string
done
```

I have ignored those, but it may be that some things get interrupted at those points.
Two of those are coming from protocols:
```
perus@ub2604:/OpenWrt/github/luci$ ./build/i18n-scan.pl protocols > koe
standard input:181: warning: RegExp literal terminated too early
standard input:105: warning: unterminated string
```

One from openfortivpn:
```
perus@ub2604:/OpenWrt/github/luci$ ./build/i18n-scan.pl protocols/luci-proto-openfortivpn/
standard input:181: warning: RegExp literal terminated too early
msgid ""
msgstr "Content-Type: text/plain; charset=UTF-8"

#: protocols/luci-proto-openfortivpn/root/usr/share/rpcd/acl.d/luci-openfortivpn.json:3
msgid "Grant access to LuCI openfortivpn procedures"
msgstr ""
```
181 is the last line, so not the exact place.
And then most of strings are removed...

After trials, the line that actually triggers the error is the regexp here:
`else if (start && !lines[i].match(/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/))`

Quoting the regexp string removes the xgettext error and produces the full string list, but I am not sure if keeps things working ok...

Regexp tester gives error about that line:
`/ An unescaped delimiter must be escaped, use \ `

I tried to fix it by adding escapes as `(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?$`, escaping the / chars with  \, and that seems to fix xgettext, too.

That line is originating from @systemcrash ES6 changes in 50a8d5325db, where he removed those escape chars before "/".  Likely not intended.

Similar "remove escape" change done to openconnect and wireguard by the same commit by @systemcrash  . 
Again the question, if those regexp changes where intended or not?
At least xgettext seems unhappy with them.

Not sure, what is the best.
But I think that maybe restoring the regexp would be the better way.
Any comments from js / regexp experts?

hnyman:
The other errors on the list are coming from these files:
(no idea about the exact location, yet)

```
modules/luci-base/htdocs/luci-static/resources/ui.js at ./build/i18n-scan.pl line 270, <S> chunk 21.
standard input:3366: warning: unterminated string
standard input:3734: warning: unterminated string
standard input:5106: warning: unterminated string

modules/luci-base/htdocs/luci-static/resources/validation.js at ./build/i18n-scan.pl line 270, <S> chunk 34.
standard input:1007: warning: unterminated string


modules/luci-mod-status/htdocs/luci-static/resources/view/status/routes.js at ./build/i18n-scan.pl line 270, <S> chunk 146.
standard input:126: warning: unterminated string

protocols/luci-proto-openconnect/htdocs/luci-static/resources/protocol/openconnect.js at ./build/i18n-scan.pl line 270, <S> chunk 213.
standard input:105: warning: unterminated string
```

AndyChiang888:
It’s strange that I haven‘t been prompted with any errors.

hnyman:
cc @jow-  @systemcrash

I tested and it looks like either 
* quoting the possibly faulty regexp strings, or
* escaping /  with  \

will fix the warnings.

These changes leave just one xgettext warning, which I think is a false positive. They are a mix of adding single quotes, double quotes or escaping. So nothing final, but just highlighting the fix possibilities.

```
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -3362,7 +3362,7 @@ const UIFileUpload = UIElement.extend(/** @lends LuCI.ui.FileUpload.prototype */
         * @returns {Promise}
         */
        handleDelete(path, fileStat, ev) {
-               const parent = path.replace(/\/[^/]+$/, '') ?? '/';
+               const parent = path.replace(/\/[^\/]+$/, '') ?? '/';
                const name = path.replace(/^.+\//, '');
                let msg;
 
@@ -5102,7 +5102,7 @@ const UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
                                for (let i = 0; i < 2; i++)
                                        for (let j = 0; j < ipaddrs.length; j++)
                                                tasks.push(this.pingDevice(i ? 'https' : 'http', ipaddrs[j])
-                                                       .then(ev => { reachable = ev.target.src.replace(/^(https?:\/\/[^/]+).*$/, '$1/') }, () => {}));
+                                                       .then(ev => { reachable = ev.target.src.replace('/^(https?:\/\/[^/]+).*$/', '$1/') }, () => {}));
 
                                return Promise.all(tasks).then(() => {
                                        if (reachable) {
diff --git a/modules/luci-base/htdocs/luci-static/resources/validation.js b/modules/luci-base/htdocs/luci-static/resources/validation.js
index abd9e00948..0d5e21273a 100644
--- a/modules/luci-base/htdocs/luci-static/resources/validation.js
+++ b/modules/luci-base/htdocs/luci-static/resources/validation.js
@@ -1003,7 +1003,7 @@ const ValidatorFactory = baseclass.extend(/** @lends LuCI.validation.ValidatorFa
                        if (v == '.' || v == '..')
                                return this.assert(false, _('valid network device name, not "." or ".."'));
 
-                       return this.assert(v.match(/^[^:/%\s]{1,15}$/), _('valid network device name between 1 and 15 characters not containing ":", "/", "%" or spaces'));
+                       return this.assert(v.match("/^[^:/%\s]{1,15}$/"), _('valid network device name between 1 and 15 characters not containing ":", "/", "%" or spaces'));
                },
 
                /**
```

```
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/routes.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/routes.js
@@ -121,7 +121,7 @@ return view.extend({
                const res = [];
 
                for (const line of routes.trim().split(/\n/)) {
-                       const [, type = 'unicast', d, f = [] ] = (line.match(/^(?:([a-z_]+|\d+) )?(default|[0-9a-f:./]+) (.+)$/) || []);
+                       const [, type = 'unicast', d, f = [] ] = (line.match("/^(?:([a-z_]+|\d+) )?(default|[0-9a-f:./]+) (.+)$/") || []);
                        const dest = d == 'default' ? (v6 ? '::/0' : '0.0.0.0/0') : d;
                        const flags = f?.trim?.().split?.(/\s+/);
 ```
```
--- a/protocols/luci-proto-openconnect/htdocs/luci-static/resources/protocol/openconnect.js
+++ b/protocols/luci-proto-openconnect/htdocs/luci-static/resources/protocol/openconnect.js
@@ -44,7 +44,7 @@ function validateCert(priv, section_id, value) {
        const beg = priv ? /^-----BEGIN (RSA )?PRIVATE KEY-----$/ : /^-----BEGIN CERTIFICATE-----$/;
        const end = priv ? /^-----END (RSA )?PRIVATE KEY-----$/ : /^-----END CERTIFICATE-----$/;
        const lines = value.trim().split(/[\r\n]/);
-       const base64 = /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/;
+       const base64 = "/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/";
        const errmsg = _('This does not look like a valid PEM file');
 ```

```
--- a/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
+++ b/protocols/luci-proto-openfortivpn/htdocs/luci-static/resources/protocol/openfortivpn.js
@@ -48,7 +48,7 @@ function validateCert(priv, section_id, value) {
        for (i = 0; i < lines.length; i++) {
                if (lines[i].match(/^-{5}BEGIN ((|RSA |DSA )PRIVATE KEY|(|TRUSTED |X509 )CERTIFICATE)-{5}$/))
                        start = true;
-               else if (start && !lines[i].match(/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/))
+               else if (start && !lines[i].match("/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/"))
                        break;
        }

```

Any idea about the best alternative, @jow-  @systemcrash  ?

Ps. The possibly false positive seems be coming from here, where I see nothing strange at the first glance.
```
		if (!L.isObject(this.menu)) {
			this.menu = request.get(L.url('admin/menu')).then(L.bind((menu) => {
				this.menu = scrubMenu(menu.json());
				session.setLocalData('menu', this.menu);

				return this.menu;
			}, this));
		}
```

AndyChiang888:
I didn't encounter any warnings or errors during the execution:
```
andy@debian:~/Build/luci$ build/mkbasepot.sh
Updating modules/luci-base/po/templates/base.pot ... done
```

What I find strange is why the same i18n-scan.pl behaves differently under different systems.

AndyChiang888:
I've updated the PR to only fix typo, as we need to figure out why this happens first.
Otherwise, the strings might get lost again the next time you sync.

hnyman:
> ...why the same i18n-scan.pl behaves differently under different systems.

The reason is likely something in the underlying xgettext that does the actual translation string work.
Which version do you have?

Mine in Ubuntu 26.04 is this:
```
perus@ub2604:/OpenWrt/github/luci$ xgettext --version
xgettext (GNU gettext-tools) 0.23.2
...
```

AndyChiang888:
The current version in Debian forky is:
```
andy@debian:~$ xgettext --version
xgettext (GNU gettext-tools) 1.0
...
```

hnyman:
Ubuntu is apparently bringing the new version into 26.10:
https://launchpad.net/ubuntu/stonking/+source/gettext

```
Stonking Stingray           (active development)
1.0-1 | proposed           (main) | 2026-06-17
0.23.2-1 | release           (main) | 2026-04-24
 
The Resolute Raccoon           (current stable release)
0.23.2-1 | release           (main) | 2025-11-14
```

But the errors actually look real to me, as also regexp checkers complain about them. So, we likely should fix the errors in any case.

hnyman:
I made the fixes with escaping the slash chars, and then syncing luci-base.
See c653c94 and 217c5f7

AndyChiang888:
Great! But I was wondering, if your xgettext gets upgraded, would need to revert https://github.com/openwrt/luci/commit/c653c94a02ea7cbca1a03cceb5beeb6f631b0fbe ?
Because the original code didn't actually have any errors in JavaScript.

AndyChiang888:
Also, wireguard is missing a string:
https://github.com/openwrt/luci/blob/c3bb51081e18874c6ee6ed4cc80b869efebcd665/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js#L38

https://github.com/openwrt/luci/blob/c3bb51081e18874c6ee6ed4cc80b869efebcd665/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js#L41

hnyman:
> Because the original code didn't actually have any errors in JavaScript.

Several regex escaping advice articles mention that slash should be escaped in javascript, as it is used as the regex delimiter in javascript.  So, the escaping should be right in any case.

The handling of a surprising slash may depend on the smartness or strictness of the various tools parsing the code. Some may "smartly" overlook the missing escape, some will complain about the errors.

> wireguard is missing a string

I noticed that there was a wireguard change earlier, but haven't investigated the reason for that.

But it looks like there is a similar thing:
The previous line before the missing string is a regex with non-escsaped slashes...

https://github.com/openwrt/luci/blob/c3bb51081e18874c6ee6ed4cc80b869efebcd665/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js#L37-L38

AndyChiang888:
I’ll have to leave this to you after all. xgettext doesn't report any errors on my side, so there's not much I can do to help.

hnyman:
> wireguard is missing a string

See 176a487 and e1bc5a0

PR #8662, opened 2026-05-26
luci-mod-network: Add option to control auto-negotiation and phy speed and duplex

Sirherobrine23:
### Description
Add some settings that were missing from the device configuration have been added to control auto-negotiation, PHY speed, duplex, and rx/tx pause.

A binding has also been added to luci.c to retrieve the supported PHY speeds, if no value is returned, 10Mbps, 100Mbps, 1Gbps, 2.5Gbps, 5Gbps, 10Gbps, and 25Gbps are set as available values for use.

### Screenshot or video of changes

mikrotik lan_cgi-bin_luci_admin_network_network (1)

[Gravação da tela 2026_05_26 14-47-40.webm](https://github.com/user-attachments/assets/e58aa58f-8227-46ba-92e7-0a359354277e)

---

## Tested on

**Device:** MikroTik RouterBOARD 760iGS (hEX S)
**OpenWrt version:** 25.12.4 (r32933-4ccb782af7)
**LuCI version:** Master (26.146.62900~8bc2621)
**Web browser(s):** Chrome 148.0.7778.178, Firefox 151.0.1

---

## Checklist
- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch. :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`).
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages.
- [ ] Incremented :up: any `PKG_VERSION` in the Makefile.
- [ ] _(Optional)_ Includes what Issue it closes (e.g. openwrt/luci#issue-number).
- [ ] _(Optional)_ Includes what it depends on (e.g. openwrt/packages#pr-number in sister repo).

PR #8731, opened 2026-06-24
luci-mod-network: add link speed column to the devices tab

tobiaspal:
## Description

While investigating a link that had negotiated down from 1 Gbit/s to 100 Mbit/s, I noticed LuCI surfaces the current link speed only in the topology-driven "Port status" widget on Status → Overview, which shows nothing for plain network devices. The negotiated speed isn't shown on the Network → Interfaces / Devices pages, so a link stuck below its capability is not visible in LuCI.

This adds a **Link Speed** column to the Devices tab showing:

- the negotiated speed (e.g. `1 Gbit/s`), or `-` when the device has no carrier or reports no speed (ppp / tunnel interfaces, etc.);
- the device's supported link speeds in a tooltip (e.g. `Supported link speeds: 10 Mbit/s, 100 Mbit/s, 1 Gbit/s`) when it exposes ethtool link modes.

This makes it possible to spot a device that isn't at its expected link speed.

The current speed comes from the device link state; the supported speeds are the distinct `network.device status` `link-supported` modes, fetched once in `load()`. The `network.device "status"` ACL grant is required for the view to call that method.

### Related PRs

- #8282 adds the same `network.device "status"` ACL grant (for a speed-selection config field). If it lands first, this PR's ACL hunk can be dropped on rebase.
- #8662 / #8270 add speed / duplex / auto-negotiation *configuration* in the same files. This change is read-only *display* and is complementary.

---

Developed with LLM assistance; reviewed and tested by me.

## Checklist

- [x] Not from my *main* / *master* branch, but a separate branch.
- [x] Each commit has a valid `Signed-off-by`.
- [x] Commit and PR title use the `<package>: title` form.
- [ ] Incremented `PKG_VERSION` — n/a (`luci.mk` auto-versions).

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
@@ -358,6 +358,7 @@ return view.extend({
 			network.getDSLModemType(),
 			network.getDevices(),
 			fs.lines('/etc/iproute2/rt_tables'),
+			L.resolveDefault(callNetworkDeviceStatus(), {}),
 			uci.changes()
 		]);
 	},
@@ -456,7 +457,7 @@ return view.extend({
 		]);
 	},
 
-	render([dslModemType, netDevs, rtTables]) {
+	render([dslModemType, netDevs, rtTables, devStatus]) {
 
 		if (this.interfaceBridgeWithIfnameSections().length)
 			return this.renderBridgeMigration();
@@ -1737,6 +1738,41 @@ return view.extend({
 		o = s.option(form.DummyValue, 'type', _('Type'));
 		o.textvalue = getDevTypeDesc;
 		o.modalonly = false;
+
+		o = s.option(form.DummyValue, '_speed', _('Link Speed'));
+		o.modalonly = false;
+		o.textvalue = function(section_id) {
+			const dev = getDevice(section_id);
+
+			if (!dev || !dev.getCarrier())
+				return '-';
+
+			const cur = dev.getSpeed();
+
+			/* getSpeed() returns Mbit/s, -1 with carrier but no negotiated rate, or null if unsupported */
+			if (!(cur > 0))
+				return '-';
+
+			/* %1000mBit/s would render 1 Gbit as "1000.00 MBit/s" (strict scale step + forced decimals), so format directly */
+			const fmt = (mbit) => (mbit >= 1000) ? (mbit / 1000) + ' Gbit/s' : mbit + ' Mbit/s';
+
+			/* distinct ethtool-supported link speeds, e.g. 1000baseT-F => 1000, sorted ascending */
+			const st = devStatus ? devStatus[dev.getName()] : null;
+			const supported = (st ? L.toArray(st['link-supported']) : []).reduce(function(speeds, mode) {
+				const n = mode.match(/^(\d+)base/);
+				if (n && speeds.indexOf(+n[1]) < 0)
+					speeds.push(+n[1]);
+				return speeds;
+			}, []).sort((a, b) => a - b);
+
+			/* current speed in the cell; all supported speeds in the tooltip when known */
+			if (supported.length)
+				return E('span', {
+					'data-tooltip': _('Supported link speeds: %s').format(supported.map(fmt).join(', '))
+				}, [ fmt(cur) ]);
+
+			return fmt(cur);
+		};
 
 		o = s.option(form.DummyValue, 'macaddr', _('MAC Address'));
 		o.modalonly = false;
--- a/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
+++ b/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json
@@ -14,7 +14,8 @@
 			"ubus": {
 				"file": [ "exec" ],
 				"iwinfo": [ "assoclist", "countrylist", "freqlist", "txpowerlist" ],
-				"luci": [ "getSwconfigFeatures", "getSwconfigPortState" ]
+				"luci": [ "getSwconfigFeatures", "getSwconfigPortState" ],
+				"network.device": [ "status" ]
 			},
 			"uci": [ "dhcp", "firewall", "luci", "network", "wireless", "system" ]
 		},


tobiaspal:
Screenshot of the column, including the popup tooltip:
luci-link-speed-mismatch

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -456,7 +457,7 @@ return view.extend({
 		]);
 	},
 
-	render([dslModemType, netDevs, rtTables]) {
+	render([dslModemType, netDevs, rtTables, , devStatus]) {
```
feckert: Isn't there a comma too many there?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1762,6 +1763,44 @@ return view.extend({
...
+
+			if (!dev || !dev.getCarrier())
+				return '-';
+
+			const cur = dev.getSpeed();
+
+			if (!(cur > 0))
+				return '-';
+
+			/* max = highest speed among the ethtool-supported link modes, e.g. 1000baseT-F => 1000 */
```
feckert: What's the json output of Ubus here? Can we simplify how it's handled? What about SFP interfaces they often support 10 Gb or even 100 Gb. I'm concerned about the code's future-proofing.

How about if we just display all supported link speeds in the tooltip and make the selected one bold? After all, it always depends on what the other end can handle. So red might not always be the right choice here.
Can you attach a screenshot showing exactly how this looks for you in different situations?

feckert (changes requested):

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -358,7 +358,8 @@ return view.extend({
 			network.getDSLModemType(),
 			network.getDevices(),
 			fs.lines('/etc/iproute2/rt_tables'),
-			uci.changes()
+			uci.changes(),
+			L.resolveDefault(callNetworkDeviceStatus(), {})
```
openwrt-ai: `callNetworkDeviceStatus()` is invoked here with no `name` argument, but `render()` then treats the result as a map keyed by device name (`devStatus[dev.getName()]`). The in-tree evidence suggests `network.device status` is a *per-device* call that returns a single flat status object, not a keyed map:

- the rpc is declared with [`params: [ 'name' ]` at interfaces.js:15-20](``https://github.com/openwrt/luci/blob/6effb505ccc22429b64846d8bfcc5243c04f38ed/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L15-L20);``
- the only other caller passes a name and reads the result as flat — [`callNetworkDeviceStatus(devName)` ... `status.pse` at interfaces.js:1593-1594](``https://github.com/openwrt/luci/blob/6effb505ccc22429b64846d8bfcc5243c04f38ed/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L1593-L1594).``

If that is correct, the name-less call returns one flat status object (or errors / returns the default device), so `devStatus[dev.getName()]` is `undefined` for every row, `max` stays `0`, and the maximum-supported-speed comparison (the main feature) never triggers. Is the name-less form confirmed to return an all-devices map keyed by device name? If not, the per-device status needs to be fetched per device (e.g. one call per `dev.getName()`), or the supported modes read from a source that does return all devices at once.

---

openwrt-ai (commented):
Reviewed 1 new commit. One inline comment on the `network.device status` call signature; the commit message accurately matches its diff.

---

jow-:
Does this gracefully handle devices without link speed information? Can you post a screenshot of the overall view?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
```diff
@@ -1762,6 +1763,44 @@ return view.extend({
...
+				return '-';
+
+			/* max = highest speed among the ethtool-supported link modes, e.g. 1000baseT-F => 1000 */
+			const st = devStatus ? devStatus[dev.getName()] : null;
+			const max = st ? L.toArray(st['link-supported']).reduce(function(mx, mode) {
+				const n = mode.match(/^(\d+)base/);
+				return n ? Math.max(mx, +n[1]) : mx;
+			}, 0) : 0;
+
+			const fmt = (mbit) => (mbit >= 1000) ? (mbit / 1000) + ' Gbit/s' : mbit + ' Mbit/s';
```
jow-: There's the `%1000mBit/s` format available to handle this automatically:

`let txt = '%1000mBit/s'.format(mbit * 1000)` should automatically produce Bit/s, KBit/s, MBit/s, GBit/s - whatever is most applicable.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
tobiaspal: This is a screenshot when the current and max link speeds match:
luci-link-speed-match
And this shows the cases when the link speed is not available (`-1` in `/sys/class/net/.../speed` (for example tailscale) or `speed` not present in `/sys/class/net/.../` (for example ppp interfaces)), and when the max speed is unknown.
luci-link-speed-unknown

Displaying all supported link speeds in the tooltip and make the selected one bold will be more informative, I'll change it.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
tobiaspal: It's intentional, skips the unused `uci.changes()` slot, but reads as a typo.

tobiaspal: Fixed the double comma in commit 54e0f338d0bcd52d2dddca09e524ffec2d52a05e, what do you think?

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
tobiaspal: Changed in commit 4a0ebc9caf8eef33b02c5cb9146be2a6fa5b1f39, no more red color.
`ubus call network.device status` exposes a `link-supported` array: `["10baseT-H","10baseT-F",...]`. Now the tooltip shows the supported speeds. Did not find a way to format bold in the tooltip.
The json has `speed`, `link-supported`, `link-advertising`, `link-partner-advertising`, which would be all good to show somewhere, but then the column would be quite crowded.
My goal is to have a way to check if any of the links are not at the expected speed.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
tobiaspal: `network.device status` called without a `name` returns an all-devices map keyed by device name (netifd), not a single flat object. On the test device `ubus call network.device status` returns `{ "lan": {…}, "wan-3bb": {…}, … }`, and `devStatus[dev.getName()]` resolves per row (the tooltip renders), so the name-less call is intentional.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js
tobiaspal: Thank you! I tried it, but it formats `1 GBit/s` to `1000.00 MBit/s`.

openwrt-ai (commented):
Reviewed 3 new commits; no new issues found.

---

tobiaspal:
This is how the Devices tab looks now after the 4 new commits:
luci-link-speed-4

(Apologize for the many commits, I just keep thinking about how to make this better. If you prefer one commit and a clean PR, I can close this and open a new PR.)

feckert:
Please fixup the 5 commits into one.
Do the following commands in your feature branch
git rebase -i HEAD~5 this will open the vi editor
replace in each line the pick with fixup execpt the top one. This will merge alle Commits into one commit
After that make a git push --force <your branch name>
Thats it.

Please make also CI/CD check for Test Formalities happy :-)

tobiaspal:
Squashed into one commit and rebased on master. Thank you!

openwrt-ai (commented):
Reviewed the squashed commit (history was rebased into one, so the prior review's commit no longer maps); no new issues found. The earlier concern about the name-less `network.device status` call returning an all-devices map keyed by device name is confirmed resolved by the author, and the direct speed formatting (vs `%1000mBit/s`) is justified in an inline comment. The squashed commit message accurately matches its diff.

---

feckert:
LGTM
Thanks merged!

PR #8733, opened 2026-06-24
luci-app-poe: add PoE control application

pieterhollander:
LuCI view under Status -> PoE driven by the realtek-poe ubus object: live per-port status/consumption, an immediate temporary on/off toggle.

Under Network -> PoE a Save & Apply form for persistent /etc/config/poe settings (enable, priority, PoE+).

Signed-off-by: Pieter Hollander <git@pieterhollander.nl>

### Description

A LuCI web interface for controlling Power over Ethernet (PoE) on OpenWrt
switches driven by the [`realtek-poe`](https://github.com/Hurricos/realtek-poe)
daemon (e.g. the HPE 1920 PoE models on the `realtek` target).

It adds **Status → POE** and **Network → PoE** pages that let you:

- view live per-port status (mode, status, priority, power consumption) plus the
  global power budget and total consumption;
- switch power on individual ports on/off immediately (via the `poe` ubus
  `manage` method — note this is **temporary** and resets on reload/reboot);
- edit the persistent configuration in `/etc/config/poe` (per-port enable,
  priority, PoE+) through a normal Save & Apply form.

If this gets merged, I'd like to back port it to 25.12 as well, so I have a GUI to configure and toggle my PoE ports.

Status -> PoE

Screenshot 2026-06-28 at 13-05-13 OpenWrt PoE

Network -> PoE

Screenshot 2026-06-28 at 13-07-10 OpenWrt PoE

## Tested on

**OpenWrt version:** OpenWrt 25.12.4 r32933-4ccb782af7
**LuCI version:** LuCI (HEAD detached at e9ebca7) branch 26.133.20346~e9ebca7
**Web browser(s):** Safari Version 26.5 (21624.2.5.11.4)

---

## Checklist

- [x] Incremented :up: any `PKG_VERSION` in the Makefile.
- [x] _(Optional)_ Includes what it depends on (e.g. openwrt/packages#pr-number in sister repo).

### DIFF
--- a/applications/luci-app-poe/htdocs/luci-static/resources/view/poe/status.js
+++ b/applications/luci-app-poe/htdocs/luci-static/resources/view/poe/status.js
@@ -1,0 +1,201 @@
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+'require view';
+'require rpc';
+'require ui';
+'require poll';
+'require dom';
+
+var callPoeInfo = rpc.declare({
+	object: 'poe',
+	method: 'info'
+});
+
+var callPoeManage = rpc.declare({
+	object: 'poe',
+	method: 'manage',
+	params: [ 'port', 'enable' ]
+});
+
+function portNum(name) {
+	var m = String(name).match(/(\d+)/);
+	return m ? +m[1] : 0;
+}
+
+/* realtek-poe reports a per-port status string such as
+   "Disabled", "Searching", "Delivering", "Fault". Anything other than
+   "Disabled" means power delivery is enabled on the port. */
+function portIsOn(port) {
+	return String((port && port.status) || '').toLowerCase() != 'disabled';
+}
+
+/* realtek-poe reports priority 0 when no explicit priority has been set on a
+   port; values 1-3 map to the Low/Medium/High choices in the config form. */
+function priorityLabel(prio) {
+	if (prio == null)
+		return '-';
+	switch (+prio) {
+	case 0: return _('Unset (0)');
+	case 1: return _('Low (1)');
+	case 2: return _('Medium (2)');
+	case 3: return _('High (3)');
+	default: return String(prio);
+	}
+}
+
+/* The daemon includes a per-port "consumption" field while the port is
+   actually delivering power; the field is absent when idle. */
+function consumptionLabel(port) {
+	return (port && port.consumption != null) ? '%.1f W'.format(port.consumption) : '-';
+}
+
+return view.extend({
+	/* No cbi form on this page, so suppress the Save/Save&Apply/Reset footer
+	   the base view would otherwise render (cf. status/processes.js). */
+	handleSaveApply: null,
+	handleSave: null,
+	handleReset: null,
+
+	handleToggle: function(port, enable) {
+		/* Mark the port as transitioning and repaint at once, so the row shows a
+		   pending state with no RPC/MCU latency. The pending entry is cleared in
+		   renderStatus once the daemon reports a status matching the intent. */
+		this.pending[port] = enable;
+		this.repaint();
+
+		return callPoeManage(port, enable).then(L.bind(function() {
+			/* The MCU takes ~1-2s to settle; poll faster than the steady 5s cadence
+			   so the pending label clears promptly. */
+			this.reconcile();
+		}, this)).catch(L.bind(function(err) {
+			delete this.pending[port];
+			this.repaint();
+			ui.addNotification(null, E('p', {},
+				_('Failed to switch PoE on %s: %s').format(port, err.message || err)), 'error');
+		}, this));
+	},
+
+	/* Re-render the status table from the last fetched info without a new RPC. */
+	repaint: function() {
+		if (this.statusNode)
+			dom.content(this.statusNode, this.renderStatus(this.lastInfo));
+	},
+
+	refreshStatus: function() {
+		return L.resolveDefault(callPoeInfo(), {}).then(L.bind(function(info) {
+			this.lastInfo = info;
+			if (this.statusNode)
+				dom.content(this.statusNode, this.renderStatus(info));
+		}, this));
+	},
+
+	/* Short burst of refreshes after a toggle to reconcile the pending state
+	   quickly, rather than waiting for the next 5s poll. */
+	reconcile: function() {
+		[ 500, 1000, 2000, 3500 ].forEach(L.bind(function(delay) {
+			window.setTimeout(L.bind(function() {
+				if (this.statusNode)
+					this.refreshStatus();
+			}, this), delay);
+		}, this));
+	},
+
+	renderStatus: function(info) {
+		info = info || {};
+		this.lastInfo = info;
+		var ports = info.ports || {};
+		var names = Object.keys(ports).sort(function(a, b) { return portNum(a) - portNum(b); });
+
+		var rows = names.map(L.bind(function(name) {
+			var p = ports[name] || {};
+			var actualOn = portIsOn(p);
+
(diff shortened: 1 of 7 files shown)


openwrt-ai (commented):

Backend coupling verified against [`Hurricos/realtek-poe` `src/main.c`](https://github.com/Hurricos/realtek-poe/blob/2bd359f858530a36ed4c63d14c4d044cd743d9b1/src/main.c) and [`files/etc/config/poe`](https://github.com/Hurricos/realtek-poe/blob/2bd359f858530a36ed4c63d14c4d044cd743d9b1/files/etc/config/poe): the per-port UCI options written by the form (`enable`, `priority`, `poe_plus`, plus `name` shown read-only) are all read by the daemon, the `manage` ubus method takes the `port`/`enable` params the view passes, and the `info` reply fields the view consumes (`budget`, `consumption`, `firmware`, and per-port `mode`, `status`, `priority`, `consumption`) all exist. `LUCI_DEPENDS:=+realtek-poe` matches the package name in the feed.

Note: the formality bot already flagged the 231-char commit-message body line; not repeating that here.

---

openwrt-ai (commented):

The only new commit since the last review (`c9db49b`) is a merge of `master` into the branch; it introduces no changes to `applications/luci-app-poe/`, so the PR's reviewed content is unchanged.

---

feckert:
Great work—I like it. I'm just wondering if it makes sense to move the "Status" section to the "Status" tab, rather than keeping it under "Configuration." :thinking:

openwrt-ai (commented):

The branch was rewritten into a single squashed commit (`bef880f`), which resolves both earlier formality failures: the merge commit (`c9db49b`) is gone, and the long commit-body line is now wrapped. The app's substantive content (`overview.js`, ACL, menu, Makefile, `.pot`) is unchanged from the previously reviewed state, and the commit message matches its diff. CI on the head SHA is still pending; nothing has failed.

---

mirko:
I'm a bit confused now, as I also considered adding PoE-functionality to luci via `/etc/config/poe` but then realised `/etc/config/poe` is being deprecated in favour of the PSE/PoE functionality in `/etc/config/network`.
See my comment here https://github.com/openwrt/netifd/pull/65#issuecomment-4526148912 and follow-ups (especially https://github.com/openwrt/netifd/pull/65#issuecomment-4526189203).
Maybe I'm misunderstanding things, though, but at least we should sync ourselves on this, I figure.

maxskokov:
@mirko raises a critical structural point, and the context in openwrt/netifd#65 makes it clear that shifting PoE management to the netifd PSE framework inside /etc/config/network is the right long-term path for OpenWrt.However, looking at the implementation here, the live telemetry and runtime port toggling driven by ubus call realtek-poe is excellent, rock-solid, and highly valuable right now. Full upstream PSE abstraction layer standardization across all legacy hardware targets will inevitably take time.It might make perfect architectural sense to decouple this valuable real-time telemetry from the legacy UCI configuration layer. If we strip out the persistent /etc/config/poe form for now to avoid introducing a deprecated configuration scheme, we could preserve the entire live consumption dashboard and move it under the Status tab (exactly as @feckert suggested). This saves the author's great frontend telemetry work while keeping the codebase compliant with the new netifd direction.

pieterhollander:
Thanks all for the feedback! I've pushed a restructured version and updated the PR description and screenshots.
The app is now split into two pages, per @feckert's suggestion.

@mirko @maxskokov — thanks for raising this.
The PSE-PD developments were new to me.
I just followed the docs and tried to build a small GUI that lets me do what I currently have to do manually through the CLI.
For the hardware this app targets, it seems PSE-PD isn't usable yet, although I did notice openwrt/openwrt#23222.

If you can agree with the new split, I'd like to open a backport PR to `openwrt-25.12` afterwards.

applications/luci-app-poe/htdocs/luci-static/resources/view/poe/status.js
```diff
@@ -0,0 +1,201 @@
...
+	render: function(info) {
+		/* pending: ports mid-toggle (name -> intended on/off); lastInfo: most
+		   recent `poe info` snapshot, used to repaint without a fresh RPC. */
+		this.pending = {};
+		this.lastInfo = info || {};
+
+		this.statusNode = E('div', {}, this.renderStatus(info));
+
+		/* keep the live status table fresh */
+		poll.add(L.bind(this.refreshStatus, this), 5);
```
efahl: Just a minor nit: Shouldn't the poll interval be derived from config's `luci.main.pollinterval` (default is 5)?  See modules/luci-base/ucode/template/header.ut for where it gets collected, and https://github.com/openwrt/luci/blob/master/modules/luci-base/htdocs/luci-static/resources/luci.js#L1093

pieterhollander: @efahl Thanks for the review, I dropped the hardcoded interval so it now follows luci.main.pollinterval.

applications/luci-app-poe/htdocs/luci-static/resources/view/poe/status.js
```diff
@@ -0,0 +1,201 @@
...
+	/* No cbi form on this page, so suppress the Save/Save&Apply/Reset footer
+	   the base view would otherwise render (cf. status/processes.js). */
+	handleSaveApply: null,
+	handleSave: null,
+	handleReset: null,
+
+	handleToggle: function(port, enable) {
+		/* Mark the port as transitioning and repaint at once, so the row shows a
+		   pending state with no RPC/MCU latency. The pending entry is cleared in
+		   statusCell once the daemon reports a status matching the intent. */
```
openwrt-ai: nit: stale comment after the split — there is no `statusCell` function. The pending entry is actually cleared in `renderStatus` (line 103). The original `overview.js` named this correctly before the move.

```suggestion
		   renderStatus once the daemon reports a status matching the intent. */
```

---

openwrt-ai (commented):
Reviewed the changes since the last review (`bef880f` → `659d662`): the app was split into separate `Status → PoE` (`status.js`) and `Network → PoE` (`overview.js`) pages, with the ACL split into `luci-app-poe` (uci read/write, for the config form) and `luci-app-poe-status` (ubus `poe info`/`manage`, for the live page), and the hardcoded 5s poll interval dropped so it follows `luci.main.pollinterval`.

The split is clean: each view's ACL grants only what it uses (overview = uci only, status = ubus only), the menu entries point at the matching ACL groups, the `.pot` line references were regenerated, and the toggle/pending/reconcile logic moved verbatim. The backend coupling against `realtek-poe` (`info`/`manage` methods and the `enable`/`priority`/`poe_plus` UCI options) is unchanged and was verified in the earlier review. Only one nit inline.

---

applications/luci-app-poe/htdocs/luci-static/resources/view/poe/status.js
pieterhollander: Applied suggestion and pushed again.

openwrt-ai (commented):
Reviewed the change since the last review (`659d662` → `66e7b05`): the only diff is the one-line fix to the `handleToggle` comment in `status.js` (`statusCell` → `renderStatus`), which is exactly the suggestion from the previous review and is now correct. No new issues found.

---

jonasjelonek:
https://github.com/openwrt/openwrt/pull/23222 has been merged. So PSE-PD is increasingly usable on Realtek switches now. At least the snapshot version of your app should rather rely on that instead of `realtek-poe`. I'm puzzled if it still makes sense to put such an effort into 25.12, I suppose we get a new version later this year which come with several improvements, especially for Realtek switches.

PR #8735, opened 2026-06-25
luci-mod-network: hide wifi modes unsupported by the driver

micpf:
## Summary

Hide wifi mode entries in the wireless UI that the underlying driver does
not advertise in `wiphy->interface_modes`.

Today the "Mode" dropdown on the radio/interface edit page lists every
mode LuCI knows about (AP, Client, Mesh, Monitor, Ad-Hoc, ...), regardless
of whether the running driver actually supports them. Selecting an
unsupported mode silently fails at netifd/hostapd time.

This series:

1. Adds a tiny rpcd ucode plugin `luci.wifi iftypes` that queries
   nl80211 `GET_WIPHY` and returns per-PHY supported `iftypes`.
2. Fetches that map once in `network.js#initNetworkState()` and stores it
   on `_state.wifi_iftypes`. Adds `Network.WifiDevice#isModeSupported(mode)`
   that maps each UCI `mode` value to the corresponding nl80211 iftype
   (`sta`/`sta-wds` → `managed`, `ap`/`ap-wds` → `ap`,
   `adhoc`/`ahdemo` → `ibss`, `mesh` → `mesh_point`, ...).
3. Gates each entry in the wireless mode dropdown on
   `radioNet.isModeSupported(...)`.

Fail-open: if the ubus call returns nothing (plugin not installed, old
rpcd, query error) `isModeSupported()` returns `true`, so the UI behaves
exactly as today.

## Motivation

Two concrete cases:

- **AP-only product variants** where the driver masks
  `NL80211_IFTYPE_STATION` out of `wiphy->interface_modes` (e.g. for
  certification reasons). Users could still pick "Client" in LuCI and
  end up with a non-functional radio. With this series, "Client" and
  the WDS-client variant disappear from the dropdown while AP / Mesh /
  Monitor remain.
- **fullmac chips** that don't advertise `monitor` or `mesh_point` no
  longer show those entries.

## Testing

- Stock build (no AP-only restriction): all existing entries still
  present, no behavior change.
- AP-only build (driver masks `NL80211_IFTYPE_STATION`):
  `ubus call luci.wifi iftypes` →
  `{"phy0":{"iftypes":["ap","monitor","mesh_point"]}, ...}`
  Mode dropdown: Client + Client (WDS) + Ad-Hoc + Pseudo Ad-Hoc are
  hidden; AP, AP (WDS), Mesh, Monitor remain.
- Removing the rpcd plugin: dropdown shows all modes again (fail-open
  verified).

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/network.js
+++ b/modules/luci-base/htdocs/luci-static/resources/network.js
@@ -69,6 +69,12 @@ const callLuciHostHints = rpc.declare({
 	expect: { '': {} }
 });
 
+const callLuciWifiIftypes = rpc.declare({
+	object: 'luci.wifi',
+	method: 'iftypes',
+	expect: { '': {} }
+});
+
 const callIwinfoAssoclist = rpc.declare({
 	object: 'iwinfo',
 	method: 'assoclist',
@@ -360,16 +366,18 @@ function initNetworkState(refresh) {
 			L.resolveDefault(callLuciBoardJSON(), {}),
 			L.resolveDefault(callLuciNetworkDevices(), {}),
 			L.resolveDefault(callLuciWirelessDevices(), {}),
+			hasWifi ? L.resolveDefault(callLuciWifiIftypes(), {}) : L.resolveDefault({}),
 			L.resolveDefault(callLuciHostHints(), {}),
 			getProtocolHandlers(),
 			L.resolveDefault(uci.load('network')),
 			hasWifi ? L.resolveDefault(uci.load('wireless')) : L.resolveDefault(),
 			L.resolveDefault(uci.load('luci'))
-		]).then(function([netifd_ifaces, board_json, luci_devs, radios, hosts]) {
+		]).then(function([netifd_ifaces, board_json, luci_devs, radios, wifi_iftypes, hosts]) {
 
 			const s = {
 				isTunnel: {}, isBridge: {}, isSwitch: {}, isWifi: {},
 				ifaces: netifd_ifaces, radios: radios, hosts: hosts,
+				wifi_iftypes: wifi_iftypes,
 				netdevs: {}, bridges: {}, switches: {}, hostapd: {}
 			};
 
@@ -3721,6 +3729,49 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */ 
 				return null;
 
 		return v;
+	},
+
+	/**
+	 * Check whether a given UCI wireless `mode` value is supported
+	 * by the underlying PHY.
+	 *
+	 * @param {string} mode
+	 * The UCI `mode` value to test.
+	 *
+	 * @returns {boolean}
+	 * Returns `true` if supported, or if support information is
+	 * unavailable; `false` only when the driver explicitly does not
+	 * advertise the matching iftype.
+	 */
+	isModeSupported: function(mode) {
+		const uciModeToIftype = {
+			ap:        'ap',
+			'ap-wds':  'ap',
+			sta:       'managed',
+			'sta-wds': 'managed',
+			adhoc:     'ibss',
+			ahdemo:    'ibss',
+			mesh:      'mesh_point',
+			monitor:   'monitor',
+			wds:       'wds'
+		};
+
+		const want = uciModeToIftype[mode];
+		if (want == null)
+			return true;
+
+		const radioname = this.ubus('radio');
+		const dev = radioname ? _state.radios[radioname] : null;
+		const phy = (dev && L.isObject(dev.iwinfo)) ? dev.iwinfo.phy : null;
+
+		if (!phy || !L.isObject(_state.wifi_iftypes))
+			return true;
+
+		const entry = _state.wifi_iftypes[phy];
+		if (!L.isObject(entry) || !Array.isArray(entry.iftypes))
+			return true;
+
+		return entry.iftypes.includes(want);
 	},
 
 	/**
--- a/modules/luci-mod-network/root/usr/share/rpcd/ucode/luci.wifi-iftypes
+++ b/modules/luci-mod-network/root/usr/share/rpcd/ucode/luci.wifi-iftypes
@@ -1,0 +1,46 @@
+#!/usr/bin/env ucode
+//
+// luci.wifi-iftypes -- expose per-PHY supported nl80211 interface modes.
+//
+//   ubus call luci.wifi iftypes
+//     -> { "phy0": { "iftypes": ["managed","ap","monitor","mesh_point"] }, ... }
+//
+// Consumed by LuCI's wireless mode UI to hide modes the driver does not
+// advertise in wiphy->interface_modes (e.g. STA disabled by the driver
+// for AP-only SKUs, or mesh/monitor unavailable on fullmac chips).
+//
+
+'use strict';
+
+import * as nl80211 from 'nl80211';
+
+function collect() {
+	let out = {};
+
+	let dumps = nl80211.request(
+		nl80211.const.NL80211_CMD_GET_WIPHY,
+		nl80211.const.NLM_F_DUMP,
+		{ split_wiphy_dump: true });
+
+	for (let d in filter(dumps ?? [], d => d.wiphy_name != null)) {
+		let phy = replace(d.wiphy_name, /^wiphy/, 'phy');
+		out[phy] ??= { iftypes: [] };
+		if (type(d.supported_iftypes) == 'object')
+			out[phy].iftypes = uniq([
+				...out[phy].iftypes,
+				...filter(keys(d.supported_iftypes), t => d.supported_iftypes[t])
+			]);
+	}
+
+	return out;
+}
+
+return {
+	'luci.wifi': {
+		iftypes: {
+			call: function() {
+				return collect();
+			}
+		}
+	}
+};
--- a/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json
+++ b/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json
@@ -37,6 +37,7 @@
 		"description": "Grant access to network status information",
 		"read": {
 			"ubus": {
+				"luci.wifi": [ "iftypes" ],
 				"luci-rpc": [ "getBoardJSON", "getHostHints", "getNetworkDevices", "getWirelessDevices" ],
 				"network": [ "get_proto_handlers" ],
 				"network.device": [ "status" ],
--- a/modules/luci-mod-network/Makefile
+++ b/modules/luci-mod-network/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=LuCI Network Administration
-LUCI_DEPENDS:=+luci-base +rpcd-mod-iwinfo +luci-lib-uqr
+LUCI_DEPENDS:=+luci-base +rpcd-mod-iwinfo +luci-lib-uqr +ucode-mod-nl80211
 
 PKG_LICENSE:=Apache-2.0
 
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -1078,9 +1078,12 @@ return view.extend({
 
 				o = ss.taboption('general', form.ListValue, 'mode', _('Mode') , !have_mesh ? '<a id="installmesh" href="%s" target="_blank" rel="noreferrer">%s</a>'
 						.format(L.url('admin/system/package-manager') + '?query=wpad-mesh', _('802.11s? Install mesh wpad') ) : '');
-				o.value('ap', _('Access Point'));
-				o.value('sta', _('Client'));
-				o.value('adhoc', _('Ad-Hoc'));
+				if (radioNet.isModeSupported('ap'))
+					o.value('ap', _('Access Point'));
+				if (radioNet.isModeSupported('sta'))
+					o.value('sta', _('Client'));
+				if (radioNet.isModeSupported('adhoc'))
+					o.value('adhoc', _('Ad-Hoc'));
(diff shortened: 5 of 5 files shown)


modules/luci-base/htdocs/luci-static/resources/network.js
```diff
@@ -3723,6 +3731,49 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */
...
+	 * unavailable; `false` only when the driver explicitly does not
+	 * advertise the matching iftype.
+	 */
+	isModeSupported: function(mode) {
+		const uciModeToIftype = {
+			ap:        'ap',
+			'ap-wds':  'ap',
+			sta:       'managed',
+			'sta-wds': 'managed',
+			adhoc:     'ibss',
```
openwrt-ai, line 3752: These map targets (`managed` for `sta`/`sta-wds`, `ibss` for `adhoc`/`ahdemo`) must exactly match the iftype key strings that the `luci.wifi iftypes` plugin emits, since `isModeSupported()` does a literal `entry.iftypes.includes(want)`. The plugin builds those keys from `keys(d.supported_iftypes)` of the `ucode-mod-nl80211` `GET_WIPHY` dump. The nl80211/ucode convention for those two iftypes is `station` and `adhoc` (the kernel enum suffixes), not `managed`/`ibss` — `managed`/`IBSS` are iw(8)/cfg80211 mode aliases, not the `NL80211_IFTYPE_*` names. If the plugin really returns `station`/`adhoc`, then `want` for `sta`/`adhoc` never matches and those modes get hidden on every driver that supports them — i.e. the feature fails *closed* for Client and Ad-Hoc, the opposite of the intended fail-open.

The commit message example shows `["managed","ap",...]`, but the PR's Testing section only shows an AP-only dump (where station is masked out, so it can't confirm the positive case). Can you paste a real `ubus call luci.wifi iftypes` from a driver that *does* advertise station/adhoc, confirming the keys are literally `managed`/`ibss`? If they come back as `station`/`adhoc`, this map (and the plugin) need to agree on one naming. `ap`, `monitor`, `mesh_point`, `wds` already match the nl80211 names and are fine.

---

modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network-wifi-iftypes.json
```diff
@@ -0,0 +1,10 @@
+{
+  "luci-mod-network-wifi-iftypes": {
+    "description": "Read supported nl80211 iftypes per PHY (used by wireless mode UI)",
+    "read": {
+      "ubus": {
+        "luci.wifi": [ "iftypes" ]
```
openwrt-ai: The `luci.wifi iftypes` call is added to `initNetworkState()` in luci-base's [`network.js`](https://github.com/openwrt/luci/blob/f4847712c9bb5b325d5fc2a84ee36447621b24f2/modules/luci-base/htdocs/luci-static/resources/network.js#L369), the shared network-state loader run by many network pages — not only the wireless view. The sibling reads it sits next to (`getWirelessDevices`, `getNetworkDevices`, …) are granted by the `luci-base-network-status` ACL group in [`luci-base.json:40`](https://github.com/openwrt/luci/blob/f4847712c9bb5b325d5fc2a84ee36447621b24f2/modules/luci-base/root/usr/share/rpcd/acl.d/luci-base.json#L40), which is the ACL effective wherever that loader runs.

Putting this grant in a brand-new standalone group (`luci-mod-network-wifi-iftypes`) shipped by luci-mod-network means it only takes effect where that group is actually attached to the session — elsewhere the call returns a permission error. It fails open so nothing breaks, but the feature silently won't work on pages that load network state without this group. Consider adding `"luci.wifi": [ "iftypes" ]` to the existing `luci-base-network-status` read block instead, alongside the other state reads this call accompanies.

---

openwrt-ai (commented):
Two inline points worth resolving before merge: (1) whether the `sta`/`adhoc` → `managed`/`ibss` iftype mapping matches the literal key strings the nl80211 dump produces (a mismatch would fail *closed* and hide Client/Ad-Hoc on supporting drivers), and (2) the ACL grant placement relative to where the `luci.wifi iftypes` call actually runs. Commit messages all match their diffs. CI had not reported any status at review time.

---

modules/luci-base/htdocs/luci-static/resources/network.js
micpf: Confirmed against the ucode binding and a live device. `ucode-mod-nl80211` (`lib/nl80211.c` lines 274–290) defines `supported_iftypes` as a nested attribute spec whose keys are the cfg80211 mode aliases (`"ibss"`, `"managed"`, …), not the `NL80211_IFTYPE_*` enum suffixes. Real output from a dual-radio device:

```
# ubus call luci.wifi iftypes
{
        "wiphy0": {
                "iftypes": [ "ap", "monitor", "mesh_point" ]
        },
        "wiphy1": {
                "iftypes": [ "managed", "ap", "monitor", "mesh_point" ]
        }
}
```

So `wiphy1` is the positive case for `managed` — `isModeSupported('sta')` will match and Client stays visible. The map in the PR is correct as-is.

modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network-wifi-iftypes.json
micpf: Good catch — agreed. Force-pushed: dropped the standalone `luci-mod-network-wifi-iftypes` ACL group and moved the grant into luci-base's `luci-base-network-status` block, alongside the other reads that accompany this call:

```diff
 		"read": {
 			"ubus": {
+				"luci.wifi": [ "iftypes" ],
 				"luci-rpc": [ "getBoardJSON", "getHostHints", "getNetworkDevices", "getWirelessDevices" ],
 				"network": [ "get_proto_handlers" ],
 				"network.device": [ "status" ],
 				"network.interface": [ "dump" ]
 			},
```

The grant now lives next to the call site's peers, so the feature is effective wherever `initNetworkState()` runs, not only on pages that pull in a luci-mod-network ACL. The ACL hunk was folded into the existing `luci-base: expose driver-reported wifi iftypes to UI` commit so each commit stays scoped to its module.

maxskokov:
The architectural pivot with moving the ACL grant into luci-base-network-status is spot on. It perfectly ensures that the lookup won't silently fail with permission errors on sibling network pages. Solid implementation of the fail-open fallback as well. LGTM!

modules/luci-base/htdocs/luci-static/resources/network.js
```diff
@@ -3723,6 +3731,49 @@ WifiNetwork = baseclass.extend(/** @lends LuCI.network.WifiNetwork.prototype */
...
+			return true;
+
+		const radioname = this.ubus('radio');
+		const dev = radioname ? _state.radios[radioname] : null;
+		const phy = (dev && L.isObject(dev.iwinfo)) ? dev.iwinfo.phy : null;
+
+		if (!phy || !L.isObject(_state.wifi_iftypes))
+			return true;
+
+		const entry = _state.wifi_iftypes[phy];
```
openwrt-ai, line 3770: The lookup key here is `phy`, taken from `dev.iwinfo.phy` (line above). That `phy` field is iwinfo's `phyname`, populated by rpcd-mod-luci at [`luci.c:1012`](https://github.com/openwrt/luci/blob/1eb634c9724e40b57bf3edea9dc2cc9295687065/libs/rpcd-mod-luci/src/luci.c#L1012) (`iw_call_str(iw->phyname, ...)`), which returns the mac80211 phy name — `phy0`, `phy1`, ….

But the `luci.wifi iftypes` plugin keys its map on `d.wiphy_name` (the nl80211 `GET_WIPHY` dump). The real-device output you pasted in the resolved thread above shows the keys coming back as `"wiphy0"` / `"wiphy1"`, not `"phy0"` / `"phy1"`. If that's the literal key shape, then `_state.wifi_iftypes["phy0"]` misses, `entry` is undefined, and the helper returns `true` for every mode — i.e. the feature silently fails open everywhere and never hides anything, which would contradict the AP-only test in the PR description (where Client/Ad-Hoc are reported as hidden).

Can you confirm the exact top-level keys returned by `ubus call luci.wifi iftypes` against the value of `iwinfo`'s `phy` field for the same radio? If `wiphy_name` is `wiphyN` while iwinfo reports `phyN`, the plugin should normalize the key (or the lookup should map between the two) so they agree.

---

openwrt-ai (commented):
One inline question on the phy-key lookup. The two prior threads (iftype string mapping `managed`/`ibss`, and the ACL grant placement) both look resolved at the current head — the grant now lives in `luci-base-network-status` and the iftype keys were confirmed against the ucode binding. The new concern is whether the iwinfo `phy` name used for the lookup matches the `wiphy_name` key the plugin emits. Commit messages match their diffs.

---

modules/luci-mod-network/root/usr/share/rpcd/ucode/luci.wifi-iftypes
```diff
@@ -0,0 +1,45 @@
...
+function collect() {
+	let out = {};
+
+	let dumps = nl80211.request(
+		nl80211.const.NL80211_CMD_GET_WIPHY,
+		nl80211.const.NLM_F_DUMP,
+		{ split_wiphy_dump: true });
+
+	for (let d in filter(dumps ?? [], d => d.wiphy_name != null)) {
+		out[d.wiphy_name] ??= { iftypes: [] };
```
maxskokov: Plugin keys on d.wiphy_name → wiphy0/wiphy1 (luci.wifi-iftypes:20), but isModeSupported() looks up by dev.iwinfo.phy → phy0/phy1. Keys never match; feature silently fails open everywhere.
Fix: normalize in the plugin (d.wiphy_name.replace(/^wiphy/, 'phy')) or in the lookup.

openwrt-ai (commented):
Re-reviewed after the force-push (3 commits rewritten since the last review).

The one substantive PR-content change since then is the plugin now normalizing its map key: `let phy = replace(d.wiphy_name, /^wiphy/, 'phy');`. The emitted keys (`phy0`, `phy1`, …) now match the `dev.iwinfo.phy` value used for the lookup in `isModeSupported()`, so the earlier phy/wiphy key-mismatch concern (raised in the prior review and by @maxskokov) is resolved. The normalization is also safe if `wiphy_name` is already `phyN` — the regex simply doesn't match.

I re-confirmed the UCI-mode→iftype map (`sta`/`sta-wds`→`managed`, `adhoc`/`ahdemo`→`ibss`, `mesh`→`mesh_point`, `ap`/`ap-wds`→`ap`, `monitor`, `wds`) against the `supported_iftypes` key names defined in ucode's nl80211 binding ([`nl80211_ifcomb_limit_types_nla`, lib/nl80211.c:275-289](https://github.com/jow-/ucode/blob/b885dd0fe1e974551fb1233ca5f3aa74d80b74d9/lib/nl80211.c#L275-L289)) — they match, so the map does not fail closed. The `+ucode-mod-nl80211` Makefile dependency also matches the package name produced by openwrt's `UcodeModule` macro.

Commit messages match their diffs. No new issues found.

---

jow-:
Shifting to ucode-nl80211 is tempting but I am somehow against introducing this feature like that (new dependency, new single purpose ubus namespace). Ideally this would either go into libiwinfo so that it is covered by the existing status procedures or it would get implemented along with a general pivot away from iwinfo towards pure nl80211.

micpf:
Thanks for the feedback — fully agree the current shape (new `ucode-nl80211` dep + a single-purpose ubus namespace) is not the right way in. Since a broader iwinfo→nl80211 pivot doesn't seem to be actively in flight, I'd like to go with your first suggestion (fold it into libiwinfo) and rework the PR accordingly. Before I invest the churn, could you sanity-check the design below?

## Proposed rework

**1. `openwrt/iwinfo`** — add a new op to `struct iwinfo_ops`:
```c
int (*iftypes)(const char *ifname, int *buf);   /* bitmask of 1 << IWINFO_OPMODE_* */
```
- Reuses the existing `IWINFO_OPMODE_*` enum and `IWINFO_OPMODE_NAMES[]` — no new naming scheme.
- `iwinfo_nl80211.c`: query `NL80211_CMD_GET_WIPHY`, walk `NL80211_ATTR_SUPPORTED_IFTYPES`, map each iftype to the corresponding `IWINFO_OPMODE_*` bit.
- `wext`/`madwifi`/`wl` backends: leave `iftypes` NULL.
- Bump `IWINFO_ABI_VERSION`.

**2. `openwrt/rpcd`** — expose it as part of the existing `iwinfo.info` response:
- New `rpc_iwinfo_call_iftypes()` mirroring `rpc_iwinfo_call_htmodes()`, using the existing `rpc_iwinfo_add_bit_array()` helper with `IWINFO_OPMODE_NAMES` / `IWINFO_OPMODE_COUNT`.
- Emits `"iftypes": ["ap","monitor","managed", ...]` inside `iwinfo.info` — **no new ubus method, no new namespace**.

**3. `openwrt/luci`** (this PR, force-pushed to a single commit):
- Drop the `luci.wifi iftypes` rpcd plugin, its ACL, and the `ucode-nl80211` dep.
- Read `iftypes` from the per-radio `iwinfo.info` result LuCI already fetches; attach to the radio state.
- `WifiDevice#isModeSupported(mode)` maps UCI mode → `IWINFO_OPMODE_*` string and checks membership.
- `wireless.js` gates the mode dropdown on `radioNet.isModeSupported(...)` (unchanged from current PR).
- Fail-open preserved: if `iftypes` is absent from the info blob (older rpcd/iwinfo, non-nl80211 backend), `isModeSupported()` returns `true`.

## Rollout

Merge order iwinfo → rpcd → luci; luci degrades gracefully to fail-open in the meantime.

Does this match what you had in mind, or would you prefer a different shape (e.g. a dedicated `iwinfo.iftypes` ubus sub-method, or exposing it purely as a new libiwinfo call without threading it through `iwinfo.info`)? Happy to adjust before I start.

PR #8738, opened 2026-06-26
Alternative view for configuring switch ports and VLANs: CSS fixes for themes other than bootstrap

mirko:
Parent-PR: https://github.com/openwrt/luci/pull/8639

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
@@ -41,6 +41,44 @@
 	--svc-vlan-tagged-bg: rgba(217, 119, 6, 0.14);
 	--svc-vlan-untagged-stripe: rgba(13, 148, 136, 0.2);
 	--svc-vlan-tagged-stripe: rgba(217, 119, 6, 0.2);
+	--svc-row-bg: var(--background-color-medium, #fafafa);
+	--svc-row-hover-bg: var(--background-color-high, #f3f3f3);
+	--svc-row-add-bg: #f7faff;
+}
+
+/* Theme overrides: several LuCI themes set large min-widths on text inputs
+ * (e.g. openwrt-2020: 20rem, material: 25rem in .cbi-value) and global
+ * button/input resets (openwrt-2020 zeros borders on *). Scope fixes here. */
+#switch-vlan-view input[type="text"],
+#switch-vlan-view input[type="number"] {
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css
```diff
@@ -376,26 +412,17 @@
...
 }
 
 .svc-vlan-label {
-	width: 100%;
-	box-sizing: border-box;
 	padding: 2px 4px;
-	border: 1px solid #ddd;
-	border-radius: 2px;
-	background: var(--background-color-high, #fff);
 }
```
openwrt-ai, line 424: This drops `width: 100%`, `max-width: 100%`, `min-width: 0` and `box-sizing: border-box` from `.svc-vlan-id-input`, relying on the new generic rule added at the top of this file to restore them. But that rule is scoped to `#switch-vlan-view input[type="text"]`, and the VLAN-ID input is rendered as `type="number"`, not `type="text"` — see [`vidInput` in switch-vlan.js:541-543](``https://github.com/openwrt/luci/blob/77bfa64ab3f4804eb3c61aecb18848f23b5020f5/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.js#L541-L543).`` So the number input no longer gets these sizing properties and falls back to the UA default width inside the 6em ID cell, which is exactly the kind of overflow this PR is trying to fix on other themes.

Either restore the sizing on the class directly:

```suggestion
.svc-vlan-id-input {
	width: 100%;
	max-width: 100%;
	min-width: 0;
	box-sizing: border-box;
	padding: 2px 4px;
	font-family: monospace;
}
```

or broaden the generic selector to also cover `input[type="number"]`.

---

openwrt-ai (commented):
Reviewed 1 new commit. One inline issue: removing the sizing properties from `.svc-vlan-id-input` while the replacement rule only targets `input[type="text"]` leaves the `type="number"` VLAN-ID input without width/box-sizing.

---

openwrt-ai (commented):

The prior review's concern (the `type="number"` VLAN-ID input losing its sizing because the generic rule only matched `input[type="text"]`) is resolved here: the selector now also covers `#switch-vlan-view input[type="number"]` for both the base and `:focus` rules.

---

efahl:
Looks good to me.

I downloaded and applied that last rev to HEAD, works fine on my x86/64 test machine.

PR #8740, opened 2026-06-27
luci-app-tailscale-community: sync upstream changes

Tokisaki-Galaxy:
### Description
- Add log viewing functionality
- Add disable firewall configuration option
- Update init script with disable_fw_config support
- Add get_logs RPC method
- Update ACL permissions
- Update i18n translations (zh_Hans, zh_Hant)
- Update translation template

---

## Tested on
test finish, all func work normal

---

## Checklist

### DIFF
--- a/applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
+++ b/applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
@@ -12,6 +12,7 @@ const callDoLogin = rpc.declare({ object: 'tailscale', method: 'do_login', param
 const callDoLogout = rpc.declare({ object: 'tailscale', method: 'do_logout' });
 const callGetSubroutes = rpc.declare({ object: 'tailscale', method: 'get_subroutes' });
 const callSetupFirewall = rpc.declare({ object: 'tailscale', method: 'setup_firewall' });
+const callGetLogs = rpc.declare({ object: 'tailscale', method: 'get_logs' });
 let map;
 
 const tailscaleSettingsConf = [
@@ -25,6 +26,7 @@ const tailscaleSettingsConf = [
 	[form.Flag, 'shields_up', _('Shields Up'), _('When enabled, blocks all inbound connections from the Tailscale network.'), { rmempty: false }],
 	[form.Flag, 'ssh', _('Enable Tailscale SSH'), _('Allow connecting to this device through the SSH function of Tailscale.'), { rmempty: false }],
 	[form.ListValue, 'dns_mode', _('DNS Mode'), _('Controls how Tailscale DNS is handled.')+'<br>'+_('Disabled: system DNS only.')+'<br>'+_('MagicDNS: Tailscale overrides resolv.conf.')+'<br>'+_('OpenWrt Forward: MagicDNS via dnsmasq forwarding.(Only support ts.net)'), { values: [['disabled', _('Disabled')], ['magicdns', 'MagicDNS'], ['openwrt_forward', _('OpenWrt Forward')]], rmempty: false }],
+	[form.Flag, 'disable_fw_config', _('Disable Firewall Configuration'), _('Disable Tailscale netfilter auto-configuration (--netfilter-mode=off).'), { rmempty: false }],
 	[form.Flag, 'enable_relay', _('Enable Peer Relay'), _('Enable this device as a Peer Relay server. Requires a public IP and an UDP port open on the router.'), { rmempty: false }]
 ];
 
@@ -239,6 +241,20 @@ function renderStatus(status) {
 	return statusTable;
 }
 
+function renderLogs(logs_data) {
+	if (!logs_data || !logs_data.logs || logs_data.logs.length === 0) {
+		return E('em', {}, _('No tailscale-related logs found.'));
+	}
+
+	const lines = logs_data.logs.map(function(line) {
+		return E('div', { 'style': 'white-space: pre; font-family: monospace; font-size: 13px; line-height: 1.5;' }, line);
+	});
+
+	return E('div', {
+		'style': 'max-height: 500px; overflow-y: auto; background: #f5f5f5; border: 1px solid #ccc; padding: 8px; border-radius: 3px;'
+	}, lines);
+}
+
 function renderDevices(status) {
 	if (!status || !status.hasOwnProperty('status')) {
 		return E('em', {}, _('Collecting data ...'));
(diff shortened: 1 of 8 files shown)


applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -161,7 +161,7 @@ methods.do_login = {
 
 		// Run the command in the background using /bin/sh -c to handle the '&' correctly
 		let login_cmd = 'tailscale login '+join(' ', loginargs);
-		popen('/bin/sh -c ' + shell_quote(login_cmd + ' &'), 'r');
+		popen('/bin/sh -c "' + login_cmd + ' &"', 'r');
```
openwrt-ai: This change reintroduces a shell-injection vector. The old form wrapped the whole command in `shell_quote(...)`, i.e. a single-quoted string, which is injection-safe because `shell_quote` escapes embedded single quotes. The new form wraps `login_cmd` in literal double quotes with no escaping. `login_cmd` embeds `loginserver` / `loginserver_authkey` which were single-quoted by `shell_quote`, but those single quotes are inert inside the outer `"..."`, and any `"`, `$`, `` ` `` or `\` in those operator-supplied values now breaks out of the double-quoted context and is interpreted by the shell. Keep the single-quote wrapping:

```suggestion
		popen('/bin/sh -c ' + shell_quote(login_cmd + ' &'), 'r');
```

---

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -357,4 +363,17 @@ methods.setup_firewall = {
 	}
 };
 
+methods.get_logs = {
+	args: { lines: 200 },
+	call: function(request) {
+		let lines = request?.args?.lines || 200;
+		let cmd = 'logread -l ' + lines + ' 2>/dev/null | grep -i -E "tailscale" || true';
```
openwrt-ai, line 366: `lines` comes straight from the RPC request (`request?.args?.lines`) and is interpolated into a shell command string. A client (any caller with the `get_logs` ACL) can pass a non-numeric value such as `"1; reboot #"` and it will be executed by the shell. Coerce to an integer before building the command:

```suggestion
		let lines = int(request?.args?.lines) || 200;
		let cmd = 'logread -l ' + lines + ' 2>/dev/null | grep -i -E "tailscale" || true';
```

---

applications/luci-app-tailscale-community/root/usr/share/rpcd/ucode/tailscale.uc
```diff
@@ -232,6 +232,12 @@ methods.get_subroutes = {
 methods.setup_firewall = {
 	call: function() {
 		try {
+			uci.load('tailscale');
+			//let disable_fw = uci.get('tailscale', 'settings', 'disable_fw_config') || '0';
+			//if (disable_fw == '1') {
+			//	return { success: true, skipped: true, message: 'Firewall auto-configuration is disabled.' };
+			//}
```
openwrt-ai: The `disable_fw_config` guard here is commented out, so `setup_firewall` still unconditionally creates the network interface, firewall zone and forwarding rules even when the user enables "Disable Firewall Configuration". `callSetupFirewall()` is invoked from the view ([tailscale.js#L451](``https://github.com/openwrt/luci/blob/2265b9f0ded846135917b9ad115c943390b1b9b6/applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js#L451)),`` so the only effect the new option actually has is `--netfilter-mode=off` in the init script. The leftover `uci.load('tailscale')` on the line above is then dead too. Either activate the guard (uncomment it) or drop the dead `uci.load`/comment block so the option does what its UI label promises.

---

applications/luci-app-tailscale-community/htdocs/luci-static/resources/view/tailscale.js
```diff
@@ -25,6 +26,7 @@ const tailscaleSettingsConf = [
 	[form.Flag, 'shields_up', _('Shields Up'), _('When enabled, blocks all inbound connections from the Tailscale network.'), { rmempty: false }],
 	[form.Flag, 'ssh', _('Enable Tailscale SSH'), _('Allow connecting to this device through the SSH function of Tailscale.'), { rmempty: false }],
 	[form.ListValue, 'dns_mode', _('DNS Mode'), _('Controls how Tailscale DNS is handled.')+'<br>'+_('Disabled: system DNS only.')+'<br>'+_('MagicDNS: Tailscale overrides resolv.conf.')+'<br>'+_('OpenWrt Forward: MagicDNS via dnsmasq forwarding.(Only support ts.net)'), { values: [['disabled', _('Disabled')], ['magicdns', 'MagicDNS'], ['openwrt_forward', _('OpenWrt Forward')]], rmempty: false }],
+	[form.Flag, 'disable_fw_config', _('Disable Firewall Configuration'), _('Completely disable all Tailscale firewall auto-configuration, including zone creation, forwarding rules, and netfilter settings.'), { rmempty: false }],
```
openwrt-ai: nit: this description promises "Completely disable all Tailscale firewall auto-configuration, including zone creation, forwarding rules, and netfilter settings", but as implemented the option only adds `--netfilter-mode=off` in the init script — the zone/forwarding creation in `setup_firewall` is not actually skipped (the guard there is commented out). Either wire up the guard or align this text with what the flag really does, so users aren't misled into thinking the zone won't be created.

---

openwrt-ai (commented):
Reviewed 1 new commit. The functional changes look mostly fine, but the new `disable_fw_config` option doesn't fully do what its label claims (the `setup_firewall` skip is commented out), and the `do_login` popen change plus the new `get_logs` method introduce shell-injection issues on operator-supplied input. Details inline. The Signed-off-by / committer-name failures are already covered by the formality bot, so I haven't repeated them here.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push amend since the last review); no new issues found.

The three concerns from the previous review have all been addressed in this revision:
- `do_login` now wraps the backgrounded command with `shell_quote(...)` instead of bare string interpolation into `/bin/sh -c "..."`, closing the shell-injection vector on the login server / auth-key inputs.
- `get_logs` now coerces the `lines` argument with `int(...)`, so non-numeric input can no longer reach the `logread` command line.
- `disable_fw_config` was re-scoped: the misleading "completely disable all firewall auto-configuration" label was narrowed to "Disable Tailscale netfilter auto-configuration (--netfilter-mode=off)", which matches what the init script actually does (`--netfilter-mode=off` when the flag is set), and the dead commented-out skip block in `setup_firewall` was removed.

The `.po`/`.pot` updates correctly retire the old msgid and add the new string. The remaining committer real-name formality check is handled by the formality bot and out of scope here.

---

PR #8743, opened 2026-06-27
luci-base: fix empty data-page attribute and login redirect on root URL

maxskokov:
### Description
When accessing LuCI via /cgi-bin/luci/ without a full path,
ctx.request_path is empty because firstchild resolution only
populates ctx.path. This causes two issues:

1. After login, http.redirect() sends the user back to the root
   URL instead of the resolved overview page (dispatcher.uc).
2. The data-page attribute in the HTML body is empty, breaking
   any CSS or JS that relies on it (header.ut in all themes).

Fix: fall back to ctx.path when ctx.request_path is empty in
both the redirect and the data-page attribute.

luci-theme-material already uses ctx.path correctly and is
unaffected.

Fixes openwrt/luci#8534.

### Maintainer _(preferred)_
@jow-

---

## Tested on
Verified by code inspection. luci-theme-material already
uses ctx.path and serves as reference implementation.

---

## Checklist
- [x] _(Optional)_ Includes what Issue it closes openwrt/luci#8534

### DIFF
--- a/modules/luci-base/ucode/dispatcher.uc
+++ b/modules/luci-base/ucode/dispatcher.uc
@@ -1044,7 +1044,7 @@ dispatch = function(_http, path) {
 				    cookie_secure = (http.getenv('HTTPS') == 'on') ? '; secure' : '';
 
 				http.header('Set-Cookie', `${cookie_name}=${session.sid}; path=${build_url()}; SameSite=strict; HttpOnly${cookie_secure}`);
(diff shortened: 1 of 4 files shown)


openwrt-ai (commented):
Reviewed 4 new commits; no new issues found.

---

jow-:
Merged, thanks!

PR #8746, opened 2026-06-28
luci-mod-network: fix DNS forward validation rejecting interface-only spec

### LINKED ISSUE #7888: luci-mod-network: 1.2.3.4@iface in DNS Forwards results in "Expecting: valid IP address"

### screenshots or captures

Image

### Actual behaviour

- get an error message in the UI
- enter the same thing in /etc/config/dhcp manually, execute `service dnsmasq reload`
- dnsmask doesn't crash and seems to be working properly

### Expected behaviour

- don't get any errors and be able to save and apply it

### Steps to reproduce

- go to network -> dhcp and dns -> forwards
- enter 1.2.3.4@iface in DNS Forwards field

### Additional Information

```shell
"release": {
		"distribution": "OpenWrt",
		"version": "24.10.2",
		"revision": "r28739-d9340319c6",
		"target": "x86/64",
		"description": "OpenWrt 24.10.2 r28739-d9340319c6",
		"builddate": "1750711236"
	}
```

Firefox

### Relevant log output

```shell

```

maxskokov:
### Description
dnsmasq supports server=1.2.3.4@eth0 (server IP bound to interface without source IP), but the regex in validateServerSpec() required a source IP address before the interface name, causing valid entries to fail with 'Expecting: valid IP address'.

Allow @interface without preceding @source in the server spec.

Fixes openwrt/luci#7888.

### Maintainer _(preferred)_
@jow-

---

## Tested on
Verified by code inspection against dnsmasq documentation.
---

## Checklist
- [x] _(Optional)_ Includes what Issue it closes openwrt/luci#7888.

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js
@@ -64,19 +64,19 @@ function validateServerSpec(sid, s) {
 	if (m[2] == '' || m[2] == '#')
 		return true;
 
-	// ipaddr%scopeid#srvport@source@interface#srcport
+	// ipaddr%scopeid#srvport[@source[@interface]]#srcport or ipaddr%scopeid#srvport@interface
 
-	m = m[2].match(/^([0-9a-f:.]+)(?:%[^#@]+)?(?:#(\d+))?(?:@([0-9a-f:.]+)(?:@[^#]+)?(?:#(\d+))?)?$/);
+	m = m[2].match(/^([0-9a-f:.]+)(?:%[^#@]+)?(?:#(\d+))?(?:@([0-9a-f:.]+)(?:@[^#]+)?(?:#(\d+))?|@[^#@0-9][^#@]*)?$/);
 
 	if (!m)
 		return _('Expecting: %s').format(_('valid IP address'));
 
 	if (validation.parseIPv4(m[1])) {
-		if (m[3] != null && !validation.parseIPv4(m[3]))
+		if (m[3] != null && !validation.parseIPv4(m[3]) && !validation.parseIPv6(m[3]))
 			return _('Expecting: %s').format(_('valid IPv4 address'));
 	}
 	else if (validation.parseIPv6(m[1])) {
-		if (m[3] != null && !validation.parseIPv6(m[3]))
+		if (m[3] != null && !validation.parseIPv4(m[3]) && !validation.parseIPv6(m[3]))
 			return _('Expecting: %s').format(_('valid IPv6 address'));
 	}
 	else {


modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js
```diff
@@ -64,9 +64,9 @@ function validateServerSpec(sid, s) {
 	if (m[2] == '' || m[2] == '#')
 		return true;
 
-	// ipaddr%scopeid#srvport@source@interface#srcport
+	// ipaddr%scopeid#srvport[@source[@interface]]#srcport or ipaddr%scopeid#srvport@interface
 
-	m = m[2].match(/^([0-9a-f:.]+)(?:%[^#@]+)?(?:#(\d+))?(?:@([0-9a-f:.]+)(?:@[^#]+)?(?:#(\d+))?)?$/);
+	m = m[2].match(/^([0-9a-f:.]+)(?:%[^#@]+)?(?:#(\d+))?(?:@([0-9a-f:.]+)(?:@[^#]+)?(?:#(\d+))?|@[^#@0-9][^#@]*)?$/);
```
openwrt-ai, line 69: The new interface-only alternative is only reached when the first alternative `@([0-9a-f:.]+)...` fails to match. Because that source-IP branch accepts any string of `[0-9a-f:.]`, an interface name composed solely of hex characters is still captured into the source-IP group (`m[3]`) rather than falling through to the new `@[^#@0-9][^#@]*` branch. For example `server=1.2.3.4@ace` parses with `m[3]="ace"`, then [the source-address check at dns.js:74-84](``https://github.com/openwrt/luci/blob/f6ada0e67a37baa68dc3c59ec1c2271bad7ed8c4/modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js#L74-L84)`` runs `parseIPv4("ace")`, which fails and still rejects the spec with "valid IPv4 address". Common names (`wan`, `lan`, `eth0`, `br-lan`) contain non-hex letters and are unaffected, so this is an edge case, but it's the same class of false rejection this PR sets out to fix. Is leaving pure-hex interface names unsupported intentional?

nit: an interface name starting with a digit (e.g. `1.2.3.4@2eth`) also fails to match, since the new branch's first char class `[^#@0-9]` excludes digits.

---

openwrt-ai (commented):
Reviewed 1 new commit. The regex change correctly fixes the primary reported case (`server=1.2.3.4@eth0`), and the commit message accurately describes the diff. One edge-case observation left inline regarding interface names that collide with the source-IP character class.

---

modules/luci-mod-network/htdocs/luci-static/resources/view/network/dns.js
maxskokov: Good catch. Fixed by also allowing m[3] to be a non-IP string (interface name): if it parses as neither IPv4 nor IPv6, the source-address check is skipped and the value is treated as an interface name. Pure-hex interface names like ace are now handled correctly.

openwrt-ai (commented):

The follow-up change (adding `!validation.parseIPv6(m[3])` to both source-address checks) correctly resolves the pure-hex interface-name case raised previously, since `parseIPv6` accepts hex-only strings and so the source check is now skipped for non-IP `m[3]` values. The previously noted nit about interface names beginning with a digit (e.g. `1.2.3.4@2eth`, excluded by the new branch's leading `[^#@0-9]` class) still applies but was already raised and is left to your discretion.

---

PR #8759, opened 2026-06-29
github: consolidate PR source branch check

BKPepe:
Remove the redundant `check_source_branch_is_not_master.yml` workflow and enable `check_branch` in HyperStickler formal workflow to check the source branch name during pull request verification.

This saves GitHub Action runner resources by consolidating the check into the existing HyperStickler job instead of spawning a separate workflow run.

_____

During https://github.com/openwrt/luci/pull/8473 it was decided to disable it by default, but I would like to remove the specific file that handles this check and use HyperStickler from @GeorgeSapkin for this purpose now. Running check_source_branch_is_not_master spawns an extra runner, which is wasteful since the formality job is already being executed. The fewer files we have to maintain, the better.

Later on, I am going to replace formal.yml with a webhook, as is done in the packages repository.


BKPepe:
See e.g. https://github.com/openwrt/luci/actions/runs/27506819451 — the total duration shows "2h 42m 7s". Yeah, it was waiting for a runner, but the run itself took less than 1 minute.

openwrt-ai (commented):

---

BKPepe:
@hnyman Can we merge this one, please?

BKPepe:
Thanks!

Issue #8749, opened 2026-06-28
Discussion: safe by default formatter


jow-:
Hi,

various recently reported security issues where instances of XSS caused by interpolating untrusted external input via `format()`'s `%s` pattern or passing bare values to `cbi_update_table()` which end up being treated as .innerHTML.

I am thinking about ways to introduce a safer by default behavior and consider the following changes:
- make `%s` escape by default (essentially making it an alias of `%h`) and introduce a complementary `%r` (raw) to opt-in into unescaped output (e.g. when some sort of HTML was already preformatted elsewhere)
- auto-escape scalar values passed as cell values to `cbi_update_table()` via `'%h'.format(...)` - user's wanting to explicitly inject HTML fragments may use `E(str)` then to opt out of the escaping.

What do you think? I anticipate some fallout when introducing it (mostly cosmetic hopefully) but this will likely be better in the long run than playing whack-a-mole with the various XSS instances popping up.

efahl:
Yeah, this centralization is way better than trying to fix all those localized instances.  Plus even if you could fix all of them right now, if you check back in a year, some more will have sneaked back in...

maxskokov:
Dug a bit deeper into how `cbi_update_table` actually handles cell values — the problem goes further than call sites.

`UITable.update()` passes `disp_val` straight into `E()`, and `dom.append()` in `luci.js` has two paths:

```js
// array of children → createTextNode, safe
node.appendChild(document.createTextNode(`${children[i]}`));

// scalar string → innerHTML, not safe
node.innerHTML = `${children}`;
```

So any scalar string cell value ends up as `innerHTML` regardless of whether it came through `%s` or `%h`. Making `%s` safer helps at the format level but doesn't close the vector at the DOM level.

Concrete examples already in tree passed to `cbi_update_table` as raw HTML strings:
- `iptables.js:168` — `'<span class="target">%s</span>'.format(target)` — notably `comment` on line 175 gets `%h` but `target`/`indev`/`outdev` don't
- `40_dhcp.js:215` — `lease.ip6addrs.join('<br />')`

The structural fix would be wrapping scalar string `disp_val` in `document.createTextNode()` inside `UITable.update()` itself, with `E()` as the explicit opt-in for HTML. Matches what you proposed for the opt-out direction.

maxskokov:
[...]

PR #8762, opened 2026-06-30
Fix iptables xss target iface

maxskokov:
### Description
`target`, `indev` and `outdev` entries in the iptables status view were interpolated into raw HTML strings via `%s`.format(), causing the values to be assigned via `innerHTML` when rendered by `cbi_update_table()`.

Both fields ultimately come from `iptables -L` output and aren't sanitized upstream, so a crafted chain or interface name could inject markup into the status page.

Replace the HTML string interpolation with `E('span', ...)`, which appends the value via `createTextNode()` instead of `innerHTML` — matching the idiomatic pattern already used for `ifacebadge` spans elsewhere (`routes.js`, `routesj.js`, `wireless.js`).

Loosely related to the discussion in #8749 about making `cbi_update_table()` safe by default — this is a self-contained fix for one concrete instance of the underlying pattern.

### Maintainer _(preferred)_
@jow-

---

## Tested on
Verified by code inspection. `E()` uses `createTextNode()` for scalar string children per `dom.append()` in `luci.js`, confirmed against existing `ifacebadge` usage in the same file (line 204) and elsewhere in the codebase.

---

## Checklist

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/iptables.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/iptables.js
@@ -165,10 +165,10 @@ return view.extend({
 						'data-format': '%.2mB',
 						'data-value': bytes
 					}, (raw ? '%d' : '%.2mB').format(bytes)),
-					target ? '<span class="target">%s</span>'.format(target) : '-',
+					target ? E('span', { 'class': 'target' }, [ target ]) : '-',
 					proto,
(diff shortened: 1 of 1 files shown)


jow-:
~~You can also use `E([ value ])` to create a fragment node, this way you do not need to introduce wrapper elements which might affect styling.~~

Edit: Sorry, I didn't look at the actual diff when I wrote this, you're replacing an existing `<span>` so nevermind.

openwrt-ai (commented):

---

PR #8763, opened 2026-06-30
luci-proto-openvpn: fix auth_user_pass option, clarify label

egc112:
Maintainer: @systemcrash

CC: @feckert

Version: Main
Compile tested: X86-64
Run tested: X86-64

Description:
The push_opt tab incorrectly registered a server-side 'auth_user_pass' Value option. So move it to the client-side FileUpload option on the general tab.

auth_user_pass is a client-only option; the corresponding server-side directive is auth-user-pass-verify, which is already correctly defined under the scripts tab as auth_user_pass_verify. Comment out the erroneous push_opt duplicate accordingly.

Also updated the general-tab auth_user_pass label to a clearer description of the expected file format (username and password on separate lines).

Signed-off-by: Erik Conijn <egc112@msn.com>

---

---

## Checklist

### DIFF
--- a/protocols/luci-proto-openvpn/htdocs/luci-static/resources/protocol/openvpn.js
+++ b/protocols/luci-proto-openvpn/htdocs/luci-static/resources/protocol/openvpn.js
@@ -73,6 +73,14 @@ const openvpnOptions = [
 	},
 	{
 		tab: 'general',
+		type: form.FileUpload,
+		root_directory: '/etc/openvpn',
+		name: 'auth_user_pass',
+		label: _('Username and password file (one per line)'),
+		placeholder: '/etc/openvpn/some-client-auth-user-pass'
+	},
+	{
+		tab: 'general',
 		type: form.Value,
 		name: 'ifconfig',
 		datatype: 'tuple(ipaddr,ipaddr)',
@@ -1471,13 +1479,6 @@ const openvpnOptions = [
 		name: 'pull',
 		label: _('Accept options pushed from server'),
 		default: 0
-	},
-	{
-		tab: 'push_opt',
-		depends: { server: "", "!reverse": true },
-		type: form.Value,
-		name: 'auth_user_pass',
-		label: _('Authenticate using username/password')
 	},
 	{
 		tab: 'push_opt',


protocols/luci-proto-openvpn/htdocs/luci-static/resources/protocol/openvpn.js
```diff
@@ -71,6 +71,14 @@ const openvpnOptions = [
 		label: _('Local private key'),
 		placeholder: '/etc/easy-rsa/keys/some-client.key'
 	},
+	{
+		tab: 'general',
+		type: form.FileUpload,
+		root_directory: '/etc/openvpn',
+		name: 'auth_user_pass',
+		label: _('Username and password file (one per line)'),
+		placeholder: '/etc/openvpn/some-client-auth-user-pass.pem'
```
openwrt-ai: nit: the `.pem` extension is misleading here. The `auth-user-pass` file is a plaintext file holding the username on line 1 and the password on line 2 (as your new label says, "one per line"), not a PEM-encoded certificate/key. The other `form.FileUpload` placeholders in this file use the real on-disk extension (e.g. `secret.key`, `tlsauth.key`); a credentials file has no conventional extension. Consider dropping the `.pem` so users aren't led to upload a cert/key:

```suggestion
		placeholder: '/etc/openvpn/some-client-auth-user-pass'
```

---

openwrt-ai (commented):
Reviewed 1 new commit; no blocking issues found. The fix is sound: the active `auth_user_pass` Value on the `push_opt` tab is correctly commented out (the `/* ... */` cleanly wraps the whole object including its trailing comma, leaving valid array syntax), and the option is re-added as a client-side `form.FileUpload` on the general tab, consistent with the sibling `cert`/`key`/`ca` entries. The commit message matches the diff. One inline nit on the placeholder extension.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The only change since the last review applies the previous inline nit — the placeholder no longer carries the misleading `.pem` extension.

---

protocols/luci-proto-openvpn/htdocs/luci-static/resources/protocol/openvpn.js
```diff
@@ -1472,13 +1480,15 @@ const openvpnOptions = [
 		label: _('Accept options pushed from server'),
 		default: 0
 	},
+	/* //auth_user_pass is a client option, the corresponding server side option is auth-user-pass-verify
```
systemcrash: Since this option presence is about handling server pushed options, should it just be `auth_user_pass_verify`?

egc112: No that is a different option.
Both should be present, 
--auth-user-pass-verify is the server side option to verify the name/password pushed by the client.
--auth-user-pass is the client side option. 
From the openvpn 2.7 manual:
```
--auth-user-pass
Authenticate with server using username/password.

Valid syntaxes:
auth-user-pass
auth-user-pass up

If up is present, it must be a file containing username/password on 2 lines or a flag named username-only to indicate no password should be prompted for. In the former case, if the password line is missing in the file, OpenVPN will prompt for one.

If up is omitted, username/password will be prompted from the console.

This option can also be inlined
<auth-user-pass>
username
[password]
</auth-user-pass>
where password is optional, and will be prompted from the console if missing.
```

I am using it as we speak

egc112: I can remove the wrongly placed auth-user-pass which I have now only commented out if you wish

egc112: I removed the commented out code which cleans things up.
In the end it is just moving this option.
It was misplaced as a server side option but it is a client side option so moved it over there.

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The only change since the last review removes the previously commented-out `push_opt` `auth_user_pass` block entirely, so the net effect is a clean move of the option from a server-side `form.Value` on the `push_opt` tab to a client-side `form.FileUpload` on the `general` tab, consistent with the sibling `ca`/`cert`/`key` entries. The commit message still matches the diff.

---

egc112:
ping @systemcrash , @feckert 

I hope you have time to review this PR.
At this moment we cannot use username/password on an OpenVPN client as the option is under the server options and thus not exposed on a client setup.
My patch is just a clean move of `auth-user-pass` from server options to client options. 
It does not belong to the server side options, the corresponding server side option is `auth-user-pass-verify` and is present as server option.

systemcrash:
OK - merged. Let's see how this works out :)

Thanks @egc112

egc112:
Thanks!

PR #8766, opened 2026-07-01
luci-app-ddns: replace dynamic_dns_lucihelper call on handleStopDDnsRule

feckert:
### Description
@champtar This should fix https://github.com/openwrt/packages/security/advisories/GHSA-hfpc-hw28-69pj

However, this change only works if the procd change is used to start the service. 
This change is included in [master](https://github.com/openwrt/packages/commit/b1502c453e50aac26ada49dd87300eb6c4449e36) and [openwrt-25.12](https://github.com/openwrt/packages/commit/04f0c5bccdeec03c1a2aabdb783648f842087dc4)

Also, I'm no longer satisfied with the app. The app needs a little love. Please try it again with a different setup.

## Checklist
- [ x ] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch. :white_check_mark:
- [ x ] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`).
- [ x ] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages.

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -220,8 +220,8 @@ return view.extend({
 			const stop = rows[i].querySelector('.cbi-section-actions .stop');
 			const cfg_enabled = uci.get('ddns', section_id, 'enabled');
 
-			reload.disabled = (status['_enabled'] == 0 || cfg_enabled == 0);
-			stop.disabled = (!service[section_id].pid);
+			reload.disabled = (cfg_enabled == 0)
+			stop.disabled = !(service[section_id] && service[section_id].pid);
 
 			const host = uci.get('ddns', section_id, 'lookup_host') || _('Configuration Error');
 			const ip = service[section_id]?.ip || _('No Data');
@@ -555,11 +555,9 @@ return view.extend({
 					'title': _('Stop this service'),
 				};
 
-			if (status['_enabled'] == 0 || cfg_enabled == 0)
+			if (cfg_enabled == 0)
 				reload_opt['disabled'] = 'disabled';
-
-			if (!resolved[section_id] || !resolved[section_id].pid ||
-					(resolved[section_id].pid && cfg_enabled == '1'))
+			if (!(resolved[section_id] && resolved[section_id].pid))
 				stop_opt['disabled'] = 'disabled';
 
 			dom.content(tdEl.lastChild, [
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -164,8 +164,7 @@ return view.extend({
 	},
 
 	handleReloadDDnsRule(m, section_id, ev) {
-		return fs.exec('/etc/init.d/ddns',
-							[ 'restart', section_id ])
+		return fs.exec('/etc/init.d/ddns', [ 'restart', section_id ])
 			.then(L.bind(m.load, m))
 			.then(L.bind(m.render, m))
 			.catch(function(e) { ui.addNotification(null, E('p', e.message)) });
@@ -179,7 +178,7 @@ return view.extend({
 	},
 
 	handleToggleDDns(m, ev) {
-        let action = this.status['_enabled'];
+		let action = this.status['_enabled'];
 		return this.callInitAction('ddns', action ? 'disable' : 'enable')
 			.then(L.bind(function () { return this.callInitAction('ddns', action ? 'stop' : 'start')}, this))
 			.then(L.bind(m.render, m))
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -172,8 +172,8 @@ return view.extend({
 	},
 
 	handleStopDDnsRule(m, section_id, ev) {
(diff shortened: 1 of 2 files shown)


champtar:
Thanks @feckert, on my phone so hard to review your change, I've pinged the initial reporter for review.

> Also, I'm no longer satisfied with the app. The app needs a little love. Please try it again with a different setup.

You mean you want to stop maintaining it ? Or you have improvement ideas but no time ?

champtar:
CI is not happy with the empty commit body

feckert:
> CI is not happy with the empty commit body

That should work now.

> You mean you want to stop maintaining it ? Or you have improvement ideas but no time ?

Well, I didn't create the app. I just took it over, because I'm also the maintainer of the ddns-script package (which I'm not happy about either), but no one else is doing it. 
The problem is simply time

applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
```diff
@@ -220,7 +220,7 @@ return view.extend({
 			const stop = rows[i].querySelector('.cbi-section-actions .stop');
 			const cfg_enabled = uci.get('ddns', section_id, 'enabled');
 
-			reload.disabled = (status['_enabled'] == 0 || cfg_enabled == 0);
+			reload.disabled = ((cfg_enabled == 0) || (service[section_id].pid));
```
openwrt-ai: The Reload button now becomes disabled whenever the service is running (`service[section_id].pid` truthy), but the static [`renderRowActions` at line 558](``https://github.com/openwrt/luci/blob/348ea9d511239027409269f610902bc0eb9005db/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L558)`` only disables it on `cfg_enabled == 0`. So on initial render Reload is enabled for a running service, then the first poll flips it to disabled — the two code paths now disagree. Reload is normally the most useful action *while* a service is running; is disabling it when `pid` is set intended? If so, the same condition should be added to `renderRowActions` for consistency; if not, drop the `|| (service[section_id].pid)` term.

---

openwrt-ai: nit: the commit subject/body describes only replacing the `dynamic_dns_lucihelper` call in `handleStopDDnsRule`, but this commit also changes the `reload.disabled` condition here and drops `status['_enabled'] == 0` from the reload gating at line 558. These two hunks are unrelated to the lucihelper removal and are undocumented in the message — consider splitting them out or mentioning them in the commit body so the scope is clear.

---

openwrt-ai (commented):
Reviewed 1 new commit. The core change (routing stop/reload through `/etc/init.d/ddns` per-section instead of `dynamic_dns_lucihelper.sh`, with the matching ACL update) is sound — the ddns init script uses `USE_PROCD` with per-section `procd_open_instance "$cfg"`, so `/etc/init.d/ddns stop <section>` targets a single instance as intended. Two notes left inline on the `reload.disabled` change, which is functionally correct but diverges from the static render path and is outside the commit's stated scope.

---

feckert:
@champtar I rebased the button handling again based on a note from the AI CI/CD on the overview page. It actually wasn't working right. The whole thing was a bit tricky. Like I said, the app needs some love. Let's see if I can find the time.

applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
```diff
@@ -220,8 +220,8 @@ return view.extend({
 			const stop = rows[i].querySelector('.cbi-section-actions .stop');
 			const cfg_enabled = uci.get('ddns', section_id, 'enabled');
 
-			reload.disabled = (status['_enabled'] == 0 || cfg_enabled == 0);
-			stop.disabled = (!service[section_id].pid);
+			reload.disabled = (cfg_enabled == 0) || (!service[section_id] && !service[section_id].pid);
```
openwrt-ai: `(!service[section_id] && !service[section_id].pid)` is broken in both directions. When `service[section_id]` is undefined — which happens for an enabled section whose service isn't currently running (the surrounding code reads `service[section_id]?.ip` with optional chaining, so undefined is expected here) — `!service[section_id]` is `true` and the `&&` then evaluates `!service[section_id].pid`, throwing `TypeError: Cannot read properties of undefined (reading 'pid')` and aborting the poll refresh. Because such a section has `cfg_enabled != 0`, the `||` doesn't short-circuit, so the crash is reachable. When `service[section_id]` is truthy the term short-circuits to `false`, so the pid check is dead and reload ends up gated only on `cfg_enabled`.

The [`stop.disabled` line directly below](``https://github.com/openwrt/luci/blob/bf6c5d62235f312ec3e71fcda1db46636a5c5878/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L224)`` already uses the correct guard — mirror it:

```suggestion
			reload.disabled = (cfg_enabled == 0) || !(service[section_id] && service[section_id].pid);
```

The same bug is in the static render path at [line 558](``https://github.com/openwrt/luci/blob/bf6c5d62235f312ec3e71fcda1db46636a5c5878/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L558):`` `(!resolved[section_id] && !resolved[section_id].pid)` should likewise become `!(resolved[section_id] && resolved[section_id].pid)`.

---

openwrt-ai (commented):
Reviewed 1 new commit (`bf6c5d6` "fix start/stop button logic"; `cfedd55` is content-identical to the previously-reviewed revision). One inline finding on the reworked reload-button gating, which throws a `TypeError` for enabled-but-not-running sections in both the poll and static render paths.

---

PR #8767, opened 2026-07-01
luci-base: fix undefined View reference in instantiateView()

maxskokov:
### Description
`instantiateView()` in `ui.js` guarded against invalid view modules with:

```js
if (!(view instanceof View))
```

`View` (capital V) is never declared or imported in `ui.js`, so this
line throws a `ReferenceError` whenever the guard is reached. Adding
`'require view as View'` would introduce a circular dependency since
view modules themselves require `ui`. Additionally, `L.require()`
returns a class instance rather than a constructor, making `instanceof`
conceptually incorrect here regardless.

Replace with a duck-type check on `render()`, which all valid LuCI view
modules are required to implement. Preserves fail-fast behavior for
invalid modules without relying on an undefined identifier.

The bug has been present since the function was introduced in
4e9cf92 (2020).

Also fixes two pre-existing `no-useless-escape` lint errors in the same
file (lines 3365 and 5105) that were surfaced by ESLint when the file
was touched.

### Maintainer _(preferred)_
@jow-

---

## Tested on
Verified by code inspection. `L.require()` returns `instance` per
`luci.js:2562`, not the constructor, confirming `instanceof` was
never applicable here.
---

## Checklist
---

Per the contributing guidelines, I'd like to request commit access.
Happy to take on more responsibility if you think the time is right.

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -3362,7 +3362,7 @@ const UIFileUpload = UIElement.extend(/** @lends LuCI.ui.FileUpload.prototype */
 	 * @returns {Promise}
 	 */
 	handleDelete(path, fileStat, ev) {
-		const parent = path.replace(/\/[^\/]+$/, '') ?? '/';
+		const parent = path.replace(/\/[^/]+$/, '') ?? '/';
 		const name = path.replace(/^.+\//, '');
 		let msg;
 
@@ -5102,7 +5102,7 @@ const UI = baseclass.extend(/** @lends LuCI.ui.prototype */ {
 				for (let i = 0; i < 2; i++)
 					for (let j = 0; j < ipaddrs.length; j++)
 						tasks.push(this.pingDevice(i ? 'https' : 'http', ipaddrs[j])
-							.then(ev => { reachable = ev.target.src.replace(/^(https?:\/\/[^\/]+).*$/, '$1/') }, () => {}));
+							.then(ev => { reachable = ev.target.src.replace(/^(https?:\/\/[^/]+).*$/, '$1/') }, () => {}));
 
 				return Promise.all(tasks).then(() => {
 					if (reachable) {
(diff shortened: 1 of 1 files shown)


modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -3362,7 +3362,7 @@ const UIFileUpload = UIElement.extend(/** @lends LuCI.ui.FileUpload.prototype */
 	 * @returns {Promise}
 	 */
 	handleDelete(path, fileStat, ev) {
-		const parent = path.replace(/\/[^\/]+$/, '') ?? '/';
+		const parent = path.replace(/\/[^/]+$/, '') ?? '/';
```
openwrt-ai, line 3365: nit: This `no-useless-escape` fix (and the second one at line 5105) is a separate change from the `instantiateView()` fix the commit subject describes. The commit message body documents only the `View` reference fix; the two lint fixes are undocumented in the message (they're mentioned in the PR body, but not the commit itself). Consider either noting them in the commit body or splitting them into their own commit so the message matches the diff's full scope.

---

openwrt-ai (commented):
Reviewed 1 new commit; verified the core fix against the tree.

The `instanceof View` → `typeof view?.render !== 'function'` change is sound: `View` is genuinely undefined in `ui.js` (only referenced in this guard), `L.require()` returns a class instance rather than a constructor (`luci.js` returns `instance` from `compileClass`), and the base `View` class defines `render()` as a required method (`luci.js:2028`), so all valid view modules pass the duck-type check. The two `no-useless-escape` fixes (`[^\/]` → `[^/]`) are correct and are the only genuine violations of that kind in the file. One nit inline about commit-message scope.

---

modules/luci-base/htdocs/luci-static/resources/ui.js
maxskokov: Fixed. Amended the commit scope.

openwrt-ai (commented):

The force-push leaves the tree byte-identical to the previously reviewed commit (`9c50037`); the only change is the commit-message body, which now documents the two `no-useless-escape` fixes — resolving the earlier commit-scope nit. Message matches the diff.

---

PR #8770, opened 2026-07-01
luci-app-strongswan: fixes and changes

feckert:
Before this changes:
before

After this changes:
after

Depends on https://github.com/openwrt/packages/pull/29882

### Maintainer _(preferred)_
@lvoegl @systemcrash 

---

## Tested on

**OpenWrt version:** master
**LuCI version:** master
**Web browser(s):** firefox 152.0.3

---

## Checklist

### DIFF
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -72,9 +72,7 @@ return view.extend({
 		m.tabbed = true;
 
 		// strongSwan General Settings
-		s = m.section(form.TypedSection, 'ipsec', _('General Settings'));
-		s.anonymous = true;
-		s.addremove = true;
+		s = m.section(form.NamedSection, 'globals', 'ipsec', _('General Settings'));
 
 		o = s.option(widgets.NetworkSelect, 'interface', _('Listening Interfaces'),
 			_('Interfaces that accept VPN traffic') + '<br /> ' +
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -76,13 +76,12 @@ return view.extend({
 		s.anonymous = true;
 		s.addremove = true;
 
-		o = s.option(widgets.NetworkSelect, 'listen', _('Listening Interfaces'),
-			_('Interfaces that accept VPN traffic'));
-		o.datatype = 'interface';
-		o.placeholder = _('Select an interface or leave empty for all interfaces');
-		o.default = 'wan';
+		o = s.option(widgets.NetworkSelect, 'interface', _('Listening Interfaces'),
+			_('Interfaces that accept VPN traffic') + '<br /> ' +
+			_('Select an interface or leave empty for all interfaces'));
 		o.multiple = true;
-		o.rmempty = false;
+		o.nocreate = true;
+		o.optional = true;
 
 		o = s.option(form.Value, 'debug', _('Debug Level'),
 			_('Trace level: 0 is least verbose, 4 is most'));
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -250,7 +250,7 @@ return view.extend({
 
 		o = s.taboption('advanced', form.ListValue, 'keyexchange', _('Keyexchange'),
 			_('Version of IKE for negotiation'));
-		o.value('ikev1', 'IKEv1 (%s)', _('deprecated'));
+		o.value('ikev1', 'IKEv1 (%s)'.format(_('deprecated')));
 		o.value('ikev2', 'IKEv2');
 		o.value('ike', 'IKE (%s, %s)'.format(_('both'), _('deprecated')));
 		o.default = 'ikev2';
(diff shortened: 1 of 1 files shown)


applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -72,22 +72,15 @@ return view.extend({
...
 		// strongSwan General Settings
-		s = m.section(form.TypedSection, 'ipsec', _('General Settings'));
-		s.anonymous = true;
-		s.addremove = true;
-
-		o = s.option(widgets.ZoneSelect, 'zone', _('Zone'),
-			_('Firewall zone that has to match the defined firewall zone'));
-		o.default = 'lan';
-		o.multiple = true;
+		s = m.section(form.NamedSection, 'globals', _('General Settings'));
```
openwrt-ai: The `NamedSection` argument order is wrong here, so the section type ends up as the string `"General Settings"` instead of `ipsec`. The constructor is `NamedSection(map, section_id, section_type, title, description)` (see [`CBINamedSection.__init__` in form.js](https://github.com/openwrt/luci/blob/e725c12b0f566eee12321668c35fdb36361d0adc/modules/luci-base/htdocs/luci-static/resources/form.js#L4159-L4164), which forwards the remaining args to [`CBIAbstractSection.__init__(map, sectionType, ...)`](https://github.com/openwrt/luci/blob/e725c12b0f566eee12321668c35fdb36361d0adc/modules/luci-base/htdocs/luci-static/resources/form.js#L866)). With the current call `section_id = 'globals'` and `sectiontype = _('General Settings')`, and no title.

Consequences: the section is created as `config 'General Settings' 'globals'` in `/etc/config/ipsec`, but the backend only iterates type `ipsec` ([`config_foreach config_ipsec ipsec` in swanctl.init](https://github.com/openwrt/packages/blob/master/net/strongswan/files/swanctl.init)), so the `listen` and `debug` values set here would never be applied. The section title also renders empty because the title argument is now missing.

Pass `ipsec` as the section type and `_('General Settings')` as the title:

```suggestion
		s = m.section(form.NamedSection, 'globals', 'ipsec', _('General Settings'));
```

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -72,22 +72,15 @@ return view.extend({
...
 
 		o = s.option(widgets.NetworkSelect, 'listen', _('Listening Interfaces'),
```
openwrt-ai: nit: `</br>` is a stray closing tag for a void element and is not valid HTML. The in-tree convention is `<br />` (used ~246 times across the tree; `</br>` appears nowhere else).

```suggestion
			_('Interfaces that accept VPN traffic') + '<br /> ' +
```

---

openwrt-ai (commented):
## Commit checks

- `e772d50` "luci-app-strongswan-swanctl: name the ipsec section type to globals" — the diff does not do what the subject says. It names the section *id* `globals` but sets the section *type* to the string `_('General Settings')` (argument-order bug on the `NamedSection` call), so the type is no longer `ipsec`. See the inline comment for the fix.

---

openwrt-ai (commented):
Reviewed 4 new commits (branch force-pushed since the last review); no new issues found.

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -72,22 +72,14 @@ return view.extend({
...
 		// strongSwan General Settings
-		s = m.section(form.TypedSection, 'ipsec', _('General Settings'));
-		s.anonymous = true;
-		s.addremove = true;
-
-		o = s.option(widgets.ZoneSelect, 'zone', _('Zone'),
-			_('Firewall zone that has to match the defined firewall zone'));
-		o.default = 'lan';
-		o.multiple = true;
+		s = m.section(form.NamedSection, 'globals', _('General Settings'));
```
openwrt-ai: The section-type argument is missing. The constructor is `NamedSection(map, section_id, section_type, title, …)`, so with only two positional args after the class `'globals'` is the section id and `_('General Settings')` becomes the **section type** (`this.sectiontype`) — not the title. Consequences: `this.sectiontype` is the translated string instead of `ipsec`; no title is passed, so the `<h3>` caption is skipped and the tab only renders because [`data-tab-title` falls back to `this.sectiontype`](https://github.com/openwrt/luci/blob/5ff2961da87016a9a4e2bb2ed2da2b39252e9652/modules/luci-base/htdocs/luci-static/resources/form.js#L2486); and the DOM `data-tab`/type would be `General Settings` (with a space). Every other in-tree caller passes the type explicitly, e.g. [`'global', 'adblock', _('Settings')`](``https://github.com/openwrt/luci/blob/5ff2961da87016a9a4e2bb2ed2da2b39252e9652/applications/luci-app-adblock/htdocs/luci-static/resources/view/adblock/overview.js#L196).`` Pass `ipsec` as the type and keep the label as the title:

```suggestion
		s = m.section(form.NamedSection, 'globals', 'ipsec', _('General Settings'));
```

---

openwrt-ai (commented):
## Commit checks

- `5ff2961` "luci-app-strongswan-swanctl: name the ipsec section type to globals" — the subject says to keep the `ipsec` section type (and the body adds `config ipsec 'globals'`), but the diff drops the type: the `NamedSection` call passes `_('General Settings')` in the section-type slot, so `this.sectiontype` becomes the label string rather than `ipsec`. See the inline comment for the one-line fix.

The other new commit (`7db75d5` "fix ike dropdown selection") is correct — `'IKEv1 (%s)'.format(_('deprecated'))` matches the sibling `ike` entry and its message.

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
feckert: I have update my PR

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The `NamedSection` argument-order bug flagged in the previous review is fixed here (`'globals'` id, `'ipsec'` type, `_('General Settings')` title), and the commit message matches the diff.

---

Issue #8785, opened 2026-07-02
Discussion: nonce-based CSP to replace unsafe-inline in luci-plugin-csp


maxskokov:
### How can we otherwise help?

Hi,

while reviewing `luci-plugin-csp` (58c7bc7) I noticed the `strict`
preset still includes `'unsafe-inline'` in `script-src`, which
defeats the XSS protection the plugin is meant to provide.

The root cause is architectural: `run_plugins()` in `luciplugins.uc`
only passes `plugin_id` to each plugin, so the CSP plugin has no
access to a per-request nonce. Without one, `unsafe-inline` is the
only way to allow LuCI's inline scripts.

I think a clean fix is feasible with minimal changes:

- generate `ctx.csp_nonce = randomid(16)` in `dispatcher.uc` where
  `runtime.env.ctx` is set (~line 933), and mirror it onto the http
  object (`http.csp_nonce = ctx.csp_nonce`)
- extend `run_plugins()` to forward an optional `extra_ctx` arg —
  backward compatible, existing plugins ignore extra args
- pass `{ csp_nonce: this.csp_nonce }` from `write_headers()` into
  `run_plugins()`, CSP plugin uses `'nonce-${ctx.csp_nonce}'`
- add `nonce="{{ ctx.csp_nonce }}"` to the ~10 inline `<script>` tags
  in core `.ut` templates (`ctx` is already available via `runtime.env`)

Before prototyping: does this approach look reasonable, or is there a
preferred way to pass per-request context into http header plugins?

cc @systemcrash  @josteink

josteink:
IIRC "unsafe-inline" was required before even the plugin system was conceived of. It's an artefact of how Luci is generally made and loaded. 

Allowing "unsafe-inline" will clearly allow some XSS attacks to still succeed, but it will prevent those from being chained into an attack loading external payloads from a foreign domain. 

As such it still provides a significant improvement as far as security goes. 

I believe removing "unsafe-inline" is a much bigger task than it is presented as in this discussion, which is why I didn't push for it when implementing the v1 CSP-plugin. Any semi-strict CSP is much better than NO CSP, which is what existed before. 

I guess some core Luci-developers may have more insights and opinions in these matters.

maxskokov:
Fair point — and the layered protection argument is valid. The nonce 
approach would be additive rather than a rewrite: the ~10 inline 
scripts in core templates are relatively contained, and `randomid()` 
is already available in `dispatcher.uc`. The main open question is 
[...]

PR #8788, opened 2026-07-03
luci-base: add nonce-based CSP support to replace unsafe-inline

maxskokov:
### Description
The CSP plugin (`luci-plugin-csp`) used `'unsafe-inline'` in `script-src`
because there was no mechanism to pass a per-request nonce to http header
plugins. This adds that mechanism across the stack with minimal changes.

`dispatcher.uc`: generate a per-request nonce via `randomid(16)` and store
it in `ctx.csp_nonce` (available in templates via `runtime.env`) and
`http.csp_nonce` (available in `write_headers()`).

`luciplugins.uc`: extend `run_plugins()` with an optional `extra_ctx` arg
forwarded to each plugin — backward compatible, existing plugins ignore it.

`http.uc`: pass `{ csp_nonce: this.csp_nonce }` into `run_plugins()` from
`write_headers()`.

Core `.ut` templates and all themes: add `nonce="{{ ctx?.csp_nonce }}"` to
inline `<script>` tags.

`luci-plugin-csp`: use `'nonce-${nonce}'` instead of `'unsafe-inline'` when
a nonce is available, falling back to `'unsafe-inline'` for backwards
compatibility.

Follows up on the discussion in #8785.

---

### Follow-up: inline event handler migration

Addressed the reviewer note that CSP2+ browsers ignore `'unsafe-inline'` in
`script-src` once a nonce is present, meaning inline event attributes
(`onclick`, `onsubmit`, `onreset`, `onkeyup`, `onchange`) and `javascript:`
URIs are blocked even when `<script>` elements are nonce-gated.

**`luci-base` / `cbi.js`**: extended `cbi_init()` to wire up all affected
handler patterns declaratively via `data-*` attributes:
`[data-href]`, `[data-cbi-submit]`, `[data-cbi-state]`, `[data-cbi-rowswap]`,
`.cbi-section-create-name` (keyup), `.cbi-cidr-toggle` (click), and
`form[name="cbi"]` submit/reset validation (error message read from
`data-error-msg`). Also moved `switchToCIDRList()` from `ipaddr.htm` into
`cbi.js`.

**`luci-lua-runtime` / `template.lua`**: exposed `csp_nonce` as a named key
in the Lua template viewns (same pattern as `token` and `REQUEST_URI`),
making it accessible as a plain variable in legacy CBI `.htm` templates.

**`luci-compat`**: removed all inline event attributes from `header.htm`,
`footer.htm`, `simpleform.htm`, `tblsection.htm`, `tsection.htm`,
`ipaddr.htm`, and `wireless_modefreq.htm`. Replaced with `data-*` attributes
handled by `cbi_init()`. Added `nonce="<%=csp_nonce%>"` to all inline
`<script>` blocks.

**`luci-app-commands`**: added `nonce="{{ ctx?.csp_nonce }}"` to the inline
`<script>` block; replaced `onclick` on Run/Download/Link buttons with
`data-action`/`data-cmd` attributes dispatched via a `DOMContentLoaded`
listener inside the nonce'd script.

---

### Maintainer _(preferred)_
@jow-

---

## Tested on
Verified by code inspection.

---

## Checklist
- [x] _(Optional)_ Includes what Issue it closes — follows up on #8785.

### DIFF
--- a/plugins/luci-plugin-csp/ucode/plugins/http/headers/c0ffee2cf63e160c091224d95d69cdff.uc
+++ b/plugins/luci-plugin-csp/ucode/plugins/http/headers/c0ffee2cf63e160c091224d95d69cdff.uc
@@ -1,29 +1,21 @@
 // SPDX-License-Identifier: Apache-2.0
-
 'use strict';
-
 import { cursor } from 'uci';
-
-function default_action(plugin_id) {
+function default_action(plugin_id, extra_ctx) {
 	const uci = cursor();
 	const mode = uci.get('luci_plugins', plugin_id, 'mode') || 'strict';
-
+	const nonce = extra_ctx?.csp_nonce ? `'nonce-${extra_ctx.csp_nonce}'` : "'unsafe-inline'";
 	const presets = {
-		'strict': "default-src 'none'; script-src 'self' 'unsafe-inline' 'unsafe-eval' 'trusted-types-eval'; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'; connect-src 'self' https://sysupgrade.openwrt.org;",
-		'permissive': "default-src 'self' https://*; script-src 'self' 'unsafe-inline' 'unsafe-eval' 'trusted-types-eval'; img-src 'self' data: blob: https://*; style-src 'self' 'unsafe-inline' https://*;",
+		'strict': `default-src 'none'; script-src 'self' ${nonce} 'unsafe-eval' 'trusted-types-eval'; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'; connect-src 'self' https://sysupgrade.openwrt.org;`,
+		'permissive': `default-src 'self' https://*; script-src 'self' ${nonce} 'unsafe-eval' 'trusted-types-eval'; img-src 'self' data: blob: https://*; style-src 'self' 'unsafe-inline' https://*;`,
 	};
-
 	let policy;
-
 	if (mode === 'custom')
 		policy = uci.get('luci_plugins', plugin_id, 'policy');
 	else
 		policy = presets[mode];
-
 	if (!policy)
 		return null;
-
 	return ['Content-Security-Policy', policy];
 }
-
 return default_action;
--- a/applications/luci-app-commands/ucode/template/commands.ut
+++ b/applications/luci-app-commands/ucode/template/commands.ut
@@ -28,7 +28,7 @@
 	` });
 -%}
 
-<script>
+<script nonce="{{ ctx?.csp_nonce }}">
 	function command_run(ev, id)
 	{
 		var field = document.getElementById(id);
@@ -113,6 +113,18 @@
 		ev.preventDefault();
 	}
 
+	document.addEventListener('DOMContentLoaded', function() {
+		document.querySelectorAll('button[data-cmd]').forEach(function(btn) {
+			btn.addEventListener('click', function(ev) {
+				var id = ev.currentTarget.dataset.cmd;
+				switch (ev.currentTarget.dataset.action) {
+				case 'run':      command_run(ev, id);      break;
+				case 'download': command_download(ev, id); break;
+				case 'link':     command_link(ev, id);     break;
+				}
+			});
+		});
+	});
 </script>
 
 {%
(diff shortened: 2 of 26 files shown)


plugins/luci-plugin-csp/ucode/plugins/http/headers/c0ffee2cf63e160c091224d95d69cdff.uc
```diff
@@ -1,29 +1,21 @@
...
 'use strict';
-
 import { cursor } from 'uci';
-
-function default_action(plugin_id) {
+function default_action(plugin_id, extra_ctx) {
 	const uci = cursor();
 	const mode = uci.get('luci_plugins', plugin_id, 'mode') || 'strict';
-
+	const nonce = extra_ctx?.csp_nonce ? `'nonce-${extra_ctx.csp_nonce}'` : "'unsafe-inline'";
```
openwrt-ai, line 7: Once a nonce is present in `script-src`, CSP2+ browsers ignore `'unsafe-inline'` in that directive — and here `'unsafe-inline'` is dropped from `script-src` entirely when a nonce is available. Nonces only whitelist `<script>` *elements*; they do **not** cover inline event-handler attributes (`onclick=`, `onchange=`, `onsubmit=`, `onkeyup=`, …) or `javascript:` URIs, which previously worked solely because of `'unsafe-inline'`. Those will now be blocked whenever the plugin runs in `strict` or `permissive` mode.

Inline handlers are still used in-tree, e.g. core [`csrftoken.ut:21`](https://github.com/openwrt/luci/blob/945be6185d4b80f78f772e7434cf0792c3db198e/modules/luci-base/ucode/template/csrftoken.ut#L21) (`onclick="window.history.back();"`), extensively across the [luci-compat CBI templates](https://github.com/openwrt/luci/blob/945be6185d4b80f78f772e7434cf0792c3db198e/modules/luci-compat/luasrc/view/cbi/footer.htm#L22) (`footer.htm`, `tblsection.htm`, `tsection.htm`, `simpleform.htm`, `header.htm`, …), and in [`luci-app-commands/commands.ut:151`](https://github.com/openwrt/luci/blob/945be6185d4b80f78f772e7434cf0792c3db198e/applications/luci-app-commands/ucode/template/commands.ut#L151). Their buttons/links would silently stop responding under the hardened policy.

The `<script>`-tag migration in this PR doesn't reach these attributes. Consider either migrating the remaining inline handlers to `addEventListener` in nonced scripts, or carrying `'unsafe-hashes'` plus the corresponding hashes in the policy for handlers that must stay inline. At minimum this behavior change is worth calling out in the commit/PR description.

---

openwrt-ai (commented):
One inline note on the CSP script-src change. The `<script>`-tag nonce plumbing itself is coherent (nonce is generated in `dispatch()` before templates render and before `write_headers()` emits the header, and the header's `'nonce-<hex>'` matches the `nonce="<hex>"` attribute value), and the single commit's message accurately describes the diff.

---

maxskokov:
Migration scope note: this PR switches the CSP `script-src` policy to nonce-only whenever `ctx.csp_nonce`/`http.csp_nonce` is present (the plugin drops `'unsafe-inline'` entirely once a nonce exists), and the inline-handler migration here covers `luci-base`, `luci-compat`, and `luci-app-commands`.

The rest of the tree isn't migrated yet, so enabling `luci-plugin-csp` in `strict`/`permissive` mode after this merges will silently break inline handlers/scripts on these pages (found by grepping for remaining `onclick=`/`onchange=`/un-nonced `<script>`):

- `applications/luci-app-lxc/htdocs/luci-static/resources/view/lxc/overview.js` — Start/Stop/Delete/Configure buttons built via `onclick='action_handler(this)'` string concatenation
- `applications/luci-app-bmx7/ucode/template/bmx7/bmxnodes.ut` — node info popup (`onclick="displayExtraInfo(...)"`)
- `modules/luci-compat/luasrc/view/cbi/delegator.htm` — bare `<script>cbi_d_update();</script>`, no `nonce`
- `modules/luci-mod-status/ucode/template/admin_status/index.ut` — inline `<script>` blocks, no `nonce`

Flagging so we can decide whether to widen this PR's scope before merge, or split the rest off into follow-up PRs (same pattern as #8762 splitting off from #8749).

openwrt-ai (commented):
Reviewed 4 new commits; no new issues found.

---

PR #8790, opened 2026-07-03
luci-mod-network: Add WPA3-Personal Compatibility Mode

hauke:
### Description

 * luci-mod-network: wireless: add WPA3-Personal Compatibility Mode
    
    Expose the sae-compat encryption mode (WPA3-Personal Compatibility Mode)
    in the wireless security settings. It advertises WPA2-PSK in the base
    RSNE while offering SAE through RSN Override elements, so that both
    legacy WPA2 and modern WPA3 clients can connect to a single SSID.
    
    The mode relies on RSN overriding, which is an access-point feature, so
    it is offered for AP mode only. It gets the passphrase field and, like
    the other WPA2/WPA3 modes, the 802.11r, KRACK-countermeasure and WPS
    options.
    
 * luci-mod-network: wireless: add gcmp256, sae_ext_key, transition_disable
    
    Expose three per-BSS WPA3 tuning options in the wireless security tab:
    
     - gcmp256: advertise the GCMP-256 pairwise cipher
     - sae_ext_key: advertise the SAE-EXT-KEY AKM (SAE using a group-dependent
       hash, aka SAE-GDH)
     - transition_disable: signal Transition Disable (WPA3 Specification v3.5
       section 13) so a client that connected once no longer downgrades
    
    GCMP-256 and SAE-EXT-KEY are mandatory for Wi-Fi 7 (EHT) but break
    interoperability with some clients. The wifi-scripts backend therefore
    defaults them on only for WPA3-Personal Compatibility Mode (sae-compat)
    on an EHT radio, and off everywhere else. The gcmp256/sae_ext_key
    checkboxes mirror that default, derived from both the encryption mode and
    the radio htmode, so the box shows the real default: leaving it untouched
    keeps the option unset (the backend decides), while toggling it writes an
    explicit override. As htmode is a radio property, not a wifi-iface one,
    it is read from the current selection of the radio's frequency widget
    (falling back to the saved config) instead of a same-section dependency.
    
    transition_disable is offered only for the WPA3-only modes where the
    signalled bitmap is non-empty.
    
This depends on https://github.com/openwrt/openwrt/pull/24041

---

## Tested on

**OpenWrt version:** OpenWrt 25.12-SNAPSHOT r33058+16-949e61ec65 
**LuCI version:** LuCI wifi-wpa3-compat-mode-25.12 branch 26.155.67407~06be235
**Web browser(s):** firefox

---

### DIFF
--- a/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
@@ -295,6 +295,30 @@ function add_dependency_permutations(o, deps) {
 	res.forEach(dep => o.depends(dep));
 }
 
+// Default gcmp256/sae_ext_key like the wifi-scripts backend: on only for
+// WPA3-Personal Compatibility Mode (sae-compat) on an EHT (Wi-Fi 7) radio, off
+// otherwise. htmode is a radio property, not a wifi-iface one, so it cannot be a
+// same-section o.defaults dependency. Read the currently selected htmode from the
+// radio's frequency/width widget so the default reacts to changes made in the
+// same modal, and only fall back to the saved config if that widget is not
+// available. On EHT the base updateDefaultValue picks the default from the
+// encryption mode (sae-compat -> on, sae/sae-mixed -> off); otherwise force off.
+// Both paths go through the base method so the checkbox is updated reactively.
+function eht_compat_default(section_id) {
+	const dev = uci.get('wireless', section_id, 'device');
+	let htmode = dev ? uci.get('wireless', dev, 'htmode') : null;
+
+	const freq = dev ? this.map.lookupOption('_freq', dev) : null;
+	if (freq)
+		htmode = freq[0].formvalue(dev)?.[0] ?? htmode;
+
+	this.defaults = (htmode && htmode.match(/^EHT/))
+		? { '1': [{ encryption: 'sae-compat' }], '0': [{ encryption: 'sae' }, { encryption: 'sae-mixed' }] }
+		: { '0': [] };
+
+	return form.Flag.prototype.updateDefaultValue.call(this, section_id);
+}
+
 // Define a class CBIWifiFrequencyValue that extends form.Value
 var CBIWifiFrequencyValue = form.Value.extend({
 	// Declare an RPC method to get the frequency list for a given device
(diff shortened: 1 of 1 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -2149,7 +2155,29 @@ return view.extend({
 						}
 
 						o = ss.taboption('encryption', form.Flag, 'wpa_disable_eapol_key_retries', _('Enable key reinstallation (KRACK) countermeasures'), _('Complicates key reinstallation attacks on the client side by disabling retransmission of EAPOL-Key frames that are used to install keys. This workaround might cause interoperability issues and reduced robustness of key negotiation especially in environments with heavy traffic load.'));
-						add_dependency_permutations(o, { mode: ['ap', 'ap-wds'], encryption: ['psk2', 'psk-mixed', 'sae', 'sae-mixed', 'wpa2', 'wpa3', 'wpa3-mixed'] });
+						add_dependency_permutations(o, { mode: ['ap', 'ap-wds'], encryption: ['psk2', 'psk-mixed', 'sae', 'sae-mixed', 'sae-compat', 'wpa2', 'wpa3', 'wpa3-mixed'] });
+
+						o = ss.taboption('encryption', form.Flag, 'gcmp256', _('GCMP-256 pairwise cipher'), _('Advertise the GCMP-256 pairwise cipher. Mandatory for Wi-Fi 7 (EHT) and recommended otherwise, but some clients and chipsets fail to associate when it is offered. Enabled by default in Compatibility Mode, disabled in WPA3-Personal and Mixed Mode.'));
```
openwrt-ai: `gcmp256` is added here as a `form.Flag`, i.e. it writes a standalone UCI boolean `option gcmp256`. The wifi-scripts backend never reads such an option: GCMP-256 is selected via the cipher suffix of the `encryption` value, which is parsed as `encryption[1]` after `split(config.encryption, '+', 2)` — see the [`switch(encryption[1])` block in iface.uc](``https://github.com/openwrt/openwrt/blob/97d6719ca75cf7e9ecd1d40e7b95f47531e8074f/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc#L76-L100)`` and [`*gcmp256) wpa_cipher="GCMP-256"` in netifd-wireless.sh](``https://github.com/openwrt/openwrt/blob/97d6719ca75cf7e9ecd1d40e7b95f47531e8074f/package/network/config/wifi-scripts/files/lib/netifd/netifd-wireless.sh#L242).`` `config.gcmp256` does not appear anywhere in wifi-scripts, and `gcmp256` is not a hostapd directive, so this checkbox has no backend consumer and does nothing.

The per-mode default claim doesn't hold either: for `sae-compat`, GCMP-256 comes from [`rsn_override_pairwise_2 = 'GCMP-256'` derived from the EHT htmode](``https://github.com/openwrt/openwrt/blob/97d6719ca75cf7e9ecd1d40e7b95f47531e8074f/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc#L60-L66),`` not from a `gcmp256` flag. To actually toggle the cipher the UI would need to influence the `encryption` string rather than expose a separate flag. Is a backend change adding a `gcmp256` option planned (and if so it should be noted in the PR description as a cross-repo dependency)?

---

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
```diff
@@ -2149,7 +2155,29 @@ return view.extend({
...
+							'1': [{ encryption: 'sae-compat' }],
+							'0': [{ encryption: 'sae' }, { encryption: 'sae-mixed' }]
+						};
+
+						o = ss.taboption('encryption', form.Flag, 'sae_ext_key', _('SAE-EXT-KEY (SAE-GDH)'), _('Advertise the SAE-EXT-KEY AKM (SAE using a group-dependent hash). Mandatory for Wi-Fi 7 (EHT) and recommended otherwise, but some clients misbehave when it is offered, in particular together with 802.11r Fast Transition. Enabled by default in Compatibility Mode, disabled in WPA3-Personal and Mixed Mode.'));
+						add_dependency_permutations(o, { mode: ['ap', 'ap-wds'], encryption: ['sae', 'sae-mixed', 'sae-compat'] });
+						o.defaults = {
+							'1': [{ encryption: 'sae-compat' }],
+							'0': [{ encryption: 'sae' }, { encryption: 'sae-mixed' }]
+						};
```
openwrt-ai: The backend schema defaults [`sae_ext_key` to `true`](``https://github.com/openwrt/openwrt/blob/97d6719ca75cf7e9ecd1d40e7b95f47531e8074f/package/network/config/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-iface.json#L965-L969),`` and that default is materialized into the config for every mode by [validate.uc's set-defaults loop](``https://github.com/openwrt/openwrt/blob/97d6719ca75cf7e9ecd1d40e7b95f47531e8074f/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/validate.uc#L106-L112),`` after which [iface.uc appends `SAE-EXT-KEY` for `auth_type == 'sae'`](``https://github.com/openwrt/openwrt/blob/97d6719ca75cf7e9ecd1d40e7b95f47531e8074f/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc#L158-L166).`` So plain WPA3-SAE already advertises SAE-EXT-KEY by default.

Two consequences of the `'0'` default for `sae`/`sae-mixed` here: (1) the commit message's claim that the default is "off for sae and sae-mixed ... matching the wifi-scripts backend" is inverted for `sae_ext_key` — the backend default is on. (2) `form.Flag` is `rmempty`, so leaving the box unchecked (value `'0'` == displayed default) writes nothing, the backend default (`true`) wins, and a user cannot actually disable SAE-EXT-KEY for `sae`/`sae-mixed` from this UI. If the intent is to let users turn it off, the default for these modes should be `'1'` (matching the backend) so unchecking emits an explicit `sae_ext_key '0'`. Is the current `'0'` default intentional?

---

openwrt-ai (commented):
Reviewed 2 new commits. Flagged two backend-coupling concerns inline (`gcmp256` flag has no backend consumer; `sae_ext_key` default appears inverted vs. the wifi-scripts schema). The formality subject-length failure is already reported by the CI bot, so it is not duplicated here.

---

hauke:
This PR is depending on https://github.com/openwrt/openwrt/pull/24041 which adds these options. 

The shell script is deprecated and not really maintained any more, the ucode configuration is the currently maintained version.

hauke:
It now checks if the bss is in sae-compat mode and is Wifi 7.

openwrt-ai (commented):

---

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
hauke: It will be read with https://github.com/openwrt/openwrt/pull/24041 
The shell script code is deprecated, we only care about the ucode handling.

modules/luci-mod-network/htdocs/luci-static/resources/view/network/wireless.js
hauke: It will be read with https://github.com/openwrt/openwrt/pull/24041

openwrt-ai (commented):

---

backslashxx:
Works well. I've been waiting for this. Thanks a lot.

PR #8792, opened 2026-07-04
luci-app-lxc: build container rows with E() instead of HTML strings

maxskokov:
### Description
CSP nonce policies block inline event attributes (`onclick`, `onchange`) even when `script-src` is nonce-gated (see #8785 / #8788). The dynamically created container rows and the configuration editor in `overview.js` relied on such attributes, assembled through `innerHTML` string concatenation. That concatenation also interpolated the container name and the raw container configuration into markup unescaped — a literal `</textarea>` inside a container's saved config would break out of the configuration editor.

This rebuilds the rows and the configuration editor with `E()`, attaching `click`/`change` listeners directly and passing the container name and configuration contents as text nodes instead of interpolated markup:

- `lxc_create_template()`: row markup (name/status/action buttons/"more" select) built via `E()`; `data-action` attributes are preserved and read back via `ev.target` in the click/change handlers, same pattern already used in `luci-app-package-manager`.
- `lxc_configure_template()`: now returns an element instead of an HTML string, so the container's saved LXC config lands in the `<textarea>` as a text node rather than being spliced into markup.
- Removed the `window.action_handler` / `window.action_more_handler` / `window.lxc_configure_handler` / `window.lxc_create` exports — they existed only so inline `onclick`/`onchange` attributes could look up the handlers globally, which no longer applies.

No functional change to `action_handler()` / `action_more_handler()` / `lxc_configure_handler()` themselves — same DOM structure (row → td → input), same `dataset` lookups.

Related to the CSP-nonce work in #8788, but split out since it's a larger, self-contained rewrite of one view rather than the plumbing change.

### Maintainer _(preferred)_
@systemcrash

---

## Tested on
Verified by code inspection:
- Confirmed `E()`'s `dom.append()` (`modules/luci-base/htdocs/luci-static/resources/luci.js`) calls `elem.addEventListener(key, fn)` for function-valued attributes — same mechanism already relied on elsewhere in the tree (e.g. `luci-app-package-manager/package-manager.js`, `luci-app-upnp`).
- Walked each handler (`action_handler`, `action_more_handler`, `lxc_configure_handler`) against the new markup to confirm `self.parentNode`/`self.parentNode.parentNode`/`dataset` lookups still resolve to the same elements as before.

Manual test recommended before merge: create a container, use Start/Stop/Delete and the "more" dropdown (configure/freeze/unfreeze/reboot) to confirm all actions still fire, and check that a container name/config containing HTML-special characters renders as literal text rather than being interpreted as markup.

---

## Checklist

### DIFF
--- a/applications/luci-app-lxc/htdocs/luci-static/resources/view/lxc/overview.js
+++ b/applications/luci-app-lxc/htdocs/luci-static/resources/view/lxc/overview.js
@@ -73,44 +73,60 @@ return L.view.extend({
 			}
 
 			info_message(output_list, '');
-			let actions = '';
-			actions += `<input type='button' onclick='action_handler(this)' data-action='start' value='${_('Start')}' class='cbi-button cbi-button-apply' />`;
-			actions += `&#160;<input type='button' onclick='action_handler(this)' data-action='stop' value='${_('Stop')}' class='cbi-button cbi-button-reset' />`;
-			actions += `&#160;<input type='button' onclick='action_handler(this)' data-action='destroy' value='${_('Delete')}' class='cbi-button cbi-button-remove' />`;
-			actions += `&#160;<select class='cbi-input-select cbi-button' style='width:10em' onchange='action_more_handler(this)'>\
-							<option selected='selected' disabled='disabled'>more</option>\
-							<option>configure</option>\
-							<option>freeze</option>\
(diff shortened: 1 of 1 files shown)


openwrt-ai (commented):

---

PR #8798, opened 2026-07-05
luci-base: protocol/dhcp: add option to disable the DHCPv4 client ID

hauke:
### Description
netifd gained a per-interface "sendclientid" boolean to suppress DHCP option 61 (client identifier), for ISPs that reject requests carrying it (e.g. Turksat KabloNet).

Expose it in the DHCP protocol form as a "Send Client ID" flag, enabled by default. Disabling it writes sendclientid '0', which makes netifd start udhcpc with -C so no client identifier is sent. The explicit "Client ID" field is hidden while sending is disabled.

Reported-at: https://github.com/openwrt/openwrt/issues/24065

This depends on https://github.com/openwrt/openwrt/pull/24087

### Maintainer _(preferred)_
@jow- 

---

## Tested on

**OpenWrt version:** OpenWrt SNAPSHOT r35210+1-8614a2ba68
**LuCI version:** LuCI luci-dhcp-sendclientid branch 26.186.53400~11788e4
**Web browser(s):** Firefox


openwrt-ai (commented):

---

PR #8799, opened 2026-07-05
luci-app-attendedsysupgrade: handle device specific image types

### LINKED ISSUE #8795: luci-app-attendedsysupgrade: TypeError: undefined is not an object (evaluating 'image.name')

### screenshots or captures

Based on reading the code, it seems it cannot find an image with type "sysupgrade" in the `selectImage` function. The `images` value:

```json
[{"filesystem":"squashfs","name":"openwrt-25.12.5-7f4088a34d41-bcm53xx-generic-asus_rt-ac88u-squashfs.trx","sha256":"dec13da00c6cfc1100c9fd2778c20d90a8caf3f24830b15a1ca24d109aa02ed4","sha256_unsigned":"dec13da00c6cfc1100c9fd2778c20d90a8caf3f24830b15a1ca24d109aa02ed4","size":11665408,"type":"trx"}]
```

### Actual behaviour

1. Clicking "Request firmware image" doesn't do anything 
2. An error is printed in the console

### Expected behaviour

1. Requesting a firmware image works

### Steps to reproduce

1. Go to System -> Attended Sysupgrade
2. Click "Search for firmware upgrade"
3. Click "Request firmware image"

### Additional Information

```shell
OpenWrt 25.12.2 r32802-f505120278 / LuCI (HEAD detached at 067535e) branch 26.082.75780~067535e
```

Safari

### Relevant log output

```shell
[Error] Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'image.name')
	handle200 (overview.js:9:207)
	(anonymous function) (overview.js:19)
```

maxskokov:
### Description

Fixes openwrt/luci#8795.

bcm53xx devices publish their single image under a device specific type such as `trx` or `chk` instead of `sysupgrade`, so `selectImage()` came up empty and `handle200()` threw an unhandled TypeError on `image.name`, leaving the "Request firmware image" button silently broken. The same applies to `sdcard` images on tegra.

This mirrors the fallback owut uses in `complete_build_info()`: when no image matches the expected types, drop factory images and accept the remaining type if it is unambiguous. When no usable image is left (some devices only publish factory or kernel images), an error dialog pointing to the support thread is shown instead of failing silently.

### Maintainer _(preferred)_
@efahl

---

## Tested on

Not tested on a device. The new selection logic was run in isolation (node) against the real 25.12.5 `profiles.json` of bcm53xx/generic, x86/64, ath79/generic and tegra/generic, all profiles with and without the EFI flag:

- 762 device/EFI combinations pick the same image as before (no behavior change)
- 62 previously broken combinations now resolve (types `trx`, `chk`, `lxl`, `bin` on bcm53xx, `sdcard` on tegra)
- 6 devices without any usable sysupgrade image now get the error dialog instead of the silent failure

eslint passes on the changed file.

---

## Checklist
- [x] Incremented :up: any `PKG_VERSION` in the Makefile. (no PKG_VERSION in this Makefile, the version comes from luci.mk)

### DIFF
--- a/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
+++ b/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
@@ -160,14 +160,27 @@ return view.extend({
 				return (e.type == 'sysupgrade' || e.type == 'combined');
 			}
 		}
-		return images.filter(filesystemFilter).filter(typeFilter)[0];
+		let candidates = images.filter(filesystemFilter);
+		let image = candidates.filter(typeFilter)[0];
+
+		if (!image) {
+			/* Some devices ship a single image used for both factory
+			 * install and sysupgrade under a device specific type, e.g.
+			 * 'trx' on bcm53xx. Like owut, fall back to the remaining
+			 * non-factory type if it is unambiguous. */
+			let remaining = candidates.filter((e) => !e.type.includes('factory'));
+			if (remaining.length && remaining.every((e) => e.type == remaining[0].type))
+				image = remaining[0];
+		}
+
+		return image;
 	},
 
 	handle200: function (response, content, data, firmware) {
 		response = response.json();
 		let image = this.selectImage(response.images, data, firmware);
 
-		if (image.name != undefined) {
+		if (image) {
 			this.sha256_unsigned = image.sha256_unsigned;
 			let sysupgrade_url = `${data.url}/store/${response.bin_dir}/${image.name}`;
 
@@ -246,6 +259,18 @@ return view.extend({
 			if (data.rebuilder) {
 				this.handleRebuilder(content, data, firmware);
 			}
+		} else {
+			ui.showModal(_('No sysupgrade image found'), [
+				E('p', _('The image builder did not return a usable sysupgrade image for this device (image types: %s).').format(
+					response.images.map((i) => i.type).join(', '))),
+				E('p', [
+					_('Please report this in'), ' ',
+					support_link, '.',
+				]),
+				E('div', { class: 'right' }, [
+					E('div', { class: 'btn', click: ui.hideModal }, _('Close')),
+				]),
+			]);
 		}
 	},
 
@@ -395,7 +420,9 @@ return view.extend({
 							response = response.json();
 							let view = document.getElementById(server);
 							let image = this.selectImage(response.images, data, firmware);
-							if (image.sha256_unsigned == this.sha256_unsigned) {
+							if (!image) {
+								view.innerText = '⚠️ %s'.format(server);
+							} else if (image.sha256_unsigned == this.sha256_unsigned) {
 								view.innerText = '✅ %s'.format(server);
 							} else {
 								view.innerHTML = `⚠️ ${server} (<a href="${server}/store/${


applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -160,14 +160,27 @@ return view.extend({
...
+
+		return image;
 	},
 
 	handle200: function (response, content, data, firmware) {
 		response = response.json();
 		let image = this.selectImage(response.images, data, firmware);
 
-		if (image.name != undefined) {
+		if (image) {
```
openwrt-ai, line 183: This correctly guards `handle200`, but the sibling rebuilder path in `handleRequest` dereferences the result of `selectImage()` the same way without a guard: [`image.sha256_unsigned` / `image.name` at overview.js#L410-L417](``https://github.com/openwrt/luci/blob/2a3fa3ddd75ff7e380377a9a74dd6c3f5693ad09/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L410-L417).`` Now that `selectImage()` can legitimately return `undefined`, a rebuilder whose response has no matching image would throw the same TypeError this PR fixes here. It only runs after the main build already resolved an image (so it's a narrow case), but a guard there would close the gap.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -160,14 +160,27 @@ return view.extend({
...
+		let candidates = images.filter(filesystemFilter);
+		let image = candidates.filter(typeFilter)[0];
+
+		if (!image) {
+			/* Some devices ship a single image used for both factory
+			 * install and sysupgrade under a device specific type, e.g.
+			 * 'trx' on bcm53xx. Like owut, fall back to the remaining
+			 * non-factory type if it is unambiguous. */
+			let remaining = candidates.filter((e) => !e.type.includes('factory'));
+			if (remaining.length && remaining.every((e) => e.type == remaining[0].type))
```
openwrt-ai, line 172: The fallback accepts any single non-`factory` type as a sysupgrade image. That is safe only as long as non-flashable partial images (e.g. `kernel`, `boot`, `rootfs`) never survive the earlier `filesystemFilter`. If a device published exactly one such type carrying a `filesystem` value matching the device rootfs, it would be offered — and flashed — as a sysupgrade target. Your testing covered 4 targets and the kernel-only devices correctly fell through to the error dialog, so this may already be handled by the filesystem mismatch; is exclusion-by-filesystem guaranteed for all partial image types, or would an explicit type allow-list be safer here?

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
```diff
@@ -246,6 +259,18 @@ return view.extend({
 			if (data.rebuilder) {
 				this.handleRebuilder(content, data, firmware);
 			}
+		} else {
+			ui.showModal(_('No sysupgrade image found'), [
+				E('p', _('The image builder did not return a usable sysupgrade image for this device (image types: %s).').format(
+					response.images.map((i) => i.type).join(', '))),
+				E('p', [
+					_('Please report this in'), ' ',
+					E('a', { href: support_url }, _('this forum thread')), '.',
```
openwrt-ai: nit: the file already defines a reusable [`support_link`](``https://github.com/openwrt/luci/blob/2a3fa3ddd75ff7e380377a9a74dd6c3f5693ad09/applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js#L19)`` node for exactly this "this forum thread" link (the header comment stresses isolating the support URL), and `handleError` reuses it. Reusing it here keeps the single source of truth.

```suggestion
					support_link, '.',
```

---

openwrt-ai (commented):
Reviewed the single new commit; its message matches the diff. A few inline notes below (one latent-TypeError sibling path, one question on the fallback's type safety, one nit). Nothing blocking.

---

maxskokov:
On the type safety: the fallback only sees images that already passed filesystemFilter, and partial images (kernel, boot, rootfs) carry no rootfs filesystem value, so they are dropped before it runs. That is why the kernel-only devices in my test fell through to the error dialog instead of being offered. Same narrowing owut does (drop factory, take the remaining type). An explicit allow-list would be stricter but diverges from owut, so I would leave that to @efahl.

openwrt-ai (commented):

The changes since my previous review resolve both inline notes: the rebuilder status path now guards the newly-possible `undefined` from `selectImage()` (`if (!image)`), and the error dialog reuses the module-level `support_link` to match the rest of the file. Commit message still matches the diff.

---

applications/luci-app-attendedsysupgrade/htdocs/luci-static/resources/view/attendedsysupgrade/overview.js
efahl: > an explicit type allow-list be safer here?

No, that would be unmaintainable.

efahl (approved):
Looks good to me.

maxskokov:
Approved by @efahl and CI is green, so this should be ready. It closes #8795, the trx/sdcard image type crash.

efahl:
This definitely needs to be backported to 25.12, too.  Not sure if it would apply cleanly to 24.10, but that would be nice if possible.

maxskokov:
@systemcrash can you merge this?

PR #8800, opened 2026-07-05
luci-app-nut: only allow admin write access and limit read access

danielfdickinson:
### Description

In order to mitigate [GHSA-4f29-2ff2-6hqf](https://github.com/openwrt/luci/security/advisories/GHSA-4f29-2ff2-6hqf): 'authenticated luci-app-nut configuration user can execute NUT monitor notify commands as root', we allow write access to the nut UCI configuration only for admin user, and also limit some additional read access for non-admin users.

In order to make this work, we split each service (server, monitor, and cgi) into 'main' and 'admin' screens.

We also remove unneeded permissions in luci-app-nut* ACLs

**Note:** This is only a preliminary mitigation. A more comprehensive solution which involves changes the nut packages in the packages repo, as well as more changes here, is a work in progress.

### Maintainer

---

## Tested on

**OpenWrt version:** OpenWrt SNAPSHOT (r35205-97d6719ca7)
**LuCI version:** LuCI Master (26.185.59369~d7a654e)
**Web browser(s):** 140.12.0esr (64-bit)

---

## Checklist
- [x] Includes what Issue it closes (e.g. openwrt/luci#issue-number).
      * [GHSA-4f29-2ff2-6hqf](https://github.com/openwrt/luci/security/advisories/GHSA-4f29-2ff2-6hqf)

## Also notify

@hnyman @jow- @hauke


danielfdickinson:
One question I have is whether it is possible to allow read access to a configuration _except_ certain options (e.g. NUT passwords in the NUT configuration) via rpcd ACL.

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js
```diff
@@ -72,18 +71,11 @@ return view.extend({
...
 		o.datatype = 'uinteger'
 		o.placeholder = 1;
 		o.optional = true;
 
-		o = s.option(form.Value, 'shutdowncmd', _('Shutdown command'));
-		o.optional = true;
-		o.placeholder = '/sbin/halt'
-
 		o = s.option(form.Value, 'notifycmd', _('Notify command'));
 		o.optional = true;
```
openwrt-ai: `notifycmd` is the exact field the advisory (GHSA-4f29-2ff2-6hqf, "execute NUT monitor notify commands as root") is about — upsmon runs it as root. This PR moves the other privileged monitor fields (`runas`, `shutdowncmd`) off this read-gated "main" screen into the admin-only screen, but `notifycmd` was left here and is also added to the admin view ([`nut_monitor_admin.js#L23`](``https://github.com/openwrt/luci/blob/0db655b31f55f3e63010e5ad2f8d145a1e3133e7/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js#L23)),`` so it now appears on both screens. A `luci-app-nut-read` user can't persist a change (no write ACL), but leaving an editable Notify command field on the non-admin screen is inconsistent with the stated goal and duplicates the admin field. Consider removing it here so it lives only on the admin screen, matching the `runas`/`shutdowncmd` treatment.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
```diff
@@ -94,22 +94,15 @@ return view.extend({
...
 		o.optional = true;
 		o.datatype = 'uinteger'
 		o.placeholder = 24;
 
 		if (have_ssl_support) {
-			o = s.option(form.Value, 'certfile', _('Certificate file (SSL)'));
-			o.optional = true;
+			o = s.option(form.FileUpload, 'certfile', _('Certificate file (SSL)'));
+			o.rmempty = true;
+			o.optional = false;
```
openwrt-ai: nit: switching `certfile` from `form.Value` to `form.FileUpload` (and making it required, `optional = false`) is unrelated to the security mitigation the commit message describes, i.e. undocumented mixed scope. It also has a functional wrinkle: this is the read-gated "main" server screen (`admin/services/nut/nut_server_main` depends only on `luci-app-nut-read`), but `FileUpload` writes uploads to `/etc/luci-uploads/*`, and the write ACL for that path is granted only in `luci-app-nut-admin`. So a non-admin `read` user is shown a file-upload widget they cannot use. Is the certfile change intended to be part of this PR, and if so should this option move to the admin screen alongside the other write-capable settings?

---

applications/luci-app-nut/root/usr/share/luci/menu.d/luci-app-nut.json
```diff
@@ -1,11 +1,16 @@
 {
 	"admin/services/nut": {
 		"title": "Network UPS Tools",
+		"order": 1,
 		"action": {
-			"type": "firstchild"
+			"type": "firstchild",
+			"recurse": true,
+			"preferred": "nut/nut_server"
```
openwrt-ai: nit: `preferred` for a `firstchild` action names a child menu node, not a view path — see luci-base which uses the bare child segment (`"preferred": "overview"` for child `admin/status/overview`). Here the children are keyed `nut_server_main`, `nut_server_admin`, etc., so this would be `"nut_server_main"`, not `"nut/nut_server"` (a view path with a slash). It happens to be harmless today because `resolve_firstchild` in `dispatcher.uc` selects purely by `order`/`node_weight` and doesn't read `preferred` at all (the value isn't consumed anywhere in the tree), and `nut_server_main` already has the lowest order. But the value as written matches no child and diverges from the convention.

---

openwrt-ai (commented):
Reviewed the ACL/screen split for the NUT security mitigation. The read/admin ACL restructure and the main/admin menu split look coherent; the old `luci-app-nut` ACL name is fully replaced and no stale references remain. Three inline notes: one about the `notifycmd` field (the one the advisory targets) still being on the read-gated main monitor screen and duplicated on the admin screen, plus two nits on the undocumented `certfile` FileUpload change and the `preferred` menu value. CI is still pending on the head commit.

---

openwrt-ai (commented):

The three points from the previous review are all addressed in this revision: `notifycmd` is now only present on the admin-gated `nut_monitor_admin` view (removed from the read-visible main monitor screen), `certfile` was reverted from `form.FileUpload` to `form.Value` (consistent with dropping the `/etc/luci-uploads/*` write ACL), and the menu `preferred` value now points to the child node `nut_server_main`.

---

danielfdickinson:
@hnyman ping?

PR #8801, opened 2026-07-06
luci-app-strongswan-swanctl: move connection mapping to ucode

lvoegl:
### Description

Moves mapping strongSwan connections and SAs into a ucode function. While moving it there, also fixes an error where `.sa` property was falsely named `.childSa` and `.childSa` was referenced as `.ChildSa`.

When opening the inner 'child' modal, the dismiss button now takes the user back to the outer modal instead of closing the modal completely.

### Maintainer

@lvoegl, Nicholas Smith

---

## Tested on

**OpenWrt version:** master 
**LuCI version:** master
**Web browser(s):** 152.0.4

---

## Checklist


feckert:
Thanks merged!

PR #8804, opened 2026-07-06
luci-app-package-manager: count only one provider for alternative deps

### LINKED ISSUE #8411: luci-app-package-manager: download size calculated incorrectly

### screenshots or captures

_No response_

### Actual behaviour

When calculating the dependencies and download size for a package that has alternative dependencies, dependencies and download sizes for all alternatives are shown

### Expected behaviour

When a package is installed that fulfills the requirement only the required downloads for that package should be shown

### Steps to reproduce

Upgrade a package with alternative dependencies, such as apk-mbedtls, which depends on wget-any

### Additional Information

```shell
OpenWrt 25.12.0 r32713-f919e7899d / LuCI openwrt-25.12 branch 26.064.72454~b89e6b3
```

_No response_

### Relevant log output

```shell

```

maxskokov:
### Description

Fixes #8411.

When a package depends on a virtual name with several providers (for example a dependency on `wget-any`, provided by `wget-ssl`, `wget-nossl` and `uclient-fetch`), the dependency details listed every provider and added each one to the download size estimate and the subdependency tree. apk installs only a single provider, so the size was overstated, and once one provider was already installed the dependency still showed the others as pending.

This picks one effective provider per dependency, an installed one if present, otherwise the first, and counts only that towards the size estimate and subdependencies. The install command is unchanged: it passes the target package to apk, which resolves providers itself, so `info.install` only ever drove the displayed estimate.

When no provider is installed yet the first one is used as the estimate. The exact provider apk would pick is not reproduced, since that would mean duplicating its preference rules, but the estimate now reflects a single provider instead of the sum of all of them.

### Maintainer
@jow-

---

## Tested on

Not tested on a device. The dependency selection and size accounting (`renderDependencies`, `renderDependencyItem`, `pkgStatus`) were run headless against fixtures shaped like the internal package model, using the reporter's `apk-mbedtls` -> `wget-any` case:

- nothing installed: before counted all three providers, now counts one
- a provider already installed: before still counted the remaining two, now counts none
- single-provider dependency chains: unchanged

eslint passes on the changed file.

---

## Checklist
- [x] This PR is not from my *main* or *master* branch, but a *separate* branch.
- [x] Each commit has a valid `Signed-off-by` row.
- [x] Each commit and PR title has a valid `<package name>: title` first line subject.
- [x] Includes what Issue it closes (openwrt/luci#8411).

### DIFF
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -523,7 +523,7 @@ function versionSatisfied(ver, ref, vop)
 	return false;
 }
 
-function pkgStatus(pkg, vop, ver, info)
+function pkgStatus(pkg, vop, ver, info, count)
 {
 	info.errors = info.errors || [];
 	info.install = info.install || [];
@@ -559,7 +559,8 @@ function pkgStatus(pkg, vop, ver, info)
 	}
 	else if (!pkg.missing) {
 		if (!vop || versionSatisfied(pkg.version, ver, vop)) {
-			info.install.push(pkg);
+			if (count !== false)
+				info.install.push(pkg);
 			return E('span', { 'class': 'label' }, _('Not installed'));
 		}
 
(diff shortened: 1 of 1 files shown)


applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -602,12 +622,13 @@ function renderDependencyItem(dep, info, flat)
 			text += ' (~%1024mB)'.format(pkg.size);
 
 		li.appendChild(E('span', { 'data-tooltip': pkg.description },
-			[ text, ' ', pkgStatus(pkg, vop, ver, info) ]));
+			[ text, ' ', pkgStatus(pkg, vop, ver, info, !satisfied && i === effective) ]));
```
openwrt-ai, line 642: `count` only gates the `Not installed` push below (line 562); the `Needs upgrade` branch at [`info.install.push(repl)`](``https://github.com/openwrt/luci/blob/27f8cf6271583692bb8f3a6050aa0550e7c161e6/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js#L541)`` is not gated by it. So if a *versioned* virtual dependency has two already-installed providers that both need upgrading, this loop calls `pkgStatus` for each (both with `count === false`) and both `repl` packages are still pushed to `info.install`, re-inflating the size estimate this PR aims to fix. This is an unusual state (apk normally keeps only one provider of a virtual installed), so it may be out of scope — but is the omission intentional, or should the `repl` push also be limited to the effective provider?

---

openwrt-ai (commented):
Reviewed 1 new commit. One inline question about the download-size accounting; the change is otherwise sound.

---

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
maxskokov: Intentional. The `repl` push is an upgrade of an already installed provider, so it is a real download that should still count even when the virtual is otherwise satisfied. Gating it on `count` would drop that upgrade size for the effective provider, since `count` is false whenever any provider is installed. The double count needs two installed providers of the same virtual at once, which apk does not keep. The inflation from #8411 comes from the not-installed push, which is the one gated here.

I also updated the fallback to follow the apk solver's own selection order: installed provider, otherwise the highest `provider-priority`, otherwise the first available one. The field is already in the model since the backend queries with `--fields all`, and the build system emits it from `PRIORITY` in `include/package-pack.mk`.

openwrt-ai (commented):

The provider-priority fallback resolves the prior review's question and mirrors the apk solver order (installed → highest `provider-priority` → first available). The `provider-priority` field name matches the dashed-key convention used elsewhere in the parser (`file-size`, `auto-installed`) and the `--info "provider-priority:..."` emit in `include/package-pack.mk`.

---

PR #8805, opened 2026-07-07
luci-proto-xfrm: replace ifid datatype check with range check

feckert:
### Description
According to the documentation, values from '1' to '4294967295' are allowed values for the 'ifid' option. The value ‘0’ is excluded, as '0' is treated as the default ID or an unassigned ID.

While we’re at it, mark the option as non-optional.

### Maintainer _(preferred)_
@hgl 

---

## Checklist


systemcrash:
Merged. Thanks @feckert

PR #8806, opened 2026-07-07
luci-proto-wireguard: pre-fill peer addresses

BdN3504:
### Description
Improve the "Generate configuration" dialog:

1. Fix the Address field being empty for existing peers: previously eips
   was read via formvalue() before parent.save() resolved, so the DOM
   state was stale. Now reads via cfgvalue() after save completes.

2. Auto-assign the next free IPv4+IPv6 address for new peers via a new
   getNextPeerAddress RPC method. It derives the subnet from the server
   interface's own addresses and finds the lowest unused host address,
   iterating across /24 block boundaries when the current block is full
   (e.g. a /8 with peers spanning multiple /24s). A client-side
   reservation list prevents duplicate allocation when creating multiple
   peers back-to-back in the same session.

3. Write the allocated addresses back to allowed_ips in UCI immediately,
   so the server-side peer entry is complete without manual editing.
   Uses uci.save() directly to avoid parent.save() re-reading the empty
   DOM widget and overwriting the allocation.

Also pre-fills the DNS Servers field with both IPv4 and IPv6 LAN
addresses, and repopulates the allowed_ips form widget on return from
the dialog so a subsequent Save does not overwrite the value.

### Maintainer _(preferred)_
@jow-

---

## Tested on
OpenWrt version: OpenWrt 25.12.5 r33051-f5dae5ece4
LuCI version: luci-proto-wireguard-26.183.16146~6aacb73
Web browser(s): Firefox 152.0.4, Chrome 149.0.7827.196

### DIFF
--- a/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
+++ b/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
@@ -4,7 +4,7 @@
 'use strict';
 
 import { cursor } from 'uci';
-import { popen } from 'fs';
+import { popen, opendir } from 'fs';
 
 
 function shellquote(s) {
@@ -54,6 +54,93 @@ const methods = {
 			const pub = command(`echo ${shellquote(priv)} | wg pubkey 2>/dev/null`);
 
 			return { keys: { priv, pub } };
+		}
+	},
+
+	getNextPeerAddress: {
+		args: { interface: 'interface' },
+		call: function(req) {
+			const ifname = req.args?.interface ?? 'wg0';
+			const uci = cursor();
+
+			uci.load('network');
+
+			// Derive subnet prefixes from the server interface's own addresses.
+			//
+			// IPv4: "10.0.0.1/24" → ipv4_prefix = "10.0.0"
+			//
+			// IPv6: "fd11:5ee:bad:c0de::ac7:9301/64"
+			//   The last 16-bit group encodes a high byte shared by all client
+			//   addresses plus the host octet.  We record the group base (high byte
+			//   shifted left) so we can reconstruct e.g. "930d" for octet 13.
+			//   ipv6_prefix    = "fd11:5ee:bad:c0de::ac7:"
+			//   ipv6_group_base = 0x9300
+			let ipv4_prefix = null;
+			let ipv6_prefix = null;
+			let ipv6_group_base = 0;
+
+			for (let addr in (uci.get('network', ifname, 'addresses') ?? [])) {
+				const bare = split(addr, '/')[0];
+
+				if (index(bare, ':') < 0) {
+					if (!ipv4_prefix) {
+						const parts = split(bare, '.');
+						if (length(parts) == 4)
+							ipv4_prefix = parts[0] + '.' + parts[1] + '.' + parts[2];
+					}
+				} else {
+					if (!ipv6_prefix) {
+						const last_colon = rindex(bare, ':');
+						if (last_colon >= 0) {
+							ipv6_prefix = substr(bare, 0, last_colon + 1);
+							ipv6_group_base = +('0x' + substr(bare, last_colon + 1)) & 0xff00;
+						}
+					}
+				}
+			}
+
+			if (!ipv4_prefix && !ipv6_prefix)
+				return { error: 'No server addresses configured on interface' };
+
+			// Collect the host octets already in use across all UCI peers.
+			const used = {};
+
+			uci.foreach('network', 'wireguard_' + ifname, (peer) => {
+				for (let cidr in (peer.allowed_ips ?? [])) {
+					const bare = split(cidr, '/')[0];
+
+					if (ipv4_prefix && index(bare, ':') < 0) {
+						const parts = split(bare, '.');
+						if (length(parts) == 4 &&
+						    parts[0] + '.' + parts[1] + '.' + parts[2] == ipv4_prefix)
+							used[+parts[3]] = true;
+
+					} else if (ipv6_prefix && index(bare, ':') >= 0) {
+						if (substr(bare, 0, length(ipv6_prefix)) == ipv6_prefix) {
+							const last_group = +('0x' + substr(bare, length(ipv6_prefix)));
+							used[last_group & 0xff] = true;
+						}
+					}
+				}
+			});
+
+			// Find the lowest free host value >= 2 (1 is reserved for the server).
+			let octet = 2;
+			while (used[octet] && octet < 255)
+				octet++;
+
+			if (octet >= 255)
+				return { error: 'No free addresses in pool' };
+
+			const result = { octet };
+
+			if (ipv4_prefix)
+				result.ipv4 = ipv4_prefix + '.' + octet + '/32';
+
+			if (ipv6_prefix)
+				result.ipv6 = ipv6_prefix + sprintf('%x', ipv6_group_base | octet) + '/128';
+
+			return result;
 		}
 	},
 
--- a/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
+++ b/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
@@ -66,18 +66,15 @@ const methods = {
 			uci.load('network');
 
 			// Derive subnet prefixes from the server interface's own addresses.
+			// Peers always receive /32 (IPv4) or /128 (IPv6) addresses, so the
+			// server's prefix length does not bound allocation.  We vary only the
+			// last octet (IPv4) or the low byte of the last 16-bit group (IPv6).
 			//
 			// IPv4: "10.0.0.1/24" → ipv4_prefix = "10.0.0"
-			//
 			// IPv6: "fd11:5ee:bad:c0de::ac7:9301/64"
-			//   The last 16-bit group encodes a high byte shared by all client
-			//   addresses plus the host octet.  We record the group base (high byte
-			//   shifted left) so we can reconstruct e.g. "930d" for octet 13.
-			//   ipv6_prefix    = "fd11:5ee:bad:c0de::ac7:"
-			//   ipv6_group_base = 0x9300
+			//   ipv6_prefix = "fd11:5ee:bad:c0de::ac7:", ipv6_group_base = 0x9300
 			let ipv4_prefix = null;
-			let ipv6_prefix = null;
-			let ipv6_group_base = 0;
+			let ipv6_prefix = null, ipv6_group_base = 0;
 
 			for (let addr in (uci.get('network', ifname, 'addresses') ?? [])) {
 				const bare = split(addr, '/')[0];
--- a/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
+++ b/protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
@@ -65,81 +65,109 @@ const methods = {
 
 			uci.load('network');
 
-			// Derive subnet prefixes from the server interface's own addresses.
-			// Peers always receive /32 (IPv4) or /128 (IPv6) addresses, so the
-			// server's prefix length does not bound allocation.  We vary only the
-			// last octet (IPv4) or the low byte of the last 16-bit group (IPv6).
-			//
-			// IPv4: "10.0.0.1/24" → ipv4_prefix = "10.0.0"
-			// IPv6: "fd11:5ee:bad:c0de::ac7:9301/64"
-			//   ipv6_prefix = "fd11:5ee:bad:c0de::ac7:", ipv6_group_base = 0x9300
-			let ipv4_prefix = null;
-			let ipv6_prefix = null, ipv6_group_base = 0;
+			let ipv4_net = null, ipv4_plen = 0, ipv4_server = null;
+			let ipv6_template = null;
 
 			for (let addr in (uci.get('network', ifname, 'addresses') ?? [])) {
-				const bare = split(addr, '/')[0];
+				const parts = split(addr, '/');
+				const arr   = iptoarr(parts[0]);
+				const plen  = +(parts[1] ?? 0);
 
-				if (index(bare, ':') < 0) {
-					if (!ipv4_prefix) {
-						const parts = split(bare, '.');
-						if (length(parts) == 4)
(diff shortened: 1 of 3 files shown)


protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
```diff
@@ -4,7 +4,7 @@
 'use strict';
 
 import { cursor } from 'uci';
-import { popen } from 'fs';
+import { popen, opendir } from 'fs';
```
openwrt-ai: nit: `opendir` is imported here but never used anywhere in this file — the new `getNextPeerAddress` method only uses `cursor()` and string builtins. Drop it to keep the import list accurate.

```suggestion
import { popen } from 'fs';
```

---

protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
```diff
@@ -57,6 +57,97 @@ const methods = {
...
+			let ipv6_group_base = 0;
+
+			for (let addr in (uci.get('network', ifname, 'addresses') ?? [])) {
+				const bare = split(addr, '/')[0];
+
+				if (index(bare, ':') < 0) {
+					if (!ipv4_prefix) {
+						const parts = split(bare, '.');
+						if (length(parts) == 4)
+							ipv4_prefix = parts[0] + '.' + parts[1] + '.' + parts[2];
```
openwrt-ai: The server's subnet mask is discarded (`split(addr, '/')[0]` above) and the prefix is taken as the first three octets, so allocation implicitly assumes a /24 for IPv4 and a single trailing host octet for IPv6 (line 96). For a server configured with e.g. `10.0.0.1/25` or `/26`, `octet` can climb to 254 and yield peer addresses outside the server's subnet, and the in-use scan below only matches peers sharing the same first three octets. Is restricting this to /24-style layouts intentional? If so it would be worth documenting; otherwise the mask should bound the host range.

---

openwrt-ai (commented):
Reviewed 4 new commits. Commit messages all match their diffs. Two inline notes below (one nit, one question about the subnet-mask assumption); the backend octet-allocation and hex/prefix logic otherwise checks out. Author-name/Signed-off-by formatting is already covered by the formality bot, so not repeated here.

---

egc112:
I would add this as a separate installer package:
https://github.com/openwrt/packages/tree/f31a9bf838353e12c7a14ee40de4eee22322015e/net/wg-installer

protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
```diff
@@ -57,6 +57,145 @@ const methods = {
...
+						}
+					}
+				}
+			});
+
+			// Also reserve IPs allocated this session but not yet saved to UCI.
+			// The client passes bare IPv4 addresses (for IPv4) or low-byte
+			// integers (for IPv6) in the reserved list.
+			for (let r in (req.args?.reserved ?? [])) {
+				if (index(r, '.') >= 0) {
```
openwrt-ai: IPv6 session reservations are silently dropped here. For an IPv6 allocation the client pushes a bare integer octet into the reserved list ([`_sessionReserved.push(next.octet)` at wireguard.js:735](``https://github.com/openwrt/luci/blob/3c3114b547c85cdbae2dcf94911483ae73e7e406/protocols/luci-proto-wireguard/htdocs/luci-static/resources/protocol/wireguard.js#L735)).`` But `index(<int>, '.')` returns `null` (not `-1`) for a non-string argument, and in ucode `null >= 0` evaluates to **true** (null coerces to 0). So a numeric octet takes the IPv4 branch, where `split(<int>, '.')` returns `null`, the `length(p) == 4` check fails, and `used_ipv6` is never updated.

Effect: creating two peers back-to-back in one page session on a server that has an IPv6 tunnel address assigns the same `::02` low byte to both (duplicate IPv6 `allowed_ips`) until the first peer is committed to UCI. The IPv4 side is unaffected because it passes a string. Gate on the type so integer octets fall through to the `else` branch:

```suggestion
				if (type(r) == 'string' && index(r, '.') >= 0) {
```

---

protocols/luci-proto-wireguard/root/usr/share/rpcd/ucode/luci.wireguard
```diff
@@ -57,6 +57,145 @@ const methods = {
...
+				const parts_cidr = split(addr, '/');
+				const bare = parts_cidr[0];
+				const plen = +(parts_cidr[1] ?? 0);
+
+				if (index(bare, ':') < 0) {
+					if (!ipv4_net) {
+						const p = split(bare, '.');
+						if (length(p) == 4) {
+							const ip32 = (+p[0] << 24) | (+p[1] << 16) | (+p[2] << 8) | +p[3];
+							const mask = plen ? (~0 << (32 - plen)) : 0;
```
openwrt-ai: nit / question: if an interface address is stored without a prefix (e.g. `list addresses '10.0.0.1'`), `plen` is 0, so `mask` is 0 and `ipv4_net` becomes 0. The allocator below then walks from `0.0.0.1` and hands out `0.0.0.2/32`, `0.0.0.3/32`, … — outside the server's subnet, and every existing IPv4 peer matches `(ip32 & 0) == 0`. The previous first-three-octets approach degraded more gracefully in this case. Is a prefix-less entry in `addresses` reachable in practice, or is a CIDR suffix guaranteed? If it can occur, consider defaulting to a /24 (or returning an error) when `plen == 0`.

---

openwrt-ai (commented):
Re-reviewed the changes since the last review (multi-block IPv4 allocation, `opendir` removal, and the session-reservation format change). Both earlier notes — the unused `opendir` import and the subnet-mask assumption — are addressed; the IPv4 allocation now honours the configured prefix length. Commit messages all match their diffs.

One new correctness issue (IPv6 session reservations silently dropped) and one edge-case question (prefix-less server address) inline.

---

openwrt-ai (commented):
Re-reviewed the reworked top two commits since the last review (the branch was force-pushed; `436aabd` / `1d45ef1` replace the previous `04dcef2` / `3c3114b`). The net change versus the last-reviewed tip is three lines in `luci.wireguard`, and both of them resolve the notes from my previous review:

- Prefix-less server address: `plen > 0` now guards the IPv4 subnet derivation, so a prefix-less address is skipped instead of producing a `~0 << 32` mask.
- IPv6 session reservations: the reserved-list loop now uses `type(r) == 'string' && index(r, '.') >= 0`, so IPv6 low-byte integers correctly fall through to `used_ipv6` instead of being mis-routed into the IPv4 branch and dropped. The client side (`_sessionReserved` pushing `ipv4_bare` strings and `octet` integers in `1d45ef1`) is consistent with this.

Commit messages match their diffs. No new issues found.

The remaining `FormalityCheck / Git & Commits` failure (invalid author name) is reported by the formality bot and out of scope here.

---

jow-:
Suggestion: rework IP calculations with `iptoarr()` / `arrtoip()`

`iptoarr()` converts a valid IP string to a byte array (4 elements for IPv4, 16 for IPv6, `null` for invalid); `arrtoip()` converts a byte array back to a string. Using them collapses most of the manual string-splitting and bit-manipulation in `getNextPeerAddress` and has a useful side-effect: validation and address-family detection happen in one call.

### What changes and why

| Current | With `iptoarr`/`arrtoip` |
|---|---|
| `index(bare, ':') < 0` to distinguish IPv4/IPv6 | `length(arr) == 4` vs `== 16` — family detection is unambiguous |
| `split(bare, '.')` + four `+p[n]` casts | `iptoarr(bare)` — null on invalid, no manual casting |
| `substr`/`rindex` to extract IPv6 prefix + `0x`-parse last group | Store server's 16-byte array as template, vary `arr[15]` directly |
| Four bit-shifts + string join for IPv4 result | `arrtoip([…])` |
| `type(r) == 'string' && index(r, '.')` to gate IPv4 session reservations | `iptoarr(r)` null-check + `length == 4` |

### Revised `getNextPeerAddress`

```js
getNextPeerAddress: {
    args: { interface: 'interface', reserved: [] },
    call: function(req) {
        const ifname = req.args?.interface ?? 'wg0';
        const uci = cursor();

        uci.load('network');

        let ipv4_net = null, ipv4_plen = 0, ipv4_server = null;
        let ipv6_template = null;   // 16-byte array of server IPv6 address

        for (let addr in (uci.get('network', ifname, 'addresses') ?? [])) {
            const parts = split(addr, '/');
            const arr   = iptoarr(parts[0]);   // null for invalid; 4 bytes=IPv4, 16=IPv6
            const plen  = +(parts[1] ?? 0);

            if (!arr)
                continue;

            if (length(arr) == 4) {
                // IPv4
                if (!ipv4_net && plen > 0) {
                    const ip32 = (arr[0] << 24) | (arr[1] << 16) | (arr[2] << 8) | arr[3];
                    const mask = ~0 << (32 - plen);
                    ipv4_net    = ip32 & mask;
                    ipv4_plen   = plen;
                    ipv4_server = ip32;
                }
            } else {
                // IPv6
                if (!ipv6_template)
                    ipv6_template = arr;   // keep the 16-byte server address as template
            }
        }

        if (ipv4_net === null && !ipv6_template)
            return { error: 'No server addresses configured on interface' };

        // Collect used IPv4 host addresses and used IPv6 low-bytes (arr[15]).
        const used_ipv4 = {};
        const used_ipv6 = {};

        uci.foreach('network', 'wireguard_' + ifname, (peer) => {
            for (let cidr in (peer.allowed_ips ?? [])) {
                const arr = iptoarr(split(cidr, '/')[0]);
                if (!arr) continue;

                if (ipv4_net !== null && length(arr) == 4) {
                    const ip32 = (arr[0] << 24) | (arr[1] << 16) | (arr[2] << 8) | arr[3];
                    const mask = ~0 << (32 - ipv4_plen);
                    if ((ip32 & mask) == ipv4_net)
                        used_ipv4[ip32] = true;

                } else if (ipv6_template && length(arr) == 16) {
                    // Peers sharing the same /120 (bytes 0–14 match server template)
                    // are tracked by their low byte (arr[15]).
                    let match = true;
                    for (let i = 0; i < 15; i++) {
                        if (arr[i] != ipv6_template[i]) { match = false; break; }
                    }
                    if (match)
                        used_ipv6[arr[15]] = true;
                }
            }
        });

        // Session reservations: client pushes bare IPv4 strings or integer
        // low-bytes for IPv6. iptoarr() handles validation and family detection.
        for (let r in (req.args?.reserved ?? [])) {
            if (type(r) == 'string') {
                const arr = iptoarr(r);
                if (arr && length(arr) == 4)
                    used_ipv4[(arr[0] << 24) | (arr[1] << 16) | (arr[2] << 8) | arr[3]] = true;
            } else {
                used_ipv6[+r] = true;
            }
        }

        const result = {};

        if (ipv4_net !== null) {
            const host_bits = 32 - ipv4_plen;
            const total     = 1 << host_bits;
            let   found     = null;

            for (let i = 1; i < total - 1; i++) {
                const last = i & 0xff;
                if (last == 0 || last == 255) continue;
                const candidate = ipv4_net + i;
                if (candidate == ipv4_server) continue;
                if (!used_ipv4[candidate]) { found = candidate; break; }
            }

            if (found === null)
                return { error: 'No free addresses in pool' };

            // arrtoip() replaces four manual bit-shifts + string concatenation.
            const bare      = arrtoip([(found >> 24) & 0xff, (found >> 16) & 0xff,
                                        (found >>  8) & 0xff,  found        & 0xff]);
            result.ipv4      = bare + '/32';
            result.ipv4_bare = bare;
        }

        if (ipv6_template) {
            let octet = 2;
            while (used_ipv6[octet] && octet < 255)
                octet++;

            if (octet >= 255)
                return { error: 'No free addresses in pool' };

            // Copy the server template, overwrite the low byte, reconstruct.
            const peer_arr  = ipv6_template[0:];   // ucode slice-copy
            peer_arr[15]    = octet;
            result.ipv6     = arrtoip(peer_arr) + '/128';
            result.octet    = octet;
        }

        return result;
    }
},
```

### Notes

- **`ipv6_group_base` disappears** — it was the high byte of the last 16-bit group, which is now simply `ipv6_template[14]`, kept implicitly by only overwriting `peer_arr[15]`.
- **IPv6 prefix matching by byte comparison** (bytes 0–14) replaces the string-prefix comparison and is more robust against non-canonical IPv6 representations that `arrtoip()` might normalise differently.
- **Session reserved list**: the old `type(r) == 'string' && index(r, '.')` guard is replaced by `iptoarr(r)` + `length == 4`. A string like `"not.an.ip"` passes the dot-index check but fails `iptoarr()`, so invalid entries are silently ignored rather than misprocessed.

jow-:
Sorry for the ai generated reply, I instructed claude to send you a counter proposal to illustrate my suggestion. It hallucinated that `ucode slice copy` expression which should be this instead:
`const peer_arr = [...ipv6_template]`

BdN3504:
> Sorry for the ai generated reply, I instructed claude to send you a counter proposal to illustrate my suggestion. It hallucinated that `ucode slice copy` expression which should be this instead: `const peer_arr = [...ipv6_template]`

Thanks for your comments, i used your suggestions to push another commit. Do you think this pr is useful? Also, is it absolutely necessary to use government names for the S.O.B. lines? I don't have a github account under my real name.

openwrt-ai (commented):

The net change since my last review is the `getNextPeerAddress` rewrite onto `iptoarr()`/`arrtoip()` (`d490d27`, `a581a00`), matching the counter-proposal in the maintainer thread. Verified it is behaviour-equivalent and more robust: `iptoarr`/`arrtoip` are ucode core built-ins (no import needed); IPv4 family detection via `length(arr) == 4` and the int32 mask/iteration are correct with `.0`/`.255`/server skipped; the IPv6 byte-array template match (bytes 0–14) replaces the earlier `substr` string-prefix comparison and handles non-canonical IPv6 correctly; and the `type(r) == 'string'` session-reservation split is consistent with the client pushing `ipv4_bare` strings and `octet` integers. Commit messages match their diffs.

The `FormalityCheck / Git & Commits` failure (invalid author name) is raised by the formality bot and out of scope here.

---

egc112:
Instead of adding this to the source code consider making an add-on package see e.g.:
https://forum.openwrt.org/t/wireguard-1-click-peer-generator/252038

PR #8815, opened 2026-07-09
luci-mod-status: channel_analysis: draw graph on first tab activation

RichieMay:
The channel graph for every radio/band was drawn via requestAnimationFrame() right after each tab was created. For tab panes still hidden at that point, offsetWidth is 0 (themes that hide inactive panes with display:none), so the column width collapsed and the graph was drawn squished until the tab was first opened.

Defer create_channel_graph() to a cbi-tab-active handler attached in render and draw each graph exactly once, guarded by a "created" flag, when its tab first becomes visible and offsetWidth is correct. Store the band channels on the graph data and start the poll from the same handler. Move the poll-start logic out of create_channel_graph, which becomes a pure drawing function.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
@@ -185,12 +185,6 @@ return view.extend({
 			}
 		}
 		createGraphHLine(G,curr_offset+step, 0.1, 1);
-
-		chan_analysis.tab.addEventListener('cbi-tab-active', L.bind(function(ev) {
-			this.active_tab = ev.detail.tab;
-			if (!this.radios[this.active_tab].loadedOnce)
-				poll.start();
-		}, this));
 	},
 
 	handleScanRefresh: function() {
@@ -475,6 +469,8 @@ return view.extend({
 					offset_tbl: {},
 					col_width: 0,
 					tab: tab,
+					created: false,
+					channels: bands[band].channels,
 				};
 
 				this.radios[ifname+band] = {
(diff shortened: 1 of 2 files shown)


modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
```diff
@@ -488,9 +484,24 @@ return view.extend({
 
 				cbi_update_table(table, [], E('em', { class: 'spinning' }, _('Starting wireless scan...')));
 
-				tabs.firstElementChild.appendChild(tab)
+				tabs.firstElementChild.appendChild(tab);
+
+				/* Draw on first tab activation: column width comes from
+				 * offsetWidth, which is 0 while the pane is hidden, so
+				 * rendering all tabs upfront squishes inactive ones. */
+				tab.addEventListener('cbi-tab-active', L.bind(function(ev) {
```
openwrt-ai, line 492: This fix targets only the stable `openwrt-24.10` branch, but the same bug is still present on `master`: it also draws every graph upfront via [`requestAnimationFrame(this.create_channel_graph …)` at channel_analysis.js:491](``https://github.com/openwrt/luci/blob/f7e59912304972e500fa6caf63d925c671b26cb6/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js#L491)`` and keeps the `cbi-tab-active` listener [inside `create_channel_graph` at :189](``https://github.com/openwrt/luci/blob/f7e59912304972e500fa6caf63d925c671b26cb6/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js#L189),`` so inactive panes render squished there too.

OpenWrt expects fixes to land on `master` first and then be cherry-picked (`git cherry-pick -x`) to the stable branch — which is also why the formality check is asking for a `(cherry picked from commit …)` line. Note that `master` has since been refactored to ES6 (`const`/`let`, method shorthand, destructuring), so this exact diff won't apply cleanly there; an equivalent fix would need to be written against `master`. Is fixing `master` planned as a follow-up?

---

openwrt-ai (commented):
Reviewed 1 new commit. The change itself is functionally correct — the initially-active tab still receives `cbi-tab-active` (dispatched via `initTabGroup`'s `requestAnimationFrame`), and `switchTab` sets `data-tab-active="true"` before dispatching, so `offsetWidth` is valid when each graph is drawn once. One note left inline about this being applied only to the stable branch.

---

PR #8820, opened 2026-07-10
luci-mod-status: channel_analysis: draw graph on first tab activation

RichieMay:
The channel graph for every radio/band was drawn via requestAnimationFrame() right after each tab was created. For tab panes still hidden at that point, offsetWidth is 0 (themes that hide inactive panes with display:none), so the column width collapsed and the graph was drawn squished until the tab was first opened.

Defer create_channel_graph() to a cbi-tab-active handler attached in render and draw each graph exactly once, guarded by a "created" flag, when its tab first becomes visible and offsetWidth is correct. Store the band channels on the graph data and start the poll from the same handler. Move the poll-start logic out of create_channel_graph, which becomes a pure drawing function.

### DIFF
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
@@ -466,7 +466,6 @@ return view.extend({
 					graph: csvg,
 					offset_tbl: {},
 					col_width: 0,
-					tab: tab,
 					created: false,
 					channels: bands[band].channels,
 				};
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
@@ -185,12 +185,6 @@ return view.extend({
 			}
 		}
 		createGraphHLine(G,curr_offset+step, 0.1, 1);
-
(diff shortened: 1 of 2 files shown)


modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js
```diff
@@ -473,6 +467,8 @@ return view.extend({
 					offset_tbl: {},
 					col_width: 0,
 					tab: tab,
```
openwrt-ai: nit: the `tab` field on `graph_data` is now dead code. Its only reader was the `chan_analysis.tab.addEventListener(...)` block inside `create_channel_graph`, which this PR removes (the new listener is attached to the local `tab` variable in `render` instead). Nothing else reads `radio.graph.tab` — `handleScanRefresh`, `add_wifi_to_graph`, and `create_channel_graph` only use `.graph`, `.offset_tbl`, and `.col_width`. Consider dropping this field.

---

openwrt-ai (commented):
Reviewed the single new commit; the refactor is internally consistent and behavior-preserving. One minor cleanup nit inline. No blocking issues found.

---

openwrt-ai (commented):

The new commit removes the dead `graph_data.tab` field flagged in the prior review; the removal is clean (the only remaining `.tab` reference is `ev.detail.tab`, an event property) and the commit message accurately describes the change.

---

PR #8833, opened 2026-07-16
luci-app-qosify: add new package

choppyc79:
Adds luci-app-qosify, a JavaScript LuCI interface for the qosify
traffic shaping daemon (CAKE + eBPF DSCP classifier).

Features:
- Overview tab with service status, shaping detection and quick
  settings for all interface options
- Inline editors for /etc/config/qosify and the DSCP classification
  rules file, with guided quick-add forms and validation
- Backup/upload/reset of config files
- Live qosify-status output

No Lua, no luci-compat. Depends on qosify only. All user-visible
strings are translatable.

Previously distributed standalone and in use since April 2025:
https://github.com/choppyc79/luci-app-qosify
https://forum.openwrt.org/t/248979

### DIFF
--- a/applications/luci-app-qosify/Makefile
+++ b/applications/luci-app-qosify/Makefile
@@ -1,0 +1,13 @@
+include $(TOPDIR)/rules.mk
+
+PKG_LICENSE:=MIT
+PKG_MAINTAINER:=Ash Clarke <clarkeaj@hotmail.co.uk>
+
+LUCI_TITLE:=LuCI interface for qosify
+LUCI_DESCRIPTION:=Web UI for the qosify CAKE/eBPF traffic shaping daemon
+LUCI_DEPENDS:=+qosify
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
+++ b/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
@@ -1,0 +1,1355 @@
+'use strict';
+'require view';
+'require fs';
+'require ui';
+'require uci';
+'require poll';
+'require rpc';
+'require dom';
+
+var UCI_PATH='/etc/config/qosify';
+var RULES_PATH='/etc/qosify/00-defaults.conf';
+var DSCP=['CS0','CS1','CS2','CS3','CS4','CS5','CS6','CS7','AF11','AF12','AF13','AF21','AF22','AF23','AF31','AF32','AF33','AF41','AF42','AF43','EF','VA','NQB','LE','DF'];
+var OVH=['none','manual','conservative','ethernet','docsis','pppoe-ptm','bridged-ptm','pppoe-vcmux','pppoe-llcsnap','pppoa-vcmux','pppoa-llc','bridged-vcmux','bridged-llcsnap','ipoa-vcmux','ipoa-llcsnap'];
+var ENCAP=['atm','noatm','ptm'];
+var MODES=['diffserv3','diffserv4','diffserv8','besteffort','precedence'];
+// qosify.init handles 'alias' with add_class and 'device' with add_interface,
+// so those section types share the option set of class / interface.
+var QAC_PANEL={defaults:'defaults','class':'class',alias:'class','interface':'interface',device:'interface'};
+var SECT=[['defaults','config defaults'],['class','config class'],['alias','config alias'],['interface','config interface'],['device','config device']];
+
+var callInit=rpc.declare({
+	object:'luci',
+	method:'setInitAction',
+	params:['name','action'],
+	expect:{result:false}
+});
+var callServiceList=rpc.declare({
+	object:'service',
+	method:'list',
+	params:['name'],
+	expect:{'':{}}
+});
+var callUciRevert=rpc.declare({
+	object:'uci',
+	method:'revert',
+	params:['config']
+});
+function isRunning(r){
+	try{var i=r.qosify.instances;for(var k in i)if(i[k].running)return true;}catch(e){}
+	return false;
+}
+
+function clsLabel(c){return c.name+(c.alias?' '+_('(alias)'):'');}
+function clsDesc(c){return _('Ingress: %s / Egress: %s').format(c.ingress||'',c.egress||'');}
+function esc(s){return (s==null?'':String(s)).replace(/[&<>"']/g,function(c){return{'&':'&amp;','<':'&lt;','>':'&gt;','"':'&quot;',"'":'&#39;'}[c]});}
+function trim(s){return (s||'').replace(/^\s+|\s+$/g,'');}
+function $(id){return document.getElementById(id);}
+
+function detectActive(out){return /qdisc cake|: active/.test(out||'');}
+
+function countRules(text){
+	var n=0,lines=(text||'').split('\n');
+	for(var i=0;i<lines.length;i++){
+		var l=lines[i],h=l.indexOf('#');
+		if(h>=0)l=l.slice(0,h);
+		if(trim(l))n++;
+	}
+	return n;
+}
+
+function validateRules(d){
+	if(/\x00/.test(d))return _('Binary content rejected');
+	var lines=d.split('\n');
+	for(var i=0;i<lines.length;i++){
+		var l=lines[i],h=l.indexOf('#');
+		if(h>=0)l=l.slice(0,h);
+		l=trim(l);
+		if(l&&!/^\S+\s+\S/.test(l))return _('Invalid rule line: %s').format(l.slice(0,40));
+	}
+	return null;
+}
+function fmtSize(n){return n<1024?n+'B':(n/1024).toFixed(1)+'K';}
+function fmtMtime(t){if(!t)return '';return new Date(t*1000).toLocaleString();}
+
+// The shaping section Quick Settings edits, or null. Prefers the first enabled
+// section, and accepts `config device` since qosify.init feeds both section
+// types through add_interface(). An anonymous section has a synthetic .name
+// (cfgXXXXXX / newXXXXXX) that never appears in the file, so name is left empty
+// for it and setOpts() locates the block by per-type ordinal instead.
+function ifSect(){
+	var a=[];
+	['interface','device'].forEach(function(t){
+		var i=0;
+		uci.sections('qosify',t,function(s){
+			a.push({type:t,id:s['.name'],name:s['.anonymous']?'':s['.name'],idx:i++,on:s.disabled!=='1'});
+		});
+	});
+	for(var j=0;j<a.length;j++)if(a[j].on)return a[j];
+	return a.length?a[0]:null;
+}
+// --- Mirrors qosify.init add_interface() and cmd_add_qdisc() in interface.c ---
+// nat defaults to 1 for interfaces and 0 for devices; host_isolate defaults on.
+function ifCfg(s,dev){
+	return {
+		bw_up:s.bandwidth_up||s.bandwidth||'',
+		bw_dn:s.bandwidth_down||s.bandwidth||'',
+		mode:s.mode||'diffserv4',
+		ingress:s.ingress!=='0',
+		egress:s.egress!=='0',
+		host_isolate:s.host_isolate!=='0',
+		autorate:s.autorate_ingress==='1',
+		nat:(s.nat==null||s.nat==='')?!dev:(s.nat!=='0')
+	};
+}
+function hasNat(v){return /(^|\s)nat(\s|$)/.test(v||'');}
+// Keys the daemon will silently drop, given the rest of the section.
+function ifLint(s,dev){
+	var w=[],c=ifCfg(s,dev);
+	if(!s.name)w.push(_('name is not set — qosify.init sends an empty device name and this section is never applied'));
+	if(!c.host_isolate&&s.nat==='1'){
+		var ne=hasNat(s.options)||hasNat(s.egress_options);
+		var ni=hasNat(s.options)||hasNat(s.ingress_options);
+		if(!ne&&!ni)w.push(_('nat is not sent: qosify only emits nat/nonat inside the host_isolate branch. CAKE does accept flows plus nat — put nat in options to apply it'));
+		else if(!ne||!ni)w.push(_('nat only reaches %s — put it in options, or in both ingress_options and egress_options').format(ne?'egress':'ingress'));
+	}
+	if(s.overhead_type!=='manual'&&(s.overhead||s.overhead_encap))w.push(_('overhead and overhead_encap are ignored unless overhead_type is manual'));
+	if(!c.ingress&&!c.egress)w.push(_('ingress and egress are both 0 — nothing is shaped'));
+	if(c.egress&&!c.bw_up)w.push(_('no bandwidth_up or bandwidth — egress CAKE runs unlimited'));
+	if(c.ingress&&!c.bw_dn)w.push(_('no bandwidth_down or bandwidth — ingress CAKE runs unlimited'));
+	['bandwidth_up','bandwidth_down','bandwidth','mode','ingress_options','egress_options','options'].forEach(function(k){
+		if(s[k]&&String(s[k]).indexOf("'")>=0)w.push(_('%s contains a quote — qosify rejects the whole value').format(k));
+	});
+	return w;
+}
+// Locate config blocks in raw UCI text: {type,name,start,end} (end = last non-blank
+// line). Headers may be bare, single- or double-quoted — all three are valid UCI.
+function unq(s){return String(s||'').replace(/^["']|["']$/g,'');}
+function cfgSections(txt){
+	var out=[],cur=null,lines=(txt||'').split('\n');
+	for(var i=0;i<lines.length;i++){
+		var m=/^\s*config\s+(\S+)(?:\s+(\S+))?\s*$/.exec(lines[i]);
+		if(m){cur={type:unq(m[1]),name:unq(m[2]),start:i,end:i};out.push(cur);}
+		else if(cur&&trim(lines[i])!=='')cur.end=i;
+	}
+	return out;
+}
+// Values are spliced into a single-quoted UCI string: strip quotes and line breaks,
+// or a stray newline injects arbitrary option/config lines into the file.
+function qv(v){return v==null?'':String(v).replace(/['"\r\n]/g,'');}
+// Set/remove options inside one config block, preserving every other byte of the
+// file (comments, ordering, lists, unknown options). kv[key]===null deletes.
+// idx = ordinal among sections of this type, used when name is empty (anonymous).
+function setOpts(txt,type,name,idx,kv){
+	var lines=(txt||'').split('\n'),secs=cfgSections(txt),s=null,n=0,i,k;
+	for(i=0;i<secs.length;i++){
+		if(secs[i].type!==type)continue;
+		if(name?secs[i].name===name:n++===idx){s=secs[i];break;}
+	}
+	if(!s){
+		var blk=["config "+type+(name?" '"+name+"'":'')];
+		for(k in kv)if(kv[k]!=null)blk.push("\toption "+k+" '"+qv(kv[k])+"'");
+		var t=(txt||'').replace(/\s+$/,'');
+		return (t?t+'\n\n':'')+blk.join('\n')+'\n';
+	}
+	var out=[lines[s.start]],seen={};
+	for(i=s.start+1;i<=s.end;i++){
+		var m=/^\s*option\s+(\S+)\s+(.*)$/.exec(lines[i]);
+		if(m&&(m[1] in kv)){
+			seen[m[1]]=1;
+			if(kv[m[1]]!=null)out.push("\toption "+m[1]+" '"+qv(kv[m[1]])+"'");
+			continue;
+		}
+		out.push(lines[i]);
+	}
+	for(k in kv)if(!seen[k]&&kv[k]!=null)out.push("\toption "+k+" '"+qv(kv[k])+"'");
+	return lines.slice(0,s.start).concat(out,lines.slice(s.end+1)).join('\n');
+}
+// Non-blocking sanity pass: flag rule targets that are neither a defined class,
+// a DSCP codepoint, nor a raw numeric value.
+function ruleWarn(txt,names){
+	var bad=[],lines=(txt||'').split('\n');
+	for(var i=0;i<lines.length;i++){
+		var l=lines[i],h=l.indexOf('#');
+		if(h>=0)l=l.slice(0,h);
+		l=trim(l);if(!l)continue;
+		var f=l.split(/\s+/);if(f.length<2)continue;
+		var v=f[1].replace(/^\+/,'');
+		if(names.indexOf(v)>=0||DSCP.indexOf(v)>=0)continue;
+		if(/^(0[xX][0-9a-fA-F]+|\d+)$/.test(v)){
+			if(parseInt(v,v.charAt(0)==='0'&&(v.charAt(1)==='x'||v.charAt(1)==='X')?16:10)<64)continue;
+		}
+		if(bad.indexOf(v)<0)bad.push(v);
+	}
+	return bad.length?_('Unknown class/DSCP target: %s').format(bad.slice(0,5).join(', ')):null;
+}
+
+var noteSeen={};
+function notify(msg,kind){
+	var key=String(msg);
+	if(noteSeen[key])return null;
+	var n=ui.addNotification(null,E('p',{},msg),kind||'info');
+	if(!n)return null;
+	noteSeen[key]=1;
+	var ms=(kind==='danger')?10000:(kind==='warning')?8000:5000;
+	setTimeout(function(){
+		delete noteSeen[key];
+		if(n&&n.parentNode)n.parentNode.removeChild(n);
+	},ms);
+	return n;
+}
+
+return view.extend({
+	handleSaveApply:null,handleSave:null,handleReset:null,
+	currentTab:'ov',
+
+	load:function(){
+		return Promise.all([
+			uci.load('qosify').catch(function(){return null;}),
+			L.resolveDefault(fs.read(RULES_PATH),''),
+			L.resolveDefault(fs.read(UCI_PATH),''),
+			L.resolveDefault(fs.stat(UCI_PATH),null),
+			L.resolveDefault(fs.stat(RULES_PATH),null),
+			L.resolveDefault(callServiceList('qosify'),{}),
+			L.resolveDefault(fs.exec('/etc/init.d/qosify',['enabled']),{code:1}),
+			L.resolveDefault(fs.stat('/usr/sbin/qosify'),null),
+			L.resolveDefault(fs.stat('/etc/init.d/qosify'),null),
+			L.resolveDefault(fs.exec('/usr/sbin/qosify-status',[]),{stdout:''})
+		]);
+	},
+
+	render:function(d){
+		var ctx={
+			rulesText:d[1]||'',
+			cfgRaw:d[2]||'',
+			cfgStat:d[3],
+			rulesStat:d[4],
+			running:isRunning(d[5]),
+			enabled:d[6].code===0,
+			hasBin:d[7]!=null,
+			hasInit:d[8]!=null,
+			qstatus:(d[9]&&d[9].stdout)||'',
+		};
+		ctx.active=detectActive(ctx.qstatus);
+
+		var root=E('div',{'class':'cbi-map','id':'qos-app'});
+		root.appendChild(E('style',{},this.css()));
+		root.appendChild(E('h2',{},'qosify'));
+		root.appendChild(E('div',{'class':'cbi-map-descr'},_('Traffic shaping and DSCP classification via qosify')));
+
+		var tabs=E('ul',{'class':'cbi-tabmenu'});
+		var tabDef=[['ov',_('Overview')],['cf',_('Config')],['ru',_('Classification Rules')],['ad',_('Advanced')],['st',_('Status')]];
+		var self=this;
+		tabDef.forEach(function(t){
+			var li=E('li',{'class':'cbi-tab-disabled','id':'th-'+t[0]},
+				E('a',{'href':'#','click':function(ev){ev.preventDefault();self.showTab(t[0]);}},t[1]));
+			tabs.appendChild(li);
+		});
+		root.appendChild(tabs);
+
+		root.appendChild(this.tabOverview(ctx));
+		root.appendChild(this.tabConfig(ctx));
+		root.appendChild(this.tabRules(ctx));
+		root.appendChild(this.tabAdvanced(ctx));
+		root.appendChild(this.tabStatus(ctx));
+
+		var hash=(location.hash||'').slice(1);
+		var map={overview:'ov',config:'cf',rules:'ru',advanced:'ad',status:'st'};
+		setTimeout(function(){self.showTab(map[hash]||'ov');},0);
+
+		this.installPollers();
+		return root;
+	},
+
+	installPollers:function(){
+		var self=this;
+		poll.add(function(){if(self.currentTab!=='ov'||self._n)return;return self.refreshOverview();},10);
+		poll.add(function(){return self.refreshStatus();},5);
+	},
+
+	showTab:function(t){
+		var dirty=this.dirty();
+		if(dirty&&(this.currentTab==='cf'||this.currentTab==='ru')&&t!==this.currentTab){
+			if(!confirm(_('You have unsaved changes. Leave this tab?')))return;
+		}
+		this.currentTab=t;
+		['ov','cf','ru','ad','st'].forEach(function(x){
+			var el=$('qos-'+x),th=$('th-'+x);
+			if(!el||!th)return;
+			if(x===t){el.style.display='block';th.className='cbi-tab';}
+			else{el.style.display='none';th.className='cbi-tab-disabled';}
+		});
+		var rev={ov:'overview',cf:'config',ru:'rules',ad:'advanced',st:'status'};
+		try{history.replaceState(null,'','#'+rev[t]);}catch(e){}
+	},
+
+	dirty:function(){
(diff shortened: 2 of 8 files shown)


applications/luci-app-qosify/po/templates/qosify.pot
openwrt-ai, line 1: This translation template is committed empty (0 bytes), but the view uses `_()` on ~100 user-visible strings. All 100 other app templates in `applications/*/po/templates/*.pot` are non-empty, so this deviates from the tree convention and means translators have nothing to work from — which also conflicts with the PR description's "All user-visible strings are translatable."

Regenerate the template so it contains the extracted `msgid` entries for every `_()` string in `main.js` (the standard LuCI mechanism is `./build/i18n-scan.pl` over the package directory), then commit the populated `qosify.pot`.

---

openwrt-ai (commented):
Reviewed 1 new commit (new package `luci-app-qosify`). Backend consumer checks out: the `qosify` daemon in openwrt.git reads `/etc/config/qosify` and `/etc/qosify/*.conf`, and the bundled defaults match its shipped config. The rpcd ACL covers every ubus/file/uci/exec path the view uses and is path-scoped; no XSS was found (the two `innerHTML` sites are static or `esc()`-wrapped) and all `fs.exec` calls use fixed paths with no user-supplied arguments.

One issue inline: the translation template is committed empty.

---

applications/luci-app-qosify/po/templates/qosify.pot
choppyc79: regenerated via ./build/i18n-scan.pl applications/luci-app-qosify, 135 entries

choppyc79 (commented):
regenerated via ./build/i18n-scan.pl applications/luci-app-qosify, 135 entries

openwrt-ai (commented):

The only change since the last review is the regenerated `po/templates/qosify.pot` (135 entries via `i18n-scan.pl`), which resolves the empty-translation-template issue flagged previously.

---

hnyman:
cc @nbd168 who authored qosify

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1085 @@
+'use strict';
+'require view';
+'require fs';
+'require ui';
+'require uci';
+'require poll';
+'require rpc';
+'require dom';
+
+var VER='2.5.6';
```
openwrt-ai: nit: this hand-maintained version constant is rendered in the UI as [`'luci-app-qosify v'+VER` at main.js:140](``https://github.com/openwrt/luci/blob/96ff5ba70147637196a8442e82d6558f1d649e45/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L140),`` but nothing in the package keeps it in sync — the Makefile has no `PKG_VERSION`, so once this is in-tree the packaged version comes from the luci.mk/git-derived version and `VER` will only change when someone remembers to edit this line. Grepping `applications/*/htdocs/luci-static/resources/view/` turns up no other in-tree luci app that carries a hardcoded version string in a view, so this looks like a leftover from the standalone distribution.

Worth either dropping the version badge, or adding a `PKG_VERSION` to the Makefile and treating that as the single source of truth. (This bump from `2.5.5` is also the only user-visible change in the force-push; the other two hunks are lint cleanups.)

---

openwrt-ai (commented):
Reviewed the force-pushed head (`5d9af89` → `96ff5ba`). The branch was squashed to a single commit; the only content change since the last review is 3 lines in `view/qosify/main.js`:

- `/^[\w\s\-\.]*$/` → `/^[\w\s.-]*$/` — semantically identical (inside a character class `.` is literal and a trailing `-` is literal), removes the useless escapes.
- `for(var i=...)` → `for(i=...)` at line 1001 — safe, `i` is already declared with `var` at line 991 in the same function scope, so this drops a redeclaration rather than creating an implicit global.
- `VER` bump — one nit inline.

The commit message matches the diff (whole new package). CI on the head SHA is green/queued; no failures to attribute.

---

hnyman:
Please do not declare in-code package version variables.  Usually there is no versioning visible inside the packages.

Either rely on LuCI's automatic versioning like most packages do, or declare a PKG_VERSION in the Makefile.

choppyc79:
> Please do not declare in-code package version variables. Usually there is no versioning visible inside the packages.
> 
> Either rely on LuCI's automatic versioning like most packages do, or declare a PKG_VERSION in the Makefile.

Dropped the version badge and the in-code VER constant — no PKG_VERSION added, so
the package relies on LuCI's automatic versioning like the other in-tree apps.
The removed string wasn't translatable, so the .pot is unchanged.

Amended into the existing commit and force-pushed.

openwrt-ai (commented):

The only change since the last review (`96ff5ba` → `c13c738`) is a 5-line deletion in `view/qosify/main.js`: the `VER='2.5.6'` constant and the version badge `<div>` appended at the end of `render()`. This addresses the maintainer request to drop in-code package versioning.

Verified:
- No remaining references to `VER` or a hardcoded version string anywhere under `applications/luci-app-qosify/` — the removal is complete, no dangling identifier.
- No `PKG_VERSION` was added to the Makefile, so the package falls back to LuCI's automatic versioning as requested.
- The removed literal (`'luci-app-qosify v'+VER`) was not wrapped in `_()`, so leaving `po/templates/qosify.pot` untouched is correct.
- Commit message still matches the diff (whole new package); no CI failures on the head SHA.

---

applications/luci-app-qosify/Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+PKG_LICENSE:=MIT
+PKG_MAINTAINER:=Ash Clarke <clarkeaj@hotmail.co.uk>
+
+LUCI_TITLE:=LuCI interface for qosify
+LUCI_DESCRIPTION:=Web UI for the qosify CAKE/eBPF traffic shaping daemon
+LUCI_DEPENDS:=+qosify
+LUCI_PKGARCH:=all
+
+include ../../luci.mk
```
openwrt-ai, line 11: `applications/luci-app-qosify/po/templates/qosify.pot` is gone from this force-push — it was present at `c13c738` (569 lines, the 135 entries you generated with `i18n-scan.pl`), and the PR now ships 7 files with no `po/` directory at all.

Every one of the 100 other packages under `applications/*/` carries a non-empty `po/templates/*.pot`, and `luci.mk` derives the `luci-i18n-qosify-*` packages from it — without the template nothing in `main.js` is translatable, which contradicts "All user-visible strings are translatable" in the PR description and re-opens the issue that was fixed earlier in this PR.

Note that restoring the old file is not enough: this commit adds a large batch of new `_()` strings (the `ifLint()` warnings, the `refTable()` per-section notes, the reworked reference/help text, the new save and lint notifications), so the template needs to be regenerated against the current `main.js`.

Was the removal intentional?

---

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1291 @@
...
+		else if(cur&&trim(lines[i])!=='')cur.end=i;
+	}
+	return out;
+}
+function qv(v){return String(v).replace(/'/g,'');}
+// Set/remove options inside one config block, preserving every other byte of the
+// file (comments, ordering, lists, unknown options). kv[key]===null deletes.
+function setOpts(txt,type,name,kv){
+	var lines=(txt||'').split('\n'),secs=cfgSections(txt),s=null,i,k;
+	for(i=0;i<secs.length;i++)if(secs[i].type===type&&(!name||secs[i].name===name)){s=secs[i];break;}
```
openwrt-ai: This lookup can never match an anonymous section, and the fallback silently duplicates it.

`ifSect()` returns `s['.name']`, which for an anonymous section — `config interface` with no name, or anything created by `uci add qosify interface` — is a synthetic id like `cfg0292bd` ([uci.js: "either an anonymous ID in the form `cfgXXXXXX` or `newXXXXXX`… or a string holding the name of the section"](https://github.com/openwrt/luci/blob/cdd57a26c3d39f147dd9968ffd004c77a9341192/modules/luci-base/htdocs/luci-static/resources/uci.js#L404-L407)). `cfgSections()` only ever sees names that literally appear in `/etc/config/qosify`, so `secs[i].name === name` fails, `s` stays null, and the `if(!s)` branch appends a brand-new `config interface 'cfg0292bd'` block at the end of the file. The user ends up with two interface sections describing the same device, and `add_interface()` in `qosify.init` will emit `json_add_object "$name"` twice for it. Line 895 compounds this by then writing `option name 'cfg0292bd'` as the netdev name.

Since `saveQs()` now bypasses `uci.set()`/`uci.apply()` entirely and rewrites the file itself, this path has no safety net. Consider having `ifSect()` skip anonymous sections (uci.js exposes a `.anonymous` flag alongside `.name`), or resolving them by ordinal index into `cfgSections()` instead of by name.

The same lookup also misses double-quoted section headers (`config interface "wan"` is valid UCI): the name group `'?([^'\s]+)'?` captures `"wan"` with the quotes still attached.

---

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1291 @@
...
+	getClasses:function(){
+		var arr=[];
+		uci.sections('qosify','class',function(s){
+			arr.push({name:s['.name'],ingress:s.ingress||'',egress:s.egress||'',
+				dscp_prio:s.dscp_prio||'',dscp_bulk:s.dscp_bulk||'',
+				prio_max_avg_pkt_len:s.prio_max_avg_pkt_len||'',
+				bulk_trigger_pps:s.bulk_trigger_pps||'',
+				bulk_trigger_timeout:s.bulk_trigger_timeout||''});
+		});
+		return arr;
```
openwrt-ai, line 677: `config alias` is now offered in the Quick Add dropdown (`SECT`, line 19) and documented in `refTable()` as a second name for a class, but `getClasses()` still enumerates `class` sections only. `qosify.init` runs both through `add_class()` — [`config_foreach add_class class` / `config_foreach add_class alias` at qosify.init:146-147](https://github.com/openwrt/openwrt/blob/c9833b993d0157bed916203d1ae342b94608a982/package/network/config/qosify/files/qosify.init#L146-L147) — so alias names are real class names to the daemon. As written, a rule pointing at an alias is reported by the new `ruleWarn()` as `Unknown class/DSCP target`, and aliases are absent from the Quick Add Rule class picker and from `dscpChoices`.

Related, from the same commit: the class panel gained `option value` (line 499), but `add_class()` resolves `json_add_string ingress "${ingress:-$value}"`, so a class defined with only `value` renders as `Ingress:  / Egress: ` in the reference box.

```suggestion
		var arr=[];
		function add(s){
			arr.push({name:s['.name'],ingress:s.ingress||s.value||'',egress:s.egress||s.value||'',
				dscp_prio:s.dscp_prio||'',dscp_bulk:s.dscp_bulk||'',
				prio_max_avg_pkt_len:s.prio_max_avg_pkt_len||'',
				bulk_trigger_pps:s.bulk_trigger_pps||'',
				bulk_trigger_timeout:s.bulk_trigger_timeout||''});
		}
		uci.sections('qosify','class',add);
		uci.sections('qosify','alias',add);
		return arr;
```

---

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1291 @@
...
+function cfgSections(txt){
+	var out=[],cur=null,lines=(txt||'').split('\n');
+	for(var i=0;i<lines.length;i++){
+		var m=/^\s*config\s+(\S+)(?:\s+'?([^'\s]+)'?)?\s*$/.exec(lines[i]);
+		if(m){cur={type:m[1],name:m[2]||'',start:i,end:i};out.push(cur);}
+		else if(cur&&trim(lines[i])!=='')cur.end=i;
+	}
+	return out;
+}
+function qv(v){return String(v).replace(/'/g,'');}
```
openwrt-ai: `qv()` removes `'` so a value cannot break out of the single-quoted string `setOpts()` splices it into, but it leaves `\n` and `\r` alone — and `setOpts()` writes the result directly to `/etc/config/qosify`. Values are not all form-validated: `sel()` re-offers any unrecognised existing config value verbatim as `<option value="…">… (current)` ([line 330](``https://github.com/openwrt/luci/blob/cdd57a26c3d39f147dd9968ffd004c77a9341192/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L330)),`` so `overhead_type`/`mode` can carry an embedded newline straight back out and inject arbitrary `option`/`config` lines into the file. Strip line breaks here as well.

```suggestion
function qv(v){return String(v).replace(/['\r\n]/g,'');}
```

---

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1291 @@
...
+		qacRow.appendChild(qacType);
+		qacRow.appendChild(E('span',{'id':'qac-nm-w','style':'display:none'},
+			E('input',{'id':'qac-name','type':'text','placeholder':'section name','style':'width:120px;font-family:monospace'})));
+		qacRow.appendChild(E('button',{'class':'cbi-button cbi-button-add','click':function(){return self.qacAdd();}},_('Add')));
+		qa.appendChild(qacRow);
+
+		// config defaults — add_defaults() in qosify.init
+		var qadDef=E('div',{'class':'qos-qa-row','id':'qac-opts-defaults'});
+		this.qaInput(qadDef,'defaults','list','/etc/qosify/*.conf',180);
+		this.qaNum(qadDef,'timeout','300',60);
```
openwrt-ai, line 518: nit: the placeholder disagrees with the reference text added in the same commit. `option timeout` maps to `qosify_map_timeout`, which the daemon resets to `3600` ([map.c](https://github.com/openwrt/qosify/blob/beeb87ec1f16aaf5d99ff8060c2455971d9d400e/map.c)), and line 551 of this file documents it as `defaults: timeout 3600`. `300` is `qosify_active_timeout`, a separate knob with no UCI option. The other placeholders in this panel (`500`, `100`, `5`) all match real defaults, so this one reads as a slip.

```suggestion
		this.qaNum(qadDef,'timeout','3600',60);
```

---

openwrt-ai (commented):
Reviewed the force-pushed head (`c13c738` → `cdd57a2`). The branch is still one commit and its message still matches the diff (whole new package), so no commit-message findings.

What changed since the last review: a substantial rework of `view/qosify/main.js` (+465/-…), the two bundled reset templates under `root/usr/share/qosify-luci/`, and the removal of `po/templates/qosify.pot`.

Verified against the qosify sources:
- `00-defaults.conf` and `qosify` are now byte-identical to [`qosify-defaults.conf`](https://github.com/openwrt/openwrt/blob/c9833b993d0157bed916203d1ae342b94608a982/package/network/config/qosify/files/qosify-defaults.conf) and [`qosify.conf`](https://github.com/openwrt/openwrt/blob/c9833b993d0157bed916203d1ae342b94608a982/package/network/config/qosify/files/qosify.conf) in openwrt.git (modulo a trailing newline) — good, the reset path now restores exactly what the package ships.
- `NQB` added to `DSCP`: correct, and the whole list matches the `dscp_names` table in qosify `map.c` including `EF`/`VA`/`NQB`/`LE`/`DF`.
- The corrected `+` help text ("override only when the DSCP field is zero") matches qosify's `QOSIFY_DSCP_FALLBACK_FLAG` handling; `dns_c:` is a real key prefix in `map.c`.
- The new defaults reference line (`timeout 3600`, `dscp_default_tcp/udp CS0`) matches `qosify_map_reset_config()`; `ifLint()`'s "overhead/overhead_encap only apply when overhead_type is manual" matches `add_interface()` in `qosify.init` (`overhead_mpu`/`overhead_vlan` are correctly excluded from that claim).
- CI is green on `cdd57a2` (build, eslint, CodeQL, FormalityCheck), so nothing CI-grounded to report.

Four issues and one nit inline — the dropped translation template and the anonymous-section handling in the new direct-to-file `saveQs()` path are the ones worth acting on before merge.

---

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1339 @@
...
+		nodes.push(E('div',{'class':'cbi-section-descr'},
+			_('Common shaping settings — written straight to %s, section %s.').format(UCI_PATH,sn?'config '+sn.type+(sn.name?" '"+sn.name+"'":' '+_('(unnamed section)')):"config interface 'wan' (will be created)")));
+		var tbl=E('table',{'class':'qos-kv','width':'100%'});
+		var bdy=E('tbody');tbl.appendChild(bdy);
+
+		function row(lbl,el){bdy.appendChild(E('tr',{},[E('td',{},lbl),E('td',{},el)]));}
+		function chk(name,val){return E('input',{'type':'checkbox','id':'q-'+name,'data-q':name,'checked':val?'checked':null});}
+		function txt(name,val,ph,style){return E('input',{'type':'text','id':'q-'+name,'data-q':name,'value':val||'','placeholder':ph||'','style':style||'width:140px;font-family:monospace'});}
+		function sel(name,val,opts,style,def,hint){
+			val=qv(val);
```
openwrt-ai, line 346: `qv()` is `String(v).replace(...)`, so `qv(undefined)` returns the literal string `"undefined"` — and both callers of `sel()` pass a uci option that is frequently absent:

- [`sel('overhead',w.overhead_type,OVH,'width:180px','none')`](``https://github.com/openwrt/luci/blob/3bd50f805b23c7428c980f21c0f37ac18dbd1784/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L360)``
- [`sel('mode',w.mode,MODES,'width:170px',null,'diffserv4')`](``https://github.com/openwrt/luci/blob/3bd50f805b23c7428c980f21c0f37ac18dbd1784/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L363)``

`w` comes straight from `uci.get('qosify',sn.id)`, where an unset option is `undefined`, not `''`. With `val="undefined"`, `sv` never matches an entry of `OVH`/`MODES`, so `known` stays false and line 350 appends `<option value="undefined" selected>undefined (current)</option>` — the dropdown shows `undefined (current)` instead of the real default. Worse, `saveQuick()` then reads that back (`ovh=get('overhead')`, `mode=get('mode')`), `kv.mode`/`kv.overhead_type` are the truthy string `"undefined"` rather than `null`, and `setOpts()` writes `option mode 'undefined'` into `/etc/config/qosify`.

This is reachable with the config the package ships: `config device wandev` in [`qosify.conf`](https://github.com/openwrt/openwrt/blob/c817d9a34f3f6a2504e197fa17312ea4448f01e6/package/network/config/qosify/files/qosify.conf) has neither `mode` nor `overhead_type`, so as soon as that section is the one `ifSect()` selects, both dropdowns are affected. Any hand-written section that omits the optional `mode`/`overhead_type` hits it too. Before this commit `val` was used raw and `val||def||''` handled `undefined` correctly.

```suggestion
			val=(val==null)?'':qv(val);
```

(Alternatively harden `qv()` itself with `if(v==null)return '';` — the other call sites already guard against null, so either works.)

---

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1339 @@
...
+	method:'revert',
+	params:['config']
+});
+function isRunning(r){
+	try{var i=r.qosify.instances;for(var k in i)if(i[k].running)return true;}catch(e){}
+	return false;
+}
+
+function clsLabel(c){return c.name+(c.alias?' (alias)':'');}
+function clsDesc(c){return 'Ingress: '+(c.ingress||'')+' / Egress: '+(c.egress||'');}
```
openwrt-ai: nit: these two helpers are the only user-visible strings in the view that are not wrapped in `_()`. `(alias)` is new in this commit, and `Ingress: ` / `Egress: ` are now centralised here from the five call sites (the class reference table, the class boxes in both the reference and the config panel). The regenerated `po/templates/qosify.pot` has 161 entries and contains neither, while the adjacent row labels [`_('Ingress')` and `_('Egress')`](``https://github.com/openwrt/luci/blob/3bd50f805b23c7428c980f21c0f37ac18dbd1784/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L365-L366)`` *are* translated — so a translated UI renders "Eingang"/"Ausgang" in Quick Settings and bare "Ingress:"/"Egress:" two panels down. That also conflicts with "All user-visible strings are translatable" in the PR description.

```suggestion
function clsLabel(c){return c.alias?_('%s (alias)').format(c.name):c.name;}
function clsDesc(c){return _('Ingress: %s / Egress: %s').format(c.ingress||'',c.egress||'');}
```

The translation template needs regenerating afterwards so the two new msgids appear.

---

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1339 @@
...
+			nat:chk('nat')?'1':'0',
+			host_isolate:chk('host_isolate')?'1':'0',
+			autorate_ingress:chk('autorate')?'1':'0',
+			ingress_options:iopts||null,
+			egress_options:eopts||null,
+			options:gopts||null,
+			option:null
+		};
+		kv.overhead=(ovh==='manual'&&ovhB)?ovhB:null;
+		if(sec&&!uci.get('qosify',sid||sec,'name'))kv.name=sec;
```
openwrt-ai: nit: the new `sec&&` guard is the right fix for the synthetic-`cfgXXXXXX` problem, but it leaves anonymous sections with no way to ever get a `name`. For an anonymous `config interface`, `ifSect()` now returns `name:''`, so `sec` is empty, this line is skipped and `kv` carries no `name` key — while [`ifLint()`](``https://github.com/openwrt/luci/blob/3bd50f805b23c7428c980f21c0f37ac18dbd1784/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L103)`` will keep reporting *"name is not set — qosify.init sends an empty device name and this section is never applied"* for exactly that section. Quick Settings has no `name` input either, so the user is told the section is broken and given no control that fixes it; the only route is the raw config editor.

Would it make sense to surface a `name` row in Quick Settings when the selected section is anonymous (or when `option name` is absent), so the warning is actionable from the same panel?

---

openwrt-ai (commented):
Reviewed the force-pushed head (`cdd57a2` → `3bd50f8`). Still a single commit whose message matches the diff (whole new package), so no commit-message findings. Changes since the last review: `po/templates/qosify.pot` restored (691 lines / 161 msgids) plus ~146 lines of `main.js`.

All four issues from the previous review are addressed:
- **Translation template** — regenerated against the current `main.js`; spot-checked that the new strings landed (`(unnamed section)`, `Port must be 1-65534 (qosify rejects 65535).`, the `ifLint()` warnings).
- **Anonymous sections in `saveQs()`** — `ifSect()` now returns `{type,id,name,idx,on}` with `name` blanked for `.anonymous` sections, and `setOpts()` falls back to a per-type ordinal. Verified the ordinal is sound: [`uci.sections()` sorts by `.index`](https://github.com/openwrt/luci/blob/3bd50f805b23c7428c980f21c0f37ac18dbd1784/modules/luci-base/htdocs/luci-static/resources/uci.js#L475-L477), i.e. file order, and `cfgSections()` counts blocks of the same type in the same order, so the two agree for mixed named/anonymous sections. `cfgSections()` also handles double-quoted headers now via `unq()`.
- **`qv()` newline injection** — now strips `'`, `"`, `\r`, `\n`; also applied to the `saveClsRow()` path that previously did its own `replace(/'/g,'')`.
- **`config alias` ignored by `getClasses()`** — aliases are enumerated alongside classes and `ingress`/`egress` fall back to `value`, matching `"${ingress:-$value}"` in `add_class()`.

Also checked the other new hunks: the `ruleWarn()` numeric branch computes its radix correctly (`&&` binds tighter than `?:`, so `0x…` → 16) and 0–63 is the right DSCP range; the sequential rewrite of `uploadFiles()` correctly reloads uci before `ruleWarn()` runs against the uploaded classes; the `dscp_*` select refresh only touches the two panels that were built from `dscpChoices` in the first place. CI is green on `3bd50f8` (build, eslint, CodeQL, FormalityCheck), so nothing CI-grounded to report.

One new issue and two nits inline — the `qv(undefined)` regression in `sel()` is the one worth acting on before merge.

---

choppyc79:
Thanks for the review — all three are fixed in the amended commit.

**1. `qv(undefined)` returning `"undefined"`**

Hardened `qv()` itself, as you suggested:

```js
function qv(v){return v==null?'':String(v).replace(/['"\r\n]/g,'');}
```

`sel()` now gets `''` for an unset option, so `val||def||''` resolves to the real default again, no `undefined (current)` entry is appended, and `saveQuick()` reads back `''` → `kv.mode`/`kv.overhead_type` are `null`, which removes the option instead of writing `option mode 'undefined'`. Verified against the shipped `config device wandev`, which has neither `mode` nor `overhead_type`. The other two `qv()` call sites already guard against empty values, so no change in behaviour there.

**2. Untranslated strings in `clsLabel()` / `clsDesc()`**

Both wrapped:

```js
function clsLabel(c){return c.name+(c.alias?' '+_('(alias)'):'');}
function clsDesc(c){return _('Ingress: %s / Egress: %s').format(c.ingress||'',c.egress||'');}
```

Used a single format string rather than two `Ingress: ` / `Egress: ` fragments so translators get the whole line in context, and all five call sites stay consistent with the adjacent `_('Ingress')` / `_('Egress')` row labels.

**3. Anonymous sections with no way to set `name`**

Added a **Device Name** row to Quick Settings, rendered whenever the selected section has no `option name` — anonymous sections and hand-written ones alike. It is prefilled with the section name (`wan` when the section is about to be created, empty for anonymous). `saveQuick()` reads it, validates `^[\w.@:-]+$`, and writes or clears `option name`; the previous `kv.name=sec` path is kept as the fallback for when the row is not rendered. The `ifLint()` warning is now actionable from the same panel instead of only via the raw config editor.

**i18n**

`po/templates/qosify.pot` regenerated with upstream `build/i18n-scan.pl` — 166 msgids, up from 161. New entries: `(alias)`, `Ingress: %s / Egress: %s`, `Device Name`, `required — qosify skips sections with no name`, `Error: name must be a device or interface name`. Validated with `msgfmt --check-format`.

Still a single commit, force-pushed with `--force-with-lease`.

applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js
```diff
@@ -0,0 +1,1349 @@
...
+		}
+
+		var enCb=chk('enabled',enChecked);
+		var enBadge=E('span',{'class':'qos-badge qos-amber','style':'margin-left:8px','id':'q-en-badge'},'');
+		this.updateEnBadge(enBadge,ctx,enChecked);
+		row(_('QoS Enabled'),[enCb,enBadge]);
+		// qosify.init passes `option name` to add_interface(); without it the daemon
+		// gets an empty device and the section is never applied, so offer it here
+		// whenever it is missing -- anonymous sections have no other way to set it.
+		if(!w.name)row(_('Device Name'),[txt('name',sn?sn.name:'wan','e.g. wan or eth0','width:140px'),
```
openwrt-ai: The prefill is the UCI *section* name, which is the right guess for `config interface` but the wrong one for `config device`.

`ifSect()` selects interface **and** device sections, and the package's own shipped config shows the two conventions differ: [`config interface wan` / `option name wan`](https://github.com/openwrt/openwrt/blob/d3191cb05603359991ecb8336662cc5408d9c2f9/package/network/config/qosify/files/qosify.conf#L27-L28) vs [`config device wandev` / `option name wan`](https://github.com/openwrt/openwrt/blob/d3191cb05603359991ecb8336662cc5408d9c2f9/package/network/config/qosify/files/qosify.conf#L44-L46) — the device section name is deliberately *not* the netdev name. So for a hand-written `config device wandev` with no `option name`, this row opens prefilled with `wandev`, and one click on Save & Apply writes `option name 'wandev'`. [`add_interface()` then does `json_add_object "$name"`](https://github.com/openwrt/openwrt/blob/d3191cb05603359991ecb8336662cc5408d9c2f9/package/network/config/qosify/files/qosify.init#L52-L53) for a netdev that does not exist, so nothing is shaped — and because `option name` is now set, `ifLint()`'s "name is not set" warning stops firing, removing the only signal the user had. Leaving the field empty for device sections keeps the warning until a real netdev name is entered.

```suggestion
		if(!w.name)row(_('Device Name'),[txt('name',sn?(sn.type==='interface'?sn.name:''):'wan','e.g. wan or eth0','width:140px'),
```

The pre-existing [`kv.name=sec` fallback at line 944](``https://github.com/openwrt/luci/blob/7648de60a440bb9079c46e63c7b4484412541105/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L944)`` makes the same guess, but it is now only reachable when `option name` is already set, so it no longer fires in practice.

nit: while here — the label reads `Device Name` for both section types, but [`refTable()` documents them as different things](``https://github.com/openwrt/luci/blob/7648de60a440bb9079c46e63c7b4484412541105/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L601-L602):`` *"name is the netifd interface"* for `interface`, *"name is a netdev"* for `device`. The panel already branches on `sn.type==='device'` for the NAT note two rows down, so a type-aware label (and placeholder — `eth0` is not a valid value for an `interface` section) would match the rest of the file.

---

openwrt-ai (commented):
Reviewed the force-pushed head (`3bd50f8` → `7648de6`). Still a single commit whose message matches the diff (whole new package), so no commit-message findings. Changes since the last review: 18 lines in `main.js` plus a regenerated `po/templates/qosify.pot`.

All three issues from the previous review are addressed:

- **`qv(undefined)`** — `qv()` now short-circuits on `null`/`undefined`, so `sel()`'s `val||def||''` resolves to the real default again, no `undefined (current)` option is appended, and `saveQuick()` no longer writes `option mode 'undefined'`. Checked the other four call sites (`setOpts()` ×3, `qacAdd()`): all of them are already guarded by `kv[k]!=null` / `if(!v)continue`, so the new early return changes nothing there.
- **`clsLabel()` / `clsDesc()`** — both wrapped in `_()`; these were the last unwrapped user-visible strings in the view. Both are rendered through `E()` text nodes, so no escaping concern with the `%s` substitution.
- **Anonymous / unnamed sections** — a `Device Name` row is now offered whenever `option name` is absent, and `saveQuick()` validates it before writing. One issue with the prefill inline.

Translation template: I diffed the `_()` string literals in `main.js` against the msgids in the regenerated `qosify.pot` — 165 on each side, exact match in both directions, so nothing is missing or stale. The five new msgids are `(alias)`, `Device Name`, `Error: name must be a device or interface name`, `Ingress: %s / Egress: %s` and `required — qosify skips sections with no name`.

CI is green on `7648de6` (build / Test x86_64, eslint, CodeQL, FormalityCheck), so nothing CI-grounded to report.

One issue inline.

---

choppyc79:
Fixed. The name field now only prefills for config interface sections, where the section name is the netifd interface by convention. config device sections get an empty field, so ifLint()'s "name is not set" warning keeps firing until a real netdev is entered rather than being silenced by a bad guess.

Also took the nit: the label and placeholder are now type-aware (Interface Name / e.g. wan vs Netdev Name / e.g. eth0), matching how refTable() describes the two, and the row reuses the existing isDev check that the NAT note already needed.

The kv.name = sec fallback in saveQuick() is gone — you're right that it was unreachable once the row renders whenever !w.name, and it encoded the same wrong guess. sid became unused and was dropped with it.

.pot regenerated with i18n-scan.pl; the only content delta is Device Name → Interface Name + Netdev Name.

openwrt-ai (commented):

Reviewed the force-pushed head (`7648de6` → `aa1ea64`). Still a single commit whose message matches the diff (whole new package), so no commit-message findings. Changes since the last review: 16 lines in `main.js` and the regenerated `po/templates/qosify.pot`.

The prefill issue from the previous review is addressed, and the fix is sound:

- `var isDev=!!(sn&&sn.type==='device')` now gates the prefill — `config device` sections get an empty field, so `option name` stays unset and [`ifLint()`'s "name is not set" warning](``https://github.com/openwrt/luci/blob/aa1ea64dc95afccf178b71f451dadfa0ab95854b/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L109)`` keeps firing until a real netdev is entered. Confirmed against `add_interface()` in openwrt.git, which does `config_get name "$cfg" name` with no fallback to the section name, so the option really is required for both section types.
- The `kv.name=sec` fallback is gone and `sid` with it. Verified it was genuinely unreachable: the row renders on `!w.name`, which is also true for `option name ''`, so there is no state where the section lacks a name and the input is absent. `buildQsSect()` is rebuilt from `refreshOverview()` (line 1316), so the row appears/disappears correctly after an external config edit too.
- `isDev` also replaces the old `natDev` for the NAT default; `!natDev` (with `natDev` possibly `null`) and `!isDev` are equivalent, so the checkbox default is unchanged.
- The type-aware label/placeholder matches [`refTable()`'s descriptions](``https://github.com/openwrt/luci/blob/aa1ea64dc95afccf178b71f451dadfa0ab95854b/applications/luci-app-qosify/htdocs/luci-static/resources/view/qosify/main.js#L608-L609)`` ("name is the netifd interface" vs "name is a netdev").

Translation template: diffed the `_()` literals in `main.js` against the msgids in the regenerated `qosify.pot` — 166 on each side, exact match in both directions. The only content delta is `Device Name` → `Interface Name` + `Netdev Name`; the rest of the churn is line-reference shifts consistent with the net +6 lines in `main.js`.

CI on `aa1ea64` shows only successes (FormalityCheck: Git & Commits, Code Patches, OpenWrt Makefiles); the build/eslint/CodeQL runs are still queued, so nothing CI-grounded to report.

---

hnyman:
I will merge this.
Nice to see that you got useful feedback from the AI, and reacted to it.

PR #8852, opened 2026-07-21
luci-mod-system: add option for force HTTP authentication

wevsty:
### Description
Inspired by
https://github.com/openwrt/luci/security/advisories/GHSA-8qcq-jgrj-gvmj

Currently, LuCI lacks restrictions on CGI programs bundled with plugins. It is absurd that if a LuCI plugin installs a CGI program, it can bypass the unified authentication with no way to mitigate this.

Adding this option to luci-mod-system allows users to at least use HTTP auth to prevent unauthorized access via LuCI.

While this can be accomplished via the command line, providing a WebUI option significantly improves user-friendliness and security hardening. This feature is not enabled by default and will not affect existing users.

## Tested on

**LuCI version:** LuCI openwrt-25.12 branch (26.183.16146~6aacb73)
**Web browser(s):** Firefox 152.0.6

---

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
@@ -14,6 +14,9 @@ return view.extend({
 		o = s.option(form.Flag, 'redirect_https', _('Redirect to HTTPS'), _('Enable automatic redirection of <abbr title="Hypertext Transfer Protocol">HTTP</abbr> requests to <abbr title="Hypertext Transfer Protocol Secure">HTTPS</abbr> port.'));
 		o.rmempty = false;
 
+		o = s.option(form.Flag, 'force_http_auth', _('Force HTTP Authentication'), _('Enforce <abbr title="Hypertext Transfer Protocol">HTTP</abbr> basic authentication for all pages using the root account.'));
+		o.rmempty = false;
+
 		return m.render();
 	}
 });


modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
```diff
@@ -1,6 +1,24 @@
 'use strict';
 'require view';
 'require form';
+'require uci';
+'require rpc';
+
+const AUTH_FILE_PATH = '/etc/force_http_auth.conf';
+const AUTH_FILE_CONTENTS = '/:root:$p$root\n'; 
+const AUTH_FILE_MODE = '0600';
```
openwrt-ai: The mode is passed as a string, but rpcd's `file.write` declares `mode` as an INT32 ([`file.c` L123](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L123)). A type-mismatched blobmsg attribute is silently dropped by `blobmsg_parse`, so `tb[RPC_F_RW_MODE]` ends up NULL and the file is created with the default `0666` (subject to umask, typically `0644`) rather than `0600` ([`file.c` L506, L526](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L506)). LuCI's own [`fs.js` passes `mode` as a number](https://github.com/openwrt/luci/blob/697a0e5c77c7d35e22d533192c10f0e3416a212f/modules/luci-base/htdocs/luci-static/resources/fs.js#L253) (`420 // 0644`) for this reason. Use a numeric literal so the intended permissions are actually applied:

```suggestion
const AUTH_FILE_MODE = 0o600;
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
```diff
@@ -1,6 +1,24 @@
 'use strict';
 'require view';
 'require form';
+'require uci';
+'require rpc';
+
+const AUTH_FILE_PATH = '/etc/force_http_auth.conf';
+const AUTH_FILE_CONTENTS = '/:root:$p$root\n';
```
openwrt-ai: nit: trailing whitespace after the semicolon.

```suggestion
const AUTH_FILE_CONTENTS = '/:root:$p$root\n';
```

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
```diff
@@ -14,6 +32,31 @@ return view.extend({
...
+			var promise;
+
+			if (formvalue === '1') {
+				promise = callFileWrite(AUTH_FILE_PATH, AUTH_FILE_CONTENTS, AUTH_FILE_MODE).then(function() {
+					uci.set('uhttpd', section_id, 'config', AUTH_FILE_PATH);
+					uci.set('uhttpd', section_id, 'force_http_auth', '1');
+				});
+			} else {
+				promise = callFileRemove(AUTH_FILE_PATH).then(function() {
+					uci.unset('uhttpd', section_id, 'config');
```
openwrt-ai: This commandeers uhttpd's shared `config` option: enabling force-auth overwrites `uhttpd.<section>.config` (line 45) and disabling it unsets the option entirely (line 50). But `config` is uhttpd's general httpd.conf path (`-c`), which a user may already have set for their own auth realms or directives. Enabling this feature would silently override that value, and later disabling it would delete it. Should the pre-existing `config` be preserved (or a dedicated mechanism used) instead of taking over the shared option?

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
```diff
@@ -14,6 +32,31 @@ return view.extend({
...
+					uci.unset('uhttpd', section_id, 'force_http_auth');
+				});
+			}
+
+			return promise.catch(function(e) {
+				console.error('Failed to configure HTTP auth: ', e);
+			});
```
openwrt-ai: The `.catch` swallows the failure by only logging to the console and resolving. If `callFileWrite`/`callFileRemove` fails (e.g. RPC/permission error), the corresponding `uci.set`/`uci.unset` in the `.then` never runs, yet the form save still reports success. For a security toggle this means the user can believe authentication is enabled when nothing was configured. Consider re-throwing (or surfacing a UI error) so the failure is visible rather than silent.

---

openwrt-ai (commented):
Reviewed 1 new commit. The mechanism (writing a `/:root:$p$root` realm file and pointing uhttpd's `config` at it) is sound — `$p$root` is a valid uhttpd auth form that resolves root's shadow hash. A few concerns are noted inline, the main one being that the `mode` argument is passed as a string and so is dropped by rpcd, leaving the auth file created world-readable/writable instead of `0600`.

Note: if the `root` account has no password set, uhttpd leaves the path unauthenticated ([`auth.c`](https://github.com/openwrt/uhttpd/blob/7b1bec45826bd78c8afc993435bdc0f1df2fe399/auth.c#L108)), so enabling this toggle silently has no effect in that case — may be worth surfacing to the user.

---

jow-:
Will LuCI itself accept the basic auth login then or would users be forced to login twice? If LuCI is going to inherit the basic auth credentials, we'd need to disable the logout functionality as it wouldn't work anymore.

wevsty:
@jow- 
> Will LuCI itself accept the basic auth login then or would users be forced to login twice?

LuCI automatically accepts the root identity established via HTTP authentication, so users do not need to log in twice.

> we'd need to disable the logout functionality as it wouldn't work anymore.

You are completely right. Once HTTP Basic Auth is enforced, browser-side credential caching prevents the standard LuCI logout session destruction from sticking (the browser immediately re-authenticates).

Disabling or conditionally hiding the Logout button would likely require changes at the luci-base or theme header level (e.g., checking uhttpd.main.force_http_auth before rendering the user menu). Let me know if you would prefer handling that in core or if we should keep this PR focused strictly on the uHTTPd configuration side.

---

Update: I have submitted a dedicated PR for openwrt/openwrt to handle this logic directly within `/etc/init.d/uhttpd` (PR link: https://github.com/openwrt/openwrt/pull/24347 ). 

Accordingly, I have simplified this LuCI PR to only handle the `force_http_auth` UCI option in the UI, keeping the LuCI side minimal while leaving the dynamic configuration and password validation to the uHTTPd init script. This also ensures consistent behavior whether configured via the Web UI or CLI.

modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
```diff
@@ -14,6 +14,9 @@ return view.extend({
 		o = s.option(form.Flag, 'redirect_https', _('Redirect to HTTPS'), _('Enable automatic redirection of <abbr title="Hypertext Transfer Protocol">HTTP</abbr> requests to <abbr title="Hypertext Transfer Protocol Secure">HTTPS</abbr> port.'));
 		o.rmempty = false;
 
+		o = s.option(form.Flag, 'force_http_auth', _('Force HTTP Authentication'), _('Enforce <abbr title="Hypertext Transfer Protocol">HTTP</abbr> basic authentication for all pages using the root account.'));
```
openwrt-ai, line 17: This adds a `force_http_auth` UCI option, but nothing in the current tree reads it. The uhttpd init script parses a fixed set of options (`config`, `realm`, `cgi_prefix`, …) with no handling for `force_http_auth` ([uhttpd.init append_arg block](https://github.com/openwrt/openwrt/blob/0f256a0a7adf5741e4a061f59a08cd01c14dc526/package/network/services/uhttpd/files/uhttpd.init#L137-L166)), and the name appears nowhere in uhttpd itself. As written, toggling this in the UI writes `uhttpd.main.force_http_auth` but has no effect on its own.

Per the earlier discussion the enforcement lives in a separate, not-yet-merged openwrt/openwrt change. That cross-repo dependency should be stated in the PR description (it is currently only in a comment thread), and there is a merge-order risk: if this LuCI PR lands first, the toggle silently does nothing. Which change consumes this option, and is it targeted for the same release?

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push that simplified the change). The prior file-write/rpcd approach was dropped; the PR now only adds a `force_http_auth` UCI flag. One inline note on the missing backend consumer / cross-repo dependency. The commit message matches the change.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/uhttpd.js
wevsty: As mentioned, this option relies on the companion uhttpd init script changes in openwrt/openwrt (PR: https://github.com/openwrt/openwrt/pull/24347). Until that core PR is merged, toggling force_http_auth in LuCI will indeed have no effect.

To mitigate the merge-order risk and avoid a broken UI state, we can hold off on merging this LuCI PR until the uhttpd changes are ready, or merge them closely together. Both are targeted for the same upcoming release cycle.

PR #8856, opened 2026-07-21
luci-base: bind rollback confirmation to the owning session

dibdot:
Minimal fix to emit the confirmation script only
when the requesting session is the one that initiated the rollback.

Reported-by: TA-MU-TA
Fixes: GHSA-76w2-p5qr-9vjv

### DIFF
--- a/modules/luci-base/ucode/template/footer.ut
+++ b/modules/luci-base/ucode/template/footer.ut
@@ -3,8 +3,9 @@
  Licensed to the public under the Apache License 2.0.
 -#}
 
-{% const rollback = dispatcher.rollback_pending() %}
-{% if (rollback || trigger_apply || trigger_revert): %}
+{% const rollback = dispatcher.rollback_pending();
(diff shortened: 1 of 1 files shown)


jow-:
Thanks for spending some effort on this @dibdot.

I do not recall the details well but afair the "loseness" was intentional. Some apply operations might disrupt device access, leading to a new login or connect from a different interface/IP and we basically wanted any interactive browser successfully hitting LuCI during the active pending apply window to confirm the apply (as in cancelling rollback).

Losing the old session, e.g. because the sysauth cookie was scoped to the old IP or hostname and we need to connect to a new IP or hostname, would entail a re-login leading to a new session ID which would be unable to cancel the pending rollback timer under your proposed change.

I guess we could record the effective username and compare that instead of the actual ID, but I remain unconvinced whether there is anything to fix in the first place.

dibdot:
Thanks for the feedback @jow!

PR #8876, opened 2026-07-26
luci-app-ddns: use default value 'wan' when interface value does not exist

AndyChiang888:
## Pull request details
* Revert "luci-app-ddns: add rmempty for interface" (commit b004197a277804ec0c8f092412b91c1d3e5936fa)
* use default value 'wan' when interface value does not exist
### Maintainer _(preferred)_

@<github-user>

---

## Tested on

**OpenWrt version:** Snapshot
**LuCI version:** master
**Web browser(s):** Chrome

---

## Checklist

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -887,12 +887,12 @@ return view.extend({
 					o.forcewrite = true;
 					o.modalonly = true;
 					o.cfgvalue = function(section_id) {
-						return uci.get('ddns', section_id, 'interface') || _('This will be autoset to the selected interface');
+						return uci.get('ddns', section_id, 'interface') || 'wan';
 					};
 					o.write = function(section_id) {
 						const opt = this.section.formvalue(section_id, 'ip_source');
 						const val = this.section.formvalue(section_id, 'ip_'+opt);
-						return uci.set('ddns', section_id, 'interface', val);
+						return uci.set('ddns', section_id, 'interface', val == 'wan' ? '' : val);
 					};
 
 					if (env['has_bindnet']) {
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -876,7 +876,6 @@ return view.extend({
 					o.modalonly = true;
 					o.multiple = false;
 					o.default = 'wan';
-					o.rmempty = false;
 					o.depends("ip_source", "web");
 					o.depends("ip_source", "script");
 					o.depends("ip_source", "interface");
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/status/include/70_ddns.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/status/include/70_ddns.js
@@ -38,7 +38,7 @@ return baseclass.extend({
 				services[key].next_update ? _(services[key].next_update) : _('Unknown'),
 				uci.get('ddns',key,'lookup_host'),
 				services[key].ip ? services[key].ip : _('No Data'),
-				(uci.get('ddns',key,'use_ipv6') == '1' ? 'IPv6' : 'IPv4') + ' / ' + uci.get('ddns',key,'interface')
+				(uci.get('ddns',key,'use_ipv6') == '1' ? 'IPv6' : 'IPv4') + ' / ' + (uci.get('ddns',key,'interface') || 'wan')
 			];
 		}), E('em', _('There is no service configured.')));
 


applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
```diff
@@ -888,12 +887,12 @@ return view.extend({
...
 					o.modalonly = true;
 					o.cfgvalue = function(section_id) {
-						return uci.get('ddns', section_id, 'interface') || _('This will be autoset to the selected interface');
+						return uci.get('ddns', section_id, 'interface') || 'wan';
 					};
 					o.write = function(section_id) {
 						const opt = this.section.formvalue(section_id, 'ip_source');
 						const val = this.section.formvalue(section_id, 'ip_'+opt);
-						return uci.set('ddns', section_id, 'interface', val);
+						return uci.set('ddns', section_id, 'interface', val == 'wan' ? '' : val);
```
openwrt-ai, line 895: Passing `''` only deletes the option for sections that already exist in uci. For a service section that is still pending creation — the normal case when a user adds a new DDNS service and saves the modal in one go — the section lives in `state.creates`, and [`uci.set()` takes the `if (n[conf]?.[sid])` branch at uci.js:610-615](https://github.com/openwrt/luci/blob/f7c7ab8570c7b45e1bcf9473a3bd0d0b6fd679b0/modules/luci-base/htdocs/luci-static/resources/uci.js#L610-L615), which only tests `val != null`. An empty string passes that test and is stored, [`save()` then forwards it in the `add` call's `values` at uci.js:934-944](https://github.com/openwrt/luci/blob/f7c7ab8570c7b45e1bcf9473a3bd0d0b6fd679b0/modules/luci-base/htdocs/luci-static/resources/uci.js#L934-L944), and rpcd writes it verbatim ([`rpc_uci_add()` value loop, uci.c:775-783](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/uci.c#L775-L783) — there is no empty-string filter). The result is a literal `option interface ''` in `/etc/config/ddns`, which is exactly the leftover this PR is trying to avoid.

`null` removes the option on both code paths.

```suggestion
						return uci.set('ddns', section_id, 'interface', val == 'wan' ? null : val);
```

---

applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
```diff
@@ -888,12 +887,12 @@ return view.extend({
 					o.forcewrite = true;
 					o.modalonly = true;
 					o.cfgvalue = function(section_id) {
-						return uci.get('ddns', section_id, 'interface') || _('This will be autoset to the selected interface');
+						return uci.get('ddns', section_id, 'interface') || 'wan';
```
openwrt-ai, line 890: This `DummyValue` is only active for `ip_source == 'network'`, and `write()` two lines below stores the value of `ip_network` — not `'wan'`. So for an existing section with e.g. `ip_network=lan` and no `interface` option, the modal now displays `wan`, but pressing Save writes `interface=lan`: the value shown contradicts the value about to be stored. The string this replaces (`This will be autoset to the selected interface`) is what communicated that.

Deriving the fallback from `ip_network` keeps the display consistent with what `write()` will store (and still yields `wan` when `ip_network` is unset, matching `o.default = 'wan'` on the `ip_network` option):

```suggestion
						return uci.get('ddns', section_id, 'interface') || uci.get('ddns', section_id, 'ip_network') || 'wan';
```

---

openwrt-ai (commented):
## Commit checks

- f7c7ab8 "luci-app-ddns: use default value 'wan' when interface value does not exist" — the subject and body only describe the read side, and state that "the config file no longer retains the default 'wan' interface value" as a pre-existing fact. It is this series that creates that state: 4e9dc03 restores the default `rmempty`, which makes [`AbstractValue.parse()` at form.js:2152](https://github.com/openwrt/luci/blob/f7c7ab8570c7b45e1bcf9473a3bd0d0b6fd679b0/modules/luci-base/htdocs/luci-static/resources/form.js#L2152-L2155) drop `interface` whenever it equals `o.default`, and this commit additionally rewrites an explicitly selected `wan` to an empty value in `_interface.write()`. That write-side normalisation is the actual behaviour change and is not mentioned anywhere in the message.

The backend premise checks out: `net/ddns-scripts/files/etc/hotplug.d/iface/ddns` uses `config_get interface $cfg interface wan`, and `config_get` expands via `${VAR:-default}`, so both an absent and an empty `interface` resolve to `wan`. Two inline notes below on the LuCI side.

---

applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
AndyChiang888: test results:
using `''` will not save `option interface ''` to `/etc/config/ddns`. Your comment is incorrect.

applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
AndyChiang888: incorrect comment, when `interface` does not exist, `ip_network` will not exist either, as both default to `wan`.

openwrt-ai (commented):

`2ce5613` is a message-only amendment — `git range-diff` against `f7c7ab8` shows the tree is unchanged. The added paragraph resolves the commit-check point from the previous round: the write-side normalisation in `_interface.write()` is now described, and the description is accurate — `_interface` only depends on `ip_source == 'network'`, so `formvalue('ip_' + opt)` there is always `ip_network`'s value.

## Retracting both of my previous inline comments

@AndyChiang888 rebutted both and both rebuttals hold. Recording the reasons here so the resolved threads stay resolved:

- **`uci.set(..., '')` would leave `option interface ''`** — wrong. A new DDNS service never reaches `_interface.write()` while it is still in `state.creates`: [`handleCreateDDnsRule()` at overview.js:468-475](``https://github.com/openwrt/luci/blob/2ce56135c533ef1399cb2b2dd4db253deb8bc515/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L468-L475)`` calls `uci.add()` *inside* `m.save()` and only opens the options modal in the `.then()`. [`Map.save()` at form.js:620-626](https://github.com/openwrt/luci/blob/2ce56135c533ef1399cb2b2dd4db253deb8bc515/modules/luci-base/htdocs/luci-static/resources/form.js#L620-L626) runs the callback, then `uci.save()`, then `this.load()`, so by the time any option `write()` runs the section is in `state.values`, not `state.creates`. `uci.set()` therefore always takes the existing-section path, where `val === ''` deletes the option. I analysed the `state.creates` branch, which this flow does not use.

- **`cfgvalue` falling back to `'wan'` while `write()` stores `ip_network`** — also wrong in practice. `ip_network` has `o.default = 'wan'` with the default `rmempty`, and after `4e9dc03` so does `interface`, so both are dropped from the config exactly when they equal `wan`; `interface` absent does imply `ip_network` absent. Configs not written by LuCI keep the same invariant — the ddns-scripts sample config sets [`interface` and `ip_network` to the same value in both services](https://github.com/openwrt/packages/blob/649dff397166ee773c334553c43b111337132548/net/ddns-scripts/files/etc/config/ddns#L18-L32) — and `_interface` is `forcewrite`, so any hand-written divergence is corrected on the next save anyway.

Backend premise re-verified against `openwrt/packages` master: [`etc/hotplug.d/iface/ddns`](https://github.com/openwrt/packages/blob/649dff397166ee773c334553c43b111337132548/net/ddns-scripts/files/etc/hotplug.d/iface/ddns#L12) uses `config_get interface $cfg interface wan`, whose `${VAR:-default}` expansion resolves both an absent and an empty `interface` to `wan`, and it is the only consumer of the `interface` option in that package.

---

AndyChiang888:
ping @systemcrash @hnyman

hnyman:
Similarly as AI, I do not quite understand the logic in line 895 
`return uci.set('ddns', section_id, 'interface', val == 'wan' ? '' : val);`
Why the value "wan" would not be saved?  What is the harm in storing it?

AndyChiang888:
Line 879 removes `o.rmempty = false;`, so when `wan` is selected, no value will be written, the same logic applies to line 895.
If line 895 is not modified, the `interface` value will always be written regardless of what it is.
Since the backend already handles the default value for `interface`, there is no need to save it when `interface` is `wan`.

hnyman:
I understand that it may be unnecessary due to automatic defaulting, but it may be confusing for the user.

It would seem more natural to me to have it in the config normally, even if the value is the default "wan". Handling that one value differently seems strange to me.

AndyChiang888:
commit b004197a277804ec0c8f092412b91c1d3e5936fa was originally a workaround introduced to fix the behavior of `o.default`.
If we simply revert b004197a277804ec0c8f092412b91c1d3e5936fa without also modifying line 895, the default value `wan` will still be written when `ip_source` is set to `network`.
This behavior is inconsistent with the case where `ip_source` is set to `web`, `script` or `interface`, which is why I made this change.
Since the default value is already handled on the backend, I don't think it is necessary to keep this value in config file. This can simplify the configuration file.

PR #8877, opened 2026-07-26
luci-app-banip: (feat) add a feed selector to the GeoIP map

renatoyamane:
The map plotted every blocked IP as one red layer. With several feeds active there was no way to  tell which Set an IP came from, or to look at one set in isolation.

Add a toolbar above the map holding a single select:

* One option per banIP Set present in ban_map.jsn, sorted by name, each showing how many map points that Set contributes. The default option reports the total and keeps the previous behaviour of plotting everything.
* Options carry the full Set name including the .v4 / .v6 suffix, so the two address families of a feed remain distinguishable in the list.
* Selecting a Set plots every mapped point of that Set, not a subset. No cap is applied, so the rendering does not depend on how f_report() orders or truncates set_elements.
* Blocked markers now live in an L.layerGroup, which makes redrawing on change a clearLayers() call instead of tracking and removing individual markers.
* The select is disabled in the markup and enabled only once at least one mappable point exists, which covers both an absent and an empty mapData without an extra branch.

Layout and theming of the new element:

* Replace the fixed 97vh map height with a flex column, so the toolbar and the map together fill the viewport exactly. This also drops the default 8px body margin that made 97vh overflow.
* Style the toolbar with the CSS system colors Canvas, CanvasText and ButtonBorder under color-scheme: light dark, instead of hardcoded values that would only suit a light theme. These follow the browser or OS preference rather than the active LuCI theme.

Everything else is untouched: coordinate deduplication, the entry validation, popup markup, the home marker code path and the key.slice(0, -3) title. In the unfiltered view the map renders exactly the same markers as before this change.

[Assisted by Anthropic Claude Opus 5.0]

### Description
Added a toolbar on the map to filter by feed

image

### Maintainer _(preferred)_
@dibdot 

---

## Tested on
[LuCI openwrt-25.12 branch (26.180.75667~128a781)]
[OpenWrt 25.12.5 (r33051-f5dae5ece4)]
Safari 26.5.2 (21624.2.5.11.8)

---

## Checklist

### DIFF
--- a/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/map.html
+++ b/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/map.html
@@ -9,8 +9,42 @@
 	<link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css"
 		integrity="sha256-p4NxAoJBhIIN+hmNHrzRCf9tD/miZyoHS5obTRR9BMY=" crossorigin="" />
 	<style>
+		html,
+		body {
+			height: 100%;
+			margin: 0;
+			color-scheme: light dark;
+		}
+
+		body {
+			display: flex;
+			flex-direction: column;
+		}
+
+		#toolbar {
+			flex: 0 0 auto;
+			display: flex;
+			flex-wrap: wrap;
+			align-items: center;
+			gap: 0.6em;
+			padding: 0.4em 0.8em;
+			box-sizing: border-box;
+			background: Canvas;
+			color: CanvasText;
+			border-bottom: 1px solid ButtonBorder;
+			font-family: sans-serif;
+			font-size: 14px;
+		}
+
+		#feedSelect {
+			min-width: 260px;
+			max-width: 70vw;
+			padding: 0.25em;
+		}
+
 		#map {
-			height: 97vh;
+			flex: 1 1 auto;
+			min-height: 0;
 			width: 100%;
 		}
 
@@ -24,6 +58,12 @@
 </head>
 
 <body>
+	<div id="toolbar">
+		<label for="feedSelect">Feed:</label>
+		<select id="feedSelect" disabled>
+			<option value="">All feeds</option>
+		</select>
+	</div>
 	<div id="map"></div>
 
 	<script src="https://unpkg.com/leaflet@1.9.4/dist/leaflet.js"
@@ -44,6 +84,7 @@
 
 		/* get mapData */
 		const mapData = sessionStorage.getItem('mapData');
+		const feedSelect = document.getElementById('feedSelect');
 		if (mapData) {
 			let uniqueCoordinates = [];
 			let homeCoordinates = [];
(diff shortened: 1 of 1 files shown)


applications/luci-app-banip/htdocs/luci-static/resources/view/banip/map.html
```diff
@@ -90,28 +131,54 @@
...
-				circle.bindPopup("<b><center>blocked IP by " + key + "</center></b>" +
-					"<p class=\"mono\">" +
-					"City: " + city + " (" + cc + ")" +
-					"<br />Latitude : " + lat +
-					"<br />Longitude: " + lon +
-					"<br />IP: " + ip +
-					"<br />AS: " + as +
-					"</p>");
+				feedCounts[coordObj.key] = (feedCounts[coordObj.key] || 0) + 1;
+			});
```
openwrt-ai, line 138: The per-Set counts and the per-Set filter are both lossy, because `uniqueCoordinates` is deduplicated on `lat`/`lon` alone, with no `key` in the predicate: [`map.html:102`](``https://github.com/openwrt/luci/blob/653b360a3a9e0ddf2726e8029a50b159e018b703/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/map.html#L102).`` The first Set to contribute a given coordinate owns it permanently; every later Set that maps an IP to the same coordinate contributes nothing.

That collision is the common case rather than a corner case. banIP emits one map entry per Set, tagged with the Set name, for each Set independently ([`banip-functions.sh:2297`](https://github.com/openwrt/packages/blob/b036203cbc27ef37c03ed1f8419d8d2b65e481f1/net/banip/files/banip-functions.sh#L2297)), and the geolocation backend returns city-level coordinates, so two feeds blocking IPs in the same city yield identical `lat`/`lon`.

Consequences for the new selector:

* `feedCounts[key]` counts coordinates a Set claimed first, not the mapped points it actually has, so the numbers in the option labels are misleading.
* Selecting Set B hides B's markers whose coordinates were already claimed by Set A — the opposite of the "look at one set in isolation" goal, and it contradicts the PR body's "Selecting a Set plots every mapped point of that Set, not a subset."
* Which Set wins is decided by `parsedData` iteration order, so the counts shift between reports for unchanged data.

Two ways out, both outside this hunk so no suggestion block: include `key` in the uniqueness predicate at line 102 so each Set keeps its own coordinate list (accepting overlapping markers in the "All" view, which is invisible at radius 2), or keep one entry per coordinate but store the set of contributing Set names on it, filter on membership, and list them all in the popup title.

---

applications/luci-app-banip/htdocs/luci-static/resources/view/banip/map.html
```diff
@@ -24,6 +58,12 @@
 </head>
 
 <body>
+	<div id="toolbar">
+		<label for="feedSelect">Feed:</label>
+		<select id="feedSelect" disabled>
+			<option value="">All feeds</option>
```
openwrt-ai, line 64: nit: the label says "Feed" but the option values are Set names carrying the `.v4`/`.v6` suffix, and the rest of luci-app-banip calls these Sets — the equivalent dropdown in the Set Reporting view is labelled [`-- Set Selection --`](``https://github.com/openwrt/luci/blob/653b360a3a9e0ddf2726e8029a50b159e018b703/applications/luci-app-banip/htdocs/luci-static/resources/view/banip/setreport.js#L99),`` and the details table header is `Set`. The commit message itself says "one option per banIP Set". banIP does distinguish the two: a feed is the source, a Set is the per-family nft set derived from it, which is what is listed here.

```suggestion
		<label for="feedSelect">Set:</label>
		<select id="feedSelect" disabled>
			<option value="">All Sets</option>
```

Line 145 needs the matching change (`"All Sets (" + ...`); the `feedSelect` / `feedCounts` identifiers are internal and can stay as they are.

---

applications/luci-app-banip/htdocs/luci-static/resources/view/banip/map.html
openwrt-ai, line 1: nit: `PKG_RELEASE` is not bumped. This package bumps it for every functional change — the two most recent ones both did it in the same commit as the code ([`4d06d15`](https://github.com/openwrt/luci/commit/4d06d15), 2 -> 3; [`7647aa9`](https://github.com/openwrt/luci/commit/7647aa9), 1 -> 2), and it currently sits at [`PKG_RELEASE:=3`](https://github.com/openwrt/luci/blob/653b360a3a9e0ddf2726e8029a50b159e018b703/applications/luci-app-banip/Makefile#L10). Without the bump the rebuilt package keeps the same version and existing installs will not pick up the new map.html on upgrade.

---

openwrt-ai (commented):
## Commit checks

- 653b360 "luci-app-banip: (feat) add a feed selector to the GeoIP map" — the message claims "Selecting a Set plots every mapped point of that Set, not a subset", but the diff filters `uniqueCoordinates`, which is deduplicated on `lat`/`lon` without regard to the Set key. A Set whose coordinates were already claimed by an earlier Set loses those points entirely, so a selection is a subset. Details inline. The message also states the change keeps the popup and home-marker paths untouched, which the diff does hold to.

---

dibdot:
Thanks for this, and for sticking with it through the rework — the second version is a clear step up from the first.

Anyway, I ended up with https://github.com/openwrt/luci/commit/859f54738e6ca6fd33c94df12b70c1e87877c868. 
I already had a broader set of map changes in progress: the dual stack viewport, the entry validation that silently discards IPs whose GeoIP lookup returns an empty city or AS, and rebuilding the popups as DOM nodes. Those touch the same code your selector sits on, so carrying both would have meant asking you to rebase onto a moving target. That is a scope question, not a quality one.

Two things from your PR went straight into mine. The system color theming with Canvas / CanvasText / ButtonBorder under color-scheme: light dark is simply the better solution. And the Set terminology from the review discussion is now used throughout. You are credited as co-author on the commit.

PR #8878, opened 2026-07-26
treewide: fix command injection, path traversal, XSS and CSRF issues

hauke:
### Description

Fixes for security issues reported to the OpenWrt security contact by Matthew
Hickey (Hacker Fantastic, https://hacker.house), together with two further
issues found while reviewing the same code paths.

One issue per commit:

| Commit | Package | Issue |
| --- | --- | --- |
| 1 | `luci-app-olsr` | Stored XSS. A mesh node announcing a crafted hostname through the olsrd nameservice plugin gets script execution in the LuCI origin when an admin opens Status → OLSR → Neighbours. No LuCI credentials needed to inject. |
| 2 | `luci-proto-openvpn` | Root command injection. `generateKey` interpolated the `keytype` argument unquoted into a `popen()` command string. |
| 3 | `luci-proto-openvpn` | Path traversal. `keyDir()` built key paths from an unvalidated `ifname`, so `makedirs()` created directories and wrote key files anywhere. `server_key` and `getSKeys` were affected in the same way. |
| 4 | `luci-app-ddns` | Root command injection. The `ddns_dateformat` UCI option was placed inside a double quoted `date` argument, so a `"` in the value started a new shell word. |
| 5 | `luci-app-ddns` | Path traversal. `get_services_log` composed the log path from an unvalidated `service_name`, while the sibling `logdir` was already checked for `../`. |
| 6 | `luci-app-ddns` | Stored XSS. `date` echoes non-specifier text verbatim, and the result is rendered as HTML by the overview page. Fixing the shell injection in commit 4 does not cover this. |
| 7 | `luci-app-commands` | Root command injection. The argument quoting allow-list contained the pipe character, so a bare `\|` argument reached the shell as an operator. |
| 8 | `luci-app-commands` | Missing CSRF protection. The run and download routes declared no `"post": true`, so they were reachable through a cross-origin GET. |
| 9 | `luci-app-commands` | Not a security fix: the download file name was derived from the already quoted `argv[0]`, so `echo hello` was offered as `.echo..txt` instead of `echo.txt`. |

Commits 1 to 5 and 7 correspond to the report and carry a `Reported-by:` tag.
Commits 6, 8 and 9 were found while working on the reported issues and carry no
such tag, since they were not reported.

Two further issues from the same report set, `luci-app-bmx7` path traversal and
`luci-app-dockerman` `ttyd_start` command injection, are already fixed in master
by 5890760a45 and ed11e49cf3 and are therefore not included here. Both still need
backporting to the release branches.

`luci-app-ddns` shares the `ddns_dateformat` option with `ddns-scripts`, which
builds shell code from the same value and is fixed separately in
openwrt/packages (see below).

### Verification

Every issue was reproduced and then re-tested against the fix on a running
OpenWrt image, by temporarily restoring the pre-fix code on the device:

- `luci-app-commands`: a bare `|` argument created a root owned file. With
  `public '1'` and `param '1'` this worked with no cookie and no token, so it was
  an unauthenticated root command execution rather than the authenticated issue
  described in the report.
- `luci-app-ddns`: `ddns_dateformat` executed `touch` as root, and the markup
  variant reached the status output. Both are rejected now, while custom formats
  such as `%d.%m.%Y %H:%M:%S` and the `%n` to `<br />` expansion still work.
- `luci-proto-openvpn`: `keytype`, `ifname` and `server_key` payloads are now
  rejected, and a legitimate `secret` key still generates correctly.
- `luci-app-olsr`: the crafted hostname survives intact through
  `ubus call olsrinfo hosts`, which is what the view fetches. Rendering the
  pre-fix and fixed markup in headless Chromium gives two `<svg>` elements with a
  firing `onload` before, and no elements at all after.

Note for the OLSR issue: the hostname is read from a whitespace separated hosts
file, so a payload containing spaces is truncated. The ``
form from the report does not survive that; `<svg/onload=...>` does.

- CSRF: run and download now answer `405` to GET and `403` to POST without a
  token, and `200` with one. The dashboard was adapted accordingly.

n/a

### Maintainer _(preferred)_

- `luci-app-olsr`: Manuel Munz <munz@comuno.net>
- `luci-app-ddns`: Ansuel Smith <ansuelsmth@gmail.com>
- `luci-app-commands`: Jo-Philipp Wich <jo@mein.io>
- `luci-proto-openvpn`: no `PKG_MAINTAINER`, added by Paul Donald in 5306e007656f

---

## Tested on

**OpenWrt version:** OpenWrt SNAPSHOT r35555-03e0a608e0 (armsr/armv8, QEMU)
**LuCI version:** master plus this branch (26.202.75735~ed23f07)
**Web browser(s):** Chromium 150.0.7871.186, used headless to verify the OLSR
rendering. The HTTP endpoints and ubus methods were exercised with curl and
`ubus call`.

---

## Checklist

- [x] _(Nice to have)_ Includes what it depends on (e.g. openwrt/packages#pr-number in sister repo).

### DIFF
--- a/applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
+++ b/applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
@@ -22,6 +22,12 @@ function shellquote(value) {
 	return "'" + replace(value, "'", "'\\''") + "'";
 }
 
+/* A DDNS service name is a UCI section name, so restrict it to the characters
+ * UCI permits. This keeps path separators and ".." out of composed log paths. */
+function is_valid_service_name(name) {
+	return length(name) > 0 && match(name, /^[a-zA-Z0-9_-]+$/) != null;
+}
+
 function get_dateformat() {
 	return uci.get('ddns', 'global', 'ddns_dateformat') || '%F %R';
 }
@@ -92,8 +98,9 @@ const methods = {
 				logdir = ddns_log_path;
 			}
 
-			// Check if service_name is provided and log file exists
-			if (request.args && request.args.service_name && stat(`${logdir}/${request.args.service_name}.log`)?.type == 'file' ) {
+			// Check if service_name is valid and the log file exists
+			if (is_valid_service_name(request.args?.service_name) &&
+			    stat(`${logdir}/${request.args.service_name}.log`)?.type == 'file' ) {
 				result = readfile(`${logdir}/${request.args.service_name}.log`);
 			}
 
--- a/applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
+++ b/applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
@@ -43,7 +43,7 @@ function trimnonewline(input) {
 }
 
 function get_date(seconds, format) {
-	return trimnonewline( popen(`date -d @${seconds} "+${format}" 2>/dev/null`, 'r')?.read?.('line') );
+	return trimnonewline( popen(`date -d @${int(seconds)} ${shellquote('+' + format)} 2>/dev/null`, 'r')?.read?.('line') );
 }
 
 // convert epoch date to given format
--- a/applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
+++ b/applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
@@ -57,6 +57,16 @@ function epoch2date(epoch, format) {
 	if (!format || length(format) < 2) {
 		format = get_dateformat();
 	}
+
+	/* date(1) echoes anything that is not a conversion specifier verbatim, and
+	 * the status views render the result as HTML - see the '%n' expansion below.
+	 * Restrict the format to the characters strftime actually needs so that a
+	 * configured value cannot inject markup, and fall back to the default
+	 * otherwise, mirroring how an unsafe ddns_logdir is handled. */
+	if (match(format, /[^%A-Za-z0-9 :\/.,_+-]/)) {
+		format = '%F %R';
+	}
+
 	format = replace(format, /%n/g, '<br />'); // Replace '%n' with '<br />'
 	format = replace(format, /%t/g, '    ');   // Replace '%t' with four spaces
 
--- a/protocols/luci-proto-openvpn/root/usr/share/rpcd/ucode/luci.openvpn.uc
+++ b/protocols/luci-proto-openvpn/root/usr/share/rpcd/ucode/luci.openvpn.uc
@@ -10,7 +10,8 @@ import { connect } from 'ubus';
 const openvpn_dir = '/etc/openvpn';
 
 /* Key types accepted by "openvpn --genkey". Anything outside this set is
- * rejected, so the value can never introduce shell metacharacters. */
+ * rejected, so the value can never introduce shell metacharacters or escape
+ * the key directory. */
 const keytypes = [
 	'secret',
 	'tls-crypt',
@@ -22,6 +23,17 @@ const keytypes = [
 
 function shellquote(s) {
 	return `'${replace(s, "'", "'\\''")}'`;
+}
+
+/* ifname is a UCI section name, so restrict it to the characters UCI allows.
+ * This keeps path separators and ".." out of the generated key paths. */
+function is_valid_iface(ifname) {
+	return length(ifname) > 0 && match(ifname, /^[a-zA-Z0-9_-]+$/) != null;
+}
+
+/* server_key is a plain file name inside the tls-crypt-v2-server directory. */
+function is_valid_keyfile(name) {
+	return length(name) > 0 && match(name, /^[a-zA-Z0-9._-]+$/) != null && index(name, '..') == -1;
 }
 
 function command(cmd) {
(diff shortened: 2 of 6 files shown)


applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
```diff
@@ -43,14 +49,24 @@ function trimnonewline(input) {
...
 	if (!format || length(format) < 2) {
 		format = get_dateformat();
 	}
+
+	/* date(1) echoes anything that is not a conversion specifier verbatim, and
+	 * the status views render the result as HTML - see the '%n' expansion below.
+	 * Restrict the format to the characters strftime actually needs so that a
+	 * configured value cannot inject markup, and fall back to the default
+	 * otherwise, mirroring how an unsafe ddns_logdir is handled. */
+	if (match(format, /[^%A-Za-z0-9 :\/.,_+-]/)) {
```
openwrt-ai, line 66: nit: the allow-list is narrower than the sink it protects, so it silently discards formats that are not dangerous. After the previous commit the format no longer reaches the shell (it is `shellquote()`d in `get_date()`), so the only characters that matter here are the ones that mean something in the `innerHTML` sink — `<`, `>`, `&`, `"` and `'`. Punctuation such as `(`, `)`, `[`, `]`, `#`, `!`, `?`, `=` and `*` is inert in that context but is rejected all the same, so an existing config with e.g. `ddns_dateformat='%F (%T)'` silently reverts to `%F %R` with no indication in the UI ([the value is echoed back as "Current setting"](``https://github.com/openwrt/luci/blob/3c7082f7cffd12252f59140ad70e7775b0ad6ca3/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L402-L404),`` which will then disagree with what the user typed into the field right above it).

Was the narrow set deliberate, or would rejecting only the five markup characters be preferable here?

---

applications/luci-app-ddns/root/usr/share/rpcd/ucode/ddns.uc
```diff
@@ -43,14 +49,24 @@ function trimnonewline(input) {
 }
 
 function get_date(seconds, format) {
-	return trimnonewline( popen(`date -d @${seconds} "+${format}" 2>/dev/null`, 'r')?.read?.('line') );
+	return trimnonewline( popen(`date -d @${int(seconds)} ${shellquote('+' + format)} 2>/dev/null`, 'r')?.read?.('line') );
```
openwrt-ai, line 52: nit: this commit message correctly notes that ddns-scripts consumes the same UCI option through `eval` ([`DATE_PROG="date +'$ddns_dateformat'"` at dynamic_dns_functions.sh:123-124](``https://github.com/openwrt/packages/blob/649dff397166ee773c334553c43b111337132548/net/ddns-scripts/files/usr/lib/ddns/dynamic_dns_functions.sh#L123-L124)),`` and the PR description says it "is fixed separately in openwrt/packages (see below)" with the "depends on … sister repo" checkbox ticked — but no link appears anywhere in the body. The sister PR looks to be openwrt/packages#30102; worth adding the reference so the two land together.

---

protocols/luci-proto-openvpn/root/usr/share/rpcd/ucode/luci.openvpn.uc
```diff
@@ -93,7 +119,11 @@ const methods = {
 	getSKeys: {
 		args: { ifname: 'ifname' },
 		call: function(req) {
-			const serverDir = `${keyDir(req.args?.ifname, 'tls-crypt-v2-server')}`;
+			const ifname = req.args?.ifname;
+
+			if (!is_valid_iface(ifname)) return { error: 'invalid ifname' };
```
openwrt-ai, line 124: nit: `getSKeys` now hard-rejects a missing `ifname`, while `generateKey` still substitutes a default for it ([`req.args?.ifname || 'unnamed'` on line 62](``https://github.com/openwrt/luci/blob/3c7082f7cffd12252f59140ad70e7775b0ad6ca3/protocols/luci-proto-openvpn/root/usr/share/rpcd/ucode/luci.openvpn.uc#L62``)) before validating. A caller that omits `ifname` can therefore still generate a server key into `/etc/openvpn/unnamed/tls-crypt-v2-server/`, but can never list it back.

The pre-existing behaviour was equally inconsistent (`keyDir(null, …)` resolved to `/etc/openvpn//tls-crypt-v2-server`, not the `unnamed` directory), so this is not a regression — but since the two methods are being hardened in the same commit, should `getSKeys` apply the same `|| 'unnamed'` fallback, or should `generateKey` drop its default and require a valid `ifname` too?

---

openwrt-ai (commented):
Reviewed 9 new commits. Each commit message matches its diff and the scoping is clean (one issue per commit, no mixed hunks). No correctness problems found in the fixes themselves; three nits inline.

Things I checked and found sound, for the record:

- The `\w`-in-bracket-expression analysis in the `luci-app-commands` commits is correct — ucode's lexer passes a macro set of just `"^"` to `parse_escape()` inside bracket expressions, so `[^\w.\/-]` really does compile to `[^w./-]`.
- `shellquote()` in all three files is safe: ucode's `replace()` replaces *all* occurrences when the pattern is a plain string (unlike JS), so multi-quote arguments do not escape the quoting.
- `kt in keytypes` works as array membership in ucode (`uc_vm_insn_in()`), and the list matches the values offered by the `_keytype` widget exactly.
- Passing the CSRF token as a query parameter on a POST is the same shape luci-base itself uses in `LuCI.poll(..., post=true)`, and `http.message.params` is seeded from `QUERY_STRING` before the body is parsed, so `test_post_security()` sees it.
- The OLSR `%q`/`%h` change matches the existing `routes.js`/`hna.js` pattern, and `neigh.hn` has no other raw sink in that view.

CI on the head commit reports no failed checks.

---

hnyman:
@hauke @jow- 
This needs rebase, and then merge?

BKPepe:
@hauke, @jow- ping :)

PR #8879, opened 2026-07-27
luci-app-commands: add optional persistent id for public URLs

### LINKED ISSUE #6703: luci-app-commands: the URL for public use should be a hash value
I would like the URL for public use to be a hash value, etc.

Thanks for providing useful tools!
I have registered several commands, such as displaying a syslog, and have opened each of them to general users.
During router maintenance, I deleted one of the commands that was no longer needed, and the URLs for the other commands were changed.
Can you please change the specification so that once the URL is published, it does not change?

## Steps to reproduce:
1 Add three commands, like echo ‘1’, echo ‘2’ and echo ‘3’
2 Copy three URLs
3 Remove second command. Now remains are echo ‘1’ and ‘3’
4 Open URL for ‘2’ 

## Actual behavior:
Show '3'

## Expected behavior:
Show '404 page not found' error.

cawcaw2023:
### Description

Command sections are stored as anonymous UCI sections, so each command's internal identifier is automatically (re-)assigned and is not guaranteed to stay the same when other commands are removed. Since this identifier is also used to build the "Run", "Download" and public "Link" URLs, previously bookmarked or published URLs can silently start pointing at a different command.

Add an optional 'id' option that lets the user assign a fixed, URL-safe identifier to a command. When set, this identifier is preferred over the internal section name when resolving a command from a URL, so the URL keeps working regardless of what happens to other commands. Existing commands and previously generated URLs keep working unchanged, since the new option is optional and falls back to the current behavior when unset.

Fixes: https://github.com/openwrt/luci/issues/6703

---

## Tested on
**OpenWrt version:** 25.12.5 r33051-f5dae5ece4
**LuCI version:** openwrt-25.12 branch 26.187.49110~99464ec
**Web browser(s):** Firefox 152.0.6 (aarch64)

I asked Claude-san to create a patch for the issue, and once the patch was ready,
I applied it to the three `commands` files in the `master` branch, then overrode
the files on the live 25.12.5 production server and verified that it worked.
---

### DIFF
--- a/applications/luci-app-commands/htdocs/luci-static/resources/view/commands.js
+++ b/applications/luci-app-commands/htdocs/luci-static/resources/view/commands.js
@@ -2,6 +2,7 @@
 
 'require view';
 'require form';
+'require uci';
 
 return view.extend({
 	render: function(data) {
@@ -22,6 +23,24 @@ return view.extend({
 		o = s.option(form.Value, 'description', _('Description'),
 			_('An optional longer description to display on the execution page'));
 		o.optional = true;
+
+		o = s.option(form.Value, 'id', _('Persistent ID'),
+			_('An optional, fixed identifier to use in the public command URL. Once set, the URL keeps working even if other commands are added, removed or reordered. Leave empty to keep relying on the automatically generated, unstable identifier. The identifier must not end with the letter "s".'));
+		o.optional = true;
+		o.datatype = 'uciname';
+		o.validate = function(section_id, value) {
+			if (value == null || value == '')
+				return true;
+
+			if (/s$/.test(value))
+				return _('The identifier must not end with the letter "s"');
+
+			let dup = uci.sections('luci', 'command').some(function(cmd) {
+				return cmd['.name'] != section_id && (cmd.id == value || cmd['.name'] == value);
+			});
+
+			return dup ? _('This identifier is already used by another command') : true;
+		};
 
 		o = s.option(form.Value, 'command', _('Command'), _('Command line to execute'));
 		o.textvalue = function(section_id) {
--- a/applications/luci-app-commands/ucode/controller/commands.uc
+++ b/applications/luci-app-commands/ucode/controller/commands.uc
@@ -117,20 +117,53 @@ function test_binary(str) {
 	return false;
 }
 
+// Resolve a command identifier as seen in a URL to the actual UCI
+// section name. If a "command" section carries a stable, user assigned
+// "id" option matching the requested identifier, that section is used.
+// Otherwise the identifier is tried verbatim, for backward compatibility
+// with links generated before the "id" option existed (it will match
+// the automatically assigned, but potentially unstable, section name).
+// An "id" ending in "s" is never honored here: the public URL handler
+// (action_public) uses a trailing "s" to distinguish its display and
+// download modes, so allowing a custom id to end in "s" would make
+// that encoding ambiguous.
+function resolve_command_id(cmdid) {
+	if (!cmdid)
+		return null;
+
+	let found = null;
+
+	uci.foreach('luci', 'command', s => {
+		if (found == null && s.id && substr(s.id, -1) != 's' && s.id == cmdid)
+			found = s['.name'];
+	});
+
+	if (found != null)
+		return found;
+
+	if (uci.get('luci', cmdid) == 'command')
+		return cmdid;
+
+	return null;
+}
+
 function parse_cmdline(cmdid, args) {
-	if (uci.get('luci', cmdid) == 'command') {
-		let cmd = uci.get_all('luci', cmdid);
-		let argv = parse_args(cmd?.command);
+	cmdid = resolve_command_id(cmdid);
 
-		if (cmd?.param == '1') {
-			if (length(args))
-				push(argv, ...(parse_args(urldecode(args)) ?? []));
-			else if (length(args = http.formvalue('args')))
-				push(argv, ...(parse_args(args) ?? []));
-		}
+	if (cmdid == null)
+		return;
 
-		return map(argv, v => match(v, /[^\w.\/|-]/) ? `'${replace(v, "'", "'\\''")}'` : v);
+	let cmd = uci.get_all('luci', cmdid);
+	let argv = parse_args(cmd?.command);
+
+	if (cmd?.param == '1') {
+		if (length(args))
+			push(argv, ...(parse_args(urldecode(args)) ?? []));
+		else if (length(args = http.formvalue('args')))
+			push(argv, ...(parse_args(args) ?? []));
 	}
+
+	return map(argv, v => match(v, /[^\w.\/|-]/) ? `'${replace(v, "'", "'\\''")}'` : v);
 }
 
 function execute_command(callback, ...args) {
@@ -240,9 +273,9 @@ return {
 			cmdid = substr(cmdid, 0, -1);
 		}
 
-		if (cmdid &&
-		    uci.get('luci', cmdid) == 'command' &&
-		    uci.get('luci', cmdid, 'public') == '1')
+		cmdid = resolve_command_id(cmdid);
+
+		if (cmdid != null && uci.get('luci', cmdid, 'public') == '1')
 		{
 			if (disp)
 				execute_command(return_html, cmdid, ...args);
--- a/applications/luci-app-commands/ucode/template/commands.ut
+++ b/applications/luci-app-commands/ucode/template/commands.ut
@@ -138,6 +138,9 @@
 		{% else %}
 			<div class="commands">
 				{% for (let command in commands): %}
+					{%
+						const cmdid = (match(command.id, /^[A-Za-z0-9_]+$/) && !match(command.id, /s$/)) ? command.id : command['.name'];
+					-%}
 				<div class="commandbox">
 					<h3>{{ entityencode(command.name) }}</h3>
 					<p>{{ _('Command:') }} <code>{{ entityencode(command.command) }}</code></p>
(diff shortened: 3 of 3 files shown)


cawcaw2023:
## Tested on
I tested on master too.
OpenWrt version: SNAPSHOT, r35551-ec8f8db38e
LuCI version: luci-light (26.207.70497~fb6b224)
luci-app-commands (26.207.70497~fb6b224)
Web browser(s): Firefox 152.0.6 (aarch64)

applications/luci-app-commands/ucode/template/commands.ut
```diff
@@ -138,20 +138,23 @@
 		{% else %}
 			<div class="commands">
 				{% for (let command in commands): %}
+					{%
+						const cmdid = command.id || command['.name'];
```
openwrt-ai: `cmdid` is interpolated raw into an HTML attribute (`id="{{ cmdid }}"`, line 151) and into JS string literals inside `onclick` (lines 154, 155, 157), while every other config-supplied value in this template goes through `entityencode()` (`command.name`, `command.command`, `command.description`).

That was safe up to now because `command['.name']` is always an auto-generated `cfgXXXXXX` name. `command.id` is free-form UCI text: the `o.datatype = 'uciname'` in [commands.js:30](``https://github.com/openwrt/luci/blob/46a4462387cf3f294f5c6097ba02291723b2a4e5/applications/luci-app-commands/htdocs/luci-static/resources/view/commands.js#L30)`` only guards the LuCI form, so a value set with `uci set luci.@command[0].id=...` or by editing `/etc/config/luci` reaches this template unchanged, and a `"` or `'` breaks out of the attribute / the JS string.

Note `entityencode()` alone would not fix the `onclick` case — `&#39;` is HTML-decoded back to `'` before the JS parser sees the attribute value. Enforcing the same charset the form's `uciname` datatype enforces, with a fallback to the section name, covers both contexts in one place:

```suggestion
						const cmdid = match(command.id, /^[A-Za-z0-9_]+$/) ? command.id : command['.name'];
```


---

applications/luci-app-commands/ucode/controller/commands.uc
```diff
@@ -240,9 +269,9 @@ return {
 			cmdid = substr(cmdid, 0, -1);
 		}
 
-		if (cmdid &&
-		    uci.get('luci', cmdid) == 'command' &&
-		    uci.get('luci', cmdid, 'public') == '1')
+		cmdid = resolve_command_id(cmdid);
```
openwrt-ai, line 276: A persistent id ending in `s` is unreachable in download mode. A few lines above, `action_public()` strips a trailing `s` from the URL segment to switch between download and display mode — that was unambiguous while ids were auto-generated `cfg` + hex names, which can never end in `s`, but a user-chosen id can.

With `id = 'logs'`, the "Download execution result" link that the Link button builds (`/command/logs`) is stripped to `log`, `resolve_command_id()` returns null, and the user gets 403. And with two commands `foo` and `foos`, `/command/foos` — the download URL of the second — is served as the display page of the first.

Resolving the unstripped identifier first and only falling back to the stripped form removes the ambiguity:

```suggestion
		if (disp && resolve_command_id(cmdid) == null) {
			disp = false;
			cmdid += 's';
		}

		cmdid = resolve_command_id(cmdid);
```

Alternatively, reject a trailing `s` in the form validator, which is simpler but constrains the user.


---

applications/luci-app-commands/htdocs/luci-static/resources/view/commands.js
```diff
@@ -23,6 +24,21 @@ return view.extend({
...
+			_('An optional, fixed identifier to use in the public command URL. Once set, the URL keeps working even if other commands are added, removed or reordered. Leave empty to keep relying on the automatically generated, unstable identifier.'));
+		o.optional = true;
+		o.datatype = 'uciname';
+		o.validate = function(section_id, value) {
+			if (value == null || value == '')
+				return true;
+
+			var dup = uci.sections('luci', 'command').some(function(s) {
+				return s['.name'] != section_id && s.id == value;
+			});
```
openwrt-ai: The duplicate check only compares against other commands' `id` values, not against their section names. Since [`resolve_command_id()`](https://github.com/openwrt/luci/blob/46a4462387cf3f294f5c6097ba02291723b2a4e5/applications/luci-app-commands/ucode/controller/commands.uc#L132-L141) prefers an `id` match over the verbatim section-name match, setting `id` to another command's auto-generated section name (`cfg024e8d` passes `uciname`) silently hijacks that command's existing bookmarked URL — exactly the breakage this PR sets out to prevent — and this validator lets it through.

Also `var` is unused elsewhere in this file (`let m, s, o;` above), and the callback parameter `s` shadows the outer `GridSection` `s`.

```suggestion
			let dup = uci.sections('luci', 'command').some(function(cmd) {
				return cmd['.name'] != section_id && (cmd.id == value || cmd['.name'] == value);
			});
```


---

openwrt-ai (commented):
Reviewed 1 new commit; the commit message matches the diff. Three inline findings, all around the fact that the command identifier used in URLs and in the dashboard markup is now free-form UCI text rather than an auto-generated `cfgXXXXXX` name.

---

applications/luci-app-commands/ucode/template/commands.ut
cawcaw2023: Fixed, thanks for catching this.

applications/luci-app-commands/ucode/controller/commands.uc
cawcaw2023: Fixed, thanks for catching this.

applications/luci-app-commands/htdocs/luci-static/resources/view/commands.js
cawcaw2023: Fixed, thanks for catching this.

applications/luci-app-commands/ucode/template/commands.ut
```diff
@@ -138,20 +138,23 @@
...
+					-%}
 				<div class="commandbox">
 					<h3>{{ entityencode(command.name) }}</h3>
 					<p>{{ _('Command:') }} <code>{{ entityencode(command.command) }}</code></p>
 					{% if (command.description): %}
 						<p><em>{{ entityencode(command.description) }}</em></p>
 					{% endif %}
 					{% if (command.param == "1"): %}
-						<p>{{ _('Arguments:') }} <input type="text" id="{{ command['.name'] }}" /></p>
+						<p>{{ _('Arguments:') }} <input type="text" id="{{ cmdid }}" /></p>
```
openwrt-ai, line 151: The DOM id of the arguments input is now a user-chosen string, so it can collide with an unrelated element on the page. [`command_run()`](https://github.com/openwrt/luci/blob/1f82545880b1dbecd9c8434af119d3804f0cdbd3/applications/luci-app-commands/ucode/template/commands.ut#L34), [`command_download()`](https://github.com/openwrt/luci/blob/1f82545880b1dbecd9c8434af119d3804f0cdbd3/applications/luci-app-commands/ucode/template/commands.ut#L75) and [`command_link()`](https://github.com/openwrt/luci/blob/1f82545880b1dbecd9c8434af119d3804f0cdbd3/applications/luci-app-commands/ucode/template/commands.ut#L90) all do `document.getElementById(id)` with the same string they use to build the URL. The theme layout defines [`topmenu`, `indicators` and `maincontent`](``https://github.com/openwrt/luci/blob/1f82545880b1dbecd9c8434af119d3804f0cdbd3/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut#L52-L56),`` plus `modemenu`, `tabmenu` and `view` — every one of those passes the `uciname` datatype and the new trailing-`s` rule.

With `option id 'maincontent'` on a `param = 1` command, `getElementById()` returns that `<div>` instead of the input, `field.value` is `undefined`, and Run silently executes without the arguments the user typed (Download builds `.../download/maincontent/undefined`). No error is shown. This was collision-free before the change because `command['.name']` is always `cfgXXXXXX`.

The URL identifier and the DOM id don't need to be the same string — keeping `command['.name']` for the element id and passing `cmdid` as a separate parameter used only for URL building decouples them. Prose only, since the change spans the three helpers above as well as this line.

---

openwrt-ai (commented):
Re-reviewed the amended commit (`46a4462` → `1f82545`). All three findings from the previous round are addressed: the template now charset-guards `cmdid` before interpolating it into the attribute and the `onclick` JS strings, the trailing-`s` ambiguity is resolved via the validator route, and the duplicate check now also compares against other sections' names. One new finding inline, on a side effect of the identifier now doubling as a DOM element id.

## Commit checks
- `1f82545` "luci-app-commands: add optional persistent id for public URLs" — the message describes the id as a "fixed, URL-safe identifier" but says nothing about the new hard restriction the diff enforces: an identifier may not end with `s` (form validator, plus `resolve_command_id()` silently ignoring such ids). That rules out common names like `logs`, `stats` or `ps`, so it is worth a sentence in the commit body and the PR description, next to the rationale already in the `resolve_command_id()` comment.

---

cawcaw2023:
Closing this PR.

While going through review, I started reconsidering the core design.
Making the identifier a free-form, user-editable field is the root
cause behind every issue raised so far (charset validation, the
trailing-"s" ambiguity, the duplicate-detection gap, and the DOM id
collision) — they all stem from the same source. I think a better
approach is to auto-generate a safe, unique identifier when a command
is created, instead of asking the user to type one, which avoids this
whole class of problem rather than patching it case by case.

I'd rather step back and redesign along those lines than keep
iterating here, so I'm closing this for now. I may open a follow-up
issue/PR once that's worked out.

Thanks for the thorough review each round, and sorry for the churn
this caused in the meantime.

PR #8880, opened 2026-07-27
luci-theme-bootstrap: fix hidden down arrow on submenus for accessibility

bpisch:
This PR is an improvement for LuCI screen reader accessibility for blind and visually impaired people. It should not affect the visual look of the page.
The down arrow displayed next to submenus is a CSS-drawn triangle, but in its content: field it had a hidden down arrow encoded as &darr; Since these HTML-encoded symbols aren't converted into real characters in CSS fields, screen readers read this out loud even if it is visually hidden. Therefore next to all submenus, screen readers announced: and darr semicolon. This change fixes the symbol encoding so that screen readers correctly announce "down arrow" when navigating onto a submenu. The visually drawn triangle is completely ignored by the screen reader, that is why some indication is needed. Later this could be improved with proper ARIA attributes, but this was the quickest fix for now to improve LuCI usability until I manage to test it further.
Change is tested on OpenWrt 25.12.5 but as this is a CSS modification it should be universal anywhere LuCI is used.

### DIFF
--- a/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
+++ b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css
@@ -1147,7 +1147,7 @@ a.menu:after {
 	width: 0;
 	height: 0;
 	display: inline-block;
-	content: "&darr;";
+	content: "\2193";
 	text-indent: -99999px;
 	vertical-align: top;
 	margin-top: 8px;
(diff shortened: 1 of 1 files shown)


openwrt-ai (commented):

---

Rondom (approved):

efahl:
Looks good to me, but...

The magic number replacing the mnemonic "darr" is going to be a target for someone to change back without looking at the commit history.  Maybe it should have a comment in the code itself that says something like 

`// Screen reader accessible "down arrow"`

to make people think before changing it. 

Question for @hauke:  Should we have some more prompts in LuCI's openwrt-ai that check for accessibility issues whenever css is modified?

bpisch:
Hello @efahl ,
Sorry for the late response. I have added the comment you requested next to the character code.

openwrt-ai (commented):

---

efahl (approved):
Looks good to me!

PR #8881, opened 2026-07-27
luci-app-strongswan-swanctl: rename and add missing options

feckert:
### Description
These are the changes that build on the PR Depends on https://github.com/openwrt/packages/pull/30105

### DIFF
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -260,11 +260,6 @@ return view.extend({
 		o.placeholder = '192.168.2.1/24';
 		o.rmempty = false;
 
-		o = s.taboption('general', form.Value, 'local_nat', _('Local NAT'),
-			_('NAT range for tunnels with overlapping IP addresses'));
-		o.datatype = 'cidr';
-		o.modalonly = true;
-
 		o = s.taboption('general', form.ListValue, 'if_id', ('XFRM Interface ID'),
 			_('XFRM interface ID set on input and output interfaces'));
 		o.load = function (section_id) {
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -375,6 +375,44 @@ return view.extend({
 		o.datatype = 'uinteger';
 		o.modalonly = true;
 
+		o = s.taboption('advanced', form.Value, 'rekeybytes', _('Rekey Bytes'),
+			_('Number of bytes processed before initiating CHILD_SA rekeying.') + ' ' +
+			_('Also used to derive lifebytes if set (110% of this value).') + ' ' +
+			_('Use "0" to disable byte based rekeying.')
+		);
+		o.datatype = 'uinteger';
+		o.placeholder = '0';
+		o.rmempty = true;
+		o.modalonly = true;
+
+		o = s.taboption('advanced', form.Value, 'lifebytes', _('Life Bytes'),
+			_('Maximum number of bytes processed before the CHILD_SA gets closed.') + ' ' +
+			_('Use "0" to disable (default).')
+		);
+		o.datatype = 'uinteger';
+		o.placeholder = '0';
+		o.rmempty = true;
+		o.modalonly = true;
+
+		o = s.taboption('advanced', form.Value, 'rekeypackets', _('Rekey Packets'),
+			_('Number of packets processed before initiating CHILD_SA rekeying.') + ' ' +
+			_('Also used to derive lifepackets if set (110% of this value).') + ' ' +
+			_('Use "0" to disable packet based rekeying (default).')
+		);
+		o.datatype = 'uinteger';
+		o.placeholder = '0';
+		o.rmempty = true;
+		o.modalonly = true;
+
+		o = s.taboption('advanced', form.Value, 'lifepackets', _('Life Packets'),
+			_('Maximum number of packets processed before the CHILD_SA gets closed.') + ' ' +
+			_('Use "0" to disable (default).')
+		);
+		o.datatype = 'uinteger';
+		o.placeholder = '0';
+		o.rmempty = true;
+		o.modalonly = true;
+
 		// Crypto Proposals
 		s = m.section(form.GridSection, 'crypto_proposal',
 			_('Encryption Proposals'),
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -193,6 +193,16 @@ return view.extend({
 		o.optional = true;
 		o.modalonly = true;
 
+		o = s.taboption('authentication', form.ListValue, 'send_cert', _('Send Certificate'),
+			_('Whether to send our own certificate to the remote peer'));
+		o.value('always');
+		o.value('ifasked');
+		o.value('never');
+		o.default = 'ifasked';
+		o.depends('authentication_method', 'pubkey');
+		o.optional = true;
+		o.modalonly = true;
+
 		o = s.taboption('advanced', form.Flag, 'mobike', _('MOBIKE'),
 			_('MOBIKE (IKEv2 Mobility and Multihoming Protocol)'));
 		o.default = '1';
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -203,6 +203,12 @@ return view.extend({
 		o.optional = true;
 		o.modalonly = true;
 
+		o = s.taboption('authentication', form.Flag, 'send_certreq', _('Send Certificate Request'),
+			_('Send certificate request payloads to offer trusted root CA certificates to the peer'));
+		o.default = '1';
+		o.depends('authentication_method', 'pubkey');
+		o.modalonly = true;
+
 		o = s.taboption('advanced', form.Flag, 'mobike', _('MOBIKE'),
 			_('MOBIKE (IKEv2 Mobility and Multihoming Protocol)'));
 		o.default = '1';
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -191,6 +191,8 @@ return view.extend({
 			_('MOBIKE (IKEv2 Mobility and Multihoming Protocol)'));
 		o.default = '1';
 		o.modalonly = true;
+		o.depends('keyexchange', 'ikev2');
+		o.depends('keyexchange', 'ike');
 
 		o = s.taboption('advanced', form.ListValue, 'fragmentation', _('IKE Fragmentation'),
 			_('Use IKE fragmentation'));
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -231,6 +231,14 @@ return view.extend({
 		o.validate = validateTimeFormat;
 		o.modalonly = true;
 
+		o = s.taboption('advanced', form.Flag, 'encap', _('ESP Encapsulation'),
+			_('To enforce UDP encapsulation of ESP packets, the IKE daemon can manipulate the NAT detection payloads.') + '<br />' +
+			_('This makes the peer believe that a NAT situation exist on the transmission path, forcing it to encapsulate ESP packets in UDP.') + '<br />' +
+			_('Usually this is not required but it can help to work around connectivity issues with too restrictive intermediary firewalls that block ESP packets.'));
+		o.modalonly = true;
+		o.default = '0';
+		o.rmempty = true;
+
 		o = s.taboption('advanced', form.ListValue, 'keyexchange', _('Keyexchange'),
 			_('Version of IKE for negotiation'));
 		o.value('ikev1', 'IKEv1 (%s)'.format(_('deprecated')));
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -332,11 +332,6 @@ return view.extend({
 		o.datatype = 'file';
 		o.modalonly = true;
 
-		o = s.taboption('advanced', form.Value, 'lifetime', _('Lifetime'),
-			_('Maximum duration of the CHILD_SA before closing'));
-		o.validate = validateTimeFormat;
-		o.modalonly = true;
-
 		o = s.taboption('advanced', form.ListValue, 'dpdaction', _('DPD Action'),
 			_('Action when DPD timeout occurs'));
 		o.value('none');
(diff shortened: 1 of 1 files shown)


applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -343,8 +364,20 @@ return view.extend({
 		o.modalonly = true;
 
 		o = s.taboption('advanced', form.Value, 'rekeytime', _('Rekey Time'),
-			_('Duration of the CHILD_SA before rekeying'));
+			_('Interval before an CHILD_SA is rekeyed.') + ' ' +
+			_('Also used to derive lifetime (110% of this value).') + ' ' +
+			_('If not configured, the default value is "4h".')
+		);
+		o.placeholder = '4h';
```
openwrt-ai: The advertised default is wrong: `4h` is the **IKE_SA** default, not the CHILD_SA one. This option is in the `tunnel` section and maps to `connections.<conn>.children.<child>.rekey_time`, whose default is `1h`:

- [`connections.<conn>.children.<child>.rekey_time = 1h or life_time - 10%`](https://github.com/strongswan/strongswan/blob/ea7a721dd8f7ff429109eb33b4f8dd5894be8f22/src/swanctl/swanctl.opt#L825)
- [`connections.<conn>.rekey_time = 4h`](https://github.com/strongswan/strongswan/blob/ea7a721dd8f7ff429109eb33b4f8dd5894be8f22/src/swanctl/swanctl.opt#L276) (the IKE_SA one, which is the separate `rekeytime` option in the remote section)

Also fixes "an CHILD_SA" -> "a CHILD_SA" while the string is still new and untranslated.

```suggestion
		o = s.taboption('advanced', form.Value, 'rekeytime', _('Rekey Time'),
			_('Interval before a CHILD_SA is rekeyed.') + ' ' +
			_('Also used to derive lifetime (110% of this value).') + ' ' +
			_('If not configured, the default value is "1h".')
		);
		o.placeholder = '1h';
```

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -230,6 +253,14 @@ return view.extend({
 		o.validate = validateTimeFormat;
 		o.modalonly = true;
 
+		o = s.taboption('advanced', form.Flag, "encap", _('ESP Encapsulation'),
+			_('To enforce UDP encapsulation of ESP packets, the IKE daemon can manipulate the NAT detection payloads.') + '</br>' +
+			_('This makes the peer believe that a NAT situation exist on the transmission path, forcing it to encapsulate ESP packets in UDP.') + '</br>' +
+			_('Usually this is not required but it can help to work around connectivity issues with too restrictive intermediary firewalls that block ESP packets.'));
+		o.modalonly = true;
+		o.default = 0;
+		o.rmempty = true;
```
openwrt-ai: Three problems in this block:

1. `</br>` is not a valid tag. Descriptions are injected via `innerHTML` (`dom.append()` sets `node.innerHTML` for string children), so this really is parsed as markup — and `</br>` appears nowhere else in the tree, while `<br />` is used 528 times.
2. `"encap"` is the only double-quoted option name in this file; everything else uses single quotes.
3. `o.default = 0` is a number, whereas `form.Flag.disabled` is the string `'0'` and every other default in this file is a string (`o.default = '1'`, `'yes'`, `'ifasked'`). It happens to survive the loose `fval == this.default` comparison in `form.Flag.parse()`, but it is inconsistent and fragile.

```suggestion
		o = s.taboption('advanced', form.Flag, 'encap', _('ESP Encapsulation'),
			_('To enforce UDP encapsulation of ESP packets, the IKE daemon can manipulate the NAT detection payloads.') + '<br />' +
			_('This makes the peer believe that a NAT situation exist on the transmission path, forcing it to encapsulate ESP packets in UDP.') + '<br />' +
			_('Usually this is not required but it can help to work around connectivity issues with too restrictive intermediary firewalls that block ESP packets.'));
		o.modalonly = true;
		o.default = '0';
		o.rmempty = true;
```

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -188,11 +186,35 @@ return view.extend({
 		o.depends('authentication_method', 'pubkey');
 		o.modalonly = true;
 
+		o = s.taboption('authentication', form.DynamicList, 'remote_ca_certs', _('Remote CA Certificates'),
+			_('Restrict the remote peer\'s certificate to be issued by one of these CAs'));
+		o.datatype = 'file';
+		o.depends('authentication_method', 'pubkey');
+		o.optional = true;
+		o.modalonly = true;
```
openwrt-ai, line 194: The commit message for 6a94ae2 says this is useful "e.g. when using eap-tls or general pubkey auth", but `depends('authentication_method', 'pubkey')` makes the field unreachable for EAP setups — the `authentication_method` ListValue above only offers `psk` and `pubkey`, so an `eap-tls` remote configured directly in UCI will have this field hidden and, worse, silently stripped on the next save of that section.

The backend does treat EAP-TLS as the main consumer of this option: [`# EAP-TLS approved client certs are restricted by remote_ca_certs option`](https://github.com/openwrt/packages/blob/220e15c69a139493c21a2e15b27f59c7ae5047c0/net/strongswan/files/swanctl.init#L642), and it handles `eap-mschapv2` / `eap-tls` as first-class `auth_method` values.

Should the dependency also cover `eap-tls` (and the `authentication_method` list be extended with the EAP methods the init script supports)?

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -85,23 +85,21 @@ return view.extend({
...
 		o.rmempty = false;
 
-		o = s.taboption('general', form.Value, 'gateway', _('Gateway (Remote Endpoint)'),
-			_('IP address or FQDN name of the tunnel remote endpoint'));
+		o = s.taboption('general', form.DynamicList, 'remote_addrs', _('Remote Endpoints'),
+			_('IP address or FQDN name of the tunnel remote endpoints.') + ' ' +
+			_('If no value is specified, "%any" is assumed.'));
 		o.datatype = 'or(hostname,ipaddr)';
-		o.rmempty = false;
+		o.placeholder = '%any'
```
openwrt-ai: nit: missing terminating semicolon — relies on ASI, and every other statement in this file is terminated. Same on line 98.

```suggestion
		o.placeholder = '%any';
```

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -206,18 +228,19 @@ return view.extend({
 		o = s.taboption('advanced', form.Value, 'keyingtries', _('Keying Retries'),
 			_('Number of retransmissions attempts during initial negotiation'));
 		o.datatype = 'uinteger';
-		o.default = '3';
+		o.placeholder = '3';
```
openwrt-ai: nit: while you are touching this option — the init script has an explicit branch for the magic value `%forever`: [`[ "$keyingtries" = "%forever" ] && swanctl_xappend2 "keyingtries = 0" || ...`](https://github.com/openwrt/packages/blob/220e15c69a139493c21a2e15b27f59c7ae5047c0/net/strongswan/files/swanctl.init#L608), but `datatype = 'uinteger'` rejects it, so that backend path can never be reached from LuCI. Was leaving it out intentional, or should the datatype be widened (e.g. `or(uinteger,"%forever")`)?

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -371,6 +404,45 @@ return view.extend({
...
+		o.datatype = 'uinteger';
+		o.placeholder = '0';
+		o.rmempty = true;
+		o.modalonly = true;
+
+		o = s.taboption('advanced', form.Value, 'lifepackets', _('Life Packets'),
+			_('Maximum number of packets processed before the CHILD_SA gets closed.') + ' ' +
+			_('Use "0" to disable (default).')
+
+		);
```
openwrt-ai, line 439: nit: stray blank line inside the argument list — the three sibling options added in the same commit (`rekeybytes`, `lifebytes`, `rekeypackets`) don't have it.

```suggestion
		o = s.taboption('advanced', form.Value, 'lifepackets', _('Life Packets'),
			_('Maximum number of packets processed before the CHILD_SA gets closed.') + ' ' +
			_('Use "0" to disable (default).')
		);
```

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -85,23 +85,21 @@ return view.extend({
 			_('Configuration is enabled or not'));
 		o.rmempty = false;
 
-		o = s.taboption('general', form.Value, 'gateway', _('Gateway (Remote Endpoint)'),
-			_('IP address or FQDN name of the tunnel remote endpoint'));
+		o = s.taboption('general', form.DynamicList, 'remote_addrs', _('Remote Endpoints'),
```
openwrt-ai, line 88: nit: series hygiene — 2115f5b renames `gateway` -> `remote_gateway` (and rewrites the title/description/`rmempty`/`placeholder`), then f75c790 immediately renames `remote_gateway` -> `remote_addrs` and converts it to a `DynamicList`. The intermediate `remote_gateway` state never ships. Even if this mirrors the commit split in openwrt/packages#30105, on the LuCI side the two could be squashed into one "rename `gateway` to `remote_addrs` and make it a list" commit — fewer translatable strings churned, and no bisect point where the UI writes an option no released backend reads.

---

openwrt-ai (commented):
Reviewed 14 commits (fresh review, head `644ca6d`). CI is green on the head commit.

Backend coupling checked against the declared dependency openwrt/packages#30105 (head `220e15c`): `remote_addrs`, `local_addrs`, `vips`, `remote_ca_certs`, `send_cert`, `send_certreq`, `encap`, `rekeybytes`, `lifebytes`, `rekeypackets`, `lifepackets` all have a consumer in `net/strongswan/files/swanctl.init`, and the removed `local_nat` / `local_gateway` / `gateway` / `local_ip` / `local_sourceip` are gone there (with uci-defaults migrations on the package side). Note that none of the renames exist in `packages` master yet, so this must not land before openwrt/packages#30105.

## Commit checks

- `512cc43` "luci-app-strongswan-swanctl: add ikev2 dependency to mobike" — the message says the option "is now only showed when IKEv2 is selected as the IKE version", but the diff also adds `o.depends('keyexchange', 'ike')`, so it is shown for the `ike` (both versions) selection too. The code looks right; the message should mention `ike`.
- `0dbeb0e` "luci-app-strongswan-swanctl: add missing placeholder" — besides adding placeholders, the diff also *removes* `o.default = '3'` (keyingtries) and `o.default = '30s'` (dpddelay). That is a behaviour change (LuCI stops writing those values into `/etc/config/ipsec`) and is not mentioned in the message.
- `a2e12c9` "luci-app-strongswan-swanctl: group CHILD_SA rekeytime and lifetime together" — the diff does considerably more than regroup two options: it rewrites the `rekeytime` description, adds `placeholder = '4h'` and `rmempty`, renames the `lifetime` label from "Lifetime" to "Life Time" and rewrites its description. Either split the reorder from the semantic/label changes, or describe them in the body.
- `6fb3207` "luci-app-strongswan-swanctl: add rekey/lifebyte and rekey/lifepackets options" — the body references `'/etc/initd./swanctl.init'`; the actual path is `/etc/init.d/swanctl`. The subject also reads "lifebyte" (singular) while the diff adds four options (`rekeybytes`, `lifebytes`, `rekeypackets`, `lifepackets`).
- `e1cfe41` "luci-app-strongswan-swanctl: remove deprecated local_nat" — "The 'local_nat' configuration option has been removed from the '/etc/init.d/swanctl' start up script recently" is not accurate yet: `local_nat` is still handled in `packages` master and is only removed by the pending openwrt/packages#30105. Worth rewording to state the dependency explicitly.

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
feckert: Do not squash this so that the change is traceable.

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
feckert: Updated commit message

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
```diff
@@ -205,19 +227,20 @@ return view.extend({
 
 		o = s.taboption('advanced', form.Value, 'keyingtries', _('Keying Retries'),
 			_('Number of retransmissions attempts during initial negotiation'));
-		o.datatype = 'uinteger';
-		o.default = '3';
+		o.datatype = 'or(uinteger, "%forever")';
+		o.placeholder = '3';
```
openwrt-ai, line 231: nit: the widened datatype is correct — I verified `or(uinteger, "%forever")` compiles (the [datatype compiler trims whitespace around each token and unquotes string literals](``https://github.com/openwrt/luci/blob/2aa91342814abda1d508e5c71ff23f6d4021947a/modules/luci-base/htdocs/luci-static/resources/validation.js#L326-L338),`` and [`or()` compares non-function arguments against the value](``https://github.com/openwrt/luci/blob/2aa91342814abda1d508e5c71ff23f6d4021947a/modules/luci-base/htdocs/luci-static/resources/validation.js#L1110-L1116)),`` so the [`%forever` branch in the init script](https://github.com/openwrt/packages/blob/7933a6fe8e85ea0bd9bfe33126e0858962eb5128/net/strongswan/files/swanctl.init#L612) is now reachable.

But nothing in the UI tells the user the magic value exists: the label says "Keying Retries", the description says "Number of retransmissions attempts", and the placeholder is `3`. A user will only discover `%forever` by typing it and noticing that validation does not reject it. Since the whole point of this hunk is to expose that value, mention it in the description:

```suggestion
		o = s.taboption('advanced', form.Value, 'keyingtries', _('Keying Retries'),
			_('Number of retransmissions attempts during initial negotiation.') + ' ' +
			_('Use "%forever" to keep retrying indefinitely.'));
		o.datatype = 'or(uinteger, "%forever")';
		o.placeholder = '3';
```

---

openwrt-ai (commented):
Re-review of the force-push (`644ca6d` → `2aa9134`). The branch was rebased, so all 14 commits have new SHAs; content-wise the delta is limited to `swanctl.js` (+11/-12). All CI checks are green on the head commit.

Everything from the previous round's inline comments has been applied: the two missing semicolons, the `keyingtries` datatype widened to accept `%forever`, `encap` (single quotes, `<br />` instead of `</br>`, `default = '0'`), the CHILD_SA `rekeytime` default corrected from `4h` to `1h` plus the "an CHILD_SA" → "a CHILD_SA" fix, and the stray blank line in `lifepackets`. The `Lifetime` → `Life Time` relabel also turns out to match the existing string in [`status.js:108`](``https://github.com/openwrt/luci/blob/2aa91342814abda1d508e5c71ff23f6d4021947a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/status.js#L108),`` so that is a consistency improvement rather than churn — good.

Backend coupling re-checked against `openwrt/packages` master (`7933a6f`): the renames this PR performs are still **not** in `packages` master — `swanctl.init` there still reads `gateway`, `local_ip`, `local_sourceip` and `local_nat`, and writes `local_addrs`/`remote_addrs`/`vips` from those. So this still must not land before openwrt/packages#30105.

## Commit checks

- `18e1a86` "luci-app-strongswan-swanctl: add remote_ca_certs option" — the reword dropped a word: "e.g. when general pubkey auth." (was "e.g. when using eap-tls or general pubkey auth."). Reads as "e.g. when using general pubkey auth."?

The following four were raised last round and are unchanged in the rebased commits — repeating only as a checklist, no new analysis:

- `e98842d` "add ikev2 dependency to mobike" — body says the option "is now only showed when IKEv2 is selected", but the diff also adds `o.depends('keyexchange', 'ike')`.
- `d696edf` "add rekey/lifebyte and rekey/lifepackets options" — body still says `'/etc/initd./swanctl.init'` (should be `/etc/init.d/swanctl`), and the subject says "lifebyte" while four options are added.
- `aee288b` "group CHILD_SA rekeytime and lifetime together" — the diff also rewrites both descriptions, adds `placeholder`/`rmempty`, and relabels `Lifetime` → `Life Time`; none of that is in the body.
- `8ea8bf5` "remove deprecated local_nat" — "has been removed from the '/etc/init.d/swanctl' start up script recently" is still not true of `packages` master; it is removed only by the pending openwrt/packages#30105.

---

applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
feckert: This option is intended solely for backwards compatibility with the corresponding configuration option for the deprecated stroke backend. Normally, this configuration is disabled by setting the value to '0’. This should not be specifically mentioned in the help section, to use this option.

PR #8885, opened 2026-07-29
luci-app-nlbwmon: use colour blind safe pie chart colours instead of shades of red

maxemig:
### Description

The current **Netlink Bandwidth Monitor** pie chart colours for _Traffic Distribution_ as well as _Application Protocols_ are a poor fit for just about anyone, in practice they render as indistinguishable shades of red.

I am proposing replacing them with the so-called [Okabe-Ito](https://jfly.uni-koeln.de/color/) palette which is one of the best palettes for users with colour blindness and also a big improvement over the status quo.

No trademark protections apply to these colours.

The two IPv6 pie charts retain their current colours as they are hard-coded and do not suffer from the same issue.

Before:
Screenshot 2026-07-29 at 15 36 29
Screenshot 2026-07-29 at 15 36 37

After:
Screenshot 2026-07-29 at 15 39 30
Screenshot 2026-07-29 at 15 39 40

### Maintainer _(preferred)_
@jow- 

---

## Tested on
**LuCI version:** LuCI (HEAD detached at 128a7812) branch (26.180.75667~128a781)
**Web browser(s):** Chrome 150
**Platform:** amd64

---

## Checklist

### DIFF
--- a/applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/display.js
+++ b/applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/display.js
@@ -14,6 +14,22 @@ const callNetworkRrdnsLookup = rpc.declare({
 	expect: { '': {} }
 });
 
+/* Okabe-Ito colour blind safe qualitative palette (minus its black entry,
+ * which would be a poor choice on the dark theme). Its colours are clearly
+ * distinguishable for most forms of colour blindness and are also a
+ * solid choice for everyone else. They will be re-used sequentially,
+ * which is standard practice for pie charts.
+ * See https://jfly.uni-koeln.de/color/ */
+const chartColors = [
+	'#e69f00', '#56b4e9', '#009e73', '#f0e442',
+	'#0072b2', '#d55e00', '#cc79a7'
+];
+
+/* We calculate the highlight colour for the table rows by appending an
+ * alpha channel to the segment colour, 0x59 translating to approx 35%
+ * opacity. */
+const chartRowAlpha = '59';
+
 const chartRegistry = {};
 let trafficPeriods = [];
 let trafficData = { columns: [], data: [] };
@@ -132,9 +148,15 @@ return view.extend({
 
 		for (let i = 0; i < data.length; i++) {
 			if (!data[i].color) {
-				const hue = 120 / (data.length-1) * i;
-				data[i].color = 'hsl(%u, 80%%, 50%%)'.format(hue);
-				data[i].label.push(hue);
+				let color = chartColors[i % chartColors.length];
+
+				/* Avoid repeating adjacent colours in the pie chart where
+				 * they would otherwise occur. */
+				if (i > 0 && i === data.length - 1 && color === data[0].color)
+					color = chartColors[(i + 1) % chartColors.length];
+
+				data[i].color = color;
+				data[i].label.push(color + chartRowAlpha);
 			}
 		}
 
@@ -747,10 +769,10 @@ return view.extend({
 			tooltipEl.style.top = pos[1] + tooltip.y - tooltip.caretHeight - tooltip.caretPadding + 'px';
 
 			const row = findParent(tooltip.text[1], '.tr');
-			const hue = tooltip.text[2];
+			const highlight = tooltip.text[2];
 
-			if (row && !isNaN(hue)) {
-				row.style.backgroundColor = 'hsl(%u, 100%%, 80%%)'.format(hue);
+			if (row && highlight) {
+				row.style.backgroundColor = highlight;
 				tooltipEl.row = row;
 			}
 		}, this);


applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/display.js
```diff
@@ -132,9 +143,8 @@ return view.extend({
 
 		for (let i = 0; i < data.length; i++) {
 			if (!data[i].color) {
-				const hue = 120 / (data.length-1) * i;
-				data[i].color = 'hsl(%u, 80%%, 50%%)'.format(hue);
-				data[i].label.push(hue);
+				data[i].color = chartColors[i % chartColors.length];
```
openwrt-ai: The commit message notes that colours are re-used sequentially, but the doughnut is a ring, so segment `data.length-1` is drawn adjacent to segment `0`. Whenever `data.length % 7 == 1` those two touching segments get the identical colour (`#e69f00`), which is exactly the ambiguity this change is trying to remove. The old `120 / (data.length-1) * i` scheme could not produce that because it spread the hue over the whole set.

`traf-pie` / `conn-pie` are fed one entry per host ([`renderHostData()`](``https://github.com/openwrt/luci/blob/5ce5388a9c78d58e3d21b2daf8b4b751a979e377/applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/display.js#L347-L356)),`` so on a typical home network `data.length` well exceeds 7 and the wrap case is hit regularly. Is that acceptable, or is it worth e.g. varying lightness on the second pass through the palette, or at least skipping index 0 on the final segment when the counts collide?

---

applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/display.js
```diff
@@ -132,9 +143,8 @@ return view.extend({
...
+				data[i].label.push(data[i].color + '59');
```
openwrt-ai: nit: `'59'` is an unexplained magic string here — nothing at this call site or at the `chartColors` definition says it is an alpha channel, and the concatenation silently requires every palette entry to stay a 6-digit `#rrggbb` literal. If someone later adds an `rgb(...)` or named colour to `chartColors`, the result is an invalid CSS colour and [`row.style.backgroundColor = highlight`](``https://github.com/openwrt/luci/blob/5ce5388a9c78d58e3d21b2daf8b4b751a979e377/applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/display.js#L763)`` just does nothing — the row highlight disappears with no error anywhere.

At minimum, state the intent on the line:

```suggestion
				/* re-use the segment colour at ~35% alpha as the table row highlight */
				data[i].label.push(data[i].color + '59');
```

(A named constant next to `chartColors` would be even clearer, but that is outside this hunk.)

---

openwrt-ai (commented):
Reviewed 1 new commit; 2 inline comments. The commit message matches the diff, and the palette values match the canonical Okabe-Ito set (black omitted, as the code comment explains). CI on `5ce5388` is green.

---

maxemig:
Ugh, onto the AI review:

The whole pre-existing code does not have a single comment so I am already exceeding that.

Regarding the theorised re-use of colours, very slim chance in practice, would be so small that no-one can read it anyway. I can try to add a mitigation for that later.

---

I will wait on human feedback/review before I touch it again.

maxemig:
Well, I addressed all of the clanker nits except that I still keep the same colours even if they occur multiple times as a deliberate design decision.

I have other ideas for improvement of this otherwise awesome solution, but this was the visually most annoying one.

applications/luci-app-nlbwmon/htdocs/luci-static/resources/view/nlbw/display.js
```diff
@@ -132,9 +148,15 @@ return view.extend({
...
 			if (!data[i].color) {
-				const hue = 120 / (data.length-1) * i;
-				data[i].color = 'hsl(%u, 80%%, 50%%)'.format(hue);
-				data[i].label.push(hue);
+				let color = chartColors[i % chartColors.length];
+
+				/* Avoid repeating adjacent colours in the pie chart where
+				 * they would otherwise occur. */
+				if (i > 0 && i === data.length - 1 && color === data[0].color)
+					color = chartColors[(i + 1) % chartColors.length];
```
openwrt-ai, line 156: nit: the comment says "adjacent colours ... where they would otherwise occur", but the condition is narrower and non-obvious: consecutive palette indices can never collide, so the only clash possible is the ring wrap between the last segment and segment 0. Spelling that out makes the `i === data.length - 1` guard self-explanatory:

```suggestion
				/* The doughnut is a ring, so the last segment touches the
				 * first one. Consecutive palette indices never collide, so
				 * this wrap-around is the only adjacent-colour clash that
				 * can occur — shift the last segment when it hits it. */
				if (i > 0 && i === data.length - 1 && color === data[0].color)
					color = chartColors[(i + 1) % chartColors.length];
```

Also nit: the commit body is still the original palette-swap rationale verbatim and doesn't mention this hunk or `chartRowAlpha` at all — worth a sentence so `git log` explains why the special case exists.

---

openwrt-ai (commented):
Re-reviewed the force-pushed head `a7aaf2d` against the previously reviewed `5ce5388`.

Both earlier findings are addressed, and I checked the new wrap-around logic exhaustively — it is correct: `chartColors[i % 7]` and `chartColors[(i+1) % 7]` can never coincide, so the last↔first pair is the only adjacency that can clash, and the replacement colour `chartColors[1]` cannot collide with either neighbour (`data[0]` is index 0, `data[n-2]` is index 6). The `i > 0` guard covers the single-segment case, and the `!isNaN(hue)` → truthy-string check in the tooltip handler behaves the same for the label arrays that carry no highlight (`no traffic`, the two hard-coded IPv6 pies).

Only a documentation nit left inline. FormalityCheck is green on `a7aaf2d`.

---

maxemig:
I am aware we are losing the previous shade of red differentiation thing which was not working for me in practice anyway and is redundant information because the percentage is also quite clear to see in terms of size which truly distinguishable colours anyway.

jow-:
Merged, thanks!

PR #8871, opened 2026-07-25
luci-app-openthread: add new application

LorbusChris:
Adds a LuCI web interface for the **OpenThread Border Router** (`openthread-br`),
proposed to be moved out of the ot-br-posix tree at the maintainers' request
(openthread/ot-br-posix#3431, openwrt/packages#29791) and rewritten for the
modern client-side LuCI framework.

**Features:** network overview with a live neighbor table, network scan & join,
network settings (channel/PAN ID/keys) including MAC-filter management, and
joiner commissioning.

**Architecture:** client-side JS views (`view.extend` / `rpc.declare` / `poll` /
`ui` / `dom`) over an rpcd **ucode** backend (`luci.openthread`) that brokers
ACL-gated access to otbr-agent's ubus API. The raw `otbr` ubus object is not
exposed to LuCI sessions. Depends on `rpcd-mod-ucode`.

The app reads otbr-agent's existing ubus reply fields, so it works against the
currently packaged `openthread-br` — no ot-br-posix change is required.

## Coordination

- **openwrt/packages#30096** (open) proposes removing the old in-package
  `luci-app-openthread` from the `openthread-br` feed. That should land together
  with this so the two feeds don't ship a same-named package.
- **openthread/ot-br-posix#3326** (open, draft) proposes dropping the in-tree copy
  there.

Both are still proposals under review; nothing has been removed anywhere yet.

The app reads otbr-agent's existing ubus reply fields, so no ot-br-posix change is
required for it to work. Note otbr-agent returns its `otError` code in a capital
`Error` field; the ucode backend maps it to lowercase `error` internally, so no
daemon-side change is implied by that either.

Interface names shown in the UI come from otbr-agent's `interfacename` reply
rather than a hardcoded `wpan0`.

On access control: the read ACL covers status, neighbour and scan methods only.
Every method that can expose or change credentials — network key, PSKc, MAC
filter, joiner management — requires the write ACL. Form inputs use
non-functional placeholders rather than example values, and the PSKd, EUI-64 and
network-key fields are validity-checked client-side before any call is issued.

## Testing

Syntax/convention validation: JS/ucode parse, ACL/menu JSON valid, `i18n-scan`
template generated, `luci.mk` app layout, and every ubus reply field the app reads
cross-checked against the names otbr-agent's ubus server actually emits. Not yet
exercised on a live device.

PR #8890, opened 2026-07-30
luci-app-openthread: read the stopped-network dataset via ubus status

LorbusChris:
Stacked on #8871 — only the top commit is new. **Draft** until its backend prerequisite exists upstream: the `status` ubus method is part of the otbr Matter-integration series being upstreamed to openthread/ot-br-posix (coordination with its author in project-chip/matter-openwrt#53; the surrounding series is openthread/ot-br-posix#3489 plus a follow-up). The ot-br-posix PR will be linked here once it is open.

The `status` method returns the committed Active Operational Dataset as TLVs. Prefer it for the stopped-network summary: one round-trip instead of three, and the values come from the dataset itself rather than the per-field getters. Without the method the stock fallback path stays in use, so the app keeps working against an unpatched otbr-agent either way.

### DIFF
--- a/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
+++ b/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
@@ -1,0 +1,564 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026, The OpenThread Authors.
+'use strict';
+
+let ubus = require('ubus').connect();
+
+function otbr(method, args) {
+	// On total ubus failure (otbr-agent not running), synthesize an error
+	// reply; 255 maps to GenericError in the UI.
+	let r = ubus.call('otbr', method, args ?? {});
+	if (r == null)
+		return { error: 255 };
+	// The daemon reports its otError code in an 'Error' field (see
+	// AppendResult() in ot-br-posix src/openwrt/ubus/otubus.cpp); normalize
+	// it to the lowercase 'error' used throughout this file.
+	if (r.error == null && r.Error != null)
+		r.error = r.Error;
+	return r;
+}
+
+function radio_restart() {
+	// Bouncing the netifd interface that runs the border router respawns
+	// otbr-agent; that is the closest analogue of restarting a radio.
+	let dump = ubus.call('network.interface', 'dump');
+	for (let iface in dump?.interface) {
+		if (iface.proto != 'thread' && iface.proto != 'openthread')
+			continue;
+		ubus.call('network.interface.' + iface.interface, 'down');
+		ubus.call('network.interface.' + iface.interface, 'up');
+		return { error: 0, interface: iface.interface };
+	}
+	return { error: 255 };
+}
+
+function summarize_state() {
+	let s = otbr('state');
+	// Distinguish "otbr-agent is not reachable" from "Thread is disabled";
+	// without this both render as a disabled network.
+	if (s.error)
+		return { error: s.error };
+	let result = { state: s.State ?? 'disabled' };
+	if (result.state != 'disabled') {
+		result.configured = true;
+		result.networkname = otbr('networkname').NetworkName;
+		result.panid = otbr('panid').PanId;
+		result.channel = otbr('channel').Channel;
+		result.rloc16 = otbr('rloc16').rloc16;
+		let leader = otbr('leaderdata').leaderdata;
+		if (leader) result.leaderdata = leader;
+	} else {
+		// A stopped network keeps its committed dataset, which the stock
+		// getters reflect even while Thread is disabled; report it so the
+		// UI can offer to enable the network rather than create one. An
+		// unprovisioned device reads as the broadcast PAN id.
+		// The stock ubus formats the PAN id "0x%04x", but parse rather
+		// than compare the spelling: a bare hex string or a plain number
+		// must not make a factory-fresh device read as provisioned, and
+		// anything unparseable counts as unprovisioned.
+		let raw = otbr('panid').PanId;
+		// ucode's numeric coercion handles "0xffff", "65535" and a plain
+		// number alike, and yields NaN for anything else; NaN != NaN, so
+		// an unparseable reply counts as unprovisioned.
+		let pan = +raw;
+		result.configured = (raw != null && pan == pan && pan != 0xffff);
+		if (result.configured) {
+			result.networkname = otbr('networkname').NetworkName;
+			result.panid = raw;
+			result.channel = otbr('channel').Channel;
+		}
+	}
+	result.interfacename = otbr('interfacename').InterfaceName;
+	return result;
+}
+
+function otctl(args) {
+	// The SRP registrations are not exposed over ubus; read them from the
+	// CLI socket the same way ot-ctl does. Root-only rpcd context.
+	let fs = require('fs');
+	let bin = fs.access('/usr/sbin/ot-ctl', 'x') ? '/usr/sbin/ot-ctl' : 'ot-ctl';
+	return fs.popen(bin + ' ' + args, 'r');
+}
+
+function srp_hostnames() {
+	// address (lowercase) -> registered SRP host name
+	let p = otctl('srp server host');
+	if (!p)
+		return {};
+	let map = {};
+	let host = null;
+	for (let line = p.read('line'); length(line); line = p.read('line')) {
+		line = trim(line);
+		let m = match(line, /^([^.\s]+)\.default\.service\.arpa\.$/);
+		if (m) {
+			host = m[1];
+			continue;
+		}
+		m = match(line, /^addresses: \[(.+)\]$/);
+		if (m && host)
+			for (let a in split(m[1], ', '))
+				map[lc(trim(a))] = host;
+	}
+	p.close();
+	return map;
+}
+
+function child_addresses() {
+	// child rloc16 (lowercase hex, no 0x) -> [registered addresses]
+	let p = otctl('childip');
+	if (!p)
+		return {};
+	let map = {};
+	for (let line = p.read('line'); length(line); line = p.read('line')) {
+		let m = match(trim(line), /^([0-9a-fA-F]{4}): (.+)$/);
+		if (m) {
+			let k = lc(m[1]);
+			map[k] ??= [];
+			push(map[k], lc(trim(m[2])));
+		}
+	}
+	p.close();
+	return map;
+}
+
+// SRP registrations change on the order of minutes; the overview polls
+// every two seconds. Cache the CLI reads so a watching browser does not
+// fork ot-ctl twice per tick.
+let srp_cache = { at: 0, hosts: {}, children: {} };
+
+function srp_lookup() {
+	let now = time();
+	if (srp_cache.at == 0 || now < srp_cache.at || now - srp_cache.at > 30) {
+		srp_cache.hosts = srp_hostnames();
+		srp_cache.children = length(srp_cache.hosts) ? child_addresses() : {};
+		srp_cache.at = now;
+	}
+	return srp_cache;
+}
+
+// One address list serves both filter modes, so an edit that is harmless
+// under one meaning locks devices out under the other. Every mutation is
+// therefore evaluated against the mode and the current contents. A read
+// that failed is not an empty list: the daemon reports an unreadable mode
+// as the literal string "error", and always emits addrlist as an array,
+// so anything else means the state is unknown and no edit is safe.
+function macfilter_mode() {
+	let mode = otbr('macfilterstate').state;
+	return (mode == null || lc(mode) == 'error') ? null : lc(mode);
+}
+
+function macfilter_list() {
+	let list = otbr('macfilteraddr').addrlist;
+	return (type(list) == 'array') ? map(list, lc) : null;
+}
+
+function macfilter_snapshot() {
+	let mode = macfilter_mode();
+	let list = macfilter_list();
+	return (mode == null || list == null) ? null : { mode: mode, list: list };
+}
+
+// An absent or empty state argument means "leave the mode as it is".
+function macfilter_want(state) {
+	return length(state) ? lc(state) : null;
+}
+
+// What makes a mode change unsafe, given the list it would reinterpret.
+// Returns null when the change is safe (or is not a change at all).
+function macfilter_hazard(want, mode, list) {
+	if (want == null || want == mode)
+		return null;
+	// Entries written for the previous meaning would flip to the opposite.
+	if (length(list) > 0)
+		return 'leftovers';
+	// An allowlist that admits nobody filters every device.
+	if (want == 'allowlist' && length(list) == 0)
+		return 'empties_allowlist';
+	return null;
+}
+
+function summarize_neighbors() {
+	let s = otbr('state');
+	if (s.error)
+		return { error: s.error };
+	let state = s.State;
+	let result = { state: state, neighbor: [] };
+
+	if (state == 'disabled')
+		return result;
+
+	let list = (state == 'child')
+		? otbr('parent').parent_list
+		: otbr('neighbor').neighbor_list;
+	for (let n in list)
+		push(result.neighbor, n);
+
+	// A Thread device announces itself to the mesh by registering a host
+	// with the SRP server on this border router; that name is the closest
+	// thing to a hostname the mesh has. Children are matched through their
+	// registered addresses. (Matter devices register their extended MAC as
+	// the host name by design, so for them this adds no friendly name.)
+	if (length(result.neighbor)) {
+		let srp = srp_lookup();
+		let hosts = srp.hosts;
+		if (length(hosts)) {
+			let caddrs = srp.children;
+			for (let n in result.neighbor) {
+				let k = lc(replace(n.Rloc16 ?? '', '0x', ''));
+				for (let a in (caddrs[k] ?? [])) {
+					if (hosts[a] != null) {
+						n.Hostname = hosts[a];
+						break;
+					}
+				}
+			}
+		}
+	}
+
+	// The filter mode and each neighbour's own standing on it are what
+	// the overview needs to offer per-device actions; the full address
+	// list stays with the other settings behind the write grant.
+	// Both are left unset when unreadable, so the page can tell "not
+	// filtered" apart from "unknown" and withhold the per-device actions.
+	result.macfilter = macfilter_mode();
+	let addrlist = macfilter_list();
+	if (addrlist != null)
+		for (let n in result.neighbor)
+			n.Filtered = index(addrlist, lc(n.ExtAddress ?? '')) >= 0;
+
+	let j = otbr('joinernum');
+	result.joinernum = j.joinernum ?? 0;
+	result.joinerlist = [];
+	if ((j.joinernum ?? 0) != 0 && j.joinerList)
+		for (let entry in j.joinerList)
+			push(result.joinerlist, entry);
+
+	return result;
+}
+
+
+
+return {
+	'luci.openthread': {
+		state_summary: {
+			args: {},
+			call: function() {
+				return summarize_state();
+			}
+		},
+		radio_restart: {
+			args: {},
+			call: function() {
+				return radio_restart();
+			}
+		},
+		neighbors_summary: {
+			args: {},
+			call: function() {
+				return summarize_neighbors();
+			}
+		},
+		settings_get: {
+			args: {},
+			call: function() {
+				let st = otbr('state');
+				// Otherwise every field comes back null and the form renders
+				// blank inputs that an admin can Apply against a dead daemon.
+				if (st.error)
+					return { error: st.error };
+				let addrs = macfilter_list();
+				return {
+					networkname: otbr('networkname').NetworkName,
+					interfacename: otbr('interfacename').InterfaceName,
+					panid: otbr('panid').PanId,
+					channel: otbr('channel').Channel,
+					extpanid: otbr('extpanid').ExtPanId,
+					mode: otbr('mode').Mode,
+					// The network key and the commissioner credential are not
+					// handed out with the rest of the settings; a page that
+					// only displays the configuration has no use for them.
+					// secret_get fetches one when the operator asks for it.
+					partitionid: otbr('partitionid').Partitionid,
+					macfilterstate: macfilter_mode(),
+					addrlist: addrs,
+					state: st.State,
+				};
+			}
+		},
+		secret_get: {
+			args: { name: 'string' },
+			call: function(req) {
+				// Reading a secret is an explicit request for that one value,
+				// so it can be logged and reasoned about on its own.
+				if (req.args?.name == 'networkkey') {
+					let r = otbr('networkkey');
+					return { error: r.error ?? 0, value: r.Networkkey };
+				}
+				if (req.args?.name == 'pskc') {
+					let r = otbr('pskc');
+					return { error: r.error ?? 0, value: r.pskc };
+				}
+				return { error: 7 /* OT_ERROR_INVALID_ARGS */ };
+			}
+		},
+		scan: {
+			args: {},
+			call: function(req) {
+				// Deferred: the active scan takes several seconds and rpcd
+				// serves all of LuCI from one event loop, so a synchronous
+				// call here would freeze every page for the scan duration.
+				let replied = false;
+				let d = ubus.defer('otbr', 'scan', {}, function(rc, reply) {
+					replied = true;
+					req.reply({
+						scan_list: reply?.scan_list ?? [],
+						error: reply?.error ?? (rc ? 255 : 0)
+					});
+				});
+				if (!d)
+					return replied ? null : { scan_list: [], error: 255 };
+				return d;
+			}
+		},
+
+		settings_apply: {
+			args: {
+				networkname: 'string',
+				channel:     0,
+				panid:       'string',
+				extpanid:    'string',
+				mode:        'string',
+				networkkey:  'string',
+				pskc:        'string',
+				macfilter:   'string',
+				macfilter_confirmed: true,
+			},
+			call: function(req) {
+				let a = req.args;
+				let st = otbr('state');
+				if (st.error)
+					return { error: st.error };
+				let state = st.State;
+				// The filter mode is the one setting whose meaning depends
+				// on data the form does not carry. Refuse a change that
+				// would reinterpret an existing list or activate an
+				// allowlist that admits nobody, unless the operator
+				// confirmed it with the list on screen.
+				let want = macfilter_want(a.macfilter);
+				if (want != null) {
+					let mode = macfilter_mode();
+					let list = macfilter_list();
+					if (mode == null || list == null)
+						return { error: 0, unchanged: true, reason: 'unreadable' };
+					// Nothing to do, and nothing to warn about.
+					if (want == mode)
+						want = null;
+					else {
+						let hazard = macfilter_hazard(want, mode, list);
+						if (hazard != null && !a.macfilter_confirmed)
+							return { error: 0, unchanged: true, reason: hazard };
+					}
+				}
+				let result;
+				if (state == 'disabled') {
+					// An empty secret means "leave as it is": the form does not
+					// carry the current values, so it cannot resubmit them.
+					let steps = [
+						[ 'setnetworkname', { networkname: a.networkname } ],
+						[ 'setchannel',     { channel:     a.channel } ],
+						[ 'setpanid',       { panid:       a.panid } ],
+						[ 'setextpanid',    { extpanid:    a.extpanid } ],
+						[ 'setmode',        { mode:        a.mode } ],
+					];
+					if (length(a.networkkey))
+						push(steps, [ 'setnetworkkey', { networkkey: a.networkkey } ]);
+					if (length(a.pskc))
+						push(steps, [ 'setpskc', { pskc: a.pskc } ]);
+					for (let step in steps) {
+						let r = otbr(step[0], step[1]);
+						if ((r?.error ?? 0) != 0)
+							return { error: r.error };
+					}
+					if (want != null) {
+						let r = otbr('macfiltersetstate', { state: want });
+						if ((r?.error ?? 0) != 0)
+							return { error: r.error };
+					}
+					result = otbr('threadstart');
+				} else {
+					let args = {
+						networkname: a.networkname,
+						extpanid:    a.extpanid,
+						panid:       a.panid,
+						channel:     sprintf('%d', a.channel),
+					};
+					if (length(a.networkkey))
+						args.networkkey = a.networkkey;
+					if (length(a.pskc))
+						args.pskc = a.pskc;
+					result = otbr('mgmtset', args);
+					if ((result?.error ?? 0) == 0 && want != null)
+						result = otbr('macfiltersetstate', { state: want });
+				}
+				return { error: result?.error ?? 0 };
+			}
+		},
+		attach: {
+			args: { panid: 'string', channel: 0, networkkey: 'string' },
+			call: function(req) {
+				let a = req.args;
+				let steps = [
+					[ 'setpanid',      { panid:      a.panid } ],
+					[ 'setchannel',    { channel:    a.channel } ],
+					[ 'setnetworkkey', { networkkey: a.networkkey } ],
+				];
+				for (let step in steps) {
+					let r = otbr(step[0], step[1]);
+					if ((r?.error ?? 0) != 0)
+						return { error: r.error };
+				}
+				let r = otbr('threadstart');
+				return { error: r?.error ?? 0 };
+			}
+		},
+		thread_start: {
+			args: {},
+			call: function() {
+				let r = otbr('threadstart');
+				return { error: r?.error ?? 0 };
+			}
+		},
+		thread_stop: {
+			args: {},
+			call: function() {
+				let r = otbr('threadstop');
+				return { error: r?.error ?? 0 };
+			}
+		},
+		leave: {
+			args: {},
+			call: function() {
+				let r = otbr('leave');
+				return { error: r?.error ?? 0 };
+			}
+		},
+		commissioner_start: {
+			args: {},
+			call: function() {
+				let r = otbr('commissionerstart');
+				return { error: r?.error ?? 0 };
+			}
+		},
+		joiner_add: {
+			args: { pskd: 'string', eui64: 'string' },
+			call: function(req) {
+				let r = otbr('joineradd', { pskd: req.args.pskd, eui64: req.args.eui64 });
+				return { error: r?.error ?? 0 };
+			}
+		},
+		joiner_remove: {
+			args: { eui64: 'string' },
+			call: function(req) {
+				let r = otbr('joinerremove', { eui64: req.args.eui64 });
+				return { error: r?.error ?? 0 };
+			}
+		},
+		macfilter_clear: {
+			args: { confirmed: true },
+			call: function(req) {
+				let snap = macfilter_snapshot();
+				if (snap == null)
+					return { error: 0, unchanged: true, reason: 'unreadable' };
+				// Emptying an active allowlist blocks every device, so it
+				// needs the operator's explicit go-ahead; the settings
+				// modal asks with the list on screen.
+				if (snap.mode == 'allowlist' && length(snap.list) > 0 && !req.args.confirmed)
+					return { error: 0, unchanged: true, reason: 'empties_allowlist' };
+				let r = otbr('macfilterclear');
+				return { error: r?.error ?? 0 };
+			}
+		},
+		macfilter_add: {
+			args: { addr: 'string', state: 'string', confirmed: true },
+			call: function(req) {
+				if (!length(req.args.addr))
+					return { error: 7 };
+				let snap = macfilter_snapshot();
+				if (snap == null)
+					return { error: 0, unchanged: true, reason: 'unreadable' };
+				let addr = lc(req.args.addr);
+				let want = macfilter_want(req.args.state);
+				let others = filter(snap.list, a => a != addr);
+				// Turning a mode on next to entries from an earlier list
+				// would give those entries the opposite meaning, and the
+				// address added here would join a list that means the
+				// opposite of the intent. Change nothing at all.
+				let hazard = macfilter_hazard(want, snap.mode, others);
+				// Only leftovers can apply to an add. The address written here
+				// is in the list this produces, so it can never leave an active
+				// allowlist empty -- and reporting that would tell the operator
+				// the opposite of what would have happened.
+				if (hazard == 'empties_allowlist')
+					hazard = null;
+				if (hazard != null)
+					return { error: 0, unchanged: true, reason: hazard };
+				// Adding under the mode already in force affects this device
+				// only, but switching the mode on decides what every other
+				// device may do -- an allowlist of one blocks the rest of the
+				// mesh. macfilter_clear and settings_apply make the caller
+				// acknowledge that; so does this.
+				if (want != null && want != snap.mode && !req.args.confirmed)
+					return { error: 0, unchanged: true, reason: 'unconfirmed_mode' };
+				let r = otbr('macfilteradd', { addr: req.args.addr });
+				let err = r?.error ?? 0;
+				// Already (24) means the entry was there; the mode may
+				// still be switched on for it.
+				if (err != 0 && err != 24)
+					return { error: err };
+				if (want != null && want != snap.mode) {
+					let sr = otbr('macfiltersetstate', { state: want });
+					if ((sr?.error ?? 0) != 0)
+						return { error: sr.error, reason: 'mode_unchanged' };
+				}
+				return { error: err };
+			}
+		},
+		macfilter_remove: {
+			args: { addr: 'string', state: 'string', confirmed: true },
+			call: function(req) {
+				if (!length(req.args.addr))
+					return { error: 7 };
+				let snap = macfilter_snapshot();
+				if (snap == null)
+					return { error: 0, unchanged: true, reason: 'unreadable' };
+				let addr = lc(req.args.addr);
+				let want = macfilter_want(req.args.state);
+				let remaining = filter(snap.list, a => a != addr);
+				// A mode switch would reinterpret whatever stays behind.
+				let hazard = macfilter_hazard(want, snap.mode, remaining);
+				if (hazard != null)
+					return { error: 0, unchanged: true, reason: hazard };
+				// An active allowlist with no entries admits nobody, so
+				// emptying one filters every neighbour rather than this
+				// single device.
+				if ((want ?? snap.mode) == 'allowlist' &&
+					length(remaining) == 0 && length(snap.list) > 0)
+					return { error: 0, unchanged: true, reason: 'empties_allowlist' };
+				// Same as in macfilter_add: a mode this call would switch on
+				// reinterprets the whole mesh, not just this address.
+				if (want != null && want != snap.mode && !req.args.confirmed)
+					return { error: 0, unchanged: true, reason: 'unconfirmed_mode' };
+				let r = otbr('macfilterremove', { addr: req.args.addr });
+				let err = r?.error ?? 0;
+				if (err != 0)
+					return { error: err };
+				if (want != null && want != snap.mode) {
+					let sr = otbr('macfiltersetstate', { state: want });
+					if ((sr?.error ?? 0) != 0)
+						return { error: sr.error, reason: 'mode_unchanged' };
+				}
+				return { error: err };
+			}
+		},
+	}
+};
--- a/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
+++ b/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
@@ -2,7 +2,8 @@
 // Copyright (c) 2026, The OpenThread Authors.
 'use strict';
 
-let ubus = require('ubus').connect();
+let ubusmod = require('ubus');
+let ubus = ubusmod.connect();
 
 function otbr(method, args) {
 	// On total ubus failure (otbr-agent not running), synthesize an error
@@ -16,6 +17,59 @@ function otbr(method, args) {
 	if (r.error == null && r.Error != null)
 		r.error = r.Error;
 	return r;
+}
+
+// A stock otbr-agent has no status method, so the overview's 2 s poll must not
+// keep paying for a call that will never succeed. Latch only on
+// STATUS_METHOD_NOT_FOUND: a missing *object* answers with a different status
+// and means otbr-agent is down, which is temporary and has to stay retryable.
+//
+// The latch lives in rpcd's per-script VM rather than in otbr-agent, so it
+// outlives an otbr-agent restart: installing a build that does have the method
+// leaves this on the getter path until rpcd is restarted too. The fallback is
+// correct there, only chattier, which is the price of not re-probing forever.
+let status_absent = false;
+
+function otbr_status() {
+	if (status_absent)
+		return null;
+	let r = ubus.call('otbr', 'status', {});
+	if (r == null) {
+		if (ubus.error(true) == ubusmod.STATUS_METHOD_NOT_FOUND)
+			status_absent = true;
+		return null;
+	}
+	return r;
+}
+
+function dataset_summary(tlvhex) {
+	// Active Operational Dataset TLVs: [type:1][length:1][value:length].
+	// Only the fields the overview shows are extracted.
+	if (type(tlvhex) != 'string' || length(tlvhex) < 4)
+		return null;
+	let out = null;
+	let i = 0;
+	let n = length(tlvhex);
+	while (i + 4 <= n) {
+		let t = hex(substr(tlvhex, i, 2));
+		let l = hex(substr(tlvhex, i + 2, 2));
+		let v = substr(tlvhex, i + 4, l * 2);
+		if (length(v) < l * 2)
+			break;
+		out ??= {};
+		if (t == 0 && l == 3)
+			out.channel = hex(substr(v, 2, 4));
+		else if (t == 1 && l == 2)
+			out.panid = '0x' + lc(v);
+		else if (t == 3) {
+			let name = '';
+			for (let k = 0; k < l * 2; k += 2)
+				name += chr(hex(substr(v, k, 2)));
+			out.networkname = name;
+		}
+		i += 4 + l * 2;
+	}
+	return out;
 }
 
 function radio_restart() {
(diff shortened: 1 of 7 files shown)


applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
```diff
@@ -0,0 +1,457 @@
...
+	} else {
+		// A stopped network keeps its committed dataset; report it so the
+		// UI can offer to enable the network rather than create one. The
+		// status method hands out the dataset TLVs directly; on a stock
+		// otbr-agent without it, fall back to the per-field getters,
+		// which reflect the committed dataset even while Thread is
+		// disabled, with the broadcast PAN id marking an unprovisioned
+		// device.
+		let ds = dataset_summary(otbr('status').ActiveDataset);
+		if (ds) {
```
openwrt-ai: The unprovisioned check that the fallback branch keeps (`lc(panid) != '0xffff'`, line 96) has no counterpart on the status path: here `configured` is set to `true` for *any* truthy `ds`. Two ways that diverges from the branch it is preferred over:

- `dataset_summary()` does `out ??= {}` (line 35) *before* testing the TLV type, so a TLV stream that parses but contains none of Channel / PAN ID / Network Name returns a truthy `{}`. `configured` then goes true while `networkname`, `panid` and `channel` are all null.
- Even with a well-formed dataset, a committed PAN ID of `0xffff` is reported as configured, where the getter path deliberately treats it as unprovisioned.

`configured` is not cosmetic: it removes the **Create** button at [`overview.js:654`](``https://github.com/openwrt/luci/blob/f5b4b0072509e0215beb73b33de4d71cd8f0d129/applications/luci-app-openthread/htdocs/luci-static/resources/view/openthread/overview.js#L654)`` and swaps the "No Thread network is configured yet…" placeholder for a network row at [`overview.js:739`](``https://github.com/openwrt/luci/blob/f5b4b0072509e0215beb73b33de4d71cd8f0d129/applications/luci-app-openthread/htdocs/luci-static/resources/view/openthread/overview.js#L739).`` So the failure mode is a Thread row reading `Network ? | PAN ID <absent>` on a device that has nothing configured, with no way to create one.

This may well be unreachable in practice — if the patched otbr-agent reports an empty/absent `ActiveDataset` on an unprovisioned device, `dataset_summary()` returns null and the getter path handles it correctly. But that is a property of a daemon method that does not exist upstream yet, and the guard is one condition. Requiring the PAN ID TLV also means a dataset the decoder cannot make sense of degrades to the getters rather than to a half-filled row:

```suggestion
		if (ds != null && ds.panid != null && lc(ds.panid) != '0xffff') {
```


---

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
```diff
@@ -0,0 +1,457 @@
...
+		if (leader) result.leaderdata = leader;
+	} else {
+		// A stopped network keeps its committed dataset; report it so the
+		// UI can offer to enable the network rather than create one. The
+		// status method hands out the dataset TLVs directly; on a stock
+		// otbr-agent without it, fall back to the per-field getters,
+		// which reflect the committed dataset even while Thread is
+		// disabled, with the broadcast PAN id marking an unprovisioned
+		// device.
+		let ds = dataset_summary(otbr('status').ActiveDataset);
```
openwrt-ai: nit: on a stock otbr-agent this probe is repaid every tick. `state_summary` is on the overview's 2 s poll ([`overview.js:966-972`](``https://github.com/openwrt/luci/blob/f5b4b0072509e0215beb73b33de4d71cd8f0d129/applications/luci-app-openthread/htdocs/luci-static/resources/view/openthread/overview.js#L966-L972)),`` and while the network is disabled every tick now makes a `status` call that returns `UBUS_STATUS_METHOD_NOT_FOUND` before falling through to the three getters — four round-trips where the previous revision made one to three. Since the answer ("this daemon has no `status`") cannot change without otbr-agent restarting, latching it in a module-level flag the first time the call fails would make the unpatched path cost exactly what it did before. That is the same reasoning as the `srp_cache` TTL added one round earlier on this file, so it seemed worth raising rather than assuming it was considered.

Also worth stating in the commit message: "one round-trip instead of three" holds only against a patched agent; against a stock one the change adds a round-trip rather than saving two.

---

openwrt-ai (commented):
First review of this PR. The base commit `155161b` is the head of #8871 and is being reviewed there, so this pass is scoped to the one new commit, `f5b4b00` — a single hunk in the rpcd ucode backend.

What I checked:

- **Commit message vs. diff.** Matches: the subject scope (`luci-app-openthread:`) is the only package touched, and the body describes exactly the status-preferred path plus getter fallback that the diff implements. Nothing to flag.
- **Fallback safety.** `otbr()` substitutes `{ error: 255 }` when `ubus.call` returns null, so `otbr('status').ActiveDataset` is null rather than a throw on a daemon without the method, and `dataset_summary()` rejects it on the `type(...) != 'string'` guard. The stock path is reached, as the commit claims.
- **TLV decoding.** Types 0/1/3 are Channel, PAN ID and Network Name per the MeshCoP dataset TLV numbering, the Channel branch correctly skips the channel-page byte, and the walk is bounded at both ends (`i + 4 <= n`, truncated-value break, `l == 0` still advances). `chr()` in ucode is byte-based ([`lib.c:711-737`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib.c#L711-L737)), so a UTF-8 network name round-trips intact rather than being re-encoded per byte.
- **Type/format parity with the branch it replaces.** `panid` is emitted as `'0x' + lc(v)`, matching the `0x%04x` string the `panid` getter returns — the `lc()` added in this commit is what makes the two agree, and `overview.js` renders it through `String()` either way. `channel` is a number on both paths.
- **ACL.** No new `luci.openthread` method, and the `status` call is made by the plugin as root, so `acl.d/luci-app-openthread.json` needs no change. Correct as-is.
- **Backend coupling.** Per `.github/llm-review-rules.md` this would normally want a consumer check, but `status` lives in ot-br-posix, which is not among the pre-cloned trees; the PR body already states the method is unmerged upstream and that the PR stays draft until it lands, so I have not flagged the dependency.
- **CI.** No checks have run on `f5b4b00` yet (status `pending`, 0 checks), so there is nothing to ground against.

Two follow-ups inline: one on `configured` losing the unprovisioned guard on the new path, one nit on the per-poll cost against an unpatched agent.

---

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
```diff
@@ -0,0 +1,592 @@
...
+		let leader = otbr('leaderdata').leaderdata;
+		if (leader) result.leaderdata = leader;
+	} else {
+		// A stopped network keeps its committed dataset; report it so the
+		// UI can offer to enable the network rather than create one. The
+		// status method hands out the dataset TLVs directly; on a stock
+		// otbr-agent without it, fall back to the per-field getters, which
+		// reflect the committed dataset even while Thread is disabled.
+		let ds = dataset_summary(otbr('status').ActiveDataset);
+		if (ds) {
```
openwrt-ai: Superseding [my earlier comment on this line](https://github.com/openwrt/luci/pull/8890#discussion_r3684678355): the rebase onto the new base commit widened the gap rather than closing it, and the fix I suggested last round no longer matches the file's idiom.

The fallback branch got stricter in this revision — it now parses the PAN id and rejects both `NaN` and the broadcast value (lines 100-102). The status branch still sets `configured = true` for *any* truthy `ds`, so the two paths now disagree in two ways:

- **Unparseable input.** `hex()` returns `NaN` for a non-hex byte pair ([`lib.c:977-994`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib.c#L977-L994)). In `dataset_summary()` a garbage `ActiveDataset` gives `l = NaN`, so `length(v) < l * 2` is false (no `break`), `out ??= {}` on line 35 has already run, and `i += 4 + l * 2` is `NaN`, which ends the loop. The function returns a truthy `{}`. The fallback branch explicitly treats an unparseable reply as unprovisioned; the status branch treats it as configured with `networkname`, `panid` and `channel` all null.
- **Broadcast PAN id.** A committed dataset whose PAN ID TLV is `0xffff` reads as configured here, where the fallback deliberately does not.

`configured` is not cosmetic — it suppresses the **Create** button at [`overview.js:719`](``https://github.com/openwrt/luci/blob/29af3c416a164e17572ab505951f32d6e236ca1e/applications/luci-app-openthread/htdocs/luci-static/resources/view/openthread/overview.js#L719)`` and swaps the "No Thread network is configured yet…" placeholder for a network row at [`overview.js:804`](``https://github.com/openwrt/luci/blob/29af3c416a164e17572ab505951f32d6e236ca1e/applications/luci-app-openthread/htdocs/luci-static/resources/view/openthread/overview.js#L804),`` so the failure mode is a row reading `Network ? | PAN ID <absent>` on a device with nothing configured and no way to create one.

Applying the same test the fallback now uses also means a dataset the decoder cannot make sense of degrades to the getters instead of to a half-filled row:

```suggestion
		let pan = ds ? +ds.panid : NaN;
		if (pan == pan && pan != 0xffff) {
```

As before, this may be unreachable against the real patched agent if it reports an absent `ActiveDataset` on an unprovisioned device — but that is a property of a daemon method that is not upstream yet, and the guard is one line.

---

openwrt-ai (commented):
Re-review at `29af3c4`, against my last pass at `f5b4b00`.

Scope: the base commit moved `155161b` → `13772da` (that is #8871's content — the mac-filter hazard checks and the numeric PAN id parse — and it is reviewed there, not here). The only commit in this PR's own scope, `29af3c4`, changed by the rebase plus a comment rewrap; its logic is unchanged from `f5b4b00`.

- **Commit message vs. diff.** `29af3c4` still matches its subject and body; nothing new to flag.
- **Carried over from last round.** [The per-poll cost nit](https://github.com/openwrt/luci/pull/8890#discussion_r3684679971) (a `status` probe every 2 s tick against a stock agent, plus "one round-trip instead of three" holding only against a patched one) is unaddressed and still applies — not repeated inline.
- **New this round.** The rebase pulled a stricter unprovisioned test into the fallback branch without giving the status branch a counterpart, so the two paths now diverge more than they did at `f5b4b00`. One inline comment, with a suggestion matching the file's new idiom; it supersedes my earlier one on the same line.
- **CI.** All three FormalityCheck runs on `29af3c4` are green.

---

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
LorbusChris: Adopted your version — the string compare I had first would have passed a garbage PAN id through. `+ds.panid` with the `pan == pan` check now matches the fallback exactly.

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
LorbusChris: Latched, with one deviation: only on `Method not found`. A null reply also covers a missing *object* — otbr-agent down — and latching that would leave the page degraded until rpcd restarted. `ubus.error()` distinguishes the two. Commit message corrected on the round-trip claim; it only held against a patched agent.

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
LorbusChris: Superseded by #3699703011, fixed there.

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
```diff
@@ -0,0 +1,622 @@
...
+		// null -- which hides the Create button on a device that has nothing.
+		// The broadcast PAN id means unprovisioned, as it does on the getters.
+		// Apply the same test the fallback uses below: hex() yields NaN on a
+		// non-hex byte pair, and dataset_summary() returns a truthy {} as soon
+		// as the walk starts, so a garbage dataset would otherwise report a
+		// configured network with every field null -- hiding the Create button
+		// on a device that has nothing. NaN != NaN; 0xffff means unprovisioned.
+		let ds = dataset_summary(otbr_status()?.ActiveDataset);
+		let pan = ds ? +ds.panid : NaN;
+		if (pan == pan && pan != 0xffff) {
```
openwrt-ai: The guard does not actually require the PAN id — my suggestion last round was wrong and I should have checked this before offering it. In ucode, unary `+` on `null` yields `0`, not `NaN`: [`uc_vm_insn_plus_minus()`, vm.c:2180-2181](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/vm.c#L2180-L2181) pushes `ucv_int64_new(0)` for `UC_NULL` (only the `default:` arm two lines down returns `NAN`). That differs from the fallback branch below, which reaches the same conclusion only because it tests `raw != null` separately on line 132.

So whenever `dataset_summary()` returns a truthy object without a PAN ID TLV, `pan` is `0`, `pan == pan` holds, `0 != 0xffff` holds, and `configured` goes true with `networkname`, `panid` and `channel` all null — the exact outcome the comment above and the commit message say this line prevents. Two reachable ways in:

- **No PAN ID TLV in an otherwise valid stream.** `out ??= {}` on line 54 runs before any type test, so a dataset carrying, say, only Extended PAN ID (type 2) and Network Key (type 5) returns `{}`.
- **Garbage dataset.** `hex()` gives `NaN` for a non-hex byte pair, `length(v) < l * 2` is false so there is no `break`, line 54 has already run, and `i += 4 + l * 2` is `NaN`, ending the loop — `{}` again.

`configured` suppresses the **Create** button at [`overview.js:741`](``https://github.com/openwrt/luci/blob/8c0e7f091a2685cbe3eeca20a33b1d88bcbe3801/applications/luci-app-openthread/htdocs/luci-static/resources/view/openthread/overview.js#L741)`` and replaces the "No Thread network is configured yet…" placeholder with a network row at [`overview.js:826-831`](``https://github.com/openwrt/luci/blob/8c0e7f091a2685cbe3eeca20a33b1d88bcbe3801/applications/luci-app-openthread/htdocs/luci-static/resources/view/openthread/overview.js#L826-L831),`` so the failure mode is still a row reading `Network ? | PAN ID <absent>` with no way to create one.

Testing the property directly also folds the `ds != null` case in, since `ds?.panid` is null either way:

```suggestion
		let pan = ds?.panid != null ? +ds.panid : NaN;
		if (pan == pan && pan != 0xffff) {
```

---

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
```diff
@@ -0,0 +1,622 @@
...
+		// Require the PAN id: dataset_summary() returns a truthy object as soon
+		// as the TLV walk starts, so a stream carrying none of the fields shown
+		// here would otherwise report a configured network with every field
+		// null -- which hides the Create button on a device that has nothing.
+		// The broadcast PAN id means unprovisioned, as it does on the getters.
+		// Apply the same test the fallback uses below: hex() yields NaN on a
+		// non-hex byte pair, and dataset_summary() returns a truthy {} as soon
+		// as the walk starts, so a garbage dataset would otherwise report a
+		// configured network with every field null -- hiding the Create button
+		// on a device that has nothing. NaN != NaN; 0xffff means unprovisioned.
```
openwrt-ai: nit: the rewrite left both drafts in place — lines 105-109 and lines 110-114 make the same argument twice ("`dataset_summary()` returns a truthy object as soon as the walk starts … would otherwise report a configured network with every field null … hides the Create button on a device that has nothing", then the broadcast-PAN-id sentence). The second version is the one that matches the code, since it names `hex()`/`NaN`; merging them keeps the `hex()` detail and the "require the PAN id" framing without the repetition:

```suggestion
		// Apply the same test the fallback uses below: require the PAN id and
		// reject the broadcast value. dataset_summary() returns a truthy {} as
		// soon as the walk starts, and hex() yields NaN on a non-hex byte pair,
		// so a garbage dataset -- or one carrying none of the fields shown here
		// -- would otherwise report a configured network with every field null,
		// hiding the Create button on a device that has nothing configured.
		// NaN != NaN; the broadcast PAN id 0xffff means unprovisioned.
```

---

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
```diff
@@ -0,0 +1,622 @@
...
+		r.error = r.Error;
+	return r;
+}
+
+// A stock otbr-agent has no status method, and that cannot change without the
+// daemon restarting -- so the overview's 2 s poll must not keep paying for a
+// call that will never succeed. Latch only on "Method not found": a missing
+// *object* means otbr-agent is down, which is temporary and has to stay
+// retryable.
+let status_absent = false;
```
openwrt-ai: nit: the latch outlives the event the comment says bounds it. "That cannot change without the daemon restarting" is true of otbr-agent, but `status_absent` lives in the rpcd script's VM, not in otbr-agent — rpcd keeps one persistent `uc_vm_t` per script and re-enters it per call ([`ucode.c:58`](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/ucode.c#L58), invoked via [`uc_vm_call()` at ucode.c:458](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/ucode.c#L458)). So the ordinary way the answer *does* change — installing an openthread-br build that has the method, which restarts otbr-agent but not rpcd — leaves the latch set, and the page stays on the getter path until rpcd is restarted too.

Consequences are mild (the fallback is correct, just chattier), so this may well be an acceptable trade for the simplicity. But the comment currently reads as if the latch is scoped to the daemon's lifetime, and it isn't — worth either saying "until rpcd restarts" outright, or clearing the flag when a subsequent `otbr` call succeeds through a different method.

---

openwrt-ai (commented):
Re-review at `8c0e7f0`, against my last pass at `29af3c4`.

Scope: the base commit moved `13772da` → `a2d15f4` (#8871's content — `overview.js` and the regenerated `.pot`; reviewed there, not here). This PR's own commit was reworked in place, `29af3c4` → `8c0e7f0`: the `status` probe moved into a latching `otbr_status()` helper, and the `configured` guard picked up the PAN-id test.

- **Latch.** Correct on the point that worried me most — the discriminator really does distinguish the two cases. `ubus.error()` renders `ubus_strerror(code)` and appends `": %s"` whenever `last_error.msg` is set ([`uc_ubus_error()`, lib/ubus.c:637-668](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L637-L668)), and `uc_ubus_call()` always sets that message ([lib/ubus.c:1411-1419](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L1411-L1419)), so a method miss on a present object yields `"Method not found: Failed to invoke function 'status' on object 'otbr'"` and the 17-char prefix matches. A missing object fails earlier in `ubus_lookup_id()` and yields `"Not found: …"`, so it does not latch, as intended. Module state persists across calls, so the latch holds for rpcd's lifetime — one nit inline on how that interacts with an otbr-agent upgrade.
- **`configured` guard.** The fix I suggested last round was wrong, and it is now in the tree. `+null` is `0` in ucode, not `NaN`, so the new test does not require the PAN id after all. Details and a corrected suggestion inline; sorry for the round trip.
- **Comments.** One nit: the rewritten block above the guard carries both drafts of the same explanation.
- **CI.** All three FormalityCheck runs on `8c0e7f0` are green.

## Commit checks

- `8c0e7f0` "luci-app-openthread: read the stopped-network dataset via ubus status" — two claims in the body do not hold:
  - *"configured requires the PAN id and rejects the broadcast value, matching what the getter path has always done."* The broadcast half is implemented; the "requires the PAN id" half is not, for the `+null` reason above. Either the code or the sentence needs to change.
  - *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* `feeds/packages/net/openthread-br` carries three patches — `100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons` — and none of them touches the ubus object. The PR body says the opposite and is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"), so the method exists in neither place today. Worth correcting before this leaves draft, since it is the sentence a reader checks to find the backend consumer.

---

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
LorbusChris: Confirmed on the interpreter rather than from either reading: `+null` is `0`, and the old guard did pass on `{}`. Fixed by testing the property as suggested, and checked the cases -- `null`, `{}`, a real NaN from `hex()`, `0xffff` and a valid id now all come out right. The fallback below is indeed safe already, since it tests `raw != null` separately.

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
LorbusChris: Merged, keeping the `hex()`/NaN detail. Also noted that `+null` is `0` in ucode, since that is the reason the test is on the property rather than on the coercion -- otherwise the next reader is likely to simplify it back.

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
LorbusChris: Reworded rather than cleared. The comment now says the latch lives in rpcd per-script VM and so outlives an otbr-agent restart: a build that gains the method stays on the getter path until rpcd restarts too, which is correct and only chattier. That is the price of not re-probing forever, and it is better stated than hidden.

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
```diff
@@ -0,0 +1,625 @@
...
+// leaves this on the getter path until rpcd is restarted too. The fallback is
+// correct there, only chattier, which is the price of not re-probing forever.
+let status_absent = false;
+
+function otbr_status() {
+	if (status_absent)
+		return null;
+	let r = ubus.call('otbr', 'status', {});
+	if (r == null) {
+		if (substr(ubus.error() ?? '', 0, 17) == 'Method not found:')
```
openwrt-ai: The latch keys off the English rendering of an external error message when the same call hands over the code. `ubus.error()` builds its string as `ubus_strerror(last_error.code)` plus `": %s"` ([`uc_ubus_error()`, lib/ubus.c:636-668](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L636-L668)), so `0, 17` is really a coupling to libubus's `__ubus_strerror` wording *and* to code 4 not taking the `STATUS_UNKNOWN_ERROR` branch just above. Both hold today — I checked, this works — but neither is part of any contract, and a reader has to reconstruct the whole chain to see why 17 is the right number.

`ubus.error(true)` returns the status code instead, and the module exports the constant ([`ADD_CONST(STATUS_METHOD_NOT_FOUND)`, lib/ubus.c:3954](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L3954)). That is also the in-tree idiom: [`uci.uc:26-30`](https://github.com/openwrt/luci/blob/52e94095a21eb834748d37f2915cf15b304b40be/modules/luci-base/ucode/controller/admin/uci.uc#L26-L30) and [`index.uc:69-73`](https://github.com/openwrt/luci/blob/52e94095a21eb834748d37f2915cf15b304b40be/modules/luci-base/ucode/controller/admin/index.uc#L69-L73) both discard the previous error with `ubus.error()` and read the new one with `ubus.error(true)`; the string form is used nowhere else in luci.

```suggestion
		if (ubus.error(true) == require('ubus').STATUS_METHOD_NOT_FOUND)
```

`require()` returns the cached module scope after the first call, so this is an object lookup per invocation. If you would rather not have it inside the function, keeping a module handle at line 5 (`let ubusmod = require('ubus'); let ubus = ubusmod.connect();`) reads the same.

---

openwrt-ai (commented):
Re-review at `52e9409`, against my last pass at `8c0e7f0`.

Scope: the base commit moved `a2d15f4` → `2ae1f9d` (#8871's content — hoisting the mac-filter Clear/Add buttons out of `macRows` so `applyMacReadability()` can gate them, and the `macModeChanged()` guard on the applied filter mode; reviewed there, not here). This PR's own commit was reworked in place, `8c0e7f0` → `52e9409`: one line of logic plus both comment blocks.

- **`configured` guard.** Fixed correctly. `ds?.panid != null ? +ds.panid : NaN` tests the property instead of the coercion, so the `+null == 0` trap from last round is closed: absent `ds`, a truthy `{}` from a TLV stream with no PAN ID TLV, and the `hex()`-NaN garbage case all now yield `NaN` and fall through to the getters, while `0xffff` is still rejected. With that in the tree, the commit message's *"configured requires the PAN id and rejects the broadcast value"* is finally accurate.
- **Latch comment.** The new second paragraph is right on the mechanism — rpcd keeps one `uc_vm_t` per script and re-enters it per call ([`rpc_ucode_script_t.vm`, ucode.c:56-62](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/ucode.c#L56-L62), invoked at [ucode.c:458](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/ucode.c#L458)) — so "until rpcd is restarted too" is the correct bound. Stating it rather than clearing the flag is a reasonable call.
- **Duplicated comment.** Merged, and the added note about `+null` being `0` is the right thing to leave behind.
- **Stale-error check (new, for the latch).** Every null-returning path in `uc_ubus_call()` goes through `err_return()`, including the `ubus_lookup_id()` miss ([lib/ubus.c:1397-1419](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L1397-L1419)), so the error read in `otbr_status()` is always the one from its own call — no risk of latching on a leftover from an earlier `otbr()` failure. And `uc_ubus_error()` clears on read, so nothing leaks the other way either.
- **New this round.** One inline comment on how the latch discriminates the two failures — the behaviour is correct, the coupling is to libubus's message text rather than to its status code.
- **CI.** All three FormalityCheck runs on `52e9409` are green.

## Commit checks

- `52e9409` "luci-app-openthread: read the stopped-network dataset via ubus status" — the message is byte-identical to `8c0e7f0`'s, so the second claim from last round is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* The package pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/2c72717c16c73df9cf05fc93ce21165452200169/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — `100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons` — none of which touches the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method exists in neither tree today. Worth correcting before this leaves draft, since it is the sentence a reader follows to find the backend consumer.

---

applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread
LorbusChris: Taken -- checked on a device first, since the distinction is the whole point of the latch: a missing method gives status 3 (`STATUS_METHOD_NOT_FOUND`) and a missing *object* gives 4, so keying on the code preserves exactly the behaviour the string match was there for, without depending on libubus wording. Kept a module handle at line 5 rather than calling `require()` per invocation. Also reworded the comment, which still said "Method not found" as though the text were the contract.

openwrt-ai (commented):
Re-review at `963dd58`, against my last pass at `52e9409`.

Scope: the base commit moved `2ae1f9d` → `55b3186` (#8871's content — `overview.js` and the regenerated `.pot`; reviewed there, not here). This PR's own commit was reworked in place, `52e9409` → `963dd58`: the module handle at lines 5-6, the latch discriminator, and the comment above it. Eleven lines, no other file touched.

- **Numeric discriminator.** Correct, and your device measurement matches the source. `uc_ubus_call()` reaches `ubus_lookup_id()` first and returns its status verbatim on a missing object ([lib/ubus.c:1399-1403](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L1399-L1403)), then returns the invoke status on a missing method ([lib/ubus.c:1412-1419](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L1412-L1419)) — `UBUS_STATUS_METHOD_NOT_FOUND` (3) and `UBUS_STATUS_NOT_FOUND` (4) respectively, so the two cases stay separable without depending on `__ubus_strerror` wording. `ubus.error(true)` returns `ucv_int64_new(last_error.code)` unconditionally under the numeric branch ([`uc_ubus_error()`, lib/ubus.c:646-648](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L646-L648)) — it does not take the `STATUS_UNKNOWN_ERROR` special case the string branch has, which was the second coupling I flagged. And `ADD_CONST(STATUS_METHOD_NOT_FOUND)` is unconditional in the module scope ([lib/ubus.c:3954](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L3954)), so `ubusmod.STATUS_METHOD_NOT_FOUND` resolves to a real value rather than `null` — worth having checked, since a `null` there would have made every error-cleared null reply latch.
- **No false latch on an empty reply.** `ok_return()` clears the error before returning ([lib/ubus.c:445](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L445)), so a `status` method that answered with no payload gives `ubus.error(true) == null`, which does not equal the constant. It falls back that poll and retries the next one, which is the right way round.
- **Module handle.** `error` is registered on both the module scope and the connection ([lib/ubus.c:3823](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L3823) and [3838](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/ubus.c#L3838)) over the same static `last_error`, so reading it off `ubus` while taking the constant off `ubusmod` is consistent, not two different error slots.
- **Comment.** Now says "answers with a different status" rather than naming the string; accurate.
- **CI.** All three FormalityCheck runs on `963dd58` are green.

No new inline comments this round.

## Commit checks

- `963dd58` "luci-app-openthread: read the stopped-network dataset via ubus status" — message is byte-identical to `52e9409`'s, so the one item from the last two rounds is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* The package pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/fbac2e7861fb30661eef52a45223bb3a3a0d6d17/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — `100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons` — none touching the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method is in neither tree today. Since the code paths themselves now look settled, this sentence is the last thing I'd want fixed before the PR leaves draft — it is what a reviewer follows to find the backend consumer.

---

openwrt-ai (commented):
Re-review at `c80a52b`, against my last pass at `963dd58`.

Scope: nothing in this PR's own scope changed. The base commit moved `55b3186` → `c1d1578` (#8871's content — the `macReadable()`/`applyMacVisibility()` rework around an unreadable mac-filter *state* rather than just an unreadable list, the per-address Remove buttons now collected fresh in `applyMacReadability()`, and the regenerated `.pot`; reviewed there, not here). This PR's own commit was rebased onto it with no other change: `git diff 55b3186 963dd58` and `git diff c1d1578 c80a52b` produce byte-identical patches, and the commit message is byte-identical too.

- **Backend file.** Unchanged since last round, so the conclusions from the previous passes still hold as-is: the numeric latch discriminator (`ubus.error(true) == ubusmod.STATUS_METHOD_NOT_FOUND`, line 38) separating a missing method (3) from a missing object (4), the module handle at lines 5-6, the `ds?.panid != null ? +ds.panid : NaN` guard, and the merged comment block. No new inline comments.
- **Base interaction.** The base commit only touches the mac-filter modal in `overview.js`; it does not go near `state_summary`, `dataset_summary()` or the `configured` consumers, so the rebase does not disturb anything this commit depends on.
- **CI.** No checks have run on `c80a52b` yet (status `pending`, 0 checks), so there is nothing to ground against this round.

## Commit checks

- `c80a52b` "luci-app-openthread: read the stopped-network dataset via ubus status" — message unchanged from `963dd58`, so the one item from the last three rounds is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* `feeds/packages/net/openthread-br` pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/ec4ca5c917337e15a055d0550b6d409fedde033b/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — `100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons` — none touching the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method is in neither tree today. The code paths have been settled for three rounds now, so this sentence is the last thing I'd want fixed before the PR leaves draft — it is what a reviewer follows to find the backend consumer.

---

openwrt-ai (commented):
Re-review at `1198d4f`, against my last pass at `c80a52b`.

Scope: nothing in this PR's own scope changed. The base commit moved `c1d1578` → `8d40ca1` (#8871's content — the `Unknown` placeholder option on the mac-filter mode select plus the `macModeNote` em, `refreshMacList()` now refreshing `s.macfilterstate` alongside `s.addrlist`, `macModeChanged()` gaining the `s.macfilterstate != null` baseline test, the narrowed Clear-button confirm condition, and the regenerated `.pot`; reviewed there, not here). This PR's own commit was rebased onto it with no other change: `git diff c1d1578 c80a52b` and `git diff 8d40ca1 1198d4f` produce byte-identical patches, and the commit message is byte-identical too.

- **Backend file.** `root/usr/share/rpcd/ucode/luci.openthread` is untouched since last round, so the conclusions from the previous passes stand as-is: the numeric latch discriminator (`ubus.error(true) == ubusmod.STATUS_METHOD_NOT_FOUND`, line 38) separating a missing method (3) from a missing object (4), the module handle at lines 5-6, the `ds?.panid != null ? +ds.panid : NaN` guard at line 130, and the merged comment block. No new inline comments.
- **Base interaction.** The base commit's `overview.js` hunks are confined to the MAC-filter modal in `handleSettings()` — `macSelect`, `macModeNote`, `refreshMacList()`, `macReadable()`/`macModeChanged()`/`applyMacVisibility()`/`applyMacReadability()`, and the `cbiValue(_('Protocol'), …)` row. It does not touch `state_summary`, `dataset_summary()`, or any of the `st.configured` / `st.panid` / `st.channel` / `st.networkname` consumers (the Create-button gate at `overview.js:799`, the network row at `:842-844`, the placeholder at `:884-889`), so the rebase does not disturb anything this commit depends on. This commit also adds no user-visible strings, so it has no stake in the regenerated `.pot`.
- **CI.** No checks have run on `1198d4f` yet (status `pending`, 0 checks), so there is nothing to ground against this round.

## Commit checks

- `1198d4f` "luci-app-openthread: read the stopped-network dataset via ubus status" — message unchanged from `c80a52b`, so the one item from the last four rounds is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* `feeds/packages/net/openthread-br` pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/ec4ca5c917337e15a055d0550b6d409fedde033b/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — `100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons` — none touching the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method is in neither tree today. The code paths have been settled for four rounds now, so this sentence is the last thing I'd want fixed before the PR leaves draft — it is what a reviewer follows to find the backend consumer.

---

openwrt-ai (commented):
Re-review at `f90de8e`, against my last pass at `1198d4f`.

Scope: nothing in this PR's own scope changed. The base commit moved `8d40ca1` → `879e162` (#8871's content — `macfilter_add`/`macfilter_remove` gaining a `confirmed` arg and the `unconfirmed_mode` refusal when the call would also switch the filter mode on, the matching `filterRefusals` string, the `macReadable()` early-returns in the three MAC-filter click handlers, the `setTimeout(applyMacReadability, 0)` after `refreshMacList()`, and the regenerated `.pot`; reviewed there, not here). This PR's own commit was rebased onto it with no other change: `git diff 8d40ca1 1198d4f` and `git diff 879e162 f90de8e` produce byte-identical patches (only the blob index lines differ), and the commit message is byte-identical too.

- **Backend file.** `root/usr/share/rpcd/ucode/luci.openthread` is untouched in this commit's own delta since last round, so the conclusions from the previous passes stand as-is: the module handle at lines 5-6, the numeric latch discriminator (`ubus.error(true) == ubusmod.STATUS_METHOD_NOT_FOUND`) separating a missing method (3) from a missing object (4), the `ds?.panid != null ? +ds.panid : NaN` guard, and the merged comment block. No new inline comments.
- **Base interaction.** The base commit's hunks in this same file are confined to the `macfilter_add` and `macfilter_remove` method bodies in the returned object — it does not go near `otbr_status()`, `dataset_summary()`, `summarize_state()`, or any of the `st.configured` / `st.panid` / `st.channel` / `st.networkname` consumers in `overview.js`. Its `overview.js` hunks are likewise confined to the MAC-filter modal in `handleSettings()` plus the `filterRefusals` map. So the rebase does not disturb anything this commit depends on, and since this commit adds no user-visible strings it has no stake in the regenerated `.pot`.
- **All prior threads.** The seven review threads on this file are resolved, and the last two rounds added no new ones. Nothing carried over inline.
- **CI.** No checks have run on `f90de8e` yet (status `pending`, 0 checks), so there is nothing to ground against this round.

## Commit checks

- `f90de8e` "luci-app-openthread: read the stopped-network dataset via ubus status" — message unchanged from `1198d4f`, so the one item from the last five rounds is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* `feeds/packages/net/openthread-br` pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/112a7e4e9606b1421494cc02377ba5d2951262dc/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — [`100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons`](https://github.com/openwrt/packages/tree/112a7e4e9606b1421494cc02377ba5d2951262dc/net/openthread-br/patches) — none touching the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method is in neither tree today. The code paths have been settled for five rounds now, so this sentence is the last thing I'd want fixed before the PR leaves draft — it is what a reviewer follows to find the backend consumer.

---

openwrt-ai (commented):
Re-review at `d6facbf`, against my last pass at `f90de8e`.

Scope: nothing changed in this PR at all beyond a rebase onto current master. Both commits were replayed — base `879e162` → `ff20643`, this PR's own commit `f90de8e` → `d6facbf` — and both are content-identical to last round: `git diff 879e162 ff20643 -- applications/luci-app-openthread` is empty, and `git diff 879e162 f90de8e` vs `git diff ff20643 d6facbf` are byte-identical patches, blob index lines included. The commit message is byte-identical too. The only delta between the two heads is the unrelated master churn the rebase pulled in (po/ translation updates and other apps).

- **Backend file.** `root/usr/share/rpcd/ucode/luci.openthread` is unchanged, so the conclusions from the previous passes stand as-is: the module handle at lines 5-6, the numeric latch discriminator (`ubus.error(true) == ubusmod.STATUS_METHOD_NOT_FOUND`) separating a missing method (3) from a missing object (4), the `ds?.panid != null ? +ds.panid : NaN` guard, and the merged comment block. No new inline comments.
- **Base interaction.** Unlike the previous five rounds, #8871's content did not move this time — the base commit's openthread tree is identical — so there is no new base behaviour for this commit to interact with.
- **Prior threads.** All seven review threads on this file are resolved; nothing carried over inline.
- **CI.** All three FormalityCheck runs on `d6facbf` are green.

## Commit checks

- `d6facbf` "luci-app-openthread: read the stopped-network dataset via ubus status" — message unchanged from `f90de8e`, so the one item from the last six rounds is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* `feeds/packages/net/openthread-br` still pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/c11ef37d558291d272ffe6afc857afa4fd538dd4/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — [`100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons`](https://github.com/openwrt/packages/tree/c11ef37d558291d272ffe6afc857afa4fd538dd4/net/openthread-br/patches) — none touching the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method is in neither tree today. The code has now been settled for six rounds, so this sentence remains the last thing I'd want fixed before the PR leaves draft — it is what a reviewer follows to find the backend consumer.

---

openwrt-ai (commented):
Re-review at `c2feeb8`, against my last pass at `d6facbf`.

Scope: nothing in this PR's own scope changed. The base commit moved `ff20643` → `51d8328` (#8871's content — `macfilter_add`/`macfilter_remove` and `macfilter_clear` flipping their `confirmed` / `macfilter_confirmed` arg declarations from `false` to `true`, and the three `overview.js` block/allow handlers passing `true` through; reviewed there, not here). This PR's own commit was rebased onto it with no other change: `git diff ff20643 d6facbf` and `git diff 51d8328 c2feeb8` are byte-identical patches apart from the blob index lines, and the commit message is byte-identical too.

- **Backend file.** [`root/usr/share/rpcd/ucode/luci.openthread`](``https://github.com/openwrt/luci/blob/c2feeb8a82672f51dc11bb1d16a15fd3a51ff38b/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread``) is unchanged in this commit's own delta, so the conclusions from the previous passes stand as-is: the module handle at [lines 5-6](``https://github.com/openwrt/luci/blob/c2feeb8a82672f51dc11bb1d16a15fd3a51ff38b/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L5-L6),`` the numeric latch discriminator at [line 38](``https://github.com/openwrt/luci/blob/c2feeb8a82672f51dc11bb1d16a15fd3a51ff38b/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L38``) separating a missing method (3) from a missing object (4), the `ds?.panid != null ? +ds.panid : NaN` guard at [line 120](``https://github.com/openwrt/luci/blob/c2feeb8a82672f51dc11bb1d16a15fd3a51ff38b/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L120),`` and the merged comment block. No new inline comments.
- **Base interaction.** The base commit's hunks in this same file are confined to the `args` declarations of `settings_apply`, `macfilter_add`/`macfilter_remove` and `macfilter_clear` in the returned object; its `overview.js` hunks are the matching `rpc.declare` param lists and the three MAC-filter confirm handlers. Nothing goes near `otbr_status()`, `dataset_summary()`, `summarize_state()` or the `st.configured` / `st.panid` / `st.channel` / `st.networkname` consumers, so the rebase does not disturb anything this commit depends on. This commit adds no user-visible strings either, so it has no stake in the regenerated `.pot`.
- **Prior threads.** All seven review threads on this file are resolved; nothing carried over inline.
- **CI.** No checks have run on `c2feeb8` yet (status `pending`, 0 checks), so there is nothing to ground against this round.

## Commit checks

- `c2feeb8` "luci-app-openthread: read the stopped-network dataset via ubus status" — message unchanged from `d6facbf`, so the one item from the last seven rounds is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* `feeds/packages/net/openthread-br` still pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/61972dae3af03f5f8ac757b9520eb6f2ebe2eb21/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — [`100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons`](https://github.com/openwrt/packages/tree/61972dae3af03f5f8ac757b9520eb6f2ebe2eb21/net/openthread-br/patches) — none touching the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method is in neither tree today. The code has been settled for seven rounds, so this sentence remains the last thing I'd want fixed before the PR leaves draft — it is what a reviewer follows to find the backend consumer.

---

openwrt-ai (commented):
Re-review at `bb734d0`, against my last pass at `c2feeb8`.

Scope: nothing in this PR's own scope changed. The base commit moved `51d8328` → `467a4bc` (#8871's content — `blockWarning()` gaining an `activatesDenylist` argument that prepends the router-wide mode-change warning, `handleBlock()` deriving `activates` from `neigh.macfilter == 'disable'` and passing it as the `confirmed` flag instead of a hardcoded `true`, and `handleAllowlistBlock()` / `handleAllow()` dropping their `true` so a stale row cannot silently switch the filter on; reviewed there, not here). This PR's own commit was rebased onto it with no other change: `git diff 51d8328 c2feeb8` and `git diff 467a4bc bb734d0` are byte-identical patches, blob index lines included, and the commit message is byte-identical too.

- **Backend file.** [`root/usr/share/rpcd/ucode/luci.openthread`](``https://github.com/openwrt/luci/blob/bb734d017da754871a04fcb9ba266a74d2b50fc2/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread``) is unchanged in this commit's own delta, so the conclusions from the previous passes stand as-is: the module handle at [lines 5-6](``https://github.com/openwrt/luci/blob/bb734d017da754871a04fcb9ba266a74d2b50fc2/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L5-L6),`` the numeric latch discriminator at [line 38](``https://github.com/openwrt/luci/blob/bb734d017da754871a04fcb9ba266a74d2b50fc2/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L38``) separating a missing method (3) from a missing object (4), the `ds?.panid != null ? +ds.panid : NaN` guard at [line 120](``https://github.com/openwrt/luci/blob/bb734d017da754871a04fcb9ba266a74d2b50fc2/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L120),`` and the merged comment block. No new inline comments.
- **Base interaction.** The base commit's delta is confined to `overview.js` (the three MAC-filter modal handlers and `blockWarning()`) plus the regenerated `.pot`; it does not touch the ucode backend at all this round, and goes nowhere near `otbr_status()`, `dataset_summary()`, `summarize_state()`, or the `st.configured` / `st.panid` / `st.channel` / `st.networkname` consumers. So the rebase does not disturb anything this commit depends on. This commit adds no user-visible strings either, so it has no stake in the regenerated `.pot`.
- **Prior threads.** All seven review threads on this file are resolved; nothing carried over inline.
- **CI.** No checks have run on `bb734d0` yet (status `pending`, 0 checks), so there is nothing to ground against this round.

## Commit checks

- `bb734d0` "luci-app-openthread: read the stopped-network dataset via ubus status" — message unchanged from `c2feeb8`, so the one item from the last eight rounds is still standing: *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have."* `feeds/packages/net/openthread-br` still pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/715c606e57664c78dddd4382ee08a3de9324cb02/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — [`100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons`](https://github.com/openwrt/packages/tree/715c606e57664c78dddd4382ee08a3de9324cb02/net/openthread-br/patches) — none touching the ubus object. The PR body is the accurate one ("Draft until its backend prerequisite exists upstream … The ot-br-posix PR will be linked here once it is open"): the method is in neither tree today. The code has been settled for eight rounds, so this sentence remains the last thing I'd want fixed before the PR leaves draft — it is what a reviewer follows to find the backend consumer.

---

openwrt-ai (commented):
Re-review at `4326260`, against my last pass at `bb734d0`.

Scope: no code changed at all this round, and the base commit did not move either — it is still `467a4bc` (#8871's content, reviewed there). `git diff bb734d0..4326260` is empty: the tree is byte-identical. The delta is entirely in this PR's own commit message, which was amended.

- **The standing commit-check item is fixed.** The sentence I flagged for eight rounds — *"Requires the ubus status method, which OpenWrt's openthread-br package carries as a patch and stock ot-br-posix does not yet have"* — is gone, replaced by a paragraph that matches the tree: *"The status method exists in neither stock ot-br-posix nor OpenWrt's openthread-br package today. It is part of the ubus series from project-chip/matter-openwrt that is being proposed to openthread/ot-br-posix; openthread-br picks it up once that lands. Until then the fallback path is the only one that runs."* Re-verified both halves: `feeds/packages/net/openthread-br` still pins ot-br-posix at [`2f3c799`](https://github.com/openwrt/packages/blob/0521bb56793387788798e4a69b84d96766ed54e2/net/openthread-br/Makefile#L7-L8) and carries exactly three patches — [`100-fix-gcc14-build`, `101-rest-Permit-to-externally-provide-cJSON-library`, `102-openwrt-fix-LuCI-icons`](https://github.com/openwrt/packages/tree/0521bb56793387788798e4a69b84d96766ed54e2/net/openthread-br/patches) — none touching the ubus object; and ot-br-posix's own `src/openwrt/ubus/otubus.cpp` registers no `status` method. The new text also agrees with the PR body's account of where the method is being upstreamed. Nothing left to flag here.
- **Trailer block.** The amend also reordered `Assisted-By:` above `Signed-off-by:` and changed its value, which brings it in line with the base commit `467a4bc` in the same series — the two commits now match rather than diverging. `Signed-off-by` is still present and last.
- **Backend file.** [`root/usr/share/rpcd/ucode/luci.openthread`](``https://github.com/openwrt/luci/blob/4326260b9e1b45da021c0a81ee233610987413f7/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread``) is untouched, so every conclusion from the previous passes carries over unchanged: the module handle at [line 5](``https://github.com/openwrt/luci/blob/4326260b9e1b45da021c0a81ee233610987413f7/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L5),`` the numeric latch discriminator at [line 38](``https://github.com/openwrt/luci/blob/4326260b9e1b45da021c0a81ee233610987413f7/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L38``) separating a missing method (3) from a missing object (4), and the `ds?.panid != null ? +ds.panid : NaN` guard at [line 120](``https://github.com/openwrt/luci/blob/4326260b9e1b45da021c0a81ee233610987413f7/applications/luci-app-openthread/root/usr/share/rpcd/ucode/luci.openthread#L120).`` No new inline comments.
- **Prior threads.** All seven review threads on this file remain resolved; nothing carried over inline.
- **CI.** No checks have run on `4326260` yet (status `pending`, 0 checks), so there is nothing to ground against this round.

With the message corrected, I have no outstanding items on this PR — the remaining blocker is the external one the body already names (the `status` method landing in ot-br-posix and then in `openthread-br`), which is why it is still draft.

---

PR #8895, opened 2026-08-01
luci-base: dispatcher: derive session token from /dev/urandom

JuliusBairaktaris:
`randomid()` built the LuCI session token from `math.rand()`. ucode seeds that generator on first use with the wall-clock time in milliseconds and then uses libc `rand()` — `lib/math.c`:

```c
if (!ucv_boolean_get(uc_vm_registry_get(vm, "math.srand_called"))) {
    gettimeofday(&tv, NULL);
    srand((tv.tv_sec * 1000) + (tv.tv_usec / 1000));
```

Each request is served by a fresh process, so the seed is simply the millisecond the request was handled — and `randomid()` is the first consumer of the generator in that process, so no state is burned before it. The resulting token is predictable to anyone who knows the approximate time it was issued.

Two consumers: the CSRF token (`dispatcher.uc`, checked in `test_post_security()`) and the config-rollback confirm token (`controller/admin/uci.uc`).

This patch reads the bytes from `/dev/urandom` instead, and returns `null` on failure so callers fail closed rather than fall back to a weak value — `session_retrieve()` requires a string token, so a session created without one is rejected.

For contrast, rpcd already generates its session IDs from `/dev/urandom` (`session.c` `rpc_random()`); this closes the gap on LuCI's own token.

### Testing

Patched `dispatcher.uc` installed on an IPQ8074 router running OpenWrt 6.18.39, with the device's LuCI copy verified byte-identical to upstream `master` before patching. Full login round-trip over HTTPS:

```
GET  /cgi-bin/luci/               -> 403 (login form)
POST /cgi-bin/luci/ (real creds)  -> 302 + Set-Cookie sysauth_https=...
GET  /admin/status/overview       -> token present, 32 hex chars
```

Isolated check of the new `randomid()` under the shipped ucode interpreter: two successive 16-byte reads, both 32 characters, hex-only, distinct. Device restored to stock afterwards (sha256 matched `/rom`), LuCI healthy.

### DIFF
--- a/modules/luci-base/ucode/dispatcher.uc
+++ b/modules/luci-base/ucode/dispatcher.uc
@@ -1,11 +1,10 @@
 // Copyright 2022 Jo-Philipp Wich <jo@mein.io>
 // Licensed to the public under the Apache License 2.0.
 
-import { open, stat, glob, lsdir, unlink, basename } from 'fs';
+import { open, readfile, stat, glob, lsdir, unlink, basename } from 'fs';
 import { striptags, entityencode } from 'html';
 import { connect } from 'ubus';
 import { cursor } from 'uci';
-import { rand } from 'math';
 import { openlog, syslog, closelog, LOG_INFO, LOG_WARNING, LOG_AUTHPRIV } from 'log';
 
 import { hash, load_catalog, change_catalog, translate, ntranslate, getuid } from 'luci.core';
@@ -487,12 +486,12 @@ function session_retrieve(sid, allowed_users) {
 }
 
 function randomid(num_bytes) {
-	let bytes = [];
+	let data = readfile('/dev/urandom', num_bytes);
 
-	while (num_bytes-- > 0)
-		push(bytes, sprintf('%02x', rand() % 256));
+	if (length(data) != num_bytes)
+		return null;
 
-	return join('', bytes);
+	return hexenc(data);
 }
 
 function session_setup(user, pass, path) {
--- a/applications/luci-app-ddns/Makefile
+++ b/applications/luci-app-ddns/Makefile
@@ -12,7 +12,7 @@ PKG_LICENSE:=Apache-2.0
 PKG_MAINTAINER:=Ansuel Smith <ansuelsmth@gmail.com>
 
 LUCI_TITLE:=LuCI Support for Dynamic DNS Client (ddns-scripts)
-LUCI_DEPENDS:=+luci-base +ddns-scripts
+LUCI_DEPENDS:=+luci-base +ddns-scripts +ucode-mod-math
 
 include ../../luci.mk
 
--- a/modules/luci-base/Makefile
+++ b/modules/luci-base/Makefile
@@ -24,7 +24,6 @@ LUCI_DEPENDS:=\
 	+ucode-mod-log \
 	+ucode-mod-uci \
 	+ucode-mod-ubus \
-	+ucode-mod-math \
 	+ucode-mod-html \
 	+liblucihttp-ucode
 


modules/luci-base/ucode/dispatcher.uc
```diff
@@ -487,10 +486,22 @@ function session_retrieve(sid, allowed_users) {
...
+		return null;
+
 	let bytes = [];
 
-	while (num_bytes-- > 0)
-		push(bytes, sprintf('%02x', rand() % 256));
+	for (let i = 0; i < num_bytes; i++)
+		push(bytes, sprintf('%02x', ord(data, i)));
 
 	return join('', bytes);
```
openwrt-ai: The manual `open()`/`read()`/`close()` plus hex loop can be replaced by `fs.readfile()` + `hexenc()`, which is both the in-tree idiom and cheaper on entropy.

In-tree precedent: [`luci.sys.uniqueid()` at sys.lua:63-66](https://github.com/openwrt/luci/blob/49fb529d027cd71ba53ca82eefdba2bff4096ddb/modules/luci-lua-runtime/luasrc/sys.lua#L63-L66) already does exactly `fs.readfile("/dev/urandom", bytes)` + hexlify.

Entropy cost: `file.read(<int>)` goes through a fully-buffered `FILE*`, so `fread()` pulls a whole `BUFSIZ` (4096 bytes) from `/dev/urandom` to hand back 16. `readfile()` explicitly avoids that — [`lib/fs.c:3043-3044`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/fs.c#L3043-L3044) switches the stream to `_IONBF` when the limit is below `BUFSIZ`, so it issues a single 16-byte read.

No suggestion block because the fix also needs `readfile` added to the `fs` import on line 4 (`open` stays, it is still used at lines 120, 149 and 425), which is outside this hunk:

```
import { open, readfile, stat, glob, lsdir, unlink, basename } from 'fs';
```

```
function randomid(num_bytes) {
	let data = readfile('/dev/urandom', num_bytes);

	if (length(data) != num_bytes)
		return null;

	return hexenc(data);
}
```


---

modules/luci-base/ucode/dispatcher.uc
```diff
@@ -487,10 +486,22 @@ function session_retrieve(sid, allowed_users) {
...
+
+	if (!fd)
+		return null;
+
+	let data = fd.read(num_bytes);
+
+	fd.close();
+
+	if (length(data) != num_bytes)
+		return null;
```
openwrt-ai, line 492: The commit message justifies the new `null` return via `session_setup()`, where `session_retrieve()` rejects a session without a string token — that consumer does fail closed. The second consumer is less clean: in [`uci_apply()` at controller/admin/uci.uc:47-71](https://github.com/openwrt/luci/blob/49fb529d027cd71ba53ca82eefdba2bff4096ddb/modules/luci-base/ucode/controller/admin/uci.uc#L47-L71) the `uci apply` with `rollback: true` has already been issued and succeeded by the time `dispatcher.randomid(16)` is called on line 57. If it now returns `null`, `action_apply_rollback` reports an error instead of a token, the browser never calls `confirm`, and the applied config silently rolls back when the 90s timeout expires — a state the old `randomid()` could not produce.

Only reachable if `/dev/urandom` is unreadable, so probably academic, but was that path considered? Aborting before the `uci apply` (or logging via `syslog(LOG_WARNING, ...)` so the rollback is diagnosable) would make the failure visible rather than looking like a spontaneous revert.

---

openwrt-ai (commented):
Reviewed 1 new commit. The underlying analysis checks out: `randomid()` was the first consumer of ucode's `math.rand()` in a fresh per-request process, and [`lib/math.c`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/math.c) seeds it from `gettimeofday()` milliseconds on first use, so the token was time-predictable. The replacement is binary-safe — `file.read(<int>)` builds the result with `ucv_string_new_length()`, and `length()`/`ord()` operate on the stored byte length, so NUL bytes in the read data are handled correctly. Removing the now-unused `math` import is right; `rand` had no other caller in this file.

Two inline notes, neither blocking. CI is green on `49fb529`.

---

jow-:
LGTM. Curious, would that also allow us to drop the ucode-mod-math dependency?

JuliusBairaktaris:
Yes — `dispatcher.uc` was luci-base's only `math` user, so `+ucode-mod-math` can drop out of `modules/luci-base/Makefile`; the one catch is that `luci-app-ddns`'s `ddns.uc` imports `isnan` from `math` while its Makefile only declares `+luci-base`, so it would have to gain the dependency in the same commit.

jow-:
Can you add a 2nd commit moving the dependency to the ddns package?

JuliusBairaktaris:
Added as `7b81757` — `+ucode-mod-math` drops out of `luci-base` and `luci-app-ddns` declares it, which is the only other consumer left in the tree.

jow-:
Merged, thank you!

PR #8896, opened 2026-08-01
luci-base: drop init_action() and the luci.setInitAction ubus method

JuliusBairaktaris:
`init_action()` and the `luci.setInitAction` ubus method both duplicate rpcd's `rc` namespace, which is what downstream code wishing to control services should be using. `rc.init` accepts the same `enable|disable|start|stop|restart|reload` set, and is the stronger of the two:

| | `luci.setInitAction` | `rc.init` |
|---|---|---|
| how the script runs | `system()` → `/bin/sh -c "env -i /etc/init.d/<name> <action> >/dev/null"` | `execl()`, no shell |
| script must be root-owned | no | yes |
| script must not be world-writable | no | yes |

That shell is why this PR started out as an argument-validation patch. `init_action()` interpolated both of its arguments into the command line, and `basename()` keeps `name` inside `/etc/init.d` but strips directory components only — `;`, backticks, `$(`, `|` and newlines survive it — while `action` reached the shell with nothing checking it at all. Nothing untrusted got there: `setInitAction` switched on a fixed action list, and for `name` the `stat()` check required the interpolated string to exist as a file under `/etc/init.d`, an existence test rather than a validation. But the helper was exported from luci-base, so any future caller forwarding an RPC argument straight through inherited root command execution. Removing it removes that for good, rather than papering over it with a regex.

### The five in-tree users

Ported first, one commit each, so every commit in the series builds:

* **luci-app-ddns**, **luci-app-lldpd** — mechanical: `object: 'rc'`, `method: 'init'`. `rc.init` has no reply payload, so the `expect: { result: false }` unwrapping goes with it; every call site already ignores the return value.
* **luci-app-sqm** — same, plus three things around it. The call now reaches the router at all: `params` names the positional arguments of the declared method, but the declaration passed the intended *values* there (`params: [ 'sqm', 'enable' ]`) and `handleEnableSQM()` was then invoked with no arguments, so every request went out as an empty argument object and was rejected — while the UI told the user the sqm init script had just been enabled on its behalf. The notification is now chained onto the call with `reject: true`, so a refused `rc.init` reports the failure instead of the success text. And the "Enable SQM" button reached the init script a second way, through `fs.exec()` and a pair of `/etc/init.d/sqm` exec grants; it takes `rc.init` too and both grants go.
* **luci-app-upnp** — nothing calls `callInitAction()`; the view drives upnpd through uci and `luci.upnp` alone. Dropped the dead declaration together with the ACL grant that came with it, which let a session holding read access to this app run any init script under `/etc/init.d`.
* **luci-app-rustdesk-server** — its rpcd backend cannot reach for the `rc` namespace the way a view can, since it runs inside rpcd itself, so it execs the init script directly. Passing `system()` an array execs without a shell, which is precisely what `init_action()` did not do; the action is whitelisted immediately above and the script name is a literal. The array form also drops the `env -i` and the `>/dev/null` that `init_action()` wrapped around the command line, since both belong to the `system()` form being removed; rpcd's environment is the boot environment rather than anything a caller supplies, but the init script does inherit rpcd's own descriptors, where `rc.init` redirects all three before `execl()`.

`luci-app-pbr`, `luci-app-adblock-fast`, `luci-app-https-dns-proxy` and `luci-app-simple-adblock` carry `setInitAction` methods of their own on their own ubus objects and are untouched.

### ACL scope

Every `rc.init` grant in the series sits in the **write** scope. rpcd matches ubus ACLs on object and method only, with no per-argument filtering, so a read-scope `rc.init` would let a session with read-only access to one app drive any init script on the box — which is the same property the upnp commit removes on privilege grounds, and what `luci-mod-system` already does with `rc.list` under read and `rc.init` under write.

### Testing

`init_action` is gone from the module's exports and the other five are intact:

```
$ ucode -R -L 'mods/*.uc' -e 'import * as sys from "luci.sys"; print(join(" ", sort(keys(sys))), "\n")'
conntrack_list init_enabled init_index init_list process_list
```

Both rpcd backends parse, the four modified ACL files are valid JSON, and `system(["/bin/false"])` returns `1`, which is the exit status the rustdesk backend now reports.

### DIFF
--- a/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
+++ b/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
@@ -33,10 +33,10 @@ return view.extend({
 	}),
 
 	callInitAction: rpc.declare({
-		object: 'luci',
-		method: 'setInitAction',
+		object: 'rc',
+		method: 'init',
 		params: [ 'name', 'action' ],
-		expect: { result: false }
+		reject: true
 	}),
 
 	callDDnsGetStatus: rpc.declare({
@@ -187,7 +187,8 @@ return view.extend({
 
 	handleRestartDDns(m, ev) {
 		return this.callInitAction('ddns', 'restart')
-			.then(L.bind(m.render, m));
+			.then(L.bind(m.render, m))
+			.catch(function(e) { ui.addNotification(null, E('p', e.message)) });
 	},
 
 	poll_status(map, data) {
--- a/applications/luci-app-ddns/root/usr/share/rpcd/acl.d/luci-app-ddns.json
+++ b/applications/luci-app-ddns/root/usr/share/rpcd/acl.d/luci-app-ddns.json
@@ -3,8 +3,7 @@
 		"description": "Grant access to ddns procedures",
 		"read": {
 			"ubus": {
-				"luci.ddns": [ "get_services_status", "get_ddns_state", "get_env", "get_services_log" ],
-				"luci": [ "setInitAction" ]
+				"luci.ddns": [ "get_services_status", "get_ddns_state", "get_env", "get_services_log" ]
 			},
 			"file": {
 				"/usr/share/ddns/default": [ "list" ],
@@ -18,7 +17,10 @@
 			"uci": [ "ddns" ]
 		},
 		"write": {
-			"uci": [ "ddns" ]
+			"uci": [ "ddns" ],
+			"ubus": {
+				"rc": [ "init" ]
+			}
 		}
 	}
 }
--- a/applications/luci-app-rustdesk-server/root/usr/share/rpcd/ucode/rustdesk-server.uc
+++ b/applications/luci-app-rustdesk-server/root/usr/share/rpcd/ucode/rustdesk-server.uc
@@ -2,7 +2,7 @@
 'use strict';
 
 import { popen, access, readfile, unlink } from 'fs';
-import { process_list, init_enabled, init_action } from 'luci.sys';
+import { process_list, init_enabled } from 'luci.sys';
 
 const BIN_DIR = '/usr/bin';
 const KEY_DIR = '/etc/rustdesk';
@@ -109,8 +109,7 @@ const methods = {
 				};
 			}
 
-			// Use luci.sys.init_action() for service control
-			let result = init_action('rustdesk-server', action);
+			let result = system([ '/etc/init.d/rustdesk-server', action ]);
 
 			return {
 				success: (result === 0),
@@ -135,8 +134,8 @@ const methods = {
 			let key_pub = KEY_DIR + '/id_ed25519.pub';
 
 			// Step 1: Stop the service first so keys are not in use
-			// init_action is synchronous - waits for service to fully stop
-			init_action('rustdesk-server', 'stop');
+			// system() is synchronous - waits for service to fully stop
+			system([ '/etc/init.d/rustdesk-server', 'stop' ]);
 
 			// Step 2: Remove existing keys
 			let priv_deleted = safeUnlink(key_priv);
--- a/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
+++ b/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
@@ -16,10 +16,10 @@ return view.extend({
 	},
 
 	handleEnableSQM: rpc.declare({
-		object: 'luci',
-		method: 'setInitAction',
-		params: [ 'sqm', 'enable' ],
-		expect: { result: false }
+		object: 'rc',
+		method: 'init',
+		params: [ 'name', 'action' ],
+		reject: true
 	}),
 
 	load: function() {
@@ -49,10 +49,12 @@ return view.extend({
 				E('button', {
 					'class': 'btn cbi-button-active',
 					'click': ui.createHandlerFn(this, function() {
-						return fs.exec('/etc/init.d/sqm', ['enable']).then(function() {
-							return fs.exec('/etc/init.d/sqm', ['start']);
-						}).then(function() {
+						return this.handleEnableSQM('sqm', 'enable').then(L.bind(function() {
+							return this.handleEnableSQM('sqm', 'start');
+						}, this)).then(function() {
 							location.reload();
+						}).catch(function(err) {
+							ui.addNotification(null, E('p', _("Failed to enable the sqm initscript: %s").format(err.message)), 'error');
 						});
 					})
 				}, _('Enable SQM'))
@@ -79,8 +81,11 @@ return view.extend({
 		o.rmempty = false;
 		o.write = L.bind(function(section, value) {
 			if (value == "1") {
-				this.handleEnableSQM();
-				ui.addNotification(null, E('p', _("The SQM GUI has just enabled the sqm initscript on your behalf. Remember to disable the sqm initscript manually under System Startup menu in case this change was not wished for.")));
+				this.handleEnableSQM('sqm', 'enable').then(function() {
+					ui.addNotification(null, E('p', _("The SQM GUI has just enabled the sqm initscript on your behalf. Remember to disable the sqm initscript manually under System Startup menu in case this change was not wished for.")));
+				}).catch(function(err) {
+					ui.addNotification(null, E('p', _("Failed to enable the sqm initscript: %s").format(err.message)), 'error');
+				});
 			}
 
 			return uci.set("sqm", section, "enabled", value);
--- a/applications/luci-app-sqm/root/usr/share/rpcd/acl.d/luci-app-sqm.json
+++ b/applications/luci-app-sqm/root/usr/share/rpcd/acl.d/luci-app-sqm.json
@@ -4,18 +4,18 @@
 		"read": {
 			"file": {
 				"/var/run/sqm/available_qdiscs": [ "list" ],
-				"/usr/lib/sqm/*.qos.help": [ "read" ],
-				"/etc/init.d/sqm enable" : [ "exec" ],
-				"/etc/init.d/sqm start" : [ "exec" ]
+				"/usr/lib/sqm/*.qos.help": [ "read" ]
 			},
 			"uci": [ "sqm" ],
 			"ubus": {
-				"file": [ "read", "list" ],
-				"luci": [ "setInitAction" ]
+				"file": [ "read", "list" ]
 			}
 		},
 		"write": {
-			"uci": [ "sqm" ]
+			"uci": [ "sqm" ],
+			"ubus": {
+				"rc": [ "init" ]
+			}
 		}
 	}
 }
--- a/modules/luci-base/ucode/sys.uc
+++ b/modules/luci-base/ucode/sys.uc
@@ -149,12 +149,3 @@ export function init_enabled(name) {
 
 	return false;
 };
-
-export function init_action(name, action) {
-	const s = stat(`/etc/init.d/${basename(name)}`);
-
-	if (s?.type != 'file' || s?.user_exec == false)
-		return false;
-
-	return system(`env -i /etc/init.d/${basename(name)} ${action} >/dev/null`);
-};
--- a/applications/luci-app-lldpd/htdocs/luci-static/resources/view/lldpd/config.js
+++ b/applications/luci-app-lldpd/htdocs/luci-static/resources/view/lldpd/config.js
@@ -10,6 +10,7 @@
 'require form';
 'require lldpd';
 'require network';
+'require ui';
 'require uci';
 'require tools.widgets as widgets';
 
@@ -26,11 +27,14 @@ const callInitList = rpc.declare({
 });
 
 const callInitAction = rpc.declare({
-	object: 'luci',
-	method: 'setInitAction',
+	object: 'rc',
+	method: 'init',
 	params: [ 'name', 'action' ],
-	expect: { result: false }
+	reject: true
 });
+
+const reportInitFailure = (err) => ui.addNotification(null,
+	E('p', _('Failed to control the lldpd initscript: %s').format(err.message)), 'error');
 
 const usage = _('See syntax <a %s>here</a>.').format('href=https://lldpd.github.io/usage.html target="_blank"');
 
@@ -99,13 +103,13 @@ return L.view.extend({
 				// Enable and start
 				callInitAction('lldpd', 'enable').then(() => {
 					return callInitAction('lldpd', 'start');
-				});
+				}).catch(reportInitFailure);
 			}
 			else {
 				// Stop and disable
 				callInitAction('lldpd', 'stop').then(() => {
 					return callInitAction('lldpd', 'disable');
-				});
+				}).catch(reportInitFailure);
 			}
 		};
 
--- a/applications/luci-app-lldpd/root/usr/share/rpcd/acl.d/luci-app-lldpd.json
+++ b/applications/luci-app-lldpd/root/usr/share/rpcd/acl.d/luci-app-lldpd.json
@@ -34,8 +34,8 @@
 				"lldpd"
 			],
 			"ubus": {
-				"luci": [
-					"setInitAction"
+				"rc": [
+					"init"
 				]
 			}
 		}
--- a/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
+++ b/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js
@@ -6,13 +6,6 @@
 'require rpc';
 'require form';
 
-const callInitAction = rpc.declare({
-	object: 'luci',
-	method: 'setInitAction',
-	params: [ 'name', 'action' ],
-	expect: { result: false }
-});
-
 const callUpnpGetStatus = rpc.declare({
 	object: 'luci.upnp',
 	method: 'get_status',
--- a/applications/luci-app-upnp/root/usr/share/rpcd/acl.d/luci-app-upnp.json
+++ b/applications/luci-app-upnp/root/usr/share/rpcd/acl.d/luci-app-upnp.json
@@ -3,8 +3,7 @@
 		"description": "Grant access to UPnP IGD & PCP/NAT-PMP",
 		"read": {
 			"ubus": {
-				"luci.upnp": [ "get_status" ],
-				"luci": [ "setInitAction" ]
+				"luci.upnp": [ "get_status" ]
 			},
 			"uci": [ "upnpd" ]
 		},
--- a/modules/luci-base/root/usr/share/rpcd/ucode/luci
+++ b/modules/luci-base/root/usr/share/rpcd/ucode/luci
@@ -6,7 +6,7 @@
 import { stdin, access, dirname, basename, open, popen, glob, lsdir, readfile, readlink, realpath, error } from 'fs';
 import { cursor } from 'uci';
 
-import { init_list, init_index, init_enabled, init_action, conntrack_list, process_list } from 'luci.sys';
+import { init_list, init_index, init_enabled, conntrack_list, process_list } from 'luci.sys';
 import { revision, branch } from 'luci.version';
 import { statvfs, uname } from 'luci.core';
 import { connect } from 'ubus';
@@ -132,29 +132,6 @@ const methods = {
 			}
 
 			return length(scripts) ? scripts : { error: 'No such init script' };
-		}
-	},
-
-	setInitAction: {
-		args: { name: 'name', action: 'action' },
-		call: function(request) {
-			switch (request.args.action) {
-			case 'enable':
-			case 'disable':
-			case 'start':
-			case 'stop':
-			case 'restart':
-			case 'reload':
-				const rc = init_action(request.args.name, request.args.action);
-
-				if (rc === false)
-					return { error: 'No such init script' };
-
-				return { result: rc == 0 };
-
-			default:
-				return { error: 'Invalid action' };
-			}
 		}
 	},
 


modules/luci-base/ucode/sys.uc
```diff
@@ -151,6 +151,9 @@ export function init_enabled(name) {
 };
 
 export function init_action(name, action) {
+	if (!match(action, /^[a-z_][a-z0-9_-]*$/))
```
openwrt-ai: `^` and `$` are line anchors here, not string anchors, so an embedded newline passes validation — and the shell treats a newline as a command separator.

ucode compiles a regex literal *without* the `s` flag using `REG_NEWLINE` ([`types.c:1412-1413`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/types.c#L1412-L1413)); the `s` flag is what clears it ([`lexer.c:509`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lexer.c#L509)). Under `REG_NEWLINE`, POSIX (and musl/TRE) make `^` match immediately after a newline and `$` immediately before one, so `action = "start\nreboot"` — or `"\nreboot"` — matches and reaches `system()`:

```
env -i /etc/init.d/foo start
reboot >/dev/null
```

I confirmed this with `regcomp("^[a-z_][a-z0-9_-]*$", REG_EXTENDED|REG_NEWLINE)`: `start\nreboot` matches; drop `REG_NEWLINE` and it does not. The injection table in the description covers `;`, `&&`, backticks, `|` and `>` but not `\n`, which is why this slipped through.

Adding `s` turns the anchors back into string anchors; the pattern has no `.`, so nothing else about it changes.

```suggestion
	if (!match(action, /^[a-z_][a-z0-9_-]*$/s))
```

---

modules/luci-base/ucode/sys.uc
```diff
@@ -151,6 +151,9 @@ export function init_enabled(name) {
...
+		return false;
+
 	const s = stat(`/etc/init.d/${basename(name)}`);
 
 	if (s?.type != 'file' || s?.user_exec == false)
```
openwrt-ai: nit: the commit message leans on this gate ("the stat() gate rejects a name that does not resolve to an executable file there"), but the executable half of it is dead code. `fs.stat()` only exposes `user_exec` under `perm` — `ucv_object_add(o, "user_exec", ...)` is added to the `perm` object, never to the top level. So `s?.user_exec` is always `null`, `null == false` is false, and a non-executable file in `/etc/init.d` still gets handed to `system()`. The two sibling helpers in this same file get it right: [`init_list()` sys.uc:122](https://github.com/openwrt/luci/blob/10db4e179d20292ac587b9f01265c759b0775f62/modules/luci-base/ucode/sys.uc#L122) and [`init_enabled()` sys.uc:146](https://github.com/openwrt/luci/blob/10db4e179d20292ac587b9f01265c759b0775f62/modules/luci-base/ucode/sys.uc#L146) both use `perm?.user_exec`.

Pre-existing rather than introduced here, but it is the very gate this patch is described as complementing, so it seems worth folding in (or at least not citing as effective):

```suggestion
	if (s?.type != 'file' || s?.perm?.user_exec == false)
```

---

modules/luci-base/ucode/sys.uc
```diff
@@ -151,6 +151,9 @@ export function init_enabled(name) {
 };
 
 export function init_action(name, action) {
```
openwrt-ai: nit: the PR description says "the `luci.setInitAction` ubus method that several ACL files still reference no longer exists in the tree" — it does still exist, at [`usr/share/rpcd/ucode/luci:138-158`](https://github.com/openwrt/luci/blob/10db4e179d20292ac587b9f01265c759b0775f62/modules/luci-base/root/usr/share/rpcd/ucode/luci#L138-L158), and it calls `init_action(request.args.name, request.args.action)` with ubus-supplied arguments. It is reachable from at least [luci-app-sqm](``https://github.com/openwrt/luci/blob/10db4e179d20292ac587b9f01265c759b0775f62/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js#L20``) and [luci-app-upnp](``https://github.com/openwrt/luci/blob/10db4e179d20292ac587b9f01265c759b0775f62/applications/luci-app-upnp/htdocs/luci-static/resources/view/upnp/upnp.js#L11``) with the matching ACLs.

The conclusion still holds — that method `switch`es on a fixed `enable|disable|start|stop|restart|reload` list, so nothing untrusted reaches the interpolation today — but the reasoning in the description is wrong about why, and the commit message body repeats the "no in-tree caller forwards an unvalidated action" framing without mentioning that there *is* an in-tree ubus caller whose safety depends entirely on that whitelist. Worth correcting before merge, since it is the basis of the severity assessment.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the change; one functional issue with the new check and two nits inline.

---

jow-:
I think this entire method should be dropped. Downstream users wishing to control services shall use the ubus `rc` namespace.

modules/luci-base/ucode/sys.uc
```diff
@@ -151,9 +151,12 @@ export function init_enabled(name) {
...
+	if (!match(action, /^[a-z_][a-z0-9_-]*$/s))
```
openwrt-ai: nit: the rewritten commit message says "The one in-tree caller, the luci.setInitAction ubus method in modules/luci-base/root/usr/share/rpcd/ucode/luci" — there is a second in-tree ucode caller. `luci-app-rustdesk-server`'s rpcd backend imports `init_action` from `luci.sys` and calls it at [rustdesk-server.uc:113](``https://github.com/openwrt/luci/blob/f626cfd63bea6df9b1c028a9ff727678bba3c786/applications/luci-app-rustdesk-server/root/usr/share/rpcd/ucode/rustdesk-server.uc#L113)`` and [:139](``https://github.com/openwrt/luci/blob/f626cfd63bea6df9b1c028a9ff727678bba3c786/applications/luci-app-rustdesk-server/root/usr/share/rpcd/ucode/rustdesk-server.uc#L139).``

The PR description asserts the opposite — "The only in-tree caller (`luci-app-rustdesk-server`'s rpcd backend) whitelists the action ... and the `luci.setInitAction` ubus method ... no longer exists in the tree" — so description and commit message now name a different single caller each, and both callers actually exist. The conclusion is unchanged (both whitelist: [setInitAction switches on a fixed list](https://github.com/openwrt/luci/blob/f626cfd63bea6df9b1c028a9ff727678bba3c786/modules/luci-base/root/usr/share/rpcd/ucode/luci#L138-L158), rustdesk checks `index(valid_actions, action)`), but since that inventory is the stated basis for "not currently exploitable" it is worth getting right in both texts before merge.

While there: the injection table in the description still lists only the 9 metacharacter cases and no `"start\nreboot"` case, even though the newline behaviour is now what the `s` flag in this line exists for.

---

modules/luci-base/ucode/sys.uc
```diff
@@ -151,9 +151,12 @@ export function init_enabled(name) {
...
+		return false;
+
 	const s = stat(`/etc/init.d/${basename(name)}`);
 
-	if (s?.type != 'file' || s?.user_exec == false)
+	if (s?.type != 'file' || s?.perm?.user_exec == false)
 		return false;
 
 	return system(`env -i /etc/init.d/${basename(name)} ${action} >/dev/null`);
```
openwrt-ai: The hardening is asymmetric: `action` now has to match a shape, but `name` still reaches `/bin/sh` raw. `basename()` only strips directory components — it does not remove `;`, backticks, `$(`, `|` or newlines, so `basename("network; reboot")` is `"network; reboot"` and the string handed to [`system()`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib.c#L3999-L4003) (`/bin/sh -c <cmdline>`) contains the metacharacters verbatim.

What actually stops that today is the `stat()` gate two lines up: the literal path `/etc/init.d/network; reboot` has to exist as a user-executable file, which it won't. So the shell never sees an injected `name` — but that is an existence check on the filesystem, not a validation of the value, and it is the same gate whose executable half was dead code until this commit. Given the patch's stated goal ("any future caller that forwards an RPC argument inherits root command execution"), is leaving `name` unvalidated deliberate, or worth the same shape check in this hunk? Something like `/^[a-zA-Z0-9_.-]+$/s` covers every name under `/etc/init.d` in the tree.

No suggestion block since the right character set for script names is your call.

---

openwrt-ai (commented):
Reviewed the force-pushed commit (`10db4e1` → `f626cfd`). Both earlier findings are addressed: the regex now carries the `s` flag, so `^`/`$` are string anchors and `"start\nreboot"` is rejected, and the `stat()` gate now reads `perm?.user_exec` like its two siblings in this file. I re-checked the accepted set against the tree — every `extra_command`/`EXTRA_COMMANDS` value in `openwrt/openwrt` and `openwrt/packages` matches `^[a-z_][a-z0-9_-]*$`, so no in-tree init script action is broken by the new check.

Two remaining points inline: one on `name` still being interpolated unvalidated, one nit on the caller inventory in the commit message and PR description.

---

JuliusBairaktaris:
@jow- Done: dropped `init_action()` and the `luci.setInitAction` ubus method, with the five in-tree users ported to `rc.init` in the preceding commits so each step of the series builds. upnp's declaration turned out to be dead and was removed rather than ported.

applications/luci-app-ddns/root/usr/share/rpcd/acl.d/luci-app-ddns.json
```diff
@@ -4,7 +4,7 @@
 		"read": {
 			"ubus": {
 				"luci.ddns": [ "get_services_status", "get_ddns_state", "get_env", "get_services_log" ],
-				"luci": [ "setInitAction" ]
+				"rc": [ "init" ]
```
openwrt-ai: `rc.init` lands in the **read** scope here. rpcd matches ubus ACLs on object+function only — there is no per-argument filtering ([`uh_foreach_matching_acl` in session.c:143-147](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/session.c#L143-L147)) — so this grants a session with *read-only* access to luci-app-ddns the ability to `enable`/`disable`/`start`/`stop` any script under `/etc/init.d`. That is the same property the upnp commit in this series removes on privilege grounds.

The other in-tree users put it in the write scope: [luci-mod-system.json:109](``https://github.com/openwrt/luci/blob/e1042ed80e6bfdb67e653314e35bb71b2b44802b/modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json#L109``) keeps `rc.list` under read and `rc.init` under write, and the lldpd commit in this very series does the same. Moving the grant into the existing `"write"` block (lines 22-24) would make the series self-consistent; same applies to luci-app-sqm. Was keeping it under read deliberate (i.e. preserving the exact reach of the old `luci.setInitAction` grant)?

---

applications/luci-app-sqm/root/usr/share/rpcd/acl.d/luci-app-sqm.json
```diff
@@ -11,7 +11,7 @@
 			"uci": [ "sqm" ],
 			"ubus": {
 				"file": [ "read", "list" ],
-				"luci": [ "setInitAction" ]
+				"rc": [ "init" ]
```
openwrt-ai: Same read-scope point as on the ddns ACL: this grant belongs in the `"write"` block below, since it lets a read-only sqm session drive any init script.

nit: while this file is being touched, the two `file` exec grants above ([lines 8-9](``https://github.com/openwrt/luci/blob/e1042ed80e6bfdb67e653314e35bb71b2b44802b/applications/luci-app-sqm/root/usr/share/rpcd/acl.d/luci-app-sqm.json#L8-L9``)) — `"/etc/init.d/sqm enable"` and `"/etc/init.d/sqm start"` with `exec` — are the pre-rpc leftovers of the same feature; nothing in `sqm.js` calls `fs.exec()` any more. They are a second, still-open path to running the sqm init script from the read scope.

---

applications/luci-app-rustdesk-server/root/usr/share/rpcd/ucode/rustdesk-server.uc
```diff
@@ -109,8 +109,7 @@ const methods = {
 				};
 			}
 
-			// Use luci.sys.init_action() for service control
-			let result = init_action('rustdesk-server', action);
+			let result = system([ '/etc/init.d/rustdesk-server', action ]);
```
openwrt-ai, line 112: nit: the array form does drop the shell (verified: `uc_system()` only prepends `/bin/sh -c` for the string form), but it also silently drops two things the old command line had — `env -i` and `>/dev/null`. The init script now runs with rpcd's full environment and with rpcd's stdin/stdout/stderr attached, whereas `rc.init`, the model for the rest of this series, explicitly redirects all three to `/dev/null` before `execl()` ([rc.c:373-385](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/rc.c#L373-L385)). Since rpcd-mod-ucode runs this script in-process, that is rpcd's own descriptors being inherited by whatever the init script spawns. Worth either matching `rc.init` here or noting the deliberate change in the commit message, which currently only discusses the shell. Same applies to line 138.

---

applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
```diff
@@ -79,7 +78,7 @@ return view.extend({
 		o.rmempty = false;
 		o.write = L.bind(function(section, value) {
 			if (value == "1") {
-				this.handleEnableSQM();
+				this.handleEnableSQM('sqm', 'enable');
```
openwrt-ai: The request now carries arguments, but the notification on the next line is still unconditional and the promise is discarded. `rpc.declare()` without `reject: true` resolves with the ubus status code instead of raising ([rpc.js:92-97](https://github.com/openwrt/luci/blob/e1042ed80e6bfdb67e653314e35bb71b2b44802b/modules/luci-base/htdocs/luci-static/resources/rpc.js#L92-L97)), so a rejected `rc.init` — e.g. `UBUS_STATUS_PERMISSION_DENIED` when `/etc/init.d/sqm` is not root-owned or is world-writable, which `rc_check_script()` enforces and `setInitAction` did not — still leaves the user reading "The SQM GUI has just enabled the sqm initscript on your behalf". Given the commit message calls out exactly that mismatch, chaining the notification onto the promise (and adding `reject: true` or checking the resolved status) would close it fully.

---

openwrt-ai (commented):
Reviewed the rewritten series (`f626cfd` → `e1042ed`, 6 new commits); the previous single validation commit is gone, so this is a fresh review of the whole PR.

The direction matches what was asked for, and the mechanical parts check out against rpcd: `rc.init`'s blobmsg policy really does name its arguments `name` and `action`, it validates the script name, whitelists the same six actions, applies `rc_check_script()` (root-owned, not world-writable) and `execl()`s without a shell — so every `rpc.declare()` in this series lines up with the callee. `init_action` is gone from `sys.uc` with no now-unused imports left behind, the `luci` rpcd script no longer imports it, no `luci.setInitAction` references or ACL grants remain in-tree (the pbr / adblock-fast / https-dns-proxy ones are their own objects, and are already scoped to `write`), the four ACL files parse, and `rc` is built into the rpcd binary itself so no new package dependency is implied.

Four points inline: two on the read/write ACL scope of the new `rc.init` grants, one on what the rustdesk exec drops relative to both `init_action()` and `rc.init`, one on the sqm notification still being unconditional.

## Commit checks
- `fed9a3d` / `7f33c1b` "luci-app-ddns/lldpd: control … through the rc ubus namespace" — both bodies end with "Both call sites already ignore the return value", but ddns has three `callInitAction()` calls across two handlers and lldpd has four inside a single `o.write`; the sentence reads as copy-paste between the two commits. The code changes themselves match their messages.

---

applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js
```diff
@@ -33,10 +33,9 @@ return view.extend({
 	}),
 
 	callInitAction: rpc.declare({
-		object: 'luci',
-		method: 'setInitAction',
-		params: [ 'name', 'action' ],
-		expect: { result: false }
+		object: 'rc',
+		method: 'init',
+		params: [ 'name', 'action' ]
```
openwrt-ai: A failed `rc.init` cannot reach the `.catch()` handlers below. `rpc.declare()` stores the `reject` option as `req.raise` ([rpc.js:332](https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/modules/luci-base/htdocs/luci-static/resources/rpc.js#L332)), and `handleCallReply()` only turns a non-zero ubus status into a thrown error when `req.raise` is set ([rpc.js:93-95](https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/modules/luci-base/htdocs/luci-static/resources/rpc.js#L93-L95)); otherwise it resolves with the status code itself. So the `.catch(function(e) { ui.addNotification(...) })` in [`handleToggleDDns()` at overview.js:184](``https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L184)`` never fires — a `PERMISSION_DENIED` from `rc_check_script()` (the root-owned / not-world-writable check `setInitAction` never had) resolves as `6`, `m.render()` runs, and the Start/Stop DDNS button silently does nothing.

That is the same mismatch the sqm commit in this series calls out and fixes with `reject: true`; ddns and lldpd kept the old non-raising form, so the failure modes `rc.init` newly introduces stay invisible here.

```suggestion
		object: 'rc',
		method: 'init',
		params: [ 'name', 'action' ],
		reject: true
```

Two follow-ups if you take this: [`handleRestartDDns()` at overview.js:187-190](``https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L187-L190)`` has no `.catch()` at all and would need one, and [`luci-app-lldpd`'s `callInitAction` at config.js:28-32](``https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/applications/luci-app-lldpd/htdocs/luci-static/resources/view/lldpd/config.js#L28-L32)`` has the identical declaration with no error handling on either branch of [`o.write` at config.js:94-110](``https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/applications/luci-app-lldpd/htdocs/luci-static/resources/view/lldpd/config.js#L94-L110).`` Was leaving those two alone deliberate (keeping the ported commits mechanical), or worth the same treatment?

---

applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
```diff
@@ -49,9 +49,9 @@ return view.extend({
...
 					'class': 'btn cbi-button-active',
 					'click': ui.createHandlerFn(this, function() {
-						return fs.exec('/etc/init.d/sqm', ['enable']).then(function() {
-							return fs.exec('/etc/init.d/sqm', ['start']);
-						}).then(function() {
+						return this.handleEnableSQM('sqm', 'enable').then(L.bind(function() {
+							return this.handleEnableSQM('sqm', 'start');
+						}, this)).then(function() {
 							location.reload();
 						});
```
openwrt-ai, line 58: This is the other `handleEnableSQM()` call path, and now that the declaration carries `reject: true` it fails silently. [`ui.createHandlerFn()` at ui.js:5680-5693](https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/modules/luci-base/htdocs/luci-static/resources/ui.js#L5680-L5693) wraps the returned promise in `Promise.resolve(...).finally(...)` with no `.catch()`, so a rejected `rc.init` here only clears the spinner: no `location.reload()`, no notification, just an unhandled rejection in the console. The flag handler 30 lines down now reports exactly this case, so the two paths disagree.

```suggestion
						return this.handleEnableSQM('sqm', 'enable').then(L.bind(function() {
							return this.handleEnableSQM('sqm', 'start');
						}, this)).then(function() {
							location.reload();
						}).catch(function(err) {
							ui.addNotification(null, E('p', _("Failed to enable the sqm initscript: %s").format(err.message)), 'error');
						});
```

---

openwrt-ai (commented):
Reviewed the force-push (`e1042ed` → `5f41c48`); the series was rebased but only three files changed content: the ddns and sqm ACLs and `sqm.js`.

All four points from the last round are addressed. `rc.init` now sits in the **write** block in both `luci-app-ddns.json` and `luci-app-sqm.json`, matching lldpd and [`luci-mod-system.json`](``https://github.com/openwrt/luci/blob/5f41c4828f5462c2534147c4f77a04c4722a93b1/modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json#L108-L111).`` The two dead `/etc/init.d/sqm enable|start` exec grants are gone, and the "Enable SQM" button now goes through `rc.init` as well — so the sqm commit message's "it takes rc.init too, and both grants go" is finally true of the diff. The remaining `read` grants in sqm (`file: [ "read", "list" ]` plus the two `/usr/lib/sqm` / `/var/run/sqm` paths) are still needed by `fs.read()`/`fs.list()` in `load()`, and no `fs.exec()` call remains in `sqm.js`. Both files parse as JSON. The flag's notification is chained onto the call with `reject: true`, and `reject` is the right option name — `rpc.declare()` maps it to `req.raise`, which is the flag `handleCallReply()` checks before raising on a non-zero ubus status.

Two inline points, both about the error paths that `reject: true` opens up: the sqm button path still swallows the rejection, and the ddns/lldpd declarations were not given the same treatment, leaving their `.catch()` handlers unreachable.

---

applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js
```diff
@@ -49,10 +49,12 @@ return view.extend({
...
-						return fs.exec('/etc/init.d/sqm', ['enable']).then(function() {
-							return fs.exec('/etc/init.d/sqm', ['start']);
-						}).then(function() {
+						return this.handleEnableSQM('sqm', 'enable').then(L.bind(function() {
+							return this.handleEnableSQM('sqm', 'start');
+						}, this)).then(function() {
 							location.reload();
+						}).catch(function(err) {
+							ui.addNotification(null, E('p', _("Failed to enable the sqm initscript: %s").format(err.message)), 'error');
 						});
```
openwrt-ai, line 58: nit: this handler is narrower than its text suggests, on two counts.

`rc.init` forks and defers the request, then completes it with a hardcoded `UBUS_STATUS_OK` in [`rc_init_cb()` at rc.c:309-316](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/rc.c#L309-L316) — the child's exit status arrives as `stat` and is discarded. So the only things that can reach this `.catch()` are the pre-`exec()` refusals: `NOT_FOUND`/`PERMISSION_DENIED` from [`rc_check_script()` at rc.c:74-85](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/rc.c#L74-L85), `INVALID_ARGUMENT` for a bad name or action, and fork/alloc failure. `/etc/init.d/sqm enable` running and exiting non-zero still resolves, and the user sees the reload.

That is a real narrowing relative to what was there before: `luci.setInitAction` returned `{ result: rc == 0 }` from `system()`'s exit status. Every call site in this series discarded it (`expect: { result: false }` and no check), so nothing regresses — but the PR description's comparison table presents `rc.init` as strictly stronger and does not mention this dimension, and the notification wording implies coverage the call cannot give. Worth a line in the description.

Second, on this path the chain is `enable` then `start`, so a refused `start` also prints "Failed to enable the sqm initscript". The flag handler 30 lines down only issues `enable`, where the same string is accurate.

Same first point applies to [`reportInitFailure` in luci-app-lldpd's config.js:36-37](``https://github.com/openwrt/luci/blob/2a083b1bad0c987a71f42e5ccff333a090bef6db/applications/luci-app-lldpd/htdocs/luci-static/resources/view/lldpd/config.js#L36-L37)`` and to the two `.catch()` handlers in [ddns overview.js:180-191](``https://github.com/openwrt/luci/blob/2a083b1bad0c987a71f42e5ccff333a090bef6db/applications/luci-app-ddns/htdocs/luci-static/resources/view/ddns/overview.js#L180-L191).``

---

openwrt-ai (commented):
Reviewed the force-push (`5f41c48` → `2a083b1`). The series was rebased onto `3fb9db3` and only three files changed content: ddns `overview.js`, lldpd `config.js` and sqm `sqm.js` — the ACL files, the upnp commit, the rustdesk backend and the luci-base removal are byte-identical to the last round.

Both points from the last review are addressed. `callInitAction` in ddns and lldpd now carries `reject: true`, which is what makes their `.catch()` handlers reachable — `rpc.declare()` maps it to `req.raise`, and [`handleCallReply()` at rpc.js:92-95](https://github.com/openwrt/luci/blob/2a083b1bad0c987a71f42e5ccff333a090bef6db/modules/luci-base/htdocs/luci-static/resources/rpc.js#L92-L95) only raises on a non-zero ubus status when that flag is set; the `L.raise()` there sits outside the `try` but is invoked from inside a `.then()`, so it rejects the call promise rather than throwing synchronously. `handleRestartDDns()` gained the `.catch()` its siblings have, lldpd gained `'require ui'` and a shared `reportInitFailure` on both branches of `o.write`, and the sqm "Enable SQM" button now reports instead of only clearing the spinner. I walked every call site in the three files: all six `rc.init` chains now terminate in a `.catch()`, so `reject: true` leaves no unhandled rejection behind. `E` needs no `'require dom'` — it is a luci-base global, and both files already use it elsewhere.

Also re-checked the parts the rebase carried over: no `init_action` / `luci.setInitAction` reference remains outside luci-app-pbr, luci-app-adblock-fast and luci-app-https-dns-proxy, which carry the method on their own ubus objects; `basename` stays used in both `sys.uc` and the `luci` rpcd script after the removals, so no import goes stale; and the ddns read-scope `/etc/init.d/ddns` exec grant is still needed by `handleReloadDDnsRule()`/`handleStopDDnsRule()`, which pass a section argument `rc.init` cannot express.

All six commit messages match their diffs. One nit inline, on how much the new failure reporting can actually observe.

---

PR #8900, opened 2026-08-02
luci-plugin-2fa: stop trusting X-Forwarded-For

JuliusBairaktaris:
[`get_client_ip()`](https://github.com/openwrt/luci/blob/fe9d6b4f09f3bc30a3f62abf00ac5542a3a98f90/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L563-L580) falls back to the leftmost `X-Forwarded-For` element whenever `REMOTE_ADDR` is loopback, and the value it returns decides two security outcomes in `check()`: whether the address is whitelisted, which skips 2FA entirely, and which bucket the login rate limiter charges.

The fallback only fires behind a reverse proxy, which is what it was written for — but that is exactly the deployment where the leftmost element is the one the *client* sent, since a proxy appends rather than prepends. So:

* `X-Forwarded-For: <whitelisted address>` returns `{ required: false, whitelisted: true }` and no OTP is ever requested. Needs `ip_whitelist_enabled=1`, which is off by default.
* A fresh header value per request gives every attempt its own rate-limit bucket, so `rate_limit_max_attempts` never trips and the six digits can be brute forced.
* The same header pointed at somebody else's address charges the failures to them and locks that address out instead.

Nothing else in LuCI reads `X-Forwarded-For`, and the plugin has no trusted-proxy configuration that could tell a proxy-supplied header apart from a forged one, so this drops the fallback and uses `REMOTE_ADDR` alone.

The fallback was loopback-gated, so only an on-box proxy is affected by the removal; there every client collapses onto `127.0.0.1` and the whitelist stops matching, which fails towards asking for the second factor. A proxy on a separate host is unchanged — `REMOTE_ADDR` is already the proxy's LAN address there, so a whitelist entry or LAN subnet covering it still exempts everyone behind it. Telling those clients apart needs a trusted-proxy setting the plugin does not have. The rate limiter would have shared one bucket across all of them, so it is now keyed on username *and* address, the way luci-base's own verify limiter is — one account's failures no longer lock out everybody behind the proxy. The limiter also runs when `REMOTE_ADDR` is empty, which the old address-keyed code skipped entirely.

@tokisaki-galaxy — your plugin, tagging you for review.

### DIFF
--- a/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
+++ b/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
@@ -298,6 +298,13 @@ function is_local_subnet(ip) {
 	return false;
 }
 
+// Bucket failed attempts per account, not per address: behind a reverse proxy
+// every client shares one REMOTE_ADDR, and an address-only bucket would let any
+// one of them lock out everybody else. Mirrors luci-base's verify limiter.
+function rate_limit_key(user, ip) {
+	return `${user || '?'}|${ip || '?'}`;
+}
+
 // Load rate limit state
 function load_rate_limit_state() {
 	let content = readfile(RATE_LIMIT_FILE);
@@ -322,19 +329,19 @@ function cleanup_rate_limit_state(state, now, window, lockout) {
 	let original_entries = 0;
 	let cleaned_entries = 0;
 
-	for (let ip, ip_state in state) {
+	for (let key, bucket in state) {
 		original_entries++;
 
-		if (type(ip_state) != 'object') {
+		if (type(bucket) != 'object') {
 			changed = true;
 			continue;
 		}
 
-		let locked_until = int(ip_state.locked_until || 0);
+		let locked_until = int(bucket.locked_until || 0);
 		let attempts = [];
 
-		if (type(ip_state.attempts) == 'array') {
-			for (let attempt in ip_state.attempts) {
+		if (type(bucket.attempts) == 'array') {
+			for (let attempt in bucket.attempts) {
 				attempt = int(attempt);
 				if (attempt > min_attempt)
 					push(attempts, attempt);
@@ -342,7 +349,7 @@ function cleanup_rate_limit_state(state, now, window, lockout) {
 		}
 
 		if (locked_until > now || length(attempts) > 0) {
-			cleaned[ip] = { attempts, locked_until };
+			cleaned[key] = { attempts, locked_until };
 			cleaned_entries++;
 		}
 		else if (locked_until < stale_before) {
@@ -387,7 +394,7 @@ function unlock_rate_limit_state() {
 	RATE_LIMIT_LOCK_HANDLE = null;
 }
 
-function evaluate_rate_limit(ip, consume_attempt) {
+function evaluate_rate_limit(key, consume_attempt) {
 	let ctx = cursor();
 
 	let rate_limit_enabled = ctx.get('luci_plugins', PLUGIN_UUID, 'rate_limit_enabled');
@@ -409,61 +416,62 @@ function evaluate_rate_limit(ip, consume_attempt) {
 		save_rate_limit_state(state);
 	let result;
 
-	if (!state[ip]) {
-		state[ip] = { attempts: [], locked_until: 0 };
+	if (!state[key]) {
+		state[key] = { attempts: [], locked_until: 0 };
 	}
 
-	let ip_state = state[ip];
+	let bucket = state[key];
 
-	if (ip_state.locked_until > now) {
-		result = { allowed: false, remaining: 0, locked_until: ip_state.locked_until };
+	if (bucket.locked_until > now) {
+		result = { allowed: false, remaining: 0, locked_until: bucket.locked_until };
 		unlock_rate_limit_state();
 		return result;
 	}
 
 	let recent_attempts = [];
-	for (let attempt in ip_state.attempts) {
+	for (let attempt in bucket.attempts) {
 		if (attempt > (now - window))
 			push(recent_attempts, attempt);
 	}
-	ip_state.attempts = recent_attempts;
+	bucket.attempts = recent_attempts;
 
-	if (length(ip_state.attempts) >= max_attempts) {
-		ip_state.locked_until = now + lockout;
-		ip_state.attempts = [];
+	if (length(bucket.attempts) >= max_attempts) {
+		bucket.locked_until = now + lockout;
+		bucket.attempts = [];
 		save_rate_limit_state(state);
-		result = { allowed: false, remaining: 0, locked_until: ip_state.locked_until };
+		result = { allowed: false, remaining: 0, locked_until: bucket.locked_until };
 		unlock_rate_limit_state();
 		return result;
 	}
 
 	if (consume_attempt)
-		push(ip_state.attempts, now);
+		push(bucket.attempts, now);
 
 	save_rate_limit_state(state);
-	result = { allowed: true, remaining: max_attempts - length(ip_state.attempts), locked_until: 0 };
+	result = { allowed: true, remaining: max_attempts - length(bucket.attempts), locked_until: 0 };
 	unlock_rate_limit_state();
 	return result;
 }
 
 // Check rate limit
-function check_rate_limit(ip) {
-	return evaluate_rate_limit(ip, false);
+function check_rate_limit(user, ip) {
+	return evaluate_rate_limit(rate_limit_key(user, ip), false);
 }
 
 // Reserve a rate-limit attempt atomically before verification
-function consume_rate_limit_attempt(ip) {
-	return evaluate_rate_limit(ip, true);
+function consume_rate_limit_attempt(user, ip) {
+	return evaluate_rate_limit(rate_limit_key(user, ip), true);
 }
 
-// Clear rate limit for an IP
-function clear_rate_limit(ip) {
+// Clear rate limit for an account
+function clear_rate_limit(user, ip) {
 	if (!lock_rate_limit_state())
 		return;
 
+	let key = rate_limit_key(user, ip);
 	let state = load_rate_limit_state();
-	if (state[ip]) {
-		delete state[ip];
+	if (state[key]) {
+		delete state[key];
 		save_rate_limit_state(state);
 	}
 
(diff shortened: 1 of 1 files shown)


plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -564,18 +564,9 @@ function verify_otp(username, otp) {
 function get_client_ip(http) {
 	let ip = null;
 
-	if (http && http.getenv) {
+	if (http && http.getenv)
 		ip = http.getenv('REMOTE_ADDR');
```
openwrt-ai, line 578: Dropping the fallback is right — nothing else in the tree trusts the header, and [`authplugins.uc:333`](https://github.com/openwrt/luci/blob/0f90d9e12eb79fd82e16b81d2bb366d193c3ba19/modules/luci-base/ucode/authplugins.uc#L333) already uses `REMOTE_ADDR` alone — but the "counts globally … fails towards asking for the second factor" framing in the commit message understates what happens behind a proxy, and it may be worth handling before merge.

With every client collapsing to `127.0.0.1`, the plugin's rate-limit bucket is keyed on that single address ([`state[ip]` at L412-L416](``https://github.com/openwrt/luci/blob/0f90d9e12eb79fd82e16b81d2bb366d193c3ba19/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L412-L416)),`` and the gate in `check()` returns `blocked: true` rather than an OTP prompt — [L585-L596](``https://github.com/openwrt/luci/blob/0f90d9e12eb79fd82e16b81d2bb366d193c3ba19/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L585-L596).`` That gate runs *before* [`is_2fa_enabled(user)` at L598](``https://github.com/openwrt/luci/blob/0f90d9e12eb79fd82e16b81d2bb366d193c3ba19/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L598),`` so with the shipped defaults (`rate_limit_enabled=1`, `max_attempts=5`, `window=60`, `lockout=300` in `root/etc/uci-defaults/luci-app-2fa`) five bad OTPs from any one client behind the proxy block *every* user's login for five minutes, including users who have no 2FA key configured. Compare with base, which keys its verify limiter on `user|ip` ([`verify_rate_limit_key`](https://github.com/openwrt/luci/blob/0f90d9e12eb79fd82e16b81d2bb366d193c3ba19/modules/luci-base/ucode/authplugins.uc#L19-L21)) and so degrades to per-user rather than global.

Same collapse hits the strict-mode path: `is_local_subnet('127.0.0.1')` is false for a LAN-derived subnet list, so on an uncalibrated clock `strict_mode=1` now blocks proxied logins outright instead of taking the `local_subnet_bypass`.

Trading a whitelist bypass for a trivial login DoS is a defensible call, but it is a different trade than the commit message describes. Would it make sense to pair this with either a trusted-proxy opt-in (honour `X-Forwarded-For` only when `REMOTE_ADDR` matches an explicitly configured proxy address, default off) or moving the rate-limit gate below `is_2fa_enabled()` so the shared bucket can only affect users who actually have 2FA on? Either way the commit message's "both fail towards asking for the second factor" line wants adjusting, since the observable outcome is a hard block.

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline comment on the proxy-deployment fallout of dropping the fallback. Commit message matches the diff otherwise.

---

BKPepe:
ping @Tokisaki-Galaxy

plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -564,18 +572,9 @@ function verify_otp(username, otp) {
...
-		if (ip && (ip == '127.0.0.1' || ip == '::1')) {
-			let xff = http.getenv('HTTP_X_FORWARDED_FOR');
-			if (xff) {
-				let parts = split(xff, ',');
-				ip = trim(parts[0]);
-			}
-		}
-	}
-
 	return ip || '';
```
openwrt-ai, line 580: The re-keying addresses the shared-bucket lockout from the previous round — [`rate_limit_key`](``https://github.com/openwrt/luci/blob/7986840e01df7e1d3a00972959a20e6dff5b54f1/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L304-L306)`` is byte-identical to base's [`verify_rate_limit_key`](https://github.com/openwrt/luci/blob/7986840e01df7e1d3a00972959a20e6dff5b54f1/modules/luci-base/ucode/authplugins.uc#L19-L21), and since `consume_rate_limit_attempt()` only ever runs from `verify()` — i.e. after `session_setup()` has accepted the password — a bucket can only be filled by someone who already holds the credentials. That is the right shape.

One correction on the commit message, though: *"That collapses every client behind a proxy onto one address. The whitelist then simply stops matching, and strict_mode stops taking the local_subnet_bypass on an uncalibrated clock - both fail closed"* only holds for an **on-box** proxy. The removed fallback was gated on `REMOTE_ADDR` being `127.0.0.1`/`::1`, so that is the only deployment it ever fired in.

For a proxy on a **separate LAN host**, `REMOTE_ADDR` is already the proxy's LAN address today and this patch changes nothing there — and both gates then fail *open*, not closed:

* [`is_ip_whitelisted()` at L216-L245](``https://github.com/openwrt/luci/blob/7986840e01df7e1d3a00972959a20e6dff5b54f1/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L216-L245)`` accepts CIDR entries via [`ip_in_cidr`](``https://github.com/openwrt/luci/blob/7986840e01df7e1d3a00972959a20e6dff5b54f1/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L206),`` so the natural "skip 2FA on LAN" config (`ip_whitelist=192.168.1.0/24`) matches the proxy and hands `{ required: false, whitelisted: true }` to every WAN client behind it.
* Same for [the `local_subnet_bypass` at L618-L621](``https://github.com/openwrt/luci/blob/7986840e01df7e1d3a00972959a20e6dff5b54f1/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L618-L621),`` since [`get_lan_subnets()`](``https://github.com/openwrt/luci/blob/7986840e01df7e1d3a00972959a20e6dff5b54f1/plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc#L249-L253)`` derives from the LAN interface address.

That hole is pre-existing and not something this patch introduces, so no objection to landing as-is — but it is the same class of bug the patch is fixing, and the commit message currently reads as if dropping the header closed it. Worth either narrowing the wording to the loopback case, or noting the LAN-proxy variant as a known follow-up so it does not get filed as solved.

---

plugins/luci-plugin-2fa/root/usr/share/ucode/luci/plugins/auth/login/bb4ea47fcffb44ec9bb3d3673c9b4ed2.uc
```diff
@@ -298,6 +298,13 @@ function is_local_subnet(ip) {
 	return false;
 }
 
+// Bucket failed attempts per account, not per address: behind a reverse proxy
+// every client shares one REMOTE_ADDR, and an address-only bucket would let any
+// one of them lock out everybody else. Mirrors luci-base's verify limiter.
+function rate_limit_key(user, ip) {
+	return `${user || '?'}|${ip || '?'}`;
```
openwrt-ai, line 305: nit: the PR description was not updated along with the amended commit and now contradicts both it and this code. The body still says *"the rate limiter counts globally"* and *"Both fail towards asking for the second factor rather than away from it"*, while the commit message explicitly rebuts that (*"The rate limiter does not: its gate runs ahead of `is_2fa_enabled()` and returns blocked rather than an OTP prompt"*) and this hunk is the fix for it. The body also never mentions the user+address re-keying, which is now the bulk of the diff — a reader going by the description alone would not expect `evaluate_rate_limit()` and its three call sites to change at all.

---

openwrt-ai (commented):
Reviewed the amended commit (force-push `0f90d9e` → `7986840`). The rate-limiter re-keying resolves the global-lockout concern from the last round and matches luci-base's own limiter; two inline comments, both about accuracy of the commit message / PR body rather than the code.

---

openwrt-ai (commented):

The force-push (`7986840` → `a8419ec`) is a docs-only code change — the new `get_client_ip()` header comment recording that REMOTE_ADDR is used deliberately, which answers the Copilot thread above. Both points from the last round are addressed: the commit message now scopes the "fail closed" claim to the on-box/loopback case that the removed fallback actually gated on, and the new closing paragraph states plainly that a separate-host proxy is unchanged and still exempts everyone behind it via a whitelist entry or LAN subnet. That matches the code. PR body no longer contradicts the commit message either.

---

Tokisaki-Galaxy:
Thanks for the report and fix
Sorry for the late reply, I little busy the past few days.
I went through the changes and they look very reasonable trusting the leftmost X-Forwarded-For element was indeed a design flaw on my part, and re-keying the rate limiter on user+address makes sense too.

JuliusBairaktaris:
@jow- ping for merge

openwrt-ai (commented):

The force-push (`a8419ec` → `37957ab`) is a rebase onto current master (`fe9d6b4` → `8b5f528`) with no content change: `git diff a8419ec..37957ab -- plugins/luci-plugin-2fa/` is empty and the commit message is byte-identical. Master did not touch `plugins/luci-plugin-2fa/` in that range either, so the patch applies to the same base content as before. All checks on `37957ab` are green (build x86_64, eslint, FormalityCheck).

Everything from the earlier rounds still stands: the rate-limit bucket keyed on `user|ip` mirrors luci-base's `verify_rate_limit_key`, and the PR body now matches the commit message.

---

PR #8902, opened 2026-08-03
luci-mod-status: add button role on HIDE buttons

bpisch:
This change helps to identify the HIDE span elements as clickable buttons for screen readers. Before this change it was not obvious that the sections on the status panel can be hidden, and clean interfaces are key for easy screen reader navigation.

## Tested on
OpenWRT 25.12 + latest Chrome on Windows + JAWS screen reader (change is universal as it utilizes ARIA)


jow-:
Merged, thanks!

PR #8905, opened 2026-08-04
luci-app-{snmpd,nut}: add missing ldd dependency for glibc builds

micpf:
### Description

Both `luci-app-snmpd` and `luci-app-nut` invoke `/usr/bin/ldd` at runtime
to probe optional library features:

- `luci-app-snmpd` — `testlibsslv3()` in `snmpd.js` runs
  `ldd /usr/sbin/snmpd` and looks for `libssl` in its output.
- `luci-app-nut` — `nut_server.js` and `nut_monitor.js` run `ldd` on
  `upsd`/`upsmon` to detect optional linked features.

Both packages already grant exec permission on `/usr/bin/ldd` via their
respective rpcd ACL files, but neither package declares a dependency on
`ldd`.

On musl builds this is harmless because `ldd` is a shim provided by the
C library itself and is always present. On glibc builds `ldd` is a
separate package produced by the toolchain that is not pulled in
automatically. In `luci-app-nut` the probes are wrapped in
`L.resolveDefault(..., [])` and degrade to "no optional features"; in
`luci-app-snmpd` there is no fallback, so a missing `/usr/bin/ldd`
rejects the view's `load()` promise and the whole SNMP settings page
fails to render.

Add `+USE_GLIBC:ldd` to `LUCI_DEPENDS` in both Makefiles so the
dependency is enforced only where it is actually required.

### Maintainer

`luci-app-snmpd`: @ckorber
`luci-app-nut`: @danielfdickinson @systemcrash

---

## Tested on

**OpenWrt version:** internal glibc-based build (aarch64)
**Result:** With `USE_GLIBC=y`, both packages now correctly pull in
the `ldd` package; the SSLv3 and upsd/upsmon feature probes work as
expected. No change on musl builds.

---

## Checklist

- [x] I have tested the changes locally.
- [x] Signed-off-by present, matching the commit author.

### DIFF
--- a/applications/luci-app-snmpd/Makefile
+++ b/applications/luci-app-snmpd/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:= Net-SNMP LuCI interface
-LUCI_DEPENDS:=+luci-base +snmpd
+LUCI_DEPENDS:=+luci-base +snmpd +USE_GLIBC:ldd
 LUCI_DESCRIPTION:=Some common net-snmp config items. In no way is this comprehensive.
 
 PKG_MAINTAINER:= Christian Korber <ck@dev.tdt.de>
--- a/applications/luci-app-nut/Makefile
+++ b/applications/luci-app-nut/Makefile
@@ -6,7 +6,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:=Network UPS Tools Configuration
-LUCI_DEPENDS:=+luci-base +nut +nut-upsmon +nut-server +nut-upsc +nut-web-cgi +nut-upscmd
+LUCI_DEPENDS:=+luci-base +nut +nut-upsmon +nut-server +nut-upsc +nut-web-cgi +nut-upscmd +USE_GLIBC:ldd
 
 PKG_LICENSE:=Apache-2.0
 PKG_MAINTAINER:=Daniel F. Dickinson <dfdpublic@wildtechgarden.ca> \


applications/luci-app-snmpd/Makefile
```diff
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 LUCI_TITLE:= Net-SNMP LuCI interface
-LUCI_DEPENDS:=+luci-base +snmpd
+LUCI_DEPENDS:=+luci-base +snmpd +USE_GLIBC:ldd
```
openwrt-ai, line 4: nit: the commit message and the PR body describe the wrong `ldd` target for this package. Both say the probe runs on libnetsnmp (PR body: `ldd /usr/lib/libnetsnmp.so.40`; commit body: "executing /usr/bin/ldd on libnetsnmp"), but [`testlibsslv3()` in snmpd.js:11-15](``https://github.com/openwrt/luci/blob/75e6b8760e9333ac3d665c0fdf178cee56cbd57c/applications/luci-app-snmpd/htdocs/luci-static/resources/view/snmpd/snmpd.js#L11-L15)`` runs `ldd /usr/sbin/snmpd`, and the ACL correspondingly grants [`"/usr/bin/ldd /usr/sbin/snmpd"`](``https://github.com/openwrt/luci/blob/75e6b8760e9333ac3d665c0fdf178cee56cbd57c/applications/luci-app-snmpd/root/usr/share/rpcd/acl.d/luci-app-snmpd.json#L8).`` No libnetsnmp path appears anywhere in the package.

Second inaccuracy in the same commit body: "Without it the SSLv3 detection silently fails." That holds for luci-app-nut, where the call is wrapped in `L.resolveDefault(..., [])` and degrades to "no SSL", but not here — `fs.exec()` in `testlibsslv3()` has no fallback, so a missing `/usr/bin/ldd` rejects `load()` and the whole SNMPD view fails to render. Worth stating accurately, since it makes the case for this dependency stronger, not weaker.

The change itself checks out: [`Package/ldd`](https://github.com/openwrt/openwrt/blob/b9411790fc7b48a25dd6c3feb8bb9903622a723d/package/libs/toolchain/Makefile#L423-L429) exists in openwrt/openwrt with `DEPENDS:=@!USE_MUSL`, musl instead ships [`/usr/bin/ldd` as a symlink to the loader](https://github.com/openwrt/openwrt/blob/b9411790fc7b48a25dd6c3feb8bb9903622a723d/package/libs/toolchain/Makefile#L577), and `+USE_GLIBC:<pkg>` inside `LUCI_DEPENDS` already has in-tree precedent in [`libs/luci-lib-nixio/Makefile`](https://github.com/openwrt/luci/blob/75e6b8760e9333ac3d665c0fdf178cee56cbd57c/libs/luci-lib-nixio/Makefile#L10).

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit subjects match the files they touch (one Makefile each), and the `+USE_GLIBC:ldd` guard is correct — `ldd` is a real openwrt/openwrt package restricted to `@!USE_MUSL`, and `USE_GLIBC`/`USE_MUSL` are exhaustive in `toolchain/Config.in`, so the conditional covers exactly the builds that need it. One inline note on commit-message accuracy.

---

applications/luci-app-snmpd/Makefile
micpf: Good catches — thanks. Amended the snmpd commit body to state the correct target (`ldd /usr/sbin/snmpd`) and to describe the actual failure mode (no fallback, whole view fails to render). Also fixed the PR body.

danielfdickinson:
Nice catch. I only learned about this musl/glibc recently. Note that for NUT that once https://github.com/openwrt/packages/pull/29893 lands, my plan is remove the ldd check and instead check for the compile-time generated `/usr/share/nut/ssl_backend` which will indicate "openssl", "nss", or "none".

openwrt-ai (commented):

The force-push was a message-only amend — `git diff 75e6b87..88b0f13` is empty, so both Makefile hunks are unchanged from the previously reviewed revision. The prior nit is resolved: 6dff230 now names the correct probe target (`ldd /usr/sbin/snmpd`, matching `testlibsslv3()` and the ACL entry) and describes the real failure mode (no fallback, `load()` rejects, view fails to render). 88b0f13's "silently fails" wording is accurate for nut, where both probes go through `L.resolveDefault(fs.exec_direct(...), [])` and degrade to "no SSL".

CI is green on the head SHA (build/x86_64, eslint, FormalityCheck).

---

PR #8907, opened 2026-08-05
luci-app-strongswan-swanctl: rename swanctl section and update I18N

feckert:
### Description
```
luci-app-strongswan-swanctl: rename swanctl section to connection

The name 'Remote/Tunnel Configuration' is too narrow. Encryption settings
can also be configured on this page. To better reflect this, the section
is renamed to 'Connection'. This makes it clear that additional settings
related to the connection can also be configured there.
```

```
uci-app-strongswan-swanctl: sync I18n

Update I18n with the latest changes
```

### DIFF
--- a/applications/luci-app-strongswan-swanctl/root/usr/share/luci/menu.d/luci-app-strongswan-swanctl.json
+++ b/applications/luci-app-strongswan-swanctl/root/usr/share/luci/menu.d/luci-app-strongswan-swanctl.json
@@ -13,7 +13,7 @@
 		}
 	},
 	"admin/vpn/strongswan-swanctl/swanctl": {
-		"title": "Remote/Tunnel",
+		"title": "Connections",
 		"order": 10,
 		"action": {
 			"type": "view",
--- a/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
+++ b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js
@@ -66,8 +66,8 @@ return view.extend({
 		if (error)
 			ui.addNotification(null, E('p', _('Some options are unavailable because swanctl failed to load: %s').format(error)), 'warning');
 
-		m = new form.Map('ipsec', _('Remote/Tunnel configuration'),
-			_('On this page, you can configure the IPsec tunnels and remotes.'));
+		m = new form.Map('ipsec', _('Connection configurations'),
+			_('On this page, you can configure the IPsec connections.'));
 		m.tabbed = true;
 
 		// Remote Configuration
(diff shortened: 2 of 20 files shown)


applications/luci-app-strongswan-swanctl/root/usr/share/luci/menu.d/luci-app-strongswan-swanctl.json
```diff
@@ -13,7 +13,7 @@
 		}
 	},
 	"admin/vpn/strongswan-swanctl/swanctl": {
-		"title": "Remote/Tunnel",
+		"title": "Connections",
```
openwrt-ai, line 16: nit: the menu entry is plural (`Connections`) while the page heading this PR sets in the same commit is singular — `_('Connection configuration')` at [`swanctl.js:69`](``https://github.com/openwrt/luci/blob/56b1e8878abd210b3e53e7d92dbbcce39951814b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/swanctl.js#L69)`` — and the commit message says "the section is renamed to 'Connection'".

The sibling page in this app pairs them in the singular: menu [`"title": "Service"`](``https://github.com/openwrt/luci/blob/56b1e8878abd210b3e53e7d92dbbcce39951814b/applications/luci-app-strongswan-swanctl/root/usr/share/luci/menu.d/luci-app-strongswan-swanctl.json#L24)`` with [`_('Service configuration')`](``https://github.com/openwrt/luci/blob/56b1e8878abd210b3e53e7d92dbbcce39951814b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/strongswan.js#L10).`` Is the plural here intentional (menu lists many connections), or should it be `Connection` to match?

Note that the plural also makes the menu entry share the msgid `Connections` with the status page's table heading at [`status.js:335`](``https://github.com/openwrt/luci/blob/56b1e8878abd210b3e53e7d92dbbcce39951814b/applications/luci-app-strongswan-swanctl/htdocs/luci-static/resources/view/strongswan-swanctl/status.js#L335)`` — visible in the regenerated `.pot`, where both references now sit under one entry.

---

openwrt-ai (commented):
Reviewed 2 new commits; one naming nit inline, no functional issues found.

The I18n sync in 56b1e88 checks out: every `#:` line reference in the regenerated `.pot` matches the actual line in the current sources (spot-checked `swanctl.js:69/147/256/368/409/474` and the `menu.d` JSON refs), the msgids removed from the `.po` files are all genuinely gone from `htdocs/`/`root/` (`Gateway (Remote Endpoint)`, `Local IP`, `Lifetime`, `Duration of the CHILD_SA before rekeying`, …), no existing translation was silently emptied, obsolete entries are retained as `#~` per the tree's usual convention, and the `.po` headers are untouched.

---

PR #8908, opened 2026-08-06
luci-base: fix lookup of non-present menu nodes

pktpls:
### Description
This surfaced when I tried developing based purely on luci-base, without any of the status, system, or network modules. The menu entries for these modules live in luci-base, while the actions (templates/views) live in the respective modules.

The header.ut template renders links to these menu entries, which fails because lookup() expects them to be present in the menu_json structure, which they are not.
The template already anticipates that lookup() might fail, just lookup() itself didn't seem to yet.

The resulting exception:

    Reference error

    In lookup(), file /usr/share/ucode/luci/dispatcher.uc, line 762, byte 24:
      called from function main (/usr/share/ucode/luci/template/themes/bootstrap/header.ut:61:51)
      called from function call ([C])
      called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:80:36)
      called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:127:40)
      called from function [arrow function] (/usr/share/ucode/luci/runtime.uc:183:57)
      called from function main (/usr/share/ucode/luci/template/header.ut:7:34)
      called from function call ([C])
      called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:80:36)
      called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:127:40)
      called from function [arrow function] (/usr/share/ucode/luci/runtime.uc:183:57)
      called from function main (/usr/share/ucode/luci/template/view.ut:1:20)
      called from function call ([C])
      called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:80:36)
      called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:127:40)
      called from function run_action (/usr/share/ucode/luci/dispatcher.uc:815:47)
      called from function [anonymous function] (/usr/share/ucode/luci/dispatcher.uc:1107:48)
      called from anonymous function (/usr/share/ucode/walter/web/index.uc:47:15)

              node = node.children[name];
      Near here -------------------^

### Maintainer _(preferred)_
@jow- 

---

## Tested on

**LuCI version:** LuCI Master (26.212.49915~fe9d6b4)
**Web browser(s):** Firefox 153.0.1

### DIFF
--- a/modules/luci-base/ucode/dispatcher.uc
+++ b/modules/luci-base/ucode/dispatcher.uc
@@ -758,7 +758,7 @@ function lookup(...segments) {
 			push(path, name);
 
 	for (let name in path) {
-		node = node.children[name];
+		node = node.children?.[name];
 
 		if (!node)
 			return null;


modules/luci-base/ucode/dispatcher.uc
```diff
@@ -758,11 +758,11 @@ function lookup(...segments) {
 			push(path, name);
 
 	for (let name in path) {
-		node = node.children[name];
-
-		if (!node)
+		if (!exists(node.children, name))
 			return null;
 
+		node = node.children[name];
```
openwrt-ai: `exists()` reports `true` for a key that is present but holds `null` — see [`uc_exists()` in lib.c:795](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib.c#L795-L811) and its documented example `exists({ qrx: null }, 'qrx') // true`. Such an entry now passes the new guard, `node` is assigned `null`, and `node.leaf` on the next line raises the very same `Reference error: left-hand side expression is null` this commit is fixing. Dropping the `if (!node)` check also means `lookup()` can return `{ node: null, url: ... }`, which callers no longer distinguish from a hit.

The same traversal earlier in this file already handles the missing-`children` case with optional chaining — [`node = node.children?.[s];` at dispatcher.uc:642](https://github.com/openwrt/luci/blob/cb4815da643ef26ffc8cb805648595af58a44497/modules/luci-base/ucode/dispatcher.uc#L642) — which fixes the crash while keeping the original falsy check and avoiding the double lookup:

```suggestion
		node = node.children?.[name];

		if (!node)
			return null;
```


---

openwrt-ai (commented):
Reviewed 1 new commit; one inline comment on the guard rewrite in `lookup()`.

---

openwrt-ai (commented):

The force-pushed version replaces the `exists()`-based guard with `node = node.children?.[name];` while keeping the original `if (!node) return null;` check, matching the existing traversal style in [`resolve_page()` at dispatcher.uc:641](https://github.com/openwrt/luci/blob/c10b5a8ba5d7d19674ba11a53a63167996a3e5a6/modules/luci-base/ucode/dispatcher.uc#L641). That resolves the earlier concern about `null`-valued keys slipping through and about `lookup()` returning `{ node: null, ... }`. Commit message matches the diff, and all three FormalityCheck runs pass on the head commit.

---

jow-:
Merged, thanks!

PR #8913, opened 2026-08-06
luci-app-51sdwan: add web interface

21hkcloud:
This adds a JavaScript LuCI view for the 51SDWAN OpenWrt edge agent.
It provides account enrollment, connection state, routing mode,
traffic statistics, connect/disconnect and logout actions with scoped
rpcd ACL permissions.

The router-side agent is proposed in:
https://github.com/openwrt/packages/pull/30176

Runtime-tested together with the agent on OpenWrt 25.12.5 x86_64.

### DIFF
--- a/applications/luci-app-51sdwan/Makefile
+++ b/applications/luci-app-51sdwan/Makefile
@@ -1,0 +1,17 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (C) 2026 Jia Liu <support@51sdwan.com>
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-app-51sdwan
+LUCI_TITLE:=LuCI support for 51SDWAN
+LUCI_URL:=https://www.51sdwan.com/en/openwrt
+LUCI_DESCRIPTION:=Web interface for the 51SDWAN OpenWrt edge agent.
+LUCI_DEPENDS:=+luci-base +51sdwan
+LUCI_PKGARCH:=all
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_MAINTAINER:=Jia Liu <support@51sdwan.com>
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js
+++ b/applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js
@@ -1,0 +1,274 @@
+'use strict';
+// SPDX-License-Identifier: GPL-2.0-or-later
+'require view';
+'require rpc';
+'require ui';
+'require poll';
+
+var callStatus = rpc.declare({
+	object: '51sdwan',
+	method: 'status',
+	expect: { '': {} }
+});
+
+var callLoginStart = rpc.declare({
+	object: '51sdwan',
+	method: 'login_start',
+	params: [ 'account', 'password' ],
+	expect: { '': {} }
+});
+
+var callLoginStatus = rpc.declare({
+	object: '51sdwan',
+	method: 'login_status',
+	params: [ 'job_id' ],
+	expect: { '': {} }
+});
+
+var callConnect = rpc.declare({ object: '51sdwan', method: 'connect', expect: { '': {} } });
+var callDisconnect = rpc.declare({ object: '51sdwan', method: 'disconnect', expect: { '': {} } });
+var callLogout = rpc.declare({ object: '51sdwan', method: 'logout', expect: { '': {} } });
+var callSetMode = rpc.declare({
+	object: '51sdwan',
+	method: 'set_mode',
+	params: [ 'mode' ],
+	expect: { '': {} }
+});
+
+function safeText(value, fallback) {
+	if (value === null || value === undefined || value === '')
+		return String(fallback === undefined ? '-' : fallback);
+	return String(value);
+}
+
+function bytes(value) {
+	return '%1024mB'.format(Number(value || 0));
+}
+
+function formatTime(epoch) {
+	if (!epoch)
+		return _('No handshake yet');
+	return new Date(Number(epoch) * 1000).toLocaleString();
+}
+
+return view.extend({
+	load: function() {
+		return callStatus();
+	},
+
+	request: function(promise, message) {
+		ui.showModal(_('51SDWAN'), [
+			E('p', { 'class': 'spinning' }, [ message || _('Processing, please wait…') ])
+		]);
+		return promise.then(function(result) {
+			ui.hideModal();
+			if (!result || result.ok === false)
+				throw new Error(safeText(result && result.error, _('Operation failed')));
+			ui.addNotification(null, E('p', {}, [ _('Operation completed') ]), 'info');
+			window.setTimeout(function() { window.location.reload(); }, 350);
+		}).catch(function(error) {
+			ui.hideModal();
+			ui.addNotification(null, E('p', {}, [ safeText(error && error.message, error) ]), 'error');
+		});
+	},
+
+	waitForLogin: function(jobId) {
+		return new Promise(function(resolve, reject) {
+			var deadline = Date.now() + 180000;
+			var check = function() {
+				callLoginStatus(jobId).then(function(result) {
+					if (!result || result.ok === false)
+						reject(new Error(safeText(result && result.error, _('Login failed'))));
+					else if (result.pending === true) {
+						if (Date.now() >= deadline)
+							reject(new Error(_('Login timed out; check the network and try again')));
+						else
+							window.setTimeout(check, 1500);
+					}
+					else
+						resolve(result);
+				}).catch(reject);
+			};
+			window.setTimeout(check, 800);
+		});
+	},
+
+	renderLogin: function() {
+		var account = E('input', {
+			'class': 'cbi-input-text sdwan-input',
+			'type': 'text',
+			'placeholder': _('Phone number or email address'),
+			'autocomplete': 'username'
+		});
+		var password = E('input', {
+			'class': 'cbi-input-password sdwan-input',
+			'type': 'password',
+			'placeholder': _('Member account password'),
+			'autocomplete': 'current-password'
+		});
+		var button = E('button', {
+			'class': 'btn cbi-button cbi-button-action sdwan-primary',
+			'click': ui.createHandlerFn(this, function() {
+				var user = account.value.trim();
+				if (!user || !password.value) {
+					ui.addNotification(null, E('p', {}, [ _('Enter the member account and password') ]), 'warning');
+					return;
+				}
+				var secret = password.value;
+				var self = this;
+				password.value = '';
+				var login = callLoginStart(user, secret).then(function(result) {
+					secret = '';
+					if (!result || result.ok === false)
+						throw new Error(safeText(result && result.error, _('Login failed')));
+					return result.pending === true ? self.waitForLogin(result.job_id) : result;
+				});
+				return this.request(login, _('Verifying the account and connecting to the gateway…'));
+			})
+		}, [ _('Log in and connect') ]);
+
+		return E('div', { 'class': 'sdwan-login-card' }, [
+			E('div', { 'class': 'sdwan-login-copy' }, [
+				E('span', { 'class': 'sdwan-eyebrow' }, [ _('MEMBER ACCESS') ]),
+				E('h2', {}, [ _('Log in to 51SDWAN') ]),
+				E('p', {}, [ _('Use an existing member account. The router is enrolled as a subscription device and receives its assigned gateway automatically.') ])
+			]),
+			E('div', { 'class': 'sdwan-login-form' }, [
+				E('label', {}, [ _('Member account') ]), account,
+				E('label', {}, [ _('Password') ]), password,
+				button,
+				E('small', {}, [ _('The password is used only for this login request and is not stored on the router.') ])
+			])
+		]);
+	},
+
+	renderDashboard: function(data) {
+		var connected = data.connected === true;
+		var mode = data.mode || 'smart';
+		var nodes = this.statusNodes = {};
+		var modeSelect = E('select', {
+			'class': 'cbi-input-select',
+			'change': ui.createHandlerFn(this, function(ev) {
+				return this.request(callSetMode(ev.target.value), _('Switching acceleration mode…'));
+			})
+		}, [
+			E('option', { 'value': 'smart', 'selected': mode === 'smart' ? '' : null }, [ _('Smart acceleration (domestic routes direct)') ]),
+			E('option', { 'value': 'global', 'selected': mode === 'global' ? '' : null }, [ _('Global acceleration (all IPv4 traffic)') ])
+		]);
+		var action = E('button', {
+			'class': 'btn cbi-button ' + (connected ? 'cbi-button-negative' : 'cbi-button-action') + ' sdwan-connect',
+			'click': ui.createHandlerFn(this, function() {
+				return this.request(
+					connected ? callDisconnect() : callConnect(),
+					connected ? _('Disconnecting safely…') : _('Establishing the secure tunnel…')
+				);
+			})
+		}, [ connected ? _('Disconnect') : _('Start acceleration') ]);
+
+		var logout = E('button', {
+			'class': 'btn cbi-button sdwan-logout',
+			'click': ui.createHandlerFn(this, function() {
+				return ui.showModal(_('Log out of 51SDWAN'), [
+					E('p', {}, [ _('Logging out stops acceleration but does not delete the device record from the member portal.') ]),
+					E('div', { 'class': 'right' }, [
+						E('button', { 'class': 'btn', 'click': ui.hideModal }, [ _('Cancel') ]),
+						E('button', {
+							'class': 'btn cbi-button-negative',
+							'click': ui.createHandlerFn(this, function() {
+								ui.hideModal();
+								return this.request(callLogout(), _('Logging out…'));
+							})
+						}, [ _('Confirm logout') ])
+					])
+				]);
+			})
+		}, [ _('Log out') ]);
+
+		nodes.gateway = E('b', {}, [ safeText(data.gateway) ]);
+		nodes.tunnel = E('b', {}, [ safeText(data.tunnel_ip) ]);
+		nodes.handshake = E('b', {}, [ formatTime(data.latest_handshake) ]);
+		nodes.traffic = E('b', {}, [ _('Down %s / Up %s').format(bytes(data.rx_bytes), bytes(data.tx_bytes)) ]);
+		nodes.account = E('div', { 'class': 'cbi-value-field' }, [ safeText(data.account) ]);
+		nodes.version = E('div', { 'class': 'cbi-value-field' }, [ safeText(data.version) ]);
+
+		return E('div', {}, [
+			E('section', { 'class': 'sdwan-status-card ' + (connected ? 'is-connected' : '') }, [
+				E('div', { 'class': 'sdwan-status-top' }, [
+					E('span', { 'class': 'sdwan-state-dot' }),
+					E('span', {}, [ connected ? _('Secure tunnel established') : _('Device logged in and ready to connect') ])
+				]),
+				E('div', { 'class': 'sdwan-status-body' }, [
+					E('div', {}, [
+						E('h2', {}, [ connected ? _('Accelerating') : _('Not connected') ]),
+						E('p', {}, connected
+							? [ safeText(data.gateway, _('51SDWAN gateway')), ' · ', safeText(data.tunnel_ip) ]
+							: [ _('Start acceleration to route LAN devices through the assigned gateway.') ])
+					]), action
+				])
+			]),
+			data.error ? E('div', { 'class': 'alert-message warning' }, [ safeText(data.error) ]) : '',
+			E('div', { 'class': 'sdwan-grid' }, [
+				E('section', { 'class': 'cbi-section sdwan-panel' }, [
+					E('h3', {}, [ _('Connection settings') ]),
+					E('div', { 'class': 'cbi-value' }, [ E('label', { 'class': 'cbi-value-title' }, [ _('Acceleration mode') ]), E('div', { 'class': 'cbi-value-field' }, [ modeSelect ]) ]),
+					E('div', { 'class': 'cbi-value' }, [ E('label', { 'class': 'cbi-value-title' }, [ _('Member account') ]), nodes.account ]),
+					E('div', { 'class': 'cbi-value' }, [ E('label', { 'class': 'cbi-value-title' }, [ _('Client version') ]), nodes.version ])
+				]),
+				E('section', { 'class': 'cbi-section sdwan-panel' }, [
+					E('h3', {}, [ _('Tunnel status') ]),
+					E('div', { 'class': 'sdwan-stat-row' }, [ E('span', {}, [ _('Gateway') ]), nodes.gateway ]),
+					E('div', { 'class': 'sdwan-stat-row' }, [ E('span', {}, [ _('Tunnel address') ]), nodes.tunnel ]),
+					E('div', { 'class': 'sdwan-stat-row' }, [ E('span', {}, [ _('Latest handshake') ]), nodes.handshake ]),
+					E('div', { 'class': 'sdwan-stat-row' }, [ E('span', {}, [ _('Tunnel traffic') ]), nodes.traffic ])
+				])
+			]),
+			E('div', { 'class': 'sdwan-footer-actions' }, [ logout ])
+		]);
+	},
+
+	updateDashboard: function(data) {
+		var nodes = this.statusNodes;
+		if (!nodes)
+			return;
+		nodes.gateway.textContent = safeText(data.gateway);
+		nodes.tunnel.textContent = safeText(data.tunnel_ip);
+		nodes.handshake.textContent = formatTime(data.latest_handshake);
+		nodes.traffic.textContent = _('Down %s / Up %s').format(bytes(data.rx_bytes), bytes(data.tx_bytes));
+		nodes.account.textContent = safeText(data.account);
+		nodes.version.textContent = safeText(data.version);
+	},
+
+	render: function(data) {
+		var self = this;
+		this.lastStatus = data || {};
+		var page = E('div', { 'class': 'cbi-map sdwan-page' }, [
+			E('style', {}, [ '\
+.sdwan-page{max-width:1120px;color-scheme:light dark}.sdwan-hero{margin-bottom:18px;padding:24px 28px;border-radius:18px;background:linear-gradient(125deg,#062f43,#087f84);color:#fff;display:flex;align-items:center;gap:16px}.sdwan-logo{width:52px;height:52px;border-radius:14px;background:linear-gradient(135deg,#18d7c0,#3b9dff);display:grid;place-items:center;font-size:21px;font-weight:800}.sdwan-hero h1{margin:0 0 4px;color:#fff}.sdwan-hero p{margin:0;color:#b5dce1}.sdwan-eyebrow{font-size:11px;letter-spacing:2px;color:#13bda9}.sdwan-login-card{display:grid;grid-template-columns:1fr 1fr;gap:32px;padding:28px;border:1px solid var(--border-color-medium,#ddd);border-radius:16px;background:var(--background-color-high,#fff);color:var(--text-color-highest,#000)}.sdwan-login-copy{padding:18px}.sdwan-login-copy h2{font-size:26px;margin:10px 0}.sdwan-login-copy p{color:var(--text-color-medium,#808080);line-height:1.7}.sdwan-login-form{display:grid;gap:9px;padding:22px;border-radius:14px;background:var(--background-color-medium,#f9f9f9)}.sdwan-login-form label{font-weight:600}.sdwan-input{box-sizing:border-box;width:100%;min-height:44px}.sdwan-primary{margin-top:8px;min-height:44px}.sdwan-login-form small{color:var(--text-color-medium,#808080);text-align:center}.sdwan-status-card{padding:22px 26px;border-radius:18px;background:linear-gradient(125deg,#082f43,#0a6570);color:#fff;box-shadow:0 8px 24px rgba(4,45,64,.16)}.sdwan-status-card.is-connected{background:linear-gradient(125deg,#06364b,#078b83)}.sdwan-status-top{display:flex;align-items:center;gap:8px;color:#bcebe5;font-size:13px}.sdwan-state-dot{width:9px;height:9px;border-radius:50%;background:#ffbd50}.is-connected .sdwan-state-dot{background:#18d7c0;box-shadow:0 0 0 5px rgba(24,215,192,.15)}.sdwan-status-body{display:flex;align-items:center;justify-content:space-between;gap:20px;margin-top:18px}.sdwan-status-body h2{margin:0 0 6px;color:#fff;font-size:27px}.sdwan-status-body p{margin:0;color:#b9dadd}.sdwan-connect{min-width:150px;min-height:44px}.sdwan-grid{display:grid;grid-template-columns:1fr 1fr;gap:16px;margin-top:18px}.sdwan-panel{margin:0!important;border:1px solid var(--border-color-medium,#ddd);border-radius:14px;background:var(--background-color-high,#fff);box-shadow:none}.sdwan-panel h3{margin-top:0}.sdwan-stat-row{display:flex;justify-content:space-between;gap:16px;padding:10px 0;border-bottom:1px solid var(--border-color-low,#eee)}.sdwan-stat-row:last-child{border:0}.sdwan-stat-row span{color:var(--text-color-medium,#808080)}.sdwan-stat-row b{text-align:right}.sdwan-footer-actions{display:flex;justify-content:flex-end;margin-top:12px}.sdwan-logout{color:var(--text-color-medium,#808080)}@media(max-width:700px){.sdwan-login-card,.sdwan-grid{grid-template-columns:1fr}.sdwan-login-copy{padding:4px}.sdwan-status-body{align-items:stretch;flex-direction:column}.sdwan-connect{width:100%}}\
+' ]),
+			E('section', { 'class': 'sdwan-hero' }, [
+				E('div', { 'class': 'sdwan-logo' }, [ '51' ]),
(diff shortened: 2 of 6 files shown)


applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
```diff
@@ -0,0 +1,235 @@
...
+function bytes(value) {
+	var number = Number(value || 0);
+	if (number < 1024) return number + ' B';
+	if (number < 1024 * 1024) return (number / 1024).toFixed(1) + ' KB';
+	if (number < 1024 * 1024 * 1024) return (number / 1024 / 1024).toFixed(1) + ' MB';
+	return (number / 1024 / 1024 / 1024).toFixed(2) + ' GB';
+}
+
+function formatTime(epoch) {
+	if (!epoch) return '尚未握手';
```
openwrt-ai: Every user-facing string in this view is a hardcoded Chinese literal that is never passed through `_()`, and the package ships no `po/` directory. All 101 other `applications/luci-app-*` packages have a `po/` tree, and no in-tree LuCI resource file contains CJK UI text — LuCI strings are authored in English and translated via gettext (`luci.mk` picks up `po/` automatically). As written this page cannot be translated and is unreadable outside a Chinese locale.

The same change is needed for the literals at lines 57, 61, 62, 79, 95, 101, 109, 118, 121, 123, 128, 129, 132, 133, 135, 146, 149, 150, 155, 157, 162, 163, 165, 172, 176, 182, 186, 187, 194, 196, 197, 200, 201, 202, 203, 204 and 218, plus the `B`/`KB`/`MB`/`GB` unit suffixes in `bytes()` at lines 40-43.

```suggestion
	if (!epoch) return _('No handshake yet');
```

---

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
```diff
@@ -0,0 +1,235 @@
...
+					E('span', {}, connected ? '安全通道已建立' : '设备已登录，等待连接')
+				]),
+				E('div', { 'class': 'sdwan-status-body' }, [
+					E('div', {}, [
+						E('h2', {}, connected ? '正在加速' : '尚未连接'),
+						E('p', {}, connected ? ((data.gateway || '51SDWAN 入口') + ' · ' + (data.tunnel_ip || '')) : '点击按钮即可让局域网设备接入专线。')
+					]), action
+				])
+			]),
+			data.error ? E('div', { 'class': 'alert-message warning' }, data.error) : '',
```
openwrt-ai: `data.error` comes straight from the `51sdwan` ubus object and is handed to `E()` as a bare string child. [`dom.append()` at luci.js:1394-1396](https://github.com/openwrt/luci/blob/827340fe60f3707b7262107d3832569b4fa36faa/modules/luci-base/htdocs/luci-static/resources/luci.js#L1394-L1396) assigns a non-array, non-node child via `node.innerHTML`, so any markup in the agent's reply is parsed as HTML rather than shown as text. Wrapping the value in an array takes the `createTextNode()` branch instead.

```suggestion
			data.error ? E('div', { 'class': 'alert-message warning' }, [ data.error ]) : '',
```

Same pattern applies to the other agent-supplied values: line 66 (`error.message`, which carries `result.error` from the agent), line 187 (`data.gateway` / `data.tunnel_ip`), line 196 (`data.account`), line 197 (`data.version`), and lines 201-202 (`data.gateway` / `data.tunnel_ip` again).

---

applications/luci-app-51sdwan/root/usr/share/rpcd/acl.d/luci-app-51sdwan.json
```diff
@@ -0,0 +1,15 @@
...
+  "luci-app-51sdwan": {
+    "description": "Access the local 51SDWAN edge agent",
+    "read": {
+      "ubus": {
+        "51sdwan": [ "status", "login_status" ]
+      }
+    },
+    "write": {
+      "ubus": {
+        "51sdwan": [ "login", "login_start", "connect", "disconnect", "set_mode", "logout" ]
```
openwrt-ai: `login` is granted write access but the view never declares or calls it — the only login entry points used are `login_start` (declared at lines 14-19 of `51sdwan.js`) and `login_status`. The PR description says the ACL is "scoped", so this extra grant looks like a leftover; dropping it keeps the ACL matched to what the frontend actually invokes.

```suggestion
        "51sdwan": [ "login_start", "connect", "disconnect", "set_mode", "logout" ]
```

---

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
```diff
@@ -0,0 +1,235 @@
...
...ems:center;gap:8px;color:#bcebe5;font-size:13px}.sdwan-state-dot{width:9px;height:9px;border-radius:50%;background:#ffbd50}.is-connected .sdwan-state-dot{background:#18d7c0;box-shadow:0 0 0 5px rgba(24,215,192,.15)}.sdwan-status-body{display:flex;align-items:center;justify-content:space-between;gap:20px;margin-top:18px}.sdwan-status-body h2{margin:0 0 6px;color:#fff;font-size:27px}.sdwan-status-body p{margin:0;color:#b9dadd}.sdwan-connect{min-width:150px;min-height:44px}.sdwan-grid{display:grid;grid-template-columns:1fr 1fr;gap:16px;margin-top:18px}.sdwan-panel{margin:0!important;border:1px solid #dfe9ec;border-radius:14px;box-shadow:none}.sdwan-panel h3{margin-top:0}.sdwan-stat-row{display:flex;justify-content:space-between;gap:16px;padding:10px 0;border-bottom:1px solid #e8eef0}.sdwan-stat-row:last-child{border:0}.sdwan-stat-row span{color:#6e858e}.sdwan-stat-row b{text-align:right}.sdwan-footer-actions{display:flex;justify-content:flex-end;margin-top:12px}.sdwan-logout{color:#70858d}@media(max-width:700px){.sdwan-login-card,.sdwan-grid{grid-template-columns:1fr}.sdwan-login-copy{padding:4px}.sdwan-status-body{align-items:stretch;flex-direction:column}.sdwan-connect{width:100%}}\
```
openwrt-ai: The injected stylesheet hardcodes a light-only palette (`background:#fff` on `.sdwan-login-card`, `background:#f5f9fa` on `.sdwan-login-form`, `border:1px solid #dce8eb`, `color:#647d87`, `color:#758c95`, `border-bottom:1px solid #e8eef0`). LuCI themes support dark mode via `:root[data-darkmode="true"]` and expose CSS custom properties for exactly this purpose — see the token block at [`bootstrap-light/cascade.css:23-121`](``https://github.com/openwrt/luci/blob/827340fe60f3707b7262107d3832569b4fa36faa/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap-light/cascade.css#L23-L121)`` and its dark override at [line 123](``https://github.com/openwrt/luci/blob/827340fe60f3707b7262107d3832569b4fa36faa/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap-light/cascade.css#L123).`` With these literals the login card and both panels render as white boxes with light-grey text when the user has dark mode on. Using `var(--background-color-high)`, `var(--text-color-low)`, `var(--border-color-medium)` and friends makes the page follow whichever theme is active.

---

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
```diff
@@ -0,0 +1,235 @@
...
+			]),
+			data && data.configured ? this.renderDashboard(data) : this.renderLogin()
+		]);
+
+		poll.add(function() {
+			return callStatus().then(function(next) {
+				if (!!next.connected !== !!data.connected || !!next.configured !== !!data.configured || next.error !== data.error)
+					window.location.reload();
+			});
+		}, 5);
```
openwrt-ai: The poll callback only ever triggers a full `window.location.reload()`, and only when `connected`, `configured` or `error` flips. The tunnel figures the PR description advertises — `latest_handshake`, `rx_bytes`, `tx_bytes`, `gateway`, `tunnel_ip` — are rendered once from the `load()` result at lines 201-204 and then never touched, so on a stable connection the traffic counters and handshake time stay frozen at their page-load values for as long as the page is open.

Keeping references to the stat `<b>` nodes and rewriting their text from `next` inside this callback (the usual LuCI live-view pattern) would make the polling actually useful and avoid discarding the whole DOM on every state change.

---

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
```diff
@@ -0,0 +1,235 @@
...
+					]), action
+				])
+			]),
+			data.error ? E('div', { 'class': 'alert-message warning' }, data.error) : '',
+			E('div', { 'class': 'sdwan-grid' }, [
+				E('section', { 'class': 'cbi-section sdwan-panel' }, [
+					E('h3', {}, '连接设置'),
+					E('div', { 'class': 'cbi-value' }, [ E('label', { 'class': 'cbi-value-title' }, '加速模式'), E('div', { 'class': 'cbi-value-field' }, modeSelect) ]),
+					E('div', { 'class': 'cbi-value' }, [ E('label', { 'class': 'cbi-value-title' }, '会员账号'), E('div', { 'class': 'cbi-value-field' }, data.account || '-') ]),
+					E('div', { 'class': 'cbi-value' }, [ E('label', { 'class': 'cbi-value-title' }, '客户端版本'), E('div', { 'class': 'cbi-value-field' }, 'v' + (data.version || '0.1.0')) ])
```
openwrt-ai: nit: the field is labelled "client version" and `data.version` is reported by the agent, but the fallback `'0.1.0'` is this LuCI package's `PKG_VERSION` (Makefile line 12). When the agent omits `version` the UI confidently displays `v0.1.0` as though it were a real agent version, and it silently goes stale the first time either version is bumped independently. The neighbouring unknown fields (lines 196, 201, 202) fall back to `'-'`; doing the same here avoids inventing a version number.

---

applications/luci-app-51sdwan/root/usr/share/luci/menu.d/luci-app-51sdwan.json
```diff
@@ -0,0 +1,13 @@
+{
+  "admin/services/51sdwan": {
```
openwrt-ai: nit: this file and `root/usr/share/rpcd/acl.d/luci-app-51sdwan.json` use 2-space indentation. In-tree, 93 of 99 `menu.d/*.json` and 98 of 102 `acl.d/*.json` files are tab-indented — worth matching so the new package lines up with the rest of the tree.

---

applications/luci-app-51sdwan/root/usr/share/luci/menu.d/luci-app-51sdwan.json
```diff
@@ -0,0 +1,13 @@
...
+    "title": "51SDWAN",
+    "order": 35,
+    "action": {
+      "type": "view",
+      "path": "services/51sdwan"
```
openwrt-ai: nit: this puts the view at `htdocs/luci-static/resources/view/services/51sdwan.js`, but no in-tree app ships a `view/services/` directory. All 87 app-owned view subdirectories are named after the package itself (`view/cloudflared/`, `view/banip/`, `view/ddns/`, …); the few shared names in use (`view/system/`, `view/network/`, `view/firewall/`) belong to core modules. Moving the file to `view/51sdwan/51sdwan.js` and setting `"path": "51sdwan/51sdwan"` keeps the package inside its own namespace.

---

openwrt-ai (commented):
Reviewed 1 new commit (`827340f`). The commit message matches the diff.

The two things most likely to block merge are the complete absence of translation support (all UI strings are hardcoded Chinese, no `_()`, no `po/`) and agent-supplied status fields being rendered through `innerHTML`; details inline.

One note on the cross-repo dependency: the `51sdwan` package that backs the `51sdwan` ubus object does not exist in the packages feed yet, so `LUCI_DEPENDS:=+51sdwan` is unsatisfiable until openwrt/packages#30176 lands. The PR description already calls this out, so this is just a reminder about merge ordering, not a change request.

---

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
21hkcloud: Fixed in 91fbfd874d. UI source strings are now English and wrapped with _(); the package includes a scanner-generated POT plus a complete zh_Hans translation catalog.

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
21hkcloud: Fixed in 91fbfd874d. Agent-supplied strings are rendered only through text nodes/textContent; no dynamic value is assigned through innerHTML.

applications/luci-app-51sdwan/root/usr/share/rpcd/acl.d/luci-app-51sdwan.json
21hkcloud: Fixed in 91fbfd874d. The unused login permission was removed and the ACL now grants only the ubus methods invoked by the view.

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
21hkcloud: Fixed in 91fbfd874d. The view now uses LuCI theme variables for backgrounds, borders, and text, including dark-mode themes.

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
21hkcloud: Fixed in 91fbfd874d. The polling callback keeps references to the status nodes and updates handshake, traffic, gateway, tunnel IP, account, version, and errors in place.

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/services/51sdwan.js
21hkcloud: Fixed in 91fbfd874d. An absent agent version now displays -, rather than inventing a package version.

applications/luci-app-51sdwan/root/usr/share/luci/menu.d/luci-app-51sdwan.json
21hkcloud: Fixed in 91fbfd874d. Both JSON files now follow the repository tab-indentation convention.

applications/luci-app-51sdwan/root/usr/share/luci/menu.d/luci-app-51sdwan.json
21hkcloud: Fixed in 91fbfd874d. The view moved to view/51sdwan/51sdwan.js and the menu path now uses the package namespace.

21hkcloud:
The LuCI review findings have been addressed in `91fbfd874d`.

- all UI source strings are English and wrapped in `_()`;
- a scanner-generated POT and complete `zh_Hans` catalog are included;
- dynamic agent values are rendered as text nodes (no dynamic `innerHTML` path);
- the ACL is limited to the methods used by the view;
- colors use LuCI theme variables and work in dark mode;
- polling updates live status/counters in place;
- unknown versions display `-`;
- JSON formatting follows the tree convention;
- the view is namespaced under `view/51sdwan/`.

Validation performed:
- official i18n scanner output is current and `msgcmp` reports the Chinese catalog complete;
- JavaScript syntax and both JSON files validate;
- the 24.10 SDK builds the application and translation packages;
- the packages load against the updated ubus agent on OpenWrt 25.12.5;
- all current formality checks pass.

The core dependency remains tracked by openwrt/packages#30176. A re-review would be appreciated when convenient.

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js
```diff
@@ -0,0 +1,281 @@
...
...round:#ffbd50}.is-connected .sdwan-state-dot{background:#18d7c0;box-shadow:0 0 0 5px rgba(24,215,192,.15)}.sdwan-status-body{display:flex;align-items:center;justify-content:space-between;gap:20px;margin-top:18px}.sdwan-status-body h2{margin:0 0 6px;color:#fff;font-size:27px}.sdwan-status-body p{margin:0;color:#b9dadd}.sdwan-connect{min-width:150px;min-height:44px}.sdwan-grid{display:grid;grid-template-columns:1fr 1fr;gap:16px;margin-top:18px}.sdwan-panel{margin:0!important;border:1px solid var(--border-color-medium);border-radius:14px;background:var(--background-color-high);box-shadow:none}.sdwan-panel h3{margin-top:0}.sdwan-stat-row{display:flex;justify-content:space-between;gap:16px;padding:10px 0;border-bottom:1px solid var(--border-color-low)}.sdwan-stat-row:last-child{border:0}.sdwan-stat-row span{color:var(--text-color-medium)}.sdwan-stat-row b{text-align:right}.sdwan-footer-actions{display:flex;justify-content:flex-end;margin-top:12px}.sdwan-logout{color:var(--text-color-medium)}@media(max-width:700px){.sdwan-login-card,.sdwan-grid{grid-template-columns:1fr}.sdwan-login-copy{padding:4px}.sdwan-status-body{align-items:stretch;flex-direction:column}.sdwan-connect{width:100%}}\
```
openwrt-ai: The six theme tokens this stylesheet now consumes — `--background-color-high`, `--background-color-medium`, `--text-color-highest`, `--text-color-medium`, `--border-color-medium`, `--border-color-low` — are declared **only** by luci-theme-bootstrap ([`bootstrap/cascade.css:23-91`](``https://github.com/openwrt/luci/blob/91fbfd874dcf1b6203783a970a7dc7bb88e40b5b/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L23-L91)).`` `grep -rl -- "--background-color-high" themes/` matches only `luci-theme-bootstrap`; luci-theme-material, luci-theme-openwrt and luci-theme-openwrt-2020 declare no CSS custom properties at all.

An unresolved `var()` with no fallback makes the declaration invalid at computed-value time, so the whole declaration is dropped. On those three themes `.sdwan-login-card`, `.sdwan-login-form` and `.sdwan-panel` end up with no background and no border, and `.sdwan-stat-row` loses its separator — i.e. the panels lose their framing entirely rather than degrading to the previous hardcoded palette.

Both existing in-tree consumers of these tokens always supply a fallback — [`switch-vlan.css:62`](``https://github.com/openwrt/luci/blob/91fbfd874dcf1b6203783a970a7dc7bb88e40b5b/modules/luci-mod-network/htdocs/luci-static/resources/view/network/switch-vlan.css#L62)`` uses `background-color: var(--background-color-high, #fff);` and [`dashboard/custom.css:19`](``https://github.com/openwrt/luci/blob/91fbfd874dcf1b6203783a970a7dc7bb88e40b5b/modules/luci-mod-dashboard/htdocs/luci-static/resources/view/dashboard/css/custom.css#L19)`` uses `border-top: 1px solid var(--border-color-medium, rgba(0, 0, 0, 0.1));`. Doing the same here (values matching the bootstrap light theme) keeps non-bootstrap themes rendering:

- `var(--background-color-high)` → `var(--background-color-high, #fff)` (2 sites: `.sdwan-login-card`, `.sdwan-panel`)
- `var(--background-color-medium)` → `var(--background-color-medium, #f9f9f9)` (`.sdwan-login-form`)
- `var(--text-color-highest)` → `var(--text-color-highest, #000)` (`.sdwan-login-card`)
- `var(--text-color-medium)` → `var(--text-color-medium, #808080)` (4 sites: `.sdwan-login-copy p`, `.sdwan-login-form small`, `.sdwan-stat-row span`, `.sdwan-logout`)
- `var(--border-color-medium)` → `var(--border-color-medium, #ddd)` (2 sites: `.sdwan-login-card`, `.sdwan-panel`)
- `var(--border-color-low)` → `var(--border-color-low, #eee)` (`.sdwan-stat-row`)

Not suggesting a patch inline because the whole stylesheet is one 2.9 KB line and a hand-edited replacement is too easy to corrupt.

---

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js
```diff
@@ -0,0 +1,281 @@
...
+function bytes(value) {
+	var number = Number(value || 0);
+	if (number < 1024)
+		return _('%s B').format(number);
+	if (number < 1024 * 1024)
+		return _('%s KiB').format((number / 1024).toFixed(1));
+	if (number < 1024 * 1024 * 1024)
+		return _('%s MiB').format((number / 1024 / 1024).toFixed(1));
+	return _('%s GiB').format((number / 1024 / 1024 / 1024).toFixed(2));
+}
```
openwrt-ai, line 46: nit: LuCI already ships this formatter — the `%m` conversion in [`String.prototype.format` at cbi.js:819](https://github.com/openwrt/luci/blob/91fbfd874dcf1b6203783a970a7dc7bb88e40b5b/modules/luci-base/htdocs/luci-static/resources/cbi.js#L819) does exactly the same 1024-based scaling and appends the `i` infix, so `'%1024mB'.format(n)` yields `512 B`, `1.50 KiB`, `3.21 MiB`, `1.05 GiB`. 100 files in the tree use it.

The bigger cost of the private helper is that it routes the unit suffixes through gettext: `%s B`, `%s KiB`, `%s MiB` and `%s GiB` are now four msgids in `po/templates/51sdwan.pot`, and their zh_Hans msgstrs in `po/zh_Hans/51sdwan.po` are byte-identical to the msgids because SI/IEC unit symbols are not translated.

```suggestion
function bytes(value) {
	return '%1024mB'.format(Number(value || 0));
}
```

The four unit msgids should then disappear from both catalogs on the next regeneration.

---

applications/luci-app-51sdwan/po/zh_Hans/51sdwan.po
```diff
@@ -0,0 +1,162 @@
+msgid ""
+msgstr ""
+"Language: zh_Hans\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
```
openwrt-ai: nit: Chinese has no plural forms. 92 of the 94 `applications/*/po/zh_Hans/*.po` files in the tree carry `nplurals=1; plural=0;`; only two (e.g. `luci-app-acme`) carry the `nplurals=2` form used here, and those look like Weblate artefacts rather than intent. With `nplurals=2` a future `ngettext`/`N_()` string in this package would need a second, never-used msgstr and would pick the wrong index.

```suggestion
"Plural-Forms: nplurals=1; plural=0;\n"
```

Also worth noting: unlike every other in-tree catalog this file carries no `#:` source references (`grep -c '^#:'` → 0). Regenerating it from `po/templates/51sdwan.pot` with the normal msgmerge step keeps it in the shape Weblate expects when it takes the package over.

---

applications/luci-app-51sdwan/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-app-51sdwan
+LUCI_TITLE:=LuCI support for 51SDWAN
+LUCI_URL:=https://www.51sdwan.com/en/openwrt
+LUCI_DESCRIPTION:=Web interface for the 51SDWAN OpenWrt edge agent.
+LUCI_DEPENDS:=+luci-base +51sdwan
+LUCI_PKGARCH:=all
+PKG_VERSION:=0.2.0
```
openwrt-ai: nit: `luci-app-51sdwan` does not exist in the tree yet — it is created by `827340f`, the first commit of this same unmerged series. Bumping `PKG_VERSION` to `0.2.0` (and rewriting `LUCI_URL` on line 8) inside `91fbfd8` records a version history for a package that has never shipped a `0.1.0`.

Squashing `91fbfd8` into `827340f` makes both hunks disappear and leaves a single "add package" commit, which is the usual shape for a new LuCI app. Note that 87 of the 101 `applications/luci-app-*` Makefiles set no `PKG_VERSION` at all and let [`luci.mk:165`](https://github.com/openwrt/luci/blob/91fbfd874dcf1b6203783a970a7dc7bb88e40b5b/luci.mk#L165) derive it from git, so dropping the line entirely is also an option.

---

openwrt-ai (commented):
Reviewed 1 new commit (`91fbfd8`) since my last review at `827340f`. All eight points from that review are addressed: the view moved to `view/51sdwan/51sdwan.js`, every user-facing string is wrapped in `_()` with a scanner-generated POT plus a zh_Hans catalog, agent-supplied values now go through `safeText()` into array children (text nodes, not `innerHTML`), the unused `login` ACL entry is gone, both JSON files are tab-indented, the poll callback updates the stat nodes in place, and the `version` fallback is `-`. Verified the POT covers all 50 source msgids and the zh_Hans catalog has no empty msgstrs; the ACL now matches exactly the seven ubus methods the view declares.

One thing the theme-variable rework introduced: the tokens it adopted are bootstrap-only, so the panels lose their background and borders on the other three in-tree themes. Details inline, along with three nits.

## Commit checks
- `91fbfd8` "luci-app-51sdwan: address interface review feedback" — the message enumerates five changes (namespace move, translation of every string, zh_Hans catalogs, text rendering + ACL removal, theme variables + poll refresh). The diff also rewrites `LUCI_URL` from `https://www.51sdwan.com/` to `https://www.51sdwan.com/en/openwrt` and bumps `PKG_VERSION` from `0.1.0` to `0.2.0`; neither hunk is described. Both go away if the commit is squashed into `827340f` (see the Makefile comment).

---

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js
21hkcloud: Fixed in 7290079ee4. Every LuCI theme token now has the suggested light-theme fallback, so bootstrap still uses its variables while material, openwrt and openwrt-2020 retain panel backgrounds, borders and readable text.

applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js
21hkcloud: Fixed in 7290079ee4. The private byte-unit formatter was replaced with LuCI's built-in %1024mB formatter, and the four unit-only msgids were removed from both catalogs.

applications/luci-app-51sdwan/po/zh_Hans/51sdwan.po
21hkcloud: Fixed in 7290079ee4. zh_Hans now declares nplurals=1; plural=0;, and the catalog was regenerated with source references. The POT and PO contain the same 48 messages with no empty Chinese translations.

applications/luci-app-51sdwan/Makefile
21hkcloud: Fixed in 7290079ee4. PKG_VERSION was dropped and the complete application history was squashed into one add-package commit rebased on current upstream master. The final commit message describes the complete application rather than an unpublished version transition.

21hkcloud:
Follow-up review items are addressed in `7290079ee4` and the PR is now a single add-package commit rebased on current master. Theme variables have cross-theme fallbacks, byte formatting uses LuCI `%1024mB`, the zh_Hans plural rule/source references are corrected, unit-only translations are removed, and `PKG_VERSION` is no longer set. Local JavaScript syntax, JSON parsing, catalog parity/completeness, source-reference and `git diff --check` validations pass.

applications/luci-app-51sdwan/po/templates/51sdwan.pot
```diff
@@ -0,0 +1,199 @@
...
+#: applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js:185
+msgid "Log out"
+msgstr ""
+
+#: applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js:171
+msgid "Log out of 51SDWAN"
+msgstr ""
+
+#: applications/luci-app-51sdwan/htdocs/luci-static/resources/view/51sdwan/51sdwan.js:172
+msgid "Logging out stops acceleration but does not delete the device record from the member portal."
```
openwrt-ai: nit: the three long msgids lost their 80-column wrapping in this revision. `build/i18n-scan.pl` passes `--no-wrap` to xgettext but then pipes everything through [`msguniq -s` at i18n-scan.pl:236](https://github.com/openwrt/luci/blob/7290079ee453a45e097133cd5e19cb49925c3105/build/i18n-scan.pl#L236), which rewraps — so canonical scanner output is wrapped. That matches what `91fbfd8` had here, and no in-tree POT has an unwrapped long msgid (`grep -hE '^msgid ' applications/*/po/templates/*.pot | awk 'length>80'` matches nothing across all 101 apps). As it stands, regenerating this catalog with the in-tree scanner produces a diff.

```suggestion
msgid ""
"Logging out stops acceleration but does not delete the device record from "
"the member portal."
```

Same applies to lines 178 and 194 of this file, and to lines 100, 182 and 198 of `po/zh_Hans/51sdwan.po`.

---

applications/luci-app-51sdwan/Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-app-51sdwan
+LUCI_TITLE:=LuCI support for 51SDWAN
+LUCI_URL:=https://www.51sdwan.com/en/openwrt
+LUCI_DESCRIPTION:=Web interface for the 51SDWAN OpenWrt edge agent.
+LUCI_DEPENDS:=+luci-base +51sdwan
+LUCI_PKGARCH:=all
+PKG_RELEASE:=1
```
openwrt-ai: nit: dropping `PKG_VERSION` left this line dead. [`luci.mk:165`](https://github.com/openwrt/luci/blob/7290079ee453a45e097133cd5e19cb49925c3105/luci.mk#L165) only folds `PKG_RELEASE` into `VERSION` when `PKG_VERSION` is set — with it unset the package version comes from `PKG_SRC_VERSION`, and [`luci.mk:115`](https://github.com/openwrt/luci/blob/7290079ee453a45e097133cd5e19cb49925c3105/luci.mk#L115) already defaults `PKG_RELEASE` to `1` regardless. 86 of the 87 `applications/luci-app-*` Makefiles that omit `PKG_VERSION` also omit `PKG_RELEASE`.

```suggestion
LUCI_PKGARCH:=all
```

---

applications/luci-app-51sdwan/Makefile
```diff
@@ -0,0 +1,19 @@
...
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=LICENSE
```
openwrt-ai: nit: none of the 101 `applications/luci-app-*` packages sets `PKG_LICENSE_FILES`, and none ships its own `LICENSE` file — the repo-wide [`LICENSE`](https://github.com/openwrt/luci/blob/7290079ee453a45e097133cd5e19cb49925c3105/LICENSE) covers the tree, and `PKG_LICENSE:=GPL-2.0-or-later` on the line above already records the per-package license for the buildroot. Dropping this line together with `applications/luci-app-51sdwan/LICENSE` would put the new package in line with every other app; is the extra copy deliberate?

---

openwrt-ai (commented):
Reviewed the squashed head `7290079` against my last review at `91fbfd8`. The series is now a single add-package commit rebased on master, and its message matches the diff.

All four items from the previous round are fixed, and I verified each: every one of the six theme tokens now carries the light-theme fallback (`--background-color-high,#fff`, `--background-color-medium,#f9f9f9`, `--text-color-highest,#000`, `--text-color-medium,#808080`, `--border-color-medium,#ddd`, `--border-color-low,#eee`), so material/openwrt/openwrt-2020 keep their panel framing; `bytes()` is now `'%1024mB'.format(...)` and the four unit-only msgids are gone from both catalogs; zh_Hans declares `nplurals=1; plural=0;` and carries `#:` references whose line numbers match the current `51sdwan.js`; POT and PO hold the same 49 msgids in the same order with no empty Chinese translations; `PKG_VERSION` is gone. FormalityCheck is green on `7290079`.

Nothing blocking left — three nits inline (POT line wrapping, a now-dead `PKG_RELEASE`, and the per-package `LICENSE`).

The `51sdwan` package backing the ubus object is still absent from the packages feed (`packages@0aba856`), so `LUCI_DEPENDS:=+51sdwan` stays unsatisfiable until openwrt/packages#30176 lands — merge-ordering note only, already covered in the PR description.

---

21hkcloud:
Addressed the three remaining review nits in `2a2cbdd226`:

- restored canonical gettext wrapping for the long POT/PO msgids;
- removed the redundant `PKG_RELEASE` assignment;
- removed the per-package `PKG_LICENSE_FILES` entry and duplicate `LICENSE` file.

Local `git diff --check`, JavaScript syntax, JSON parsing and long-msgid wrapping checks pass.

openwrt-ai (commented):

Diffed `7290079..2a2cbdd` — the amended head touches only four files and fixes exactly the three nits from my last review:

- the three long msgids in `po/templates/51sdwan.pot` and `po/zh_Hans/51sdwan.po` are now wrapped in canonical gettext form (continuation lines under 78 columns, trailing space kept on the split), matching how the rest of the tree's catalogs wrap long strings;
- `PKG_RELEASE:=1` is gone — `luci.mk` already defaults it via `PKG_RELEASE?=1`;
- `PKG_LICENSE_FILES:=LICENSE` and the per-package `LICENSE` file are removed; no other `applications/luci-app-*/Makefile` sets `PKG_LICENSE_FILES`, and `PKG_LICENSE:=GPL-2.0-or-later` plus the Makefile SPDX header still carry the licensing.

Re-verified after the change: POT and PO still hold the same 49 msgids in the same order, no empty `msgstr`, and the `#:` line references are untouched (the view file did not change in this delta). The commit message matches the squashed add-package diff. All three FormalityCheck runs are green on `2a2cbdd`.

The `51sdwan` package backing the ubus object is still absent from the packages feed (`packages@112a7e4`), so `LUCI_DEPENDS:=+51sdwan` remains unsatisfiable until openwrt/packages#30176 lands — merge-ordering note only, already covered in the PR description.

---

PR #8915, opened 2026-08-06
luci-mod-network: use device-specific MTU limits

dormancygrace:
## Description

Use the per-device `min_mtu` and `max_mtu` values exposed by Linux instead of validating every network device against the hardcoded 576 to 9200 range.

The values are collected with one rtnetlink link dump per `getNetworkDevices` call, returned by `luci-rpc`, preserved in the shared network device model, and used for form validation. Devices or kernels which do not expose the attributes keep the existing fallback limits. The kernel maximum is also applied to the IPv6 MTU field. The field is omitted on link-layer devices which cannot carry IPv6's 1280-byte minimum.

The general MTU field retains the previous safe 576-byte floor when the kernel reports a lower link-layer minimum.

The backend, model and form changes are split into their respective package commits. The `rpcd-mod-luci` version is bumped so existing installations pull the updated RPC module.

## Testing

- `git diff --check`
- ESLint with Node.js 22 on the changed JavaScript files
- Cross-compiled `rpcd-mod-luci` for `ramips/mt7620` (mipsel_24kc)

### DIFF
--- a/libs/rpcd-mod-luci/src/luci.c
+++ b/libs/rpcd-mod-luci/src/luci.c
@@ -644,10 +644,138 @@ lease_next(void)
 }
 
 
+struct network_device_mtu_limit {
+	struct network_device_mtu_limit *next;
+	char name[IFNAMSIZ];
+	uint32_t min_mtu;
+	uint32_t max_mtu;
+};
+
+struct network_device_mtu_context {
+	struct network_device_mtu_limit *limits;
+	int pending;
+};
+
+static int
+rpc_luci_parse_network_device_mtu_limits(struct nl_msg *msg, void *arg)
+{
+	struct network_device_mtu_context *ctx = arg;
+	struct nlmsghdr *hdr = nlmsg_hdr(msg);
+	struct ifinfomsg *ifi = NLMSG_DATA(hdr);
+	struct network_device_mtu_limit *limit;
+	struct nlattr *tb[IFLA_MAX + 1];
+
+	if (hdr->nlmsg_type != RTM_NEWLINK)
+		return NL_SKIP;
+
+	nlmsg_parse(hdr, sizeof(*ifi), tb, IFLA_MAX, NULL);
+
+	if (!tb[IFLA_IFNAME] || (!tb[IFLA_MIN_MTU] && !tb[IFLA_MAX_MTU]))
+		return NL_SKIP;
+
+	limit = calloc(1, sizeof(*limit));
+	if (!limit)
+		return NL_SKIP;
+
+	snprintf(limit->name, sizeof(limit->name), "%s",
+	         nla_get_string(tb[IFLA_IFNAME]));
+
+	if (tb[IFLA_MIN_MTU])
+		limit->min_mtu = nla_get_u32(tb[IFLA_MIN_MTU]);
+
+	if (tb[IFLA_MAX_MTU])
+		limit->max_mtu = nla_get_u32(tb[IFLA_MAX_MTU]);
+
+	limit->next = ctx->limits;
+	ctx->limits = limit;
+
+	return NL_SKIP;
+}
+
+static int
+rpc_luci_network_device_mtu_done(struct nl_msg *msg, void *arg)
+{
+	struct network_device_mtu_context *ctx = arg;
+
+	ctx->pending = 0;
+	return NL_STOP;
+}
+
+static int
+rpc_luci_network_device_mtu_error(struct sockaddr_nl *nla,
+                                  struct nlmsgerr *err, void *arg)
+{
+	struct network_device_mtu_context *ctx = arg;
+
+	ctx->pending = 0;
+	return NL_STOP;
+}
+
 static void
-rpc_luci_parse_network_device_sys(const char *name, struct ifaddrs *ifaddr)
+rpc_luci_get_network_device_mtu_limits(struct network_device_mtu_context *ctx)
+{
+	struct ifinfomsg ifm = { .ifi_family = AF_UNSPEC };
+	struct nl_sock *sock = NULL;
+	struct nl_msg *msg = NULL;
+	struct nl_cb *cb = NULL;
+
+	sock = nl_socket_alloc();
+	if (!sock || nl_connect(sock, NETLINK_ROUTE))
+		goto out;
+
+	cb = nl_cb_alloc(NL_CB_DEFAULT);
+	if (!cb)
+		goto out;
+
+	msg = nlmsg_alloc_simple(RTM_GETLINK, NLM_F_REQUEST | NLM_F_DUMP);
+	if (!msg)
+		goto out;
+
+	nlmsg_append(msg, &ifm, sizeof(ifm), 0);
+	nl_cb_set(cb, NL_CB_VALID, NL_CB_CUSTOM,
+	          rpc_luci_parse_network_device_mtu_limits, ctx);
+	nl_cb_set(cb, NL_CB_FINISH, NL_CB_CUSTOM,
+	          rpc_luci_network_device_mtu_done, ctx);
+	nl_cb_err(cb, NL_CB_CUSTOM, rpc_luci_network_device_mtu_error, ctx);
+
+	ctx->pending = 1;
+
+	if (nl_send_auto_complete(sock, msg) < 0)
+		ctx->pending = 0;
+
+	while (ctx->pending && nl_recvmsgs(sock, cb) >= 0)
+		;
+
+out:
+	if (sock)
+		nl_socket_free(sock);
+
+	if (cb)
+		nl_cb_put(cb);
+
+	if (msg)
+		nlmsg_free(msg);
+}
+
+static struct network_device_mtu_limit *
+rpc_luci_find_network_device_mtu_limit(struct network_device_mtu_context *ctx,
+                                       const char *name)
+{
+	struct network_device_mtu_limit *limit;
+
+	for (limit = ctx->limits; limit; limit = limit->next)
+		if (!strcmp(limit->name, name))
+			return limit;
+
+	return NULL;
+}
+
+static void
+rpc_luci_parse_network_device_sys(const char *name, struct ifaddrs *ifaddr,
+                                  struct network_device_mtu_context *mtu_ctx)
 {
 	char link[64], buf[512], *p;
+	struct network_device_mtu_limit *mtu_limit;
 	unsigned int ifa_flags = 0;
 	struct sockaddr_ll *sll;
 	struct ifaddrs *ifa;
@@ -714,6 +842,13 @@ rpc_luci_parse_network_device_sys(const char *name, struct ifaddrs *ifaddr)
 	n = atoi(readstr("/sys/class/net/%s/mtu", name));
 	if (n > 0)
 		blobmsg_add_u32(&blob, "mtu", n);
+
+	mtu_limit = rpc_luci_find_network_device_mtu_limit(mtu_ctx, name);
+	if (mtu_limit && mtu_limit->min_mtu > 0)
+		blobmsg_add_u32(&blob, "min_mtu", mtu_limit->min_mtu);
+
+	if (mtu_limit && mtu_limit->max_mtu > 0)
+		blobmsg_add_u32(&blob, "max_mtu", mtu_limit->max_mtu);
 
 	n = atoi(readstr("/sys/class/net/%s/tx_queue_len", name));
 	if (n > 0)
@@ -863,11 +998,14 @@ rpc_luci_get_network_devices(struct ubus_context *ctx,
                              const char *method,
                              struct blob_attr *msg)
 {
+	struct network_device_mtu_context mtu_ctx = {};
+	struct network_device_mtu_limit *limit;
 	struct ifaddrs *ifaddr;
 	struct dirent *e;
 	DIR *d;
 
 	blob_buf_init(&blob, 0);
+	rpc_luci_get_network_device_mtu_limits(&mtu_ctx);
 
 	d = opendir("/sys/class/net");
 
@@ -882,13 +1020,20 @@ rpc_luci_get_network_devices(struct ubus_context *ctx,
 				break;
 
 			if (e->d_type != DT_DIR && e->d_type != DT_REG)
-				rpc_luci_parse_network_device_sys(e->d_name, ifaddr);
+				rpc_luci_parse_network_device_sys(e->d_name, ifaddr,
+				                                  &mtu_ctx);
 		}
 
 		if (ifaddr != NULL)
 			freeifaddrs(ifaddr);
 
 		closedir(d);
+	}
+
+	while (mtu_ctx.limits) {
+		limit = mtu_ctx.limits;
+		mtu_ctx.limits = limit->next;
+		free(limit);
 	}
 
 	ubus_send_reply(ctx, req, blob.head);
--- a/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
+++ b/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
@@ -1104,8 +1104,15 @@ return baseclass.extend({
 		o.datatype = 'uinteger';
 		o.depends({ type: 'bridge', multicast_querier: '1' });
 
+		const min_mtu = dev ? dev.getMinMTU() : null;
+		const max_mtu = dev ? dev.getMaxMTU() : null;
+		const effective_max_mtu = max_mtu || 9200;
+		const effective_min_mtu = effective_max_mtu < 576
+			? Math.min(min_mtu || effective_max_mtu, effective_max_mtu)
+			: Math.min(Math.max(min_mtu || 576, 576), effective_max_mtu);
+
 		o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu', _('MTU'));
-		o.datatype = 'range(576, 9200)';
+		o.datatype = 'range(%d, %d)'.format(effective_min_mtu, effective_max_mtu);
 		o.validate = function(section_id, value) {
 			const parent_mtu = (dev && dev.getType() == 'vlan') ? (parent_dev ? parent_dev.getMTU() : null) : null;
 
@@ -1275,9 +1282,11 @@ return baseclass.extend({
 		o.sysfs = '/proc/sys/net/ipv6/conf/%s/drop_unsolicited_na'.format(devname || 'default');
 		o.depends('ipv6', /1/);
 
-		o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu6', _('IPv6 MTU'));
-		o.datatype = 'max(9200)';
-		o.depends('ipv6', /1/);
+		if (effective_max_mtu >= 1280) {
+			o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu6', _('IPv6 MTU'));
+			o.datatype = 'range(1280, %d)'.format(effective_max_mtu);
+			o.depends('ipv6', /1/);
+		}
 
 		o = this.replaceOption(s, 'devgeneral', form.Value, 'dadtransmits', _('DAD transmits'), _('Amount of Duplicate Address Detection probes to send'));
 		o.placeholder = '1';
--- a/libs/rpcd-mod-luci/Makefile
+++ b/libs/rpcd-mod-luci/Makefile
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=rpcd-mod-luci
-PKG_VERSION:=20240305
+PKG_VERSION:=20260807
 PKG_RELEASE:=1
 PKG_MAINTAINER:=Jo-Philipp Wich <jo@mein.io>
 
--- a/modules/luci-base/htdocs/luci-static/resources/network.js
+++ b/modules/luci-base/htdocs/luci-static/resources/network.js
@@ -394,6 +394,8 @@ function initNetworkState(refresh) {
 					type:     dev.type,
 					devtype:  dev.devtype,
 					mtu:      dev.mtu,
+					min_mtu:  dev.min_mtu,
+					max_mtu:  dev.max_mtu,
 					qlen:     dev.qlen,
 					wireless: dev.wireless,
 					parent:   dev.parent,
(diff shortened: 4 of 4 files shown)


modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1104,8 +1104,13 @@ return baseclass.extend({
...
 
-		o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu', _('MTU'));
-		o.datatype = 'range(576, 9200)';
+		const min_mtu = dev ? dev.getMinMTU() : null;
+		const max_mtu = dev ? dev.getMaxMTU() : null;
+		const mtu_range = (min_mtu && max_mtu) ?
+			_('Supported MTU range: %d–%d bytes').format(min_mtu, max_mtu) : null;
+
+		o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu', _('MTU'), mtu_range);
+		o.datatype = 'range(%d, %d)'.format(min_mtu || 576, max_mtu || 9200);
```
openwrt-ai: Taking the kernel values verbatim drops the two sanity bounds this range previously provided, in both directions.

**Floor:** `ether_setup()` sets `dev->min_mtu = ETH_MIN_MTU` (68), so essentially every ethernet-like device reports 68. The form will now accept an MTU of 68 on a WAN/LAN device — legal for the kernel, but it breaks IPv4 (RFC 791 minimum reassembly buffer is 576) and IPv6 (1280). The old 576 floor was deliberate.

**Ceiling:** the useful half of this change is *lowering* the ceiling for NICs that only do 1500 (`ether_setup()` also sets `max_mtu = ETH_DATA_LEN`, and only jumbo-capable drivers raise it). But virtual devices go the other way: `br_dev_setup()` and `veth_setup()` both override `dev->max_mtu = ETH_MAX_MTU` (65535). On a stock OpenWrt box `br-lan` therefore reports `max_mtu = 65535`, so this field will accept 65535 for the bridge and the hint will advertise "Supported MTU range: 68–65535 bytes" — neither is a real limit, since the effective bridge MTU is bounded by its ports. Is capping at the previous 9200 (or at `min()` over the bridge ports) intended as a follow-up, or should the bridge case be special-cased here?

The floor part is unambiguous, so at minimum:

```suggestion
		o.datatype = 'range(%d, %d)'.format(Math.max(min_mtu || 576, 576), max_mtu || 9200);
```


---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1104,8 +1104,13 @@ return baseclass.extend({
 		o.datatype = 'uinteger';
 		o.depends({ type: 'bridge', multicast_querier: '1' });
 
-		o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu', _('MTU'));
-		o.datatype = 'range(576, 9200)';
+		const min_mtu = dev ? dev.getMinMTU() : null;
+		const max_mtu = dev ? dev.getMaxMTU() : null;
+		const mtu_range = (min_mtu && max_mtu) ?
+			_('Supported MTU range: %d–%d bytes').format(min_mtu, max_mtu) : null;
```
openwrt-ai: nit: the hint requires *both* values (`min_mtu && max_mtu`), but the `datatype` on the next line falls back to each one independently. A device that reports only `max_mtu` (e.g. a driver that leaves `min_mtu` at 0, which the rpcd side then drops because of `if (n > 0)`) gets its MTU rejected against a kernel-reported ceiling that is never shown in the UI. Gating the hint the same way the validation is gated — showing the effective pair that `datatype` actually uses — would keep the two in sync.

---

libs/rpcd-mod-luci/src/luci.c
```diff
@@ -715,6 +715,14 @@ rpc_luci_parse_network_device_sys(const char *name, struct ifaddrs *ifaddr)
 	if (n > 0)
 		blobmsg_add_u32(&blob, "mtu", n);
 
+	n = atoi(readstr("/sys/class/net/%s/min_mtu", name));
+	if (n > 0)
+		blobmsg_add_u32(&blob, "min_mtu", n);
+
+	n = atoi(readstr("/sys/class/net/%s/max_mtu", name));
+	if (n > 0)
+		blobmsg_add_u32(&blob, "max_mtu", n);
```
openwrt-ai: This changes the `getNetworkDevices` payload but [`libs/rpcd-mod-luci/Makefile#L10-L11`](https://github.com/openwrt/luci/blob/7d5612b5af8700f1f8f4e76f0b3fe1610bca84ff/libs/rpcd-mod-luci/Makefile#L10-L11) still carries `PKG_VERSION:=20240305` / `PKG_RELEASE:=1`. Without a bump, existing installs will not pull the rebuilt `rpcd-mod-luci`, so the new `luci-base`/`luci-mod-network` will keep hitting the 576/9200 fallback silently — which looks like the feature simply not working. The repo's habit for exactly this situation is a version bump alongside the source change (e.g. [eb93ea0 "rpcd-mod-luci: bump version"](https://github.com/openwrt/luci/commit/eb93ea0ac244397aaba57a9c9db7414c518c7693), [a26b533](https://github.com/openwrt/luci/commit/a26b53314db468e2a3c0c5ea4d8db4939d8eb09c)).

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1104,8 +1104,13 @@ return baseclass.extend({
...
+
+		o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu', _('MTU'), mtu_range);
```
openwrt-ai: nit: the sibling `mtu6` option further down in the same function still hardcodes the old ceiling — [`tools/network.js:1278-1279`](``https://github.com/openwrt/luci/blob/7d5612b5af8700f1f8f4e76f0b3fe1610bca84ff/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js#L1278-L1279)`` (`o.datatype = 'max(9200)'`). The IPv6 MTU is bounded by the same `dev->max_mtu`, so after this change the two fields on the same form disagree about what the device supports. Was leaving `mtu6` alone intentional (e.g. because IPv6 MTU has its own 1280 floor you did not want to conflate), or just out of scope for this pass?

---

openwrt-ai (commented):
## Commit checks

- `7d5612b` "luci-mod-network: use device-specific MTU limits" — the subject scopes the change to `luci-mod-network`, but the commit also touches `libs/rpcd-mod-luci/src/luci.c` (new sysfs reads in the rpcd C plugin) and `modules/luci-base/htdocs/luci-static/resources/network.js` (new `Device.getMinMTU()`/`getMaxMTU()` API). `CONTRIBUTING.md` asks for a subject prefixed with the affected component, and this tree normally splits per-package changes into separate commits — here that would be three: one `rpcd-mod-luci:` (backend attributes + version bump), one `luci-base:` (model accessors), one `luci-mod-network:` (form validation). Splitting also makes the cross-repo/package dependency explicit, which matters because the frontend silently degrades to the old fallback range when the backend half is not installed.


---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
dormancygrace: The kernel attributes describe per-netdevice L2 bounds, not protocol or topology bounds. This form configures generic L2 devices, so imposing 576 would reject valid low-MTU non-IP uses and would no longer show the real min_mtu requested by this change. Bridge port constraints are dynamic and are still enforced by the kernel/netifd when applying the configuration. I renamed the hint to “Kernel MTU range” to avoid presenting it as an end-to-end path limit.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
dormancygrace: Fixed. The form now computes one effective pair, uses it for both validation and the hint, and shows the hint when either kernel value is available.

libs/rpcd-mod-luci/src/luci.c
dormancygrace: Fixed. rpcd-mod-luci is now a separate package-scoped commit and PKG_VERSION is bumped to 20260807, so package upgrades will install the updated RPC module.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
dormancygrace: Fixed. mtu6 now uses the same device-specific upper bound while retaining the IPv6 minimum of 1280 bytes.

dormancygrace:
Addressed the review: split the backend, model and form changes into three package-scoped commits; bumped rpcd-mod-luci; synchronized the displayed and validated ranges; and applied the device-specific ceiling to IPv6 MTU. The branch was force-pushed and the PR description was updated.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
dormancygrace: Follow-up: after reconsidering router UI compatibility, the generic MTU field now keeps the established 576 floor with Math.max(kernel_min_mtu, 576). The kernel-reported maximum is still enforced. IPv6 MTU uses its protocol minimum of 1280.

libs/rpcd-mod-luci/src/luci.c
```diff
@@ -715,6 +777,13 @@ rpc_luci_parse_network_device_sys(const char *name, struct ifaddrs *ifaddr)
 	if (n > 0)
 		blobmsg_add_u32(&blob, "mtu", n);
 
+	rpc_luci_get_network_device_mtu_limits(name, &min_mtu, &max_mtu);
```
openwrt-ai: This allocates a netlink socket, connects it, sends a request and blocks on the reply **once per network device**: `rpc_luci_parse_network_device_sys()` is called from the `readdir("/sys/class/net")` loop at [luci.c:954](https://github.com/openwrt/luci/blob/49af7953f5b2b02fb860556a74e82d1e0fb4e353/libs/rpcd-mod-luci/src/luci.c#L954), so a box with 30 netdevs now does 30 socket create/connect/send/recv/free cycles on every `getNetworkDevices` call — and that call is polled by the LuCI status and network pages.

The existing netlink user in this file, [`rpc_luci_get_host_hints_nl()`](https://github.com/openwrt/luci/blob/49af7953f5b2b02fb860556a74e82d1e0fb4e353/libs/rpcd-mod-luci/src/luci.c#L1443-L1495), does the opposite: one socket, one `NLM_F_DUMP` request, all results collected in the callback. The same shape works here — a single `RTM_GETLINK | NLM_F_DUMP` issued once in `rpc_luci_get_network_devices()` before the readdir loop, keyed by ifindex or ifname into a small lookup that `rpc_luci_parse_network_device_sys()` then consults. That also removes the per-device blocking `nl_recvmsgs_default()` from the hot path.

---

libs/rpcd-mod-luci/src/luci.c
```diff
@@ -644,10 +644,72 @@ lease_next(void)
...
+}
+
+static void
+rpc_luci_get_network_device_mtu_limits(const char *name, uint32_t *min_mtu,
+                                       uint32_t *max_mtu)
+{
+	struct ifinfomsg ifm = { .ifi_family = AF_UNSPEC };
+	struct nl_sock *sock = NULL;
+	struct nl_msg *msg = NULL;
+	uint32_t limits[3] = { if_nametoindex(name), 0, 0 };
```
openwrt-ai: nit: packing three unrelated values — the input ifindex plus the two output MTUs — into `uint32_t limits[3]` means the callback at line 648 has to address them by magic index (`limits[0]` is the filter key, `limits[1]`/`limits[2]` are results). A small file-local struct with named members would make both halves self-explanatory, and would let `ifi_index` be compared against an `int` without the signed/unsigned mix on line 655.

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1104,8 +1104,13 @@ return baseclass.extend({
 		o.datatype = 'uinteger';
 		o.depends({ type: 'bridge', multicast_querier: '1' });
 
+		const min_mtu = dev ? dev.getMinMTU() : null;
+		const max_mtu = dev ? dev.getMaxMTU() : null;
+		const effective_min_mtu = Math.max(min_mtu || 576, 576);
+		const effective_max_mtu = max_mtu || 9200;
```
openwrt-ai: The 576 floor is applied without regard to the device's reported ceiling, so a device whose `max_mtu` is below 576 produces `range(576, <max>)` with min > max — a field no value can satisfy, where the old hardcoded `range(576, 9200)` at least accepted something. Devices like 6LoWPAN/802.15.4 netdevs report `IFLA_MAX_MTU` well under 576, and any future device that does will silently become uneditable rather than falling back.

Clamping the floor to the ceiling keeps the intended "safe 576 minimum" for normal devices while staying satisfiable for the rest:

```suggestion
		const effective_max_mtu = max_mtu || 9200;
		const effective_min_mtu = Math.min(Math.max(min_mtu || 576, 576), effective_max_mtu);
```

The same applies to the IPv6 MTU field on line 1284 — `range(1280, %d)` with `effective_max_mtu` below 1280 is likewise unsatisfiable, so that one needs `Math.max(effective_max_mtu, 1280)` (or the field hidden entirely when the device cannot carry an IPv6 MTU).

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1104,8 +1104,13 @@ return baseclass.extend({
...
+
 		o = this.replaceOption(s, 'devgeneral', form.Value, 'mtu', _('MTU'));
```
openwrt-ai, line 1114: nit: the PR description says the limits are "preserved in the shared network device model, **displayed below the MTU field**, and used for form validation", and the follow-up comment says "synchronized the displayed and validated ranges" — but this revision passes no description argument, so nothing is displayed. The previous revision (`7d5612b`) did render a `Supported MTU range: %d–%d bytes` hint here and it was dropped in the force-push. Either restore the hint or drop the display claim from the PR body so the two agree.

---

openwrt-ai (commented):
## Commit checks

- `b50f6e0` "rpcd-mod-luci: expose device MTU limits" — the message says "Linux exposes the supported MTU range through `min_mtu` and `max_mtu` **sysfs attributes**", but the diff does not read sysfs: it opens a `NETLINK_ROUTE` socket and parses `IFLA_MIN_MTU`/`IFLA_MAX_MTU` from an `RTM_GETLINK` reply. There are no `min_mtu`/`max_mtu` files under `/sys/class/net/<dev>/` (the kernel only exposes `mtu` there), which is presumably why the implementation moved to rtnetlink between revisions — but the commit message still describes the earlier sysfs approach. The PR body already says "read from rtnetlink"; the commit message should match.

The per-package split requested on the previous revision is done correctly — all three commit subjects match the files they touch.

---

libs/rpcd-mod-luci/src/luci.c
dormancygrace: Fixed in 135e2e1: getNetworkDevices now issues one RTM_GETLINK dump, builds an interface-name lookup, and reuses it throughout the sysfs enumeration. Cross-compilation for ramips/mt7620 passes.

libs/rpcd-mod-luci/src/luci.c
dormancygrace: Fixed as part of the dump conversion. The callback now uses named fields in network_device_mtu_limit and network_device_mtu_context; there are no magic array indexes or signed ifindex comparison.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
dormancygrace: Fixed in 3ae81b1. The effective IPv4 floor is clamped to the device ceiling, and the IPv6 MTU field is omitted when the device maximum is below the IPv6 minimum.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
dormancygrace: Fixed the PR description to match the intended UI: the separate range hint remains removed, while the field validation itself uses the kernel limits.

dormancygrace:
Fixed in the current head. Commit `135e2e1e37` now describes the actual implementation: one `RTM_GETLINK` dump, with no sysfs claim. The PR body already uses the same rtnetlink wording.

libs/rpcd-mod-luci/src/luci.c
```diff
@@ -644,10 +644,138 @@ lease_next(void)
...
+{
+	struct network_device_mtu_context *ctx = arg;
+
+	ctx->pending = 0;
+	return NL_STOP;
+}
+
+static int
+rpc_luci_network_device_mtu_error(struct sockaddr_nl *nla,
+				  struct nlmsgerr *err, void *arg)
```
openwrt-ai: nit: this continuation line is indented with tabs, while every other multi-line function signature in this file aligns with spaces under the opening paren — e.g. [`rpc_luci_get_network_devices()` at luci.c:998-1002](https://github.com/openwrt/luci/blob/3ae81b10d3eac7bd81a423447c98d7c1af5c2203/libs/rpcd-mod-luci/src/luci.c#L998-L1002) and [`rpc_luci_get_wireless_devices()` at luci.c:1175-1178](https://github.com/openwrt/luci/blob/3ae81b10d3eac7bd81a423447c98d7c1af5c2203/libs/rpcd-mod-luci/src/luci.c#L1175-L1178). The rest of this same commit already uses spaces (`nl_cb_set(...)` on line 736, `snprintf(...)` on line 681), so the file is now mixed.

```suggestion
                                  struct nlmsgerr *err, void *arg)
```

Same applies to line 762 (`rpc_luci_find_network_device_mtu_limit`, 39 spaces) and line 774 (`rpc_luci_parse_network_device_sys`, 34 spaces).

---

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
```diff
@@ -1104,8 +1104,13 @@ return baseclass.extend({
 		o.datatype = 'uinteger';
 		o.depends({ type: 'bridge', multicast_querier: '1' });
 
+		const min_mtu = dev ? dev.getMinMTU() : null;
+		const max_mtu = dev ? dev.getMaxMTU() : null;
+		const effective_max_mtu = max_mtu || 9200;
+		const effective_min_mtu = Math.min(Math.max(min_mtu || 576, 576), effective_max_mtu);
```
openwrt-ai: nit: refinement of the clamp I suggested on the previous revision — it makes the range satisfiable, but degenerate. For a device whose kernel ceiling is below 576, `effective_min_mtu` collapses onto `effective_max_mtu`, so the `datatype` on line 1113 becomes `range(N, N)` and the field accepts exactly one value even though the kernel reports a much lower `min_mtu` for the same device. Those are precisely the devices this branch exists for, and for them the kernel's own minimum is a better floor than its maximum:

```suggestion
		const effective_min_mtu = effective_max_mtu < 576
			? Math.min(min_mtu || effective_max_mtu, effective_max_mtu)
			: Math.max(min_mtu || 576, 576);
```

Normal devices are unaffected (the `>= 576` branch is the current expression verbatim).

---

openwrt-ai (commented):
Reviewed 3 new commits (force-push over `49af795`); no commit-message issues — the sysfs wording flagged last round is gone and each subject matches the files it touches.

The rtnetlink rework looks right: one `RTM_GETLINK | NLM_F_DUMP` per `getNetworkDevices`, the same `nl_cb_alloc`/`nl_cb_set`/`nl_cb_err` + `pending` shape as the existing `rpc_luci_get_host_hints_nl()`, and the limit list is freed on every path. Two nits inline.

---

libs/rpcd-mod-luci/src/luci.c
dormancygrace: Fixed all three continuation lines to use the space-aligned style already used by the surrounding multi-line signatures. The repository ESLint check also passes under Node 24.

modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js
dormancygrace: Fixed the sub-576 case to use the kernel-reported minimum (or the device maximum when no minimum is reported). I retained the existing outer clamp in the normal >=576 branch so even inconsistent min/max data still yields a satisfiable range. Ordinary router interfaces therefore keep the agreed 576 floor.

openwrt-ai (commented):
Reviewed 3 new commits (force-push over `3ae81b1`). The only code change since the last revision is the sub-576 branch in `tools/network.js`, and it holds up: I exercised the new expression over the interesting `(min_mtu, max_mtu)` pairs — `(68,1500)`→`range(576,1500)`, `(128,300)`→`range(128,300)`, `(0,300)`→`range(300,300)`, `(400,300)`→`range(300,300)`, `(null,null)`→`range(576,9200)` — and every result is satisfiable with the kernel minimum preferred over the ceiling on low-MTU devices, which was the point. The three continuation-line re-indents in `luci.c` match the space-aligned style of the surrounding signatures. FormalityCheck is green on `8674732`.

## Commit checks

- `8674732` "luci-mod-network: use device-specific MTU limits" — the commit also touches `libs/rpcd-mod-luci/src/luci.c` (the three continuation-line re-indents at lines 705, 761 and 774), which is outside the `luci-mod-network:` subject scope and re-formats whitespace that `135e2e1` introduced earlier in the same series. Squashing those three hunks into `135e2e1` keeps each commit inside its own package — the split you did on the previous revision — and removes an introduce-then-fix pair from the series. `135e2e1` and `07836c5` match the files they touch.

---

dormancygrace:
Good point. I moved the three `rpcd-mod-luci` continuation-line formatting changes into the first backend commit, so the final `luci-mod-network` commit now touches only `tools/network.js`.

The aggregate tree is byte-identical to the previously green `8674732947` head (`git diff 8674732947..8ed21a4afc` is empty); only commit ownership changed. Updated head: `8ed21a4afc`.

openwrt-ai (commented):
Reviewed 3 new commits (force-push over `8674732`); no new issues found.

I confirmed the rebase claim locally: `git diff 8674732947..8ed21a4afc` is empty, so the tree is unchanged and only commit ownership moved. The commit-check nit from the last round is resolved — `f6eaaae` now carries all of `libs/rpcd-mod-luci/` (Makefile bump + `luci.c`, including the three continuation-line re-indents), `ac41ffc` only `luci-base`'s `network.js`, and `8ed21a4` only `luci-mod-network`'s `tools/network.js`. Each subject matches its files, and the introduce-then-fix pair is gone. All three FormalityCheck runs are green on `8ed21a4`.

---

openwrt-ai (commented):
Reviewed 3 new commits; no new issues found.

---

PR #8918, opened 2026-08-07
luci-app-wifischedule: fix untranslated validation error message

DH-41:
### Description
Fix an issue where the error message was not being translated because the .format() method was incorrectly included inside the `_()` i18n wrapper function, preventing the i18n scanner from extracting the translation key.

---

## Tested on

**OpenWrt version:** SNAPSHOT
**LuCI version:** master
**Web browser(s):** Chrome

---

## Checklist


jow-:
Merged, thanks!

PR #8919, opened 2026-08-07
luci-base: include original stacktrace in render exception

### LINKED ISSUE #8909: luci-base: backtraces from template render errors are shadowed

### screenshots or captures

_No response_

### Actual behaviour

When the rendering of a template throws an exception, it is caught and re-thrown in `runtime.render_any()`, which overwrites the original backtrace.

Example, the exception would have actually been from `dispatcher.lookup()`.
```
Error

In [anonymous function](), file /usr/share/ucode/luci/runtime.uc, line 133, byte 10:
  called from function run_action (/usr/share/ucode/luci/dispatcher.uc:815:47)
  called from function [anonymous function] (/usr/share/ucode/luci/dispatcher.uc:1107:48)
  called from anonymous function (/www/cgi-bin/luci:39:13)

              die(ex);
  Near here --------^
```

### Expected behaviour

The backtrace reflects the actual location of the exception.

Example, actual backtrace of the same exception from `dispatcher.lookup()`:
```
Reference error

In lookup(), file /usr/share/ucode/luci/dispatcher.uc, line 762, byte 24:
  called from function main (/usr/share/ucode/luci/template/themes/bootstrap/header.ut:61:51)
  called from function call ([C])
  called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:80:36)
  called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:127:40)
  called from function [arrow function] (/usr/share/ucode/luci/runtime.uc:183:57)
  called from function main (/usr/share/ucode/luci/template/header.ut:7:34)
  called from function call ([C])
  called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:80:36)
  called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:127:40)
  called from function [arrow function] (/usr/share/ucode/luci/runtime.uc:183:57)
  called from function main (/usr/share/ucode/luci/template/view.ut:1:20)
  called from function call ([C])
  called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:80:36)
  called from function [anonymous function] (/usr/share/ucode/luci/runtime.uc:127:40)
  called from function run_action (/usr/share/ucode/luci/dispatcher.uc:815:47)
  called from function [anonymous function] (/usr/share/ucode/luci/dispatcher.uc:1107:48)
  called from anonymous function (/usr/share/ucode/walter/web/index.uc:47:15)

          node = node.children[name];
  Near here -------------------^
```

### Steps to reproduce

1. Trigger any kind of exception in a template
2. Observe that the template file isn't part of the backtrace

### Additional Information

```shell
{
        "kernel": "6.18.41",
        "hostname": "OpenWrt",
        "system": "AMD EPYC",
        "rootfs_type": "ext4",
        "release": {
                "distribution": "OpenWrt",
                "version": "SNAPSHOT",
                "revision": "r35634-d3191cb056",
                "target": "x86/64",
                "description": "OpenWrt SNAPSHOT r35634-d3191cb056",
                "builddate": "1785620585"
        }
}
```

_No response_

### Relevant log output

```shell

```

pktpls:
### Description
In order to properly handle nested variable scopes, when an exception is thrown during the rendering of the template, it is caught and then re-thrown after some scope bookkeeping is done.

This means the exception loses its stacktrace information. The exception now has the stacktrace of the re-throw location, not the original stacktrace which would be much more useful.

Let's at least include that stacktrace in the message itself. It's not a clean solution, but better than not having this info at all.

Fixes #8909 

### Maintainer _(preferred)_
@jow- 

---

## Tested on

**LuCI version:** LuCI master
**Web browser(s):** Firefox 153

### DIFF
--- a/modules/luci-base/ucode/runtime.uc
+++ b/modules/luci-base/ucode/runtime.uc
@@ -130,7 +130,16 @@ const Class = {
 		}
 		catch (ex) {
 			pop(this.scopes);
-			die(ex);
+
+			// To enable the scope bookkeeping right above,
+			// we catch and re-throw the exception, losing its original stacktrace.
+			// To keep the original stacktrace, we inject it into the message itself.
+			// The exception gets re-thrown multiple times from nested render calls,
+			// so if its message already contains a stacktrace, we don't do anything.
+			if (index(ex.message, "Near here") < 0)
+				die(`${ex.message}\n\n${ex.stacktrace[0].context}`);
+
+			die(ex)
 		}
 
 		pop(this.scopes);


jow-:
I am considering a change to ucode's `die()` which would just re-throw a given exception object verbatim. Means if you do a

```javascript
try {
    something_failing();
}
catch (ex) {
    die(ex);
}
```

... then the thrown exception would retain the original message, type and stacktrace.

I'd prefer this solution even if it means that it'll take until the next OpenWrt ucode update that the stacktraces look correct.

pktpls:
> I'd prefer this solution even if it means that it'll take until the next OpenWrt ucode update that the stacktraces look correct.

Yes that's clearly the better solution :-) I'll carry this commit here locally until then.

PR #8920, opened 2026-08-07
themes: fix custom CSS via menu.d node

pktpls:
#### themes: fix custom CSS via menu.d node

The themes have supported custom CSS via menu.d nodes since they
were first ported to ucode. There were two little problems though:

- The css property wasn't added to the node schema in luci.dispatcher.
- The themes used the wrong variable name.

The advantage is that the client will load the custom CSS right away,
instead of later when the view initializes. In my local development,
it has prevented visual flickering on page load.

Example menu.d entry:

    "falter": {
      "title": "Settings",
      "order": 5,
      "action": {
        "type": "view",
        "path": "falter/settings"
      },
      "css": "view/falter/falter.css"
    }

There are lots of potential users for this, since many Luci applications
inject their CSS via client-side JS. Even luci-mod-network does.

#### luci-theme-bootstrap: remove broken title class on `<body>`

When the bootstrap theme was first ported to ucode, the dispatched
menu.d title was intended to be added to the <body> CSS classes.

It never worked because it uses the wrong variable name.
It also doesn't seem useful and other themes don't do it.

So let's just remove it.

### Maintainer _(preferred)_
@jow- 

---

## Tested on

**LuCI version:** LuCI master
**Web browser(s):** Firefox 153

### DIFF
--- a/modules/luci-base/ucode/dispatcher.uc
+++ b/modules/luci-base/ucode/dispatcher.uc
@@ -350,6 +350,7 @@ function build_pagetree() {
 		action: 'object',
 		auth: 'object',
 		cors: 'bool',
+		css: 'string',
 		depends: 'object',
 		order: 'int',
 		setgroup: 'string',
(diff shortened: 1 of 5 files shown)


jow-:
Merged, thanks!

PR #8925, opened 2026-08-09
[openwrt-24.10] luci-app-natmap: allow reading status file under /tmp/run

ragingbulld:
Backport of #8923 to openwrt-24.10.

**Root cause:** rpcd's file plugin now checks ACL against the real path of symlinks (openwrt/rpcd e37ed9d), so /var/run/natmap/* (resolving to /tmp/run/natmap/*) is rejected and the LuCI status view shows none on 24.10.8.

**Fix:** add `/tmp/run/natmap/*` to the read ACL.

Ref: heiher/natmap#134

### DIFF
--- a/applications/luci-app-natmap/root/usr/share/rpcd/acl.d/luci-app-natmap.json
+++ b/applications/luci-app-natmap/root/usr/share/rpcd/acl.d/luci-app-natmap.json
@@ -3,7 +3,8 @@
 		"description": "Grant access to LuCI app natmap",
 		"read": {
 			"file": {
-				"/var/run/natmap/*": [ "read" ]
+				"/var/run/natmap/*": [ "read" ],
+				"/tmp/run/natmap/*": [ "read" ]
 			},
 			"ubus": {
 				"service": [ "list" ]


openwrt-ai (commented):
## Commit checks

- `388e727` "luci-app-natmap: allow reading status under /tmp/run" — the `(cherry picked from commit f4c8d21a5660dfe228a1910d7213801f28d6276c)` trailer does not reference a commit in `openwrt/luci`. `f4c8d21a` is the head commit of the still-open master PR #8923 (`git merge-base --is-ancestor f4c8d21a5660dfe228a1910d7213801f28d6276c origin/master` → not an ancestor). Once #8923 is applied to master the resulting commit will carry a different SHA, so as it stands the trailer points at a commit that will never exist in the repository. Suggest holding this backport until the master commit lands, then regenerating it with `git cherry-pick -x <merged-sha>` so the trailer records the real master SHA.

Otherwise the backport looks faithful: the resulting `luci-app-natmap.json` is byte-identical to the one produced by #8923, the commit message body matches the upstream commit verbatim, and the dual `/var/run/...` + `/tmp/run/...` grant matches the existing in-tree pattern used by [`luci-app-adblock.json`](``https://github.com/openwrt/luci/blob/388e72790162e8e136235a98bae5d966a3947a50/applications/luci-app-adblock/root/usr/share/rpcd/acl.d/luci-app-adblock.json#L44-L59),`` `luci-app-banip` and `luci-app-travelmate`. The root cause described in the PR body checks out against rpcd's [`rpc_file_access()` re-authorization of the `realpath()` result in `file.c`](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L264-L289), which re-runs the ACL check against the resolved path — so both entries are required, not just the `/tmp/run` one.

---

heiher (approved):
LGTM. Thanks!

jow-:
Merged, thanks!

PR #8926, opened 2026-08-09
luci-app-nut: add/update SSL configuration for full support

danielfdickinson:
### Description

Both the configuration for NSS-based SSL and the configuration for
OpenSSL-based SSL were incomplete, so add missing options, and enhance
existing settings.

Additionally, the method of checking for being compiled with SSL changes
with https://github.com/openwrt/packages/pull/29893. We now have a file
/usr/share/nut/ssl_backend which contains either 'nss', 'openssl', or
'none'. We use this to show the applicable configuration options in the
UI when it is available. For compatibility with previous version of the
nut package, we keep the 'ldd' method of determining availability of
SSL, for now.

That same pull request adds server-side CERTPATH and so on handling in
https://github.com/danielfdickinson/packages/blob/c0ec3741ce64b8613b8324575f907342303563fc/net/nut/files/nut-server-config.sh.functions#L94-L107

Finally, SSL settings are security-sensitive so move them to the 'admin'
NUT screens to reduce visibility, and to prepare for a future commit to
the NUT package which will separate the UCI configs for admin and
non-privileged users.

### Maintainer

---

## Tested on

**OpenWrt version:** OpenWrt SNAPSHOT r35715-94a21b3fe9
**LuCI version:** LuCI Master 26.220.05649~98417a6
**Web browser(s):** Firefox 140.13.0esr (64-bit)

### Server and Self-client

**OpenWrt version:** OpenWrt SNAPSHOT r35906-3d1645ee2
**LuCI version:** LuCI Master 26.239.32285~01b60d0
**Web browser(s):** Firefox 140.14.0esr (64-bit)

### Client

**OpenWrt version:** OpenWrt SNAPSHOT r35921-f0204d78e1
**LuCI version:** LuCI Master 26.238.54490~7a6ff75
**Web browser(s):** Firefox 140.14.0esr (64-bit)

---

## Checklist
- [ ] Includes what Issue it closes (e.g. openwrt/luci#issue-number).
  Prefers but does not depend on: https://github.com/openwrt/packages/pull/29893

### DIFF
--- a/applications/luci-app-nut/Makefile
+++ b/applications/luci-app-nut/Makefile
@@ -12,7 +12,7 @@ PKG_LICENSE:=Apache-2.0
 PKG_MAINTAINER:=Daniel F. Dickinson <dfdpublic@wildtechgarden.ca> \
 		Paul Donald <newtwen+github@gmail.com>
 
-EXTRA_DEPENDS:=nut (>= 2.8.4-3)
+LUCI_EXTRA_DEPENDS:=nut (>=2.8.4-r3)
 
 include ../../luci.mk
 
--- a/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js
+++ b/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js
@@ -1,9 +1,6 @@
 'use strict';
 'require form';
-'require fs';
 'require view';
-
-const upsmon_tool = '/usr/sbin/upsmon';
 
 function ESIFlags(o) {
 	o.value('EXEC', _('Execute notify command'));
@@ -50,19 +47,8 @@ function MonitorUserOptions(s) {
 }
 
 return view.extend({
-	load: function() {
-		return Promise.all([
-			L.resolveDefault(fs.exec_direct('/usr/bin/ldd', [upsmon_tool]), []).catch(function(err) {
-				throw new Error(_('Unable to run ldd: %s').format(err.message));
-			}).then(function(stdout) {
-				return stdout.includes('libssl.so');
-			}),
-		])
-	},
-
-	render: function(loaded_promises) {
+	render: function() {
 		let m, s, o;
-		const have_ssl_support = loaded_promises[0];
 
 		m = new form.Map('nut_monitor', _('NUT Monitor'),
 			_('Network UPS Tools Monitoring Configuration'));
@@ -94,16 +80,6 @@ return view.extend({
 		o.datatype = 'uinteger'
 		o.optional = true;
 		o.placeholder = 15;
-
-		if (have_ssl_support) {
-			o = s.option(form.Value, 'certpath', _('CA Certificate path'), _('Path containing ca certificates to match against host certificate'));
-			o.optional = true;
-			o.placeholder = '/etc/ssl/certs'
-
-			o = s.option(form.Flag, 'certverify', _('Verify all connection with SSL'), _('Require SSL and make sure server CN matches hostname'));
-			o.optional = true;
-			o.default = false;
-		}
 
 		s = m.section(form.TypedSection, 'notifications', _('Notifications settings'));
 		s.optional = true;
--- a/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
+++ b/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
@@ -1,14 +1,29 @@
 'use strict';
 'require form';
+'require fs';
 'require view';
+
+const ups_ssl_backend_file = '/usr/share/nut/ssl_backend';
+const upsmon_daemon = '/usr/sbin/upsmon';
 
 return view.extend({
 	load: function() {
-
+		return Promise.all([
+			fs.trimmed(ups_ssl_backend_file).then(function(backend) {
+				if (backend)
+					return backend;
+				// nut older than 2.8.5-5 does not ship ssl_backend, so probe the binary instead
+				return L.resolveDefault(fs.exec_direct('/usr/bin/ldd', [upsmon_daemon]), '').then(function(stdout) {
+					return stdout.includes('libssl.so') ? 'old_nut_openssl' : 'none';
+				});
+			}),
+		])
 	},
 
-	render: function() {
+	render: function(loaded_promises) {
 		let m, s, o;
+
+		const ssl_support_type = loaded_promises[0];
 
 		m = new form.Map('nut_monitor', _('NUT Monitor'),
 			_('Network UPS Tools Monitoring Configuration'));
@@ -26,6 +41,51 @@ return view.extend({
 		o = s.option(form.Value, 'shutdowncmd', _('Shutdown command'));
 		o.optional = true;
 		o.placeholder = '/usr/sbin/nutshutdown'
+
+		if (ssl_support_type == 'openssl') {
+			o = s.option(form.FileUpload, 'certfile', _('Client certificate chain and private key'), _('PEM file containing client certificate chain and private key (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl') {
+			// For OpenSSL, certpath is a PEM file, not a database directory as with NSS
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			// For NSS, certpath is a database directory, not a PEM file as with OpenSSL.
+			// We do not fill this field by default as the default database has no certificates,
+			// and must be filled to be useful. Doing this through the UI is a future task.
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'), _('An empty database was created in /etc/nut/cert_db at package install. To use it, SSH in, fill it using certutil, and put /etc/nut/cert_db in this field.'));
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certident', _('Certificate name and password'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has a space in it. An empty password must be entered as an empty pair of double-quotes ("")'));
+			o.placeholder = '"certificate name" "certificate password"';
+			o.password = true;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Flag, 'certverify', _('Verify all connections with SSL'), _('Require SSL and verify host certificate'));
+			o.optional = true;
+			o.default = false;
+			o.rmempty = false
+
+			o = s.option(form.Flag, 'forcessl', _('Require SSL to connect'), _('Require SSL even if not verifying host certificates'));
+			o.optional = true;
+			o.default = false;
+			o.rmempty = false;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.DynamicList, 'certhost', _('Per-host override for certident, certverify, and forcessl'), _('hostname "certificate name" certverify (as 0 or 1) forcessl (as 0 or 1)'));
+			o.optional = true;
+			o.placeholder = 'hostname "certificate name" certverify forcessl';
+		}
 
 		return m.render();
 	}
--- a/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
+++ b/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
@@ -5,7 +5,6 @@
 
 const driver_path = '/usr/libexec/nut/';
 const old_driver_path = '/lib/nut/';
-const ups_daemon = '/usr/sbin/upsd';
 
 function resolveDriverList(path) {
 	return Promise.resolve(L.resolveDefault(fs.list(path), []).then(function(entries) {
@@ -24,11 +23,6 @@ function resolveDriverList(path) {
 return view.extend({
 	load: function() {
 		return Promise.all([
-			L.resolveDefault(fs.exec_direct('/usr/bin/ldd', [ups_daemon]), []).catch(function(err) {
-				throw new Error(_('Unable to run ldd: %s').format(err.message));
-			}).then(function(stdout) {
-				return stdout.includes('libssl.so');
-			}),
 			resolveDriverList(driver_path),
 			resolveDriverList(old_driver_path),
 		])
@@ -36,8 +30,8 @@ return view.extend({
 
 	render: function(loaded_promises) {
 		let m, s, o;
-		const have_ssl_support = loaded_promises[0];
-		const driver_list = (loaded_promises[1].length > 0) ? loaded_promises[1] : loaded_promises[2];
+
+		const driver_list = (loaded_promises[0].length > 0) ? loaded_promises[0] : loaded_promises[1];
 
 		m = new form.Map('nut_server', _('NUT Server'),
 			_('Network UPS Tools Server Configuration'));
@@ -98,11 +92,6 @@ return view.extend({
 		o.optional = true;
 		o.datatype = 'uinteger'
 		o.placeholder = 24;
-
-		if (have_ssl_support) {
-			o = s.option(form.Value, 'certfile', _('Certificate file (SSL)'));
-			o.optional = true;
-		}
 
 		// Drivers global settings
 		s = m.section(form.NamedSection, 'driver_global', 'driver_global', _('Driver Global Settings'));
--- a/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
+++ b/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
@@ -1,14 +1,29 @@
 'use strict';
 'require form';
+'require fs';
 'require view';
+
+const ups_ssl_backend_file = '/usr/share/nut/ssl_backend';
+const ups_daemon = '/usr/sbin/upsd';
 
 return view.extend({
 	load: function() {
-
+		return Promise.all([
+			fs.trimmed(ups_ssl_backend_file).then(function(backend) {
+				if (backend)
+					return backend;
+				// nut older than 2.8.5-5 does not ship ssl_backend, so probe the binary instead
+				return L.resolveDefault(fs.exec_direct('/usr/bin/ldd', [ups_daemon]), '').then(function(stdout) {
+					return stdout.includes('libssl.so') ? 'old_nut_openssl' : 'none';
+				});
+			}),
+		])
 	},
 
-	render: function() {
+	render: function(loaded_promises) {
 		let m, s, o;
+
+		const ssl_support_type = loaded_promises[0];
 
 		m = new form.Map('nut_server', _('NUT Server'),
 			_('Network UPS Tools Server Configuration'));
@@ -24,6 +39,43 @@ return view.extend({
 		o = s.option(form.Value, 'statepath', _('Path to state file'));
 		o.optional = true;
 		o.placeholder = '/var/run/nut'
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl') {
+			o = s.option(form.FileUpload, 'certfile', _('Server certificate chain and private key'), _('PEM file containing server certificate chain and private key (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'openssl') {
+			// For OpenSSL, certpath is a PEM file, not a database directory as with NSS
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			// For NSS, certpath is a database directory, not a PEM file as with OpenSSL.
+			// We do not fill this field by default as the default database has no certificates,
+			// and must be filled to be useful. Doing this through the UI is a future task.
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'), _('An empty database was created in /etc/nut/cert_db at package install. To use it, SSH in, fill it using certutil, and put /etc/nut/cert_db in this field.'));
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certident', _('Certificate name and password'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has a space in it. An empty password must be entered as an empty pair of double-quotes ("")'));
+			o.placeholder = '"certificate name" "certificate password"';
+			o.password = true;
+
+			o = s.option(form.ListValue, 'certrequest', _('Certificate request level'), _('Whether to request a client certificate and whether to validate a requested client certificate'));
+			o.value('0', _('NO (do not request or verify a client certificate)'));
+			o.value('1', _('REQUEST (any client certificate)'));
+			o.value('2', _('REQUIRE (and verify)'));
+
+			o = s.option(form.Flag, 'disable_weak_ssl', _('Disable weak SSL'), _('Require at least TLSv1.2 for SSL communications'));
+			o.rmempty = false;
+			o.optional = false;
+			o.default = true;
+		}
 
 		return m.render();
 	}
--- a/applications/luci-app-nut/po/templates/nut.pot
+++ b/applications/luci-app-nut/po/templates/nut.pot
@@ -1,101 +1,130 @@
 msgid ""
 msgstr "Content-Type: text/plain; charset=UTF-8"
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:244
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:67
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:65
+msgid ""
+"\"Certificate name\" and \"certificate password\" need to be entered in "
+"double-quotes (\") if the value has a space in it. An empty password must be "
+"entered as an empty pair of double-quotes (\"\")"
+msgstr ""
+
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:233
 msgid "Additional Shutdown Time(s)"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:74
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:68
 msgid "Addresses on which to listen"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:57
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:51
 msgid "Allowed actions"
 msgstr ""
 
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:62
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:60
+msgid ""
+"An empty database was created in /etc/nut/cert_db at package install. To use "
+"it, SSH in, fill it using certutil, and put /etc/nut/cert_db in this field."
+msgstr ""
+
 #: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_cgi.js:21
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:25
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:22
 msgid "As configured by NUT"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:67
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:61
 msgid "Auxiliary"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:69
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:63
 msgid "Auxiliary (Deprecated)"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:183
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:172
 msgid "Bytes to read from interrupt pipe"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:99
-msgid "CA Certificate path"
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:53
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:51
+msgid "CA certificate(s)"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:103
-msgid "Certificate file (SSL)"
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:67
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:65
+msgid "Certificate name and password"
+msgstr ""
+
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:69
+msgid "Certificate request level"
+msgstr ""
+
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:46
+msgid "Client certificate chain and private key"
 msgstr ""
 
 #: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_cgi_admin.js:17
 msgid "Control UPS via CGI"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:113
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:89
 msgid "Custom notification message for message type"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:93
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:79
 msgid "Deadtime"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:114
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:103
 msgid "Default for UPSes without this field."
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:210
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:199
 msgid "Delay for kill power command"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:219
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:208
 msgid "Delay to power on UPS if power returns after kill power"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:172
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:161
 msgid "Description (Display)"
+msgstr ""
+
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:74
+msgid "Disable weak SSL"
 msgstr ""
 
 #: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_cgi.js:33
 msgid "Display name"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:202
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:191
 msgid "Do not lock port when starting driver"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:143
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:132
 msgid "Driver"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:138
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:127
 msgid "Driver Configuration"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:108
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:97
 msgid "Driver Global Settings"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:240
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:229
 msgid "Driver Shutdown Order"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:133
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:266
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:122
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:255
 msgid "Driver waits for data to be consumed by upsd before publishing more."
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:20
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:35
 msgid "Drop privileges to this user"
 msgstr ""
 
@@ -103,41 +132,41 @@ msgstr ""
 msgid "Enable"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:160
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:149
 msgid ""
 "Enables a hotplug script that makes all ttyUSB devices (e.g. serial USB) "
 "group read-write as user %s"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:9
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:6
 msgid "Execute notify command"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:150
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:139
 msgid ""
 "For USB HID based UPSes, USB Product Id is required for NUT to set "
 "permissions so the driver can access the UPS"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:153
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:142
 msgid ""
 "For USB HID based UPSes, USB Vendor Id is required for NUT to set "
 "permissions so the driver can access the UPS"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:156
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:145
 msgid ""
 "For USB HID based UPSes, allows use of multiple USB devices of the same type "
 "(USB product and USB vendor are identical). Other UPSes may also use this "
 "value."
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:60
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:54
 msgid "Forced Shutdown"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:70
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:16
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:56
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:31
 msgid "Global Settings"
 msgstr ""
 
@@ -146,7 +175,7 @@ msgstr ""
 msgid "Go to NUT CGI"
 msgstr ""
 
-#: applications/luci-app-nut/root/usr/share/rpcd/acl.d/luci-app-nut.json:15
+#: applications/luci-app-nut/root/usr/share/rpcd/acl.d/luci-app-nut.json:13
 msgid "Grant admin UCI access for luci-app-nut"
 msgstr ""
 
@@ -158,7 +187,7 @@ msgstr ""
 msgid "Host"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:89
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:75
 msgid "Host Sync"
 msgstr ""
 
@@ -166,77 +195,81 @@ msgstr ""
 msgid "Hostname or IP address"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:28
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:25
 msgid "Hostname or address of UPS"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:78
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:72
 msgid "IP Address"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:144
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:133
 msgid "If this list is empty you need to %s"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:12
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:9
 msgid "Ignore"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:175
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:164
 msgid "Ignore Low Battery"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:63
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:57
 msgid "Instant commands"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:179
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:168
 msgid "Interrupt Only"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:183
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:172
 msgid "Interrupt Size"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:196
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:185
 msgid "Manufacturer (Display)"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:187
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:176
 msgid "Max USB HID Length Reported"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:92
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:86
 msgid "Maximum Age of Data"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:118
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:107
 msgid "Maximum Retries"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:114
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:191
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:103
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:180
 msgid "Maximum Start Delay"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:97
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:91
 msgid "Maximum connections"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:118
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:107
 msgid "Maximum number of times to try starting a driver."
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:128
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:117
 msgid "Maximum time in seconds between refresh of UPS status"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:74
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:60
 msgid "Minimum required number or power supplies"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:199
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:188
 msgid "Model (Display)"
+msgstr ""
+
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:70
+msgid "NO (do not request or verify a client certificate)"
 msgstr ""
 
 #: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_cgi.js:13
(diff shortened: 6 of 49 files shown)


applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -1,15 +1,26 @@
...
 return view.extend({
 	load: function() {
-
+		return Promise.all([
+			L.resolveDefault(fs.read(ups_ssl_backend_file), []).catch(function(err) {
+				throw new Error(_('Unable to read ssl_backend: %s').format(err.message));
+			}).then(function(ssl_type) {
+				return ssl_type;
+			}),
+		])
```
openwrt-ai, line 12: `fs.read()` returns the raw file contents, newline included. A shell-generated `/usr/share/nut/ssl_backend` will almost certainly end in `\n`, so `ssl_support_type == 'openssl'` / `== 'nss'` in `render()` never matches and the whole SSL section silently disappears — the exact failure this PR is meant to fix. [`fs.trimmed()`](https://github.com/openwrt/luci/blob/75d1349be5ab046bcec6d9d4faf4781e3ab2ee96/modules/luci-base/htdocs/luci-static/resources/fs.js#L324-L328) exists for this case and already resolves to `''` on error.

Two more issues in the same block: `L.resolveDefault()` never rejects, so the `.catch()` and its `_('Unable to read ssl_backend: %s')` string are unreachable (that string is now carried in all 48 `.po` files for nothing), and the `[]` default should be a string rather than an array. The `.then()` is an identity function.

The same fix applies to `nut_server_admin.js` lines 10-16.

```suggestion
		return Promise.all([
			fs.trimmed(ups_ssl_backend_file),
		])
```

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -27,6 +38,44 @@ return view.extend({
...
+		if (ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'));
+			o.optional = true;
+			o.default = '/etc/nut/cert_db';
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certident', _('Certificate name and password'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has as space in it. An empty password must be entered as and empty pair of double-quotes ("")'));
+			o.placeholder = '"certificate name" "certificate password"';
+			o.datatype = "tuple(string, string)"
```
openwrt-ai: This datatype rejects exactly the input the help text asks for. [`tuple()`](https://github.com/openwrt/luci/blob/75d1349be5ab046bcec6d9d4faf4781e3ab2ee96/modules/luci-base/htdocs/luci-static/resources/validation.js#L916-L921) splits on `/\s+/` and fails unless the token count equals the type count, and [`string`](``https://github.com/openwrt/luci/blob/75d1349be5ab046bcec6d9d4faf4781e3ab2ee96/modules/luci-base/htdocs/luci-static/resources/validation.js#L1305-L1309``) with no argument always returns true — so the only thing `tuple(string, string)` enforces is "exactly two whitespace-separated tokens". The description one line above tells the user to quote values containing spaces, and `"my cert" ""` splits into three tokens and fails validation.

Same problem for `certhost` on line 76 — `tuple(or(hostname,hostport), string, range(0, 1), range(0, 1))` breaks as soon as the quoted certificate name contains a space — and for the identical `certident` datatype at `nut_server_admin.js:57`. Either drop the datatype or use a validator that understands the quoting.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +36,38 @@ return view.extend({
...
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certident', _('"certificate name" "certificate password"'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has as space in it'));
+			o.placeholder = '"certificate name" "certificate password"';
+			o.datatype = "tuple(string, string)"
+			o.password = true;
+
+			o = s.option(form.ListValue, 'certrequest', _('Certificate request level'), _('Whether to request a client certificate and whether to validate a requested client certifcte'));
+			o.value('0', _('NO (do not request or verify a client certificate'));
```
openwrt-ai: Typo `certifcte` → `certificate`, and the `NO (...` label is missing its closing parenthesis. Both strings are already extracted into the 48 `.po` files, so worth fixing before merge.

```suggestion
			o = s.option(form.ListValue, 'certrequest', _('Certificate request level'), _('Whether to request a client certificate and whether to validate a requested client certificate'));
			o.value('0', _('NO (do not request or verify a client certificate)'));
```

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +36,38 @@ return view.extend({
...
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'));
+			o.default = '/etc/nut/cert_db';
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certident', _('"certificate name" "certificate password"'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has as space in it'));
```
openwrt-ai: nit: `has as space` → `has a space`. Also, this uses the raw placeholder string as the option *title*, while `nut_monitor_admin.js:58` titles the same option `Certificate name and password` — the two admin screens end up labelling the identical setting differently. The suggestion aligns them.

```suggestion
			o = s.option(form.Value, 'certident', _('Certificate name and password'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has a space in it'));
```

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -27,6 +38,44 @@ return view.extend({
...
+		}
+
+		if (ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'));
+			o.optional = true;
+			o.default = '/etc/nut/cert_db';
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certident', _('Certificate name and password'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has as space in it. An empty password must be entered as and empty pair of double-quotes ("")'));
```
openwrt-ai: nit: two typos in the description — `has as space` → `has a space`, and `as and empty pair` → `as an empty pair`.

```suggestion
			o = s.option(form.Value, 'certident', _('Certificate name and password'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has a space in it. An empty password must be entered as an empty pair of double-quotes ("")'));
```

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -27,6 +38,44 @@ return view.extend({
...
 		o.placeholder = '/usr/sbin/nutshutdown'
 
+		if (ssl_support_type == 'openssl') {
+			o = s.option(form.FileUpload, 'certfile', _('Server certificate chain and private key'), _('PEM file containing server certificate chain and private key (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
```
openwrt-ai: nit: this block is byte-for-byte identical to `nut_server_admin.js:40-46`, but this view configures `upsmon.conf` — the client side. Calling `CERTFILE` "Server certificate chain and private key" reads as if the user should upload upsd's key here.

`certpath` looks off too: the in-tree default config documents it as a directory for the monitor (`# option certpath /path/to/ca/dir`, [packages `net/nut/files/nut_monitor:31`](https://github.com/openwrt/packages/blob/112a7e4e9606b1421494cc02377ba5d2951262dc/net/nut/files/nut_monitor#L31)), and the code this PR removes used `form.Value` with a `/etc/ssl/certs` placeholder — a `form.FileUpload` can only produce a single file path under `/etc/luci-uploads`. Is the widget/wording change on the monitor side intentional?

---

applications/luci-app-nut/root/usr/share/rpcd/acl.d/luci-app-nut.json
```diff
@@ -4,15 +4,18 @@
 		"read": {
 			"file": {
 				"/lib/nut/": [ "list" ],
-				"/usr/libexec/nut/": [ "list" ],
-				"/usr/bin/ldd /usr/sbin/upsmon": ["exec"],
-				"/usr/bin/ldd /usr/sbin/upsd": ["exec"]
+				"/usr/libexec/nut/": [ "list" ]
 			},
 			"uci": [ "nut_cgi", "nut_monitor", "nut_server" ]
```
openwrt-ai, line 9: The commit message claims the SSL settings are moved to the admin screens "instead of being readable by a non-privileged user", but this line still grants `luci-app-nut-read` unrestricted `uci` read on `nut_monitor` and `nut_server`. rpcd's uci ACL is per-config-file only — [`rpc_uci_read_access()`](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/uci.c#L311-L320) checks `rpc_session_access(sid, "uci", config, "read")` with no section or option granularity — so a read-only user can still fetch `certident` (the certificate password, `o.password = true` in both admin views) directly over ubus. Moving the form field to a different page hides it in the UI but does not restrict the data.

Either the read group has to stop covering these configs (or the sensitive values need to live in a config the read group isn't granted), or the commit message should drop the security claim.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +36,38 @@ return view.extend({
 		o.optional = true;
 		o.placeholder = '/var/run/nut'
 
+		if (ssl_support_type == 'openssl') {
+			o = s.option(form.FileUpload, 'certfile', _('Server certificate chain and private key'), _('PEM file containing server certificate chain and private key (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
```
openwrt-ai, line 41: Backend consumer check on the upsd side: in packages master, [`nut-server-config.sh.functions:70-82`](https://github.com/openwrt/packages/blob/112a7e4e9606b1421494cc02377ba5d2951262dc/net/nut/files/nut-server-config.sh.functions#L70-L82) reads only `maxage`, `maxconn` and `certfile` from the `upsd` section, so `certpath`, `certident`, `certrequest`, `disable_weak_ssl` and the new `debug_min` in `nut_server.js:96` currently have no consumer and would silently do nothing.

The upsmon side is already covered — `CERTPATH CERTFILE CERTIDENT CERTHOST DEBUG_MIN` and `CERTVERIFY FORCESSL` are all in [`nut-monitor-config.sh.functions:32-34`](https://github.com/openwrt/packages/blob/112a7e4e9606b1421494cc02377ba5d2951262dc/net/nut/files/nut-monitor-config.sh.functions#L32-L34) — so it's only upsd that is missing. Presumably openwrt/packages#29893 adds the upsd handling along with `/usr/share/nut/ssl_backend`; worth confirming, since the `EXTRA_DEPENDS` bump to `nut (>= 2.8.5-5)` is the only thing gating it (packages master is currently at 2.8.5-4).

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +36,38 @@ return view.extend({
...
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'));
+			o.default = '/etc/nut/cert_db';
```
openwrt-ai: nit: the identical NSS `certpath` option in `nut_monitor_admin.js:52-54` sets `o.optional = true`; this one doesn't, so the same setting renders as a required field on the server screen and an optional one on the monitor screen.

```suggestion
			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'));
			o.optional = true;
			o.default = '/etc/nut/cert_db';
```

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js
```diff
@@ -99,10 +93,10 @@ return view.extend({
 		o.datatype = 'uinteger'
 		o.placeholder = 24;
 
-		if (have_ssl_support) {
-			o = s.option(form.Value, 'certfile', _('Certificate file (SSL)'));
-			o.optional = true;
-		}
+		o = s.option(form.Value, 'debug_min', _('Minimum debug level to use'), _('The value present in reloaded configuration files is applied instantly and overrides any previously set value; a missing (or commented away) value however does not change the previously active logging verbosity'));
```
openwrt-ai: nit: `debug_min` here and in `nut_monitor.js:88` is unrelated to the SSL configuration that the subject and body describe — the commit message never mentions it. Either add a line about it to the message or split it into its own commit.

---

openwrt-ai (commented):
Reviewed 1 new commit. Main functional finding: `fs.read()` is not newline-trimmed, so the `ssl_support_type` comparisons in both admin views will not match a shell-written `/usr/share/nut/ssl_backend` and the SSL sections would never render — see the inline comment on `nut_monitor_admin.js`. Also flagged the `tuple()` datatypes rejecting the quoted form the help text prescribes, the unchanged `luci-app-nut-read` uci ACL vs. the commit message's security claim, and the missing upsd-side consumers for several new options. The `.po`/`.pot` regeneration looks clean and CI is still pending on this head, so nothing to report there.

## Commit checks
- 75d1349 "luci-app-nut: add/update SSL configuration for full support" — the message covers the SSL rework, the `ssl_backend` switch and the ldd removal, but not the new `debug_min` option added to `nut_server.js` and `nut_monitor.js`. It also states the SSL settings are no longer "readable by a non-privileged user", which the ACL change does not actually accomplish (see the inline comment on `acl.d/luci-app-nut.json`).

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
danielfdickinson: Added in openwrt/packages#29893 at

https://github.com/danielfdickinson/packages/blob/c0ec3741ce64b8613b8324575f907342303563fc/net/nut/files/nut-server-config.sh.functions#L94-L107

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js
```diff
@@ -51,18 +48,11 @@ function MonitorUserOptions(s) {
...
 	load: function() {
-		return Promise.all([
-			L.resolveDefault(fs.exec_direct('/usr/bin/ldd', [upsmon_tool]), []).catch(function(err) {
-				throw new Error(_('Unable to run ldd: %s').format(err.message));
-			}).then(function(stdout) {
-				return stdout.includes('libssl.so');
-			}),
-		])
+
 	},
```
openwrt-ai: nit: leftover from the removed `ldd` probe — an empty `load()` that overrides the base class with the exact same behaviour. [`LuCI.view.load()`](https://github.com/openwrt/luci/blob/6c73922a0e3fbfa12c91893b9696d5621a1591e4/modules/luci-base/htdocs/luci-static/resources/luci.js#L1995) is already a no-op, so the override (and the stray blank line inside it) can just go.

```suggestion
```

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +32,40 @@ return view.extend({
...
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			/* Update certpath for NSS to reflect its difference from the use of certpath for OpenSSL */
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'));
+			o.optional = true;
+			o.placeholder = '/etc/nut/cert_db';
```
openwrt-ai: Switching `o.default = '/etc/nut/cert_db'` to `o.placeholder` fixes the optional/required parity with the monitor view, but it also changes what the field promises: in LuCI a placeholder reads as "this is the value that applies if you leave the box empty". I can't find any such fallback in the nut package — `/etc/nut/cert_db` appears nowhere in packages master, and [`nut-server-config.sh.functions`](https://github.com/openwrt/packages/blob/d04b36bbba607e65072ecd0abfae5a89af468607/net/nut/files/nut-server-config.sh.functions#L79-L82) / [`nut-monitor-config.sh.functions:32`](https://github.com/openwrt/packages/blob/d04b36bbba607e65072ecd0abfae5a89af468607/net/nut/files/nut-monitor-config.sh.functions#L32) only emit `CERTPATH` when the UCI option is set. With NSS selected and the field left blank, no `CERTPATH` line is written at all and SSL won't come up.

Does openwrt/packages#29893 add `/etc/nut/cert_db` as a default in the config generator? If not, `o.default` (which pre-fills and persists the value) is the semantically correct property here. Same question for `nut_monitor_admin.js:50-52`.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +32,40 @@ return view.extend({
 		o.optional = true;
 		o.placeholder = '/var/run/nut'
 
+		if (ssl_support_type == 'openssl') {
+			o = s.option(form.FileUpload, 'certfile', _('Server certificate chain and private key'), _('PEM file containing server certificate chain and private key (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+
+			/* Update certpath for OpenSSL to reflect its difference from the use of certpath for NSS */
```
openwrt-ai: nit: this reads as a changelog entry about the patch ("Update certpath ...") rather than as a comment about the code, which is what a future reader will see. It's also the only `/* */` comment in a file that otherwise uses `//` (`// Server global settings` on line 23). Stating the fact instead keeps it useful after merge:

```suggestion
		// For OpenSSL, certpath is a PEM file, not a database directory as with NSS
```

Same at line 47 below and at `nut_monitor_admin.js:42` and `:49`.

---

openwrt-ai (commented):
Re-reviewed the force-push (`75d1349` → `6c73922`). All of the previous round's findings are addressed: `fs.trimmed()` replaces the untrimmed `fs.read()` + unreachable `.catch()` in both admin views, the `tuple()` datatypes that rejected the quoted syntax are gone, the `certifcte` / `has as space` / `as and empty pair` typos and the unbalanced `NO (...` label are fixed, the monitor-side `certfile` is now labelled "Client", the NSS `certpath` optional flag is consistent across both admin views, `debug_min` is dropped so the diff matches the subject, and the commit message no longer claims the ACL restricts read access. The `.pot`/`.po` regeneration is consistent with the new sources — no stale `Unable to read ssl_backend` / `Minimum debug level` / `ldd` strings remain. CI is green on this head, so nothing to report there.

Three remaining items inline, all minor: a leftover empty `load()` override in `nut_monitor.js`, a question about `placeholder` vs `default` for the NSS `certpath`, and a comment-style nit on the new `/* Update certpath ... */` lines.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +32,40 @@ return view.extend({
...
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			// For NSS, certpath is a database directory, not a PEM file as with OpenSSL
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'));
+			o.optional = true;
+			o.default = '/etc/nut/cert_db';
```
openwrt-ai: Correction to my previous round — I said `o.default` "pre-fills and persists the value". The pre-fill part is right, the persist part is not, so this swap doesn't change what lands in UCI.

[`AbstractValue.parse()`](https://github.com/openwrt/luci/blob/3a6958f5ed975150e2c95f875e480c46575fc469/modules/luci-base/htdocs/luci-static/resources/form.js#L2152-L2155) *removes* the option whenever `fval == this.default && (this.optional || this.rmempty)`, and [`rmempty` defaults to `true`](https://github.com/openwrt/luci/blob/3a6958f5ed975150e2c95f875e480c46575fc469/modules/luci-base/htdocs/luci-static/resources/form.js#L1431) on top of the explicit `o.optional = true` here. So the field renders pre-filled with `/etc/nut/cert_db`, the user saves, and `certpath` is still absent from the config — the same end state as `placeholder`, but now with a UI that looks like it is set. It also means an existing explicit `option certpath '/etc/nut/cert_db'` gets deleted on the next save of this form.

On packages master, `CERTPATH` falls through to the `*)` branch and is written only when the UCI option is non-empty ([`upsmon_conf_get_write()` returns 2 and writes nothing otherwise](https://github.com/openwrt/packages/blob/a56b754b01d85d1f6f7b98d7f0adbd9d508f2601/net/nut/files/nut-monitor-config.sh.functions#L83-L109)), so with NSS selected and the box untouched no `CERTPATH` line reaches `upsmon.conf`/`upsd.conf`.

If openwrt/packages#29893 makes the config generator fall back to `/etc/nut/cert_db` when the option is absent, `placeholder` was the correct property and this should go back to it. If it does not, the value has to actually be written, which requires both flags off:

```suggestion
			o.optional = false;
			o.rmempty = false;
			o.default = '/etc/nut/cert_db';
```

Same applies to `nut_monitor_admin.js:51-52`.

---

openwrt-ai (commented):
Re-reviewed the force-push (`6c73922` → `3a6958f`). The delta is three files / six lines and it addresses all three items from the last round: the empty `load()` override in `nut_monitor.js` is gone, the `/* Update certpath ... */` changelog-style comments are now `//` statements of fact matching the file's existing style, and the NSS `certpath` `placeholder` became `default` on both admin views. The commit message matches the diff. CI is green on this head (build x86_64, eslint, CodeQL, FormalityCheck all passing), so nothing to report there.

One item inline, and it is a correction of my own previous advice: `o.default` combined with `o.optional = true` does not persist the value either — `form.js` explicitly removes the option when the form value equals the default. So the `placeholder` → `default` swap changes only the rendering, not what ends up in UCI.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -27,6 +34,45 @@ return view.extend({
...
+		}
+
+		if (ssl_support_type == 'nss') {
+			// For NSS, certpath is a database directory, not a PEM file as with OpenSSL.
+			// We do not fill this field by default as the default database has no certificates,
+			// and must be filled to be useful. Doing this through the UI is a future task.
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'), _('An empty database was created in /etc/nut/cert_db at package install. To use it SSH in, fill it using certutil, and put /etc/nut/cert_db in this field.'));
```
openwrt-ai: nit: "To use it SSH in, fill it using certutil" is missing the comma after the introductory clause, so it parses as "use it SSH" on first read. Worth fixing now rather than later — this string has just been extracted into all 45 `.po` files, and any later edit re-churns every one of them.

```suggestion
			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'), _('An empty database was created in /etc/nut/cert_db at package install. To use it, SSH in, fill it using certutil, and put /etc/nut/cert_db in this field.'));
```

The same string is duplicated at [`nut_server_admin.js:50`](``https://github.com/openwrt/luci/blob/e37d312f51a576ac5dc24ec2daa1b927c4c976fa/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js#L50)`` and needs the identical change, otherwise the two views produce two distinct msgids for the same help text.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +32,41 @@ return view.extend({
...
+			o.optional = true;
+
+			// For OpenSSL, certpath is a PEM file, not a database directory as with NSS
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			// For NSS, certpath is a database directory, not a PEM file as with OpenSSL
```
openwrt-ai: nit: this round added the sentence-terminating period to the twin of this line in [`nut_monitor_admin.js:49`](``https://github.com/openwrt/luci/blob/e37d312f51a576ac5dc24ec2daa1b927c4c976fa/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js#L49)`` but not here, so the two otherwise byte-identical comment blocks now differ by one character.

```suggestion
			// For NSS, certpath is a database directory, not a PEM file as with OpenSSL.
```

---

openwrt-ai (commented):
Re-reviewed the delta since `3a6958f` — the branch was rebased onto `98417a6` and the single commit force-pushed to `e37d312`. Outside the `.po`/`.pot` regeneration the change is seven lines across the two admin views.

The previous round's item is resolved, and resolved correctly: rather than trying to make `default`/`placeholder` persist `/etc/nut/cert_db`, the NSS `certpath` field is now left genuinely empty with the rationale spelled out in an inline comment ("the default database has no certificates, and must be filled to be useful") and a help string telling the user how to populate it. That sidesteps the `parse()`-removes-value-equal-to-default behaviour entirely, so there is nothing left to flag there.

Two things I checked specifically because of the rebase:

- The rebase base `98417a6` added `+USE_GLIBC:ldd` to `LUCI_DEPENDS`, which this PR's own removal of the `ldd` probe makes dead. The rebase handled it — `98417a6..e37d312` drops that token from the Makefile, and no `ldd` reference survives anywhere in `applications/luci-app-nut/` (Makefile, ACL, or views).
- The `.pot` regeneration is consistent with the sources: the new `An empty database was created…` msgid is present with correct `nut_monitor_admin.js:52` / `nut_server_admin.js:50` references, the shifted `nut_monitor.js` line numbers now match the file after the `load()` removal from the previous round, and all 44 `.po` files carry the same reference updates with no translations misattached.

Commit message is unchanged from the last round and still matches the diff. GitHub reports no check runs on this head yet, so nothing to report from CI.

Two nits inline, both cosmetic and both in strings/comments introduced by this delta.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -27,6 +34,45 @@ return view.extend({
...
+			o = s.option(form.FileUpload, 'certpath', _('CA certificate(s)'), _('PEM file containing the CA certificate(s) (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'nss') {
+			// For NSS, certpath is a database directory, not a PEM file as with OpenSSL.
+			// We do not fill this field by default as the default database has no certificates,
+			// and must be filled to be useful. Doing this through the UI is a future task.
+			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'), _('An empty database was created in /etc/nut/cert_db at package install. To use, it SSH in, fill it using certutil, and put /etc/nut/cert_db in this field.'));
```
openwrt-ai: The comma landed one word early: "To use, it SSH in, fill it using certutil" has no subject-verb agreement — "it SSH in" is not a clause. The intent (from the previous round) was "To use it, SSH in, ..." — the introductory clause is "To use it", and "SSH in" is the imperative that follows, matching the two imperatives after it ("fill it", "put ... in this field").

```suggestion
			o = s.option(form.Value, 'certpath', _('Path to NSS certificate and key databases'), _('An empty database was created in /etc/nut/cert_db at package install. To use it, SSH in, fill it using certutil, and put /etc/nut/cert_db in this field.'));
```

The identical string is at [`nut_server_admin.js:50`](``https://github.com/openwrt/luci/blob/2218fe0d4e32d1aa641d567d2a5ce17d7aa95ced/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js#L50)`` and needs the same edit, otherwise the two views emit two distinct msgids for the same help text. Since this is the second re-churn of the same msgid across all 44 `.po` files, worth getting right in one pass.

---

openwrt-ai (commented):
Re-reviewed the delta since `e37d312` (force-pushed to `2218fe0d`). Outside the `.po`/`.pot` regeneration the change is three lines across the two admin views — both items from the last round.

- The comment-parity nit is fixed correctly: `nut_server_admin.js:41` now carries the same sentence-terminating period as its twin in `nut_monitor_admin.js:43`, and the two blocks are byte-identical again.
- The grammar nit is *not* fixed correctly — the comma was inserted one word to the left of where it belongs, producing "To use, it SSH in", which is worse than the original. Inline comment with the suggestion; it needs applying in both views.

The `.pot`/`.po` regeneration is consistent with the (new) source string: the msgid is rewrapped in the template and in all 43 `.po` files with the same `nut_monitor_admin.js:52` / `nut_server_admin.js:50` references, no `msgstr` was left attached to the old msgid, and no copy of the previous wording survives anywhere under `applications/luci-app-nut/`.

Commit message is unchanged from the last round and still matches the diff. CI is green on this head (build x86_64, eslint, CodeQL, all three FormalityCheck jobs), so nothing to report there.

---

openwrt-ai (commented):

Re-reviewed the delta since `2218fe0d` (force-pushed to `7b479dc`). Outside the `.po`/`.pot` regeneration the change is one line, duplicated in the two admin views, and it is the grammar nit from the last round — now fixed correctly: "To use, it SSH in" → "To use it, SSH in". Applied identically in [`nut_monitor_admin.js:52`](``https://github.com/openwrt/luci/blob/7b479dcbea9d183fa3a3843ed348f0760809919d/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js#L52)`` and [`nut_server_admin.js:50`](``https://github.com/openwrt/luci/blob/7b479dcbea9d183fa3a3843ed348f0760809919d/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js#L50),`` and the surrounding NSS `certpath` blocks in the two views remain byte-identical.

The `.pot`/`.po` regeneration is consistent with the new source string: the msgid is rewrapped in the template and in all 42 `.po` files, every file carries exactly the same three removed / two added msgid lines and no other change (no `msgstr` line was touched anywhere in the delta), the `nut_monitor_admin.js:52` / `nut_server_admin.js:50` references are unchanged, and no copy of the old wording survives anywhere under `applications/luci-app-nut/`.

Commit message is unchanged from the last round and still matches the diff. CI is green on this head (build x86_64, eslint, CodeQL, all three FormalityCheck jobs).

---

danielfdickinson:
Now that https://github.com/openwrt/packages/pull/29893 has landed, this PR is relevant / ready for review.

danielfdickinson:
Oh and I see there is a conflict to resolve.

danielfdickinson:
@BKPepe Found another thing to you may (or may not) want to update in your bot - LUCI_EXTRA_DEPENDS needs to be of the form `package (>=1.2.3-r4)`, with appropriate substitutions allowed. I had a failed build because I had the package version format wrong and space after >=. Obviously not urgent, but I though I'd mention it as a nice to have 'lint' of the LuCI Makefile.

openwrt-ai (commented):

Re-reviewed the delta since `7b479dc` (rebased from `98417a6` onto `77dad3f` and force-pushed to `8196c15`). The commit's file list is byte-for-byte the same as the last round except for two hunks, and the commit message is unchanged.

**Makefile** — `EXTRA_DEPENDS:=nut (>= 2.8.5-5)` → `LUCI_EXTRA_DEPENDS:=nut (>=2.8.5-r5)`. Both halves of this check out:

- The variable rename is a real fix, not cosmetic. `EXTRA_DEPENDS` is only ever emitted from inside the `Package/$(PKG_NAME)` define, via [`luci.mk:166`](https://github.com/openwrt/luci/blob/8196c1515c162d3e6411748cf36e1fc8ded352bb/luci.mk#L166) — a bare `EXTRA_DEPENDS:=` set before `include ../../luci.mk` is just a global make variable and never reaches the package metadata, so the constraint was inert before this. `luci-app-adguardhome` is the only other in-tree user of the working form and it matches (`LUCI_EXTRA_DEPENDS:=adguardhome (>=0.107.73-r3)`); `luci-app-radicale3` still carries the inert `EXTRA_DEPENDS` spelling.
- The `-r5` version format is right. `include/package-defaults.mk` builds the package version as `VERSION:=$(PKG_VERSION)-r$(PKG_RELEASE)`, and `net/nut/Makefile` in packages master is now `PKG_VERSION:=2.8.5` / `PKG_RELEASE:=5`, so `2.8.5-r5` is exactly what the constraint has to compare against. The previous `2.8.5-5` would not have matched.

Now that openwrt/packages#29893 has landed I also re-ran the backend-consumer check that was open from the first round — every UCI option the two admin views write has a consumer in the merged package: `CERTFILE`, `CERTPATH`, `CERTIDENT`, `CERTHOST`, `CERTVERIFY`, `FORCESSL` are all handled in `nut-monitor-config.sh.functions`, and `CERTFILE`, `CERTPATH`, `CERTIDENT`, `CERTREQUEST`, `DISABLE_WEAK_SSL` in `nut-server-config.sh.functions`, each gated on the same `ssl_backend` values (`openssl` / `nss`) the views gate their sections on. `/usr/share/nut/ssl_backend` is installed by the package Makefile and is readable through the `luci-app-nut-admin` ACL. Nothing frontend-only left.

**po/sv/nut.po** — the rebase pulled in a Weblate refresh of the Swedish catalogue, so this file now carries both the upstream retranslations and this commit's regeneration. I checked the merge did not damage anything:

- The `.pot` is exactly consistent with the head sources: all 137 source references resolve to an existing file and an in-range line, and for all 149 entries the referenced line actually contains the msgid.
- All 44 `.po` files carry precisely the `.pot`'s 149 msgids — no orphaned entries, no missing ones, in any language.
- The only `msgstr` values dropped in the Swedish file are the ones whose msgids this commit removes or rewords (`Sökväg till CA-certifikat`, `Certifikatfil (SSL)`, `Sökväg med CA-certifikat…`, `Kräv SSL och kontrollera…`, `Det gick inte att köra ldd: %s`). No live translation was lost and no fuzzy marker was introduced.
- `ldd` survives only as `#~`-commented obsolete entries in a handful of catalogues, which is normal msgmerge output; no `ldd` reference remains in the Makefile, ACL, or views.

GitHub reports no check runs on this head yet, so nothing to report from CI. `mergeable_state` is `clean`, so the conflict you spotted is resolved.

---

BKPepe:
> @BKPepe Found another thing to you may (or may not) want to update in your bot - LUCI_EXTRA_DEPENDS needs to be of the form `package (>=1.2.3-r4)`, with appropriate substitutions allowed. I had a failed build because I had the package version format wrong and space after >=. Obviously not urgent, but I though I'd mention it as a nice to have 'lint' of the LuCI Makefile.

It was caught by @hnyman in https://github.com/openwrt/luci/pull/8904#issuecomment-5201666176, it needs to be discussed first, though.

danielfdickinson:
> It was caught by @hnyman in [#8904 (comment)](https://github.com/openwrt/luci/pull/8904#issuecomment-5201666176)

Thanks for pointing me to that discussion.

In this case, though, it's actually about the format of EXTRA_DEPENDS (might apply beyond LuCI); It would be nice if the bot caught a wrong EXTRA_DEPENDS format (I originally had `package (>= 1.2.3-4)` - the build does not like either the bare `-4` (needs to be `-r4`) nor the space between `>=` and the version.

openwrt-ai (commented):

Re-reviewed the delta since `8196c15` (force-pushed to `abed243`). This round is a pure rebase: `git range-diff 8196c15^..8196c15 abed243^..abed243` reports the two commits as identical (`1: 8196c15 = 1: abed243`), so the patch content and the commit message are byte-for-byte what I reviewed last time. The base moved from `77dad3f` to `459941f`, and `git diff 77dad3f..459941f -- applications/luci-app-nut/` is empty, so nothing upstream touched this package in between and there was no content to re-merge.

Checks I re-ran against the new head rather than assuming they carried over:

- **`.pot` ↔ sources.** All 176 source references in `po/templates/nut.pot` resolve to a file that exists at `abed243` and to an in-range line, and for every entry the referenced line actually contains the msgid text. No drift from the rebase.
- **`.pot` ↔ catalogues.** All 42 `.po` files carry exactly the template's 149 msgids — no orphaned entries, none missing, in any language. Obsolete `#~` entries were excluded from the comparison, as msgmerge is expected to leave those.
- **Makefile dependency constraint.** `LUCI_EXTRA_DEPENDS:=nut (>=2.8.5-r5)` still matches `net/nut/Makefile` in packages master (`PKG_VERSION:=2.8.5` / `PKG_RELEASE:=5`), and [`luci.mk:166`](https://github.com/openwrt/luci/blob/abed243659240db13f74fdf16325f6ad99ddff76/luci.mk#L166) still emits it into the package metadata unchanged on the new base, so the constraint remains live.

CI is green on this head — build x86_64, eslint, CodeQL (both runs), and all three FormalityCheck jobs. `mergeable_state` is `clean`.

---

danielfdickinson:
In order to keep this working with 25.12.x I am going to rework it a bit to include detecting whether we are old SSL (not /usr/share/nut/ssl_backend) or new and adjust accordingly.

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -1,15 +1,32 @@
...
+			L.resolveDefault(fs.list().contains(ups_ssl_backend_dir) ? fs.trimmed(ups_ssl_backend_file) : fs.exec_direct('/usr/bin/ldd', [upsmon_daemon]), []).catch(function(err) {
+				throw new Error(_('Unable to run ldd: %s').format(err.message));
+			}).then(function(stdout) {
+				if (stdout.includes('libssl.so')) {
+					return "openssl"
+				} else {
+					return "none"
+				};
+			}),
+		])
```
openwrt-ai: The new backward-compat probe does not work; four separate problems in this block, all of which the single suggestion below fixes.

1. **`fs.list().contains(...)` throws a `TypeError` synchronously**, so `load()` rejects and the page never renders. [`fs.list(path)`](https://github.com/openwrt/luci/blob/c2fbac7c964f2d4fb3a182debea090da179dbc0c/modules/luci-base/htdocs/luci-static/resources/fs.js#L179-L181) returns a `Promise`, and `Promise` has no `contains` method. Even awaited it would not help: it resolves to an array of `FileStatEntry` objects (not path strings), and `Array` has `includes`, not `contains`.

2. **The `ssl_backend` file's value is discarded.** Even with (1) fixed, the `.then()` maps the result through `stdout.includes('libssl.so')`, so `fs.trimmed()` returning `'nss'` or `'openssl'` both fall through to `'none'`. The NSS branch in `render()` becomes unreachable and the OpenSSL branch is only reachable via the `ldd` fallback — i.e. the `ssl_backend` mechanism this commit is built around never takes effect. The `libssl.so` test must apply only to the `ldd` output.

3. **The `.catch()` is unreachable.** [`L.resolveDefault`](https://github.com/openwrt/luci/blob/c2fbac7c964f2d4fb3a182debea090da179dbc0c/modules/luci-base/htdocs/luci-static/resources/luci.js#L3117-L3119) is `Promise.resolve(value).catch(() => defvalue)`, so it has already swallowed the rejection by the time `.catch()` runs — the `Unable to run ldd: %s` error can never be raised. (This is the same construct that was removed in the `75d1349` → `6c73922` round.)

4. **`ups_ssl_backend_dir` is wrong and unusable.** `/usr/share/nut/ssl` is not a path the nut package creates ([`net/nut/Makefile:139`](https://github.com/openwrt/packages/blob/ff5519f75032bcbcfe9db97a49c96f6ba5d65e89/net/nut/Makefile#L139) installs `/usr/share/nut/ssl_backend`), and `nut_server_admin.js:6` sets the same constant to `/usr/share/nut` instead — the two views disagree. Separately, a directory check would need a `list` ACL: rpcd gates both `file.list` and `file.stat` on the `list` permission ([`file.c:664`](https://github.com/openwrt/rpcd/blob/06b7d17b6dee9e9e97654dfd3ad5b1cfe22f7d43/file.c#L664), [`file.c:734`](https://github.com/openwrt/rpcd/blob/06b7d17b6dee9e9e97654dfd3ad5b1cfe22f7d43/file.c#L734)), and `acl.d/luci-app-nut.json` grants only `read` on the file. Probing with `fs.trimmed()` — which is documented to resolve to `''` rather than reject — avoids needing any new ACL entry.

```suggestion
const ups_ssl_backend_file = '/usr/share/nut/ssl_backend';
const upsmon_daemon = '/usr/sbin/upsmon';

return view.extend({
	load: function() {
		return Promise.all([
			fs.trimmed(ups_ssl_backend_file).then(function(backend) {
				if (backend)
					return backend;

				// nut older than 2.8.5-5 does not ship ssl_backend, so probe the binary instead
				return L.resolveDefault(fs.exec_direct('/usr/bin/ldd', [upsmon_daemon]), '').then(function(stdout) {
					return stdout.includes('libssl.so') ? 'openssl' : 'none';
				});
			}),
		])
```


---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -1,15 +1,32 @@
...
+			L.resolveDefault(fs.list().contains(ups_ssl_backend_dir) ? fs.trimmed(ups_ssl_backend_file) : fs.exec_direct('/usr/bin/ldd', [ups_daemon]), []).catch(function(err) {
+				throw new Error(_('Unable to run ldd: %s').format(err.message));
+			}).then(function(stdout) {
+				if (stdout.includes('libssl.so')) {
+					return "openssl"
+				} else {
+					return "none"
+				};
+			}),
+		])
```
openwrt-ai: Same four defects as the twin block in `nut_monitor_admin.js:6-22` — see that comment for the analysis. Note additionally that `ups_ssl_backend_dir` is `'/usr/share/nut'` here but `'/usr/share/nut/ssl'` in the monitor view, so the two views would not even have agreed on what to probe.

```suggestion
const ups_ssl_backend_file = '/usr/share/nut/ssl_backend';
const ups_daemon = '/usr/sbin/upsd';

return view.extend({
	load: function() {
		return Promise.all([
			fs.trimmed(ups_ssl_backend_file).then(function(backend) {
				if (backend)
					return backend;

				// nut older than 2.8.5-5 does not ship ssl_backend, so probe the binary instead
				return L.resolveDefault(fs.exec_direct('/usr/bin/ldd', [ups_daemon]), '').then(function(stdout) {
					return stdout.includes('libssl.so') ? 'openssl' : 'none';
				});
			}),
		])
```


---

applications/luci-app-nut/po/templates/nut.pot
```diff
@@ -295,270 +328,318 @@ msgstr ""
...
+msgid "Path to NSS certificate and key databases"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:24
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js:31
 msgid "Path to state file"
 msgstr ""
 
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:92
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js:71
```
openwrt-ai: The `.pot`/`.po` files were not regenerated after this round's `load()` rework, so the catalogues no longer match the sources in two ways.

**Stale line references.** Re-adding the `ldd` fallback grew both `load()` bodies by exactly 10 lines, and every reference into the two admin views is now off by that amount. This one points at line 71, but `Per-host override for certident, certverify, and forcessl` is at [`nut_monitor_admin.js:81`](``https://github.com/openwrt/luci/blob/c2fbac7c964f2d4fb3a182debea090da179dbc0c/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js#L81).`` Checking all 176 references in this template against the head sources, 40 are wrong — all of them the `nut_monitor_admin.js` / `nut_server_admin.js` ones, each short by 10 (e.g. `NUT Monitor` is listed at `:20`, actually at `:30`; `Certificate request level` at `nut_server_admin.js:59`, actually `:69`). Every other file's references still resolve correctly. The same stale references are replicated across all 42 `.po` files.

**Missing msgid.** `_('Unable to run ldd: %s')`, newly reintroduced at `nut_monitor_admin.js:14` and `nut_server_admin.js:14`, has no entry in this template at all — it is the only source string of the 141 in the package that is absent. (It also cannot currently be reached; see the `.catch()` point in the `nut_monitor_admin.js` comment. If that path is dropped rather than fixed, the string goes away and only the line references need refreshing.)

Please regenerate the template and the catalogues once the view sources have settled.

---

applications/luci-app-nut/Makefile
```diff
@@ -12,7 +12,7 @@ PKG_LICENSE:=Apache-2.0
 PKG_MAINTAINER:=Daniel F. Dickinson <dfdpublic@wildtechgarden.ca> \
 		Paul Donald <newtwen+github@gmail.com>
 
-EXTRA_DEPENDS:=nut (>= 2.8.4-3)
+LUCI_EXTRA_DEPENDS:=nut (>=2.8.4-r3)
```
openwrt-ai, line 16: Relaxing this from `nut (>=2.8.5-r5)` back to `nut (>=2.8.4-r3)` makes the 25.12.x compatibility goal work for the *detection* side, but it also lets this UI be installed against a nut whose config scripts do not consume most of what the SSL sections write. On 2.8.4-r3 there is no `ssl_backend` file, the `ldd` probe reports `openssl`, and `render()` then exposes `certpath`, `certident`, `certrequest` and `disable_weak_ssl` — but by this PR's own description the server-side `CERTPATH` handling arrives only with openwrt/packages#29893, i.e. with 2.8.5-5 ([`nut-server-config.sh.functions:94-107`](https://github.com/openwrt/packages/blob/ff5519f75032bcbcfe9db97a49c96f6ba5d65e89/net/nut/files/nut-server-config.sh.functions#L94-L107)). Those options would silently do nothing on the older package.

Did you confirm which of the new options 2.8.4-r3's `nut-server-config.sh` / `nut-monitor-config.sh` actually emit? If it is only `certfile`/`certverify`/`forcessl`, it may be worth gating the rest on the `ssl_backend` file being present (i.e. treat "detected via `ldd`" as a distinct, narrower capability level) rather than showing the full 2.8.5 option set.

---

openwrt-ai (commented):
Re-reviewed the delta since `abed243` (force-pushed to `c2fbac7`). This round implements the backward-compatibility rework you described in [your comment of 26 Aug](https://github.com/openwrt/luci/pull/8926#issuecomment-5424531895): the `ldd` probe comes back alongside the `ssl_backend` file, `+USE_GLIBC:ldd` is restored in `LUCI_DEPENDS`, the two `ldd` exec entries are back in the `luci-app-nut-admin` ACL, `LUCI_EXTRA_DEPENDS` drops from `nut (>=2.8.5-r5)` to `nut (>=2.8.4-r3)`, and the `certhost` help string is rewrapped. The commit message was updated to match, and it does match the diff.

The detection logic itself does not work, though. `fs.list().contains(...)` throws synchronously (`fs.list()` returns a `Promise`; `Promise` has no `contains`), so `load()` rejects and neither admin page renders. Behind that, the `.then()` funnels the `ssl_backend` file contents through a `libssl.so` substring test, so `'nss'` and `'openssl'` both come out as `'none'` — the mechanism the commit is built around never selects anything. The `.catch()` after `L.resolveDefault()` is also unreachable again (it was removed for this reason in the `75d1349` → `6c73922` round), and `ups_ssl_backend_dir` is set to two different, nonexistent paths in the two views. All four are in one region and one suggestion block fixes them; details inline on `nut_monitor_admin.js`, with the twin suggestion on `nut_server_admin.js`.

`eslint` and CodeQL pass because none of this is a syntax or taint issue — the `TypeError` is a runtime property access.

Two further items inline: the `.pot`/`.po` files predate this round's `load()` growth (40 of 176 references in the template point 10 lines short, replicated across all 42 catalogues, and the reintroduced `Unable to run ldd: %s` msgid is missing entirely), and a question about whether the relaxed `nut (>=2.8.4-r3)` constraint leaves several of the SSL options without a backend consumer on the older package.

Everything else in the delta checks out: the `certhost` help-string rewording is applied consistently in both views and in every catalogue, and no other source string drifted. CI is green on this head (build x86_64, eslint, CodeQL, all three FormalityCheck jobs).

---

danielfdickinson:
As this no longer has a hard dependency on a not yet merged PR, and issues fixed, taking off draft.

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -27,6 +42,51 @@ return view.extend({
 		o.optional = true;
 		o.placeholder = '/usr/sbin/nutshutdown'
 
+		if (ssl_support_type == 'openssl') {
+			o = s.option(form.FileUpload, 'certfile', _('Client certificate chain and private key'), _('PEM file containing client certificate chain and private key (OpenSSL)'));
```
openwrt-ai, line 46: `certfile` is the only shared option whose gate differs between the two admin views. Its enclosing `if` (line 45, just above this hunk) is still `ssl_support_type == 'openssl'`, whereas the server view gates the same option on `ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl'` (`nut_server_admin.js:43`). Every other option shared by the two views — `certpath`, `certident`, `certverify`/`forcessl` — was widened symmetrically in this round, so this reads as a missed line: on an old-nut (ldd-detected) build the monitor page offers the CA file at line 53 but no client certificate/key to pair with it.

Is the asymmetry deliberate — i.e. does nut 2.8.4-r3's `nut-monitor-config.sh.functions` genuinely not map `certfile`, even though the server-side script of the same version does? If not, line 45 wants `|| ssl_support_type == 'old_nut_openssl'` too.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
```diff
@@ -27,6 +42,51 @@ return view.extend({
...
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'nss') {
+			o = s.option(form.Value, 'certident', _('Certificate name and password'), _('"Certificate name" and "certificate password" need to be entered in double-quotes (") if the value has a space in it. An empty password must be entered as an empty pair of double-quotes ("")'));
+			o.placeholder = '"certificate name" "certificate password"';
+			o.password = true;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl' || ssl_support_type == 'nss') {
```
openwrt-ai, line 72: `old_nut_openssl` here exposes `forcessl`, which the pre-PR ldd-gated block never offered — that block only had [`certpath` and `certverify` (nut_monitor.js:98-106)](``https://github.com/openwrt/luci/blob/7a6ff759fac36016accddba32fe776d02ed25fd7/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js#L98-L106).`` The same happens on the server side, where `old_nut_openssl` now also gets `certpath` while pre-PR only [`certfile` (nut_server.js:102-103)](``https://github.com/openwrt/luci/blob/7a6ff759fac36016accddba32fe776d02ed25fd7/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js#L102-L103)`` was shown.

Since `LUCI_EXTRA_DEPENDS` now accepts `nut (>=2.8.4-r3)`, `old_nut_openssl` can be any package release from 2.8.4-r3 up to (but not including) 2.8.5-5. I can confirm the consumers exist in packages master (`NUT_UPSMON_OPTIONS`/`NUT_UPSMON_BOOL_OPTIONS` in `nut-monitor-config.sh.functions` handle `CERTPATH`/`CERTVERIFY`/`FORCESSL`), but that is the 2.8.5-5 script — I can't see the 2.8.4-r3 one from here. Did you verify that the older `nut-monitor-config.sh.functions` maps `forcessl`, and the older `nut-server-config.sh.functions` maps `certpath`? If either is unhandled on the old package, the option is frontend-only there and the `old_nut_openssl` set should shrink to what the pre-PR views exposed.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js
```diff
@@ -25,6 +40,43 @@ return view.extend({
 		o.optional = true;
 		o.placeholder = '/var/run/nut'
 
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl') {
+			o = s.option(form.FileUpload, 'certfile', _('Server certificate chain and private key'), _('PEM file containing server certificate chain and private key (OpenSSL)'));
+			o.rmempty = true;
+			o.optional = true;
+		}
+
+		if (ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl') {
```
openwrt-ai: nit: this closes an `if` and immediately reopens one with a byte-identical condition (`ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl'`), so the split has no effect. The monitor view needs two blocks here because its conditions genuinely differ; on the server side they don't. Dropping the seam keeps `certfile` and `certpath` in one block:

```suggestion
```


---

openwrt-ai (commented):
Re-reviewed the delta since `c2fbac7` (force-pushed to `b18655b`). Outside the `.po`/`.pot` regeneration the change is 23 added / 19 removed lines across the two admin views, and it addresses the main finding from the last round.

**The detection logic is now correct.** `fs.list().contains(...)` is gone; `fs.trimmed(ups_ssl_backend_file)` is called unconditionally and, per [`fs.js:324`](https://github.com/openwrt/luci/blob/b18655be695f60ce3404597c1cb942360f7554dd/modules/luci-base/htdocs/luci-static/resources/fs.js#L324), it wraps its own `read()` in `L.resolveDefault(..., '')` and is documented as never rejecting — so a missing `ssl_backend` resolves to `''` and falls through to the `ldd` probe rather than killing `load()`. The `libssl.so` substring test no longer swallows the file contents (`'nss'` and `'openssl'` are now returned verbatim), the unreachable `.catch()` after `L.resolveDefault()` is gone along with the `Unable to run ldd: %s` msgid, and the two bogus `ups_ssl_backend_dir` constants are removed. `L.resolveDefault(fs.exec_direct(...), '')` on the fallback path means a missing/denied `ldd` yields `'none'` instead of a rejection, and the ACL still carries both `ldd` exec entries plus `/usr/share/nut/ssl_backend` read under `luci-app-nut-admin`.

**The `.pot`/`.po` staleness from the last round is fixed.** I re-verified against the head sources: all 176 references in `po/templates/nut.pot` resolve to an existing file and an in-range line, and for all 149 msgids the referenced line actually contains the string (the three apparent mismatches in my check were my own `\"` unescaping, not drift). All 42 `.po` files carry exactly the template's 149 msgids — none missing, none orphaned, obsolete `#~` entries excluded. No `ldd` msgid survives in any catalogue as a live entry, and no `ldd` reference remains in the views beyond the intentional `exec_direct` call.

The commit message is unchanged and still matches the diff, including the "we keep the `ldd` method ... for now" paragraph. CI is green on this head (build x86_64, eslint, CodeQL, Analyze JavaScript and JSON, all three FormalityCheck jobs).

Three items inline, all about the new `old_nut_openssl` gating rather than the detection itself: a `certfile` gate that was widened in the server view but not the monitor view, a question about whether nut 2.8.4-r3 actually consumes the two options newly exposed under `old_nut_openssl`, and a nit on a redundant `if` split.

---

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
danielfdickinson: The asymmetry is deliberate. Upstream versions of NUT prior to 2.8.5 were asymmetric between the SSL options exposed for upsd and upsmon, as well as being asymmetric by backend, even more so than 2.8.5.

applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js
danielfdickinson: I have confirmed that 2.8.4-r3 consumes certpath / certverify / forcessl but not certfile _for upsmon_, and consumes only certfile _for upsd_.

openwrt-ai (commented):

Re-reviewed the delta since `b18655b` (force-pushed to `dc323e6`). The rebase base is unchanged (`7a6ff75` for both), and `git diff b18655b..dc323e6` is a single line in `nut_server_admin.js`: the `certpath` gate narrows from `ssl_support_type == 'openssl' || ssl_support_type == 'old_nut_openssl'` to `ssl_support_type == 'openssl'`.

That is the right resolution of the two open questions from the last round, and it lines up exactly with the facts you supplied in [#discussion_r3878861505](https://github.com/openwrt/luci/pull/8926#discussion_r3878861505) — 2.8.4-r3 consumes `certpath`/`certverify`/`forcessl` but not `certfile` for upsmon, and only `certfile` for upsd. Checking the whole `old_nut_openssl` surface against that at head:

- **Server** ([`nut_server_admin.js`](``https://github.com/openwrt/luci/blob/dc323e659c32f2e2f12eccd4a879ff92622ad8de/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server_admin.js#L43-L54))`` — `old_nut_openssl` now reaches only `certfile` (line 43). `certpath` (49), `certident`/`certrequest`/`disable_weak_ssl` (64) are all excluded. Matches "only `certfile` for upsd" exactly.
- **Monitor** ([`nut_monitor_admin.js`](``https://github.com/openwrt/luci/blob/dc323e659c32f2e2f12eccd4a879ff92622ad8de/applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor_admin.js#L45-L88))`` — `old_nut_openssl` reaches `certpath` (51) and `certverify`/`forcessl` (72), and is excluded from `certfile` (45), `certident` (66) and `certhost` (84). Matches the upsmon side exactly.

So the two views are now asymmetric in precisely the two places the old package is asymmetric, and every option exposed under `old_nut_openssl` has a consumer on 2.8.4-r3. The `certfile` gate asymmetry I asked about last round is confirmed deliberate and stays as is; the redundant-`if` nit is resolved as a side effect — the two blocks on the server side now carry genuinely different conditions, so the seam is load-bearing.

The `.po`/`.pot` files are untouched in this delta, which is correct: the change replaces one line in place, so no msgid and no line number moved. I re-verified against the head sources anyway — all 176 references in `po/templates/nut.pot` resolve to an existing file and an in-range line, and for every entry the referenced line still contains the msgid.

Commit message is unchanged and still matches the diff. CI is green on this head (build x86_64, eslint, CodeQL, Analyze JavaScript and JSON, all three FormalityCheck jobs).

---

openwrt-ai (commented):

Re-reviewed the delta since `dc323e6` (force-pushed to `cf83ef12`). This round is a pure rebase: the base moved from `7a6ff75` to `4c0a4eda`, and every non-catalogue file the commit touches is byte-for-byte identical to what I reviewed last time — I compared `Makefile`, `nut_monitor.js`, `nut_monitor_admin.js`, `nut_server.js`, `nut_server_admin.js` and `acl.d/luci-app-nut.json` blob-by-blob between the two heads and the diff is empty. The commit message is unchanged.

The only content change is in two catalogues, and it comes from upstream rather than from this commit: `git diff 7a6ff75..4c0a4eda -- applications/luci-app-nut/` is exactly `po/lt/nut.po` and `po/zh_Hant/nut.po`, a Weblate refresh that landed on master between the two rebases. Since this commit also regenerates those files, I checked the merge did not damage either one:

- **No live translation was lost.** For both languages, every msgid that carries a `msgstr` at `4c0a4eda` and still exists at `cf83ef12` keeps the same `msgstr` — zero in-place blanking. The only translated entries that disappear are the six msgids this commit itself removes or rewords (`CA Certificate path`, `Certificate file (SSL)`, `Path containing ca certificates to match against host certificate`, `Require SSL and make sure server CN matches hostname`, `Verify all connection with SSL`, `Unable to run ldd: %s`).
- **The new upstream work survived the regeneration.** The Lithuanian catalogue keeps the newly translated USB HID / ACL / menu strings it gained upstream (132 → 149 entries, none blanked), and the Chinese (Traditional) `主機名` → `主機名稱` retranslations are present at head.
- **`.pot` ↔ sources.** All 176 references in `po/templates/nut.pot` resolve to a file that exists at `cf83ef12` and to an in-range line, and for all 149 entries the referenced line actually contains the msgid. No drift from the rebase.
- **`.pot` ↔ catalogues.** All 42 `.po` files carry exactly the template's 149 msgids — none missing, none orphaned, obsolete `#~` entries excluded.
- `ldd` still appears only where this round intends it to: `+USE_GLIBC:ldd` in `LUCI_DEPENDS`, the two `/usr/bin/ldd` exec entries under `luci-app-nut-admin`, and the `fs.exec_direct` fallback in each admin view.

The one red check (`build / Test x86_64`) is infrastructure, not this PR: the SDK tarball failed its checksum after download (`openwrt-sdk-x86-64_gcc-14.4.0_musl.Linux-x86_64.tar.zst: FAILED`), before any luci package was built. eslint, CodeQL, Analyze JavaScript and JSON and all three FormalityCheck jobs pass.

---

danielfdickinson:
Ping @systemcrash @hnyman

applications/luci-app-nut/po/tr/nut.po
```diff
@@ -310,282 +343,345 @@ msgstr "Ağ UPS Araçları"
...
 msgstr "Örnek alma sıklığı(s)"
 
 #: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_cgi.js:28
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:32
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:83
-#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:233
-#, fuzzy
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_monitor.js:29
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:77
+#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:222
```
openwrt-ai, line 438: nit: the rebase kept the retranslated `msgstr` that [`4d9e47d6`](https://github.com/openwrt/luci/commit/4d9e47d63f4f4bd7e07dd0aefaf3adc8b7fbc10e) brought in but dropped the `#, fuzzy` flag that came with it, so a string Weblate marked as needing review is silently promoted to approved. `msgmerge -U -N` — what [`build/i18n-update.pl:76`](https://github.com/openwrt/luci/blob/ca911e36084b40b406fb47b3247573039a91abdf/build/i18n-update.pl#L76) runs — preserves existing fuzzy flags, so this looks like a hand-resolved rebase conflict rather than regeneration output; it is the only fuzzy marker that changed across the 42 catalogues.

```suggestion
#: applications/luci-app-nut/htdocs/luci-static/resources/view/nut/nut_server.js:222
#, fuzzy
```

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

openwrt-ai (commented):

---

PR #8927, opened 2026-08-09
luci-app-pbr: preserve resolver_set selection

pesa1234:
Since commit 974b5864e05e ("luci-base: fix uci write bug when input value equals default"), form values matching their default are removed when rmempty is enabled.

luci-app-pbr selects the resolver_set default dynamically according to the resolver support available on the device. Unlike the other PBR options with defaults, pbr treats a missing resolver_set as disabled.

Saving the page can therefore silently disable dnsmasq nft set handling and make domain policies fall back to one-time resolveip lookups.

Use "none" as the base default and disable rmempty so LuCI always stores the selected resolver.

### Maintainer

@stangri @egc112 

---

## Tested on

**OpenWrt version:** Snapshot
**LuCI version:** Latest

---

### DIFF
--- a/applications/luci-app-pbr/htdocs/luci-static/resources/view/pbr/overview.js
+++ b/applications/luci-app-pbr/htdocs/luci-static/resources/view/pbr/overview.js
@@ -113,6 +113,8 @@ return view.extend({
 			text
 		);
 		o.value("none", _("Disabled"));
+		o.default = "none";
+		o.rmempty = false;
 		if (reply.platform.adguardhome_ipset_support) {
 			o.value("adguardhome.ipset", _("AdGuardHome ipset"));
 			o.default = "adguardhome.ipset";
--- a/applications/luci-app-pbr/Makefile
+++ b/applications/luci-app-pbr/Makefile
@@ -7,7 +7,7 @@ PKG_NAME:=luci-app-pbr
 PKG_LICENSE:=AGPL-3.0-or-later
 PKG_MAINTAINER:=Stan Grishin <stangri@melmac.ca>, Erik Conijn <egc112@msn.com>
 PKG_VERSION:=1.2.2
-PKG_RELEASE:=20
+PKG_RELEASE:=21
 
 LUCI_TITLE:=Policy Based Routing Service Web UI
 LUCI_URL:=https://github.com/mossdef-org/luci-app-pbr/


applications/luci-app-pbr/htdocs/luci-static/resources/view/pbr/overview.js
```diff
@@ -113,6 +113,8 @@ return view.extend({
 			text
 		);
 		o.value("none", _("Disabled"));
+		o.default = "none";
+		o.rmempty = false;
```
openwrt-ai, line 117: `o.default = "none"` is immediately overridden by the three platform-conditional [`o.default = ...` assignments on lines 118-129](``https://github.com/openwrt/luci/blob/762807d6574f1558a0cda3f84d5d038408abfa0a/applications/luci-app-pbr/htdocs/luci-static/resources/view/pbr/overview.js#L118-L129),`` so on any device with resolver support the effective default is still a real resolver, not `none`. Combined with the new `rmempty = false` that produces a silent config flip in the opposite direction of the one this patch fixes:

- when the option is unset, [`ListValue.renderWidget` preselects `this.default`](https://github.com/openwrt/luci/blob/762807d6574f1558a0cda3f84d5d038408abfa0a/modules/luci-base/htdocs/luci-static/resources/form.js#L4703), i.e. `dnsmasq.nftset` on a supporting device;
- with `rmempty = false` the rmempty branch is skipped and [`parse()` falls through to `!isEqual(cval, fval)` → `write()`](https://github.com/openwrt/luci/blob/762807d6574f1558a0cda3f84d5d038408abfa0a/modules/luci-base/htdocs/luci-static/resources/form.js#L2152-L2163);
- pbr treats a missing `resolver_set` as disabled — [`resolver_set=` is initialised empty](https://github.com/openwrt/packages/blob/d04b36bbba607e65072ecd0abfae5a89af468607/net/pbr/files/etc/init.d/pbr#L146) and [`[ "$resolver_set" = 'none' ] && unset resolver_set`](https://github.com/openwrt/packages/blob/d04b36bbba607e65072ecd0abfae5a89af468607/net/pbr/files/etc/init.d/pbr#L791), with the `''|none` case short-circuiting the resolver helper.

So on a config where `resolver_set` is absent, merely opening and saving this page now enables dnsmasq nft set handling that was previously off.

Which behaviour is intended? If "missing means disabled" is the semantic being encoded, the three platform-conditional default overrides should be dropped so that `none` genuinely is the base default. If instead the platform-detected default is meant to win on configs that have no `resolver_set`, then this line has no effect and the commit message's "use `none` as the base default" doesn't describe what the patch does.

---

openwrt-ai (commented):
Reviewed 1 new commit; one question on the interaction between the new `rmempty = false` and the platform-conditional defaults. The commit message matches its diff, the `PKG_RELEASE` bump is present, and the `resolver_set` option is confirmed to have a backend consumer in `net/pbr` (`openwrt/packages`).

---

applications/luci-app-pbr/htdocs/luci-static/resources/view/pbr/overview.js
pesa1234: The analysis is correct, but an absent value is not an explicit “Disabled” choice; `none` is. When unset, the platform-detected default should be stored, matching the behavior before 974b5864e05e. `o.default = "none"` is only the fallback when no resolver is supported.

egc112:
Thanks for bringing this to our attention.

The Luci-Base "fix" causing this  is luckily only in Master branch as far as I can tell and is hopefully not backported as it will cause more problems when Luci and back-end have different defaults.

Your PR is against version 1.2.2 which is EOL as we are working on an ucode port 1.2.3 so changes will be made there and I am working on it as we speak. 

Again many thanks

BKPepe:
@egc112 Is there any update on this one?

egc112:
Yes there is, 1.2.2 is E.O.L and superseded by the ucode port which has the fix from @pesa1234  included.
See: https://github.com/mossdef-org/luci-app-pbr/commit/33c16b9176a8dcd35262c5f36686c61f2de10b8a
@stangri will push this to OpenWRT Master branch the coming days

PR #8930, opened 2026-08-10
luci-app-watchcat: add support for 'enabled' option

Ser9ei:
Adds a checkbox to enable/disable each watchcat instance.

### Description

Adds a checkbox to enable/disable each watchcat instance.
Depends on: https://github.com/openwrt/packages/pull/30234

### Maintainer: @jow-
---

## Tested on
**OpenWrt Version:** 25.12.5
**OpenWrt Target/Subtarget:** mediatek/filogic
**OpenWrt Device:** Xiaomi Mi Router AX3000T

---

## Checklist
- [x] Includes what Issue it closes (e.g. openwrt/luci#issue-number).

### DIFF
--- a/applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js
+++ b/applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js
@@ -18,6 +18,12 @@ return view.extend({
 
 		s.tab('general', _('General Settings'));
 
+		o = s.taboption('general', form.Flag, 'enabled',
+			_('Enabled'),
+			_('Enable or disable this Watchcat instance'));
+		o.default = '1';
+		o.rmempty = false;
+
 		o = s.taboption('general', form.ListValue, 'mode',
 			_('Mode'),
 			_("Ping Reboot: Reboot this device if a ping to a specified host fails for a specified duration of time. <br /> \
--- a/applications/luci-app-watchcat/po/templates/watchcat.pot
+++ b/applications/luci-app-watchcat/po/templates/watchcat.pot
@@ -1,7 +1,7 @@
 msgid ""
 msgstr "Content-Type: text/plain; charset=UTF-8"
 
-#: applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js:94
+#: applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js:100
 msgid ""
 "<i>Applies to Ping Reboot and Periodic Reboot modes</i> <br /> When "
 "rebooting the router, the service will trigger a soft reboot. Entering a non-"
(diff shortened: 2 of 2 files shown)


applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js
```diff
@@ -18,6 +18,12 @@ return view.extend({
 
 		s.tab('general', _('General Settings'));
 
+		o = s.taboption('general', form.Flag, 'enabled',
+			_('Enabled'),
+			_('Enable or disable this watchcat instance'));
+		o.default = '1';
```
openwrt-ai, line 24: `option enabled` only lands in `/etc/config/watchcat` once the user saves this page — [`AbstractValue.parse()`](https://github.com/openwrt/luci/blob/9f169b678e037dbb0c6ec1015cc5efba7baa739d/modules/luci-base/htdocs/luci-static/resources/form.js#L2148-L2168) writes it because `rmempty = false`, but pre-existing sections that are never re-saved keep no `enabled` line at all. Current master of the backend never reads the option ([`config_watchcat()` in watchcat.init](https://github.com/openwrt/packages/blob/master/utils/watchcat/files/watchcat.init) has no `config_get_bool enabled`), and the shipped [`watchcat.config`](https://github.com/openwrt/packages/blob/master/utils/watchcat/files/watchcat.config) default section has none either.

Does openwrt/packages#30234 use `config_get_bool enabled "$1" enabled "1"` (i.e. absent == enabled)? If it defaults to `0`, every existing installation's watchcat instances would silently stop after the package upgrade. I can't read that PR from this session, so flagging it as a question rather than an assertion.

---

applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js
```diff
@@ -18,6 +18,12 @@ return view.extend({
 
 		s.tab('general', _('General Settings'));
 
+		o = s.taboption('general', form.Flag, 'enabled',
+			_('Enabled'),
+			_('Enable or disable this watchcat instance'));
```
openwrt-ai: nit: every other user-visible string in this file spells the product name capitalised — `_('Watchcat')` on [line 11](``https://github.com/openwrt/luci/blob/9f169b678e037dbb0c6ec1015cc5efba7baa739d/applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js#L11)`` and "you can have Watchcat restart your ModemManager interface" on [line 111-112](``https://github.com/openwrt/luci/blob/9f169b678e037dbb0c6ec1015cc5efba7baa739d/applications/luci-app-watchcat/htdocs/luci-static/resources/view/watchcat.js#L111-L112).`` Lowercase `watchcat` here is inconsistent for translators.

```suggestion
			_('Enable or disable this Watchcat instance'));
```

Changing the msgid means `po/templates/watchcat.pot` needs regenerating so its entry matches.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff, and the regenerated `po/templates/watchcat.pot` is consistent with the JS change (all existing references shifted by +6 lines, the two new msgids inserted in correct sort order at `watchcat.js:22`/`:23`). Two inline notes below.

---

openwrt-ai (commented):

The amended commit only capitalises `Watchcat` in the new help string and regenerates the matching `po/templates/watchcat.pot` msgid — the previous nit is resolved and the `.pot` stays consistent (alphabetical position unchanged, all line references still correct). Commit message still matches the diff, FormalityCheck is green.

The open thread about the backend consumer for `option enabled` (openwrt/packages#30234) is unchanged and still applies.

---

PR #8936, opened 2026-08-12
luci-base: form.js: fix stripTags NotFoundError and modify source node

jjm2473:
### Description

This PR addresses two issues:

1. NotFoundError:
 For example `s='<a><br></a>'`, then `x=dom.parse('<div><a><br></a></div>')`, this `br` is not a direct child node of `x`, the subsequent call `x.replaceChild(document.createTextNode('\n'), br)` will result in an error:

  ```
Uncaught (in promise) NotFoundError: Failed to execute 'replaceChild' on 'Node': The node to be replaced is not a child of this node.
    at eval (form.js:312:6)
    at NodeList.forEach (<anonymous>)
    at ClassConstructor.stripTags (form.js:311:28)
    at ClassConstructor.renderFrame (form.js:4434:30)
  ```

2. modify source node
 `stripTags` modifies the source node, causing switch port status missing `<br>` on first render.
 Clone source node as AI suggests: https://github.com/openwrt/luci/pull/8937#discussion_r3768291647

### Maintainer _(preferred)_

@systemcrash 

---

## Tested on

**OpenWrt version:** iStoreOS 25.12.5 2026081215
**LuCI version:** LuCI istoreos-25.12 branch 26.195.42598~4a1ef8c
**Web browser(s):** Chrome (151.0.7922.76)

---

## Checklist

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -309,7 +309,7 @@ const CBIAbstractElement = baseclass.extend(/** @lends LuCI.form.AbstractElement
 		const x = dom.elem(s) ? s : dom.parse(`<div>${s}</div>`);
 
 		x.querySelectorAll('br').forEach((br) => {
-			x.replaceChild(document.createTextNode('\n'), br);
+			br.parentNode.replaceChild(document.createTextNode('\n'), br);
 		});
 
 		return (x.textContent ?? x.innerText ?? '').replace(/([ \t]*\n)+/g, '\n');
--- a/modules/luci-base/htdocs/luci-static/resources/form.js
+++ b/modules/luci-base/htdocs/luci-static/resources/form.js
@@ -306,7 +306,7 @@ const CBIAbstractElement = baseclass.extend(/** @lends LuCI.form.AbstractElement
 		if (typeof(s) == 'string' && !s.match(/[<>&]/))
 			return s;
 
-		const x = dom.elem(s) ? s : dom.parse(`<div>${s}</div>`);
+		const x = dom.elem(s) ? s.cloneNode(true) : dom.parse(`<div>${s}</div>`);
 
 		x.querySelectorAll('br').forEach((br) => {
 			br.parentNode.replaceChild(document.createTextNode('\n'), br);


modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -309,7 +309,10 @@ const CBIAbstractElement = baseclass.extend(/** @lends LuCI.form.AbstractElement
 		const x = dom.elem(s) ? s : dom.parse(`<div>${s}</div>`);
 
 		x.querySelectorAll('br').forEach((br) => {
-			x.replaceChild(document.createTextNode('\n'), br);
+			if (br.parentElement)
+				br.parentElement.replaceChild(document.createTextNode('\n'), br);
+			else
+				x.replaceChild(document.createTextNode('\n'), br);
```
openwrt-ai: The branch is unnecessary — `br.parentNode` is the correct target in every case and makes the fallback redundant.

`br` always comes from `x.querySelectorAll('br')`, so it is always a descendant of `x` and always has a parent node. Two situations exist:

- `x` is an element (the `dom.parse(`<div>${s}</div>`)` path, or an element passed in via `dom.elem(s)`): the parent is an element, so `parentElement === parentNode`.
- `x` is a `DocumentFragment` and the `<br>` is a direct child of it: `parentElement` is `null` — this is the only case the `else` reaches — and `parentNode` is the fragment, i.e. exactly what `x.replaceChild()` uses. Note [`dom.elem()`](https://github.com/openwrt/luci/blob/61436bb1e63093824e9c1eb40ca3a1a876eb9ac7/modules/luci-base/htdocs/luci-static/resources/luci.js#L1249-L1251) only tests for `nodeType`, so a fragment does reach here.

Using `parentNode` unconditionally covers both, and matches how the rest of the tree does node replacement (e.g. [`cbi.js:366`](https://github.com/openwrt/luci/blob/61436bb1e63093824e9c1eb40ca3a1a876eb9ac7/modules/luci-base/htdocs/luci-static/resources/cbi.js#L366), [`tools/network.js:1383`](``https://github.com/openwrt/luci/blob/61436bb1e63093824e9c1eb40ca3a1a876eb9ac7/modules/luci-mod-network/htdocs/luci-static/resources/tools/network.js#L1383)).``

```suggestion
			br.parentNode.replaceChild(document.createTextNode('\n'), br);
```


---

openwrt-ai (commented):
Reviewed 1 new commit; one inline simplification suggested. The commit message accurately describes the change, and the underlying `NotFoundError` diagnosis is correct.

---

modules/luci-base/htdocs/luci-static/resources/form.js
```diff
@@ -306,10 +306,10 @@ const CBIAbstractElement = baseclass.extend(/** @lends LuCI.form.AbstractElement
 		if (typeof(s) == 'string' && !s.match(/[<>&]/))
 			return s;
 
-		const x = dom.elem(s) ? s : dom.parse(`<div>${s}</div>`);
+		const x = dom.elem(s) ? s.cloneNode(true) : dom.parse(`<div>${s}</div>`);
 
 		x.querySelectorAll('br').forEach((br) => {
-			x.replaceChild(document.createTextNode('\n'), br);
+			br.parentNode.replaceChild(document.createTextNode('\n'), br);
```
openwrt-ai, line 312: nit: the PR description's example for issue 1 no longer matches the code or the commit message, and as written it does not reproduce the error.

The description says:

> For example `s='<br>'`, then `x=dom.parse('<br>')`, this `br` is not a direct child node of `x`

But the string branch calls [`dom.parse(`<div>${s}</div>`)`](https://github.com/openwrt/luci/blob/619ce76e6f979b5b0f455991d1a95336f148bd38/modules/luci-base/htdocs/luci-static/resources/form.js#L309), so for `s = '<br>'` the parsed `x` is the wrapper `<div>` and the `<br>` *is* its direct child — `x.replaceChild()` would have succeeded. The commit message of 19a5954 has the corrected example (`s = '<a><br></a>'`, nested `<br>`), which is the one that actually throws. Worth syncing the PR body so the two don't disagree.

---

openwrt-ai (commented):
Reviewed 2 new commits (the branch was force-pushed, so 61436bb was rewritten as 19a5954, which adopts the `br.parentNode` simplification from the previous review).

Both changes look correct:

- 19a5954 — `br.parentNode.replaceChild()` is right for every case, since `br` always comes from `x.querySelectorAll('br')` and therefore always has a parent, whether `x` is an element or a `DocumentFragment`.
- 619ce76 — `cloneNode(true)` on the `dom.elem(s)` branch is the right scope for the mutation fix: only that branch receives a caller-owned node, while the `dom.parse()` branch already builds a throwaway tree. `textContent` on the detached clone is unaffected by the clone.

Both commit messages match their diffs. One nit inline about the PR description.

---

modules/luci-base/htdocs/luci-static/resources/form.js
jjm2473: My example always `s='<a><br></a>'`, check again.

PR #8941, opened 2026-08-13
luci-app-rustdesk-server: fixes firmware update causes configuration loss.

### LINKED ISSUE #8935: luci-app-rustdesk-server: retaining configuration updates will revert the configuration

### screenshots or captures

_No response_

### Actual behaviour

1. Compile the firmware using OpenWrt source code that includes the RustDesk server.
2. Flash the firmware to the device.
3. Access the OpenWrt interface.
4. Enable the RustDesk server, then save and apply the settings.
5. Ensure the RustDesk server is running.
6. Perform a firmware update while preserving configuration settings by uploading the previously compiled firmware.
7. Access the OpenWrt interface and observe that the previously enabled RustDesk server is now unchecked (disabled); it must be manually enabled again.

### Expected behaviour

Retaining configuration when upgrading the firmware allows you to preserve your previous RustDesk settings.

### Steps to reproduce

see Actual behaviour

### Additional Information

```shell
{
        "kernel": "6.12.94",
        "hostname": "CgyyOARoute",
        "system": "Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz",
        "model": "QEMU Standard PC (Q35 + ICH9, 2009)",
        "board_name": "qemu-standard-pc-q35-ich9-2009",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "25.12-SNAPSHOT",
                "revision": "r33143-d78c0fd16d",
                "target": "x86/64",
                "description": "OpenWrt 25.12-SNAPSHOT r33143-d78c0fd16d",
                "builddate": "1786123572"
        }
}
```

Chrome

### Relevant log output

```shell

```

smallprogram:
Fixes #8935
This fixes the issue where configuration settings were lost during a firmware upgrade.
Signed-off-by: David Mandy <smallprogramzhusir@gmail.com>

### Description

fixes firmware update causes configuration loss.

---

## Tested on

**OpenWrt version:** OpenWrt 25.12.2
**Web browser(s):** Chrome

---

## Checklist

- [x] _(Nice to have)_ Includes what Issue it closes (e.g. openwrt/luci#issue-number).

### DIFF
--- a/applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
+++ b/applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
@@ -3,11 +3,13 @@
 # luci-app-rustdesk-server UCI defaults
 # This script runs on first install to set up sensible defaults
 
-uci -q batch <<-EOF >/dev/null
-	set rustdesk-server.@rustdesk-server[0]=rustdesk-server
-	set rustdesk-server.@rustdesk-server[0].enabled='0'
-	set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
-	commit rustdesk-server
+if ! uci -q get rustdesk-server.@rustdesk-server[0] >/dev/null; then
+	uci -q batch <<-EOF >/dev/null
+		set rustdesk-server.@rustdesk-server[0]=rustdesk-server
+		set rustdesk-server.@rustdesk-server[0].enabled='0'
+		set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
+		commit rustdesk-server
 EOF
+fi
 
 exit 0


applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
```diff
@@ -3,11 +3,14 @@
 # luci-app-rustdesk-server UCI defaults
 # This script runs on first install to set up sensible defaults
 
+[ -f /etc/config/rustdesk-server ] || touch /etc/config/rustdesk-server
+if ! uci -q get rustdesk-server.@rustdesk-server[0] >/dev/null; then
```
openwrt-ai: The `touch` can never fire: the package ships [`root/etc/config/rustdesk-server`](``https://github.com/openwrt/luci/blob/bc36bd593150e3f1756a0f9c48e2b1240c1269e6/applications/luci-app-rustdesk-server/root/etc/config/rustdesk-server#L1-L3``) and registers it via [`Package/$(PKG_NAME)/conffiles`](https://github.com/openwrt/luci/blob/bc36bd593150e3f1756a0f9c48e2b1240c1269e6/applications/luci-app-rustdesk-server/Makefile#L13-L15), and uci-defaults scripts are run from [`default_postinst`](https://github.com/openwrt/openwrt/blob/91a1245ad74b7c2882a53dac3b84ef90abfbc906/package/base-files/files/lib/functions.sh#L384-L390) only after the file list has been unpacked, so `/etc/config/rustdesk-server` is already there.

And if it somehow were missing, an empty file would not rescue the batch anyway — `@rustdesk-server[0]` only resolves against an already existing section, so `set rustdesk-server.@rustdesk-server[0]=rustdesk-server` would still fail. Creating an anonymous section needs `add rustdesk-server rustdesk-server` followed by `@rustdesk-server[-1]`, the way [`90-hev-socks5-tunnel`](https://github.com/openwrt/packages/blob/c11ef37d558291d272ffe6afc857afa4fd538dd4/net/hev-socks5-tunnel/files/90-hev-socks5-tunnel#L12-L18) does it.

Drop the line:

```suggestion
if ! uci -q get rustdesk-server.@rustdesk-server[0] >/dev/null; then
```

---

openwrt-ai, line 6: With this guard added, the script becomes a no-op on *every* path, not just after sysupgrade: the shipped conffile already contains `config rustdesk-server` with `enabled '0'` and `enabled_relay '0'`, so the section exists on a fresh install too and the branch is never taken.

Was deleting `root/etc/uci-defaults/50-luci-rustdesk-server` outright considered instead? That fixes #8935 the same way — there is nothing left to re-run and reset values on the next boot — and it avoids leaving a script in the tree that can only ever be dead code. If the script is being kept deliberately (e.g. as a safety net for installs predating the conffile), it would help to say so in the commit message.

---

applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
```diff
@@ -3,11 +3,14 @@
...
 uci -q batch <<-EOF >/dev/null
 	set rustdesk-server.@rustdesk-server[0]=rustdesk-server
 	set rustdesk-server.@rustdesk-server[0].enabled='0'
 	set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
 	commit rustdesk-server
 EOF
+fi
```
openwrt-ai: nit: the new `if` body is left at column 0, so nothing visually marks the conditional block. Comparable in-tree uci-defaults scripts indent it — see [`90-hev-socks5-tunnel`](https://github.com/openwrt/packages/blob/c11ef37d558291d272ffe6afc857afa4fd538dd4/net/hev-socks5-tunnel/files/90-hev-socks5-tunnel#L12-L18). `<<-EOF` strips leading tabs from both the here-doc body and the terminator, so indenting is safe here.

```suggestion
	uci -q batch <<-EOF >/dev/null
		set rustdesk-server.@rustdesk-server[0]=rustdesk-server
		set rustdesk-server.@rustdesk-server[0].enabled='0'
		set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
		commit rustdesk-server
	EOF
fi
```

---

openwrt-ai (commented):
Reviewed 1 new commit; the diagnosis is right — uci-defaults scripts ship in the new rootfs and re-run on the first boot after sysupgrade, so the unguarded batch was resetting `enabled`/`enabled_relay` back to `0`. Comments inline on the implementation.

---

applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
```diff
@@ -3,11 +3,13 @@
...
-	set rustdesk-server.@rustdesk-server[0].enabled='0'
-	set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
-	commit rustdesk-server
+if ! uci -q get rustdesk-server.@rustdesk-server[0] >/dev/null; then
+	uci -q batch <<-EOF >/dev/null
+		set rustdesk-server.@rustdesk-server[0]=rustdesk-server
+		set rustdesk-server.@rustdesk-server[0].enabled='0'
+		set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
+		commit rustdesk-server
 EOF
```
openwrt-ai, line 12: nit: the re-indent in c33e042 moved the here-doc body but left the terminator at column 0, so the block now reads as unbalanced. It still works — `<<-` strips leading tabs from the delimiter line as well as the body — so this is cosmetic only, but indenting it to match the `uci -q batch` line it closes is what the rest of the block now does.

```suggestion
	EOF
```

---

applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
```diff
@@ -3,11 +3,13 @@
 # luci-app-rustdesk-server UCI defaults
 # This script runs on first install to set up sensible defaults
 
-uci -q batch <<-EOF >/dev/null
-	set rustdesk-server.@rustdesk-server[0]=rustdesk-server
-	set rustdesk-server.@rustdesk-server[0].enabled='0'
-	set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
-	commit rustdesk-server
+if ! uci -q get rustdesk-server.@rustdesk-server[0] >/dev/null; then
```
openwrt-ai, line 6: nit: this guard is now spread over three commits for what is a single 7-line change — bc36bd5 introduces it together with a `touch` line, ebd811f deletes that `touch` line again, and c33e042 only re-indents the here-doc body. Neither of the last two stands on its own; they are review fixups to the first.

Squashing all three into one commit would make the series bisectable and would also clear the `FormalityCheck / Git & Commits` failure, since the offending `Update applications/...` subject on ebd811f disappears with the commit.

---

openwrt-ai (commented):
Reviewed 2 new commits. The `touch` removal and the indentation change both address the earlier inline comments, and the guard itself is correct now.

The design question from the previous review is still open, though: the shipped conffile [`root/etc/config/rustdesk-server`](``https://github.com/openwrt/luci/blob/c33e0422801d7c1fc337b3336191008b27b597fc/applications/luci-app-rustdesk-server/root/etc/config/rustdesk-server#L1-L3``) already contains the `rustdesk-server` section with `enabled '0'` and `enabled_relay '0'`, so `@rustdesk-server[0]` resolves on a fresh install too and the guarded batch is unreachable on every path. That thread was resolved without a reply — if the script is being kept deliberately rather than deleted, the commit message is the place to say why.

## Commit checks

- c33e042 "luci-app-rustdesk-server: fixes firmware update causes config loss" — the subject and body describe the sysupgrade fix, but this commit's diff contains only the re-indentation of the here-doc body; the actual fix landed in bc36bd5. Squashing the series (see inline) makes the message match the change.

---

efahl:
I think if you squash the commits, and add 
`Fixes: https://github.com/openwrt/luci/issues/8935` 
to the commit message, you'll have it.

smallprogram:
> I think if you squash the commits, and add `Fixes: https://github.com/openwrt/luci/issues/8935` to the commit message, you'll have it.

Resubmitted the PR. #8946

PR #8942, opened 2026-08-13
luci-proto-openthread: add protocol support for the border router

LorbusChris:
Adds a netifd protocol handler for the `openthread` protocol, so an interface running a Thread
border router can be configured from the network pages instead of only from uci.

The protocol handler itself ships with openthread-br, which already installs /lib/netifd/proto/openthread.sh at packages master.

Options, matching what the proto handler consumes:

| Tab | Option | What |
|---|---|---|
| General | Backbone network | the network carrying mDNS, TREL and border routing for the mesh, usually LAN |
| General | Thread network device | the netdev `otbr-agent` creates for the Thread interface; declared here because the stock device picker is hidden for virtual protocols |
| General | Radio URL | how `otbr-agent` reaches the 802.15.4 radio |
| Advanced | Operational dataset | hex-encoded active dataset committed at startup; normally left empty, since the network is formed or joined from the application |
| Advanced | On-mesh prefixes | prefixes announced to the Thread network |
| Advanced | Verbose logging | flag |

One thing worth calling out, since it is the only non-obvious part: the Thread radio is a tun-like
netdev with no `DEVTYPE` in sysfs, so the generic `Device` class falls back to the ethernet icon.
The protocol therefore owns its device instances and reports `wifi` / "Thread Radio", the same
pattern `luci-proto-relay` uses. Without it the interface list shows a wired icon for a radio.

### Testing

Running on a Turris Omnia with an 802.15.4 USB dongle: the interface appears with the Thread type
and icon, the backbone network and radio URL round-trip through uci, and the advanced options are
written only when set.

### Related

Independent of #8871 (`luci-app-openthread`) — different package, no shared files — and of the
openthread-br changes in the packages feed. It can be reviewed and merged on its own.

### DIFF
--- a/protocols/luci-proto-openthread/Makefile
+++ b/protocols/luci-proto-openthread/Makefile
@@ -1,0 +1,13 @@
+# This is free software, licensed under the Apache License, Version 2.0 .
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Support for the Thread border router protocol
+LUCI_DEPENDS:=+openthread-br
+
+PKG_MAINTAINER:=Christian Glombek <c.glombek@cosa.systems>
+PKG_LICENSE:=Apache-2.0
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/protocols/luci-proto-openthread/htdocs/luci-static/resources/protocol/openthread.js
+++ b/protocols/luci-proto-openthread/htdocs/luci-static/resources/protocol/openthread.js
@@ -1,0 +1,97 @@
+'use strict';
+'require form';
+'require network';
+'require tools.widgets as widgets';
+
+// The Thread radio is a tun-like netdev with no DEVTYPE in sysfs, so the
+// generic Device class falls back to the ethernet icon. Own the device
+// instances (same pattern as luci-proto-relay) so the interface list shows
+// a wireless icon and a meaningful type.
+var ThreadDevice = {
+	getType: function() {
+		return 'wifi';
+	},
+
+	getTypeI18n: function() {
+		return _('Thread Radio');
+	}
+};
+
+function threadDev(proto, name) {
+	var m = name ? name.match(/^([^:/]+)/) : null;
+	return m ? network.instantiateDevice(m[1], proto, ThreadDevice) : null;
+}
+
+network.registerErrorCode('MISSING_BACKBONE_NETWORK', _('Backbone network is not configured'));
+network.registerErrorCode('MISSING_BACKBONE_IFNAME',  _('Unable to determine the backbone network device'));
+network.registerErrorCode('MISSING_DEVICE',           _('Thread network device is not configured'));
+network.registerErrorCode('MISSING_RADIO_URL',        _('Radio URL is not configured'));
+network.registerErrorCode('MISSING_SVC_MDNSD',        _('The mdnsd service is not running'));
+network.registerErrorCode('MISSING_UBUS_OBJ',         _('Unable to reach the otbr ubus object'));
+
+return network.registerProtocol('openthread', {
+	getI18n: function() {
+		return _('Thread');
+	},
+
+	getDevice: function() {
+		return threadDev(this, this._get('device'));
+	},
+
+	getL2Device: function() {
+		return threadDev(this, this._ubus('device'));
+	},
+
+	getL3Device: function() {
+		return threadDev(this, this._ubus('l3_device'));
+	},
+
+	getPackageName: function() {
+		return 'openthread-br';
+	},
+
+	renderFormOptions: function(s) {
+		var o;
+
+		o = s.taboption('general', widgets.NetworkSelect, 'backbone_network',
+			_('Backbone network'),
+			_('The network whose interface carries mDNS, TREL and border routing for the Thread mesh; usually the LAN.'));
+		o.exclude = s.section;
+		o.nocreate = true;
+		o.rmempty = false;
+
+		// no_device makes this protocol virtual, so the stock device picker
+		// is inactive here -- and an inactive picker deletes the option on
+		// save. Aliasing an own option onto `device` keeps the field
+		// editable and, through the sibling-alias guard in
+		// AbstractValue.remove(), stops the picker from unsetting it.
+		o = s.taboption('general', form.Value, '_thread_device',
+			_('Thread network device'),
+			_('Name of the network device otbr-agent creates for the Thread interface.'));
+		o.ucioption = 'device';
+		o.rmempty = false;
+		o.placeholder = 'wpan0';
+
+		o = s.taboption('general', form.Value, 'radio_url',
+			_('Radio URL'),
+			_('How otbr-agent reaches the 802.15.4 radio.'));
+		o.rmempty = false;
+		o.placeholder = 'spinel+hdlc+uart:///dev/ttyACM0?uart-baudrate=460800';
+
+		o = s.taboption('advanced', form.Value, 'dataset',
+			_('Operational dataset'),
+			_('Hex-encoded active operational dataset committed at startup. Usually left empty: the network is formed or joined from the OpenThread application instead.'));
+		o.optional = true;
+		o.datatype = 'hexstring';
+		o.password = true;
+
+		o = s.taboption('advanced', form.DynamicList, 'prefix',
+			_('On-mesh prefixes'),
+			_('Prefixes announced to the Thread network.'));
+		o.optional = true;
+
+		o = s.taboption('advanced', form.Flag, 'verbose',
+			_('Verbose logging'));
+		o.optional = true;
+	}
+});


protocols/luci-proto-openthread/htdocs/luci-static/resources/protocol/openthread.js
```diff
@@ -0,0 +1,77 @@
...
+		return _('Thread Radio');
+	}
+};
+
+function threadDev(proto, name) {
+	var m = name ? name.match(/^([^:/]+)/) : null;
+	return m ? network.instantiateDevice(m[1], proto, ThreadDevice) : null;
+}
+
+return network.registerProtocol('openthread', {
```
openwrt-ai, line 32: The proto handler raises six custom error codes that this file never registers, so every setup failure renders as `Unknown error (MISSING_RADIO_URL)` in the interface status instead of a readable message — which is exactly the case a user hitting a missing radio URL or a stopped mdnsd will see.

`proto_openthread_setup()` emits `MISSING_BACKBONE_NETWORK`, `MISSING_BACKBONE_IFNAME`, `MISSING_DEVICE`, `MISSING_RADIO_URL`, `MISSING_SVC_MDNSD` and `MISSING_UBUS_OBJ` ([openthread-proto.sh:66-93](https://github.com/openwrt/packages/blob/61972dae3af03f5f8ac757b9520eb6f2ebe2eb21/net/openthread-br/files/openthread-proto.sh#L66-L93)), and none of them is in the base [`proto_errors` table at network.js:8-22](https://github.com/openwrt/luci/blob/4fd96e09e62fb8e5d2cc822ff71c653cb8b28e6e/modules/luci-base/htdocs/luci-static/resources/network.js#L8-L22); [`getErrors()` at network.js:2443](https://github.com/openwrt/luci/blob/4fd96e09e62fb8e5d2cc822ff71c653cb8b28e6e/modules/luci-base/htdocs/luci-static/resources/network.js#L2443) then falls back to the raw code. Protocols that ship their own handler codes register them at module scope — see [map.js:7-9](``https://github.com/openwrt/luci/blob/4fd96e09e62fb8e5d2cc822ff71c653cb8b28e6e/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/map.js#L7-L9``) and [qmi.js:21-23](``https://github.com/openwrt/luci/blob/4fd96e09e62fb8e5d2cc822ff71c653cb8b28e6e/protocols/luci-proto-qmi/htdocs/luci-static/resources/protocol/qmi.js#L21-L23).``

```suggestion
network.registerErrorCode('MISSING_BACKBONE_NETWORK', _('Backbone network is not configured'));
network.registerErrorCode('MISSING_BACKBONE_IFNAME', _('Unable to determine the backbone network device'));
network.registerErrorCode('MISSING_DEVICE', _('Thread network device is not configured'));
network.registerErrorCode('MISSING_RADIO_URL', _('Radio URL is not configured'));
network.registerErrorCode('MISSING_SVC_MDNSD', _('The mdnsd service is not running'));
network.registerErrorCode('MISSING_UBUS_OBJ', _('Unable to reach the otbr ubus object'));

return network.registerProtocol('openthread', {
```


---

protocols/luci-proto-openthread/htdocs/luci-static/resources/protocol/openthread.js
```diff
@@ -0,0 +1,77 @@
...
+			_('Radio URL'),
+			_('How otbr-agent reaches the 802.15.4 radio.'));
+		o.rmempty = false;
+		o.placeholder = 'spinel+hdlc+uart:///dev/ttyACM0?uart-baudrate=460800';
+
+		o = s.taboption('advanced', form.Value, 'dataset',
+			_('Operational dataset'),
+			_('Hex-encoded active operational dataset committed at startup. Usually left empty: the network is formed or joined from the OpenThread application instead.'));
+		o.optional = true;
+		o.datatype = 'hexstring';
```
openwrt-ai, line 85: The active operational dataset is a credential, not just an opaque blob: its TLVs carry the Thread Network Key and the PSKc, and `$OTCTL dataset set active "$dataset"` in the handler commits them verbatim. Rendering it in a plain text input shows the mesh key in cleartext to anyone with the LuCI page open (and to shoulder-surfing/screenshots), while every other protocol option holding a secret in this tree sets `o.password = true` — e.g. [6in4.js:82](``https://github.com/openwrt/luci/blob/4fd96e09e62fb8e5d2cc822ff71c653cb8b28e6e/protocols/luci-proto-ipv6/htdocs/luci-static/resources/protocol/6in4.js#L82``) for the tunnel update key and [vpnc.js:59](``https://github.com/openwrt/luci/blob/4fd96e09e62fb8e5d2cc822ff71c653cb8b28e6e/protocols/luci-proto-vpnc/htdocs/luci-static/resources/protocol/vpnc.js#L59``) for the pre-shared key.

```suggestion
		o.optional = true;
		o.datatype = 'hexstring';
		o.password = true;
```


---

protocols/luci-proto-openthread/Makefile
```diff
@@ -0,0 +1,13 @@
+# This is free software, licensed under the Apache License, Version 2.0 .
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Support for the Thread border router protocol
+LUCI_DEPENDS:=+openthread-br
```
openwrt-ai, line 6: nit: the PR body says the handler arrives with "the update in openwrt/packages#30091, which installs `/lib/netifd/proto/openthread.sh`", but `openthread-br` at packages master already installs it — [Makefile:99](https://github.com/openwrt/packages/blob/61972dae3af03f5f8ac757b9520eb6f2ebe2eb21/net/openthread-br/Makefile#L99) does `$(INSTALL_BIN) ./files/openthread-proto.sh $(1)/lib/netifd/proto/openthread.sh` — and the option set that [`proto_openthread_init_config()`](https://github.com/openwrt/packages/blob/61972dae3af03f5f8ac757b9520eb6f2ebe2eb21/net/openthread-br/files/openthread-proto.sh#L36-L46) declares there (`prefix`, `verbose`, `backbone_network`, `dataset`, `radio_url`) is already exactly what this form writes.

If that is right, this `LUCI_DEPENDS` is satisfied by the feed as it stands and #30091 is not a merge blocker — worth correcting the description so a maintainer doesn't hold the PR waiting on it. If #30091 changes the option names or adds ones this form should expose, could you say which, so the coupling is reviewable?

---

openwrt-ai (commented):
Reviewed 1 new commit; commit message matches the diff. Three inline comments — the option set, tab layout, `hexstring` datatype and `optional`/`rmempty` flags all line up with what `proto_openthread_init_config()`/`proto_openthread_setup()` in `openthread-br` consume, and the `_net_device` selector in `interfaces.js` covers the `device` option the handler requires, so the missing pieces are the unregistered handler error codes and the unmasked dataset field.

---

protocols/luci-proto-openthread/htdocs/luci-static/resources/protocol/openthread.js
```diff
@@ -0,0 +1,85 @@
...
+			_('The network whose interface carries mDNS, TREL and border routing for the Thread mesh; usually the LAN.'));
+		o.exclude = s.section;
+		o.nocreate = true;
+		o.rmempty = false;
+
+		o = s.taboption('general', form.Value, 'radio_url',
+			_('Radio URL'),
+			_('How otbr-agent reaches the 802.15.4 radio.'));
+		o.rmempty = false;
+		o.placeholder = 'spinel+hdlc+uart:///dev/ttyACM0?uart-baudrate=460800';
```
openwrt-ai, line 79: `option device` can neither be set from this form nor survive a save through it, so the `MISSING_DEVICE` string registered on line 27 is an error the user has no way to clear in the UI.

`proto_openthread_init_config()` sets [`no_device=1`](https://github.com/openwrt/packages/blob/ae3874f304d85d8978422f8e466105b9bebf2e7f/net/openthread-br/files/openthread-proto.sh#L36-L46), netifd turns that into `PROTO_FLAG_NODEV` ([proto-shell.c:111-113](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/proto-shell.c#L111-L113)) and reports `no_device: true` from `get_proto_handlers` ([proto.c:617](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/proto.c#L617)), so [`isVirtual()` at network.js:840-842](https://github.com/openwrt/luci/blob/1a90b724d5b6899aaf35ec2503067cc13593b128/modules/luci-base/htdocs/luci-static/resources/network.js#L840-L842) is true for `openthread`. Both places that expose the uci `device` option gate on exactly that:

- the interface modal builds `_net_device`'s deps [only from non-virtual protocols (interfaces.js:1338-1348)](``https://github.com/openwrt/luci/blob/1a90b724d5b6899aaf35ec2503067cc13593b128/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L1338-L1348),`` and
- the *Add new interface* dialog does the same ([interfaces.js:1417-1426](``https://github.com/openwrt/luci/blob/1a90b724d5b6899aaf35ec2503067cc13593b128/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L1417-L1426)).``

The second-order effect is worse than a missing field: since `_net_device` is inactive, [`parse()` at form.js:2163-2168](https://github.com/openwrt/luci/blob/1a90b724d5b6899aaf35ec2503067cc13593b128/modules/luci-base/htdocs/luci-static/resources/form.js#L2163-L2168) takes the `!this.retain` branch and calls `remove()`, so opening an existing `openthread` interface and hitting Save *deletes* a hand-written `option device 'wpan0'` and the interface stops coming up.

The handler hard-requires the option — [`[ -n "$device" ] || … MISSING_DEVICE` at openthread-proto.sh:71](https://github.com/openwrt/packages/blob/ae3874f304d85d8978422f8e466105b9bebf2e7f/net/openthread-br/files/openthread-proto.sh#L71) — and the package README lists it as required ([README.md:83](https://github.com/openwrt/packages/blob/ae3874f304d85d8978422f8e466105b9bebf2e7f/net/openthread-br/README.md#L83), [README.md:91](https://github.com/openwrt/packages/blob/ae3874f304d85d8978422f8e466105b9bebf2e7f/net/openthread-br/README.md#L91)), right above the ["Creating a network in LuCI appears to be broken for the moment"](https://github.com/openwrt/packages/blob/ae3874f304d85d8978422f8e466105b9bebf2e7f/net/openthread-br/README.md#L128) note this PR presumably means to retire. `getDevice()` above already reads `this._get('device')`, so the form is the only piece missing.

Declaring the option here fixes both halves at once: the user gets a field, and the sibling-alias guard in [`remove()` at form.js:2224-2244](https://github.com/openwrt/luci/blob/1a90b724d5b6899aaf35ec2503067cc13593b128/modules/luci-base/htdocs/luci-static/resources/form.js#L2224-L2244) then finds an active option mapping to `device` and stops `_net_device` from unsetting it.

```suggestion
		o = s.taboption('general', form.Value, 'device',
			_('Thread network device'),
			_('Name of the network device otbr-agent creates for the Thread interface.'));
		o.rmempty = false;
		o.placeholder = 'wpan0';

		o = s.taboption('general', form.Value, 'radio_url',
			_('Radio URL'),
			_('How otbr-agent reaches the 802.15.4 radio.'));
		o.rmempty = false;
		o.placeholder = 'spinel+hdlc+uart:///dev/ttyACM0?uart-baudrate=460800';
```

The option table in the PR description would want a matching row.

---

openwrt-ai (commented):
Re-reviewed the range `4fd96e0..1a90b72`. The two code findings from the previous round are addressed correctly: the six `registerErrorCode()` calls match the codes `proto_openthread_setup()` actually raises one-for-one and land at module scope like `map.js`/`qmi.js` do, and `o.password = true` on the dataset field is the right knob (it reaches `ui.Textfield` via `form.Value.renderWidget()`). The Makefile/`LUCI_DEPENDS` point is resolved by the updated PR description. Commit message still matches the diff.

One inline comment on the remaining gap: `openthread` is a `no_device` protocol, which makes LuCI hide the device picker and — on save — actively remove the `option device` the handler requires.

---

protocols/luci-proto-openthread/htdocs/luci-static/resources/protocol/openthread.js
```diff
@@ -0,0 +1,95 @@
...
+
+		// no_device makes this protocol virtual, so the stock device picker
+		// is inactive here -- and an inactive picker deletes the option on
+		// save. Declaring the option keeps the field editable and, through
+		// the sibling-alias guard, stops the picker from unsetting it.
+		o = s.taboption('general', form.Value, 'device',
+			_('Thread network device'),
+			_('Name of the network device otbr-agent creates for the Thread interface.'));
+		o.rmempty = false;
+		o.placeholder = 'wpan0';
```
openwrt-ai, line 73: The sibling-alias guard the comment relies on does not actually fire for this option, so `_net_device` still deletes `option device` on save. My previous round's suggestion was wrong on this point — apologies for sending you down this path.

The guard skips any sibling that has no explicit `ucioption`: [`if (sibling === this || sibling.ucioption == null) continue;` at form.js:2228-2229](https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/modules/luci-base/htdocs/luci-static/resources/form.js#L2228-L2229). This option is declared as `form.Value, 'device'`, i.e. `option = 'device'` with `ucioption` left `undefined`, so when the inactive `_net_device` (`ucioption = 'device'`, [interfaces.js:631-632](``https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L631-L632))`` hits the [`!this.retain` branch at form.js:2167-2168](https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/modules/luci-base/htdocs/luci-static/resources/form.js#L2167-L2168) and calls `remove()`, the loop `continue`s past this option and reaches `this.map.data.unset(…, 'device')`.

Nothing puts the value back on a plain Save: `_net_device` is added at [interfaces.js:631](``https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L631),`` well before `ifc.renderFormOptions(s)` at [interfaces.js:1216](``https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/modules/luci-mod-network/htdocs/luci-static/resources/view/network/interfaces.js#L1216),`` so it parses first — and this option's own parse then compares against [`cfgvalue()`, which returns the *cached* `load()` result (form.js:1923-1932)](https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/modules/luci-base/htdocs/luci-static/resources/form.js#L1923-L1932), not live uci. An untouched `wpan0` therefore takes the `isEqual(cval, fval)` path at [form.js:2163](https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/modules/luci-base/htdocs/luci-static/resources/form.js#L2163) and never writes, leaving the option removed.

Setting `ucioption` explicitly is what every other protocol owning `device` does — [`_modem_device` in 3g.js:73-74](``https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/protocols/luci-proto-3g/htdocs/luci-static/resources/protocol/3g.js#L73-L74),`` [modemmanager.js:51-52](``https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js#L51-L52),`` [`_batadv_vlan_device` in batadv_vlan.js:48-50](``https://github.com/openwrt/luci/blob/341e507488670ea52db237a84d6ac8f18f01977f/protocols/luci-proto-batman-adv/htdocs/luci-static/resources/protocol/batadv_vlan.js#L48-L50)`` — always under an underscore-prefixed form-option name.

```suggestion
		// no_device makes this protocol virtual, so the stock device picker
		// is inactive here -- and an inactive picker deletes the option on
		// save. Aliasing an own option onto `device` keeps the field
		// editable and, through the sibling-alias guard in
		// AbstractValue.remove(), stops the picker from unsetting it.
		o = s.taboption('general', form.Value, '_thread_device',
			_('Thread network device'),
			_('Name of the network device otbr-agent creates for the Thread interface.'));
		o.ucioption = 'device';
		o.rmempty = false;
		o.placeholder = 'wpan0';
```

`getDevice()` on line 43 reads `this._get('device')` and is unaffected by the form-option rename.

---

openwrt-ai (commented):
Re-reviewed the range `1a90b72..341e507`. The only change is the new `device` option in `renderFormOptions()`, which addresses the previous round's finding — but not completely: without an explicit `o.ucioption`, the sibling-alias guard in `AbstractValue.remove()` skips the option, so `_net_device` still unsets `option device` on save. One inline comment with the fix; the correction is on my previous suggestion, which omitted it.

Commit message still matches the diff. Formality CI is green on `341e507`.

---

openwrt-ai (commented):

Re-reviewed the range `341e507..0bc0731`. The only change is the `device` form option being renamed to `_thread_device` with an explicit `o.ucioption = 'device'`, which closes the previous round's finding: `remove()` now skips the `continue` at [form.js:2228-2229](https://github.com/openwrt/luci/blob/0bc0731480bb1c427d442840a6e4d1f225a842c7/modules/luci-base/htdocs/luci-static/resources/form.js#L2228-L2229) for this sibling, matches on `this_opt`/`sibling_opt` both being `device`, and returns early instead of reaching `unset()` — so the inactive `_net_device` no longer deletes `option device` on save. Both options live on the same section `s`, and `load()`/`write()` resolve through [`this.ucioption ?? this.option` at form.js:1878-1881](https://github.com/openwrt/luci/blob/0bc0731480bb1c427d442840a6e4d1f225a842c7/modules/luci-base/htdocs/luci-static/resources/form.js#L1878-L1881), so the field still round-trips the same uci key. `getDevice()` is unaffected by the rename, as expected.

Commit message still matches the diff (the series was squashed to a single commit; subject scope `luci-proto-openthread:` matches the two added files). Formality CI is green on `0bc0731`.

---

openwrt-ai (commented):

Re-reviewed `0bc0731..dac3882`. The head moved only because the branch was rebased onto master (`4c0a4ed`) — `git diff 0bc0731..dac3882 -- protocols/luci-proto-openthread/` is empty, so both added files are byte-identical to the version I reviewed last round, and the single commit's message is unchanged and still matches its diff (subject scope `luci-proto-openthread:` covers the two added files). Nothing landed on master in that window that touches `luci.mk` or protocol conventions.

The remaining nit from the first round is resolved in the updated PR description, which now states that `openthread-br` at packages master already installs `/lib/netifd/proto/openthread.sh`, so `LUCI_DEPENDS:=+openthread-br` is satisfied by the feed as it stands and no cross-repo merge order is implied.

One thing I re-checked rather than flagged: the file uses `var` and `function () {}` while some recently-touched protocol files (e.g. [`relay.js`](``https://github.com/openwrt/luci/blob/dac3882616696aac9b9dcc2cdb822460246c459c/protocols/luci-proto-relay/htdocs/luci-static/resources/protocol/relay.js#L9-L36))`` have been modernised to method shorthand. 42 of the 44 files under `protocols/*/htdocs/luci-static/resources/protocol/` still use `var`, so this matches the prevailing style of its neighbours and is not a deviation.

All three FormalityCheck runs are green on `dac3882`.

---

PR #8944, opened 2026-08-14
luci-proto-modemmanager: fix iptype default value

micpf:
### Description

The `iptype` and `init_iptype` `ListValue` options in `modemmanager.js` declared `ipv4v6` as the default, while the `modemmanager` package itself defaults to `ipv4` (see [net/modemmanager README.md](https://github.com/openwrt/packages/blob/master/net/modemmanager/README.md?plain=1#L40-L41)).

Since LuCI's form widgets skip writing a UCI option to `/etc/config/network` when the selected value equals the field's declared default, selecting **IPv4/IPv6 (both)** in the UI never actually persisted `option iptype 'ipv4v6'`. On save/apply, modemmanager then fell back to its own default (`ipv4`), silently downgrading the connection to IPv4-only — making the `ipv4v6` option effectively impossible to use through LuCI.

This PR aligns both `iptype` and `init_iptype` defaults with the modemmanager package default (`ipv4`), and drops the now-inaccurate `- defaults to IPv4` hint from the `ipv4v6` label.

### Issue reported by

Originally reported and diagnosed by @moawut on the forum:
https://forum.openwrt.org/t/luci-proto-modemmanager-support-thread/23696/120

### DIFF
--- a/protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
+++ b/protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
@@ -140,10 +140,10 @@ return network.registerProtocol('modemmanager', {
 		o.password = true;
 
 		o = s.taboption('general', form.ListValue, 'iptype', _('IP Type'));
-		o.value('ipv4v6', _('IPv4/IPv6 (both - defaults to IPv4)'))
 		o.value('ipv4', _('IPv4 only'));
 		o.value('ipv6', _('IPv6 only'));
-		o.default = 'ipv4v6';
+		o.value('ipv4v6', _('IPv4/IPv6 (both)'));
+		o.default = 'ipv4';
 
 		o = s.taboption('advanced', form.Value, 'mtu', _('Override MTU'));
 		o.placeholder = dev ? (dev.getMTU() || '1500') : '1500';
@@ -197,9 +197,9 @@ return network.registerProtocol('modemmanager', {
 
 		o = s.taboption('general', form.ListValue, 'init_iptype', _('Initial EPS Bearer IP Type'));
 		o.depends('init_epsbearer', 'custom');
-		o.value('ipv4v6', _('IPv4/IPv6 (both - defaults to IPv4)'))
 		o.value('ipv4', _('IPv4 only'));
 		o.value('ipv6', _('IPv6 only'));
-		o.default = 'ipv4v6';
+		o.value('ipv4v6', _('IPv4/IPv6 (both)'));
+		o.default = 'ipv4';
 	}
 });


protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
```diff
@@ -140,10 +140,10 @@ return network.registerProtocol('modemmanager', {
 		o.password = true;
 
 		o = s.taboption('general', form.ListValue, 'iptype', _('IP Type'));
-		o.value('ipv4v6', _('IPv4/IPv6 (both - defaults to IPv4)'))
-		o.value('ipv4', _('IPv4 only'));
+		o.value('ipv4', _('IPv4 only'))
```
openwrt-ai: nit: the statement-terminating semicolon is missing. It was missing on the old `ipv4v6` line too and got carried over when the lines were reordered — since these lines are being rewritten anyway, worth fixing. ASI covers it today, but every other `o.value(...)` in this file terminates explicitly.

```suggestion
		o.value('ipv4', _('IPv4 only'));
```

Same applies to line 200 in the `init_iptype` block.

---

protocols/luci-proto-modemmanager/htdocs/luci-static/resources/protocol/modemmanager.js
```diff
@@ -140,10 +140,10 @@ return network.registerProtocol('modemmanager', {
...
 		o.value('ipv6', _('IPv6 only'));
-		o.default = 'ipv4v6';
+		o.value('ipv4v6', _('IPv4/IPv6 (both)'));
```
openwrt-ai, line 145: nit: besides changing the default and the label, this hunk also reorders the dropdown entries (`ipv4v6` moves from first to last, so the visible order changes from both/v4/v6 to v4/v6/both). The commit message mentions only the default alignment and the dropped `- defaults to IPv4` hint. Putting the default first is reasonable, but it is a user-visible change worth one line in the commit body — or drop it to keep the diff strictly to the described fix. Same reordering happens in the `init_iptype` block at lines 200-202.

---

openwrt-ai (commented):
Reviewed 1 new commit. The diagnosis holds up against the code: [`CBIAbstractValue.parse()`](https://github.com/openwrt/luci/blob/8352451ce8984f35c20e69fa20d35d9466b308cb/modules/luci-base/htdocs/luci-static/resources/form.js#L2152) removes the option when `fval == this.default && (this.optional || this.rmempty)`, and [`rmempty` defaults to `true`](https://github.com/openwrt/luci/blob/8352451ce8984f35c20e69fa20d35d9466b308cb/modules/luci-base/htdocs/luci-static/resources/form.js#L1431), so with `default = 'ipv4v6'` the selection could indeed never be persisted. Backend side checks out too: `iptype` and `init_iptype` are declared in `proto_modemmanager_init_config()` and only passed to `mmcli` via `${iptype:+ip-type=${iptype}}`, i.e. omitted entirely when unset, leaving ModemManager's own IPv4 default — matching the `net/modemmanager` README. Two nits inline.

---

micpf:
Both nits addressed:
- Added the missing terminating semicolons on the `ipv4` value lines in both `iptype` and `init_iptype` blocks.
- Noted the entry reorder (new default `ipv4` listed first) explicitly in the commit message.

Force-pushed the updated commit.

openwrt-ai (commented):

The force-push resolves both prior nits and changes nothing else: the delta against `8352451` is exactly the two missing statement-terminating semicolons on the `o.value('ipv4', _('IPv4 only'));` lines (145 and 200) plus the new commit-message paragraph documenting the dropdown reorder. The functional change (`default = 'ipv4'`, dropped `- defaults to IPv4` hint) is unchanged from the version already reviewed. All checks on the head commit are green.

---

feckert:
Thanks merged

PR #8188, opened 2026-01-03
luci-app-rustdesk-server: add new application

luminoso:
luci-app-rustdesk-server: add new application

Add LuCI web interface for managing RustDesk Server on OpenWrt.

This is my first attempt to contribute to Luci. If I've missed anything or any pattern is wrong, please let me know.

> RustDesk is an open-source remote desktop application, providing a self-hosted alternative to proprietary solutions like TeamViewer and AnyDesk. It enables secure remote access without relying on third-party servers.

This package provides configuration for the RustDesk server components:
- hbbs: ID/Rendezvous server for connection brokering
- hbbr: Relay server for NAT traversal

Features:
- UI with live status polling
- Service management (start/stop/restart/enable/disable)
- Public key display with copy button and regeneration
- Firewall rule auto-management based on port settings
- Tabbed configuration for ID Server (hbbs) and Relay Server (hbbr)
- Log viewer with auto-refresh
- Input validation (paths, ports, CIDR notation)
- i18n ready with POT template

Screenshot:
image

Video:
[Screencast_20260103_184715.webm](https://github.com/user-attachments/assets/6c9ae8ce-e958-4c2b-aebc-52b95ab0aa68)

Tasks:

- [x] This PR is not from my *main* or *master* branch :poop:, but a *separate* branch :white_check_mark:
- [x] Each commit has a valid :black_nib: `Signed-off-by: <my@email.address>` row (via `git commit --signoff`)
- [x] Each commit and PR title has a valid :memo: `<package name>: title` first line subject for packages
- [x] Incremented :up: any `PKG_VERSION` in the Makefile
- [x] Tested on:
  - [x] OpenWrt 23.05.5 r24106-10cc5fcd00 / LuCI openwrt-23.05 branch git-25.222.75657-7ce34fe, ipq806x/generic, TP-Link Archer C2600
  - [x] OpenWrt 24.10.5 r29087-d9c5716d1d / LuCI openwrt-24.10 branch 25.340.26705~d88390b, mediatek/filogic, Cudy WR3000S v1
- [x] \( Preferred ) Screenshot or mp4 of changes:
- [x] Description: (describe the changes proposed in this PR)

References:

- https://github.com/jiliukang/RustDesk-Openwrt

PR #8946, opened 2026-08-14
luci-app-rustdesk-server: fixes firmware update causes config loss

### LINKED ISSUE #8935: luci-app-rustdesk-server: retaining configuration updates will revert the configuration

smallprogram:
Fixes #8935
This fixes the issue where configuration settings were lost during a firmware upgrade.

### Description

fixes firmware update causes configuration loss.

---

## Tested on

**OpenWrt version:** OpenWrt 25.12.2
**Web browser(s):** Chrome

---

## Checklist

- [x] _(Nice to have)_ Includes what Issue it closes (e.g. openwrt/luci#issue-number).


applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
```diff
@@ -3,11 +3,13 @@
 # luci-app-rustdesk-server UCI defaults
 # This script runs on first install to set up sensible defaults
 
-uci -q batch <<-EOF >/dev/null
-	set rustdesk-server.@rustdesk-server[0]=rustdesk-server
-	set rustdesk-server.@rustdesk-server[0].enabled='0'
-	set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
-	commit rustdesk-server
+if ! uci -q get rustdesk-server.@rustdesk-server[0] >/dev/null; then
```
openwrt-ai: This guard is never satisfied, so the whole script becomes dead code rather than a first-install path.

The package already ships [`root/etc/config/rustdesk-server`](``https://github.com/openwrt/luci/blob/402beb39a6e11420c48e968db1d5073aa115bb53/applications/luci-app-rustdesk-server/root/etc/config/rustdesk-server#L1-L3``) containing exactly the section and both options this batch would create:

```
config rustdesk-server
	option enabled '0'
	option enabled_relay '0'
```

and it is registered as a conffile in [`Makefile:13-15`](https://github.com/openwrt/luci/blob/402beb39a6e11420c48e968db1d5073aa115bb53/applications/luci-app-rustdesk-server/Makefile#L13-L15). So:

* fresh install — the file is part of the rootfs, the section exists before uci-defaults runs, guard is false;
* sysupgrade with config retention — the conffile is preserved, guard is false.

That leaves no path on which the batch executes (barring a user who manually deleted every section). Since the shipped conffile already provides the defaults, the cleaner fix for #8935 is to drop `root/etc/uci-defaults/50-luci-rustdesk-server` entirely instead of keeping a script that can never run — the conffile mechanism is what actually preserves the user's settings across upgrade.


---

applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
```diff
@@ -3,11 +3,13 @@
...
+	uci -q batch <<-EOF >/dev/null
+		set rustdesk-server.@rustdesk-server[0]=rustdesk-server
+		set rustdesk-server.@rustdesk-server[0].enabled='0'
+		set rustdesk-server.@rustdesk-server[0].enabled_relay='0'
+		commit rustdesk-server
 EOF
```
openwrt-ai: nit: if the script is kept, the here-doc terminator is left flush-left while the body gained a second indent level. `<<-` strips leading tabs from the delimiter line too, so indenting it to match the block reads better and still terminates correctly.

```suggestion
	EOF
```


---

applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
openwrt-ai, line 1: nit: `PKG_RELEASE` in [`Makefile:4`](https://github.com/openwrt/luci/blob/402beb39a6e11420c48e968db1d5073aa115bb53/applications/luci-app-rustdesk-server/Makefile#L4) isn't bumped. This package is one of the few luci apps that pins `PKG_VERSION` explicitly, and [`luci.mk:165`](https://github.com/openwrt/luci/blob/402beb39a6e11420c48e968db1d5073aa115bb53/luci.mk#L165) derives the package version as `$(PKG_VERSION)-r$(PKG_RELEASE)` whenever `PKG_VERSION` is set — the git-derived `PKG_SRC_VERSION` fallback that normally makes bumps unnecessary doesn't apply here.

So this fix ships as `20260313-r1`, the same version already installed, and existing users won't pick it up via `apk`/`opkg upgrade`. Bumping `PKG_RELEASE` to `2` (or moving `PKG_VERSION` to today's date) would make the fix reachable.


---

openwrt-ai (commented):
Reviewed 1 new commit; commit message matches the diff. Three inline comments — the main one is that the new guard can never be true, because the package already ships the same section in `/etc/config/rustdesk-server` as a conffile.

---

applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
smallprogram: I've made the changes as you suggested.

smallprogram:
request review and merge @jow-

applications/luci-app-rustdesk-server/root/etc/uci-defaults/50-luci-rustdesk-server
openwrt-ai, line 1: nit: the package README still documents this file as part of the layout, so the architecture tree no longer matches what ships.

[`README.md:33`](https://github.com/openwrt/luci/blob/29abd530fd742a246c4d902c687931d2bbd5c7e0/applications/luci-app-rustdesk-server/README.md#L33) still has:

```
    │   └── uci-defaults/50-luci-rustdesk-server  # First-run setup
```

Dropping that line (and turning the `init.d/rustdesk-server` entry above it into the last `└──` branch of the `etc/` subtree) would keep the README in sync with the removal.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `402beb3` → `29abd53`); the commit message still matches the diff.

The two substantive points from the previous review are addressed: the never-reachable guard is gone because the whole `uci-defaults` script is removed, and `PKG_VERSION` now moves `20260313` → `20260815` so the fix is actually reachable by `apk`/`opkg upgrade`. The shipped [`root/etc/config/rustdesk-server`](``https://github.com/openwrt/luci/blob/29abd530fd742a246c4d902c687931d2bbd5c7e0/applications/luci-app-rustdesk-server/root/etc/config/rustdesk-server#L1-L3``) plus its `conffiles` registration in [`Makefile:13-15`](https://github.com/openwrt/luci/blob/29abd530fd742a246c4d902c687931d2bbd5c7e0/applications/luci-app-rustdesk-server/Makefile#L13-L15) still provide the defaults on fresh install and preserve user values across sysupgrade, so nothing is lost by the removal.

One documentation leftover flagged inline.

---

smallprogram:
This Claude Code is amazing. @efahl

openwrt-ai (commented):
Reviewed 1 new commit (force-push `29abd53` → `45bc5e0`); no new issues found.

The only delta since the last review is the README architecture tree, which now drops the `uci-defaults/50-luci-rustdesk-server` line and promotes `init.d/rustdesk-server` to the last `└──` branch of the `etc/` subtree — exactly the leftover flagged previously. No remaining references to the removed script anywhere in the package, and the commit message still matches the diff. CI on `45bc5e0` is green (build, eslint, FormalityCheck).

---

efahl (approved):
Looks good to me.

smallprogram:
How can I get this PR merged?

BKPepe (changes requested):
Hi everyone,

I believe that pull request [https://github.com/openwrt/luci/pull/8188](https://github.com/openwrt/luci/pull/8188) shouldn't have been merged in the first place. CC: @luminoso

If AI is going to be used, things should be seen through to the end or even better, implemented so they work out of the box on the first try, rather than half-baking everything.

The problem here is that we're constantly trying to patch leaks as they pop up. Ideally, the config and init script should be packaged alongside the binary, and that binary belongs in the OpenWrt packages repository. Instead, the current approach tells the user: "Install LuCI and it will just work" except it doesn't, because you still have to manually download something elsewhere. That is not user-friendly at all.

I might sound a bit strict and this might not be a popular take, but even this commit https://github.com/openwrt/luci/commit/1b47d9c3535e30404afe90919a8504a4f31c9c6a shouldn't have been merged. It’s the exact kind of flood-damage quick fix I'm talking about; all that's really needed here is conffiles.

Look, this UI either needs a complete overhaul or it should be removed entirely, because I honestly wonder how many people actually use it in its current state.

I know a lot of what I wrote here isn't directly relevant to your change, but if we don't address it now, we never will and I couldn't help but notice once I opened it up and saw what else was tucked away in those files. I'm sorry. I know this isn't strictly on you and that you're just trying to help, but 🤷

luminoso:
@BKPepe 

Thanks for raising this. I still fully plan to package the binaries for rustdesk app so that installing it or the app includes all the needed binaries out of the box. It’s certainly long overdue, and I’m aiming to have this ready in the next month or so, since I have to study how to do packaging for openwrt.

As the keep.d, I haven't worked extensively with keep.d mechanics yet, so if you or anyone else has recommendations on the cleanest implementation, please feel free to share your thoughts. Right now seems that packaging the binaries would solve having the seamless user experience of "install luci and it will just work" done it in one go.

luminoso:
I believe this can be closed in favour of the split the review asked for: the server itself is now a package, openwrt/packages#30572, and the app change that depends on it is https://github.com/openwrt/luci/pull/9058. The uci-defaults and keep.d files are gone from the app.

smallprogram:
> I believe this can be closed in favour of the split the review asked for: the server itself is now a package, [openwrt/packages#30572](https://github.com/openwrt/packages/pull/30572), and the app change that depends on it is #9058. The uci-defaults and keep.d files are gone from the app.

As you wish—it's up to you.

PR #8947, opened 2026-08-15
luci-mod-system, luci-mod-status: permit sysupgrade --create-backup and legacy iptables-save

JuliusBairaktaris:
The firewall status page runs the legacy iptables-save pair whenever those binaries
exist, and only the non-legacy pair is in the session ACL. The `.catch()` covers a
failed `stat()`, so a device that carries the legacy binaries never falls back to the
non-legacy command.

    modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js, load():

            fs.stat('/usr/sbin/iptables-legacy-save').then(function() {
                    return L.resolveDefault(fs.exec_direct('/usr/sbin/iptables-legacy-save'), '');
            }).catch(function() {
                    return L.resolveDefault(fs.exec_direct('/usr/sbin/iptables-save'), '');
            }),

cgi-io matches the requested command against the caller's session ACL and refuses the
exec. `handleCgiIoReply()` in `luci-base/htdocs/luci-static/resources/fs.js` turns the
403 into a `PermissionError` and `L.resolveDefault(..., '')` turns that into an empty
string, so the legacy tables come back empty. The luci-mod-status half is a fix against
current master on its own.

    main.c, main_exec() (cgi-io master):

            /* If there was no ACL match, check for a match on the executable */
            if (!allowed && !session_access(fields[1], "file", exe, "exec")) {
                    free(args);
                    return failure(403, 0, "Access to command denied by ACL");

The luci-mod-system half only takes effect once openwrt/cgi-io#6 lands. On cgi-io
master the archive is created by cgi-io itself, gated on the `cgi-io backup` scope and
not on a file exec grant, so the new entry is unused today.

    main.c, main_backup() (cgi-io master):

            if (!fields[1] || !session_access(fields[1], "cgi-io", "backup", "read"))
                    return failure(403, 0, "Backup permission denied");
            ...
            execl("/sbin/sysupgrade", "/sbin/sysupgrade",
                  "--create-backup", "-", NULL);

openwrt/cgi-io#6 moves that exec to rpcd's file method so it keeps working when uhttpd
runs as a non-root user (openwrt/openwrt#24740), and openwrt/rpcd#40 returns the
child's stdout as a file descriptor so the archive is streamed. rpcd matches on the
command line, so the grant names the exact string, the same form as the
`/sbin/sysupgrade --list-backup` entry already in that ACL. Both PRs are still open.

Depends on openwrt/cgi-io#6 and openwrt/rpcd#40.

I used AI to help me write the change. The empty-table behaviour is read from the
code chain above; I have not reproduced it on a device. Should the luci-mod-status
commit go in on its own now, since it fixes current master, and the
luci-mod-system one wait for cgi-io#6?

### DIFF
--- a/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
+++ b/modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
@@ -94,7 +94,9 @@
 				"/usr/sbin/ip6tables --line-numbers -w -nvxL -t *": [ "exec" ],
 				"/usr/sbin/ip6tables": [ "list" ],
 				"/usr/sbin/iptables-save": [ "exec" ],
-				"/usr/sbin/ip6tables-save": [ "exec" ]
+				"/usr/sbin/ip6tables-save": [ "exec" ],
+				"/usr/sbin/iptables-legacy-save": [ "exec" ],
+				"/usr/sbin/ip6tables-legacy-save": [ "exec" ]
 			},
 			"ubus": {
 				"file": [ "stat" ]
--- a/modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json
+++ b/modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json
@@ -176,7 +176,8 @@
 				"/proc/mtd": [ "read" ],
 				"/proc/partitions": [ "read" ],
 				"/proc/sys/kernel/hostname": [ "read" ],
-				"/sbin/sysupgrade --list-backup": [ "exec" ]
+				"/sbin/sysupgrade --list-backup": [ "exec" ],
+				"/sbin/sysupgrade --create-backup -": [ "exec" ]
 			},
 			"ubus": {
 				"file": [ "exec", "read", "stat" ]


JuliusBairaktaris:
@jow- — review ping.

One-line ACL grant permitting exactly `/sbin/sysupgrade --create-backup -`, in the same `luci-mod-system-flash` read section as the existing `--list-backup` grant. rpcd matches the bare executable first and then `executable + " " + each param`, so this is an exact match with no glob metacharacters.

Needed by openwrt/cgi-io#6, which is in turn needed by openwrt/openwrt#24558.

modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json
```diff
@@ -176,7 +176,8 @@
 				"/proc/mtd": [ "read" ],
 				"/proc/partitions": [ "read" ],
 				"/proc/sys/kernel/hostname": [ "read" ],
-				"/sbin/sysupgrade --list-backup": [ "exec" ]
+				"/sbin/sysupgrade --list-backup": [ "exec" ],
+				"/sbin/sysupgrade --create-backup -": [ "exec" ]
```
openwrt-ai, line 180: The ACL string itself looks right: rpcd checks the bare executable first and then `executable + " " + each param` joined with spaces ([`rpc_file_exec_run()` file.c:1055-1083](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L1055-L1083)), so this is an exact match for today's `execl("/sbin/sysupgrade", "/sbin/sysupgrade", "--create-backup", "-", NULL)` in cgi-io, and the `read` section next to `--list-backup` is the right home for it.

What I can't square is the transport. `--create-backup -` emits a gzip stream on stdout, and rpcd's `file.exec` cannot carry that payload:

- stdout is copied into a NUL-terminated blobmsg **string** ([`rpc_ustream_to_blobmsg()` file.c:889-909](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L889-L909)), so the archive is cut at the first NUL byte;
- the ustream read buffer is capped at `RPC_FILE_MAX_SIZE` = 4096 * 64 = 256 KB ([file.c:46](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L46), [file.c:57-59](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L57-L59)), and filling it aborts the call with `UBUS_STATUS_NOT_SUPPORTED` ([file.c:982-989](https://github.com/openwrt/rpcd/blob/e37ed9d814699098eb7e26c8b33c054840782dfb/file.c#L982-L989));
- rpcd master has no fd-passing/streaming exec variant, so there is no way around the buffer.

That limit is precisely why cgi-io exists in the first place — [`fs.exec_direct()` docs, fs.js:404-407](https://github.com/openwrt/luci/blob/7b773aa325b5d03270d99f89dc42f4e950d18296/modules/luci-base/htdocs/luci-static/resources/fs.js#L404-L407) describe `/cgi-bin/cgi-exec` as the path for "large command outputs which might exceed the ubus message size limits or which contain binary data", and [flash.js:83](``https://github.com/openwrt/luci/blob/7b773aa325b5d03270d99f89dc42f4e950d18296/modules/luci-mod-system/htdocs/luci-static/resources/view/system/flash.js#L83``) posts the backup form straight at `cgi-backup`.

So: does openwrt/cgi-io#6 still spawn and stream the child's stdout to the HTTP client itself, using rpcd only for the privileged spawn / ACL check, or does it go through `ubus call file exec`? If the latter, any backup over 256 KB (i.e. most of them) fails outright and smaller ones come back truncated at the first NUL. I couldn't read that PR from here, so this may already be handled — worth stating in the commit message either way, since the ACL grant is the only visible half.

The same question applies to the `luci-mod-status` hunk in this PR: `iptables-save` / `ip6tables-save` output on a large ruleset also runs past 256 KB.

---

modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
```diff
@@ -94,7 +94,9 @@
 				"/usr/sbin/ip6tables --line-numbers -w -nvxL -t *": [ "exec" ],
 				"/usr/sbin/ip6tables": [ "list" ],
 				"/usr/sbin/iptables-save": [ "exec" ],
-				"/usr/sbin/ip6tables-save": [ "exec" ]
+				"/usr/sbin/ip6tables-save": [ "exec" ],
+				"/usr/sbin/iptables-legacy-save": [ "exec" ],
```
openwrt-ai, line 98: nit: the PR title and description only cover the `luci-mod-system` backup grant; this second commit adds `luci-mod-status` entries that aren't mentioned anywhere in the body. Worth widening the title/description (or splitting), so the merge doesn't look like it touched an unrelated module.

For the record the grant itself checks out: [nftables.js:163-173](``https://github.com/openwrt/luci/blob/7b773aa325b5d03270d99f89dc42f4e950d18296/modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js#L163-L173)`` probes `/usr/sbin/iptables-legacy-save` with `fs.stat()` (covered by the `/*: [ "list" ]` grant in `luci-base`) and then `fs.exec_direct()`s the bare path with no params, which matches the bare-executable check in rpcd. Both alternatives are real install paths ([iptables/Makefile:65 and :485](https://github.com/openwrt/openwrt/blob/fb4c485122985919b974d31f220b94f9609e7e5c/package/network/utils/iptables/Makefile#L65)), and although they are symlinks to `xtables-legacy-multi`, rpcd's exec path only canonicalizes textually and never resolves symlinks, so the grant on the link name is the correct one.

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit messages match their diffs, both files stay valid JSON, and the ACL strings are in the correct `read` sections. Two inline notes — the main one is a question about whether rpcd's `file.exec` can actually carry a backup tarball, since it is fully buffered and returns stdout as a NUL-terminated string.

---

modules/luci-mod-system/root/usr/share/rpcd/acl.d/luci-mod-system.json
JuliusBairaktaris: > So: does openwrt/cgi-io#6 still spawn and stream the child's stdout to the HTTP client itself, using rpcd only for the privileged spawn / ACL check, or does it go through `ubus call file exec`?

It goes through `ubus call file exec`. On cgi-io master the fork and the stream
are cgi-io's own - the child's stdout is a plain pipe spliced straight into the
HTTP body, with no cap on what passes through:

    cgi-io main.c, main_exec() on master:
        default:
            close(fds[1]);
            ...
            len = splice(fds[0], NULL, 1, NULL, READ_BLOCK, SPLICE_F_MORE);

On the openwrt/cgi-io#6 branch the exec and backup paths no longer fork. cgi-io
asks rpcd for the exec with `stream` set and splices the file descriptor that
comes back on the reply; `main_exec()` on that branch has the same shape:

    cgi-io main.c, main_backup() on the openwrt/cgi-io#6 branch:
        blobmsg_add_u8(&buf, "stream", true);
        ...
        len = splice(data.fd, NULL, 1, NULL, READ_BLOCK, SPLICE_F_MORE);

The NUL cut and the 256 KB cap apply to stdout returned as a blobmsg string,
which is why openwrt/rpcd#40 comes first: with `stream` set, stdout goes to an
anonymous memfd handed back as a file descriptor, so none of it passes through
blobmsg.

    rpcd file.c, rpc_file_exec_reply() on the openwrt/rpcd#40 branch:
        if (c->stream && c->memfd >= 0) {
            lseek(c->memfd, 0, SEEK_SET);
            ubus_request_set_fd(c->context, &c->request, c->memfd);

The memfd is not unbounded either. `rpc_file_exec_run()` sets RLIMIT_FSIZE to
16 MB in the child before `execv()`, so the process writing the archive dies
with SIGXFSZ at the limit. sysupgrade runs its tar-into-gzip pipeline under
`set -o pipefail` and exits non-zero, and cgi-io turns a non-zero `code` into a
500 before it writes the first header, so the over-limit case is an error rather
than a truncated archive. I haven't tried a config large enough to hit 16 MB.

> The same question applies to the `luci-mod-status` hunk in this PR:
> `iptables-save` / `ip6tables-save` output on a large ruleset also runs past
> 256 KB.

That hunk stays on cgi-io and is unaffected. `nftables.js` reaches the binaries
through `fs.exec_direct()`, which posts to `/cgi-bin/cgi-exec`, so the output
takes the fork-and-splice path quoted above and never becomes a blobmsg string.
The ACL section says the same thing: `luci-mod-status-firewall` grants
`"cgi-io": [ "exec" ]` and only `"file": [ "stat" ]` on ubus, so there is no
`ubus call file exec` on that path and the 256 KB cap does not apply.

modules/luci-mod-status/root/usr/share/rpcd/acl.d/luci-mod-status.json
JuliusBairaktaris: > nit: the PR title and description only cover the `luci-mod-system` backup grant; this second commit adds `luci-mod-status` entries that aren't mentioned anywhere in the body.

You're right. The title and the description now name both modules.

The commit message was also wrong about why the grant is needed: it is not
pre-emptive for the cgi-io change, it fixes master today. cgi-io master matches
the requested command against the session ACL before it forks:

    cgi-io main.c, main_exec() on master:
        allowed = session_access(fields[1], "file", cmdline, "exec");
        ...
        if (!allowed && !session_access(fields[1], "file", exe, "exec")) {
            free(args);
            return failure(403, 0, "Access to command denied by ACL");

Together with the `/*: [ "list" ]` grant you found in luci-base, the `fs.stat()`
probe succeeds on a device with the legacy binaries, the following
`fs.exec_direct()` gets a 403, and `L.resolveDefault(..., '')` renders an empty
legacy ruleset instead of falling back to the non-legacy pair.

So this commit fixes master on its own and does not depend on openwrt/cgi-io#6
or openwrt/rpcd#40.

openwrt-ai (commented):

---

PR #8948, opened 2026-08-16
luci-mod-system: improve Dropbear interface binding selection

Ser9ei:
### Motivation

The current Dropbear LuCI configuration exposes interface binding through a boolean
"Bind to Interface" option and two different Interface fields:

- when disabled, Interface means: listen on IP addresses assigned to the selected
  network interface;
- when enabled, DirectInterface means: bind directly to the selected network interface.

This works correctly internally, but the UI is confusing because the same-looking Interface selector represents different Dropbear configuration options depending on the state of the flag.

The current UI can therefore be ambiguous:

    Bind to Interface: [x]
    Interface: [lan]

versus:

    Bind to Interface: [ ]
    Interface: [lan]

Although both configurations display an Interface selector, they result in different Dropbear behaviour.

Additionally, the UI flag was persisted as “option _direct” in UCI, although the dropbear service never uses it.

### Description
**New UI**

Replace the boolean "Bind to Interface" option with a three-way "Bind to" selection:

    Bind to
        ( ) All interfaces (unspecified)
        ( ) IP addresses of interface
            Interface: [ lan ]
        ( ) Network interface
            Interface: [ lan ]

The three modes map to the existing Dropbear configuration:

1. All interfaces
   Neither Interface nor DirectInterface is configured.

2. IP addresses of interface
   The existing Interface option is used.

3. Network interface
   The existing DirectInterface option is used.

**Implementation**

The change is limited to the LuCI dropbear configuration.

A virtual "_bind_to" ListValue controls the UI. Its value is derived from the existing Interface/DirectInterface options and is not persisted.

Added a one-time "uci-defaults" migration to remove the legacy "_direct" UI state from existing Dropbear configs.

IMG_0156
IMG_0157
IMG_0158

### Maintainer _(preferred)_
-

---

## Tested on

**LuCI version:** LuCI openwrt-25.12 branch (26.209.73834~b61f907)
**Web browser(s):** Firefox 153.0.4

---

## Checklist

### DIFF
--- a/modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
+++ b/modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
@@ -17,16 +17,39 @@ return view.extend({
 		o = s.option(form.Flag, 'enable', _('Enable Instance'), _('Enable <abbr title="Secure Shell">SSH</abbr> service instance'));
 		o.default  = o.enabled;
 
-		o = s.option(form.Flag, '_direct', _('Bind to Interface'));
-		o.default = o.disabled;
+		// Virtual option: derives UI mode from Interface/DirectInterface,
+		// is not stored in UCI; inactive real options are removed on save
+		o = s.option(form.ListValue, '_bind_to', _('Bind to'), _('Select how the SSH service should be bound to network interfaces or IP addresses'));
+		o.widget = 'radio';
+		o.value('all', _('All interfaces (unspecified)'));
+		o.value('interface', _('IP addresses of interface'));
+		o.value('direct', _('Network interface'));
+		o.default = 'all';
+		o.cfgvalue = function(section) {
+			if (this.section.cfgvalue(section, 'DirectInterface'))
+				return 'direct';
+			if (this.section.cfgvalue(section, 'Interface'))
+				return 'interface';
+			return 'all';
+		};
+		o.forcewrite = true;
+		o.write = function(section) {
+			this.remove(section);
+		};
 
-		o = s.option(widgets.NetworkSelect, 'DirectInterface', _('Interface'), _('Listen only on the given interface or, if unspecified, on all'));
+		o = s.option(widgets.NetworkSelect, 'DirectInterface', _('Interface'), _('Listen only on the given interface'));
 		o.nocreate = true;
-		o.depends('_direct', '1');
+		o.depends('_bind_to', 'direct');
+		o.validate = function(section, value) {
+			return value ? true : _('Please select an interface');
+		};
 
-		o = s.option(widgets.NetworkSelect, 'Interface', _('Interface'), _('Listen on up to 10 IPs on the given interface or, if unspecified, on all interfaces'));
+		o = s.option(widgets.NetworkSelect, 'Interface', _('Interface'), _('Listen on up to 10 IPs on the given interface'));
 		o.nocreate = true;
-		o.depends('_direct', '0');
+		o.depends('_bind_to', 'interface');
+		o.validate = function(section, value) {
+			return value ? true : _('Please select an interface');
+		};
 
 		o = s.option(form.Value, 'Port', _('Port'));
 		o.datatype    = 'port';
--- a/modules/luci-mod-system/root/etc/uci-defaults/90_luci-mod-system-dropbear
+++ b/modules/luci-mod-system/root/etc/uci-defaults/90_luci-mod-system-dropbear
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+. /lib/functions.sh
+
+# remove the "_direct" UI state from existing Dropbear configs
+
+is_commit=0
+
+dropbear_update() {
+	local _direct
+
+	config_get _direct "$1" _direct
+	[ -n "$_direct" ] || return 0
+
+	if uci -q delete dropbear."$1"._direct; then
+		is_commit=1
+	fi
+}
+
+config_load "dropbear"
+config_foreach dropbear_update dropbear
+[ $is_commit -eq 0 ] || uci commit dropbear
+
+exit 0


Ser9ei:
@systemcrash & @jow-, could you please take a look when you have a moment?

This builds on the previous dropbear listen options UI (boolean "Bind to Interface") and replaces it with an explicit three-way selector.

modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
```diff
@@ -17,16 +17,39 @@ return view.extend({
...
+			this.remove(section);
+		};
 
 		o = s.option(widgets.NetworkSelect, 'DirectInterface', _('Interface'), _('Listen only on the given interface or, if unspecified, on all'));
 		o.nocreate = true;
-		o.depends('_direct', '1');
+		o.depends('_bind_to', 'direct');
+		o.validate = function(section, value) {
+			return value ? true : _('Please select an interface');
+		};
```
openwrt-ai, line 45: The dropdown still offers an explicit `unspecified` choice, which now duplicates the new `All interfaces (unspecified)` radio and can only ever produce this validation error. `widgets.NetworkSelect` adds that entry whenever `rmempty` or `optional` is set ([widgets.js:507-508](``https://github.com/openwrt/luci/blob/301bf5e3775c8b4030228411a59fde4672aaa162/modules/luci-base/htdocs/luci-static/resources/tools/widgets.js#L507-L508)),`` and `rmempty` defaults to `true` ([form.js:1431](https://github.com/openwrt/luci/blob/301bf5e3775c8b4030228411a59fde4672aaa162/modules/luci-base/htdocs/luci-static/resources/form.js#L1431)) — neither option sets it here.

Setting `rmempty = false` drops the entry from the choice list *and* makes the option required, so the custom `validate()` becomes redundant: `parse()` already rejects an empty active value with `Option "Interface" must not be empty.` ([form.js:2152-2161](https://github.com/openwrt/luci/blob/301bf5e3775c8b4030228411a59fde4672aaa162/modules/luci-base/htdocs/luci-static/resources/form.js#L2152-L2161)). Clearing on mode switch is unaffected — an inactive option is removed via the `!this.retain` branch ([form.js:2167-2169](https://github.com/openwrt/luci/blob/301bf5e3775c8b4030228411a59fde4672aaa162/modules/luci-base/htdocs/luci-static/resources/form.js#L2167-L2169)) regardless of `rmempty`.

```suggestion
		o.rmempty = false;
```

Same applies to lines 50-52 for `Interface`.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
```diff
@@ -17,16 +17,39 @@ return view.extend({
...
+			if (this.section.cfgvalue(section, 'Interface'))
+				return 'interface';
+			return 'all';
+		};
+		o.forcewrite = true;
+		o.write = function(section) {
+			this.remove(section);
+		};
 
 		o = s.option(widgets.NetworkSelect, 'DirectInterface', _('Interface'), _('Listen only on the given interface or, if unspecified, on all'));
```
openwrt-ai: The `, if unspecified, on all` clause no longer describes reachable behaviour after this change: "unspecified" is now the separate `All interfaces (unspecified)` radio, and the `validate()` added below turns an empty selection into an error rather than a bind-to-all fallback. Leaving the old wording keeps the ambiguity the PR is trying to remove.

```suggestion
		o = s.option(widgets.NetworkSelect, 'DirectInterface', _('Interface'), _('Listen only on the given interface'));
```

Line 47 has the same stale clause — `_('Listen on up to 10 IPs on the given interface or, if unspecified, on all interfaces')` should become `_('Listen on up to 10 IPs on the given interface')`.

---

modules/luci-mod-system/htdocs/luci-static/resources/view/system/dropbear.js
```diff
@@ -17,16 +17,39 @@ return view.extend({
 		o = s.option(form.Flag, 'enable', _('Enable Instance'), _('Enable <abbr title="Secure Shell">SSH</abbr> service instance'));
 		o.default  = o.enabled;
 
-		o = s.option(form.Flag, '_direct', _('Bind to Interface'));
-		o.default = o.disabled;
+		// Virtual option: derives UI mode from Interface/DirectInterface,
+		// is not stored in UCI and clears the real options on change
+		o = s.option(form.ListValue, '_bind_to', _('Bind to'), _('Select how the SSH service should be bound to network interfaces or IP addresses'));
```
openwrt-ai, line 22: nit: the replaced `_direct` flag was actually persisted to `/etc/config/dropbear`, unlike the new `_bind_to`. `form.Flag.parse()` writes any non-default value ([form.js:5153-5156](https://github.com/openwrt/luci/blob/301bf5e3775c8b4030228411a59fde4672aaa162/modules/luci-base/htdocs/luci-static/resources/form.js#L5153-L5156)) and `uci.set()` only filters option names starting with `.`, not `_` ([uci.js:607](https://github.com/openwrt/luci/blob/301bf5e3775c8b4030228411a59fde4672aaa162/modules/luci-base/htdocs/luci-static/resources/uci.js#L607)). So every instance where a user had ticked "Bind to Interface" carries a stale `option _direct '1'` that nothing reads any more (the dropbear init script only consumes `Interface`/`DirectInterface`), and this PR never removes it.

Would it be worth clearing it here, e.g. `this.map.data.unset('dropbear', section, '_direct')` alongside the `this.remove(section)` in the `write` handler below, so configs get tidied up on the next save?

Unrelated to that: the comment says the virtual option "clears the real options on change", but the clearing actually comes from the framework removing `Interface`/`DirectInterface` once their `depends()` stop matching — `write()` here only suppresses `_bind_to` itself.

---

openwrt-ai (commented):
Reviewed 1 new commit. The three-way selector approach looks sound — `_bind_to` is correctly kept out of UCI (the `write` override is needed, since the default `write()` would happily persist `_bind_to` just like the old `_direct` did), the `DirectInterface`-before-`Interface` precedence in `cfgvalue()` matches the dropbear init script's own precedence, and mode switching clears the unused option through the normal inactive-option removal path. No new UCI options are introduced, so there is no backend-coupling gap.

Three inline notes, mainly about the interaction between the new radio and the interface dropdowns.

---

Ser9ei:
Updated:
* kept the custom validation for "Interface" and "DirectInterface", since "rmempty = false" would implicitly select the first interface (and possible first in the list will be not a "lan") instead of leaving the selection unspecified;
* clarified both option descriptions to remove the obsolete "if unspecified, on all";
* added a one-time "uci-defaults" migration to remove the legacy "_direct" UI state from existing Dropbear configs.

modules/luci-mod-system/root/etc/uci-defaults/90_luci-mod-system-dropbear
openwrt-ai, line 1: nit: `/etc/uci-defaults/` is a single flat namespace shared by every installed package, and this filename doesn't say which package owns it — it reads as if it belongs to the `dropbear` package, which already installs its own [`/etc/uci-defaults/50-dropbear`](https://github.com/openwrt/openwrt/blob/00f7f33d2f458484416e87a066ee2e171c1578aa/package/network/services/dropbear/Makefile#L276-L277) from openwrt.git.

Every other uci-defaults file shipped by a LuCI package names its LuCI package: [`50_luci-mod-admin-full`](https://github.com/openwrt/luci/blob/af34b7672c57438baaffaac3bae60268cbbcae64/modules/luci-mod-network/root/etc/uci-defaults/50_luci-mod-admin-full), `40_luci-statistics`, `95-luci-app-banip-housekeeping`, `30_luci-theme-*`, etc. Something like `90_luci-mod-system-dropbear` would keep that pattern and make the owner obvious when the file shows up in `luci-mod-system.list`.

---

modules/luci-mod-system/root/etc/uci-defaults/90_luci-mod-system-dropbear
```diff
@@ -0,0 +1,24 @@
...
+	[ -n "$_direct" ] || return 0
+
+	if uci -q delete dropbear."$1"._direct; then
+		is_commit=1
+	fi
+}
+
+config_load "dropbear"
+config_foreach dropbear_update dropbear
+[ $is_commit -eq 0 ] || uci commit dropbear
```
openwrt-ai, line 22: nit: the `is_commit` bookkeeping doesn't buy anything — both runners of uci-defaults scripts already `uci commit` unconditionally right after sourcing them: [`uci_apply_defaults()` in `/etc/init.d/boot`](https://github.com/openwrt/openwrt/blob/00f7f33d2f458484416e87a066ee2e171c1578aa/package/base-files/files/etc/init.d/boot#L14-L18) on the boot path and [`default_postinst()`](https://github.com/openwrt/openwrt/blob/00f7f33d2f458484416e87a066ee2e171c1578aa/package/base-files/files/lib/functions.sh#L383-L389) on the package-install path. That's why the in-tree precedents for exactly this kind of stale-option cleanup just stage the delete and stop — [`95-luci-app-attendedsysupgrade-housekeeping`](``https://github.com/openwrt/luci/blob/af34b7672c57438baaffaac3bae60268cbbcae64/applications/luci-app-attendedsysupgrade/root/etc/uci-defaults/95-luci-app-attendedsysupgrade-housekeeping#L7-L9)`` does `uci -q delete` with no commit at all, and openwrt's own [`dropbear.defaults`](https://github.com/openwrt/openwrt/blob/00f7f33d2f458484416e87a066ee2e171c1578aa/package/network/services/dropbear/files/dropbear.defaults#L11-L12) relies on the same thing.

An unconditional `uci -q commit dropbear` is a no-op when the delta is empty, so the flag can go (also drops the unquoted `$is_commit` in `[ ]`). Using `config_foreach` is still right here, since `s.addremove = true` means there can be several `dropbear` sections.

```suggestion
dropbear_update() {
	local _direct

	config_get _direct "$1" _direct
	[ -n "$_direct" ] || return 0

	uci -q delete dropbear."$1"._direct
}

config_load "dropbear"
config_foreach dropbear_update dropbear
uci -q commit dropbear
```

---

openwrt-ai (commented):
Re-reviewed after the force-push (`301bf5e` → `af34b76`). The two earlier notes about the stale `or, if unspecified, on all` clauses and the misleading code comment are addressed, and the `rmempty = false` suggestion is fairly answered in [your comment](https://github.com/openwrt/luci/pull/8948#issuecomment-5314652677) — dropping the `unspecified` entry would make `NetworkSelect` preselect whatever network happens to come first, so keeping the custom `validate()` is the better trade. I traced the empty-value path to confirm the custom validator actually fires: `Validator.validate()` only short-circuits on an empty value when the field is *not* optional, otherwise it falls through to `vfunc` ([validation.js:208-236](``https://github.com/openwrt/luci/blob/af34b7672c57438baaffaac3bae60268cbbcae64/modules/luci-base/htdocs/luci-static/resources/validation.js#L208-L236)),`` so "Please select an interface" does get raised.

The new `uci-defaults` migration is correct as written. Worth recording for whoever merges: `_direct` did ship in a release — it is on `openwrt-25.12` via a cherry-pick of `37aa280` — so the migration is not dead code. It also reaches users on both paths (sysupgrade first boot sees the new file since it has no overlay whiteout, and an `apk`/`opkg` upgrade of `luci-mod-system` alone gets it through `default_postinst`). `config_load` handles the case where `/etc/config/dropbear` is absent without erroring, `_direct` survives the `uci export` → `config_set` round-trip fine despite the leading underscore, and the `90_` prefix orders it after `50-dropbear`.

Two nits inline, both on the new script; nothing blocking.

---

Ser9ei:
Update:

Renamed the uci-defaults file to 90_luci-mod-system-dropbear.
Kept the "is_commit + conditional commit" so the script stays self-contained if run manually. No issue simplifying further if a human reviewer prefers that.

openwrt-ai (commented):

The force-push (`af34b76` → `c10f903`) contains only the rename `90_dropbear_remove_direct` → `90_luci-mod-system-dropbear`, which addresses the naming nit; the script body and `dropbear.js` are byte-identical to what I reviewed before. The commit message still matches the diff. Nothing references the old filename, and `luci.mk` installs everything under `root/` automatically, so no `.list`/Makefile update is needed. The 0644 mode is fine — both runners source the file rather than exec it ([`uci_apply_defaults()` in `/etc/init.d/boot`](https://github.com/openwrt/openwrt/blob/00f7f33d2f458484416e87a066ee2e171c1578aa/package/base-files/files/etc/init.d/boot#L14-L15) and [`default_postinst()`](https://github.com/openwrt/openwrt/blob/00f7f33d2f458484416e87a066ee2e171c1578aa/package/base-files/files/lib/functions.sh#L383-L389)) — and matches the majority of in-tree LuCI uci-defaults files.

The `is_commit` nit is [answered in your comment](https://github.com/openwrt/luci/pull/8948#issuecomment-5324020847) and left for a human reviewer to call, so I'm not re-raising it. FormalityCheck is green on `c10f903`.

---

Ser9ei:
CI is waiting for approval (first-time contributor). @hnyman would you
mind approving the workflows when convenient?

PR #8951, opened 2026-08-17
luci-app-librespeed: add new application

BKPepe:
### Description

Web UI for librespeed-cli measurements: a Test page with a live speedometer
and 24-hour chart, a History page with speed/latency series, filters and
CSV/JSON export, and Settings for scheduling and history retention.

Talks to the librespeed-common rpcd backend over ubus. Live progress uses the
client's `--json-stream` when available, with an interface-counter fallback.

image

image

image

### Maintainer _(preferred)_
@BKPepe

---

## Tested on

**OpenWrt version:** OpenWrt SNAPSHOT (r0+35801-c0ffca4c63)
**LuCI version:** LuCI master branch (26.228.64482~c26820a)

---

## Checklist
- [x] _(Nice to have)_ Includes what it depends on (openwrt/packages#30294 in sister repo).

### DIFF
--- a/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
+++ b/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
@@ -1,0 +1,550 @@
+'use strict';
+'require baseclass';
+'require ui';
+
+/* Shared between the Test page's recent-history block and the full History
+ * page, so the two can never draw the same data differently. */
+
+/* Labels are translated here, where the extractor can see them; consumers
+ * take them display-ready, the way RANGES and PHASES do. */
+const METRICS = [
+	[ 'download_mbps', _('Download'), 'Mbps' ],
+	[ 'upload_mbps',   _('Upload'),   'Mbps' ],
+	[ 'ping_ms',       _('Ping'),     'ms'   ],
+	[ 'jitter_ms',     _('Jitter'),   'ms'   ]
+];
+
+/* Series that share a unit share an axis; mixing Mbps and ms on one scale
+ * would flatten the milliseconds into the floor. */
+const GROUPS = {
+	speed:   [ 'download_mbps', 'upload_mbps' ],
+	latency: [ 'ping_ms', 'jitter_ms' ]
+};
+
+/* Labels are translated here, where the extractor can see them: switcher()
+ * takes them display-ready. The key is the seconds, never the label. */
+const RANGES = [
+	[ _('24h'), 86400    ],
+	[ _('7d'),  604800   ],
+	[ _('30d'), 2592000  ],
+	[ _('1y'),  31536000 ],
+	[ _('All'), 0        ]
+];
+
+function esc(s) {
+	return String(s == null ? '' : s).replace(/"/g, '""');
+}
+
+function download(name, mime, text) {
+	const blob = new Blob([ text ], { type: mime }),
+	      url = window.URL.createObjectURL(blob),
+	      link = E('a', { 'style': 'display:none', 'href': url, 'download': name });
+
+	document.body.appendChild(link);
+	link.click();
+	document.body.removeChild(link);
+	window.URL.revokeObjectURL(url);
+}
+
+function metricOf(key) {
+	return METRICS.find(m => m[0] == key);
+}
+
+/* A metric keeps one color everywhere: the first of its group wears the
+ * theme's primary color, the second the fixed companion shade. */
+function colorIndexOf(key) {
+	for (const g in GROUPS) {
+		const i = GROUPS[g].indexOf(key);
+		if (i >= 0)
+			return i;
+	}
+	return 0;
+}
+
+/* Math.min.apply passes the whole series as arguments, and a year of
+ * quarter-hourly samples is more of them than an engine has to accept in one
+ * call -- the chart would throw instead of drawing. */
+function extent(vals) {
+	let min = vals[0], max = vals[0];
+
+	for (let i = 1; i < vals.length; i++) {
+		if (vals[i] < min) min = vals[i];
+		if (vals[i] > max) max = vals[i];
+	}
+
+	return [ min, max ];
+}
+
+/* One scale shared by the renderer and the hover layer: both must agree on
+ * where a sample sits, or the tooltip would point beside the line. Spans
+ * every requested series, and the day's min/max spread when aggregated. */
+function chartScale(entries, metrics, resolution) {
+	const W = 600, H = 240, L = 46, R = 10, T = 14, B = 24;
+
+	if (!entries || entries.length < 1)
+		return null;
+
+	const vals = [];
+
+	entries.forEach(e => {
+		metrics.forEach(k => {
+			if (typeof e[k] == 'number')
+				vals.push(e[k]);
+			if (resolution == '1d') {
+				if (typeof e[k + '_min'] == 'number')
+					vals.push(e[k + '_min']);
+				if (typeof e[k + '_max'] == 'number')
+					vals.push(e[k + '_max']);
+			}
+		});
+	});
+
+	if (!vals.length)
+		return null;
+
+	let [ min, max ] = extent(vals);
+
+	/* Vertical padding: 10% of the spread, but never less than 2% of the
+	 * value itself -- a connection steady within tenths of a Mbps must not
+	 * be zoomed into dramatic noise. A flat zero line still needs a
+	 * nonzero span to be drawable. */
+	const pad = Math.max((max - min) * 0.1, Math.abs(max) * 0.02) || 1;
+	min -= pad; max += pad;
+
+	return {
+		W: W, H: H, L: L, R: R, T: T, B: B,
+		min: min, max: max,
+		xs: i => entries.length > 1
+			? L + i * (W - L - R) / (entries.length - 1)
+			: (L + W - R) / 2,
+		ys: v => T + (H - T - B) * (1 - (v - min) / (max - min))
+	};
+}
+
+/* Timestamps for humans: daily aggregates carry a plain date, everything else
+ * becomes the locale's date and time; garbage shows as itself. */
+function chartStampFull(e, resolution) {
+	if (!e || !e.timestamp)
+		return '?';
+	if (resolution == '1d' || e.timestamp.indexOf('T') < 0)
+		return e.timestamp;
+	const d = new Date(e.timestamp);
+	return isNaN(d.getTime()) ? e.timestamp : d.toLocaleString();
+}
+
+function chartStampShort(e, resolution) {
+	if (!e || !e.timestamp)
+		return '';
+	if (resolution == '1d' || e.timestamp.indexOf('T') < 0)
+		return e.timestamp;
+	const d = new Date(e.timestamp);
+	return isNaN(d.getTime()) ? e.timestamp : d.toLocaleDateString();
+}
+
+function fmtVal(v, unit) {
+	return (typeof v == 'number')
+		? (v.toFixed(unit == 'Mbps' ? 2 : 1) + ' ' + unit) : '–';
+}
+
+return baseclass.extend({
+	METRICS: METRICS,
+	GROUPS: GROUPS,
+	RANGES: RANGES,
+
+	/* Linked plainly, the way the other applications do: uhttpd serves
+	 * static resources with ETag revalidation, so an upgraded stylesheet
+	 * reaches the browser without a cache-busting version to maintain. */
+	cssLink() {
+		return E('link', {
+			'rel': 'stylesheet',
+			'href': L.resource('librespeed/librespeed.css')
+		});
+	},
+
+	/* One token->label table for the schedule intervals: Settings builds
+	 * its choices from it and the Test page looks the status label up, so
+	 * the two cannot drift. The init script accepts more shapes than these
+	 * six, so readers must fall back to the raw token. */
+	INTERVALS: {
+		'15m': _('Every 15 minutes'),
+		'30m': _('Every 30 minutes'),
+		'1h': _('Hourly'),
+		'6h': _('Every 6 hours'),
+		'12h': _('Every 12 hours'),
+		'1d': _('Daily')
+	},
+
+	/* Exported alongside the chart: every timestamp a page prints should
+	 * come through here, or the date-only guard gets lost in a copy. */
+	chartStampFull: chartStampFull,
+
+	/* Plain figures of one series: what is normal, how much it wobbles, and
+	 * where it stands now. The story sentences are built from these. On
+	 * daily aggregates the extremes come from the _min/_max columns the
+	 * chart also draws -- min/max of the daily means would understate the
+	 * spread and contradict the band directly above the sentence. */
+	seriesStats(entries, metric, resolution) {
+		const vals = (entries || []).map(e => e[metric]).filter(v => typeof v == 'number');
+
+		if (!vals.length)
+			return null;
+
+		const avg = vals.reduce((a, b) => a + b, 0) / vals.length;
+		const lows = (resolution == '1d')
+			? (entries || []).map(e => typeof e[metric + '_min'] == 'number'
+				? e[metric + '_min'] : e[metric]).filter(v => typeof v == 'number')
+			: vals;
+		const highs = (resolution == '1d')
+			? (entries || []).map(e => typeof e[metric + '_max'] == 'number'
+				? e[metric + '_max'] : e[metric]).filter(v => typeof v == 'number')
+			: vals;
+
+		return {
+			count: vals.length,
+			avg: avg,
+			min: extent(lows)[0],
+			max: extent(highs)[1],
+			current: vals[vals.length - 1]
+		};
+	},
+
+	/* A row of toggle buttons; the active one is highlighted. The state is
+	 * also spoken: colour alone says nothing to a screen reader, so each
+	 * button carries aria-pressed and the group a name for what it picks. */
+	switcher(view, items, current, onpick, label) {
+		return E('div', { 'role': 'group', 'aria-label': label || null },
+			items.map(it =>
+				E('button', {
+					'class': it[0] == current ? 'cbi-button cbi-button-action' : 'cbi-button',
+					'style': 'margin-right:.3em',
+					'aria-pressed': it[0] == current ? 'true' : 'false',
+					'click': ui.createHandlerFn(view, function() {
+						return onpick.call(view, it[0]);
+					})
+				/* Callers hand in display-ready labels: a runtime _() here
+				 * would both miss the extractor and translate twice. */
+				}, [ it[1] != null ? it[1] : it[0] ])));
+	},
+
+	/* The one chart component: hand-built SVG, the way the realtime status
+	 * graphs do it, colored by the stylesheet so it wears the active theme.
+	 *
+	 * opts.series      keys drawn as lines (one unit family per chart)
+	 * opts.resolution  'raw' or '1d'; aggregates draw their min-max band
+	 * opts.hover       guide line + a full-measurement tooltip
+	 * opts.legend      { all, onToggle }: checkbox legend that switches
+	 *                  series on and off -- a control, not a caption
+	 *
+	 * The x axis is always time: a time series must never be reordered, or
+	 * the story it tells dissolves. Sorting belongs to the table.
+	 * Returns false when there is nothing to draw. */
+	renderChart(container, entries, opts) {
+		container.innerHTML = '';
+
+		const series = (opts.series || []).filter(k => metricOf(k));
+		const sc = chartScale(entries, series, opts.resolution);
+
+		if (!sc) {
+			if (opts.legend)
+				this.renderLegend(container, entries, opts);
+			return false;
+		}
+
+		const W = sc.W, H = sc.H, L = sc.L, R = sc.R, T = sc.T, B = sc.B;
+
+		const grid = [ 0.25, 0.5, 0.75 ].map(f => {
+			const y = T + (H - T - B) * f,
+			      v = sc.max - (sc.max - sc.min) * f;
+			return '<line x1="%d" y1="%f" x2="%d" y2="%f" stroke="currentColor" stroke-opacity="0.15"/>'
+					.format(L, y, W - R, y) +
+				'<text x="%d" y="%f" font-size="10" fill="currentColor" fill-opacity="0.6" text-anchor="end">%s</text>'
+					.format(L - 4, y + 3, v.toFixed(1));
+		}).join('');
+
+		let body = '';
+
+		series.forEach(k => {
+			const ci = colorIndexOf(k);
+			const valid = entries
+				.map((e, i) => (typeof e[k] == 'number') ? [ sc.xs(i), sc.ys(e[k]), e[k] ] : null)
+				.filter(p => p != null);
+
+			if (!valid.length)
+				return;
+
+			const pts = valid.map(p => '%f,%f'.format(p[0], p[1])).join(' ');
+
+			/* Aggregated entries carry the day's spread beside the mean,
+			 * drawn as a band under the mean line. */
+			if (opts.resolution == '1d') {
+				const lo = entries.map((e, i) => (typeof e[k + '_min'] == 'number')
+					? '%f,%f'.format(sc.xs(i), sc.ys(e[k + '_min'])) : null).filter(p => p != null);
+				const hi = entries.map((e, i) => (typeof e[k + '_max'] == 'number')
+					? '%f,%f'.format(sc.xs(i), sc.ys(e[k + '_max'])) : null).filter(p => p != null);
+
+				if (lo.length)
+					body += '<polygon class="librespeed-band librespeed-fill-%d" points="%s %s"/>'
+						.format(ci, lo.join(' '), hi.reverse().join(' '));
+			}
+
+			/* The soft fill reads well under one line; under two it is mud. */
+			if (series.length == 1 && valid.length > 1)
+				body += '<polygon class="librespeed-area librespeed-fill-%d" points="%s %f,%f %f,%f"/>'
+					.format(ci, pts, valid[valid.length - 1][0], H - B, valid[0][0], H - B);
+
+			/* The dashed average is the series' "what is normal" reference. */
+			const avg = valid.reduce((a, p) => a + p[2], 0) / valid.length;
+			body += '<line class="librespeed-avgline librespeed-stroke-%d" x1="%d" y1="%f" x2="%d" y2="%f"/>'
+				.format(ci, L, sc.ys(avg), W - R, sc.ys(avg));
+			body += '<text x="%d" y="%f" font-size="9" fill="currentColor" fill-opacity=".55" text-anchor="end">%h %s</text>'
+				.format(W - R, sc.ys(avg) - 4, _('avg'),
+					avg.toFixed(avg >= 100 ? 0 : 1));
+
+			body += (valid.length > 1)
+				? '<polyline class="librespeed-line librespeed-stroke-%d" points="%s"/>'.format(ci, pts)
+				: '<circle class="librespeed-dot librespeed-fill-%d" cx="%f" cy="%f" r="3"/>'
+					.format(ci, valid[0][0], valid[0][1]);
+
+			/* Dots make a sparse series readable; with hundreds of points
+			 * they would just thicken the line, so they bow out. */
+			if (valid.length > 1 && valid.length <= 60)
+				body += valid.map(p =>
+					'<circle class="librespeed-dot librespeed-fill-%d" cx="%f" cy="%f" r="2"/>'
+						.format(ci, p[0], p[1])).join('');
+		});
+
+		/* A date-only timestamp names a local calendar day; new Date() would
+		 * read it as UTC midnight and shift it in negative offsets. */
+		const first = entries[0],
+		      last = entries[entries.length - 1],
+		      mid = entries[Math.floor((entries.length - 1) / 2)];
+		const s1 = chartStampShort(first, opts.resolution),
+		      s2 = chartStampShort(mid, opts.resolution),
+		      s3 = chartStampShort(last, opts.resolution);
+		const midLabel = (entries.length > 2 && s2 != s1 && s2 != s3)
+			? '<text x="%f" y="%d" font-size="10" fill="currentColor" fill-opacity="0.6" text-anchor="middle">%h</text>'
+				.format((L + W - R) / 2, H - 6, s2)
+			: '';
+
+		const unit = metricOf(series[0]) ? metricOf(series[0])[2] : '';
+
+		const chart = E('div', {});
+		chart.innerHTML =
+			'<svg viewBox="0 0 %d %d" style="width:100%%; max-height:280px" xmlns="http://www.w3.org/2000/svg">'
+				.format(W, H) +
+			grid + body +
+			'<text x="%d" y="%d" font-size="10" fill="currentColor" fill-opacity="0.6">%h</text>'
+				.format(L, H - 6, s1) +
+			midLabel +
+			'<text x="%d" y="%d" font-size="10" fill="currentColor" fill-opacity="0.6" text-anchor="end">%h (%s)</text>'
+				.format(W - R, H - 6, s3, unit) +
+			'</svg>';
+		container.appendChild(chart);
+
+		if (opts.hover !== false)
+			this.attachChartHover(chart, entries, series, opts.resolution, sc);
+
+		if (opts.legend)
+			this.renderLegend(container, entries, opts);
+
+		return true;
+	},
+
+	/* The legend doubles as the series switch: each entry is a checkbox with
+	 * the series' color and its average over the visible range. */
+	renderLegend(container, entries, opts) {
+		const leg = E('div', { 'class': 'librespeed-legend' });
+
+		opts.legend.all.forEach(k => {
+			const m = metricOf(k);
+			const st = this.seriesStats(entries, k, opts.resolution);
+			const cb = E('input', { 'type': 'checkbox' });
+
+			cb.checked = opts.series.indexOf(k) >= 0;
+			cb.addEventListener('change', () => opts.legend.onToggle(k, cb.checked));
+
+			leg.appendChild(E('label', { 'class': 'librespeed-legend-item' }, [
+				cb,
+				E('span', { 'class': 'librespeed-swatch librespeed-bg-%d'.format(colorIndexOf(k)) }),
+				E('span', {}, [ m[1] ]),
+				st ? E('span', { 'class': 'librespeed-muted' },
+					[ ' ' + fmtVal(st.avg, m[2]) + ' ' + _('avg') ]) : ''
+			]));
+		});
+
+		container.appendChild(leg);
+	},
+
+	/* Hover layer: a guide line, a highlighted sample per series, and an HTML
+	 * tooltip carrying the whole measurement -- every metric, the server and
+	 * how the test travelled -- because whoever inspects one point wants to
+	 * know about that test, not one number of it. Daily aggregates show the
+	 * day's spread of the drawn series instead. */
+	attachChartHover(container, entries, series, resolution, sc) {
+		const svg = container.querySelector('svg');
+
+		if (!svg || !sc)
+			return;
+
+		container.classList.add('librespeed-chart-holder');
+
+		const NS = 'http://www.w3.org/2000/svg';
+		const guide = document.createElementNS(NS, 'line');
+		guide.setAttribute('class', 'librespeed-guide');
+		guide.setAttribute('y1', sc.T);
+		guide.setAttribute('y2', sc.H - sc.B);
+		guide.style.display = 'none';
+
+		const dots = series.map(k => {
+			const d = document.createElementNS(NS, 'circle');
+			d.setAttribute('class', 'librespeed-dot librespeed-fill-%d'.format(colorIndexOf(k)));
+			d.setAttribute('r', '3.5');
+			d.style.display = 'none';
+			return d;
+		});
+
+		const hit = document.createElementNS(NS, 'rect');
+		hit.setAttribute('x', sc.L);
+		hit.setAttribute('y', sc.T);
+		hit.setAttribute('width', sc.W - sc.L - sc.R);
+		hit.setAttribute('height', sc.H - sc.T - sc.B);
+		hit.setAttribute('fill', 'transparent');
+
+		svg.appendChild(guide);
+		dots.forEach(d => svg.appendChild(d));
+		svg.appendChild(hit);
+
+		const tip = E('div', { 'class': 'librespeed-tooltip' });
+		container.appendChild(tip);
+
+		const row = (lab, txt) => E('div', { 'class': 'librespeed-tooltip-row' }, [
+			E('span', { 'class': 'librespeed-muted' }, [ lab ]),
+			E('span', {}, [ txt ])
+		]);
+
+		const hide = () => {
+			guide.style.display = 'none';
+			dots.forEach(d => (d.style.display = 'none'));
+			tip.style.display = 'none';
+		};
+
+		hit.addEventListener('mousemove', ev => {
+			const rect = svg.getBoundingClientRect();
+			const x = (ev.clientX - rect.left) * sc.W / rect.width;
+			const step = entries.length > 1
+				? (sc.W - sc.L - sc.R) / (entries.length - 1) : 1;
+			const i = Math.max(0, Math.min(entries.length - 1,
+				Math.round((x - sc.L) / step)));
+			const e = entries[i];
+
+			const px = sc.xs(i);
+			let anchorY = null;
+
+			guide.setAttribute('x1', px);
+			guide.setAttribute('x2', px);
+			guide.style.display = '';
+
+			series.forEach((k, si) => {
+				if (typeof e[k] == 'number') {
+					const py = sc.ys(e[k]);
+					dots[si].setAttribute('cx', px);
+					dots[si].setAttribute('cy', py);
+					dots[si].style.display = '';
+					anchorY = (anchorY == null) ? py : Math.min(anchorY, py);
+				}
+				else
+					dots[si].style.display = 'none';
+			});
+
+			if (anchorY == null)
+				return hide();
+
+			tip.innerHTML = '';
+			tip.appendChild(E('div', { 'style': 'font-weight:600; margin-bottom:.2em' },
+				[ chartStampFull(e, resolution) ]));
+
+			if (resolution == '1d') {
+				series.forEach(k => {
+					const m = metricOf(k);
+					if (typeof e[k] != 'number')
+						return;
+					let txt = fmtVal(e[k], m[2]);
+					if (typeof e[k + '_min'] == 'number' &&
+					    typeof e[k + '_max'] == 'number')
+						txt += ' (%s – %s)'.format(
+							e[k + '_min'].toFixed(m[2] == 'Mbps' ? 0 : 1),
+							e[k + '_max'].toFixed(m[2] == 'Mbps' ? 0 : 1));
+					tip.appendChild(row(m[1], txt));
+				});
+			}
+			else {
+				METRICS.forEach(m => {
+					if (typeof e[m[0]] == 'number')
+						tip.appendChild(row(m[1], fmtVal(e[m[0]], m[2])));
+				});
+
+				if (e.server && e.server.name)
+					tip.appendChild(E('div', { 'style': 'margin-top:.2em' },
+						[ e.server.name ]));
+
+				const meta = [];
+				if (e.interface)
+					meta.push(e.interface);
+				if (e.family)
+					meta.push(e.family == 'ipv6' ? 'IPv6' : 'IPv4');
+				if (e.proto)
+					meta.push(e.proto.toUpperCase());
+				if (meta.length)
+					tip.appendChild(E('div', { 'class': 'librespeed-muted' },
+						[ meta.join(' · ') ]));
+			}
+
+			tip.style.display = 'block';
+
+			const cr = container.getBoundingClientRect();
+			const pxs = rect.left - cr.left + px * rect.width / sc.W;
+			const pys = rect.top - cr.top + anchorY * rect.height / sc.H;
+			let left = pxs - tip.offsetWidth / 2;
+			left = Math.max(0, Math.min(cr.width - tip.offsetWidth, left));
+			tip.style.left = left + 'px';
+			tip.style.top = Math.max(0, pys - tip.offsetHeight - 14) + 'px';
+		});
+		hit.addEventListener('mouseleave', hide);
+	},
+
+	/* Daily aggregates carry each metric's spread beside the mean; the CSV
+	 * must not lose it, so aggregated exports gain the _min/_max columns. */
+	exportCSV(entries, resolution) {
+		const agg = (resolution == '1d');
+		const keys = METRICS.map(m => m[0]);
+		const n = v => (typeof v == 'number') ? v : '';
+
+		let head = 'timestamp,resolution,interface,family,proto,server,' + keys.join(',');
+
+		if (agg)
+			head += ',' + keys.map(k => k + '_min,' + k + '_max').join(',');
+
+		const lines = [ head ];
+
+		(entries || []).forEach(e => {
+			/* Every text field is quoted, so esc()'s quote-doubling is correct
+			 * everywhere and embedded commas cannot shift columns. */
+			let row = '"%s","%s","%s","%s","%s","%s",'.format(
+				esc(e.timestamp), agg ? '1d' : 'raw', esc(e.interface),
+				esc(e.family), esc(e.proto), esc(e.server && e.server.name)) +
+				keys.map(k => n(e[k])).join(',');
+
+			if (agg)
+				row += ',' + keys.map(k => n(e[k + '_min']) + ',' + n(e[k + '_max'])).join(',');
+
+			lines.push(row);
+		});
+
+		download('librespeed-history.csv', 'text/csv', lines.join('\n') + '\n');
+	},
+
+	exportJSON(entries, resolution) {
+		download('librespeed-history.json', 'application/json',
+			JSON.stringify({ resolution: resolution || 'raw', entries: entries || [] }, null, '\t') + '\n');
+	}
+});
--- a/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/librespeed.css
+++ b/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/librespeed.css
@@ -1,0 +1,344 @@
+/* The theme's accent where the theme publishes one: --primary-color-medium
+ * exists only in luci-theme-bootstrap. Every other in-tree theme sees the
+ * fallback, which must clear 3:1 against a light page -- the arc and the
+ * history line are the only graphical encoding of the data. */
+:root {
+	--librespeed-accent: var(--primary-color-medium, #3c8dbc);
+}
+
+/* Dark is an attribute in LuCI, not a media preference: bootstrap stamps
+ * data-darkmode on the root and pins it for its light/dark variants, while
+ * the themes that actually reach the fallback have no dark mode at all --
+ * an OS-level preference must not darken their light pages. */
+:root[data-darkmode="true"] {
+	--librespeed-accent: var(--primary-color-medium, #6ab0de);
+}
+
+/* The one piece of CSS this application owns: layout and type scale. Colors,
+ * borders, buttons and spacing inside sections stay with the theme. */
+
+.librespeed-columns {
+	display: grid;
+	grid-template-columns: minmax(0, 2fr) minmax(18rem, 1fr);
+	gap: 1.5rem;
+	align-items: start;
+}
+
+@media (max-width: 768px) {
+	.librespeed-columns {
+		grid-template-columns: 1fr;
+	}
+}
+
+.librespeed-metrics {
+	display: grid;
+	grid-template-columns: repeat(auto-fit, minmax(7.5rem, 1fr));
+	gap: 1rem;
+	text-align: center;
+	margin: 1rem 0;
+}
+
+.librespeed-value {
+	font-size: 2.2em;
+	font-weight: 500;
+	line-height: 1.2;
+}
+
+.librespeed-value-sm {
+	font-size: 1.5em;
+	line-height: 1.2;
+}
+
+/* Muted through opacity rather than a hardcoded grey, so it dims whatever
+ * foreground color the active theme uses. */
+.librespeed-muted {
+	opacity: .65;
+}
+
+.librespeed-caps {
+	font-size: .85em;
+	letter-spacing: .08em;
+	text-transform: uppercase;
+}
+
+.librespeed-center {
+	text-align: center;
+	padding: 1.5em 1em;
+}
+
+.librespeed-progress {
+	width: 60%;
+	max-width: 20em;
+	display: inline-block;
+	vertical-align: middle;
+	accent-color: var(--librespeed-accent);
+}
+
+/* The chart takes its color from the theme where one is exposed, and falls
+ * back to a neutral accent where none is. */
+.librespeed-chart-line {
+	stroke: var(--librespeed-accent);
+	fill: none;
+	stroke-width: 1.5;
+}
+
+.librespeed-chart-fill {
+	fill: var(--librespeed-accent);
+	fill-opacity: .1;
+	stroke: none;
+}
+
+.librespeed-chart-band {
+	fill: var(--librespeed-accent);
+	fill-opacity: .18;
+	stroke: none;
+}
+
+.librespeed-chart-dot {
+	fill: var(--librespeed-accent);
+}
+
+.librespeed-toolbar {
+	display: flex;
+	flex-wrap: wrap;
+	align-items: center;
+	gap: .5em;
+	margin: .4em 0;
+}
+
+.librespeed-push {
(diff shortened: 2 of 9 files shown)


applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
```diff
@@ -0,0 +1,498 @@
...
+		const chart = E('div', {});
+		chart.innerHTML =
+			'<svg viewBox="0 0 %d %d" style="width:100%%; max-height:280px" xmlns="http://www.w3.org/2000/svg">'
+				.format(W, H) +
+			grid + body +
+			'<text x="%d" y="%d" font-size="10" fill="currentColor" fill-opacity="0.6">%s</text>'
+				.format(L, H - 6, s1) +
+			midLabel +
+			'<text x="%d" y="%d" font-size="10" fill="currentColor" fill-opacity="0.6" text-anchor="end">%s (%s)</text>'
+				.format(W - R, H - 6, s3, unit) +
```
openwrt-ai: `s1`/`s3` are raw backend strings assembled into `innerHTML`. `chartStampShort()` returns `e.timestamp` verbatim whenever it has no `T` or does not parse as a date, so a history entry whose `timestamp` contains markup is injected into the page unescaped. `%s` in LuCI's `String.format` does no escaping — [`case 's': subst = param;` in cbi.js:772](https://github.com/openwrt/luci/blob/4475d333f83828ac6fb8b34b810564e4d0d91d5f/modules/luci-base/htdocs/luci-static/resources/cbi.js#L772-L774) — but [`%h` runs `esc(param, html_esc)`](https://github.com/openwrt/luci/blob/4475d333f83828ac6fb8b34b810564e4d0d91d5f/modules/luci-base/htdocs/luci-static/resources/cbi.js#L783-L785).

The same applies to `s2` in the `midLabel` template on line 276.

```suggestion
			'<text x="%d" y="%d" font-size="10" fill="currentColor" fill-opacity="0.6">%h</text>'
				.format(L, H - 6, s1) +
			midLabel +
			'<text x="%d" y="%d" font-size="10" fill="currentColor" fill-opacity="0.6" text-anchor="end">%h (%s)</text>'
				.format(W - R, H - 6, s3, unit) +
```

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
```diff
@@ -0,0 +1,498 @@
...
+
+		if (agg)
+			head += ',' + keys.map(k => k + '_min,' + k + '_max').join(',');
+
+		const lines = [ head ];
+
+		(entries || []).forEach(e => {
+			let row = '%s,%s,%s,%s,%s,"%s",'.format(
+				esc(e.timestamp), agg ? '1d' : 'raw', esc(e.interface),
+				esc(e.family), esc(e.proto), esc(e.server && e.server.name)) +
```
openwrt-ai: Only the `server` field is wrapped in quotes, but `esc()` (line 30) doubles `"` on *every* field it is applied to. For the unquoted `timestamp`/`interface`/`family`/`proto` columns that doubling is meaningless, and any value containing a comma, a quote or a newline silently shifts the remaining columns of that row. `interface` and `proto` come straight from the history file, so this is not purely hypothetical.

Quoting all six text fields makes `esc()`'s doubling correct everywhere:

```suggestion
		let row = '"%s","%s","%s","%s","%s","%s",'.format(
			esc(e.timestamp), agg ? '1d' : 'raw', esc(e.interface),
			esc(e.family), esc(e.proto), esc(e.server && e.server.name)) +
```

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
```diff
@@ -0,0 +1,498 @@
...
+	/* A row of toggle buttons; the active one is highlighted. */
+	switcher(view, items, current, onpick) {
+		return E('div', {}, items.map(it =>
+			E('button', {
+				'class': it[0] == current ? 'cbi-button cbi-button-action' : 'cbi-button',
+				'style': 'margin-right:.3em',
+				'click': ui.createHandlerFn(view, function() {
+					return onpick.call(view, it[0]);
+				})
+			}, [ _(it[1] != null ? it[1] : it[0]) ])));
```
openwrt-ai: `_()` is called on a runtime value, so xgettext never sees these strings and they can never be translated. Confirmed against the generated template: `msgid "24h"`, `"7d"`, `"30d"` and `"1y"` are all absent from `po/templates/librespeed.pot`, even though `RANGES` (lines 22–28) is fed through here by history.js:253.

It also translates twice: history.js:251 passes `_('Speed')`/`_('Latency')`, i.e. already-localised text, which is then looked up again here.

Either translate at the call sites and drop `_()` here, or keep `_()` here and wrap the `RANGES` labels in literal `_()` calls where the array is declared so the extractor can find them.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,644 @@
...
+			/* The walk through the stages, with the current one highlighted --
+			 * this is what says why nothing moves yet while connecting. */
+			const phaseIdx = PHASES.findIndex(p => p[0] == status.phase);
+
+			rows.push(E('div', { 'class': 'librespeed-steps librespeed-caps' },
+				PHASES.map((p, i) => E('span', {
+					'class': 'librespeed-step' +
+						(i == phaseIdx ? ' librespeed-step-active' :
+							(phaseIdx >= 0 && i < phaseIdx ? ' librespeed-step-done' : ''))
+				}, [ _(p[0]) ]))));
```
openwrt-ai: Same runtime-`_()` problem as common.js:178: `p[0]` is `'ping'`, `'download'` or `'upload'` from `PHASES` (lines 90–94), and none of those msgids exist in `po/templates/librespeed.pot`. The three phase names are prominent user-visible labels in the progress strip, so they will stay English in every locale.

Give `PHASES` a fourth, literal, translatable label element (`_('Ping')`, `_('Download')`, `_('Upload')` — the last two are already extracted from common.js) and render that instead of `_(p[0])`.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/librespeed.css
```diff
@@ -0,0 +1,300 @@
...
+ * by a hardcoded hue -- a dashed line and a striped swatch survive any theme
+ * and read for color-blind users too. */
+.librespeed-stroke-0 { stroke: var(--primary-color, #5bc0de); }
+.librespeed-stroke-1 { stroke: var(--primary-color, #5bc0de); }
+polyline.librespeed-stroke-1 { stroke-dasharray: 6 3; }
+.librespeed-avgline.librespeed-stroke-1 { stroke-dasharray: 2 4; }
+.librespeed-fill-0 { fill: var(--primary-color, #5bc0de); }
+.librespeed-fill-1 { fill: var(--primary-color, #5bc0de); }
+circle.librespeed-fill-1 { fill-opacity: .55; }
+.librespeed-bg-0 { background: var(--primary-color, #5bc0de); }
```
openwrt-ai: `--primary-color` is not defined by any in-tree theme, so every one of these declarations always resolves to the hardcoded `#5bc0de` fallback. That defeats the stated intent here ("One theme color for every series") and in common.js:53–54 ("the first of its group wears the theme's primary color").

The only variables the themes actually publish are [`--primary-color-high/-medium/-low` in luci-theme-bootstrap](``https://github.com/openwrt/luci/blob/4475d333f83828ac6fb8b34b810564e4d0d91d5f/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L93-L96``) (with a [dark-mode override](``https://github.com/openwrt/luci/blob/4475d333f83828ac6fb8b34b810564e4d0d91d5f/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L134-L136));`` material, openwrt and openwrt-2020 define none, so those will keep the fallback regardless.

```suggestion
.librespeed-stroke-0 { stroke: var(--primary-color-medium, #5bc0de); }
.librespeed-stroke-1 { stroke: var(--primary-color-medium, #5bc0de); }
polyline.librespeed-stroke-1 { stroke-dasharray: 6 3; }
.librespeed-avgline.librespeed-stroke-1 { stroke-dasharray: 2 4; }
.librespeed-fill-0 { fill: var(--primary-color-medium, #5bc0de); }
.librespeed-fill-1 { fill: var(--primary-color-medium, #5bc0de); }
circle.librespeed-fill-1 { fill-opacity: .55; }
.librespeed-bg-0 { background: var(--primary-color-medium, #5bc0de); }
```

Same substitution applies to the gradient on line 232 and to `.librespeed-fg-0/-1` (269–270) and `.librespeed-accent-0/-1` (293–294).

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,644 @@
...
+			meta.push(result.proto == 'https' ? _('encrypted (HTTPS)') : _('plain HTTP'));
+		if (result.timestamp)
+			meta.push(fmtTime(result.timestamp));
+
+		const line = E('p', { 'class': 'librespeed-muted', 'style': 'margin-top:.5em' },
+			[ meta.join(' · ') ]);
+
+		/* Only offered when the backend actually returned one -- telemetry may
+		 * be off, in which case there is nothing to link to. */
+		if (result.share) {
```
openwrt-ai: `result.share` goes straight into an `href` with no scheme check. It originates from the LibreSpeed telemetry server's reply (relayed by the rpcd backend), i.e. from outside the router, so a hostile or compromised telemetry endpoint can return `javascript:…` and get it executed in the LuCI origin — with the session already authenticated — as soon as the user clicks "View result".

Restricting it to http(s) at the guard costs nothing:

```suggestion
		if (result.share && /^https?:\/\//i.test(result.share)) {
```

---

applications/luci-app-librespeed/po/templates/librespeed.pot
```diff
@@ -0,0 +1,535 @@
+msgid ""
+msgstr "Content-Type: text/plain; charset=UTF-8"
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js:216
```
openwrt-ai: nit: the template is stale — every reference into the three view files is off by exactly 4 lines, which is the length of the `/* The alias comes from the require above… */` + `/* global lscommon */` block that was added to each of them afterwards.

`%d measurements` is cited as `test.js:216` but lives at test.js:220; `All` as `history.js:71` but lives at history.js:75; `1 year` as `settings.js:125` but lives at settings.js:129. References into common.js (e.g. `avg` at common.js:322) are correct, since that file has no such block.

Please regenerate the template from the current sources.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,644 @@
...
+							(phaseIdx >= 0 && i < phaseIdx ? ' librespeed-step-done' : ''))
+				}, [ _(p[0]) ]))));
+
+			/* The dial range starts from the last known result -- a familiar
+			 * line gets a familiar dial from the first tick -- and only grows
+			 * when this run's peak outruns it. */
+			/* The dial range is frozen at the start of the run -- a familiar
+			 * line gets a familiar dial -- and only ever expands, once the
+			 * rate actually exceeds it. Recomputing it downwards mid-run
+			 * would pull the needle back while the speed grows. */
```
openwrt-ai: nit: two comment blocks describing the same three lines of code, and they contradict each other — the first says the dial range "starts from the last known result … and only grows", the second says it "is frozen at the start of the run". The code below matches the second. Looks like a leftover from a rewrite.

```suggestion
			/* The dial range is frozen at the start of the run -- a familiar
			 * line gets a familiar dial -- and only ever expands, once the
			 * rate actually exceeds it. Recomputing it downwards mid-run
			 * would pull the needle back while the speed grows. */
```

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/settings.js
```diff
@@ -0,0 +1,175 @@
+'use strict';
+'require view';
+'require form';
+'require rpc';
+'require tools.widgets as widgets';
+'require librespeed.common as lscommon';
+
+/* The alias comes from the require above; the repo eslint config only
+ * knows the stock module names. */
+/* global lscommon */
```
openwrt-ai: nit: `lscommon` is never referenced in this file — unlike test.js and history.js, settings.js uses neither `lscommon.cssLink()` nor any of the chart helpers. The `require` and the accompanying `/* global */` pragma can both go.

---

openwrt-ai (commented):
Reviewed 1 new commit; findings posted inline.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
```diff
@@ -0,0 +1,502 @@
...
+		return E('div', {}, items.map(it =>
+			E('button', {
+				'class': it[0] == current ? 'cbi-button cbi-button-action' : 'cbi-button',
+				'style': 'margin-right:.3em',
+				'click': ui.createHandlerFn(view, function() {
+					return onpick.call(view, it[0]);
+				})
+			/* Callers hand in display-ready labels: a runtime _() here
+				 * would both miss the extractor and translate twice. */
+				}, [ it[1] != null ? it[1] : it[0] ])));
```
openwrt-ai: The new comment block landed at the wrong indent depth: the opening `/*` sits at three tabs, its continuation line and the closing `}` at four. Everything here closes the object literal opened by `E('button', {` at three tabs, so all three lines belong at three tabs.

```suggestion
			/* Callers hand in display-ready labels: a runtime _() here
			 * would both miss the extractor and translate twice. */
			}, [ it[1] != null ? it[1] : it[0] ])));
```

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
```diff
@@ -0,0 +1,502 @@
...
+	speed:   [ 'download_mbps', 'upload_mbps' ],
+	latency: [ 'ping_ms', 'jitter_ms' ]
+};
+
+const RANGES = [
+	[ '24h', 86400    ],
+	[ '7d',  604800   ],
+	[ '30d', 2592000  ],
+	[ '1y',  31536000 ],
+	[ 'All', 0        ]
```
openwrt-ai: Dropping the runtime `_()` from `switcher()` fixed the extractor problem for the group buttons, but only one of the two call sites was updated to compensate. [history.js:271](``https://github.com/openwrt/luci/blob/1d1d9420561ce3908f96350683d4970e222fcb42/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js#L271)`` now passes `_('Speed')`/`_('Latency')`, but [history.js:274](``https://github.com/openwrt/luci/blob/1d1d9420561ce3908f96350683d4970e222fcb42/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js#L274)`` feeds `r[0]` from `RANGES` through unchanged, so the five range buttons are now hard-wired English.

`All` is a visible regression rather than just a missed opportunity: the msgid exists in the template (extracted from [`_('All')` in `filterSelect`, history.js:75](``https://github.com/openwrt/luci/blob/1d1d9420561ce3908f96350683d4970e222fcb42/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js#L75)),`` so before this change the range button did get translated at runtime and now no longer does — while the filter dropdown right next to it still does.

`r[0]` is only ever used as a display label (`switcher`'s key is `String(r[1])`), so wrapping it here is safe, and it matches what the `PHASES` table in test.js:90-94 now does:

```suggestion
	[ _('24h'), 86400    ],
	[ _('7d'),  604800   ],
	[ _('30d'), 2592000  ],
	[ _('1y'),  31536000 ],
	[ _('All'), 0        ]
```

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js
```diff
@@ -0,0 +1,354 @@
...
+		this.controls = E('div', {});
+
+		/* ui.Table gives sortable headers for free -- every cell below is a
+		 * [sort key, display] pair, so Time orders by epoch even though it
+		 * shows a locale string, and numbers order numerically. The table is
+		 * where sorting belongs; the chart above never reorders time. */
+		this.table = new ui.Table([
+			_('Time'), _('Server'), _('Interface'), _('Protocol'),
+			_('Download [Mbps]'), _('Upload [Mbps]'), _('Ping [ms]'), _('Jitter [ms]')
+		], { id: 'librespeed-history' }, E('em', {}, [ _('No measurements yet.') ]));
```
openwrt-ai: nit: this `ui.Table` placeholder is unreachable now that the empty state exists. The table lives inside `dataNode`, and `redraw()` returns at line 184-192 with `dataNode` hidden whenever `entries.length` is 0, so the table is never rendered without rows.

It also leaves translators with two near-identical msgids differing only by the trailing period — `"No measurements yet."` here and `"No measurements yet"` from the new empty-state heading at line 320. Dropping the placeholder argument removes both problems.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit against the previous head (`4475d33`). The XSS fix on the chart labels, the CSV quoting, the `https?:` scheme guard on the share link, the `--primary-color-medium` switch, the `PHASES` label table, the stale duplicate comment and the unused `librespeed.common` require in settings.js are all addressed, and the `.pot` line references now match the sources. Three follow-ups posted inline; CI is green on `1d1d942`.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,643 @@
...
+
+		const drew = lscommon.renderChart(chart, entries, {
+			series: [ 'download_mbps' ],
+			resolution: data && data.resolution,
+			hover: true
+		});
+
+		if (!drew) {
+			this.recentNode.appendChild(E('p', { 'class': 'librespeed-muted' },
+				[ _('No measurements yet') ]));
```
openwrt-ai: The history page just got this exact branch relabelled (`No data for the selected metric.`, history.js:211) with the reasoning that reaching `!drew` does not mean "no measurements" — but the identical branch here kept the old wording, and here it is actually contradicted on screen.

`renderChart` returns false whenever [`chartScale` finds no numeric values for the requested series](``https://github.com/openwrt/luci/blob/ca267d5ff028f8732ade61fb0800af74fe305a17/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js#L90-L91),`` not only when `entries` is empty. This block requests only `download_mbps`, so a 24-hour window holding measurements without a download figure lands here while [`this.recentCount` two lines up still reads "%d measurements"](``https://github.com/openwrt/luci/blob/ca267d5ff028f8732ade61fb0800af74fe305a17/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js#L219-L220)`` and the section subtitle says `Download · last 24 hours`. Older measurements outside the window produce the same claim.

Naming the window also restores the sentence period that was dropped only to share a msgid with the History page's `<h?>`-style empty-state heading (history.js:322), where no period is correct.

```suggestion
				[ _('No download data in the last 24 hours.') ]));
```

The template needs regenerating afterwards so this msgid and its reference replace the shared `No measurements yet` entry.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
```diff
@@ -0,0 +1,504 @@
...
+'require ui';
+
+/* Shared between the Test page's recent-history block and the full History
+ * page, so the two can never draw the same data differently. */
+
+const METRICS = [
+	[ 'download_mbps', 'Download', 'Mbps' ],
+	[ 'upload_mbps',   'Upload',   'Mbps' ],
+	[ 'ping_ms',       'Ping',     'ms'   ],
+	[ 'jitter_ms',     'Jitter',   'ms'   ]
```
openwrt-ai: nit: `RANGES` (line 24) and `PHASES` in test.js now carry literal `_()` so the extractor sees them, but `METRICS` is the one label table left holding bare literals that are fed through a runtime `_()` — at [common.js:324](``https://github.com/openwrt/luci/blob/ca267d5ff028f8732ade61fb0800af74fe305a17/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js#L324),`` [431](``https://github.com/openwrt/luci/blob/ca267d5ff028f8732ade61fb0800af74fe305a17/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js#L431),`` [437](``https://github.com/openwrt/luci/blob/ca267d5ff028f8732ade61fb0800af74fe305a17/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js#L437),`` [history.js:143](``https://github.com/openwrt/luci/blob/ca267d5ff028f8732ade61fb0800af74fe305a17/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js#L143)`` and [history.js:157](``https://github.com/openwrt/luci/blob/ca267d5ff028f8732ade61fb0800af74fe305a17/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js#L157).``

These four happen to translate today only because the same words are extracted from unrelated places in test.js — the template credits `Ping` to test.js:91/481 and `Jitter` to test.js:482 alone, nothing in common.js. Rewording the dial captions or the phase table would silently turn every metric label in the charts, tooltips and legend cards back to English.

```suggestion
	[ 'download_mbps', _('Download'), 'Mbps' ],
	[ 'upload_mbps',   _('Upload'),   'Mbps' ],
	[ 'ping_ms',       _('Ping'),     'ms'   ],
	[ 'jitter_ms',     _('Jitter'),   'ms'   ]
```

The five `_(m[1])` call sites above should then drop their `_()` and pass `m[1]` straight through, exactly as `switcher()` was changed to do, and the template regenerated.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit against the previous head (`1d1d942`). All three earlier follow-ups are addressed: the `switcher()` comment indentation is fixed, `RANGES` now carries literal `_()` so the range buttons are extractable and translated again, and the unreachable `ui.Table` placeholder is gone. I re-validated the template mechanically — all 157 source references resolve to the right line, there are no missing msgids, and no orphans. CI is green on `ca267d5`.

Two follow-ups posted inline, both about the same relabelling pass.

---

BKPepe:
Hey @1715173329, since you're maintaining the librespeed-go backend in the openwrt/packages repository, what do you think about adding the backend configuration here?

I'd like to do that in an upcoming version once this is merged, to avoid making more changes right now. For the time being, I'd like to keep this PR focused on the client.

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js
```diff
@@ -0,0 +1,403 @@
...
+
+		const rows = entries.slice().reverse();
+		const pages = Math.max(1, Math.ceil(rows.length / pageSize));
+
+		if (this.page >= pages)
+			this.page = pages - 1;
+
+		this.renderPager(rows.length, pages);
+
+		this.table.update(rows.slice(this.page * pageSize, (this.page + 1) * pageSize).map(e => [
```
openwrt-ai: Paging the rows before handing them to `ui.Table` turns the sortable headers into a per-page sort. `update()` stores exactly what it is given — [`this.data = data;` at ui.js:3892](https://github.com/openwrt/luci/blob/ed953ce028957cb0858cdcb3d0eef42f50989bef/modules/luci-base/htdocs/luci-static/resources/ui.js#L3892) — and a header click re-sorts only that stored array: [`this.update(this.data, this.placeholder);` at ui.js:4137](https://github.com/openwrt/luci/blob/ed953ce028957cb0858cdcb3d0eef42f50989bef/modules/luci-base/htdocs/luci-static/resources/ui.js#L4137).

So on a 30-day range with ~3000 measurements, clicking `Download [Mbps]` descending gives the fastest of the 50 newest rows, not the fastest of the range, and the pager label right below still says `3000 measurements`. The comment at line 306-309 above the table promises the opposite ("The table is where sorting belongs"), and before this change it held, because the whole set was passed in.

Either sort `rows` yourself with the table's active sort state before slicing (`this.table.getActiveSortState()` plus `deriveSortKey` give you the same ordering `update()` would apply), or construct the table with `sortable: false` so the headers stop advertising an ordering they cannot deliver.

---

applications/luci-app-librespeed/po/templates/librespeed.pot
```diff
@@ -0,0 +1,593 @@
...
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js:318
+msgid "Latency"
+msgstr ""
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/settings.js:72
+msgid "Leave empty to run every day."
+msgstr ""
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js:684
```
openwrt-ai: nit: the template was not regenerated after the last edit to `test.js`. Everything up to `Schedule status` (test.js:657) still matches, but the four references past the point where the three-line `/* Heading and count on one line… */` comment was inserted are now off by two or three lines:

| msgid | template says | actual |
| --- | --- | --- |
| `LibreSpeed` | test.js:684 | test.js:687 |
| `Recent history` | test.js:673 (line 363) | test.js:675 |
| `Router speed test` | test.js:685 (line 371) | test.js:688 |
| `Measures the internet connection from this router.` | test.js:687 (line 267) | test.js:690 |

No msgids are missing and there are no orphans, so it is only the source references — please regenerate the template from the current sources so they point at the right lines again.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,701 @@
...
+		 * or focus. data-tooltip is LuCI's own mechanism -- a plain title
+		 * attribute leaves it to the browser, which shows nothing on some --
+		 * and the dotted underline is the affordance, since the theme styles
+		 * tooltips only inside form fields. The panel stays a compact status
+		 * list, and the reasoning still lives where the choice is made. */
+		const kv = rows => E('table', { 'class': 'table' },
+			rows.map(r => E('tr', { 'class': 'tr' }, [
+				E('td', { 'class': 'td librespeed-muted', 'style': 'width:45%' }, [ r[0] ]),
+				E('td', { 'class': 'td' }, [
+					r[2] ? E('span', { 'class': 'librespeed-hint', 'data-tooltip': r[2] }, [ r[1] ]) : r[1]
```
openwrt-ai: The comment above promises "on hover **or focus**", but a bare `<span>` is not focusable, so the focus half never fires. LuCI installs both listeners globally — [`document.addEventListener('focus', this.showTooltip.bind(this), true)` at ui.js:4178](https://github.com/openwrt/luci/blob/ed953ce028957cb0858cdcb3d0eef42f50989bef/modules/luci-base/htdocs/luci-static/resources/ui.js#L4176-L4179) — but the element has to be reachable by the tab order for the event to ever reach it. Today the only explanation of `HTTP (forced)` / `HTTPS (forced)` is mouse-only.

```suggestion
					r[2] ? E('span', { 'class': 'librespeed-hint', 'tabindex': '0', 'data-tooltip': r[2] }, [ r[1] ]) : r[1]
```

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,701 @@
...
+		 * comparison instead. History is where series get combined. */
+		this.recentSwitch = lscommon.switcher(this,
+			lscommon.GROUPS.speed.map(k => {
+				const m = lscommon.METRICS.find(x => x[0] == k);
+				return [ k, m[1] ];
+			}), this.recentMetric,
+			function(v) {
+				this.recentMetric = v;
+				this.renderRecent(this.recentData);
+			});
```
openwrt-ai: nit: this switcher is discarded before it is ever seen. `renderRecent()` runs eight lines down at line 586 and its first act on the node is `this.recentSwitch.replaceChildren(...lscommon.switcher(...).childNodes)` (lines 227-235), rebuilding the identical button row from the identical `GROUPS.speed.map(...)` expression. The wrapper `<div>` carries no class or style, so nothing distinguishes the two copies — they just both have to be kept in sync by hand.

An empty container is enough; `renderRecent()` fills it on every call, including the first.

```suggestion
		this.recentSwitch = E('div', {});
```

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit against the previous head (`ca267d5`). Both earlier follow-ups are addressed: `METRICS` now carries literal `_()` and the five `_(m[1])` call sites pass the label straight through, and the Test page's empty-chart branch was reworded — the metric switcher makes the metric-neutral "No data in the last 24 hours." the right wording there. The new work in this push is the recent-history metric switcher, the `ui.Table` swap on the Test page, History pagination, and the tooltip-bearing configuration panel. CI is green on `ed953ce`.

Four follow-ups posted inline: one behavioural regression from the new paging, one stale template, and two smaller ones.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js
```diff
@@ -0,0 +1,428 @@
...
+		 * count below still speaks for every measurement. */
+		let rows = entries.slice().reverse().map(e => [
+			[ e.epoch ?? 0, when(e) ],
+			(e.server && e.server.name) || '–',
+			e.interface || '–',
+			protoCell(e),
+			[ num(e.download_mbps), fmtNum(e.download_mbps, 2) ],
+			[ num(e.upload_mbps), fmtNum(e.upload_mbps, 2) ],
+			[ num(e.ping_ms), fmtNum(e.ping_ms, 1) ],
+			[ num(e.jitter_ms), fmtNum(e.jitter_ms, 1) ]
```
openwrt-ai: Now that sorting drives the whole range rather than one page, it is worth noting that the four numeric columns order wrongly whenever two values share an integer part.

`deriveSortKey` is handed the entire `[raw, display]` cell — [`this.deriveSortKey(row[sorting[0]], sorting[0])` at ui.js:3869](https://github.com/openwrt/luci/blob/fbeecb75e5b538e3a574c9c70f3f84a1e70ef493/modules/luci-base/htdocs/luci-static/resources/ui.js#L3869), mirrored by the block just below — and no branch in it matches an array, so it ends at [`return String(value)` at ui.js:4052-4057](https://github.com/openwrt/luci/blob/fbeecb75e5b538e3a574c9c70f3f84a1e70ef493/modules/luci-base/htdocs/luci-static/resources/ui.js#L4052-L4057), producing `"94.4,94.40"`. `L.naturalCompare` is [`new Intl.Collator(undefined, { numeric: true }).compare` at luci.js:2246](https://github.com/openwrt/luci/blob/fbeecb75e5b538e3a574c9c70f3f84a1e70ef493/modules/luci-base/htdocs/luci-static/resources/luci.js#L2246), which compares each digit run on its own, so the fraction `4` loses to `35`:

```
-1,–  0.5,0.50  0.45,0.45  8.07,8.07  8.7,8.70  9.5,9.50  94.4,94.40  94.35,94.35  120,120.00
```

Ascending Download puts 94.4 before 94.35 and 0.5 before 0.45; Ping and Jitter do the same with one decimal (8.7 before 8.07). The manual sort here and `ui.Table.update()` agree with each other — they just agree on the wrong order, so the range-wide sort inherits it.

The raw half only needs the same fixed decimal count as the display half; equal-width fraction runs then compare correctly. `num` (history.js:223) has no other caller:

```js
const num = (v, d) => (typeof v == 'number') ? v.toFixed(d) : '';
```

with the cells becoming `[ num(e.download_mbps, 2), fmtNum(e.download_mbps, 2) ]` and so on — line 252 needs it too, not just the three anchored here. That yields `,–  0.45  0.50  8.07  8.70  9.50  94.35  94.40  120.00`.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js
```diff
@@ -0,0 +1,428 @@
...
+				.map(x => x[1]);
+
+		const pages = Math.max(1, Math.ceil(rows.length / pageSize));
+
+		if (this.page >= pages)
+			this.page = pages - 1;
+
+		this.renderPager(rows.length, pages);
+		this.table.update(rows.slice(this.page * pageSize, (this.page + 1) * pageSize));
+
```
openwrt-ai: nit: stray blank line before the closing brace — left over from moving the pager and `update()` call down here. It is the only one in the package; every other method in the four JS files closes straight after its last statement.

```suggestion
		this.table.update(rows.slice(this.page * pageSize, (this.page + 1) * pageSize));
```

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit against the previous head (`ed953ce`). All four earlier follow-ups are addressed:

- the paging/sorting regression is fixed — `redraw()` now sorts the whole range with the table's own key derivation before slicing, and the new header-click listener on `tableNode` resets to page 1 and re-sorts the range. The listener ordering works out: `handleSort` is bound on the `tr.table-titles` child and `createHandlerFn` applies it synchronously, so the bubbled listener on the `<table>` already sees the new sort state;
- the template is regenerated. I re-validated it mechanically against the current sources — all 169 source references resolve to the right line, no msgids are missing, and the only entries without a JS literal are the three that come from the ACL and menu JSON;
- the configuration-panel hints carry `tabindex`, so the focus half of the tooltip fires;
- the duplicate `recentSwitch` is gone, and the plain `E('div', {})` matches what `switcher()` returns anyway (its wrapper carries no class), so nothing is lost by letting `renderRecent()` fill it.

CI is green on `fbeecb7`. Two follow-ups posted inline, one on the numeric sort keys the new range-wide sort now relies on, one nit.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,743 @@
...
+			}), this.recentMetric,
+			function(v) {
+				this.recentMetric = v;
+				this.renderRecent(this.recentData);
+			}, _('Metric')).childNodes);
+
+		/* The stability story of the last day, told in one sentence and four
+		 * figures: what is normal, how much it wobbles, where it is now. */
+		const metric = this.recentMetric || 'download_mbps';
+		const st = lscommon.seriesStats(entries, metric);
```
openwrt-ai: `seriesStats()` grew a third `resolution` parameter in this push, and every other call site was updated to pass it — [history.js:122-123](``https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js#L122-L123)`` and [common.js:350](``https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js#L350)`` — but this one still calls it with two arguments, so `resolution` is `undefined` and the `'1d'` branch never runs.

The Test page does treat `1d` as reachable for this window: the chart just below is given `resolution: data && data.resolution` (line 270) and `when()` passes it to the date-only guard (line 304). When the backend answers the 24-hour window with daily aggregates, `st.min`/`st.max` here fall back to the min/max of the daily *means*, so `Variable — ranged from %d to %d Mbps.` and the `best`/`worst` figures at lines 295-296 understate the spread against the min/max band the chart directly above them draws — the exact contradiction the new comment on [`seriesStats`, common.js:154-158](``https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js#L154-L158)`` says the parameter exists to prevent.

```suggestion
		const st = lscommon.seriesStats(entries, metric, data && data.resolution);
```

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/librespeed.css
```diff
@@ -0,0 +1,341 @@
...
+ * history line are the only graphical encoding of the data. */
+:root {
+	--librespeed-accent: var(--primary-color-medium, #3c8dbc);
+}
+
+@media (prefers-color-scheme: dark) {
+	:root {
+		--librespeed-accent: var(--primary-color-medium, #6ab0de);
+	}
+}
```
openwrt-ai, line 15: `prefers-color-scheme` is not how a LuCI page signals that it is dark, so this override fires at the wrong times.

luci-theme-bootstrap keys its dark palette off an attribute — [`:root[data-darkmode="true"]` at cascade.css:123](``https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L123``) — and only the auto variant ties that attribute to the OS. [header.ut:12](``https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut#L12``) pins `darkpref` to `'true'`/`'false'` for `bootstrap-dark`/`bootstrap-light`, and the [`matchMedia` block at header.ut:22-31](``https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut#L22-L31``) is then not emitted at all. luci-theme-material, -openwrt and -openwrt-2020 have no dark mode in any form.

Those are precisely the themes that reach the fallback. On material/openwrt/openwrt-2020 with the OS set to dark, this block swaps in `#6ab0de` on a permanently white page — the low-contrast case the comment above says the fallback exists to avoid — and `bootstrap-light` on a dark OS does the same. Conversely `bootstrap-dark` on a light OS keeps `#3c8dbc`.

Keying on the attribute covers every case, the auto variant included, since its script sets `data-darkmode` from that same media query:

```suggestion
/* The theme's accent where the theme publishes one: --primary-color-medium
 * is defined by luci-theme-bootstrap (and luci-theme-footstrap on current
 * master); the other in-tree themes see the fallback, which must clear 3:1
 * against the page -- the arc and the history line are the only graphical
 * encoding of the data. Dark mode is the data-darkmode attribute, not the
 * OS preference: bootstrap-dark/-light pin it regardless of
 * prefers-color-scheme, and material/openwrt/openwrt-2020 are light-only. */
:root {
	--librespeed-accent: var(--primary-color-medium, #3c8dbc);
}

:root[data-darkmode="true"] {
	--librespeed-accent: var(--primary-color-medium, #6ab0de);
}
```

Side note on the comment's premise: the branch is based on a master that predates luci-theme-footstrap, which also defines the variable — [footstrap cascade.css:12](``https://github.com/openwrt/luci/blob/459941fb2fdb7365b0169ea06e9148afdd0319e0/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L12``) on this PR's base commit — and it uses `data-darkmode` as well.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js
```diff
@@ -0,0 +1,462 @@
...
+		/* Without data the empty state IS the page: everything that only
+		 * makes sense with measurements -- cards, chart, table, export --
+		 * stays out of the way entirely, and the one useful action is a
+		 * button to go run a test. Filters that merely exclude everything
+		 * get their own wording and no button: data exists, go widen them. */
+		const noneAtAll = !this.entries.length;
+
+		if (!entries.length) {
+			const histOff = noneAtAll && this.config &&
+				this.config.history && this.config.history.enabled === false;
```
openwrt-ai, line 210: This requires `enabled` to be a real JavaScript `false`, but the other reader of the same `config` payload in this PR assumes something weaker: [`sched.enabled ? _('Yes') : _('No')` at test.js:666](``https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js#L666)`` is plain truthiness. The two disagree on the encoding, and at most one of them survives whatever the backend actually emits:

- backend emits a boolean — both are right;
- backend emits `0`/`1` — test.js is right, this branch never fires and the empty state keeps offering `Start test` on a router that will never record one, which is the whole point of the new `callConfig()` in `load()`;
- backend passes UCI's raw `'0'`/`'1'` through — this branch never fires *and* test.js reports a disabled schedule as `Yes`.

`librespeed-common` is not in the packages feed yet (openwrt/packages#30294), so I could not check which shape `config` returns. Which is it, and would it be worth making both sites agree — e.g. a small `isOff(v)` helper in `librespeed.common` that treats `false`, `0` and `'0'` alike — so a later change to the backend's serialisation cannot silently disarm one of them?

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js
```diff
@@ -0,0 +1,540 @@
...
+		'6h': _('Every 6 hours'),
+		'12h': _('Every 12 hours'),
+		'1d': _('Daily')
+	},
+
+	/* Exported alongside the chart: every timestamp a page prints should
+	 * come through here, or the date-only guard gets lost in a copy. */
+	chartStampFull: chartStampFull,
+
+	seriesStats(entries, metric, resolution) {
```
openwrt-ai: nit: `INTERVALS` and `chartStampFull` were inserted between the `seriesStats` doc comment and `seriesStats` itself, so the comment now reads as documentation for the interval table — and the `_min`/`_max` paragraph it just gained describes code twenty lines further down.

```suggestion
	/* One token->label table for the schedule intervals: Settings builds
	 * its choices from it and the Test page looks the status label up, so
	 * the two cannot drift. The init script accepts more shapes than these
	 * six, so readers must fall back to the raw token. */
	INTERVALS: {
		'15m': _('Every 15 minutes'),
		'30m': _('Every 30 minutes'),
		'1h': _('Hourly'),
		'6h': _('Every 6 hours'),
		'12h': _('Every 12 hours'),
		'1d': _('Daily')
	},

	/* Exported alongside the chart: every timestamp a page prints should
	 * come through here, or the date-only guard gets lost in a copy. */
	chartStampFull: chartStampFull,

	/* Plain figures of one series: what is normal, how much it wobbles, and
	 * where it stands now. The story sentences are built from these. On
	 * daily aggregates the extremes come from the _min/_max columns the
	 * chart also draws -- min/max of the daily means would understate the
	 * spread and contradict the band directly above the sentence. */
	seriesStats(entries, metric, resolution) {
```

This shifts the six `INTERVALS` labels up by five lines, so the template's references into `common.js` need regenerating afterwards. (I re-validated the current template mechanically: all 139 entries' source references resolve to the right line, no msgids are missing, and the only ids without a JS literal are the five that come from the ACL and menu JSON — so it is correct as it stands today.)

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,743 @@
...
+				this.runTop = niceTop(status.mbps * 1.02);
+			this.wasRunning = true;
+
+			/* The spinner is LuCI's own; while the client still negotiates
+			 * with the server there is nothing to point the gauge at. */
+			if (phaseIdx < 0)
+				rows.push(E('p', { 'class': 'spinning', 'style': 'margin-top:.75em' },
+					[ _('Connecting to the test server…') ]));
+			else {
+				const holder = E('div', { 'aria-hidden': 'true' });
```
openwrt-ai, line 432: nit: the reason given for the live region above ("announcing the rebuilt gauge SVG would read the whole dial out several times a run") does not apply to `aria-hidden` — `statusNode` is not a live region, so nothing in it was ever announced automatically; only the new `liveNode` is. What `aria-hidden` changes is browse mode, and it removes the running rate along with the tick labels: the `%.2f` readout at line 178-179 is the only place the current Mbps figure appears while a test runs, and `liveNode` carries just the phase name (`Ping`/`Download`/`Upload`). The overall percent and the elapsed time below stay readable, the speed no longer is.

Hiding the ten tick labels is clearly right; if the intent is to keep the whole SVG out of the way, folding the rate into the announced string (or leaving the readout outside the `aria-hidden` holder) would keep the figure reachable. Was the readout meant to go with the dial?

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit against the previous head (`fbeecb7`). Both earlier follow-ups are addressed:

- the numeric sort keys now carry the same fixed decimal count as their display half, so `deriveSortKey`'s `String(value)` fallback produces equal-width digit runs and `L.naturalCompare` orders `94.35` before `94.40` again; the display halves also became `<span>` nodes, which closes the `dom.append` → `innerHTML` path for the server name;
- the stray blank line before `redraw()`'s closing brace is gone.

New work in this push: the schedule-interval table shared between Settings and the Test page, `resolution`-aware `seriesStats()` extremes, the `--librespeed-accent` indirection, the `stopped`/`elapsed` status handling, `N_()` plurals, the ARIA pass (live region, `aria-pressed`, `aria-current`), the filter reconciliation in `renderControls()`, the history-off empty state, the widened ACL, and moving the schedule preview out of the `Map` root — that last one is correct, [`renderContents()` does `dom.content(mapEl, null)`](https://github.com/openwrt/luci/blob/44c91d30083fb9c33264e4dbc04e2f3383f76ac3/modules/luci-base/htdocs/luci-static/resources/form.js#L678-L682) on every `save()`/`reset()`, so the old `node.appendChild(box)` really did lose the preview.

I re-validated the template mechanically against the current sources: all 139 entries' source references resolve to the right line, no msgids are missing, and no orphans beyond the five that come from the ACL and menu JSON. CI is green on `44c91d3`.

Five follow-ups posted inline: one missed `resolution` argument, one dark-mode media query that does not match how LuCI signals dark, one config-encoding question, and two nits.

---

applications/luci-app-librespeed/po/templates/librespeed.pot
```diff
@@ -0,0 +1,622 @@
...
+msgid ""
+"Encrypted transfers, which on hardware without AES acceleration may bound "
+"the result."
+msgstr ""
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js:163
+msgid "Every 12 hours"
+msgstr ""
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js:159
```
openwrt-ai: The `common.js` references were shifted by hand rather than regenerated, so all eight of them are now off by exactly three lines. They moved by −5 — the distance the `seriesStats` doc comment travelled when it was relocated below `chartStampFull` — but this push also removed the four-line `CSS_VERSION` block above `metricOf()` and grew the `cssLink()` comment from two lines to three. Net −8, template says −5.

| msgid | template says | actual |
| --- | --- | --- |
| `Every 15 minutes` | common.js:159 | common.js:156 |
| `Every 30 minutes` | common.js:160 | common.js:157 |
| `Hourly` | common.js:161 | common.js:158 |
| `Every 6 hours` | common.js:162 | common.js:159 |
| `Every 12 hours` | common.js:163 | common.js:160 |
| `Daily` | common.js:164 | common.js:161 |
| `avg` | common.js:290 | common.js:287 |
| `avg` | common.js:361 | common.js:358 |

`common.js:161` and `:164` are the amusing ones — they now point at `_('Hourly')`'s and `_('Daily')`'s *neighbours*, so a translator following the reference lands on the wrong interval.

Everything else in the template is fine: I re-validated it mechanically against the current sources — the other 131 entries' references all resolve to the right line, no msgids are missing, and there are no orphans beyond the five that come from the ACL and menu JSON. Please regenerate it from the current sources so these eight point at the right lines again.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
```diff
@@ -0,0 +1,752 @@
...
+				holder.innerHTML = gauge(
+					(typeof status.mbps == 'number' && status.mbps > 0)
+						? status.mbps : null,
+					this.runTop,
+					status.phase == 'upload' ? 1 : 0);
+				rows.push(holder);
+				if (typeof status.mbps == 'number' && status.mbps > 0)
+					rows.push(E('span', { 'class': 'librespeed-visually-hidden' },
+						[ '%.2f Mbps'.format(status.mbps) ]));
```
openwrt-ai, line 441: nit: `.librespeed-visually-hidden` has a second user now, but its definition still claims to have only one — [`/* Off-screen but spoken: the live region announcing the running phase. */` at librespeed.css:335](``https://github.com/openwrt/luci/blob/84decc27e5c88eb031a8f516e073823d1967d225/applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/librespeed.css#L335-L336).`` This span is deliberately *not* in a live region (`statusNode` is a plain sibling of `liveNode`, both appended at line 714-715), so a reader who takes the CSS comment at its word would conclude the rate is announced on every poll, which is the opposite of what the comment four lines up here says.

Something like "Off-screen but reachable: the live region, and the gauge figures repeated for browse mode" would cover both callers.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit against the previous head (`44c91d3`). All five earlier follow-ups are addressed:

- `renderRecent()` now passes `data && data.resolution` into `seriesStats()`, so the `'1d'` branch reaches the Test page's stability sentence and its `best`/`worst` figures agree with the min/max band the chart above them draws;
- the dark-mode override is keyed on `:root[data-darkmode="true"]` instead of `prefers-color-scheme`, which is what [bootstrap's `header.ut` actually stamps](``https://github.com/openwrt/luci/blob/84decc27e5c88eb031a8f516e073823d1967d225/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut#L12-L31``) — the light-only themes that reach the fallback no longer darken on an OS preference, and `bootstrap-dark` on a light OS now gets the dark accent;
- both readers of `config` are strict about `enabled` and say so in the source, so the encoding question is settled from the author's side;
- the `seriesStats` doc comment sits above `seriesStats` again rather than above `INTERVALS`;
- the gauge's rate is repeated as visually-hidden text, so hiding the dial no longer takes the running Mbps figure with it.

Also new in this push: `CSS_VERSION` is gone and the stylesheet is linked plainly. That checks out — [uhttpd emits `ETag` and `Last-Modified` on every static response](https://github.com/openwrt/uhttpd/blob/HEAD/file.c#L351-L362) and honours `If-None-Match`, and plain `L.resource()` is what luci-app-nlbwmon and luci-app-banip do.

CI is green on `84decc2`. Two follow-ups posted inline, one on the template, one nit.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit against the previous head (`84decc2`). Both earlier follow-ups are addressed, and nothing else changed in this push:

- the template is regenerated — the six `common.js` `INTERVALS` references shifted from 159–164 to 156–161 and the two `_('avg')` references from 290/361 to 287/358, which is where those literals now sit. I re-validated the whole `.pot` mechanically against the current sources: all 178 source references resolve to the right line, none of the 175 `_()`/`N_()` literals in the JS is missing a msgid, and the only three entries without a JS literal are `Test`, `Settings` and `Grant access to LibreSpeed measurements`, which come from the menu and ACL JSON;
- the `.librespeed-visually-hidden` comment now covers both users of the class — the `role="status"` live region and the gauge rate repeated as plain text — rather than just the live region.

CI is green on `8522919`, and the commit subject still matches the diff. Reviewed 1 new commit; no new issues found.

---

1715173329:
> Hey @1715173329, since you're maintaining the librespeed-go backend in the openwrt/packages repository, what do you think about adding the backend configuration here?
> 
> I'd like to do that in an upcoming version once this is merged, to avoid making more changes right now. For the time being, I'd like to keep this PR focused on the client.

This LuCI (as a client) looks really lovely to me, and it's already quite complete. Adding server/backend configuration would make it feel a bit bloated/overdesigned to me. Of course, that's up to you.

BKPepe:
@systemcrash , @efahl guys, can you take a look?

efahl:
I certainly can't do a better job of picking at the code than the clankers have already done, it looks good to me!  Very polished.

systemcrash (approved):

BKPepe:
Can this be merged? 🙏

applications/luci-app-librespeed/po/templates/librespeed.pot
```diff
@@ -0,0 +1,622 @@
...
+msgid ""
+"Encrypted transfers, which on hardware without AES acceleration may bound "
+"the result."
+msgstr ""
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js:160
+msgid "Every 12 hours"
+msgstr ""
+
+#: applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/common.js:156
```
openwrt-ai: The template was not regenerated after `extent()` was inserted at common.js:64, so the eight references below that point are all off by 13 lines:

| msgid | template says | actual |
| --- | --- | --- |
| `Every 15 minutes` | common.js:156 | common.js:169 |
| `Every 30 minutes` | common.js:157 | common.js:170 |
| `Hourly` | common.js:158 | common.js:171 |
| `Every 6 hours` | common.js:159 | common.js:172 |
| `Every 12 hours` | common.js:160 | common.js:173 |
| `Daily` | common.js:161 | common.js:174 |
| `avg` | common.js:287 | common.js:300 |
| `avg` | common.js:358 | common.js:371 |

The `METRICS`/`RANGES` references (common.js:11–14, 27–31) sit above the insertion point and are still correct, as is everything pointing into the three view files; no msgids are missing and there are no orphans beyond the five from the ACL and menu JSON. Please regenerate the template so these eight point at the right lines again.

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

applications/luci-app-librespeed/htdocs/luci-static/resources/librespeed/librespeed.css
openwrt-ai: fixed, thanks

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
openwrt-ai: fixed, thanks

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/test.js
openwrt-ai: solved by repeating the gauge rate as visually-hidden text rather than moving it out of the `aria-hidden` holder, thanks

---

applications/luci-app-librespeed/htdocs/luci-static/resources/view/librespeed/history.js
openwrt-ai: answered — both readers are now strict (`=== false` here, `=== true` at test.js:675/677) and the comment states the backend emits a real boolean, thanks

---

BKPepe:
@feckert Can you merge this one? :) Is it possible to backport this to OpenWrt 25.12 somehow?

openwrt-ai (commented):

---

applications/luci-app-librespeed/po/templates/librespeed.pot
openwrt-ai: fixed, thanks

---

BKPepe:
@dibdot :pray: Can you merge this one for me? :)

dibdot:
I would validate the used absolute paths for history / retention, but that's more a backend issue. Let's merge this ... ;-)

BKPepe:
Thanks @dibdot! :) Is there any chance to backport this to OpenWrt 25.12? I am thinking to create PR. Would be great to get this to users as soon as possible. At least for Turris users in their next Turris OS release.

dibdot:
No problem to cherrypick this luci-app for you - is librespeed-common already in 25.12. package repo?

BKPepe:
Sounds great! Woah. :))

No, no. Librespeed-common is not there yet.

dibdot:
Just ping me whenever the base package lands in the package 25.12. branch ... ;-)

Maxpower905:
Thanks for adding luci-app-librespeed — it is a very useful addition to LuCI!

Could you consider adding an advanced option in the LuCI settings to disable TLS certificate verification for the LibreSpeed test?

I encountered a LibreSpeed server whose certificate is not trusted by the OpenWrt CA store. The `librespeed-cli` test works correctly with the existing `--skip-cert-verify` option, but the LuCI Router speed test currently fails with:

`tls: failed to verify certificate: x509: certificate signed by unknown authority`

I confirmed that adding:

`set -- "$@" --skip-cert-verify`

to `/usr/libexec/librespeed-run` makes the LuCI test work correctly.

Since `librespeed-cli` already supports this option, perhaps it could be exposed as **“Skip TLS certificate verification”** in the Settings page, disabled by default and clearly marked as insecure.

This would avoid users having to manually modify `/usr/libexec/librespeed-run`, and the setting could be passed to the runner via UCI.

PR #8953, opened 2026-08-18
luci-app-package-manager: make the per-row action button compact

BKPepe:
Maintainer: @jow-

# Description:

The package table reserves 15% of its width for the action cell, causing the single action button to stretch unnecessarily.

Let the cell size to its content instead, giving the description more space while keeping the action buttons aligned. The change is scoped to the package table and does not affect the narrow-screen layout.
Screenshot or video of changes: 

# Before:
01-desktop-before

# After:
02-desktop-after
03-desktop-after-mixed-buttons

Tested on:
OpenWrt version: OpenWrt SNAPSHOT r0+35801-c0ffca4c63
LuCI version: luci-base 26.228.64482~c26820a
Web browser(s): Chrome 151.0.7922.138

### DIFF
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -51,6 +51,11 @@ const css = '								\
 		white-space: nowrap;			\
 	}									\
 										\
+	/* bootstrap sets 15%; shrink */	\
+	#packages .td.cbi-section-actions {	\
+		width: auto;					\
+	}									\
+										\
 	ul.deps, ul.deps ul, ul.errors {	\
 		margin-left: 1em;				\
 	}									\


applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -51,6 +51,13 @@ const css = '								\
 		white-space: nowrap;			\
 	}									\
 										\
+	/* Theme sets 15%; shrink to fit */	\
+	#packages .td.cbi-section-actions {	\
+		width: auto;					\
+		padding-left: 1em;				\
+		padding-right: 1em;				\
```
openwrt-ai: These two longhands do reach the narrow-screen layout, contrary to what the commit message and PR description state.

luci-theme-bootstrap's mobile stylesheet sets the horizontal padding for this cell in [`mobile.css:138-143`](``https://github.com/openwrt/luci/blob/579fe6c7ecd24b3bc89cc3a692f8500f8d45bc14/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/mobile.css#L138-L143):``

```css
	.td.cbi-section-actions {
		overflow: initial;
		max-width: 100%;
		padding: 3px 2px;
		flex-basis: 100%;
	}
```

That selector is `(0,2,0)`; the new rule is `(1,2,0)` because of `#packages`, and a `@media` block adds no specificity — so `padding-left/right: 1em` wins inside the media query too. On phones the action cell is stacked at `flex-basis: 100%`, so instead of the theme's 2px it now gets ~2em of horizontal padding eaten out of the full row width (`box-sizing: border-box` from [`mobile.css:40-52`](``https://github.com/openwrt/luci/blob/579fe6c7ecd24b3bc89cc3a692f8500f8d45bc14/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/mobile.css#L40-L52``) keeps it from overflowing, but the button is visibly inset).

Note also that on desktop `.table .td` already carries `padding: 10px 10px 9px` ([`cascade.css:981-987`](``https://github.com/openwrt/luci/blob/579fe6c7ecd24b3bc89cc3a692f8500f8d45bc14/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L981-L987)),`` so `width: auto` alone already yields a compact, padded cell. Either drop the two padding longhands, or move them into a min-width media query so the stacked layout keeps the theme's values — whichever matches the intended look in your screenshots.

---

applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
```diff
@@ -51,6 +51,13 @@ const css = '								\
 		white-space: nowrap;			\
 	}									\
 										\
+	/* Theme sets 15%; shrink to fit */	\
```
openwrt-ai: nit: "Theme" is only true for one of the four in-tree themes. `width: 15%` on `.td.cbi-section-actions` is set exclusively by luci-theme-bootstrap ([`cascade.css:2086-2090`](``https://github.com/openwrt/luci/blob/579fe6c7ecd24b3bc89cc3a692f8500f8d45bc14/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L2086-L2090)).`` The other three set no width on that cell — [material `cascade.css:1355-1358`](``https://github.com/openwrt/luci/blob/579fe6c7ecd24b3bc89cc3a692f8500f8d45bc14/themes/luci-theme-material/htdocs/luci-static/material/cascade.css#L1355-L1358),`` [openwrt `cascade.css:919-923`](``https://github.com/openwrt/luci/blob/579fe6c7ecd24b3bc89cc3a692f8500f8d45bc14/themes/luci-theme-openwrt/htdocs/luci-static/openwrt.org/cascade.css#L919-L923),`` [openwrt-2020 `cascade.css:1552-1554`](``https://github.com/openwrt/luci/blob/579fe6c7ecd24b3bc89cc3a692f8500f8d45bc14/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css#L1552-L1554)`` — so there `width: auto` is a no-op and the rule's only net effect is the extra 1em horizontal padding. Worth naming the theme so the next reader doesn't go hunting, and worth confirming the padding change in the other three themes is intended.

```suggestion
	/* bootstrap sets 15%; shrink */	\
```

---

openwrt-ai (commented):
Reviewed 1 new commit; the commit message matches the diff. Two inline notes on how the new rule interacts with the themes' own `.td.cbi-section-actions` styling.

---

openwrt-ai (commented):

Both points from the previous review are addressed: the `padding-left`/`padding-right` longhands that leaked into the `mobile.css` stacked layout are gone, and the comment now names the theme. What remains is `width: auto`, which overrides luci-theme-bootstrap's `.td.cbi-section-actions { width: 15% }` on desktop and is inert on narrow screens (the cell is a flex item with `flex-basis: 100%` there) and in the other three themes (they set no width on that cell) — so the commit message and PR description now match the diff's actual effect.

---

Self-Hosting-Group:
I was wondering about the different wording in the package manager: the `Updates` tab and the `Upgrade…` action.

PR #8958, opened 2026-08-20
luci-app-51ddns: add remote access agent interface

21hkcloud:
## Description

Add a LuCI interface for the open-source 51DDNS OpenWrt agent.

The page is available under **Services > 51DDNS Remote Access** and provides:

- account-token configuration with password masking;
- service status and installed agent version;
- current plan name, expiry and remaining time;
- a direct link to the 51DDNS console.

The application depends on `51ddns-agent`, which is submitted separately in openwrt/packages#30158. The LuCI application is intentionally not bundled in the agent source archive.

## Documentation and trust model

English installation, operation, privacy, troubleshooting and removal documentation:
https://www.51ddns.com/openwrt

Agent source and client privacy notice:
https://github.com/21hkcloud/51ddns-openwrt

The UI explicitly identifies 51DDNS as a vendor-hosted remote-access service. The agent uses the feed-provided `frpc` package and runs as a dedicated unprivileged user in a procd/ujail sandbox.

## Testing

- The page and RPC helper use translatable English strings.
- ACL access is limited to the 51DDNS UCI configuration, service status and the dedicated read-only RPC helper.
- The app has been tested with the source-built agent on OpenWrt 24.10 and 25.12 snapshots.

### DIFF
--- a/applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js
+++ b/applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js
@@ -1,0 +1,141 @@
+'use strict';
+'require view';
+'require form';
+'require rpc';
+'require uci';
+
+const callServiceList = rpc.declare({
+	object: 'service',
+	method: 'list',
+	params: [ 'name' ],
+	expect: { '': {} },
+});
+
+const callLocalStatus = rpc.declare({
+	object: 'luci.51ddns',
+	method: 'status',
+	expect: { '': {} },
+});
+
+const callAgentInfo = rpc.declare({
+	object: 'luci.51ddns',
+	method: 'info',
+	expect: { '': {} },
+});
+
+function serviceState(data) {
+	const instances = data?.['51ddns-agent']?.instances || {};
+	const running = Object.values(instances).some(instance => instance?.running);
+
+	return {
+		running,
+		label: running ? _('Running') : _('Not running'),
+		className: running ? 'alert-message success' : 'alert-message warning',
+	};
+}
+
+function formatDate(value) {
+	const date = new Date(value || '');
+
+	if (Number.isNaN(date.getTime()))
+		return _('Synchronizing');
+
+	return date.toLocaleString([], {
+		year: 'numeric',
+		month: '2-digit',
+		day: '2-digit',
+		hour: '2-digit',
+		minute: '2-digit',
+	});
+}
+
+function remainingState(value) {
+	const expires = new Date(value || '');
+
+	if (Number.isNaN(expires.getTime()))
+		return { label: _('Synchronizing'), className: 'alert-message notice' };
+
+	const milliseconds = expires.getTime() - Date.now();
+	if (milliseconds <= 0)
+		return { label: _('Expired'), className: 'alert-message error' };
+
+	const hours = Math.ceil(milliseconds / 3600000);
+	const days = Math.ceil(hours / 24);
+	const label = hours > 48
+		? N_(days, 'About %d day', 'About %d days').format(days)
+		: N_(hours, '%d hour', '%d hours').format(hours);
+
+	return {
+		label,
+		className: milliseconds <= 3 * 86400000
+			? 'alert-message warning'
+			: 'alert-message success',
+	};
+}
+
+return view.extend({
+	load() {
+		return Promise.all([
+			uci.load('51ddns'),
+			L.resolveDefault(callServiceList('51ddns-agent'), {}),
+			L.resolveDefault(callLocalStatus(), {}),
+			L.resolveDefault(callAgentInfo(), {}),
+		]);
+	},
+
+	render(data) {
+		const state = serviceState(data[1]);
+		const local = data[2] || {};
+		const info = data[3] || {};
+		const plan = local.plan || null;
+		const remaining = remainingState(plan?.expires_at);
+		const map = new form.Map(
+			'51ddns',
+			_('51DDNS Remote Access'),
+			_('Enter the account token once to register this router automatically. Device identity and relay settings are managed by the 51DDNS control plane.'),
+		);
+		const section = map.section(form.NamedSection, 'main', 'agent', _('Quick setup'));
+		section.addremove = false;
+
+		const status = section.option(form.DummyValue, '_status', _('Service status'));
+		status.rawhtml = true;
+		status.cfgvalue = () =>
+			`<span class="${state.className}"><strong>${state.label}</strong></span>`;
+
+		const version = section.option(form.DummyValue, '_version', _('Agent version'));
+		version.cfgvalue = () => info.version || _('Unavailable');
+
+		const planName = section.option(form.DummyValue, '_plan_name', _('Current plan'));
+		planName.cfgvalue = () => plan?.product_name || _('No active plan');
+
+		const expiresAt = section.option(form.DummyValue, '_expires_at', _('Expires at'));
+		expiresAt.cfgvalue = () => plan ? formatDate(plan.expires_at) : _('Not bound');
+
+		const remainingTime = section.option(form.DummyValue, '_remaining_time', _('Time remaining'));
+		remainingTime.rawhtml = true;
+		remainingTime.cfgvalue = () => {
+			if (!plan)
+				return `<span class="alert-message warning"><strong>${_('Bind or purchase a plan')}</strong></span>`;
+
+			return `<span class="${remaining.className}"><strong>${remaining.label}</strong></span>`;
+		};
+
+		const consoleButton = section.option(form.Button, '_console', _('Console'));
+		consoleButton.inputtitle = _('Open 51DDNS Console');
+		consoleButton.inputstyle = 'apply';
+		consoleButton.onclick = () =>
+			window.open('https://console.51ddns.com/console#/workbench', '_blank', 'noopener,noreferrer');
+
+		const enabled = section.option(form.Flag, 'enabled', _('Enable'));
+		enabled.rmempty = false;
+		enabled.default = enabled.disabled;
+
+		const token = section.option(form.Value, 'account_token', _('Account token'));
+		token.password = true;
+		token.rmempty = false;
+		token.placeholder = '51d_...';
+		token.description = _('Copy the token from the 51DDNS console. All devices in the same account share this token.');
+
+		return map.render();
+	},
+});
--- a/applications/luci-app-51ddns/po/templates/51ddns.pot
+++ b/applications/luci-app-51ddns/po/templates/51ddns.pot
@@ -1,0 +1,108 @@
+msgid ""
+msgstr "Content-Type: text/plain; charset=UTF-8"
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:66
+msgid "%d hour"
+msgid_plural "%d hours"
+msgstr[0] ""
+msgstr[1] ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:94
+#: applications/luci-app-51ddns/root/usr/share/luci/menu.d/luci-app-51ddns.json:3
+msgid "51DDNS Remote Access"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:65
+msgid "About %d day"
+msgid_plural "About %d days"
+msgstr[0] ""
+msgstr[1] ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:133
+msgid "Account token"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:105
+msgid "Agent version"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:118
+msgid "Bind or purchase a plan"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:123
+msgid "Console"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:137
+msgid ""
+"Copy the token from the 51DDNS console. All devices in the same account "
+"share this token."
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:108
+msgid "Current plan"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:129
+msgid "Enable"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:95
+msgid ""
+"Enter the account token once to register this router automatically. Device "
+"identity and relay settings are managed by the 51DDNS control plane."
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:60
+msgid "Expired"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:111
+msgid "Expires at"
+msgstr ""
+
+#: applications/luci-app-51ddns/root/usr/share/rpcd/acl.d/luci-app-51ddns.json:3
+msgid "Grant access to the 51DDNS configuration"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:109
+msgid "No active plan"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:112
+msgid "Not bound"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:32
+msgid "Not running"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:124
+msgid "Open 51DDNS Console"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:97
+msgid "Quick setup"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:32
+msgid "Running"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:100
+msgid "Service status"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:41
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:56
+msgid "Synchronizing"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:114
+msgid "Time remaining"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:106
+msgid "Unavailable"
+msgstr ""
--- a/applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
+++ b/applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
@@ -1,0 +1,59 @@
+#!/bin/sh
+
+. /usr/share/libubox/jshn.sh
+
+print_status() {
+	status_has_plan=0
+	status_product_name=''
+	status_expires_at=''
+	status_payload=''
+
+	if [ -s /var/lib/51ddns/status.json ]; then
+		status_payload="$(cat /var/lib/51ddns/status.json 2>/dev/null)"
+		json_load "$status_payload" 2>/dev/null
+
+		if json_select plan >/dev/null 2>&1; then
+			status_has_plan=1
+			json_get_var status_product_name product_name
+			json_get_var status_expires_at expires_at
+		fi
+		json_cleanup
+	fi
+
+	json_init
+	if [ "$status_has_plan" = 1 ]; then
+		json_add_object plan
+		json_add_string product_name "$status_product_name"
+		json_add_string expires_at "$status_expires_at"
+		json_close_object
+	fi
+	json_dump
+	json_cleanup
+}
+
+case "$1" in
+	list)
+		echo '{"status":{},"info":{}}'
+		;;
+	call)
+		case "$2" in
+			status)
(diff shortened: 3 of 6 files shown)


applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
```diff
@@ -0,0 +1,33 @@
...
+				if [ -s /var/lib/51ddns/status.json ]; then
+					cat /var/lib/51ddns/status.json
+				else
+					echo '{"online":false}'
+				fi
+				;;
+			info)
+				version="$(/usr/bin/51ddns-agent --version 2>/dev/null | head -n 1)"
+				printf '{"version":"%s"}\n' "$(json_escape "$version")"
+				;;
```
openwrt-ai, line 49: The hand-rolled `json_escape` only escapes `\` and `"`. Any control character in the agent's `--version` output — a tab, a CR from a `\r\n` line ending, a backspace — is emitted raw into the string and produces invalid JSON, which makes the whole `info` ubus call fail rather than just showing a mangled version string. `head -n 1` strips the LF but not a trailing CR.

Other rpcd helpers in this repo use jshn for exactly this reason, e.g. [`luci.squid`](https://github.com/openwrt/luci/blob/746fff9fde693b4b2678edd7ee18da18de911bb2/applications/luci-app-squid/root/usr/libexec/rpcd/luci.squid#L5) and [`olsrvizinfo`](https://github.com/openwrt/luci/blob/746fff9fde693b4b2678edd7ee18da18de911bb2/applications/luci-app-olsr-viz/root/usr/libexec/rpcd/olsrvizinfo#L2).

```suggestion
			info)
				version="$(/usr/bin/51ddns-agent --version 2>/dev/null | head -n 1)"
				json_init
				json_add_string version "$version"
				json_dump
				;;
```

This also needs `. /usr/share/libubox/jshn.sh` near the top of the file, and the `json_escape` helper on lines 3-5 can then be dropped.

---

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
```diff
@@ -0,0 +1,33 @@
...
+		;;
+	call)
+		case "$2" in
+			status)
+				if [ -s /var/lib/51ddns/status.json ]; then
+					cat /var/lib/51ddns/status.json
+				else
+					echo '{"online":false}'
+				fi
+				;;
```
openwrt-ai, line 42: `status` relays the whole of `/var/lib/51ddns/status.json` verbatim as the ubus reply, but the view only consumes `plan.product_name` and `plan.expires_at`. Two consequences worth confirming:

1. Everything the agent happens to write into that file becomes readable by any session holding the `luci-app-51ddns` read ACL. Can `status.json` contain the account token, a device secret or a relay credential? If so, this widens the exposure well beyond what the page needs — emitting only the fields the view reads (`online`, `plan.product_name`, `plan.expires_at`) would keep the ACL honest.
2. The file is passed through without any validation, so if the agent rewrites it non-atomically a partial read is returned as the ubus payload and the call fails to parse instead of falling back to `{"online":false}`.

---

applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js
```diff
@@ -0,0 +1,141 @@
...
+		return { label: _('Synchronizing'), className: 'alert-message notice' };
+
+	const milliseconds = expires.getTime() - Date.now();
+	if (milliseconds <= 0)
+		return { label: _('Expired'), className: 'alert-message error' };
+
+	const hours = Math.ceil(milliseconds / 3600000);
+	const label = hours > 48
+		? _('About %d days').format(Math.ceil(hours / 24))
+		: _('%d hours').format(hours);
```
openwrt-ai: nit: these two strings are count-dependent, so translators into languages with more than two plural forms can't render them correctly through `_()`. luci-base exposes [`N_(n, singular, plural)`](https://github.com/openwrt/luci/blob/746fff9fde693b4b2678edd7ee18da18de911bb2/modules/luci-base/htdocs/luci-static/resources/cbi.js#L167) for this (see [nftables.js:401](``https://github.com/openwrt/luci/blob/746fff9fde693b4b2678edd7ee18da18de911bb2/modules/luci-mod-status/htdocs/luci-static/resources/view/status/nftables.js#L401)),`` and it is already whitelisted as a global in `eslint.config.mjs`.

```suggestion
	const hours = Math.ceil(milliseconds / 3600000);
	const days = Math.ceil(hours / 24);
	const label = hours > 48
		? N_(days, 'About %d day', 'About %d days').format(days)
		: N_(hours, '%d hour', '%d hours').format(hours);
```

The `.pot` template needs regenerating afterwards so the entries pick up their `msgid_plural` form.

---

applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js
```diff
@@ -0,0 +1,141 @@
...
+		const plan = local.plan || null;
+		const remaining = remainingState(plan?.expires_at);
+		const map = new form.Map(
+			'51ddns',
+			_('51DDNS Remote Access'),
+			_('Enter the account token once to register this router automatically. Device identity and relay settings are managed by the 51DDNS control plane.'),
+		);
+		const section = map.section(form.NamedSection, 'main', 'agent', _('Quick setup'));
+		section.anonymous = true;
+		section.addremove = false;
```
openwrt-ai, line 98: nit: `anonymous` is only consulted by `TypedSection`/`GridSection`; [`NamedSection`](https://github.com/openwrt/luci/blob/746fff9fde693b4b2678edd7ee18da18de911bb2/modules/luci-base/htdocs/luci-static/resources/form.js#L4159-L4165) never reads it, so line 97 has no effect.

```suggestion
		const section = map.section(form.NamedSection, 'main', 'agent', _('Quick setup'));
		section.addremove = false;
```

Separately: with `addremove = false` this page can only edit a pre-existing `config agent 'main'` section — if `/etc/config/51ddns` is missing or has no such section the form renders empty and the user has no way to enter the token. This package ships no `root/etc/config/51ddns`, so I assume `51ddns-agent` provides it with that exact section name; can you confirm that's the case in openwrt/packages#30158?

---

applications/luci-app-51ddns/po/templates/51ddns.pot
```diff
@@ -0,0 +1,100 @@
...
+msgid "Account token"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:105
+msgid "Agent version"
+msgstr ""
+
+#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:64
+msgid "About %d days"
+msgstr ""
```
openwrt-ai: nit: `"About %d days"` sorts before `"Account token"`, so this template can't have come out of the in-tree scanner — it pipes everything through [`msguniq -s`](https://github.com/openwrt/luci/blob/746fff9fde693b4b2678edd7ee18da18de911bb2/build/i18n-scan.pl#L236), which emits msgids in strict byte order. Please regenerate the template with the repo's i18n scanner rather than hand-maintaining it, otherwise the next automated refresh will churn this file.

```suggestion
#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:64
msgid "About %d days"
msgstr ""

#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:133
msgid "Account token"
msgstr ""

#: applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js:105
msgid "Agent version"
msgstr ""
```

(If you take the `N_()` change suggested on `overview.js`, ignore this hunk and just regenerate — the entries change shape anyway.)

---

openwrt-ai (commented):
Reviewed 1 new commit; see inline comments. The commit message matches its changes, and the backend dependency on `51ddns-agent` (openwrt/packages#30158) is already called out in the PR description, so I did not flag it as a missing consumer.

---

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
21hkcloud: Fixed in c46c3c1. The RPC helper now uses jshn to serialize the version response, and the hand-written JSON escaper has been removed.

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
21hkcloud: Fixed in c46c3c1. The RPC helper validates the status JSON and rebuilds a strict whitelist containing only online, plan.product_name, and plan.expires_at. Malformed input falls back to online=false. The agent writes this file atomically through a temporary file and rename.

applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js
21hkcloud: Fixed in c46c3c1. Both hour and day labels now use N_() plural forms, and the translation template has been regenerated.

applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js
21hkcloud: Fixed in c46c3c1. The ineffective anonymous assignment was removed. I also confirmed that openwrt/packages#30158 installs files/51ddns.config as /etc/config/51ddns and that it defines config agent main.

applications/luci-app-51ddns/po/templates/51ddns.pot
21hkcloud: Fixed in c46c3c1. The POT was regenerated with ./build/i18n-scan.pl applications/luci-app-51ddns, so it is scanner-sorted and includes the new plural entries.

21hkcloud:
Addressed all five review findings in c46c3c1: jshn-based RPC serialization, validated and whitelisted status output, plural-aware expiry labels, the NamedSection cleanup with backend config confirmation, and a scanner-generated POT. Local ESLint, shell syntax, ShellCheck, git diff checks, and RPC whitelist/fallback tests pass. All three OpenWrt Formality checks are green on the current head.

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
```diff
@@ -0,0 +1,65 @@
...
+	fi
+
+	json_init
+	json_add_boolean online "$status_online"
+	if [ -n "$status_product_name" ] || [ -n "$status_expires_at" ]; then
+		json_add_object plan
+		json_add_string product_name "$status_product_name"
+		json_add_string expires_at "$status_expires_at"
+		json_close_object
+	fi
```
openwrt-ai: The `plan` object is now emitted based on whether its fields are non-empty rather than on whether the agent actually reported a plan, which changes the page's behaviour compared to the previous verbatim relay. If `status.json` contains `"plan": { "product_name": "", "expires_at": "" }` — a plan that exists but hasn't synced yet — `plan` is dropped from the reply, so [`local.plan`](``https://github.com/openwrt/luci/blob/c46c3c1b09b1066176da495b91710c19e4a46143/applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js#L90)`` becomes `null` and the view renders "Not bound" and "Bind or purchase a plan" instead of "Synchronizing". Before this commit the object was passed through and the view showed the syncing state correctly.

Key the object on the `json_select` result instead:

```suggestion
			if json_select plan 2>/dev/null; then
				status_has_plan=1
				json_get_var status_product_name product_name
				json_get_var status_expires_at expires_at
			fi
		fi
		json_cleanup
	fi

	json_init
	json_add_boolean online "$status_online"
	if [ "$status_has_plan" = 1 ]; then
		json_add_object plan
		json_add_string product_name "$status_product_name"
		json_add_string expires_at "$status_expires_at"
		json_close_object
	fi
```

Add `status_has_plan=0` next to the other initialisations on line 6 so it matches the style of the surrounding resets.

---

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
```diff
@@ -0,0 +1,65 @@
...
+
+print_status() {
+	status_online=0
+	status_product_name=''
+	status_expires_at=''
+	status_payload=''
+
+	if [ -s /var/lib/51ddns/status.json ]; then
+		status_payload="$(cat /var/lib/51ddns/status.json 2>/dev/null)"
+		if json_load "$status_payload" 2>/dev/null; then
```
openwrt-ai: Does this `if` actually catch malformed JSON? `json_load` is a shell wrapper that `eval`s the output of `jshn -r`, and on a parse failure `jshn` writes its diagnostic to stderr and prints nothing, so the `eval` of an empty string still exits 0 and the true branch is taken regardless. That would explain why no other rpcd helper in this repo tests its return value — [`luci.pbr:243`](https://github.com/openwrt/luci/blob/c46c3c1b09b1066176da495b91710c19e4a46143/applications/luci-app-pbr/root/usr/libexec/rpcd/luci.pbr#L243) and [`luci.wificalling-gateway:12`](``https://github.com/openwrt/luci/blob/c46c3c1b09b1066176da495b91710c19e4a46143/applications/luci-app-wificalling-gateway/root/usr/libexec/rpcd/luci.wificalling-gateway#L12)`` both call it bare and validate the extracted values afterwards instead.

The `{"online":false}` fallback you describe in the PR comment still happens either way, because the subsequent `json_get_var`/`json_select` calls find nothing — but if the guard is inert it reads as validation that isn't there. Worth either dropping it in favour of the in-tree pattern, or confirming that `json_load` really does propagate a non-zero status on your target versions.

---

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
+
+. /usr/share/libubox/jshn.sh
```
openwrt-ai, line 3: nit: this line, and everything else in c46c3c1, only rewrites code that 746fff9 introduced in the same unmerged series — the hand-rolled `json_escape` never existed in the tree, so "address review feedback" documents a review round rather than a change to LuCI. For a new-package PR the usual shape is a single `luci-app-51ddns: add remote access agent interface` commit; [CONTRIBUTING.md:40](https://github.com/openwrt/luci/blob/c46c3c1b09b1066176da495b91710c19e4a46143/CONTRIBUTING.md?plain=1#L40) points at `git rebase -i` for exactly this. Please squash before merge so the history doesn't carry an introduce-then-fix pair.

---

openwrt-ai (commented):
Reviewed 1 new commit (c46c3c1); see inline comments.

The jshn conversion, the `N_()` plural labels and the removal of the ineffective `section.anonymous` all look right — `N_(n, singular, plural)` matches [cbi.js:167](https://github.com/openwrt/luci/blob/c46c3c1b09b1066176da495b91710c19e4a46143/modules/luci-base/htdocs/luci-static/resources/cbi.js#L167) and the scanner keyword spec `N_:2,3` in [build/i18n-scan.pl:16](https://github.com/openwrt/luci/blob/c46c3c1b09b1066176da495b91710c19e4a46143/build/i18n-scan.pl#L16). I re-checked the regenerated `51ddns.pot` by hand: msgids are now in strict byte order, the wrapping and `msgid_plural`/`msgstr[0..1]` shape match other in-tree templates, and every `#:` line reference resolves to the right line of `overview.js` at this head. The commit message matches its changes.

One functional regression remains in the new whitelist, plus two smaller points inline.

---

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
21hkcloud: Fixed in 500603debb. The RPC helper now tracks whether `json_select plan` succeeds and emits the `plan` object even when both fields are temporarily empty, preserving the Synchronizing state.

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
21hkcloud: Fixed in 500603debb. The inert conditional around `json_load` was removed; the helper now follows the in-tree pattern and validates the extracted fields before rebuilding the whitelisted response.

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
21hkcloud: Fixed in 500603debb. The introduce-then-fix pair and the latest status fixes were squashed into one `luci-app-51ddns: add remote access agent interface` commit with a meaningful body and Signed-off-by trailer.

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
```diff
@@ -0,0 +1,66 @@
...
+	if [ -s /var/lib/51ddns/status.json ]; then
+		status_payload="$(cat /var/lib/51ddns/status.json 2>/dev/null)"
+		json_load "$status_payload" 2>/dev/null
+		json_get_var status_online online
+		case "$status_online" in
+			1|true) status_online=1 ;;
+			*) status_online=0 ;;
+		esac
+
+		if json_select plan 2>/dev/null; then
```
openwrt-ai: `2>/dev/null` is the wrong stream here: when `json_select` fails it prints its `WARNING: Variable 'plan' does not exist or is not an array/object` diagnostic on **stdout**, not stderr. rpcd's exec plugin parses this script's stdout as the ubus reply, so that warning line is prepended to the `json_dump` output and the whole `status` call stops being valid JSON.

That fires on two paths that are supposed to be normal:

* `status.json` without a `plan` key — the view treats this as expected ([`local.plan || null`](``https://github.com/openwrt/luci/blob/500603debbbc5f2a557457c18b60342a7702d25d/applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js#L90)`` → "No active plan");
* malformed `status.json` — now that the `if json_load` guard is gone, the failed load leaves no `plan` element, so the `{"online":false}` fallback described in the PR comment never reaches the client either; `L.resolveDefault` just swallows the parse error and the page renders as if nothing was reported.

The in-tree idiom is to suppress stdout too, or to set `_json_no_warning=1` — see [`_procd_add_jail_mount` in procd.sh:212](https://github.com/openwrt/openwrt/blob/54f7b240b4e80c4df8ee157742f49879da66511b/package/system/procd/files/procd.sh#L212), [`json_select_object` in uci-defaults.sh:16](https://github.com/openwrt/openwrt/blob/54f7b240b4e80c4df8ee157742f49879da66511b/package/base-files/files/lib/functions/uci-defaults.sh#L16), and script-side users such as [collectd.init:415](https://github.com/openwrt/packages/blob/b785b5c631d4a100fabe0fba1c34146be3275246/utils/collectd/files/collectd.init#L415) and [travelmate-functions.sh:1137](https://github.com/openwrt/packages/blob/b785b5c631d4a100fabe0fba1c34146be3275246/net/travelmate/files/travelmate-functions.sh#L1137), which all redirect stdout as well.

```suggestion
		if json_select plan >/dev/null 2>&1; then
```

---

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
```diff
@@ -0,0 +1,66 @@
...
+			status_has_plan=1
+			json_get_var status_product_name product_name
+			json_get_var status_expires_at expires_at
+		fi
+		json_cleanup
+	fi
+
+	json_init
+	json_add_boolean online "$status_online"
```
openwrt-ai: nit: nothing consumes `online`. The view's only read of the status reply is [`local.plan`](``https://github.com/openwrt/luci/blob/500603debbbc5f2a557457c18b60342a7702d25d/applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js#L90)`` — the "Service status" row comes from `serviceState(data[1])`, i.e. procd's view of whether the process is running, not from the agent's own notion of being connected. So lines 6 and 15-19 are currently dead weight (I mistakenly listed `online` as consumed in the earlier whitelist comment).

Was it meant to feed the status row? A running-but-not-connected agent currently shows plain "Running", which is the case where the agent's own flag would add something. Either wire it in or drop it and the normalisation block with it.

---

openwrt-ai (commented):
Reviewed the squash into 500603d (diff since c46c3c1 is the `plan` tracking in the rpcd helper).

The `status_has_plan` flag does fix the "plan present but not yet synced" regression, and squashing the introduce-then-fix pair addressed the history nit — the single commit's message now matches its contents, so no commit checks.

One remaining problem in the same hunk: `json_select`'s failure diagnostic goes to stdout, so `2>/dev/null` doesn't keep it out of the ubus reply. Details inline, plus one nit about the unused `online` field.

---

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
21hkcloud: Fixed in e780f2c8cf: `json_select plan` now redirects both stdout and stderr (`>/dev/null 2>&1`), so a missing `plan` object cannot leak diagnostics into the ubus JSON response.

applications/luci-app-51ddns/root/usr/libexec/rpcd/luci.51ddns
21hkcloud: Fixed in e780f2c8cf: removed the unused `online` field and its normalization from the RPC response. Service state remains the single source used by the LuCI UI.

21hkcloud:
Addressed the latest review in e780f2c8cf:

- suppress both stdout and stderr from failed `json_select plan` calls, preserving valid ubus JSON output
- remove the unused `online` RPC field and normalization
- keep the PR as a single signed-off commit

Validation: ShellCheck, Bash syntax check, and `git diff --check` all pass.

openwrt-ai (commented):
Reviewed 1 new commit (e780f2c); no new issues found.

The diff since 500603d is confined to the rpcd helper and resolves both of the previous round's findings:

- `json_select plan >/dev/null 2>&1` now suppresses stdout as well, so the `WARNING: Variable 'plan' does not exist...` line can no longer be prepended to the `json_dump` output on the two normal paths (no `plan` key, or a failed `json_load`). `json_get_var` is silent on a missing key, so no other call in `print_status` can leak into the reply.
- `online` and its normalisation block are gone; `git grep online` over `applications/luci-app-51ddns/` at this head returns nothing, so no leftover reader, ACL entry or `.pot` msgid dangles. With no plan the helper now emits a bare `{}`, which [`data[2] || {}` / `local.plan || null`](``https://github.com/openwrt/luci/blob/e780f2c8cf5cf7b604bc99c06fabe0bd2f1a87af/applications/luci-app-51ddns/htdocs/luci-static/resources/view/51ddns/overview.js#L88-L90)`` already handles as "No active plan".

Squash preserved — still a single `luci-app-51ddns: add remote access agent interface` commit, and its message body ("serializing RPC responses with jshn and exposing only validated status fields") matches the file contents, so no commit checks. The helper keeps mode 100755. All three FormalityCheck runs are green on this head.

---

PR #8962, opened 2026-08-21
luci-app-nes-emulator: add NES emulator interface

communism420:
### Description

Add a JavaScript LuCI interface for the NES emulator service.

The application provides service status and controls, authenticated ROM upload
and loading, emulator settings, and access to the thin browser game client.
Emulation, rendering, JPEG encoding, and audio generation remain in the native
`nesd` service on the router. No ROM or BIOS files are included.

A shared application resource centralizes the long-running RPC policy used by
all three views. The package directly depends on its external runtime helpers;
installing it also selects and installs the native `nes-emulator` package.

The RPCD bridge uses BusyBox-compatible file operations. Client-supplied load
paths are passed to `nesd`, whose final validation canonicalizes the path,
confines it to configured ROM roots, opens it with `O_NOFOLLOW`, and validates
the extension, size, and ROM header before loading.

Token creation and rotation share a lock with the native init script. Token,
upload, and startup locks live in the root-owned mode-`0700`
`/var/run/nes-emulator` tmpfs directory. Each mode-`0600`, root-owned,
single-link lock file is safely prepared with no-clobber creation and
`umask 077`, opened without truncation through catchable `command exec`
redirection, and revalidated after opening. Descriptor 9 owns the upload lock,
8 the token lock, and 7 the startup lock. Every external child in the token
critical sections, including the outer command-substitution shells, closes all
three descriptors; startup/upload children close the held request locks too.

Rotation is serialized with startup and rechecks a daemon that may have started
concurrently. Its RPCD-side lock phase has an exact maximum of seven one-second
sleeps: two for each of the two token-lock encounters plus three for the startup
lock. A contended native restart can add at most nine native token-lock sleeps,
for a coordinated lock-wait maximum of 16.

A shared application resource declares the start-capable `access`, `start`,
`load`, and `import` calls, plus `rotate_token`, with `nobatch: true` and an RPC
timeout of at least 120 seconds. Each request therefore captures its own timeout
instead of inheriting the first request in an animation-frame batch. The helper
restores the previous global value immediately after the synchronous RPC
invocation, so unrelated calls retain their configured timeout. The local
Start/Load guard is computed from the effective RPC timeout plus five seconds;
transactional Import/Rotate operations await their real result instead of
racing a false client-side timeout.

Depends on [openwrt/packages#30350](https://github.com/openwrt/packages/pull/30350).
This pull request remains a draft until that native package is available to the
LuCI build, as required by the LuCI pull request template for dependent
submissions.

The English POT template was generated with LuCI's `i18n-scan.pl`; translation
`.po` files remain managed through Weblate.

OpenWrt NES browser client

The screenshot uses the project's freely distributable demo ROM; no commercial
game assets are included.

### Maintainer _(preferred)_

@communism420

---

## Tested on

**LuCI version:** hardware-tested PR head `5028c07a` on the OpenWrt 25.12.5 LuCI stack  
**Current review heads:** LuCI `038ad57ad8`; native `62ec2dff9`  
**Device:** ASUS RT-AX52 Pro (`mediatek/filogic`, `aarch64_cortex-a53`)  
**Web browser(s):** Chromium desktop checks; end-to-end desktop browser smoke test passed

The on-device smoke test used LuCI head `5028c07a` and native head `b4424df4`,
both built with the official OpenWrt 25.12.5 `mediatek/filogic` SDK. LuCI pages,
service status/controls, settings, ROM upload/loading, streamed gameplay, video,
audio, controller input, save/load state, and the FCEUX-style FPS overlay all
worked on the router.

The hardware-tested emulator and FCEUmm sources remain unchanged. Current heads
add the review-driven protected lock lifecycle, external-storage diagnostics,
bounded wait accounting, isolated/unbatched long-operation RPC handling, and
complete fd 7/8/9 child-process hygiene.

The source revision exporting these heads is covered by the successful
[project CI run](https://github.com/communism420/openwrt-nes-emulator/actions/runs/32956033619).
It includes ShellCheck, BusyBox resource-operation contracts, dynamic unbatched
long-RPC timeout/restoration tests, shared-resource packaging/export checks,
lock-fd inheritance checks, browser-client contracts, native path-confinement
integration tests, exact export validation, a clean
native build, and the complete black-box regression suite.

The unchanged package recipes and native binary also passed a clean current
Snapshot x86/64 SDK package check/build. The application passes the current LuCI
ESLint configuration, package check, and compilation. Generated APK metadata
was inspected for direct `cgi-io`, `jshn`, `jsonfilter`, `luci-base`,
`nes-emulator`, and `rpcd` dependencies. The RPCD bridge is mode `0755`.

---

## Checklist

- [x] Depends on openwrt/packages#30350.

### DIFF
--- a/applications/luci-app-nes-emulator/Makefile
+++ b/applications/luci-app-nes-emulator/Makefile
@@ -1,0 +1,16 @@
+# SPDX-License-Identifier: MIT
+
+include $(TOPDIR)/rules.mk
+
+PKG_LICENSE:=MIT
+PKG_MAINTAINER:=Yaroslav Vereshchagin <yarik.vereshchagin1996@gmail.com>
+
+LUCI_TITLE:=LuCI support for NES Emulator
+LUCI_DESCRIPTION:=Web interface for the NES emulator service
+LUCI_DEPENDS:=+luci-base +rpcd +jshn +jsonfilter +cgi-io +nes-emulator
+LUCI_PKGARCH:=all
+LUCI_MAINTAINER:=$(PKG_MAINTAINER)
+
+include ../../luci.mk
+
+# call BuildPackage - OpenWrt buildroot signature
--- a/applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/overview.js
+++ b/applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/overview.js
@@ -1,0 +1,519 @@
+'use strict';
+'require view';
+'require rpc';
+'require ui';
+'require nes-emulator as nesEmulator';
+
+const callCanControl = rpc.declare({
+	object: 'session',
+	method: 'access',
+	params: [ 'scope', 'object', 'function' ],
+	expect: { '': { access: false } }
+});
+
+const callStatus = rpc.declare({
+	object: 'nes-emulator',
+	method: 'status',
+	expect: { '': {} }
+});
+
+const callRoms = rpc.declare({
+	object: 'nes-emulator',
+	method: 'roms',
+	expect: { '': { roms: [] } }
+});
+
+const callLoad = nesEmulator.declareLongRunningRpc({
+	object: 'nes-emulator',
+	method: 'load',
+	params: [ 'path' ],
+	expect: { '': {} }
+});
+
+const callImport = nesEmulator.declareLongRunningRpc({
+	object: 'nes-emulator',
+	method: 'import',
+	params: [ 'staged', 'name' ],
+	expect: { '': {} }
+});
+
+const callReserveUpload = rpc.declare({
+	object: 'nes-emulator',
+	method: 'reserve_upload',
+	expect: { '': {} }
+});
+
+const callDiscardUpload = rpc.declare({
+	object: 'nes-emulator',
+	method: 'discard_upload',
+	params: [ 'staged' ],
+	expect: { '': {} }
+});
+
+const callStart = nesEmulator.declareLongRunningRpc({
+	object: 'nes-emulator',
+	method: 'start',
+	expect: { '': {} }
+});
+
+const callStop = rpc.declare({
+	object: 'nes-emulator',
+	method: 'stop',
+	expect: { '': {} }
+});
+
+const callPause = rpc.declare({
+	object: 'nes-emulator',
+	method: 'pause',
+	expect: { '': {} }
+});
+
+const callEmuReset = rpc.declare({
+	object: 'nes-emulator',
+	method: 'reset',
+	expect: { '': {} }
+});
+
+function withTimeout(promise, ms) {
+	return new Promise((resolve, reject) => {
+		const timer = setTimeout(
+			() => reject(new Error(_('Request timed out'))),
+			ms
+		);
+		Promise.resolve(promise).then(
+			value => {
+				clearTimeout(timer);
+				resolve(value);
+			},
+			error => {
+				clearTimeout(timer);
+				reject(error);
+			}
+		);
+	});
+}
+
+function requireSuccess(reply) {
+	if (!reply)
+		throw new Error(_('Empty response'));
+	if (reply.error || reply.ok === false)
+		throw new Error(reply.error || _('Operation failed'));
+	return reply;
+}
+
+return view.extend({
+	load() {
+		return callCanControl('ubus', 'nes-emulator', 'access')
+			.catch(() => ({ access: false }));
+	},
+
+	render(data) {
+		this.canControl = Boolean(data && data.access);
+		this.actionBusy = false;
+		this.lastStatus = null;
+		// Invalidate requests owned by an older rendering of this view.
+		this.refreshGeneration = (this.refreshGeneration || 0) + 1;
+		const body = E('div', { class: 'cbi-map' }, [
+			E('h2', {}, _('NES Emulator')),
+			E('div', { class: 'cbi-map-descr' },
+				_('FCEUmm, picture rendering and optional JPEG encoding all run on the router CPU. The browser is only a display, audio sink and controller.')),
+			E('div', { class: 'cbi-section' }, [
+				E('h3', {}, _('Status')),
+				E('div', { id: 'nes-status', class: 'cbi-value' },
+					E('em', {}, _('Loading status…'))),
+				E('div', { class: 'cbi-page-actions' }, [
+					E('button', {
+						class: 'btn cbi-button cbi-button-apply',
+						'data-nes-action': '1',
+						'data-nes-kind': 'start',
+						disabled: true,
+						click: ui.createHandlerFn(this, 'onStart')
+					}, _('Start')),
+					' ',
+					E('button', {
+						class: 'btn cbi-button cbi-button-action',
+						'data-nes-action': '1',
+						'data-nes-kind': 'pause',
+						disabled: true,
+						click: ui.createHandlerFn(this, 'onPause')
+					}, _('Pause / Resume')),
+					' ',
+					E('button', {
+						class: 'btn cbi-button cbi-button-action',
+						'data-nes-action': '1',
+						'data-nes-kind': 'reset',
+						disabled: true,
+						click: ui.createHandlerFn(this, 'onEmuReset')
+					}, _('Reset console')),
+					' ',
+					E('button', {
+						class: 'btn cbi-button cbi-button-remove',
+						'data-nes-action': '1',
+						'data-nes-kind': 'stop',
+						disabled: true,
+						click: ui.createHandlerFn(this, 'onStop')
+					}, _('Stop emulation')),
+					' ',
+					E('button', {
+						class: 'btn cbi-button',
+						click: ui.createHandlerFn(this, 'onRefreshAll')
+					}, _('Refresh'))
+				])
+			]),
+			this.canControl ? E('div', { class: 'cbi-section' }, [
+				E('h3', {}, _('Upload ROM from this computer')),
+				E('p', {},
+					_('The authenticated LuCI uploader stages the file, then rpcd validates its size, extension and NES header before an atomic import. Maximum size: 16 MiB.')),
+				E('div', { class: 'cbi-page-actions' }, [
+					E('button', {
+						class: 'btn cbi-button cbi-button-apply',
+						'data-nes-action': '1',
+						'data-nes-kind': 'upload',
+						disabled: !this.canControl,
+						click: ui.createHandlerFn(this, 'onUpload')
+					}, _('Choose, upload & load ROM'))
+				])
+			]) : '',
+			E('div', { class: 'cbi-section' }, [
+				E('h3', {}, _('ROMs on router')),
+				E('div', { id: 'nes-rom-table' },
+					E('em', {}, _('Loading ROM list…')))
+			]),
+			E('div', { class: 'cbi-section' }, [
+				E('h3', {}, _('Play')),
+				this.canControl ? E('p', {}, [
+					E('a', {
+						href: L.url('admin/services/nes-emulator/play')
+					}, _('Open Play tab (control permission required)'))
+				]) : E('p', { class: 'alert-message warning' },
+					_('Read-only access: game controls, ROM loading and uploads are disabled.')),
+				this.canControl && window.location.protocol === 'https:'
+					? E('details', {}, [
+						E('summary', {}, _('Connection help')),
+						E('p', {},
+							_('LuCI is using HTTPS, while the local game client uses HTTP. If the browser blocks the new window or upgrades it to HTTPS, allow HTTP for this router address or use a trusted reverse proxy.'))
+					])
+					: ''
+			])
+		]);
+
+		requestAnimationFrame(() => this.onRefreshAll());
+		return body;
+	},
+
+	statusText(status) {
+		if (!status || status.error)
+			return E('em', {}, status && status.error
+				? status.error
+				: _('Daemon is not reachable'));
+
+		const parts = [
+			status.core || 'FCEUmm',
+			status.rom ? _('ROM: %s').format(status.rom) : _('no ROM'),
+			status.running ? _('running') : _('stopped'),
+			status.on_demand ? _('on demand') : '',
+			status.paused ? _('paused') : '',
+			status.stream_format || '',
+			status.stream_fps
+				? _('%s stream FPS').format(status.stream_fps)
+				: ''
+		].filter(Boolean);
+
+		return E('span', {}, parts.join(' · '));
+	},
+
+	romTable(list, error, notice, truncated) {
+		const rows = (list || []).map(item => {
+			const path = typeof item === 'string'
+				? item
+				: (item.path || item.name || '');
+			const label = typeof item === 'string'
+				? item
+				: (item.name || item.path || '');
+			const unreadable = typeof item === 'object' &&
+				item !== null && item.readable === false;
+			const reason = unreadable
+				? (item.error ||
+					_('Not readable by nesd; upload through LuCI or set group nesd and mode 0640.'))
+				: '';
+
+			return E('tr', { class: 'tr' }, [
+				E('td', { class: 'td' }, unreadable
+					? [
+						label,
+						E('br'),
+						E('small', { class: 'warning' }, reason)
+					]
+					: label),
+				E('td', { class: 'td' }, [
+					E('button', {
+						class: 'btn cbi-button cbi-button-apply',
+						'data-nes-action': '1',
+						'data-nes-kind': 'load',
+						'data-nes-unavailable': unreadable ? '1' : '0',
+						disabled: Boolean(
+							this.actionBusy ||
+							!path ||
+							!this.canControl ||
+							unreadable
+						),
+						title: reason,
+						click: ui.createHandlerFn(this, 'onLoad', path)
+					}, _('Load'))
+				])
+			]);
+		});
+
+		if (!rows.length) {
+			rows.push(E('tr', { class: 'tr' }, [
+				E('td', { class: 'td', colspan: 2 },
+					E('em', {}, error || (this.canControl
+						? _('No ROMs — upload one above')
+						: _('No ROMs found'))))
+			]));
+		}
+		if (truncated || notice) {
+			rows.push(E('tr', { class: 'tr' }, [
+				E('td', { class: 'td', colspan: 2 },
+					E('small', {}, [
+						truncated ? _('ROM list was truncated. ') : '',
+						notice || ''
+					]))
+			]));
+		}
+
+		return E('table', { class: 'table' }, [
+			E('tr', { class: 'tr table-titles' }, [
+				E('th', { class: 'th' }, _('File')),
+				E('th', { class: 'th' }, _('Action'))
+			]),
+			...rows
+		]);
+	},
+
+	notifyError(title, error) {
+		ui.addNotification(
+			null,
+			E('p', {}, _('%s: %s').format(title, error.message || error)),
+			'error'
+		);
+	},
+
+	statusHasGame(status) {
+		if (!status || status.error)
+			return false;
+		if (typeof status.game_loaded === 'boolean')
+			return status.game_loaded;
+		return Boolean(status.rom || status.rom_path);
+	},
+
+	stateAllowsAction(kind, status) {
+		if (kind === 'load' || kind === 'upload')
+			return true;
+		if (!status || status.error)
+			return false;
+
+		const loaded = this.statusHasGame(status);
+		switch (kind) {
+		case 'start':
+			return loaded && !status.running;
+		case 'pause':
+		case 'reset':
+		case 'stop':
+			return loaded && Boolean(status.running);
+		default:
+			return true;
+		}
+	},
+
+	syncActionButtons() {
+		document.querySelectorAll('[data-nes-action]').forEach(button => {
+			const kind = button.getAttribute('data-nes-kind') || '';
+			const unavailable =
+				button.getAttribute('data-nes-unavailable') === '1';
+			button.disabled = Boolean(
+				this.actionBusy ||
+				!this.canControl ||
+				unavailable ||
+				!this.stateAllowsAction(kind, this.lastStatus)
+			);
+			if (kind === 'pause') {
+				button.textContent = this.lastStatus &&
+					this.lastStatus.paused
+					? _('Resume')
+					: _('Pause');
+			}
+		});
+	},
+
+	setActionBusy(busy) {
+		this.actionBusy = busy;
+		this.syncActionButtons();
+	},
+
+	async runAction(action, successText, timeoutMs = 30000) {
+		if (this.actionBusy)
+			return;
+		this.setActionBusy(true);
+		try {
+			requireSuccess(await withTimeout(action(), timeoutMs));
+			if (successText) {
+				ui.addNotification(null, E('p', {}, successText), 'info');
+			}
+			await this.onRefreshAll();
+		}
+		catch (error) {
+			this.notifyError(_('Operation failed'), error);
+		}
+		finally {
+			this.setActionBusy(false);
+		}
+	},
+
+	async onLoad(path) {
+		if (!this.canControl || !path)
+			return;
+		await this.runAction(
+			() => callLoad(path),
+			_('Loaded %s').format(path),
+			nesEmulator.getActionTimeout()
+		);
+	},
+
+	async onUpload() {
+		let staged = '';
+
+		if (!this.canControl || this.actionBusy)
+			return;
+		this.setActionBusy(true);
+		try {
+			const reservation = requireSuccess(
+				await withTimeout(callReserveUpload(), 20000)
+			);
+			staged = reservation.staged || '';
+			if (!/^\/tmp\/nes-emulator-upload\/[0-9a-f]{32}\.rom$/.test(staged))
+				throw new Error(_('Invalid upload reservation'));
+			const reply = await ui.uploadFile(
+				staged,
+				null,
+				_('Supported: iNES/NES 2.0, UNIF and FDS; maximum 16 MiB')
+			);
+			// rpcd import is transactional but cannot be cancelled. Waiting for
+			// its real result avoids a false timeout followed by a racing discard.
+			const result = await callImport(
+				staged,
+				reply.name || 'upload.nes'
+			);
+			if (result && result.stored)
+				staged = '';
+			if (result && result.stored &&
+			    (result.error || result.ok === false)) {
+				ui.addNotification(
+					null,
+					E('p', {}, _('ROM was stored but could not be loaded: %s (%s)')
+						.format(
+							result.path || reply.name || _('uploaded ROM'),
+							result.error || _('Operation failed')
+						)),
+					'warning'
+				);
+				await this.onRefreshAll();
+				return;
+			}
+			requireSuccess(result);
+			staged = '';
+			ui.addNotification(
+				null,
+				E('p', {}, _('ROM uploaded and loaded: %s')
+					.format(result.path || reply.name)),
+				'info'
+			);
+			await this.onRefreshAll();
+		}
+		catch (error) {
+			if (staged)
+				await withTimeout(callDiscardUpload(staged), 20000)
+					.catch(() => {});
+			this.notifyError(_('Upload failed'), error);
+		}
+		finally {
+			this.setActionBusy(false);
+		}
+	},
+
+	onStart() {
+		if (!this.canControl ||
+		    !this.stateAllowsAction('start', this.lastStatus))
+			return Promise.resolve();
+		return this.runAction(
+			callStart,
+			_('Emulation started'),
+			nesEmulator.getActionTimeout()
+		);
+	},
+
+	onStop() {
+		if (!this.canControl ||
+		    !this.stateAllowsAction('stop', this.lastStatus))
+			return Promise.resolve();
+		return this.runAction(callStop, _('Emulation stopped'));
+	},
+
+	onPause() {
+		if (!this.canControl ||
+		    !this.stateAllowsAction('pause', this.lastStatus))
+			return Promise.resolve();
+		return this.runAction(callPause);
+	},
+
+	onEmuReset() {
+		if (!this.canControl ||
+		    !this.stateAllowsAction('reset', this.lastStatus))
+			return Promise.resolve();
+		return this.runAction(callEmuReset, _('Console reset'));
+	},
+
+	async onRefreshAll() {
+		const generation = (this.refreshGeneration || 0) + 1;
+		this.refreshGeneration = generation;
+		const statusRequest = withTimeout(callStatus(), 5000)
+			.catch(error => ({ error: error.message || String(error) }))
+			.then(status => {
+				if (generation !== this.refreshGeneration)
+					return;
+				this.lastStatus = status;
+				const statusNode = document.getElementById('nes-status');
+				if (statusNode)
+					statusNode.replaceChildren(this.statusText(status));
+				this.syncActionButtons();
+			});
+		const romRequest = withTimeout(callRoms(), 15000)
+			.catch(error => ({
+				roms: [],
+				error: error.message || String(error)
+			}))
+			.then(roms => {
+				if (generation !== this.refreshGeneration)
+					return;
+				const romNode = document.getElementById('nes-rom-table');
+				if (romNode) {
+					romNode.replaceChildren(
+						this.romTable(
+							roms.roms || [],
+							roms.error,
+							roms.notice,
+							roms.truncated
+						)
+					);
+				}
+				this.syncActionButtons();
+			});
+
+		await Promise.all([ statusRequest, romRequest ]);
+		return generation === this.refreshGeneration;
+	},
+
+	handleSaveApply: null,
+	handleSave: null,
+	handleReset: null
+});
--- a/applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/settings.js
+++ b/applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/settings.js
@@ -1,0 +1,274 @@
+'use strict';
+'require view';
+'require form';
+'require ui';
+'require nes-emulator as nesEmulator';
+
+const callRotateToken = nesEmulator.declareLongRunningRpc({
+	object: 'nes-emulator',
+	method: 'rotate_token',
+	expect: { '': {} }
+});
+
+function validateDataDir(sectionId, value) {
+	if (!value || value[0] !== '/')
+		return _('An absolute directory path is required');
+	if (value.includes('\0') || value.includes('\n') ||
+	    value.includes(':') || value.includes(';') ||
+	    value.includes(',') || value.includes('//'))
+		return _('The directory contains unsupported characters');
+
+	const parts = value.split('/');
+	if (parts.includes('.') || parts.includes('..'))
+		return _('Relative path components are not allowed');
+
+	const forbidden = [
+		'/', '/bin', '/boot', '/dev', '/etc', '/home', '/lib',
+		'/lib64', '/mnt', '/opt', '/overlay', '/proc', '/root',
+		'/run', '/sbin', '/sys', '/tmp', '/usr', '/var'
+	];
+	if (forbidden.includes(value.replace(/\/+$/, '') || '/'))
+		return _('Choose a dedicated subdirectory, not a system root');
+	return true;
+}
+
+function validateOptionalDataDir(sectionId, value) {
+	if (!value)
+		return true;
+	return validateDataDir(sectionId, value);
+}
+
+function validateOrigin(sectionId, value) {
+	if (!value)
+		return true;
+	try {
+		const parsed = new URL(value);
+		if ((parsed.protocol !== 'http:' && parsed.protocol !== 'https:') ||
+		    parsed.username || parsed.password || parsed.search ||
+		    parsed.hash || parsed.pathname !== '/' ||
+		    value !== parsed.origin)
+			throw new Error();
+		return true;
+	}
+	catch (error) {
+		return _('Enter an exact Origin such as https://192.168.1.1');
+	}
+}
+
+function validateRom(sectionId, value) {
+	if (!value)
+		return true;
+	const dirResult = validateDataDir(sectionId, value);
+	if (dirResult !== true)
+		return dirResult;
+	if (!/\.(nes|fds|unf|unif)$/i.test(value))
+		return _('Expected a .nes, .fds, .unf or .unif file');
+	return true;
+}
+
+return view.extend({
+	render() {
+		const map = new form.Map(
+			'nes-emulator',
+			_('NES Emulator settings'),
+			_('nesd contains a statically linked FCEUmm core. Emulation, PPU rendering and JPEG encoding stay on the router CPU.')
+		);
+
+		const section = map.section(
+			form.TypedSection,
+			'nes-emulator',
+			_('Service')
+		);
+		section.anonymous = true;
+		section.addremove = false;
+
+		let option;
+
+		option = section.option(
+			form.Flag,
+			'enabled',
+			_('Start at boot'),
+			_('Keep disabled unless the emulator should start on every boot. Loading a ROM starts it on demand for the current uptime.')
+		);
+		option.default = option.disabled;
+		option.rmempty = false;
+
+		option = section.option(
+			form.DummyValue,
+			'_core',
+			_('Emulator core')
+		);
+		option.cfgvalue = () => _('FCEUmm (statically linked into nesd)');
+
+		option = section.option(
+			form.Value,
+			'rom_dir',
+			_('Primary ROM directory'),
+			_('Validated LuCI uploads are stored here')
+		);
+		option.default = '/etc/nes-emulator/roms';
+		option.rmempty = false;
+		option.validate = validateDataDir;
+
+		option = section.option(
+			form.Flag,
+			'extra_rom_dirs_enabled',
+			_('Enable extra ROM directories'),
+			_('Hard-disable scanning and daemon access to every extra directory while keeping the saved paths ready for re-enabling.')
+		);
+		option.default = option.disabled;
+		option.rmempty = false;
+
+		option = section.option(
+			form.DynamicList,
+			'extra_rom_dir',
+			_('Extra ROM directories'),
+			_('Read-only USB or SD paths to scan, for example /mnt/sda1/roms')
+		);
+		option.placeholder = '/mnt/sda1/roms';
+		option.validate = validateOptionalDataDir;
+		option.retain = true;
+		option.depends('extra_rom_dirs_enabled', '1');
+
+		option = section.option(
+			form.Value,
+			'system_dir',
+			_('System directory')
+		);
+		option.default = '/etc/nes-emulator/system';
+		option.rmempty = false;
+		option.validate = validateDataDir;
+
+		option = section.option(
+			form.Value,
+			'save_dir',
+			_('Save directory'),
+			_('Battery-backed SRAM and full save states are written here using atomic replacement')
+		);
+		option.default = '/etc/nes-emulator/saves';
+		option.rmempty = false;
+		option.validate = validateDataDir;
+
+		option = section.option(
+			form.Value,
+			'bind',
+			_('Bind address'),
+			_('Use 0.0.0.0 for LAN access. Requests still require the generated token.')
+		);
+		option.datatype = 'ipaddr';
+		option.default = '0.0.0.0';
+		option.rmempty = false;
+
+		option = section.option(
+			form.Value,
+			'port',
+			_('HTTP / WebSocket port')
+		);
+		option.datatype = 'port';
+		option.default = '29876';
+		option.rmempty = false;
+
+		option = section.option(
+			form.Value,
+			'allowed_origin',
+			_('Restrict browser Origin'),
+			_('Optional exact Origin for a cross-origin custom client. The built-in game page is same-origin with nesd, so normally leave this empty.')
+		);
+		option.placeholder = 'https://client.example';
+		option.validate = validateOrigin;
+
+		option = section.option(
+			form.Button,
+			'_rotate_token',
+			_('Access token'),
+			_('The token is generated automatically and is required by every API and WebSocket client.')
+		);
+		option.inputtitle = _('Rotate token');
+		option.inputstyle = 'action';
+		option.onclick = async () => {
+			ui.showModal(
+				_('Rotating access token'),
+				E('p', {}, _('Restarting nesd…'))
+			);
+			try {
+				const reply = await callRotateToken();
+				if (!reply || reply.error || reply.ok === false)
+					throw new Error(reply && reply.error || _('Operation failed'));
+				ui.hideModal();
+				ui.addNotification(
+					null,
+					E('p', {}, _('Access token rotated; old game windows were disconnected.')),
+					'info'
+				);
+			}
+			catch (error) {
+				ui.hideModal();
+				ui.addNotification(
+					null,
+					E('p', {}, _('Token rotation failed: %s')
+						.format(error.message || error)),
+					'error'
+				);
+			}
+		};
+
+		option = section.option(
+			form.ListValue,
+			'stream_format',
+			_('Video stream format'),
+			_('Each raw RGB565 frame is 120 KiB: video uses about 2.0 Mbit/s at the default 2 FPS and 59.0 Mbit/s at 60 FPS. PCM 48 kHz stereo int16 adds about 1.54 Mbit/s, for roughly 3.5 or 60.5 Mbit/s total before WebSocket and Wi-Fi overhead. Raw is pixel-perfect but can consume substantial LAN airtime. JPEG reduces video bandwidth but is encoded in software on the router.')
+		);
+		option.value('raw', _('Raw RGB565 (pixel-perfect)'));
+		option.value('jpeg', _('Software JPEG (lower bandwidth)'));
+		option.default = 'raw';
+		option.rmempty = false;
+
+		option = section.option(
+			form.Value,
+			'jpeg_quality',
+			_('JPEG quality')
+		);
+		option.datatype = 'range(1,100)';
+		option.default = '92';
+		option.rmempty = false;
+		option.depends('stream_format', 'jpeg');
+
+		option = section.option(
+			form.Value,
+			'stream_fps',
+			_('Maximum streamed FPS'),
+			_('Allowed range: 1–60 FPS. This only limits frame delivery; FCEUmm keeps the ROM region’s native 50/60 FPS timing. Raw bandwidth scales directly with this value, so use 1–2 FPS on slow or shared Wi-Fi.')
+		);
+		option.datatype = 'range(1,60)';
+		option.default = '2';
+		option.rmempty = false;
+
+		option = section.option(
+			form.Flag,
+			'show_fps',
+			_('Show FPS counter'),
+			_('Draw only the browser-painted FPS number as a FCEUX-like pixel OSD directly over the NES canvas in its top-right corner, without a separate browser widget. The number reflects delivery, decode and paint slowdowns rather than the ROM’s native frame rate or the configured stream limit. The setting is applied after Save & Apply and the game window reconnects.')
+		);
+		option.default = option.enabled;
+		option.rmempty = false;
+
+		option = section.option(
+			form.Flag,
+			'show_touch_controls',
+			_('Show on-screen controls'),
+			_('Show the clickable and touch-friendly NES buttons in the game window. Disable this on a computer to free screen space; keyboard and gamepad controls remain available. Save & Apply, then reconnect the game window to apply the setting.')
+		);
+		option.default = option.enabled;
+		option.rmempty = false;
+
+		option = section.option(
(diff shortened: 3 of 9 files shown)


applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1516 @@
...
+		ext="$(
+			exec 9>&-
+			printf '%s' "${name##*.}" |
+				tr 'ABCDEFGHIJKLMNOPQRSTUVWXYZ' 'abcdefghijklmnopqrstuvwxyz'
+		)"
+		case "$ext" in nes|fds|unf|unif) ;; *) continue ;; esac
+		size="$(
+			exec 9>&-
+			du -b "$full" 2>/dev/null |
+				awk 'NR == 1 { print $1 }'
```
openwrt-ai: `du -b` is a GNU coreutils option. BusyBox `du` — the only `du` on a stock OpenWrt image, and `coreutils-du` is not in `LUCI_DEPENDS` — accepts only `-aHLdclsxhmk` and exits with a usage error here. Because the result is piped into `awk`, the pipeline status is `awk`'s, so the failure is silent and `$size` ends up empty; the `case "$size" in ''|*[!0-9]*) continue` guard on the next line then drops every ROM, and the offline ROM list renders empty on any router where nesd is stopped.

`wc -c` also reports apparent size rather than allocated blocks, which is what the size and quota checks actually want:

```suggestion
			wc -c <"$full" 2>/dev/null | tr -d ' \t'
```

The same `du -b` is used at lines 971, 1203 and 1300, each with the identical silent-empty-string failure: line 971 makes overwriting an existing ROM always return "ROM quota exceeded", line 1203 makes every import fail with "ROM size must be between 16 bytes and 16 MiB". Line 951 uses `-exec du -b '{}' ';'`, where the empty scan file leaves `used` at 0 and turns the 128 MiB `ROM_QUOTA_SIZE` into a no-op — `-exec wc -c '{}' ';'` prints `SIZE PATH`, which the existing `$1` awk sum already handles.

For what it's worth, neither openwrt.git nor luci.git uses `du -b` anywhere today.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1516 @@
...
+	fi
+	# Revalidate at commit time. The completed temp file is now included in
+	# the directory scan, so concurrent direct /api/upload traffic cannot
+	# make this import exceed the quota or free-space reserve.
+	if ! storage_is_available "$dest" 0; then
+		rm -f "$tmp" "$staged" 9>&-
+		json_error "ROM quota exceeded or less than 8 MiB would remain free"
+		return
+	fi
+	if ! mv -fT "$tmp" "$dest" 9>&-; then
```
openwrt-ai: `mv -T` is GNU coreutils as well; BusyBox `mv` implements only `-f`, `-i`, `-n` (plus `-v`), so this fails with a usage error on a stock OpenWrt image and every import ends in "cannot store ROM in the configured directory" — and since the error path also removes `$staged`, the user has to re-upload just to hit the same failure.

`-T` is not needed here: lines 1269-1275 already reject a `$dest` that exists but is not a regular file, so a plain `mv -f` cannot silently turn into a move-into-directory.

```suggestion
	if ! mv -f "$tmp" "$dest" 9>&-; then
```

Like `du -b`, `mv -T` appears nowhere in openwrt.git or luci.git today.

---

applications/luci-app-nes-emulator/Makefile
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: MIT
+
+include $(TOPDIR)/rules.mk
+
+PKG_RELEASE:=1
+PKG_LICENSE:=MIT
+PKG_LICENSE_FILES:=files/LICENSE-MIT
```
openwrt-ai: `PKG_LICENSE_FILES` resolves relative to `$(PKG_BUILD_DIR)`, and luci.mk's `Build/Prepare` copies only `luasrc ucode htdocs root src` into it — see [the copy loop at `luci.mk:181`](https://github.com/openwrt/luci/blob/32686a850252b3ef7c1540c32e877e491dea82e2/luci.mk#L181). `files/` is never copied, so this declaration points at a path that does not exist in the build directory, and `files/LICENSE-MIT` is not installed into the package either — it is dead weight in the tree as it stands.

[`luci-theme-footstrap`](https://github.com/openwrt/luci/blob/32686a850252b3ef7c1540c32e877e491dea82e2/themes/luci-theme-footstrap/Makefile#L110-L115) is the one package in tree that hits this and documents the workaround: keep the license text at the package root and copy it in from a `Build/Prepare/<name>` hook.

Simplest fix is to drop this line — the large majority of `luci-app-*` Makefiles declare `PKG_LICENSE` only, with no license file — and drop `files/LICENSE-MIT` with it. Otherwise move the file to the package root and add the hook.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1516 @@
...
+			exit 0
+		}
+		start_daemon || {
+			json_error "${START_ERROR:-nesd could not be started}"
+			exit 0
+		}
+		json_init
+		json_add_string path "$path"
+		body="$(json_dump_without_upload_fd)"
+		resp="$(api_post /api/load "$body")"
```
openwrt-ai, line 1574: `path` comes straight from the client and is forwarded to `/api/load` verbatim: it is not checked against `$ROM_DIR` or the configured extra roots, and it skips the `valid_data_dir` / `readlink -f` / extension / header checks that `import_rom_locked` applies to an imported ROM.

That stands out because every other client-supplied path in this script is confined — `staged_path_is_valid` for uploads, `valid_data_dir` plus resolved-equals-configured for the ROM directory — so this reads like a deliberate exception rather than an oversight. Is nesd expected to confine `path` to the configured roots itself? If so it would help to say so in a comment here, since the ROM list the UI offers is already restricted to those roots and any caller holding the `write` ACL can pass an arbitrary path.

---

applications/luci-app-nes-emulator/Makefile
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: MIT
+
+include $(TOPDIR)/rules.mk
+
+PKG_RELEASE:=1
```
openwrt-ai: nit: `PKG_RELEASE` has no effect without `PKG_VERSION`. luci.mk sets `VERSION:=$(if $(PKG_VERSION),$(if $(PKG_RELEASE),$(PKG_VERSION)-r$(PKG_RELEASE),$(PKG_VERSION)),$(PKG_SRC_VERSION))`, so with no `PKG_VERSION` the package version falls back to the git-derived `PKG_SRC_VERSION` and this line is ignored. All 15 `applications/*/Makefile` that set `PKG_RELEASE` also set `PKG_VERSION`; the other 87 set neither.

```suggestion
```

Alternatively add a `PKG_VERSION` above it if you want a manually versioned package.

---

openwrt-ai (commented):
Reviewed 1 new commit; comments inline.

The headline items are two GNU-coreutils-only options in the rpcd bridge (`du -b`, `mv -fT`) that BusyBox does not implement — they would break the ROM listing, the quota accounting and every ROM import on a stock OpenWrt image, mostly silently. The stated dependency on openwrt/packages#30350 and the draft status are noted; nothing in this review depends on that package landing first.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in 5028c07a. All `du -b` uses are gone. Regular non-symlink file sizes now come from a shared `wc -c` helper, and quota scanning uses BusyBox-compatible `find -exec wc -c`. The resource contract exercises both single-file and directory accounting.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in 5028c07a. The commit now uses BusyBox-compatible `mv -f`; the existing destination guard still rejects anything that exists but is not a regular non-symlink file.

applications/luci-app-nes-emulator/Makefile
communism420: Fixed in 5028c07a. `PKG_LICENSE_FILES` and the unreachable `files/LICENSE-MIT` copy have been removed from the LuCI submission; the Makefile retains the normal `PKG_LICENSE:=MIT` declaration.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Yes, `nesd` is the final confinement boundary. Its `/api/load` path flows through `resolve_rom_path()` / `validate_rom_path()`: `realpath()` canonicalization, containment under a configured ROM root, extension validation, `O_NOFOLLOW`, maximum-size checks, and ROM header validation before loading. The RPCD code now documents that boundary. The integration suite also sends a valid ROM outside every configured root and an in-root symlink escape; both are rejected with HTTP 400. Relevant native validation: https://github.com/communism420/openwrt-nes-emulator/blob/v1.0.0-r19/package/nes-emulator/src/http.c#L2532-L2566

applications/luci-app-nes-emulator/Makefile
communism420: Fixed in 5028c07a. The ineffective `PKG_RELEASE` assignment has been removed; this application now follows the normal git-derived LuCI package version.

openwrt-ai (commented):

The delta since `32686a85` addresses all five earlier findings, and I could not find a regression introduced by the fixes:

- The new `file_size_bytes()` helper is BusyBox-safe (`wc -c <file`, no GNU-only flags), guards against symlinks and non-regular files, normalises BusyBox `wc`'s width padding with `tr -d ' \t'`, and rejects non-numeric output — so the three former `du -b` call sites (`scan_rom_root`, `storage_is_available`'s `old_size`, and the two size checks in `import_rom_locked`) now fail closed rather than silently yielding an empty string. It also follows the script's existing `exec 7>&- 9>&-` fd-hygiene convention.
- The quota scan's `find … -exec wc -c '{}' ';'` emits `SIZE PATH`, which the existing `$1`-summing awk consumes unchanged; the `$1 !~ /^[0-9]+$/ { exit 2 }` guard still propagates a non-zero status through the `END` block, so a malformed scan file makes `storage_is_available` return failure instead of a bogus total.
- `mv -f` is BusyBox-compatible and the "destination is not a regular file" guard remains in place ahead of it.
- The `PKG_LICENSE_FILES` line and `files/LICENSE-MIT` are both gone, leaving no dangling reference; `PKG_LICENSE:=MIT` plus the SPDX header matches the common `luci-app-*` pattern. The ineffective `PKG_RELEASE` is removed.
- The forwarded `load` path now documents nesd as the confinement boundary at the call site, which was what that thread asked for.

Commit message matches the diff, and all three FormalityCheck runs are green on `5028c07a`.

---

communism420:
Hardware testing update: this exact LuCI head (5028c07a) was built with the official OpenWrt 25.12.5 mediatek/filogic SDK and tested successfully on an ASUS RT-AX52 Pro (aarch64_cortex-a53) together with the native package from openwrt/packages#30350. LuCI pages, service controls, settings, ROM upload/loading, streamed video and audio, controller input, save/load state, and the FCEUX-style FPS overlay all worked. The PR description now records the test evidence. This PR intentionally remains Draft until the dependent native package is merged, per the LuCI template.

communism420:
Coordination update: native package review identified that both init and the LuCI RPCD bridge must lock the shared `auth.token` through the same protected path. Current head `db1acc3c` therefore updates the RPCD bridge to use `/etc/nes-emulator/.auth.token.lock`, catches lock-open failures under BusyBox ash, serializes rotation with RPC startup, and rechecks a daemon that may have started directly while rotation waited. No UI, ACL, menu, or translation file changed.

The exact amended native/LuCI exports passed the full project contracts and a clean current OpenWrt Snapshot x86/64 SDK package check/build. The LuCI APK still directly depends on `nes-emulator`. This PR remains Draft until openwrt/packages#30350 is merged, as required for dependent LuCI submissions.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1531 @@
...
+# Authenticated rpcd bridge. LuCI talks only to rpcd; rpcd talks to nesd on
+# the router and never exposes init-script execution through the ACL.
+
+. /lib/functions.sh 2>/dev/null
+. /usr/share/libubox/jshn.sh 2>/dev/null
+
+CONFIG_FILE=/etc/config/nes-emulator
+TOKEN_FILE=/etc/nes-emulator/auth.token
+UPLOAD_DIR=/tmp/nes-emulator-upload
+TOKEN_LOCK_FILE=/etc/nes-emulator/.auth.token.lock
```
openwrt-ai: Moving the token lock under `/etc/` puts it on the read-write overlay, i.e. on flash. `ensure_auth_token` runs unconditionally on **every** rpcd invocation (`ensure_auth_token` at line 315 calls `acquire_token_lock` at line 319, before the `case "$1"` dispatch), and `command exec 8>"$TOKEN_LOCK_FILE"` opens it `O_CREAT|O_TRUNC`, which dirties the inode and costs a JFFS2/UBIFS metadata write on every `status`, `roms`, `access`, `start`, `stop` and upload call the UI makes. The old `/var/lock` path is tmpfs and had no such cost.

The hazard being avoided is real — `/var/lock` is [`chmod 1777` in base-files' `boot` init](https://github.com/openwrt/openwrt/blob/3d902d07a9d3ac8dff5807aec18d54bbd25c2d28/package/base-files/files/etc/init.d/boot#L26-L27) — but this new file gets none of the hardening the start lock gets. `prepare_start_lock_file()` (line 267) creates it under `set -C` + `umask 077` and then runs `validate_regular_metadata`, and `acquire_start_lock()` re-validates after opening (line 289). `acquire_token_lock()` does neither, and because the script's own `umask 077` is only applied *after* `ensure_auth_token` has already run, the very first creation inherits rpcd's umask and lands as `0644`.

A root-owned tmpfs lock created with the `prepare_start_lock_file` pattern would give both protection and zero flash writes. Separately, `/etc/nes-emulator` is the *native* package's directory, so a runtime file dropped there survives `apk del nes-emulator`. I could not verify the native side of this (openwrt/packages#30350 isn't in the feed yet) — is the init script really flocking this exact path, or would a protected tmpfs path work for both?

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1531 @@
...
+UPLOAD_LOCK_FILE=/var/lock/nes-emulator-upload.lock
```
openwrt-ai: The upload lock is left in `/var/lock` with the exposure the token lock was just moved out to avoid. `/var/lock` is [`chmod 1777`](https://github.com/openwrt/openwrt/blob/3d902d07a9d3ac8dff5807aec18d54bbd25c2d28/package/base-files/files/etc/init.d/boot#L26-L27), and `acquire_upload_lock()` opens this path unguarded at line 252 — `exec 9>"$UPLOAD_LOCK_FILE" || return 1` — with no `set -C` pre-creation and no `validate_regular_metadata`, unlike `START_LOCK_FILE`.

Any local unprivileged account (including the `nesd` service user) can create that name before rpcd first touches it. As a symlink, root's `exec 9>` follows it and truncates whatever it points at; as a directory, the open fails and — see the separate note on line 199 — the whole rpcd script dies. The sticky bit stops others *deleting* the file, not creating it first.

If the token lock warranted a protected path, this one does too: same treatment (root-owned pre-created file plus post-open metadata re-validation, as `prepare_start_lock_file` already demonstrates).

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1531 @@
...
+	[ "$UPLOAD_LOCK_HELD" -eq 1 ] ||
+		[ "$START_LOCK_HELD" -eq 1 ] ||
+		trap - 0 1 2 15
+}
+
+acquire_token_lock() {
+	local attempts=0
+
+	[ "$TOKEN_LOCK_HELD" -eq 0 ] || return 0
+	command exec 8>"$TOKEN_LOCK_FILE" || return 1
```
openwrt-ai: The `command` prefix is correct and genuinely needed here — but it was applied to only one of the three `exec` redirections, and the other two have exactly the same defect.

`exec` is a POSIX *special* built-in, so a redirection error on it terminates a non-interactive shell and the `|| return 1` is never reached. `command` strips the special-ness. Confirmed on dash (BusyBox ash is the same lineage):

```
-- command exec failure test:
t2.sh: 1: cannot create /nonexistent/lock: Directory nonexistent
caught failure, rc=2    <- script continued, rc=1
-- plain exec failure test:
t3.sh: 1: cannot create /nonexistent/lock: Directory nonexistent
   script rc=2          <- shell exited, `|| ...` never ran
```

Still bare:

- line 252 — `exec 9>"$UPLOAD_LOCK_FILE" || return 1`
- line 288 — `exec 7<>"$START_LOCK_FILE" || return 1`

Under BusyBox ash both `|| return 1` branches are dead code: an open failure (an attacker-planted directory at the `/var/lock` path, a lock file whose metadata check passed but whose open fails, `/var/lock` missing on an early-boot call) kills the rpcd script before any JSON is written, so LuCI receives an empty reply instead of the intended `"upload storage is busy"` / `"another NES Emulator startup operation is still in progress; retry shortly"` error. Both need the same `command exec` treatment.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1531 @@
...
+		json_error "another NES Emulator startup operation is still in progress; retry shortly"
+		return
+	}
+	/etc/init.d/nes-emulator running >/dev/null 2>&1 9>&- 7>&- &&
+		was_running=1
+	acquire_token_lock || {
+		release_start_lock
+		json_error "authentication token is busy"
+		return
+	}
```
openwrt-ai: Serialising rotation against startup is the right fix, but stacking the two waits pushes `rotate_token` past the LuCI RPC client's timeout.

`acquire_start_lock` retries 20 times with `sleep 1` (lines 293-300) and `acquire_token_lock` retries 10 times with `sleep 1` (lines 200-207), so under contention this function can now block ~30 s before it even reaches the `restart`. LuCI's RPC client aborts at 20 s by default — [`rpc.js:38`](https://github.com/openwrt/luci/blob/db1acc3ca373ba70120dd9605f694f66f8ffdbb7/modules/luci-base/htdocs/luci-static/resources/rpc.js#L38) (`timeout: (L.env.rpctimeout ?? 20) * 1000`) — and `settings.js`'s `_rotate_token` `onclick` sets no override.

The result is the worst possible outcome for this particular operation: the browser gives up, the modal shows `Token rotation failed: Request timeout`, while server-side the rotation actually completes and disconnects every open game window. Before this change `rotate_token` waited on the token lock only (10 s max), which fit inside the budget.

Either bound the combined wait under 20 s, or raise the client timeout for this one call — [`luci-app-wwand` does the latter](``https://github.com/openwrt/luci/blob/9a6dc4403674a876c0afd4dd5b45b87992369002/applications/luci-app-wwand/htdocs/luci-static/resources/wwand/netsel.js#L176-L178)`` by saving/restoring `L.env.rpctimeout` around a long-running RPC.

---

openwrt-ai (commented):
Reviewed the delta from `5028c07a` to `db1acc3c` (rpcd bridge only: shared token lock path, `command exec`, start-lock serialisation of `rotate_token`).

The direction is right — `command exec` is genuinely required for the `|| return 1` to be reachable under ash, the start-lock/token-lock ordering is consistent with `start_daemon` so there is no deadlock, and the post-rotation re-check of a concurrently-started daemon closes the race it describes. Four things the delta leaves open, inline:

- the token lock now lives on the flash overlay and is written on every single RPC call, and it skips the hardening `prepare_start_lock_file` applies to the start lock;
- the upload lock keeps the exact `/var/lock` (mode 1777) exposure the token lock was moved away from;
- the `command` prefix was applied to one of three `exec` redirections; the other two still silently kill the script instead of returning a JSON error;
- `acquire_start_lock` (20 s) plus `acquire_token_lock` (10 s) exceeds LuCI's 20 s default RPC timeout, so a contended rotation reports failure to the user after it has already succeeded.

Commit message matches the diff. All three FormalityCheck runs are green on `db1acc3c`. The stated dependency on openwrt/packages#30350 and the draft status are noted; the native package is not in the feed yet, so I could not verify the init script's side of the shared lock path.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in `d3041a4f`, coordinated with native head `891dfd45`. Both writers now share `/var/run/nes-emulator/auth.token.lock` in a root-owned 0700 tmpfs directory. The 0600 single-link root-owned file is created with `set -C` and `umask 077`, opened read/write without truncation through `command exec`, and revalidated after open. This removes the per-RPC flash inode write and leaves no persistent runtime lock under `/etc`.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in `d3041a4f`. Token, upload, and startup locks now live together under the protected root-owned `/var/run/nes-emulator` tmpfs directory. The upload lock uses the common no-clobber 0600 preparation helper, `command exec 9<>`, and an exact post-open uid/gid/mode/link-count validation.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in `d3041a4f`. All three lock opens now use catchable, non-truncating `command exec` redirections (`7<>`, `8<>`, and `9<>`) after safe preparation and before post-open metadata validation. The BusyBox ash regression contract forces a missing-parent open failure for every fd and verifies that each function returns instead of terminating the script.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in `d3041a4f`. Ordinary calls retain the previous 10-attempt token-lock budget, but `call:rotate_token` selects a dedicated 3-attempt budget before the unconditional token check and uses a 4-attempt startup-lock budget. Because each loop sleeps only between attempts, the exact combined maximum is `(3-1) + (4-1) + (3-1) = 7` one-second sleeps, comfortably below the 20-second LuCI RPC timeout. The resource contract executes the method-specific selection and asserts that exact total.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1573 @@
...
+	START_ERROR=''
+	if [ -z "$HTTP_CLIENT" ]; then
+		START_ERROR="the matching /usr/bin/nesd loopback RPC client is missing; reinstall nes-emulator and luci-app-nes-emulator together"
+		return 1
+	fi
+
+	if daemon_is_ready; then
+		return 0
+	fi
+	if ! acquire_start_lock; then
```
openwrt-ai, line 697: The new per-method lock budget was wired up for `rotate_token` only. `start_daemon` still uses the default `START_LOCK_MAX_ATTEMPTS=20`, so the four methods that reach it can still blow past the same 20 s client timeout the comment at line 394-395 uses as the design constraint.

Worst case for `call:access`, `call:start`, `call:load` and `call:import`:

- `ensure_auth_token` (line 1455) runs before dispatch with the *default* `TOKEN_LOCK_MAX_ATTEMPTS=10` — the `case "$1:$2"` override at line 1452 only lowers it for `rotate_token` — so up to 9 × `sleep 1`;
- `acquire_start_lock` here retries 20 times, i.e. up to 19 more × `sleep 1`, before returning and producing `"another NES Emulator startup operation is still in progress; retry shortly"`;
- on the success path the readiness loop at [lines 696-704](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L696-L704)`` adds up to 15 more.

19 s on the start lock alone is already at the limit, and ~28 s once the pre-dispatch token wait is contended. [`rpc.js:38`](https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/modules/luci-base/htdocs/luci-static/resources/rpc.js#L38) is still `timeout: (L.env.rpctimeout ?? 20) * 1000`, and none of the four declarations override it — [`play.js` `callAccess`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/play.js#L5-L9)`` and [`overview.js` `callLoad` / `callImport` / `callStart`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/overview.js#L25-L56).`` That reproduces the exact symptom the rotate_token thread was about: the browser aborts and reports a failure while the start is still progressing server-side.

Was leaving these callers on the 20/10 budget deliberate? If a cold start genuinely needs to wait longer than the RPC client will, an `L.env.rpctimeout` override on those calls would be the other half of the fix; otherwise the start-lock budget needs to fit alongside the pre-dispatch token wait the way `ROTATE_START_LOCK_MAX_ATTEMPTS` does.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1573 @@
...
+		exec 8>&-
+		return 1
+	}
+	while ! flock -n 8 2>/dev/null; do
+		attempts=$((attempts + 1))
+		if [ "$attempts" -ge "$TOKEN_LOCK_MAX_ATTEMPTS" ]; then
+			exec 8>&-
+			return 1
+		fi
+		sleep 1
```
openwrt-ai: nit: this is the only one of the three lock-wait loops whose `sleep` keeps the lock fds open — [`acquire_upload_lock` uses `sleep 1 9>&-`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L309``) and [`acquire_start_lock` uses `sleep 1 9>&- 7>&-`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L334).`` Here the `sleep` child inherits fd 8, opened just above at line 240, and on the `rotate_token` path fd 7 as well, since `acquire_token_lock` is reached at line 402 while the start flock is held.

Each child lives at most one second, so this is not a real lock leak, but it contradicts the convention the file states for itself at lines 41-43.

```suggestion
		sleep 1 9>&- 8>&- 7>&-
```

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1573 @@
...
+SERVICE_INSTANCE_EXIT_CODE=''
+SERVICE_NESD_GID=''
+META_MODE=''
+META_NLINK=''
+META_UID=''
+META_GID=''
+
+# fd 9 owns the upload flock and fd 7 owns the startup flock. Keep both in this
+# rpcd shell only: no command-substitution shell, service process or external
+# jshn process may keep either lock alive after this request exits.
```
openwrt-ai: nit: fd 8 holds a real flock too (`acquire_token_lock`, line 240), but this inventory still names only fds 9 and 7 — and the older helpers follow it literally: [`json_load_without_upload_fd`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L48),`` [`json_dump_without_upload_fd`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L56``) and [`read_metadata`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L63``) close 7 and 9 only, and [`load_service_nesd_gid`](``https://github.com/openwrt/luci/blob/d3041a4f1f2be086bb23230dda10f55924aca71a/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L140``) closes 9 only. Both of the latter two do run under the token flock: `ensure_auth_token` calls `read_auth_token` at line 355 after `acquire_token_lock` has succeeded, and that in turn calls `load_service_nesd_gid` and `validate_regular_metadata`/`read_metadata`.

The two helpers added in this revision, `prepare_lock_directory` and `prepare_lock_file`, do close all three (lines 109 and 127), so the two sets now disagree with each other and with this comment. Every affected child (`ls`, `id`, `jshn`) is short-lived, so this is consistency rather than a leak — but the stated invariant should at least cover the fd it now applies to:

```suggestion
# fd 9 owns the upload flock, fd 8 owns the token flock and fd 7 owns the startup
# flock. Keep all three in this rpcd shell only: no command-substitution shell,
# service process or external jshn process may keep any of them alive after this
# request exits.
```

---

openwrt-ai (commented):
Reviewed the delta from `db1acc3c` to `d3041a4f` (rpcd bridge only: shared `/var/run/nes-emulator` lock directory, common `prepare_lock_file` helper, `command exec` on all three lock opens, parameterised lock budgets).

All four findings from the previous round are genuinely addressed. Moving the locks onto tmpfs removes the per-RPC flash write and gets the upload lock out of the world-writable `/var/lock`; `/var/run` is `/tmp/run` on stock OpenWrt ([`base-files/Makefile:227`](https://github.com/openwrt/openwrt/blob/7f6895da277d3f0c155c9c6120be669807d116b0/package/base-files/Makefile#L227), created root-owned `0755` by [`init.d/boot`](https://github.com/openwrt/openwrt/blob/7f6895da277d3f0c155c9c6120be669807d116b0/package/base-files/files/etc/init.d/boot#L29)), so the `0700` root-owned lock directory is not reachable by an unprivileged local account. The `command exec` prefix is now on all three redirections, the `<>` opens no longer truncate, and each is re-validated after opening. Start-lock ordering stays consistent between `start_daemon` and `rotate_token`, so there is no new deadlock. The arithmetic in the new comment at line 394-395 checks out: `(3-1) + (4-1) + (3-1) = 7` sleeps for a contended `rotate_token`.

Three things inline, one of them substantive:

- the bounded budget was applied to `rotate_token` only — `access`, `start`, `load` and `import` still get `START_LOCK_MAX_ATTEMPTS=20` plus the default 10-attempt pre-dispatch token wait, which reproduces the timeout symptom the earlier thread was about;
- two fd-hygiene nits: the `sleep` in `acquire_token_lock` is the only lock-wait loop that does not close the lock fds, and the fd inventory comment plus four older helpers still predate fd 8 becoming a lock.

Commit message matches the diff. All three FormalityCheck runs are green on `d3041a4f`. The stated dependency on openwrt/packages#30350 and the draft status are noted; the native package is still not in the feed, so I could not verify the init script's side of the shared `/var/run/nes-emulator` directory — in particular that it creates it with exactly `0700` root:root, which `prepare_lock_directory` requires on every call.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in `d34fd840`: the token-lock retry now uses `sleep 1 9>&- 8>&- 7>&-`, so the sleep child retains none of the request lock descriptors. The lock contract checks this exact closure.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in `d34fd840`. The inventory now documents fd 9 (upload), fd 8 (token), and fd 7 (startup). `json_load_without_upload_fd`, `json_dump_without_upload_fd`, `read_metadata`, and `load_service_nesd_gid` close all three in their external children; the resource contract asserts each cited helper.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in `d34fd840`. `access`, `start`, `load`, and `import` now use declarations which temporarily raise `L.env.rpctimeout` to at least 120 seconds, invoke the declared RPC, and immediately restore the prior global value after `rpc.js` captures it. Start/Load also use a 125-second local action guard; transactional Import still awaits its real result. The dynamic contracts record 120 seconds for every affected call and verify restoration.

applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/overview.js
```diff
@@ -0,0 +1,540 @@
...
+	method: 'roms',
+	expect: { '': { roms: [] } }
+});
+
+const callLoad = declareLongRunningRpc({
+	object: 'nes-emulator',
+	method: 'load',
+	params: [ 'path' ],
+	expect: { '': {} }
+});
```
openwrt-ai: `declareLongRunningRpc` raises a global, but the timeout that actually reaches the XHR is chosen per *batch*, not per call, so the raise can be silently discarded.

`rpc.call` captures the value into the request options synchronously ([`rpc.js:37-41`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/rpc.js#L37-L41)) — that part is fine. But since these declarations don't set `nobatch`, the request is queued rather than sent ([`isQueueableRequest`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/luci.js#L532-L545), [queue push at `luci.js:738-740`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/luci.js#L738-L740)), and `flushRequestQueue` builds the batched request from `requestQueue[0][0]` ([`luci.js:566`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/luci.js#L566)) — the *first* request queued in that animation frame supplies `timeout` for every request in the batch.

So when any other RPC lands in the same frame's queue, one of two things happens: the long call falls back to the other request's 20 s (exactly the symptom this change is meant to fix), or the unrelated call inherits 120 s — which also contradicts "unrelated concurrent calls keep their configured timeout" in the PR description.

That is not hypothetical for `access`, the call that can genuinely take ~40 s on a cold `start_daemon` (token wait + start-lock wait + readiness loop). `play.js` issues it from `view.load()`, which runs off the `luci-loaded` event ([`luci.js:1959-1966`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/luci.js#L1959-L1966)), and `ui.changes.init()` is registered on that same event and immediately fires `uci.changes()` ([`ui.js:4182`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/ui.js#L4182), [`ui.js:5136`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/ui.js#L5136)). Both land in the same queue before the next frame, and the batch timeout is whichever got there first.

Declaring these calls `nobatch: true` makes the timeout deterministic and keeps the 120 s away from unrelated calls — that is what [`callIwinfoScan`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/modules/luci-base/htdocs/luci-static/resources/network.js#L79-L85), the one long-running RPC declared in luci-base, does:

```suggestion
const callLoad = declareLongRunningRpc({
	object: 'nes-emulator',
	method: 'load',
	params: [ 'path' ],
	nobatch: true,
	expect: { '': {} }
});
```

The same applies to `callImport` (lines 54-59) and `callStart` (74-78) below, to `callAccess` in `play.js` (lines 25-29) and to `callRotateToken` in `settings.js` (lines 27-31).

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1574 @@
...
+write_auth_token() {
+	local token="$1" temporary
+
+	token_is_valid "$token" || return 1
+	validate_token_parent || return 1
+	if [ -e "$TOKEN_FILE" ] || [ -L "$TOKEN_FILE" ]; then
+		validate_regular_metadata "$TOKEN_FILE" -rw-r----- 0 \
+			"$SERVICE_NESD_GID" || return 1
+	fi
+	temporary="$(mktemp "${TOKEN_FILE}.tmp.XXXXXX" 2>/dev/null)" ||
```
openwrt-ai: nit: the fd-8 additions landed on the four helpers I happened to name last round, but the token-lock critical section itself still hands fd 8 to its own children — and unlike those four, these run with the lock held by construction.

Everything below `acquire_token_lock` in [`ensure_auth_token` (line 354)](``https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L350-L360)`` and in [`rotate_token` (line 403)](``https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L403-L411)`` forks external commands without closing it:

- `write_auth_token` here — `mktemp` (214), `chown` (217), `chmod` (218), `mv` (219)
- `new_token` — `hexdump` (172-173)
- `validate_token_parent` — `readlink -f` (183)
- `read_auth_token` — `cat` (195), `wc -c` / `wc -l` / `tr` (197-198)
- `remove_legacy_token` — the three `uci` invocations (343-346)

On the `rotate_token` path fd 7 is held across all of these too. Every child is short-lived, so this is consistency with the invariant at lines 41-43 ("no command-substitution shell, service process or external jshn process may keep a lock alive") rather than a real leak — but these are the sites where the lock is actually held, so if the contract is worth stating it is worth applying here.

---

applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/settings.js
```diff
@@ -0,0 +1,294 @@
...
+			LONG_RUNNING_RPC_TIMEOUT
+		);
+		try {
+			return call(...args);
+		}
+		finally {
+			L.env.rpctimeout = savedTimeout;
+		}
+	};
+}
```
openwrt-ai: nit: this block is byte-identical in all three views — `play.js` lines 5-23 and `overview.js` lines 6-26 — so the 120 s constant and the wrapper now have to be kept in sync by hand (the `nobatch` point on the `overview.js` thread, for instance, has to be applied in three places).

LuCI's pattern for code shared between the views of one app is an app resource module: [`luci-app-lldpd` keeps `lldpd.js` at `htdocs/luci-static/resources/lldpd.js`](https://github.com/openwrt/luci/blob/d34fd8400b40cfa18adbc3c9fef9bac1c9974915/applications/luci-app-lldpd/htdocs/luci-static/resources/lldpd.js) and pulls it into both its config and status views with `'require lldpd'`. A `resources/nes-emulator.js` exporting the helper would leave one copy.

---

openwrt-ai (commented):
Reviewed the delta from `d3041a4f` to `d34fd840`: the dynamic 120 s `L.env.rpctimeout` raise for `access`/`start`/`load`/`import`/`rotate_token`, the 125 s local action guard on Start/Load, and the fd-8 hygiene follow-ups in the rpcd bridge.

The save/raise/restore wrapper does capture the timeout at the right moment — `rpc.call` reads `L.env.rpctimeout` synchronously inside the `new Promise` executor, so the `finally` restore cannot race it, and the Start/Load client guard (125 s) is correctly above the RPC budget (120 s), while Import and Rotate await their real result. The rpcd side of the delta is fine: the fd inventory comment now matches reality, the token-lock retry `sleep` closes 7/8/9 like the other two loops, and the four cited helpers were updated.

One substantive point inline: raising a *global* is not equivalent to a per-call timeout, because luci-base batches RPC requests per animation frame and takes the whole batch's options — including `timeout` — from the first request queued. For `access`, which is issued from `view.load()` alongside luci-base's own `uci.changes()` on the `luci-loaded` event, that is likely to hand the batch a 20 s timeout again, i.e. the original symptom. `nobatch: true` on the long-running declarations makes it deterministic and is what luci-base does for `iwinfo scan`. Two nits: the remaining fd-8 sites are the ones inside the token-lock critical section, and the timeout helper is triplicated across the three views.

Commit message matches the diff, and all three FormalityCheck runs are green on `d34fd840`. The native package is still not in the packages feed, so the init-script side of the shared `/var/run/nes-emulator` lock directory remains unverified here; the draft status and the openwrt/packages#30350 dependency are noted.

---

applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/overview.js
communism420: Fixed in f8ba9e65fa. The shared long-running RPC helper clones every declaration with `nobatch: true`, so `load`, `import`, `start`, `access`, and `rotate_token` each capture their own timeout and cannot share it with an animation-frame batch. The previous global value is restored immediately after the synchronous RPC invocation. The regression contracts invoke all five declarations and assert both `nobatch` and timeout restoration.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in f8ba9e65fa. The listed `hexdump`, `readlink`, `cat`, `wc`/`tr`, `mktemp`, `chown`, `chmod`, `mv`, `rm`, and `uci` children now close fd 7/8/9. I also closed the outer `new_token` command-substitution shells at both lock-held call sites and the enabled-state lookup while the startup lock is held. The RPCD resource contract checks these exact critical-section sites.

applications/luci-app-nes-emulator/htdocs/luci-static/resources/view/nes-emulator/settings.js
communism420: Fixed in f8ba9e65fa. The duplicated constant and wrapper now live once in `htdocs/luci-static/resources/nes-emulator.js`, imported by Overview, Play, and Settings. The module also owns the unbatched declaration policy and computes the local Start/Load guard from the effective RPC timeout, including configured values above 120 seconds.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1595 @@
...
+			"$SERVICE_NESD_GID" || return 1
+	fi
+	temporary="$(
+		exec 7>&- 8>&- 9>&-
+		mktemp "${TOKEN_FILE}.tmp.XXXXXX" 2>/dev/null
+	)" ||
+		return 1
+	if ! printf '%s\n' "$token" >"$temporary" ||
+	   ! chown root:nesd "$temporary" 2>/dev/null 7>&- 8>&- 9>&- ||
+	   ! chmod 0640 "$temporary" 7>&- 8>&- 9>&- ||
```
openwrt-ai: nit: this delta dropped the `2>/dev/null` that used to be on the `chmod`. The previous head had `! chmod 0640 "$temporary" 2>/dev/null ||`; adding the fd closures replaced the stderr redirection instead of appending to it. The neighbouring `chown` on line 234 kept its `2>/dev/null`, so a failing `chmod` is now the only one of the four that writes to rpcd's stderr.

```suggestion
	   ! chmod 0640 "$temporary" 2>/dev/null 7>&- 8>&- 9>&- ||
```


---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
```diff
@@ -0,0 +1,1595 @@
...
+
+ensure_auth_token() {
+	local token
+
+	validate_token_parent || return 1
+	acquire_token_lock || return 1
+	if read_auth_token; then
+		token="$AUTH_TOKEN"
+	elif [ ! -e "$TOKEN_FILE" ] && [ ! -L "$TOKEN_FILE" ]; then
+		token="$(new_token 7>&- 8>&- 9>&-)"
```
openwrt-ai: nit: the `7>&- 8>&- 9>&-` here is now redundant and is the odd one out stylistically. `new_token` already closes all three inside its own command substitution ([lines 172-175](``https://github.com/openwrt/luci/blob/f8ba9e65fa54edd55bce2dd8a8d4b29f4c050410/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L172-L175)),`` and it is a shell function whose only external child is that `hexdump` — so the redirection on the function invocation closes nothing that stays open. Every other site this commit touched uses the `$( exec 7>&- 8>&- 9>&-; … )` form (e.g. [`read_auth_token` lines 200-203](``https://github.com/openwrt/luci/blob/f8ba9e65fa54edd55bce2dd8a8d4b29f4c050410/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L200-L203)),`` which makes this one read as if the function itself needed the guard. Same applies to line 426 in `rotate_token`. Either drop the redirection at both call sites or, if you want it belt-and-braces, keeping it is harmless — but worth making the two consistent with each other and with the rest of the file.

---

openwrt-ai (commented):
Reviewed the delta from `d34fd840` to `f8ba9e65`: the new shared `nes-emulator.js` resource with `nobatch: true` on the long-running declarations, the three views switching to it, and the remaining fd 7/8/9 closures in the rpcd bridge.

The substantive finding from the previous round is properly fixed. `nobatch: true` reaches the transport — `rpc.declare` forwards `options.nobatch` into [`rpc.call`](https://github.com/openwrt/luci/blob/f8ba9e65fa54edd55bce2dd8a8d4b29f4c050410/modules/luci-base/htdocs/luci-static/resources/rpc.js#L349), which passes it to `request.post`, and [`isQueueableRequest`](https://github.com/openwrt/luci/blob/f8ba9e65fa54edd55bce2dd8a8d4b29f4c050410/modules/luci-base/htdocs/luci-static/resources/luci.js#L539) then refuses to queue it. So `access` issued from `play.js` `load()` no longer inherits the animation-frame batch's 20 s timeout from whatever luci-base queued first; each of the five calls carries its own 120 s budget. The save/raise/restore wrapper still works because `rpc.call` reads `L.env.rpctimeout` synchronously inside the `new Promise` executor, ahead of the wrapper's `finally`.

The shared-resource mechanics check out too: `'require nes-emulator as nesEmulator'` matches the require regex (`(\S+)` accepts the hyphen, same as [`luci-app-olsr-viz`](``https://github.com/openwrt/luci/blob/f8ba9e65fa54edd55bce2dd8a8d4b29f4c050410/applications/luci-app-olsr-viz/htdocs/luci-static/resources/view/olsr-viz/olsr-viz-view.js#L7)),`` and [`require()` returns `new _class()`](https://github.com/openwrt/luci/blob/f8ba9e65fa54edd55bce2dd8a8d4b29f4c050410/modules/luci-base/htdocs/luci-static/resources/luci.js#L2556), so the `baseclass.extend` prototype members resolve as `nesEmulator.getActionTimeout` / `.declareLongRunningRpc`. A top-level `resources/*.js` for an app has in-tree precedent (`lldpd.js`, `olsr-viz.js`) and `luci.mk` copies `htdocs` wholesale, so nothing extra is needed to package it. `Number(L.env.rpctimeout) || 20` is not dead code — it is what keeps `Math.max` from returning `NaN` when the env value is unset.

On the rpcd side I walked the token critical section for a remaining holder and did not find one: `read_auth_token`, `write_auth_token`, `validate_token_parent`, `remove_legacy_token` and `new_token` now fork every external child with 7/8/9 closed, `printf` is a builtin, and the two `/etc/init.d/nes-emulator running` probes in `rotate_token` run when fd 8 is already released. `ROTATE_TOKEN_LOCK_MAX_ATTEMPTS` is applied at the dispatcher (line 1475), so the seven-sleep arithmetic in the comment holds.

Two nits inline, one of them a small regression this delta introduced (`chmod` lost its `2>/dev/null`).

Commit message matches the diff, and all three FormalityCheck runs are green on `f8ba9e65`. `nes-emulator` is still absent from the packages feed, so the init-script side of the shared `/var/run/nes-emulator` lock directory remains unverified here; the draft status and the openwrt/packages#30350 dependency are noted.

---

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Fixed in [`038ad57ad8`](https://github.com/openwrt/luci/commit/038ad57ad85184505509ccc9a9f19847bb3ef56c). Restored `2>/dev/null` on `chmod` while retaining the fd 7/8/9 closures, and added an exact regression contract for the full command.

applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator
communism420: Addressed in [`038ad57ad8`](https://github.com/openwrt/luci/commit/038ad57ad85184505509ccc9a9f19847bb3ef56c). Both call sites now use the same multiline command-substitution form, with `exec 7>&- 8>&- 9>&-` inside the outer substitution shell before `new_token`. This matches the cited style and genuinely closes that shell's copies (BusyBox ash may save function-call redirections on another fd for restoration); the nested `hexdump` child remains protected too. The two exact call-site forms are covered by regression contracts.

openwrt-ai (commented):

The delta from `f8ba9e65` to `038ad57a` is nine lines in the rpcd bridge and does exactly what the two open nits from the previous round asked for, with nothing else mixed in:

- [`chmod 0640 "$temporary" 2>/dev/null 7>&- 8>&- 9>&-`](``https://github.com/openwrt/luci/blob/038ad57ad85184505509ccc9a9f19847bb3ef56c/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L235``) restores the stderr redirection that the fd closures had displaced, so all four children in `write_auth_token`'s failure chain are consistent again. (`mv -f` on the next line still has no `2>/dev/null`, but it never had one and rpcd stderr only reaches syslog, so nothing to do there.)
- Both `new_token` call sites — [`ensure_auth_token`](``https://github.com/openwrt/luci/blob/038ad57ad85184505509ccc9a9f19847bb3ef56c/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L375-L378)`` and [`rotate_token`](``https://github.com/openwrt/luci/blob/038ad57ad85184505509ccc9a9f19847bb3ef56c/applications/luci-app-nes-emulator/root/usr/libexec/rpcd/nes-emulator#L428-L431)`` — now use the `$( exec 7>&- 8>&- 9>&-; … )` form that the other ~20 sites in the file use, instead of a redirection on the function invocation.

The rewrite is behaviour-preserving in the ways that matter: closing an fd that is not open is a no-op, so `exec 7>&- 9>&-` is harmless in `ensure_auth_token` where only fd 8 is held; the substitution's exit status is still `new_token`'s, and a failed `exec` inside the substitution yields an empty `$token`, which the existing `[ "${#token}" -ne 64 ]` guard catches on both paths. `new_token` is still resolvable from the substitution shell, and its own inner `hexdump` substitution keeps its closures. `dash -n` and `sh -n` parse the resulting script cleanly.

Commit message matches the diff, and all three FormalityCheck runs are green on `038ad57a`. `nes-emulator` is still absent from the packages feed, so the init-script side of the shared `/var/run/nes-emulator` lock directory — in particular that it creates it as exactly `0700` root:root, which `prepare_lock_directory` requires on every call — remains unverified here; the draft status and the openwrt/packages#30350 dependency are noted.

---

PR #8964, opened 2026-08-22
luci-theme-openwrt-2020: minor fixes

mcprat:
## Pull request details

1. scrolling issue for syslog (see commit 24fbd754aa9cc470679253dd7cadb5249492022e)
2. more balanced brighter color for checkbox check (visibility for dark mode)

---

## Tested on
**Web browser(s): Win 10 64-bit Chrome 151.0.7922.169**

### DIFF
--- a/applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js
+++ b/applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js
@@ -17,7 +17,7 @@ return view.extend({
 
 					res.file = res.file || '';
 					res.content = 'content' in res ? res.content.trim() : '';
-					res.rows = res.content.split('\n', 20).length;
+					res.rows = res.content.split('\n', 20).length + 1;
 					return res;
 				})
 			);
@@ -35,7 +35,7 @@ return view.extend({
 						'style': 'width: 100%',
 						'readonly': true,
 						'wrap': 'off',
-						'rows': data.rows >= 20 ? 20 : data.rows + 1
+						'rows': data.rows >= 20 ? data.rows + 1 : 20
 					}, data.content)
 				)
 			]);
--- a/modules/luci-base/htdocs/luci-static/resources/tools/views.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/views.js
@@ -140,7 +140,7 @@ var CBILogreadBox = function(logtag, name) {
 			if (element) {
 				const log = await this.retrieveLog();
 				element.value = log?.value;
-				element.rows = log?.rows;
+				element.rows = log?.rows + 1;
 			}
 		},
 
@@ -277,7 +277,7 @@ var CBILogreadBox = function(logtag, name) {
 						'style': 'font-size:12px',
 						'readonly': 'readonly',
 						'wrap': 'off',
-						'rows': loglines?.rows,
+						'rows': loglines?.rows + 1
 					}, [ loglines?.value ]),
 					E('div', {'style': 'padding-bottom: 20px'}, [scrollUpButton])
 				])
--- a/modules/luci-base/htdocs/luci-static/resources/ui.js
+++ b/modules/luci-base/htdocs/luci-static/resources/ui.js
@@ -527,7 +527,7 @@ const UITextarea = UIElement.extend(/** @lends LuCI.ui.Textarea.prototype */ {
 			'placeholder': this.options.placeholder,
 			'style': style,
 			'cols': this.options.cols,
-			'rows': this.options.rows,
+			'rows': this.options.rows ? this.options.rows + 1 : this.options.rows,
 			'wrap': this.options.wrap ? 'soft' : 'off'
 		}, [ value ]));
 
--- a/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css
+++ b/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css
@@ -981,7 +981,7 @@ button[disabled], button.disabled, .btn[disabled], .btn.disabled {
 input[type="checkbox"],
 input[type="radio"] {
 	--bd-color: var(--main-dark-color);
-	--fg-color: var(--main-dark-color);
+	--fg-color: var(--main-bright-color);
 	-webkit-appearance: none; /* nonstandard, should remove in future */
 	appearance: none;
 	width: 1em;
--- a/applications/luci-app-antiblock/htdocs/luci-static/resources/view/antiblock/log.js
+++ b/applications/luci-app-antiblock/htdocs/luci-static/resources/view/antiblock/log.js
@@ -25,7 +25,7 @@ return view.extend({
 		if (element) {
 			const log = await this.retrieveLog();
 			element.value = log.value;
-			element.rows = log.rows;
+			element.rows = log.rows + 1;
 		}
 	},
 
@@ -78,7 +78,7 @@ return view.extend({
 					'style': 'font-size:12px',
 					'readonly': 'readonly',
 					'wrap': 'off',
-					'rows': loglines.rows
+					'rows': loglines.rows + 1
 				}, [loglines.value]),
 				E('div', { 'style': 'padding-top: 20px' }, [scrollUpButton])
 			])
--- a/applications/luci-app-antiblock/htdocs/luci-static/resources/view/antiblock/statistics.js
+++ b/applications/luci-app-antiblock/htdocs/luci-static/resources/view/antiblock/statistics.js
@@ -25,7 +25,7 @@ return view.extend({
         if (element) {
             const log = await this.retrieveLog();
             element.value = log.value;
-            element.rows = log.rows;
+            element.rows = log.rows + 1;
         }
     },
 
@@ -61,7 +61,7 @@ return view.extend({
                     'style': 'font-size:12px',
                     'readonly': 'readonly',
                     'wrap': 'off',
-                    'rows': loglines.rows
+                    'rows': loglines.rows + 1
                 }, [loglines.value])
             ])
         ]);
--- a/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
+++ b/applications/luci-app-package-manager/htdocs/luci-static/resources/view/package-manager.js
@@ -981,7 +981,7 @@ function handleConfig(ev)
 			body.push(E('h5', {}, '%h'.format(file)));
 			body.push(E('textarea', {
 				'name': file,
-				'rows': Math.max(Math.min(L.toArray(conf[file].match(/\n/g)).length, 10), 3)
+				'rows': Math.max(Math.min(L.toArray(conf[file].match(/\n/g)).length, 10), 3) + 1
 			}, '%h'.format(conf[file])));
 		});
 
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
@@ -119,7 +119,7 @@ return view.extend({
 		if (element) {
 			const log = await this.retrieveLog();
 			element.value = log.value;
-			element.rows = log.rows;
+			element.rows = log.rows + 1;
 		}
 	},
 
@@ -284,7 +284,7 @@ return view.extend({
 					'style': 'font-size:12px',
 					'readonly': 'readonly',
 					'wrap': 'off',
-					'rows': loglines.rows
+					'rows': loglines.rows + 1
 				}, [ loglines.value ]),
 				E('div', {'style': 'padding-bottom: 20px'}, [scrollUpButton])
 			])


mcprat:
ping @jow- @hnyman

themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css
```diff
@@ -1106,6 +1106,7 @@ textarea {
 	font-family: monospace;
 	font-size: .9rem;
 	padding: .2rem;
+	overflow-y: hidden;
 }
```
openwrt-ai: This lands in the global `textarea` rule, so it removes the vertical scrollbar and disables wheel scrolling for *every* textarea in LuCI, not just the log views.

The log views are safe because they size `rows` to the content ([`views.js:275-281`](https://github.com/openwrt/luci/blob/500895691b08c46ab9da396a3ea3e8e4f1b0fbbd/modules/luci-base/htdocs/luci-static/resources/tools/views.js#L275-L281)) and therefore never overflow. Fixed-height editors do overflow: the crontab editor is [`rows: 25`](``https://github.com/openwrt/luci/blob/500895691b08c46ab9da396a3ea3e8e4f1b0fbbd/modules/luci-mod-system/htdocs/luci-static/resources/view/system/crontab.js#L30)`` and the rc.local editor is [`rows: 20`](``https://github.com/openwrt/luci/blob/500895691b08c46ab9da396a3ea3e8e4f1b0fbbd/modules/luci-mod-system/htdocs/luci-static/resources/view/system/startup.js#L117).`` With this rule a crontab longer than 25 lines has no scrollbar and cannot be wheel-scrolled — the rest is only reachable by moving the caret with the keyboard.

Two points in favour of scoping it: the precedent this PR cites, [24fbd754](https://github.com/openwrt/luci/commit/24fbd754aa9cc470679253dd7cadb5249492022e), applied the identical fix to `textarea#syslog` and not to bare `textarea`; and the sibling theme deliberately sets [`textarea { overflow: auto; }`](``https://github.com/openwrt/luci/blob/500895691b08c46ab9da396a3ea3e8e4f1b0fbbd/themes/luci-theme-bootstrap/htdocs/luci-static/bootstrap/cascade.css#L258-L259).`` Both the syslog and dmesg views still render `<textarea id="syslog">`, so the 2019 selector still covers the reported case here.

```suggestion
}

textarea#syslog {
	overflow-y: hidden;
}
```

---

themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css
```diff
@@ -981,7 +981,7 @@ button[disabled], button.disabled, .btn[disabled], .btn.disabled {
 input[type="checkbox"],
 input[type="radio"] {
 	--bd-color: var(--main-dark-color);
-	--fg-color: var(--main-dark-color);
+	--fg-color: var(--main-bright-color);
```
openwrt-ai, line 984: This lowers the checked-state contrast in the theme's own (light) rendering. `--fg-color` paints the checkmark glyph via [`background: var(--fg-color)` on `:checked::after` at line 1034](``https://github.com/openwrt/luci/blob/500895691b08c46ab9da396a3ea3e8e4f1b0fbbd/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css#L1034),`` and the page background is white ([`body { background: var(--secondary-bright-color) }`](``https://github.com/openwrt/luci/blob/500895691b08c46ab9da396a3ea3e8e4f1b0fbbd/themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css#L50-L51),`` `#FFFFFF`). `#00B5E2` on white is roughly 2.4:1, under the 3:1 that WCAG 2.1 SC 1.4.11 asks for on non-text UI components; the current `--main-dark-color` (`#002B49`) is roughly 14.6:1.

This theme ships no `prefers-color-scheme` handling, so the only rendering it controls is the light one — and the commit message says the motivation is a third-party extension that inverts the page. Is trading the default rendering for that the intent? A `@media (prefers-color-scheme: dark)` override of `--fg-color`, or brightening only the `::before` box outline and leaving the glyph dark, would fix the dark case without regressing the light one.

---

openwrt-ai (commented):
## Commit checks

- 5008956 "luci-theme-openwrt-2020: use lighter color for checkboxes" — the subject and body talk about checkboxes only, but the changed declaration sits in the shared `input[type="checkbox"], input[type="radio"]` block and `input[type="radio"]:checked::after` also paints with `var(--fg-color)`, so radio buttons are recoloured too. Either mention radios in the message or scope the override to `input[type="checkbox"]`.


---

themes/luci-theme-openwrt-2020/htdocs/luci-static/openwrt2020/cascade.css
mcprat: maybe we should consider changing the color because of WCAG 2.1 SC 1.4.11, but thats for another day...

mcprat:
I see that using `overflow-y` causes other problems elsewhere and got reverted

bumping the "rows" value also fixes the problem while not blocking a scrollbar from appearing in other contexts.

I think this is ready now

modules/luci-base/htdocs/luci-static/resources/tools/views.js
```diff
@@ -140,7 +140,7 @@ var CBILogreadBox = function(logtag, name) {
 			if (element) {
 				const log = await this.retrieveLog();
 				element.value = log?.value;
-				element.rows = log?.rows;
+				element.rows = log?.rows + 1;
```
openwrt-ai, line 143: `retrieveLog()` already returns the padded value — [`rows: loglines?.length + 1` at views.js:126](https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/modules/luci-base/htdocs/luci-static/resources/tools/views.js#L126). So `log?.rows` is *already* `lines + 1`, and this hunk makes the syslog box `lines + 2`. The commit message says "add 1 to the dynamic size", not 2, so this looks unintended — the `+ 1` the message notes "already exists in some instances of similar Javascript files" is this same producer-side one, not a separate concern.

```suggestion
				element.rows = log?.rows;
```

The identical double-apply is in the four other log views, each of which already pads at the producer:

| producer (already `+ 1`) | sites this PR adds a second `+ 1` |
| --- | --- |
| [views.js:126](https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/modules/luci-base/htdocs/luci-static/resources/tools/views.js#L126) | this line, and line 280 |
| [dmesg.js:109](``https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js#L109``) | dmesg.js:122, dmesg.js:287 |
| [antiblock/log.js:16](``https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/applications/luci-app-antiblock/htdocs/luci-static/resources/view/antiblock/log.js#L16)`` | log.js:28, log.js:81 |
| [antiblock/statistics.js:16](``https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/applications/luci-app-antiblock/htdocs/luci-static/resources/view/antiblock/statistics.js#L16)`` | statistics.js:28, statistics.js:64 |

If `lines + 2` really is what fixed the scrolling for you in testing, then the fix belongs in the one producer expression per file rather than in the two consumers, and the commit message should say +2.

---

modules/luci-base/htdocs/luci-static/resources/ui.js
```diff
@@ -527,7 +527,7 @@ const UITextarea = UIElement.extend(/** @lends LuCI.ui.Textarea.prototype */ {
 			'placeholder': this.options.placeholder,
 			'style': style,
 			'cols': this.options.cols,
-			'rows': this.options.rows,
+			'rows': this.options.rows + 1,
```
openwrt-ai: `rows` is optional here and defaults to `null` — [`rows: null` in `__init__` at ui.js:511](https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/modules/luci-base/htdocs/luci-static/resources/ui.js#L511). `null + 1` is `1`, so every `ui.Textarea` constructed without an explicit `rows` now renders a **one-line-tall** textarea. Before this change the `null` was dropped by [`attr[key] == null` in `dom.attr()` at luci.js:1503](https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/modules/luci-base/htdocs/luci-static/resources/luci.js#L1503), so no `rows` attribute was emitted at all and the theme CSS governed the height.

The `form.TextValue` path is hit too but lands differently: [`rows: this.rows` at form.js:5356](https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/modules/luci-base/htdocs/luci-static/resources/form.js#L5356) passes `undefined` for options that never set `.rows`, and `Object.assign` overwrites the `null` default with that `undefined`, so those get `rows="NaN"` — an invalid attribute value the browser discards. Affected in-tree options that never assign `.rows` include [cloudflared config.js:62](``https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/applications/luci-app-cloudflared/htdocs/luci-static/resources/view/cloudflared/config.js#L62)`` and [keepalived peers.js:83](``https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/applications/luci-app-keepalived/htdocs/luci-static/resources/view/keepalived/peers.js#L83),`` plus samba4, ksmbd, ocserv, lldpd, smartdns, clamav and nlbwmon.

Only pad when a row count was actually supplied, so the unset case keeps being omitted:

```suggestion
			'rows': this.options.rows != null ? this.options.rows + 1 : null,
```

---

applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js
```diff
@@ -17,7 +17,7 @@ return view.extend({
 
 					res.file = res.file || '';
 					res.content = 'content' in res ? res.content.trim() : '';
-					res.rows = res.content.split('\n', 20).length;
+					res.rows = res.content.split('\n', 20).length + 1;
```
openwrt-ai, line 20: Padding here misses the case that actually has the bug and over-pads the one that doesn't. The consumer already adds the row itself in its else-branch — [`'rows': data.rows >= 20 ? 20 : data.rows + 1` at files.js:38](``https://github.com/openwrt/luci/blob/32970c53fc94d6659253154812e53cbfee8f2fe0/applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js#L38).``

Because `split('\n', 20)` caps the array at 20 entries, `res.rows` maxes out at 20 before this change:

- **≥ 20 lines** (the case that double-scrolls): `res.rows` becomes 21, still `>= 20`, so the ternary still yields exactly `20` rows for 20 displayed lines. Unchanged — and still buggy.
- **≤ 18 lines**: now `lines + 2` instead of `lines + 1`.
- **19 lines**: `res.rows` becomes 20, which now trips the `>= 20` branch, so the cap boundary moves one line earlier.

It also makes `res.rows` no longer mean "number of lines", which is what line 38 reads it as. Revert this line:

```suggestion
					res.rows = res.content.split('\n', 20).length;
```

and fix the capped branch at line 38 instead — `'rows': Math.min(data.rows, 20) + 1` gives 21 rows for the capped case and keeps `lines + 1` everywhere below it.

---

openwrt-ai (commented):
Re-reviewed after the force-push: the CSS `overflow-y` approach was replaced by a JS `rows + 1` approach, so the first commit is entirely new. The previous commit-check on the checkbox commit is addressed — `32970c5` now documents the radio-button side effect.

## Commit checks

- db70415 "treewide: fix textarea double scrolling issue" — the body says "add 1 to the dynamic size of the `textarea` boxes", but in five of the seven files the value being incremented has already been padded by its producer, so the effective change is +2, not +1. The message also reads the pre-existing `+ 1` as belonging to unrelated files ("this `+ 1` already exists in some instances of similar Javascript files") when it is in fact the same padding, in the same files this commit touches. Either drop the duplicate hunks or state that +2 is intended. Details inline.

---

modules/luci-base/htdocs/luci-static/resources/tools/views.js
mcprat: this is on purpose, to encourage this behavior no matter what example is used for a new application

mcprat:
@hnyman I handled the Claude comments, would you mind taking a look?

applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js
```diff
@@ -35,7 +35,7 @@ return view.extend({
 						'style': 'width: 100%',
 						'readonly': true,
 						'wrap': 'off',
-						'rows': data.rows >= 20 ? 20 : data.rows + 1
+						'rows': data.rows >= 20 ? data.rows + 1 : 20
```
openwrt-ai, line 38: Flipping the ternary makes every short file render a 20-row box. The commit message reads the original as "reversed logic", but the `20` in [`res.content.split('\n', 20)` at line 20](``https://github.com/openwrt/luci/blob/60e8555ff891efb8f826844d174522cc214990d1/applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js#L20)`` is what shows it was a *maximum*: the producer deliberately stops counting at 20 because the consumer never wanted more than 20 rows. Capping the count only makes sense against a cap.

Concretely, with `res.rows` now being `min(lines, 20) + 1`:

- **empty or 1-line `aria2.conf`** — `res.rows` is 2, `2 >= 20` is false, so the box is **20 rows** tall for one line of content. Before this PR it was 2.
- **19 lines** — `res.rows` is 20, so the `>= 20` branch fires one line early and yields 21.
- **20+ lines** — `res.rows` saturates at 21, so `data.rows + 1` is 22 no matter whether the file has 20 lines or 5000. So the "exact number + 1" branch never actually tracks the line count; the cap is still in force, just at 22 now. A genuine "minimum 20, exact count above that" would also have to drop the `20` limit from the `split()`.

Neither effect is mentioned in the commit body, and neither relates to the double-scrolling bug this commit is about. If the intent was to keep the cap and just add the padding row, this restores that and stays consistent with the `+ 1` already applied at line 20:

```suggestion
						'rows': Math.min(data.rows, 21)
```

If a 20-row minimum really is wanted for the aria2 views, that is a separate UI change and worth its own commit rather than riding along in a treewide scrolling fix.

---

openwrt-ai (commented):
Re-reviewed the two changes since 32970c5.

The `ui.js` fix is correct — `this.options.rows ? this.options.rows + 1 : this.options.rows` keeps the unset (`null`/`undefined`) case falling through to `dom.attr()`'s null check, so no `rows` attribute is emitted and the previous one-row / `NaN` regressions are gone.

The double-`+ 1` question on the log views is answered by the updated commit body and the reply on that thread, so I'm not re-raising it. One remaining issue on the aria2 ternary, inline.

Both commit messages match their diffs; the checkbox commit now documents the radio-button side effect.

---

applications/luci-app-aria2/htdocs/luci-static/resources/view/aria2/files.js
mcprat: probably an intermediate commit would be best to change the behavior

DragonBluep:
It seems that checkbox no longer display the check mark after this PR. Environment: windows 11 + edge
image

hnyman:
Thanks for the report.
I tested with my own router installing openwrt-2020 theme with the nornal light mode, and you are right, the checkmark is not visible at all (Win11, Firefox).
Reverting the --main-bright-color change brought the checkmark back.

I will revert the change.

mcprat:
PLEASE do not revert the change 

the problem is the blue background on that single page

mcprat:
can we FIX regressions instead of going BACKWARDS?

hnyman:
I reverted it, as your change hurt the default use case.

If you want it back, please also test your change with the normal light mode. And if necessary, author also the needed changes to the pages.

mcprat:
you still think I didn't test it... funny

the regression affects only those 3 boxes in the picture

PR #8965, opened 2026-08-22
luci-app-ppe-qos: add a page for the PPE hardware QoS classifiers

JuliusBairaktaris:
A LuCI page for `ppe-qos`, which configures the Qualcomm PPE's hardware QoS
classifiers: marking rules that put a DSCP on a flow with the switch's own ACL
engine, a DSCP-to-priority map applied per switch port with `dcb app`, and the
driver's small-packet priority. The page edits all three from one place under
Network.

It depends on `ppe-qos` (`applications/luci-app-ppe-qos/Makefile`,
`LUCI_DEPENDS:=+ppe-qos`), which is openwrt/openwrt#24806 and is not in
openwrt/openwrt yet, so this cannot build or merge before that lands.

The page's class description says a mark only changes the switch queue if the
map turns it into a priority of 4 or more, and that it reaches Wi-Fi either
way because Wi-Fi reads the DSCP for its own access categories. It also says
that shaper rates and the bulk queue depth are not here; they belong to SQM
(`hw_ppe.qos`).

I cross-checked the UCI option names against that backend. The page writes
`small_pkt_len`, `small_pkt_prio` and `wan_port`:

    applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js:
        o = s.option(form.Value, 'small_pkt_len', _('Small packet: maximum L3 length'),
        ...
        o = s.option(form.ListValue, 'small_pkt_prio', _('Small packet: priority'));
        ...
        o = s.option(form.Value, 'wan_port', _('Uplink switch port'),

which is what the init reads (openwrt/openwrt#24806):

    package/network/config/ppe-qos/files/ppe-qos.init, apply():
        config_get len global small_pkt_len 1000
        config_get prio global small_pkt_prio 5

    package/network/config/ppe-qos/files/ppe-qos.init, wan_port():
        config_get dev global wan_port

I installed the page on an AX3600; the JS is served over HTTPS and emits the
option names the init reads. I have not driven the interactive save round
trip; there is no browser on the bench, so the form-to-UCI write path is
unexercised.

### DIFF
--- a/applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
+++ b/applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
@@ -1,0 +1,137 @@
+'use strict';
+'require view';
+'require form';
+
+var CLASSES = [
+	[ 'cs0',  _('CS0 - best effort') ],
+	[ 'cs1',  _('CS1 - scavenger / bulk') ],
+	[ 'cs3',  _('CS3 - broadcast video') ],
+	[ 'af31', _('AF31 - multimedia streaming') ],
+	[ 'cs4',  _('CS4 - real-time interactive (games)') ],
+	[ 'af41', _('AF41 - multimedia conferencing (calls)') ],
+	[ 'cs5',  _('CS5 - signalling, DNS') ],
+	[ 'ef',   _('EF - voice bearer') ],
+	[ 'cs6',  _('CS6 - network control') ]
+];
+
+function prio_options(o) {
+	for (var i = 0; i <= 7; i++)
+		o.value(i, i >= 4 ? _('Priority %d (served past the bulk queue)').format(i)
+				  : _('Priority %d (bulk queue)').format(i));
+}
+
+return view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('ppe-qos', _('PPE Hardware QoS'),
+			_('Which egress queue of the switch a flow rides. A marking rule puts a DSCP on the flow, ' +
+			  'a map turns that DSCP into an internal priority, and the priority picks the queue: ' +
+			  'priorities 4 and up are served past the bulk queue of a shaped port, so marked traffic ' +
+			  'keeps idle latency while the bulk queue stays deep enough for full throughput. ' +
+			  'Marking is done by the switch itself, so it reaches flows the PPE has offloaded. ' +
+			  'Shaper rates and the bulk queue depth belong to SQM (hw_ppe.qos).'));
+
+		s = m.section(form.NamedSection, 'global', 'global', _('Global'));
+
+		o = s.option(form.Flag, 'enabled', _('Enable marking rules'),
+			_('Turns the rules below off without deleting them. The DSCP maps and the small-packet ' +
+			  'priority stay in effect either way.'));
+		o.default = '1';
+		o.rmempty = false;
+
+		o = s.option(form.Value, 'small_packet_len', _('Small packet: maximum L3 length'),
+			_('Frames up to this IP length take the priority below whatever their marking: ' +
+			  'ACKs, handshakes, DNS, VoIP and game traffic jump the bulk queue. 0 disables. ' +
+			  'A value just under the MTU means "everything that is not a full-size bulk frame".'));
+		o.datatype = 'and(uinteger,range(0,1500))';
+		o.placeholder = '128';
+
+		o = s.option(form.ListValue, 'small_packet_prio', _('Small packet: priority'));
+		prio_options(o);
+		o.default = '5';
+
+		o = s.option(form.Value, 'wan_port', _('Uplink switch port'),
+			_('Left empty this is taken from the WAN interface, with any VLAN id stripped, ' +
+			  'since the switch knows only the port. Set it if that guess is wrong.'));
+		o.datatype = 'netdevname';
+		o.placeholder = _('auto');
+		o.optional = true;
+
+		/* Marking rules. */
+		s = m.section(form.GridSection, 'rule', _('Marking rules'),
+			_('Each rule is applied in both directions: as written on the ports your clients sit ' +
+			  'behind, and mirrored on the uplink port so the returning half of the same connection ' +
+			  'is marked too. Ports may be single or a range (5000-5500), and several may be listed. ' +
+			  'The switch matches addresses and ports only - there is no matching by domain name.'));
+		s.addremove = true;
+		s.anonymous = true;
+		s.sortable = true;
+		s.nodescriptions = true;
+
+		o = s.option(form.Flag, 'enabled', _('Enable'));
+		o.editable = true;
+		o.default = '1';
+
+		o = s.option(form.Value, 'name', _('Name'));
+		o.rmempty = false;
+
+		o = s.option(form.ListValue, 'proto', _('Protocol'));
+		o.value('tcpudp', _('TCP and UDP'));
+		o.value('tcp', 'TCP');
+		o.value('udp', 'UDP');
+		o.default = 'tcpudp';
+
+		o = s.option(form.Value, 'src_ip', _('Source address'),
+			_('Optional. Addresses or CIDRs, space separated; IPv4 and IPv6 may be mixed.'));
+		o.datatype = 'list(ipmask)';
+		o.optional = true;
+		o.modalonly = true;
+
+		o = s.option(form.Value, 'dest_ip', _('Destination address'),
+			_('Optional, same syntax as the source address.'));
+		o.datatype = 'list(ipmask)';
+		o.optional = true;
+		o.modalonly = true;
+
+		o = s.option(form.Value, 'src_port', _('Source port'),
+			_('Optional. A port, a range, or several separated by spaces.'));
+		o.datatype = 'list(portrange)';
+		o.optional = true;
+		o.modalonly = true;
+
+		o = s.option(form.Value, 'dest_port', _('Destination port'),
+			_('Optional, same syntax as the source port. This is the usual way to name a service.'));
+		o.datatype = 'list(portrange)';
+		o.optional = true;
+
+		o = s.option(form.ListValue, 'class', _('Class'),
+			_('A class only changes the queue if the map below turns it into a priority of 4 or more. ' +
+			  'It also reaches Wi-Fi, which reads the DSCP for its own access categories.'));
+		CLASSES.forEach(function(c) { o.value(c[0], c[1]); });
+		o.rmempty = false;
+
+		/* DSCP to priority. */
+		s = m.section(form.GridSection, 'dscp', _('DSCP to priority'),
+			_('What a mark is worth once the switch reads it. Applied with dcb app on every switch ' +
+			  'port, so it classifies traffic marked by these rules and traffic that arrives already ' +
+			  'marked alike. Anything unlisted stays at priority 0 and rides the bulk queue.'));
+		s.addremove = true;
+		s.anonymous = true;
+
+		o = s.option(form.Value, 'dscp', _('DSCP'));
+		o.datatype = 'and(uinteger,range(0,63))';
+		o.rmempty = false;
+		o.value('32', _('32 (CS4 - games)'));
+		o.value('34', _('34 (AF41 - calls)'));
+		o.value('40', _('40 (CS5 - signalling, DNS)'));
+		o.value('46', _('46 (EF - voice)'));
+		o.value('48', _('48 (CS6 - control)'));
+
+		o = s.option(form.ListValue, 'prio', _('Priority'));
+		prio_options(o);
+		o.rmempty = false;
+
+		return m.render();
+	}
+});
--- a/applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
+++ b/applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
@@ -1,0 +1,141 @@
+'use strict';
+'require view';
+'require form';
+
+var CLASSES = [
+	[ 'cs0',  _('CS0 - best effort') ],
+	[ 'cs1',  _('CS1 - scavenger / bulk') ],
+	[ 'cs3',  _('CS3 - broadcast video') ],
+	[ 'af31', _('AF31 - multimedia streaming') ],
+	[ 'cs4',  _('CS4 - real-time interactive (games)') ],
+	[ 'af41', _('AF41 - multimedia conferencing (calls)') ],
+	[ 'cs5',  _('CS5 - signalling, DNS') ],
+	[ 'ef',   _('EF - voice bearer') ],
+	[ 'cs6',  _('CS6 - network control') ]
+];
+
+function prio_options(o) {
+	for (var i = 0; i <= 7; i++)
+		o.value(i, i >= 4 ? _('Priority %d (served past the bulk queue)').format(i)
+				  : _('Priority %d (bulk queue)').format(i));
+}
+
+return view.extend({
+	render: function() {
+		var m, s, o;
+
+		m = new form.Map('ppe-qos', _('PPE Hardware QoS'),
+			_('Which egress queue of the switch a flow rides. A marking rule puts a DSCP on the flow, ' +
+			  'a map turns that DSCP into an internal priority, and the priority picks the queue: ' +
+			  'priorities 4 and up are served past the bulk queue of a shaped port, so marked traffic ' +
+			  'keeps idle latency while the bulk queue stays deep enough for full throughput. ' +
+			  'Marking is done by the switch itself, so it reaches flows the PPE has offloaded. ' +
+			  'Shaper rates and the bulk queue depth belong to SQM (hw_ppe.qos).'));
+
+		s = m.section(form.NamedSection, 'global', 'global', _('Global'));
+
+		o = s.option(form.Flag, 'enabled', _('Enable'),
+			_('Turns the whole configuration off without deleting it. The marking rules, the DSCP ' +
+			  'maps and the small-packet priority are all removed while this is off.'));
+		o.default = '1';
+		o.rmempty = false;
+
+		o = s.option(form.Value, 'small_pkt_len', _('Small packet: maximum L3 length'),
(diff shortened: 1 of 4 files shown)


applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,133 @@
...
+		o = s.option(form.ListValue, 'small_packet_prio', _('Small packet: priority'));
+		for (var i = 0; i <= 7; i++)
+			o.value(i, _('Priority %d').format(i));
+		o.default = '5';
+
+		o = s.option(form.Value, 'wan_port', _('Uplink switch port'),
+			_('Left empty this is taken from the WAN interface, with any VLAN id stripped, ' +
+			  'since the switch knows only the port. Set it if that guess is wrong.'));
+		o.placeholder = _('auto');
+		o.optional = true;
```
openwrt-ai, line 60: `wan_port` is the only option on the page with no `datatype`, so it accepts arbitrary free text — spaces, `;`, `$(...)` — which is then written to UCI and consumed by the `ppe-qos` shell/init code that runs `dcb app dev <port>`. Since the description says the value is derived from the WAN interface's device with the VLAN id stripped, [`netdevname`](https://github.com/openwrt/luci/blob/10332fbc87ddbc507bd500028fff7db6718611ea/modules/luci-base/htdocs/luci-static/resources/validation.js#L1000) is the matching validator (1–15 chars, no `:`, `/`, `%` or whitespace):

```suggestion
		o.placeholder = _('auto');
		o.datatype = 'netdevname';
		o.optional = true;
```

If the backend actually expects a numeric switch port index rather than a device name, use `uinteger` instead — but it should not stay unvalidated either way.

---

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,133 @@
...
+			_('Turns the rules below off without deleting them. The DSCP maps and the small-packet ' +
+			  'priority stay in effect either way.'));
+		o.default = '1';
+		o.rmempty = false;
+
+		o = s.option(form.Value, 'small_packet_len', _('Small packet: maximum L3 length'),
+			_('Frames up to this IP length take the priority below whatever their marking: ' +
+			  'ACKs, handshakes, DNS, VoIP and game traffic jump the bulk queue. 0 disables. ' +
+			  'A value just under the MTU means "everything that is not a full-size bulk frame".'));
+		o.datatype = 'range(0,1500)';
```
openwrt-ai: `range()` validates a *decimal*, not an integer — [`range(min, max)`](https://github.com/openwrt/luci/blob/10332fbc87ddbc507bd500028fff7db6718611ea/modules/luci-base/htdocs/luci-static/resources/validation.js#L1021) calls `parseDecimal()`, so `128.5` passes here and gets written to UCI as a packet length. The sister QoS app uses the `and(uinteger,…)` idiom for exactly this, e.g. [`sqm.js:98`](``https://github.com/openwrt/luci/blob/10332fbc87ddbc507bd500028fff7db6718611ea/applications/luci-app-sqm/htdocs/luci-static/resources/view/network/sqm.js#L98).``

```suggestion
		o.datatype = 'and(uinteger,range(0,1500))';
```

Same applies to the `dscp` value on line 117 (`range(0,63)` accepts `46.5`).

---

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,133 @@
...
+		s.anonymous = true;
+
+		o = s.option(form.Value, 'dscp', _('DSCP'));
+		o.datatype = 'range(0,63)';
+		o.rmempty = false;
+		o.value('46', _('46 (EF - voice)'));
+		o.value('40', _('40 (CS5 - signalling, DNS)'));
+		o.value('34', _('34 (AF41 - calls)'));
+		o.value('32', _('32 (CS4 - games)'));
+		o.value('48', _('48 (CS6 - control)'));
```
openwrt-ai: nit: the suggestion list runs 46, 40, 34, 32, 48 — descending except for CS6, which is appended out of order. The `CLASSES` array on lines 6-14 that feeds the marking rules is ordered ascending by class, so the two dropdowns present the same DSCP scale in opposite/inconsistent orders. Ascending matches `CLASSES`:

```suggestion
		o.value('32', _('32 (CS4 - games)'));
		o.value('34', _('34 (AF41 - calls)'));
		o.value('40', _('40 (CS5 - signalling, DNS)'));
		o.value('46', _('46 (EF - voice)'));
		o.value('48', _('48 (CS6 - control)'));
```

---

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,133 @@
...
+		o = s.option(form.Value, 'small_packet_len', _('Small packet: maximum L3 length'),
+			_('Frames up to this IP length take the priority below whatever their marking: ' +
+			  'ACKs, handshakes, DNS, VoIP and game traffic jump the bulk queue. 0 disables. ' +
+			  'A value just under the MTU means "everything that is not a full-size bulk frame".'));
+		o.datatype = 'range(0,1500)';
+		o.placeholder = '128';
+
+		o = s.option(form.ListValue, 'small_packet_prio', _('Small packet: priority'));
+		for (var i = 0; i <= 7; i++)
+			o.value(i, _('Priority %d').format(i));
```
openwrt-ai: nit: this is the same 0-7 priority scale as the `prio` selector on lines 126-128, but only that one tells the user where the 4-and-up cutoff is. Here the default is 5 with no hint why, and the page's whole premise is that "4 or more" is the meaningful threshold. Labelling both identically makes the connection visible:

```suggestion
		for (var i = 0; i <= 7; i++)
			o.value(i, i >= 4 ? _('Priority %d (served past the bulk queue)').format(i)
					  : _('Priority %d (bulk queue)').format(i));
```

---

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,133 @@
...
+
+		o = s.option(form.ListValue, 'proto', _('Protocol'));
+		o.value('tcpudp', _('TCP and UDP'));
+		o.value('tcp', 'TCP');
+		o.value('udp', 'UDP');
+		o.default = 'tcpudp';
+
+		o = s.option(form.Value, 'src_ip', _('Source address'),
+			_('Optional. Addresses or CIDRs, space separated; IPv4 and IPv6 may be mixed.'));
+		o.datatype = 'list(neg(ipmask))';
```
openwrt-ai: Does the ppe-qos ACL backend actually support negated matches? [`neg()`](https://github.com/openwrt/luci/blob/10332fbc87ddbc507bd500028fff7db6718611ea/modules/luci-base/htdocs/luci-static/resources/validation.js#L1177) strips a leading `!` and accepts the value, so this field (and `dest_ip` on line 87, `src_port` on line 93, `dest_port` on line 99) will happily take `!192.168.1.0/24`. None of the four descriptions mention negation, and the section intro says "The switch matches addresses and ports only". The datatype is copied from luci-app-firewall, where nftables does support inversion; a switch ACL engine typically does not. If ppe-qos cannot invert, dropping `neg(` here would stop the UI accepting rules the backend will silently mis-handle or reject.

---

openwrt-ai (commented):
Reviewed 1 new commit; the commit message matches what the diff does. Comments inline.

The `ppe-qos` backend is not in `openwrt/openwrt` yet, so the UCI option names (`small_packet_len`, `small_packet_prio`, `wan_port`, `class`, `dscp`, `prio`) could not be cross-checked against a consumer — the dependency on openwrt/openwrt#24806 is already noted in the PR body.

---

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,137 @@
...
+		o = s.option(form.Value, 'small_packet_len', _('Small packet: maximum L3 length'),
+			_('Frames up to this IP length take the priority below whatever their marking: ' +
+			  'ACKs, handshakes, DNS, VoIP and game traffic jump the bulk queue. 0 disables. ' +
+			  'A value just under the MTU means "everything that is not a full-size bulk frame".'));
+		o.datatype = 'and(uinteger,range(0,1500))';
+		o.placeholder = '128';
+
+		o = s.option(form.ListValue, 'small_packet_prio', _('Small packet: priority'));
+		prio_options(o);
+		o.default = '5';
```
openwrt-ai, line 53: Selecting Priority 5 here deletes `small_packet_prio` from UCI rather than writing it: [form.js:2152](https://github.com/openwrt/luci/blob/fee8122417a49e9b99fa514e22e468837be74121/modules/luci-base/htdocs/luci-static/resources/form.js#L2152) removes an option when `fval == this.default` and `rmempty` is still at its default of true. So the device then runs on whatever `ppe-qos` compiles in as its own default, while the page keeps showing 5.

That is only harmless if the backend default is also 5. Every other option that carries a meaningful value is pinned — the `enabled` flag right above on line 41 has the same `default = '1'` plus an explicit `o.rmempty = false`, and `class`, `dscp` and `prio` all set it too. `small_packet_prio` is the one option with a default that is not pinned; is that deliberate, or should it get `o.rmempty = false` like its neighbour?

---

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,137 @@
...
+		s.anonymous = true;
+
+		o = s.option(form.Value, 'dscp', _('DSCP'));
+		o.datatype = 'and(uinteger,range(0,63))';
+		o.rmempty = false;
+		o.value('32', _('32 (CS4 - games)'));
+		o.value('34', _('34 (AF41 - calls)'));
+		o.value('40', _('40 (CS5 - signalling, DNS)'));
+		o.value('46', _('46 (EF - voice)'));
+		o.value('48', _('48 (CS6 - control)'));
```
openwrt-ai, line 133: nit: AF31 is offered as a marking class (`CLASSES`, line 9) but DSCP 26 is missing from these suggestions, so a user who marks a flow "multimedia streaming" finds nothing to select here and has to know that AF31 means 26 and type it. The other three unlisted classes are self-explanatory — CS0/CS1/CS3 are exactly the ones the section text means by "anything unlisted stays at priority 0" — which makes AF31 look like an oversight rather than the same policy.

The field is a `form.Value` with choices, so typing 26 does work; this is only about the suggestion list matching the classes the page itself offers:

```suggestion
		o.value('26', _('26 (AF31 - streaming)'));
		o.value('32', _('32 (CS4 - games)'));
		o.value('34', _('34 (AF41 - calls)'));
		o.value('40', _('40 (CS5 - signalling, DNS)'));
		o.value('46', _('46 (EF - voice)'));
		o.value('48', _('48 (CS6 - control)'));
```

---

openwrt-ai (commented):
Re-reviewed after the force-push (`10332fb` → `fee8122`). All five points from the previous round are addressed: `wan_port` now has `netdevname`, both bare `range()` datatypes became `and(uinteger,range(...))` (idiom confirmed in-tree at [routes.js:93](``https://github.com/openwrt/luci/blob/fee8122417a49e9b99fa514e22e468837be74121/modules/luci-mod-network/htdocs/luci-static/resources/view/network/routes.js#L93)),`` the DSCP suggestions are in ascending order, the two priority selectors share the new `prio_options()` helper, and `neg()` is gone from the four address/port datatypes. The amended commit message still matches what the diff does, and the added `PKG_MAINTAINER`/`PKG_LICENSE` follow the placement used by `luci-app-sqm`.

Two remaining points inline, both minor.

The `ppe-qos` backend is still absent from `openwrt/openwrt` master, so the UCI option names (`small_packet_len`, `small_packet_prio`, `wan_port`, `class`, `dscp`, `prio`) and the accepted `class` keywords remain uncheckable against a consumer — the dependency on openwrt/openwrt#24806 is noted in the PR body.

---

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
JuliusBairaktaris: > `small_packet_prio` is the one option with a default that is not pinned; is that deliberate, or should it get `o.rmempty = false` like its neighbour?

No pin needed for that one. The backend default is 5 at every layer it could fall through to, so an option dropped from UCI lands on the same value the page shows (all three in openwrt/openwrt#24806):

    package/network/config/ppe-qos/files/ppe-qos.config:
        option small_pkt_prio '5'

    package/network/config/ppe-qos/files/ppe-qos.init, apply():
        config_get prio global small_pkt_prio 5

    target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_acl.c:
        static ushort ppe_small_pkt_prio = 5;

The mechanism you describe does bite one option, though not the one flagged: the per-rule `enabled` flag. The backend's per-rule default is the opposite of the global one:

    package/network/config/ppe-qos/files/ppe-qos.init, apply_rule():
        config_get_bool enabled "$cfg" enabled 0
        [ "$enabled" = 1 ] || return 0

So with `rmempty` at its default, `form.js` dropped `enabled` from every rule saved with the box ticked and the init then skipped all of them. `o.rmempty = false` added there.

Checking this did turn up a worse bug than either pin: the page wrote `small_packet_len` and `small_packet_prio`, and the init reads `small_pkt_len` and `small_pkt_prio`. Neither option ever reached the backend, so the priority selector was inert whether or not it was written. Both are renamed now, and the length placeholder is corrected to 1000, which is what the config ships.

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
JuliusBairaktaris: > AF31 is offered as a marking class (`CLASSES`, line 9) but DSCP 26 is missing from these suggestions

AF31 is deliberately outside the switch map the backend ships, so listing 26 among the suggestions would advertise a priority the device does not give it. The suggestions mirror the `config dscp` sections, which are 46, 40, 32, 34 and 48:

    package/network/config/ppe-qos/files/ppe-qos.config (openwrt/openwrt#24806):
        config dscp
            option dscp '34'	# af41 - multimedia conferencing (calls)
            option prio '4'

The class is still worth offering, because a mark reaches the air as well as the switch queue, and mac80211 puts DSCP 26 in the Video access category:

    net/wireless/util.c, cfg80211_classify8021d():
        case 26:
        case 28:
        case 30:
                /* Multimedia Streaming: AF31, AF32, AF33 */
                ret = 4;

`ieee802_1d_to_ac[4]` in `net/mac80211/wme.c` is `IEEE80211_AC_VI`.

The field is a `form.Value` with choices, so typing 26 works for anyone who does add a map entry for it. I would rather leave the list showing what the device actually has configured, so I am leaving this one as it is.

JuliusBairaktaris:
Cross-checking the page against the `ppe-qos` init turned up two option names that do not match: the page wrote `small_packet_len` and `small_packet_prio`, the init reads `small_pkt_len` and `small_pkt_prio`, so neither ever reached the backend. Both renamed, and the length placeholder corrected from 128 to 1000, which is what the shipped config sets.

The per-rule `enabled` flag had `default = '1'` with no `rmempty = false`, so `form.js` removed the option on save while the init's per-rule default is `config_get_bool enabled "$cfg" enabled 0` - every rule saved from the page would have been skipped. Pinned.

Three descriptions were wrong about what the init does. Marking rules apply in both directions only when the global `mark_upload` option is set, which the shipped config leaves off. Turning the global option off runs `teardown`, which flushes `dcb app` on every port and writes 0 to `small_pkt_len`, so neither the DSCP maps nor the small-packet priority stay in effect. The small-packet length has to stay under a full-size segment rather than just under the MTU, because a bulk segment that qualified would ride the priority band. All three corrected, and the global option relabelled "Enable".

Checked on an AX3600 (IPQ8074). With `option small_pkt_len '1000'` and `option small_pkt_prio '5'` in the config, `/sys/module/qca_ppe/parameters/` carries `small_pkt_len` 1000 and `small_pkt_prio` 5. The fixed page, installed at `/www/luci-static/resources/view/network/ppe-qos.js` and served over HTTPS, emits `class`, `dest_ip`, `dest_port`, `dscp`, `enabled`, `name`, `prio`, `proto`, `small_pkt_len`, `small_pkt_prio`, `src_ip`, `src_port` and `wan_port`, matching the three global names the init reads. I have not driven the interactive save round trip; there is no browser on the bench.

I answered the `small_packet_prio` pin and the DSCP 26 suggestion in their threads.

applications/luci-app-ppe-qos/htdocs/luci-static/resources/view/network/ppe-qos.js
```diff
@@ -0,0 +1,141 @@
...
+		o.datatype = 'netdevname';
+		o.placeholder = _('auto');
+		o.optional = true;
+
+		/* Marking rules. */
+		s = m.section(form.GridSection, 'rule', _('Marking rules'),
+			_('Each rule is installed on the uplink port, matched as the returning half of the ' +
+			  'connection with source and destination swapped, so the download direction is marked. ' +
+			  'The upload half is installed on the ports your clients sit behind only when the ' +
+			  'mark_upload option of the ppe-qos config is set, which it is not by default. ' +
```
openwrt-ai, line 67: nit: `mark_upload` sits in the same `config global` section as `enabled`, `small_pkt_len` and `wan_port`, but it is the only one the page names without offering — turning upload marking on means hand-editing `/etc/config/ppe-qos`, which cuts against the page editing the whole config from one place. Should it get a `form.Flag` in the Global section?

---

openwrt-ai (commented):
One inline nit.

---

PR #8969, opened 2026-08-23
luci-proto-bfd: add BFD protocol handler

Wolf480pl:
### Description
Add luci-proto-bfd - a BFD protocol handler, so that users can configure bfdd peers through luci.
This depends on openwrt/packages#30364

Screenshot 2026-08-23 at 16-38-47 integra - LuCI
Screenshot 2026-08-23 at 15-57-27 integra - LuCI

### Maintainer
Not sure... I'm adding myself (@wolf480pl) as the maintainer of the luci-protocol-bfd package, but @lucize maintains the bfdd package on openwrt/packages side

---

## Tested on
**OpenWrt version:** OpenWrt 24.10.2 r28739-d9340319c6
**LuCI version:** openwrt-24.10 branch (25.340.26705~d88390b)
**Web browser(s):** Firefox 145.0.2 (64-bit)

(Once I get a separate router for testing, I'll try testing on 25.12 and master as well)

---

## Checklist
- ~~_(Nice to have)_ Includes what Issue it closes (e.g. openwrt/luci#issue-number).~~
- [x] _(Nice to have)_ Includes what it depends on (e.g. openwrt/packages#pr-number in sister repo).

### DIFF
--- a/protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
+++ b/protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
@@ -1,0 +1,82 @@
+'use strict';
+'require form';
+'require network';
+'require tools.widgets as widgets';
+
+return network.registerProtocol('bfd', {
+	getI18n: function() {
+		return _('Bidirectional Forwarding Detection (BFD) Session');
+	},
+
+	getPackageName: function() {
+		return 'proto-bfd';
+	},
+
+	renderFormOptions: function(s) {
+		var o;
+		var proto = this;
+
+		// -- general ---------------------------------------------------------------------
+
+		o = s.taboption('general', form.Value, 'peer_address', _("Remote IP address"), _("The IP address of the remote BFD peer."));
+		o.rmempty = false;
+		o.datatype = 'or(ip4addr("nomask"),ip6addr("nomask"))';
+
+		o = s.taboption('general', form.Flag, 'multihop', _("Multihop"), _("Allow BFD packats to traverse multiple hops. Use when BFD peer is not a direct neighbour of this router."));
+
+		o = s.taboption('general', form.Value, 'local_address', _("Local IPv4 address"), _("The local IP address used to communicate with the BFD peer (mandatory on multihop)."));
+		o.rmempty = false;
+		o.datatype = 'or(ip4addr("nomask"),ip6addr("nomask"))';
+		o.depends('multihop', '1');
+		o.load = function(section_id) {
+			return network.getDevices().then(L.bind(function(devs) {
+				var addrs = devs.flatMap(d => {
+					var linkName = d.getName();
+					function addLinkName(addr) {
+						if (addr.startsWith("fe8")) {
+							addr = addr + "%" + linkName;
+						}
+						return addr;
+					}
+					var addrs4 = d.getIPAddrs().map(a => a.split('/')[0]);
+					var addrs6 = d.getIP6Addrs().map(a => a.split('/')[0]).map(addLinkName);
+					return addrs4.concat(addrs6)
+				}).sort()
+
+				for (var i = 0; i < addrs.length; i++) {
+					this.value(addrs[i]);
+				}
+				return form.Value.prototype.load.apply(this, [section_id]);
+			}, this));
+		};
+
+		o = s.taboption('general', form.Value, 'vrf_name', _("VRF Name"), _("Optional. The VRF used to communicate with the BFD peer."));
+		o.optional = true;
+		o.datatype = 'string';
+		o.depends('multihop', '1');
+
+		// -- advanced ---------------------------------------------------------------------
+
+		o = s.taboption('advanced', form.Value, 'detect_multiplier', _("Detect Multiplier"), _("Number of consecutive missed BFD packets after which the session is considered down."));
+		o.optional = true;
+		o.placeholder = 3;
+		o.datatype = 'uinteger';
+
+		o = s.taboption('advanced', form.Value, 'receive_interval', _("Receive Interval (ms)"), _("Minimum number of milliseconds between BFD packets we're willing to receive from the peer."));
+		o.optional = true;
+		o.placeholder = 300;
+		o.datatype = 'uinteger';
+
+		o = s.taboption('advanced', form.Value, 'transmit_interval', _("Transmit Interval (ms)"), _("Minimum number of milliseconds between BFD packet we transmit."));
+		o.optional = true;
+		o.placeholder = 300;
+		o.datatype = 'uinteger';
+
+		o = s.taboption('advanced', form.Flag, 'echo_mode', _("Echo Mode"), _("Echo mode sends packets with ourselves as destination, hoping the peer will reflect them without noticing."));
+
+		o = s.taboption('advanced', form.Value, 'echo_interval', _("Echo Interval (ms)"), _("Minimum number of milliseconds between BFD Echo packets, both for transmitting and receiving/reflecting. Set to zero to tell our peer we don't want incoming echos."));
+		o.optional = true;
+		o.placeholder = 50;
+		o.datatype = 'uinteger';
+	}
+});
--- a/protocols/luci-proto-bfd/Makefile
+++ b/protocols/luci-proto-bfd/Makefile
@@ -1,0 +1,26 @@
+# Copyright 2026 Wolf480pl <wolf480@interia.pl>
+#
+# Modified by Wolf480pl <wolf480@interia.pl>,
+# based on luci-proto-gre, which was:
+#
+#   Based on luci-proto-ipip.
+#   Credited author of luci-proto-ipip is Roger Pueyo Centelles <roger.pueyo@guifi.net>
+#   Copyright 2016 Roger Pueyo Centelles <roger.pueyo@guifi.net>
+#
+#   Modified by Jan Betik <jan.betik@svine.su>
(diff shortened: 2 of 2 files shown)


Wolf480pl:
> Author name format is invalid ('Wolf480pl'). Please set your full name (first and last, e.g. 'Jane Doe').

Unlike in openwrt/packages, here the pull request template explicitly says 
> - Your S.O.B. *may* be a nickname.

Since S.O.B. must match Author name, I thought this means commit author also may be a nickname. Is that not the case?

This is a nickname I've been using on the internet for 10+ years, and contributed to other projects using it. Not some random account created yesterday.

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
```diff
@@ -0,0 +1,82 @@
...
+'require network';
+'require tools.widgets as widgets';
+
+return network.registerProtocol('bfd', {
+	getI18n: function() {
+		return _('Bidirectional Forwarding Detection (BFD) Session');
+	},
+
+	getPackageName: function() {
+		return 'proto-bfd';
```
openwrt-ai, line 12: There is no `proto-bfd` package. `getPackageName()` is documented as "the name of the package to download, required for the protocol to function" ([`network.js:2464-2477`](``https://github.com/openwrt/luci/blob/99b2a01da28033bd21249749a76636fcf7e2bc8d/modules/luci-base/htdocs/luci-static/resources/network.js#L2464-L2477)),`` and in openwrt/packages#30364 the netifd handler (`/lib/netifd/proto/bfd.sh`) is installed by the `bfdd` package itself — which is also what `LUCI_DEPENDS:=+bfdd` in the Makefile refers to.

```suggestion
		return 'bfdd';
```

---

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
```diff
@@ -0,0 +1,82 @@
...
+		o.load = function(section_id) {
+			return network.getDevices().then(L.bind(function(devs) {
+				var addrs = devs.flatMap(d => {
+					var linkName = d.getName();
+					function addLinkName(addr) {
+						if (addr.startsWith("fe8")) {
+							addr = addr + "%" + linkName;
+						}
+						return addr;
+					}
```
openwrt-ai, line 40: The zone-suffixed values this produces cannot pass the option's own `datatype`. `ip6addr("nomask")` matches against `/^([0-9a-fA-F:.]+)$/` ([`validation.js:533-538`](``https://github.com/openwrt/luci/blob/99b2a01da28033bd21249749a76636fcf7e2bc8d/modules/luci-base/htdocs/luci-static/resources/validation.js#L533-L538)),`` so `%` is rejected — picking any link-local entry from this dropdown fails validation with "valid IP address" and the form cannot be saved.

Either drop the `%<link>` suffix here, or replace the `datatype` on line 29 with a custom `validate()` that strips the zone before delegating to the address check.

Separately, the prefix test is too narrow: link-local is `fe80::/10` (`fe8`, `fe9`, `fea`, `feb`), so `startsWith("fe8")` only covers `fe80::/16`.

---

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
```diff
@@ -0,0 +1,82 @@
...
+
+		// -- general ---------------------------------------------------------------------
+
+		o = s.taboption('general', form.Value, 'peer_address', _("Remote IP address"), _("The IP address of the remote BFD peer."));
+		o.rmempty = false;
+		o.datatype = 'or(ip4addr("nomask"),ip6addr("nomask"))';
+
+		o = s.taboption('general', form.Flag, 'multihop', _("Multihop"), _("Allow BFD packats to traverse multiple hops. Use when BFD peer is not a direct neighbour of this router."));
+
+		o = s.taboption('general', form.Value, 'local_address', _("Local IPv4 address"), _("The local IP address used to communicate with the BFD peer (mandatory on multihop)."));
```
openwrt-ai, line 27: The label says IPv4 but the `datatype` on line 29 accepts IPv6 as well, and the `load()` below populates the dropdown with `getIP6Addrs()` too. Looks like a leftover from luci-proto-gre, which is IPv4-only.

```suggestion
		o = s.taboption('general', form.Value, 'local_address', _("Local IP address"), _("The local IP address used to communicate with the BFD peer (mandatory on multihop)."));
```

---

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
```diff
@@ -0,0 +1,82 @@
...
+		o.rmempty = false;
+		o.datatype = 'or(ip4addr("nomask"),ip6addr("nomask"))';
+		o.depends('multihop', '1');
```
openwrt-ai, line 30: Gating on `multihop` makes `local_address` unreachable for single-hop sessions, but the backend does not treat it that way: `bfd.sh` in openwrt/packages#30364 passes `${local_address:+-l "$local_address"}` unconditionally, and bfdd's own `bfdd.template.json` documents `local-address` as `"mandatory on multihop"` — i.e. permitted, just not required, otherwise. The option's own help text says the same thing. Is hiding it for single-hop intentional? As written a user with several addresses on the bound device has no way to pin the source address for a single-hop session.

Same question for `vrf_name` on line 56 — `bfd.sh` emits `vrf-name` regardless of `multihop`.

---

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
```diff
@@ -0,0 +1,82 @@
...
+		var o;
+		var proto = this;
+
+		// -- general ---------------------------------------------------------------------
+
+		o = s.taboption('general', form.Value, 'peer_address', _("Remote IP address"), _("The IP address of the remote BFD peer."));
+		o.rmempty = false;
+		o.datatype = 'or(ip4addr("nomask"),ip6addr("nomask"))';
+
+		o = s.taboption('general', form.Flag, 'multihop', _("Multihop"), _("Allow BFD packats to traverse multiple hops. Use when BFD peer is not a direct neighbour of this router."));
```
openwrt-ai, line 25: nit: typo "packats" — this string goes into the translation catalogs, so it is worth fixing before merge.

```suggestion
		o = s.taboption('general', form.Flag, 'multihop', _("Multihop"), _("Allow BFD packets to traverse multiple hops. Use when BFD peer is not a direct neighbour of this router."));
```

Line 70 has a related one: "Minimum number of milliseconds between BFD packet we transmit" should be "packets".

---

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
```diff
@@ -0,0 +1,82 @@
...
+		return _('Bidirectional Forwarding Detection (BFD) Session');
+	},
+
+	getPackageName: function() {
+		return 'proto-bfd';
+	},
+
+	renderFormOptions: function(s) {
+		var o;
+		var proto = this;
```
openwrt-ai, line 17: nit: `proto` is never used in this function.

```suggestion
```

Same for the `tools.widgets as widgets` import on line 4 — no widget from it is referenced (unlike luci-proto-gre, which uses `widgets.NetworkSelect`).

---

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
```diff
@@ -0,0 +1,82 @@
...
+					function addLinkName(addr) {
+						if (addr.startsWith("fe8")) {
+							addr = addr + "%" + linkName;
+						}
+						return addr;
+					}
+					var addrs4 = d.getIPAddrs().map(a => a.split('/')[0]);
+					var addrs6 = d.getIP6Addrs().map(a => a.split('/')[0]).map(addLinkName);
+					return addrs4.concat(addrs6)
+				}).sort()
```
openwrt-ai, line 44: nit: missing statement terminators; every other statement in this file is semicolon-terminated.

```suggestion
					return addrs4.concat(addrs6);
				}).sort();
```

---

openwrt-ai (commented):
Reviewed 1 new commit; 7 inline comments. Backend coupling checks out: every UCI option here (`peer_address`, `local_address`, `multihop`, `vrf_name`, `detect_multiplier`, `receive_interval`, `transmit_interval`, `echo_mode`, `echo_interval`) is read by `proto_bfd_init_config`/`proto_bfd_setup` in `net/bfdd/files/bfd.sh` from openwrt/packages#30364, and that dependency is called out in the PR description as required.

The main functional issue is the link-local zone suffix added in `local_address`'s `load()`, which the option's own datatype rejects.

---

protocols/luci-proto-bfd/htdocs/luci-static/resources/protocol/bfd.js
Wolf480pl: bfdd itself ignores `vrf_name` and `local_address` when multihop is is disabled https://github.com/rzalamena/bfdd/blob/921bfa05b0e0020eb910cffc2a6edf37543b84cf/bfd.c#L775-L792

It's an inconsistence between bfdd's documentation and implementation, though I guess I should pick one interpretation and stick to it instead of carrying the inconsistency into openwrt.

PR #8970, opened 2026-08-23
luci-base,luci-mod-status: make log text search case-insensitive

### LINKED ISSUE #8952: Case sensitivity of log text search in System Log and Kernel Log
The text search in both the System Log and Kernel Log views is currently case-sensitive.

In particular, was the current case-sensitive behavior intentional, or would case-insensitive text search be preferable?

If case-insensitive search is considered appropriate, should it be applied to both "CBILogreadBox" and the Kernel Log implementation in "dmesg.js" for consistent behavior?

@systemcrash @karlp - I would appreciate your thoughts on this.

Image

Ser9ei:
### Description

Enable case-insensitive text search in the System Log and Kernel Log views.
Closes: https://github.com/openwrt/luci/issues/8952

### Maintainer
@systemcrash, @jow-

---

## Tested on
**LuCI version:** LuCI openwrt-25.12 branch (26.209.73834~b61f907)
**Web browser(s):** Firefox 153.0.4

---

## Checklist

### DIFF
--- a/modules/luci-base/htdocs/luci-static/resources/tools/views.js
+++ b/modules/luci-base/htdocs/luci-static/resources/tools/views.js
@@ -116,10 +116,13 @@ var CBILogreadBox = function(logtag, name) {
 					return line.toLowerCase().includes(this.logTagFilter?.toLowerCase());
 				});
 
-				loglines = loglines.filter(line => {
-					const match = line.includes(this.logTextFilter);
-					return this.invertLogTextSearch ? !match : match;
-				});
+				if (this.logTextFilter) {
+					const filter = this.logTextFilter.toLowerCase();
+					loglines = loglines.filter(line => {
+						const match = line.toLowerCase().includes(filter);
+						return this.invertLogTextSearch ? !match : match;
+					});
+				}
 
 				return {
 					value: loglines?.join('\n'),
--- a/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
+++ b/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js
@@ -95,8 +95,9 @@ return view.extend({
 
 			// Filter by text
 			if (this.logTextFilter) {
+				const filter = this.logTextFilter.toLowerCase();
 				loglines = loglines.filter(({ text }) => {
-					const match = text.includes(this.logTextFilter);
+					const match = text.toLowerCase().includes(filter);
 					return this.invertLogTextSearch ? !match : match;
 				});
 			}


modules/luci-base/htdocs/luci-static/resources/tools/views.js
```diff
@@ -116,8 +116,9 @@ var CBILogreadBox = function(logtag, name) {
 					return line.toLowerCase().includes(this.logTagFilter?.toLowerCase());
 				});
 
+				const filter = this.logTextFilter.toLowerCase();
 				loglines = loglines.filter(line => {
-					const match = line.includes(this.logTextFilter);
+					const match = line.toLowerCase().includes(filter);
 					return this.invertLogTextSearch ? !match : match;
 				});
```
openwrt-ai: nit: parity gap with the sister implementation this PR also touches — `dmesg.js` wraps the same text filter in `if (this.logTextFilter)` ([dmesg.js:97-101](``https://github.com/openwrt/luci/blob/54967fb7b93060665e691cf8a2a1db461c2fddc3/modules/luci-mod-status/htdocs/luci-static/resources/view/status/dmesg.js#L97-L101)),`` but `CBILogreadBox` runs it unconditionally. With an empty search box and the "Not" checkbox ticked, `line.toLowerCase().includes('')` is always `true`, so the inversion hides every line and the System Log renders empty.

This is pre-existing (it behaved the same before the `toLowerCase()` change), but since you're already aligning the two views' search semantics here, adding the guard would make them consistent:

```suggestion
				if (this.logTextFilter) {
					const filter = this.logTextFilter.toLowerCase();
					loglines = loglines.filter(line => {
						const match = line.toLowerCase().includes(filter);
						return this.invertLogTextSearch ? !match : match;
					});
				}
```

Worth folding into this series or leaving for a separate fix, your call.

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit messages match their diffs (`luci-base` → `tools/views.js`, `luci-mod-status` → `status/dmesg.js`), and the two changed sites are the only case-sensitive log-text filters in the tree — `logTagFilter` was already lowercased, and the facility/severity filters compare against fixed dropdown values. Hoisting `toLowerCase()` out of the filter callback is the right shape, and both `logTextFilter` fields are initialised to `''` and only ever assigned from an `<input>.value`, so the unguarded `.toLowerCase()` in `views.js` cannot throw.

One inline nit about empty-filter behaviour in `CBILogreadBox`.

---

Ser9ei:
Added: Skip empty text filters in luci-base, matching luci-mod-status.

Notes on a separate pre-existing issue:

While verifying the empty-filter + “Not including” behaviour, I noticed another pre-existing problem in CBILogreadBox (views.js).
When any filter control is changed while the text filter is empty, the displayed log set changes and older entries appear, entries with timestamps that were not present in the previous view at all.

This is stable and reproducible even when no new messages are being written to the log. It is not caused by the ring buffer simply advancing with new messages.

openwrt-ai (commented):

The only change since the last review is the `if (this.logTextFilter)` guard in `CBILogreadBox`, which is exactly what the earlier nit asked for — the two views' text-search semantics now match, and an empty search box with "Not including" ticked no longer blanks the System Log. Both commit messages still match their diffs, and the `luci-base` body was updated to mention the added empty-filter skip. CI on the head commit is green.

The unrelated behaviour you described in the PR comment (older entries appearing when a filter control changes) looks like a separate pre-existing issue in the log-fetch path rather than anything introduced here; a follow-up PR or issue is the right place for it.

---

openwrt-ai (commented):

---

modules/luci-base/htdocs/luci-static/resources/tools/views.js
openwrt-ai: fixed, thanks

---

danielfdickinson:
I think this should be a toggle not forced on every user. There are times case-sensitive search really helps narrow down the search.

Ser9ei:
@danielfdickinson, In practice, the need for case-sensitive log search is fairly rare, most UIs default to case-insensitive.
It may be better to ask the maintainers for their opinion on a “Match case” option in the original issue.

danielfdickinson:
@jow- agrees, so I withdraw my comment.

systemcrash:
toLowerCase is too naive in this case and does not take other utf-8 or foreign language cases into account. I think the equivalent is toLocaleLowerCase() 

But case sensitivity is generally preferred, and search works on partial strings. So at a minimum case-sensitive should be preferred if we are to introduce an 'ignore case' boolean.

Ser9ei:
Thanks for the feedback, @systemcrash. I suggest discussing the
design on the original issue rather than on the implementation.

PR #8971, opened 2026-08-23
luci-base: fix auth plugin asset loading

FuSan21:
`normalize_assets()` builds its path pattern with `sprintf()` and hands the resulting string to `match()`. ucode's `match()` only accepts a regexp:

```c
if (ucv_type(pattern) != UC_REGEXP || !subject)
	return NULL;
```

It returns null whatever the subject is, `!match(...)` is therefore always true, and every asset is dropped by the `continue` under it. As far as I can tell no auth plugin has ever had its scripts rendered on the login page since the mechanism was added in 4a308bab.

It fails silently, which makes it unpleasant to track down. The `html` from the same `check()` result is unaffected, so the plugin's login UI renders normally but without the script that drives it, and nothing is logged. `plugins/luci-plugin-auth-example` declares `assets` too, so the shipped example is affected as well.

Reproducer:

```
$ ucode -e 'print(match("/a/b", "^/a/") ? "MATCH" : "NULL", "\n")'
NULL
$ ucode -e 'print(match("/a/b", /^\/a\//) ? "MATCH" : "NULL", "\n")'
MATCH
```

The check is a plain prefix test, so the first commit uses `index()` rather than compiling a regexp inside the loop. It returns the offset of the first occurrence or `-1`, which makes `!= 0` the equivalent of the anchored `^`, and it matches the existing idiom at `http.uc:397`.

The second commit is a follow-up from review. Because this series is what makes `normalize_assets()` return a non-empty result for the first time, the surviving `src` now reaches `<script src="{{ asset.src }}"></script>` unescaped in all three in-tree sysauth templates (`luci-base`, `bootstrap`, `footstrap`). The existing filter rejects `..` and whitespace but not quotes or angle brackets, so a `src` ending `x.js"></script><script>...` would close the attribute — no whitespace required. `normalize_assets()` is the single point every theme passes through, so the rejected character class is widened there rather than escaping in each template, which also covers out-of-tree themes carrying the same markup.

Found while getting a third-party WebAuthn auth plugin working. Verified on qualcommax/ipq807x with luci-base 26.234.21935: `auth_assets` arrives empty at the login template and no `<script>` tag is emitted, and correcting the path check makes the tag render and the passkey login complete.

### DIFF
--- a/modules/luci-base/ucode/authplugins.uc
+++ b/modules/luci-base/ucode/authplugins.uc
@@ -157,7 +157,7 @@ function normalize_assets(uuid, assets) {
 		if (type(src) != 'string')
 			continue;
 
-		if (!match(src, sprintf("^/luci-static/plugins/%s/", uuid)))
+		if (index(src, sprintf("/luci-static/plugins/%s/", uuid)) != 0)
 			continue;
 
 		if (match(src, /\.\.|[\r\n\t ]/))
--- a/modules/luci-base/ucode/authplugins.uc
+++ b/modules/luci-base/ucode/authplugins.uc
@@ -160,7 +160,7 @@ function normalize_assets(uuid, assets) {
 		if (index(src, sprintf("/luci-static/plugins/%s/", uuid)) != 0)
 			continue;
 
-		if (match(src, /\.\.|[\r\n\t ]/))
+		if (match(src, /\.\.|[\r\n\t "'<>`]/))
 			continue;
 
 		push(rv, { type: 'script', src: src });


modules/luci-base/ucode/authplugins.uc
```diff
@@ -157,7 +157,7 @@ function normalize_assets(uuid, assets) {
 		if (type(src) != 'string')
 			continue;
 
-		if (!match(src, sprintf("^/luci-static/plugins/%s/", uuid)))
+		if (!match(src, regexp(sprintf("^/luci-static/plugins/%s/", uuid))))
```
openwrt-ai: nit: `regexp()` sits inside the `for (let asset in assets)` loop, so the same loop-invariant pattern is recompiled for every asset. Since the check is a plain prefix test, `index()` does the same job without compiling anything, and it also removes the need for the metacharacter-safety argument in the commit message (no regex, no metacharacters, regardless of how `uuid` is validated):

```suggestion
		if (index(src, sprintf("/luci-static/plugins/%s/", uuid)) != 0)
```

`index()` on two strings returns the byte offset of the first occurrence or `-1`, so `!= 0` is the exact equivalent of the anchored `^`. If you prefer to keep the regexp form, hoisting the compile above the loop would be enough.

---

modules/luci-base/ucode/authplugins.uc
```diff
@@ -157,7 +157,7 @@ function normalize_assets(uuid, assets) {
...
 			continue;
 
 		if (match(src, /\.\.|[\r\n\t ]/))
```
openwrt-ai: Question, since this PR is what makes `normalize_assets()` produce a non-empty result for the first time: the surviving `src` is interpolated raw into an HTML attribute at [`sysauth.ut:67`](https://github.com/openwrt/luci/blob/8700150960bd720a269560e0bf2881fcfd7b2254/modules/luci-base/ucode/template/sysauth.ut#L67) (`<script src="{{ asset.src }}"></script>`), and ucode's `{{ }}` does not escape — the same template escapes explicitly where it matters, e.g. [`entityencode(duser, true)` at sysauth.ut:31](https://github.com/openwrt/luci/blob/8700150960bd720a269560e0bf2881fcfd7b2254/modules/luci-base/ucode/template/sysauth.ut#L31).

This filter rejects `..` and `[\r\n\t ]` but not `"`, `<` or `>`, so a `src` such as `/luci-static/plugins/<uuid>/x.js"></script><script>...</script>` passes both checks and breaks out of the attribute. No spaces are required for that payload, so the whitespace class doesn't stop it.

I don't think this is a privilege boundary — `result.assets` comes from a plugin's own `check()`, which is server-side ucode that could emit script through `html` anyway. But the two existing checks read like this function is meant to be the sanitizer, and a plugin that derives `src` from request input would now hand that straight to the login page. Was allowing quotes/angle brackets through intentional, or would adding them to the rejected set (or `entityencode`-ing at the template) be worth folding in here?

---

openwrt-ai (commented):
The diagnosis checks out. `uc_match()` bails with `if (ucv_type(pattern) != UC_REGEXP || !subject) return NULL;` ([lib.c:3224](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/lib.c#L3224)), so the string pattern made the check unconditionally false and every asset was dropped. `plugin.uuid` is indeed only assigned after `match(uuid, /^[a-f0-9]{32}$/)` passes during `load()`, so the interpolation-safety claim holds, and `regexp()` is a core builtin available under `'use strict'`. This was also the only `match()` call in the tree passing a non-regexp pattern, so no sibling instances need the same fix.

Two inline notes, neither blocking.

---

modules/luci-base/ucode/authplugins.uc
FuSan21: Done, switched to `index()`. It also lets the commit message drop the metacharacter-safety argument entirely, which is a better place to end up than justifying why the interpolation happens to be safe.

modules/luci-base/ucode/authplugins.uc
FuSan21: Folded in as a second commit. I went with widening the rejected set rather than escaping at the template because three in-tree sysauth templates carry the same `<script src="{{ asset.src }}">` markup (luci-base, bootstrap, footstrap) and out-of-tree themes copy it; `normalize_assets()` is the one point they all pass through. Agreed it isn't a privilege boundary, since `check()` is server-side ucode that can emit script through `html` regardless — but the function reads like it's meant to be the sanitizer, and it should hold for a plugin deriving `src` from request input. Now rejecting `"`, `'`, `<`, `>` and backtick alongside the existing `..`/whitespace class.

modules/luci-base/ucode/authplugins.uc
```diff
@@ -157,10 +157,10 @@ function normalize_assets(uuid, assets) {
 		if (type(src) != 'string')
 			continue;
 
-		if (!match(src, sprintf("^/luci-static/plugins/%s/", uuid)))
+		if (index(src, sprintf("/luci-static/plugins/%s/", uuid)) != 0)
```
openwrt-ai, line 160: nit: the PR description is now out of date with respect to this line. It still says "Compiling the pattern with `regexp()` first fixes it. `uuid` is already validated against `/^[a-f0-9]{32}$/` before the plugin is loaded, so the interpolated value can't contain metacharacters" — but the commit that landed uses `index()`, compiles nothing, and its message explicitly drops that argument ("comparing two plain strings keeps the check independent of how uuid is validated"). The description also doesn't mention the second commit (`b179d85`, the quoting-character rejection) at all, so someone reading only the PR body would not know the series contains a hardening change. Worth refreshing the body before merge so the merge-time record matches the commits.

The change itself is right: `uc_index()` returns `0` only for a match at offset 0 for two `UC_STRING` operands ([lib.c:432-459](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/lib.c#L432-L459)), so `!= 0` is the exact equivalent of the anchored `^`, and it matches the existing in-tree idiom at [`http.uc:397`](https://github.com/openwrt/luci/blob/b179d85a5e9da04f351f08116b43866b651b0c20/modules/luci-base/ucode/http.uc#L397).

---

openwrt-ai (commented):
Both notes from the previous round are addressed.

The `index()` switch is correct and is the in-tree idiom for a prefix test. The widened character class parses as intended — inside a regexp literal ucode's lexer reads a bracket expression through to the closing `]` and pushes `"`, `'` and `` ` `` verbatim ([lexer.c:400-412](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/lexer.c#L400-L412)), so no escaping is needed there. Filtering in `normalize_assets()` rather than in the templates is the right call: all three in-tree sysauth templates (`modules/luci-base/ucode/template/sysauth.ut:67`, `themes/luci-theme-bootstrap/.../sysauth.ut:52`, `themes/luci-theme-footstrap/.../sysauth.ut:134`) carry the same unescaped `src="{{ asset.src }}"`, and out-of-tree themes copying that markup are covered for free. With `<`, `>`, `"`, `'` and backtick rejected, the attribute-breakout payload from the earlier round no longer survives, and the fixed `/luci-static/plugins/<uuid>/` prefix rules out scheme-based `src` values.

Both commit messages match their diffs. One nit inline about the PR description having drifted from the code.

---

modules/luci-base/ucode/authplugins.uc
FuSan21: Refreshed it: the `regexp()`/metacharacter paragraph is replaced with the `index()` rationale, and there's now a paragraph covering the second commit and why the filtering sits in `normalize_assets()` rather than the templates.

openwrt-ai (commented):
## Commit checks

- `fa2a6ee` "Merge branch 'master' into authplugins-asset-regexp" — merge commits aren't allowed in a PR series and this one is the sole cause of the red `FormalityCheck / Git & Commits`. Rebase the two commits onto current master and force-push so the branch is linear again; master hasn't touched `authplugins.uc` or any sysauth template since the branch point, so the rebase is a no-op content-wise.

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

systemcrash:
OK - thanks for the fix!

PR #8972, opened 2026-08-23
luci-app-shunt: add LuCI support for shunt

dibdot:
Maintainer: @dibdot

Description:
Companion frontend for the shunt policy-based routing daemon (openwrt/packages#30365). Provides an overview with live service state, per-policy rule and route counters and learned-address totals, a policy editor covering all selectors (client address and MAC, destination, domain, port, protocol), set reporting with per-element counters, and a filtered processing log. Status data comes from a ucode rpcd object shipped with the base package.

🧪 Run Testing Details
Bananapi BPI-R3, mediatek/filogic, OpenWrt SNAPSHOT (r35906-3d1645ee26)

### DIFF
--- a/applications/luci-app-shunt/htdocs/luci-static/resources/view/shunt/overview.js
+++ b/applications/luci-app-shunt/htdocs/luci-static/resources/view/shunt/overview.js
@@ -1,0 +1,441 @@
+'use strict';
+'require dom';
+'require view';
+'require poll';
+'require fs';
+'require ui';
+'require uci';
+'require rpc';
+'require form';
+'require tools.widgets as widgets';
+
+const getStatus = rpc.declare({
+	object: 'luci.shunt',
+	method: 'status'
+});
+
+const getPackages = rpc.declare({
+	object: 'rpc-sys',
+	method: 'packagelist',
+	params: ['all'],
+	expect: { packages: {} }
+});
+
+function handleAction(ev) {
+	if (ev === 'restart') {
+		const map = document.querySelector('.cbi-map');
+		return dom.callClassMethod(map, 'save')
+			.then(L.bind(ui.changes.apply, ui.changes))
+			.then(function () {
+				return fs.exec_direct('/etc/init.d/shunt', [ev]);
+			});
+	}
+	return fs.exec_direct('/etc/init.d/shunt', [ev]);
+}
+
+function fmtMark(mark) {
+	if (mark == null) {
+		return '-';
+	}
+	return '0x%08x'.format(mark);
+}
+
+function fmtCount(n) {
+	if (n == null) {
+		return '?';
+	}
+	return '%d'.format(n);
+}
+
+function card(label, value, sub, cls) {
+	return E('div', { 'class': 'shunt-card' }, [
+		E('div', { 'class': 'shunt-label' }, label),
+		E('div', { 'class': 'shunt-value ' + (cls || '') }, value),
+		sub ? E('div', { 'class': 'shunt-sub' }, sub) : ''
+	]);
+}
+
+function dot(state) {
+	return E('span', { 'class': 'shunt-dot shunt-dot-' + state });
+}
+
+function renderState(st) {
+	if (!st || st.running == null) {
+		return E('div', { 'class': 'shunt-state' }, [
+			dot('off'), E('span', {}, _('no answer from the backend'))
+		]);
+	}
+
+	if (st.running && st.applied) {
+		return E('div', { 'class': 'shunt-state' }, [
+			dot('ok'), E('span', {}, _('running, policy applied'))
+		]);
+	}
+
+	if (st.running && !st.applied) {
+		return E('div', { 'class': 'shunt-state' }, [
+			dot('warn'), E('span', {}, _('running, but no ruleset in the kernel'))
+		]);
+	}
+
+	if (!st.running && st.applied) {
+		return E('div', { 'class': 'shunt-state' }, [
+			dot('warn'), E('span', {}, _('ruleset present, service not running'))
+		]);
+	}
+
+	return E('div', { 'class': 'shunt-state' }, [
+		dot('off'), E('span', {}, _('stopped'))
+	]);
+}
+
+// The observer's verdict identifiers are module contract - translated for
+// display only, with the raw key kept beside each label.
+const DROP_LABEL = {
+	'nomatch': _('Domain is in no policy'),
+	'qtype': _('Not an address query'),
+	'noaddr': _('Answer carried no usable address'),
+
+	'dns:E_RCODE': _('Error reply, e.g. NXDOMAIN'),
+	'dns:E_NOTRESP': _('Not a response'),
+	'dns:E_TRUNC': _('Truncated response'),
+	'dns:E_QDCOUNT': _('Not exactly one question'),
+	'dns:E_SHORT': _('Message ends mid-structure'),
+	'dns:E_MSGLEN': _('Message too long'),
+	'dns:E_QPTR': _('Compression pointer in the question'),
+	'dns:E_LABEL': _('Reserved label type'),
+	'dns:E_NAMELEN': _('Name too long'),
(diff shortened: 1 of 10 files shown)


applications/luci-app-shunt/htdocs/luci-static/resources/view/shunt/overview.js
```diff
@@ -0,0 +1,436 @@
...
+				E('div', { 'id': 'shunt-snoop' }, ''),
+				E('div', { 'id': 'shunt-issues' }, '')
+			]);
+		}, this);
+
+		poll.add(function () {
+			return L.resolveDefault(getStatus(), null).then(update);
+		}, 2);
+
+		update(result[0]);
```
openwrt-ai: This call is a no-op, so the status preloaded in `load()` is thrown away. `update()` resolves its targets with `document.getElementById()`, but the `#shunt-status` subtree is only built when `m.render()` runs (line 430) and is inserted into the document *after* `render()`'s promise resolves — see [`View.__init__` in luci.js:1966-1970](https://github.com/openwrt/luci/blob/ca7adf0c296557c62489b86434c80402589fc799/modules/luci-base/htdocs/luci-static/resources/luci.js#L1966-L1970). At this point nothing is attached, every `setNodes()` finds `null` and returns, and the cards stay on `-` until the first poll tick up to 2 s later.

`setreport.js:252` in this same PR already solves it with `requestAnimationFrame`; same fix here:

```suggestion
		requestAnimationFrame(function () {
			update(result[0]);
		});
```

---

applications/luci-app-shunt/htdocs/luci-static/resources/view/shunt/setreport.js
```diff
@@ -0,0 +1,262 @@
...
+					_('No client is selected, so every client is covered.')),
+				renderCards(sets, ['s'], _('Static Destinations'), null),
+				renderLearned(sets)
+			]);
+		};
+
+		const reload = function () {
+			const sel = document.getElementById('shunt-policy');
+
+			return getSets(sel ? sel.value : '').then(render_sets);
```
openwrt-ai: Unlike `load()` on line 162, this call is not wrapped in `L.resolveDefault()`. `luci.shunt` is provided by the `shunt` package's rpcd object, so the call rejects whenever the service is stopped or rpcd has not loaded the object — and both the Refresh button and the policy `select` route through `ui.createHandlerFn`, which only attaches [`.finally()` and no `.catch()` (ui.js:5689)](https://github.com/openwrt/luci/blob/ca7adf0c296557c62489b86434c80402589fc799/modules/luci-base/htdocs/luci-static/resources/ui.js#L5689). The result is an unhandled promise rejection, a spinner that stops with no feedback, and stale contents left on screen.

```suggestion
			return L.resolveDefault(getSets(sel ? sel.value : ''), {}).then(render_sets);
```

---

applications/luci-app-shunt/htdocs/luci-static/resources/view/shunt/overview.js
```diff
@@ -0,0 +1,436 @@
...
+			setNodes('shunt-state', renderState(st));
+			setNodes('shunt-version', E('span', {}, ['%s / %s'.format(
+				pkgs['luci-app-shunt'] || _('n/a'), pkgs['shunt'] || _('n/a'))]));
+			setNodes('shunt-learned', E('span', {}, [
+				svc ? '%d'.format(svc.dedupe || 0) : '-']));
+			setNodes('shunt-poll', E('span', {}, [svc
+				? (svc.poll.resolv
+					? _('%d name(s) every %ds').format(svc.poll.names, svc.poll.interval)
+					: _('unavailable - ucode-mod-resolv missing'))
+				: '-']));
```
openwrt-ai: `svc` is checked but `svc.poll` is dereferenced unguarded, so a status payload without a `poll` key throws a `TypeError` inside the poll callback — which aborts `update()` before `setNodes('shunt-policies'/'shunt-snoop'/'shunt-issues')` run, leaving the rest of the page frozen every tick. `renderSnoop()` guards the sibling key explicitly (`if (!svc || !svc.snoop)`, line 127) and `svc.dedupe` is guarded on line 316; this is the one sub-object taken on faith from the rpc reply.

```suggestion
			setNodes('shunt-poll', E('span', {}, [svc?.poll
				? (svc.poll.resolv
					? _('%d name(s) every %ds').format(svc.poll.names, svc.poll.interval)
					: _('unavailable - ucode-mod-resolv missing'))
				: '-']));
```

---

applications/luci-app-shunt/htdocs/luci-static/resources/view/shunt/policies.js
```diff
@@ -0,0 +1,143 @@
...
+				netifd name is resolved to its device; any other device name is used as entered.'));
+		o.rmempty = false;
+
+		uci.sections('network', 'interface').forEach(function (n) {
+			if (n['.name'] !== 'loopback') {
+				o.value(n['.name'], '%s (%s)'.format(n['.name'], _('interface')));
+			}
+		});
+
+		o = s.option(form.ListValue, 'fallback', _('Fallback Behavior'),);
```
openwrt-ai: nit: stray trailing comma in the argument list (harmless since ES2017, but it reads like a dropped description argument — every other `s.option()` in this file passes one).

```suggestion
		o = s.option(form.ListValue, 'fallback', _('Fallback Behavior'));
```

---

applications/luci-app-shunt/htdocs/luci-static/resources/view/shunt/overview.js
```diff
@@ -0,0 +1,436 @@
...
+			E('th', { 'class': 'th' }, _('Fallback'))
+		])
+	];
+
+	st.policies.forEach(function (p) {
+		rows.push(E('tr', { 'class': 'tr' }, [
+			E('td', { 'class': 'td' }, [p.name]),
+			E('td', { 'class': 'td' }, [p.interface || '-']),
+			E('td', { 'class': 'td' }, [fmtMark(p.mark)]),
+			E('td', { 'class': 'td' }, ['%d'.format(p.rt_table)]),
```
openwrt-ai: nit: this is the only numeric column that bypasses `fmtCount()`. `'%d'.format(null)` yields `0` (the formatter coerces via `Math.floor(+param || 0)`), so a policy whose `rt_table` the backend omits renders a plausible-looking table `0` instead of the `?` used for `rules`/`routes` on the next two lines — and `mark` right above already maps missing to `-` via `fmtMark()`.

```suggestion
			E('td', { 'class': 'td' }, [fmtCount(p.rt_table)]),
```

---

openwrt-ai (commented):
Reviewed 1 new commit. Commit message matches the diff; CI on `ca7adf0` is green. Five inline notes (three issues, two nits) — nothing structural: the Makefile, ACL, menu entry, uci-defaults housekeeping and `logtemplate.js` all follow the existing `luci-app-banip` / `luci-app-adblock` layout.

The backend dependency on openwrt/packages#30365 (`luci.shunt` ubus object, the `shunt` UCI options) is called out in the PR description, so the option names could not be cross-checked against a merged consumer here.

---

openwrt-ai (commented):

The force-push resolves all five points from the previous review: `rt_table` now goes through `fmtCount()`, `svc?.poll` is guarded, `overview.js` and `setreport.js` both defer the initial paint to `requestAnimationFrame`, `getSets()` in `reload()` is wrapped in `L.resolveDefault(..., {})` (and `render_sets` degrades cleanly on `{}`), and the stray trailing comma in `policies.js` is gone. `shunt.pot` was regenerated and its line references match the new sources. Optional chaining is already used in `luci-base` (`uci.js`, `ui.js`), so `svc?.poll` is in line with the tree. CI on `11b6524` is green across all eight checks.

The backend dependency on openwrt/packages#30365 (`luci.shunt` ubus object and the `shunt` UCI options) is still unmerged and called out in the PR description, so the option names remain uncross-checkable against a merged consumer.

---

PR #8974, opened 2026-08-23
themes: add granular individual classes to alerts

pktpls:
### Description
Putting this up for discussion - let me know if there's a better way to achieve this. I will do the same in the other themes if there's consensus.

This allows applications to control how specific alerts are displayed.

One example is an application which ships with its own flow for setting
a root password. It'd be confusing to show an alert on top of this flow,
which sends users to a different password flow.

Until now you'd need to fork the theme to achieve this.

### Maintainer _(preferred)_

@jow- 

---

## Tested on

**OpenWrt version:** snapshot
**LuCI version:** main
**Web browser(s):** Firefox

### DIFF
--- a/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
+++ b/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
@@ -55,7 +55,7 @@
 
 		<div id="maincontent" class="container">
 			{% if (getuid() == 0 && getspnam('root')?.pwdp === ''): %}
-				<div class="alert-message warning">
+				<div class="alert-message warning password-absent">
 					<h4>{{ _('No password set!') }}</h4>
 					<p>{{ _('There is no password set on this router. Please configure a root password to protect the web interface.') }}</p>
 					{% if (dispatcher.lookup("admin/system/admin")): %}
@@ -65,7 +65,7 @@
 			{% endif %}
 
 			{% if (boardinfo.rootfs_type == "initramfs"): %}
-				<div class="alert-message warning">
+				<div class="alert-message warning recovery-mode">
 					<h4>{{ _('System running in recovery (initramfs) mode.') }}</h4>
 					<p>{{ _('No changes to settings will be stored and are lost after rebooting. This mode should only be used to install a firmware upgrade') }}</p>
 					{% if (dispatcher.lookup("admin/system/flash")): %}
@@ -75,7 +75,7 @@
 			{% endif %}
 
 			<noscript>
-				<div class="alert-message warning">
+				<div class="alert-message warning js-required">
 					<h4>{{ _('JavaScript required!') }}</h4>
 					<p>{{ _('You must enable JavaScript in your browser or LuCI will not work properly.') }}</p>
 				</div>


themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
```diff
@@ -55,7 +55,7 @@
 
 		<div id="maincontent" class="container">
 			{% if (getuid() == 0 && getspnam('root')?.pwdp === ''): %}
-				<div class="alert-message warning">
+				<div class="alert-message warning password-absent">
```
openwrt-ai, line 58: Since you're asking for a better way: each of these three notices is a singleton in the rendered page, and this file already exposes its stable hooks as ids rather than as marker classes — [`#topmenu` / `#indicators` / `#maincontent` at header.ut:52-56](``https://github.com/openwrt/luci/blob/c32fa5534f205b904a7471871cc7ad5a0b0c2e60/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut#L52-L56``) and [`#tabmenu` at header.ut:84](``https://github.com/openwrt/luci/blob/c32fa5534f205b904a7471871cc7ad5a0b0c2e60/themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut#L84).`` Was an id (`<div id="notice-no-password" class="alert-message warning">`) considered instead? The existing classes here (`alert-message`, `warning`, `right`, `btn`) are presentational and shared, whereas these three new ones are unique identifiers with no styling of their own — an id says that, gives the third-party rule enough specificity to win without `!important`, and keeps generic names like `recovery-mode` out of the global class namespace where another app's stylesheet may already use them.

Whichever shape you settle on, one thing worth knowing before you replicate it across the other themes: footstrap doesn't have three copies of this markup to patch. All three of its notices render through one shared partial, [`partials/notice.ut:15`](``https://github.com/openwrt/luci/blob/c32fa5534f205b904a7471871cc7ad5a0b0c2e60/themes/luci-theme-footstrap/ucode/template/themes/footstrap/partials/notice.ut#L15),`` driven by [`partials/notices.ut`](``https://github.com/openwrt/luci/blob/c32fa5534f205b904a7471871cc7ad5a0b0c2e60/themes/luci-theme-footstrap/ucode/template/themes/footstrap/partials/notices.ut#L11-L34),`` so it needs a new pass-through parameter on the partial rather than a copy of this hunk.

---

themes/luci-theme-bootstrap/ucode/template/themes/bootstrap/header.ut
```diff
@@ -75,7 +75,7 @@
 			{% endif %}
 
 			<noscript>
-				<div class="alert-message warning">
+				<div class="alert-message warning js-required">
```
openwrt-ai, line 78: nit: the three names are each formed differently — `password-absent` (noun-adjective), `recovery-mode` (state), `js-required` (a requirement, not a description of the element it sits on). These become the identifiers third-party CSS targets across all five themes and will effectively be frozen once apps rely on them, so one predictable pattern would be easier to document and grep for (e.g. `notice-no-password` / `notice-initramfs` / `notice-noscript`, matching what actually gates each block).

`password-absent` also reads inverted from every other in-tree name for the same condition: the heading right below it says "No password set!", and footstrap computes the same test as [`empty_password` at footstrap header.ut:30](``https://github.com/openwrt/luci/blob/c32fa5534f205b904a7471871cc7ad5a0b0c2e60/themes/luci-theme-footstrap/ucode/template/themes/footstrap/header.ut#L30).`` Same applies to lines 58 and 68.

---

openwrt-ai (commented):
## Commit checks

- c32fa55 "themes: add granular individual classes to alerts" — the subject scope is `themes:` but the diff only touches `luci-theme-bootstrap`. In-tree convention uses the package name when a single theme changes ([084fae3](https://github.com/openwrt/luci/commit/084fae3db599bcf51e0c68b6f4ef37e49bd9b794) "luci-theme-bootstrap: remove broken title class on &lt;body&gt;") and reserves `themes:` for changes that actually span them (your own [7c6d8ff](https://github.com/openwrt/luci/commit/7c6d8ff7f8b894d1c8c3382af4b79b7cea68c709), which touched four themes plus the dispatcher). Either narrow the subject to `luci-theme-bootstrap:` or land the other themes in the same series.

Note for reviewers: the three files under `ucode/template/themes/` are not three copies — `bootstrap-dark` and `bootstrap-light` are symlinks to `bootstrap`, so all three shipped themes in this package pick the change up automatically.

---

pktpls:
I have found another way: my CSS hides all warnings (`#maincontent>.warning`), and then my JS reinstates the initramfs warning. This one can be found based on text content across translations (they all contain "initramfs").

PR #8976, opened 2026-08-24
luci-app-dockerman: handle Docker multiplexed stream logs

jjm2473:
Docker API can return container logs using the
application/vnd.docker.multiplexed-stream format [1].
The response consists of framed stdout/stderr records rather than plain text.

Currently these responses are not decoded correctly, resulting in
garbled or missing log output.

Parse the multiplexed stream format in both the ucode RPC handler and the
LuCI API client, then combine the frame payloads before passing the logs
to the existing ANSI-to-HTML renderer.

[1] https://docs.docker.com/reference/api/engine/version/v1.43/#stream-format

### Description
`docker.container logs` does not work with "application/vnd.docker.multiplexed-stream" response.

1. Create a container **without pseudo-TTY (`-t`)** `docker run --name testlog alpine ls -la`
2. Run `ubus call docker.container logs '{"id":"testlog","query":{"stdout":true,"stderr":true,"tail":100}}'`

Before this patch:
```json
{
        "code": 200,
        "headers": {
                "api-version": "1.47",
                "content-type": "application/vnd.docker.multiplexed-stream",
                "docker-experimental": "false",
                "ostype": "linux",
                "server": "Docker/27.3.1 (linux)",
                "date": "Mon, 24 Aug 2026 13:15:12 GMT",
                "connection": "close",
                "transfer-encoding": "chunked"
        },
        "body": "\u0001"
}
```

After patch:
```json
{
        "code": 200,
        "headers": {
                "api-version": "1.47",
                "content-type": "application/vnd.docker.multiplexed-stream",
                "docker-experimental": "false",
                "ostype": "linux",
                "server": "Docker/27.3.1 (linux)",
                "date": "Mon, 24 Aug 2026 13:16:07 GMT",
                "connection": "close",
                "transfer-encoding": "chunked"
        },
        "body": [
                {
                        "type": 1,
                        "payload": "total 64\n"
                },
                {
                        "type": 1,
                        "payload": "drwxr-xr-x    1 root     root          4096 Aug 24 12:29 .\n"
                },
                {
                        "type": 1,
                        "payload": "drwxr-xr-x    1 root     root          4096 Aug 24 12:29 ..\n"
                },
                {
                        "type": 1,
                        "payload": "-rwxr-xr-x    1 root     root             0 Aug 24 12:29 .dockerenv\n"
                },
                {
                        "type": 1,
                        "payload": "drwxr-xr-x    2 root     root          4096 Oct  8  2025 bin\n"
                },
               ...
        ]
}
```

Before:
image

After:
image

### Maintainer _(preferred)_

@systemcrash 

---

## Tested on

**Web browser(s):** Chrome 151.0.7922.76 

(Both JS API and RPC were tested)

---

## Checklist

### DIFF
--- a/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
+++ b/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
@@ -116,6 +116,28 @@ function processLines(buffer, onChunk) {
 	return buffer;
 }
 
+// Parse the docker multiplexed stream format, https://docs.docker.com/reference/api/engine/version/v1.43/#stream-format
+function parse_multiplexed_stream(buffer) {
+	const output = [];
+	// Only three stream types are defined: 0=stdin, 1=stdout, 2=stderr.
+	const decoders = [new TextDecoder(), new TextDecoder(), new TextDecoder()];
+
+	while (buffer.length > 0) {
+		if (buffer.length < 8) break; // Not enough data for header
+		let stream_type = buffer[0];
+		let payload_length = ((buffer[4] << 24) | (buffer[5] << 16) | (buffer[6] << 8) | buffer[7]) >>> 0; // Convert to unsigned
+		if (buffer.length < 8 + payload_length) break; // Not enough data for payload
+		if (stream_type <= 2) {
+			let payload = buffer.subarray(8, 8 + payload_length);
+			output.push({ type: stream_type, payload: decoders[stream_type].decode(payload, { stream: true }) });
+		} else {
+			console.warn(`Unknown stream type ${stream_type} in multiplexed stream, ignoring`);
+		}
+		buffer = buffer.subarray(8 + payload_length);
+	}
+
+	return output;
+}
 
 function call_docker(method, path, options = {}) {
 	return loadPromise.then(() => {
@@ -221,6 +243,17 @@ function call_docker(method, path, options = {}) {
 				const headersObj = {};
 				for (const [key, value] of response.headers.entries()) {
 					headersObj[key] = value;
+				}
+
+				if ("application/vnd.docker.multiplexed-stream" === headersObj['content-type']) {
+					return response.arrayBuffer().then(buffer => {
+						const parsed = parse_multiplexed_stream(new Uint8Array(buffer));
+						return {
+							code: response.status,
+							body: parsed,
+							headers: headersObj,
+						};
+					});
 				}
 
 				return response.text().then(text => {
--- a/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
+++ b/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
@@ -1685,7 +1685,12 @@ return dm2.dv.extend({
 					return false;
 				}
 
-				const logText = response?.body || _('No logs available');
+				let logText = response?.body;
+				if ("application/vnd.docker.multiplexed-stream" === response?.headers?.['content-type']) {
+					logText = (response?.body || []).map(frame => frame.payload).join('');
+				}
+				logText = logText || _('No logs available');
+
 				// Convert ANSI codes to HTML and set innerHTML
 				logsDiv.innerHTML = dm2.ansiToHtml(logText);
 				logsDiv.scrollTop = logsDiv.scrollHeight;
--- a/applications/luci-app-dockerman/root/usr/share/rpcd/ucode/docker_rpc.uc
+++ b/applications/luci-app-dockerman/root/usr/share/rpcd/ucode/docker_rpc.uc
@@ -100,6 +100,23 @@ function coerce_values_to_string(obj) {
 	return obj;
 };
 
+// Parse the docker multiplexed stream format, https://docs.docker.com/reference/api/engine/version/v1.43/#stream-format
+function parse_multiplexed_stream(buffer) {
+	let output = [];
+
+	while (length(buffer) > 0) {
+		if (length(buffer) < 8) break; // Not enough data for header
+		let stream_type = ord(buffer, 0);
+		let payload_length = ord(buffer, 4) << 24 | ord(buffer, 5) << 16 | ord(buffer, 6) << 8 | ord(buffer, 7);
+		if (length(buffer) < 8 + payload_length) break; // Not enough data for payload
+		let payload = substr(buffer, 8, payload_length);
+		push(output, { type: stream_type, payload: payload });
+		buffer = substr(buffer, 8 + payload_length);
+	}
+
+	return output;
+};
+
 function call_docker(method, path, options) {
 	options = options || {};
 	const headers = options.headers || {};
@@ -277,6 +294,8 @@ function call_docker(method, path, options) {
 		} else {
 			response_body = data;
 		}
+	} else if (response_headers['content-type'] === 'application/vnd.docker.multiplexed-stream' && response_body) {
+		response_body = parse_multiplexed_stream(response_body);
 	}
 
 	return {


applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
```diff
@@ -228,20 +244,24 @@ function call_docker(method, path, options = {}) {
...
-							'application/x-ndjson',
-							'application/json-seq'].includes(contentType) || safeText.includes('\n')) {
-							const lines = safeText.split(/\r?\n/).filter(Boolean);
-							try {
-								parsed = lines.map(l => JSON.parse(l));
-							} catch (err2) {
-								// Fall back to raw text if parsing fails
-								parsed = text;
+					if ("application/vnd.docker.multiplexed-stream" == contentType) {
+						parsed = parse_multiplexed_stream(safeText);
```
openwrt-ai: `safeText` comes from `response.text()`, which UTF-8-decodes the body — but the multiplexed stream is binary, so character offsets are not byte offsets and the frame header cannot be read back reliably.

Concretely, the 4-byte big-endian size is read with `charCodeAt` at [api.js:126](``https://github.com/openwrt/luci/blob/c90e845a763c0265df3eecbccce86ef1912a582f/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js#L126).`` Any payload of 128 bytes or more puts a byte >= 0x80 into that size field; `response.text()` decodes it as invalid UTF-8 and substitutes U+FFFD, so `charCodeAt(7)` yields 65533 instead of e.g. 200. `parse_multiplexed_stream` then trips the `buffer.length < 8 + payload_length` guard and breaks out, returning an empty array — logs containing any line longer than 127 bytes render as nothing. Multi-byte UTF-8 inside the payload desynchronises frame boundaries the same way, because `payload_length` is a byte count while `slice()` counts UTF-16 units.

The [ucode implementation](``https://github.com/openwrt/luci/blob/c90e845a763c0265df3eecbccce86ef1912a582f/applications/luci-app-dockerman/root/usr/share/rpcd/ucode/docker_rpc.uc#L104-L118)`` is unaffected — ucode strings are byte strings and `ord()`/`substr()` are byte-indexed — so this only bites when `preferApi` selects the direct-fetch branch rather than the rpcd one shown in the PR description.

Fixing it needs the raw bytes: branch on the content type before `response.text()` and read this response with `response.arrayBuffer()`, walk the frames over a `Uint8Array`, and `TextDecoder().decode()` each payload slice.

---

applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
```diff
@@ -1685,7 +1685,10 @@ return dm2.dv.extend({
 					return false;
 				}
 
-				const logText = response?.body || _('No logs available');
+				let logText = response?.body || _('No logs available');
+				if ("application/vnd.docker.multiplexed-stream" == response?.headers['content-type']) {
```
openwrt-ai: The optional chain stops at `response`, so a response object without a `headers` property throws a `TypeError` here instead of falling through to the plain-text path. The same file already guards for exactly that case in `readFile` — [container.js:1646](``https://github.com/openwrt/luci/blob/c90e845a763c0265df3eecbccce86ef1912a582f/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js#L1646)`` does `if (!response.headers || response?.headers?.entries?.length == 0)`.

nit: both touched files otherwise use strict `===` for these comparisons (e.g. `headers['Connection']?.toLowerCase() === 'upgrade'` in api.js, and `response_headers['content-type'] === '...'` in docker_rpc.uc), so the yoda `==` is out of step with the surrounding style — same at api.js:247.

```suggestion
				if (response?.headers?.['content-type'] === 'application/vnd.docker.multiplexed-stream') {
```

---

applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
```diff
@@ -1685,7 +1685,10 @@ return dm2.dv.extend({
...
+					logText = (response?.body || []).map((frame)=>frame.payload).join('');
```
openwrt-ai: This unconditionally discards the `_('No logs available')` fallback computed on the line above. When the container has no output, `body` is either `[]` (JS API path) or `''` (rpcd path, where [docker_rpc.uc:297](``https://github.com/openwrt/luci/blob/c90e845a763c0265df3eecbccce86ef1912a582f/applications/luci-app-dockerman/root/usr/share/rpcd/ucode/docker_rpc.uc#L297``) skips the parse for an empty body); both end up as `''` after the join, and `ansiToHtml('')` returns `''`, so the pane just goes blank instead of showing the message.

```suggestion
					logText = (response?.body || []).map((frame) => frame.payload).join('') || _('No logs available');
```

---

openwrt-ai (commented):
Reviewed 1 new commit; see inline comments. The ucode-side parser matches the byte semantics of `ord()`/`substr()` correctly; the JS-side parser in `api.js` operates on a UTF-8-decoded string and does not.

---

applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
```diff
@@ -1685,7 +1685,12 @@ return dm2.dv.extend({
 					return false;
 				}
 
-				const logText = response?.body || _('No logs available');
+				let logText = response?.body;
+				if ("application/vnd.docker.multiplexed-stream" === response?.headers['content-type']) {
```
openwrt-ai: The optional chain stops at `response`, so the bracket access still throws when `response` is truthy but not the `{code, headers, body}` shape — which the rpcd path can produce.

[`container_logs` in common.js:66-70](``https://github.com/openwrt/luci/blob/93ba06d01ebef34d66576aaba0c43504cb67c25b/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/common.js#L66-L70)`` declares the RPC without `reject`, and rpc.js maps that to [`raise: options.reject`](https://github.com/openwrt/luci/blob/93ba06d01ebef34d66576aaba0c43504cb67c25b/modules/luci-base/htdocs/luci-static/resources/rpc.js#L332), so [the non-zero ubus status is not raised](https://github.com/openwrt/luci/blob/93ba06d01ebef34d66576aaba0c43504cb67c25b/modules/luci-base/htdocs/luci-static/resources/rpc.js#L93-L97) and the promise resolves with the bare status *number* (`ret = msg.result[0]`) — e.g. 6 on permission denied, or when the `docker.container` object is missing. `response?.code >= 300` is then `undefined >= 300` (false), execution reaches this line, `response?.headers` is `undefined`, and `['content-type']` raises `TypeError: Cannot read properties of undefined`. The user gets that message in the log pane instead of a useful error. The pre-patch `response?.body || _('No logs available')` did not have this failure mode.

```suggestion
				if ("application/vnd.docker.multiplexed-stream" === response?.headers?.['content-type']) {
```

---

applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
```diff
@@ -116,6 +116,23 @@ function processLines(buffer, onChunk) {
...
+	const utf8decoder = new TextDecoder();
+
+	while (buffer.length > 0) {
+		if (buffer.length < 8) break; // Not enough data for header
+		let stream_type = buffer[0];
+		let payload_length = ((buffer[4] << 24) | (buffer[5] << 16) | (buffer[6] << 8) | buffer[7]) >>> 0; // Convert to unsigned
+		if (buffer.length < 8 + payload_length) break; // Not enough data for payload
+		let payload = buffer.slice(8, 8 + payload_length);
+		output.push({ type: stream_type, payload: utf8decoder.decode(payload) });
+		buffer = buffer.slice(8 + payload_length);
```
openwrt-ai: nit: `Uint8Array.prototype.slice` copies, so advancing the cursor with `buffer = buffer.slice(...)` re-copies the entire remaining log on every frame — quadratic in frame count. Docker emits one frame per write, so a `tail=1000` fetch means ~1000 copies of a shrinking buffer. `subarray()` returns a view over the same `ArrayBuffer` with no copy, and `TextDecoder.decode()` accepts a view just as well, so both call sites can use it.

```suggestion
		let payload = buffer.subarray(8, 8 + payload_length);
		output.push({ type: stream_type, payload: utf8decoder.decode(payload) });
		buffer = buffer.subarray(8 + payload_length);
```

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `93ba06d`, diffed against `c90e845`). All three findings from the previous round are addressed: `parse_multiplexed_stream` now operates on a `Uint8Array` from `response.arrayBuffer()` with a `TextDecoder` per payload, so byte offsets and frame boundaries are correct for payloads ≥ 128 bytes and for multi-byte UTF-8; the `>>> 0` on the length field is right; and the `No logs available` fallback is restored after the frame join. Two remaining inline notes, one of them a nit.

---

applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
```diff
@@ -116,6 +116,23 @@ function processLines(buffer, onChunk) {
...
+	const output = [];
+	const utf8decoder = new TextDecoder();
+
+	while (buffer.length > 0) {
+		if (buffer.length < 8) break; // Not enough data for header
+		let stream_type = buffer[0];
+		let payload_length = ((buffer[4] << 24) | (buffer[5] << 16) | (buffer[6] << 8) | buffer[7]) >>> 0; // Convert to unsigned
+		if (buffer.length < 8 + payload_length) break; // Not enough data for payload
+		let payload = buffer.subarray(8, 8 + payload_length);
+		output.push({ type: stream_type, payload: utf8decoder.decode(payload) });
```
openwrt-ai: nit: decoding each frame with a non-streaming `TextDecoder` mangles a UTF-8 sequence that straddles a frame boundary. Docker's json-file logger splits a single log message into multiple frames at a fixed byte size, and the split is not character-aware, so a multi-byte character can end up with its lead byte in frame N and its continuation bytes in frame N+1. `decode()` without `{ stream: true }` flushes at the end of each call, so both halves become U+FFFD — one visible garbled character per split, in non-ASCII logs long enough to be split.

The rpcd path doesn't have this failure mode: [`parse_multiplexed_stream` in docker_rpc.uc:103-116](``https://github.com/openwrt/luci/blob/a3151882840a763dbf46d690b9ffe6e03202dc70/applications/luci-app-dockerman/root/usr/share/rpcd/ucode/docker_rpc.uc#L103-L116)`` keeps payloads as byte strings and [container.js:1690](``https://github.com/openwrt/luci/blob/a3151882840a763dbf46d690b9ffe6e03202dc70/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js#L1690)`` rejoins them before rendering, so the character is reassembled intact.

Since the consumer joins all payloads anyway, the smallest fix is to make the decode streaming (`utf8decoder.decode(payload, { stream: true })`) so a truncated tail sequence is carried into the next frame — though that does mix state across stdout/stderr frames, so if the per-`type` split is meant to stay independent, a decoder per stream type would be the accurate form. Your call which is worth the complexity here.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `a315188`, diffed against `93ba06d`). Both notes from the previous round are addressed: the `?.['content-type']` optional chain no longer throws when the rpcd path resolves with a bare ubus status number, and the frame walk now advances with `subarray()` views instead of copying `slice()`, removing the quadratic re-copy. Commit message matches the change; CI on the head SHA is green or still running. One low-severity inline note left.

---

applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
```diff
@@ -116,6 +116,28 @@ function processLines(buffer, onChunk) {
...
+	const decoders = [new TextDecoder(), new TextDecoder(), new TextDecoder()];
+
+	while (buffer.length > 0) {
+		if (buffer.length < 8) break; // Not enough data for header
+		let stream_type = buffer[0];
+		let payload_length = ((buffer[4] << 24) | (buffer[5] << 16) | (buffer[6] << 8) | buffer[7]) >>> 0; // Convert to unsigned
+		if (buffer.length < 8 + payload_length) break; // Not enough data for payload
+		if (stream_type >=0 && stream_type <= 2) {
+			let payload = buffer.subarray(8, 8 + payload_length);
+			output.push({ type: stream_type, payload: decoders[stream_type].decode(payload, {stream: true}) });
```
openwrt-ai: nit: `stream_type >= 0` can never be false — `stream_type` is `buffer[0]` where `buffer` is the `Uint8Array` built at [api.js:250](``https://github.com/openwrt/luci/blob/c13ebfe10c14449217291692d8cdfcd73a10177a/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js#L250),`` so element reads are already clamped to 0–255. Only the upper bound does any work.

Two spacing points in the same hunk: `>=0` is missing the space the rest of the file uses, and `{stream: true}` drops the inner padding that both the sibling object literal on that same line (`{ type: ... }`) and [api.js:196](``https://github.com/openwrt/luci/blob/c13ebfe10c14449217291692d8cdfcd73a10177a/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js#L196)`` (`decoder.decode(value, { stream: true })`) use.

```suggestion
		if (stream_type <= 2) {
			let payload = buffer.subarray(8, 8 + payload_length);
			output.push({ type: stream_type, payload: decoders[stream_type].decode(payload, { stream: true }) });
```

---

applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
```diff
@@ -116,6 +116,28 @@ function processLines(buffer, onChunk) {
 	return buffer;
 }
 
+// Parse the docker multiplexed stream format, https://docs.docker.com/reference/api/engine/version/v1.43/#stream-format
+function parse_multiplexed_stream(buffer) {
+	const output = [];
+	// Only three stream types are defined: 0=stdin, 1=stdout, 2=stderr.
+	const decoders = [new TextDecoder(), new TextDecoder(), new TextDecoder()];
```
openwrt-ai, line 123: nit: now that every `decode()` call passes `{ stream: true }`, no decoder is ever flushed, so any bytes a decoder is still holding when the loop exits are dropped silently rather than surfacing as U+FFFD.

Two ways to get there: the `buffer.length < 8` / `buffer.length < 8 + payload_length` guards break out with a truncated tail frame whose earlier sibling frame ended mid-sequence, and a well-formed stream whose final frame ends on an incomplete UTF-8 sequence. In both cases the partial character just disappears from the log pane. The whole body is read in one shot at [api.js:249-250](``https://github.com/openwrt/luci/blob/c13ebfe10c14449217291692d8cdfcd73a10177a/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js#L249-L250),`` so there is no later call that would pick the held bytes back up — the decoders are recreated per invocation.

Low impact given Docker's json-file driver emits whole messages, but a terminating flush after the `while` (an empty `decoders[i].decode()` per stream type, appending anything non-empty) would restore the usual TextDecoder contract. Leaving the form to you since it depends on whether you want the flushed remnant attributed to a particular frame or just appended.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `c13ebfe`, diffed against `a315188`). The remaining note from the previous round is addressed, and in the more accurate of the two forms offered: per-stream-type `TextDecoder`s with `{ stream: true }`, so a UTF-8 sequence split across frames is reassembled without mixing state between stdout and stderr. Commit message still matches the change, and all eight checks on the head SHA are green.

Two nits left, both in `parse_multiplexed_stream`.

---

applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/api.js
jjm2473: `parse_multiplexed_stream` does not make multiple calls to the appended log; instead, it requests a specific number of lines each time and renders only those lines.

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `45109fc`, diffed against `c13ebfe`); no new issues found.

The delta is exactly the two nits from the previous round: the dead `stream_type >= 0` half of the range test is gone, and `{ stream: true }` now carries the inner padding used elsewhere in the file. The arrow-function reformat in `container.js:1690` is cosmetic and matches the surrounding style. Commit message still matches the change, and all eight checks on the head SHA are green.

Also noting your reply on the trailing-flush thread — with the whole body read in one shot and a fresh set of decoders per call, the remnant only matters for a stream whose final frame ends mid-sequence, which Docker's json-file driver won't produce. Happy to leave it.

---

jjm2473:
ping @systemcrash

BKPepe (changes requested):
The code change itself makes sense, but the commit message needs improvement.

It currently says what the commit adds, but it does not explain why the change is needed. The commit history should document the problem being fixed, not just describe the implementation.

What about something like this?
```
luci-app-dockerman: handle Docker multiplexed stream logs

Docker API can return container logs using the
application/vnd.docker.multiplexed-stream format. The response consists
of framed stdout/stderr records rather than plain text.

Currently these responses are not decoded correctly, resulting in
garbled or missing log output.

Parse the multiplexed stream format in both the ucode RPC handler and the
LuCI API client, then combine the frame payloads before passing the logs
to the existing ANSI-to-HTML renderer.
```

openwrt-ai (commented):

---

PR #8977, opened 2026-08-24
luci-app-dockerman: fix whitespace rendering in container logs

jjm2473:
Container logs contain significant whitespace, which gets collapsed by
normal HTML rendering, and that "white-space: pre-wrap;" preserves
that whitespace while still allowing long lines to wrap.

### Description

Fix logs missing spacebar on render.

Before:
image

After:
image

### Maintainer _(preferred)_

@systemcrash 

---

## Tested on

**Web browser(s):** Chrome 151.0.7922.76 

---

## Checklist

### DIFF
--- a/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
+++ b/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
@@ -1190,7 +1190,7 @@ return dm2.dv.extend({
 				]),
 				E('div', {
 					'id': 'container-logs-text',
-					'style': 'width: 100%; font-family: monospace; padding: 10px; border: 1px solid #ccc; overflow: auto;',
+					'style': 'width: 100%; font-family: monospace; padding: 10px; border: 1px solid #ccc; overflow: auto; white-space: pre-wrap;',
 					'innerHTML': ''
 				})
 			]);


applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/common.js
```diff
@@ -1280,13 +1280,14 @@ const ansiToHtml = function(text) {
...
 		const map = {
+			' ': '&nbsp;',
 			'&': '&amp;',
 			'<': '&lt;',
 			'>': '&gt;',
 			'"': '&quot;',
 			"'": '&#039;'
 		};
-		return str.replace(/[&<>"']/g, m => map[m]);
+		return str.replace(/[&<>"' ]/g, m => map[m]);
```
openwrt-ai: Escaping *every* space as `&nbsp;` fixes the symptom but introduces a wrapping regression, and it does so for all three callers of `ansiToHtml()`, not just the logs pane.

`&nbsp;` is a non-breaking space, so the browser will no longer break a line at any space in the log output. Long log lines (stack traces, `docker pull` progress, JSON payloads) will stop wrapping and overflow horizontally instead — in the logs pane that turns into a horizontal scrollbar, since the container is `overflow: auto`.

The actual cause is that the logs container has no `white-space` rule, so it renders with `white-space: normal` and collapses runs of spaces / drops leading indentation:

[`#container-logs-text` style at container.js:1193](``https://github.com/openwrt/luci/blob/70abada9373267a5eacb88c80dde124541c6d19c/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js#L1193)``

```
'style': 'width: 100%; font-family: monospace; padding: 10px; border: 1px solid #ccc; overflow: auto;',
```

Adding `white-space: pre-wrap;` there preserves spaces *and* keeps wrapping — which is exactly what the sibling console pane already does, [`#ws-console-output` at container.js:1129](``https://github.com/openwrt/luci/blob/70abada9373267a5eacb88c80dde124541c6d19c/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js#L1129).`` Note that because that pane already sets `pre-wrap`, its spaces render correctly today, so this patch is a pure regression for it (and for the build-progress `msgTxt` element in this file at line 1105).

Two more things this approach leaves behind: tabs are collapsed by `white-space: normal` too and are not covered here, and the helper is still commented "Escape HTML special characters" while a space is not an HTML special character.

Suggested direction: drop the `escapeHtml` change and add `white-space: pre-wrap;` to the `#container-logs-text` style instead.

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline issue on the approach taken (see comment). Commit message matches its diff, and CI on `70abada` is green.

---

applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
```diff
@@ -1190,7 +1190,7 @@ return dm2.dv.extend({
 				]),
 				E('div', {
 					'id': 'container-logs-text',
-					'style': 'width: 100%; font-family: monospace; padding: 10px; border: 1px solid #ccc; overflow: auto;',
+					'style': 'width: 100%; font-family: monospace; padding: 10px; border: 1px solid #ccc; overflow: auto; white-space: pre-wrap;',
```
openwrt-ai, line 1193: nit: with this change two of the three `ansiToHtml()` sinks set `white-space: pre-wrap` — this one and [`#ws-console-output` at container.js:1449](``https://github.com/openwrt/luci/blob/329f2840833cb913e07ae62db2768679013e01b4/applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js#L1449).`` The third one still doesn't: the image pull/build progress line is a bare `<p>`, [`msgTxt` at common.js:1066](``https://github.com/openwrt/luci/blob/329f2840833cb913e07ae62db2768679013e01b4/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/common.js#L1066),`` fed by [`msgTxt.innerHTML = ansiToHtml(msg?.stream)` at common.js:1105](``https://github.com/openwrt/luci/blob/329f2840833cb913e07ae62db2768679013e01b4/applications/luci-app-dockerman/htdocs/luci-static/resources/dockerman/common.js#L1105),`` so `docker build` step output shown there still collapses runs of spaces and drops leading indentation for exactly the same reason this pane did.

Is leaving that one out intentional (progress lines are single-line, so indentation doesn't matter much), or worth the same one-property fix while you're here?

---

openwrt-ai (commented):
Reviewed 1 new commit. The `&nbsp;` escaping approach flagged last round has been replaced with `white-space: pre-wrap;` on the logs container, which addresses the wrapping regression and the tab-collapsing gap as well — `ansiToHtml()` already rewrites `\r\n`/`\r`/`\n` to `<br/>` before the string reaches the DOM, so `pre-wrap` won't double up line breaks. Commit message matches its diff. One nit inline; CI on `329f284` is green (x86_64 build still running).

---

applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
jjm2473: Not related.

jjm2473:
ping @systemcrash

applications/luci-app-dockerman/htdocs/luci-static/resources/view/dockerman/container.js
BKPepe: Even though, it is not related, it should be fixed or at least there should be issue what needs to be done.

jjm2473: @BKPepe 
I tested pulling a Docker image, and there was always just a single line of logs and a progress bar; the indentation shouldn't be an issue.
image

BKPepe (changes requested):
The code change makes sense, but the commit message does not explain the actual problem being fixed.

Saying that this was "recommended by AI" is not useful as commit history. Please describe what is wrong with the log rendering and why white-space: pre-wrap fixes it.

For example, the message could explain that container logs contain significant whitespace, which gets collapsed by normal HTML rendering, and that pre-wrap preserves that whitespace while still allowing long lines to wrap.

The commit should document the technical reason for the change, not where the suggestion came from.

openwrt-ai (commented):
## Commit checks
- `61f8734f` "luci-app-dockerman: fix whitespace rendering in container logs" — the body's second clause is a dangling fragment (`..., and that "white-space: pre-wrap;" preserves ...`), because the "the message could explain that" framing it was copied from was dropped. Make it two sentences, e.g. "Container logs contain significant whitespace, which gets collapsed by normal HTML rendering. `white-space: pre-wrap` preserves that whitespace while still allowing long lines to wrap."

---

openwrt-ai (commented):

---

Issue #8961, opened 2026-08-21
Feature request: luci-theme-footstrap: show hostname on login screen


GoetzGoerisch:
### What would you like to see in luci?

@VizzleTF Thank you for this great theme! 

Coming from `Material` as my default I'm missing the hostname display on the login screen. This is missing in `bootstrap` as well.

Request:

Add hostname display to login screen.

Thank you for consideration

VizzleTF:
Hello @GoetzGoerisch 
Thank you for feedback. Will be in next version

GoetzGoerisch:
Thank you.

One additional nit the icon of a time based FW rule does not convey the message.

<img width="278" height="67" alt="Image" src="https://github.com/user-attachments/assets/8b3e7138-2af6-49cf-af58-1402b7042e5d" />

BKPepe:
Closing as it should be fixed by this https://github.com/openwrt/luci/pull/8981

PR #8903, opened 2026-08-03
luci-theme-footstrap: add new theme

VizzleTF:
### Description

Add Footstrap, a standalone theme for OpenWrt 24.10 and newer. It depends on `luci-base` and nothing else and ships no CSS framework; pages are rendered by luci-base's own view JS.

- Sidebar or top bar, switched by the user and remembered per browser.
- Light and dark modes, palettes, density and colour axes, applied before the first paint.
- Search over the menu tree.
- Client-side navigation that swaps the view instead of reloading the page. It falls back to a full load when a page's stylesheet would otherwise carry over.
- Settings page built from `.cbi-value` rows, `ui.Select` and `ui.RangeSlider`.

Developed at https://github.com/VizzleTF/luci-theme-footstrap

Navigation is client-side: a click swaps the view rather than reloading the document, so a page opens without the full-load round trip. Measured against bootstrap on a Netcore N60 Pro running 25.12.2, five runs, medians:

| | bootstrap | footstrap |
|---|---|---|
| 38-page tour | 11 306 ms | 4933 ms |
| router CPU time | 37.3 s | 18.4 s |
| median page | | 3.03× faster |

Method and per-page results:
https://github.com/VizzleTF/luci-theme-footstrap/blob/main/docs/benchmark.md

Discussed on the forum since July:
https://forum.openwrt.org/t/footstrap-new-theme-for-luci-2x-faster-than-bootstrap/251930

#### Desktop
overview-desktop-dark
appearance-desktop-light

#### Mobile
network-mobile-dark

### Maintainer _(preferred)_

@VizzleTF

---

## Tested on
**OpenWrt version:** OpenWrt 25.12.4 and OpenWrt 24.10.8 (aarch64_generic containers), plus OpenWrt 25.12.2 (r32802-f505120278) on a Netcore N60 Pro
**LuCI version:** this branch, built with the x86_64 master and 24.10 SDKs
**Web browser(s):** Chromium 151, Firefox 145, Safari 26
And a single release is downloaded 200–240 times

PR #8978, opened 2026-08-24
luci-theme-footstrap: keep the reader's place on every engine

VizzleTF:
A LuCI poll refreshes a section with `dom.content()`, which empties the container before it refills it. For that moment the document is shorter than the offset the reader sits at, the engine clamps the offset, and what happens on the way back is the engine's business: Chromium lands where it started, WebKit overshoots by 60px on every tick. On a live Safari that is the page creeping while you read it.

The theme decided whether to help by asking `CSS.supports("overflow-anchor")`, which answers "does this engine anchor at all" — a different question, and one every current engine now answers yes to. It measures instead: two frames after the mutation, the reference it was already holding is asked where it ended up. An engine that got it right reports zero drift and nothing happens; what is left over is what nobody put back.

Three more faults sat behind the same symptom, all in *which element* the reference is taken on — `elementFromPoint` answering with `#view` itself wherever the hit lands in a gap, a point above the first section answering with `.fs-content` outside the host, and a page that is one table (Processes, Routes, the realtime lists) having that table as a direct child of `#view`, so the climb out of it landed on the host and gave up. On that last shape nothing was holding the reader on **any** engine.

Also in this sync, each reported and each fixed on a luci-base class rather than on the page it was seen on:

* a meter's value sat on its own label once its column became a card (luci-mod-dashboard's Wireless list on a phone, 6–9px at every density);
* a page-title button row touched the heading above it (Status → Channel Analysis, 0px measured);
* the per-section Delete button touched the tab bar below it — every named section: SQM's queues, the firewall's zones;
* a block a view builds itself fused with the card below it (a view may return a bare widget where a section is expected, as luci-app-irqbalance does for its `/proc/interrupts` snapshot). Stock bootstrap measures the same 0px there; the difference is that a section is a stretch of page in bootstrap and a card here.

And two from before this sync: the login page prints the hostname it belongs to (#8961) and carries a top-level heading, and the Appearance tab builds its own range control where `ui.RangeSlider` is missing, which is 23.05 — the whole tab died there on one missing widget.

Measured on OpenWrt 25.12 and 24.10 containers, chromium/firefox/webkit, both layouts, 390 and 1440, all three density settings: 216 anchoring runs with the reader staying put, and the page audit reporting no new findings.

### DIFF
--- a/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
+++ b/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
@@ -295,6 +295,7 @@ to{opacity:1;transform:none}
 [data-darkmode="true"] [data-page="admin-statistics-graphs"] [data-plugin] img{filter:invert(100%) hue-rotate(180deg)}
 #view div[style] > svg{background-color:var(--fs-panel)}
 #view div[style] > svg line[style]{stroke:var(--fs-text) !important}
+#view div[style] > svg text[style]{fill:var(--fs-text) !important;text-shadow:none !important;font-weight:var(--fs-weight-bold)}
 }
 @layer theme{:where([data-fs-chrome]:not([data-fs-chrome] *)){font-style:normal;font-variant:normal;letter-spacing:normal;word-spacing:normal;text-transform:none;text-indent:0;text-align:start;cursor:auto}
 .fs-brand{display:flex;align-items:center;gap:var(--fs-space-2-5);color:var(--fs-text);font-weight:var(--fs-weight-bold)}
@@ -465,6 +466,8 @@ footer a:hover{color:var(--fs-accent)}
 .cbi-value-field > .cbi-progressbar::after{inset-inline:calc(100% + var(--fs-space-2)) auto;bottom:auto;top:50%;transform:translateY(-50%)}
 .cbi-section .table:has(.tr.table-titles) .td .cbi-progressbar{margin-inline-end:var(--fs-meter-value);--fs-meter-value:var(--fs-space-10);min-width:60px}
 .cbi-section .table:has(.tr.table-titles) .td .cbi-progressbar::after{inset-inline:calc(100% + var(--fs-space-2)) auto;bottom:auto;top:50%;transform:translateY(-50%)}
+.cbi-section .table.fs-stacked:has(.tr.table-titles) .td .cbi-progressbar{margin-inline-end:0;margin-block-start:calc(var(--fs-type-xs) * var(--fs-leading))}
+.cbi-section .table.fs-stacked:has(.tr.table-titles) .td .cbi-progressbar::after{inset-inline:auto 0;bottom:calc(100% + var(--fs-space-1));top:auto;transform:none}
 .cbi-section .table:not([id]):not(.cbi-section-table) .tr:has(.cbi-progressbar) .td{vertical-align:middle}
 }
 @layer theme{.cbi-section,.cbi-map > .cbi-section-node,fieldset.cbi-section,.panel{background:var(--fs-panel);border:var(--fs-hairline);border-radius:var(--fs-radius-lg);box-shadow:var(--fs-shadow)}
@@ -473,6 +476,8 @@ legend{font-size:var(--fs-type-lg);font-weight:var(--fs-weight-bold);color:var(-
 .cbi-section > .cbi-title{margin-bottom:var(--fs-space-3)}
 .cbi-section > .cbi-title h3,.cbi-section > h3:first-child,.cbi-section > legend{font-size:var(--fs-type-lg);font-weight:var(--fs-weight-bold);color:var(--fs-text);margin:0;padding:0;border:0}
 .cbi-section > h3:first-child,.cbi-section > legend{margin-bottom:var(--fs-space-3)}
+.cbi-section .table + h3{margin-top:var(--fs-space-5)}
+.cbi-section h3 + .table{margin-top:var(--fs-space-3)}
 .cbi-section .table:not([id]):not(.fs-dt),.cbi-section-node .table:not([id]):not(.fs-dt),.cbi-modal .table,.table .table,.cbi-value .table{border:0;border-radius:0;overflow:visible;background:none}
 .cbi-section .table{margin:0;width:100%}
 .table.cbi-section-table input[type="password"],.table.cbi-section-table input[type="text"],.table.cbi-section-table textarea,.table.cbi-section-table select,.table.cbi-section-table .cbi-select{width:100%;min-width:auto}
@@ -634,7 +639,9 @@ h6{font-size:var(--fs-type-xs);color:var(--fs-dim);text-transform:uppercase}
 .cbi-map > h2:has(+ .cbi-map-descr),#view > h2:has(+ .cbi-map-descr){border-bottom:0;border-radius:var(--fs-radius-lg) var(--fs-radius-lg) 0 0;padding-bottom:var(--fs-space-1);margin-bottom:0;box-shadow:none}
 .cbi-map > h2 + .cbi-map-descr,#view > h2 + .cbi-map-descr{background:var(--fs-panel);border:var(--fs-hairline);border-top:0;border-radius:0 0 var(--fs-radius-lg) var(--fs-radius-lg);box-shadow:var(--fs-shadow);margin:0 0 var(--fs-card-gap);padding:0 var(--fs-space-4) var(--fs-space-4);color:var(--fs-dim);font-size:var(--fs-type);line-height:var(--fs-leading)}
 #view :is(textarea,input[type="text"],input[type="password"],.cbi-dropdown) + .right{margin-top:var(--fs-space-2-5)}
-.cbi-title-buttons{margin-bottom:var(--fs-card-gap)}
+.cbi-title-buttons{margin-block:var(--fs-space-2) var(--fs-card-gap)}
+.cbi-section-remove{margin-bottom:var(--fs-space-2)}
+.cbi-map > *:not(.cbi-section):not(.cbi-map-descr):not(.cbi-page-actions):not(h2):not(legend):not(.cbi-tabmenu):not(.cbi-map-tabbed):not(:empty){margin-bottom:var(--fs-card-gap)}
 }
 @layer theme{:root[data-layout="top"] .fs-sidebar{min-height:var(--fs-bar-h);height:auto;gap:var(--fs-space-3);flex-wrap:nowrap;padding:0 max(20px,calc((100% - var(--fs-content-max)) / 2 + var(--fs-content-pad)));box-shadow:var(--fs-shadow-bar)}
 :root[data-layout="top"] .fs-sidebar ul.nav{order:2;flex:1 1 auto;margin-inline-start:var(--fs-space-2)}
@@ -947,9 +954,10 @@ input,textarea,select,.cbi-dropdown > ul > li{font-size:var(--fs-type-lg)}
 }
 }
 @layer page{.fs-login{width:min(400px,calc(100% - 32px));margin:7vh auto var(--fs-space-10);background:var(--fs-panel);border:var(--fs-hairline);border-radius:var(--fs-radius-lg);padding:var(--fs-space-6) var(--fs-space-6) var(--fs-space-6);box-shadow:var(--fs-shadow-pop)}
+.fs-login-host{margin:0 0 var(--fs-space-6);text-align:center;font-size:var(--fs-type-lg);font-weight:var(--fs-weight-bold);color:var(--fs-dim);overflow-wrap:anywhere}
 .fs-login .cbi-map,.fs-login .cbi-section,.fs-login .cbi-section-node{background:none;border:0;padding:0;margin:0;box-shadow:none}
-.fs-login .cbi-map > h2,.fs-login .cbi-map-descr{background:none;border:0;border-radius:0;box-shadow:none;padding:0}
-.fs-login .cbi-map > h2{font-size:var(--fs-type-xl);font-weight:var(--fs-weight-bold);letter-spacing:-.01em;margin:0 0 var(--fs-space-1-5)}
+.fs-login .cbi-map > h1,.fs-login .cbi-map-descr{background:none;border:0;border-radius:0;box-shadow:none;padding:0}
+.fs-login .cbi-map > h1{font-size:var(--fs-type-xl);font-weight:var(--fs-weight-bold);letter-spacing:-.01em;margin:0 0 var(--fs-space-1-5)}
 .fs-login .cbi-map-descr{color:var(--fs-dim);font-size:var(--fs-type);margin:0 0 var(--fs-space-5)}
 .fs-login .cbi-value{display:block;border:0;padding:0 0 var(--fs-space-4);margin:0}
 .fs-login .cbi-value-title{display:block;text-align:start;width:auto;max-width:none;padding:0 0 var(--fs-space-1-5);font-size:var(--fs-type-xs);font-weight:var(--fs-eyebrow-weight);letter-spacing:var(--fs-eyebrow-tracking);text-transform:uppercase;color:var(--fs-eyebrow-color)}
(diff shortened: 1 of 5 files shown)


themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js
```diff
@@ -259,12 +277,26 @@ function noteMotion() {
...
+}
+
 (function watchMotion() {
 	const opts = { passive: true, capture: true };
 	window.addEventListener('scroll', noteMotion, opts);
-	window.addEventListener('wheel', noteMotion, opts);
-	window.addEventListener('touchstart', noteMotion, opts);
-	window.addEventListener('touchmove', noteMotion, opts);
+	for (const name of [ 'wheel', 'touchstart', 'touchmove', 'mousedown', 'keydown' ])
+		window.addEventListener(name, noteUser, opts);
```
openwrt-ai, line 299: Routing `mousedown` and `keydown` through `noteUser()` also routes them into `noteMotion()`, which unconditionally sets `_movingUntil = Date.now() + SCROLL_IDLE`. So `scrolling()` now answers "yes" for 400ms after *any* click and after *every* keystroke, and that answer gates a lot more than `lateDrift()`:

- [`holdFloor()`](https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L127) — the content floor stops being re-measured
- [`rememberRest()`](https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L426) — the reference stops being refreshed
- [`anchorRef()`](https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L532) and the new gate in `applyAnchor()` — no correction at all
- every fitter that consults `scrolling()` and calls `deferMeasurement()`

Typing into a form re-arms the 400ms window on each keypress, so on a page with inputs the fit passes are deferred for as long as the user is typing and only land in one burst afterwards — which is the "whole deferred pass in one burst" failure the comment above `SCROLL_IDLE` describes.

The comment's justification for the two new events is the scrollbar thumb and Page Down, but both of those produce real `scroll` events, and `scroll` is already wired to `noteMotion` on the line above. Only `_userUntil` needs them:

```suggestion
let _userUntil = 0;
function noteIntent() {
	_userUntil = Date.now() + SCROLL_IDLE;
}
function noteUser() {
	noteIntent();
	noteMotion();
}

(function watchMotion() {
	const opts = { passive: true, capture: true };
	window.addEventListener('scroll', noteMotion, opts);
	for (const name of [ 'wheel', 'touchstart', 'touchmove' ])
		window.addEventListener(name, noteUser, opts);
	for (const name of [ 'mousedown', 'keydown' ])
		window.addEventListener(name, noteIntent, opts);
```


---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js
```diff
@@ -517,6 +626,58 @@ function anchorEnabled() {
...
+	_lateFrame = requestAnimationFrame(() => {
+		_lateFrame = requestAnimationFrame(() => {
+			_lateFrame = 0;
+			/* NOT `scrolling()`: the engine's own compensation moves the offset and therefore starts
+			 * the motion sampler, so gating on that skipped every tick this exists for (measured —
+			 * the check reported "busy" on all of them). Two narrower questions instead: is the
+			 * reader driving (a gesture), and is the offset STREAMING (moving repeatedly, which a
+			 * one-shot compensation never does but a scroll always does). Either one means hands off. */
+			if (!anchorEnabled() || Date.now() < _userUntil || (scrolling() && _steps > 1)) return;
+			if (!ref.el.isConnected || _restPage !== pageStamp()) return;
```
openwrt-ai, line 668: `lateDrift()` bails out on `!ref.el.isConnected`, but `ref` is `_rest`, which this PR just taught to carry a surviving fallback — and the reason it was added is the case that happens here. [`anchorRef()`'s own comment](``https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L586-L588``) says `dom.content()` replaces a section's children, "so the element the hit landed on is usually gone by the time the correction runs", and [`anchorFor()`](``https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L476-L483``) accordingly falls back to `_rest.sec` / `_rest.secTop`.

`anchorFor()` is now only reachable on the non-anchoring path (see the `ENGINE_ANCHORS ? null : anchorFor()` call site), so the `sec` fallback never runs on any engine that anchors — i.e. on every current engine, per the commit message. On the WebKit overshoot this PR is fixing, the poll tick that replaces the section's children is exactly the tick where `ref.el` is disconnected and the residual correction returns without doing anything.

Should this take the same fallback, e.g. use `ref.sec`/`ref.secTop` when `ref.el` is gone but `ref.sec` is still connected?

---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js
```diff
@@ -517,6 +626,58 @@ function anchorEnabled() {
...
+			const drift = ref.el.getBoundingClientRect().top - ref.top;
+			if (Math.abs(drift) < 1) return;			/* the engine put it back */
+			if (Math.abs(drift) > (window.innerHeight || 800)) return;
+			const sc = scroller();
+			const at = sc ? sc.scrollTop : window.scrollY;
+			if (sc) sc.scrollTop = at + drift; else window.scrollTo(0, at + drift);
+			/* the write may have been clamped short; the reader is where they are now */
+			_restAt = scrollTop();
```
openwrt-ai, line 676: The write moves the offset by `drift`, but only `_restAt` is brought forward — `_rest.top` and `_rest.at` still describe the page as it stood *before* the correction. `run()` re-took `_rest` two frames earlier (via [`rememberRest()` in `run()`](``https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L93)),`` so the object the *next* tick captures as `settled` is off by exactly the amount that was just paid, and `lateDrift()` measures it again.

That is the failure [`scheduleAnchor()` explicitly guards against on the other path](``https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L691-L694):`` *"AFTER the correction, never before … or the next tick measures a drift that has already been paid and pays it twice."* The `ENGINE_ANCHORS` path has no equivalent step.

Note that a plain `rememberRest()` call here would not work either — the write itself starts the motion sampler, so `rememberRest()` would return early on its `scrolling()` guard. Refreshing `_rest.top`/`_rest.at` in place from the post-write rects would, if you want to keep the reference rather than drop it.

---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js
```diff
@@ -588,9 +756,28 @@ function observeContent() {
...
+		 * scroll WebKit hands back the new `scrollTop` before the layout that goes with it. The drift
+		 * then measures the reader's own move and the correction undoes it — measured on the 24.10
+		 * stand at 1440 in the top layout at Compact: the page went back to 0 from 591, every run.
+		 * What that engine needs is the RESIDUE, two frames later, once its own anchoring has run
+		 * and the layout is settled — which is lateDrift().
+		 *
+		 * Where it does not anchor, nobody else will put the reader back within the frame, so the
+		 * immediate correction stays, measured against the reference from the last still page. */
+		const settled = _rest;
+		const ref = ENGINE_ANCHORS ? null : anchorFor();
```
openwrt-ai, line 772: nit: this is now the only call site of `anchorFor()`, and it guards on `ENGINE_ANCHORS` before calling — which makes [`anchorFor()`'s own first line, `if (ENGINE_ANCHORS) return anchorRef();`](``https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L451),`` unreachable. Worth dropping it (and adjusting the function's doc comment, which still describes returning "the post-mutation one everywhere else"), otherwise the next reader has to re-derive that the branch is dead.

---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -972,20 +980,18 @@ body[data-page="admin-status-overview"] #view > h2[name="content"],body[data-pag
...
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:nth-child(3){order:1;flex:0 0 100%;width:100%;margin:var(--fs-space-1) 0 var(--fs-space-1-5)}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:first-child{flex:0 0 auto;width:100%;display:flex;align-items:center;justify-content:space-between;gap:var(--fs-space-1-5);font-size:var(--fs-type);font-weight:var(--fs-weight);color:var(--fs-text)}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:first-child::after{content:"";width:8px;height:8px;border-radius:50%;background:var(--fs-good);flex:0 0 auto}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:nth-child(3){order:1;flex:0 0 100%;width:100%;margin:var(--fs-space-0-5) 0 var(--fs-space-1)}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:first-child{flex:0 0 auto;width:100%;min-width:0;overflow:hidden;text-overflow:ellipsis;line-height:var(--fs-leading);font-size:var(--fs-type);font-weight:var(--fs-weight);color:var(--fs-text)}
```
openwrt-ai, line 988: `text-overflow:ellipsis` does nothing here: the box has no `white-space:nowrap` and is not a `-webkit-box` with a line clamp, so the name wraps and is silently cut off by `overflow:hidden` with no ellipsis and no line limit at all.

This also contradicts the justification for the new code in `fs-overview.js`, whose comment states *"styles/pages/20-overview.css clamps a port card's name to two lines with an ellipsis"* — the shipped rule clamps to nothing. The theme's own idiom for this is at [`.fs-wordmark`, cascade.css:344](``https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L344):`` `display:-webkit-box;-webkit-box-orient:vertical;-webkit-line-clamp:N;overflow:hidden`.

```suggestion
body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-head:first-child{flex:0 0 auto;width:100%;min-width:0;display:-webkit-box;-webkit-box-orient:vertical;-webkit-line-clamp:2;overflow:hidden;line-height:var(--fs-leading);font-size:var(--fs-type);font-weight:var(--fs-weight);color:var(--fs-text)}
```


---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -972,20 +980,18 @@ body[data-page="admin-status-overview"] #view > h2[name="content"],body[data-pag
...
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(2) br{display:none}
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]):has(img[src*="_down.svg"]) > .ifacebox-body:nth-child(2){color:var(--fs-dim)}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4){order:2;text-align:end;flex:1 0 auto}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4) > .cbi-tooltip-container{font-family:var(--fs-font-mono);font-size:var(--fs-type-xs) !important;color:var(--fs-dim);line-height:var(--fs-leading);white-space:nowrap;text-align:end !important}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4){order:2;text-align:end;flex:1 0 100%}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4) > .cbi-tooltip-container{font-family:var(--fs-font-mono);font-size:calc(round(10px * var(--fs-density-type),1px)) !important;color:var(--fs-dim);line-height:var(--fs-leading-tight);white-space:nowrap;text-align:end !important}
```
openwrt-ai, line 994: `round()` is used unguarded here, which deviates from the convention this file sets for itself: every other `round()` in the sheet lives inside [`@supports (width:round(1px,1px))` at cascade.css:13](``https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L13),`` with a plain `calc(Npx * var(--fs-density-type))` definition preceding it as the fallback.

On an engine without `round()` the whole declaration is invalid and dropped — including the `!important` — so the port-card tooltip inherits its size instead. That is exactly the older-WebKit audience this PR is otherwise widening support for. A preceding plain declaration restores the pattern:

```suggestion
body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4) > .cbi-tooltip-container{font-family:var(--fs-font-mono);font-size:calc(10px * var(--fs-density-type)) !important;font-size:calc(round(10px * var(--fs-density-type),1px)) !important;color:var(--fs-dim);line-height:var(--fs-leading-tight);white-space:nowrap;text-align:end !important}
```


---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-appearance.js
```diff
@@ -124,9 +124,51 @@ function build() {
...
+		render() {
+			this.sliderEl = E('input', {
+				type: 'range', min: this.options.min, max: this.options.max,
+				step: this.options.step || 'any', value: this.value
+			});
+			this.valueEl = E('output', { class: 'cbi-range-slider-value' }, String(this.value));
+			const node = E('div', { class: 'cbi-range-slider' }, [
+				this.sliderEl,
+				this.valueEl,
+				this.options.calcunits ? E('span', { class: 'cbi-range-slider-calc-units' }, this.options.calcunits) : null
```
openwrt-ai, line 156: The polyfill renders the units span whenever `calcunits` is set, but the real widget renders it only when `calculate` produced a value — see [`ui.js:2904`](https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/modules/luci-base/htdocs/luci-static/resources/ui.js#L2904), where `calcunitsEl` is gated on `this.calculatedvalue`, which is `null` unless [`options.calculate` is a function (ui.js:2887)](https://github.com/openwrt/luci/blob/71cedff88098f4706b94e6d34a49bf313979ba38/modules/luci-base/htdocs/luci-static/resources/ui.js#L2887-L2889).

Every `sliderCtl()` caller passes `unit` and no `calculate` (radius `px`, tint `%`, pattern size `px`, pattern strength `%`, photo dim `%`), so the Appearance tab shows a unit suffix on 23.05 and no unit suffix on 24.10/master — the opposite of the "same DOM, same class names … nothing below this line knows which of the two it got" the comment above promises. `.cbi-range-slider-calc-units` is styled at cascade.css:802, so it is visible on the polyfill path.

Worth deciding which behaviour is intended: either drop the span so the two paths match, or (if the unit really should show) put it in a class the theme owns and render it on both paths, so a router that upgrades 23.05 → 24.10 does not lose it.

---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -295,6 +295,7 @@ to{opacity:1;transform:none}
 [data-darkmode="true"] [data-page="admin-statistics-graphs"] [data-plugin] img{filter:invert(100%) hue-rotate(180deg)}
 #view div[style] > svg{background-color:var(--fs-panel)}
 #view div[style] > svg line[style]{stroke:var(--fs-text) !important}
+#view div[style] > svg text[style]{fill:var(--fs-text) !important;text-shadow:none !important;font-weight:var(--fs-weight-bold)}
```
openwrt-ai, line 298: nit: this hunk appears in neither the commit message nor the PR body. The message enumerates its scope carefully (anchoring, meter value in a card, page-title button row, per-section Delete, view-built block gap, login hostname + heading, `RangeSlider` on 23.05), and a global restyle of SVG text in statistics graphs — `fill` and `text-shadow` forced, plus a weight change — is a separate, user-visible change from all of them.

Two other undocumented hunks in the same commit:

- the port-card rework further down (`cascade.css:983-994`): the up/down status dot `::after` is **removed** from `.ifacebox-head:first-child`, `max-width` moves from `none` to `200px`, the grid minimum becomes density-scaled, and the speed line goes from `flex:1 0 auto` to `flex:1 0 100%`. Dropping the status indicator in particular looks like a decision worth a line in the message.
- `nameTooltips()` in `fs-overview.js` — a new native tooltip on every port card.

Per the patch-hygiene convention, either mention these in the commit message or split them out.

---

openwrt-ai (commented):
Reviewed 1 new commit (`71cedff`). CI is green on the head SHA, so nothing below is CI-grounded.

## Commit checks

- 71cedff "luci-theme-footstrap: keep the reader's place on every engine" — the message covers the anchoring rework, the four luci-base spacing fixes, the login page and the 23.05 `RangeSlider`, but three further changes in the same commit are not described anywhere in it or in the PR body: the statistics-graph SVG text restyle, the Overview port-card rework (including removal of the up/down status dot), and the new `nameTooltips()` in `fs-overview.js`. Details inline.

The substantive findings are on the `ENGINE_ANCHORS` path in `fs-fit.js`: `lateDrift()` skips the `sec`/`secTop` fallback that this PR added for the common "the poll replaced the section's children" case, and it corrects the offset without bringing `_rest` forward — the double-payment that `scheduleAnchor()` already documents on the other path. There is also a scope widening in `noteUser()` that makes every click and keystroke read as "the page is moving".


---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-appearance.js
```diff
@@ -124,9 +124,51 @@ function build() {
 		node.addEventListener('widget-change', () => apply(w.getValue()));
 		return node;
 	};
+	/* THE ONE WIDGET THIS THEME CANNOT ASSUME, and the reason 23.05 is a supported release again.
```
BKPepe, line 127: Since when? O_o

VizzleTF: "23.05 is a supported release again" means "23.05 is a supported release again by the theme". 
It means you can install it to 23.05.

BKPepe: I don't think we should add 23.05 support here.

The Footstrap theme was merged only five days ago (https://github.com/openwrt/luci/commit/6f08de768f4ddc3701144680b833773f50252c51 by https://github.com/openwrt/luci/pull/8903) to the master branch. 23.05 is EOL, so I don't see a good reason why we should now deliberately extend a brand-new upstream theme to an unsupported release.

We are upstream here, so I don't think we should add extra compatibility code and take on the maintenance burden for old releases unless there is a really compelling reason to do so. 🤷

VizzleTF: I got many request for that from users. 
I think its be too complicated to maintain 2 versions: upstream here and separate extended in main repo.
Also i don't see problem with that small adaptation.

BKPepe: This commit is already doing a lot of things at once. Most of the commit is code changes and fixes, with a lot of detailed comments explaining individual cases. I don't think we should make it even more complicated just to accommodate an EOL release.

We are the upstream project. We don't need to keep adding compatibility code for old releases just because it happens to be possible. If a release is EOL, I think the default should be that new code targets supported releases unless there is a strong reason otherwise.

So unless there is a specific requirement that makes 23.05 support necessary, I'd rather keep the theme as it was originally merged and not add this compatibility layer.

VizzleTF: Okay. i will drop it.

hnyman:
Yeah, do not complicate things by trying to support the 23.05 release that has been end-of-life for some time now. 
Note that also 24.10 is likely going to be EoL next month...

PR #8979, opened 2026-08-25
luci-app-mwan3: add resolveDefault to rule.js

ckorber:
### Description

rule.js reports an error stating that access to command is denied. This is caused by the call to `fs.exec_direct` for `/usr/libexec/luci-mwan3 ipset dump` in the load function. Dump does not write names on my setup leading to the error.

Thus, hardening the call for cases, when the result is empty, is reasonable.

*Before*
image

*After*
image

### Maintainer _(preferred)_

@systemcrash @feckert 

---

## Tested on

**LuCI version:** luci openwrt-25.12
**Web browser(s):** firefox

---

### DIFF
--- a/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/network/rule.js
+++ b/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/network/rule.js
@@ -8,7 +8,7 @@
 return view.extend({
 	load: function() {
 		return Promise.all([
-			fs.exec_direct('/usr/libexec/luci-mwan3', ['ipset', 'dump']),
+			L.resolveDefault(fs.exec_direct('/usr/libexec/luci-mwan3', ['ipset', 'dump']), ''),
 			uci.load('mwan3')
 		]);
 	},


applications/luci-app-mwan3/root/usr/share/rpcd/acl.d/luci-app-mwan3.json
```diff
@@ -53,7 +53,8 @@
 		"description": "Grant access for luci-app-mwan3 configuration",
 		"read": {
 			"cgi-io": [
-				"exec"
+				"exec",
+				"exec_direct"
```
openwrt-ai: `exec_direct` is a LuCI JS method name, not an rpcd ACL object — so this entry looks like a no-op rather than a fix.

The names under `"cgi-io"` are the cgi-io *applets* whose session access is checked (`exec` for `/cgi-bin/cgi-exec`, `upload`, `download`, `backup`, `read`). [`fs.exec_direct()`](https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/modules/luci-base/htdocs/luci-static/resources/fs.js#L443-L461) POSTs to `L.env.cgi_base + '/cgi-exec'`, i.e. it is served by the very same `exec` applet that is already granted on line 56. Nothing ever asks rpcd for a `cgi-io`/`exec_direct` object, so granting it changes no decision.

Supporting evidence in-tree: no ACL file in this repo grants `exec_direct` (`grep -rn exec_direct --include=*.json .` is empty), while every other app that calls `fs.exec_direct()` grants only `exec` — e.g. [luci-app-nlbwmon.json:5](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/applications/luci-app-nlbwmon/root/usr/share/rpcd/acl.d/luci-app-nlbwmon.json#L5),`` [luci-app-package-manager.json:5](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/applications/luci-app-package-manager/root/usr/share/rpcd/acl.d/luci-app-package-manager.json#L5),`` [luci-mod-network.json:5](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/modules/luci-mod-network/root/usr/share/rpcd/acl.d/luci-mod-network.json#L5).`` The sibling group in this same file ([`luci-app-mwan3-status`, line 5](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/applications/luci-app-mwan3/root/usr/share/rpcd/acl.d/luci-app-mwan3.json#L5``)) also drives `fs.exec_direct()` from [detail.js:7](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/detail.js#L7)`` with `exec` alone.

Since ACL files are only re-read when rpcd reloads, a plausible explanation for "after adding it, it works" is that reinstalling the package/restarting rpcd is what took effect, not the new token. Could you re-test by only restarting rpcd on an unmodified tree?

If the denial does reproduce, the check that is failing is the *second* one — the `"file"` scope match on the full command line on line 59. `fs.exec_direct('/usr/libexec/luci-mwan3', ['ipset', 'dump'])` builds the string `/usr/libexec/luci-mwan3 ipset dump`, which should match, unless `/usr/libexec/luci-mwan3` is a symlink on the test system (rpcd canonicalises the executable path before matching) or the helper is not installed at that path at all. Worth confirming which of the two `Access to command denied` sites fires before changing the ACL.

Separately, and regardless of the ACL question: [rule.js:11](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/network/rule.js#L11)`` is the only `fs.exec_direct()` call in this app that is *not* wrapped in `L.resolveDefault()`, which is why a failure there blanks the whole Rules page instead of degrading gracefully — its siblings [detail.js:7](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/detail.js#L7)`` and [troubleshooting.js:7](``https://github.com/openwrt/luci/blob/66b465cfc5d3a1d95ef7f43ad50321e006a64476/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/troubleshooting.js#L7)`` both wrap it. Hardening that call would fix the reported user-visible symptom (a hard error banner) on any system where `ipset` output is unavailable.

```suggestion
				"exec"
```

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches its diff, but the change itself looks like a no-op — see the inline comment on the ACL entry.

---

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/network/rule.js
```diff
@@ -8,7 +8,7 @@
 return view.extend({
 	load: function() {
 		return Promise.all([
-			fs.exec_direct('/usr/libexec/luci-mwan3', ['ipset', 'dump']),
+			L.resolveDefault(fs.exec_direct('/usr/libexec/luci-mwan3', ['ipset', 'dump']), ''),
```
openwrt-ai, line 11: This makes the page degrade gracefully, which is right and matches the siblings ([detail.js:7](``https://github.com/openwrt/luci/blob/e381d3bb4eed97da83de4b4aa2370a8700009ea4/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/detail.js#L7),`` [troubleshooting.js:7](``https://github.com/openwrt/luci/blob/e381d3bb4eed97da83de4b4aa2370a8700009ea4/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/status/troubleshooting.js#L7)),`` and the `''` default is safe downstream — `''.split(/\n/)` yields `['']`, and the `ips.length > 0` guard at [rule.js:112](``https://github.com/openwrt/luci/blob/e381d3bb4eed97da83de4b4aa2370a8700009ea4/applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/network/rule.js#L112)`` skips it.

But it converts the reported failure into a silent one rather than fixing it: on the system where you saw the denial, the IPset dropdown will now come up with only `-- Please choose --` and no discovered ipsets, with nothing in the UI saying why. The ACL side looks correct in-tree — [`luci-app-mwan3.json:59`](``https://github.com/openwrt/luci/blob/e381d3bb4eed97da83de4b4aa2370a8700009ea4/applications/luci-app-mwan3/root/usr/share/rpcd/acl.d/luci-app-mwan3.json#L59``) grants `"/usr/libexec/luci-mwan3 ipset dump": ["exec"]` under `read`, which is exactly the scope/object/function triple cgi-io's exec applet checks, and the `admin/network/mwan3` menu tree depends on that same `luci-app-mwan3` group — so the grant should be reachable from this page.

Did you ever establish what was actually being denied on your box (helper missing/symlinked at a different path, stale rpcd ACL cache, an older `luci-app-mwan3.json` installed than the one in the tree)? If the root cause is still unknown, this hardening is worth having on its own, but the PR should probably stop claiming to fix the ACL error — see the nit below.

---

openwrt-ai, line 11: nit: the PR metadata is stale after the force-push and no longer describes this diff. The title is still `luci-app-mwan3: fix acl error` and the body still says *"This is caused by the call to `fs.exec_direct` ... After adding `exec_direct` in acl it works"*, but no ACL file is touched any more — the only change is this `L.resolveDefault()` wrapper. The commit message itself ("add resolveDefault to rule.js" / "Hardening the call for cases if the command is unavailable") is accurate; it's the PR title and description that disagree with it.

Worth realigning the title with the commit subject and rewriting the description to say this hardens the call rather than fixes an ACL grant, so the merge doesn't carry a misleading rationale for anyone bisecting this later.

---

openwrt-ai (commented):
Re-reviewed after the force-push. The ACL change from the previous head is gone and replaced with the `L.resolveDefault()` hardening — the new commit message matches its diff, so no commit-check findings. CI is green on `e381d3b`. Two inline notes: the underlying denial still appears unexplained (the in-tree ACL grant looks correct), and the PR title/body are stale relative to what the diff now does.

---

applications/luci-app-mwan3/htdocs/luci-static/resources/view/mwan3/network/rule.js
ckorber: In my case `/usr/libexec/luci-mwan3 ipset dump` did not produce names of ipset. So then nothing was reported.

PR #8980, opened 2026-08-25
luci-app-irqbalance: show the interrupts as a table, not a textarea

VizzleTF:
The page printed `/proc/interrupts` into a read-only `ui.Textarea` and the service
state into a bare `<p>`. This renders the same data with stock LuCI widgets, so a
theme can lay it out and a phone can read it.

What changes:

- **Status** becomes form rows — service state, CPU count, hardware IRQ count,
  busiest interrupt — instead of a paragraph alone in a section.
- **The snapshot** becomes two tables: hardware IRQs, and the software and
  inter-processor counters. One column per CPU, a total, and the device the line
  belongs to. Cells carry `data-title`, which is what themes key their
  narrow-screen layout off.
- **Interrupts that never fired** collapse behind a checkbox: 6 rows instead of 18
  on a four-core router, with the count stated beside the toggle.
- **The numbers refresh in place** on the existing poll rather than being rebuilt,
  so the page does not move under the reader.
- **Exclude IRQs** offers each IRQ with the device that owns it, instead of listing
  valid values in the description to match by hand against the snapshot.

Settings are unchanged: same options, same defaults, same validation. No
stylesheet is added and no colour is hardcoded — running/not running uses the
`--success-color-high` / `--error-color-high` variables themes export, with a
literal fallback.

Measured on a live GL-MT6000 (OpenWrt 25.12.5): 6 of 18 hardware IRQs collapsed,
18 expanded, counters advancing each tick with the scroll offset unmoved.

Context: the page came up in #8903, where the spacing under the textarea was
reported and before/after screenshots of this rewrite were posted.

### DIFF
--- a/applications/luci-app-irqbalance/htdocs/luci-static/resources/view/irqbalance.js
+++ b/applications/luci-app-irqbalance/htdocs/luci-static/resources/view/irqbalance.js
@@ -2,117 +2,298 @@
 'require view';
 'require fs';
 'require form';
-'require ui';
+'require poll';
 'require rpc';
+'require ui';
 
 const callServiceList = rpc.declare({
-    object: 'service',
-    method: 'list',
-    params: ['name'],
-    expect: { '': {} },
+	object: 'service',
+	method: 'list',
+	params: [ 'name' ],
+	expect: { '': {} },
 });
 
 function getServiceStatus() {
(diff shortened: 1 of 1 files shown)


VizzleTF:
@hnyman here we go

hnyman:
Thanks.
I tested this on two of my routers, and looked ok, so merging in.

PR #8981, opened 2026-08-25
luci-theme-footstrap: fixes from the field and from review

VizzleTF:
Replaces #8978. No 23.05 compatibility: that release is EOL, the theme has
dropped support for it, and the local `ui.RangeSlider` copy is gone
([#8978 (comment)](https://github.com/openwrt/luci/pull/8978#issuecomment-5409598124)).

Six commits:

1. **Port status cards** fill the row they are given — `minmax(126px, 200px)`
   counted tracks from the max, so 11 ports drew three ragged rows. The card
   drops its up/down dot, cuts the name with an ellipsis plus a tooltip, keeps
   one layout whether the port is linked or not, and sizes its figures with a
   token of their own (`--fs-type-2xs`).
2. **The reader stays put across a poll.** WebKit comes back 60px past the
   start after `dom.content()` swaps a section; the theme measured the wrong
   thing and picked the wrong reference on three page shapes. Held by 216 runs
   over three engines × two releases × two layouts × two widths × three
   densities.
3. **Spacing and hiding on luci-base classes:** a view-built block gets a
   section's gap (`luci-app-irqbalance` returns a bare widget), a meter's value
   stays off its label in a card, the title button row and per-section Delete
   get air, a value row draws no separator when none follows — and `hidden`
   hides, which a class's `display` was overriding, retiring four restatements
   of the attribute with it.
4. **Login page and graph labels:** the page renders without chrome, so it had
   an h2 and no h1; it now also names the router (#8961). Realtime axis labels
   were 1.16:1 against a light panel — recoloured through a child combinator,
   so Channel Analysis keeps the per-BSSID colour that ties each label to its
   curve.
5. **The settings tab comes back after a Save** — the redraw rebuilds the tab
   group and marks it ready with an attribute nothing was watching (#8903).
6. **Five exports nothing in this tree calls are gone.** They existed so gates in
   the theme's own repository could reach a module-private function; the
   functions stay, only the export line goes.
**Checked on** OpenWrt 25.12 (apk) and 24.10 (opkg), ImmortalWrt 25.12 and
24.10, and a hardware router. CSS changes are diffed against a baseline run of
the same sheet rather than eyeballed.


BKPepe (changes requested):
I appreciate that you split the changes up instead of putting everything into one large commit like in #8978, but I think this goes a bit too far in the other direction.

Twelve commits for what is essentially one theme update feels overly granular. I'd prefer to see related changes squashed into a few logical commits, rather than having a separate commit for every small fix.

I'd say 3–5 commits would be a much better balance: enough separation to keep the history meaningful, but without making it unnecessarily fragmented.

hnyman:
One comment is about the verbose comments in footstrap in general.

LuCI has js, lua & css minifying as default (defined in luci.mk, and small dedicated apps for the actions), so comments get stripped from the final binaries. 
But I do not think that we actually have ucode minifying. Does @jow- know?

So, your verbose comments get removed from your .js files , but I think that the ucode .ut comments ship along the compiled package, increasing the compiled size quite much. In some files, the comments are a clear majority of the content. They are good in documenting the reasoning for the code, but do increase the size.

It is not that many kB, but probably a reason why this theme's compiled version is much larger than the other themes. Actually this theme's .apk file is as large as the four other themes together...

> [luci-theme-bootstrap-26.235.67112~5cb5db6.apk](https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/luci/luci-theme-bootstrap-26.235.67112~5cb5db6.apk)	19.2 KB	Mon Aug 24 21:16:18 2026
> [luci-theme-footstrap-26.235.67112~5cb5db6.apk](https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/luci/luci-theme-footstrap-26.235.67112~5cb5db6.apk)	93.4 KB	Mon Aug 24 21:21:18 2026
> [luci-theme-material-26.235.67112~5cb5db6.apk](https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/luci/luci-theme-material-26.235.67112~5cb5db6.apk)	23.7 KB	Mon Aug 24 21:21:17 2026
> [luci-theme-openwrt-2020-26.235.67112~5cb5db6.apk](https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/luci/luci-theme-openwrt-2020-26.235.67112~5cb5db6.apk)	44.9 KB	Mon Aug 24 21:21:21 2026
> [luci-theme-openwrt-26.235.67112~5cb5db6.apk](https://downloads.openwrt.org/snapshots/packages/aarch64_cortex-a53/luci/luci-theme-openwrt-26.235.67112~5cb5db6.apk)	10.8 KB	Mon Aug 24 21:21:17 2026

In the light of the generic size inflation during the years, not that much, but still something to think in future. What level of commenting, or how can we clean them off?

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -295,6 +296,7 @@ to{opacity:1;transform:none}
 [data-darkmode="true"] [data-page="admin-statistics-graphs"] [data-plugin] img{filter:invert(100%) hue-rotate(180deg)}
 #view div[style] > svg{background-color:var(--fs-panel)}
 #view div[style] > svg line[style]{stroke:var(--fs-text) !important}
+#view div[style] > svg text[style]{fill:var(--fs-text) !important;text-shadow:none !important;font-weight:var(--fs-weight-bold)}
```
openwrt-ai: `svg text[style]` is a descendant selector, so it also repaints Status → Channel Analysis, where the inline `fill` is the one place in the realtime graphs that carries meaning: [`text.setAttribute('style', 'fill:'+color+…)` at channel_analysis.js:70](``https://github.com/openwrt/luci/blob/1bf17fab3be195c8e96ca9c0e6ac8f85cef034fd/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js#L70)`` derives a per-BSSID colour and paints both the curve and its SSID label with it. The curve keeps its colour (it is a `<polyline>`, which the sibling `line[style]` rule does not match), so after this rule every label is `--fs-text` and nothing ties a label back to its curve.

The grey `#eee` axis labels this rule is actually for are all direct children of the `<svg>` root — [`const G = svg.firstElementChild` at bandwidth.js:36](``https://github.com/openwrt/luci/blob/1bf17fab3be195c8e96ca9c0e6ac8f85cef034fd/modules/luci-mod-status/htdocs/luci-static/resources/view/status/bandwidth.js#L36)`` is the `<svg>` (`svg` there is the styled `div` returned by `loadSVG`), every `G.appendChild(text)` lands there, and the static `<text id="label_75">` in `svg/bandwidth.svg` is a direct child too. The coloured labels sit one level deeper, inside the `<g>` built at [channel_analysis.js:63-77](``https://github.com/openwrt/luci/blob/1bf17fab3be195c8e96ca9c0e6ac8f85cef034fd/modules/luci-mod-status/htdocs/luci-static/resources/view/status/channel_analysis.js#L63-L77).`` A child combinator separates the two:

```suggestion
#view div[style] > svg > text[style]{fill:var(--fs-text) !important;text-shadow:none !important;font-weight:var(--fs-weight-bold)}
```

---

themes/luci-theme-footstrap/ucode/template/themes/footstrap/sysauth.ut
```diff
@@ -24,12 +24,45 @@
...
 <form method="post" class="fs-login">
+	{#
+	 The router's name, above everything else in the card — an admin with three of these open has no
+	 other way to tell the tabs apart once the form is focused. `?? 'OpenWrt'` is the fallback
+	 partials/head.ut and partials/brand.ut already use, so the line never renders empty; the
+	 striptags/entityencode pair is theirs too, and load-bearing: the hostname is admin-controlled
+	 text printed to an UNAUTHENTICATED page, and one containing `&` broke entity parsing here
+	 before it broke anything else (see the <title> note in partials/head.ut).
+
+	 A <p> and NOT a heading: the card's h2 is `Authorization Required` and stays the page's one
```
openwrt-ai: nit: this comment says the card's heading is an `h2`, but the same commit promotes it to `<h1 name="content">` a few lines below (line 115) — and the whole point of the `<p>` is that it must not outrank that heading, so the stale level makes the rationale read backwards.

```suggestion
	 A <p> and NOT a heading: the card's h1 is `Authorization Required` and stays the page's one
```

---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-overview.js
```diff
@@ -53,6 +53,34 @@ function sectionTitle(sec) {
...
+ *
+ * styles/pages/20-overview.css cuts a port card's name at one line with an ellipsis (`white-space:
+ * nowrap` comes from base/95-luci.css; what the page rule adds is `min-width: 0` and the pair that
+ * makes the overflow readable) — that is what lets every card take the width the row can spare
+ * instead of the width of the longest interface name on the device. The name is the one thing on
+ * the card that cannot be guessed from the rest of it, so the element carries its own full text as
+ * a native tooltip.
+ *
+ * NOTHING HERE READS LAYOUT. Asking "was this one actually truncated?" means `scrollWidth` against
+ * `clientWidth` per card, i.e. a forced synchronous layout inside the path a poll tick lands on
```
openwrt-ai, line 66: nit: introduce-then-fix inside the series. 4df041c added this block describing a *two-line* clamp and a `scrollHeight`/`clientHeight` test; 42de08f — the very next commit, whose subject is "keep the reader's place across a poll" — rewrites it to one line and `scrollWidth`/`clientWidth`. The CSS in 4df041c never clamped to two lines (`overflow:hidden;text-overflow:ellipsis` on `.ifacebox-head:first-child`, with `white-space:nowrap` inherited from the `.ifacebox` base rule), so the comment was wrong the moment it was written. Fold the correction back into 4df041c rather than shipping the wrong version and fixing it one commit later.

---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -972,20 +980,20 @@ body[data-page="admin-status-overview"] #view > h2[name="content"],body[data-pag
...
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(2) br{display:none}
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]):has(img[src*="_down.svg"]) > .ifacebox-body:nth-child(2){color:var(--fs-dim)}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4){order:2;text-align:end;flex:1 0 auto}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4) > .cbi-tooltip-container{font-family:var(--fs-font-mono);font-size:var(--fs-type-xs) !important;color:var(--fs-dim);line-height:var(--fs-leading);white-space:nowrap;text-align:end !important}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4){order:2;text-align:end;flex:1 0 100%}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4) > .cbi-tooltip-container{font-family:var(--fs-font-mono);font-size:calc(10px * var(--fs-density-type)) !important;color:var(--fs-dim);line-height:var(--fs-leading-tight);white-space:nowrap;text-align:end !important}
```
openwrt-ai: nit: `calc(10px * var(--fs-density-type))` is the only hardcoded `font-size` in the whole sheet — every other size in `cascade.css` resolves through the `--fs-type-*` ladder, and this rule previously used `var(--fs-type-xs)`. The commit message argues the size is "declared plainly in the rule … the way 02-tokens.css states the type ladder", but 10px is *outside* that ladder (`--fs-type-xs` is 11px), and it is now restated twice, here and in the `@supports` override below. If a sub-xs step is genuinely needed for the port figures, a token (`--fs-type-2xs`) would keep the density axis in one place; otherwise the 1px saving over `--fs-type-xs` is worth a line in the commit message explaining why the ladder could not hold it.

---

openwrt-ai (commented):
## Commit checks

- `42de08f` "luci-theme-footstrap: keep the reader's place across a poll" — the message is entirely about scroll anchoring, but the commit also carries three `cascade.css` hunks that belong to the *next* commit and are described by its message: the `.table.fs-stacked:has(.tr.table-titles) .td .cbi-progressbar` pair (e697a41's "a meter's value sat on its own label once its column became a card") and `.cbi-title-buttons` / `.cbi-section-remove` (e697a41's "a page-title button row touched the heading above it, and the per-section Delete button touched the tab bar below it"). It additionally rewrites a `fs-overview.js` comment introduced one commit earlier (see inline).
- `e697a41` "luci-theme-footstrap: fix spacing and hiding on luci-base classes" — the message lists four spacing fixes, but the diff is 3 insertions / 2 deletions and contains only two of them (the `.cbi-map > *` gap and the `.cbi-value` hairline) plus the `[hidden]` rule. The meter-in-a-card and title-button/Delete items it describes are in `42de08f`.
- `38c414e` "luci-theme-footstrap: fix the login page heading and the graph labels" — adds `.cbi-section .table + h3{margin-top:var(--fs-space-5)}` and `.cbi-section h3 + .table{margin-top:var(--fs-space-3)}`. Those are neither the login heading nor the graph labels, and no commit message in the series mentions them.

Since the split is already under discussion in #8981 (review by @BKPepe), it is worth landing the spacing hunks in the spacing commit while re-rolling — right now three of the five subjects do not describe their own diffs.


---

VizzleTF:
> how can we clean them off?

I clean everything + do minification in my repo. I can strip all comments in PRs. I do it in [releases ](https://github.com/VizzleTF/luci-theme-footstrap/releases/tag/v0.14.2) so theme weights around 70kb.

> I'd say 3–5 commits would be a much better balance

@BKPepe  done

hnyman:
> I clean everything + do minification in my repo. 

Do you have a ucode minification routine that could be imported here?
The comments are useful to have in the source code here, so stripping them all away directly at sources might be harmful. But being able to strip them at compilation would be useful.

Examples:
* https://github.com/openwrt/luci/blob/master/luci.mk#L247
* https://github.com/openwrt/luci/tree/master/contrib/package/csstidy
* https://github.com/openwrt/luci/tree/master/modules/luci-base/src/jsmin.c
* https://github.com/openwrt/packages/tree/master/lang/lua/luasrcdiet/Makefile#L3

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -40,6 +40,7 @@
 abbr[title],acronym[title]{border-bottom:1px dotted;font-weight:inherit;cursor:help}
 table{border-collapse:collapse;border-spacing:0}
 ol,ul{list-style:none}
+[hidden]:not([hidden="until-found"]){display:none !important}
```
openwrt-ai, line 43: This general rule retires more than the one point fix the commit message names. `!important` beats a non-important declaration regardless of cascade layer, so three further `[hidden]` restatements in this same sheet are now dead:

- [`.fs-search-note[hidden]{display:none}` at cascade.css:905](``https://github.com/openwrt/luci/blob/a241def23892d4bcbf460c50a36d07a818d8b1ce/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L905``)
- [`.fs-ap-body[hidden]{display:none}` at cascade.css:1042](``https://github.com/openwrt/luci/blob/a241def23892d4bcbf460c50a36d07a818d8b1ce/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L1042``)
- [`.cbi-value[hidden],.fs-ap-bgprev[hidden],.fs-ap-bgrow .btn[hidden]{display:none}` at cascade.css:1069](``https://github.com/openwrt/luci/blob/a241def23892d4bcbf460c50a36d07a818d8b1ce/themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css#L1069``)

The commit message says "the theme's own point fix for the same fault -- a restated `display:none` for the search overlay -- goes away with the general one", and only `.fs-search-ov[hidden]` was dropped. Dropping the other three too would keep the claim true and leave one place that decides what `hidden` means.

The same change also falsifies a rationale comment that is left in place: [fs-appearance.js:383-385](``https://github.com/openwrt/luci/blob/a241def23892d4bcbf460c50a36d07a818d8b1ce/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-appearance.js#L383-L385)`` still says the restatement exists "at a specificity that beats `.cbi-value`'s own display — the UA's bare `[hidden]` rule loses to it", which is no longer the situation once the theme ships its own `!important` rule.

---

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -972,20 +980,18 @@ body[data-page="admin-status-overview"] #view > h2[name="content"],body[data-pag
...
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(2) br{display:none}
 body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]):has(img[src*="_down.svg"]) > .ifacebox-body:nth-child(2){color:var(--fs-dim)}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4){order:2;text-align:end;flex:1 0 auto}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4) > .cbi-tooltip-container{font-family:var(--fs-font-mono);font-size:var(--fs-type-xs) !important;color:var(--fs-dim);line-height:var(--fs-leading);white-space:nowrap;text-align:end !important}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4){order:2;text-align:end;flex:1 0 100%}
+body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]) > .ifacebox-body:nth-child(4) > .cbi-tooltip-container{font-family:var(--fs-font-mono);font-size:var(--fs-type-2xs) !important;color:var(--fs-dim);line-height:var(--fs-leading-tight);white-space:nowrap;text-align:end !important}
```
openwrt-ai, line 993: nit: introduce-then-fix inside the re-rolled series. `0604294` writes this rule with `calc(10px * var(--fs-density-type))` stated twice (plainly plus an `@supports` override) and spends a paragraph of its message defending that shape — "Their size is declared plainly in the rule and rounded in an @supports block, the way 02-tokens.css states the type ladder". `a241def`, the last commit, then deletes both statements and the `@supports` block and replaces them with `--fs-type-2xs`.

Since the split is being re-rolled anyway (#8981 review by @BKPepe), landing `--fs-type-2xs` directly in `0604294` and dropping that paragraph would leave the series with no commit that ships a value it argues for and then removes. `a241def` would disappear entirely.

---

themes/luci-theme-footstrap/ucode/template/themes/footstrap/sysauth.ut
```diff
@@ -24,12 +24,45 @@
...
+	 other way to tell the tabs apart once the form is focused. `?? 'OpenWrt'` is the fallback
+	 partials/head.ut and partials/brand.ut already use, so the line never renders empty; the
+	 striptags/entityencode pair is theirs too, and load-bearing: the hostname is admin-controlled
+	 text printed to an UNAUTHENTICATED page, and one containing `&` broke entity parsing here
+	 before it broke anything else (see the <title> note in partials/head.ut).
+
+	 A <p> and NOT a heading: the card's h1 is `Authorization Required` and stays the page's one
+	 heading. An outline is about the page's SUBJECT, and a heading above that one would announce
+	 the login form as two sections — a screen reader walking the headings would hear the router's
+	 name as a section title with the form nested under it. It is printed above the h1 and centred
```
openwrt-ai, line 61: nit: this h2 -> h1 correction is carried by `a241def`, whose subject is "give the port card figures a size token" — the only other file that commit touches is `cascade.css`, and nothing in its message mentions `sysauth.ut`. The comment it corrects was introduced two commits earlier by `458c87ad`, which is also the commit that promotes the heading to `<h1>` on line 115.

Fold it into `458c87ad` so the wrong level is never committed, rather than shipping it and correcting it at the end of the series.

---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js
```diff
@@ -638,6 +855,17 @@ return baseclass.extend({
...
+	/* -> the offset this file last took a reference at, or null before it has taken one.
+	 *
+	 * FOR THE GATES, and it is not a convenience: every correction here is measured against a
+	 * reference captured while the page was still, so a probe that grows the page before that
+	 * reference exists measures the guard rather than the anchor. There is no way to ask that from
+	 * outside — "is it scrolling" answers a different question, and answers "no" both before the
+	 * motion sampler starts and after it finishes, which in WebKit are 1.5 seconds apart. Waiting a
+	 * flat interval instead made tools/scroll-anchor.mjs report a jump on every WebKit run and none
+	 * on the other two engines, with the theme identical on all three. */
+	restAt: () => _restAt,
```
openwrt-ai: This adds a public method to the shipped `fs-fit` baseclass that nothing in `openwrt/luci` calls — `git grep restAt` over the tree at this head matches only the private `_restAt` inside this file and this accessor. Its own doc comment names its consumer as `tools/scroll-anchor.mjs`, which is not part of this repository (`themes/luci-theme-footstrap/` here contains only `LICENSE`, `Makefile`, `htdocs`, `po`, `root`, `ucode`).

Since the theme is minified on package build, a test-only accessor with no in-tree caller is dead weight that also reads as API to anyone extending the module. Either drop it from the exported object and let the harness reach `_restAt` another way, or say in the comment that the consumer lives out of tree so the next reader does not go looking for `tools/`.

---

openwrt-ai (commented):
Re-reviewed after the re-roll from twelve commits to six. All four inline comments from the previous round are addressed in the new series: the graph-label rule now uses a child combinator so Channel Analysis keeps its per-BSSID colours (confirmed against `channel_analysis.js` — the axis labels go in through `createGraphText(G, …)` where `G` is the `<svg>` root, while the per-BSSID labels sit inside a `<g>`), the `sysauth.ut` heading-level comment is corrected, `--fs-type-2xs` replaces the hardcoded size, and the `fs-overview.js` clamp comment now matches the CSS. The net content delta since `1bf17fa` is small and looks right.

Two of the three commit/diff mismatches from the previous round are fixed by the re-roll — the meter-in-a-card and title-button/Delete hunks now sit in the spacing commit that describes them. What is left:

## Commit checks

- `7260b51` "luci-theme-footstrap: fix spacing and hiding on luci-base classes" — still carries `.cbi-section .table + h3{margin-top:var(--fs-space-5)}` and `.cbi-section h3 + .table{margin-top:var(--fs-space-3)}`. These moved from `38c414e` into this commit, but no message in the series describes them: the four items this one lists are the `.cbi-map > *` gap, the meter in a card, the title-button row plus per-section Delete, and the `.cbi-value` hairline, and none of them is a heading adjacent to a table. Either add a fifth bullet or drop the two rules.
- `a241def` "luci-theme-footstrap: give the port card figures a size token" — the message is entirely about `--fs-type-2xs` and `cascade.css`, but the commit also rewrites two lines of `sysauth.ut` (see inline). That commit should not exist at all once the token lands in `0604294` and the comment fix lands in `458c87ad`.
- All six commits — the message bodies are hard-wrapped twice, leaving orphaned continuation lines: `0604294` has "150px of", "from", "shares", "rows"; `326beb3` has "reference", "broken", "the", "flick"; `458c87ad` has "and", "one, at", "wordmark and", "for", "legible"; and so on in the other three. This predates the re-roll rather than being introduced by it, but since the series is being rewritten for @BKPepe's squash request anyway, a re-wrap at a single width would be worth doing in the same pass.

---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-appearance.js
```diff
@@ -762,11 +763,35 @@ function tabGroup(view) {
  * Idempotent through the marker: an observer fires for every mutation, and the form's own
  * construction is a mutation. Without the check it would append itself for as long as it kept
  * noticing itself. */
+/* WHAT WAKES THIS UP WHEN NOTHING ELSE WILL.
```
BKPepe, line 766: Is this comment really needed? :-D Isn't there better way how to handle this?

BKPepe (changes requested):
Hey, you're really going to have to train that AI of yours. I have nothing against using AI. It helps me too, but your texts are just way too long, and nobody is going to read them. When I look at the commit descriptions, they're basically essays. :)

Personally, I don't think it's ideal to have references in the commit, because after every force-push it just spams them.

Looking at the ```Checks``` or ```Files``` changed, I see this:
```
'prefs' is not defined
```

There are also simply way too many comments in the code, as @hnyman correctly pointed out. Unfortunately, that's kind of an AI habit, but do we really need 6 commits for such a small amount of changes?

I'd also be quite interested to know whether a [security audit](https://support.claude.com/en/articles/11932705-automated-security-reviews-in-claude-code) was performed.

VizzleTF:
> your texts are just way too long

Agreed - there are a lot of comments and they are long. I squash them from time to time, but llm grows it back.

They also serve a second purpose: they keep the LLM from going in circles over questions that were already settled - the scroll anchoring for engines that do not restore the offset themselves (WebKit) is the obvious one - and from deleting workarounds that look wrong but are load-bearing. Most of them are exactly that: a note on why the code departs from what you would normally write.

I will trim the PR texts and keep commit messages short from now on if that bothers you.

> I don't think it's ideal to have references in the commit

Okay, i will not doing that

> do we really need 6 commits for such a small amount of changes?

No, thats why i make only 1 commit at first) 

> Is this comment really needed?

No.

>  Isn't there better way how to handle this?

Yeah, and its already in code. This part was for test, i forgot about it - will remove it.

> I'd also be quite interested to know whether a [security audit](https://support.claude.com/en/articles/11932705-automated-security-reviews-in-claude-code) was performed.

image
image


> Looking at the `Checks` or `Files` changed, I see this:
> 
> ```
> 'prefs' is not defined
> ```

That warning is the repo eslint config, not the theme: it does not read the 'require X as Y' pragma and instead hardcodes a globals list, so an alias only lints clean if someone added it there by hand. prefs was never added.

The fix belongs in eslint.config.mjs: derive the aliases from the pragma instead of listing them. I can send that as a separate PR.

themes/luci-theme-footstrap/htdocs/luci-static/footstrap/cascade.css
```diff
@@ -972,20 +979,18 @@ body[data-page="admin-status-overview"] #view > h2[name="content"],body[data-pag
 @container fs-view (max-width:800px){.fs-ovl{grid-template-columns:1fr}
 .fs-ovl > .fs-ovl-sys,.fs-ovl > .fs-ovl-mem,.fs-ovl > .fs-ovl-sto{grid-column:1;grid-row:auto}
 }
-body[data-page="admin-status-overview"] div:has(> .ifacebox img[src*="/port_"]){gap:var(--fs-space-2-5);grid-template-columns:repeat(auto-fit,minmax(126px,200px)) !important;margin-bottom:0 !important}
-body[data-page="admin-status-overview"] .ifacebox:has(img[src*="/port_"]){margin:0 !important;min-width:0 !important;max-width:none !important;width:auto !important;justify-self:stretch;align-self:stretch;text-align:start;background:var(--fs-panel2);border:var(--fs-hairline);border-radius:var(--fs-radius-lg);padding:var(--fs-space-2) var(--fs-space-3);display:flex;flex-direction:row;flex-wrap:wrap;column-gap:var(--fs-space-2);row-gap:var(--fs-space-1);align-items:baseline}
+body[data-page="admin-status-overview"] div:has(> .ifacebox img[src*="/port_"]){gap:var(--fs-space-2);--fs-port-min:calc(76px * var(--fs-density-type) + var(--fs-space-3) * 2);grid-template-columns:repeat(auto-fit,minmax(var(--fs-port-min),1fr)) !important;margin-bottom:0 !important}
```
openwrt-ai, line 982: nit: the three floors the commit message quotes — "94px at Normal, 80 at Compact and 109 at Large" — do not come out of this formula. `--fs-space-3` is `calc(12px * var(--fs-density-space))`, and the density pair is `1`/`1`, `.9`/`.65`, `1.15`/`1`, so `--fs-port-min` resolves to:

- Normal: `76 + 12*2` = **100px**
- Compact: `68.4 + 7.8*2` = **84px**
- Large: `87.4 + 12*2` = **111.4px**

The padding term scales on `--fs-density-space` while the content term scales on `--fs-density-type`, which is what pulls Compact apart from the quoted 80. Were 94/80/109 measured against an earlier padding token, or is 76px the wrong content figure? Either the numbers in the message or the formula here should move, since the message presents them as the measured result of this rule.

---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-router.js
```diff
@@ -1566,13 +1566,10 @@ return baseclass.extend({
 	/* exported for the unit suite in the theme's own repository (tests/router-contract.test.mjs —
 	 * it is not part of the package, which ships no tests), where it is driven against a
 	 * hand-broken `L`: the one way to see the OFF branch without a router that ships one */
-	contractBreaks,
 	/* likewise, and likewise out-of-package: interval-pause.test.mjs drives the navigation sweep
 	 * around a visibilitychange and session-expiry.test.mjs reads the verdict the interceptors
 	 * reached. navigate() is the real caller of the first and `_expired` gates the second — nothing
 	 * else may call either. */
```
openwrt-ai, line 1572: Both doc blocks are left behind by the exports they described. `prefetchSegs` now carries three comments in a row, and the first two document members that no longer exist and point at `tests/router-contract.test.mjs` / `interval-pause.test.mjs` / `session-expiry.test.mjs`, none of which are in this repository — so the next reader sees "exported for the unit suite" attached to `prefetchSegs`.

```suggestion
```

Same orphaning in the other two files this commit touches:

- [fs-chrome.js:579-582](``https://github.com/openwrt/luci/blob/48d0d05a2bbb9b2e9138f9bf779e6feff98bf9b3/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-chrome.js#L579-L582)`` — "exported for the unit suite … tests/chrome-geometry.test.mjs … the cut is pure arithmetic" now reads as documentation for `wireRail`; it described the removed `columnWidth`.
- [fs-fit.js:858-866](``https://github.com/openwrt/luci/blob/48d0d05a2bbb9b2e9138f9bf779e6feff98bf9b3/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-fit.js#L858-L866``) — the whole "-> the offset this file last took a reference at" block documented `restAt` and is left floating above `forgetRest`'s own comment.

---

themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-router.js
```diff
@@ -1566,13 +1566,10 @@ return baseclass.extend({
...
-	clearViewIntervals,
-	sessionExpired,
```
openwrt-ai, line 1575: Dropping this export leaves `sessionExpired()` with no caller anywhere in the tree. [Its definition at fs-router.js:347](``https://github.com/openwrt/luci/blob/48d0d05a2bbb9b2e9138f9bf779e6feff98bf9b3/themes/luci-theme-footstrap/htdocs/luci-static/resources/fs-router.js#L347``) is now the only mention of the name — every in-module consumer reads `_expired` directly (lines 325, 343, 959, 1546).

The commit message says "The functions stay where they were -- each is used inside its own module", which holds for the other four (`contractBreaks` is called at line 1407, `clearViewIntervals` at 1198, `columnWidth` at 279 and 571, `_restAt` throughout `fs-fit.js`) but not for this one. Either delete the function along with its export, or say in the message that it is kept deliberately for the out-of-tree suite — the theme is shipped minified and jsmin does no dead-code elimination, so it stays in the package as-is.

---

openwrt-ai (commented):
Re-reviewed after the second re-roll (`a241def` → `48d0d05`, seven commits to six). The only content change since the last round is eight deletions — the three surviving `[hidden]` restatements in `cascade.css` and the five module exports — both raised last round and both correct. The two introduce-then-fix pairs are gone: `--fs-type-2xs` now lands in the commit that introduces the rule, and the `sysauth.ut` comment fix in the commit that promotes the heading. The `calcunits` removal checks out against `ui.RangeSlider`, which only renders that span when a `calculate` function produced a value. All checks are green on this head.

## Commit checks

- `8918bd1` "pack the port status cards into their row" — the message still carries the paragraph arguing for the *previous* roll's shape: "Their size is declared plainly in the rule and rounded in an @supports block, the way 02-tokens.css states the type ladder: an engine without round() drops the whole declaration, flag and all". The diff no longer does that — it adds `--fs-type-2xs` to the token block — and the next paragraph, which also opens "Their size is …", says so. The first is left over from `0604294` and contradicts the commit it now sits on.
- `8ad08a1` "fix spacing and hiding on luci-base classes" — still carries `.cbi-section .table + h3` and `.cbi-section h3 + .table` (lines 480-481). The message lists four items plus the `[hidden]` rule, and none of them is a heading adjacent to a table. Carried over unchanged by the re-roll: add a fifth item or drop the two rules.

---

hnyman:
I will merge this in, although AI still has some points. You can fix them in the next polishing round.

